Issue #1944, opened 2020-03-29
Make pc.Application#loadScene public


aidinabedi:
### Description

Currently, there is no way to load a scene without fetching and parsing the same JSON scene file twice. To load load a scene properly (both settings and hierarchy), a developer must do the following:
```javascript
app.loadSceneSettings("1000.json", function (err1) {
    if (!err1) {
        app.loadSceneHierarchy("1000.json", function (err2) {
            if (!err2) {
              // success
            } else {
              // error
            }
        });
    } else {
      // error
    }
});
```

This is unnecessarily error-prone and inefficient because there is already a function that does exactly both simultaneously `pc.Application#loadScene`. However, it is currently marked as private and not available in the API references. Additionally, the private function is also used by the official "Web Download" script `__start__.js`, further emphasizing its need to be public.

PR #2009, opened 2020-04-23
Updated tests and docs for sceneRegistry loadScene* functions

yaustar:


yaustar:
Hold off on merging this, I just found out that the examples from https://developer.playcanvas.com/en/api/pc.SceneRegistryItem.html are also outdated so I will fix this first

yaustar:
Made loadScene private as it looked like it was made public accidentally during an API update to pc.Application. 

Updated some tests and documentation.

src/callbacks.js
```diff
@@ -51,25 +51,53 @@
...
+ * @private
+ * @deprecated
  * @callback pc.callbacks.LoadSettings
  * @description Callback used by {@link pc.Application#loadSceneSettings}.
  * @param {string|null} err - The error message in the case where the loading or parsing fails.
  */
 
 /**
+ * @private
+ * @deprecated
```
willeastcott, line 72: Why deprecate the old callbacks? Just update the description to say the updated function names of the new API.

src/framework/scene-registry.js
```diff
@@ -224,6 +224,7 @@ Object.assign(pc, function () {
     };
 
     /**
+     * @private
```
willeastcott, line 227: I have a feeling @aidinabedi make this function public on the `pc.Application` API because engine-only devs need it. Any reason to make it private here?

yaustar: Ahh, interesting about making it public for engine devs (I see the original PR https://github.com/playcanvas/engine/issues/1944 )

Will have to rethink this.

yaustar:
Made loadScene public again. Removed unused callbacks.

src/callbacks.js
```diff
@@ -52,24 +52,26 @@
...
 
 /**
  * @callback pc.callbacks.LoadSettings
- * @description Callback used by {@link pc.Application#loadSceneSettings}.
+ * @description Callback used by {@link pc.SceneRegistry#loadSceneSettings}.
  * @param {string|null} err - The error message in the case where the loading or parsing fails.
  */
 
 /**
+ * @private
```
willeastcott, line 67: Shouldn't this be public?

willeastcott:
Might wanna tweak the PR name now. 😄

yaustar:
Good call 😅

aidinabedi:
Hello @yaustar!

Happy to see the docs for `pc.SceneRegistry` being updated! If I may make a suggestion:

Please take a look at the `callback` parameter in each `pc.SceneRegistry#loadScene*` method. You will notice that they are all of type `{Function}`, which provides limited documentation and no actual callback argument type info. This is why the `pc.callbacks.Load*` definitions were added. They properly document each callback argument and more. Please consider replacing `{Function}` with the corresponding `pc.callbacks.Load*` type. You can compare with the same `callback` parameter in the now deprecated `pc.Application#loadScene*` methods.

Cheers,
Aidin

willeastcott:
Good point @aidinabedi! I've gone ahead and made that change. Thanks! 😄

PR #2019, opened 2020-04-26
Specular over alpha

### LINKED ISSUE #2018: Physically-correct highlights on transparent surfaces
Currently, PlayCanvas doesn't have a mode to account for specular (incl reflection) highlights when applying opacity. In photorealistic applications, this often results in transparent materials appearing flat and physically incorrect, unfortunately.

Below is a 50% transparent metallic glossy sphere in the Editor. Observe the gray color of the highlight:

transparent_material__0003_PlayCanvas

Exact same material in different game engines that account for specular highlights over transparency:

* **Unity:**
transparent_material__0001_Unity

* **Unreal:**
transparent_material__0000_Unreal

* **Babylon:**
transparent_material__0002_Babylon

aidinabedi:
Fixes #2018

PR changes:
- Add new property `pc.StandardMaterial#useSpecularOverAlpha` (default value: `false`)
- Define macro `SPECULAROVERALPHA` in frag shader when `useSpecularOverAlpha` enabled
- Calculate `specularOverAlpha` term differently based on specular mode (`combineDiffuse*.frag`)
- Increase `dAlpha` with `specularOverAlpha` squared (gives best visual result)
- Add shader utility function `getLuminance` (in `base.frag`)

----

All changes have been tested using a [public PlayCanvas project](https://playcanvas.com/project/681384/overview/specular-over-alpha-test) (simply run it with and without the PR):

### With PR:
image
(Press SPACE, double-click, or double-tap to toggle `useSpecularOverAlpha` on and off)

### Without PR:
image

### DIFF
--- a/src/graphics/program-lib/standard.js
+++ b/src/graphics/program-lib/standard.js
@@ -879,6 +879,10 @@ pc.programlib.standard = {
         var codeBegin = code;
         code = "";
 
+        if (options.useSpecularOverAlpha) {
+            code += '#define SPECULAROVERALPHA 1\n';
+        }
+
         if (options.clearCoat > 0) {
             code += '#define CLEARCOAT 1\n';
         }
--- a/src/graphics/program-lib/chunks/base.frag
+++ b/src/graphics/program-lib/chunks/base.frag
@@ -15,3 +15,8 @@ vec3 saturate(vec3 x) {
     return clamp(x, vec3(0.0), vec3(1.0));
 }
 
+float getLuminance(vec3 color) {
+    // https://en.wikipedia.org/wiki/Relative_luminance
+    return saturate(dot(color, vec3(0.2126, 0.7152, 0.0722)));
+}
+
--- a/src/graphics/program-lib/chunks/combineDiffuseSpecular.frag
+++ b/src/graphics/program-lib/chunks/combineDiffuseSpecular.frag
@@ -1,5 +1,9 @@
 vec3 combineColor() {
     vec3 combinedSpecular = dSpecularLight + dReflection.rgb * dReflection.a;
+
+    float specularOverAlpha = getLuminance(combinedSpecular * dSpecularity);
+    dAlpha = saturate(dAlpha + specularOverAlpha * specularOverAlpha);
+
     return mix(dAlbedo * dDiffuseLight, combinedSpecular, dSpecularity);
 }
 
--- a/src/graphics/program-lib/chunks/combineDiffuseSpecular.frag
+++ b/src/graphics/program-lib/chunks/combineDiffuseSpecular.frag
@@ -1,8 +1,10 @@
 vec3 combineColor() {
     vec3 combinedSpecular = dSpecularLight + dReflection.rgb * dReflection.a;
 
+#ifdef SPECULAROVERALPHA
     float specularOverAlpha = getLuminance(combinedSpecular * dSpecularity);
     dAlpha = saturate(dAlpha + specularOverAlpha * specularOverAlpha);
+#endif
 
     return mix(dAlbedo * dDiffuseLight, combinedSpecular, dSpecularity);
 }
--- a/src/graphics/program-lib/chunks/combineDiffuseSpecular.frag
+++ b/src/graphics/program-lib/chunks/combineDiffuseSpecular.frag
@@ -1,4 +1,5 @@
 vec3 combineColor() {
-    return mix(dAlbedo * dDiffuseLight, dSpecularLight + dReflection.rgb * dReflection.a, dSpecularity);
+    vec3 combinedSpecular = dSpecularLight + dReflection.rgb * dReflection.a;
+    return mix(dAlbedo * dDiffuseLight, combinedSpecular, dSpecularity);
 }
 
--- a/src/graphics/program-lib/chunks/combineDiffuseSpecularNoConserve.frag
+++ b/src/graphics/program-lib/chunks/combineDiffuseSpecularNoConserve.frag
@@ -1,5 +1,9 @@
 vec3 combineColor() {
     vec3 combinedSpecular = (dSpecularLight + dReflection.rgb * dReflection.a) * dSpecularity;
+
+    float specularOverAlpha = getLuminance(combinedSpecular);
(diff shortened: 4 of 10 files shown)


src/graphics/program-lib/standard.js
```diff
@@ -879,6 +879,10 @@ pc.programlib.standard = {
         var codeBegin = code;
         code = "";
 
+        if (options.useSpecularOverAlpha) {
+            code += '#define SPECULAROVERALPHA 1\n';
```
willeastcott, line 883: I don't think you need to define this as 1. I believe you can simply do: `#define SPECULAROVERALPHA`

Yeah, I see that `CLEARCOAT` is defined as 1 below, but I guess that's unnecessary too. :smile: 

cc @raytranuk

willeastcott:
This looks great! Aside from my _very_ minor comment about the usage of `#define`, this looks good to me.

Maksims:
Is it possible to implement this flag, with affecting either specular or alpha calculation chunks, instead of `combine` chunks?

Reasoning: less we affects chunks - the better. As advanced users do tend to extend rendering capabilities using chunks. And any modifications to them - can lead to breaking their shaders during development.
So less changes, or at least more isolated changes - the better.

src/graphics/program-lib/standard.js
raytranuk: yes - I will remove the 1 from CLEARCOAT too

mvaligursky:
I'm thinking that perhaps we could give material property some better name than "useSpecularOverAlpha".
Perhaps 'fadeSpecularByAlpha'. It would default to true, which is old behavior, where material (both diffuse & specular) fade using alpha value. When set to false, specular would not fade by alpha.

willeastcott:
Or to group it with other specular properties in the API reference, start it with `specular`. Something like: `specularIgnoreOpacity`.

raytranuk:
It is correct that the specular reflections should not be affected by opacity, so we could make it so the opacity only affect diffuse when a flag (`specularIgnoreOpacity` seems like a good flag name) is set - in terms of implementation: if using `BLEND_PREMULTIPLIED`, multiplying diffuse by Opacity before the combine would work - if using `BLEND_NORMAL`, multiplying specular reflections by one over Opacity would work (we would also need a workaround for divide by zero).

slimbuck:
Is there a reason you didn't use premultiplied alpha on the diffuse component instead of hacking around with luminance?

I suspect this approach to specular may cause issues for us in future.

slimbuck:
I think a better solution would to to add an option to the material to exclude specular contribution from the premul alpha calculation.

slimbuck:
This line https://github.com/playcanvas/engine/blob/master/src/graphics/program-lib/chunks/outputAlphaPremul.frag#L1 is premultiplying diffuse and specular contributions and shouldn't.

aidinabedi:
@Maksims 
> less we affects chunks - the better. As advanced users do tend to extend rendering capabilities using chunks. And any modifications to them - can lead to breaking their shaders during development.

Agreed. Will look for alternative.

---

@mvaligursky
> I'm thinking that perhaps we could give material property some better name than "useSpecularOverAlpha".

I choose the name based on the same concept in Babylon. The only engine I could find that actually had a name and switch for this.

---

@willeastcott
> Or to group it with other specular properties in the API reference, start it with `specular`. Something like: `specularIgnoreOpacity`.

Sure. Suggestion: `specularOverrideOpacity` ?

---

@raytranuk @slimbuck 
> if using BLEND_PREMULTIPLIED: multiplying only diffuse component by Opacity before the final combine would work

Good catch. Need to handle premultiplied.

---

Thank you everyone for the feedback! I appreciate it. Will try to address everything code-wise end of the week.

src/graphics/program-lib/standard.js
aidinabedi: No problem. Will fix.

raytranuk:
@aidinabedi - thanks again for this PR - it has been hotly discussed in the PlayCanvas dev team and we think that correctly displaying specular reflections on translucent materials will probably be what we want to be used generally for translucent materials in the future. From some discussion in the PlayCanvas dev team we have a few comments/requests:
1) We think the naming should be something like: `opacityFadesSpecularReflections` and the default value set to `true`.
2) We think the implementation would be more efficient if done before the combine shader chunks stage - this is important because, as mentioned above, we expect `opacityFadesSpecularReflections` to be set to `false` in materials in future projects.
3) We believe the opacity material property is also being used to fade in and out objects and not just as a material property - and so we would need a way to fade in/out specular reflections without popping - an good option we discussed to do this is to expose a new `fade` property that can be faded down instead of opacity if `opacityFadesSpecularReflections` is set to `false` in newer projects.
This could be quite a large change to your PR - so if you prefer we can also implement this in a separate PR?

willeastcott:
`opacityFadesSpecularReflections` is 31 characters. I wonder if I can create a custom ESLint rule that disallows property/variable names longer than 20 characters. 😄

aidinabedi:
Thank you @raytranuk for the comments. I agree, adding a new `fade` property makes sense. Overall, it sounds like the team has a very specific implementation in mind. I have no issue if anyone wants to take over. I'm not territorial :)

In fact, it would free up some of my time to make PRs for other implemented features that I've been hoping to contribute back at some point. I can keep the [actual issue](https://github.com/playcanvas/engine/issues/2018) open but close this PR, it's mainly an internal solution I quickly wrote for our own needs that I thought I'd share. Feel free to reuse the example project.

@willeastcott I hear you. At least it's no `combineDiffuseSpecularNoReflSeparateAmbientPS` ;)

raytranuk:
> `opacityFadesSpecularReflections` is 31 characters. I wonder if I can create a custom ESLint rule that disallows property/variable names longer than 20 characters. 😄

@willeastcott - seems `Refl` is used as a short form of `Reflections` in Engine code, so if we do that then `opacityFadesSpecularRefl` takes it down to 24 characters - if we shorten `Specular` to `Spec` then `opacityFadesSpecRefl`is exactly 20 characters?

@aidinabedi - great - we will get a PR together for `opacityFadesSpecRefl` soon and will keep you updated - and I look forward to seeing your next PR.

Issue #1045, opened 2017-11-28
Wireframe data wrong on complex model


scarletsky:
Here is a [demo project](https://playcanvas.com/editor/scene/569334), using 100k and 200k vertices models.
Go to launch page and then you can see the result.

Screenshot:

![image](https://user-images.githubusercontent.com/2386165/33307795-d7e34d50-d452-11e7-808c-01feed7392cf.png)

![image](https://user-images.githubusercontent.com/2386165/33307826-fd2d29fa-d452-11e7-9b7e-9f6a7c559976.png)

willeastcott:
This is because of this line:

https://github.com/playcanvas/engine/blob/master/src/scene/model.js#L259

We force 16-bit indices for wireframe meshes. And you're getting int overflow with your model because it has >65535 vertices. @guycalledfrank - we could just switch to 32-bit when necessary, but there is only about 95% support for 32-bit indices in WebGL 1. The correct approach would be to generate more than one mesh maybe?

guycalledfrank:
We already use 32-bit indices for models that don't fit into 16. So perhaps switching to 32 in generateWireframe would be consistent?

willeastcott:
Except 32-bit index support is not guaranteed for WebGL 1.0. I believe support is around 96% or something, although webglstats.com is down at the moment so I can't verify.

slimbuck:
This should be fixed in #2023

PR #2023, opened 2020-04-27
Glb morph targets

### LINKED ISSUE #179: Morph Target Support
We should add morph target support to the engine. We'll need POSITION0..7 semantics (preumably 8 is enough?), GLSL to do the interp. Obviously, we want the FBX conversion tool to read this data and write it to model JSON, but since the engine needs to work with procedurally created meshes, it's not a blocker.

slimbuck:
Fixes #179

The following changes are included in this PR:
- fixed morph target loading in the GLB viewer
- moved generateWireframe() function to the mesh
- updated the glb-viewer with new features and look


willeastcott:
Does this also fix https://github.com/playcanvas/engine/issues/1045?

slimbuck:
Most likely, but untested. Will do so now.

tools/glb-viewer/src/debug.js
```diff
@@ -0,0 +1,198 @@
...
+    mesh.vertexBuffer = new pc.VertexBuffer(device, vertexFormat, 1024, pc.BUFFER_DYNAMIC);
+    mesh.primitive[0].type = pc.PRIMITIVE_LINES;
+    mesh.primitive[0].base = 0;
+    mesh.primitive[0].indexed = false;
+    mesh.primitive[0].count = 0;
+
+    // construct the material
+    var material = new pc.BasicMaterial();
+    material.blend = false;
+    material.blendType = pc.BLEND_NORMAL;
```
willeastcott, line 48: On the previous line, you disable blending. Then, setting `blendType` to `pc.BLEND_NORMAL` will reenable blending. Did you want `pc.BLEND_NONE`? https://github.com/playcanvas/engine/blob/master/src/scene/materials/material.js#L178

src/scene/mesh.js
```diff
@@ -630,6 +629,65 @@ Object.assign(pc, function () {
...
+                    case pc.INDEXFORMAT_UINT32:
+                        return new Uint32Array(indexBuffer.storage);
+                    default:
+                        return null;
+                }
+            };
+
+            var lines = [];
+            var format;
+            if (this.indexBuffer[0]) {
```
mvaligursky, line 650: maybe should be if (this.indexBuffer && this.indexBuffer[0]) to handle case when indexBuffer in undefined

slimbuck:
Sorry I jumped the gun on this PR, will makes the suggested changes.

src/scene/mesh.js
slimbuck: this.indexBuffer is initialized to [null] at construction and (far as I know) should always be an array.

Adding the test could be safer, but it also implies to the reader that indexBuffer could validly be undefined or null. Is that the case?

slimbuck: According to the doc the array could be null, so perhaps the test should be:
```this.indexBuffer.length > 0 && indexBuffer[0]``` ?

Comment (mvaligursky) [reply src/scene/mesh.js:650]: it's public variable, so the user could set it to just null.
I don't remember specific test case, but I had this issue with my code where indexBuffer was null, so it's safer to test it.

Comment (slimbuck) [reply src/scene/mesh.js:650]: I would argue that's a user error. What if a user sets indexBuffer to a string or map or anything else? We can't test all variables all the time and I think the code should reflect what the system considers to be valid state.

Comment (slimbuck) [reply src/scene/mesh.js:650]: And the valid state is defined by the property doc:
```
     * @property {pc.IndexBuffer[]} indexBuffer An array of index buffers. For unindexed meshes, this array can
     * be empty. The first index buffer in the array is used by {@link pc.MeshInstance}s with a renderStyle
     * property set to pc.RENDERSTYLE_SOLID. The second index buffer in the array is used if renderStyle is
     * set to pc.RENDERSTYLE_WIREFRAME.
```

mvaligursky: fair enough.

slimbuck: I actually think adding a check for null would do a disservice, because then a reader of the code would think that null is a valid state.

slimbuck: I guess this is where typescript would help us; the setter function would have a signature disallowing any user from setting indexbuffer to null.

tools/glb-viewer/src/debug.js
slimbuck: this is being addressed in https://github.com/playcanvas/engine/pull/2024

src/scene/mesh.js
slimbuck: Additional array check is being added in https://github.com/playcanvas/engine/pull/2024

Issue #2021, opened 2020-04-27
Give more control and information to the developer when loading resources


yaustar:
Related issues: 
#737 Asset loader should retry failed assets
#494 assets.load and assets.loadFromUrl to return XHR object if one created internally

Currently the resource loader is not resilient to unstable networks (such as mobile) and doesn't give much control or information (such as progress) back to the client.

This makes it very difficult for the client to know if a load has stalled or going very slowly to make the decision to retry or notify the user if the network connection has dropped.

Taking `pc.SceneRegistry.loadSceneHierarchy` for example:
https://github.com/playcanvas/engine/blob/master/src/framework/scene-registry.js#L144
```
    SceneRegistry.prototype.loadSceneHierarchy = function (url, callback) {
        var self = this;

        // Because we need to load scripts before we instance the hierarchy (i.e. before we create script components)
        // Split loading into load and open
        var handler = this._app.loader.getHandler("hierarchy");

        // include asset prefix if present
        if (this._app.assets && this._app.assets.prefix && !pc.ABSOLUTE_URL.test(url)) {
            url = pc.path.join(this._app.assets.prefix, url);
        }

        handler.load(url, function (err, data) {
            if (err) {
                if (callback) callback(err);
                return;
            }

            // called after scripts are preloaded
            var _loaded = function () {
                self._app.systems.script.preloading = true;
                var entity = handler.open(url, data);
                self._app.systems.script.preloading = false;

                // clear from cache because this data is modified by entity operations (e.g. destroy)
                self._app.loader.clearCache(url, "hierarchy");

                // add to hierarchy
                self._app.root.addChild(entity);

                // initialize components
                pc.ComponentSystem.initialize(entity);
                pc.ComponentSystem.postInitialize(entity);

                if (callback) callback(err, entity);
            };

            // load priority and referenced scripts before opening scene
            self._app._preloadScripts(data, _loaded);
        });
    };
```

The request is made to the handler which makes the call to `pc.HierarchyHandler.load` 
https://github.com/playcanvas/engine/blob/master/src/resources/hierarchy.js#L10
```
        load: function (url, callback) {
            if (typeof url === 'string') {
                url = {
                    load: url,
                    original: url
                };
            }

            var assets = this._app.assets;

            pc.http.get(url.load, {
                retry: this.retryRequests
            }, function (err, response) {
                if (!err) {
                    pc.TemplateUtils.waitForTemplatesInScene(
                        response,
                        assets,
                        callback);
                } else {
[...]

PR #2025, opened 2020-04-27
Draft concept of handling resource loading

yaustar:
This is just for the scene register for the moment but gives an idea on one of the possible ways we can handle the API for this issue for both the scenes and the assets.

Reference #2021 


vkalpias:
Would it be adequate to just return the XHR object instead of creating a new handle class?

yaustar:
We can do but I don't think we return a 'raw' low level object as part of the engine API. For example, mouse events are wrapped in a PlayCanvas event object.

yaustar:
Won't do upon review.

PR #2031, opened 2020-04-28
re-refactor reflections to work with seemore custom reflection chunks…

raytranuk:
… and clear coat


src/graphics/program-lib/standard.js
```diff
@@ -880,7 +880,7 @@ pc.programlib.standard = {
         code = "";
 
         if (options.clearCoat > 0) {
-            code += '#define CLEARCOAT 1\n';
+            code += '#define CLEARCOAT\n';
```
willeastcott, line 883: I don't understand why you need a #define for clear coat. I mean, of `clearCoat` is 0, the generated shader has none of the GLSL anyway because of the `if`statements on the JS side.

PR #2036, opened 2020-04-29
Fix getting transforms of inputSource after XR session is ended

Maksims:
Fix error with getting transforms on input sources when XR session is ended.


src/xr/xr-input-source.js
```diff
@@ -224,6 +224,9 @@ Object.assign(pc, function () {
     XrInputSource.prototype._updateTransforms = function () {
         var dirty;
```
willeastcott, line 225: Any reason to keep `dirty`?

src/xr/xr-input-source.js
```diff
@@ -224,6 +224,9 @@ Object.assign(pc, function () {
...
 
+        if (! this._manager.active)
+            return;
+
         if (this._dirtyLocal) {
             dirty = true;
```
willeastcott, line 231: And this occurrence?

src/xr/xr-input-source.js
Maksims: Yes, it uses it slightly bellow to decide to do transform update or not.

src/xr/xr-input-source.js
Maksims: It happens when session is ended, but due to WebXR API async nature of session end event, it is possible to call `app.xr.end()` but then getting actual callback of XRSession end later. Although camera pointer is removed immediately on `end` call. So input sources are not removed immediately as they are removed by XRSession events.

We can start removing them by hand when `end` is manually executed.

willeastcott:
Hey @Maksims - is this ready to be merged?

PR #2049, opened 2020-05-04
Engine mini stats

slimbuck:
This PR adds a graphical display of the frame time, CPU load and GPU load (GPU only where supported, mostly non-safari desktop browsers).

Mini stats can toggle between 3 sizes:
- small, no graph 100x16
- medium, graph 128x32
- large, graph 256x64

mini-stats

Click on the mini stats to cycle through the supported sizes, mouse over it for fully opaque view.

The information displayed is as follows:
- GPU shows milliseconds spent rendering in PlayCanvas per frame
- CPU shows approximate milliseconds spent updating the game per frame
- Frame shows the frame time as passed in by the browser

Each graph plots the stacked timings per frame with the update part of a frame in red and the render portion of a frame in green. (The Frame graph is always red).

A new preprocessor define called RELEASE_BUILD was added to build.js. The mini stats is excluded from this build.


build/build.js
```diff
@@ -168,7 +168,8 @@ var preprocess = function (dependencies) {
         else 
             src = pp.process({
                 PROFILER: profiler || debug,
-                DEBUG: debug
+                DEBUG: debug,
+                RELEASE_BUILD: compilerLevel != COMPILER_LEVEL[0]
```
willeastcott, line 172: Any reason not to have this as `RELEASE` instead of `RELEASE_BUILD'?

willeastcott: And just to be clear, we have the following build configs:

* DEBUG: playcanvas.dbg.js
* PROFILER: playcanvas.prf.js
* RELEASE: playcanvas.js
* RELEASE (MINIFIED): playcanvas.min.js

So MiniStats is just in the first two?

slimbuck: I should probably have, but at the time RELEASE just seemed too general.

slimbuck: MiniStats is available in all but the last RELEASE(MINIFIED) since this is the only build which specifies compilerLevel.

willeastcott: I get your point, but the same would hold for `DEBUG`. I think we should ensure consistency of naming here. And I guess I'm fine with `playcanvas.js` containing MiniStats. We can always modify this decision later if required.

willeastcott: Just thinking about it....so if `playcanvas.js` isn't `DEBUG`, `PROFILER` or `RELEASE`....what is it? 😄

willeastcott: I mean, `playcanvas.js` and `playcanvas.min.js` are functionally identical (and so they should be). You should have near 100% confidence that your game will behave the same on both. I'm a little bit worried about diverging these two library flavors.

slimbuck: Renamed RELEASE_BUILD -> RELEASE.

mvaligursky:
just curious - what's the CPU overhead of it?
In theory we could just pass texture of 2 x width instead of 50 x width (not real dimensions, just example) and change shader to expand the quads instead of filling the texture on CPU). Not sure what's better overall.

slimbuck:
@mvaligursky the CPU is doing two things: generating the quad vertices per frame (about 20 quads in total) and updating a single column of pixels in the graph texture.

Can you give an example of 2x width? Not sure what you mean.

slimbuck:
Most importantly, when the mini-stats is in small mode then the graphs are not being updated. So in that case the overhead excludes updating textures.

slimbuck:
@mvaligursky Oh! I think I understand what you're saying. Yes that could be done... nice idea. I would prefer trying this as a round two.

mvaligursky:
Ah if it updates only one column a frame, then no big deal, I thought it updates 2d area.
And so I was proposing to only have texture where for each frame there's two values - CPU time and GPU time, and fixed VB with quads (two per frame) .. and vertex shader would stretch those quads depending on those timing values.

slimbuck:
I'm attempting something slightly different: 1d texture with update timing in R and render timing in G (and whatever else in B). Then single pixel update per-graph per-frame. Nearly there!

slimbuck:
> it's -> its

face palm

Maksims:
And how to call, enable, create those stats? Does it have new API to be called?

slimbuck:
Hi @Maksims, It's currently enabled as an option on the application constructor, however we will be removing the mini-stats from the playcanvas engine package and so API might change.

Maksims:
There is no way to enable it by command from already running application with !release engine?

src/framework/application.js
```diff
@@ -293,6 +294,9 @@ Object.assign(pc, function () {
         this._soundManager = new pc.SoundManager(options);
         this.loader = new pc.ResourceLoader(this);
 
+        // construct ministats if it's present
+        this.miniStats = pc.MiniStats ? new pc.MiniStats(this) : null;
```
slimbuck, line 298: not really sure whether this is best approach

willeastcott: I guess the problem here is that it adds a hard dependency between the engine and ministats.

tools/glb-viewer/src/viewer.js
```diff
@@ -7,7 +7,8 @@ var Viewer = function (canvas) {
     // create the application
     var app = new pc.Application(canvas, {
         mouse: new pc.Mouse(canvas),
-        touch: new pc.TouchDevice(canvas)
+        touch: new pc.TouchDevice(canvas),
+        miniStats: true
```
willeastcott, line 11: Why is this `miniStats` property needed?

slimbuck: It's no longer needed, removed. Also renamed glb-viewer -> viewer.

src/framework/application.js
slimbuck: removed

willeastcott:
One final request - are you gonna add this to all the engine examples too? 😄

slimbuck:
You drive a hard bargain Eastcott!

PR #2050, opened 2020-05-05
RenderLine fix when no options specified

### LINKED ISSUE #2044: Error when trying to render a line with 2 colors
Works fine when drawing a single line color. Error happens when trying to use 2 colors:

```
launch.js:7876 TypeError: Cannot read property 'layer' of undefined
    at Application._addLines (playcanvas-stable.dbg.js:48205)
    at Application.renderLine (playcanvas-stable.dbg.js:48273)
```

The example, [shown in the docs](https://developer.playcanvas.com/en/api/pc.Application.html#renderLine), will generate the error.
```
var start = new pc.Vec3(0, 0, 0);
var end = new pc.Vec3(1, 0, 0);
var startColor = new pc.Color(1, 1, 1);
var endColor = new pc.Color(1, 0, 0);
app.renderLine(start, end, startColor, endColor);
```

mvaligursky:
fixes #2044

### DIFF
--- a/src/scene/immediate.js
+++ b/src/scene/immediate.js
@@ -135,7 +135,7 @@ Object.assign(pc.Application.prototype, function () {
 
     function _addLines(position, color, options) {
         var layer = (options && options.layer) ? options.layer : this.scene.layers.getLayerById(pc.LAYERID_IMMEDIATE);
-        var depthTest = (options && options.depthTest) ? options.depthTest : true;
+        var depthTest = (options && options.depthTest !== undefined) ? options.depthTest : true;
         var mask = (options && options.mask) ? options.mask : undefined;
 
         this._initImmediate();
--- a/src/scene/immediate.js
+++ b/src/scene/immediate.js
@@ -266,8 +266,10 @@ Object.assign(pc.Application.prototype, function () {
         } else if (arg3) {
             // options passed in
             options = arg3;
-        } else {
-            // no arg3, use default options
+        }
+
+        // use default options
+        if (!options) {
             options = {
                 layer: this.scene.layers.getLayerById(pc.LAYERID_IMMEDIATE),
                 depthTest: true
--- a/src/scene/immediate.js
+++ b/src/scene/immediate.js
@@ -134,12 +134,12 @@ Object.assign(pc.Application.prototype, function () {
     }
 
     function _addLines(position, color, options) {
-        if (options.layer === undefined) options.layer = this.scene.layers.getLayerById(pc.LAYERID_IMMEDIATE);
-        if (options.depthTest === undefined) options.depthTest = true;
+        var layer = (options && options.layer) ? options.layer : this.scene.layers.getLayerById(pc.LAYERID_IMMEDIATE);
+        var depthTest = (options && options.depthTest) ? options.depthTest : true;
+        var mask = (options && options.mask) ? options.mask : undefined;
 
         this._initImmediate();
 
-        var layer = options.layer;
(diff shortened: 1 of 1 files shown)


src/scene/immediate.js
```diff
@@ -266,8 +266,10 @@ Object.assign(pc.Application.prototype, function () {
         } else if (arg3) {
             // options passed in
             options = arg3;
-        } else {
-            // no arg3, use default options
+        }
+
+        // use default options
+        if (!options) {
```
willeastcott, line 272: Would it be better to fix `_addLines` to handle an undefined options object? That way, you would avoid allocating a JS object on every call to this function.

mvaligursky: done, good idea.

src/scene/immediate.js
```diff
@@ -134,29 +134,29 @@ Object.assign(pc.Application.prototype, function () {
     }
 
     function _addLines(position, color, options) {
-        if (options.layer === undefined) options.layer = this.scene.layers.getLayerById(pc.LAYERID_IMMEDIATE);
-        if (options.depthTest === undefined) options.depthTest = true;
+        var layer = (options && options.layer) ? options.layer : this.scene.layers.getLayerById(pc.LAYERID_IMMEDIATE);
+        var depthTest = (options && options.depthTest) ? options.depthTest : true;
```
willeastcott, line 138: What if `options.depthTest` is set to `false`?

src/scene/immediate.js
```diff
@@ -134,29 +134,29 @@ Object.assign(pc.Application.prototype, function () {
     }
 
     function _addLines(position, color, options) {
-        if (options.layer === undefined) options.layer = this.scene.layers.getLayerById(pc.LAYERID_IMMEDIATE);
-        if (options.depthTest === undefined) options.depthTest = true;
+        var layer = (options && options.layer) ? options.layer : this.scene.layers.getLayerById(pc.LAYERID_IMMEDIATE);
```
willeastcott, line 137: What if I pass `pc.LAYERID_WORLD` which is set to 0?  
https://github.com/playcanvas/engine/blob/master/src/scene/scene.js#L150

src/scene/immediate.js
```diff
@@ -134,29 +134,29 @@ Object.assign(pc.Application.prototype, function () {
...
+        var depthTest = (options && options.depthTest !== undefined) ? options.depthTest : true;
+        var mask = (options && options.mask) ? options.mask : undefined;
```
willeastcott, line 139: This one's a bit trickier. We don't document `pc.MeshInstance#mask` very well. There are some enums here:
https://github.com/playcanvas/engine/blob/master/src/scene/scene.js#L576
This line will convert 0 to undefined. Either way, the `if` statement below will fail. So I think this is fine.

src/scene/immediate.js
mvaligursky: that should be ok .. you pass in Layer object, not id. That's how this is called:

     * var worldLayer = app.scene.layers.getLayerById(pc.LAYERID_WORLD);
     * app.renderLine(start, end, color, {
     *     layer: worldLayer
     * });

src/scene/immediate.js
mvaligursky: yep that's what I think as well, this should be fine.

src/scene/immediate.js
willeastcott: Oh yes, you're right. 👍

PR #2073, opened 2020-05-12
Extended the Outline post effect

leonidaspir:
Added **thickness** as an attribute and **color alpha** support to the outline post process effect.


willeastcott:
Thanks @leonidaspir! All yours @raytranuk. 😄

raytranuk:
@leonidaspir - thanks for this enhancement - looks fine to me - something to note is that cost will increase exponentially with increase in thickness - and at a certain threshold thickness (depending on the GPU) it would be more performant to do this in 2 passes (if still done as a post process).

PR #1997, opened 2020-04-15
Physics engine optimizations

willeastcott:
This PR optimizes the rigidbody and collision components. For some physics-heavy applications, the speed up is very significant (up to ~2x is possible).

Specifically, the PR:

* Moves all physics update into the `pc.RigidBodyComponentSystem`. The collision component system now does zero update. This means that the order in which the physics world is updated can be strictly controlled.
* Avoids all the per-frame checks for whether each body needs to be processed.
* Moves the updating of kinematic body transforms to *before* the stepping of the physics simulation.
* Optimizes the `_updateKinematic` function to be much faster. It no longer takes velocities into account and just tracks the world transform of the entity.
* Moves the long-deprecated `pc.RigidBodyComponent#bodyType` to backwards-compatibility.js.

Benchmark (Win10 laptop + Chrome):

The ['Creating Rigid Bodies in Code' scene](https://launch.playcanvas.com/480301) has gone from ~16.5ms to ~9.5ms per frame on the CPU.

Fixes #1992 

I confirm I have signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #2074, opened 2020-05-13
Changed collision checks to properly check on each each internal physics world time step.

### LINKED ISSUE #2072: [Physics] Collision Checking Incorrect
Related forum posts:
[Events not firing properly when increasing fixedtimestep](https://forum.playcanvas.com/t/collision-events-not-firing-properly-when-increasing-fixedtimestep/6241)
[Ammo js not returning collision callbacks](https://forum.playcanvas.com/t/solved-ammo-js-not-returning-collision-callbacks/7064)

The update function in the Rigidbody system file is checking for collisions after stepSimulation is called, but collisions can occur and then stop occurring during a single stepSimulation called if either 

a. The frame rate is low 
b. You decrease the fixedTimeStep to increase simulation accuracy (for example to 1 / 120) 

The easiest way to reproduce this is to increase the fixedTimeStep to something small like 1 / 240 and log collision events, it's clearly noticeable that many collision events are not being fired. 

This is because the stepSimulation will take how many ever substeps in increments of the fixedTimeStep it needs to in order to fill the deltaTime you pass it. So if you are using the dispatcher and the manifolds to check for collisions, if a collision occurs during one of these substeps but then is not occurring at the last substep then Playcanvas will fail to fire a collision event. 

Bullet has a built in method for handling this that is exposed in Ammo as dynamicsWorld.setInternalTickCallback. This will allow you to register a callback (the separated checkCollision function) that fires on each actual simulation step. This is what this looks like. I've tested all of the following code and it fixes the issue. If you'd like a pull request for this stuff let me know. 

```
var callback = Ammo.addFunction(this.checkForCollisions.bind(this), 'vif');
this.dynamicsWorld.setInternalTickCallback(callback);
```

The code for collision checking would then be removed from the update function as follows

```
// onUpdate(dt) function in the rigid-body/system.js

// #ifdef PROFILER
this._stats.physicsStart = pc.now();
// #endif

// Check to see whether we need to update gravity on the dynamics world
var gravity = this.dynamicsWorld.getGravity();
if (
  gravity.x() !== this.gravity.x ||
  gravity.y() !== this.gravity.y ||
  gravity.z() !== this.gravity.z
) {
  gravity.setValue(this.gravity.x, this.gravity.y, this.gravity.z);
  this.dynamicsWorld.setGravity(gravity);
}

// Update the transforms of all bodies
this.dynamicsWorld.stepSimulation(dt, this.maxSubSteps, this.fixedTimeStep);

// Update the transforms of all entities referencing a body
var components = this.store;
for (var id in components) {
  if (components.hasOwnProperty(id)) {
    var entity = components[id].entity;
    var componentData = components[id].data;
    if (
      componentData.body &&
      componentData.body.isActive() &&
      componentData.enabled &&
      entity.enabled
    ) {
      if (componentData.type === pc.BODYTYPE_DYNAMIC) {
        entity.rigidbody.syncBodyToEntity();
      } else if (componentData.type === pc.BODYTYPE_KINEMATIC) {
        entity.rigidbody._updateKinematic(dt);
      }
    }
  }
}

// #ifdef PROFILER
this._stats.physicsTime = pc.now() - this._stats.physicsStart;
// #endif

```

New check for collisions function 

```
      checkForCollisions: function (dw, timestep) {
        // Check for collisions and fire callbacks
        var dispatcher = this.dynamicsWorld.getDispatcher();
        var numManifolds = dispatcher.getNumManifolds();
        var i, j;

        frameCollisions = {};

        // loop through the all contacts and fire events
        for (i = 0; i < numManifolds; i++) {
          var manifold = dispatcher.getManifoldByIndexInternal(i);

          var body0 = manifold.getBody0();
          var body1 = manifold.getBody1();

          var wb0 = Ammo.castObject(body0, Ammo.btRigidBody);
          var wb1 = Ammo.castObject(body1, Ammo.btRigidBody);

          var e0 = wb0.entity;
          var e1 = wb1.entity;

          // check if entity is null - TODO: investigate when this happens
          if (!e0 || !e1) {
            continue;
          }

          var flags0 = wb0.getCollisionFlags();
          var flags1 = wb1.getCollisionFlags();

          var numContacts = manifold.getNumContacts();
          var forwardContacts = [];
          var reverseContacts = [];
          var newCollision, e0Events, e1Events, e0BodyEvents, e1BodyEvents;

          if (numContacts > 0) {
            // don't fire contact events for triggers
            if (
              flags0 & pc.BODYFLAG_NORESPONSE_OBJECT ||
              flags1 & pc.BODYFLAG_NORESPONSE_OBJECT
            ) {
              e0Events =
                e0.collision &&
                (e0.collision.hasEvent("triggerenter") ||
                  e0.collision.hasEvent("triggerleave"));
              e1Events =
                e1.collision &&
                (e1.collision.hasEvent("triggerenter") ||
                  e1.collision.hasEvent("triggerleave"));
              e0BodyEvents =
                e0.rigidbody &&
                (e0.rigidbody.hasEvent("triggerenter") ||
                  e0.rigidbody.hasEvent("triggerleave"));
              e1BodyEvents =
                e1.rigidbody &&
                (e1.rigidbody.hasEvent("triggerenter") ||
                  e1.rigidbody.hasEvent("triggerleave"));

              // fire triggerenter events for triggers
              if (e0Events) {
                newCollision = this._storeCollision(e0, e1);
                if (newCollision && !(flags1 & pc.BODYFLAG_NORESPONSE_OBJECT)) {
                  e0.collision.fire("triggerenter", e1);
                }
              }

              if (e1Events) {
                newCollision = this._storeCollision(e1, e0);
                if (newCollision && !(flags0 & pc.BODYFLAG_NORESPONSE_OBJECT)) {
                  e1.collision.fire("triggerenter", e0);
                }
              }

              // fire triggerenter events for rigidbodies
              if (e0BodyEvents) {
                if (!newCollision) {
                  newCollision = this._storeCollision(e1, e0);
                }

                if (newCollision) {
                  e0.rigidbody.fire("triggerenter", e1);
                }
              }

              if (e1BodyEvents) {
                if (!newCollision) {
                  newCollision = this._storeCollision(e0, e1);
                }

                if (newCollision) {
                  e1.rigidbody.fire("triggerenter", e0);
                }
              }
            } else {
              e0Events = this._hasContactEvent(e0);
              e1Events = this._hasContactEvent(e1);
              var globalEvents = this.hasEvent("contact");

              if (globalEvents || e0Events || e1Events) {
                for (j = 0; j < numContacts; j++) {
                  var btContactPoint = manifold.getContactPoint(j);

                  var contactPoint = this._createContactPointFromAmmo(
                    btContactPoint
                  );
                  var reverseContactPoint = null;
                  if (e0Events || e1Events) {
                    reverseContactPoint = this._createReverseContactPointFromAmmo(
                      btContactPoint
                    );
                    forwardContacts.push(contactPoint);
                    reverseContacts.push(reverseContactPoint);
                  }

                  if (globalEvents) {
                    // fire global contact event for every contact
                    var result = this._createSingleContactResult(
                      e0,
                      e1,
                      contactPoint
                    );
                    this.fire("contact", result);
                  }
                }

                if (e0Events) {
                  var forwardResult = this._createContactResult(
                    e1,
                    forwardContacts
                  );
                  newCollision = this._storeCollision(e0, e1);

                  if (e0.collision) {
                    e0.collision.fire("contact", forwardResult);
                    if (newCollision) {
                      e0.collision.fire("collisionstart", forwardResult);
                    }
                  }

                  if (e0.rigidbody) {
                    e0.rigidbody.fire("contact", forwardResult);
                    if (newCollision) {
                      e0.rigidbody.fire("collisionstart", forwardResult);
                    }
                  }
                }

                if (e1Events) {
                  var reverseResult = this._createContactResult(
                    e0,
                    reverseContacts
                  );
                  newCollision = this._storeCollision(e1, e0);

                  if (e1.collision) {
                    e1.collision.fire("contact", reverseResult);
                    if (newCollision) {
                      e1.collision.fire("collisionstart", reverseResult);
                    }
                  }

                  if (e1.rigidbody) {
                    e1.rigidbody.fire("contact", reverseResult);
                    if (newCollision) {
                      e1.rigidbody.fire("collisionstart", reverseResult);
                    }
                  }
                }
              }
            }
          }
        }
        this._cleanOldCollisions();

        // Reset contact pools
        this.contactPointPool.freeAll();
        this.contactResultPool.freeAll();
        this.singleContactResultPool.freeAll();
      },
```

maxbbb:
Fixes #2072 

Separates the check collision logic into its own function and lets Ammo handle calling it on each internal time step. Provides warning if user is using older version that does not support setInternalTickCallback and falls back to calling after each stepSimulation if unavailable. 

### DIFF
--- a/src/framework/components/rigid-body/system.js
+++ b/src/framework/components/rigid-body/system.js
@@ -185,7 +185,9 @@ Object.assign(pc, function () {
                     var checkForCollisionsPointer = Ammo.addFunction(this._checkForCollisions.bind(this), 'vif');
                     this.dynamicsWorld.setInternalTickCallback(checkForCollisionsPointer);
                 } else {
+                    // #ifdef DEBUG
                     console.warn("WARNING: This version of ammo.js can potentially fail to report contacts. Please update it to the latest version.");
+                    // #endif
                 }
 
                 // Lazily create temp vars
@@ -534,8 +536,8 @@ Object.assign(pc, function () {
          * @function
          * @name pc.RigidBodyComponentSystem#_checkForCollisions
          * @description Checks for collisions and fires collision events
-         * @param {pc.Entity} dynamicsWorld - The pointer to the dynamics world that invoked this callback.
-         * @param {pc.Entity} timeStep - The amount of simulation time processed in the last simulation tick.
+         * @param {object | number} dynamicsWorld - Either The integer pointer to the dynamics world that invoked this callback or the underlying world.
+         * @param {number} timeStep - The amount of simulation time processed in the last simulation tick.
          * @returns {void}
          */
         _checkForCollisions: function (dynamicsWorld, timeStep) {
@@ -725,7 +727,7 @@ Object.assign(pc, function () {
             }
 
             if (!this.dynamicsWorld.setInternalTickCallback)
-                this._checkForCollisions();
+                this._checkForCollisions(this.dynamicsWorld, dt);
 
             // #ifdef PROFILER
             this._stats.physicsTime = pc.now() - this._stats.physicsStart;
--- a/src/framework/components/rigid-body/system.js
+++ b/src/framework/components/rigid-body/system.js
@@ -181,6 +181,13 @@ Object.assign(pc, function () {
                 var solver = new Ammo.btSequentialImpulseConstraintSolver();
                 this.dynamicsWorld = new Ammo.btDiscreteDynamicsWorld(dispatcher, overlappingPairCache, solver, collisionConfiguration);
 
+                if (this.dynamicsWorld.setInternalTickCallback) {
+                    var checkForCollisionsPointer = Ammo.addFunction(this._checkForCollisions.bind(this), 'vif');
+                    this.dynamicsWorld.setInternalTickCallback(checkForCollisionsPointer);
+                  } else {
+                    console.warn("WARNING: This version of ammo.js can potentially fail to report contacts. Please update it to the latest version.");
+                }
+
                 // Lazily create temp vars
                 ammoRayStart = new Ammo.btVector3();
                 ammoRayEnd = new Ammo.btVector3();
@@ -522,38 +529,16 @@ Object.assign(pc, function () {
             return r && (r.hasEvent("collisionstart") || r.hasEvent("collisionend") || r.hasEvent("contact"));
         },
 
-        onUpdate: function (dt) {
-            // #ifdef PROFILER
-            this._stats.physicsStart = pc.now();
-            // #endif
-
-            // Check to see whether we need to update gravity on the dynamics world
-            var gravity = this.dynamicsWorld.getGravity();
-            if (gravity.x() !== this.gravity.x || gravity.y() !== this.gravity.y || gravity.z() !== this.gravity.z) {
-                gravity.setValue(this.gravity.x, this.gravity.y, this.gravity.z);
-                this.dynamicsWorld.setGravity(gravity);
-            }
-
-            // Update the transforms of all bodies
-            this.dynamicsWorld.stepSimulation(dt, this.maxSubSteps, this.fixedTimeStep);
-
-            // Update the transforms of all entities referencing a body
-            var components = this.store;
-            for (var id in components) {
-                if (components.hasOwnProperty(id)) {
-                    var entity = components[id].entity;
-                    var componentData = components[id].data;
-                    if (componentData.body && componentData.body.isActive() && componentData.enabled && entity.enabled) {
-                        if (componentData.type === pc.BODYTYPE_DYNAMIC) {
-                            entity.rigidbody.syncBodyToEntity();
-                        } else if (componentData.type === pc.BODYTYPE_KINEMATIC) {
-                            entity.rigidbody._updateKinematic(dt);
-                        }
-                    }
-
-                }
-            }
-
+        /**
+         * @private
+         * @function
+         * @name pc.RigidBodyComponentSystem#_checkForCollisions
+         * @description Checks for collisions and fires collision events 
+         * @param {pc.Entity} dynamicsWorld - The pointer to the dynamics world that invoked this callback.
+         * @param {pc.Entity} timeStep - The amount of simulation time processed in the last simulation tick.
+         * @returns {void}
+         */
+        _checkForCollisions: function (dynamicsWorld, timeStep) {
             // Check for collisions and fire callbacks
             var dispatcher = this.dynamicsWorld.getDispatcher();
             var numManifolds = dispatcher.getNumManifolds();
@@ -705,6 +690,42 @@ Object.assign(pc, function () {
             this.contactPointPool.freeAll();
             this.contactResultPool.freeAll();
             this.singleContactResultPool.freeAll();
+        },
+
+        onUpdate: function (dt) {
+            // #ifdef PROFILER
+            this._stats.physicsStart = pc.now();
+            // #endif
+
+            // Check to see whether we need to update gravity on the dynamics world
+            var gravity = this.dynamicsWorld.getGravity();
+            if (gravity.x() !== this.gravity.x || gravity.y() !== this.gravity.y || gravity.z() !== this.gravity.z) {
+                gravity.setValue(this.gravity.x, this.gravity.y, this.gravity.z);
+                this.dynamicsWorld.setGravity(gravity);
+            }
+
+            // Update the transforms of all bodies
+            this.dynamicsWorld.stepSimulation(dt, this.maxSubSteps, this.fixedTimeStep);
+
+            // Update the transforms of all entities referencing a body
+            var components = this.store;
+            for (var id in components) {
+                if (components.hasOwnProperty(id)) {
+                    var entity = components[id].entity;
+                    var componentData = components[id].data;
+                    if (componentData.body && componentData.body.isActive() && componentData.enabled && entity.enabled) {
+                        if (componentData.type === pc.BODYTYPE_DYNAMIC) {
+                            entity.rigidbody.syncBodyToEntity();
+                        } else if (componentData.type === pc.BODYTYPE_KINEMATIC) {
+                            entity.rigidbody._updateKinematic(dt);
+                        }
+                    }
+
+                }
+            }
+
+            if (!this.dynamicsWorld.setInternalTickCallback)
+                this._checkForCollisions();
 
             // #ifdef PROFILER
             this._stats.physicsTime = pc.now() - this._stats.physicsStart;
--- a/src/framework/components/rigid-body/system.js
+++ b/src/framework/components/rigid-body/system.js
@@ -184,7 +184,7 @@ Object.assign(pc, function () {
                 if (this.dynamicsWorld.setInternalTickCallback) {
                     var checkForCollisionsPointer = Ammo.addFunction(this._checkForCollisions.bind(this), 'vif');
                     this.dynamicsWorld.setInternalTickCallback(checkForCollisionsPointer);
-                  } else {
+                } else {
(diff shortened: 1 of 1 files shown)


src/framework/components/rigid-body/system.js
```diff
@@ -522,38 +529,16 @@ Object.assign(pc, function () {
...
-                }
-            }
-
+        /**
+         * @private
+         * @function
+         * @name pc.RigidBodyComponentSystem#_checkForCollisions
+         * @description Checks for collisions and fires collision events
+         * @param {pc.Entity} dynamicsWorld - The pointer to the dynamics world that invoked this callback.
+         * @param {pc.Entity} timeStep - The amount of simulation time processed in the last simulation tick.
```
willeastcott, line 538: Type is wrong. Should be `number`.

src/framework/components/rigid-body/system.js
```diff
@@ -522,38 +529,16 @@ Object.assign(pc, function () {
...
-
-                }
-            }
-
+        /**
+         * @private
+         * @function
+         * @name pc.RigidBodyComponentSystem#_checkForCollisions
+         * @description Checks for collisions and fires collision events
+         * @param {pc.Entity} dynamicsWorld - The pointer to the dynamics world that invoked this callback.
```
willeastcott, line 537: Type is wrong. Should be `object`.

src/framework/components/rigid-body/system.js
```diff
@@ -705,6 +690,42 @@ Object.assign(pc, function () {
...
+                        } else if (componentData.type === pc.BODYTYPE_KINEMATIC) {
+                            entity.rigidbody._updateKinematic(dt);
+                        }
+                    }
+
+                }
+            }
+
+            if (!this.dynamicsWorld.setInternalTickCallback)
+                this._checkForCollisions();
```
willeastcott, line 728: Personally, I would pass `this.dynamicsWorld` and `dt` here, since that's the function signature now.

Maksims:
From warning text:
> This version of ammo.js can potentially fail to report contacts.
Is this due to this change, it always been a case?

Does this change can break any existing projects in development?

Also, please put warning log, under "Debug" build ifdef. Spamming logs in release builds, are discouraged.

src/framework/components/rigid-body/system.js
```diff
@@ -522,38 +529,16 @@ Object.assign(pc, function () {
...
+         * @function
+         * @name pc.RigidBodyComponentSystem#_checkForCollisions
+         * @description Checks for collisions and fires collision events
+         * @param {pc.Entity} dynamicsWorld - The pointer to the dynamics world that invoked this callback.
+         * @param {pc.Entity} timeStep - The amount of simulation time processed in the last simulation tick.
+         * @returns {void}
+         */
+        _checkForCollisions: function (dynamicsWorld, timeStep) {
             // Check for collisions and fire callbacks
             var dispatcher = this.dynamicsWorld.getDispatcher();
```
willeastcott, line 543: I'd change `this.dynamicsWorld` to just `dynamicsWorld` since it's a parameter of the function now.

src/framework/components/rigid-body/system.js
```diff
@@ -181,6 +181,13 @@ Object.assign(pc, function () {
                 var solver = new Ammo.btSequentialImpulseConstraintSolver();
                 this.dynamicsWorld = new Ammo.btDiscreteDynamicsWorld(dispatcher, overlappingPairCache, solver, collisionConfiguration);
 
+                if (this.dynamicsWorld.setInternalTickCallback) {
+                    var checkForCollisionsPointer = Ammo.addFunction(this._checkForCollisions.bind(this), 'vif');
+                    this.dynamicsWorld.setInternalTickCallback(checkForCollisionsPointer);
+                } else {
+                    console.warn("WARNING: This version of ammo.js can potentially fail to report contacts. Please update it to the latest version.");
```
willeastcott, line 188: Wrap this with #ifdef DEBUG and #endif. For example: https://github.com/playcanvas/engine/blob/master/src/scene/graph-node.js#L1050

src/framework/components/rigid-body/system.js
willeastcott: Also, I would say 'The dynamics world that...` instead of `The pointer to the dynamics world...` since this is JavaScript rather than C++. 😄

willeastcott:
No @Maksims, this doesn't break projects on older versions of ammo.js. And yep, I've logged a comment in my review about wrapping the warning with DEBUG define. 👍 

Thanks for the review!

src/framework/components/rigid-body/system.js
maxbbb: So because of the way WASM handles that callback, the dyanmicsWorld is returned at an integer pointer and you would have to make an extra call every step to get the underlying data which seems like it would have an impact on performance. They could be removed from the function signature and it will still work fine, but I figured in the future if you wanted to keep track of exact simulation time you would need the timeStep and would have to define the dynamicsWorld to get that.

src/framework/components/rigid-body/system.js
maxbbb: Given the point about the integer pointer, would @param {object | number} be appropriate here?

willeastcott: Sounds like it should be `number` 😄

src/framework/components/rigid-body/system.js
willeastcott: Gotcha. OK, then let's leave this 'as is'.

src/framework/components/rigid-body/system.js
maxbbb: If it's passed in when this function is called from the update function for versions of Ammo that don't have setInternalTickCallback then it will be an object.

src/framework/components/rigid-body/system.js
willeastcott: OK, forget this since `this.dynamicWorld` does not equal the `number` value passed from ammo.js.

src/framework/components/rigid-body/system.js
maxbbb: If I do that the linter complains about the else statement not being on the same line as the closing bracket of the if statement 😞

maxbbb: Linter didn't seem to have a problem with just wrapping the inside of the else. Pushed what should resolve the rest.

willeastcott: How about:

```
if (this.dynamicsWorld.setInternalTickCallback) {
    var checkForCollisionsPointer = Ammo.addFunction(this._checkForCollisions.bind(this), 'vif');
    this.dynamicsWorld.setInternalTickCallback(checkForCollisionsPointer);
}

#ifdef DEBUG
if (!this.dynamicsWorld.setInternalTickCallback) {
    console.warn("WARNING: This version of ammo.js can potentially fail to report contacts. Please update it to the latest version.");
}
#endif
```

willeastcott: Oh cool, sounds good.

willeastcott:
Hey @maxbbb, I've updated this slightly based on what I was saying before about fixing the function signature of the callback:

https://github.com/playcanvas/engine/pull/1997/commits/3a42c0f5c11a8f151ea00c8ec7380cb8571cd9da

Seems to work fine. 😄

maxbbb:
Yeah I can see for consistency why that would be the best approach. I haven't done any profiling on it so the performance difference might be incredibly small, but in an application where you decrease the fixedTimeStep to something like 1 / 240 that function is getting called a lot and I'm not sure of the cost of the wrapPointer function. I'll do some profiling and let you know.

willeastcott:
I suspect that the cost of wrapping the pointer is tiny compared to the cost of executing the body of `_checkForCollisions`, even when there are no contacts to process.

PR #2078, opened 2020-05-14
small reflection chunks cleanup

slimbuck:
Reflection chunks all pass material_reflectivity into the calcReflection() function just to have it package the value and return it to the caller. This seems unnecessary and so this PR sets the value in a more straightforward way.

@raytranuk - do you think this is an improvement? If not, I'm also happy to pull it.

### DIFF
--- a/src/graphics/program-lib/chunks/reflectionCC.frag
+++ b/src/graphics/program-lib/chunks/reflectionCC.frag
@@ -1,6 +1,6 @@
 #ifdef CLEARCOAT
 uniform float material_clearCoatReflectivity;
 void addReflectionCC() {    
-    ccReflection += calcReflection(ccReflDirW, ccGlossiness, material_clearCoatReflectivity); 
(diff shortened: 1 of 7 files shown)


willeastcott:
Is this change fully backwards compatible for anybody who's overridden these chunks?

slimbuck:
It depends whether users have overridden addReflection() or calcReflection(). This change alters the signature of calcReflection(), which is a breaking change. This is a small improvement, so perhaps not worth the risk.

willeastcott:
Well, if `addReflection` is considered "API" and `calcReflection` is internal to the chunks, the this change is fine, right?

raytranuk:
It looks fine - for some reason I didn't notice it was just passed out again to be used at a later stage - and as Will mentioned, addReflection() is what older projects used to override.

slimbuck:
Tested with Seemore and After the flood projects, they both seem to work fine.

PR #2080, opened 2020-05-14
[Physics] Fix Ammo dynamics world memory never being released 

### LINKED ISSUE #2064: Ammo js isn't released when application is destroyed
This bug happens when you have multiple components running one after the other on the browser. Each component creates a playcanvas application and as it's unmounted from DOM it destroys its pc application with `app.destroy()`. After about 16 of these iterations, ammo.js throws an out of memory exception. `ammo.wasm.js:292 Uncaught abort("OOM"). Build with -s ASSERTIONS=1 for more info`. On firefox, I have worked around it by setting `window.Ammo = null` and `window.AmmoLib = null` after `app.destroy()` but this doesn't work on chrome. On chrome this workaround makes it throw an error after only about 2 iterations of `memory access out of bounds` from ammo.js. Haven't tried it on other browsers but am having the feeling that ammo.js isn't properly released when an application is destroyed in playcanvas.

maxbbb:
Fixes #2064 

The rigidbody component system did not have a destroy function so each time Ammo was created the previous Ammo dynamics world objects were never released. 

### DIFF
--- a/src/framework/components/rigid-body/system.js
+++ b/src/framework/components/rigid-body/system.js
@@ -735,7 +735,7 @@ Object.assign(pc, function () {
         },
 
         destroy: function () {
-            if (Ammo) {
+            if (typeof Ammo !== 'undefined') {
                 Ammo.destroy(this.dynamicsWorld);
                 Ammo.destroy(this.solver);
                 Ammo.destroy(this.overlappingPairCache);
--- a/src/framework/components/rigid-body/system.js
+++ b/src/framework/components/rigid-body/system.js
@@ -175,11 +175,11 @@ Object.assign(pc, function () {
         onLibraryLoaded: function () {
             // Create the Ammo physics world
             if (typeof Ammo !== 'undefined') {
-                var collisionConfiguration = new Ammo.btDefaultCollisionConfiguration();
-                var dispatcher = new Ammo.btCollisionDispatcher(collisionConfiguration);
-                var overlappingPairCache = new Ammo.btDbvtBroadphase();
-                var solver = new Ammo.btSequentialImpulseConstraintSolver();
-                this.dynamicsWorld = new Ammo.btDiscreteDynamicsWorld(dispatcher, overlappingPairCache, solver, collisionConfiguration);
+                this.collisionConfiguration = new Ammo.btDefaultCollisionConfiguration();
+                this.dispatcher = new Ammo.btCollisionDispatcher(this.collisionConfiguration);
+                this.overlappingPairCache = new Ammo.btDbvtBroadphase();
+                this.solver = new Ammo.btSequentialImpulseConstraintSolver();
+                this.dynamicsWorld = new Ammo.btDiscreteDynamicsWorld(this.dispatcher, this.overlappingPairCache, this.solver, this.collisionConfiguration);
 
                 if (this.dynamicsWorld.setInternalTickCallback) {
                     var checkForCollisionsPointer = Ammo.addFunction(this._checkForCollisions.bind(this), 'vif');
@@ -732,6 +732,19 @@ Object.assign(pc, function () {
             // #ifdef PROFILER
             this._stats.physicsTime = pc.now() - this._stats.physicsStart;
             // #endif
+        },
+
+        destroy: function () {
+            Ammo.destroy(this.dynamicsWorld);
+            Ammo.destroy(this.solver);
+            Ammo.destroy(this.overlappingPairCache);
+            Ammo.destroy(this.dispatcher);
+            Ammo.destroy(this.collisionConfiguration);
+            this.dynamicsWorld = null;
+            this.solver = null;
+            this.overlappingPairCache = null;
+            this.dispatcher = null;
+            this.collisionConfiguration = null;
         }
     });
 
--- a/src/framework/components/rigid-body/system.js
+++ b/src/framework/components/rigid-body/system.js
@@ -735,16 +735,18 @@ Object.assign(pc, function () {
         },
 
         destroy: function () {
-            Ammo.destroy(this.dynamicsWorld);
-            Ammo.destroy(this.solver);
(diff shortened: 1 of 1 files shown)


willeastcott:
Brilliant! Thanks for this. 😄

maxbbb:
No problem, quick question though is there any reason in particular that the Ammo existence check is a typeof for undefined? I'm not sure if there is a situation in which it could be null but certainly that would cause problems if there was. Also this fix does not release the actual Ammo instance itself; Not sure if setting it to undefined would cause it to be garbage collected or not since it's a WASM module but that's something that you might want to look into.

willeastcott:
I don't think there's a specific reason. I mean, I guess it could be `if (Ammo) {` but, hey, it works. 😄

ssejjemba:
This fix introduced a new issue for me. Now when I destroy my application and create a new one, the rigidbody component system complains `playcanvas.js:32492: Uncaught TypeError: Cannot read property 'getGravity' of null`. Looks to me like it's still sniffing from the null reference... I think a new instance is not being created or something... what do you guys think?

willeastcott:
Are there easily reproducible steps to see what you're seeing? It's hard to debug this just by inspecting the code.

ssejjemba:
Well I am using the engine in a react project, but basically I have a re-usable component that creates a playcanvas application each mount and destroys it on unmout. prior to this fix it was crashing after 16 mounts and unmout cycles but with this fix due to memory overload... after this fix.. it can't mount the second component as `this.dynamicsWorld` is null from the new destruction fix... even after creating a `new pc.Application()` which is supposed to create new instances of all these according to my understanding... I am going to try to create a public mini project to reproduce this and paste the link here...

absconcept:
Thank you @ssejjemba , and thank you @willeastcott for the rapid response.

ssejjemba:
@willeastcott  just patched up a simple demo to reproduce the intents of my project. The project code can be found [here](https://github.com/ssejjemba/ReactPlaycanvas). I didn't add any fancy progress loaders so there is a blank canvas as the model is loaded but the idea is, to reproduce the bug you simply press next to change screen component (from playcanvas app to simple form) then press next again (from simple form to playcanvas) this will crash when creating a new playcanvas application as described above. This crash disappears if I don't destroy app when components are switching but that would mean the memory overload problem will then re-appear after some cycles. A link to a running demo of the project can be found [here](https://react-playcanvas.firebaseapp.com/)

maxbbb:
Swapping the `playcanvas.js` file in the node_module with the latest build from master solved the problem when I cloned your repo. You can clone the playcanvas repo and run `npm build` to get that file and test it for yourself. Slightly unrelated, but are you planning on moving these models around with physics? If not you should ignore Ammo altogether. If you are, I saw that you're using the model as the collision mesh and I would would be careful with that as these look like pretty high poly models and that will slow down the application drastically.

absconcept:
Hi Max, thanks for that, we use the physics for Ray casting, so we can place a pin on the 3D mesh. like the attached :) (the green blob is the 'pin' )

Good to know that the latest update fixes it, lets see what Daniel finds out.

image

ssejjemba:
Thanks @maxbbb for the help, what I want to ask is if I am using the npm module version of playcanvas which is `"playcanvas": "^1.27.4"` shouldn't this be pointing at the latest build from master on the engine repo. Asking this because as much as your solution would work on my local builds, I don't have any control over the node_modules of the server build.... Also I am kinda curious why it's different `playcanvas.js`, Is it something intentional or .... as I am puzzled as to why the release in the npm module is different from the master script of the same release...

willeastcott:
No, the release NPM packages are built from the [stable branch](https://github.com/playcanvas/engine/commits/stable). There have been several commits to master since the 1.27.4 release.

ssejjemba:
ohhhhh.... I see... so I have to wait for the next stable release..... Okay, I'll hope it's soon as it's my best option right now... Thanks @willeastcott for the instant reply...

ssejjemba:
@maxbbb I have just updated with the latest playcanvas release 1.28.3... but I still face the same issue `Uncaught TypeError: Cannot read property 'getGravity' of null  at RigidBodyComponentSystem.onUpdate (playcanvas.js:33472)`

absconcept:
Hi, any news on fixing this? or shall we fork our own branch and hack a fix?

absconcept:
Ok Guys, we are ditching Ammo.js as it seems to cause all sorts of problems! and this is hurting us, This ticket can be closed :)

willeastcott:
Sorry to hear that you're ditching ammo.js @absconcept. I'm open to finding a resolution. But I just need a super-simple reproducible test case. Ideally based on one of the engine examples perhaps.

ssejjemba:
none of the engine examples use playcanvas in a component based nature like we are doing. The example above uses Reactjs for this component-based nature of our requirements. But other than that the actual playcanvas component is easily reproducable. steps:
1.) Load ammo.js libraries to enable physics
2.) Mount canvas to DOM
3.) Create a new application `var app1 = new pc.Application(canvas)`
4.) Load asset1 to the canvas with rigid body and collision components on entity
5.) Destroy asset1 and application.
6.) Unmount canvas from DOM
5.) Repeat steps 2 - 6 for app2 and asset2

expected results:
The app should mount and unmount canvas drawing the appropriate models on DOM through both cycles.

actual results:
The app crashes on the second cycle with the error mentioned in the above comments.

Removing step 5 solves this issue but then crashes the app if you are to have more than about 16 cycles with the ammo js out of memory error that this issue was supposed to fix.

absconcept:
To help with figuring this out @willeastcott here is a Dev LINK [here](https://survey.sip.csiolabs.me/3dtag-debug?respondent_id=githubdebug
)

Load it, click next, then once on the 3D can, click next, then click back, then next and keep doing it, then it will crash, all down to ammo.js

I hope this helps :)

Issue #317, opened 2015-06-17
Support loading cubemap json 


vkalpias:
We used to be able to load a cubemap json for engine-only users. This feature was removed with the new resource loader. We need to add it back because exported models with cubemaps will not load.

daredevildave:
What is the format of the cubemap json and how to I generate one?

vkalpias:
The cubemap json is already generated when we export a model archive and one of the materials references a cubemap. The material cubemap property is changed to point to the cubemap json. The cubemap json simply contains the cubemap asset data. For example:

```
{
   "textures":[
      "../18062/f0.png",
      "../18065/f1.png",
      "../18067/f2.png",
      "../18061/f3.png",
      "../18063/f4.png",
      "../18064/f5.png"
   ],
   "minFilter":5,
   "name":"New Cubemap",
   "magFilter":1,
   "rgbm":true,
   "anisotropy":5
}
```

If the cubemap is prefiltered then the json will also contain a 'prefiltered' property that points to the dds file.

willeastcott:
Sorry, I'm not sure exactly what's missing here. We have a cubemap resource handler. The Editor seems to generate a valid zip for a cubemap. What do we need to do and (roughly) how long would it take to fix? (BTW, I just tried downloading a cubemap and loading it into one of the engine examples but the load function passed to loadFromUrl never gets called).

daredevildave:
When loaded from the editor, cubemaps reference textures by asset id. When they are downloaded for use outside of the editor (as a json file) they reference textures as a relative path. 

The resource handler needs to support path-based cubemaps which it doesn't at the moment.

PR #2082, opened 2020-05-18
handle cubemap asset faces specified as urls

### LINKED ISSUE #1902: CubemapHandler doesn't load faces
When creating a cubemap asset, the CubemapHandler doesn't create any assets for the face textures, so they are never loaded.

Example:
```js
const cubemapAsset = new pc.Asset('helipad', 'cubemap', {
    url: "../viewer/assets/cubemap/6079289/Helipad.dds"
}, {
  "textures": [
      "./assets/cubemap/6079292/Helipad_posx.png",
      "./assets/cubemap/6079290/Helipad_negx.png",
      "./assets/cubemap/6079293/Helipad_posy.png",
      "./assets/cubemap/6079298/Helipad_negy.png",
      "./assets/cubemap/6079294/Helipad_posz.png",
      "./assets/cubemap/6079300/Helipad_negz.png"
  ],
  "magFilter": 1,
  "minFilter": 5,
  "anisotropy": 1,
  "name": "Helipad",
  "rgbm": true,
  "prefiltered": "Helipad.dds"
});
app.assets.add(cubemapAsset);
app.assets.load(cubemapAsset);
cubemapAsset.ready(function () {
  app.scene.skyboxMip = 0;
  app.setSkybox(cubemapAsset);
});
```

With the above example, `cubemapAsset.loadFaces` is set to `true`, but the assets for the face textures are never created so a `skyboxMip` of `0` will always show a black skybox since the textures haven't been loaded.

Looking at the source code, I believe this is the part where it is failing: https://github.com/playcanvas/engine/blob/master/src/resources/cubemap.js#L196-L198

Changing the last two lines to the following seems to load the face textures properly:
```js
asset = new pc.Asset(id, "texture", { url: id });
assets.add(asset);
assets.load(asset);
assets.once("load:" + asset.id, assetReady);
assets.once("add:" + asset.id, assetAdded);
```

slimbuck:
Fixes #1902 and #317

The cubemap handler only supported loading faces specified as asset id. This PR adds support for faces specified as strings (urls) which is desirable for engine-only apps.

@Maksims - I would like to remove the 'loadFaces' property and check on the cubemap. Can you remember why it was added?


Maksims:
> @Maksims - I would like to remove the 'loadFaces' property and check on the cubemap. Can you remember why it was added?

Don't remember exactly, but probably to assist this scenario:
1. Cubemap asset has prefiltered data and used around scene.
2. If skybox uses mip 0, then it has to load faces (full size files).
3. If cubemap used as light cookie, then it has to load faces (full size files).

Otherwise, do not load faces.

slimbuck:
Thanks @Maksims. Is there a difference between specifying cubemap faces with loadFaces=false and not specifying faces?

Maksims:
> Thanks @Maksims. Is there a difference between specifying cubemap faces with loadFaces=false and not specifying faces?

Not sure I properly get the question. If faces are defined for asset, then they are defined and might be required later in application logic. Although they might not be required to preload but their prefiltered mipmapped cubemap might be required as preloaded.

What we found is this: you often have a cubemap, then you check individual textures of faces not to preload, but it would preload it anyway, as cubemap was set to preload. But those faces were not used anywhere, plus their individual assets (faces) were set not to preload.

slimbuck:
Thanks @Maksims that does make sense. I guess it's a consequence of a cubemap asset being both a cubemap texture and a set of lightprobe data. I am hoping to clean some of this up, but not yet sure how.

Maksims:
Indeed. Bear in mind, prefiltered cubemap also has Spherical Harmonics information for ambientSH property on materials.
SHs - is something we will need in engine, either stored as grids in assets, or some other ways.

slimbuck:
I would like to split out the prefiltered data into some 'light probe' resource which can be user configured with any of the supported setups (i.e. 6x prefiltered cubemaps, 1x prefiltered cubemap, ambientSH + dual paraboloid map, etc). The material/shader system would then interface with this directly.

PR #2084, opened 2020-05-19
GPU Morph Targets

mvaligursky:
This change moves morph target evaluation from CPU to GPU.

morphing

Pros:
* Performance. A large test mesh (20k vertices) with few morph targets took 200ms on CPU, and now it takes less than 1ms on GPU.

Cons:
* Flexibility. CPU solution allowed unlimited number of concurently active morph targets. GPU implementation allows
either 4 morph targets using position and normal, or 8 morph targets using only positions, or only normals. This is due to the implementation
using vertex attributes to provide active morph target data to GPU. On WebGl1 devices, there are 16 attributes available, and 8 are dedicated
to morphing. Note that more than 4 or 8 morph targets can be used on a mesh, but 4 or 8 of them with the highest weights are used to render each frame, while morph targets with lower weights are ignored.
* Memory use. CPU implementation used sparse morph target, but GPU implementation needs to use full morph targets.

**New APIs:**
* there are no new API's introduced

**Changed APIs:**
* small changes to Morph / MorphTarget / MorphInstance, primarily to pass GraphicsDevice where needed, and removed support for morphing of tangents as those are no longer needed by the engine.

**Other Changes:**
* parts of morph.js file has been separated to morph-target.js and morph-instance.js 
* evaluation of AABB of MeshInstance/MorphInstance/Morph/MorphTarget has been rewritten to be more flexible (any vertex format for skin weights / indices), better performance without temporary vertex arrays (mosty allowed due to non-sparse morph targets).
* not using morph targets from Draco compressed glb files as those do not work due to draco compression not preserving sparse morph target indices
* functionality to disable vertex attributes on a device
* adjusted json-model.js to use changed morhping api and convert sparse to full morph targets. Also removed support for tangent morphing.

**Engine examples:**
* {@link graphics/mesh-morph.html}

### DIFF
--- a/examples/graphics/mesh-morph.html
+++ b/examples/graphics/mesh-morph.html
@@ -149,9 +149,9 @@
             time += dt;
 
             // modify weights of all 3 morph targets along some sin curve with different frequency
-            self.meshInstance.morphInstance.setWeight(0, Math.abs(Math.sin(time)));
-            self.meshInstance.morphInstance.setWeight(1, Math.abs(Math.sin(time * 0.3)));
-            self.meshInstance.morphInstance.setWeight(2, Math.abs(Math.sin(time * 0.7)));
+            morphInstance.setWeight(0, Math.abs(Math.sin(time)));
+            morphInstance.setWeight(1, Math.abs(Math.sin(time * 0.3)));
+            morphInstance.setWeight(2, Math.abs(Math.sin(time * 0.7)));
 
             // orbit camera around
             camera.setLocalPosition(6 * Math.sin(time * 0.2), 4, 6 * Math.cos(time * 0.2));
--- a/examples/graphics/mesh-morph.html
+++ b/examples/graphics/mesh-morph.html
@@ -1,0 +1,163 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>PlayCanvas Mesh Morph</title>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0">
+    <link rel="icon" type="image/png" href="../playcanvas-favicon.png" />
+    <script src="../../build/output/playcanvas.js"></script>
+    <script src="../../build/output/playcanvas-extras.js"></script>
+    <style>
+        body { 
+            margin: 0;
+            overflow-y: hidden;
+        }
+    </style>
+</head>
+
+<body>
+    <!-- The canvas element -->
+    <canvas id="application-canvas"></canvas>
+
+    <!-- The script -->
+    <script>
+        var canvas = document.getElementById("application-canvas");
+
+        // Create the application and start the update loop
+        var app = new pc.Application(canvas);
+        app.start();
+
+        // Set the canvas to fill the window and automatically change resolution to be the same as the canvas size
+        app.setCanvasFillMode(pc.FILLMODE_FILL_WINDOW);
+        app.setCanvasResolution(pc.RESOLUTION_AUTO);
+
+        window.addEventListener("resize", function () {
+            app.resizeCanvas(canvas.width, canvas.height);
+        });
+
+        var miniStats = new pc.MiniStats(app);
+
+        // Create an entity with a directional light component
+        var light = new pc.Entity();
+        light.addComponent("light", {
+            type: "directional"
+        });
+        app.root.addChild(light);
+        light.setLocalEulerAngles(45, 30, 0);
+
+        // Create an entity with a camera component
+        var camera = new pc.Entity();
+        camera.addComponent("camera", {
+            clearColor: new pc.Color(0.1, 0.1, 0.1)
+        });
+        app.root.addChild(camera);
+
+        // helper function to return the shortest distance from point [x, y, z] to a plane defined by [a, b, c] normal
+        var shortestDistance = function (x, y, z, a, b, c) 
+        { 
+            var d = Math.abs((a * x + b * y + c * z)); 
+            var e = Math.sqrt(a * a + b * b + c * c); 
+            return d / e;
+        } 
+
+        // helper function that creates a morph target from original positions, normals and indices, and a plane normal [nx, ny, nz]
+        var createMorphTarget = function (positions, normals, indices, nx, ny, nz) {
+
+            // modify vertices to separate array
+            var modifiedPositions = new Float32Array(positions.length);
+            var dist, i, displacement;
+            var limit = 0.2;
+            for (i = 0; i < positions.length; i += 3) {
+                // distance of the point to the specified plane
+                dist = shortestDistance(positions[i], positions[i + 1], positions[i + 2], nx, ny, nz);
+
+                // modify distance to displacement amoint - displace nearby points more than distant points
+                displacement = pc.math.clamp(dist, 0, limit);
+                displacement = pc.math.smoothstep(0, limit, dist);
+                displacement = 1 - displacement;
+                
+                // generate new position by extruding vertex along normal by displacement
+                modifiedPositions[i] = positions[i] + normals[i] * displacement;
+                modifiedPositions[i + 1] = positions[i + 1] + normals[i + 1] * displacement;
+                modifiedPositions[i + 2] = positions[i + 2] + normals[i + 2] * displacement;
+            }
+
+            // generate normals based on modified positions and indices
+            var modifiedNormals = new Float32Array(pc.calculateNormals(modifiedPositions, indices));
+
+            // generate delta positions and normals - as morph targets store delta between base position / normal and modified position / normal
+            for (i = 0; i < modifiedNormals.length; i++) {
+                modifiedPositions[i] -= positions[i];
+                modifiedNormals[i] -= normals[i];
+            }
+
+            // create a morph target
+            return new pc.MorphTarget(app.graphicsDevice, {
+                deltaPositions: modifiedPositions,
+                deltaNormals: modifiedNormals
+            });
+        };
+
+        // create the base mesh - a sphere, with higher amount of vertices / triangles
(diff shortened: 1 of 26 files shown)


examples/graphics/mesh-morph.html
```diff
@@ -0,0 +1,163 @@
...
+        // update function called once per frame
+        var self = this;
+        var time = 0;
+        app.on("update", function (dt) {
+            time += dt;
+
+            // modify weights of all 3 morph targets along some sin curve with different frequency
+            self.meshInstance.morphInstance.setWeight(0, Math.abs(Math.sin(time)));
+            self.meshInstance.morphInstance.setWeight(1, Math.abs(Math.sin(time * 0.3)));
+            self.meshInstance.morphInstance.setWeight(2, Math.abs(Math.sin(time * 0.7)));
```
willeastcott, line 154: You can just do `morphInstance` instead of `self.meshInstance.morphInstance`.

mvaligursky: done, thanks!

src/graphics/program-lib/chunks/normal.vert
```diff
@@ -6,6 +6,24 @@ vec3 getNormal() {
     #else
         dNormalMatrix = matrix_normal;
     #endif
-    return normalize(dNormalMatrix * vertex_normal);
+
+    vec3 tempNormal = vertex_normal;
```
slimbuck, line 10: is vertex_normal used elsewhere? shouldn't it be updated directly by morph weights?

src/graphics/program-lib/chunks/normal.vert
```diff
@@ -6,6 +6,24 @@ vec3 getNormal() {
...
+
+    #ifdef MORPHING
+        #ifdef MORPHING_NRM03
+            tempNormal += morpth_weights_a[0] * morph_nrm0;
```
slimbuck, line 14: morpth_weights_a -> morph_weights_a

src/scene/morph-instance.js
```diff
@@ -0,0 +1,137 @@
...
+         * @function
+         * @name pc.MorphInstance#update
+         * @description Selects active morph targets and prepares morph for rendering. Called automatically by renderer.
+         */
+        update: function () {
+
+            var targets = this.morph._targets;
+
+            // collect active targets
+            this._activeTargets.length = 0;
```
slimbuck, line 80: it should be more efficient to:
```
this._activeTargets.length = targets.length
...
this._activeTargets[i] =...
```

You could also reuse the previous update's { absWeight: ... } objects

src/shape/bounding-box.js
```diff
@@ -10,14 +10,14 @@ Object.assign(pc, function () {
...
-     * @param {pc.Vec3} [halfExtents] - Half the distance across the box in each axis. The constructor takes a reference of this parameter.
+     * @param {pc.Vec3} [center] - Center of box. The constructor takes a copy of this parameter.
+     * @param {pc.Vec3} [halfExtents] - Half the distance across the box in each axis. The constructor takes a copy of this parameter.
      * @property {pc.Vec3} center Center of box.
      * @property {pc.Vec3} halfExtents Half the distance across the box in each axis.
      */
     var BoundingBox = function BoundingBox(center, halfExtents) {
-        this.center = center || new pc.Vec3(0, 0, 0);
-        this.halfExtents = halfExtents || new pc.Vec3(0.5, 0.5, 0.5);
+        this.center = center ? center.clone() : new pc.Vec3(0, 0, 0);
```
slimbuck, line 19: I think this is a breaking change - I believe there is code that -assumes- a reference is taken and updates it accordingly.

src/shape/bounding-sphere.js
```diff
@@ -12,11 +12,11 @@ Object.assign(pc, function () {
      * @example
      * // Create a new bounding sphere centered on the origin with a radius of 0.5
      * var sphere = new pc.BoundingSphere();
-     * @param {pc.Vec3} [center] - The world space coordinate marking the center of the sphere. The constructor takes a reference of this parameter.
+     * @param {pc.Vec3} [center] - The world space coordinate marking the center of the sphere. The constructor takes a copy of this parameter.
      * @param {number} [radius] - The radius of the bounding sphere. Defaults to 0.5.
      */
     function BoundingSphere(center, radius) {
-        this.center = center || new pc.Vec3(0, 0, 0);
+        this.center = center ? center.clone() : new pc.Vec3(0, 0, 0);
```
slimbuck, line 19: Ditto

src/shape/frustum.js
```diff
@@ -15,8 +15,8 @@ Object.assign(pc, function () {
      * @param {pc.Mat4} viewMatrix - The inverse of the world transformation matrix for the frustum.
      */
     var Frustum = function Frustum(projectionMatrix, viewMatrix) {
-        projectionMatrix = projectionMatrix || new pc.Mat4().setPerspective(90, 16 / 9, 0.1, 1000);
-        viewMatrix = viewMatrix || new pc.Mat4();
+        projectionMatrix = projectionMatrix ? projectionMatrix.clone() : new pc.Mat4().setPerspective(90, 16 / 9, 0.1, 1000);
```
slimbuck, line 18: Ditto, this can break existing code. (Though I think this is the correct way of doing things).

src/shape/oriented-box.js
```diff
@@ -11,12 +11,12 @@ Object.assign(pc, function () {
      * @classdesc Oriented Box.
      * @property {pc.Mat4} [worldTransform] The world transform of the OBB.
      * @param {pc.Mat4} [worldTransform] - Transform that has the orientation and position of the box. Scale is assumed to be one.
-     * @param {pc.Vec3} [halfExtents] - Half the distance across the box in each local axis. The constructor takes a reference of this parameter.
+     * @param {pc.Vec3} [halfExtents] - Half the distance across the box in each local axis. The constructor takes a copy of this parameter.
      */
     var OrientedBox = function OrientedBox(worldTransform, halfExtents) {
-        this.halfExtents = halfExtents || new pc.Vec3(0.5, 0.5, 0.5);
+        this.halfExtents = halfExtents ? halfExtents.clone() : new pc.Vec3(0.5, 0.5, 0.5);
```
slimbuck, line 17: Potentially breaking

src/shape/plane.js
```diff
@@ -7,12 +7,12 @@ Object.assign(pc, function () {
...
      * @description Create an infinite plane.
-     * @param {pc.Vec3} [point] - Point position on the plane. The constructor takes a reference of this parameter.
-     * @param {pc.Vec3} [normal] - Normal of the plane. The constructor takes a reference of this parameter.
+     * @param {pc.Vec3} [point] - Point position on the plane. The constructor takes a copy of this parameter.
+     * @param {pc.Vec3} [normal] - Normal of the plane. The constructor takes a copy of this parameter.
      */
     var Plane = function Plane(point, normal) {
-        this.normal = normal || new pc.Vec3(0, 0, 1);
-        this.point = point || new pc.Vec3(0, 0, 0);
+        this.normal = normal ? normal.clone() : new pc.Vec3(0, 0, 1);
```
slimbuck, line 14: Also potentially breaking

src/shape/ray.js
```diff
@@ -8,16 +8,16 @@ Object.assign(pc, function () {
...
-     * @param {pc.Vec3} [direction] - The direction of the ray. The constructor takes a reference of this parameter.
+     * @param {pc.Vec3} [direction] - The direction of the ray. The constructor takes a copy of this parameter.
      * Defaults to a direction down the world negative Z axis (0, 0, -1).
      * @property {pc.Vec3} origin The starting point of the ray.
      * @property {pc.Vec3} direction The direction of the ray.
      */
     var Ray = function Ray(origin, direction) {
-        this.origin = origin || new pc.Vec3(0, 0, 0);
-        this.direction = direction || new pc.Vec3(0, 0, -1);
+        this.origin = origin ? origin.clone() : new pc.Vec3(0, 0, 0);
```
slimbuck, line 19: Ditto

src/graphics/program-lib/chunks/normal.vert
mvaligursky: this seems the only place where it's used in generic sense. There's a specialized chunk for skinning only, and for instancing only.

You cannot modify vertex_normal, I tried that, those are input attributes and you get shader compiler error.

This seems consistent with previous state of this file - we modify vertex_normal by skinning and instancing, I just added morphing.

src/graphics/program-lib/chunks/normal.vert
mvaligursky: fixing! hw did I missed that. At least it's consistent across all uses, so it works ;)

src/shape/bounding-box.js
mvaligursky: good point, moving these changes to separate PR

src/shape/frustum.js
mvaligursky: good point, moving these changes to separate PR

src/shape/plane.js
mvaligursky: good point, moving these changes to separate PR

src/shape/ray.js
mvaligursky: good point, moving these changes to separate PR

src/shape/oriented-box.js
mvaligursky: good point, moving these changes to separate PR

src/shape/bounding-sphere.js
mvaligursky: good point, moving these changes to separate PR

src/scene/morph-instance.js
mvaligursky: good suggestion, done, please do one last review on this change

PR #2086, opened 2020-05-19
Reset state of button on enable

### LINKED ISSUE #2085: Disabling a button on the event callback leaves it in a 'hovered' state when re-enabled
Test project: https://playcanvas.com/editor/scene/924089

Code:
```
UiManager.prototype.initialize = function() {
    var self = this;
    this.listenToClickButtonEntity.button.on('click', function () {
        console.log('click');
        self.listenToClickButtonEntity.enabled = false;
    });
    
    this.enableClickButtonEntity.button.on('click', function ()  {
        self.listenToClickButtonEntity.enabled = true;
    });
    
    this.listenToDownButtonEntity.button.on('pressedstart', function () {
        console.log('pressedstart');
        self.listenToDownButtonEntity.enabled = false;
    });
    
    this.enableDownButtonEntity.button.on('click', function ()  {
        self.listenToDownButtonEntity.enabled = true;
    });
};
```

Kapture 2020-05-19 at 16 11 31

In the case of listening for the 'pressedstart' event, it fires again when re-enabled.

yaustar:
Fixes #2085 

### DIFF
--- a/src/framework/components/button/component.js
+++ b/src/framework/components/button/component.js
@@ -476,6 +476,11 @@ Object.assign(pc, function () {
         },
 
         onEnable: function () {
+            // Reset input state
+            this._isHovering = false;
+            this._hoveringCounter = 0;
+            this._isPressed = false;
+
             this._imageReference.onParentComponentEnable();
             this._toggleHitElementListeners('on');
             this._forceReapplyVisualState();


Maksims:
Should those changes be applied onDisable instead of onEnable?
This is to ensure appropriate events to be fired related to change of hover states.

yaustar:
Yeah, I guess that would make sense as the subscriber would expect to have the end events 🤔

yaustar:
Just checked the the events being fired and they are the same regardless if that code change is on onEnable or onDisable.

It does bring up another bug where 'pressedend' would fire when the button is reenabled.
Kapture 2020-05-21 at 18 45 06

PR #2087, opened 2020-05-19
Cleanup loadFromUrl, construct relative asset URLs correctly

### LINKED ISSUE #1813: AssetsRegistry, prefix with url caching inconsistencies
There is a prefix feature in AssetsRegistry. When loading assets, it is mostly respected, but there are some areas where it is not respected well, leading to issues. For example, we [cache list of loaded assets by URLs not respecting prefix](https://github.com/playcanvas/engine/blob/master/src/asset/asset-registry.js#L196-L197). And it does not respect variants as well. Then when we load model, we load materials from mappings, and [trying to retrieve material from index](https://github.com/playcanvas/engine/blob/master/src/resources/model.js#L127-L130), which uses getFileUrl, which then respects [variants](https://github.com/playcanvas/engine/blob/master/src/asset/asset.js#L168) and adds [prefix](https://github.com/playcanvas/engine/blob/master/src/asset/asset.js#L175-L176).
So it will never work, once prefix is set to anything.

**Question:** Do we add prefix into cache keys, and ensure we subscribe to full paths with prefixes (as is now).
Or do we change the way we try to retrieve asset from that cache, by making url without prefix to match current cache keys?

slimbuck:
Fixes #1813

This PR includes the following:
- simplified loadFromUrl
- added a helper function to generate URL from a path relative to another asset
- handle destroying assets with multiple resources correctly


src/asset/asset-registry.js
```diff
@@ -480,184 +472,127 @@ Object.assign(pc, function () {
                 filename: filename || name,
                 url: url
             };
-            var data = {};
 
             var asset = self.getByUrl(url);
             if (!asset) {
-                asset = new pc.Asset(name, type, file, data);
+                asset = new pc.Asset(name, type, file, {});
```
willeastcott, line 478: You don't need to pass the empty object for the last parameter here.

willeastcott: The API ref is actually wrong. The `data` parameter should be marked as optional.

slimbuck: Thanks Will, I updated the code and the param jsdoc.

PR #2106, opened 2020-05-28
Basic/simple/unobtrusive Parcel support

### LINKED ISSUE #2103: PlayCanvas needs a proper build system
Currently, the engine is built using the [`build\build.js`](https://github.com/playcanvas/engine/blob/master/build/build.js), a Node.js script that loads all scripts that appear in [`build\dependencies.txt`](https://github.com/playcanvas/engine/blob/master/build/dependencies.txt) and concatenates them. This concatenated script is then compiled (optimized/minified) with the Google Closure Compiler. This works OK. However, it's time we considered the benefits of a proper JS module builder.

Options are probably:

* [Webpack](https://webpack.js.org/)
* [Rollup](https://rollupjs.org/)
* [Parcel](https://parceljs.org/)
* [Browserify](http://browserify.org/)
* [esbuild](https://github.com/evanw/esbuild)

There are lots of blog articles advocating the different options. For example, [here's one](https://blog.bitsrc.io/choosing-the-right-javascript-bundler-in-2020-f9b1eae0d12b) advocating for Parcel (it has an interesting comparison table).

Some benefits:

* We can easily output different flavors of the engine (browser, Node.js, module).
* We can trivially leverage lots of plugins (Babel, Typescript, etc).
* We can easily generate source maps.
* Potentially reduce engine footprint (current module structure is quite verbose and could be simplified).

(Please comment below and if you have additional benefits or even disadvantages and I'll add them to this issue description. Feel free to advocate for your build system preference to help us make a decision).

kungfooman:
Fixes #2103

I tried to change as few files as possible, so it is quite unobtrusive, yet useful.

image

In "watch" mode via `npm run parcel:watch` (observing saved files and recompile on the fly) it even just takes half a second:

image

So `npm run parcel:watch` would be the best bet for quick development cycles (no need to constantly switch to a console/shell for triggering a rebuild).


willeastcott:
This looks promising! Some observations/comments:

* The Parcel generated playcanvas.js is ~60KB larger than playcanvas.min.js built with the `build:min` script. This isn't insignificant.
* The `parcel:build` script requires you to run `build` first in order to generate `graphics/program-lib/chunks/generated-shader-chunks.js`.
* The output engine build will need the copyright notice added at the top.
* The engine source has lots of occurrences of:

```javascript
// #ifdef DEBUG

// #endif
```

Those codeblocks are included in the engine generated from `build:debug`. We'd need to support that. Same deal for the PROFILER build.
* What engine builds do we want to generate? Currently, `playcanvas.js` can be used in both Node.js and the browser. Does the Parcel build also work with Node?

kungfooman:
Thank you for your feedback, in order to fix most of these issues I started to implement a custom Parcel bundler, because it requires deeper integration with Parcel

kungfooman:
I am down to a ~7kb difference, so I wondered about the actual code difference, and I compared the minified files. Boils down to this and I started an issue for it: https://github.com/parcel-bundler/parcel/issues/4676

Zipped files have ~1kb difference:

image

willeastcott:
Very interesting! I've subscribed to that issue. Doesn't Parcel use terser to do the minification though? So wouldn't it be an issue with that software instead?

kungfooman:
Oops, good point, should probably be an issue in https://github.com/terser/terser then

willeastcott:
I'd be quite pleased if we could get rid of Closure Compiler actually, because our build process currently requires the JDK to be installed. And that slows down each run of the continuous integration.

build/JSAssetPlayCanvas.js
```diff
@@ -0,0 +1,68 @@
+var Preprocessor = require("preprocessor");
+const JSAsset = require("parcel-bundler/src/assets/JSAsset");
+//const { parse } = require("ifdef-loader/preprocessor");
+
+/*
+JSAssetPlayCanvas JSAssetPlayCanvas {
+  id: 'framework/components/registry.js',
+  name: 'C:\\web\\playcanvas-engine\\src\\framework\\components\\registry.js',
```
willeastcott, line 8: I'm sure you are well aware of all this, but this script is very specific to your machine. 😄

kungfooman:
I solved most issues you mentioned, just the node build doesn't work. I didn't change the old build system, so this PR is basically just for people with specific needs, e.g:

1) Who want to test Parcel to start migration towards ES6 or TypeScript
2) To get rid of JDK
3) Use `npm run parcel:watch` for quicker development cycles (full rebuild in half a second upon file save)

> * What do we need to do to be able to generate a proper module build?

Problem with the node build currently is the `Object.assign(pc, ...)` pattern, which causes `pc` to be undefined. Pretty sure that would be fixable aswell somehow without code change, but probably the best solution would be ES6 import/module style (and that would be quite some code refactoring for every single JS file).

willeastcott:
I don't quite get how `Object.assign(pc, ...)` causes `pc` to be undefined. Could you explain in a bit more detail please?

One thing I'm actually keen to do is get rid of that pattern. It bloats the library and makes initial parse slower. Basically, the entire engine codebase should not have to (internally) access anything via the `pc.` namespace. At the end of the library code, it should add all public API to `pc`.

kungfooman:
Sorry for the strange wording, `./src/playcanvas.js` is the entry point for Parcel and it looks like this:

```js
import "./polyfill/array-find";
import "./polyfill/math-log2";
import "./polyfill/math-sign";
import "./polyfill/object-assign";
import "./polyfill/pointer-lock";
import "./polyfill/request-animation-frame";
import "./polyfill/string";
import { pc } from "./core/core";
console.log("pc is", pc)
import "./core/color";
import "./core/guid";
```

And it is converted to something like:

```js
require("./polyfill/array-find");
require("./polyfill/math-log2");
require("./polyfill/math-sign");
require("./polyfill/object-assign");
require("./polyfill/pointer-lock");
require("./polyfill/request-animation-frame");
require("./polyfill/string");
var _core = require("./core/core");
require("./core/color");
require("./core/guid");
console.log("pc is", _core.pc);
module.exports = _core.pc;
```

So if we don't hack `pc` into the globals, then the `Object.assign(pc, ...)` pattern in `require("./core/color");` will fail to execute because `pc is undefined` (`pc` is just not included in any ES6 way like `import {pc} from "core/core"`).

I added this code now in `core/core.js`, so it works for node and browser, to prevent this:

```js
// Parcel requirement for node, otherwise pc is undefined
if (typeof global !== 'undefined')
    global.pc = pc;
```

So now in node we can do `require("./build/output/playcanvas.js")`, like:

image

I completely see your point for getting rid of the `Object.assign(pc, ...)` pattern, I just did a little test, replaced every occurance of `pc.` with nothing, that saves ~8kb of file size. Kinda funny that size difference is the current overhead from Parcel compared to `npm run build:min`:

image

Not sure how well JS engines are optimizing this, but `new pc.Vec3` is also adding the `Vec3` key lookup overhead everywhere e.g.

kungfooman:
I believe this is as unobtrusive as possible now 😅

But it requires: `cd build-parcel` and `npm install` (so `parcel-bundler` is only installed for people who actually want to use it)

willeastcott:
This PR fails the code lint step at the moment because of the `playcanvas.js` file. We'd either need to update the lint rules or exclude `playcanvas.js` from the checks.

So is there anything beyond fixing that that needs to be done here? What's next? Should we have two alternative build systems or should we pick one?

kungfooman:
> This PR fails the code lint step at the moment because of the `playcanvas.js` file. We'd either need to update the lint rules or exclude `playcanvas.js` from the checks.

Thanks, fixed that by moving the `playcanvas.js` entry file into `build-parcel` dir itself.

> So is there anything beyond fixing that that needs to be done here? What's next? Should we have two alternative build systems or should we pick one?

I would just like a transition time, so other devs can just test it for a while:
 * Gaining experience with Parcel/parcel-bundler
 * Implementing missing features and fixing potential bugs
 * Figuring out pros/cons of each build system

And then just biting the bullet and convert to TypeScript at some point 😂

willeastcott:
Hmm. Looks like CI is failing still. Can you merge from master branch? That should fix it.

kungfooman:
Yep, looks like it worked now

willeastcott:
OK, well, this is pretty cool. I think we're in a position to merge this now. But let's deploy v1.28.0 tomorrow and I'll merge straight afterwards. Exciting!

PR #2108, opened 2020-05-29
Cubemap load textures

### LINKED ISSUE #1983: BASIS textures assets not usable with create cubemap
When creating a CUBEMAP asset from BASIS compressed textures, the cubemap not created. If I use JPG files, everything works fine.
I don't need create prefiltered cubemap, I need use cubemap with custom shader as 'samplerCube'  uniform on posteffect.

https://forum.playcanvas.com/t/basis-texture-for-cubemap/12505

slimbuck:
Fixes #1983

The cubemap handler was assuming face textures are always img types (png/jpg).

This PR changes the handler to support non-img type images too.

### DIFF
--- a/src/resources/cubemap.js
+++ b/src/resources/cubemap.js
@@ -176,14 +176,25 @@ Object.assign(pc, function () {
 
                     // when all faces checked
                     if (count === 6 && levelsUpdated) {
-                        // set cubemap sources
-                        cubemap._width = sources[0].width;
-                        cubemap._height = sources[0].height;
-                        cubemap._format = sources[0].format;
-                        cubemap._levels[0] = sources.map(function (t) {
-                            return t._levels[0];
+
+                        // extract level data from source textures
+                        var levels = [];
+                        for (var mip = 0; mip < sources[0]._levels.length; ++mip) {
+                            levels.push(sources.map(function (s) {
+                                return s._levels[mip];
+                            }));
+                        }
+
+                        // reconstruct cubemap with new data
+                        assetCubeMap.resource = new pc.Texture(self._device, {
+                            name: "cubemap-faces",
+                            cubemap: true,
+                            rgbm: assetCubeMap.data.rgbm,        // take rgbm flag from asset data
+                            width: sources[0].width,
+                            height: sources[0].height,
+                            format: sources[0].format,
+                            levels: levels
                         });
-                        cubemap.dirtyAll();
 
                         // trigger load event (resource changed)
                         assets.fire('load', assetCubeMap);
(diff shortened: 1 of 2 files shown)


src/resources/cubemap.js
```diff
@@ -170,13 +167,41 @@ Object.assign(pc, function () {
...
+                        // reconstruct cubemap with new data
+                        assetCubeMap.resource = new pc.Texture(self._device, {
+                            name: "cubemap-faces",
+                            cubemap: true,
+                            rgbm: sources[0].rgbm,
+                            width: sources[0].width,
+                            height: sources[0].height,
+                            format: sources[0].format,
+                            levels: levels
+                        });
```
willeastcott, line 204: Does this overwrite a cubemap already set on `assetCubeMap.resource`? Does that leak anything?

src/resources/cubemap.js
```diff
@@ -170,13 +167,42 @@ Object.assign(pc, function () {
...
-                        cubemap.setSource(sources);
+                    if (levelsUpdated) {
+
+                        // build levels
+                        var levels = [];
+                        for (var mip = 0; mip < sources[0]._levels.length; ++mip) {
+                            levels.push(sources.map(function (s) {
+                                // use 'this' to overcome eslint no-loop-func when referencing mip in loop
+                                return s._levels[this];
+                            }, mip));
```
willeastcott, line 192: 🤢 How about:

 `// eslint-disable-line no-loop-func`

src/resources/cubemap.js
slimbuck: That's a good point.

The previous approach used texture.setSource() which internally modifies the texture's _width and _height based on the levels passed in.

This approach can't work for non-img sources (due to there being no width and height but also the pixel format is not necessarily RGBA8, _compressed flag could be different etc etc) and so we must construct the texture from scratch.

On first load the cubemap texture being replaced is a 4x4 blank texture created when cubemap loading starts. Ideally this just wouldn't be created in the first place. However there may be users depending on that behaviour.

I'm not sure that it's safe to call assetCubeMap.resource.destroy on the old cubemap because that texture may be in use elsewhere. Is there another option?

src/resources/cubemap.js
slimbuck: lol I tried that, didn't seem to work. I'll try again.

slimbuck: worked, thanks!

src/resources/cubemap.js
slimbuck: On second thought, I think destroying the old texture should be fine. Code updated accordingly.

willeastcott: I don't know, sorry. 😟  But I'm not sure why somebody would rely on the 4x4 blank texture behavior...

leonidaspir:
This seems to have been merged but currently I observe that both BASIS textures and legacy compressed textures break a cubemap used for skybox.

image

Here is how it should look:

image

And here is how it looks after I compress the source files:

image

If I delete the prefiltered data, and try to prefilter again while the faces are compressed then It results in a black cubemap:

image

PR #2114, opened 2020-05-30
Particle system optimizations

willeastcott:


src/framework/components/particle-system/component.js
```diff
@@ -574,10 +574,69 @@ Object.assign(pc, function () {
...
+                }
+                if (numSteps) {
+                    numSteps = Math.min(numSteps, emitter.maxSubSteps);
+                    for (i = 0; i < numSteps; i++) {
+                        emitter.addTime(emitter.fixedTimeStep, false);
+                    }
+                    var stats = this.system.app.stats.particles;
+                    stats._updatesPerFrame += numSteps;
+                    stats._frameTime += emitter._addTimeTime;
+                    emitter._addTimeTime = 0;
```
yaustar, line 631: Does this need to be wrapped in `#ifdef PROFILER`?

src/framework/components/particle-system/component.js
```diff
@@ -696,27 +755,43 @@ Object.assign(pc, function () {
...
-            if (this.system.app.scene.layers) {
-                this.system.app.scene.layers.off("add", this.onLayerAdded, this);
-                this.system.app.scene.layers.off("remove", this.onLayerRemoved, this);
+            var data = this.data;
+            var system = this.system;
+
+            // Remove from enabled particle systems array
+            var idx = system._enabledParticleSystems.indexOf(this);
+            if (idx > -1) {
+                system._enabledParticleSystems.splice(idx, 1);
```
yaustar, line 780: Minor: I would prefer this to be done through a function/property as `_enabledParticleSystems` is considered private to the `system`. But if we do this elsewhere, that's fine.

src/framework/components/particle-system/component.js
```diff
@@ -574,10 +574,69 @@ Object.assign(pc, function () {
             }
         },
 
+        _update: function (dt) {
+            var emitter = this.emitter;
+            if (!emitter.meshInstance.visible) return;
+
```
yaustar, line 580: If the entity/component is disabled, is `emitter.meshInstance.visible` false anyway and therefore the emitter is not updated?

raytranuk (commented):
A description of what this optimizes and high level as to how, would useful - from what I gather: iteration per particle is done per emitter instead of per system - and I also see some code tidy up. Profiling numbers before and after would be great too.

willeastcott:
This has gone a bit stale so I'm closing it (for now).

PR #2115, opened 2020-05-31
chore: build next

### LINKED ISSUE #2103: PlayCanvas needs a proper build system

hxhieu:
Fixes #2103 

This is a proposal for the new build system, still need to replicate all original build flags.
I have only created the minified build and also update the tests for this proof-of-concept.

It is using [esbuild](https://github.com/evanw/esbuild) with zero configuration, just to minify the source.

The new min file is ~40KB > than the current build - I guess because of the shader code is not yet minified since they are not JS code. We can investigate in minifying them just as same as current build process.

image

And all tests passed on the output minified file

image


willeastcott:
This is very interesting. But I don't think this will be useful until it's actually processing the JS files beyond minification. One thing I'm keen to do after developing a new build system is adopting either ES6 or TS. The Parcel PR sets us up for that a bit better.

hxhieu:
@willeastcott Thanks for the comment, esbuild is doing totally fine with either ES6 or TS and it is the fastest TS transpiler on Earth at the moment.

Let me sleep first then I will update this PR 😆

kungfooman:
Too early for testing I guess, I just have this error:

image

How will esbuild handle `// #ifdef DEBUG` etc., would every sophisticated project need a custom Go compilation/build or is there a plugin API for each asset type?

hxhieu:
@kungfooman sorry about the convenience, it was Windows line endings issue. So I have pushed a crude fix for it for now.

And to answer your question - no esbuild has not supported plugins just yet and I will need to hand code the directives replacement for now. Please head over the [ issue discussion](https://github.com/playcanvas/engine/issues/2103#issuecomment-636544676) to get my response on the next step.

But hey thanks for promoting Parcel and I hope we could eventually improve your implementation as I could see that we still have a lot of room for improvements over there.

PR #2122, opened 2020-06-03
Viewer - morph target controls changes

mvaligursky:
- morph target sliders are per mesh now instead of global list, allowing individual control
- scroll bar for left control panel, allowing large list of morph target controls


slimbuck:
cool!

PR #2125, opened 2020-06-04
split out texture parsers and coalesce texture types into a single enum

slimbuck:
This PR splits out the texture resource handler parsers for file formats.

It also introduces a new enum for texture type:
```
TEXTURETYPE_GENERAL: 1,
TEXTURETYPE_RGBM: 2,
TEXTURETYPE_RGBE: 3,
TEXTURETYPE_SWIZZLEGGGR: 4,

```

This enum replaces the previous boolean flags on pc.Texture for rgbm and swizzleGGGR with a 'type' parameter instead.

### DIFF
--- a/src/graphics/graphics.js
+++ b/src/graphics/graphics.js
@@ -1014,6 +1014,38 @@
          */
         TEXTURELOCK_WRITE: 2,
 
+        /**
+         * @constant
+         * @name pc.TEXTURETYPE_GENERAL
+         * @type {number}
+         * @description Texture is a general type.
+         */
+        TEXTURETYPE_GENERAL: 1,
+
+        /**
+         * @constant
+         * @name pc.TEXTURETYPE_RGBM
+         * @type {number}
+         * @description Texture stores high dynamic range data in RGBM format
+         */
+        TEXTURETYPE_RGBM: 2,
+
+        /**
+         * @constant
+         * @name pc.TEXTURETYPE_GENERAL
+         * @type {number}
+         * @description Texture stores high dynamic range data in RGBE format
+         */
+        TEXTURETYPE_RGBE: 3,
+
+        /**
+         * @constant
+         * @name pc.TEXTURETYPE_GENERAL
+         * @type {number}
+         * @description Texture stores normalmap data swizzled in GGGR format
+         */
+        TEXTURETYPE_SWIZZLEGGGR: 4,
+
         TEXHINT_NONE: 0,
         TEXHINT_SHADOWMAP: 1,
         TEXHINT_ASSET: 2,
--- a/src/graphics/texture.js
+++ b/src/graphics/texture.js
@@ -48,7 +48,9 @@ Object.assign(pc, function () {
      * @param {boolean} [options.mipmaps] - When enabled try to generate or use mipmaps for this texture. Default is true
      * @param {boolean} [options.cubemap] - Specifies whether the texture is to be a cubemap. Defaults to false.
      * @param {boolean} [options.volume] - Specifies whether the texture is to be a 3D volume (WebGL2 only). Defaults to false.
+     * @param {number} [options.type] - Specifies the image type, see {@link pc.TEXTURETYPE_GENERAL}
      * @param {boolean} [options.rgbm] - Specifies whether the texture contains RGBM-encoded HDR data. Defaults to false.
+     * @param {boolean} [options.rgbe] - Specifies whther the texture contains RGBE-encoded HDR data. Defaults to false.
      * @param {boolean} [options.swizzleGGGR] - Specifies whether the texture contains swizzled GGGR data for use with tangent space normal
      * maps. The R component is stored in alpha and G is stored in RGB. This packing can result in higher quality when the texture data
      * is compressed. Defaults to false.
@@ -100,8 +102,7 @@ Object.assign(pc, function () {
         this._depth = 1;
 
         this._format = pc.PIXELFORMAT_R8_G8_B8_A8;
-        this.rgbm = false;
-        this.swizzleGGGR = false;
+        this.type = pc.TEXTURETYPE_GENERAL;
 
         this._cubemap = false;
         this._volume = false;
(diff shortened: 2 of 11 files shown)


src/graphics/graphics.js
```diff
@@ -1014,6 +1014,38 @@
          */
         TEXTURELOCK_WRITE: 2,
 
+        /**
+         * @constant
+         * @name pc.TEXTURETYPE_GENERAL
+         * @type {number}
+         * @description Texture is a general type.
+         */
+        TEXTURETYPE_GENERAL: 1,
```
mvaligursky, line 1023: maybe should be texture semantic instead of type?

src/graphics/texture.js
```diff
@@ -545,6 +554,33 @@ Object.assign(pc, function () {
         }
     });
 
+    Object.defineProperty(Texture.prototype, 'rgbm', {
```
mvaligursky, line 557: you could use Object.defineProperties (see mesh.js) to group them together.

src/resources/parser/texture/img.js
```diff
@@ -0,0 +1,89 @@
...
+                callback(null, image);
+            };
+
+            image.onerror = function () {
+                // Retry a few times before failing
+                if (retryTimeout) return;
+
+                if (retryRequests && ++retries <= maxRetries) {
+                    var retryDelay = Math.pow(2, retries) * 100;
+                    console.log(pc.string.format("Error loading Texture from: '{0}' - Retrying in {1}ms...", originalUrl, retryDelay));
```
mvaligursky, line 65: this should be warning maybe, as we're not giving up yet

src/graphics/texture.js
slimbuck: I was going to submit a separate PR changing all texture properties, but I've just done it here instead.

src/graphics/graphics.js
slimbuck: I prefer type over semantic, which sounds to me like a vertex thing. I'm happy to change if there is consensus though, @willeastcott @raytranuk what do you think?

willeastcott: `pc.TEXTURETYPE_RGB`? 😆  I've got more of a problem with `GENERAL` than `TYPE`. But I prefer `TYPE` to `SEMANTIC`.

willeastcott: Also, what about `SRGB` textures? Does this fall under this enum? Or a separate flag?

slimbuck: Agreed on the general naming, it bugged me too :) Though RGB is not entirely accurate either, could be RGBA.

Linear vs gamma is a little subtle. Usually it is inferred from the texture usage (normal maps are assumed 'linear' and diffuse are assumed srgb), but also the texture's pixel format (floating point textures are stored in linear space and everything else is in gamma space).

willeastcott: Maybe we should add `TEXTURETYPE_RGB` and `TEXTURETYPE_RGBA` then. :)

slimbuck: I don't know of a need to differentiate between those two...

willeastcott: That's better than `GENERAL`.... 😄

src/graphics/texture.js
```diff
@@ -48,7 +48,9 @@ Object.assign(pc, function () {
      * @param {boolean} [options.mipmaps] - When enabled try to generate or use mipmaps for this texture. Default is true
      * @param {boolean} [options.cubemap] - Specifies whether the texture is to be a cubemap. Defaults to false.
      * @param {boolean} [options.volume] - Specifies whether the texture is to be a 3D volume (WebGL2 only). Defaults to false.
+     * @param {number} [options.type] - Specifies the image type, see {@link pc.TEXTURETYPE_DEFAULT}
      * @param {boolean} [options.rgbm] - Specifies whether the texture contains RGBM-encoded HDR data. Defaults to false.
+     * @param {boolean} [options.rgbe] - Specifies whther the texture contains RGBE-encoded HDR data. Defaults to false.
```
willeastcott, line 53: Doesn't specifying the `type` property make specifying `rgbm` or `rgbe` redundant?

src/graphics/texture.js
```diff
@@ -133,8 +134,16 @@ Object.assign(pc, function () {
...
-            this.swizzleGGGR = (options.swizzleGGGR !== undefined) ? options.swizzleGGGR : this.swizzleGGGR;
+
+            if (options.type !== undefined) {
+                this.type = options.type;
+            } else if (options.rgbm !== undefined) {
+                this.type = !!options.rgbm ? pc.TEXTURETYPE_RGBM : pc.TEXTURETYPE_DEFAULT;
+            } else if (options.rgbe !== undefined) {
+                this.type = !!options.rgbe ? pc.TEXTURETYPE_RGBE : pc.TEXTURETYPE_DEFAULT;
+            } else if (options.swizzleGGGR !== undefined) {
+                this.type = !!options.swizzleGGGR ? pc.TEXTURETYPE_SWIZZLEGGGR : pc.TEXTURETYPE_DEFAULT;
```
willeastcott, line 145: OK, so from this code, I can see that `rgbm`, `rgbe` and `swizzleGGGR` are there for backwards compatibility. So it seems kind of weird to _add_ `rgbe` to the API at this point. If anything, I would remove them from the docs and print a deprecation warning if we want people to migrate to use the `type` property.

src/graphics/texture.js
slimbuck: Yes rgbe and swizzleGGGR flags are legacy.

src/resources/parser/texture/legacy-dds.js
```diff
@@ -0,0 +1,159 @@
...
+            texture.upload();
+
+            return texture;
+        }
+    });
+
+    return {
+        LegacyDdsParser: LegacyDdsParser
+    };
+}());
```
willeastcott, line 159: Can we not unify our DDS code. Seems weird we have two loaders.

src/graphics/texture.js
slimbuck: Do we remove the docs for the deprecated flags or tag them as @deprecated?

src/resources/parser/texture/legacy-dds.js
slimbuck: Yes I plan to do this in a followup PR.

src/graphics/texture.js
willeastcott: I normally just remove them. I'm not sure the `deprecated` tag does anything to the output docs HTML.

src/graphics/texture.js
slimbuck: rgbm and swizzleGGGR has been deprecated.

src/graphics/texture.js
slimbuck: done

PR #2130, opened 2020-06-04
Anim component updates

ellt92:
- Fix for the `AnimComponent#stateGraphAsset` property
- Fix for the `AnimController#playable` property
- Dirty flag any materials that are animated
- Add the `AnimComponent#baseLayer` property as many workflows will only make use of a single layer. This stops users having to search for that layer using `AnimComponent#findAnimationLayer`.


src/framework/components/anim/component.js
```diff
@@ -344,6 +345,21 @@ Object.assign(pc, function () {
...
             }
+        },
+        /**
+         * @private
+         * @name pc.AnimComponent#baseLayer
+         * @type {pc.AnimComponentLayer}
+         * @readonly
+         * @description Returns the base layer of the state graph
+         */
+        baseLayer: {
```
mvaligursky, line 356: good idea!

src/framework/components/anim/binder.js
```diff
@@ -215,7 +215,13 @@ Object.assign(pc, function () {
                     propertyComponent[entityPropertySetterFunctionName](this._getProperty(propertyComponent, [entityProperty]));
                 };
                 return new pc.AnimTarget(entityPropertySetter.bind(this), animDataType, animDataComponents);
+            } else if (propertyHierarchy.indexOf('material') !== -1) {
+                return new pc.AnimTarget(function (values) {
+                    setter(values);
+                    propertyComponent.material.dirty = true;
```
mvaligursky, line 221: I think it'd be safer to call material.update();

src/framework/components/anim/controller.js
```diff
@@ -238,8 +238,8 @@ Object.assign(pc, function () {
             get: function () {
                 var playable = true;
                 var i;
-                for (i = 0; i < this._states.length; i++) {
-                    if (!this._states[i].playable) {
+                for (i = 0; i < this.states.length; i++) {
+                    if (!this._states[this.states[i]].playable) {
```
willeastcott, line 242: This does look a bit odd. Are `this.states` really ids or names or something?

ellt92: Yeah the controller exposes a list of available state names to the component but they're stored internally as AnimState objects. I could loop over `this._states` keys instead here if it makes things clearer.

ellt92: Using `this._stateNames` here instead.

PR #2131, opened 2020-06-04
Add asset options object, fix gltf texture loading

slimbuck:
glTF files containing urls to external resources must have crossOrigin specified. This PR adds an options object to pc.Asset for purposes of specifying load time options to the asset handler.

(In future the options object can also be used for specifying other load-specific settings to the resource handlers too).

### DIFF
--- a/src/asset/asset.js
+++ b/src/asset/asset.js
@@ -61,6 +61,7 @@ Object.assign(pc, function () {
      * var asset = new pc.Asset("a texture", "texture", {
      *     url: "http://example.com/my/assets/here/texture.png"
      * });
+     * @param {object} [options] - a JSON object containing load-time options specific to the asset.
      * @property {string} name The name of the asset
      * @property {number} id The asset id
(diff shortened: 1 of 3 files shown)


src/asset/asset.js
```diff
@@ -61,6 +61,7 @@ Object.assign(pc, function () {
      * var asset = new pc.Asset("a texture", "texture", {
      *     url: "http://example.com/my/assets/here/texture.png"
      * });
+     * @param {object} [options] - a JSON object containing load-time options specific to the asset.
```
willeastcott, line 64: You don't document the options. 😄

slimbuck: Yeah that's an interesting one. The options structure will depend on the asset type. For textures so far it's one crossOrigin member. Do you have any ideas how we go about documenting that?

willeastcott: List the options and in brackets put `(only applies to assets of type 'texture')`?

PR #2160, opened 2020-06-12
Texture based skinning precision issue on iOS

### LINKED ISSUE #1267: Animation flickering/trembling on mobile
I’m having an issue on mobile devices where the animation flickers, when running on a PC the animation is pretty smooth but when testing the same setup on iPad or Samsung S8 there is a noticeable flicker happening.

There’s no code added, only the model and an “idle” clip on it.

I assume the problem lies with the way the model / animation was prepared / exported, the generated model .json file's size is: 32.2 MB.

Here’s the link to the editor:
https://playcanvas.com/editor/scene/625770

And a live build:
https://playcanv.as/p/GCxiY7lI/

WebGL Report:
image

image

1. Open the live build link on mobile device and on a PC.
2. Notice the issue with the animation flickering/trembling on mobile device compared to it being smooth on PC.

mvaligursky:
forcing vertex shader uniforms related to texture based skinning to high precision, as it seems to default to mid precision giving us shaky animations on iOS

Fixes #1267

it seems https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices confirms this, as it states these are default precisions in vertex shader:

precision highp float;
precision highp int;
precision lowp sampler2D;
precision lowp samplerCube;

### DIFF
--- a/src/graphics/program-lib/chunks/skinTex.vert
+++ b/src/graphics/program-lib/chunks/skinTex.vert
@@ -1,8 +1,8 @@
-attribute vec4 vertex_boneWeights;
-attribute vec4 vertex_boneIndices;
+attribute highp vec4 vertex_boneWeights;
+attribute highp vec4 vertex_boneIndices;
 
-uniform sampler2D texture_poseMap;
-uniform vec2 texture_poseMapSize;
+uniform highp sampler2D texture_poseMap;
+uniform highp  vec2 texture_poseMapSize;
 
 mat4 getBoneMatrix(const in float i)
 {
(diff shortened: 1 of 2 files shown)


src/graphics/program-lib/chunks/skinTex.vert
```diff
@@ -1,8 +1,8 @@
-attribute vec4 vertex_boneWeights;
-attribute vec4 vertex_boneIndices;
+attribute highp vec4 vertex_boneWeights;
+attribute highp vec4 vertex_boneIndices;
```
willeastcott, line 2: What happens on devices that don't support hiphp?

mvaligursky: we only use those chunks when float textures are support, which should mean highp is supported.

willeastcott: Isn't it just the `texture_poseMap` that needs to be `highp`?

PR #2161, opened 2020-06-12
Update the GraphNode findbytag parameter types so it can take one or more strings or string arrays

TheJonRobinson:
`GraphNode`'s `findbytag` typings only specify a string parameter, but in fact takes a variable argument list of `string` or `string[]` parameters. This updates the JSDoc so the docs and generated Typescript typings reflect this.


willeastcott:
Thanks! 😄

PR #2162, opened 2020-06-12
Simplified computation of bone texture size

mvaligursky:
- smaller code
- handles more than 1024 bones
- tested by looping over 1024 bones and comparing results of both, it is equivalent.


willeastcott:
Nice! I :heart: PRs like this!

PR #2163, opened 2020-06-12
Minor optimizations for engine startup time

willeastcott:
This PR includes some minor optimizations affecting engine startup time:

* Lazily creates materials in `pc.SpriteComponentSystem`
* Only create `pc.ObjectPool`s in the `pc.RigidBodyComponentSystem` if ammo.js is loaded
* Remove redundant sets of `this.app` in subclasses of `pc.ComponentSystem` (since it's set in the superclass)
* Reduce the number of calls to `gl.getExtension`


src/graphics/device.js
```diff
@@ -732,7 +732,7 @@ Object.assign(pc, function () {
                     gl.bindVertexArray = ext.bindVertexArrayOES.bind(ext);
                 }
                 this.extColorBufferFloat = null;
-                this.extDisjointTimerQuery = null;
+                this.extDisjointTimerQuery = getExtension('EXT_disjoint_timer_query');
```
willeastcott, line 735: Hey @slimbuck - am I right in saying that GPU timings in the MiniStats didn't work with WebGL 1 before the above change?

mvaligursky: I know it didn't work on iOS

willeastcott: That's because `EXT_disjoint_timer_query` isn't available on iOS's WebGL 1.0 implementation. Not because WebGL 1.0 doesn't provide `EXT_disjoint_timer_query`.

slimbuck: Actually, when I wrote this Firefox was publishing its _webgl2_ disjoint timer under `EXT_disjoint_timer_query` string. So with this code both Chrome and Firefox worked. I have just tested this again and it looks like Firefox has removed the query timer entirely, sobface. So the change you've made here is legit.

slimbuck: Due to this weird setup MiniStats was only written to support `EXT_disjoint_timer_query_webgl2`. For now, it's probably safest to leave webgl1 disjoint timer NULL.

willeastcott: OK, I've restored previous behavior. But it does look kind of weird to me since `EXT_disjoint_timer_query_webgl2` is a superset of `EXT_disjoint_timer_query` I believe. And FYI, my Firefox (v77) on Win10 exposes `EXT_disjoint_timer_query`.

slimbuck: Would you mind mentioning the firefox behaviour in a comment (as I should have done originally)?

willeastcott: Done. 😄

PR #2164, opened 2020-06-13
Deprecate pc.string.format and pc.log API

willeastcott:
This PR formerly deprecates `pc.string.format`. JavaScript has a perfectly good and succinct string concatenation operator '`+`'. This operator is _far_ more performant that calling `pc.string.format` which internally creates a RegEx _for each argument_. It's incredibly inefficient.

It also moves `pc.log` API to the backwards compatibility module. 👋 


src/backwards-compatibility.js
```diff
@@ -135,6 +135,56 @@ Object.assign(pc.input, {
     TouchEvent: pc.TouchEvent
 });
 
+pc.log = {
+    write: function (text) {
+        console.log(text);
```
slimbuck, line 140: should these also print out deprecated warnings?

willeastcott: I considered it but thought it was weird that a console logging API, logging warning too. But yeah, TBH, it probably should.

willeastcott: OK, I've added deprecation warnings for `pc.log.*`.

PR #2172, opened 2020-06-18
Anim binder material texture support

ellt92:
Enables support for `pc.AnimCurve` paths that point to material texture maps. The binder will look for a loaded asset using the supplied id and assign the resource to the particular material map.


src/anim/anim.js
```diff
@@ -717,6 +717,34 @@ Object.assign(pc, function () {
...
+                if (!object ||
+                    !object.model ||
+                    !object.model.model ||
+                    !object.model.model.morphInstances) {
+                    return null;
+                }
+                var meshInstances = object.model.meshInstances;
+                var meshInstance;
+                for (var i = 0; i < meshInstances.length; ++i) {
+                    if (meshInstances[i].node.name === node.name) {
```
mvaligursky, line 735: is this name compare needed? We take meshInstance from the node, so I assume they all point to the same node?

ellt92: We get the mesh instance from the nodes ancestor model. But that model may contain multiple mesh instances so this finds the one we're looking for.

src/anim/anim.js
```diff
@@ -717,6 +717,36 @@ Object.assign(pc, function () {
...
+                        meshInstance = meshInstances[i];
+                        break;
+                    }
+                }
+                if (!meshInstance) {
+                    return null;
+                }
+                var func = function (value) {
+                    var textureAsset = pc.app.assets.get(value[0]);
```
willeastcott, line 743: Don't use `pc.app`. If you don't have the application at hand, you can use `Application.getApplication()`. Remember, `pc.` is now banned in internal engine code. 😄

src/anim/anim.js
```diff
@@ -717,6 +717,36 @@ Object.assign(pc, function () {
                     }
                 };
                 return new pc.AnimTarget(func, 'vector', morphInstance.morph._targets.length);
+            },
+            'materialTexture': function (node, textureName) {
+                var object = node;
+                while (object && object.constructor !== pc.Entity) {
```
willeastcott, line 723: `pc.Entity` -> `Entity`

PR #2178, opened 2020-06-19
ParticleSystem rendering in UI space

mvaligursky:
ParticleSystem component can be used on Entity inside hierarchy with parent being Screen component, allowing particles getting rendered integrated in UI layer (under button, on top of button ..)

instructions:
- add entity with ParticleSystem into hierarchy with root Screen. Note that it cannot be added as a child of screen, and its parent needs to be one of the Element components (Image, Group ..)
- tick on 'ScreenSpace' option on the ParticleSystem
- make sure the Layer is set to UI (same layer as the Screen) - otherwise sorting does not work

engine example: engine/examples/#user-interface/button-particle.html

### DIFF
--- a/src/scene/particle-system/particle-emitter.js
+++ b/src/scene/particle-system/particle-emitter.js
@@ -180,6 +180,7 @@ Object.assign(pc, function () {
         setProperty("initialVelocity", 1);
         setProperty("wrap", false);
         setProperty("localSpace", false);
+        setProperty("screenSpace", false);
         setProperty("wrapBounds", null);
         setProperty("colorMap", ParticleEmitter.DEFAULT_PARAM_TEXTURE);
         setProperty("normalMap", null);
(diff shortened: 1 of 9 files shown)


src/scene/particle-system/particle-emitter.js
```diff
@@ -820,6 +824,10 @@ Object.assign(pc, function () {
                     fog: (this.emitter.scene && !this.emitter.noFog) ? this.emitter.scene.fog : "none",
                     wrap: this.emitter.wrap && this.emitter.wrapBounds,
                     localSpace: this.emitter.localSpace,
+
+                    // in Editor, screen space particles (children of 2D Screen) are still rendered in 3d space
+                    screenSpace: inTools ? false : this.emitter.screenSpace,
```
vkalpias, line 829: I think this should be set by the Editor (like how the Editor makes the screen a 3D screen and not the engine). We made an effort in the past to get rid of code like this which is basically editor code sprinkled in the engine.

mvaligursky: done in the linked PR, closing this one due to complexity merging in rollup changes.

PR #2191, opened 2020-06-23
Restore an optimized version of pc.string.format

willeastcott:
Restores `pc.string.format` to API (for now). But I've taken the opportunity to optimize it and condense the code so it's not quite as odious as before. Implementation is now:
```
    format: function (s) {
        for (var i = 1; i < arguments.length; i++) {
            s = s.replace('{' + (i - 1) + '}', arguments[i]);
        }
        return s;
    },
```

This is approximately twice as fast as the previous RegExp-based implementation and the code is obviously smaller.

I've also removed the closure in the string.js module which is no longer required under Rollup.

### DIFF
--- a/src/core/string.js
+++ b/src/core/string.js
@@ -118,16 +118,16 @@ var string = {
      */
     ASCII_LETTERS: ASCII_LETTERS,
 
-    /** 
-     * @function    
-     * @name pc.string.format   
-     * @description Return a string with {n} replaced with the n-th argument.   
-     * @param {string} s - The string to format.    
-     * @param {object} [arguments] - All other arguments are substituted into the string.   
-     * @returns {string} The formatted string.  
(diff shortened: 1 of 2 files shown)


Maksims:
Use of `arguments` is known for preventing functions to be optimised by V8 in Chrome. This is one of the reasons why we do `arg1, arg2, arg3 ...` as arguments in event emitters.

willeastcott:
Interesting. But I can't make assumptions about the maximum number of args here unfortunately. This new implementation is 2x faster, which is something. But it's still literally 600 times slower than just doing a string concat. 

I just made a jsperf to investigate:

https://jsperf.com/pc-string-format

image

Looks like not using the arguments array would give another 20% boost. But yeah, I think I can live with the hit to have a compact, generalized function.

PR #2196, opened 2020-06-24
Texture Based Mesh Morphing

mvaligursky:
Implementation of mesh morphing based on storing morph targets in float textures, as opposed to storing them in vertex attributes.

final

**Implementation details:**
- Supported platform requirements:
    - webgl 1 or 2
    - float / half float texture support
    -  float / half float texture blending support
    - float / half float render target support
    - highp fragment shader precision
    - at least 2 vertex shader textures
- this allows "unlimited" number of active morph targets per mesh (up to 4096x4096 texture size to store morph targets)
- texture based implementation is the preferred option if supported by hardware (half-float, then float precision), otherwise attribute based morphing is used.

**Future work:**
- allow callback during Morph creation by glb-parser and let user specify which render technique should be used (attribute vs texture based) and precision (float / half float)
- create cpu based fallback if more than 8 active targets are needed and texture based morphing is not supported

**Performance:**
- memory: texture based morphing is storing morph targets as partially sparse (so less memory here), but has additional render target texture for position / normal allocated (so additional memory here)
- performance (2019 Mac and Google Pixel 3A, GPU time in ms, 100 morph high res meshes, no shadows)
Screen Shot 2020-06-05 at 2 29 43 PM
- texture based morph target blending takes place only one time per frame and so when rendering the same mesh multiple times (to shadow maps ..), performance of texture based morphing is likely to be even better than current test

**Engine example (when released):**
* show 10 morph targets modifying position and normal of a single mesh {@link http://playcanvas.github.io/#graphics/mesh-morph-many.html}

Supersedes #2123 which was closed due to complex merging required, related to other large changes to engine.


src/graphics/device.js
```diff
@@ -776,6 +808,7 @@ Object.assign(GraphicsDevice.prototype, {
 
         this.extDebugRendererInfo = getExtension('WEBGL_debug_renderer_info');
         this.extTextureFloatLinear = getExtension("OES_texture_float_linear");
+        this.extFloatBlend = getExtension("EXT_float_blend");
```
willeastcott, line 811: Is this required for texture based morphing? It's not available on iOS.

src/scene/morph-instance.js
```diff
@@ -65,13 +147,160 @@ Object.assign(MorphInstance.prototype, {
...
+        var submitBatch = function (usedCount, blending) {
+
+            // factors
+            this.morphFactor.setValue(this._shaderMorphWeights);
+
+            // alpha blending - first pass gets none, following passes are additive
+            device.setBlending(blending);
+            if (blending) {
+                device.setBlendFunction(pc.BLENDMODE_ONE, pc.BLENDMODE_ONE);
+                device.setBlendEquation(pc.BLENDEQUATION_ADD);
```
willeastcott, line 206: You can't use `pc.` in the internal engine code any more.

src/scene/morph-instance.js
```diff
@@ -65,13 +147,160 @@ Object.assign(MorphInstance.prototype, {
...
+            }
+
+            // render quad with shader for required number of textures
+            var shader = this._getShader(usedCount);
+            pc.drawQuadWithShader(device, renderTarget, shader, undefined, undefined, blending);
```
willeastcott, line 211: No `pc.`

src/scene/materials/standard-material-options-builder.js
```diff
@@ -54,6 +54,7 @@ StandardMaterialOptionsBuilder.prototype._updateSharedOptions = function (option
     options.useInstancing = objDefs && (objDefs & SHADERDEF_INSTANCING) !== 0;
     options.useMorphPosition = objDefs && (objDefs & SHADERDEF_MORPH_POSITION) !== 0;
     options.useMorphNormal = objDefs && (objDefs & SHADERDEF_MORPH_NORMAL) !== 0;
+    options.useMorphTextureBased = objDefs && (objDefs & pc.SHADERDEF_MORPH_TEXTURE_BASED) !== 0;
```
willeastcott, line 57: No `pc.`

src/scene/morph-instance.js
```diff
@@ -6,27 +15,74 @@
...
+        this.maxSubmitCount = this.device.maxTextures;
+
+        // array for max number of weights
+        this._shaderMorphWeights = new Float32Array(this.maxSubmitCount);
+
+        // create render targets to morph targets into
+        var createRT = function (name, textureVar) {
+
+            // render to appropriate, RGBA formats, we cannot render to RGB float / half float format in WEbGL
+            var format = morph._renderTextureFormat === pc.Morph.FORMAT_FLOAT ? pc.PIXELFORMAT_RGBA32F : pc.PIXELFORMAT_RGBA16F;
```
willeastcott, line 45: No `pc.`

src/scene/morph-instance.js
```diff
@@ -6,27 +15,74 @@
...
+            this[textureVar] = morph._createTexture(name, format);
+            return new pc.RenderTarget({
```
willeastcott, line 47: No `pc.`

src/scene/morph-instance.js
```diff
@@ -65,13 +147,160 @@ Object.assign(MorphInstance.prototype, {
...
+
+    // creates complete shader for texture based morphing
+    _getShader: function (count) {
+
+        var shader = this.shaderCache[count];
+
+        // if shader is not in cache, generate one
+        if (!shader) {
+            var fs = this._getFragmentShader(count);
+            shader = pc.shaderChunks.createShaderFromCode(this.device, textureMorphVertexShader, fs, "textureMorph" + count);
```
willeastcott, line 185: No `pc.`

src/scene/morph.js
```diff
@@ -31,19 +69,162 @@ Object.defineProperties(Morph.prototype, {
...
+            return false;
+        }
+
+        this.morphTextureWidth = morphTextureWidth;
+        this.morphTextureHeight = morphTextureHeight;
+
+        // texture format based vars
+        var halfFloat = false;
+        var numComponents = 3;  // RGB32 is used
+        var float2Half = pc.math.float2Half;
```
willeastcott, line 177: No `pc.`

src/graphics/device.js
mvaligursky: float blending is only used when float render targets are available, so not used on iOS - there a half float format is used instead.

src/scene/morph-instance.js
mvaligursky: I'll remove it, but I've just tested it and it works.

willeastcott: Maybe 'can't' should be 'shouldn't'. :smile:

mvaligursky: done

src/scene/materials/standard-material-options-builder.js
mvaligursky: fixing

src/scene/morph-instance.js
mvaligursky: fixed

src/scene/morph-instance.js
mvaligursky: fixed

src/scene/morph-instance.js
mvaligursky: fixed

src/scene/morph.js
mvaligursky: fixing

PR #2206, opened 2020-06-25
Draft Shader Graph script interface to use with Node Material subclass

raytranuk:
This PR includes: A new interface/namespace: `pc.shadergraph` - which enables `NodeMaterials` to be created and configured in script.

Simple example:
   ```
 //start building shader graph
    pc.shadergraph.start();

    //create a simple pixel shader graph - just a 2D texture sample
    var texSampleNode = pc.shadergraph.textureSample2D('diffSamp',diffuseMap, pc.shadergraph.uv0);

    //hook up pixel shader output
    pc.shadergraph.connectFragOut(texSampleNode);

    //create a custom vertex offset shader graph node
    var wobbleVSNode = pc.shadergraph.customNode('wobbleVS', document.getElementById("wobbleVS").textContent);

    //hook up custom node inputs
    pc.shadergraph.connectCustom(wobbleVSNode, 'wp', pc.shadergraph.worldPosVS);
    pc.shadergraph.connectCustom(wobbleVSNode, 'wn', pc.shadergraph.worldNormVS);
    pc.shadergraph.connectCustom(wobbleVSNode, 't', pc.shadergraph.param('float', 'uTime', 0.0));

    //hook up vertex offset output
    pc.shadergraph.connectVertexOffset(wobbleVSNode);

    //end graph and assign built graph to material
    var material = pc.shadergraph.end();
```

pc.shadergraph API
current:
- pc.shadergraph.start() - call this to start building a shader graph
- pc.shadergraph.textureSample2D(name, texture, uvNodePort) - create and return a textureSample2D node which will sample texture with uv from connected uvNodePort  
- pc.shadergraph.uv0 - return a node that outputs the value of texture coordinate channel 0
- pc.shadergraph.connectFragOut(nodePort) - connect nodePort of type vec4 to the pixel shader output
- pc.shadergraph.customNode(name, code) - create and return a custom node of name name with code supplied in code string 
- pc.shadergraph.connectCustom(node, portName, nodePort) - connect a nodePort to a custom node port of name portName
- pc.shadergraph.worldPosVS - return a node that ouputs the value of the vertex postion in world space
- pc.shadergraph.worldNormVS - return a node that ouputs the value of the vertex normal in world space
- pc.shadergraph.param(type, name, value) - create a graph input parameter of type with name and value  
- pc.shadergraph.connectVertexOffset(nodePort) - connect nodePort of type vec3 to the vertex offset output
- pc.shadergraph.end() - call this to finish building a shader graph and return a material which uses the graph

MVP no shader code (proposed):
- pc.shadergraph.splitVec(v) - create a node that takes a vector input and outputs individual components
- pc.shadergraph.joinVec(x,y,z,w) - create a node that takes individual components and outputs a vector
- pc.shadergraph.add(a, b) - create a node that performs addition operation on parameters a and b - parameters must be of the same type 
- pc.shadergraph.sub(a, b) - create a node that performs subtraction operation on parameters a and b - parameters must be of the same type 
- pc.shadergraph.mul(a, b) - create a node that performs multiplication operation on parameters a and b - parameters must be of the same type 
- pc.shadergraph.div(a, b) - create a node that performs division operation on parameters a and b - parameters must be of the same type
- pc.shadergraph.dot(a, b) - create a node that performs dot product operation on parameters a and b - parameters must be a vector type 
- pc.shadergraph.cross(a, b) - create a node that performs a cross product operation on parameters a and b - parameters must be a vector type 
- pc.shadergraph.if(c, a, b) - create a node that outputs a if c is true and b if c is false - parameter c must be of boolean type and parameters a and b must be of the same type
- pc.compare(a, op, b) - create a node that compares the value of a and the value of b using an operation defined in op, and returns true or false. Valid operations are: '==', '!=', '>', '<', '>=', '<=', '&&', '||'

### CODE REVIEW

File: src/graphics/device.js [hunk 1]
```diff
@@ -2092,6 +2092,10 @@ Object.assign(pc, function () {
                             this.instancedAttribs[locationId] = false;
                         }
                     }
+                    else
+                    {
+                        gl.disableVertexAttribArray(attribute.locationId); //this allows for a single variant for NodeMaterials
+                    }
```
mvaligursky, line 2098: there's different (similar) line like this in master (used by morphing), hopefully that would work for you

raytranuk: Thanks - just merged master and will test with your version.

raytranuk: It works

Maksims:
I've recently started some R&D in terms of shader graphs, with agnostic to engine approach. And thought of: what would make a good implementation. It will be definitely interesting to know **what major factors are driving design and implementations**.

Here are my IMHOs:
1. **Simple to store.** Simple JSON, only to store necessary for shader building.
2. **Fast to parse** - simplicity in data as well as its layout, is important to keep big O as low as possible with edge cases.
3. **Memory friendly.** Generation of shader code from graph, should not lead to major garbage collection stalls. There are two paths can be taken here: stateless and object based approaches. In stateless code is generated directly from data using temporary structures. And object based - would be generating an objects out of data, and updating them in case of need to regenerate shader.
4. **Simple to work with.** Tools - are major part of shader graphs, and depending on underlying design, tools very much depend a lot on this. Also, there should be a workflow with no tools available as well. This includes machine generated graphs.
5. **Cross-engine compatibility** - this one, is my personal IMHO. Just like GLTF came to WebGL industry, I believe same way should happen with shader graphs. More and more engines are getting developed, and WebGPU is coming closer. Providing unified way for technical artists to express them self in Web 3D industry - is a major win for everyone.
6. **Separate engine builder data from graph data.** There are a lot of meta information for tools, and engine implementations. While graph data stays the same. By separating graph data from bits that are required to build shader from graph, we provide flexibility here. Node types and their connections - is graph data. But e.g. shader code - is engine specific implementation parts.

raytranuk:
@Maksims - thanks for your thoughts on design for a system like this.

This PR is still an early implementation and focuses on a script interface for coders. Making it convenient to build shader graphs in code - including writing custom shader nodes - was a discussed design consideration. I'd also expect that the same interface would be usable by a visual graphing tool - probably without the need for the `shadergraph` convenience functionality which is mainly targeted at making scripts more concise.

> * **Simple to store.** Simple JSON, only to store necessary for shader building.

Agreed - node properties should be simple to serialize in and out - something that will help is to use ids or indices rather than object references. I can change this.

> * **Fast to parse** - simplicity in data as well as its layout, is important to keep big O as low as possible with edge cases.

Directed acyclic graphs tend to be straight forward to traverse and linear time? What edge cases do you envisage?

> * **Memory friendly.** Generation of shader code from graph, should not lead to major garbage collection stalls. There are two paths can be taken here: stateless and object based approaches. In stateless code is generated directly from data using temporary structures. And object based - would be generating an objects out of data, and updating them in case of need to regenerate shader.

I think I get what you mean here - currently there are temporary or transient values that are stored in the node objects when the graph is traversed to generate code - instead, temporary structures which reference (or index) the nodes should be used. I will implement this.

> * **Simple to work with.** Tools - are major part of shader graphs, and depending on underlying design, tools very much depend a lot on this. Also, there should be a workflow with no tools available as well. This includes machine generated graphs.

I agree good tools are very important - and ultimately, the expectation is that the majority of shader graphs will be created using visual graphing tools rather than code - only the low-level built-in nodes will be coded and the interface with engine rendering systems will be code. I also expect some higher level custom nodes to be written with code in some projects - and perhaps even before a visual shader editor is available? Machine generated graphs - I assume this means descriptions of shaders/materials in a form other graphs - perhaps like a list of options from the StandardMaterial?

> * **Cross-engine compatibility** - this one, is my personal IMHO. Just like GLTF came to WebGL industry, I believe same way should happen with shader graphs. More and more engines are getting developed, and WebGPU is coming closer. Providing unified way for technical artists to express them self in Web 3D industry - is a major win for everyone.

Did you know there is a proposal for a GLTF material extension for material graphs?: https://docs.google.com/document/d/1Y6JFE2FV164IFDe7_cYhp2gzhSapB76fUNPgmsI6DDY/edit#heading=h.ptayfzbvu0py - if it extends how GLTF material attributes are generated, it is very likely to be supported. I'm actually currently working on new shader nodes to perform rendering functionality (lighting/shadows/reflections) and I'm working with the GLTF material specification (including the clear coat extension).

> * **Separate engine builder data from graph data.** There are a lot of meta information for tools, and engine implementations. While graph data stays the same. By separating graph data from bits that are required to build shader from graph, we provide flexibility here. Node types and their connections - is graph data. But e.g. shader code - is engine specific implementation parts.

Agreed - separation of meta data for the visualization in the tools would be good.

Some other design considerations recently discussed - suggested by @mvaligursky - was to make the built-in nodes available as assets in a package or bundle that can be optionally importable into projects - and also new versions can be released without breaking old versions.

Thanks again for the feedback - I'll make some of the changes discussed here.

src/resources/material.js
```diff
@@ -36,8 +19,9 @@ function MaterialHandler(app) {
     this._device = app.graphicsDevice;
 
     this._placeholderTextures = null;
-
-    this._parser = new JsonStandardMaterialParser();
+   
+    this._binder = null;
```
slimbuck, line 23: Assuming a single project is free to mix and match material systems, you will need to instantiate all parsers and binders here (or perhaps lazily) because this single instance of MaterialHandler loads all materials.

src/resources/material.js
```diff
@@ -67,7 +51,25 @@ Object.assign(MaterialHandler.prototype, {
         });
     },
 
+    _assignParserAndBinderBasedOnSubClass: function (data) {
+        if (data.shaderGraph) //maybe there is a better way to tell if node material path should be taken?
+        {
```
slimbuck, line 56: bracket should be end of previous line :)

src/resources/material.js
```diff
@@ -125,196 +93,17 @@ Object.assign(MaterialHandler.prototype, {
         asset.data.name = asset.name;
         asset.resource.name = asset.name;
 
-        this._bindAndAssignAssets(asset, assets);
+        if (!this._parser || !this._binder) 
+        {
```
slimbuck, line 97: bracket!

src/resources/material.js
raytranuk: thanks @slimbuck - latest commit lazily instantiates in arrays of subclass binders/parsers

src/resources/material.js
raytranuk: I will lint it

src/resources/material.js
raytranuk: I shall lint it

src/resources/material.js
raytranuk: linted

src/resources/material.js
raytranuk: linted

raytranuk:
This is the old Draft PR description (I will likely refer to this):
> NB Now working (not robustly yet) in editor:
> image
> 
> NB. Now upgraded to ES6 format
> 
> Draft PR for WIP but basically functional system - mainly comprised of an object/namespace: `shadergraph` and 2 object classes: `ShaderGraphNode` and `NodeMaterial`
> 
> The `shadergraph` object/namespace makes it easier to create and configure custom and "built-in" shader graph nodes - e.g in the engine example (https://github.com/playcanvas/engine/blob/node-material/examples/graphics/shader-node.html) - this line:
> 
> ```
> var texSampleNode = pc.shadergraph.textureSample2D(diffuseMap, pc.shadergraph.uv0);
> ```
> 
> actually creates a sub graph consisting of 3 nodes: a texture parameter node, a built in constant node and a texturesample node - here is how this would be done without the `shadergraph` object/namespace:
> 
> ```
> var texSampleNode = new pc.ShaderGraphNode('void texSample(in sampler2D texture, in vec2 uv, out vec4 rgba, out vec3 color, out float alpha) {\n vec4 sample=texture2D(texture, uv);\n rgba=sample;\n color=sample.rgb;\n alpha=sample.a;\n}');
> texSampleNode.texture = new pc.ShaderGraphNode('sampler2D', 'uDiffuseMap', diffuseMap).uDiffuseMap;
> texSampleNode.uv = new pc.ShaderGraphNode('vec2 uv0() { return vUv0; }').ret;
> ```
> 
> `ShaderGraphNode` class objects links together to form a shader graph - there are 3 categories of shader graph node:
> 
> 1. Parameter - these have values that can be set dynamically set without regenerating/recompiling shader code (shader uniforms) - initialized using `ShaderGraphNode.setParam()`
> 2. Operator - these perform operations on input values and produce one or more output value - initialized using `ShaderGraphNode.setFunctionString()`
> 3. Root - this is the output node of the graph (or sub graph) - it has inputs used as the shader outputs (e.g gl_FragColor) or outputs of a sub graph.
> 
> `NodeMaterial` class inherits from `Material`, is initialized using a Root `ShaderGraphNode`, and injects the generated shader code and parameter list from the shader graph into a material - which can be assigned to mesh instances as usual.
> 
> NB. I'll expand on this PR description with more details as this system develops further - any early feedback welcome!
> 
> I confirm I have signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

raytranuk:
Closing this NodeMaterial PR - moving to new StandardNodeMaterial PR for MPV

PR #2208, opened 2020-06-25
Update sound component to current format

willeastcott:
This PR updates the sound component to use the current recommended structure for an engine component. The `data` block is now no longer used.

It also moves ModelComponent#setVisible to the deprecated.js file.

### DIFF
--- a/src/framework/components/sound/component.js
+++ b/src/framework/components/sound/component.js
@@ -50,15 +50,6 @@ function SoundComponent(system, entity) {
     this._slots = {};
 
     this._playingBeforeDisable = {};
-
-    this.on('set_slots', this.onSetSlots, this);
-    this.on('set_volume', this.onSetVolume, this);
-    this.on('set_pitch', this.onSetPitch, this);
-    this.on("set_refDistance", this.onSetRefDistance, this);
-    this.on("set_maxDistance", this.onSetMaxDistance, this);
-    this.on("set_rollOffFactor", this.onSetRollOffFactor, this);
-    this.on("set_distanceModel", this.onSetDistanceModel, this);
-    this.on("set_positional", this.onSetPositional, this);
 }
 SoundComponent.prototype = Object.create(Component.prototype);
 SoundComponent.prototype.constructor = SoundComponent;
--- a/src/framework/components/sound/component.js
+++ b/src/framework/components/sound/component.js
@@ -1,3 +1,5 @@
+import { DISTANCE_LINEAR } from '../../../audio/constants.js';
+
 import { Component } from '../component.js';
 
 import { SoundSlot } from './slot.js';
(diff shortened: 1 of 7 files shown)


src/framework/components/sound/system.js
```diff
@@ -56,50 +46,62 @@ Component._buildAccessors(SoundComponent.prototype, _schema);
...
+        var srcComponent = entity.sound;
+        var srcSlots = srcComponent.slots;
 
         // convert 'slots' back to
         // simple option objects
-        newData.slots = {};
-
-        for (key in oldData.slots) {
-            var oldSlot = oldData.slots[key];
-            if (oldSlot instanceof SoundSlot) {
```
willeastcott, line 81: Hey @vkalpias , I didn't see how a component being cloned could have a JSON data block instead of an actual `SoundSlot` so I removed that case. Is that OK?

src/framework/components/sound/component.js
```diff
@@ -13,160 +15,107 @@ import { SoundSlot } from './slot.js';
...
-            var slot = slots[key];
-            // change pitch of non-overlapping instances
-            if (!slot.overlap) {
-                var instances = slot.instances;
-                for (var i = 0, len = instances.length; i < len; i++) {
-                    instances[i].pitch = slot.pitch * newValue;
+            var slots = this._slots;
+            for (var key in slots) {
+                var slot = slots[key];
+                // change refDistance of non-overlapping instances
```
vkalpias, line 69: refDistance not relevant to this comment

src/framework/components/sound/system.js
```diff
@@ -56,50 +45,62 @@ Component._buildAccessors(SoundComponent.prototype, _schema);
 
 Object.assign(SoundComponentSystem.prototype, {
     initializeComponentData: function (component, data, properties) {
-        properties = ['volume', 'pitch', 'positional', 'refDistance', 'maxDistance', 'rollOffFactor', 'distanceModel', 'slots', 'enabled'];
-        ComponentSystem.prototype.initializeComponentData.call(this, component, data, properties);
+        properties = [
```
vkalpias, line 48: Is there a reason for changing the order of the properties? Usually the order matters in initialization not sure if this will break something...

src/framework/components/sound/system.js
vkalpias: that's OK I think

src/framework/components/sound/system.js
```diff
@@ -148,7 +149,9 @@ Object.defineProperty(SoundComponentSystem.prototype, 'volume', {
 Object.defineProperty(SoundComponentSystem.prototype, 'context', {
     get: function () {
         if (!hasAudioContext()) {
+            // #ifdef DEBUG
```
vkalpias, line 152: Wondering if we should leave that on always to know what's happening on devices that don't support this (where we usually don't use the debug engine)

willeastcott: Web Audio is supported absolutely everywhere except IE11. Personally, I don't think the engine should ever `console.log/warn/error` in a release build _except_ for the engine startup message.

src/framework/components/sound/component.js
willeastcott: Fixed.

src/framework/components/sound/system.js
willeastcott: Restored to original order.

PR #2210, opened 2020-06-26
Fixed WebXR session support check

yaustar:
Fixes https://forum.playcanvas.com/t/webxr-ar-camera-not-working/13775/11

### DIFF
--- a/src/xr/xr-manager.js
+++ b/src/xr/xr-manager.js
@@ -6,9 +6,7 @@ import { Quat } from '../math/quat.js';
 import { Vec3 } from '../math/vec3.js';
 import { Vec4 } from '../math/vec4.js';
 
-import { XRTYPE_INLINE } from './constants.js';
-import { XRTYPE_VR } from './constants.js';
-import { XRTYPE_AR } from './constants.js';
+import { XRTYPE_INLINE, XRTYPE_VR, XRTYPE_AR } from './constants.js';
 import { XrHitTest } from './xr-hit-test.js';
 import { XrInput } from './xr-input.js';
 import { XrLightEstimation } from './xr-light-estimation.js';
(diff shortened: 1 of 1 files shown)


src/xr/xr-manager.js
```diff
@@ -6,6 +6,8 @@ import { Quat } from '../math/quat.js';
 import { Vec3 } from '../math/vec3.js';
 import { Vec4 } from '../math/vec4.js';
 
+import { XRTYPE_INLINE } from './constants.js';
+import { XRTYPE_VR } from './constants.js';
```
willeastcott, line 10: Please use comma notation here:
```
import { XRTYPE_AR, XRTYPE_INLINE, XRTYPE_VR } from './constants.js';
```

src/xr/xr-manager.js
```diff
@@ -37,11 +39,12 @@ function XrManager(app) {
...
-        XRTYPE_AR: false,
-        XRTYPE_VR: false,
-        XRTYPE_INLINE: false
-    };
+    this._available = { };
+
+    // Add all the supported session types
+    this._available[XRTYPE_INLINE] = false;
+    this._available[XRTYPE_VR] = false;
+    this._available[XRTYPE_AR] = false;
```
willeastcott, line 47: What's this change for?

yaustar: The previous code:
```
    this._available = {
        XRTYPE_AR: false,
        XRTYPE_VR: false,
        XRTYPE_INLINE: false
    };
```
Had named the keys as 'XRTYPE_AR', 'XRTYPE_VR' and 'XRTYPE_INLINE' rather than the string values that are in the constants.js file.

This change uses the string values of the constants instead :)

willeastcott: Ah interesting. Fair enough.

PR #2213, opened 2020-06-27
AssemblyScript/WebAssembly for performance critical math code

### LINKED ISSUE #2201: Integrate WebAssembly support build system
Allow an easy way to write c++ code, build it into WebAssembly and integrate it into the Engine. This would allow us to execute some data heavy operations faster.

kungfooman:
Fixes #2201

This is pretty much the fundamentals to get started with AssemblyScript, so everyone can explore the possibilities and issues, like:

image


mvaligursky:
Thanks @kungfooman -  I'll have a closer look / play next week, much appreciated.

assembly/Quat.ts
```diff
@@ -0,0 +1,364 @@
...
+        ex *= halfToRad;
+        ey *= halfToRad;
+        ez *= halfToRad;
+
+        var sx = Mathf.sin(ex);
+        var cx = Mathf.cos(ex);
+        var sy = Mathf.sin(ey);
+        var cy = Mathf.cos(ey);
+        var sz = Mathf.sin(ez);
+        var cz = Mathf.cos(ez);
```
MaxGraey, line 190: You could speedup this via:
```ts
Mathf.sincos(ex);
var sx = Mathf.sincos_sin;
var cx = Mathf.sincos_cos;
Mathf.sincos(ey);
var sy = Mathf.sincos_sin;
var cy = Mathf.sincos_cos;
Mathf.sincos(ez);
var sz = Mathf.sincos_sin;
var cz = Mathf.sincos_cos;
```

assembly/Mat4.ts
```diff
@@ -0,0 +1,717 @@
...
+        ey *= pc_math.DEG_TO_RAD;
+        ez *= pc_math.DEG_TO_RAD;
+
+        // Solution taken from http://en.wikipedia.org/wiki/Euler_angles#Matrix_orientation
+        var s1 = Mathf.sin(-ex);
+        var c1 = Mathf.cos(-ex);
+        var s2 = Mathf.sin(-ey);
+        var c2 = Mathf.cos(-ey);
+        var s3 = Mathf.sin(-ez);
+        var c3 = Mathf.cos(-ez);
```
MaxGraey, line 413: same here

assembly/Mat4.ts
```diff
@@ -0,0 +1,717 @@
...
+    }
+
+    setFromAxisAngle(axis: Vec3, angle: f32): Mat4 {
+        angle *= pc_math.DEG_TO_RAD;
+
+        var x: f32 = axis.x;
+        var y: f32 = axis.y;
+        var z: f32 = axis.z;
+        var c: f32 = Mathf.cos(angle);
+        var s: f32 = Mathf.sin(angle);
```
MaxGraey, line 373: and here

Maksims:
Are there tests and benchmarks on this?
What are benefits for up to date platforms, and if there is any sacrifice for older platforms?

Also how does this affects profiling and debug-ability of code?

kungfooman:
@Maksims Thanks for the interest, I don't see any downsides for older platforms, because I simply add another custom build target (generating playcanvas.assemblyscript.js), and it is the explicit decision of the developer to use that target.

I figured that the Math api changed a bit in the meanwhile (I did the ES5 to AssemblyScript conversion like 16 months ago), so I am in the process of figuring out the differences and integrating the existing math unit tests.

I got the unit tests working with node so far, and it has one "serious" bug, which I am investigating:

```pc.Quat
getEulerAngles failed:  AssertionError: expected 116.5650405883789 to equal 180```

(Not sure yet where this difference comes from)

I will make a video about the testing/debugging process later with some devtools profiling for comparison

**Comment by MaxGraey** [comment]
It seems to me in AS it makes sense to make only a few functions that require complex calculations such as: `Mat4#getEulerAngles`, `Mat4#setFromEulerAngles`, `Quat#getAxisAngle`, `Quat#getEulerAngles`, `Quat#setFromEulerAngles`, `Quat#slerp`.

In these cases, there is complete gananty that seepup of math functions will overlap js <-> wasm interop overhead

File: package.json [hunk 4]
```diff
@@ -1,77 +1,82 @@
...
+    "serve": "http-server build -a localhost -p 51000",
+    "test": "karma start tests/karma.conf.js -- --single-run --release",
+    "test:watch": "karma start tests/karma.conf.js",
+    "test:debug": "karma start tests/karma.conf.js -- --single-run=false",
+    "test:tsd": "npm run tsd && tsc --pretty false build/playcanvas.d.ts",
+    "tsd": "jsdoc -c conf-tsd.json",
+    "uglify": "uglifyjs build/playcanvas.js --compress --mangle --output build/playcanvas.min.js",
+    "unzipbundle": "gzip -x tests/assets/bundle.gz",
+    "asbuild:untouched": "asc assembly/index.ts -b build/untouched.wasm -t build/untouched.wat --sourceMap --measure --debug",
+    "asbuild:optimized": "asc assembly/index.ts -b build/optimized.wasm -t build/optimized.wat --sourceMap --measure --optimize"
```
MaxGraey, line 76:
```suggestion
    "asbuild:optimized": "asc assembly/index.ts -b build/optimized.wasm -t build/optimized.wat --sourceMap --measure -O3 --noAssert"
```

examples/animation/test.js
```diff
@@ -0,0 +1,114 @@
...
+	navigator: {
+		userAgent: "node"
+	}
+};
+
+require("./Loader");
+//console.log("Loader", Loader)
+
+async function load_wasm() {
+	var wasmFile = fs.readFileSync("../../build/untouched.wasm");
```
MaxGraey, line 62: I guess it's just temporary for debugging? Otherwise it could be
```suggestion
	var wasmFile = fs.readFileSync("../../build/optimized.wasm");
```

assembly/Quat.ts
```diff
@@ -213,10 +216,22 @@ export class Quat {
...
         // Remove the scale from the matrix
-        var lx: f32 = 1.0 / Mathf.sqrt(m00 * m00 + m01 * m01 + m02 * m02);
-        var ly: f32 = 1.0 / Mathf.sqrt(m10 * m10 + m11 * m11 + m12 * m12);
-        var lz: f32 = 1.0 / Mathf.sqrt(m20 * m20 + m21 * m21 + m22 * m22);
-
+        //var lx: f32 = 1.0 / Mathf.sqrt(m00 * m00 + m01 * m01 + m02 * m02);
+        //var ly: f32 = 1.0 / Mathf.sqrt(m10 * m10 + m11 * m11 + m12 * m12);
+        //var lz: f32 = 1.0 / Mathf.sqrt(m20 * m20 + m21 * m21 + m22 * m22);
+        lx = m00 * m00 + m01 * m01 + m02 * m02;
+        if (lx === 0)
```
MaxGraey, line 223:
```suggestion
        if (Mathf.abs(lx) <= 1e-6)
```

examples/animation/test.js
kungfooman: Thanks! It's a script for executing the unit tests, so eventually it should test both versions

MaxGraey: I see

package.json
kungfooman: Thanks! Changed it

assembly/Mat4.ts
kungfooman: Thanks! Fixed

assembly/Mat4.ts
kungfooman: Thanks! Fixed aswell

assembly/Quat.ts
kungfooman: Any other opinions on this one? I am not sure

kungfooman:
It all works so far, all math unit tests just generate two warnings (0.9999994234 vs 1)

Made a demonstration here:

[IMAGE ALT TEXT HERE](https://www.youtube.com/watch?v=zY7d7e5U3mo)

Debugging the WASM module in Chrome is possible (like seen in video), but inspecting values doesn't work for me

assembly/Quat.ts
MaxGraey: Or `abs(x) <= 7e-46` https://webassembly.studio/?f=nw1g86emsnl

kungfooman:
Wow, I am so happy about all the last performance commits...

My test scene is the 1+64 Maila models (the woman):

image

Default JS code dropped from ~32ms per frame to ~25ms:

image

AssemblyScript code dropped from ~26ms to ~20ms:

image

mvaligursky:
There are some very promising results @kungfooman, thanks for working on this!

I am also interested in impact on other and older platforms / browsers / devices. In ideal case we find a solution that works on most devices vs having to ship two builds. If user wants to support older devices, which many of our customers are (majority of mobile customers), they would not have any benefits on any devices.

Also curious about debugging experience in other browsers as well.

malDuffin:
Great work Hermann!

There is another example project in the first post here ( 2nd link ), it
would be very interesting to hear about any performance improvements you
get compared to the original version ( 1st  link )
https://forum.playcanvas.com/t/is-it-possible-to-batch-skinned-animated-characters-in-playcanvas/9339

assembly/Quat.ts
```diff
@@ -0,0 +1,388 @@
...
+    copy(rhs: Quat): Quat {
+        this.x = rhs.x;
+        this.y = rhs.y;
+        this.z = rhs.z;
+        this.w = rhs.w;
+        return this;
+    }
+
+    equals(rhs: Quat): boolean {
+        return ((this.x === rhs.x) && (this.y === rhs.y) && (this.z === rhs.z) && (this.w === rhs.w));
```
MaxGraey, line 38:
```suggestion
        return this.x === rhs.x && this.y === rhs.y && this.z === rhs.z && this.w === rhs.w;
```

assembly/Quat.ts
```diff
@@ -0,0 +1,388 @@
...
+                axis.z *= -1;
+                rad *= -1;
+            }
+        } else {
+            // If s is zero, return any axis (no rotation - axis does not matter)
+            axis.x = 1;
+            axis.y = 0;
+            axis.z = 0;
+        }
+        return rad * pc_math.RAD_TO_DEG;
```
MaxGraey, line 61:
```suggestion
        let ax: f32 = 1, ay: f32 = 0, az: f32 = 0;
        if (s !== 0) {
            s = 1.0 / s;
            ax = this.x * s;
            ay = this.y * s;
            az = this.z * s;
            if (ax < 0 || ay < 0 || az < 0) {
                // Flip the sign
                ax *= -1;
                ay *= -1;
                az *= -1;
                rad *= -1;
            }
        }
        axis.x = ax;
        axis.y = ay;
        axis.z = az;
        return rad * pc_math.RAD_TO_DEG;
```

assembly/Quat.ts
```diff
@@ -0,0 +1,388 @@
...
+        var z: f32;
+
+        var qx = this.x;
+        var qy = this.y;
+        var qz = this.z;
+        var qw = this.w;
+
+        var a2: f32 = 2.0 * (qw * qy - qx * qz);
+        if (a2 <= -0.99999) {
+            x = 2 * Mathf.atan2(qx, qw);
```
MaxGraey, line 76: I guess it should be
```suggestion
            x = -2 * Mathf.atan2(qx, qw);
```

assembly/Quat.ts
```diff
@@ -0,0 +1,388 @@
...
+        if (Mathf.abs(sinHalfTheta) < 0.001) {
+            this.w = (lw * 0.5 + rw * 0.5);
+            this.x = (lx * 0.5 + rx * 0.5);
+            this.y = (ly * 0.5 + ry * 0.5);
+            this.z = (lz * 0.5 + rz * 0.5);
+            return this;
+        }
+
+        var ratioA = Mathf.sin((1.0 - alpha) * halfTheta) / sinHalfTheta;
+        var ratioB = Mathf.sin(alpha * halfTheta) / sinHalfTheta;
```
MaxGraey, line 361:
```suggestion
        var angle = alpha * halfTheta;
        var ratioA = Mathf.sin(halfTheta - angle) / sinHalfTheta;
        var ratioB = Mathf.sin(angle) / sinHalfTheta;
```

assembly/Quat.ts
kungfooman: Thanks! I understand to remove divisions with faster multiplications, but is there a point of declaring ax/ay/az besides "coding style taste"? Otherwise I would like the code to look nearly exactly like the original JS code

MaxGraey: It also reduced memory accessing to instance fields

kungfooman:
@malDuffin Thanks! A large part of your demo is spent in mul2, so surely it will be faster using AssemblyScript

image

@mvaligursky Thanks! Made a "debugging video experience" for Edge and Firefox: https://www.youtube.com/watch?v=wMOMfIIsK6Y&feature=youtu.be

Edge is basically a fork of Chrome and Firefox devtools don't even support WASM source maps

About support for old devices: I tested this on my 2013'ish Nexus 7 tablet via Chrome, and it works without any noticeable problems for me. Just gotta make sure to not allocate useless amounts of memory (like 10mb only is fine, 300mb is not).

About shipping multiple versions... WebAssembly has no "sync api", but all PlayCanvas examples depend on a synchronously loaded `playcanvas.js` file due to lack of Separation of Concerns (mixing HTML + JS, instead of `blend.html` and `blend.js` for example) and that pattern is exemplified everywhere because it's so easy to get people started.

One solution could be to use `<script type="text/paperscript" canvas="myCanvas">` (like in PaperScript), then PlayCanvas can simply instantiate the code once the WebAssembly version is loaded.

@MaxGraey Thanks so much for all the reviewing and helping!

assembly/Quat.ts
kungfooman: Would it make sense to multiply both with `1 / sinHalfTheta` here aswell? Two divisions vs one division and two multiplications 😅

MaxGraey: > Would it make sense to multiply both with 1 / sinHalfTheta here aswell? 

No, usually two float point divisions better

mvaligursky:
Hi guys - thanks for your work on it, it's definitely very interesting, and the performance gains are pretty promising.
I've created new issue https://github.com/playcanvas/engine/issues/2402 to try and explore this further, from slightly different angle. Your input and / or help with this would be appreciated.

slimbuck:
I can imagine switching from f64 (js default) to f32 could alter the behaviour of some apps.

Is it possible to introduce a typedef (or similar) so we can easily test both versions of this code, f32 and f64?

kungfooman:
Thanks for the interest @slimbuck, redirecting the answer from @MaxGraey (Discord):

```
yes. You could do something like this:
type real = DEF_F32_OR_F64;

export class Vec3 {
    x: real;
    y: real;
    z: real;
}

and use later:
asc ... --use DEF_F32_OR_F64=f32
or
asc ... --use DEF_F32_OR_F64=f64

but it also required redefine Math <-> Mathf. I guess you could try it like:
asc ... --use DEF_F32_OR_F64=f32, Math=NativeMathf

and
asc ... --use DEF_F32_OR_F64=f64, Math=NativeMath

and always use Math instead Mathf
```

I will have to do some JS rewriting to support 64bit aswell, stuff like:

```js

Mat4.prototype.assignDataView = function () {
    this.data = new Float32Array(assemblyscript.module.memory.buffer, this.ptr, 16);
};
```

and

```js
Object.defineProperty(Vec3.prototype, 'x', {
    get: function () {
        return assemblyscript.module.F32[this.ptr >> 2]; // the shifting is same as dividing by 4, used to quickly lookup the value in assemblyscript.module.F32
    },
    set: function (newValue) {
        assemblyscript.module.F32[this.ptr >> 2] = newValue;
    }
});
```

kungfooman:
I fully integrated 32-bit and 64-bit versions with unit tests and this is the result:

image

Two unit tests fail due 32-bit (in-)precision, but in practice I never saw a failed scene because of that. 🙈

mvaligursky:
nice, fantastic to have as an option!

kungfooman:
I made a live demo now, so everyone can check it out easily on multiple devices/browsers:

http://playcanvas.killtube.org/playcanvas-engine/examples/#graphics/hierarchy.html

That `graphics/hierarchy.html` example is super strange... for some reason (maybe some Chrome internal issue), the `AS64` version is extremly laggy when it comes to these `gl.uniformMatrix3fv` / `uniformMatrix4fv` methods (probably not the only performance bottleneck, still figuring around):

image

```js
Chrome:
    JS   version: CPU  ~29ms   GPU  ~16ms
    AS32 version: CPU  ~27ms   GPU  ~15ms
    AS64 version: CPU ~120ms   GPU ~102ms

Firefox (no GPU profiling nor useful performance analyzer):
    JS   version: CPU ~42ms
    AS32 version: CPU ~42ms
    AS64 version: CPU ~62ms
```

Firefox is usually always slower than Chrome, but the `AS64` version running in Firefox can even surpass Chrome...

mvaligursky:
Interesting tests. For some reason Point Cloud Simulation seems a lot slower using AS? That's the second more CPU intensive example there.

kungfooman:
Good point, I have no clue where these strange numbers in the DevTools analyzer come from, but they seem to eat all the time:

image

MaxGraey:
As I told [here](https://github.com/playcanvas/engine/pull/2213#issuecomment-651031481) js<->wam interop really big bottleneck especially on chrome. So make sense translate to AS / wasm only complex methods which have process time much bigger than interop time. Also make sense batch all operations inside something like `ndarray` but in term of CG. It will reduce interop frequiency. For example Godot has `Vector3Array`, `QuaternionArray` and etc.

mvaligursky:
that's main reason here https://github.com/playcanvas/engine/issues/2402 I'd be keen to explore different approach to wasm where we have separate .as files compiled to wasm to allow us to offload hot functions to it vs everything.

kungfooman:
I just had way too much code (object lookups etc.) in the getters/setters, just optimized it:

image

And assigning it in the constructor:

```js

Vec3.prototype.assignDataView = function () {
    // #ifdef X32
    this.arr = new Float32Array(assemblyscript.module.memory.buffer, this.ptr, 3);
    // #endif

    // #ifdef X64
    this.arr = new Float64Array(assemblyscript.module.memory.buffer, this.ptr, 3);
    // #endif
};
```

`Vec3#set` was slow too, I could replace it with:

```js
Vec3.prototype.set = function (x, y, z) {
    //vec3_set(this.ptr, x, y, z);
    this.arr[0] = x;
    this.arr[1] = y;
    this.arr[2] = z;
    return this;
};
```

(still optimizing more, already down from 10ms to 5ms)

I am just not sure how optimizing a few "hot" functions shall reduce the work load of methods like `Mat4#mulAffine2` during glTF animations. Uploading 2x Mat4 and returning 1x Mat4 or simply not do it? Now I just need to pass three pointers. The reason I work on this is because of the sheer amount of math in complex hierarchy calculations (where I already showed that WebAssembly is way faster, and in the future hopefully even more so, when WebGL<->WebAssembly Memory archives higher performance, like `gl.uniformMatrix4fv`).

mvaligursky:
Specifically, to optimize animation matrix update .. which is a loop over matrix multiplies, what I'm hoping to do is:
- in js code collect local matrices from the hierarchy to a float32array
- keep inverse bind poses in float32array as well (these are constants)
- then per character call "AffineMultArrays" and pass both arrays in and get an array back which can be directly written to texture / shader constants.

But further to this, and that's even bigger deal, is that at some point I'd like to have worker thread(s) with those hot wasm functions on them, and use atomics to write a small job system where I can kick off these jobs from main thread, wasm workers would pick them up and execute them, and then later in the frame when needed for rendering on the main thread, we could wait for those. This bit is more complex, but if done right, we can completely hide the cost of the matrix multiplies by running them on worker threads while main thread continues on other things.

This would be applied to other things such as cpu particle update, culling, possibly even command buffer generation, making the engine multithreaded.

MaxGraey:
> I made a live demo now, so everyone can check it out easily on multiple devices/browsers:
http://playcanvas.killtube.org/playcanvas-engine/examples/#graphics/hierarchy.html

It doesn't work on Safari due to safari hasn't `WebAssembly.instantiateStreaming` (only `WebAssembly.instantiate`). AS loader btw handle this.

kungfooman:
@MaxGraey Thank you for testing! I have used a rather old and slightly customized version of the AssemblyScript loader (slightly customized the latest Loader again, so it works without a bundler, just `<script src="Loader.js"></script>`)

Could you test again? http://playcanvas.killtube.org/playcanvas-engine/examples/#graphics/hierarchy.html

Now I feel a bit dissatisfied because there is no demo that shows off the performance improvement for 64 animated/complex glTF models, quite the contrary, just getting slower for all the WASM Interop and browser regressions 😅

MaxGraey:
Yeah, now everything works

kungfooman:
Got `GraphNode` compiled, now I just need to write the JS layer over it (and test):

image

My idea is to collect the entire hierarchy in WebAssembly memory and then just have a `WebWorker` crunching the numbers at whatever FPS (e.g. at 20 FPS instead of max 60 FPS hierarchy calculations, if otherwise 16.6 ms / 60 FPS can't be reached).

Problem I see is browser support for `new WebAssembly.Memory({ shared: true });` (at 37.31%):

https://caniuse.com/mdn-javascript_builtins_webassembly_memory_memory_shared

If not going for the `WebWorker` idea (or just as a use-if-available option), having the `GraphNode` in WebAssembly should reduce lots of WASM Interop aswell (`GraphNode` actions are what usually dominates the performance analysis Flame Charts for me).

willeastcott:
While this is really cool + interesting, I'm struggling to see this being merged at any point. Would it be sensible to close this PR?

PR #2217, opened 2020-06-29
Skinning and morphing matrices optimization

### LINKED ISSUE #2202: Optimize bone skinning system - matrices
At the moment this is the skinning matrix operations:
- scene hierarchy store Mat4 worldMatrices as an input for skinning
- we invert root node world matrix of the character hierarchy
- for each bone of the character:
    - we multiply its world matrix with inverted root matrix to have it in root space
    - we multiply this result with default pose
    - when mesh is visible, we write all these matrices to GPU (either float texture, or uniforms)

The last loop is 2 Mat4 multiplies, and a copy, costing us about 0.5ms on CPU for a test character.

Proposal:
- change shader code to use transposed matrix of 4x3 size, allowing us to drop 4 floats.
- change matrix multiplies to use 4x3 matrix part only, saving 25% of the cost
- perhaps inline transpose into it (write results transposed) to avoid extra copy.

mvaligursky:
Fixes #2202

Skinning / batching code on GPU now uses 4x3 matrix, instead of full 4x4 matrix.

Details:
- optimized affine transformation matrix multiplication added: Mat4.mulTransform2, used by skinning matrix array preparation. In test case of 20 characters, this resulted in this code taking 2.0ms instead of 2.7ms in Mac (more on mobile)
- skinning shader:
    - uniform based: uses 3 x Vec4 per bone instead of Mat4, allowing more bones 
    - texture based: 3 pixel per bone instead of 4. For 4 bone skinning, this means 4 texture samples less per vertex
    - both shader types: slightly more complicated math to generate transposed matrix
- morphing shader:
    - similar to skinning, but uses single bone per vertex instead of 4

### DIFF
--- a/src/graphics/graphics.js
+++ b/src/graphics/graphics.js
@@ -1071,7 +1071,9 @@ export var UNIFORMTYPE_FLOATARRAY = 17;
 export var UNIFORMTYPE_TEXTURE2D_SHADOW = 18;
 export var UNIFORMTYPE_TEXTURECUBE_SHADOW = 19;
 export var UNIFORMTYPE_TEXTURE3D = 20;
-export var UNIFORMTYPE_VEC4ARRAY = 21;
+export var UNIFORMTYPE_VEC2ARRAY = 21;
+export var UNIFORMTYPE_VEC3ARRAY = 22;
+export var UNIFORMTYPE_VEC4ARRAY = 23;
 
 // map of engine pc.TYPE_*** enums to their corresponding typed array constructors and byte sizes
 export var typedArrayTypes = [Int8Array, Uint8Array, Int16Array, Uint16Array, Int32Array, Uint32Array, Float32Array];
--- a/src/graphics/graphics.js
+++ b/src/graphics/graphics.js
@@ -1071,6 +1071,7 @@ export var UNIFORMTYPE_FLOATARRAY = 17;
 export var UNIFORMTYPE_TEXTURE2D_SHADOW = 18;
 export var UNIFORMTYPE_TEXTURECUBE_SHADOW = 19;
 export var UNIFORMTYPE_TEXTURE3D = 20;
+export var UNIFORMTYPE_VEC4ARRAY = 21;
 
 // map of engine pc.TYPE_*** enums to their corresponding typed array constructors and byte sizes
 export var typedArrayTypes = [Int8Array, Uint8Array, Int16Array, Uint16Array, Int32Array, Uint32Array, Float32Array];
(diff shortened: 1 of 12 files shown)


src/graphics/graphics.js
```diff
@@ -1071,6 +1071,7 @@ export var UNIFORMTYPE_FLOATARRAY = 17;
 export var UNIFORMTYPE_TEXTURE2D_SHADOW = 18;
 export var UNIFORMTYPE_TEXTURECUBE_SHADOW = 19;
 export var UNIFORMTYPE_TEXTURE3D = 20;
+export var UNIFORMTYPE_VEC4ARRAY = 21;
```
willeastcott, line 1074: If we're exposing VEC4ARRAY, should we do VEC2 + VEC3 too for completeness?

raytranuk:
I don't see any problems - but I have a quick question: do you have any numbers on the performance difference in the shaders? I'm assuming we are not generally vertex shader bound - but there are places where that can happen and its worth just knowing what the impact would be?

src/graphics/graphics.js
mvaligursky: good idea, I'll add that.

mvaligursky:
> I don't see any problems - but I have a quick question: do you have any numbers on the performance difference in the shaders? I'm assuming we are not generally vertex shader bound - but there are places where that can happen and its worth just knowing what the impact would be?

On the Mac I cannot see a difference in GPU load. Overall it has 4 less texture samples per vertex, and maybe 15 arithmetic instructions. Considering we still have 12 texture samples left, I'd say this could even help it remove the bandwidth issue.

src/math/mat4.js
```diff
@@ -271,6 +271,75 @@ Object.assign(Mat4.prototype, {
         return this;
     },
 
+    /**
+     * @function
+     * @name pc.Mat4#mulTransform2
```
willeastcott, line 276: Just a suggestion, but would `mulAffine2` be better naming?

willeastcott: Also, I'm wondering if this needs to be public API initially since it's only used internally. Thoughts?

mvaligursky: I started with mulAffine2 yesterday but wasn't sure it'd be accessible to people. I'm happy to go back to that.

I'd like this to be a public API - this should be the commonly used matrix multiply function to use in most cases. But I agree most common use would be in the engine - end users should usually just use Vec3 + quaternion.

src/graphics/graphics.js
mvaligursky: done

src/math/mat4.js
mvaligursky: renamed.

PR #2222, opened 2020-06-29
Check for null and undefined explicitly in element.text

### LINKED ISSUE #2220: Setting 0 to element.text results in an empty string being set
https://github.com/playcanvas/engine/blob/master/src/framework/components/element/text-element.js#L1217

value when `0` resolves to false

yaustar:
Fixes #2220

### DIFF
--- a/src/framework/components/element/text-element.js
+++ b/src/framework/components/element/text-element.js
@@ -1209,7 +1209,7 @@ Object.defineProperty(TextElement.prototype, "text", {
     },
     set: function (value) {
         this._i18nKey = null;
-        var str = value && value.toString() || "";
+        var str = value != null && value.toString() || "";
         this._setText(str);
     }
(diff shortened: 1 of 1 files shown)


src/framework/components/element/text-element.js
```diff
@@ -1209,7 +1209,7 @@ Object.defineProperty(TextElement.prototype, "text", {
     },
     set: function (value) {
         this._i18nKey = null;
-        var str = value && value.toString() || "";
+        var str = value != null && value.toString() || "";
```
willeastcott, line 1212: Personally, I'd do:
```
var str = (value == null) ? "" : value.toString();
```

See [here](https://stackoverflow.com/questions/2647867/how-can-i-determine-if-a-variable-is-undefined-or-null) for reference.

PR #2227, opened 2020-06-30
Moved compound and trigger update back to 1.27.4 order

### LINKED ISSUE #2152: Compound physics crash in Ammo when entity.setPosition is used every frame
Does not occur in v1.27.4.

Demo project: https://playcanvas.com/project/693422/overview/1280-physics-type-change (press Space)

By setting the position of the entity every frame, it's dirtying the world transform which in turn removes/adds shapes to the compound parent (in updateCompoundChildTransform)

I don't know why that causes an overflow in ammo stepsimulation and only if the rigidbody is dynamic, a compound and not moving/collision.

Calling entity.getPosition 'fixes' this by undirtying the world transform via sync() call in graphnode.

yaustar:
Fixes #2152

The order of updates of compound and trigger objects are back in the order that they were in 1.27.4

### DIFF
--- a/src/framework/components/collision/component.js
+++ b/src/framework/components/collision/component.js
@@ -273,6 +273,10 @@ Object.assign(CollisionComponent.prototype, {
         if (this.entity.rigidbody) {
             if (this.entity.rigidbody.enabled) {
                 this.entity.rigidbody.enableSimulation();
+
+                if (this.data.type === 'compound') {
+                    this.system._compounds.push(this);
+                }
             }
         } else if (this._compoundParent && this !== this._compoundParent) {
             if (this._compoundParent.shape.getNumChildShapes() === 0) {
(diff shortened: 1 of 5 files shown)


willeastcott:
So my optimization of the physics engine included moving all physics update into the `pc.RigidBodyComponentSystem#onUpdate`. In a local branch, I have taken this a step further and moved all physics into a separate class called `PhysicsWorld`. This PR takes us back to before the refactor. Is it really not possible to change the order of updates in `pc.RigidBodyComponentSystem#onUpdate` to resolve this?

yaustar:
When I was debugging the issue, the cause of it was moving the compound update between the script system update and render. I thought that restoring 1.27.4 would be the safest option.

I hadn't considered at changing the order or an alternative fix in the fear of creating new bugs 😅

I/We can take another look as no active project I know is being blocked by this.

PR #2231, opened 2020-07-01
ImageBitmap loader

### LINKED ISSUE #2185: Consider using ImageBitmap API to load images
API: https://developer.mozilla.org/en-US/docs/Web/API/ImageBitmap
ImageBitmap provides an asynchronous and resource efficient pathway to prepare textures for rendering in WebGL

Issue: I created a project in Editor with single cube, material, and attached 6 large jpg textures on the material. Textures are marked as not for preload, to start the application ASAP and load textures later. Each time a texture is loaded, profiler shows around 100ms being spent on main thread to upload the texture to WebGl, large part of which is Image Decode.

Screen Shot 2020-06-22 at 1 55 13 PM

It's possible this cost could be removed by using the mentioned API.

Some other references:
https://wiki.whatwg.org/wiki/ImageBitmap_Options
https://threejs.org/docs/#api/en/loaders/ImageBitmapLoader

Note: Not supported by all browsers.

slimbuck:
Fixes #2185

Uses ImageBitmap interface to load images in Chrome. ImageBitmap spends less time on the main thread decoding images compared to Image.

Notes/Issues:
- ImageBitmap is disabled on Firefox, which doesn't yet support the options object on construction.
- ImageBitmap interface uses promises, probably the only place in the engine right now.
- ImageBitmap takes flipY and premultipliedAlpha flags in the constructor and ignores the WebGL device state to control these.

### DIFF
--- a/src/resources/parser/texture/img.js
+++ b/src/resources/parser/texture/img.js
@@ -1,4 +1,5 @@
 import { path } from '../../../core/path.js';
+import { http } from '../../../net/http.js';
 
 import { PIXELFORMAT_R8_G8_B8, PIXELFORMAT_R8_G8_B8_A8, TEXHINT_ASSET } from '../../../graphics/graphics.js';
 import { Texture } from '../../../graphics/texture.js';
@@ -92,17 +93,26 @@ Object.assign(ImgParser.prototype, {
     },
 
     _loadImageBitmap: function (url, originalUrl, crossOrigin, callback) {
-		fetch( url ).then( function ( res ) {
-			return res.blob();
-		} ).then( function ( blob ) {
-			return createImageBitmap( blob, {
-                premultiplyAlpha: 'none'
-            } );
-		} ).then( function ( imageBitmap ) {
-			callback(null, imageBitmap);
-		} ).catch( function ( e ) {
-			callback(e);
-        } );
+        var options = {
+            cache: true,
+            responseType: "blob",
+            retry: this.retryRequests
+        };
+        pc.http.get(url, options, function (err, blob) {
+            if (err) {
+                callback(err);
+            } else {
+                createImageBitmap(blob, {
+                    premultiplyAlpha: 'none'
+                })
+                .then( function (imageBitmap) {
+                    callback(null, imageBitmap);
+                })
+                .catch( function (e) {
+                    callback(e);
+                });
+            }
+        });
     }
 });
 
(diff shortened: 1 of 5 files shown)


src/resources/parser/texture/img.js
```diff
@@ -84,6 +89,20 @@ Object.assign(ImgParser.prototype, {
         };
 
         image.src = url;
+    },
+
+    _loadImageBitmap: function (url, originalUrl, crossOrigin, callback) {
```
willeastcott, line 94: Do we need to duplicate the retry logic here too?

slimbuck: Yes, possibly. I may just use http instead of promise and bring the download method in line with other image types.

willeastcott: Yeah, sounds sensible.

src/resources/parser/texture/img.js
```diff
@@ -15,6 +16,7 @@ function ImgParser(registry, retryRequests) {
     // by default don't try cross-origin, because some browsers send different cookies (e.g. safari) if this is set.
     this.crossOrigin = registry.prefix ? 'anonymous' : null;
     this.retryRequests = !!retryRequests;
+    this.useImageBitmap = typeof ImageBitmap !== 'undefined' && /Firefox/.test( navigator.userAgent ) === false;
```
willeastcott, line 19: This will prevent Firefox from uses its ImageBitmap support when it works as required presumably? We probably need a better 'test' to check for working ImageBitmap support.

slimbuck: I tried changing the test to:
```javascript
createImageBitmap(new ImageData(1, 1), {
    premultiplyAlpha: 'none'
})
.then(function (imageData) {
    self.useImageBitmap = true;
})
.catch(function (err) {
    self.useImageBitmap = false;
});
```

but unfortunately the async nature of .catch means the flag is set only later on, by which point the loader has already started processing texture requests.

willeastcott: Sorry, can you explain exactly what's broken in Firefox? 😄

slimbuck: :D Firefox throws if you attempt to call createImageBitmap with an options object (which is the means of specifying flipY and premultiplyAlpha).

https://bugzilla.mozilla.org/show_bug.cgi?id=1367251

slimbuck: So I was attempting to construct an object with options. Unfortunately such a call on Firefox fails asynchronously and so you find out the result after you've already started loading textures.

willeastcott: OK, well, I'm all right with that UA check for now.

willeastcott: But consider adding your voice to that Bugzilla thread, saying the PlayCanvas engine is integrating this and we don't want to blanket ban using ImageBitmap in FF.

slimbuck: done

PR #2233, opened 2020-07-01
ParticleSystem rendering in UI space

mvaligursky:
ParticleSystem component can be used on Entity inside hierarchy with parent being Screen component, allowing particles getting rendered integrated in UI layer (under button, on top of button ..)

Instructions:

* Add entity with ParticleSystem into hierarchy with root Screen. Note that it cannot be added as a child of screen, and its parent needs to be one of the Element components (Image, Group ..)
* Tick on 'ScreenSpace' option on the ParticleSystem
* Make sure the Layer is set to UI (same layer as the Screen) - otherwise sorting does not work

Engine example: engine/examples/#user-interface/button-particle.html

Supersedes #2178

### DIFF
--- a/src/framework/components/particle-system/component.js
+++ b/src/framework/components/particle-system/component.js
@@ -109,7 +109,7 @@ var depthLayer;
  * @property {boolean} depthWrite If enabled, the particles will write to the depth buffer. If disabled, the depth buffer is left unchanged and particles will be guaranteed to overwrite one another in the order in which they are rendered.
  * @property {boolean} noFog Disable fogging.
  * @property {boolean} localSpace Binds particles to emitter transformation rather then world space.
- * @property {boolean} screenSpace Renders particles in 2D screen space. This needs to be set when particle system is part of hierarchy with {@link pc.ScreenComponent} at its root, and allows particle system to integrate with the rendering of {@link pc.ElementComponent}s. Note that an entity with ParticleSystem component cannot be parented directly to {@link pc.ScreenComponent}, but has to be a child of a {@link pc.ElementComponent}, for example {@link pc.LayoutGroupComponent}.
+ * @property {boolean} screenSpace Renders particles in 2D screen space. This needs to be set when particle system is part of hierarchy with {@link pc.ScreenComponent} as its ancestor, and allows particle system to integrate with the rendering of {@link pc.ElementComponent}s. Note that an entity with ParticleSystem component cannot be parented directly to {@link pc.ScreenComponent}, but has to be a child of a {@link pc.ElementComponent}, for example {@link pc.LayoutGroupComponent}.
  * @property {number} numParticles Maximum number of simulated particles.
  * @property {number} rate Minimal interval in seconds between particle births.
  * @property {number} rate2 Maximal interval in seconds between particle births.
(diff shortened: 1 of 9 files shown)


src/framework/components/particle-system/component.js
```diff
@@ -108,7 +109,7 @@ var depthLayer;
  * @property {boolean} depthWrite If enabled, the particles will write to the depth buffer. If disabled, the depth buffer is left unchanged and particles will be guaranteed to overwrite one another in the order in which they are rendered.
  * @property {boolean} noFog Disable fogging.
  * @property {boolean} localSpace Binds particles to emitter transformation rather then world space.
- * @property {number} numParticles Maximum number of simulated particles.
+ * @property {boolean} screenSpace Renders particles in 2D screen space. This needs to be set when particle system is part of hierarchy with {@link pc.ScreenComponent} at its root, and allows particle system to integrate with the rendering of {@link pc.ElementComponent}s. Note that an entity with ParticleSystem component cannot be parented directly to {@link pc.ScreenComponent}, but has to be a child of a {@link pc.ElementComponent}, for example {@link pc.LayoutGroupComponent}. * @property {number} numParticles Maximum number of simulated particles.
```
vkalpias, line 112: Looks like numParticles was deleted

mvaligursky: fixed

src/framework/components/particle-system/component.js
```diff
@@ -108,6 +109,7 @@ var depthLayer;
  * @property {boolean} depthWrite If enabled, the particles will write to the depth buffer. If disabled, the depth buffer is left unchanged and particles will be guaranteed to overwrite one another in the order in which they are rendered.
  * @property {boolean} noFog Disable fogging.
  * @property {boolean} localSpace Binds particles to emitter transformation rather then world space.
+ * @property {boolean} screenSpace Renders particles in 2D screen space. This needs to be set when particle system is part of hierarchy with {@link pc.ScreenComponent} at its root, and allows particle system to integrate with the rendering of {@link pc.ElementComponent}s. Note that an entity with ParticleSystem component cannot be parented directly to {@link pc.ScreenComponent}, but has to be a child of a {@link pc.ElementComponent}, for example {@link pc.LayoutGroupComponent}.
```
willeastcott, line 112: The screen doesn't have to be at the root, right? It just has to be an ancestor in the hierarchy, right?

willeastcott: Also, what's the motivation for having the `screenSpace` property anyway? If the particlesystem component is a descendant of an entity with a screen component, shouldn't screen space be assumed automatically?

mvaligursky: I changed the doc to use "ancestor", thanks.

mvaligursky: > Also, what's the motivation for having the `screenSpace` property anyway? If the particlesystem component is a descendant of an entity with a screen component, shouldn't screen space be assumed automatically?

The motivation is avoiding the complexity / performance impact of that. If a child entity is added to a parent, the code would need to check if any of the descendant entities contain ParticleSystem and if there is a Screen component ascendant. setParent could get expensive.

PR #2234, opened 2020-07-01
Adding support for json script attributes with a schema (1 level deep)

vkalpias:
This PR adds support for JSON script attributes. The engine already had some support for a 'json' type which was meant to be a string which would be `JSON.parse`ed. However this PR changes that mechanism and instead allows users to define JSON script attributes like so:

(example):
```
MyScript.attributes.add('product', {
  type: 'json',
  schema: [{
    name: 'serialNumber',
    type: 'string'
  }, {
    name: 'price',
    type: 'number',
    default: 1
  }, {
    name: 'thumbnail',
    type: 'asset',
    assetType: 'texture'   
  }]
})
```

The above can then be accessed by code like so:

```
MyScript.prototype.initialize = function() {
    console.log(this.product.serialNumber, this.product.price, this.product.thumbnail);
}
```
If you wanted an array of 'products' you could do the same as above but also add `array: true` like so:

```
MyScript.attributes.add('products', {
  type: 'json',
  array: true,
  schema: [{
    name: 'serialNumber',
    type: 'string'
  }, {
    name: 'price',
    type: 'number',
    default: 1
  }, {
    name: 'thumbnail',
    type: 'asset',
    assetType: 'texture'   
  }]
})
```

The above can then be accessed by code like so:

```
MyScript.prototype.initialize = function() {
    for (var i = 0; i < this.products.length; i++) {
        console.log('product index', i, this.products[i].serialNumber, this.products[i].price, this.products[i].thumbnail);
    }
    
}
```


yaustar:
It would be great to have an example of how to use the JSON schema in the JS docs of the script attributes but not essential.

Maksims:
Specifically in User Manual: https://developer.playcanvas.com/en/user-manual/scripting/script-attributes/

src/script/script-attributes.js
```diff
@@ -33,14 +34,31 @@ var rawToValue = function (app, args, value, old) {
             }
             return null;
         case 'json':
-            if (typeof value === 'object') {
-                return value;
+            var result = {};
+
+            if (!value || typeof value !== 'object') {
```
slimbuck, line 39: could move this inside the block starting line 43

yaustar:
> Specifically in User Manual: https://developer.playcanvas.com/en/user-manual/scripting/script-attributes/

I would like it in both, I and other users go to the API section for the syntax of how to use a particular part of the engine API (eg raycasting being a good example). Whereas the User Manual is for the longer form of why and how it can be used.

src/script/script-attributes.js
vkalpias: Done

PR #2240, opened 2020-07-03
iOS particle velocity fix

ellt92:
The particle systems velocity curves are packed into 1D textures that are then unpacked by the tex1Dlod_lerp function. The sampling of these curves are imprecise on iOS currently. Setting the precision of the texture sampler to high looks to resolve this issue.

### DIFF
--- a/src/graphics/program-lib/chunks/particle.vert
+++ b/src/graphics/program-lib/chunks/particle.vert
@@ -10,11 +10,11 @@ float saturate(float x) {
     return clamp(x, 0.0, 1.0);
 }
 
-vec4 tex1Dlod_lerp(sampler2D tex, vec2 tc) {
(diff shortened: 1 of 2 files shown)


slimbuck:
Could you perhaps add a little context on what problem this solves?

ellt92:
@slimbuck The particle velocity curves are packed into a 1D texture that are then unpacked by the tex1Dlod_lerp function. Setting the curves value to 0 produces static particles on desktop browsers but small movements on iOS. Setting the precision of the texture sampler to high produces static particles on iOS too.

ellt92:
> could you please apply this to other places if needed.
> At least in particle.vert, tex1Dlod_lerp I think
Sure, i'll update that function too.

PR #2244, opened 2020-07-06
Anim component updates for the editor

ellt92:
- Anim component can now be linked to a state graph asset by id. Any changes to the state graph will trigger a reload of the anim component.
- Adding a model to an entity triggers a reload of the anim component.
- The anim component can now load a set of anim tracks into its controller based on a set of animation asset ids.
- No longer require the specific name 'DEFAULT_LAYER' to interact with the base layer.
- Check animation clips exist in the evaluator before updating them.


src/framework/components/anim/component.js
```diff
@@ -72,6 +72,50 @@ Object.assign(AnimComponent.prototype, {
...
+
+    setAnimationAssets: function () {
+        var animationAssets = {};
+        for (var i = 0; i < this.data.layers.length; i++) {
+            var layer = this.data.layers[i];
+            var layerName = layer.name;
+            for (var j = 0; j < layer.states.length; j++) {
+                var stateName = layer.states[j];
+                if (stateName !== 'START' && stateName !== 'END') {
+                    animationAssets[layerName + ':' + stateName] = {
```
slimbuck, line 86: it would be more efficient to assign to `this.data.animationAssets` directly and then call `this.loadAnimationAssets();` after the loop instead. what you've done here is clean though, so if there won't be many layers and states and it's very fast anyway, then fine to leave as-is.

src/framework/components/anim/component.js
```diff
@@ -72,6 +72,50 @@ Object.assign(AnimComponent.prototype, {
...
+                if (!animationAsset || !animationAsset.asset) {
+                    this.removeNodeAnimations(stateName, layer.name);
+                    continue;
+                }
+                var assetId = animationAsset.asset;
+                var asset = this.system.app.assets.get(assetId);
+                if (asset.resource) {
+                    this.assignAnimation(stateName, asset.resource, layer.name);
+                } else {
+                    asset.on('load', function (layerName, stateName) {
```
slimbuck, line 110: wondering whether you want to use `asset.once` here?

ellt92: Yeah once is more fitting.

src/framework/components/anim/component.js
ellt92: 👍 I've went ahead with this approach for efficiency.

src/framework/components/anim/controller.js
```diff
@@ -23,6 +23,9 @@ Object.defineProperties(AnimState.prototype, {
     animations: {
         get: function () {
             return this._animations;
+        },
+        set: function (value) {
+            this._animations = value;
```
willeastcott, line 28: Any reason why `animations` isn't just a raw property (which would be faster to access and less verbose)?

src/framework/components/anim/controller.js
```diff
@@ -253,6 +256,14 @@ Object.defineProperties(AnimController.prototype, {
             return playable;
         }
     },
+    playing: {
+        get: function () {
+            return this._playing;
+        },
+        set: function (value) {
+            this._playing = value;
```
willeastcott, line 264: Also could be a raw property.

src/framework/components/anim/controller.js
```diff
@@ -253,6 +256,14 @@ Object.defineProperties(AnimController.prototype, {
             return playable;
         }
     },
+    playing: {
+        get: function () {
+            return this._playing;
+        },
+        set: function (value) {
```
yaustar, line 263: A bit odd to have a setter for a property that sounds like it should be read only.

src/framework/components/anim/controller.js
ellt92: No other reason that consistency, I was keeping all vars declared privately & only exposing the ones that are necessary. If efficiency is a concern I can remove the private declarations.

src/framework/components/anim/controller.js
willeastcott: Hmm, yeah, good point. I wouldn't expect that to be settable from the naming...

src/framework/components/anim/controller.js
willeastcott: It's not a big deal. Just thought I'd point it out. 😄

src/framework/components/anim/controller.js
ellt92: This will be private api only. The component has a public read only 'playable' property.

src/framework/components/anim/controller.js
ellt92: I'll look at this when doing a performance pass of the anim component then :)

PR #2248, opened 2020-07-07
disable imagebitmap

slimbuck:
Cubemap faces are not correctly unflipped under imageBitmap, so disable for now.


Maksims:
This seems to be fixed in Chrome: https://chromium.googlesource.com/chromium/src/+/f440137cd96a18f64b9931e8f369670197204d55

willeastcott:
Cool - but how do you cater for older versions of Chrome that do the wrong thing?

Maksims:
Feature can have a better path only for newer versions. Although using `navigator.userAgent` is not great, indeed.

PR #2250, opened 2020-07-07
Add support for OES_vertex_array_object

### LINKED ISSUE #2204: Implement Vertex Array Object support
To improve performance of rendering, consider implementing support of VAOs.

There are polyfills we can use if not all needed platforms support it:
https://github.com/greggman/oes-vertex-array-object-polyfill

mvaligursky:
Using VAOs to set up vertex buffer attributes mapping to vertex shader, to minimize number of calls to WebGL and improve performance.

Screen Shot 2020-07-07 at 12 51 02 PM

Implementation details:
- if drawing is using a single vertex buffer (VB) - which is the majority of cases, a VAO object is created and attached to the VB and is used for rendering
- if drawing is using multiple vertex buffers - which is the case of both implementation of morphing (morph targets in attributes, or vertex_id attribute), and also hardware instancing (per instance vertex attributes) - VAO is generated for the combination of VBs and is stored in cache in Device class.
- IE11 does not support the extension, and so polyfill is used to provide it
- we now use fixed mapping from attribute name to attribute location (see semanticToLocation in graphics.js) to allow VAO for VB to be used for all shaders, instead of requiring VAO per shader variant. WebGl gives us 16 locations on all devices, and so currently existing 30 semantics are mapped to 16 locations with some overlap, but this is unlikely to cause problems as the overlap takes place on rarely used semantic names (ATTR10-ATTR15 overlap TEXCOORD2-TEXCOORD7 range)

Performance:
- testing on a scene with just over 5000 objects, frame duration on CPU dropped from 15.3ms to 12.5ms, which is saving of about 18%. Impact will be lower on scenes with smaller number of rendered objects.

Fixes #2204

### DIFF
--- a/examples/graphics/hierarchy.html
+++ b/examples/graphics/hierarchy.html
@@ -70,8 +70,9 @@
                 var offset = spacing * (gridSize - 1) * 0.5;
                 for (var x = 0; x < gridSize; x++) {
                     for (var y = 0; y < gridSize; y++) {
+                        var shape = Math.random() < 0.5 ? "box" : "sphere";
                         var position = new pc.Vec3(x * spacing - offset, spacing, y * spacing - offset);
-                        var entity = createPrimitive("box", position, new pc.Vec3(scale, scale, scale));
+                        var entity = createPrimitive(shape, position, new pc.Vec3(scale, scale, scale));
 
                         parent.addChild(entity);
                         entities.push(entity);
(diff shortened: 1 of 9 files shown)


Maksims:
Not sure it is related to this PR, but testing latest master engine, with large demos: Seemore, Space Base, and few others, crashes WebGL context in Chrome and Chromium, on both of my GPUs (GTX 1060 and Intel HD Graphics 630).

willeastcott:
I'd recommend you go through the test plan @mvaligursky.

mvaligursky:
Thanks for heads up @Maksims , I see missing floor and error message in Seemore, and I'm investigating.

mvaligursky:
I have the fix PR @Maksims - I tried Seemore, Space Base and After the flood and they work now .. any others I should test?

Maksims:
Checked few projects, they work now well!

PR #2252, opened 2020-07-07
Glb texture coordinate fix

slimbuck:
We address here the issue with glTF having texture coordinates with top-left origin while the engine has bottom-right origin. We previously dealt with this by _not_ flipping glTF embedded textures (where all others were being flipped). The change in this PR is to flip glTF texture coordinates instead and also flip textures, thus bringing them inline with the rest of the engine.

 The following changes are made:
- flip glTF texture V coordinates at load time
- flip glTF textures to match the rest of the engine
- fix for cubemap faces under ImageBitmap
- reenable ImageBitmap
- also simplify glTF vertex buffer creation function

### DIFF
--- a/src/resources/parser/glb-parser.js
+++ b/src/resources/parser/glb-parser.js
@@ -171,7 +171,10 @@ var generateIndices = function (numVertices) {
     return dummyIndices;
 };
 
-var generateNormals = function (sourceDesc, vertexDesc, positions, numVertices, indices) {
+var generateNormals = function (sourceDesc, indices) {
+
+    var positions = sourceDesc[SEMANTIC_POSITION];
+    var numVertices = positions.count;
 
     if (!indices) {
         indices = generateIndices(numVertices);
@@ -182,19 +185,14 @@ var generateNormals = function (sourceDesc, vertexDesc, positions, numVertices, 
     var normals = new Float32Array(normalsTemp.length);
     normals.set(normalsTemp);
 
-    vertexDesc.push({
-        semantic: SEMANTIC_NORMAL,
-        components: 3,
-        type: TYPE_FLOAT32
-    });
-
     sourceDesc[SEMANTIC_NORMAL] = {
         buffer: normals.buffer,
         size: 12,
         offset: 0,
         stride: 12,
-        count: numVertices
-    };
+        count: numVertices,
+        components: 3,
+        type: TYPE_FLOAT32    };
 };
 
 var flipTexCoordVs = function (vertexBuffer) {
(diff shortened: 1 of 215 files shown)


src/resources/parser/glb-parser.js
```diff
@@ -182,22 +185,81 @@ var generateNormals = function (sourceDesc, vertexDesc, positions, numVertices,
...
         offset: 0,
         stride: 12,
-        count: numVertices
+        count: numVertices,
+        components: 3,
+        type: TYPE_FLOAT32
+    };
+};
+
+var flipTexCoordVs = function (vertexBuffer) {
```
mvaligursky, line 199: t would be probably simpler and faster to pass semantic as a parameter to this function, and call it twice (uv0, uv1) then doing array in array loop when copying

src/resources/parser/glb-parser.js
```diff
@@ -182,22 +185,81 @@ var generateNormals = function (sourceDesc, vertexDesc, positions, numVertices,
...
+                case TYPE_FLOAT32:
+                    floatOffsets.push({ offset: element.offset / 4 + 1, stride: element.stride / 4 });
+                    break;
+                case TYPE_UINT16:
+                    shortOffsets.push({ offset: element.offset / 2 + 1, stride: element.stride / 2 });
+                    break;
+                case TYPE_UINT8:
+                    byteOffsets.push({ offset: element.offset + 1, stride: element.stride });
+                    break;
+            }
```
mvaligursky, line 219: maybe print error message in default:

slimbuck: The glTF spec only supports these 3 data types (it's not an implementation choice on our part) don't think a message is necessary?

src/resources/parser/glb-parser.js
slimbuck: Not sure I follow. This function could be called twice, but as far as I can see that just moves code from here to the caller. Perhaps I misunderstand?

src/resources/parser/glb-parser.js
mvaligursky: perfect in that case, not needed.

src/resources/parser/glb-parser.js
mvaligursky: I take it back. I was looking at this

```
var flip = function (offsets, type, one) {
        var typedArray = new type(vertexBuffer.storage);
        for (i = 0; i < offsets.length; ++i) {
            var index = offsets[i].offset;
            var stride = offsets[i].stride;
            for (j = 0; j < vertexBuffer.numVertices; ++j) {
                typedArray[index] = one - typedArray[index];
                index += stride;
            }
        }
    };
```
and thought inside look is the one that loops over offsets. It's the other way, so all good.

slimbuck: kk

PR #2267, opened 2020-07-10
[FIX] Fix to ScreenSpace ParticleSystem billboard rotation

mvaligursky:
- keeping them rotated in ScreenSpace and ignoring camera rotation


willeastcott:
I'll defer to @ellt92 to review this. 😄

PR #2278, opened 2020-07-13
Update the anim state graph to support the new data schema

ellt92:

### DIFF
--- a/src/framework/components/anim/state-graph.js
+++ b/src/framework/components/anim/state-graph.js
@@ -3,6 +3,35 @@ function AnimStateGraph(data) {
     this._parameters = {};
     var i;
     if (!Array.isArray(data.layers)) {
+        // Layers as an object
+        // var data = {
+        //     "layers": {
+        //         "0": {
+        //             "name": "Base",
+        //             "states": [0, 1],
+        //             "transitions": [0]
+        //         }
+        //     },
+        //     "states": {
+        //         "0": {
+        //             "name": "START",
+        //             "speed": 1
+        //         },
+        //         "1": {
+        //             "name": "New State",
+        //             "speed": 1,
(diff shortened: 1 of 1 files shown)


src/framework/components/anim/state-graph.js
```diff
@@ -1,23 +1,46 @@
...
-            });
-        } else {
-            this._layers = data.layers;
-        }
-        this._parameters = {};
-        for (var paramId in data.parameters) {
-            var param = data.parameters[paramId];
-            this._parameters[param.name] = { type: param.type, value: param.value };
+    var i;
+    if (!Array.isArray(data.layers)) {
```
slimbuck, line 5: Why this test here? Is data.layers sometimes an array? Shouldn't it be testing whether data.layers is an object instead (as the code below assumes)?

src/framework/components/anim/state-graph.js
```diff
@@ -1,23 +1,46 @@
...
-                transitions: data.transitions
-            });
-        } else {
-            this._layers = data.layers;
-        }
-        this._parameters = {};
-        for (var paramId in data.parameters) {
-            var param = data.parameters[paramId];
-            this._parameters[param.name] = { type: param.type, value: param.value };
+    var i;
```
slimbuck, line 4: Might be good to document what data is in this case?

src/framework/components/anim/state-graph.js
ellt92: Yeah, I missed out the assignment of data.layers in the case it's an array. I've added this in now.

src/framework/components/anim/state-graph.js
ellt92: 👍

PR #2289, opened 2020-07-15
Deprecate pc.makeArray

willeastcott:
`pc.makeArray` is private API. But it's pretty useless considering its behavior can be achieved with `Array.prototype.slice.call`. It's been officially moved to `deprecated.js`.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -30,27 +30,6 @@ var data = { }; // Storage for exported entity data
 /**
  * @private
  * @function
- * @name pc.makeArray
- * @description Convert an array-like object into a normal array.
- * For example, this is useful for converting the arguments object into an array.
- * @param {object} arr - The array to convert.
- * @returns {Array} An array.
- */
-function makeArray(arr) {
-    var i,
-        ret = [],
-        length = arr.length;
-
-    for (i = 0; i < length; ++i) {
-        ret.push(arr[i]);
-    }
-
-    return ret;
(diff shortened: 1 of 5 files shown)


willeastcott:
That's one option. But it may be that it's unnecessary to ever remove deprecated stuff and instead just provide a legacy build and a non-legacy build. That's now easy with rollup - we just add another target minus the `deprecated.js` module. We could even make that a publish option in the Editor.

Maksims:
Taking in account some deprecated stuff is still within code, for example WebVR - is deprecated, but still there.
One of the way to go, is to add another **define**, called `LEGACY`, we can then wrap deprecated stuff into LEGACY define, and that clean new engine can be compiled.

But probably, legacy should be still a default engine for most people, to avoid unnecessary confusion "why it suddenly breaks". And provide Launch and Publish options, to run without LEGACY stuff.

willeastcott:
Yeah, that's certainly one option. We do need to deal with:

* WebVR
* AudioSource component
* Legacy Script component
* Zone component

Maksims:
Sad to see Zone component to go away, it definitely worth re-thinking it, as often wanted to separate spaces in zones, and do some clustering, and other logic based on them (fill zone with probes e.g.).

willeastcott:
It's not guaranteed we'll get rid of the zone component. But right now, there's little to no benefit from it.

PR #2294, opened 2020-07-15
Support for a more compressed scene file format

zachpeterpaul:
This PR implements the engine-side handling of scenes and templates in a compressed (optimized) format. Scenes are compressed if a user selects the corresponding option in the PlayCanvas Editor during a Project Publish or Project Download operation.

For instance, a 4MB original scene (367KB gzipped) can be optimized to 1.6MB (154KB gzipped).

This is achieved chiefly by using 1 or 2 character field names, and storing the position, rotation and scale data in separate flat arrays that take advantage of repeating patterns, rather than as separate vectors with each entity.

Each compressed scene is accompanied by an array of original field names, indexed by the numerical value of the new (short) key in the compressed format. Original (longer) field names are restored early during scene parsing, so that the rest of the engine can work as before. This takes around 5ms per 1MB of original scene data.

We also use the fact that in the engine position, rotation and scale are set by value and not by reference to a vector (e.g. entitly.setLocalPosition). So there is no need to restore the original vectors. Position, rotation and scale are set directly from the flat array in the compressed scene.

Savings resulting from using the compressed format:

* Over a 3G network, reducing download size by 300KB saves ~1 second.
* Time to unzip in the browser is proportional to the .gz file size.
* Time to parse json is reduced because of the smaller size of the result and a simpler object structure.

These savings will be even larger when we further simplify the format by removing common parts of entity components.


mvaligursky:
Looks good, simpler than I expected.

willeastcott:
Hi @zachpeterpaul, can you give a slightly expanded description to this PR please? It'll help the open source community understand what this PR is and the benefits it will bring (include some numbers too ideally).

mvaligursky (approved):

PR #2296, opened 2020-07-17
Deprecate pc.Frustum#update

willeastcott:
Currently, the API has `pc.Frustum#update` which takes two matrices (view and projection). It would be cleaner to have a function which sets the frustum planes from a single generic matrix. Therefore, this PR deprecates `pc.Frustum#update` and replaces it with the better named `pc.Frustum#setFromMat4` (which is consistent with `pc.Quat#setFromMat4` too). This eliminates a matrix mult op every frame in the forward renderer too.

### DIFF
--- a/src/shape/frustum.js
+++ b/src/shape/frustum.js
@@ -1,27 +1,17 @@
-import { Mat4 } from '../math/mat4.js';
-
 /**
  * @class
  * @name pc.Frustum
- * @classdesc A frustum is a shape that defines the viewing space of a camera.
+ * @classdesc A frustum is a shape that defines the viewing space of a camera. It can be
+ * use to determine visibility of points and bounding spheres. Typically, you would not
+ * create a Frustum shape directly, but instead query {@link pc.CameraComponent#frustum}.
  * @description Creates a new frustum shape.
  * @example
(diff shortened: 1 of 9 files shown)


src/shape/frustum.js
```diff
@@ -1,114 +1,117 @@
...
         // Extract the numbers for the RIGHT plane
-        this.planes[0][0] = vpm[3] - vpm[0];
-        this.planes[0][1] = vpm[7] - vpm[4];
-        this.planes[0][2] = vpm[11] - vpm[8];
-        this.planes[0][3] = vpm[15] - vpm[12];
+        plane = planes[0];
+        plane[0] = vpm[3] - vpm[0];
+        plane[1] = vpm[7] - vpm[4];
+        plane[2] = vpm[11] - vpm[8];
+        plane[3] = vpm[15] - vpm[12];
```
raytranuk, line 43: Almost repeating code - could this be made more concise with a function?

willeastcott: Yep. Next PR will refactor `pc.Plane` so that `pc.Frustum` can use those instead. But I thought it would be better to break it into two steps.

raytranuk: seems fine then :)

PR #2299, opened 2020-07-19
Refactor camera code

willeastcott:
This PR updates `pc.CameraComponent[System]` to the current recommended format.

It also includes a general refactor to improve the general code structure. Hopefully, performance is slightly better also.

Successfully tested against:

* All engine examples
* Seemore
* After The Flood

### DIFF
--- a/src/framework/components/camera/component.js
+++ b/src/framework/components/camera/component.js
@@ -161,7 +161,8 @@ CameraComponent.prototype.constructor = CameraComponent;
     { name: 'rect', readonly: false },
     { name: 'renderTarget', readonly: false },
     { name: 'scissorRect', readonly: false },
-    { name: 'viewMatrix', readonly: true }
+    { name: 'viewMatrix', readonly: true },
+    { name: 'vrDisplay', readonly: false }
 ].forEach(function (property) {
     var name = property.name;
     var options = {};
@@ -228,41 +229,6 @@ Object.defineProperty(CameraComponent.prototype, "priority", {
             var layer = this.system.app.scene.layers.getLayerById(layers[i]);
             if (!layer) continue;
             layer._sortCameras();
-        }
-    }
-});
-
-/**
- * @private
- * @deprecated
- * @name pc.CameraComponent#vrDisplay
- * @type {pc.VrDisplay}
- * @description The {@link pc.VrDisplay} that the camera is current displaying to. This is
- * set automatically by calls to {@link pc.CameraComponent#enterVr} or
- * {@link pc.CameraComponent#exitVr}. Setting this property to a display directly enables
- * the camera to use the transformation information from a display without rendering
- * stereo to it, e.g. for "magic window" style experiences.
- * @example
- * // enable magic window style interface
- * var display = this.app.vr.display;
- * if (display) {
- *     this.entity.camera.vrDisplay = display;
- * }
- *
- * var camera = this.entity.camera;
- * camera.enterVr(function (err) {
- *     if (err) return;
- *     var display = camera.vrDisplay; // access presenting pc.VrDisplay
- * });
- */
-Object.defineProperty(CameraComponent.prototype, "vrDisplay", {
-    get: function () {
-        return this._camera.vrDisplay;
-    },
-    set: function (newValue) {
-        this._camera.vrDisplay = newValue;
-        if (newValue) {
-            newValue._camera = this._camera;
         }
     }
 });
--- a/src/framework/components/camera/component.js
+++ b/src/framework/components/camera/component.js
@@ -159,6 +159,7 @@ CameraComponent.prototype.constructor = CameraComponent;
     { name: 'projection', readonly: false },
     { name: 'projectionMatrix', readonly: true },
     { name: 'rect', readonly: false },
+    { name: 'renderTarget', readonly: false },
     { name: 'scissorRect', readonly: false },
     { name: 'viewMatrix', readonly: true }
 ].forEach(function (property) {
--- a/src/framework/components/camera/component.js
+++ b/src/framework/components/camera/component.js
@@ -198,7 +198,7 @@ Object.defineProperty(CameraComponent.prototype, "layers", {
             layer.removeCamera(this);
         }
 
-        this._camera.layers = newValue.slice(0);
+        this._camera.layers = newValue;
 
         if (!this.enabled || !this.entity.enabled) return;
 
--- a/src/framework/components/camera/component.js
+++ b/src/framework/components/camera/component.js
@@ -1,19 +1,22 @@
-import { CLEARFLAG_COLOR, CLEARFLAG_DEPTH, CLEARFLAG_STENCIL } from '../../../graphics/graphics.js';
-
 import { ASPECT_AUTO } from '../../../scene/constants.js';
+import { Camera } from '../../../scene/camera.js';
 
 import { Component } from '../component.js';
+
+import { PostEffectQueue } from './post-effect-queue.js';
 
 /**
  * @component
  * @class
  * @name pc.CameraComponent
  * @augments pc.Component
- * @classdesc The Camera Component enables an Entity to render the scene. A scene requires at least one
- * enabled camera component to be rendered. Note that multiple camera components can be enabled
- * simultaneously (for split-screen or offscreen rendering, for example).
+ * @classdesc The Camera Component enables an Entity to render the scene. A scene
+ * requires at least one enabled camera component to be rendered. Note that multiple
+ * camera components can be enabled simultaneously (for split-screen or offscreen
+ * rendering, for example).
  * @description Create a new Camera Component.
- * @param {pc.CameraComponentSystem} system - The ComponentSystem that created this Component.
+ * @param {pc.CameraComponentSystem} system - The ComponentSystem that created this
+ * Component.
  * @param {pc.Entity} entity - The Entity that this Component is attached to.
  * @example
  * // Add a pc.CameraComponent to an entity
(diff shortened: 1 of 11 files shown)


src/framework/components/camera/system.js
```diff
@@ -75,118 +44,113 @@ CameraComponentSystem.prototype.constructor = CameraComponentSystem;
...
-        var data = {};
-        for (var i = 0, len = properties.length; i < len; i++) {
+        for (var i = 0; i < properties.length; i++) {
             var property = properties[i];
-            data[property] = _data[property];
-        }
-
-        if (data.layers && Array.isArray(data.layers)) {
-            data.layers = data.layers.slice(0);
-        }
```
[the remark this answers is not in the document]
willeastcott: No, because the code now just does:
```
component[property] = value;
```
Which sets the layers correctly.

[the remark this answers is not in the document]
yaustar: Does that take a clone of the layers/Do we still need a clone of the layers?

src/scene/camera.js
```diff
@@ -125,8 +400,8 @@ Object.assign(Camera.prototype, {
         }
 
         if (this._projMatDirty || this._viewMatDirty || this._viewProjMatDirty) {
```
slimbuck, line 402: could this go into a viewProjMat getter and then shared with the similar block in screenToWorld()?

src/scene/camera.js
```diff
@@ -203,313 +478,26 @@ Object.assign(Camera.prototype, {
...
-     * @function
-     * @name pc.Camera#getClearOptions
-     * @description Retrieves the options used to determine how the camera's render target will be cleared.
-     * @returns {object} The options determining the behaviour of render target clears.
-     */
-    getClearOptions: function () {
-        return this._clearOptions;
-    },
-
     _evaluateProjectionMatrix: function () {
```
slimbuck, line 481: wondering why this isn't just wrapped into projectionMatrix getter?

src/framework/components/camera/component.js
```diff
@@ -29,157 +32,204 @@ import { Component } from '../component.js';
  * @example
  * // Update a property on a camera component
  * entity.camera.nearClip = 2;
- * @property {number} projection The type of projection used to render the camera. Can be:
+ * @property {number} projection The type of projection used to render the camera.
```
slimbuck, line 35: Is it possible to change pc.PROJECTION_ enum to use strings instead of numbers? Much nicer for debugging!

src/framework/components/camera/component.js
```diff
@@ -29,157 +32,204 @@ import { Component } from '../component.js';
...
- * The order of the values is [x, y, width, height].
- * @property {pc.PostEffectQueue} postEffects The post effects queue for this camera. Use this to add or remove post effects from the camera.
- * @property {boolean} frustumCulling Controls the culling of mesh instances against the camera frustum, i.e. if objects outside of camera should be omitted from rendering.
- * If true, culling is enabled.
- * If false, all mesh instances in the scene are rendered by the camera, regardless of visibility. Defaults to false.
- * @property {pc.callbacks.CalculateMatrix} calculateTransform Custom function you can provide to calculate the camera transformation matrix manually. Can be used for complex effects like reflections. Function is called using component's scope.
+ * The order of the values is [x, y, width, height]. Defaults to [0, 0, 1, 1].
+ * @property {pc.PostEffectQueue} postEffects The post effects queue for this camera.
+ * Use this to add or remove post effects from the camera.
+ * @property {boolean} frustumCulling Controls the culling of mesh instances against
```
slimbuck, line 82: defaults to false?!

src/framework/components/camera/data.js
```diff
@@ -1,49 +1,5 @@
...
-    this.clearStencilBuffer = true;
-    this.nearClip = 0.1;
-    this.farClip = 1000;
-    this.fov = 45;
-    this.orthoHeight = 100;
-    this.projection = PROJECTION_PERSPECTIVE;
-    this.priority = 0;
-    this.rect = new Vec4(0, 0, 1, 1);
-    this.scissorRect = new Vec4(0, 0, 1, 1);
     this.enabled = true;
```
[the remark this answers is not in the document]
willeastcott: It's the way all the other newly structured components format the ComponentData class. `enabled` is from the base `Component` class. When all components are in the currently recommended format, I suspect we can remove all these ComponentData classes entirely.

src/framework/components/camera/system.js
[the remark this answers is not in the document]
willeastcott: Check out the refactored `camera.js` which now does:

```
Object.defineProperty(Camera.prototype, 'layers', {
    get: function () {
        return this._layers;
    },
    set: function (newValue) {
        this._layers = newValue.slice(0);
    }
});
```

src/framework/components/camera/component.js
willeastcott: Yes! I like this idea. But I'd rather not make this part of this PR. An issue requesting this is fine but begs the question about doing a proper audit of _all_ engine constants and make their type consistent across the whole engine.

slimbuck: great, happy to know we can do this at some point.

src/framework/components/camera/component.js
willeastcott: It did previously - I've just kept existing behavior. We can flip it to true. In actual fact, if you create a new camera component in the Editor, it defaults to `true`. So this discrepancy is a bit weird and should probably be changed.

src/scene/forward-renderer.js
```diff
@@ -779,28 +782,44 @@ Object.assign(ForwardRenderer.prototype, {
...
+        this.cameraParams[2] = (1 - f / n) * 0.5;
+        this.cameraParams[3] = (1 + f / n) * 0.5;
+        this.cameraParamsId.setValue(this.cameraParams);
 
         var device = this.device;
         device.setRenderTarget(target);
         device.updateBegin();
 
-        var rect = camera.getRect();
+        var rect = camera.rect;
```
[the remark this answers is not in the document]
willeastcott: Yeah, this is the bit of the PR I'm not happy with. The `Camera` class used to hold the `Object` passed in to `GraphicsDevice#clear`. To simplify the `Camera` class, I mapped it's property to those exposed by the `CameraComponent` API. But this now means I have to construct the clear options object in the forward renderer. I considered creating a new clear function that takes parameters. e.g.:

```
GraphicsDevice#clearNew(clearColor: boolean, color: Color, clearDepth: boolean, depth: number, clearStencil: boolean, stencil: number);
```

But I couldn't think of a good name for it! Got any ideas?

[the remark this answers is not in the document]
slimbuck: perhaps you could just call clearView() directly with an option to not set color and depth write? (which looks like the only difference between the two).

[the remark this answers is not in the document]
willeastcott: Done!

src/scene/camera.js
willeastcott: Yes, you're right. The matrix calculation could be refactored too. This PR is already so big though and I'm nervous about making more changes here and would rather do a follow up re-factor. 😄  Let's let the dust settle from this and then take another look.

src/scene/camera.js
willeastcott: It does need tidying up. But this PR is really, mainly about adopting the new component format and nicely dividing component functionality (property management) from camera functionality (pc.Camera). Let's address this in a follow up PR.

Christopher-Hayes:
This code has stopped working in the past day or so, do you guys know what this can be replaced with to achieve the same result? The function doesn't exist.
```
    this.camera.camera.camera.requestDepthMap();
```
I searched that function in the git blame and it turns up nothing, so not sure how it was indirectly affected. It does appear to be related to the [post-effect-queue.js](https://github.com/playcanvas/engine/blob/v1.33.1/src/framework/components/camera/post-effect-queue.js) [changes](https://github.com/playcanvas/engine/pull/2299).

willeastcott:
Hi @Christopher-Hayes. So you're using the undocumented `pc.Camera` API there instead of the [`pc.CameraComponent` API](https://developer.playcanvas.com/en/api/pc.CameraComponent.html). You can actually do:

```
var projectionMatrix = this.entity.camera.projectionMatrix;
var viewMatrix = this.entity.camera.viewMatrix;
```

As for `pc.Camera#requestDepthMap`, it was removed because it literally did nothing except increment a variable on the `pc.Camera` instance, so I removed it. These days, to request a depth map for a post effect, you do:

```
        this.needsDepthBuffer = true;
```

Such as in the [Bokeh effect](https://github.com/playcanvas/engine/blob/master/examples/assets/scripts/posteffects/posteffect-bokeh.js#L19). `_requestDepthMap` is definitely not public API.

Hope that helps. 😄

Christopher-Hayes:
Thanks! That helps a lot, I'll switch over to using Camera Component. Was working with some code Max had brought in from a previous project. Probably one of a few people who know about these undocumented APIs haha.

src/scene/lightmapper.js
```diff
@@ -412,15 +411,12 @@ Object.assign(Lightmapper.prototype, {
         // Create pseudo-camera
         if (!lmCamera) {
             lmCamera = new Camera();
-            lmCamera._node = new GraphNode();
-            lmCamera.clearColor[0] = 0;
-            lmCamera.clearColor[1] = 0;
-            lmCamera.clearColor[2] = 0;
-            lmCamera.clearColor[3] = 0;
-            lmCamera.clearDepth = 1;
```
yaustar, line 420: We don't clear the depthBuffer in the previous version of the lightmapper. Bug or designed behaviour?

willeastcott: Check out the next line. `clearFlags` used to be set to just `CLEARFLAG_COLOR`. So previously, the depth bugger was never cleared before. So no change in behavior.

PR #2312, opened 2020-07-23
Expose glTF bufferView to callbacks

slimbuck:
This PR simplifies the way bufferViews are handled in GlbParser and adds support for `preprocess`, `processAsync` and `postprocess` callbacks. This callback support will allow us to easily integrate bufferView-based compression schemes, for example https://github.com/zeux/meshoptimizer.

ContainerHandler's asynchronous callbacks were changed. The `processAsync` handlers are now always called, but if the handler invokes the continuation without an error or result (i.e. `callback(null, null);`), this indicates the resource was unhandled and GlbParser continues to load the resource as usual.

Also fixed a bug with MiniStats sometimes failing to update it's dynamic texture.

### DIFF
--- a/extras/mini-stats/mini-stats.js
+++ b/extras/mini-stats/mini-stats.js
@@ -311,9 +311,9 @@ Object.assign(Graph.prototype, {
             }
 
             // write latest sample to the texture
+            var gl = this.device.gl;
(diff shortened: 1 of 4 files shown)


slimbuck:
done! :D

PR #2315, opened 2020-07-25
[DOCS] Improve docs for pc.Application#renderLines

willeastcott:
Fixes https://github.com/playcanvas/developer.playcanvas.com/issues/114

### DIFF
--- a/src/framework/application.js
+++ b/src/framework/application.js
@@ -2015,14 +2015,37 @@ Object.assign(Application.prototype, {
     /**
      * @function
      * @name pc.Application#renderLines
-     * @description Draw an array of lines.
-     * @param {pc.Vec3[]} position - An array of points to draw lines between.
-     * @param {pc.Color[]} color - An array of colors to color the lines. This must be the same size as the position array.
+     * @description Renders an arbitrary number of discrete line segments. The lines
+     * are not connected by each subsequent point in the array. Instead, they are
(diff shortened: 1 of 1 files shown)


src/framework/application.js
```diff
@@ -2015,14 +2015,37 @@ Object.assign(Application.prototype, {
...
      * @name pc.Application#renderLines
-     * @description Draw an array of lines.
-     * @param {pc.Vec3[]} position - An array of points to draw lines between.
-     * @param {pc.Color[]} color - An array of colors to color the lines. This must be the same size as the position array.
+     * @description Renders an arbitrary number of discrete line segments. The lines
+     * are not connected by each subsequent point in the array. Instead, they are
+     * individual segments specified by two points. Therefore, the lengths of the
+     * supplied position and color arrays must be the same and also must be a multiple
+     * of 2. The colors of the ends of each line segment will be interpolated along
+     * the length of each line.
```
yaustar, line 2023: I be tempted to say that each pair of points represents a single line rather than using the word segments. Although, the example does make it clearer below.

willeastcott: Yeah, I used segments to indicate that the resulting line isn't necessarily continuous. It was actually quite tricky to draw up a clear, concise description, but let's go with this for now. It's better than what was there before. 😄

PR #2319, opened 2020-07-27
Anim component api updates

ellt92:
Introduces new API to Anim component layers to manage clip times & exposes the pc.AnimClip looping property to AnimState. 

New API:
```javascript
// Returns the currently active states duration.
pc.AnimComponentLayer#activeStateDuration: number;
// The active states time in seconds
pc.AnimComponentLayer#activeStateCurrentTime: number;
```
New Anim State Graph Schema:
**State Graph Asset Schema:**
```json
{
    "state": {
        "type": "object",
        "description": "Defines a state which will become a node in the state graph",
        "properties": {
            "name": {
                "type": "String",
                "description": "The name of the state. Used to link animations to a specific state and find states used in transitions"
            },
            "speed": {
                "type": "Number",
                "description": "The speed at which the animation linked to this state should play at"
            },
            "loop": {
                "type": "Boolean",
                "description": "Whether the state should continuously loop playback of its clips"
            }
        },
        "required": ["name"]
    }
}
```

### DIFF
--- a/src/framework/components/anim/controller.js
+++ b/src/framework/components/anim/controller.js
@@ -7,11 +7,12 @@ import {
     ANIM_STATE_START, ANIM_STATE_END
 } from './constants.js';
 
-function AnimState(controller, name, speed) {
+function AnimState(controller, name, speed, loop) {
     this._controller = controller;
     this._name = name;
     this._animations = [];
     this._speed = speed || 1.0;
+    this._loop = loop === undefined ? true : loop;
 }
 
 Object.defineProperties(AnimState.prototype, {
@@ -31,6 +32,11 @@ Object.defineProperties(AnimState.prototype, {
     speed: {
         get: function () {
             return this._speed;
+        }
+    },
+    loop: {
+        get: function () {
+            return this._loop;
         }
     },
     playable: {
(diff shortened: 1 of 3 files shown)


src/framework/components/anim/controller.js
```diff
@@ -269,6 +276,36 @@ Object.defineProperties(AnimController.prototype, {
...
     },
+    activeStateDuration: {
+        get: function () {
+            if (this.activeStateName === ANIM_STATE_START || this.activeStateName === ANIM_STATE_END)
+                return 0.0;
+
+            var maxDuration = 0.0;
+            for (var i = 0; i < this.activeState.animations.length; i++) {
+                var activeClip = this._animEvaluator.findClip(this.activeState.animations[i].name);
+                if (activeClip && activeClip.track.duration > maxDuration) {
```
slimbuck, line 287: Small point, might be neater to go:
```javascript
maxDuration = math.Max(maxDuration, activeClip.track.duration);
```

Also, ```this.activeState.animations``` is referenced multiple times, might be worth storing it in a var.

ellt92: Cheers, updated these 👍

PR #2321, opened 2020-07-28
[FIX] VAO usage fix

### LINKED ISSUE #2311: 3D Screen elements disappear when entering WebVR
So far, this seems to be restricted to WebVR projects and not WebXR.

It seems to be due to a change made in 1.32.1

Test project: https://playcanvas.com/project/704962/overview/akuna-fork-from-forums

1. Go to https://launch.playcanvas.com/960120?use_local_engine=https://code.playcanvas.com/playcanvas-1.32.1.js
2. Open devtools and toggle mobile view
3. Refresh page and see the white rect element
4. Enter VR via the bottom right icon
5. See that the white rect element disappears and remains so even when leaving VR.

Console is spammed with
image

WebVR Labs also has this problem: https://playcanvas.com/editor/scene/484674

What it should look like
Screenshot 2020-07-22 at 17 06 41

With 1.32.1
Screenshot 2020-07-22 at 17 09 36

mvaligursky:
Fixes #2311 (both projects)

- unbind VAO at the end of block of rendering, to protect VAO bind on device getting modified by locking VB / IB and similar operations
- cleanup to text-element - removed duplicate and calling `mesh.destroy()` instead, also making sure `meshInstance.material` is set to null in both cases (one was not)
- improved usage of `device._removedByInstancing` - this only counts saving by autoinstancing (which is currently not used), and not when user sets up instancing.

### DIFF
--- a/src/framework/components/element/text-element.js
+++ b/src/framework/components/element/text-element.js
@@ -384,7 +384,6 @@ Object.assign(TextElement.prototype, {
                 // destroy old mesh
                 if (meshInfo.meshInstance) {
                     this._removeMeshInstance(meshInfo.meshInstance);
-                    meshInfo.meshInstance.material = null;
                 }
 
                 // if there are no letters for this mesh continue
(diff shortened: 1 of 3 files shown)


willeastcott:
Net of 4 lines deleted. Nice. :smile:

PR #1692, opened 2019-09-12
AssetsRegistry.loadFromUrl material now will load textures too

Maksims:
Fixes #1096

Now textures are loaded too, when loading material asset from url.

I confirm I have signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #2325, opened 2020-07-29
Load textures when loading material from URL

### LINKED ISSUE #2322: A problem like #1096
It looks like the https://github.com/playcanvas/engine/issues/1096  is reappear,maybe  I used it in a wrong  way.
I also asked a question on https://forum.playcanvas.com/t/changing-a-material-with-downloaded-material/6029

### When I used v1.31.0, I also encountered this problem, and when I tried to use v1.32.3 (the latest version), the bug was still not solved, What I do now is to use this method to load materials dynamically without Editor. This material was downloaded from the Editor and put on my own server. It is normal to change the material in the editor. However, when I do not use the Editor to load the material dynamically(I did something like this:
pc.app.assets.loadFromUrl…), the effect is not the same, so I print and check it and I found that the“ materia.sphereMap " is null, I compared it and found that texture is missing. It does not load normally

```
// I loaded the model with code like this:
var url = “…/........./paint_whit.json”;
app.assets.loadFromUrl(url, “model”, function (err, asset) {
   console.log(asset);
   self.car = new pc.Entity();
   self.car.addComponent(“model”, {
      type: “asset”,
      asset: -2,
      castShadows: true
   });
   app.root.addChild(self.car);
});
// It all worked well,but when I did like this to load my materia:
var url = “…/......../TG86_carpaint_red_01_mat.json”;
app.assets.loadFromUrl(url, “material”, function (err, asset) {
   console.log(asset);   
   self.material = asset.resource;
   self.material.update();
});
// the“ materia.sphereMap " is null,which means the texture was not loaded,but it should be
```

###  this is screenshots of my material in my project which I download from Playcanvas Editor
materia2020-07-29 15 36 27

yaustar:
Fixes #2322

### DIFF
--- a/src/asset/asset-registry.js
+++ b/src/asset/asset-registry.js
@@ -450,7 +450,13 @@ Object.assign(AssetRegistry.prototype, {
 
         var startLoad = function (asset) {
             asset.once("load", function (loadedAsset) {
-                callback(null, loadedAsset);
+                if (type === 'material') {
+                    self._loadTextures(loadedAsset, function (err, textures) {
+                        callback(err, loadedAsset);
+                    });
+                } else {
+                    callback(null, loadedAsset);
+                }
             });
             asset.once("error", function (err) {
                 callback(err);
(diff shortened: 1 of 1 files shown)


yaustar:
Please do not merge

yaustar:
Amended fix to #1692

PR #2326, opened 2020-07-29
Transform-feedback particle simulation engine example

mvaligursky:
- new engine example
- transform feedback integration with SpectorJS
- error when transform feedback is used with non-interleaved buffer (has to be interleaved)

Screen Shot 2020-07-29 at 4 54 02 PM

### DIFF
--- a/src/graphics/transform-feedback.js
+++ b/src/graphics/transform-feedback.js
@@ -84,7 +84,7 @@ function TransformFeedback(inputBuffer, usage) {
 
     // #ifdef DEBUG
     if (!inputBuffer.format.interleaved && inputBuffer.format.elements.length > 1) {
-        console.error("Vertex buffer used by TransformFeedback needs to be inverleaved.");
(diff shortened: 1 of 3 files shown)


src/graphics/transform-feedback.js
```diff
@@ -82,6 +82,12 @@ function TransformFeedback(inputBuffer, usage) {
     this.device = inputBuffer.device;
     var gl = this.device.gl;
 
+    // #ifdef DEBUG
+    if (!inputBuffer.format.interleaved && inputBuffer.format.elements.length > 1) {
+        console.error("Vertex buffer used by TransformFeedback needs to be inverleaved.");
```
slimbuck, line 87: inverleaved -> interleaved

mvaligursky:
> Did you test this runs on Safari?

Transform-feedback needs webgl2, so no Safari .. but it's handled by example and nothing renders (no crash).

src/graphics/transform-feedback.js
mvaligursky: done, cheers

PR #2334, opened 2020-08-04
Fix for additive particles with fog and on windows

### LINKED ISSUE #2302: Particles not rendered correctly with fog and particles with additive blend type
https://playcanvas.com/project/704291/overview/fog-and-particle-affects-bug

Add fog to the scene from rendering settings
Add a particle system
Change the blend type on the particle systems to additive

Kapture 2020-07-20 at 11 35 43

raytranuk:
Fixes #2302

Also (related) fix added to additive particles on Windows

### DIFF
--- a/src/graphics/program-lib/chunks/fogExp.frag
+++ b/src/graphics/program-lib/chunks/fogExp.frag
@@ -1,9 +1,9 @@
 uniform vec3 fog_color;
 uniform float fog_density;
-float d_fog_factor=1.0; //for additive blend, this should be 0.0
+float dFogFactor=1.0;
 vec3 addFog(vec3 color) {
     float depth = gl_FragCoord.z / gl_FragCoord.w;
     float fogFactor = exp(-depth * fog_density);
     fogFactor = clamp(fogFactor, 0.0, 1.0);
-    return mix(fog_color*d_fog_factor, color, fogFactor);
+    return mix(fog_color*dFogFactor, color, fogFactor);
 }
--- a/src/graphics/program-lib/chunks/fogExp.frag
+++ b/src/graphics/program-lib/chunks/fogExp.frag
@@ -5,5 +5,5 @@ vec3 addFog(vec3 color) {
     float depth = gl_FragCoord.z / gl_FragCoord.w;
     float fogFactor = exp(-depth * fog_density);
     fogFactor = clamp(fogFactor, 0.0, 1.0);
-    return mix(fog_color*dBlendModeFogFactor, color, fogFactor);
+    return mix(fog_color * dBlendModeFogFactor, color, fogFactor);
 }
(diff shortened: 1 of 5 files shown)


src/graphics/program-lib/chunks/particle_blendAdd.frag
```diff
@@ -1,2 +1,3 @@
+    d_fog_factor=0.0; //for additive blend, this should be 0.0
```
willeastcott, line 1: Right, but the comment doesn't really tell me why. 😄

src/graphics/program-lib/chunks/fogExp.frag
```diff
@@ -1,9 +1,9 @@
 uniform vec3 fog_color;
 uniform float fog_density;
-
+float d_fog_factor=1.0; //for additive blend, this should be 0.0
```
willeastcott, line 3: I would just call this `dFogFactor`. Isn't it just uniforms (and attributes) that follow the underscore naming convention?

src/graphics/program-lib/chunks/particle_blendAdd.frag
raytranuk: removed the comment - since there were no fog comments before

src/graphics/program-lib/chunks/fogExp.frag
raytranuk: renamed to follow the naming convention correctly

raytranuk: (I think) - I left in the 'd' - because its used like other 'd' globals

raytranuk: hmm - maybe dFogBlendFactor could be better? Or dFogBlendTypeFactor?

willeastcott: You went with `dFogFactor` which is fine by me.

raytranuk: I changed it again - its now dBlendModeFogFactor - I'm thinking this makes it a bit more useful as it will appear when  searching for code related blend modes or fog?

raytranuk:
> I'm gonna nitpick here (so apologies in advance)....but do try to stick to existing code style. We have eslint to enforce lots of rules for JS, but not for GLSL chunks. So it's more important to be vigilant about this with GLSL. In these chunks (and I suspect all/most of the other chunks), there are spaces around operators (`=`, `*`).

I think I suffer from "space between operators blindness" - I just don't notice either way - and have been relying on lint or IDEs to correct it! Would be good to at least have some spacing automated tool checker for shaders? Fixes checked in.

src/graphics/program-lib/chunks/fogExp.frag
willeastcott: Looks good.

PR #2348, opened 2020-08-10
[Fix] improve compatibility of example with older mobile devices

mvaligursky:


examples/graphics/hardware-instancing.html
```diff
@@ -87,38 +87,40 @@
...
-            // generate random positions / scales and rotations
-            pos.set(Math.random() * radius - radius * 0.5, Math.random() * radius - radius * 0.5, Math.random() * radius - radius * 0.5);
-            scl.set(0.1 + Math.random() * 0.1, 0.1 + Math.random() * 0.3, 0.1 + Math.random() * 0.1);
-            rot.setFromEulerAngles(i * 30, i * 50, i * 70);
-            matrix.setTRS(pos, rot, scl);
-
-            // copy matrix elements into array of floats
-            for (var m = 0; m < 16; m++)
-                matrices[matrixIndex++] = matrix.data[m];
+        if (this.app.graphicsDevice.supportsInstancing) {
```
slimbuck, line 90: what does the example do now if instancing isn't supported?

mvaligursky: renders the background/mini-stats. The same as transform-feedback.

PR #2353, opened 2020-08-13
Added example code for curve and curve set

yaustar:
Also fixes: https://github.com/playcanvas/developer.playcanvas.com/issues/119

### DIFF
--- a/src/math/curve.js
+++ b/src/math/curve.js
@@ -22,14 +22,12 @@ import { CurveEvaluator } from './curve-evaluator.js';
  * Defaults to {@link pc.CURVE_SMOOTHSTEP}.
  *
  * @example
- * var curve = new pc.Curve(
- *     [
- *         [0, 0],
- *         [0.33, 2],
- *         [0.66, 2.6],
- *         [1, 3]
- *     ]
- * );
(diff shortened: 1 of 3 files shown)


src/math/curve.js
```diff
@@ -23,12 +23,13 @@ import { CurveEvaluator } from './curve-evaluator.js';
  *
  * @example
  * var curve = new pc.Curve(
- * [
- *     [0, 0],
- *     [0.33, 2],
- *     [0.66, 2.6],
- *     [1, 3]
- * ]);
+ *     [
```
Maksims, line 26: CurveSet has:
```javascript
new pc.CurveSet([
    [
        [0, 0],
        [0, 0],
    ],
    [
        [0, 0],
        [0, 0]
    ]
]);
```
So it clearly shows it is array of groups of tuples.
But for Curve, it should not look similar, as there is one depth less, so it probably should be:
```javascript
new pc.Curve([
    [0, 0],
    [0, 0]
])
```

yaustar: Ah, good shout. That explains why the linter was complaining

willeastcott (approved):

PR #2354, opened 2020-08-14
Anim optimizations

### LINKED ISSUE #2351: Optimize setters in AnimSystem
This function probably could be optimized, as it takes considerable amount in a frame with lots of characters. It typically executes for position, scale and rotation updates for each node in the hierarchy when character is animated, and it does string operations and constructs and array of a single item - all these things are constant and perhaps should be prepared at init time. Or perhaps there should only be a single instance of setter for position - there probably does not need to be one per node?

 ```
var entityPropertySetter = function (values) {
                // first set new values on the property as before
                setter(values);
                // create the function name of the entity properties setter
                var entityPropertySetterFunctionName = "set" +
                    entityProperty.substring(0, 1).toUpperCase() +
                    entityProperty.substring(1);
                // call the setter function for entities updated property using the newly set property value
                propertyComponent[entityPropertySetterFunctionName](this._getProperty(propertyComponent, [entityProperty]));
            };

```
and likely this as well, it should be simplified and the loop unrolled.

```
_vecSetter: function (propertyComponent, propertyHierarchy) {
        var vectorKeys = ['x', 'y', 'z', 'w'];
        var setter = function (values) {
            for (var i = 0; i < values.length; i++) {
                this._setProperty(propertyComponent, propertyHierarchy.concat(vectorKeys[i]), values[i]);
            }
        };
        return setter.bind(this);
    },
```

### PR DESCRIPTION
**Pull request by ellthompson**
Fixes #2351 

This PR optimizes the setting of animated entity properties, with particular impact to scenes with a large amount of character animations.

These optimizations were tested on a scene with 100 characters, each with an anim component & glb animations driving the character model.

Before - onAnimationUpdate ~33.6ms per frame:
image

After - onAnimationUpdate ~8.7ms per frame:
image

### DIFF
--- a/src/framework/components/anim/binder.js
+++ b/src/framework/components/anim/binder.js
@@ -77,49 +77,48 @@ Object.assign(AnimComponentBinder.prototype, {
         }
         return currEntity._parent.findByPath(entityHierarchy.join('/'));
     },
+    _setComponentProperty(propertyComponent, propertyHierarchy) {
+        // trigger the component property setter function
+        propertyComponent[propertyHierarchy[0]] = propertyComponent[propertyHierarchy[0]];
+    },
 
     _floatSetter: function (propertyComponent, propertyHierarchy) {
         var propertyParent = this._getProperty(propertyComponent, propertyHierarchy, true);
         var propertyKey = propertyHierarchy[propertyHierarchy.length - 1];
-        var propertyParentKey = propertyHierarchy[propertyHierarchy.length - 2];
         var setter = function (values) {
             propertyParent[propertyKey] = values[0];
-            propertyParent.parent[propertyParentKey] = propertyParent;
+            this._setComponentProperty(propertyComponent, propertyHierarchy);
         };
         return setter.bind(this);
     },
     _booleanSetter: function (propertyComponent, propertyHierarchy) {
         var propertyParent = this._getProperty(propertyComponent, propertyHierarchy, true);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/framework/components/anim/binder.js [hunk 1]
```diff
@@ -91,31 +91,32 @@ Object.assign(AnimComponentBinder.prototype, {
         return setter.bind(this);
     },
     _colorSetter: function (propertyComponent, propertyHierarchy) {
-        var colorKeys = ['r', 'g', 'b', 'a'];
+        var color = this._getProperty(propertyComponent, propertyHierarchy);
```
slimbuck, line 94: might be worth also caching object and property in the float and boolean case?

ellthompson: 👍  I've went and updated the boolean / float setters.

src/framework/components/anim/binder.js
```diff
@@ -77,55 +77,62 @@ Object.assign(AnimComponentBinder.prototype, {
         }
         return currEntity._parent.findByPath(entityHierarchy.join('/'));
     },
+    _setComponentProperty(propertyComponent, propertyHierarchy) {
+        // trigger the component property setter function
+        propertyComponent[propertyHierarchy[0]] = propertyComponent[propertyHierarchy[0]];
```
slimbuck, line 82: Would you mind explaining why this is needed?

willeastcott (approved):

mvaligursky:
great job for speeding this up!

PR #2359, opened 2020-08-17
Specify basis URL config

slimbuck:
It would be better for the playcanvas viewer to download and instantiate the basis wasm module on demand rather than always doing so at startup.

This PR adds the option of specifying a basis wasm module URL config ahead of time, ready for use when a basis file is encountered and the module is initialised on demand.

Also, log a warning message if download is requested but no URL config is found.

### DIFF
--- a/src/resources/basis.js
+++ b/src/resources/basis.js
@@ -480,7 +480,7 @@ function basisDownloadFromConfig(callback) {
                       downloadConfig.fallbackUrl,
                       callback);
     } else {
-        // grab config from global PC config structure
+        // get config from global PC config structure
         var modules = (window.config ? window.config.wasmModules : window.PRELOAD_MODULES) || [];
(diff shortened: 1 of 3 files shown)


src/resources/basis.js
```diff
@@ -494,4 +517,11 @@ function basisTranscode(url, data, callback, options) {
     }
 }
 
-export { basisDownload, basisDownloadFromConfig, basisInitialize, basisTargetFormat, basisTranscode };
+export {
+    basisDownload,
+    basisSetDownloadConfig,
```
willeastcott, line 522: I made this list alphabetical. And you BROKE it! 😢

slimbuck: Lol! Without thinking, I just assumed the order here would match the order they appear in the file. I'll fix it.

PR #2373, opened 2020-08-23
Add documentation for shader parameter methods in pc.MeshInstance

aidinabedi:
PR adds jsdoc documentation for the following:
- `pc.MeshInstance#getParameter`
- `pc.MeshInstance#setParameter`
- `pc.MeshInstance#setParameters`
- `pc.MeshInstance#deleteParameter`

These methods are analogous to the similar (`xxxParameter`) methods in `pc.Material` which are already documented.

### DIFF
--- a/src/scene/mesh-instance.js
+++ b/src/scene/mesh-instance.js
@@ -411,10 +411,25 @@ Object.assign(MeshInstance.prototype, {
         return this.parameters;
     },
 
+    /**
+     * @function
+     * @name pc.MeshInstance#getParameter
+     * @description Retrieves the specified shader parameter from a mesh instance.
+     * @param {string} name - The name of the parameter to query.
+     * @returns {object} The named parameter.
+     */
     getParameter: function (name) {
         return this.parameters[name];
     },
 
+    /**
+     * @function
+     * @name pc.MeshInstance#setParameter
+     * @description Sets a shader parameter on a mesh instance.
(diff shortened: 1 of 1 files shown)


mvaligursky:
I'd prefer to remove the documentation for pc.MeshInstance#setParameters. The function is not used internally in the engine, similar code executes directly here:

https://github.com/playcanvas/engine/blob/5ad21854a054bc3708f438a2a5b5b762ad65904e/src/scene/forward-renderer.js#L1877-L1886

the same applies to Material.setParameters - I think I'll remove it.

Also, could you please make it clear in comment of MeshInstance#setParameter that the parameter overrides parameter of the same name set on Material.

aidinabedi:
> I'd prefer to remove the documentation for pc.MeshInstance#setParameters. The function is not used internally in the engine, similar code executes directly here:

No problem.

> Also, could you please make it clear in description of MeshInstance#setParameter that the parameter overrides parameter of the same name set on Material.

Alright, will fix.

mvaligursky:
Hi @aidinabedi - I'll close this PR. I'm just making changes to these functions (removing passFlags parameter from both material and meshinstance), and will document them when done. Thanks for the contribution though, it is appreciated.

mvaligursky:
here is the PR that replaces this one.
https://github.com/playcanvas/engine/pull/2391

PR #2375, opened 2020-08-23
Fix normals when mesh has both two-sided material and negative scale

### LINKED ISSUE #2374: Incorrect normals when mesh has both two-sided material and negative scale
### Description

Currently, PlayCanvas uses `gl_FrontFacing` to support two-sided materials or a mesh with negative scale. Unfortunately, if both are combined the `gl_FrontFacing` method will fail, see below, and lighting is calculated with a incorrect normal.

### Steps to Reproduce

1. Open project: https://playcanvas.com/project/712754/overview/twosided-negativescale-test
2. Open scene `Main` and launch it.

### What I expected to see
image

### What I see instead
image

aidinabedi:
Fixes #2374

### DIFF
--- a/src/graphics/program-lib/programs/standard.js
+++ b/src/graphics/program-lib/programs/standard.js
@@ -1316,15 +1316,20 @@ var standard = {
         code = this._fsAddStartCode(code, device, chunks, options);
 
         if (needsNormal) {
+            code += "   getViewDir();\n";
+        }
+
+        if (needsNormal) {
             if (options.twoSidedLighting) {
-                code += "   dVertexNormalW = gl_FrontFacing ? vNormalW : -vNormalW;\n";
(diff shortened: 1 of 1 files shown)


raytranuk (approved):
Looks fine - this is a good way to deal with negative scale flipping winding order - by not using winding order at all. 

NB. There may be related issue where back face culling is wanted, but front faces will be culled if negative scale is used (I will log an issue)

raytranuk:
> NB. There may be related issue where back face culling is wanted, but front faces will be culled if negative scale is used (I will log an issue)

Just checked and there is no culling issue because its already dealt with by flipping the culling mode if the scale is negative - see: https://github.com/playcanvas/engine/blob/master/src/scene/forward-renderer.js#L1612

raytranuk:
Accidentally closed this!

PR #2384, opened 2020-08-25
set entity._template to true for template root

zachpeterpaul:
This PR adds a private boolean field on Entity called _template. 
It can be used to differentiate between the entities of a template root instance,
used internally by the template instantiation mechanism, and the template instances
created by users at runtime (their entities have this flag set to false).
Template instances are created by cloning the root instance.


zachpeterpaul:
It is set on all entities of a template, I don’t think it can have other children at parse time

mvaligursky (approved):

Issue #2061, opened 2020-05-07
Support for glTF extension KHR_lights_punctual


godiagonal:
### Description
[KHR_lights_punctual](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_lights_punctual) is a glTF 2.0 extension supported by most major WebGL libraries. We propose that the PlayCanvas built-in glTF parser should support this as it is the current de-facto standard for storing light source information in glTF.
​
### Step to reproduce
1. Open project https://playcanvas.com/project/684177/overview/khr_lights_punctual-test
1. Open scene “PointLightTest” and launch it
1. Open scene “SpotLightTest” and launch it
​
### What I expect to see and what I saw instead
Note that the mocked lights for the “Expected” results are an approximation and might not match exactly what a proper implementation would yield.

Point lights:
![Screenshot 2020-05-07 at 09 39 26](https://user-images.githubusercontent.com/4249058/81267731-37b66700-9047-11ea-852d-4969063f7b1f.png)

Spot lights:
![Screenshot 2020-05-07 at 09 39 13](https://user-images.githubusercontent.com/4249058/81267741-3ab15780-9047-11ea-9fe3-2ae56363bb5e.png)

willeastcott:
Blocked on #2063.

cx20:
I too am waiting for `KHR_lights_punctual` to be supported. It would be great to have support for the extensions listed below.

https://github.com/cx20/gltf-test#extension-test-models
![image](https://user-images.githubusercontent.com/1833979/81282569-4e37df00-9096-11ea-9ddb-5bfc0ff64b4c.png)

Test model:
https://github.com/cx20/gltf-test/tree/master/tutorialModels/Lights

cx20:
I have confirmed that this extension is available in the latest version of PlayCanvas and updated the status of [gltf-test](https://github.com/cx20/gltf-test#extension-test-models).

[PlayCnavas v1.44.0-dev + Light.gltf result:](https://cx20.github.io/gltf-test/examples/playcanvas/index.html?category=tutorialModels&model=Lights&scale=1.0&type=glTF)
![image](https://user-images.githubusercontent.com/1833979/123293919-58919280-d54f-11eb-9b17-7f071285e398.png)

![image](https://user-images.githubusercontent.com/1833979/123293336-dc974a80-d54e-11eb-8460-6c54b20e2a0d.png)

willeastcott:
[...]

PR #2385, opened 2020-08-26
Support for glTF scenes, cameras and lights

### LINKED ISSUE #2062: Support for glTF cameras
### Description
Currently, when loading a glTF which contains one or more camera objects ([as supported in the specification](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#cameras)) no actual camera is added to the scene. This is an important standard glTF 2.0 feature that we need (and is supported by most other major WebGL libraries).
​
### Steps to reproduce
1. Open project https://playcanvas.com/project/684225/overview/gltf-cameras-test
2. Open scene “CameraTest” and launch it
​
### What I expect to see and what I saw instead
Screenshot 2020-05-07 at 11 16 35

### LINKED ISSUE #2063: Load glTF scenes as entity hierarchies
Currently, glTF scenes are loaded as `pc.Model` objects. A `pc.Model` has a hierarchy made up of `pc.GraphNode` objects. A `pc.Entity` is a sub-class of `pc.GraphNode`, the main difference being that an entity can have components but a graph node cannot. Until we make this change, we cannot fix issues #2061 and #2062. Note that this issue has significant overlap with #1035.

In the Editor, this would mean importing 3D scenes as template assets rather than model assets.

godiagonal:
Related to #2063
Fixes #2062 #2061 

These are customizations of the GLB parser that we (Animech) have had to make to satisfy the requirements of one of our ongoing projects. I noticed there are other PRs related to #2063 that might be more in line with how you want the parser to work, but I thought I would share our implementation anyway since it has been working well in our use case.

## Features
* Support for scenes
  * Adds props `pc.ContainerResource.scenes`, `pc.ContainerResource.scene` and `pc.ContainerResource.nodes`.
  * Scenes and nodes are exported as `pc.Entity` hierarchies.
  * The `scenes` and `scene` props of the glTF are used to create the exported scenes.
  * Can be processed using asset option callbacks `preprocess`, `process` and `postprocess`.
* Models are created per glTF `mesh` instead of just one per glTF.
  * `pc.ModelComponent` is automatically added to all nodes referencing a mesh.
  * Adds `pc.ContainerResource.models`, a list of `pc.Model` assets per glTF mesh. Useful if using glTF extensions that 
reference meshes.
  * Removes `pc.ContainerResource.scenes.model`.
* Support for cameras
  * `pc.CameraComponent` is automatically added to all nodes referencing a camera.
  * Adds `pc.ContainerResource.cameras`, a list of all `pc.CameraComponent` instances. Does not match index of the glTF `cameras` array.
  * Can be processed using asset option callbacks `preprocess`, `process` and `postprocess`.
* Support for `KHR_lights_punctual`
  * `pc.LightComponent` is automatically added to all nodes with extension `KHR_lights_punctual`.
  * Adds `pc.ContainerResource.lights`, a list of all `pc.LightComponent` instances. Does not match index of the glTF `lights` array.
  * Can be processed using asset option callbacks `preprocess`, `process` and `postprocess`.
* Animations grouped by node
  * Adds `pc.ContainerResource.animationIndicesByNode`, a mapping of animation indices to nodes. This can be used to add an animation component of choice (`pc.AnimComponent` or `pc.AnimationComponent`) to nodes that should be animated. The animations must be assigned to the referenced nodes to work, since the animation tracks use relative paths to target child nodes.
* Updates JSDoc for `pc.ContainerResource`.

## Rationale
Exporting scenes rather than a model better correspond to the glTF 2.0 spec which treats glTF as a scene format. It makes it possible to populate the scene with models, cameras, lights etc in the parser.

## Other changes
For the new GLB parser implementation to work I also had to make the following changes.
* Updated `pc.Model.clone()` so that references to external bones (nodes outside the internal model graph) of skin instances are kept in the cloned model.

## Test instructions
Launch [this project](https://playcanvas.com/editor/scene/976797) using a custom build of the engine from this branch. Use [this launch url](https://launch.playcanvas.com/976797?debug=true&use_local_engine=https://launch.playcanvas.com/api/assets/files/playcanvas.js?id=34573773&branchId=a1fc2544-c925-470b-b5c8-d17e5cd4553c).

The project showcases a glTF asset containing a single scene with animations, cameras and lights being loaded by this version of the GLB parser.


mvaligursky:
Thanks, we definitely were keen to add this and more. We have multiple changes in flight in the glb-parser related to this as well, so this might take a while to get merged in / worked on.

willeastcott:
@mvaligursky With the progress we've made importing models as asset templates, are we now in a position to import cameras and lights?

mvaligursky:
We can import cameras and lights from fbx files in the editor - which gets converted into a glb file for meshes, and a template file for hierarchy with render / light / camera components.
But once the render component goes public, we should still consider updating this PR to use render component (instead of model component) and use it to allow us to load hierarchies of entities with cameras and lights directly in the engine.

willeastcott:
Yeah, I'm thinking more of creating `CameraComponent`s and `LightComponent`s in asset templates on import. If that's the route we're going (and presumably we can get there soon), can this PR be closed?

Maksims:
This definitely needs Editor work with UX to define import settings, as there will be cases where creating lights or cameras won't be desirable, and vice versa.

mvaligursky:
Current implementation for fbx-> glb+ templates imports everything. Import options is discussed a lot and we'd like to implement it for many reasons. But in this specific case, templates kind of handle it for us. If you for example import fbx, it generates gbl with geometry only, and hierarchy with components create a template. So you can place it into the scene, have Entity hierarchy with components referencing meshes, lights, cameras ... If you don't want some / all lights, you simply delete them from the template. When you drag & drop newly exported fbx with say changes to meshes, it won't bring back your lights, as template has 'remove these entities with lights' overrides on them. So the end result is what you want.

But that does not decrease the need for import settings in the future for many other reasons.

mvaligursky:
@willeastcott - we have good new "fbx -> gbl + template" pipeline (not released to public yet), for which this PR is not needed.

But it'd be still useful to allow external glb files to be loaded (they will not have template created), and be able to use cameras and lights, and even render component. So this PR would be useful for that, even thought a large part of it will not be needed anymore or will need to change. I'd like to keep it for reference of light / camera support from glb though.

mvaligursky:
Closing this, as the engine implementation of these has already been done now (based on this PR, thanks so much for making it @godiagonal), or any remaining work has been captured in the issues.

PR #2389, opened 2020-08-26
MeshInstance frustum culling - small refactor

mvaligursky:
- small refactor - moving _isVisible from ForwardRenderer to MeshInstance, where it belongs
- small optimization in BoundingBox.setFromTransformedAabb - saves about 5% in my test.

### DIFF
--- a/src/scene/mesh-instance.js
+++ b/src/scene/mesh-instance.js
@@ -1,4 +1,5 @@
 import { BoundingBox } from '../shape/bounding-box.js';
+import { BoundingSphere } from '../shape/bounding-sphere.js';
 
 import {
     BLEND_NONE, BLEND_NORMAL,
@@ -13,7 +14,7 @@ import {
 
 var _tmpAabb = new BoundingBox();
 var _tempBoneAabb = new BoundingBox();
-var _tempSphere = { center: null, radius: 0 };
+var _tempSphere = new BoundingSphere(null, 0);
 
 /**
  * @class
--- a/src/scene/mesh-instance.js
+++ b/src/scene/mesh-instance.js
@@ -13,6 +13,7 @@ import {
 
 var _tmpAabb = new BoundingBox();
 var _tempBoneAabb = new BoundingBox();
+var _tempSphere = { center: null, radius: 0 };
 
 /**
  * @class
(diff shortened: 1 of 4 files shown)


src/scene/mesh-instance.js
```diff
@@ -372,6 +373,32 @@ Object.assign(MeshInstance.prototype, {
         // Deprecated
     },
 
+    // test if meshInstance is visible by camera. It requires the frustum of the camera to be up to date, which forward-renderer
+    // takes care of. This function should  not be called elsewhere.
+    _isVisible: function (camera) {
```
willeastcott, line 378: Just playing devil's advocate here, but I'm wondering if this function should take a `pc.Frustum` instead of a camera.

src/scene/mesh-instance.js
```diff
@@ -13,6 +13,7 @@ import {
 
 var _tmpAabb = new BoundingBox();
 var _tempBoneAabb = new BoundingBox();
+var _tempSphere = { center: null, radius: 0 };
```
willeastcott, line 16: I would make this a proper `pc.BoundingSphere` for clarity. If this was Typescript, the compiler would complain. 😄

willeastcott:
I love seeing this kind of incremental tidy up, @mvaligursky! 🎉

src/scene/mesh-instance.js
mvaligursky: the custom visibility function accepts camera, so that would change the pubic API
this.isVisibleFunc(camera)

mvaligursky: or at least required larger changes

src/scene/mesh-instance.js
mvaligursky: done, makes sense

src/scene/mesh-instance.js
willeastcott: Ooo. Good point. But just to be clear, `pc.Camera` is *not* (currently) public API. And neither is `isVisibleFunc` AFAIK.

mvaligursky: hey good to know. I looked at how the isVisibleFunc is used by elements, and in 
isVisibleForCamera: function (camera)
they use other properties from camera than the frustum, so I don't think this can change

willeastcott: Fair enough. We *might* have to make `pc.Camera` public at some point then. Or alternatively merge `pc.Camera` with `pc.CameraComponent`.

src/scene/mesh-instance.js
```diff
@@ -13,6 +14,7 @@ import {
 
 var _tmpAabb = new BoundingBox();
 var _tempBoneAabb = new BoundingBox();
+var _tempSphere = new BoundingSphere(null, 0);
```
willeastcott, line 17: No point in passing these parameters - I wouldn't bother specifying them.

mvaligursky: excelent point!

willeastcott (approved):

PR #2391, opened 2020-08-27
[Fix] Shader parameters set on Material or MeshInstance are used in all passes

mvaligursky:
In previous implementation, a passFlag parameter would be used to control which passes (depth, main camera ...) a parameter would apply to, and by default it would not apply parameters in shadow passes, making shaders changing positions of vertices not behave correctly.
I've change this behavior and now all parameters are available in all passes - if shader needs it, it will have access to up to date data. The cost for this is negligible, only a single assignment and increment executes in addition to existing dictionary search.

This PR also removes some duplicate functions from forward-renderer.

### DIFF
--- a/src/framework/components/sprite/component.js
+++ b/src/framework/components/sprite/component.js
@@ -379,10 +379,9 @@ Object.assign(SpriteComponent.prototype, {
 
             // update outer scale
             if (this._meshInstance) {
-                // use outerScale in ALL passes (depth, picker, etc) so the shape is correct
                 this._outerScaleUniform[0] = this._outerScale.x;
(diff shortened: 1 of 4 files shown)


willeastcott:
Can you test After the Flood with this PR before we merge please?

mvaligursky:
> Looks great! A net reduction in code is always good. 😄 Can you test After the Flood with this PR before we merge please?

I tried and it works fine .. it forces parameters for all passes for tree rendering, so that won't be needed when this is merged in, but those are ignored.

willeastcott (approved):

PR #2397, opened 2020-08-28
WebXR optionalFeatures options and change to `start` method

### LINKED ISSUE #2158: No way to pass optional features into a WebXR session
### Description

Using WebXR on Chrome, the DOM isn't available or shown at all. Since Chrome 83, a new optional feature has been included to include a DOM overlay, e.g (from [https://immersive-web.github.io/webxr-samples/ar-barebones.html](https://immersive-web.github.io/webxr-samples/ar-barebones.html)):

```
navigator.xr.requestSession('immersive-ar', {
    optionalFeatures: ['dom-overlay'],
    domOverlay: {root: document.getElementById('overlay')}
}).then(onSessionStarted, onRequestSessionError);
```

However looking through the engine code for XrManager, there's no way to pass in additional data: 
```
var optionalFeatures = [];

if (type === pc.XRTYPE_AR)
    optionalFeatures.push('light-estimation');

navigator.xr.requestSession(type, {
    requiredFeatures: [spaceType],
    optionalFeatures: optionalFeatures
}).then(function (session) {
```
Could there be a way to pass in the optionalFeatures? Perhaps in entity.camera.startXr?

Maksims:
Fixes #2158

And changed the way `start` method works. Old way still works.

**Old:**
```js
app.xr.start(camera, pc.XRTYPE_VR, pc.XRSPACE_LOCAL, function() { });
```
**New:**
```js
app.xr.start(camera, pc.XRTYPE_VR, pc.XRSPACE_LOCAL, {
    callback: function() { }
});
```

Now we have an options object to provide additional options, such as: callback and `optionalFeatures` to requestSession, this provides ability for developers to access experimental WebXR implementations in various browsers, without need to wait till it is integrated into engine as a first-class API.

Here is an example for enabling [WebXR Layers API](https://immersive-web.github.io/layers/):

```js
app.xr.start(camera, pc.XRTYPE_VR, pc.XRSPACE_LOCAL, {
    optionalFeatures: [ 'layers' ]
});
```

### DIFF
--- a/examples/xr/ar-basic.html
+++ b/examples/xr/ar-basic.html
@@ -93,8 +93,12 @@
             if (app.xr.supported) {
                 var activate = function () {
                     if (app.xr.isAvailable(pc.XRTYPE_AR)) {
-                        c.camera.startXr(pc.XRTYPE_AR, pc.XRSPACE_LOCALFLOOR, function (err) {
-                            if (err) message("WebXR Immersive AR failed to start: " + err.message);
+                        c.camera.startXr({
+                            type: pc.XRTYPE_AR,
+                            spaceType: pc.XRSPACE_LOCALFLOOR,
+                            callback: function (err) {
+                                if (err) message("WebXR Immersive AR failed to start: " + err.message);
+                            }
                         });
                     } else {
                         message("Immersive AR is not available");
--- a/examples/xr/ar-basic.html
+++ b/examples/xr/ar-basic.html
@@ -93,9 +93,7 @@
             if (app.xr.supported) {
                 var activate = function () {
                     if (app.xr.isAvailable(pc.XRTYPE_AR)) {
-                        c.camera.startXr({
-                            type: pc.XRTYPE_AR,
-                            spaceType: pc.XRSPACE_LOCALFLOOR,
+                        c.camera.startXr(pc.XRTYPE_AR, pc.XRSPACE_LOCALFLOOR, {
                             callback: function (err) {
                                 if (err) message("WebXR Immersive AR failed to start: " + err.message);
                             }
--- a/examples/xr/ar-hit-test.html
+++ b/examples/xr/ar-hit-test.html
@@ -81,9 +81,7 @@
             if (app.xr.supported) {
                 var activate = function () {
                     if (app.xr.isAvailable(pc.XRTYPE_AR)) {
-                        c.camera.startXr({
-                            type: pc.XRTYPE_AR,
-                            spaceType: pc.XRSPACE_LOCALFLOOR,
+                        c.camera.startXr(pc.XRTYPE_AR, pc.XRSPACE_LOCALFLOOR, {
                             callback: function (err) {
                                 if (err) message("WebXR Immersive AR failed to start: " + err.message);
                             }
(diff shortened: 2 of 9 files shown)


willeastcott:
Normally, I author functions to have required parameters as individually specified, and then collect optional parameters in a final options object. In this case, the `camera` parameter is always required, right? I guess `type` and `spaceType` could be optional (and maybe default to `pc.XRTYPE_VR` and `pc.XRSPACE_LOCAL` say). So wouldn't a better function prototype be:

````
app.xr.start(camera); // Start XR with VR and Local Space

// Start AR with hit test support
app.xr.start(camera, {
    type: pc.XRTYPE_AR,
    optionalFeatures: [  'hit-test' ]
});
```

yaustar:
Having optional features as individual properties means that developers will still be reliant on the engine releases to support any new WebXR modes or features unfortunately.

Maybe we can create some constants for these optional features to make it easier? Or have preset constants for the most common applications?

Maksims:
> In this case, the `camera` parameter is always required, right?

Sure we can. So `CameraComponent.startXr()` - could be just like so?
Although not specifying Type and Space, can lead to assumtions. First one is AR is as valid XR type as VR. And reference space, for VR it is most of the time either LOCAL and I can expect UNBOUNDED to be popular in future as well. But for AR it is LOCALFLOOR most common. So there is no specific default here. And there is deliberate incentive to teach developer of different reference spaces as a lot of XR logic derives from it.

**Should I change to: `start(camera, options);`? While I still believe Type and Space to be mandatory.**

> Also, what about having optional features as individual properties:

Currently, there is no overhead of enabling features by default, such as Hit Test when AR session is started. I will expect same for many other APIs. So developer might get confused why some features are not working, while they are behind the flag (unnecessary).
If you've mentioned it as instead of `optionalFeatures`, then reasoning is this: we do not know what platform is supporting, in terms of `optionalFeatures` flag. If we develop first-class API for some API, for example hand tracking or light estimation, then we inject to `optionalFeatures` inside engine. But some APIs, like Layers, or any other future ones, are not implemented, so it is up to developer to add them as `optionalFeatures` flags. We basically don't know what APIs might be added there, and I would avoid adding toggles for every possible APIs as this space is constantly changing.

Maksims:
> Having optional features as individual properties means that developers will still be reliant on the engine releases to support any new WebXR modes or features unfortunately.

This is exactly why we need simple `optionalFeatures` array, to provide developers with access to experimentations. And not hide it behind engine releases.

willeastcott:
> Should I change to: start(camera, options);? While I still believe Type and Space to be mandatory.

Well, if type and space are also mandatory, it should be:

```
start(camera, type, space, options);
```

Maksims:
> Well, if type and space are also mandatory, it should be:
> 
> ```js
> start(camera, type, space, options);
> ```

Agree, updated PR.

Regarding `features`, it is intentionally `optionalFeatures`, as there are two lists: optionalFeatures and requiredFeatures, later we don't expose as it will fail starting session if such feature is not supported. So we rather incentivize developers to add features progressively or fallback gracefully instead of crashing. When developer searches our documentation with specifically `optionalFeatures` (as it is in WebXR Specs), they will find what they need.

yaustar (approved):

PR #2419, opened 2020-09-09
Warning added when element input is enabled, but no pc.ElementInput is created to fire input events

raytranuk:
Warning added when element input is enabled, but no pc.ElementInput is created to fire input events

This will greatly help debugging a problem I wasted hours on - where element input event were not firing without any indication of why.

### DIFF
--- a/src/framework/components/element/component.js
+++ b/src/framework/components/element/component.js
@@ -1445,8 +1445,8 @@ Object.defineProperty(ElementComponent.prototype, "useInput", {
                 this.system.app.elementInput.removeElement(this);
             }
         } else {
-            if (this._useInput === true){
-                console.warn("Elements will not get any input events because this.system.app.elementInput is not initialized")
+            if (this._useInput === true) {
+                console.warn("Elements will not get any input events because this.system.app.elementInput is not initialized");
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/framework/components/element/component.js
```diff
@@ -1444,6 +1444,10 @@ Object.defineProperty(ElementComponent.prototype, "useInput", {
             } else {
                 this.system.app.elementInput.removeElement(this);
             }
+        } else {
+            if (this._useInput === true){
```
yaustar, line 1448: While the CI is complaining, space before the `{` :P

src/framework/components/element/component.js
```diff
@@ -1444,6 +1444,10 @@ Object.defineProperty(ElementComponent.prototype, "useInput", {
...
+                console.warn("Elements will not get any input events because this.system.app.elementInput is not initialized")
```
yaustar, line 1449: Should it be 'not created/null rather than 'not initialized'. The latter sounds like a function needs to be called on `this.system.app.elementInput` to initialize it.

yaustar (approved):
Looks good but needs the lint fixes :)

src/framework/components/element/component.js
raytranuk: I will change it

mvaligursky (approved):

PR #2429, opened 2020-09-15
Refraction engine example and small related engine fix

mvaligursky:
- fix to grabTexture functionality in the engine: WebGl supports RGBA framebuffers and not RGB, so use the same format for the grab texture
- engine example demonstrating grabTexture feature to render glass distortion

Screen Shot 2020-09-15 at 10 59 44 AM

Link to demo (when released): http://playcanvas.github.io/#graphics/grab-pass.html


examples/examples.js
```diff
@@ -43,6 +43,7 @@ var categories = [
             "portal",
             "post-effects",
             "render-to-texture",
+            "refraction",
```
willeastcott, line 46: So you can achieve refraction with a regular standard material using a box-projected cubemap. This is maybe better named as 'dynamic-refraction' or 'grab-pass' IMHO. And if I picked on of those, I'd go with 'grab-pass'.

mvaligursky: done, cheers.

slimbuck (approved):

PR #2434, opened 2020-09-18
[AnimComponent] State transitions fix

ellthompson:
- Transitions currently set new anim clips to a weight of zero for the first frame which sets all newly animated bones to zero vectors. This PR increases the initial weight of new anim clips by the current deltaTime normalised by the transition duration. This ensures all newly animated bones are at least partially transformed by the anim clip.

- Also enables the previous state to continue animating during a transition, rather than taking its pose.

### DIFF
--- a/src/framework/components/anim/controller.js
+++ b/src/framework/components/anim/controller.js
@@ -485,7 +485,9 @@ Object.assign(AnimController.prototype, {
             var interpolatedTime = this._currTransitionTime / this._totalTransitionTime;
             for (i = 0; i < this._transitionPreviousStates.length; i++) {
                 // interpolate the weights of the most recent previous state and all other previous states based on the progress through the previous transition
-                if (i !== this._transitionPreviousStates.length - 1) {
+                if (!this._isTransitioning) {
+                    this._transitionPreviousStates[i].weight = 1.0;
+                } else if (i !== this._transitionPreviousStates.length - 1) {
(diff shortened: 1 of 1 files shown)


src/framework/components/anim/controller.js
```diff
@@ -485,7 +485,9 @@ Object.assign(AnimController.prototype, {
             var interpolatedTime = this._currTransitionTime / this._totalTransitionTime;
             for (i = 0; i < this._transitionPreviousStates.length; i++) {
                 // interpolate the weights of the most recent previous state and all other previous states based on the progress through the previous transition
-                if (i !== this._transitionPreviousStates.length - 1) {
+                if (!this._isTransitioning) {
+                    this._transitionPreviousStates[i].weight = 1.0;
```
willeastcott, line 489: Just curious, but why use 1.0 instead of 1? In JS, aren't they identical?

mvaligursky (approved):

src/framework/components/anim/controller.js
ellthompson: I've got into the habit of distinguishing ints and floats so I can notice what that var is being used as. They are identical.

willeastcott: Fair enough. It's not a big deal. 😄

willeastcott (approved):

PR #2435, opened 2020-09-18
Optimized culling solution for skinned meshes

### LINKED ISSUE #2352: Optimize visibility calculation for skinned characters
Visibility calculation for characters has a very high cost. It calls MeshInstance.aabb, which transforms bounding boxes of each bone into world space, ads them up, computes final  bounding sphere and uses this for frustum check.

In my test case with 100 characters of 70 bones each, this takes about 8ms on the MacBook Pro and it is about 10% of the frame.

Would could perhaps at init time evaluate the largest possible bounding box / let user specify it, and use this for early / the only culling.

Alternatively, instead of computing total bounding spheres for each bones, which has no early out, we could cull per bone, which would early out for any visible bone. Or test visibility of root bone of characters at least, before computing total bounding volume for bones.

mvaligursky:
Fixes #2352

new API:
**ModelComponent.aabb** - allows oversized bounds to be specified, and turns off their update if not null

Performance impact - tested in a scene with 100 characters, about 100 bones each:
- culling. Originally took **9ms** to update aabb for characters. With oversized aabb specified, this drops to **1.6ms**.
- bone updates. Bone matrices are now processed (matrices multiplied, bones prepared for GPU consumption) only when skinned mesh is visible, instead of always. When all 100 characters are invisible, this saved about **8ms** in the test scene.

Integration to Editor still TBD.

### DIFF
--- a/src/framework/components/model/component.js
+++ b/src/framework/components/model/component.js
@@ -47,6 +47,8 @@ import { Component } from '../component.js';
  * @property {boolean} lightmapped If true, this model will be lightmapped after using lightmapper.bake().
  * @property {number} lightmapSizeMultiplier Lightmap resolution multiplier.
  * @property {boolean} isStatic Mark model as non-movable (optimization).
+ * @property {pc.BoundingBox} aabb If set, the bounding box is used as a bounding box for visibility culling of attached mesh instances. This is an optimization,
+ * allowing oversized bounding box to be specified for skinned characters in order to avoid per frame bounding box computations based on bone positions.
  * @property {pc.MeshInstance[]} meshInstances An array of meshInstances contained in the component's model. If model is not set or loaded for component it will return null.
  * @property {number} batchGroupId Assign model to a specific batch group (see {@link pc.BatchGroup}). Default value is -1 (no group).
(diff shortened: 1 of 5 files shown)


src/framework/components/model/component.js
```diff
@@ -47,6 +47,8 @@ import { Component } from '../component.js';
  * @property {boolean} lightmapped If true, this model will be lightmapped after using lightmapper.bake().
  * @property {number} lightmapSizeMultiplier Lightmap resolution multiplier.
  * @property {boolean} isStatic Mark model as non-movable (optimization).
+ * @property {pc.BoundingBox} aabb If set, the bounding box is used as a bounding box for visibility culling of attached mesh instances. This is an optimization,
```
slimbuck, line 50: might be clearer if this is also called overrideAabb?

mvaligursky: I had it originally called that, but decided just aabb is more suitable

willeastcott: Yeah, I'd probably agree. My order of preference would be:

1. `aabb`
2. `customAabb`
3. `overrideAabb`

slimbuck (approved):

PR #2445, opened 2020-09-24
Use buffers setup in draw call for left eye when drawing right eye

### LINKED ISSUE #2444: Instanced meshes only rendered in one eye in WebVR
### Description

Instanced meshes only render in one eye in WebVR. The shadows render in both. An example project is included here. https://playcanvas.com/project/722033/overview/batchingomnivor

### Steps to Reproduce
1.  Connect WebXR compatible headset and open WebXR compatible browser
2. Open https://playcanvas.com/project/722033/overview/batchingomnivor
3. Notice that instanced cubes are visible
4. Click on page to start VR
5. Notice instanced cubes are only visible in one eye

evan-omnivor:
Fixes #2444

The right eye should use the same buffers setup by the left eye. The draw call for the left eye clears the GraphicsDevice's vertex buffer array so calling draw with keepBuffers set to false for the right eye was not drawing all meshes.

### DIFF
--- a/src/scene/forward-renderer.js
+++ b/src/scene/forward-renderer.js
@@ -1308,13 +1308,7 @@ Object.assign(ForwardRenderer.prototype, {
         if (instancingData) {
             if (instancingData.count > 0) {
                 this._instancedDrawCalls++;
-                device.setVertexBuffer(instancingData.vertexBuffer);
-                device.draw(mesh.primitive[style], instancingData.count);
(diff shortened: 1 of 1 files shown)


mvaligursky:
Thanks for the report and the fix, great work. I've create a new PR with slightly different change (want to keep the auto instancing functionality around) .. so I'll close this.

PR #2454, opened 2020-09-30
Add PARTICLE #define to particle shaders

Maksims:
In order to branch shader chunks logic, it is useful to know in shader if it is used for particles or something else.
Use case where I've faced issues is for example writing custom Fog shader chunk. When making for dependant on vPositionW (world position of fragment) in particle shader this variable is not available. So overriding global shader chunk for fog in such case is not possible, or requires extra work to add vPositionW to vertex shaders.

In such case, it is useful to know in shader if it is used for particle or anything else.

This change has no effect on existing project.


raytranuk (approved):
looks useful

mvaligursky (approved):

PR #2459, opened 2020-10-02
[FIX] Minor corrections to Typescript definitions

### LINKED ISSUE #2458: small fixes for typescript definitions
I ran into these two small fixes I needed to do to the generated .d.ts file for the engine:

Add to Application:
`lightmapper: pc.Lightmapper;`

Change the first param of `Lightmapper` back to allow null:
`bake(nodes: pc.Entity[] | null, mode?: number): void;`

Also, the current `tags.add()` only allows a single string parameter, not an array, and not an arbitrary list of params.

(Sorry, I don’t have time to fix this properly and submit a PR, and wanted to drop it here before I forget)

willeastcott:
Fixes #2458 

### DIFF
--- a/src/core/tags.js
+++ b/src/core/tags.js
@@ -193,7 +193,7 @@ Object.assign(Tags.prototype, {
      * @function
      * @name pc.Tags#add
      * @description Add a tag, duplicates are ignored. Can be array or comma separated arguments for multiple tags.
-     * @param {string} name - Name of a tag, or array of tags.
+     * @param {string|string[]} name - Name of a tag, or array of tags.
      * @returns {boolean} True if any tag were added.
      * @example
      * tags.add('level-1');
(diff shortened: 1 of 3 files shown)


src/core/tags.js
```diff
@@ -193,7 +193,7 @@ Object.assign(Tags.prototype, {
      * @function
      * @name pc.Tags#add
      * @description Add a tag, duplicates are ignored. Can be array or comma separated arguments for multiple tags.
-     * @param {string} name - Name of a tag, or array of tags.
+     * @param {string|string[]} name - Name of a tag, or array of tags.
```
blairmacintyre, line 196: does the processor all a "rest" argument, so we can capture the 3rd "any number of strings"?  Would `{string|string[]|...string}` be legal?

mvaligursky (approved):

src/core/tags.js
willeastcott: Gosh, this is a tricky one. I can add `...string` which generates `...name: string[]`. But doing `...string|string[]` outputs `name: string[]` which is not what we want. So I've heavily relaxed the types to have `...*` which does work for all possible params (although of course does allow for other types like booleans and numbers say. If you have a better approach, let me know, otherwise, we can just merge this.

PR #2462, opened 2020-10-05
Fixed First Person Jump

### LINKED ISSUE #2461: Weird jump behaviour in First Person Camera example
### Description

In the First Person Camera example, when the player jumps and falls, it goes up normally and falls with a normal behaviour in the first half of the fall but as it approaches the ground it slows down and you can can see it falls really slowly until it touches the ground. This happens every time you jump.

https://playcanvas.github.io/#camera/first-person.html

GamerWael:
When the player jumps and falls, it goes up normally and falls with a normal behaviour in the first half of the fall but as it approaches the ground it slows down and you can can see it falls really slowly until it touches the ground. This happens every time you jump.

This commit fixes that.

Fixes #2461 

### DIFF
--- a/examples/assets/scripts/camera/first-person-camera.js
+++ b/examples/assets/scripts/camera/first-person-camera.js
@@ -8,7 +8,7 @@ CharacterController.attributes.add('jumpImpulse', { type: 'number', default: 400
 
 // initialize code called once per entity
 CharacterController.prototype.initialize = function () {
-    this.groundCheckRay = new pc.Vec3(0, -1.2, 0);
+    this.groundCheckRay = new pc.Vec3(0, -0.5, 0);
     this.rayEnd = new pc.Vec3();
 
     this.groundNormal = new pc.Vec3();


mvaligursky (commented):
should this maybe be more generic and based on some size of bounding box or the character or some parameter the user can change?

willeastcott:
Unfortunately, this change doesn't fix the problem. The character controller is supposed to be used with a capsule collision shape. A default capsule is 2m high. The ray is 1.2m in length, fired down from the center of the capsule. So 0.2m of the ray exits the bottom of the capsule and if it penetrates the ground, the controller is considered to be on the ground. The whole ray solution here is not great. Because the capsule can be on stairs for example and the ray might not reach the ground if the capsule of hanging over a stair.

All you're doing here is making the ray so short it doesn't even exit the bottom of the controller's collision shape. So the controller is much more likely to get stuck on uneven terrain.

There is a problem here, but this isn't the solution, IMHO.

PR #2473, opened 2020-10-15
First draft removing the markers from text mesh generation

yaustar:
Fixes rendering the LTR and RTL indicators

Not sure if this is the best way to approach this. It feels like that the RTL setup in the project should remove the non-visible markers

### DIFF
--- a/src/framework/components/element/text-element.js
+++ b/src/framework/components/element/text-element.js
@@ -49,7 +49,6 @@ function TextElement(element) {
     // public
     this._text = "";            // the original user-defined text
     this._symbols = [];         // array of symbols with unicode processing and markup removed
-    this._visibleSymbols = [];  // array of symbols that is visible to the user
     this._colorPalette = [];    // per-symbol color palette
     this._symbolColors = null;  // per-symbol color indexes. only set for text with markup.
     this._i18nKey = null;
@@ -156,7 +155,7 @@ var NON_VISIBLE_CHAR = new Set([
     '\u202D', // left-to-right override
     '\u202E', // right-to-left override
     '\u202C', // pop directional formatting
-    '\u2069', // pop directional isolate
+    '\u2069'  // pop directional isolate
 ]);
 
 Object.assign(TextElement.prototype, {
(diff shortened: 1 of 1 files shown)


src/framework/components/element/text-element.js
```diff
@@ -493,6 +505,17 @@ Object.assign(TextElement.prototype, {
...
         }
 
+        // Remove the non visible symbols
+        var inputIndex = 0;
+        for (i = 0; i < this._symbols.length; i++) {
+            if (!NON_VISIBLE_CHAR.has(this._symbols[i])) {
+                this._symbols[inputIndex] = this._symbols[i];
+                inputIndex++;
+            }
+        }
```
willeastcott, line 515: Am I missing something here? The `this._symbols` array is created empty. This `for` loop iterates it but it has zero elements. And elements are only added inside. And `inputIndex` is only non-zero if code inside the loop is executed. I can't see how `this._symbols` is ever anything other than the empty array. This might be a 'brain still not in gear in the morning' moment on my part.

yaustar: `this._symbols` is populated at the top of the function: https://github.com/playcanvas/engine/blob/master/src/framework/components/element/text-element.js#L262 :)

willeastcott: Oh right, so you're effectively throwing away the initially assigned empty array and overwriting the property with a new array. Might as well initially assign it as null perhaps?

yaustar: I've should have explained what the bug was in the PR 😅. My bad!

One of our clients has a RTL localisation text that also has these 'markers' which changes a section of the text be LTR

Unfortunately, these markers get rendered as empty squares:
image

Before this loop executes, `_symbols` contains an array of each character in the text including the markers. This loop is an attempt to remove the markers without allocating a new array and save on garbage collection.

willeastcott: OK. Well, this doesn't look too bad performance wise. Any reasons/concerns that'd make you hesitate to merge this?

slimbuck: actually, I think symbols should be removed before the call to _calculateCharsPerTexture().

also, this will probably break text markup as symbolColors needs updating.

yaustar:
Added symbolColors fix

yaustar:
Closed. Fixed via: https://github.com/playcanvas/engine/pull/2490#pullrequestreview-514797035

Issue #2470, opened 2020-10-13
Should we set picker.layer.passThrough to true ?


scarletsky:
### Description

When we call `Picker#prepare`, it will add opaque and transparent meshInstances into `picker.layer.opaqueMeshInstances`, and then call `ForwardRenderer#renderComposition` to render it.

However, `LayerComposition#_update` is called in `ForwardRenderer#renderComposition`, and this method may **RESET** the `picker.layer.opaqueMeshInstances` and `picker.layer.transparentMeshInstances` in a very complex scenario, such as changing the skybox, changing meshInstance's material etc.

This will cause the renderer do not render the transparent meshInstances in the picker layer, and `Picker#getSelection` will return wrong result when clicking transparent meshInstances.

The related code is here:

```diff
LayerComposition.prototype._update = function () {
    // ...
    if (this._dirtyBlend) {
        // TODO: make it fast
        result |= COMPUPDATED_BLEND;
        var opaqueOld, transparentOld, opaqueNew, transparentNew;
        for (i = 0; i < len; i++) {
            layer = this.layerList[i];
            if (layer.passThrough) continue;
            opaqueOld = layer.opaqueMeshInstances;
            transparentOld = layer.transparentMeshInstances;
            opaqueNew = [];
            transparentNew = [];
            for (j = 0; j < opaqueOld.length; j++) {
                if (opaqueOld[j].material && opaqueOld[j].material.blendType !== BLEND_NONE) {
                    transparentNew.push(opaqueOld[j]);
                } else {
                    opaqueNew.push(opaqueOld[j]);
                }
            }
            for (j = 0; j < transparentOld.length; j++) {
                if (transparentOld[j].material && transparentOld[j].material.blendType !== BLEND_NONE) {
                    transparentNew.push(transparentOld[j]);
                } else {
                    opaqueNew.push(transparentOld[j]);
                }
            }
            layer.opaqueMeshInstances.length = opaqueNew.length;
            for (j = 0; j < opaqueNew.length; j++) {
+               layer.opaqueMeshInstances[j] = opaqueNew[j];
            }
            layer.transparentMeshInstances.length = transparentNew.length;
            for (j = 0; j < transparentNew.length; j++) {
+                layer.transparentMeshInstances[j] = transparentNew[j];
            }
        }
        this._dirtyBlend = false;
    }
}
```

A possible solution is to set `picker.layer.passThrough = true`. And I am not sure whether it will cause other problem too.

[...]

Issue #2468, opened 2020-10-12
Picker#getSelection on empty area will return wrong result if camera's clearColor is [0, 0, 0, 0]


scarletsky:
### Description

When camera's clearColor is `[0, 0, 0, 0]`, `Picker#getSelection` on empty area will always return the first meshInstance in the draw call list.

### Reason

In `Picker#onLayerPreRender`, it will set `camera._clearOptions`:

```diff
Picker.prototype.onLayerPreRender = function (layer, sourceLayer, sourceRt) {
    // ...
+    layer.cameras[0].camera._clearOptions = this.clearOptions;
    // ...
};
```

In `ForwardRenderer#renderComposition ` and `ForwardRenderer#clearView`, it *WILL NOT* use `camera._clearOptions`, and It will use `camera._clearColor` instead:

```diff
// ...
clearView: function (camera, target, clear, forceWrite, options) {
    // ...
    if (clear) {
+        device.clear(options ? options : {
+            color: [camera._clearColor.r, camera._clearColor.g, camera._clearColor.b, camera._clearColor.a],
+            // ...
+        });
    }
}

renderComposition: function (comp) {
    // ...
    if (!(layer._preRenderCalledForCameras & (1 << cameraPass))) {
    // ...
        if (layer.overrideClear) {
+            this.clearView(camera.camera, layer.renderTarget, true, true, layer._clearOptions);
        }
    }
    if (camera) {
        rt = layer.renderTarget;
        if (!processedThisCameraAndRt) {
+            if (!layer.overrideClear) this.clearView(camera.camera, layer.renderTarget, true, true);
              // ...
        }
    }
}
```

As a result, when you click on empty area, the pixel color is `[0, 0, 0, 0]`, and this is the first draw call's pick color.

### Solution

Respect the `camera._clearOptions` in `ForwardRenderer#clearView`:

```diff
clearView: function (camera, target, clear, forceWrite, options) {
    // ...
+   if (!options) {
+        options = camera._clearOptions;
+   }
    // ...
}
```

mvaligursky:
nice investigation! Thanks, we'll try to sort it out soon.

PR #2476, opened 2020-10-16
Improvements and fixes to Picker, new engine example

mvaligursky:
new engine example (link valid after engine release):
http://playcanvas.github.io/#graphics/area-picker.html

Screen Shot 2020-10-16 at 4 47 56 PM

Fixes https://github.com/playcanvas/engine/issues/2470
Fixes https://github.com/playcanvas/engine/issues/2468

Thanks to @scarletsky for contributions!

### DIFF
--- a/src/scene/forward-renderer.js
+++ b/src/scene/forward-renderer.js
@@ -813,6 +813,10 @@ Object.assign(ForwardRenderer.prototype, {
         device.setScissor(x, y, w, h);
 
         if (clear) {
+            // use camera clear options if any
+            if (!options)
+                options = camera._clearOptions;
+
             device.clear(options ? options : {
                 color: [camera._clearColor.r, camera._clearColor.g, camera._clearColor.b, camera._clearColor.a],
                 depth: camera._clearDepth,
--- a/src/scene/pick.js
+++ b/src/scene/pick.js
@@ -47,6 +47,9 @@ function Picker(app, width, height) {
     this.pickColor = new Float32Array(4);
     this.pickColor[3] = 1;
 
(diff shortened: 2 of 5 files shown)


src/scene/forward-renderer.js
```diff
@@ -813,6 +813,10 @@ Object.assign(ForwardRenderer.prototype, {
         device.setScissor(x, y, w, h);
 
         if (clear) {
+            // use camera clear options if any
+            if (!options)
+                options = camera._clearOptions;
```
slimbuck, line 818: what's the difference between camera._clearOptions and all the settings used below (camera._clearColor et al)?

src/scene/pick.js
```diff
@@ -299,6 +308,9 @@ Picker.prototype.prepare = function (camera, scene, arg) {
     // save old camera state
     this.onLayerPreRender(this.layer, sourceLayer, sourceRt);
 
+    // clear registered meshes, rendering will register them against
```
slimbuck, line 311: against->again?

mvaligursky: fixed, thanks!

src/scene/forward-renderer.js
mvaligursky: There are two ways .. one is settings on camera, one is override settings on camera. 
Don't ask ... a lot of this code seems over engineered to me.

slimbuck: Ok I take the question back!

slimbuck (approved):
LGTM and also excited to see the new demo!

PR #2480, opened 2020-10-20
Brush painting engine example using rendering to texture

mvaligursky:
engine example demonstrating how a 2D painting using render target can be done.

available here after release: http://playcanvas.github.io/#graphics/painter.html

Screen Shot 2020-10-20 at 4 34 26 PM

### DIFF
--- a/examples/examples.js
+++ b/examples/examples.js
@@ -36,6 +36,7 @@ var categories = [
             "model-outline",
             "model-shapes",
             "model-textured-box",
+            "painter",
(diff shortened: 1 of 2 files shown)


yaustar (approved):
It be good to have an option to make it interactive but otherwise, good! Thinking about making a project example on playcanvas.com to do bullet decals :)

mvaligursky:
I'm not sure picker is the best way to do bullet decals, performance impact for games is not very good.

PR #2489, opened 2020-10-22
Added a material flag to disable material opacity fading out specular and added a value to specifically alpha fade out materials (including specular)

### LINKED ISSUE #2018: Physically-correct highlights on transparent surfaces

raytranuk:
Fixes #2018 

Implementation is based on #2019 - but with different naming and differing location of some code changes and support for clear coat specular included.

Screen grab from the material-translucent-specular engine example:
image

### DIFF
--- a/examples/graphics/material-translucent-specular.html
+++ b/examples/graphics/material-translucent-specular.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
-    <title>PlayCanvas translucent material specular</title>
+    <title>PlayCanvas Translucent Material Specular</title>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0">
     <link rel="icon" type="image/png" href="../playcanvas-favicon.png" />
@@ -25,6 +25,7 @@
 
         // Create the application and start the update loop
         var app = new pc.Application(canvas);
+        app.start();
 
         // Set the canvas to fill the window and automatically change resolution to be the same as the canvas size
         app.setCanvasFillMode(pc.FILLMODE_FILL_WINDOW);
@@ -40,21 +41,24 @@
         app.scene.toneMapping = pc.TONEMAP_ACES;
         // Set the skybox to the 128x128 cubemap mipmap level
         app.scene.skyboxMip = 1;
+        app.scene.skyboxIntensity = 1;
 
         // Create an entity with a camera component
         var camera = new pc.Entity();
         camera.addComponent("camera");
-        camera.translate(0, 0, 3);
+        camera.translate(0, 0, 8);
+        camera.rotate(0, 0, 0);
         app.root.addChild(camera);
 
-        // Create an entity with a directional light component
-        var light = new pc.Entity();
-        light.addComponent("light", {
-            type: "directional",
-            color: new pc.Color(1, 0.8, 0.25)
-        });
-        app.root.addChild(light);
-        light.setLocalEulerAngles(85, -100, 0);
+         // Create an entities with a directional light components
+        for (var i = 0; i < 3; i++) {
+            var light = new pc.Entity();
+            light.addComponent("light", {
+                type: "directional"
+            });
+            app.root.addChild(light);
+            light.rotateLocal(60+10*i, 30+90*i, 0);
+        }
 
         // Load a cubemap asset. This DDS file was 'prefiltered' in the
         // PlayCanvas Editor and then downloaded.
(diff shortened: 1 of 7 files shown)


examples/graphics/material-translucent-specular.html
```diff
@@ -0,0 +1,195 @@
...
+                    clearCoatMaterial.clearCoatGlossiness = 0.9;
+
+                    clearCoatMaterial.blendType = pc.BLEND_PREMULTIPLIED;
+                    clearCoatMaterial.opacity = 0.5;
+                    clearCoatMaterial.opacityFadesSpecRefl = false;
+                    clearCoatMaterial.alphaWrite = false;
+
+                    clearCoatMaterial.update();
+
+                    createSphere(0.5, 0, 0, clearCoatMaterial);
```
mvaligursky, line 166: you could probably create a function for the shared material set up and only set up differences individually to make it cleaner

raytranuk: I completely switched out the engine example - see picture in the description

src/scene/materials/standard-material-options-builder.js
```diff
@@ -125,6 +125,8 @@ StandardMaterialOptionsBuilder.prototype._updateMaterialOptions = function (opti
     options.lightMapFormat = stdMat.lightMap ? (stdMat.lightMap.type === TEXTURETYPE_RGBM ? 1 : (stdMat.lightMap.format === PIXELFORMAT_RGBA32F ? 2 : 0)) : null;
     options.specularAntialias = stdMat.specularAntialias && (!!stdMat.normalMap) && (!!stdMat.normalMap.mipmaps) && !isPackedNormalMap;
     options.conserveEnergy = stdMat.conserveEnergy;
+    options.opacityFadesSpecRefl = stdMat.opacityFadesSpecRefl;
```
willeastcott, line 128: Any opinions about this `opacityFadesSpecRefl` naming? What about `opacityFadeSpecular`. I'm not a huge fan of cropping words in property names. @mvaligursky, you love naming debates.... 😄

src/scene/materials/standard-material.js
```diff
@@ -262,6 +262,8 @@ import { standardMaterialCubemapParameters, standardMaterialTextureParameters }
  * * occludeSpecularFloat: defines if {@link pc.StandardMaterial#occludeSpecularIntensity} constant should affect specular occlusion.
  * * alphaTest: enable alpha testing. See {@link pc.Material#alphaTest}.
  * * alphaToCoverage: enable alpha to coverage. See {@link pc.Material#alphaToCoverage}.
+ * * opacityFadesSpecRefl: used to specify whether specular and reflections are faded out using opacity. Default is true. When set to false use alphaFade to fade out materials.
+ * * alphaFade: used to fade out materials when opacityFadesSpecRefl is set to false.
```
willeastcott, line 266: These properties are documented as inputs for `onUpdateShader` - but are they not just public on the `StandardMaterial` itself?

src/resources/parser/glb-parser.js
```diff
@@ -919,9 +919,12 @@ var createMaterial = function (gltfMaterial, textures, disableFlipV) {
 
     var material = new StandardMaterial();
 
-    // glTF dooesn't define how to occlude specular
+    // glTF doesn't define how to occlude specular
     material.occludeSpecular = true;
 
+    // glTF doesn't expect material opacity to fade specular lighting
+    material.opacityFadesSpecRefl = true;// false
```
willeastcott, line 926: Delete `// false`

src/scene/materials/standard-material.js
willeastcott: Also, I'm not really understanding this `alphaFade` property. Is this equivalent to `opacity` and used instead of `opacity` when `opacityFadesSpecRefl` is false?

raytranuk: They should be public - I will add them

src/resources/parser/glb-parser.js
raytranuk: I will remove these changes from this PR - and add to a subsequent PR to discuss whether PBR glTF materials should fade specular or not.

src/scene/materials/standard-material-options-builder.js
raytranuk: The problem is that the engine makes a distinction between specular lighting that is referred to as specular and environment map reflections referred to as reflections - and this flag affects both - so it may be unclear if environment map reflections are affected or if they are treated separately?

src/scene/materials/standard-material.js
raytranuk: done - thanks!

src/resources/parser/glb-parser.js
raytranuk: removed changes

src/scene/materials/standard-material-options-builder.js
raytranuk: Although it is true that both specular lighting and environment map reflections model specular reflections - which can be referred in a shorter form to as: specular (as I did for the PR description)

raytranuk: (I just changed it to `opacityFadesSpecular` - since I'm making changes to fix other stuff in the PR anyway)

mvaligursky: I think we should drop 's' from 'Fades', and just have opacityFadeSpecular

mvaligursky:
It looks good to me, even though I'd drop the s from opacityFadesSpecular.

Can you please create an issue on Editor repo to expose this in the Editor and link it here.
And please list new API changes in the description of PR.

mvaligursky (approved):

src/scene/materials/standard-material-options-builder.js
raytranuk: @mvaligursky approved the PR - so I'm marking this as resolved and will merge

src/graphics/program-lib/programs/standard.js
```diff
@@ -1567,6 +1571,15 @@ var standard = {
             }
         }
 
+        if (options.opacityFadesSpecular === false) {
+            if (options.blendType === BLEND_NORMAL || options.blendType === BLEND_PREMULTIPLIED) {
+                code += "float specLum = dot((dSpecularLight + dReflection.rgb * dReflection.a) * dSpecularity, vec3( 0.2126, 0.7152, 0.0722 ));\n";
```
Maksims, line 1576: Can this be moved into a shader chunk?

Maksims:
Currently this PR does not change visual behaviour of existing projects - which is good.
As I understand newly created materials in Editor, now will have fading set to false, while old data without those flags will default to true. Correct?

src/graphics/program-lib/programs/standard.js
raytranuk: I thought about putting into a chunk but decided against it because the javascript logic isolates a single glsl line - and it seems more readable in this form to me.

raytranuk:
@Maksims - I think the proposal will be that new materials have specular not fade with material opacity - the final decision will be made when the editor PR is reviewed - but it makes sense to me to do as you described above.

src/graphics/program-lib/programs/standard.js
Maksims: The question is, will other developers will want to override its behaviour, or this will *never* be the case?

slimbuck:
Hi @raytranuk,

Just checked out this PR and I have a few questions...

As far as I understand, the usual method of rendering glass-type specular is by enabling [premultiplied](https://github.com/playcanvas/engine/blob/master/src/scene/materials/material.js#L198) blend mode and outputting roughly the following from the shader:
```hlsl
gl_FragColor = vec4(diffuse * alpha + specular, alpha);
```
The premultiplied blend mode is applied as follows:
```hlsl
final = gl_FragColor.rgb + final * (1.0 - gl_FragColor.a);
```
Which, expanded is roughly:
```hlsl
final = diffuse * alpha + specular + final * (1.0 - alpha);
```

This gives precisely what we're after: blended diffuse component with (unblended) specular.

My questions:
1) Reading the shader code I'm not clear how the above is accomplished? I might be wrong, but it looks like the shader calculates a rough approximation of the above based on diffuse and specular colour and works in both normal blend mode and premultiplied-alpha mode (though neither are correct?). Seems to me the approximation may behave unexpectedly under certain scenarios, like very differently coloured diffuse and specular components. Why approximate the blending result when it should be straightforward (and faster!) to do it _correctly_?

2) Why introduce a new `alphaFade` value? How is it different from the material's existing specular colour tint? While I understand its purpose in spirit, it also seems quite confusing.

PR #2493, opened 2020-10-23
normalize interpolated vertex normal when there are no tangents and d…

raytranuk:
…erived tangents can be calculated

Fixes bad looking specular on some Bitmojis

### DIFF
--- a/src/graphics/program-lib/programs/standard.js
+++ b/src/graphics/program-lib/programs/standard.js
@@ -1339,11 +1339,20 @@ var standard = {
         code = this._fsAddStartCode(code, device, chunks, options);
 
         if (needsNormal) {
-            if (options.twoSidedLighting) {
-                code += "   dVertexNormalW = gl_FrontFacing ? vNormalW * twoSidedLightingNegScaleFactor : -vNormalW * twoSidedLightingNegScaleFactor;\n";
+            if (!options.hasTangents && device.extStandardDerivatives) {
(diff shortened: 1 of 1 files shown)


src/graphics/program-lib/programs/standard.js
```diff
@@ -1339,11 +1339,20 @@ var standard = {
...
+                } else {
+                    code += "   dVertexNormalW = normalize(vNormalW);\n";
+                }
             } else {
-                code += "   dVertexNormalW = vNormalW;\n";
+                if (options.twoSidedLighting) {
+                    code += "   dVertexNormalW = gl_FrontFacing ? vNormalW * twoSidedLightingNegScaleFactor : -vNormalW * twoSidedLightingNegScaleFactor;\n";
+                } else {
+                    code += "   dVertexNormalW = vNormalW;\n";
+                }
```
mvaligursky, line 1353: why don't we normalize here as well .. are these normalized already?

raytranuk: they get normalized or not based on fastTbn option - actually... we may want to use the fastTbn option to skip normalization even in the case where there are no tangents - I shall make that change.

raytranuk: I added a change to respect the fastTbn material option - if the fastTbn option is true then the normalize is not performed

mvaligursky (approved):

PR #2495, opened 2020-10-23
Fixed WebVR transform update from vrDisplay

yaustar:
Fixes bug where the camera local transform was not updated to match the direction that display was facing.

### DIFF
--- a/src/framework/components/camera/system.js
+++ b/src/framework/components/camera/system.js
@@ -139,11 +139,12 @@ Object.assign(CameraComponentSystem.prototype, {
             for (var id in components) {
                 var component = components[id];
 
-                if (component.enabled && component.entity.enabled) {
-                    var vrDisplay = component.vrDisplay;
+                if (component.data.enabled && component.entity.enabled) {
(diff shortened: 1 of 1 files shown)


src/framework/components/camera/system.js
```diff
@@ -139,11 +139,12 @@ Object.assign(CameraComponentSystem.prototype, {
             for (var id in components) {
                 var component = components[id];
 
-                if (component.enabled && component.entity.enabled) {
-                    var vrDisplay = component.vrDisplay;
+                if (component.data.enabled && component.entity.enabled) {
+                    var cameraComponent = component.entity.camera;
```
yaustar, line 143: This seems very wrong to get the component from the entity but couldn't see a clean way to add to the data class

willeastcott: This is fine. The screen component does something similar: https://github.com/playcanvas/engine/blob/master/src/framework/components/screen/system.js#L85

willeastcott (approved):

PR #2500, opened 2020-10-27
Standard Node Material and Shader Graph 

raytranuk:
### This PR:
- Adds private classes: `pc.StandardNodeMaterial`, `pc.ShaderNodeRegistry`, `pc.ShaderGraphBuilder` and `pc.ShaderGraphNode`
- Adds a private engine example: `shader-graph-chunks-builder` demonstrating the use of `pc.ShaderNodeRegistry` and `pc.ShaderGraphBuilder` with `pc.StandardNodeMaterial`

### example code:
```
    //create a shader graph builder
    var builder = new pc.ShaderGraphBuilder(app);

    //add input parameters
    var timeParam = builder.addParam('float', 'uTime', 0.0);
    var emissiveTexParam = builder.addParam('sampler2D', 'uEmissiveMap', emissiveMap);
    var emissiveColConst = builder.addConst('vec3', 'cEmissiveColorConst', new pc.Vec3(0.2, 0.3, 0.3));

    //create and hook up graph nodes
    var scrollNode = builder.addNode('add', timeParam, builder.addNode('uv0'));
    var emissiveSampleNode = builder.addNode('texSample', emissiveTexParam, scrollNode);
    var emissiveCalcNode = builder.addNode('mul', emissiveColConst, {node:emissiveSampleNode, port:'rgb'});

    //hook up outputs
    builder.addOutput({node:emissiveSampleNode, port:'g'}, 'float', 'sgAlpha');
    builder.addOutput(emissiveCalcNode, 'vec3', 'sgEmission');

    //return built shader graph chunk
    var shaderGraphChunk = builder.getShaderGraphChunk();
```
### Additional Implementation Notes

- In order to derive `pc.StandardNodeMaterial` and its shader generator, `pc.StandardMaterial` and the `standard` shader generator were altered - however, to try to reduce backwards compatibility issues and try to reduce future maintenance workload, the changes were kept to a minimum.
- Shader graph asset handling was removed from this PR, and will be added later alongside editor support in a Shader Graph Editor MVP PR.


src/scene/materials/shader-graph-node.js
```diff
@@ -0,0 +1,614 @@
...
+
+    hasValidGraphData: function (graphRootTime) {
+        var i;
+        if (this.graphData.ioPorts && this.graphData.ioPorts.length > 0 && (this.graphData.customFuncGlsl || ( this.graphData.subGraphs && this.graphData.subGraphs.length > 0 && this.graphData.connections))) {
+            for (i = 0; i < this.graphData.ioPorts.length; i++) {
+                if (this.graphData.ioPorts[i] && this.graphData.ioPorts[i].name && this.graphData.ioPorts[i].type) {
+                    if (this.graphData.ioPorts[i].type === 'sampler2D' && !this.graphData.ioPorts[i].name.startsWith('OUT_') ) {
+                        if (!(this.graphData.ioPorts[i].valueTex && this.graphData.ioPorts[i].valueTex instanceof Texture)) {
+                            return false;
+                        }
```
mvaligursky, line 98: you should do something like this:
`var port = this.graphData.ioPorts[i];
and then access it like this:
ports.type ...
`

examples/graphics/shader-graph-chunks-builder.html
```diff
@@ -0,0 +1,134 @@
...
+                            meshInstance.material.alphaTest = 0.5;
+                            meshInstance.material.update();
+                        }
+                    });
+
+                    //update time parameter and alpha test every frame
+                    app.on("update", function (dt) {
+                        time += dt;
+                        shaderGraphMaterials.forEach(function (material) {
+                            material.setShaderGraphParameter('uTime', time);
```
mvaligursky, line 123: should we have simpler name for this instead of **setShaderGraphParameter**

src/scene/materials/shader-graph-node.js
raytranuk: I did exactly that elsewhere - not sure why I missed this loop - now fixed - thanks!

examples/graphics/shader-graph-chunks-builder.html
raytranuk: Any suggestions? This is the equivalent to setParameter - but for the shader graph embedded in the (standard node) material - so the function is setting a shader graph parameter...

mvaligursky: I wonder if you could maybe even override setParameter and do some quick check to see where the parameter needs to be set?

examples/graphics/shader-graph-chunks-builder.html
```diff
@@ -0,0 +1,137 @@
...
+                    
+                    //register loaded core nodes
+                    pc.shaderChunks.registerCoreShaderGraphNodes(coreNodeList);
+
+                    //create a shader graph builder
+                    var builder = new pc.ShaderGraphBuilder();
+
+                    //add input parameters
+                    var timeParam = builder.addParam('float', 'uTime', 0.0);
+                    var emissiveTexParam = builder.addParam('sampler2D', 'uEmissiveMap', emissiveMap);
```
mvaligursky, line 85: it seems the value is not optional parameter for .addParam - internally in ShaderGraphNode._addIoPort a type of it is tested and used. Do we need to be able to build a shader graph without having a value set?

mvaligursky: I tried removing the emissiveMap parameter and setting it later and that seems to work. If that's supported, maybe the parameter needs to be marked as optional.

src/scene/materials/shader-graph-node.js
```diff
@@ -0,0 +1,615 @@
...
+
+    addInput: function (type, name, value) {
+        return this._addIoPort(type, 'IN_' + name, value);
+    },
+
+    addOutput: function (type, name, value) {
+        return this._addIoPort(type, 'OUT_' + name, value);
+    },
+
+    addConstant: function (type, value) {
```
mvaligursky, line 201: curious if this could take [optional] name parameter to be used to make generated code more readable? (to be used instead of 'this.graphData.ioPorts.length' when supplied)

examples/graphics/shader-graph-chunks-builder.html
raytranuk: okay - I'll try overriding setParameter

examples/graphics/shader-graph-chunks-builder.html
raytranuk: yes - it optional - but won't be unless its set later (before rendering) - I'll mark it as optional

mvaligursky: so should ShaderGraphNode._addIoPort use _type_ instead of _value_ in those ifs? Without value it falls back to what comment suggests is output port.

src/scene/materials/shader-graph-node.js
raytranuk: yes - I can add that

examples/graphics/shader-graph-chunks-builder.html
raytranuk: I will adjust the comment

raytranuk: done - made value optional when graph params (node ports) are added - and updated comment - thanks!

examples/graphics/shader-graph-chunks-builder.html
raytranuk: Overriding the setParameter function works and is definitely nicer - thanks!

src/scene/materials/shader-graph-node.js
raytranuk: I added an optional userId - from past experience building material graphs, its rare to name constants - unless you convert it to a parameter - but this could be useful for debugging

src/graphics/shader-node-registry.js
```diff
@@ -0,0 +1,43 @@
...
+ */
+ShaderNodeRegistry.prototype.register = function (coreNodeList) {
+    if (coreNodeList) {
+        if (typeof coreNodeList === 'string') {
+            coreNodeList = JSON.parse(coreNodeList);
+        }
+        var self = this;
+        Object.keys(coreNodeList).forEach(function (key) {
+            self._nodes[key] = new ShaderGraphNode(coreNodeList[key].code);
+            self._list.push(self._nodes[key]);
```
mvaligursky, line 38: We should filter the duplicates from the list. Perhaps with a warning in debug mode.

mvaligursky: I have not checked how this being used yet, just curious why the list is needed when you have dictionary like access with _nodes. Likely useful I'm sure, just checking.

examples/graphics/shader-graph-chunks-builder.html
```diff
@@ -0,0 +1,140 @@
...
+
+            // load a texture asset
+            app.assets.loadFromUrl("../assets/textures/clouds.jpg", "texture", function (err, asset) {    
+                var emissiveMap = asset.resource;
+
+                //load shader graph core nodes asset
+                app.assets.loadFromUrl("../assets/core_nodes.json", "text", function (err, asset) {
+                    var coreNodeList = asset.resource;
+                    
+                    //register loaded core nodes
```
mvaligursky, line 80: please use single space after // just like required in the engine code

src/scene/materials/shader-graph-builder.js
```diff
@@ -0,0 +1,128 @@
...
+    addParam: function (type, name, value) {
+        var ioPort = this._graph.addInput(type, name, value);
+        return ioPort;
+    },
+    /**
+     * @private
+     * @function
+     * @name pc.ShaderGraphBuilder#addNode
+     * @description creates and adds a core node to the shader graph and connects up inputs
+     * @param {string} name - name of parameter
```
mvaligursky, line 44: should be "name of node" I think

src/scene/materials/shader-graph-builder.js
```diff
@@ -0,0 +1,128 @@
...
+     * @param {object} [arguments] - first argument is name string, then 0-N output node ports or params to connect to new core node inputs - number and port type of arguments must match.
```
mvaligursky, line 45: "{object} [arguments]" - I think this might create typescript definition allowing a single argument only. It might need some way to indicate list of arguments, not sure what the right way to do this is.

src/scene/materials/shader-graph-builder.js
```diff
@@ -0,0 +1,128 @@
...
+     * @example
+     * var node0 = pc.shadergraph.addNode('uv0');
+     * var node1 = pc.shadergraph.addNode('texSample', param0, node0 );
+     * var node2 = pc.shadergraph.addNode('splitVec4', node1 );
+     * var node3 = pc.shadergraph.addNode('mul3', param1, node42 );
+     */
+    addNode: function (name) {
+        var args = arguments;
+        var coreNode = this._nodes[name];
+        var sgIndex = this._graph.addSubGraph(coreNode);
```
mvaligursky, line 56: I think the public API should be more robust and check / report issues. In this case coreNode can be null, which would cause exception later. addSubGraph should handle the case of duplicate and print error / fail and similar, to help users diagnose issues.

src/scene/materials/shader-graph-builder.js
```diff
@@ -0,0 +1,128 @@
...
+                } else if (arg.type) {
+                    // ioPort (ioPort)
+                    this._graph.connect(-1, arg.name, sgIndex, sgInputs[argIndex].name);
+                } else {
+                    // specific port of a node
+                    this._graph.connect(arg.node, 'OUT_' + arg.port, sgIndex, sgInputs[argIndex].name);
+                }
+            }
+        } else {
+            console.warn("arguments do not match core node");
```
mvaligursky, line 78: Please make the message more informative. Node name, required and provided number of arguments.

src/scene/materials/shader-graph-builder.js
```diff
@@ -0,0 +1,128 @@
...
+            return ioPort.name.startsWith('IN_');
+        });
+
+        if (sgInputs.length === (args.length - 1)) {
+            for (var argIndex = 0; argIndex < (args.length - 1); argIndex++) {
+                var arg = args[argIndex + 1];
+                if (typeof(arg) === 'number') {
+                    // ret port of a node
+                    var argNodeIndex = arg;
+                    this._graph.connect(argNodeIndex, 'OUT_ret', sgIndex, sgInputs[argIndex].name);
```
mvaligursky, line 68: I'd like this to be more robust as it's public API - this should check .connect was successful, and print errors with details if not.

src/scene/materials/shader-graph-builder.js
```diff
@@ -0,0 +1,128 @@
...
+     * pc.shadergraph.output(arg, 'float', 'dAlpha');
+     * pc.shadergraph.output(arg, 'vec3', 'dAlbedo');
+     * pc.shadergraph.output(arg, 'vec3', 'dNormalMap');
+     * pc.shadergraph.output(arg, 'float', 'dGlossiness');
+     * pc.shadergraph.output(arg, 'vec3', 'dSpecularity');
+     * pc.shadergraph.output(arg, 'vec3', 'dEmission');
+     */
+    addOutput: function (arg, type, name) {
+        // assumes this is only called once per graph output TODO: verify this
+        var ioPort = this._graph.addOutput(type, name);
```
mvaligursky, line 102: yes lets verify this ;)

src/scene/materials/shader-graph-builder.js
```diff
@@ -0,0 +1,128 @@
...
+
+        if (typeof(arg) === 'number') {
+            // ret port of a node
+            var argNodeIndex = arg;
+            this._graph.connect(argNodeIndex, 'OUT_ret', -1, ioPort.name);
```
mvaligursky, line 107: again - please add validation .connect was successful, and useful error message if not

src/scene/materials/shader-graph-builder.js
```diff
@@ -0,0 +1,128 @@
...
+     * @function
+     * @name pc.ShaderGraphBuilder#addParam
+     * @description adds a parameter input to graph
+     * @param {string} type - type of parameter
+     * @param {string} name - name of parameter
+     * @param {any} value - optional value of parameter
+     * @returns {any} returns the created input port
+     */
+    addParam: function (type, name, value) {
+        var ioPort = this._graph.addInput(type, name, value);
```
mvaligursky, line 36: handle and print error when .addInput fails (duplicates, incorrect type ...)

src/scene/materials/shader-graph-builder.js
```diff
@@ -0,0 +1,128 @@
...
+     * @private
+     * @function
+     * @name pc.ShaderGraphBuilder#addOutput
+     * @description connect a node port to graph material output
+     * @param {any} arg - node port to output
+     * @param {string} type - type of output
+     * @param {string} name - name of output
+     * @example
+     * // valid outputs are:
+     * pc.shadergraph.output(arg, 'vec3', 'vertOff');
```
mvaligursky, line 92: I think valid outputs should be stored in a global array as data.
Something like
`var validOutputs = [ { name: "dAlpha", type: "float"}, ... ];`
This would give us:
- easier validation, and remove the need to pass in type
- would allow us to implement function to do this instead code like this being used in few places in other files:
`        if (rootShaderGraph.getIoPortByName('OUT_dAlpha') || rootShaderGraph.getIoPortByName('OUT_dNormalMap') || rootShaderGraph.getIoPortByName('OUT_dGlossiness') || rootShaderGraph.getIoPortByName('OUT_dSpecularity') || rootShaderGraph.getIoPortByName('OUT_dAlbedo') || rootShaderGraph.getIoPortByName('OUT_fragOut') || rootShaderGraph.getIoPortByName('OUT_dEmission')) {
`

src/scene/materials/standard-node-material.js
```diff
@@ -0,0 +1,100 @@
...
+function StandardNodeMaterial(mat, chunk) {
+    StandardMaterial.call(this);
+
+    if (mat) {
+        StandardMaterial.prototype._cloneInternal.call(mat, this);
+    }
+
+    if (chunk) {
+        this._shaderGraphChunk = chunk;
+    }
```
mvaligursky, line 27: remove the if line, so _shaderGraphChunk is always assigned to (either valid chunk, or undefined)

src/scene/materials/standard-node-material.js
```diff
@@ -0,0 +1,100 @@
...
+}
+
+StandardNodeMaterial.prototype = Object.create(StandardMaterial.prototype);
+StandardNodeMaterial.prototype.constructor = StandardNodeMaterial;
+
+Object.assign(StandardNodeMaterial.prototype, {
+
+    clone: function () {
```
mvaligursky, line 35: this is public function (overrides Material. ) and should have documentation even if the same, otherwise it won't show up in manual

src/scene/materials/standard-node-material.js
```diff
@@ -0,0 +1,100 @@
...
+        var clone = new StandardNodeMaterial();
+        StandardMaterial.prototype._cloneInternal.call(this, clone);
+
+        clone._shaderGraphChunk = this._shaderGraphChunk;
+
+        return clone;
+    },
+
+    setParameter: function (name, data) {
```
mvaligursky, line 44: this is public function (overrides Material. ) and should have documentation even if the same, otherwise it won't show up in manual

src/scene/materials/standard-node-material.js
```diff
@@ -0,0 +1,100 @@
...
+        StandardMaterial.prototype.setParameter.call(this, name, data);
+
+        if (this._shaderGraphChunk) {
+            var rootShaderGraph = this._shaderGraphChunk;
+
+            var portName = 'IN_' + name + '_' + rootShaderGraph.id;
```
mvaligursky, line 50: I wonder if there's a way to do this without allocations

src/scene/materials/standard-node-material.js
```diff
@@ -0,0 +1,100 @@
...
+        }
+    },
+
+    updateUniforms: function () {
+        StandardMaterial.prototype.updateUniforms.call(this);
+
+        if (this._shaderGraphChunk) {
+            var rootShaderGraph = this._shaderGraphChunk;
+
+            rootShaderGraph.updateShaderGraphUniforms(this);
```
mvaligursky, line 62: instead of updateShaderGraphUniforms, maybe call it updateUniforms

examples/graphics/shader-graph-chunks-builder.html
```diff
@@ -0,0 +1,140 @@
...
+                    //hook up outputs
+                    builder.addOutput({node:emissiveSampleNode, port:'g'}, 'float', 'dAlpha');
+                    builder.addOutput(emissiveCalcNode, 'vec3', 'dEmission');
+
+                    //return built shader graph chunk
+                    var shaderGraphChunk = builder.getShaderGraphChunk();
+
+                    //replace model's materials with standard node materials
+                    var originalMaterials=[];
+                    var shaderGraphMaterials=[];
```
mvaligursky, line 105: spaces around =

src/graphics/program-lib/programs/standard-node.js
```diff
@@ -0,0 +1,93 @@
...
+        graphCode[2] = '';
+        if (rootShaderGraph.getIoPortByName('OUT_dAlpha') || rootShaderGraph.getIoPortByName('OUT_dNormalMap') || rootShaderGraph.getIoPortByName('OUT_dGlossiness') || rootShaderGraph.getIoPortByName('OUT_dSpecularity') || rootShaderGraph.getIoPortByName('OUT_dAlbedo') || rootShaderGraph.getIoPortByName('OUT_fragOut') || rootShaderGraph.getIoPortByName('OUT_dEmission')) {
+            graphCode[2] += "#define SG_PS\n";
+            graphCode[2] += rootDeclGLSL;
+        }
+    }
+
+    if (options._shaderGraphChunk) {
+        graphCode[3] = '';
+        if (rootShaderGraph.getIoPortByName('OUT_dAlpha') || rootShaderGraph.getIoPortByName('OUT_dNormalMap') || rootShaderGraph.getIoPortByName('OUT_dGlossiness') || rootShaderGraph.getIoPortByName('OUT_dSpecularity') || rootShaderGraph.getIoPortByName('OUT_dAlbedo') || rootShaderGraph.getIoPortByName('OUT_fragOut') || rootShaderGraph.getIoPortByName('OUT_dEmission')) {
```
mvaligursky, line 54: this function should be made more generic based on possible shader node outputs being specified in data, mentioned elsewhere

src/graphics/program-lib/programs/standard-node.js
```diff
@@ -0,0 +1,93 @@
...
+    if (options._shaderGraphChunk) {
+        rootShaderGraph = options._shaderGraphChunk;
+
+        rootDeclGLSL = rootShaderGraph.generateRootDeclGlsl();
+        rootCallGLSL = rootShaderGraph.generateRootCallGlsl();
+    }
+
+    var graphCode = [];
+
+    if (options._shaderGraphChunk) {
```
mvaligursky, line 27: any reason this condition is repeated few times below, instead of all code being inside a single if statement?

src/graphics/program-lib/programs/standard-node.js
```diff
@@ -0,0 +1,93 @@
...
+        graphCode[0] = '';
+        if (rootShaderGraph.getIoPortByName('OUT_vertOff')) {
```
mvaligursky, line 29: also this line is called again bellow, perhaps cache result, or combine those blocks?

src/graphics/program-lib/programs/standard.js
```diff
@@ -465,6 +467,8 @@ var standard = {
         var needsNormal = lighting || reflections || options.ambientSH || options.prefilteredCubemap || options.heightMap || options.enableGGXSpecular;
         var shadowPass = options.pass >= SHADER_SHADOW && options.pass <= 17;
 
+        needsNormal = needsNormal || graphCode;
```
mvaligursky, line 470: does it always need normal?

src/graphics/program-lib/programs/standard-node.js
```diff
@@ -0,0 +1,93 @@
...
+    var rootCallGLSL = '';
+
+    if (options._shaderGraphChunk) {
+        rootShaderGraph = options._shaderGraphChunk;
+
+        rootDeclGLSL = rootShaderGraph.generateRootDeclGlsl();
+        rootCallGLSL = rootShaderGraph.generateRootCallGlsl();
+    }
+
+    var graphCode = [];
```
mvaligursky, line 25: instead of array which we have fixed 4 slots in that we access as array[1] and similar, make it an object with more descriptive names perhaps, making it easier on receiver side to understand the code. For example:
```
var graphCode = {};
graphCode.definitions = ...;
graphCode.functions = ...;
```

src/graphics/program-lib/programs/standard.js
```diff
@@ -1417,7 +1434,36 @@ var standard = {
...
             if (!getGlossinessCalled) code += "   getGlossiness();\n";
-            if (options.fresnelModel > 0) code += "   getFresnel();\n";
+            if (options.fresnelModel > 0) {
+                if (!graphCode) code += "   getFresnel();\n";
+            }
+        }
+
+        if (graphCode) {
+            if (!options.normalMap) {
+                code += "   dNormalMap = vec3(0);\n";
```
mvaligursky, line 1444: maybe should be 0.0 instead of 0, might generate error that int cannot be converted to float. Not sure.

src/graphics/shader-node-registry.js
raytranuk: I will add a warning - the core nodes JSON asset should be considered an engine private asset and is invalid if core nodes are defined more than once - but being JSON format some developers will probably modify it - and a warning will also help engine development to identify and debug bad merges. The private _list is something I added for use (and is used) with the material editor prototype - and is quite useful for that.

src/scene/materials/shader-graph-builder.js
raytranuk: yes - copy & paste typo - will fix

src/scene/materials/shader-graph-node.js
```diff
@@ -0,0 +1,615 @@
+import { Texture } from '../../graphics/texture.js';
+import { Vec2 } from '../../math/vec2.js';
+import { Vec3 } from '../../math/vec3.js';
+import { Vec4 } from '../../math/vec4.js';
+
```
mvaligursky, line 5: overall, I would like to see more comments in this file.

src/scene/materials/shader-graph-builder.js
raytranuk: I saw this format used in JSDocs in another system - I will check (google it) if it right.

src/scene/materials/shader-graph-node.js
```diff
@@ -0,0 +1,615 @@
...
+        clone.graphData.customFuncGlsl = this.graphData.customFuncGlsl;
+
+        return clone;
+    },
+
+    updateShaderGraphUniforms: function (mat) {
+        for (var n = 0; n < this.graphData.ioPorts.length; n++) {
+            var ioPort = this.graphData.ioPorts[n];
+
+            if (ioPort.name.startsWith('IN_') || (ioPort.name.startsWith('CONST_') && ioPort.type === 'sampler2D')) {
```
mvaligursky, line 62: ideally perhaps we have a list that allows us to avoid these string operations / testing here. We should have a list of uniforms to just loop over, and set parameter on material.

src/scene/materials/shader-graph-node.js
```diff
@@ -0,0 +1,615 @@
...
+                var chunkId = '_' + this.id;
+
+                switch (ioPort.type) {
+                    case 'sampler2D':
+                        mat.setParameter(ioPort.name + chunkId, ioPort.valueTex);
```
mvaligursky, line 67: this should be evaluated once and stored as well: ioPort.name + chunkId

src/scene/materials/shader-graph-builder.js
raytranuk: A check would be nice - when/if the API is public a badly typed in name could result in exceptions - what could be better is a enum of core nodes (which isn't possible yet because the complete list of core nodes hasn't yet been locked down) - for now - when used privately with the editor - its less of a problem?

src/scene/materials/shader-graph-builder.js
raytranuk: I will add a bit more - NB this won't be a problem for MVP - the editor's core node schemas are generated by the core node asset so they always match

src/scene/materials/shader-graph-builder.js
raytranuk: ShaderGraphNode won't be made public - additional checks could be made in the builder when/if it gets made public (after MVP).

src/scene/materials/shader-graph-node.js
```diff
@@ -0,0 +1,615 @@
...
+                    case 'samplerCube':
+                    default:
+                        // error
+                        break;
+                }
+            }
+        }
+    },
+
+    hasValidGraphData: function (graphRootTime) {
```
mvaligursky, line 90: this does not have comment about the purpose and it's currently not called, so I'm not entirely sure what its purpose is - will this called when asset with graph is loaded?

src/scene/materials/shader-graph-node.js
```diff
@@ -0,0 +1,615 @@
...
+
+                // infinite recursion protection
+                // if no graphRootTime set, set it to current time
+                if (!graphRootTime) {
+                    graphRootTime = new Date().getTime();
+                    this._tmpRootTime = graphRootTime;
+                }
+                for (i = 0; i < this.graphData.subGraphs.length; i++) {
+                    if (this.graphData.subGraphs[i] && this.graphData.subGraphs[i] instanceof ShaderGraphNode && this.graphData.subGraphs[i].name) {
+                        if (this.graphData.subGraphs[i]._tmpRootTime === graphRootTime) {
```
mvaligursky, line 151: it seem that the time is used as unique Id here? If yes, I think we should use some variable we keep incrementing instead, as the time has low precision (2ms on some browsers) and so will not give us unique id at all times.

src/scene/materials/shader-graph-builder.js
raytranuk: Not necessary while builder is privately used by the editor - where all outputs are specified in a output node - after MVP I'd like to support custom outputs and will have to do checks for clashes at that point

src/scene/materials/shader-graph-node.js
```diff
@@ -0,0 +1,615 @@
...
+
+            if (inOrOutAndTypeAndName[0] === "") inOrOutAndTypeAndName.shift();
+
+            if (inOrOutAndTypeAndName[0] === "out") {
+                this.addOutput(inOrOutAndTypeAndName[1], inOrOutAndTypeAndName[2]);
+            }
+            if (inOrOutAndTypeAndName[0] === "in") {
+                this.addInput(inOrOutAndTypeAndName[1], inOrOutAndTypeAndName[2]);
+            } else {
+                // unsupported parameter
```
mvaligursky, line 230: perhaps log an error

src/scene/materials/shader-graph-node.js
```diff
@@ -0,0 +1,615 @@
...
+    },
+
+    addSubGraph: function (graph) {
+        var ret = this.graphData.subGraphs.length;
+        this.graphData.subGraphs.push(graph);
+        return ret;
+    },
+
+    connect: function (srcIndex, srcVarName, dstIndex, dstVarName) {
+        var connection = { srcIndex: srcIndex, srcVarName: srcVarName, dstIndex: dstIndex, dstVarName: dstVarName };
```
mvaligursky, line 242: perhaps we could make an actual class for the connection object

src/scene/materials/shader-graph-builder.js
raytranuk: As mentioned above, used privately with the editor, it is the editor that does validation for connections - preventing invalid connections from being made - after MVP we can come up with a system to do it with code (maybe even at code build time?).

src/scene/materials/shader-graph-builder.js
raytranuk: definitely after MVP

src/scene/materials/shader-graph-node.js
```diff
@@ -0,0 +1,615 @@
...
+
+        this.graphData.connections.push(connection);
+    },
+
+    _getIoPortValueString: function (ioPort) {
+        var ret;
+
+        if (ioPort.type === 'float') {
+            ret = 'float(' + ioPort.valueX + ')';
```
mvaligursky, line 251: do we need to make sure values for float type contain '.' - otherwise shader compiler will see them as int and often generate incompatible type error.

src/scene/materials/shader-graph-builder.js
raytranuk: yes - when the API is made public, a public enum should be added - probably the same with the core nodes - as mentioned before, used privately with the editor there will be no mismatch - until custom outputs are added to the editor (after MVP)

src/scene/materials/standard-node-material.js
raytranuk: true - will remove

src/scene/materials/standard-node-material.js
raytranuk: ah - I shall put the JSDoc block back in

src/scene/materials/standard-node-material.js
raytranuk: I will restore the JSDoc block

src/scene/materials/shader-graph-node.js
```diff
@@ -0,0 +1,615 @@
...
+            } else if ( func.endsWith('VS') ) {
+                funcString += '#endif //SG_VS\n';
+            }
+
+            depGraphList.push(funcString);
+        }
+        var tLen = depGraphList.length; // need this because pop() reduces array length!
+        for (i = 0; i < tLen; i++) {
+            generatedGlsl += depGraphList.pop();
+        }
```
mvaligursky, line 414: while (depGraphList.length) {
    generatedGlsl += depGraphList.pop();
}

src/scene/materials/shader-graph-node.js
```diff
@@ -0,0 +1,615 @@
...
+        var generatedGlsl = '';
+
+        // generate input and output names for function call and run through outputs to declare variables
+        var inNames = {};
+        var outNames = {};
+
+        for (var i = 0; i < this.graphData.ioPorts.length; i++) {
+            var matId = '_' + this.id;
+            var ioPort = this.graphData.ioPorts[i];
+            if (ioPort.name.startsWith('IN_')) {
```
mvaligursky, line 429: Personally I'd be tempted to make Port a class, and when its constructed with name, I would store booleans on it:
```
this.isInput = this.name.startsWith('IN_');
this.isOutput = this.name.startsWith('OUT_');
...
```

src/graphics/program-lib/programs/standard-node.js
raytranuk: I'll think about how to do this - however this code defines the chunks that gets inserted into the Standard Material's via its shader generator - so I wanted to keep it as simple as possible - with some of the same searchable terms in the standard material.

src/graphics/program-lib/programs/standard-node.js
raytranuk: Its just a hangover from actually being embedded in the standard material shader generator - where it was useful for ironing out edge cases in the logic - I will remove the duplication - as put comments in to help maintenance when StandardMaterial is updated

src/scene/materials/shader-graph-node.js
```diff
@@ -0,0 +1,615 @@
...
+            while (subGraphList.length < this.graphData.subGraphs.length && whileLoopCount < this.graphData.connections.length) {
+                whileLoopCount++;
+
+                for (i = 0; i < this.graphData.subGraphs.length; i++) {
+                    if (subGraphOnListFlags[i] != true) {
+                        var allInputsOnList = true;
+                        if (dstConnectedMap[i] != undefined) {
+                            for (var n = 0; n < dstConnectedMap[i].length; n++) {
+                                var connectedSrcIndex = dstConnectedMap[i][n];
+                                if (subGraphOnListFlags[connectedSrcIndex] != true) {
```
mvaligursky, line 561: !== instead of !=

src/graphics/program-lib/programs/standard.js
raytranuk: technically it doesn't - so I will add a check if the shader graph needs normal

src/graphics/program-lib/programs/standard-node.js
raytranuk: makes sense - I will improve this

src/graphics/program-lib/programs/standard.js
raytranuk: no error - GLSL allows this as constructors are used for type conversion

mvaligursky:
I did a code review and added some inline questions / suggestion. Here are more generic points:

- can user create a new node (subgraph) that acts like a core node in a way that it can be used by other graphs? Does getShaderGraphChunk allow this perhaps?
- documentation description format (example: "@description _get shader graph chunk_") - Sentences should start with capital letter and end with full stop.
- call define SG_PS something more descriptive, what does SG stand for?
- maybe ShaderGraphBuilder.AddNode could take optional name parameter, allowing better name generation for temporary variables, making final code more readable?
- I wonder what the plan is to support WebGPU when the time comes. Are you thinking about translating glsl->webgpu shader language, or adding direct support for it inside the node system?
- how to add a constant to the graph (not a uniform, which has higher cost - cpu needs to set it every frame, takes uniform slot, storage on each material ..) - I don't see API for this, unless I missed it.

src/scene/materials/shader-graph-node.js
raytranuk: I can do a pass and add comments to anything that seems non-obvious

src/scene/materials/shader-graph-node.js
raytranuk: yes - I will add _lists

src/scene/materials/shader-graph-builder.js
mvaligursky: We review this code now, and the plan is to make it public at some point, and so validation is needed. We won't be reviewing it again when we simply turn interface public, so I believe the validation needs to be done at this stage and not later.

src/scene/materials/shader-graph-node.js
raytranuk: hmm - it could be but may not be much use - name is persistent, but chunkId is transient

src/scene/materials/shader-graph-node.js
raytranuk: yes - this was used by the asset handler - which will dynamically swap in asset resources - probably needed for next stage of MVP

src/scene/materials/shader-graph-node.js
raytranuk: okay - I'll switch it to a counter

src/scene/materials/shader-graph-node.js
raytranuk: nice - I'll swap in this code - thanks!

src/scene/materials/shader-graph-node.js
raytranuk: I am considering refactoring to split out ports into categories - or maintain a transient private structure to accelerate access and/or group the split categories for convenient iteration.

raytranuk:
> I did a code review and added some inline questions / suggestion. Here are more generic points:
> 
> 1. can user create a new node (subgraph) that acts like a core node in a way that it can be used by other graphs? Does getShaderGraphChunk allow this perhaps?
**Although the underlying system does support user created subgraphs, the MVP plan is not to expose them yet - to reduce editor-side work scope for MVP.**
> 2. documentation description format (example: "@description _get shader graph chunk_") - Sentences should start with capital letter and end with full stop.
**Public documentation uniformity is of course very important.** 
> 3. call define SG_PS something more descriptive, what does SG stand for?
**SG stands for Shader Graph - I can change it.**
> 4. maybe ShaderGraphBuilder.AddNode could take optional name parameter, allowing better name generation for temporary variables, making final code more readable?
**Shader code readability, unfortunately, is not something that can be maintained with most implementations of shader graphs - but after MVP I can something like this - graph readability is something that that is more possible - and there are editor features that will need to be developed after MVP**
> 5. I wonder what the plan is to support WebGPU when the time comes. Are you thinking about translating glsl->webgpu shader language, or adding direct support for it inside the node system?
**The option I was leaning toward was a WebGPU version of the Core Nodes and some minor changes to the Shader Graph systems.**
> 6. how to add a constant to the graph (not a uniform, which has higher cost - cpu needs to set it every frame, takes uniform slot, storage on each material ..) - I don't see API for this, unless I missed it.
**This was needed for the engine example so not exposed in the builder - but is supported at a lower level so will be exposed for use in the MVP editor PR.**

Thanks again for the PR review - I will do all the things I mention in my replies as promptly as I can.

src/scene/materials/shader-graph-node.js
mvaligursky: yep either of these would work well

src/scene/materials/shader-graph-builder.js
raytranuk: Fair point - I will add some validation.

raytranuk: I will also say that we should be more thorough about making previously private APIs public than just removing the `@private` JSDoc keyword

src/scene/materials/shader-graph-builder.js
raytranuk: added validation code to the builder class

src/scene/materials/shader-graph-builder.js
raytranuk: added validation code to the builder class

src/scene/materials/shader-graph-builder.js
raytranuk: added validation code to the builder class

src/scene/materials/shader-graph-builder.js
raytranuk: added validation code to the builder class

src/scene/materials/shader-graph-builder.js
raytranuk: After more thought, the shader graph builder and the shader graph shouldn't the outputs - this should be done at a higher level by the system that binds the outputs - which for the MVP will the standard node material - in the future, the shader graph will be used by other systems like the particle materials and post processes. Therefore, I altered the examples given in the JSDoc

src/scene/materials/standard-node-material.js
raytranuk: these blocks are back in - but I had to mark them as `@private` because the class is private

src/scene/materials/standard-node-material.js
raytranuk: this is now lazily cached to avoid allocation here

examples/graphics/shader-graph-chunks-builder.html
raytranuk: done - thanks!

src/graphics/program-lib/programs/standard-node.js
raytranuk: I have improved this - I haven't turned it into a generic data driven system because it is still tied to the non-data driven standard material.

src/graphics/program-lib/programs/standard-node.js
raytranuk: I have improved this - I haven't turned it into a generic data driven system because it is still tied to the non-data driven standard material.

src/graphics/program-lib/programs/standard-node.js
raytranuk: I have improved this - I haven't turned it into a generic data driven system because it is still tied to the non-data driven standard material.

src/graphics/program-lib/programs/standard.js
raytranuk: updated

src/graphics/program-lib/programs/standard-node.js
raytranuk: I have improved this - I haven't turned it into a generic data driven system because it is still tied to the non-data driven standard material - and IMO it would become harder to maintain/debug.

src/scene/materials/shader-graph-node.js
raytranuk: done - thanks!

src/scene/materials/shader-graph-node.js
raytranuk: done - thanks!

src/scene/materials/shader-graph-node.js
raytranuk: done - thanks!

src/scene/materials/standard-node-material.js
raytranuk: restored - thanks!

src/scene/materials/standard-node-material.js
raytranuk: done - thanks!

src/scene/materials/shader-graph-node.js
raytranuk: added a log - also revealed a small logic error - thanks!

src/scene/materials/shader-graph-node.js
raytranuk: more comments added

src/scene/materials/shader-graph-node.js
raytranuk: I've added optimizations remove/cache common string operations - however the main purpose of the class is to generate a glsl strings - so it doesn't make sense to go too far IMO

src/scene/materials/shader-graph-node.js
raytranuk: optimized

src/scene/materials/shader-graph-node.js
raytranuk: class created - does look better - thanks!

src/scene/materials/shader-graph-node.js
raytranuk: I have refactored and removed prefixes entirely and replaced with PORT_TYPE_ enum - overall it is improved I think - thanks!

src/graphics/shader-node-registry.js
raytranuk: warning added

examples/graphics/shader-graph-chunks-builder.html
raytranuk: done

src/scene/materials/shader-graph-builder.js
raytranuk: Validation has been added

src/scene/materials/shader-graph-builder.js
raytranuk: Validation - with improved error messages - has been added

src/scene/materials/shader-graph-builder.js
raytranuk: https://jsdoc.app/tags-param.html#multiple-types-and-repeatable-parameters - seems it should be {...*}

raytranuk: done

src/scene/materials/shader-graph-builder.js
raytranuk: thanks

raytranuk:
> >    **Public documentation uniformity is of course very important.**
I did a pass - should be almost ready - we can/should still do a pass when/if the docs are made public 
> >    **SG stands for Shader Graph - I can change it.**
changed to SHADERGRAPH_PS - I kept PS because it is commonly used in engine materials and shader chunks 
> >    **The option I was leaning toward was a WebGPU version of the Core Nodes and some minor changes to the Shader Graph systems.**
I've removed GLSL references in the code - at some point any remaining GLSL format strings ideally should be moved to asset files.

raytranuk:
> 6. how to add a constant to the graph (not a uniform, which has higher cost - cpu needs to set it every frame, takes uniform slot, storage on each material ..) - I don't see API for this, unless I missed it.

Constants are now exposed in the builder - enjoy

src/scene/materials/shader-graph-builder.js
```diff
@@ -0,0 +1,221 @@
...
+     * @param {...*} arguments - First argument is name string, then 0-N output node ports or params to
+     * connect to new core node inputs - number and port type of arguments must match.
+     * @returns {number} Returns the created node id.
+     * @example
+     * var node0 = builder.addNode('uv0');
+     * var node1 = builder.addNode('texSample', param0, node0 );
+     * var node2 = builder.addNode('splitVec4', node1 );
+     * var node3 = builder.addNode('mul3', param1, node42 );
+     */
+    addNode: function (name) {
```
mvaligursky, line 126: I wonder if this should be changed to take an array of ports to connect to inputs, instead of just list of arguments.

currently we call it
`var emissiveCalcNode = builder.addNode('mul3', emissiveColConst, {node:emissiveSampleNode, port:'rgb'});`

but we could change it to 
```
addNode: function (name, inputs)

// and call it as 
var emissiveCalcNode = builder.addNode('mul3', 
    [ emissiveColConst, {node:emissiveSampleNode, port:'rgb'} ]
 );
```

this would make it easier to read and add additional parameters at a later stage, for example:

```
addNode: function (name, inputs, precision)

// and call it as 
var emissiveCalcNode = builder.addNode('mul3', [ emissiveColConst, {node:emissiveSampleNode, port:'rgb'} ] );

var emissiveCalcNode = builder.addNode('mul3', 
    [ emissiveColConst, {node:emissiveSampleNode, port:'rgb'} ],
    pc.PRECISION_HIGH,
    ...
 );

```

It would also make it easier to do meaningful type checking using typescript.

raytranuk:
Dropping this PR back to draft to finish off Core Node Generators.

src/scene/materials/shader-graph-builder.js
raytranuk: Thanks for this suggestion - I have done this as part of the Core Node Generators WIP commit.

extras/shader-nodes/texSample.js
```diff
@@ -0,0 +1,24 @@
...
+// ports
+texSample.meta.ports = [];
+texSample.meta.ports[0] = { label: 'TEXTURE', id: 0 };
+texSample.meta.ports[1] = { label: 'UV', id: 1 };
+texSample.meta.ports[2] = { label: 'RGBA', id: 2 };
+texSample.meta.ports[3] = { label: 'RGB', id: 3 };
+texSample.meta.ports[4] = { label: 'R', id: 4 };
+texSample.meta.ports[5] = { label: 'G', id: 5 };
+texSample.meta.ports[6] = { label: 'B', id: 6 };
+texSample.meta.ports[7] = { label: 'A', id: 7 };
```
mvaligursky, line 16: write these in mode condensed form I think. Example:

```
var texSample = {
code = "something",
meta = {
    label: "blah",
    ports = [
         { label: 'TEXTURE', id: 0 },
         { label: 'UV', id: 1 
    ]
    }
}
```

extras/shader-nodes/add.js
```diff
@@ -0,0 +1,17 @@
+import { flexOp } from "./flexOp";
```
mvaligursky, line 1: for testing it's ok, but don't add them to /extras directly, create a new folder for these and add build step similar to extras to build them into separate js file.

mvaligursky: see the end of rollup.config.js

extras/shader-nodes/texSample.js
mvaligursky: even one more idea would be to separate meta data that are needed for shader generation from metadata only used in editor .. so that at a later stage we can exclude them from the file. 
```
var texSample = {
code = "something",
meta = {
    ports = [.  // maybe not needed if only ids?
         { id: 0 },
         { id: 1 
    ]
    },
editor = {
    label: "blah",
    ports = [
         { label: 'TEXTURE'},
         { label: 'UV'}
    ]
}
```

extras/shader-nodes/add.js
raytranuk: Done (adding to it extras was just for testing)

raytranuk:
NB Was working on dynamic switches - which also benefit from the Core Node Generator flexibility of having N arguments - when I realized I had all the pieces to implement static switches - so I did - and they are functional. Will test more - but looks like Render Pass, Platform/API/Extension, Quality Setting and user defined static switches will work.

extras/shader-nodes/texSample.js
raytranuk: Thanks for this - I will tidy this up soon - before implementing any more core shaders nodes

raytranuk:
After implementing the flexible operator and flexible switch in core node generation - it seemed that adding flexibility into all connections would also be possible - so I have added:
- per connection swizzle - just specify the swizzle when passing parameters to addNodes()
- automatic channel truncation/extension/broadcast - this works with or without swizzle

extras/shader-nodes/texSample.js
raytranuk: Tidied up - split up editor data from other data (explicit port ids) - should be possible to split editor data into `playcanvas-shader-nodes-editor.js` later?

mvaligursky: I think add generator function directly inside, vs having to have separate cross.gen block.

I thought it could be easier to build two js files from these, and in one simply strip the editor part .. similarly to how we strip spacing and comments from shaders and code.

Either way, should be doable.

src/scene/materials/standard-node-material.js
```diff
@@ -0,0 +1,189 @@
...
+    },
+
+    setParameters: function (device, names, restoreGraphDefaults) {
+        StandardMaterial.prototype.setParameters.call(this, device, names);
+
+        // apply / restore graph chunk overrides
+        if (this.shaderGraphChunk) {
+            if (restoreGraphDefaults === true) {
+                // restore
+                this.shaderGraphChunk.setParameters(device, Object.keys(this._graphParamOverrides), null);
```
mvaligursky, line 88: Object.keys allocates array, which is not ideal. We should either store overrides in an array directly, or just pass the object as is and iterate over it.

src/scene/materials/standard-node-material.js
```diff
@@ -0,0 +1,189 @@
...
+        this.dirtyShader = false;
+    }
+});
+
+Object.defineProperty(StandardNodeMaterial.prototype, 'shaderGraphChunk', {
+    get: function () {
+        return this._shaderGraphChunk;
+    },
+    set: function (value) {
+        if (!value || this._shaderGraphChunk === value)
```
mvaligursky, line 169: just curious about the !value here. Should the code that removes current chunk execute? When I set something to null, I'd expect current value to be removed.

src/scene/materials/standard-node-material.js
```diff
@@ -0,0 +1,189 @@
...
+            return;
+
+        // remove this material from current chunk's material list
+        if (this._shaderGraphChunk) {
+            var self = this;
+            this._shaderGraphChunk._materials = this._shaderGraphChunk._materials.filter(function (mat) {
```
mvaligursky, line 175: ideally we do this in place without extra array allocation

src/scene/materials/shader-graph-builder.js
```diff
@@ -0,0 +1,350 @@
...
+            // specific port of a node - maybe with swizzle
+            this._graph.connect(arg.node, arg.port, -1, ioPort.name, arg.swizzle);
+        }
+    },
+    /**
+     * @private
+     * @function
+     * @name pc.ShaderGraphBuilder#getShaderGraphChunk
+     * @description Get shader graph chunk.
+     * @returns {any} Shader graph chunk.
```
mvaligursky, line 343: should the output type be a real type rather then any?

src/scene/materials/standard-node-material.js
```diff
@@ -0,0 +1,189 @@
...
+
+/**
+ * @private
+ * @class
+ * @name pc.StandardNodeMaterial
+ * @augments pc.StandardMaterial
+ * @classdesc StandardNodeMaterial is sub class of the StandardMaterial, and adds shader graph
+ * interop functionality.
+ * @param {pc.StandardMaterial} mat - Optional material which is cloned.
+ * @param {any} chunk - Optional shader graph node chunk to be used.
```
mvaligursky, line 17: does it take "any" parameter?

src/scene/materials/shader-graph-node.js
```diff
@@ -0,0 +1,860 @@
...
+        var ioPort = this.getIoPortByName(name);
+        if (ioPort && value !== undefined) {
+            if (!((value instanceof Texture || typeof(value) === 'number') ? value === ioPort._value : value.equals(ioPort._value))) {
+                this._addIoPort(ioPort.type, name, value, ioPort.ptype );
+            }
+        }
+    },
+
+    getIoPortByName: function (name) {
+        return (this._portMap[name] === undefined) ? undefined : this.graphData.ioPorts[this._portMap[name]];
```
mvaligursky, line 524: this._portMap[name] should not execute twice

src/scene/materials/shader-graph-node.js
```diff
@@ -0,0 +1,860 @@
...
+    },
+
+    getIoPortUniformName: function (name) {
+        return (this._portMap[name] === undefined) ? undefined : this._getCachedPort(this._portMap[name]).nameId;
```
mvaligursky, line 528: same here, this._portMap[name] should not execute twice

src/scene/materials/shader-graph-node.js
```diff
@@ -0,0 +1,860 @@
...
+        return clone;
+    },
+
+    _flushCachedPorts: function () {
+        this._portCache = {};
+    },
+
+    _getCachedPort: function (i) {
+        var ioPort = this.graphData.ioPorts[i];
+        if (!this._portCache[ioPort.name]) this._portCache[ioPort.name] = { isUniform: (ioPort.ptype === PORT_TYPE_IN || (ioPort.ptype === PORT_TYPE_CONST && ioPort.type === 'sampler2D')), nameId: ioPort.name + '_' + this.id, port: ioPort, index: i };
```
mvaligursky, line 132: make it a multiline :)

examples/graphics/shader-graph-chunks-overrides.html
```diff
@@ -0,0 +1,215 @@
...
+                    app.assets.loadFromUrl("../assets/textures/particles-numbers.png", "texture", function (err, asset) {    
+                        var numbersMap = asset.resource;            
+
+                        // build one graph
+                        // create a shader graph builder
+                        var builder = new pc.ShaderGraphBuilder(app);
+
+                        // add input parameters
+                        var timeParam = builder.addParam('float', 'uTime', 0.0);
+                        var texParamA = builder.addParam('sampler2D', 'uTexMapA', logoMap);
```
mvaligursky, line 107: the value is described as optional (and should be optional), but crashes when not specified.

examples/graphics/shader-graph-chunks-builder.html
```diff
@@ -0,0 +1,176 @@
...
+                    var cloudsMap = asset.resource;            
+                app.assets.loadFromUrl("../assets/textures/particles-numbers.png", "texture", function (err, asset) {    
+                    var numbersMap = asset.resource;            
+
+                    // create a shader graph builder
+                    var builder = new pc.ShaderGraphBuilder(app);
+
+                    // add input parameters
+                    var timeParam = builder.addParam('float', 'uTime', 0.0);
+                    var texParamA = builder.addParam('sampler2D', 'uTexMapA', cloudsMap);
```
mvaligursky, line 88: the value (cloudsMap) is described as optional (and should be optional), but crashes when not specified.

examples/graphics/shader-graph-chunks-builder.html
```diff
@@ -0,0 +1,176 @@
...
+                            switchValue = 1-switchValue;
+                        } 
+
+                        shaderGraphMaterials.forEach(function (material) {
+                            if (flipSwitch)
+                            {
+                                material.overrideGraphSwitch(switchName, switchValue);
+                                material.update();
+                            }
+                            material.overrideGraphParameter('uTime', time);
```
mvaligursky, line 160: I'd prefer these would be set as any other material properties using setParameter to make it simpler for users. A programmer might not know if it is parameter or override, and that can be changed and code would need to be updated.

shader-nodes/shader-nodes/cross.js
```diff
@@ -0,0 +1,37 @@
...
+// generator
+cross.gen = function ( argTypes, options ) {
+    var code = this.code;
+
+    if (options && options.precision) {
+        // precision - alter name to create variant
+        code = code.replace('vec3 cross', 'vec3 cross_' + options.precision);
+        // precision - tmp: useful comment
+        code = code.replace( '{\n', '{\n// precision ' + options.precision + ' float;\n');
+    }
```
mvaligursky, line 32: I think you should try something more elegant than starting with a code string and then applying some hard to read string.replace calls.

Maybe even template strings directly (see https://hacks.mozilla.org/2015/05/es6-in-depth-template-strings-2/) - you'd need to merge master to your branch to get ES6 support. Or something along similar format to allow easy replacements are runtime, and good readability.

(adding as a screenshot to get better highlighting)
Screen Shot 2020-12-31 at 2 29 53 PM

src/scene/materials/shader-graph-builder.js
```diff
@@ -0,0 +1,350 @@
...
+            } else {
+                if (!arg.port) {
+                    arg = { node: arg.node, port: 'ret', swizzle: arg.swizzle };
+                }
+
+                // node and port - maybe with swizzle
+                var core = this._addedNodes[arg.node];
+                if (!core) {
+                    console.error(err + ": invalid node index: " + arg.node);
+                    return false;
```
mvaligursky, line 150: the function returns an array .. so for type consistency, this should return null instead of false.

mvaligursky:
small comment, but I'll bring it up just for consideration. This is few lines of generated shader:

```
vec2 temp_vec2_0;
vec2 temp_vec2_1;
temp_vec2_0 = uv0(  );
temp_vec2_1 = add( float(uTime), vec2(temp_vec2_0) );
```

could this be generated instead? For readability:

```
vec2 temp_vec2_0 = uv0(  );
vec2 temp_vec2_1 = add( float(uTime), vec2(temp_vec2_0) );
```

src/graphics/shader-node-registry.js
```diff
@@ -0,0 +1,88 @@
...
+        for (var keyIndex = 0; keyIndex < optionKeys.length; keyIndex++) {
+            variantKey += '_' + optionKeys + '_' + options[optionKeys[keyIndex]];
+        }
+    }
+
+    return hashCode(variantKey);
+};
+
+ShaderNodeRegistry.prototype.get = function (name, argTypes, options) {
+    if (this._nodeCache[name]) {
```
mvaligursky, line 61: you should store the result of this instead of executing it many times, as each of these is a dictionary lookup.
`this._nodeCache[name]`

examples/graphics/shader-graph-chunks-builder.html
raytranuk: hmm - probably best to keep it optional, and have the builder put in a placeholder if not supplied - I'll see if I can add a placeholder (this is what materials do (in the material handler) I think)

examples/graphics/shader-graph-chunks-builder.html
raytranuk: If using the same function name it wouldn't be possible to validate and give warnings if the parameter exists on the graph or not - which could be useful to developers - it also could lead to bugs that occur if the intention wasn't to override a parameter or switch, leading to odd behavior or performance - all of this can also be said of the mesh instance overrides which do use the same function name - so at least it will be consistently tricky to debug :)

src/graphics/shader-node-registry.js
```diff
@@ -0,0 +1,88 @@
...
+    var variantKey = 'varKey';
+
+    for (var argIndex = 0; argIndex < argTypes.length; argIndex++) {
+        variantKey += '_' + argTypes[argIndex];
+    }
+
+    if (options) {
+        var optionKeys = Object.keys(options);
+        for (var keyIndex = 0; keyIndex < optionKeys.length; keyIndex++) {
+            variantKey += '_' + optionKeys + '_' + options[optionKeys[keyIndex]];
```
mvaligursky, line 53: I think `+ optionKeys +` is not what you want there, as optionKeys is an array and that add all array keys there, should be `+ optionKeys[keyIndex] +` I suspect?
There is probably better way to walk this anyways, without calling `Object.keys` which creates an array, and without having to do lookup on each step `options[optionKeys[keyIndex]]`

src/scene/materials/standard-node-material.js
raytranuk: Good point - maintaining an array alongside the object makes the most sense.

examples/graphics/shader-graph-chunks-builder.html
mvaligursky: well a user of a shader does not need to understand internals of the material / graph implementation .. in the future they get a shader graph material from a store, or somebody else on the team builds it, and they just need to set parameters.

src/scene/materials/standard-node-material.js
raytranuk: I will change it to splice()

src/scene/materials/standard-node-material.js
raytranuk: yeah not sure why its here - I shall remove it

src/scene/materials/shader-graph-builder.js
raytranuk: It should be `ShaderGraphNode` - I'll update it

src/scene/materials/standard-node-material.js
raytranuk: It should be ShaderGraphNode - I'll update it

src/scene/materials/shader-graph-node.js
raytranuk: I will add a var

src/scene/materials/shader-graph-node.js
raytranuk: I will add a var

src/scene/materials/shader-graph-node.js
raytranuk: okay

shader-nodes/shader-nodes/cross.js
raytranuk: ooh nice - I'll merge master and use this - we can also improve a lot of the shader chunk stuff in the generator

src/scene/materials/shader-graph-builder.js
raytranuk: I'll change it

examples/graphics/shader-graph-chunks-builder.html
raytranuk: > small comment, but I'll bring it up just for consideration. These are a few lines of generated shader:
> 
> ```
> vec2 temp_vec2_0;
> vec2 temp_vec2_1;
> temp_vec2_0 = uv0(  );
> temp_vec2_1 = add( float(uTime), vec2(temp_vec2_0) );
> ```
> 
> could this be generated instead? For readability:
> 
> ```
> vec2 temp_vec2_0 = uv0(  );
> vec2 temp_vec2_1 = add( float(uTime), vec2(temp_vec2_0) );
> ```

It could - I would have to add a flag of some sort to indicate the output is a return value - but certainly doable - and make the shader a bit shorter

raytranuk:
> small comment, but I'll bring it up just for consideration. These are a few lines of generated shader:
> 
> ```
> vec2 temp_vec2_0;
> vec2 temp_vec2_1;
> temp_vec2_0 = uv0(  );
> temp_vec2_1 = add( float(uTime), vec2(temp_vec2_0) );
> ```
> 
> could this be generated instead? For readability:
> 
> ```
> vec2 temp_vec2_0 = uv0(  );
> vec2 temp_vec2_1 = add( float(uTime), vec2(temp_vec2_0) );
> ```

Should be doable - would need to add a flag to indicate the output is a return value.

src/graphics/shader-node-registry.js
raytranuk: Not sure what you mean here - the node registry is just a single instance per app - and this core node cache is only used when building graphs - which happens only on shader graph editing/building or generation.

shader-nodes/shader-nodes/cross.js
raytranuk: I'm planning to look into this - but is it necessary for this PR? This PR isn't supposed to contain the final list of core nodes or core node generators - just the underlying systems to allow for them to be developed (alongside the MVP graph editor). In fact it would be great (and probably necessary) if core nodes could be developed by several developers and eventually by external contributors

src/graphics/shader-node-registry.js
mvaligursky: What I meant what that instead of calling this many times:
`this._nodeCache[name]`
ideally you store it in a var and reuse:
`var namedNode = this._nodeCache[name];`

shader-nodes/shader-nodes/cross.js
mvaligursky: My opinion is that any code that makes it into code base needs to be a good quality, as we wont be reviewing it again. I agree all core nodes can be developed by multiple programmers at a later stage, but here we're looking at a system that allows us to do it effectively. But I understand that it's not easy to develop everything in a larger PR to best quality, and so I'll call @willeastcott for an opinion here as well.

src/scene/materials/standard-node-material.js
raytranuk: refactored

src/scene/materials/standard-node-material.js
raytranuk: removed

src/scene/materials/standard-node-material.js
raytranuk: refactored - thanks!

src/scene/materials/shader-graph-builder.js
raytranuk: updated

src/scene/materials/standard-node-material.js
raytranuk: updated

src/scene/materials/shader-graph-node.js
raytranuk: improved

src/scene/materials/shader-graph-node.js
raytranuk: improved

src/scene/materials/shader-graph-node.js
raytranuk: changed

examples/graphics/shader-graph-chunks-builder.html
raytranuk: refactored - Texture parameters are now treated separately - with additional validation and placeholder.

examples/graphics/shader-graph-chunks-builder.html
raytranuk: I've made general parameters use setParameters - but I am treating switches differently - because they are different.

src/scene/materials/shader-graph-builder.js
raytranuk: done

src/graphics/shader-node-registry.js
raytranuk: yes - it was a typo - refactored to fix this and be more efficient - and looks simpler

src/graphics/shader-node-registry.js
raytranuk: ah - okay - good point - and done.

shader-nodes/shader-nodes/cross.js
raytranuk: IMO replace() isn't low quality code - it is what was possible without newer es6 language features - and is used in the engine in the shader generators for all material texture map UVs.

raytranuk: I've replaced `replace()` (and some other string generation) with string templates - it does increase readability a bit.

src/scene/materials/shader-graph-builder.js
```diff
@@ -0,0 +1,393 @@
...
+
+    this.app = app;
+
+    this._graph = new ShaderGraphNode();
+    this._graph.name = 'graphRoot_' + id;
+
+    // for validation
+    this._addedNamedInputs = {};
+    this._addedNodes = [];
+    this._addedOutputs = {};
```
mvaligursky, line 30: curious why these names start with 'added'?
Maybe just _namedInputs, _nodes, _outputs?

src/scene/materials/shader-graph-builder.js
```diff
@@ -0,0 +1,393 @@
...
+                pixels[i * 4 + c] = [128, 128, 128, 255];
+            }
+        }
+        _placeholderTex.unlock();
+    }
+
+    this._placeholderTex = _placeholderTex;
+};
+
+ShaderGraphBuilder.prototype.constructor = ShaderGraphBuilder;
```
mvaligursky, line 54: is this line needed or just some leftover from when there was different base class?

src/scene/materials/shader-graph-builder.js
```diff
@@ -0,0 +1,393 @@
...
+        });
+        _placeholderTex.name = 'placeholder';
+
+        // fill pixels with grey
+        var pixels = _placeholderTex.lock();
+        for (var i = 0; i < 4; i++) {
+            for (var c = 0; c < 4; c++) {
+                pixels[i * 4 + c] = [128, 128, 128, 255];
+            }
+        }
```
mvaligursky, line 47: This does not fill the texture with those pixels .. you're assigning an array of 4 elements to each pixel, that does do what you need. The final pixels array is just full of zeros.
Maybe look .fill, as you get typedArray from .lock()
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/fill

src/scene/materials/shader-graph-builder.js
```diff
@@ -0,0 +1,393 @@
...
+        _placeholderTex.unlock();
+    }
+
+    this._placeholderTex = _placeholderTex;
```
mvaligursky, line 51: I think don't assign _placeholderTex to this._placeholderTex and just use _placeholderTex directly when you need to access it. It's static, and so no need to assign it to instance of the class.

src/scene/materials/shader-graph-builder.js
raytranuk: I guess they can be renamed - they are just private data used for validation - maybe they can be tidied up by putting them into a _validationData struct?

src/scene/materials/shader-graph-builder.js
raytranuk: Ah yes - made a mistake in adjusting this code when copying from the asset handler - I will switch to fill - thanks!

src/scene/materials/shader-graph-builder.js
raytranuk: ok - will do

src/scene/materials/shader-graph-builder.js
mvaligursky: good idea.

src/scene/materials/shader-graph-builder.js
raytranuk: I think it was indeed a leftover - removing

shader-nodes/shader-nodes/flexOp.js
```diff
@@ -0,0 +1,65 @@
...
+        for (var argIndex = 0; argIndex < argTypes.length; argIndex++) {
+            // convert all arguments to return type
+            code += `${retType}(arg${argIndex}`;
+
+            // expand if needed - extra components are set to zero - NB scalar values are broadcast to all components
+            var argComp = type2Comp[argTypes[argIndex]];
+            var extend = type2Comp[retType] - argComp;
+            code += (extend > 0 && argComp !== 1) ? `, ${comp2Type[expand]}(0))` : ')';
+
+            code += (argIndex === argTypes.length - 1) ? ';\n' : ` ${opCode} `;
```
mvaligursky, line 56: it feels like this should be on a higher level when the generated node function gets called .. so that it works any node parameters (and not only flex nodes), without having to implement this for each core node / subgraph?

raytranuk: Auto type conversion actually already does work on all nodes in the graph - in combination with (an optional) per connection swizzle - which makes it very flexible - so you can feed a scalar value into vector port or vice versa or a vec2 into a vec3 or a vec4 into a vec3 etc and it gets automatically extended or truncated: https://github.com/playcanvas/engine/blob/3b7e6dbf40c1f1e1a844234e5dd9579da712cf7e/src/scene/materials/shader-graph-node.js#L423 - its pretty easy to try in the engine examples - I can also add one that demos Auto type conversion? It should also become more intuitive when implemented in the graph editor with colors (like in Lens Studio) and/or numbers (link in Unity) indicating the type conversion.

The flexOp is actually distinct from general auto type conversion because it and can take a flexible number of arguments of any (scalar or vector) type and changes the type of the operation itself and based on the argument with the largest number of components - this means other arguments may need to be extended.

Issue #2496, opened 2020-10-26
Weird grabpass and texture_grabPass behavior


devcem:
### Description
I was working on grabpass shader and noticed it doesn't work on my friends's computer, on my phone and safari. But works properly on my computer (Chrome latest version) for some reason. Then I wanted to prepare a reproducing demo but this time it doesn't work at all even on my computer.

For some reason on different project or computers "texture_grabPass" doesn't provide anything.

## Demo that works my computer on first try :
https://playcanv.as/b/QOQGf2Z9/

## Demo that I prepared for reproducing step, simpler 
https://playcanv.as/b/6wOtRyWu/

Error : launch.js:7988 ERROR: FRAMEBUFFER_INCOMPLETE_ATTACHMENT
[.WebGL-0x7fd05bae6e00]GL ERROR :GL_INVALID_OPERATION : glCopyTexImage2D: no valid color image

 I couldn't figure out, maybe it's my mistake, but on simpler version the image I put as a sampler2D works properly. But "texture_grabPass" doesn't work.

mvaligursky:
Hi @devcem - are you able to share the Editor project for the simpler repro as well?
Thanks!

devcem:
> Hi @devcem - are you able to share the Editor project for the simpler repro as well?
> Thanks!

Hello,
Here is the editor version of simpler repo : 
https://playcanvas.com/editor/project/731620

mvaligursky:
I'm still looking for the best solution, but it seems that changing the 'transparent canvas' setting in Rendering section can fix it. I'm still working on a solution that works with either state of that setting - and have one apart from Chrome.

PR #2504, opened 2020-10-28
Grabpass compatibility improvement with  alpha / antialiasing settings combinations

mvaligursky:
- previously alpha==false case was not handled
- also antialising && !alpha has special handling as Chromium throws errors (even though it works)

fixes https://github.com/playcanvas/engine/issues/2496

### DIFF
--- a/src/graphics/device.js
+++ b/src/graphics/device.js
@@ -685,7 +685,7 @@ var GraphicsDevice = function (canvas, options) {
     this._spectorCurrentMarker = "";
     // #endif
 
-    this.createGrabPass();
+    this.createGrabPass(options.alpha, options.antialias);
 
     VertexFormat.init(this);
 };
@@ -1020,11 +1020,14 @@ Object.assign(GraphicsDevice.prototype, {
         this.transformFeedbackBuffer = null;
     },
 
-    createGrabPass: function () {
+    createGrabPass: function (alpha, antialias) {
         if (this.grabPassTexture) return;
 
+        // blitFramebuffer is throwing errors on Chromium when antialias is true and alpha is false, use webgl1 path in that case
+        this.grabPassUseWebGl2Path = this.webgl2 && (!(alpha === false && antialias === true));
+
         var grabPassTexture = new Texture(this, {
-            format: PIXELFORMAT_R8_G8_B8_A8,
+            format: alpha === false ? PIXELFORMAT_R8_G8_B8 : PIXELFORMAT_R8_G8_B8_A8,
             minFilter: FILTER_LINEAR,
             magFilter: FILTER_LINEAR,
             addressU: ADDRESS_CLAMP_TO_EDGE,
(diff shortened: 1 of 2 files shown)


src/graphics/device.js
```diff
@@ -685,7 +685,7 @@ var GraphicsDevice = function (canvas, options) {
     this._spectorCurrentMarker = "";
     // #endif
 
-    this.createGrabPass();
+    this.createGrabPass(options.alpha, options.antialias);
```
willeastcott, line 688: So, I might be wrong here, but isn't it the case that a WebGL implementation is not required to honor the context creation options? So you could *request* alpha, but not necessarily get it? Same deal with antialising.

mvaligursky: That is possible. What other way to detect this is there? These functions are not supported on all platforms:
https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/getFramebufferAttachmentParameter
https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/getRenderbufferParameter

willeastcott: Hey @kenrussell, do you have any tips for this?

kenrussell: Per the spec, https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.2.1 , alpha is required to be honored exactly as specified. antialias is essentially a hint, and might not be honored.

I'm not sure exactly what information you need but you can figure out whether antialiasing was enabled for the back buffer by querying the SAMPLES parameter:

<pre>
let c = document.createElement('canvas');
let gl = c.getContext('webgl');
gl.getParameter(gl.SAMPLES)
> 4

let c2 = document.createElement('canvas');
let gl2 = c2.getContext('webgl', { antialias: false });
gl2.getParameter(gl.SAMPLES)
> 0
</pre>

If there seems to be some bug in Chromium with the combination of antialias:true and alpha:false, please provide a small test case and file a bug on crbug.com . Please include about:gpu from your system when doing so, and ping me the bug ID. Thanks.

willeastcott: Thanks for the words of wisdom @kenrussell! Much appreciated. 😄

mvaligursky: @kenrussell - thanks for the info, that was helpful.
I have created an issue now - only can reproduce in Chrome on Mac, there was similar issue in the past already, see the description: https://bugs.chromium.org/p/chromium/issues/detail?id=1156756

src/graphics/device.js
```diff
@@ -2123,10 +2148,16 @@ Object.assign(GraphicsDevice.prototype, {
                 texture._parameterFlags = 0;
             }
 
+            // grag framebuffer to be used as a texture - this returns false when not supported for current render pass
```
slimbuck, line 2151: grag?

mvaligursky: ;) fixing.

willeastcott (approved):

PR #2512, opened 2020-10-30
Add tweening example

willeastcott:
Adds an example showing how to do tweening in PlayCanvas using [tween.js](https://github.com/tweenjs/tween.js/), a popular open-source tweening library.

tween

This PR includes a script that can be dropped in to the Editor and provides the following interface:

image

You'll also need to add the tween.js library to your project that can be downloaded from [here](https://github.com/tweenjs/tween.js/blob/master/dist/tween.umd.js).

### DIFF
--- a/examples/assets/scripts/animation/tween.js
+++ b/examples/assets/scripts/animation/tween.js
@@ -125,10 +125,10 @@ Tween.prototype.initialize = function () {
     this.tweenInstances = [];
     this.tweenCallbacks = [];
 
-    var makeStartCallback = function (i) {  
-       return function() {  
-          this.start(i);
-       };
+    var makeStartCallback = function (i) {
+        return function () {
+            this.start(i);
+        };
     };
 
     for (i = 0; i < this.tweens.length; i++) {
@@ -192,8 +192,8 @@ Tween.prototype.start = function (idx) {
     var app = this.app;
     var tween = this.tweens[idx];
 
-    var easingTypes = [ 'In', 'Out', 'InOut' ];
-    var easingFuncs = [ 'Linear', 'Quadratic', 'Cubic', 'Quartic', 'Quintic', 'Sinusoidal', 'Exponential', 'Circular', 'Elastic', 'Back', 'Bounce'];
+    var easingTypes = ['In', 'Out', 'InOut'];
+    var easingFuncs = ['Linear', 'Quadratic', 'Cubic', 'Quartic', 'Quintic', 'Sinusoidal', 'Exponential', 'Circular', 'Elastic', 'Back', 'Bounce'];
 
     var easingFunc;
     if (tween.easingFunction === 0) {
--- a/examples/assets/scripts/animation/tween.js
+++ b/examples/assets/scripts/animation/tween.js
@@ -1,0 +1,318 @@
+var Tween = pc.createScript('tween');
+
+Tween.attributes.add('tweens', {
+    type: 'json',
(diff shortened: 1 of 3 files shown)


examples/assets/scripts/animation/tween.js
```diff
@@ -0,0 +1,318 @@
...
+
+Tween.attributes.add('tweens', {
+    type: 'json',
+    schema: [
+        {
+            name: 'autoPlay',
+            title: 'Autoplay',
+            description: 'Play tween immediately.',
+            type: 'boolean',
+            default: false
```
willeastcott, line 11: Hey @vkalpias - I can't get defaults working for array-based JSON attributes. Can you provide an example? I'd like to update the User Manual to cover this topic as well.

vkalpias: I believe the default is set on the JSON attribute, not the individual schema fields. e.g.:

```
Test.attributes.add('json',{
    type: 'json',
    schema: [{
        name: 'field1',
        type: 'number'
    }, {
        name: 'field2',
        type: 'string'
    }],
    default: {
        field1: 4,
        field2: 'hello world'
    }
});

```

willeastcott: So this is wrong? https://developer.playcanvas.com/en/user-manual/scripting/script-attributes/#json-attribute

willeastcott: And that also works for `array: true`?

examples/assets/scripts/animation/tween.js
```diff
@@ -0,0 +1,318 @@
...
+        for (i = 0; i < this.tweenCallbacks.length; i++) {
+            if (this.tweenCallbacks[i]) {
+                app.off(this.tweenCallbacks[i].event, this.tweenCallbacks[i].cb, this);
+                this.tweenCallbacks[i] = null;
+            }
+        }
+
+        for (i = 0; i < this.tweens.length; i++) {
+            var tween = this.tweens[i];
+            if (tween.event.length > 0) {
```
slimbuck, line 180: no autoplay here?

examples/assets/scripts/animation/tween.js
yaustar: @vkalpias The defaults on the schema work fine in the Editor:
```
var Sandbox = pc.createScript('sandbox');
Sandbox.attributes.add('settings', {
    type: 'json',
    schema: [{
        name: 'gravity',
        type: 'number',
        default: -9.8
    }, {
        name: 'startingHealth',
        type: 'number',
        default: 20
    }, {
        name: 'godMode',
        type: 'boolean',
        default: false
    }]
});
```

willeastcott: Yep, it just doesn't work for the engine-only case.

willeastcott: Defaults now fixed here: https://github.com/playcanvas/engine/pull/2525  Resolving...

examples/assets/scripts/animation/tween.js
willeastcott: No, because we're just rewiring the (potentially changed) events here. Even if `autoPlay` on a tween was flipped from `false` to `true`, I'm not sure the tween should be started. Autoplay should occur on script initialization alone IMHO.

vkalpias (approved):

slimbuck (approved):

PR #2517, opened 2020-11-02
Add debug message when using a destroyed texture

slimbuck:
Following on from #2515, issue an error message in development builds when a destroyed texture is referenced in the scene.

### DIFF
--- a/src/graphics/device.js
+++ b/src/graphics/device.js
@@ -1785,6 +1785,12 @@ Object.assign(GraphicsDevice.prototype, {
     },
 
     uploadTexture: function (texture) {
+        // #ifdef DEBUG
+        if (!texture.device) {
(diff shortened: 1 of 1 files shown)


mvaligursky (approved):

src/graphics/device.js
```diff
@@ -1785,6 +1785,12 @@ Object.assign(GraphicsDevice.prototype, {
     },
 
     uploadTexture: function (texture) {
+        // #ifdef DEBUG
+        if (!texture.device) {
+            console.error("attempting to use a texture that has been destroyed.");
```
willeastcott, line 1790: Isn't this going to potentially spam the JS console since this function is called a lot?

slimbuck: yeah possibly...

willeastcott (approved):

Issue #189, opened 2015-03-20
Box-projected cubemaps: rotation


Maksims:
Currently Box-Projected CubeMaps can't be rotated.  
Means that if you have one same room with same cubemap on them, you cannot simply duplicate that room and rotate, as box projection won't inherit rotation from room, nor it will work with non-parallel walls which creates loads of restrictions on where box-projection can be used.

Consider adding vec3 (euler) of rotation for box-projected cubemap, additionally to inheriting world rotation of entity.

willeastcott:
Another cubemap related feature request to consider, @slimbuck.

Maksims:
Just to note, there can be 2 modes: local and world in terms of transform, both have uses.

willeastcott:
Closing because we now have `Scene#skyboxRotation` (which is a `Quat`):

https://developer.playcanvas.com/api/pc.Scene.html#skyboxRotation

PR #2527, opened 2020-11-04
Add support for Cube Map Rotation and expose Skybox Rotation

### LINKED ISSUE #2505: Allow env cube maps to be rotated
At the moment they cannot be rotated in the world (to say align sun in the cube map with some specific world direction where user wants to place directional light).

We should implement this (possibly as an optional shader chunk).

At the moment we also flip cube map lookup on x-axis. We should perhaps remove the instruction and adjust the 3x3 matrix instead.

```
vec3 calcReflection(vec3 tReflDirW, float tGlossiness) {
    vec3 lookupVec = fixSeams(cubeMapProject(tReflDirW));
    lookupVec.x *= -1.0;  // <===========
    return $textureCubeSAMPLE(texture_cubeMap, lookupVec).rgb;
}

```

This would allow us to use this for dynamically rendered cube maps, which do not need the flip at all.

raytranuk:
Fixes #2505

Public API additions:
```
 * @class
 * @name pc.Scene
 *
 * @property {pc.Quat} skyboxRotation The rotation of the skybox to be displayed. Defaults to pc.Quat.IDENTITY
```
Example:
```
    var r = new pc.Quat();
    r.setFromEulerAngles(42, 42, 0);
    app.scene.skyboxRotation = r;
```


src/graphics/program-lib/chunks/ambientPrefilteredCube.frag
```diff
@@ -4,7 +4,11 @@ uniform samplerCube texture_prefilteredCubeMap4;
 #endif
 
 void addAmbient() {
+#ifdef CUBEMAPROT
+    vec3 fixedReflDir = fixSeamsStatic(dNormalW*cubeMapRotationMatrix, 1.0 - 1.0 / 4.0);
+#else
     vec3 fixedReflDir = fixSeamsStatic(dNormalW, 1.0 - 1.0 / 4.0);
+#endif
     fixedReflDir.x *= -1.0;
```
mvaligursky, line 12: can we remove this line which flips the .x of reflection vector? TO use dynamic cubemaps, this line needs to not be applied, and I was hoping to handle it by using the matrix. Ideally we add this flip into the matrix on the cpu and expose it as a material option, so that it can be disabled for realtime cubemaps.

slimbuck: Since we don't want to add an extra matrix multiply to all shaders, it probably makes sense for `skyboxRotation` to be exposed as 3x3 matrix instead of quaternion so you can scale as necessary.

raytranuk: Switched from Quat to Mat3 and embedded the -x flip in the Matrix - also moved the cube map rotation property to pc.Texture. I had to change a lot of shader chunks in order to embed the -x flip and keep identical behavior for reflection and ambient shader chunks that do not use cube maps. Its quite likely to break some custom shader chunks...

slimbuck (commented):
As a user, I think it's confusing that setting an identity rotation matrix results in a mirror transform. I'd much rather keep the -1x always and optionally multiply in the rotation matrix.

Also, is it possible to have this rotation logic (the -1x and optional rotation matrix) encapsulated into single chunk/function? This way we can modify/customise easily in future.

slimbuck:
Alternatively you could apply the -1x to the matrix when it's being set on the shader.

raytranuk:
> As a user, I think it's confusing that setting an identity rotation matrix results in a mirror transform. I'd much rather keep the -1x always and optionally multiply in the rotation matrix.
>
> Also, is it possible to have this rotation logic (the -1x and optional rotation matrix) encapsulated into single chunk/function? This way we can modify/customise easily in future.
>
> Alternatively you could apply the -1x to the matrix when it's being set on the shader.

I will move the rotation logic into a function and put it in a new chunk.

For the -1x flip, maybe the best way for users is to expose an extra boolean flag which defaults to true to determine is the -1x should be applied to the rotation - and maybe switch back to Quat from Mat3 for the rotation property?

mvaligursky:
> For the -1x flip, maybe the best way for users is to expose an extra boolean flag which defaults to true to determine is the -1x should be applied to the rotation - and maybe switch back to Quat from Mat3 for the rotation property?

yep I agree with this. Ideally we don't convert quat->matrix each frame, so it would need to store additional Mat3 as well?

src/framework/application.js
```diff
@@ -1695,6 +1695,7 @@ Object.assign(Application.prototype, {
      * @param {number|null} [settings.render.skybox] - The asset ID of the cube map texture to be used as the scene's skybox. Defaults to null.
      * @param {number} settings.render.skyboxIntensity - Multiplier for skybox intensity.
      * @param {number} settings.render.skyboxMip - The mip level of the skybox to be displayed. Only valid for prefiltered cubemap skyboxes.
+     * @param {pc.Mat3} settings.render.skyboxRotationMatrix - Rotation of skybox.
```
willeastcott, line 1698: Shouldn't this be:
`* @param {number[]} settings.render.skyboxRotation - Rotation of skybox.`
At least, it should be according to the example code below.

src/graphics/program-lib/chunks/cubeMapRotate.frag
```diff
@@ -0,0 +1,19 @@
...
+#endif
+}
+
+vec3 cubeMapRotate1(vec3 refDir) {
+#ifdef CUBEMAPROT
+    return (refDir*cubeMapRotationMatrix)*vec3(-1,1,1);
+#else
+    return refDir;
+#endif
+}
```
willeastcott, line 19: Remember to stick to our coding style:
* space after comma
* spaces around operators
🤓

src/graphics/program-lib/chunks/reflectionDpAtlas.frag
```diff
@@ -26,7 +26,7 @@ vec2 getDpAtlasUv(vec2 uv, float mip) {
 }
 
 vec3 calcReflection(vec3 tReflDirW, float tGlossiness) {
-    vec3 reflDir = normalize(cubeMapProject(tReflDirW));
+    vec3 reflDir = normalize(cubeMapProject(tReflDirW)*vec3(-1,1,1));
```
willeastcott, line 29: Style (operator)

src/graphics/program-lib/chunks/skyboxHDRRot.frag
```diff
@@ -0,0 +1,11 @@
+varying vec3 vViewDir;
+
+uniform samplerCube texture_cubeMap;
+uniform mat3 cubeMapRotationMatrix;
+
+void main(void) {
+    vec3 color = processEnvironment($textureCubeSAMPLE(texture_cubeMap, fixSeamsStatic((vViewDir*vec3(-1,1,1))*cubeMapRotationMatrix, $FIXCONST)).rgb);
```
willeastcott, line 7: Style (operators and commas)

willeastcott:
Will this fully address #189?

src/framework/application.js
raytranuk: thanks - will also have to do a conversion from array to the the object - which is actually supposed to be pc.Quat not pc.Mat3 I will fix it

src/graphics/program-lib/chunks/cubeMapRotate.frag
raytranuk: falling foul of my reliance on lint - I'll do a spacing pass on all the shaders chunks I touched

src/graphics/program-lib/chunks/reflectionDpAtlas.frag
raytranuk: I'll do a spacing pass on all the shaders chunks I touched

src/graphics/program-lib/chunks/skyboxHDRRot.frag
raytranuk: I'll do a spacing pass on all the shaders chunks I touched

src/graphics/program-lib/chunks/reflectionDpAtlas.frag
willeastcott: Yeah, unfortunately, we can't run ESLint over the GLSL code unfortunately. So we have to catch these things just with our 👁️👁️

raytranuk:
> Just curious why the docs for each property have been made single line? We normally try to make an effort for JSDoc blocks to stick to a reasonable number of columns (the coding standards never specified this maximum though). But it's good to be able to read them without having to scroll horizontally.

I restored it to multi-line

src/graphics/program-lib/chunks/skyboxHDRRot.frag
raytranuk: shader chunk spacing should all be fixed now - thanks!

src/graphics/program-lib/chunks/reflectionDpAtlas.frag
raytranuk: shader chunk spacing should all be fixed now - thanks!

src/graphics/program-lib/chunks/cubeMapRotate.frag
raytranuk: shader chunk spacing should all be fixed now - thanks!

src/framework/application.js
raytranuk: should be fixed now - thanks!

raytranuk:
> Will this fully address #189?

It won't address #189 - this was touched on in the graphics sync - to allow a cube map to be reused at multiple different rotations, we would have to store the rotations of each instance - which we think should be in something like instances of light-probe components that reference cube maps.

willeastcott:
OK, understood.

src/graphics/texture.js
```diff
@@ -65,6 +65,8 @@ import {
  * @param {string} [options.type] - Specifies the image type, see {@link pc.TEXTURETYPE_DEFAULT}
  * @param {boolean} [options.fixCubemapSeams] - Specifies whether this cubemap texture requires special
  * seam fixing shader code to look right. Defaults to false.
+ * @param {boolean} [options.isRenderTarget] - Specifies whether texture or cubemap was used as a render target.
+ * Defaults to false.
```
willeastcott, line 69: Why does this need to be public API?

raytranuk: It doesn't - so I have just made it private

examples/graphics/material-physical-cubemap-rotation.html
```diff
@@ -0,0 +1,136 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>PlayCanvas Physical Material</title>
```
mvaligursky, line 4: Rename to match the file name.
Do we need new very similar example or should we merge those two into a single example?

examples/graphics/material-physical-cubemap-rotation.html
```diff
@@ -0,0 +1,136 @@
...
+        var x = 0;
+        var y = 0;
+        var rot = new pc.Quat();
+
+        mouse.on('mousemove', function (event) {
+            if (event.buttons[pc.MOUSEBUTTON_LEFT]) {
+                x += event.dx;
+                y += event.dy;
+
+                // entity.setLocalEulerAngles(0.2 * y, 0.2 * x, 0);
```
mvaligursky, line 128: remove the comment?

examples/graphics/material-physical-cubemap-rotation.html
willeastcott: I would probably agree here. I'm not sure this warrants a dedicated example to demonstrate a single property.

src/graphics/program-lib/chunks/cubeMapRotate.frag
```diff
@@ -0,0 +1,19 @@
+#ifdef CUBEMAPROT
+uniform mat3 cubeMapRotationMatrix;
+#endif
+
+vec3 cubeMapRotate0(vec3 refDir) {
+#ifdef CUBEMAPROT
+    return refDir*cubeMapRotationMatrix;
+#else
+    return refDir * vec3(-1, 1, 1);
```
mvaligursky, line 9: maybe return vec3(-refDir.x, refDir.yz);

src/graphics/program-lib/chunks/cubeMapRotate.frag
```diff
@@ -0,0 +1,19 @@
+#ifdef CUBEMAPROT
+uniform mat3 cubeMapRotationMatrix;
+#endif
+
+vec3 cubeMapRotate0(vec3 refDir) {
```
mvaligursky, line 5: should these more descriptive suffix than 0 and 1?

src/graphics/program-lib/chunks/skyboxHDRRot.frag
```diff
@@ -0,0 +1,11 @@
+varying vec3 vViewDir;
+
+uniform samplerCube texture_cubeMap;
+uniform mat3 cubeMapRotationMatrix;
+
+void main(void) {
+    vec3 color = processEnvironment($textureCubeSAMPLE(texture_cubeMap, fixSeamsStatic((vViewDir * vec3(-1, 1, 1))*cubeMapRotationMatrix, $FIXCONST)).rgb);
```
mvaligursky, line 7: space around *

examples/graphics/material-physical-cubemap-rotation.html
raytranuk: hmm - I will merge it into one - the skybox rotation could be seen as adding interactivity to the example to better show how physical materials change with lighting.

examples/graphics/material-physical-cubemap-rotation.html
raytranuk: I shall remove this - and some other changes

src/graphics/program-lib/chunks/cubeMapRotate.frag
raytranuk: good idea - I will add NegX and PosX

src/graphics/program-lib/chunks/skyboxHDRRot.frag
raytranuk: well spotted - will fix

src/graphics/program-lib/chunks/cubeMapRotate.frag
raytranuk: yes - I will change it

src/scene/scene.js
```diff
@@ -349,8 +391,12 @@ Scene.prototype.applySettings = function (settings) {
     this._skyboxIntensity = settings.render.skyboxIntensity === undefined ? 1 : settings.render.skyboxIntensity;
     this._skyboxMip = settings.render.skyboxMip === undefined ? 0 : settings.render.skyboxMip;
 
-    this._resetSkyboxModel();
-    this.updateShaders = true;
+    if (settings.render.skyboxRotation !== undefined) {
+        this.skyboxRotation = new Quat(settings.render.skyboxRotation);
+    } else {
+        this._resetSkyboxModel();
+        this.updateShaders = true;
```
mvaligursky, line 398: should _resetSkyboxModel be in else branch?
when settings.render.skyboxRotation si undefined, it won't get called .. but it used to

src/framework/application.js
```diff
@@ -1695,6 +1695,7 @@ Object.assign(Application.prototype, {
      * @param {number|null} [settings.render.skybox] - The asset ID of the cube map texture to be used as the scene's skybox. Defaults to null.
      * @param {number} settings.render.skyboxIntensity - Multiplier for skybox intensity.
      * @param {number} settings.render.skyboxMip - The mip level of the skybox to be displayed. Only valid for prefiltered cubemap skyboxes.
+     * @param {number[]} settings.render.skyboxRotation - Rotation of skybox.
```
mvaligursky, line 1698: should be Quat instead of number[] ?

examples/graphics/material-physical-cubemap-rotation.html
raytranuk: done - thanks

examples/graphics/material-physical-cubemap-rotation.html
raytranuk: done - thanks

src/graphics/program-lib/chunks/cubeMapRotate.frag
raytranuk: also done - thanks

src/graphics/program-lib/chunks/skyboxHDRRot.frag
raytranuk: fixed - thanks

src/framework/application.js
raytranuk: this is correct - I had it as Quat before, but Will pointed out that render settings done as number arrays, and are objects initialized from arrays - look at the example of Color render settings.

src/scene/scene.js
raytranuk: I'm using the object Object.defineProperty() accessors to do extra stuff in the set() accessor - which also calls the reset. So if settings.render.skyboxRotation is undefined - the _resetSkyboxModel is called - and if settings.render.skyboxRotation is defined the _resetSkyboxModel is called in the accessor - so behavior is correct - but maybe I can improve it by using a common private function...

src/graphics/program-lib/chunks/cubeMapRotate.frag
raytranuk: changed - thanks!

src/scene/scene.js
raytranuk: made improvements - thanks!

src/graphics/program-lib/chunks/cubeMapRotate.frag
```diff
@@ -0,0 +1,19 @@
...
+#endif
+}
+
+vec3 cubeMapRotatePosX(vec3 refDir) {
+#ifdef CUBEMAPROT
+    return (refDir * cubeMapRotationMatrix) * vec3(-1, 1, 1);
+#else
+    return refDir;
+#endif
+}
```
willeastcott, line 19: Add a newline at the end here or concatenation might caused compilation errors.

src/scene/scene.js
```diff
@@ -65,6 +68,7 @@ import { StandardMaterial } from './materials/standard-material.js';
  * @property {pc.Texture} skyboxPrefiltered8 The prefiltered cubemap texture (size 8x8) used as the scene's skybox, if mip level 5. Defaults to null.
  * @property {pc.Texture} skyboxPrefiltered4 The prefiltered cubemap texture (size 4x4) used as the scene's skybox, if mip level 6. Defaults to null.
  * @property {number} skyboxIntensity Multiplier for skybox intensity. Defaults to 1.
+ * @property {pc.Quat} skyboxRotation The rotation of the skybox to be displayed. Defaults to pc.Quat.IDENTITY
```
willeastcott, line 71: Pedantic alert! 🚨  All other @property lines end with a full stop. 😄

willeastcott: Even better would be:

```
Defaults to {@link pc.Quat.IDENTITY}.
```

raytranuk: That is better - fixed

src/graphics/program-lib/chunks/cubeMapRotate.frag
raytranuk: done - thanks

src/graphics/program-lib/chunks/cubeMapProjectBox.frag
```diff
@@ -13,5 +15,6 @@ vec3 cubeMapProject(vec3 nrdir) {
 
     vec3 posonbox = vPositionW + nrdir * fa;
     vec3 envBoxPos = (envBoxMin + envBoxMax) * 0.5;
-    return posonbox - envBoxPos;
+
+    return (posonbox - envBoxPos) * vec3(-1, 1, 1);
```
mvaligursky, line 19: does this mean the dynamic cubemap wont be usable for box projection as that flips x?

raytranuk: Box projection uses `cubeMapRotatePosX()` so doesn't flip with the rotate - but needs to flip here because we used to flip after `cubeMapProject()` - the end result is that for box projection, with no rotation the cost is still the same as before, but with rotation the cost isn't absorbed by the rotation.

raytranuk: I've refactored the code - separating out rotation from -x flip

src/scene/scene.js
```diff
@@ -116,6 +120,12 @@ function Scene() {
     this._skyboxIntensity = 1;
     this._skyboxMip = 0;
 
+    this._skyboxRotation = new Quat();
+    this._skyboxRotationMatrix = new Mat3();
```
slimbuck, line 124: looks like this can be initialised to null?

src/scene/scene.js
```diff
@@ -116,6 +120,12 @@ function Scene() {
...
+    this._skyboxRotationMatrix.data[0] *= -1.0;
```
slimbuck, line 125: should this be here?

src/scene/scene.js
```diff
@@ -372,6 +399,17 @@ Scene.prototype._updateSkybox = function (device) {
             return;
         }
 
+        // update rotation matrix after used texture is determined, but before material is updated
+        if (!this._skyboxRotationMatrix4) this._skyboxRotationMatrix4 = new Mat4();
```
slimbuck, line 403: can this block of code be moved to line 435 (inside the test) since that's where the matrix is used?

src/graphics/program-lib/chunks/ambientPrefilteredCube.frag
```diff
@@ -4,7 +4,9 @@ uniform samplerCube texture_prefilteredCubeMap4;
 #endif
 
 void addAmbient() {
-    vec3 fixedReflDir = fixSeamsStatic(dNormalW, 1.0 - 1.0 / 4.0);
+    vec3 fixedReflDir = fixSeamsStatic(cubeMapRotate(dNormalW), 1.0 - 1.0 / 4.0);
+#ifndef DYNCUBEMAP
```
slimbuck, line 8: Wouldn't it be clearer to readers if this macro was named such that it describes better what it does? Perhaps REFLECT_CUBE_X or similar?

src/scene/scene.js
raytranuk: I was treating this _skyboxRotation more like a value than an object - which lines up with the code in: `Scene.prototype.applySettings()` - I can change it but is it better to do a new object allocation in `Scene.prototype.applySettings()`?

src/scene/scene.js
raytranuk: I missed that - thanks - I think I can also make the matrix lazily allocated

src/scene/scene.js
raytranuk: yes I think it can be moved - I will move it

src/graphics/program-lib/chunks/ambientPrefilteredCube.frag
raytranuk: I thought about this - I would prefer the macro name to give an idea what it is used for - not what it does. Its used to specify if the cubemap is dynamic and when the cubemap is not dynamic it skips an operation

src/scene/scene.js
raytranuk: done - thanks

src/scene/scene.js
raytranuk: moved - thanks

src/scene/scene.js
raytranuk: After more thought I do prefer it be treated like a value vs object in this case - I did remove the 3X3 matrix allocation though

src/graphics/program-lib/chunks/ambientPrefilteredCube.frag
raytranuk: hmm - how about RIGHT_HANDED_CUBEMAP - this is the real difference between dynamically rendered cubemaps and the pregenerated cubemaps - and what this macro and the flip operation deals with?

src/scene/scene.js
slimbuck: Oh I see there are two matrices. I would say you should handle Matrix and Matrix4 similarly - either initialise both on construction or both on demand. (Unless the two matrices are different in some way I haven't noticed).

src/graphics/program-lib/chunks/ambientPrefilteredCube.frag
slimbuck: yes!!!

mvaligursky: I like that!

slimbuck (approved):

willeastcott (approved):

Maksims:
PR name is misleading - this is not cubemap rotation, but skybox rotation.

Is it possible to rotate cubemap that is applied on material directly?
Box projection cubemaps been limited by not able to rotate them, leading to limited use cases. One of them is moving objects, e.g. train with box projected cubemap inside.

raytranuk:
> PR name is misleading - this is not cubemap rotation, but skybox rotation.
> 
> Is it possible to rotate cubemap that is applied on material directly?
> Box projection cubemaps been limited by not able to rotate them, leading to limited use cases. One of them is moving objects, e.g. train with box projected cubemap inside.

@Maksims - You are right, this PR now only exposes skybox rotation, even though the shader code does support any cubemap to be rotated. A subsequent PR will also expose non-skybox cube map rotation - and we are discussing the best place to store the rotation. I'll adjust the name of this PR.

src/graphics/program-lib/chunks/ambientPrefilteredCube.frag
raytranuk: renamed macros - thanks!

PR #2522, opened 2020-11-02
Prefiltered lighting updates

slimbuck:
This PR makes the following changes:
- update `reprojectTexture` to obey/implement `pc.Texture.fixCubemapSeams` and make sample count configurable
- standard material's high quality prefiltering lookup (used on most android devices) has been updated to perform 2 cubemap samples instead of 6 and calculates edge seams correctly:

filter_comp

I confirm I have signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #2528, opened 2020-11-04
Further prefilter fixes

slimbuck:
Follow on from #2522 :

Use a slower but better quality basis creation and normalise vector correctly.

### DIFF
--- a/src/graphics/program-lib/chunks/reproject.frag
+++ b/src/graphics/program-lib/chunks/reproject.frag
@@ -156,7 +156,7 @@ vec3 getDirectionCubemap() {
         vec = vec3(-st.x, -st.y, -1);
     }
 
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

src/graphics/program-lib/chunks/reproject.frag
```diff
@@ -167,6 +167,12 @@ mat3 matrixFromVector(vec3 n) { // frisvad
     return mat3(b1, b2, n);
 }
```
willeastcott, line 168: Is this now unused? Any reason not to delete it and make `matrixFromVectorSlow` into `matrixFromVector`?

slimbuck: I would like to keep it around for more accurate performance testing, but ultimately if we don't expose both as user options it can be removed.

Issue #1081, opened 2018-02-02
Rename shininess to gloss


willeastcott:
Shininess is a legacy of the past. It takes a value in the range 0 to 100 which is totally out of step with the other StandardMaterial scalar properties. Also, since a glossMap is essentially per-pixel glossiness, we should rename.

We should:

* Create a new property called glossiness (which is essentially shininess mapped to the 0 to 1 range).
* Alias shininess to glossiness (dividing the value by 100). The setter/getter should print a warning to the JS console that that property is deprecated and glossiness should be used instead.

Note that a property called shininess will remain in the material JSON format, at least for now.

slimbuck:
done in https://github.com/playcanvas/engine/pull/4949

willeastcott:
@slimbuck  Really? How come this is still here:

https://github.com/playcanvas/engine/blob/129062e8602e89b05d2d7311828fad619355b27b/src/scene/materials/standard-material.js#L242-L245

mvaligursky:
Yeah, we should deprecate it .. move this function to deprecated as well.
https://github.com/playcanvas/engine/blob/129062e8602e89b05d2d7311828fad619355b27b/src/scene/materials/standard-material.js#L1113-L1121

PR #2535, opened 2020-11-06
Feature/arealights

### LINKED ISSUE #774: Area Lights
Add Area Lights with different techniques, even if their functionality is partly limited (no shadows for example).

It would be great to have: `Sphere`, `Capsule`, `Plane` area light shapes.

marklundin:
### Implementation of LTC based area lights. 

The PR adds a new light shape property on all light types that when set to `pc.LIGHTSHAPE_RECT` or `pc.LIGHTSHAPE_DISK` or `pc.LIGHTSHAPE_SPHERE` enables area lighting using Linearly Transformed Cosines. It supports both Diffuse, Specular and Clearcoat using the physical/standard material.

image

Lights can be controlled using the entity transformations together with light attributes such as intensity and colour.

diffuse concrete surfaces
_diffuse materials_

non uniform glossy + diffuse
_Non uniform glossy + diffuse materials_

I expect there to be a few comments on this, but I thought I'd try this as a first attempt to start a discussion. 

Open questions;

1. **Luts** - The LUT's required by the implementation are very large and are computationally expensive to generate at runtime.
update: LUTs are currently optionally included as a javascript file - but this may change before we make this feature public.

2. **Floating point texture** - It currently uses a FP texture, which could be a half float depending on hardware
update: Half Floats and 8 bit per channel textures are now supported - allowing this feature to run on all platforms
 
3. **Banding** - There's an issue with banding which is as expected which is more noticeable on rough materials. It could be good to add some blue noise dithering behind a flag at the end of the material, but this is a separate issue.
update: This is yet to be addressed.

Next steps. I've left these open until the core implementation is done

1. Lightmapping
3. Other light shapes
4. Textured lights

Additional Notes:
1. Area light shadows are currently approximated with punctual light shadows - this should be improved in a subsequent PR. 
2. Directional area light specular is currently approximated with positional area light calculations - this should also be improved.

Also, I'm not great at the Pull Request workflow, so apologies if I mess this all up :D

Fixes #774.


marklundin:
Ah, also I have a demo scene here that should work with the custom engine

https://playcanvas.com/project/730762/overview/arealight

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "playcanvas",
-  "version": "1.35.0-dev",
+  "version": "1.35.0",
```
willeastcott, line 3: Can you revert your package-lock.json please?

src/graphics/ltc-lut.js
```diff
@@ -0,0 +1,155 @@
...
...131983, 0.159543, - 0.13253, 0.307151, 0.152541, 0.176432, - 0.143611, 0.31564, 0.174673, 0.201723, - 0.15548, 0.33538, 0.199842, 0.229721, - 0.167166, 0.355256, 0.227097, 0.250206, - 0.178238, 0.360047, 0.256014, 0.282118, - 0.189905, 0.378761, 0.28855, 0.312821, - 0.201033, 0.39181, 0.323348, 0.341482, - 0.211584, 0.397716, 0.360564, 0.377368, - 0.221314, 0.410141, 0.400004, 0.418229, - 0.230474, 0.423485, 0.442371, 0.444881, - 0.239443, 0.418874, 0.488796, 0.488899, - 0.245987, 0.427545, 0.535012, 0.520317, - 0.253948, 0.422147, 0.589678, 0.568566, - 0.256616, 0.42719, 0.637683, 0.599607, - 0.26376, 0.415114, 0.703363, 0.64222, - 0.268687, 0.408715, 0.771363, 0.685698, - 0.2694, 0.399722, 0.83574, 0.732327, - 0.266642, 0.388651, 0.897764, 0.769873, - 0.267712, 0.369198, 0.983312, 0.806733, - 0.263479, 0.346802, 1.06222, 0.843466, - 0.254575, 0.321368, 1.13477, 0.873008, - 0.242749, 0.29211, 1.20712, 0.908438, - 0.22725, 0.262143, 1.27465, 0.936321, - 0.207621, 0.228876, 1.33203, 0.950353, - 0.187932, 0.19484, 1.40439, 0.96442, - 0.165154, 0.163178, 1.4732, 0.979856, - 0.139302, 0.127531, 1.53574, 0.982561, - 0.11134, 0.0903457, 1.59982, 0.996389, - 0.0808124, 0.0489007, 1.6577];
```
willeastcott, line 90: Yeah, this file is 310KB uncompressed and 97KB gzipped - so we'll have to figure out a good solution to mitigate this.

willeastcott:
Would be great to add a screenshot to the PR description.

src/graphics/ltc-lut.js
marklundin: I'll have to look into the overhead of generating this at runtime. I'm guessing it's not insignificant, but at least it could be non blocking

marklundin: For reference the Lut generation code is here https://github.com/selfshadow/ltc_code/blob/texture/fit/fitLTC.cpp

I've not looked into the complexity, but it could be a good candidate for a worker or potentially, maybe a wasm lib?

willeastcott: Yeah, a worker sounds like a good option. But it'd be good to get a feel for how long the code takes to execute as compiled C++.

raytranuk:
> Ah, also I have a demo scene here that should work with the custom engine
> 
> https://playcanvas.com/project/730762/overview/arealight

@marklundin - I took a look at the demo - great to see it working!

One thing I noticed is that there is a crash when shadows are enabled - I'm that aware shadows are not part of this PR, but ignoring the setting and not crashing would be preferable.

marklundin:
> > Ah, also I have a demo scene here that should work with the custom engine
> > https://playcanvas.com/project/730762/overview/arealight
> 
> @marklundin - I took a look at the demo - great to see it working!
> 
> One thing I noticed is that there is a crash when shadows are enabled - I'm that aware shadows are not part of this PR, but ignoring the setting and not crashing would be preferable.

Ahh thanks for pointing this out. Yes I'm a little unfamiliar with all the different code paths, and I'm aware these all create different shaders, so I'll have to dig into this a little more

kungfooman:
Just did a quick investigation, generating those two LTC lookup tables:

```g++ fitLTC.cpp dds.cpp -o fitLTC.exe -I..\external\CImg -Dcimg_display=0```

Started `fitLTC.exe` and it took several minutes (AMD Ryzen 5 at 3.6 GHZ)

Actual amount of numbers:

```
> ltc.g_ltc_1.length
16384
> ltc.g_ltc_2.length
16384
```

I thought that maybe the printing/exporting the lookup tables itself may require most of the time (IO), so I commented all `cout` and export functions, but it still takes minutes.

So I guess it boils down to compressing 32768 floats as best as possible or finding a quicker generation algorithm.

**Comment by marklundin** [comment]
> Just did a quick investigation, generating those two LTC lookup tables:
> 
> `g++ fitLTC.cpp dds.cpp -o fitLTC.exe -I..\external\CImg -Dcimg_display=0`
> 
> Started `fitLTC.exe` and it took several minutes (AMD Ryzen 5 at 3.6 GHZ)
> 
> Actual amount of numbers:
> 
> ```
> > ltc.g_ltc_1.length
> 16384
> > ltc.g_ltc_2.length
> 16384
> ```
> 
> I thought that maybe the printing/exporting the lookup tables itself may require most of the time (IO), so I commented all `cout` and export functions, but it still takes minutes.
> 
> So I guess it boils down to compressing 32768 floats as best as possible or finding a quicker generation algorithm.

That's good to know. From a cursory look it didn't seem that expensive but I presumed there was a logical reason why it was precomputed

**Comment by raytranuk** [comment]
> > > Ah, also I have a demo scene here that should work with the custom engine
> > > https://playcanvas.com/project/730762/overview/arealight
> > 
> > 
> > @marklundin - I took a look at the demo - great to see it working!
> > One thing I noticed is that there is a crash when shadows are enabled - I'm that aware shadows are not part of this PR, but ignoring the setting and not crashing would be preferable.
> 
> Ahh thanks for pointing this out. Yes I'm a little unfamiliar with all the different code paths, and I'm aware these all create different shaders, so I'll have to dig into this a little more

Adding `&& type !== LIGHTTYPE_AREA` to this line:

https://github.com/playcanvas/engine/blob/dd3a10654bb659753e17d2b4b86bba92f8d577a0/src/scene/forward-renderer.js#L1240

 will probably stop the crash I was getting.

**Comment by marklundin** [comment]
Updated the [demo](https://launch.playcanvas.com/1019217?debug=true&use_local_engine=https://pc-arealights.netlify.app/build/playcanvas.js) with something that shows it off a little more.  _(Big 4K textures, so long load time. Probably not great on mobile)_

**Comment by mvaligursky** [comment]
we could store those LUTS in binary files (as float32), and load them like this perhaps
https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Sending_and_Receiving_Binary_Data

but instead of
`var byteArray = new Uint8Array(arrayBuffer);`
use
`var float32Array = new Float32Array(arrayBuffer); `

**Comment by marklundin** [comment]
This is what I'm leaning towards @mvaligursky . As a binary it could also be compressed, asynchronously decompressed and stored locally via a service worker, however all of that brings it's own complexity such as including some decompression lib into the engine, which may be problematic on mobile.

Also most webservers gzip files right? Would we be fighting against that by making this a binary?

**Comment by mvaligursky** [comment]
I think 128kb is small. We would not want every user to have to download this, and so making this part of the core engine is not ideal. But if in a project a user decides to use area lights (which I've tested works well on mobile as well! I didn't expect that), downloading this is small cost - comparing to most other assets in the project, like textures or meshes. We can likely have this being served as zipped up by web server, but as you said, it might not compress much.

I think there are two approaches that I'd be happy to go with:
1. either store it as .separate js file, that user can load into project / drag & drop into editor / or we build is similar to extra to a new js bundle. This will be larger text file (compared to binary file), but compressed it will be similar to compressed binary float file.
2. store it as a binary blob and load it as float32. I think the actual size that needs to be downloaded would be similar size to gzipped text file.

So the only different is based on what is easier for users and us to include, and I suspect perhaps option 1. could be easier - the users could just include (drag & drop) js file which we prepare that registers some data with the application / window. Later we could automate this, similarly to how we automate physics engine integration. In editor, when you select area light, you'd have Import button that does what is needed to include that js file. 

Thoughts on this @willeastcott ?

**Comment by willeastcott** [comment]
Well now - this is a tough one. I kind of like the idea of putting the data in a JS file because then it will be concatenated with the other JS at build time instead of incurring another HTTP requests. But I also like the purity of a BIN file containing just the float data. I'd probably go with the latter. In an engine-only case, the app could load a base64 encoded string burned into the code, or load the BIN file at startup and we'd need some kind of registration function/property (`app.scene.areaLightLut = data;`). In the Editor, if you create a light and set type to 'Area', it will show a warning if the LUT data is not set in the Scene Settings (bin asset picker). Under that, there'd be an import area light LUT button. Something like that. :smile:

File: src/scene/forward-renderer.js [hunk 3]
```diff
@@ -1179,7 +1237,7 @@ Object.assign(ForwardRenderer.prototype, {
             light = lights[i];
             type = light._type;
             if (light.castShadows && light.enabled && light.shadowUpdateMode !== SHADOWUPDATE_NONE) {
-                if (type !== LIGHTTYPE_DIRECTIONAL) {
+                if (type !== LIGHTTYPE_DIRECTIONAL && type !== LIGHTTYPE_AREA) {
```
mvaligursky, line 1240: It will probably be good to implement frustum culling for area lights as well, otherwise not many local area lights will be able to be placed in the scene as they will all need to be "used".

Do these lights have range, similar to point / spot lights? If yes, perhaps area light culling could be based on spot light with range and 180deg cone?

marklundin: There is no implicit attenuation or falloff calculated, but it does seem like a good idea to cap it. A couple of things to consider; area lights can be different shapes (disk, line etc), so any culling would need to accommodate this. I guess this can just be added on a case by case basis? Also, they also commonly allow double sided illumination which is inexpensive to calculate, so in the quad instance the hemisphere would become a sphere

mvaligursky: That sounds good. The culling function calls light.getBoundingSphere where it handles spot and point light differently, we'd just need to add some handling of area light per shape type. For now either half sphere or sphere should be returned as you described.

src/graphics/program-lib/programs/standard.js
```diff
@@ -1445,12 +1462,16 @@ var standard = {
                 code += "   addDirLightMap();\n";
             }
 
+
+            if (hasAreaLights){
+                code += "   float roughness = max((1.0 - dGlossiness) * (1.0 - dGlossiness), 0.001);\n";
```
marklundin, line 1467: @mvaligursky Is this calculation of roughness correct? I took this out of one of the chunks but would be good to verify

raytranuk: This is the glossiness to roughness conversion I use for the GGX specular implementation (which is a model that is not widely used yet) - its based off what I heard was used a few years back in Frostbite - I was planning to revisit and recalibrate this if/when we switch using roughness.

mvaligursky:
We had a discussion about how to integrate LUTs for the area lights with the engine, and concluded that perhaps for this PR we'll keep it simple. We can have API on engine allowing these to be assigned, and later we can change delivery mechanism, and as long as API does not change, all should work.
So please add property to src/scene/scene.js, perhaps called areaLightLuts, which would take object, with multiple Luts in it, allowing in the future to provide luts for different light shapes. So maybe
```
{
   quadLut0 = [];
   quadLut1 = [];
   circleLut = [];
   ...
}
```
and for now just store luts in this format in separate .js file, add it to the project and call API to assign it.

Separately, I thought you asking whether to use areaWidth/areaHight property, or use unit size quad and apply scale to it. Considering other shapes in the future (circle, tube, sphere?), perhaps it would be simpler to use the unit size light and scale instead of specifying different properties per light shape? It would also allow light to be part of mesh (say a tube light part of room), and it would handle scaling of this as a hierarchy. Just not sure if non-uniform scale for all shapes would work? Thoughts on this?

marklundin:
The LUTs themselves are consistent across different light shapes, so just the 2 arrays should suffice for any new shapes.

For the implementation, is it ok have the external JS file directly add the LUTs on the scene? As in;

`scene.areaLightLuts = {lut1, lut2}` 

And then later on in the render raise an error if the area light is used but areaLightsLuts is undefined?

For the scaling, I think this makes sense and I'm already using the translation and rotation, so this should be more intuitive

mvaligursky:
> scene.areaLightLuts = {lut1, lut2}

That sounds good to me.

marklundin:
Hey @willeastcott @mvaligursky has there been an update on the playcanvas launcher to with regard to using `use_local_engine`? Just trying to test and the local dev server works fine, but seems we can't use it against a publicly hosted file.

yaustar:
@marklundin Unfortunately, yes. We had to restrict use_local_engine to only use localhost or code.playcanvas.com due to security concerns.

For a publicly hosted project, the only workaround that we can suggest is to add the engine file into the project and set it to load after the engine.

eg https://playcanvas.com/project/735658/overview/arealight-fork
image

Not ideal but it's the best we can do right now.

marklundin:
No worries! I thought something else was going on, but that makes sense 👍

src/graphics/ltc-lut.js
marklundin: @willeastcott as per @mvaligursky's suggestion, I've removed the LUT's from the engine itself. It is now a [separate dependency](https://playcanvas.com/editor/code/730762?tabs=37932022) that can be copied into asset registry. I imagine there will be better ways of integrating this, but for the moment, I think this gives the most minimally invasive way of doing it, yet flexible enough to switch it up down the line.

I've also removed the explicit width and height from the area lights and rely just on the entity transformation, which makes the code a lot simpler.

There is also the case of adding a flag for allowing the quad to project both sides, but I've left this out for now and can submit this as part of separate PR.

kungfooman:
I wondered about the nature of the LUT's and just did a simple plotting:

LTC1 (First 3 numbers: position, last number: color as hue):

ltc1

LTC2 (I realized that the entire Z column is 0, so I completely dropped it and colored vertices by their position):

ltc2

I see quite some "similarity", so maybe it could just be quickly approximated, without even knowing the original algorihm.

If the LUT ever gets committed, at least the size could be 112 KB: `((32768 * 7/8) * 4) / 1024`

marklundin:
Amazing work! The first LUT is the BRDF, which in this instance is a GGX
one. The second I believe is an approximation of the fresnel, which could
be swapped for a schlicks approximation. However I've never seen this done
in practice, so I left in in there for now

marklundin:
@mvaligursky Is there anything else you need on this PR?

mvaligursky:
@marklundin - I think the light culling was the only missing thing before people will do more detailed review / approval.
You also mentioned double sided lighting .. is this something you'd want to add or leave it for follow up PRs?

mvaligursky:
and also half texture precision maybe - or at least we'll need to make sure the light is simply off when float textures are not supported for now

mvaligursky:
@marklundin - are you able to give me write access to your branch? I've made initial simple engine example for area lights which we use for automated testing when we release engine, and would like to probably help out on some parts.

marklundin:
Ahh that would be super helpful. I'll sort that out for you

raytranuk:
I've been thinking about how things could move forward if we add a new Area light type to the engine.

What we are currently seeing is a few `if (... || light.type === LIGHTTYPE_AREA )` and `if (... && light.type !== LIGHTTYPE_AREA)` which is not so bad to start with - but next steps would include additional light source shapes, whether the light source is omni-directional or hemispherical, and whether the light source is positional or non-positional - e.g the Sun light is modeled as non-positional global - these properties of the area light will affect what type of culling occurs and what kind of shadow is projected. 

Omni-directional area light sources are very similar to` LIGHTTYPE_POINT` - and will share some code paths - so we could start seeing:`if (... || (light.type === LIGHTTYPE_AREA && light.isOmniDirectional === false))` or `if (... && (light.type !== LIGHTTYPE_AREA || light.isOmniDirectional === true))` to include or exclude area lights in omni-directional or non-omni-directional code paths. 

Global non-positional light sources are very similar to `LIGHTTYPE_DIRECTIONAL` - so we may see code like `if (... || (light.type === LIGHTTYPE_AREA && light.areaLightSubType === NON_POSITIONAL))` to choose orthographic projection for shadows or `if (... && (light.type !== LIGHTTYPE_AREA || light.areaLightSubType !== UNI_DIRECTIONAL ))` to skip parts of culling that deals with cones/hemispheres.

Another thing to consider is that it is often essential in real-time rendering to optimize down expensive higher fidelity techniques to cheaper approximations where possible - and we should plan for this in PlayCanvas. In many cases an acceptable approximation for area lights is punctual lights - the bigger the area light, the greater the distance from the light before the approximation is acceptable - a good example is that the Sun is a big spherical light source: until relatively recently it has been acceptable to approximate it as a punctual directional light in most games - but now, some AAA games are using an area light (circle shape rather sphere shape) for the Sun. What we may start seeing is code like: `if (... || (light.type === LIGHTTYPE_AREA && light.areaLightSubType === NON_POSITIONAL && light.areaLightLOD(renderPass) === 1.0))` to select whether a non-positional global light should be treated as a punctual light in render passes that don't need the extra cost - which is probably all passes but the main pass and light map baking - and in the main pass, probably perform a LOD calculation based on: user defined settings, light source size and distance from the light.

The main problem I'm seeing with adding Area lights as a sibling type to the existing punctual types is that, in the long run, it will effectively overlap the punctual light types - but as a sibling type - effectively it will be the same as the current `pc.Light` light class but with an additional property: light source shape.

mvaligursky:
> Ahh that would be super helpful. I'll sort that out for you

I still have issue pushing to this repo and emailed you directly.

marklundin:
I will take a look on the branch permissions

mvaligursky:
@marklundin - I created engine example project (examples/graphics/area-lights.html). If you have local serving for the engine, just go to examples project, and pick it from the menu. Few notes:
- it seems area light is in the opposite direction compared to spot light, it'd be good to fix. In the example change "area" to "spot" to see the issue.
- compared to spot again, it seems I need to set intensity of area light to maybe 10x of the intensity of spot light to get similar visual effect .. any idea why that is?
- at the moment the range does not work the area light .. ideally it would work similarly to when you select spot light and play with the range value.

marklundin:
Thanks for this @mvaligursky 

> it seems area light is in the opposite direction compared to spot light, it'd be good to fix. In the example change "area" to "spot" to see the issue.

Sounds good. I think the default plane must be flipped.

> compared to spot again, it seems I need to set intensity of area light to maybe 10x of the intensity of spot light to get similar visual effect .. any idea why that is?

Does changing the size of the area light have any effect? It may be that the intensity needs to account for the size of the area light, as in should this be lumens or lux? Not sure the answer to this, but maybe just consistent with spot light.

> at the moment the range does not work the area light .. ideally it would work similarly to when you select spot light and play with the range value.

Yes, there's no attenuation in the shader at the moment, however thinking about this, any fall off would also need to account for the width/height of the area light. For example, if the falloff was 10 units, but the area light was uniformly scaled to 20, then presumably it would need to account for that. There's not really a correct answer here, but would be good to know your thoughts on this.

mvaligursky:
I just checked the spot light - its direction is aligned with the negative Y axis, and you likely use positive Y axis for the area light? Easy to change I hope.

I tried few relative small sizes (3x3 up to 5x5) and they're considerably darker than spot light. I agree this probably needs to be set up just so it feels right and consistent? Any thoughts on this @raytranuk ?

Regarding the range / falloff distance scaling, that's tricky one. For other lights range is not scaled and is specified in world units, so perhaps to be consistent here this should not scale as well, even though if we scale the quad size, it feels like it should scale with it. In order to be consistent, I would probably not scale this.

marklundin:
Interesting comments @raytranuk, thanks for the insight. Caveat to the following; this is all very much on the fringe of my understanding, so feel free to point out incorrect assumptions.

> Global non-positional light sources are very similar to LIGHTTYPE_DIRECTIONAL - so we may see code like if (... || (light.type === LIGHTTYPE_AREA && light.areaLightSubType === NON_POSITIONAL)) to choose orthographic projection for shadows or if (... && (light.type !== LIGHTTYPE_AREA || light.areaLightSubType !== UNI_DIRECTIONAL )) to skip parts of culling that deals with cones/hemispheres.
 
I had not considered non-positional lights before, but your example of the sun modeled as a disk makes sense. From looking into it further, it seems like many engines separate punctual lights from area lights

> Another thing to consider is that it is often essential in real-time rendering to optimize down expensive higher fidelity techniques to cheaper approximations where possible

I think it's important to understand the relative cost of the proposed area lights solution. From what I understand the analytical area lights in engines such as Frostbite are quite complex and expensive. I am only going by their [2015 paper](https://seblagarde.files.wordpress.com/2015/07/course_notes_moving_frostbite_to_pbr_v32.pdf) which may now be redundant, however, this importantly separates out the two implementations for performance reasons. It would therefore be wise to only weigh the additional overhead in complexity if it was deemed necessary to implement. 

> The main problem I'm seeing with adding Area lights as a sibling type to the existing punctual types is that, in the long run, it will effectively overlap the punctual light types - but as a sibling type - effectively it will be the same as the current pc.Light light class but with an additional property: light source shape.

Do you see a proposed way forward with this?

raytranuk:
> I had not considered non-positional lights before, but your example of the sun modeled as a disk makes sense. From looking into it further, it seems like many engines separate punctual lights from area lights
>
It does seem other engines do indeed choose to have a separated implementation - however, I'd like to think PlayCanvas tries harder than most engines to not break public APIs that were introduced, in newer engine versions - but also may require a slightly different approach to adding new APIs to avoid difficulty maintaining or having to change them later.

> > Another thing to consider is that it is often essential in real-time rendering to optimize down expensive higher fidelity techniques to cheaper approximations where possible
> 
> I think it's important to understand the relative cost of the proposed area lights solution. From what I understand the analytical area lights in engines such as Frostbite are quite complex and expensive. I am only going by their [2015 paper](https://seblagarde.files.wordpress.com/2015/07/course_notes_moving_frostbite_to_pbr_v32.pdf) which may now be redundant, however, this importantly separates out the two implementations for performance reasons. It would therefore be wise to only weigh the additional overhead in complexity if it was deemed necessary to implement.
> 
There may be several options for optimization / approximations with disc shaped area lights - one example is a single value for penumbra size when dealing with shadows (which is one of the most noticeable when rendering Sun light - since all the softness we see in Sun light shadow from opaque objects is actually due to the Sun having a non-zero radius). Even reasonably close up (on Earth) you could use super simple wrap lighting for diffuse - and surfaces with high roughness or not close up to the camera could also only apply punctual specular calculations - this leaves the additional cost and fidelity in the areas that makes the most impact to the player/user. Quite an interesting thing to think about.

> > The main problem I'm seeing with adding Area lights as a sibling type to the existing punctual types is that, in the long run, it will effectively overlap the punctual light types - but as a sibling type - effectively it will be the same as the current pc.Light light class but with an additional property: light source shape.
> 
> Do you see a proposed way forward with this?

It could mean a refactor to go from Area Light as a sibling type to Area Light as a new light source shape property - to start with we will only have 2 shapes: punctual and rectangle and only positional lights would be able to use both shapes. An alternative to adding extra `if` tests in `pc.Light` class is to extend it into a derived class - perhaps: `pc.AreaLight` which would include a light source shape property, and as more shapes Area Light types are support will override more of the pc.Light base class.

mvaligursky:
My preference is to keep area light as a separate type as is at the moment. There are then 2 ways I see to handle LOD:

1. User adds custom lighting rig with different light setups and either switches between whole setups at load time, or does in a per light basic as needed. This is easy for directional light, as there's typically only very small number of these. Advantage for area lights is that user keeps decision over how to approximate their area lights by punctual lights. They might use single point / spot light, or few of them to larger area lights.

2. In the engine we have all lights in an array, and for rendering split it into lights by type. During this step we can evaluate LOD (based on distance to camera and similar), and simply add the light to an array of different light type (making sure the properties for that type are set up). This is a way we have some automated LOD system for distant area lights.

raytranuk:
> My preference is to keep area light as a separate type as is at the moment. There are then 2 ways I see to handle LOD:

How would you handle adding global area light support - equivalent to a directional light with a disc (or rectangle etc) shape? An additional sibling-type or an additional property in the Area light?

> 1. User adds custom lighting rig with different light setups and either switches between whole setups at load time, or does in a per light basic as needed. This is easy for directional light, as there's typically only very small number of these. Advantage for area lights is that user keeps decision over how to approximate their area lights by punctual lights. They might use single point / spot light, or few of them to larger area lights.

IMO Not having to author multiple sets of lights is clearly better for the user - in teams were you only have small teams lighting artists (or even no dedicated ones) you are multiplying up workload - and mirroring tweaks or adjustments to multiple sets of lights is not fun. Even if some projects have a specific situation where they can get better performance / quality by manually setting up multiple sets of lights, we could still support that with a well designed LOD settings - could be as simple as having a max and min distance at which a light becomes active for the user to set.

Also how would you handle a scenario, where you want some objects to render cheaper using a Punctual directional light and others, higher quality, with an Area directional light? Would the user have to add 2 lights to the scene with almost the same same parameters, but different light type and keep them manually synced up? How about shadows - would we render 2 separate shadow maps or would a system be implemented for them to share one?

> 2. In the engine we have all lights in an array, and for rendering split it into lights by type. During this step we can evaluate LOD (based on distance to camera and similar), and simply add the light to an array of different light type (making sure the properties for that type are set up). This is a way we have some automated LOD system for distant area lights.

If implemented like this, it is very likely you would see bad light popping or you would have to switch really far out when there is literally no difference in lighting result - being able transition earlier - but not make it a jarring pop so that the user/player doesn't notice - is something that is often used in high quality games to maintain perception of high quality but still run at good performance and scalable to platforms with differing capabilities.

mvaligursky:
> How would you handle adding global area light support - equivalent to a directional light with a disc (or rectangle etc) shape? An additional sibling-type or an additional property in the Area light?

I think perhaps having it as different shape for the same area light type. "Quad" "Sphere" "Tube" "Sun"

> IMO Not having to author multiple sets of lights is clearly better for the user

I'm not sure why it is clearly better for the user. From my experience, some users like lot of control, and some not - and we can handle both cases with manual or automatic system, the internal design of light type does not make much of a difference.

> Also how would you handle a scenario, where you want some objects to render cheaper using a Punctual directional light and others, higher quality, with an Area directional light? Would the user have to add 2 lights to the scene with almost the same same parameters, but different light type and keep them manually synced up? How about shadows - would we render 2 separate shadow maps or would a system be implemented for them to share one?

We can handle LOD as few additional properties for the area light type.

> If implemented like this, it is very likely you would see bad light popping

You can implement area light size scaling in both systems easily. The way classes are set up does not limit the functionality, this is only about how it's exposed to the user.

mvaligursky:
> The main problem I'm seeing with adding Area lights as a sibling type to the existing punctual types is that, in the long run, it will effectively overlap the punctual light types - but as a sibling type - effectively it will be the same as the current `pc.Light` light class but with an additional property: light source shape.

Just one detail to this which probably causes some confusion - pc.Light class is a single class that stores properties for all light types. We don't have light class per type and so there won't be a duplication between them. We can change light type by just changing the type property. It's easy to add LOD system to this - in simplest case, you only do something like this:

`entity.light.type = cameraDistance < treshold: LIGHTTYPE_AREA : LIGHTTYPE_SPOT;`

slimbuck:
For the benefit of us onlookers, could someone perhaps illustrate practically the difference between the (two?) options under discussion here? If I understand it correctly there are two alternative data layouts when adding area lights to the engine. So practically what do the two options look like? Thanks!

raytranuk:
> For the benefit of us onlookers, could someone perhaps illustrate practically the difference between the (two?) options under discussion here? If I understand it correctly there are two alternative data layouts when adding area lights to the engine. So practically what do the two options look like? Thanks!

The difference is small but important - the current proposed implementation adds an additional type `pc.LIGHTTYPE_AREA` - what I think would be better is adding a property `light.shape` because, in the long run all our current punctual light types will support additional light source shapes - so adding a new type will be extraneous - and if projects start using it, we would then have to support that extraneous type going forward or break some projects.

I also think `light.shape` will create less complex code and more maintainable code - if we look at 3 of the next steps for area lights: adding support for bi-directional or uni-directional area lights, support for non-positional global lights like the Sun and adding a disc shape - the current proposal as I understand it using the `pc.LIGHTTYPE_AREA` type: 
1) Add a `light.areaLightShape` property that is only exposed if `pc.LIGHTTYPE_AREA` is selected - the shape property will need to have values of something like: `pc.AREA_LIGHT_RECT`, `pc.AREA_LIGHT_RECT_BI_DIRECTIONAL`, `pc.AREA_LIGHT_DISC`, `pc.AREA_LIGHT_DISC_BI_DIRECTIONAL`, `pc.AREA_LIGHT_SUN`, and `pc.AREA_LIGHT_NON_POSITIONAL_RECT`.
2) Add checks using `pc.LIGHTTYPE_AREA` and `light.areaLightShape` to determine if non-positional or positional code paths should be taken and if bi-directional or uni-directional code paths should be taken, in systems like culling and shadows.
3) Add checks for `light.areaLightShape` and code paths to support the new `pc.AREA_LIGHT_RECT_BI_DIRECTIONAL`, `pc.AREA_LIGHT_DISC`, `pc.AREA_LIGHT_DISC_BI_DIRECTIONAL`, `pc.AREA_LIGHT_SUN`, and `pc.AREA_LIGHT_NON_POSITIONAL_RECT` shapes.

With a `light.shape` property instead of a `pc.LIGHTTYPE_AREA` type, the steps to implement a rectangle positional area light is:
1) Add a `light.shape` property that can be a value of `pc.LIGHTSHAPE_RECT` or `pc.LIGHTSHAPE_NONE`.
2) Add checks for `light.shape` and code paths to support the `pc.LIGHTSHAPE_RECT` light source shape in positional (point and spot) light code paths.

Then, to add support for bi-directional or uni-directional area lights, non-positional global area lights and a disc shape:  
1) Add a `pc.LIGHTSHAPE_DISC` value.
2) Add checks for `light.shape` in the global non-positional (directional) light code path that is affected by light source shape. NB: bi or uni directionality can be inferred from `light.type`
3) Add checks and code paths to support the `pc.LIGHTSHAPE_DISC` light source shape.

slimbuck:
Right, so if the engine will ultimately support various area light shapes `none, rect, disk` as a logical extension of the existing punctual light types `point, spot, directional` (meaning the two settings are orthogonal), then it probably makes sense separate them?

And tbh it makes sense to me that our existing lights implicitly have shape 'punctual' which we are now extending to areas.

This should work well:
```javascript
// soft directional light
light.type = LIGHTTYPE_DIRECTIONAL;
light.shape = LIGHTSHAPE_RECT;

// soft point light
light.type = LIGHTTYPE_POINT;
light.shape = LIGHTTYPE_RECT;

// hard directional light
light.type = LIGHTTYPE_DIRETIONAL;
light.shape = LIGHTTYPE_PUNCTUAL;
```

What do you think @mvaligursky, any objections going this route?

Maksims:
If it will be a `shape` property, then how it will affect shadows?

slimbuck:
Problem is there won't just be one RECT area light, there will be three:
- a single-sided version with position (as this PR adds)
- a double-sided version with position
- a directional-only version without position

So the list would become:
Directional
Spot
Point
RectSingleSided
RectTwoSided
RectDirectional
...etc

Not really sure what the shadow support will be for area lights (it may vary, I guess), but that's a setting on the light anyway? Not sure how that impact this discussion?

mvaligursky:
my preference is 4 light types:
- Point
- Spot
- Directional
- Area
     area light will additionally expose Shape: Rect, RectTwoSided, Disk (Sun), Sphere ...
- But I would be happy to add more types, as @Maksims suggested, as that would make it even more straightforward for following two points as well. So maybe: Rect, Disk, Sphere ...

advantages:
- cleaner interface for the user. It's a stretch of imagination for the user who wants to create large area light to pick spot light and pick rect shape, or for double sided area light to start wit a point light.
- cleaning engine user. The engine splits array of lights into separate arrays by the type of light, and operates on these.

slimbuck:
Yeah you're right @mvaligursky and @Maksims: unfortunately light type `POINT` (and the others) specifically refer to punctual lights and users won't expect area lighting to fall under this.

Still, @raytranuk has a point that if we ultimately plan to support the full suit of area light types then we must plan accordingly and probably don't want to permutate the combinations i.e. we don't want: RECT, RECT_TWO_SIDED, RECT_DIRECTIONAL, DISK, DISK_TWO_SIDED, DISK_DIRECTIONAL, CAPSULE, CAPSULE_TWO_SIDED, CAPSULE_DIRECTIONAL, etc...

It would be better to split this into two attributes, for example:
**areaShape**: RECT, DISK, CAPSULE
**areaType**: SINGLE_SIDED, TWO_SIDED, DIRECTIONAL

I have some questions on this though:
- do we really plan to support all combinations of **areaShape** and **areaType**, is it even necessary? For example, will CAPSULE work as SINGLE_SIDED? cc @raytranuk 
- how would the engine handle an as-yet unsupported combination of **areaShape** and **areaType**? the editor UI can enforce valid combinations, but what about engine-only users?

marklundin:
From a shader persepective, other area light shapes and double sided are
reasonably trivial to implement given the existing chunks. Determining
culling etc is more complex, but it would be great to implement.

However I would also really like to add textured area lights. This would be
great with live video content, creating a screen for example. However this
only really makes sense for quads and maybe disks. I suppose the existing
cookie interface could work.

Maksims:
Not sure combination of area shape and type has valid combinations. I seems to fail to imagine what capsule with different areaTypes are. Capsule area lights - is same shape as capsule primitive, and there are no sides.
While rect can be single sided or two sided. Directional - can you specify what it is?
Sphere - same, no areaType variations here, just a sphere, which is capsule with no elongation.

I imagine in the future, Rect type of area light will support a texture. Which already is on light (cookie). And allow to create any shaped rects, so no special Disk type is required.

raytranuk:
>do we really plan to support all combinations of areaShape and areaType, is it even necessary? For example, will CAPSULE >work as SINGLE_SIDED? cc @raytranuk

All light sources are omnidirectional - the current spot lights are just a useful approximation for a omnidirectional light source with built-in angular occlusion - which is used to reduce shadow map rendering cost (1 map instead of 6) and cull more aggressively (cone shape) - the current directional light type is a useful approximation for very distant omnidirectional light source - which also affects shadow map rendering and disables culling. `SINGLE_SIDED` is just a light source with built-in angular occlusion on one side - and so can use the same code path as the current spot light type for culling - I see no reason to restrict the user by disallowing a light source to have a shape, like `CAPSULE`, and built-in occlusion if they want it to - in fact, I've seen real tube lights that have one side painted with reflective paint (I'll see if I can find a picture). So there really isn't a need for a `areaType` property at all - we can just use the current 3 light types.

>I would argue it is not intuitive to set directional + rect, or spot + rect, or point + rect?
>This sounds confusing from user point of view.

Shape is a light source property just like color or intensity and is orthogonal (as @slimbuck mentioned) to type - so if setting a directional light with yellow color and intensity of 2.0 is fine for a user - why is specifying a rectangle or disc shape any different? If we encode other orthogonal properties into light type then we would have: `pc.LIGHTTYPE_DIRECTIONAL_YELLOW_TWO_INTENSITY` or `pc.LIGHTTYPE_SPOT_WHITE_ONE_INTENSITY` - even if a ridiculously big list of these would allow the user to select type, color and intensity in one go, its still not a good approach.

> It's a stretch of imagination for the user who wants to create large area light to pick spot light and pick rect shape, or for double sided area light to start with a point light.

You think users won't be able to understand setting a light source shape on a light source? Why would that be so hard to understand? All lights sources actually do have a shape just like they have a color or intensity.

> cleaner engine use. The engine splits array of lights into separate arrays by the type of light, and operates on these.

These 3 arrays of lights already have orthogonal properties like shadow casting and filtering - which are only checked for in the places that are affected by these properties and can affect code path and shader uniforms - light source shape property will the same. Currently the light type is used to define culling (point: sphere culling, spot: cone+sphere culling, directional: no culling) and shadow projection (directional: orthographic, point+spot: perspective) and shader uniforms to be set (spot has cone angles directional doesn't have position) - is it really cleaner to add an extra array of area lights to run through and then check again, based on a area light sub-type, what culling each light should have or what shadow projection or extra shader uniforms should be set? I think it's the opposite to clean code.

>If it will be a shape property, then how it will affect shadows?

The initial area light implementation, using a light source shape property, will have no affect on shadows - however, IMO one of the things that will make area lights really useful is to implement good area light shadow projection for runtime generated light maps - this will bring us on a par (without bounced light) with other engines like UE4: (pictures from: https://docs.unrealengine.com/en-US/Engine/Rendering/LightingAndShadows/LightTypes/RectLights/index.html)
Static light maps generates these shadows: 
image
But for dynamic shadows you get:
image
 
You can see in the picture above that our current dynamic shadows would not be a great approximation for area light shadows, so we probably should look at extending the variance shadow mapping we have to use summed area variance shadow maps: https://developer.nvidia.com/gpugems/gpugems3/part-ii-light-and-shadows/chapter-8-summed-area-variance-shadow-maps - which could result in dynamic shadows like this:
image - and probably fall back to heavily blurred variance map/sharper shadows for weaker platforms or where the user decides shadow quality vs performance.

slimbuck:
If a POINT light has an area light source, is it not by definition no longer a POINT light? :)

slimbuck:
(this doesn't bug me nearly as much for spot or directional lights tbh, but making POINT an area light seems a little unintuitive)

raytranuk:
> If a POINT light has an area light source, is it not by definition no longer a POINT light? :)

Good point (haha) - in the editor we could label `POINT` lights `OMNIDIRECTIONAL` or `OMNI` or `POINT/OMNI`- then later rename it in engine and keep legacy support for POINT - this is similar to what was done and what is next for the shininess material property (labelled Glossiness in editor): https://github.com/playcanvas/engine/issues/1081

Maksims:
I would approach this first from user point of view. And the API design should be first priority for usability, and it will influence the engine design. Not vice-versa.

So for user, it is very simple idea of going through single list of types: directional (sun), spot (ceiling light), point (actual point light), capsule (light tube, classic office light), rect (TV screen).
It is simple to remember, easy to understand. No combinations. And UI will be very clean, by switching light type, it will show/hide certain settings just like it does now.

If there will be multiple combinations of different properties. It is hard to remember and is less direct to get desired result. Is it directional+rect or spot+rect for TV screen? Is it spot+sphere or point+sphere for light bull? And for capsule, it completely falls out of current three.

So to keep it simple for user, also keep it open for future types and easy additions, it would be beneficial to use light type property to define light type. And not multiple combination of properties that complicate API.

What engine does inside - is different story, and it can be anything regardless of an API.

raytranuk:
> And for capsule, it completely falls out of current three.

No - a capsule light would be a 3D shape - like a sphere is a 3D shape - 3D shapes would generally only be used as omni-directional lights, but the user should be able to specify angular occlusion (as I mentioned above) and there is a use case for using non-positional lighting with a 3D shape too.
 
> What engine does inside - is different story, and it can be anything regardless of an API.

The discussion in this PR is about adding a rectangle light source shape implementation to the engine - and looking further ahead, adding more light source shapes, including for (non-positional global) directional lights, and trying to avoid subsequently having to either break a badly thought out public API, or take on the maintenance burden of supporting it - adding large lists of permutations or the mess of redundant extra sub-type properties and related logic.

>It is simple to remember, easy to understand. No combinations. And UI will be very clean, by switching light type, it will show/hide certain settings just like it does now.

Whether we should, or want to, provide a simple list of pre-configured lights in a drop down list for new/basic users, is a discussion for an editor PR. I think most users will be able to work out, from a list of 3 types and a few shapes, the right light properties they should use - but maybe we can provide an easier pre-configured list - just like we have a few primitive shapes for model (that are just treated as meshes in the engine) - so less advanced/new users who want to make a quick demo can just choose: I want a TV screen light or I want a Sun light. We could add an editor issue to track this feature request.

BTW You mention the directional light type is for Sun light - but already there are high quality games released with a directional light type with a disc light source shape rather than a punctual light source shape for Sun light - so in a pre-configured list, would you have one directional (sun) light configuration, or directional_punctual (sun) and directional_disc (sun) configurations?

Maksims:
Directional light source with disc - would be a specific property of directional light.
Please, list all possible combinations, in order to see how many there will be "invalid" combinations. Best thing is to write it out as possible API, and it will make it clear.

raytranuk:
Since light shape is a light source property, we don't have to enforce restrictions to the user - but some combinations will be less useful or of no use at all - just like if you set the light source intensity property to 0.0 or the light source color property to (0.0,0.0,0.0) - and as with any big feature still under development, not every combination will be supported immediately or ever need to be supported - and this is common in engines that are still under active development, for example: dynamic shadows do not yet work with area lights in Unity and anisotropic lighting response doesn't yet work with area lights in UE4. What these engines do is add some validation, and documentation, to ensure the user is informed as to what can be used with what initially - which I think we will have to do also.

Maksims:
Indeed. Supporting everything from day one is hard.
Most important, is to make sure developer is facing clear and simple API to work with, and don't have extra learning curve, or memorisation needed to benefit from new features.

mvaligursky:
I've done some updates to mostly LUT tables:
- initialized from light.js, when its set to area light type
- handles float / half float / rgba8 format (last one with some clipping, but visually seems ok - added TODO to improve)

I’ve noticed shader only uses .xy from lut2, .zw is unused - @marklundin - is there any need for those in the future or can I strip them out from luts?

raytranuk:
@marklundin - Just wanted to keep you up-to-date: After some offline discussion in the PlayCanvas team, I have been assigned to make some final changes to your excellent contribution, so it aligns with the future plans for the PlayCanvas lighting system - and probably subsequently expose your work in the PlayCanvas editor. I'll try to work on the engine side in the next couple of weeks - and the editor probably further down the line. Thanks again for putting time into making this PR.

marklundin:
Thanks so much @raytranuk @mvaligursky for picking the torch up on this. I've been swamped the past few weeks so haven't been able to dedicate the time to this, but it seems you have everything in hand! Excited to see this come together. Once it's in the engine I've a proposal for textured area lights, which may need some pre-computation on the editor side (similar to the prefiltered env maps) but I will leave that for another day!

Maksims:
Just to add into conversation on light types, the way other engines do:

**Unreal Engine**: different type per light: directional, sky, spot, point, rect. Each light can have properties such as light source radius, or length (for point light, making it line). There is no concept of shape.

**Unity**: point, spot, directional, area. So each light is separate type. Shape is only available for area light.

Not sure we should invent the wheel here as well, it will be good to follow common practice with light types.

mvaligursky:
I would prefer to stay with these options, and shapes (shapes mostly to be done in the future, for now single hidden default shape)

**Point** - light from a point
- shapes: none

**Spot** - light from a point, constrained by shape
- shapes: cone (default), pyramid, box

**Directional** - infinitely far directional light
- it would have a disk diameter / radius property eventually to implement disk based directional light

**Area** - light source is a surface
- shapes: rectangle (default), disk, tube

I think keeping Area light as a separate type makes it clearer for the users. It also makes it easier to handle in the engine, as its behavior is defined by a single property (type). As @Maksims mentioned, it also makes it follow industry standard.

Regarding the code complexity - I don't see a clear advantage in either solution, they both will need different ifs to handle the complexity.

raytranuk:
I've updated this PR so that area lights are fully integrated into the engine's lighting systems.

NB:
- Area light shadows are approximated using punctual light shadows
- Directional area light specular is approximated using positional calculations

src/scene/constants.js
```diff
@@ -183,11 +183,11 @@ export var LAYERID_UI = 4;
 export var LIGHTTYPE_DIRECTIONAL = 0;
 /**
  * @constant
- * @name pc.LIGHTTYPE_POINT
+ * @name pc.LIGHTTYPE_OMNI
```
mvaligursky, line 186: it seems you have removed pc.LIGHTTYPE_POINT? Unless that was moved elsewhere. This probably breaks existing scripts? Test with engine tests, few of them would use it.

raytranuk: The API for creating lights uses the string "point" - which is aliased to LIGHTTYPE_OMNI - I didn't find anywhere that uses LIGHTTYPE_POINT directly

raytranuk: Unfortunately pc.LIGHTTYPE_POINT is part of the public API - so I'll put it back in - and mark pc.LIGHTTYPE_OMNI as private for now. Later we should mark pc.LIGHTTYPE_POINT as deprecated but still around for legacy support

mvaligursky: good point. Either way, we need to keep it alive, as its public and likely some people use it. Maybe add it to deprecated.js, something similar to 
`export var RIGIDBODY_TYPE_STATIC = BODYTYPE_STATIC;`

raytranuk: done

raytranuk:
> ideally we'd test format of lut here (or read some flag based on format) instead of this condition which needs to match functionality of LUT creation

I moved the logic for LUT format to device startup so that it is constant and can be used in shader generation - which can be done before LUT texture initialization.

examples/graphics/area-lights.html
```diff
@@ -0,0 +1,269 @@
...
+                castShadows: shadows,
+                innerConeAngle: 80,
+                outerConeAngle: 85,
+                shadowBias: 0.1,
+                normalOffsetBias: 0.1,
+                shadowResolution: 2048                
+            });
+            
+            light.setLocalScale( scale, scale, scale);
+            light.setLocalEulerAngles( (type === "directional") ? 90 : 0, 0, 0);
```
mvaligursky, line 92: just curious about why this is needed here?

examples/graphics/area-lights.html
```diff
@@ -0,0 +1,269 @@
...
+            lightParent.addChild(light);
+
+            // emissive material that is the light source color
+            var brightMaterial = new pc.StandardMaterial();
+            brightMaterial.emissive = color;
+            brightMaterial.useLighting = false;
+            brightMaterial.cull = (shape===pc.LIGHTSHAPE_RECT) ? pc.CULLFACE_NONE : pc.CULLFACE_BACK;
```
mvaligursky, line 99: spaces around ===

src/graphics/program-lib/chunks/fresnelSchlick.frag
```diff
@@ -1,7 +1,22 @@
 // Schlick's approximation
 uniform float material_fresnelFactor; // unused
 
+#ifdef HAS_AREA_LIGHTS
```
mvaligursky, line 4: should this be HAS_AREA_LIGHTS or just AREA_LIGHTS?
(looking at CLEARCOAT there as an example)

src/graphics/program-lib/chunks/ltc.frag
```diff
@@ -0,0 +1,447 @@
...
+	coords.coord1 = lightPos - halfWidth - halfHeight;
+	coords.coord2 = lightPos - halfWidth + halfHeight;
+	coords.coord3 = lightPos + halfWidth + halfHeight;
+	return coords;
+}
+//used for simple sphere light falloff
+float dSphereRadius;
+Coords getSphereLightCoords(vec3 lightPos, vec3 halfWidth, vec3 halfHeight){
+	Coords coords;
+	float radius = length(halfWidth);
```
mvaligursky, line 101: should we use max of width and height?

src/graphics/program-lib/chunks/ltc.frag
```diff
@@ -0,0 +1,447 @@
...
+	vec3 f = normalize(lightPos-view_position);
+	vec3 w = normalize(cross(f, halfHeight));
+	vec3 h = normalize(cross(f, w));
+
+	coords.coord0 = lightPos + w*radius - h*radius;
+	coords.coord1 = lightPos - w*radius - h*radius;
+	coords.coord2 = lightPos - w*radius + h*radius;
+	coords.coord3 = lightPos + w*radius + h*radius;
+	return coords;
+}
```
mvaligursky, line 114: spaces around operators

src/graphics/program-lib/chunks/ltc.frag
```diff
@@ -0,0 +1,447 @@
...
+    // Algorithm A
+    {
+        float A_a = 1.0;
+        float C_a = Delta.x;
+        float D_a = -2.0*B*Delta.x + Delta.y;
+
+        // Take the cubic root of a normalized complex number
+        float Theta = atan(sqrt(Discriminant), -D_a)/3.0;
+
+        float x_1a = 2.0*sqrt(-C_a)*cos(Theta);
```
mvaligursky, line 208: spacing

src/graphics/program-lib/programs/standard.js
```diff
@@ -998,6 +1000,22 @@ var standard = {
...
         var usePerspZbufferShadow = false;
         var light;
+
+        var hasAreaLights = options.lights.some(function (light){
+            return light._shape !== LIGHTSHAPE_PUNCTUAL;
+        });
+
+        if (device._areaLightLutFormat === PIXELFORMAT_R8_G8_B8_A8) {
+            // use offset and scale for rgb8 format luts
+            code += "#define HAS_R8_G8_B8_A8_LUTS\n";
```
mvaligursky, line 1010: and similarly here, not sure the HAS part is needed in HAS_R8_G8_B8_A8_LUTS
maybe: AREA_LUTS_R8_G8_B8_A8

src/graphics/program-lib/programs/standard.js
```diff
@@ -1454,18 +1483,53 @@ var standard = {
...
+                code += "   dSpecularLight *= dSpecularity;\n";
+
+                code += "   float roughness = max((1.0 - dGlossiness) * (1.0 - dGlossiness), 0.001);\n";
+            }
+
+            // light source shape support
+            var shapeStringMap = [];
+            shapeStringMap[LIGHTSHAPE_RECT] = 'Rect';
+            shapeStringMap[LIGHTSHAPE_DISK] = 'Disk';
+            shapeStringMap[LIGHTSHAPE_SPHERE] = 'Sphere';
```
mvaligursky, line 1503: this should be moved outside to be initialized one time only

src/scene/forward-renderer.js
```diff
@@ -1061,7 +1119,7 @@ Object.assign(ForwardRenderer.prototype, {
         var i;
         var point, spot;
 
-        var pnts = sortedLights[LIGHTTYPE_POINT];
+        var pnts = sortedLights[LIGHTTYPE_OMNI];
```
mvaligursky, line 1122: maybe rename pnts to omnis

mvaligursky:
I had a look and this looks good to me. I added few small comments. I'm approving this, but please wait for 1-2 more approvals before merging.

Additional comments:
1. please test this in local environment inside the Editor to make sure existing functionality is working as expected
2. create Editor issue if not already to expose shape property and add gizmo rendering for area light shapes
	- ideally we would expose area lights in the editor behind flag pretty early, to allow us to conveniently test this functionality more comprehensively

mvaligursky (approved):

src/graphics/program-lib/chunks/fresnelSchlick.frag
raytranuk: I was influenced by the naming of the javascript var in the standard shader generator which sets the macro - I can change it

src/graphics/program-lib/chunks/ltc.frag
raytranuk: makes sense

src/graphics/program-lib/chunks/ltc.frag
raytranuk: I will add spaces - NB this code was copied directly from the original implementation (linked at the top of the file) - spacing between `+` and `-`  is different than `*` and `/` - resulting in IMO very readable code - e.g `A*B + C*D` vs our standard which is `A * B + C * D`

willeastcott: I would say stick to the standard. If you want to argue for a change to the standard, then we migrate the whole codebase. We should avoid deviations made on a whim.

examples/graphics/area-lights.html
raytranuk: it wasn't actually needed - was a overhang from using lookAt() - removed

src/graphics/program-lib/programs/standard.js
raytranuk: moved it into a private function - making it more consistent with the standard material generator

src/scene/forward-renderer.js
raytranuk: renamed in all code except point light shadows - which are assuming punctual shape

examples/assets/scripts/lights/area-light-lut.js
```diff
@@ -0,0 +1,12 @@
...
...9861, 0.0107655, 0, 7.12986e-05, 0.999763, 0.0108497, 0, 0.000743959, 0.999616, 0.0109716, 0, 0.00227361, 0.999408, 0.0111408, 0, 0.0046983, 0.999112, 0.0113659, 0, 0.00800158, 0.998637, 0.0116475, 0, 0.0121493, 0.996223, 0.0117231, 0, 0.0171023, 0.994006, 0.0118064, 0, 0.0228218, 0.992444, 0.0120254, 0, 0.0292711, 0.991028, 0.0123314, 0, 0.036417, 0.98803, 0.0124954, 0, 0.0442295, 0.984816, 0.0126538, 0, 0.0526815, 0.981399, 0.0128537, 0, 0.0617492, 0.977085, 0.0129694, 0, 0.0714114, 0.972154, 0.013091, 0, 0.0816495, 0.966617, 0.0131166, 0, 0.0924472, 0.960628, 0.0131583, 0, 0.10379, 0.953295, 0.0131094, 0, 0.115665, 0.94575, 0.0129966, 0, 0.128062, 0.937654, 0.0128796, 0, 0.140972, 0.927716, 0.0126477, 0, 0.154387, 0.917932, 0.0123889, 0, 0.168301, 0.907719, 0.012131, 0, 0.182709, 0.89584, 0.0118013, 0, 0.197608, 0.883526, 0.0114145, 0, 0.212994, 0.870301, 0.0110075, 0, 0.228867, 0.856272, 0.0106019, 0, 0.245227, 0.842251, 0.0101938, 0, 0.262074, 0.826466, 0.00973254, 0, 0.279412, 0.810859, 0.0092846, 0, 0.297244, 0.795051, 0.00883304, 0, 0.315575, 0.780053, 0.00840272, 0, 0.334412, 0.76575, 0.00796438, 0, 0.35376, 0.752298, 0.00752526, 0, 0.373631, 0.739153, 0.00711486, 0, 0.394
```
willeastcott, line 9: I don't really like the fact that this asset/data make assumptions about the engine. This binds the data to a specific engine and means we can't change where the data is expected to be found. Also, `pc.app` was never actually officially made API. I'd prefer it was either in a global var or maaaaybe placed on the `pc` object. But I'm open to suggestions. All I know is that when I look at this, I get a bit worried. 😨

src/scene/constants.js
```diff
@@ -196,6 +204,42 @@ export var LIGHTTYPE_POINT = 1;
...
+export var LIGHTSHAPE_RECT = 1;
+
+/**
+ * @constant
+ * @private
+ * @name pc.LIGHTSHAPE_DISK
+ * @type {number}
+ * @description Disk shape of light source.
+ */
+export var LIGHTSHAPE_DISK = 2;
```
willeastcott, line 232: I saw this and immediately thought it should be DISC. But then I read that DISK is American English. So let's stick with this. Although interesting to see Unity goes with Disc: https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@8.0/api/UnityEngine.Rendering.HighDefinition.AreaLightShape.html

raytranuk: Using American English confuses fewer people - when I first added it, I added it as DISC - but changed it to DISK.

src/scene/light.js
```diff
@@ -311,6 +320,112 @@ Object.assign(Light.prototype, {
...
+            var ret = new Uint8ClampedArray(count);
+            for (var i = 0; i < count; i++) {
+                ret[i] = data[i] * 255;
+            }
+
+            return ret;
+        }
+
+        // create texture if not already
+        var app = Application.getApplication();
```
willeastcott, line 383: The scene layer should not have a dependency on the `Application`. This forms a circular dependency.

examples/assets/scripts/lights/area-light-lut.js
raytranuk: This code was added by Martin - I would maybe just have done it with a global - but maybe Martin has reasons? @mvaligursky - thoughts?

mvaligursky: no reason .. that's what we discussed on one of the syncs, feel free to change it.

src/scene/light.js
mvaligursky: this is existing pattern in this file already, as constructor does not take device and so this is how its obtained. I think this needs more global solution as it's taking place in few places.

willeastcott: OK. I just think we need to resolve this soon because otherwise, we'll simply continue in the wrong direction.

mvaligursky: created: https://github.com/playcanvas/engine/issues/2651

examples/assets/scripts/lights/area-light-lut.js
raytranuk: cool - changed to a global var

willeastcott (approved):

raytranuk (approved):
@marklundin - I updated your PR description to line up with the final implementation in the PR. Well done and thank you again for the contribution to the PlayCanvas engine.

willeastcott:
Yeah, thanks for this awesome contribution @marklundin! It's so exciting to see this get merged. 😄 🚀

marklundin:
Great news @willeastcott and @raytranuk, excited to see this land! Any timeline on editor support?

willeastcott:
So timeline will be:

* Deploy v1.38.0 - guesstimate: 11th January, which should give us the previous week to test on our dev server.
* Develop Editor UI changes.
* Test Editor Area Light UI.
* Deploy Editor Area Light UI behind a flag. We'll whitelist you straight away. Maybe 18th Jan - but we'll keep you updated.

willeastcott:
Note that once the engine is deployed, you'll be able to drive the new functionality in a script as a stop gap. 😄

PR #2537, opened 2020-11-06
Fix Bokeh Depth of Field post process

### LINKED ISSUE #1704: Accessing the DepthMap in WebGL1 vs WebGL2
I am attempting to access the DepthMap from a custom shader (PostEffect) and I am unable to properly grab when falling back to WebGL1. In WebGL2 it works as expected.

### Description

* URL for sample scene: https://playcanvas.com/editor/scene/813310
* Tested on Firefox/Chrome
* Related forum thread: https://forum.playcanvas.com/t/depth-buffer-in-material-shader/9120/6

Tried it in two ways:

1. Enabled the **needsDepthBuffer** property in the PostEffect class. Result:

WebGL1: no depth map is rendered
WebGL2: depth map is correctly rendered

1. Disabled the **needsDepthBuffer** property in the PostEffect class and added a custom pc.Script named **RenderDepth** which is used to incrementCounter() on the depthLayer. Result:

WebGL1: no depth map is rendered unless I enable realtime shadows on a directional light. Then a depthMap is rendered and made available as **uDepthMap** uniform but not for the camera perspective but for the light perspective (which is not what is expected).
WebGL2: depth map is correctly rendered

Any idea if I am doing something wrong?

raytranuk:
Closes #1704

The bokeh post process can now be referenced as an example of how to access depth in WebGL1 and 2

The post process example was also updated to use the fixed bokeh better.

### DIFF
--- a/examples/assets/scripts/posteffects/posteffect-bokeh.js
+++ b/examples/assets/scripts/posteffects/posteffect-bokeh.js
@@ -39,11 +39,12 @@ Object.assign(pc, function () {
             ].join("\n"),
             fshader: [
                 "precision " + graphicsDevice.precision + " float;",
+                (graphicsDevice.webgl2) ? "#define GL2" : "",
+                pc.shaderChunks.screenDepthPS,
                 "",
                 "varying vec2 vUv0;",
                 "",
                 "uniform sampler2D uColorBuffer;",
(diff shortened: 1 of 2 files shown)


mvaligursky (approved):
looks good to me, glad this works now!

PR #2550, opened 2020-11-11
Added public API for Container resource

yaustar:
JS docs for the properties of a container resource


mvaligursky:
Maybe we should hold off exposing model api, the others are fine. With the changes to render component, we have new API to create entity with model, or entity with render on it.

yaustar:
Ah, there's interesting to hear that the API might change.

yaustar:
Removed model

mvaligursky:
it might not change (and likely will not), but we might not make it public and only expose the new api

mvaligursky (approved):

willeastcott (approved):

PR #2551, opened 2020-11-11
Fixed bug where animations are not correctly removed from component when asset is unloaded

yaustar:
Repro project: https://playcanvas.com/editor/scene/1029320
Press 3 to unload the animation asset. 
Press 4 to load the animation asset again.

Fixes exception errors with a clients project

### DIFF
--- a/src/framework/components/animation/component.js
+++ b/src/framework/components/animation/component.js
@@ -236,6 +236,12 @@ Object.assign(AnimationComponent.prototype, {
     onAssetChanged: function (asset, attribute, newValue, oldValue) {
         var i;
         if (attribute === 'resource' || attribute === 'resources') {
+            // If the attribute is 'resources', newValue can be an empty array when the
+            // asset is unloaded. Therefore, we should assign null in this case
+            if (newValue && newValue.length === 0 && attribute === 'resources') {
+                newValue = null;
+            }
+
             // replace old animation with new one
             if (newValue) {
                 var restarted = false;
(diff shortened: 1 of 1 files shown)


ellthompson (approved):

src/framework/components/animation/component.js
```diff
@@ -236,6 +236,12 @@ Object.assign(AnimationComponent.prototype, {
     onAssetChanged: function (asset, attribute, newValue, oldValue) {
         var i;
         if (attribute === 'resource' || attribute === 'resources') {
+            // If the attribute is 'resources', newValue can be an empty array when the
+            // asset is unloaded. Therefore, we should assign null in this case
+            if (newValue && newValue.length === 0 && attribute === 'resources') {
```
slimbuck, line 241: You probably want to change the order here. If newValue is a `resource` instead of `resources`, accessing length member is probably wrong.

Just change the order to:
```javascript
if (attribute === 'resources' && newValue && newValue.length === 0) {
```

yaustar: I had the string comparison last as I wanted the 'more' expensive comparison last. `new.length` will be undefined with an Object type so `newValue.length === 0` would be false.

It's also inline with the logic in this function: https://github.com/playcanvas/engine/blob/master/src/framework/components/animation/component.js#L248

Happy to change it though, I'm not overly keen on the order but really wanted the string comparison last 😬

slimbuck: Javascript strings are interned so the string test is actually a pointer comparison. (i.e. it's fast)

vkalpias: > Javascript strings are interned so the string test is actually a pointer comparison. (i.e. it's fast)

I think that depends on the browser?

PR #2555, opened 2020-11-12
Move glTF material conversion shader code into shader chunks

### LINKED ISSUE #2404: Integrate glb-parser material chunks into the engine
In order to implement glb materials, glb parser contains some additional shader chunks:
https://github.com/playcanvas/engine/blob/0e3022cef2075f6c8fd7d49ffddd54521a68c0d9/src/resources/parser/glb-parser.js#L794-L881

These should be integrated into the engine chunk system, allowing glb parser to just set some options on material.

raytranuk:
Fixes #2404


src/graphics/program-lib/chunks/clearCoatGloss.frag
```diff
@@ -21,5 +21,9 @@ void getClearCoatGlossiness() {
     ccGlossiness *= saturate(vVertexColor.$VC);
     #endif
 
+    #if GLTF_MAT_GENERATED
+    dGlossiness = 1.0 - dGlossiness;
```
slimbuck, line 25: is this a better solution than setting roughness like this on the material directly at load time in the glb parser?

raytranuk: Up until the invert, glossiness is actually really roughness - so we could indeed invert it earlier - but we would have to invert all the components that can make up the roughness including the roughness factor - which is easy and roughness map texture channel which is trickier. Alternatively we could just invert the texture sample - and convert roughness factor?

raytranuk: Hmm looks like I made a mistake in this PR - not dealing with specular-glossiness extension correctly - will fix this

raytranuk: spec-gloss fixed and only roughness texture inverted in shader chunk

raytranuk:
NB engine visual test now pass

src/resources/parser/glb-parser.js
```diff
@@ -1039,14 +1039,15 @@ var createMaterial = function (gltfMaterial, textures, disableFlipV) {
...
-            material.clearCoatGlossiness = ccData.clearcoatRoughnessFactor;
+            material.clearCoatGlossiness = 1.0 - ccData.clearcoatRoughnessFactor;
         } else {
             material.clearCoatGlossiness = 0;
         }
         if (ccData.hasOwnProperty('clearcoatRoughnessTexture')) {
             var clearcoatRoughnessTexture = ccData.clearcoatRoughnessTexture;
             material.clearCoatGlossMap = textures[clearcoatRoughnessTexture.index];
             material.clearCoatGlossMapChannel = 'g';
+            material.hasRoughnessMap = true;
```
Maksims, line 1050: It is probably `isRoughnessMap` based on conventions used in engine, also as it is same texture (not separate).

raytranuk: makes sense - switched to is instead of has - thanks!

raytranuk:
Switching to draft - as we are now re-considering a switch to roughness proper

willeastcott:
Reassigning to you @slimbuck. But this is so old, it feels like we should maybe just close it? Up to you...

willeastcott:
Unilaterally deciding to close this now due to inactivity.

PR #2558, opened 2020-11-16
fix shader compile error when forceUv1 option is used

### LINKED ISSUE #2539: Lightmap baking generates incorrect shader, causing WebGL errors in Safari.
Below is a minimal project and a screenshot showing the WebGL errors.
The errors are seen in Safari Mac/iOS.
Here is a forum thread https://forum.playcanvas.com/t/webgl-errors-safari-mac-ios-console/15699
This may be the code issue https://github.com/playcanvas/engine/blob/1db4031ab7d8cad5fb137ac456223c6af90e2c0d/src/scene/procedural.js#L915

https://playcanvas.com/project/734901/overview/blank-project-test

c6e70c08ec0083a13954f103279d8990626a58bc_2_1310x1000

The only changes I made to the default blank project are:

Light
- checked bake light map
- unchecked Affect Dynamic

Box
- checked lightmapped

Plane
- checked lightmapped

raytranuk:
Fixes #2539

### DIFF
--- a/src/graphics/program-lib/programs/standard.js
+++ b/src/graphics/program-lib/programs/standard.js
@@ -632,7 +632,7 @@ var standard = {
 
         if (options.forceUv1) {
             useUv[1] = true;
-            useUnmodifiedUv[1] = useUnmodifiedUv[1] ? useUnmodifiedUv[1] : true;
+            useUnmodifiedUv[1] = (useUnmodifiedUv[1] !== undefined) ? useUnmodifiedUv[1] : true;
         }
 
         for (i = 0; i < maxUvSets; i++) {
--- a/src/graphics/program-lib/programs/standard.js
+++ b/src/graphics/program-lib/programs/standard.js
@@ -637,9 +637,10 @@ var standard = {
                 attributes["vertex_texCoord" + i] = "TEXCOORD" + i;
                 code += chunks["uv" + i + "VS"];
                 codeBody += "   vec2 uv" + i + " = getUv" + i + "();\n";
-            }
-            if (useUnmodifiedUv[i]) {
-                codeBody += "   vUv" + i + " = uv" + i + ";\n";
+
+                if (useUnmodifiedUv[i] === true || useUnmodifiedUv[i] === undefined) {
+                    codeBody += "   vUv" + i + " = uv" + i + ";\n";
+                }
             }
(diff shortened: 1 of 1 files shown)


src/graphics/program-lib/programs/standard.js
```diff
@@ -637,9 +637,10 @@ var standard = {
                 attributes["vertex_texCoord" + i] = "TEXCOORD" + i;
                 code += chunks["uv" + i + "VS"];
                 codeBody += "   vec2 uv" + i + " = getUv" + i + "();\n";
-            }
-            if (useUnmodifiedUv[i]) {
-                codeBody += "   vUv" + i + " = uv" + i + ";\n";
+
+                if (useUnmodifiedUv[i] === true || useUnmodifiedUv[i] === undefined) {
```
mvaligursky, line 641: feels bit strange to do it if it's set to undefined. Should this maybe be fixed elsewhere in the code and make this to be set to true?

raytranuk: I'll move it to the place where forceUv1 is applied (just above)

src/graphics/program-lib/programs/standard.js
```diff
@@ -630,7 +630,10 @@ var standard = {
             }
         }
 
-        if (options.forceUv1) useUv[1] = true;
+        if (options.forceUv1) {
+            useUv[1] = true;
+            useUnmodifiedUv[1] = useUnmodifiedUv[1] ? useUnmodifiedUv[1] : true;
```
mvaligursky, line 635: isn't this line equivalent to just `useUnmodifiedUv[1] = true;` ?

raytranuk: I don't think it is - the varying can be set to something else if it is used and there is a uv transformation - so it should skip the code that checks the useUnmodifiedUv and sets vUv1.

raytranuk: oh wait - I fix this

raytranuk: fixed.

mvaligursky (approved):

raytranuk:
Also tested with: https://playcanvas.com/project/737968/overview/force-uv1-test

PR #2566, opened 2020-11-18
resource id fix

### LINKED ISSUE #2501: Invalid URL from cubemap resource handler
Example scene: https://playcanvas.com/editor/scene/480301.

Open the Cubemap folder in the Asset Panel and notice all the errors about a 'scene not found'. The issue is that each asset has its ID as the URL so the browser tries to load it as a relative URL.

This is set in `engine/src/resources/cubemap.js` line 321.

slimbuck:
Fixes #2501

### DIFF
--- a/src/resources/cubemap.js
+++ b/src/resources/cubemap.js
@@ -213,6 +213,12 @@ Object.assign(CubemapHandler.prototype, {
         return true;
     },
 
+    // convert string id to int
+    resolveId: function (value) {
+        var valueInt = parseInt(value, 10);
+        return ((valueInt === value) || (valueInt.toString() === value)) ? valueInt : value;
+    },
+
     loadAssets: function (cubemapAsset, callback) {
         // initialize asset structures for tracking load requests
         if (!cubemapAsset.hasOwnProperty('_handlerState')) {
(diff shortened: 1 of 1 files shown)


src/resources/cubemap.js
```diff
@@ -297,7 +297,7 @@ Object.assign(CubemapHandler.prototype, {
             } else if (self.compareAssetIds(assetId, loadedAssetIds[i])) {
                 // asset id hasn't changed from what is currently set
                 onReady(i, loadedAssets[i]);
-            } else if (parseInt(assetId, 10) === assetId) {
+            } else if (parseInt(assetId, 10) == assetId) {
```
willeastcott, line 300: Why do === to ==?

willeastcott, line 300: This is a bit scary. I really dislike double equals and so does ESlint it seems:

https://eslint.org/docs/rules/eqeqeq

Some coder in the future might just decide to 'improve' the code and switch to triple equals, not understanding the rationale. Is there not a clearer way to deal with this?

slimbuck: I can add a comment with explanation?

slimbuck: something like:
```javascript
} else if (parseInt(assetId, 10) == assetId) {  // NOTE: use '==' to handle assetId being 123 or '123'.
```

slimbuck: wrapped the test in a function and used explicit '===' instead

willeastcott (approved):

PR #2574, opened 2020-11-21
WebXR Image Tracking

Maksims:
Implements experimental draft extension for WebXR Image Tracking: https://github.com/immersive-web/marker-tracking/blob/main/explainer.md
This API is a subject to change, until spec is more solid.
It is available in Chrome for Android 89, with enabled chrome://flags#webxr-incubations flag.

Image Tracking allows to provide few images and their widths, and then track them in real world even if their positions change, tracking will assume that actively tracked images can move in real world.

### New APIs:
```js
// pc.XrManager
app.xr.imageTracking // interface to access image tracking

// pc.XrImageTracking
imageTracking.supported // true if image tracking is supported
imageTracking.available // true if image tracking is currently available
imageTracking.images // list of pc.XrTrackedImage's that contain tracked information

imageTracking.add(image, width); // add image for tracking
imageTracking.remove(trackedImage); // remove tracked image from tracking
// modifying tracked images currently is possible only outside of AR session.
// This is by design of API, but a raised issue: https://github.com/immersive-web/marker-tracking/issues/2

imageTracking.on('error', function(err) { }); // fired when image tracking encountered an error

// pc.XrTrackedImage
trackedImage.image // image that been provided for tracking
trackedImage.width // width that been provided, can be changed outside of AR session
trackedImage.trackable // True if image is trackable. It might be untrackable by underlying system if image was provided is not meeting some requirements, like: at least 300x300 resolution.
trackedImage.tracking // True if image is tracking, or False if image tracking is lost.
trackedImage.emulated // True if image is not actively tracking, but recently was actively tracked, and transformation is emulated by underlying system

trackedImage.getPosition(); // get world space position of tracked image
trackedImage.getRotation(); // get world space rotation of tracked image

trackedImage.on('tracked', function() { }); // fired when image becomes actively tracked
trackedImage.on('untracked', function() { }); // fired when image is no more tracked
```

### Test project:
https://playcanvas.com/project/739875/overview/ar-image-tracking
In order to test, fork project, and replace example images with desired ones.

### Video:
https://twitter.com/mrmaxm/status/1330232934192853003


src/xr/xr-image-tracking.js
```diff
@@ -0,0 +1,162 @@
...
+ * @name pc.XrImageTracking#error
+ * @param {Error} error - Error object related to failure of image tracking.
+ * @description Fired when XR session is started, but image tracking failed to process provided images.
+ */
+
+/**
+ * @function
+ * @name pc.XrImageTracking#add
+ * @description Add image for image tracking, as well as width that helps underlying system to estimate proper transformation. Modifying tracked images list is only possible before AR session is started.
+ * @param {object} image - Image that is matching real world image as close as possible. Resolution of images should be at least 300x300. High resolution does NOT improve tracking performance. Color of image is irelevant, so greyscale images can be used. Images with too many geometric features or repeating patterns will reduce tracking stability.
```
willeastcott, line 42: Can we be more specific with the type here? `HTMLImageElement` maybe? That's what we use for pc.Texture#setSource:

https://developer.playcanvas.com/en/api/pc.Texture.html#setSource

src/xr/xr-tracked-image.js
```diff
@@ -0,0 +1,142 @@
+import { EventHandler } from '../core/event-handler.js';
+import { Vec3 } from '../math/vec3.js';
+import { Quat } from '../math/quat.js';
+
+/**
+ * @class
+ * @name pc.XrTrackedImage
+ * @classdesc The tracked image interface that is created by the Image Tracking system and is provided as a list from {@link pc.XrImageTracking#images}. It contains information about the tracking state as well as the position and rotation of the tracked image.
+ * @description The tracked image interface that is created by the Image Tracking system and is provided as a list from {@link pc.XrImageTracking#images}. It contains information about the tracking state as well as the position and rotation of the tracked image.
+ * @param {object} image - Image that is matching the real world image as closely as possible. Resolution of images should be at least 300x300. High resolution does NOT improve tracking performance. Color of image is irrelevant, so greyscale images can be used. Images with too many geometric features or repeating patterns will reduce tracking stability.
```
willeastcott, line 10: Again, should this be `HTMLImageElement`?

src/xr/xr-tracked-image.js
```diff
@@ -0,0 +1,142 @@
...
+    var self = this;
+
+    if (this._bitmap) {
+        return {
+            image: this._bitmap,
+            widthInMeters: this._width
+        };
+    }
+
+    return this._image.decode()
```
willeastcott, line 58: Do we need some level of error handling in this function? What is the decode or conversion to image bitmap fails for some reason?

Maksims: This is actually caught by https://github.com/playcanvas/engine/pull/2574/files#diff-9ded8a4095318b7afb9e35db6f2332793c8df84957dc591309f6a7a332dce45eR108

src/xr/xr-image-tracking.js
Maksims: Actually, looking better into this, Image Tracking consumes ImageBitmap, which can be constructed from many different image data sources: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/createImageBitmap#Parameters
So I should change it to handle such cases.

LeXXik:
This is a great contribution and really appreciated! I don't think Playcanvas would have WebXR support at the level it is today without Maksims PRs. This also made me think, what would happen if we don't have Maksims available tomorrow for one reason or another? Is there anyone else in the team, who would be able to take over the support of experimental API to make sure the main branch doesn't break after the API changes? Would it make sense to move WebXR to own branch, until WebXR API is stable or perhaps implement only approved API?

Maksims:
Good concerns @LeXXik, although, not planning to go anywhere ;)
The integration in engine is developed that way, that only affects users who actually start XR Session, so all those experimental APIs, do not affect non XR users.

Also, WebXR core API is stable, although there are [various API extensions](https://immersive-web.github.io/) are delivered as separate specs. They should be evaluated on individual basis. I'm actually planning to potentially change current way of session start, where developer have to specify which APIs he want optionally to enable, that way when starting XR session, it will be barebones, with options to make it more feature rich.

src/xr/xr-image-tracking.js
willeastcott: OK, sounds good.

willeastcott (approved):

Issue #2433, opened 2020-09-16
Migrate engine to typescript


slimbuck:
We would like to start migrating the engine codebase to typescript in order to leverage its well known advantages.

Given the engine codebase size though, it is unfeasible to convert everything in one go. Instead we must find a way of migrating the engine bit by bit (by file or module). Presumably we would start with the lowest-level code (i.e. the leafs of the code dependency tree) and work our way up to the highest level code. A good candidate might be the math lib.

One question is: how can we generate a set of docs and bindings for the engine when it comprises both JS and TS source? Has anyone attempted or managed this before?

Some of this might already be answered in #2213.

kungfooman:
I started the migration process once (with help of @jtenner), but in effect all upstream changes were so hard to merge and it just turned into an experiment (and due lack of time, it just died out): https://github.com/KILLTUBE/PlayCanvasTS

Also tried to generate C++ from TypeScript, so I would be able to use either AssemblyScript or Emscripten to generate either WebAssembly or native binaries, for reusing the PlayCanvas math code in different game engines/bindings (e.g. as binding for embeddable Duktape JavaScript interpreter).

`One question is: how can we generate a set of docs and bindings for the engine when it comprises both JS and TS source?`

Can you explain further? Not sure what you mean by "bindings" and which bindings have to be generated. TypeScript boils down by definition to JavaScript, the "simplest" thing would be to rename *.js to *.ts (it works, it will just rain errors tho).

slimbuck:
Hi @kungfooman,

Thanks for the info. By bindings I mean the playcanvas.d.ts typescript definitions we generate from jsdoc (and also the API reference documentation).

I'm not sure how we go about combining jsdocs from JS and TS files. (I haven't even tried this, perhaps it's a matter or running all source through tsc with some special flags?).

jtenner:
tsdoc is always an option. However, writing the source in TypeScript in the first place is usually better, because then d.ts can be generated as a byproduct without the help of jsdoc.

jtenner:
I realize that saying "starting with typescript" is unhelpful in this case, but perhaps just making a fork and switching to the typescript compiler in general should be a great first step.

Typescript will bundle regular js modules too.

mvaligursky:
two possible interesting links:

https://www.bookstack.cn/read/typescriptlang-3.7/spilt.7.055dccf6c3f8aae3.md
[...]

PR #2577, opened 2020-11-24
Migrate PlayCanvas to Typescript

willeastcott:
This PR migrates PlayCanvas from ES5 JavaScript to Typescript.

**What's done**
Initially, just the math and extras libraries have been fully ported.

**What's left to do**
* Migrate the rest of the JS files in `src`.
* Decide how we will continue to generate `playcanvas.d.ts` while the library is being migrated.
* Ensure the docs build process still works with `.ts` files instead of `.js` files.

**What are the benefits?**

* **ESnext language features now usable.** We've been stuck in an IE11/ES5 world for years now. This PR will suddenly enable us to use all the latest language features (arrow functions, default parameters, etc).
* **Guaranteed ES5 output.** It's no longer possible for ES6 to creep into the built library. Using ES6 now gets compiled to ES5 and using ES6 functions (like `Array.prototype.flat` for example - actually used [here](https://github.com/playcanvas/engine/blob/master/extras/mini-stats/gpu-timer.js#L87) in the extras lib) now throw an error at build time:
image
* **Great tools support for refactoring etc.** There's a rich set of tools for working with Typescript like [Visual Studio Code](https://code.visualstudio.com/) that really enhance the developer experience. You can jump to definition of variables and classes, get inline errors/warnings as you type code and so on.
* **Easier to scale development.** As the number of developers working on this codebase increases, switching to Typescript enables everyone to work in a more structured and controlled environment, that is less likely to introduce inconsistencies and bugs.
* **Easier path to experiment with AssemblyScript?** There's already a [PR](https://github.com/playcanvas/engine/pull/2213) open that explores the benefits of AssemblyScript. It will be easier to iterate on that with the codebase already in Typescript.

**What are the negatives?**
* The Rollup build process is around 25% slower (because of the Typescript compilation step). This might be mitigated by specifying a cache directory to the Typescript compiler, but this is yet to be investigated.
* The generated `playcanvas.js` might be very marginally larger (see next section for the reason).

**How has the built `playcanvas.js` changed?**
`playcanvas.js` is remarkably similar to what it was before. The main change is that classes have gone from:

```
	function Vec2(x, y) {
		// code...
	}
	Object.assign(Vec2.prototype, {
		add: function (rhs) {
			// code...
		}
	});
```

To:

```
	var Vec2 =  (function () {
		function Vec2(x, y) {
			// code...
		}
		Vec2.prototype.add = function (rhs) {
			// code...
		};
		return Vec2;
	}());
```

So `Object.assign` is no longer used to add function to the prototype and there's an additional closure to define the class. But it's still nice and readable.

**Feedback wanted!**
Let us know what you think. Would you like to see this PR merged? Do you think PlayCanvas should avoid Typescript? Do you have any tips for how to improve this PR to leverage Typescript better? Anything else? Your feedback is very much welcome!!


yaustar:
Would we need sourcemaps for the non-minified versions of the engine or would the generated code be easy enough to read and debug?

mvaligursky:
any way we could use class instead of function for classes?
`class Point {}`

Maksims:
I would add another negative: all current contributors, who do not know TypeScript, have to learn TypeScript to keep contributing.
This also applies to potential contributors.

willeastcott:
@yaustar  As I say in the initial PR description, the generated code is remarkably similar to the current `playcanvas.js`.

@mvaligursky The Typescript code _does_ use `class`. But when compiled to `playcanvas.js`, it converts it to a ES5-compatible function-based class definition.

@Maksims Agreed! It is a completely valid negative that non-TS coders will have to learn some additional syntax. But it's not a whole new language. It's JS with some type decoration. See the [vec3.ts](https://github.com/playcanvas/engine/pull/2577/files#diff-a8ec219ad2beb4f41ec1fe03c6dfc2a0f6ea850f07f038489d31896dd391724b) module for the minimal changes required. Also, I wouldn't say that the goal is _just_ to access 'modern' JS language features, although that's nice. The other benefits I list are also super-important (which Babel alone can't help with).

LeXXik:
A downside, as was mentioned, is that it adds a small learning curve for the current contributors, who are not familiar with Typescript. 
On the other hand, I found Typescript helps with a robustness of the system interoperability. All attributes become strictly typed, so only by this you already can eliminate most of unexpected results or crashes when components start interacting with each other. I remember there was another issue regarding a development of a validation system to catch invalid inputs early. TS would most certainly help in that regard.

goldenratio:
I don't think, 'learning curve' reason is still valid in 2020. If you can write JS, you can write TS.

Looks awesome!!

slimbuck:
This is wonderful @willeastcott!

It is slightly unfortunate that we are compelled to document types twice: once in the jsdoc and again in typescript's type specifier. Ideally the documented types would be inferred from typescript itself. Failing that it would be great if the tools could warn when the two types are mismatched. Does this already happen? Any idea how we could migrate the docs to use typescript types?

kungfooman:
TSDoc does not want types in JSDoc, so all that should be removed:

https://microsoft.github.io/tsdoc/#
image

I guess the question then is how to replace the JSDoc setup with TSDoc for `*.ts` files

mvaligursky:
> I guess the question then is how to replace the JSDoc setup with TSDoc for `*.ts` files

some related links to this in https://github.com/playcanvas/engine/issues/2433

kungfooman:
Just tested better-docs and I am running into errors:

RegEx to get rid of JSDoc types:

Search: `@(param|returns|property) \{(.+)\}`
Replace: `@$1`

image

Output:


image

(maybe there is an easy fix for this)

willeastcott:
Very interesting, @kungfooman - better-docs looks like a good solution to use while we port the whole codebase. And we can then switch over to tsdoc when we're done. I wonder if any tools are available that can convert jsdoc to tsdoc comments. I couldn't find anything.

If you're able to figure out the issue with the TYPE-ERRORs, that'd be a big help. I'm currently trying to figure out how to get rollup to generate playcanvas.d.ts and sourcemaps. It's a real pain!

Maksims:
Is there a path of migrating to TypeScript gradually, as I understand JS is valid in TS, so to simplify migration, and make sure this PR doesn't grow too large making it hard to merge after some time.
And that way if TypeScript will be now defacto standard, can this be more announced, specified in contributors docs, and coding guidelines, so that new PRs will be coming in TS.

kungfooman:
`I wonder if any tools are available that can convert jsdoc to tsdoc comments`

Do you mean the removal of type info from JSDoc comments?

I digged into better-docs yesterday to figure out the issue, today I just learned more about TSDoc to see how alternatives work, but TSDoc can't even extract type info from TS 🤔 

image

So having "double type information" (TS types + JSDoc) currently seems to be the best solution, will try to make better-docs work soon

kungfooman:
Using `better-docs` for this PR just requires four actions now:

1) Update `conf-api.json`:

image

2) Install better-docs: `npm i better-docs`

3) Merge: https://github.com/playcanvas/playcanvas-jsdoc-template/pull/18

4) Update `playcanvas-jsdoc-template` for this repo

I compared a bunch of old JS and TS doc pages and I can't see any difference:

image

willeastcott:
Awesome work, @kungfooman! I'll merge your PR and deploy it... 😄

willeastcott:
OK @kungfooman, [this commit](https://github.com/playcanvas/engine/pull/2577/commits/45f1bf7393bc0dd50c08c359d75172ab20e82341) has restored API reference generation by following your advice. Thanks! 🎉

willeastcott:
@Maksims Yeah, I think that's the best way to proceed. This PR is almost in a mergeable state. Once merged, we can migrate each `src` subfolder one by one over time.

kungfooman:
TS is just shoehorning JS, and it will leave us with less flexibility in the long run.

Just imagine a simple case like this:

image

If you want to fix the "issue" (in the mind of some folks at Microsoft), suddenly two simple lines become this TS mess:

image

(This is not just adding types! This is actually rewriting a `function` into a `var`, so the shoehorning works)

And then JavaScript has quite a `C/C++` legacy, where `true`/`false` basically just means `1` or `0`. The folks at Microsoft don't agree with that either: 

image

image

And then TS also restricts the use of closures for methods, like this case:

image

You solved that by putting the captured variables in the file context, but now they are overwriting each other and could lead to undefined behaviour, when multiple methods just share the same captured variables (just because TS is incapable at representing JS as they claim they do):

image

Of course we could just write TS in the "prototype" way, but then we are at the first issue again (the interface shoehorning).

And please note how they call everything ErRoRs. This is legit JavaScript, and TypeScript claims "WE ARE JUST TYPES AND COMPILE TO JAVASCRIPT", but that couldn't be further from the truth. It wants you to rewrite your code into what they consider "legit". And what they consider legit is not JavaScript. And people like JS because it is as dynamic as it is, and not some shoehorned OOP sugar.

MaxGraey:
@kungfooman This is not basically true. TS never try force you write in different style. Ofc it more preferred when you use modern classes, interfaces and etc. In this case you don't need extra efforts. 

So how to resolve you example? Just declare constructor for Foo:

```ts
declare class Foo {
    constructor();
}

function Foo(): void {

}

const foo = new Foo();
```

And you successfully compile this.

kungfooman:
Thank you for your example @MaxGraey, I just wanted to show the issue with captured variables for methods, because that pattern is used a few times and I don't know a proper way to express it with the same ease as with JS prototypes:

image

The compressed fix you suggested:

image

People basically have to write more "fix it up code" than real code (because TS only cares about the `class` paradigm sugar). If you want a compressed line of code, you write another/longer line just as sugar.

willeastcott:
That closure pattern is actually no longer required in the codebase since I ported the engine to ES6 modules (for Rollup). We can now put module 'globals' at the top of each module, safe in the knowledge that Rollup will rename them as necessary to avoid name clashes with the globals of other modules.

kungfooman:
@willeastcott It creates some mental overhead to consider when suddenly temporary variables are shared across methods. Methods are no longer "separate" units and are able to mess each others values up. I realize the module boundary ("globals" in `Mat3` are invisible for `Mat4` e.g.), but in this case they are all in `Mat4.ts`.

Lets consider this example:

image

image

(oops, `10` seems to be `0` now)

It may seem a bit "far fetched", but it imo still violates the principle of least surprise. How would one find such a bug even, once every `return function` in the code base is rewritten in this paradigm (there are more than a dozen)?

If we adopt this pattern, then every method should at least get its own temporary variables:

```ts
// Reduce GC via preallocated variables for specific methods
let getScale_x = new Vec3();
let getScale_y = new Vec3();
let getScale_z = new Vec3();
let getEulerAngles_scale = new Vec3();
```

trusktr:
These are great steps. Congrats on the move. I'd suggest convert to class syntax, so

```js
	function Vec2(x, y) {
		this.x = 0
		// ...
	}
	Object.assign(Vec2.prototype, {
		add: function (rhs) {
			// code...
		}
	});
```

becomes

```js
class Vec2() {
	x = 0
	// ...
	add(rhs) {
		// ...
	}
}
```

trusktr:
@kungfooman if the methods are all synchronous, variables named for every method are not needed. If the most cache variables that the methods ever use happens to be 3, for example, then

```js
let x = new Vec3();
let y = new Vec3();
let z = new Vec3();
```

is all that's needed for all methods.

I think the benefits of TS far outweigh this matter of cache variables being outside the place where methods are defined. It's easy to live with it for the benefits.

willeastcott:
@trusktr That's exactly the format being used in this PR:

https://github.com/playcanvas/engine/pull/2577/files#diff-24213396f60c3fc2a7a134a5aac2783bcf3bd5f44b474e0c50926c1796f9f62f

@kungfooman Yeah, I see your point - I guess we'd just need to be careful. 😄  That closure technique is maybe only use 4 or 5 times throughout the engine I think - it's not a big deal IMHO.

Maksims:
I've ran a poll in twitter, of course it is biased and does not validate any assumtions, but is interesting to see the results:
https://twitter.com/mrmaxm/status/1340658224333205504

image

MaxGraey:
> I've ran a poll in twitter, of course it is biased and does not validate any assumtions, but is interesting to see the results:

Hmm, but how this correlate to library which could be written on JS, TS, wasm or node.js native addon under the hood? For example many web developers use sass which actually uses LibSass written on C (now they decided move to SartSass and use Dart lang btw).

I mean users could still use JavaScript on their code space. But with library / framework they are got a lot of benefits - better intellisense, correct ts definitions (for users who picked TS).

Maksims:
> Hmm, but how this correlate to library which could be written on JS, TS, wasm or node.js native addon under the hood? For example many web developers use sass which actually uses LibSass written on C (now they decided move to SartSass and use Dart lang btw).

This conversation is not about engine users, it is about engine developers (community here).

> I mean users could still use JavaScript on their code space. But with library / framework they are got a lot of benefits - better intellisense, correct ts definitions (for users who picked TS).

We have great intellisense already, thanks to JSDocs and Tern.js, TypeScript would make small difference here. 

> Maintainers got error-prone refactorings and self-documented codebase.

JSDocs does great job making developers to consciously write docs. Some things could be auto-generated true, not descriptions and examples though, but is it a major problem right now?

Recently found a good article of someone who had large projects, open source as well, in both: JS and TS. And there is a good analysis on cons and pros based on experience: https://medium.com/javascript-scene/the-typescript-tax-132ff4cb175b

mvaligursky:
It's certainly an interesting article you linked @Maksims . Searching for the mentioned 38% bug reduction in airbnb codebase led me to this - perhaps we could see if that could be useful https://www.infoq.com/news/2020/08/airbnb-typescript-migration/

The author of the article you linked mentioned an average of up to 20% reduction of bugs, assuming you have other systems to catch bugs first - like unit testing. As our units test coverage is not very huge, perhaps we can get to even bigger wins?

Maksims:
> The author of the article you linked mentioned an average of up to 20% reduction of bugs, assuming you have other systems to catch bugs first - like unit testing. As our units test coverage is not very huge, perhaps we can get to even bigger wins?

You might have missread it, but he mentioned *theoretical maximum of possible bug reduction of 20%*, if nothing else is preventing bugs, like we have in engine: linter + tests + closure compiler.
With reference link to source: http://earlbarr.com/publications/typestudy.pdf

MaxGraey:
> You might have missread it, but he mentioned theoretical maximum of possible bug reduction of 20%, if nothing else is preventing bugs, like we have in engine: linter + tests + closure compiler.

That article have couple of updates. See `Update: 2019–02–11` which describe real feedback from Airbnb. Airbnb recently reported a 38% reduction in bugs by adding TypeScript to their development process

MaxGraey:
Let's look at recent bugs in three.js which will never appeared if they used TS:

1) https://github.com/mrdoob/three.js/issues/20826
2) https://github.com/mrdoob/three.js/issues/20769
3) https://github.com/mrdoob/three.js/issues/20054

Maksims:
Even during this PR, two bugs that has been caught:
1. https://github.com/playcanvas/engine/blob/master/src/shape/bounding-box.js#L89 - this is simply harmless.
2. https://github.com/playcanvas/engine/blob/master/src/shape/bounding-sphere.js#L44 - this is lack of test coverage.

> Let's look at recent bugs in three.js which will never appear if they used TS:
> 
> 1. [mrdoob/three.js#20826](https://github.com/mrdoob/three.js/issues/20826)
> 2. [mrdoob/three.js#20769](https://github.com/mrdoob/three.js/issues/20769)
> 3. [mrdoob/three.js#20054](https://github.com/mrdoob/three.js/issues/20054)

Are those bugs catchable by good test coverage?
First one for sure.

MaxGraey:
> Are those bugs catchable by good test coverage?

But if you could catch this bugs before write tests it's much better, isn't? Also make 100% caverage is pretty hard especially for huge and complex projects

Maksims:
> > Are those bugs catchable by good test coverage?
> 
> But if you could catch this bugs before write tests it's much better, isn't? Also make 100% test coverage is pretty hard especially for huge and complex projects

Nothing is ever 100%. Nor TS will magically catch all the bugs. It simply won't. Even in this PR, it caught only one breaking bug. Which would be caught by anyone ever using this method.

willeastcott:
I'm worried that we're conflating the benefits of modern JavaScript with the benefits of Typescript. One way or another, we *should not* stay on ES5.

I have a suggestion. Why not split this endeavor into two phases:

1. Introduce Babel into the build Rollup build process. Port the engine codebase to ES6 classes (and adopt arrow functions, `let` and `const` and other post ES5 language features).
2. Recreate this PR and reassess the benefits.

I suspect we'll then have a more focused discussion on the specific benefits of Typescript.

mvaligursky:
I'd like to have access to interfaces, which typescript provides, to enforce some rules / function declarations in the engine.

For example IDestroyable and IContextLoser (name to be decided) to enforce functions to handle destroy and loseContext.

I am also thinking about how to go go about the WebGPU device implementation, and having interface to enforce common function / property declarations between renderers would be great. Otherwise we're just hoping both renderers are kept in sync.

Are there any good alternatives in JS?

Maksims:
> I'd like to have access to interfaces, which typescript provides, to enforce some rules / function declarations in the engine.
> 
> For example IDestroyable and IContextLoser (name to be decided) to enforce functions to handle destroy and loseContext.
> 
> I am also thinking about how to go go about the WebGPU device implementation, and having interface to enforce common function / property declarations between renderers would be great. Otherwise we're just hoping both renderers are kept in sync.
> 
> Are there any good alternatives in JS?

We've followed more flat structure in code for a long time, similar to other engines. And generally in the web, OOP - is not popular and not used often. Functional approach, and more flat dependencies/functionality - to keep things less tangled, and easier to work with.

OOP in the web, is not popular, probably for reasons.

mvaligursky:
Interface does not change final code in any way, it's all about compile time validation / enforcement.
I agree going crazy with class hierarchies is not good for performance in many languages.

kungfooman:
@mvaligursky TypeScript definitions are mixable with JSDoc:

Example:

*interfaces.d.ts*

```ts
interface Destroyable {
    destroy(): void;
    isDestroyed(): boolean;
}
```

Then add this to any class:

````js
@implements {Destroyable}
```

Enable typechecking for JS via *tsconfig.json*:

```json
{
    "compilerOptions": {
        "allowJs": true,
        "checkJs": true,
        "outDir": "./nowhere",
        "target": "ES6",
        "noImplicitAny": true
    }
}

```

Test it on any class:

![image](https://user-images.githubusercontent.com/5236548/103366987-38af7380-4ac4-11eb-827a-1991569ca2e8.png)

mvaligursky:
@kungfooman that sounds great! I'll give it a test. I assume this can only test function existence, but not matching parameters?

kungfooman:
@mvaligursky I tested with parameters:

image

Correct error messages for missing interface functionality:

image

But it requires `npm run tsd` to generate `build/playcanvas.d.ts` (otherwise `pc.Vec3` is unknown, and TypeScript seems to fail hard when I import it from `math/vec3.js` - the entire interface stops working)

And in **interfaces.d.ts**:

```ts
interface Destroyable {
    someBoolean: boolean;
    someVector: pc.Vec3;
    destroy(): void;
    isDestroyed(): boolean;
}

```

willeastcott:
Closing this since the PlayCanvas codebase has now been (mostly) migrated to modern JavaScript. See PR #2641 for details. There's just too many conflicts to bother resolving.

When that effort is 100% complete, we can revisit the pros and cons of a switch to Typescript in a new PR.

PR #2583, opened 2020-12-01
Shader graph asset

raytranuk:
Dependent on PR: https://github.com/playcanvas/engine/pull/2500


src/resources/shader-graph.js
```diff
@@ -254,15 +239,12 @@ Object.assign(ShaderGraphHandler.prototype, {
                 }
                 // if there is a name, then it is a core node
                 for (i = 0; i < Object.keys(data.graphData.subGraphs).length; i++) {
```
Maksims, line 241: For loops second and third parts, are executed on every tick of a loop. Please, do cache `Object.keys(..)` in value before, as for large loops this can be costly.

willeastcott: Good tip. But this is now closed and we're not planning on merging it. 😄

PR #2586, opened 2020-12-02
[Fix] Layer.addMeshInstances can add duplicate

mvaligursky:
problem:
- adding meshInstance (MI) to a layer would only test an array it belongs to to see if it's already added.
- if material of the meshInstance has changed alpha state since last update (LayerComposition moving MIs between opaque and transparent arrays), this would miss MI in the other array and add duplicate MI to a Layer, causing issue down the line when it's removed, but still visible due to only single instance getting removed. And double rendering.

solution:
- check if mesh instance is in either arrays before adding it

Also did light refactor of Layer.removeMeshInstances - moved duplicate code to its own function.

### DIFF
--- a/src/scene/layer.js
+++ b/src/scene/layer.js
@@ -410,7 +410,7 @@ Layer.prototype.addMeshInstances = function (meshInstances, skipShadowCasters) {
 
         // test for meshInstance in both arrays, as material's alpha could have changed since LayerComposition's update to avoid duplicates
         // TODO - following uses of indexOf are expensive, to add 5000 meshInstances costs about 70ms on Mac. Consider using Set.
-        if (this.opaqueMeshInstances.indexOf(m) < 0 && this.transparentMeshInstances.indexOf(m)) {
+        if (this.opaqueMeshInstances.indexOf(m) < 0 && this.transparentMeshInstances.indexOf(m) < 0) {
             arr.push(m);
         }
 
--- a/src/scene/layer.js
+++ b/src/scene/layer.js
@@ -400,6 +400,7 @@ Layer.prototype.addMeshInstances = function (meshInstances, skipShadowCasters) {
     var casters = this.shadowCasters;
     for (var i = 0; i < meshInstances.length; i++) {
         m = meshInstances[i];
+
         mat = m.material;
         if (mat.blendType === BLEND_NONE) {
             arr = this.opaqueMeshInstances;
@@ -407,8 +408,12 @@ Layer.prototype.addMeshInstances = function (meshInstances, skipShadowCasters) {
             arr = this.transparentMeshInstances;
         }
 
+        // test for meshInstance in both arrays, as material's alpha could have changed since LayerCompositions to avoid duplicates
         // TODO - following uses of indexOf are expensive, to add 5000 meshInstances costs about 70ms on Mac. Consider using Set.
-        if (arr.indexOf(m) < 0) arr.push(m);
+        if (this.opaqueMeshInstances.indexOf(m) < 0 && this.transparentMeshInstances.indexOf(m)) {
+            arr.push(m);
+        }
+
         if (!skipShadowCasters && m.castShadow && casters.indexOf(m) < 0) casters.push(m);
 
         if (!this.passThrough && sceneShaderVer >= 0 && mat._shaderVersion !== sceneShaderVer) { // clear old shader if needed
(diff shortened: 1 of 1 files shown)


src/scene/layer.js
```diff
@@ -400,15 +400,20 @@ Layer.prototype.addMeshInstances = function (meshInstances, skipShadowCasters) {
...
+
         mat = m.material;
         if (mat.blendType === BLEND_NONE) {
             arr = this.opaqueMeshInstances;
         } else {
             arr = this.transparentMeshInstances;
         }
 
+        // test for meshInstance in both arrays, as material's alpha could have changed since LayerComposition's update to avoid duplicates
         // TODO - following uses of indexOf are expensive, to add 5000 meshInstances costs about 70ms on Mac. Consider using Set.
```
yaustar, line 412: As we are doing two indexofs, should we consider this TODO as we've now added to the cost?

src/scene/layer.js
```diff
@@ -400,15 +400,20 @@ Layer.prototype.addMeshInstances = function (meshInstances, skipShadowCasters) {
...
-        if (arr.indexOf(m) < 0) arr.push(m);
+        if (this.opaqueMeshInstances.indexOf(m) < 0 && this.transparentMeshInstances.indexOf(m)) {
```
slimbuck, line 413: `< 0` missing?

src/scene/layer.js
```diff
@@ -431,59 +463,28 @@ Layer.prototype.addMeshInstances = function (meshInstances, skipShadowCasters) {
...
-        // remove from shadows
-        if (skipShadowCasters) continue;
-        j = casters.indexOf(m);
-        if (j >= 0) casters.splice(j, 1);
+        // remove from casters
+        if (!skipShadowCasters) {
+            j = casters.indexOf(m);
+            if (j >= 0)
+                casters.splice(j, 1);
+        }
```
yaustar, line 485: Thanks! Much more readable!

src/scene/layer.js
mvaligursky: totally.
thanks!

mvaligursky: fixed

src/scene/layer.js
mvaligursky: that's separate PR, I don't want to get to it now.

src/scene/layer.js
```diff
@@ -400,15 +400,20 @@ Layer.prototype.addMeshInstances = function (meshInstances, skipShadowCasters) {
...
         if (mat.blendType === BLEND_NONE) {
             arr = this.opaqueMeshInstances;
         } else {
             arr = this.transparentMeshInstances;
         }
 
+        // test for meshInstance in both arrays, as material's alpha could have changed since LayerComposition's update to avoid duplicates
         // TODO - following uses of indexOf are expensive, to add 5000 meshInstances costs about 70ms on Mac. Consider using Set.
-        if (arr.indexOf(m) < 0) arr.push(m);
+        if (this.opaqueMeshInstances.indexOf(m) < 0 && this.transparentMeshInstances.indexOf(m) < 0) {
```
slimbuck, line 413: perhaps you could optimise this in the general case: test the opaque array first, then semitrans array for opaque materials and visa-versa for semitrans materials?

slimbuck (approved):
approving with comment

Maksims:
Would using Set here be a reasonable option?

LeXXik:
> Would using Set here be a reasonable option?

Set has a weak support on IE. Doing simple arrays check would be safer, if IE support is needed.

src/scene/layer.js
mvaligursky: there is && in the middle, so both need to execute, not || where only one branch executes

mvaligursky:
I've used Set in few PRs already - as long as you stick to functions supported in IE11, it works great.
Yes, Set would be a part of solution here - maybe we'd still need array, I'd need to see what else this array is used for.
Separate PR in the future. This was done for now to fix bug, not optimize.

src/scene/layer.js
slimbuck: `&&` is also short-circuited though? if `a` is false in `a && b`, then `b` does not need to be evaluated.

mvaligursky: true, the optimization could be done. But I'm not sure it's worth the code complexity - we'll do proper Set based optimization instead I hope. Most meshes are opaque, so we can the benefit in most cases already.

PR #2587, opened 2020-12-02
Max Asset Retries

vkalpias:
This PR changes the following:

* Change pc.ResourceLoader#enableRetry to take the maximum number of retries as an argument. 
* Change the retryRequests flags in the resource handlers with a maxRetries number. 
* Added maxAssetRetries project setting.

### DIFF
--- a/examples/assets/scripts/parsers/hdr-texture.js
+++ b/examples/assets/scripts/parsers/hdr-texture.js
@@ -46,8 +46,8 @@ Object.assign(Reader.prototype, {
  * @implements {pc.TextureParser}
  * @classdesc Texture parser for hdr files.
  */
-function HdrParser(registry, retryRequests) {
-    this.retryRequests = retryRequests;
+function HdrParser(registry, maxRetries) {
+    this.maxRetries = Math.max(maxRetries, 0) || 0;
 }
 
 Object.assign(HdrParser.prototype, {
@@ -55,7 +55,8 @@ Object.assign(HdrParser.prototype, {
         var options = {
             cache: true,
             responseType: "arraybuffer",
-            retry: this.retryRequests
+            retry: this.maxRetries > 0,
+            maxRetries: this.maxRetries
         };
         pc.http.get(url.load, options, callback);
     },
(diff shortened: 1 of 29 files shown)


examples/assets/scripts/parsers/hdr-texture.js
```diff
@@ -46,16 +46,17 @@ Object.assign(Reader.prototype, {
  * @implements {pc.TextureParser}
  * @classdesc Texture parser for hdr files.
  */
-function HdrParser(registry, retryRequests) {
-    this.retryRequests = retryRequests;
+function HdrParser(registry, maxRetries) {
+    this.maxRetries = Math.max(maxRetries, 0) || 0;
```
willeastcott, line 50: Shouldn't you set maxRetries as optional? And default it to 0?

vkalpias: It is optional and defaults to 0

src/resources/parser/texture/basis.js
```diff
@@ -11,16 +11,17 @@ import { basisTargetFormat, basisTranscode } from '../../basis.js';
  * @implements {pc.TextureParser}
  * @classdesc Parser for basis files.
  */
-function BasisParser(registry, retryRequests) {
-    this.retryRequests = !!retryRequests;
+function BasisParser(registry, maxRetries) {
+    this.maxRetries = Math.max(maxRetries, 0) || 0;
```
willeastcott, line 15: Again, make maxRetries optional.

examples/assets/scripts/parsers/hdr-texture.js
willeastcott: Yeah, actually, true. But man, that code is confusing. Because `Math.max(undefined, 0)` is `NaN` and `Nan || 0` is `0`.

vkalpias: Checking for undefined, NaN, negative values or other invalid values can become pretty verbose otherwise.. I preferred the shorter code which does it all :)

Maksims:
Does this PR enables auto retry by default? Would be good.

vkalpias:
It doesn't but the Editor will now expose a max retries setting. We should probably default that to 5 or so...

examples/assets/scripts/parsers/hdr-texture.js
willeastcott: Well, it's one line, but up to you. :)

willeastcott (approved):

PR #2590, opened 2020-12-03
Directional lightmap fix

### LINKED ISSUE #2545: Non-directional lightmap baking results in black scene
Directional lightmaps work in Editor projects, but non-directional seem to end up black.

Repro:
create new project. Set up light to bake lightmaps. Set up plane and cube meshes to receive lightmaps and static.
Start the project. WIth light set to directional lightmap, all works. WIth light set to non-directional lightmap, the scene is black.

slimbuck:
Fixes #2545

The issue was lack of specular for metal objects:
change

Also simplified the shader chunk by sampling textures twice.

### DIFF
--- a/src/graphics/program-lib/chunks/lightmapDir.frag
+++ b/src/graphics/program-lib/chunks/lightmapDir.frag
@@ -14,6 +14,7 @@ void addLightMap() {
         float nlight = (flight / max(vlight, 0.01)) * 0.5;
 
         dDiffuseLight += color * nlight * 2.0;
-        dSpecularLight += color * getLightSpecular();
     }
+
+    dSpecularLight += color * getLightSpecular();
 }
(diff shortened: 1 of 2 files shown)


mvaligursky:
The issue was with non-directional lightmaps, not with directional - is that fixed as well?

slimbuck:
Yes this fixes all light types. The issue was the 'Lightmap Direction' option. With that disabled specular was not being calculated.

slimbuck:
('Directional lightmap' is referring to the lightmap containing directions, not the directional light).

willeastcott (approved):

raytranuk:
Does this work in cases when a material is diffuse only, and `useSpecular` is false?

slimbuck:
Yeah, actually `useSpecular` is enforced when directional maps are involved, see https://github.com/playcanvas/engine/blob/master/src/graphics/program-lib/programs/standard.js#L446

raytranuk (approved):
Looks good to me - fixes the problem and faster.

slimbuck:
Also tested with https://launch.playcanvas.com/446278. Actually doesn't seem any faster on macos, but works fine.

mvaligursky (approved):

PR #2595, opened 2020-12-07
RenderComponent engine examples

mvaligursky:
- Multiple engine examples converted from using model component to render mesh instances, to using new render component, which is being made public here: https://github.com/playcanvas/engine/pull/2943
- some engine examples using model component were moved to deprecated folder to allow for continued testing, while not visible in example browser


examples/graphics/batching-dynamic.html
```diff
@@ -55,19 +55,22 @@
...
 
             var entity = new pc.Entity();
-            entity.addComponent("model", {
-                type: shapeName
-            });
 
-            // randomly assign a material
-            entity.model.material = Math.random() < 0.5 ? material1 : material2;
+            // randomly create render or model component
+            var componentType = Math.random() < 0.5 ? "render" : "model";
```
willeastcott, line 65: I know it makes a nice functional test to do this here, but the primary purpose of the examples is to show the correct way to use the API. I don't think anyone is going to do what's here in their own applications. I would stick to using render components instead of model components. And we should maybe have a 'deprecated' examples folder that contains usage of legacy APIs that is not included in the examples browser sidebar.

examples/examples.js
```diff
@@ -24,6 +24,7 @@ var categories = [
             "layers",
             "lights",
             "lights-baked",
+            "lights-baked-render",
```
willeastcott, line 27: I would just make this `lights-baked` and put the previous one in a `deprecated` folder.

examples/graphics/render-asset.html
```diff
@@ -0,0 +1,114 @@
...
+        app.assets.loadFromUrl("../assets/models/playcanvas-cube.glb", "container", function (err, asset) {
+
+            // get the instance of it set up with render component and add it to scene
+            cubeEntities[0] = asset.resource.instantiateRenderEntity();
+            cubeEntities[0].setLocalPosition(7, 12, 0);
+            cubeEntities[0].setLocalScale(3, 3, 3);
+            app.root.addChild(cubeEntities[0]);
+
+            // get another instance of it set up with render component and add it to scene
+            cubeEntities[1] = asset.resource.instantiateRenderEntity();
```
willeastcott, line 51: Is this faster/better than doing `cubeEntities[1] = cubeEntities[0].clone();`?

examples/loaders/loader-glb.html
```diff
@@ -35,25 +35,30 @@
...
         var url = "../assets/models/playcanvas-cube.glb";
         app.assets.loadFromUrl(url, "container", function (err, asset) {
             app.start();
 
-            var entity = new pc.Entity();
-            entity.addComponent("model", {
-                type: "asset",
-                asset: asset.resource.model,
+            // create instance using model component
+            var entity1 = asset.resource.instantiateModelEntity({
```
willeastcott, line 44: Again, I wouldn't confuse things. I would proceed with the render component at the recommended way here.

examples/graphics/batching-dynamic.html
mvaligursky: makes sense.

examples/graphics/render-asset.html
mvaligursky: It's matching the other case here .. performance should be similar.
But I've changed it to .clone() to test this path .. and that exposed an issue .. I'll have separate PR for this in a minute.

raytranuk (commented):
I'm mostly seeing "model" replaced by "render" - and sometimes things are a bit tidier because materials and other parameters are passed on render component creation. A question: I'm aware that underlyingly its a new component - but are these the only differences to the user?

mvaligursky:
On the API level differences are not large - as you said, mostly simpler code. The actual implementation is what makes the difference:
- instead of a single model using a single Entity, render components distribute meshes to nodes they're attached on in Maya and similar. So there is no hidden hierarchy of nodes that components cannot be attached to, it's a hierarchy of entities and each entity can have a render component with some parts of the mesh.
- no extra nodes per Model - in the hierarchy engine example test this halves the number of nodes and doubles the performance of their update
- as all this is entity based, you can load a model into the Editor and see hierarchy inside and meshes attached where they should be, including child lights / cameras and similar - you can individually control which parts of that mesh is visible or have different shadow flags and similar

examples/graphics/model-asset.html
```diff
@@ -45,12 +45,11 @@
         app.assets.loadFromUrl(url, "container", function (err, asset) {
             app.start();
 
-            entity = new pc.Entity();
-            entity.addComponent("model", {
-                type: "asset",
-                asset: asset.resource.model,
+            // create an entity with model assets
+            var entity = asset.resource.instantiateModelEntity({
```
willeastcott, line 49: Presumably this should be `instantiateRenderEntity`?

willeastcott (approved):

examples/graphics/model-asset.html
mvaligursky: yes, that was the idea, thanks! :)

Issue #2409, opened 2020-09-06
pc.MeshInstance.castShadow API docs not clear


yaustar:
https://github.com/playcanvas/engine/blob/master/src/scene/mesh-instance.js#L30

It doesn't really control whether or not it casts shadows as changing the property without updating the shadowcasters on the layers won't do anything.

We should update the docs to make it more clear how this property is used in that respect.

As the model component controls this logic, maybe the property shouldn't be public anymore?

mvaligursky:
Good suggestion @yaustar , I've done just that.

willeastcott:
I should point out that the engine was originally designed such that you could potentially build an app using just the APIs in `src/graphics` or `src/graphics` + `src/scene` or `src/graphics` + `src/scene` + `src/framework`. I believe this layered design is useful in keeping the engine modular - and avoiding circular dependencies in the code.

It seems that it may be the case that when the layer system was introduced in 2017, the `pc.MeshInstance#castShadow` flag stopped working as expected. Before that time, you could enabled shadow casting for a whole model, say, and then disable casting for some specific mesh instance in the model.

So sure, you could just make the property private now. And I won't strongly fight against that. But ideally, we'd just fix the flag to work as advertised IMHO.

PR #2612, opened 2020-12-14
MeshInstance. castShadow is no longer public

mvaligursky:
- as it does not work in isolation. The public interface for this is on Model/Render component.

Fixes: https://github.com/playcanvas/engine/issues/2409


willeastcott (approved):

mvaligursky:
I'd go ahead with this after looking at it some more. The flag itself (meshInstance.castShadow) is there for compatibility reasons, but does not work by itself. On the graphics level API, one needs to use Layer.add(remove)ShadowCasters to get this functionality. MeshInstance does not store layers its with new Layer system, and so cannot do this behind the scene.

willeastcott:
OK, let's merge if the property simply doesn't work.

willeastcott (approved):

PR #2616, opened 2020-12-15
Separated batching.js into file per class it contained

mvaligursky:
no other changes


src/scene/batching/batch.js
```diff
@@ -0,0 +1,26 @@
...
+ * @name pc.Batch
+ * @classdesc Holds information about batched mesh instances. Created in {@link pc.BatchManager#create}.
+ * @param {pc.MeshInstance[]} meshInstances - The mesh instances to be batched.
+ * @param {boolean} dynamic - Whether this batch is dynamic (supports transforming mesh instances at runtime).
+ * @param {number} batchGroupId - Link this batch to a specific batch group. This is done automatically with default batches.
+ * @property {pc.MeshInstance[]} origMeshInstances An array of original mesh instances, from which this batch was generated.
+ * @property {pc.MeshInstance} meshInstance A single combined mesh instance, the result of batching.
+ * @property {pc.Model} model A handy model object.
+ * @property {boolean} dynamic Whether this batch is dynamic (supports transforming mesh instances at runtime).
+ * @property {number} [batchGroupId] Link this batch to a specific batch group. This is done automatically with default batches.
```
willeastcott, line 14: I know you've just copy/pasted the code and not modified this....but it looks weird that `batchGroupId` is marked here as optional. Is that a doc bug?

mvaligursky: We don't use it without this parameter in the engine, but a user can create a Batch without id to batch some meshInstances and then destroy the group - without having the have an Id for it registered with the engine.

willeastcott (approved):

PR #2618, opened 2020-12-15
Fixed bug where new sound instances aren't kept

yaustar:
When `sound.positional` is set by the developer, the new instances created are no longer kept in slot.instances. This means that the sound is audible but the slot.isPlaying property is false because it has no instances and the developer can no longer stop the sound.

Reproduced in: https://playcanvas.com/project/747427/overview/sound-positional-issues

Open up dev console in the browser

Press 1 to play the sound, 4 to toggle positional property.

Notice that isPlaying is false in the console (press 3) and that pressing 2 to stop the sound doesn't stop the sound.

### DIFF
--- a/src/framework/components/sound/component.js
+++ b/src/framework/components/sound/component.js
@@ -120,19 +120,17 @@ Object.defineProperty(SoundComponent.prototype, "positional", {
             var slot = slots[key];
             // recreate non overlapping sounds
             if (!slot.overlap) {
-                var oldInstances = slot.instances;
-                var newInstances = [];
+                var instances = slot.instances;
+                var oldLength = instances.length;
 
                 // When the instance is stopped, it gets removed from the slot.instances array
                 // so we are going backwards to compenstate for that
 
-                // We are also going to create a new array to hold the new instances created
-                // and assign to slot at the end
-                for (var i = oldInstances.length - 1; i >= 0; i--) {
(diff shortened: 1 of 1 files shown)


vkalpias:
Let me know if I'm wrong but it seems to me that the old code should work if you just reverse the loop? Which might avoid creating a new array? Although it probably will still splice instances from the old array and then resize it to the old size again so not sure if it's better.

yaustar:
> Let me know if I'm wrong but it seems to me that the old code should work if you just reverse the loop? Which might avoid creating a new array? Although it probably will still splice instances from the old array and then resize it to the old size again so not sure if it's better.

I can give it a try but I think we get the same issue as on stop, the instance is spliced out of the array. The test above is with one sound instance so reversing and going forward would be the same thing in this case :)

yaustar:
Oh, I did think about pushing the new instances to the end of the array and doing something to ensure that only the old ones are removed but it didn't seem as clear for something that is probably done rarely for the sake of not allocating a new array.

yaustar:
@vkalpias I've got an idea on how to avoid creating a new array if you would like me to implement that instead? It will make the logic a bit harder to follow though 😅

yaustar:
I've updated the PR to avoid the array allocation as Vaios mentioned above.

vkalpias (approved):

PR #2621, opened 2020-12-16
[FIX] Engine and mini-stats handle context lost and restored correctly

mvaligursky:


willeastcott (approved):

willeastcott:
This is a nicely crafted PR - love the general rearchitecting/tidy up of how this works. Much more elegant. 🚀

PR #2623, opened 2020-12-16
Scene data caching API 

yaustar:
Edit Thu 18 Feb 2021: 

Add API `loadSceneData` and `unloadSceneData` to the Scene Registry to allow uses to store/cache scene data in the registry. This reduces the number of HTTP calls to the scene JSON URL if a scene needs to be loaded multiple times and can also be done ahead of the actual scene load to prevent waiting for a network request.

`loadSceneHierarchy` and `loadSceneSettings` do not store/cache scene data to maintain existing expected behaviour. A project may have lots of scenes that are only loaded once and would not want all that data stored.

TemplateUtils has been removed after talking with @zachppaul and @vkalpias as it was not being used. I've tested against some template heavy projects that I have access to.

The resource handlers for scene settings, hierarchy and scene have been refactored as the load function was identical after TemplateUtils was removed. They should be refactored more as they are very similar but as the Editor launch.js adds it's own handlers and uses the open function of the engine ones, I left that alone for this PR so there isn't a dependency of releasing both.

**Publishing will also need to be tested due to the difference of use of handlers in launch.js and in a published release. I've tested against locally hosted builds.**

----
An attempt at improving the scene loading code to allow for 'preloading' of data and cache the scene data so that multiple scene loads won't keep calling network requests for the scene JSON.

This would allow developers to preload the scene data so that it can created into the hierarchy without waiting for a network request. It also means that loading scene settings and scene hierarchy can be done in one network request and negates the need for loadScene which never should be used except in one specific use case.

There is a performance hit if the user is just loading scene settings and nothing else compared to the previous version as hierarchy handler also prepares templates.
I think this would be okay as it's rare for developers to just load scene settings.

Existing API is preserved.

Example usages:

Loading scene with settings
```
    var self = this;
    var oldHierarchy = this.app.root.findByName ('Root');
    var scene = this.app.scenes.find(sceneName);

    this.app.scenes.loadSceneHierarchy(scene, function(err, parent) {
        if (!err) {
            oldHierarchy.destroy();
            self.app.scenes.loadSceneSettings(scene, function(err) {
                if (err) {
                    console.error(err);
                }
            });
        } else {
            console.error(err);
        }
    });
```

Preloading scene data and load scene hierarchy
```
        var scene = this.app.scenes.find('Scene 1');
        this.app.scenes.loadSceneData(scene, function (err, sceneItem) {
            if (err) {
                console.log(err);
            }
        });

       // later
      this.app.scenes.loadSceneHierarchy(scene, function(err, parent) {
        if (!err) {
            oldHierarchy.destroy();
        } else {
            console.error(err);
        }
    });
      
```


yaustar:
TODO: Write tests

src/framework/scene-registry.js
```diff
@@ -126,44 +142,127 @@ SceneRegistry.prototype.remove = function (name) {
...
+    // Because we need to load scripts before we instance the hierarchy (i.e. before we create script components)
+    // Split loading into load and open
+    var handler = this._app.loader.getHandler("hierarchy");
+
+    // include asset prefix if present
+    if (this._app.assets && this._app.assets.prefix && !ABSOLUTE_URL.test(url)) {
+        url = path.join(this._app.assets.prefix, url);
+    }
+
+    if (sceneItem._loading) {
```
yaustar, line 189: I would like to change this into a subscriber model, similar to assets.

yaustar:
I'm also considering refactoring the hierarchy, scene and scene settings handler as they share logic. Maybe making a single handler? 

image

marklundin:
Just to add to this; there is a usecase that I've been struggling to implement correctly which I think would benefit from rethinking the scene loading. 

I have two scenes 'intro' and 'main'. All the assets used in the 'main' scene are tagged  'main', and all the intro scene assets are tagged 'intro'. I am then able to filter out an array of assets used for each scene and control/prioritise loading of assets. In my ase, I want to start preloading all the assets from the smaller  'intro' scene, then immediately (before intro has loaded) start preloading the asset used in 'main'.  The idea being that once the assets are available from 'intro' I'm able to display that to the user whilst still waiting for the main scene to load in the background. As soon as the assets from the 'main' scene are available, I want to swap the scenes.

Currently this is sort of working, however in order to play the intro Scene, n addition to loading the assets, you have to load the [scene].json file. So although it's assets are already available, it's json scene file isn't, and you therefore have to wait for it to load before you can start the scene. This is ok in a general case, but in my scenario, there is already a stack of network requests to load the assets from the 'main' scene; and because the browser caps the number of concurrent requests per domain, it has to wait for all of the 'main' scene assets to finish load before it can load the intro scene json file and play the intro scene. This obviously defeats the point of preloading the intro assets.

Ideally the scene definition and app config could be loaded from local data as this would give more control over custom load requirements. If you had an in-memory scene definition and asset registry, it should just be able to run it. I'm sure this is probably possible in the current system, but it would be great to have a clean API

yaustar:
Thanks for the feedback Mark! 

> Currently this is sort of working, however in order to play the intro Scene, you have to load the [scene].json file aswell as the assets. So although it's assets are already available in memory, it's json scene file isn't,

Yes, this was the main point that I wanted to tackle with this PR. Every scene load is a new http request regardless whether it has been loaded before or not. It also exposes the ability to preload the scene data without creating the scene entities themselves. This allows the developer to preload all scenes and make scene switching quicker down the line in the lifetime of the session.

In theory, you can do:
```
this.app.scenes.loadSceneData(scene, function (err, sceneItem) {
     if (err) {
         // load scene assets
         self.loadAssets(assets, function(err) {
               self.app.scenes.loadSceneHierarchy(/* */);
         });
     }
});
```
Not sure if that solves your current problem? Were you thinking more of being able to have scenes preloaded?

> Also it's not totally clear when to use loadScene vs loadSceneHierachy.
loadScene should never be used IMO unless you are an engine user and you are loading the very first scene. It was exposed to public API just for this reason and we are a bit stuck on not being able to change the function name :( 

loadScene loads both the hierarchy and the settings with a single HTTP request but doesn't initialise or postInitialise components. 

> Ideally the scene definition and app config could be loaded from local data as this would give more control over custom load requirements.

Not sure what you mean by this, can you elaborate?

marklundin:
Thanks for this @yaustar. I'm going to test some of this in one of out projects. Am I correct in thinking that given two scenes, the following would work

```javascript
 var scene = this.app.scenes.find('Scene 1');

//Preloads both scenes without loading anything
this.app.scenes.loadSceneData(Scene1)
this.app.scenes.loadSceneData(Scene2)

// At some point later, given that both do not fail
this.app.scenes.loadSceneHierarchy(Scene1)

```

yaustar:
> Am I correct in thinking that given two scenes, the following would work, in that it would load and run both scenes

Yes.

> One question; how does this relate to assets tagged with the preload flag? When are these loaded as these are application wide, and not scene specific

Not quite sure what you mean. The changes here don't affect preloaded assets AFAIK.

marklundin:
Ahh it's the preload() that does this I think

yaustar:
Updated first post :)

vkalpias (approved):

vkalpias (approved):

PR #2624, opened 2020-12-17
[FIX] Fix to primitive caching to handle multi device/application scenario

mvaligursky:
fixes non-public issue introduced by refactoring for RenderComponent

### DIFF
--- a/src/scene/procedural.js
+++ b/src/scene/procedural.js
@@ -36,7 +36,7 @@ var primitiveUv1Padding = 4.0 / 64;
 var primitiveUv1PaddingScale = 1.0 - primitiveUv1Padding * 2;
 
 // cached mesh primitives
-var shapePrimitiveMap = new Map();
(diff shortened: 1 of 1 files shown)


src/scene/procedural.js
```diff
@@ -1063,7 +1063,15 @@ function createBox(device, opts) {
 // returns Primitive data, used by ModelComponent and RenderComponent
 function getShapePrimitive(device, type) {
 
-    var primData = shapePrimitiveMap.get(type);
+    // find in cache
+    var primData = null;
+    for (var i = 0; i < shapePrimitives.length; i++) {
+        if (shapePrimitives[i].type === type && shapePrimitives[i].device === device) {
```
willeastcott, line 1069: LOL. So the more apps on the page, the slower each app gets. 😆  But I'm fine going with this to get the deployment out.

willeastcott (approved):

PR #2634, opened 2020-12-22
[FIX] 'attr' event does not use copy of old value

vkalpias:
When firing an 'attr' event from a script, use a copy of the 'old' value e.g. for vectors and colors

Fixes https://github.com/playcanvas/editor/issues/222

### DIFF
--- a/src/script/script-attributes.js
+++ b/src/script/script-attributes.js
@@ -254,6 +254,13 @@ ScriptAttributes.prototype.add = function (name, args) {
         },
         set: function (raw) {
             var old = this.__attributes[name];
+            // keep copy of old for the event below
+            var oldCopy = old;
+            // json types might have a 'clone' field in their
+            // schema so make sure it's not that
+            if (old && args.type !== 'json' && old.clone) {
+                oldCopy = old.clone();
+            }
 
             // convert to appropriate type
             if (args.array) {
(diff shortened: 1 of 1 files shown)


mvaligursky (approved):

Maksims:
It should try to avoid allocations, by pre-allocating types of value, and only copying of value in it. It is ok if old value lifetime is guaranteed only during a callback.

willeastcott:
I totally get your point, @Maksims, but if this is just behavior we document and expect developers to read it, we might be disappointed. I bet some developers will store off the value and use it again later. 😞

Maksims:
> I totally get your point, @Maksims, but if this is just behavior we document and expect developers to read it, we might be disappointed. I bet some developers will store off the value and use it again later. 😞

It's true. Many functions in our engine already follow such path of not allowing to reuse or modify things. Like getPosition returns vector that should not be modified by developer and is subject to change later.
WebXR APIs (actual browser APIs) follow similar rules as well, where things are only actual during sync call to them, but not outside of tick or callback.

It is a question of sacrificing performance for everyone, or ask some to follow the rules if they don't.

willeastcott (approved):

willeastcott:
I can totally see Max's point here. This would have been a major problem is `Vec2/3/4` and `Color` classes still allocated a `Float32Array`. But they're very lightweight constructors now. Let's just merge for now and if it because a problem with GC, we can revisit. I'm just glad the bug is fixed.

Maksims:
Ok, if we going to clone stuff, then can we first check if there is any event subscribtion so that we don't allocate if there is not event to be fired.

willeastcott:
Good point. What do you think @vkalpias?

vkalpias:
> Good point. What do you think @vkalpias?

https://github.com/playcanvas/engine/pull/2677

Issue #2633, opened 2020-12-21
Investigate performance of ForwardRenderer.updateLitShaders


mvaligursky:
This reproduces in test scene FarmLP (https://sketchfab.com/3d-models/farm-lp-9f114cea5d604d1982f98ec646cffb5e) but probably any other with lots of meshInstances. In testing this was imported as Glb-template. Turning any light on / off in Editor takes about a second.

What happens: 
- in updateLitShaders we loop over all meshInstances, and for each material clearVariants is called
- material loops over all meshInstance that use the material (which is most of the scene again), and clears stuff up

so from O(n) we get to O(n * n)

Ideally the inner loop over meshInstances can be early skipped. Maybe mat.updateShader should be updated inside updateLitShaders?

![Screen Shot 2020-12-21 at 4 27 00 PM](https://user-images.githubusercontent.com/59932779/102798711-5f292c80-43a9-11eb-9659-30f8ad7e7433.png)

PR #2638, opened 2020-12-23
Optimized updateShaders in forward-renderer

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/2633

### DIFF
--- a/src/scene/forward-renderer.js
+++ b/src/scene/forward-renderer.js
@@ -2312,7 +2312,7 @@ Object.assign(ForwardRenderer.prototype, {
 
         // Clear material shaders
         this.getUniqueMaterials(_tempMaterialSet, drawCalls);
-        for (let mat of _tempMaterialSet) {
+        for (var mat of _tempMaterialSet) {
             if (mat.updateShader !== Material.prototype.updateShader) {
                 mat.clearVariants();
                 mat.shader = null;
@@ -2327,7 +2327,7 @@ Object.assign(ForwardRenderer.prototype, {
 
         // Clear material shaders
         this.getUniqueMaterials(_tempMaterialSet, drawCalls);
-        for (let mat of _tempMaterialSet) {
+        for (var mat of _tempMaterialSet) {
             if (mat.updateShader !== Material.prototype.updateShader) {
 
                 // skip unlit standard and particles materials
--- a/src/scene/forward-renderer.js
+++ b/src/scene/forward-renderer.js
@@ -2293,7 +2293,7 @@ Object.assign(ForwardRenderer.prototype, {
         // #endif
     },
 
-    // funtion returns unique set of materials from drawCalls
+    // function returns unique set of materials from drawCalls
     // material parameter is expected to be a Set to allow for fast duplicates removal
     getUniqueMaterials: function (materials, drawCalls) {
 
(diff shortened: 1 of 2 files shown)


src/scene/forward-renderer.js
```diff
@@ -2290,59 +2293,55 @@ Object.assign(ForwardRenderer.prototype, {
         // #endif
     },
 
-    updateShaders: function (drawCalls) {
-        // #ifdef PROFILER
-        var time = now();
-        // #endif
+    // funtion returns unique set of materials from drawCalls
```
willeastcott, line 2296:
```suggestion
    // function returns unique set of materials from drawCalls
```

src/scene/scene.js
```diff
@@ -140,7 +140,7 @@ function Scene() {
         lastStaticPrepareWriteTime: 0,
         lastStaticPrepareTriAabbTime: 0,
         lastStaticPrepareCombineTime: 0,
-        updateShadersTime: 0
+        updateShadersTime: 0 // deprecated
```
willeastcott, line 143: Just curious, but why is this now deprecated?

mvaligursky: it was only used around the functions I optimized, which take no time now .. so no need to report duration

src/scene/forward-renderer.js
```diff
@@ -2290,59 +2293,55 @@ Object.assign(ForwardRenderer.prototype, {
...
+        return materials;
+    },
+
+    updateShaders: function (drawCalls) {
+
         // Clear material shaders
-        for (i = 0; i < materials.length; i++) {
-            var mat = materials[i];
+        this.getUniqueMaterials(_tempMaterialSet, drawCalls);
+        for (var mat of _tempMaterialSet) {
```
Maksims, line 2315: There are two loops here with extra function call. Why not iterate once through draw calls with track of updated materials using a set (as is now), so eliminating for second iteration through set of materials.

src/scene/forward-renderer.js
```diff
@@ -2290,59 +2293,55 @@ Object.assign(ForwardRenderer.prototype, {
...
-                if (mat.updateShader !== Material.prototype.updateShader) {
-                    if (mat.useLighting === false || (mat.emitter && !mat.emitter.lighting)) {
-                        // skip unlit standard and particles materials
-                        continue;
-                    }
-                    mat.clearVariants();
-                    mat.shader = null;
+        // Clear material shaders
+        this.getUniqueMaterials(_tempMaterialSet, drawCalls);
+        for (var mat of _tempMaterialSet) {
```
Maksims, line 2330: Same as above comment, regarding two loops.

src/scene/forward-renderer.js
mvaligursky: fixed, thanks

src/scene/forward-renderer.js
mvaligursky: a fine suggestion. I've done that now.

src/scene/forward-renderer.js
mvaligursky: same as above

src/scene/scene.js
willeastcott: OK, fair enough. 🚀

willeastcott (approved):

PR #2640, opened 2020-12-25
Fix API docs, as markup links in properties are not handled well

Maksims:


willeastcott:
In what way are they not handled well?

yaustar:
Odd, I don't remember seeing this when I added that link originally?

image

Maksims:
> In what way are they not handled well?

Check html generated for session property here: https://developer.playcanvas.com/en/api/pc.XrManager.html

And notice this column.

yaustar (approved):

PR #2641, opened 2020-12-26
Transition to modern JavaScript

willeastcott:
Inspired by [this talk](https://www.youtube.com/watch?v=cLxNdLK--yI&feature=youtu.be) from Chrome Dev Summit.

This PR migrates a portion of the PlayCanvas codebase from strict ES5 to support the very latest JS language features. Initially, the code has been updated to leverage:

* Classes
* Static class properties
* Default parameters
* `const` (sparingly for now)

Code updated so far is:

* `bundle`
* Most of `core`
* `math`
* `resources`
* `shape`
* `template`
* `extras` (MiniStats)

This is all made possible by adding Babel to the engine's rollup build process.

The rollup configuration has also been updated to generate `playcanvas.mjs`, an ES Module build of the engine. So now we have:

* Raw codebase: ES2021
* `playcanvas.mjs`: Transpiled to ES6 (ECMAScript 2015)
* `playcanvas.js`: Transpiled to ES5 (ECMAScript 2009)

This PR follows up on the [previous unmerged PR](https://github.com/playcanvas/engine/pull/2577) that migrated the engine codebase to Typescript. There was much debate about the overall benefit of adopting Typescript. Typescript provides two main benefits:

1. static type checking
2. ability to adopt the latest JS language features

But we can achieve 2 with Babel. Therefore, it is probably easier to migrate to TS-like JS (ES6 classes in particular), and then revisit the benefits of Typescript. *If* Typescript is adopted at some point, this will be a stepping stone almost directly in that direction.

Let's look at output in terms of `pc.Vec2` for example. Previously, it was:

```javascript
	function Vec2(x, y) {
		if (x && x.length === 2) {
			this.x = x[0];
			this.y = x[1];
		} else {
			this.x = x || 0;
			this.y = y || 0;
		}
	}
	Object.assign(Vec2.prototype, {
		add: function (rhs) {
			this.x += rhs.x;
			this.y += rhs.y;
			return this;
		},
```

Now it's:

```javascript
	var Vec2 = function () {
		function Vec2(x, y) {
			if (x && x.length === 2) {
				this.x = x[0];
				this.y = x[1];
			} else {
				this.x = x || 0;
				this.y = y || 0;
			}
		}

		var _proto = Vec2.prototype;

		_proto.add = function add(rhs) {
			this.x += rhs.x;
			this.y += rhs.y;
			return this;
		};
```

So essentially, a closure is introduced and `Object.assign` is removed. Still just as easy to debug. 

Feedback most definitely welcome!

Closes #2642

### DIFF
--- a/src/core/color.js
+++ b/src/core/color.js
@@ -16,7 +16,7 @@ import { math } from '../math/math.js';
  */
 class Color {
     constructor(r = 0, g = 0, b = 0, a = 1) {
-        var length = r && r.length;
+        var length = r.length;
         if (length === 3 || length === 4) {
             this.r = r[0];
             this.g = r[1];
--- a/src/core/color.js
+++ b/src/core/color.js
@@ -88,11 +88,11 @@ class Color {
      * @param {number} [a] - The value for the alpha (0-1), defaults to 1.
      * @returns {pc.Color} Self for chaining.
      */
-    set(r, g, b, a) {
+    set(r, g, b, a = 1) {
         this.r = r;
         this.g = g;
         this.b = b;
-        this.a = (a === undefined) ? 1 : a;
+        this.a = a;
 
         return this;
     }
--- a/src/core/color.js
+++ b/src/core/color.js
@@ -14,31 +14,31 @@ import { math } from '../math/math.js';
  * @property {number} b The blue component of the color.
  * @property {number} a The alpha component of the color.
  */
-function Color(r, g, b, a) {
-    var length = r && r.length;
-    if (length === 3 || length === 4) {
-        this.r = r[0];
-        this.g = r[1];
-        this.b = r[2];
-        this.a = r[3] !== undefined ? r[3] : 1;
-    } else {
-        this.r = r || 0;
-        this.g = g || 0;
-        this.b = b || 0;
-        this.a = a !== undefined ? a : 1;
+class Color {
+    constructor(r = 0, g = 0, b = 0, a = 1) {
+        var length = r && r.length;
+        if (length === 3 || length === 4) {
+            this.r = r[0];
+            this.g = r[1];
+            this.b = r[2];
+            this.a = r[3] !== undefined ? r[3] : 1;
+        } else {
+            this.r = r;
+            this.g = g;
+            this.b = b;
+            this.a = a;
+        }
     }
-}
 
-Object.assign(Color.prototype, {
     /**
      * @function
      * @name pc.Color#clone
      * @description Returns a clone of the specified color.
      * @returns {pc.Color} A duplicate color object.
      */
-    clone: function () {
+    clone() {
         return new Color(this.r, this.g, this.b, this.a);
-    },
+    }
 
     /**
      * @function
@@ -54,14 +54,14 @@ Object.assign(Color.prototype, {
      *
      * console.log("The two colors are " + (dst.equals(src) ? "equal" : "different"));
      */
-    copy: function (rhs) {
+    copy(rhs) {
         this.r = rhs.r;
         this.g = rhs.g;
         this.b = rhs.b;
         this.a = rhs.a;
 
         return this;
-    },
+    }
 
     /**
      * @function
@@ -74,9 +74,9 @@ Object.assign(Color.prototype, {
      * var b = new pc.Color(1, 1, 0, 1);
      * console.log("The two colors are " + (a.equals(b) ? "equal" : "different"));
      */
-    equals: function (rhs) {
+    equals(rhs) {
         return this.r === rhs.r && this.g === rhs.g && this.b === rhs.b && this.a === rhs.a;
-    },
+    }
 
     /**
      * @function
@@ -88,15 +88,14 @@ Object.assign(Color.prototype, {
      * @param {number} [a] - The value for the alpha (0-1), defaults to 1.
      * @returns {pc.Color} Self for chaining.
      */
-    set: function (r, g, b, a) {
+    set(r, g, b, a) {
         this.r = r;
         this.g = g;
         this.b = b;
         this.a = (a === undefined) ? 1 : a;
 
         return this;
-    },
-
+    }
 
     /**
      * @function
@@ -117,14 +116,14 @@ Object.assign(Color.prototype, {
      * r.lerp(a, b, 0.5); // r is 0.5, 0.5, 0.25
      * r.lerp(a, b, 1);   // r is equal to b
      */
-    lerp: function (lhs, rhs, alpha) {
+    lerp(lhs, rhs, alpha) {
         this.r = lhs.r + alpha * (rhs.r - lhs.r);
         this.g = lhs.g + alpha * (rhs.g - lhs.g);
         this.b = lhs.b + alpha * (rhs.b - lhs.b);
         this.a = lhs.a + alpha * (rhs.a - lhs.a);
 
         return this;
-    },
+    }
 
     /**
      * @function
@@ -134,7 +133,7 @@ Object.assign(Color.prototype, {
      * This is the same format used in HTML/CSS.
      * @returns {pc.Color} Self for chaining.
      */
-    fromString: function (hex) {
+    fromString(hex) {
         var i = parseInt(hex.replace('#', '0x'), 16);
         var bytes;
         if (hex.length > 7) {
@@ -147,7 +146,7 @@ Object.assign(Color.prototype, {
         this.set(bytes[0] / 255, bytes[1] / 255, bytes[2] / 255, bytes[3] / 255);
 
         return this;
-    },
+    }
 
     /**
      * @function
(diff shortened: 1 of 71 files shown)


Maksims:
> * `playcanvas.js`: Transpiled to ES2015

I think this is mistake. As ES2015 (should be written as `ECMAScript 2015`), is ES6. And IE11 does not support ES6. It supports only ES5 (ECMA2009).

Maksims:
I've tested more of Babel and Closure, what it outputs, and for ES5 both seem to do a good job, but for ES6, they start to differ.

# Example 1:

Source:
```js
class Ray {
    constructor(origin = new Vec3(0, 0, 0), direction = new Vec3(0, 0, -1)) {
        this.origin = origin;
        this.direction = Object.values(direction);
    }
}
```

Closure Compiler:
```js
class Qd {
    constructor(a=new z(0,0,0), b=new z(0,0,-1)) {
        this.origin = a;
        this.direction = Object.values(b)
    }
}
```

Babel + Terser:
```js
class Ray {
    constructor() {
        var e = arguments.length > 0 && void 0 !== arguments[0] ? arguments[0] : new Vec3(0,0,0)
          , t = arguments.length > 1 && void 0 !== arguments[1] ? arguments[1] : new Vec3(0,0,-1);
        this.origin = e,
        this.direction = Object.values(t)
    }
}
```

### Issues:
1. default parameters - are converted to ES5 format
2. use of `arguments` leads to unoptimized (unable to compile to machine code) code in most JS engines.
3. it seem to not mangling classes like Closure does, which leads to higher file size

# Example 2:
Source:
```js
containsPoint(point) {
    for (const plane of this.planes) {
        if (plane[0] * point.x + plane[1] * point.y + plane[2] * point.z + plane[3] <= 0) {
            return false;
        }
    }
    return true;
}
```

Closure Compiler:
```js
containsPoint(a) {
    for (const b of this.planes)
        if (0 >= b[0] * a.x + b[1] * a.y + b[2] * a.z + b[3])
            return !1;
    return !0
}
```

Babel + Terser:
```js
containsPoint(e) {
    for (var t of this.planes)
        if (t[0] * e.x + t[1] * e.y + t[2] * e.z + t[3] <= 0)
            return !1;
    return !0
}
```

### Issues:
1. It seem to not like let/const.

# File sizes:
Right now, I've tested only on engine with small part moved to ES6+ code, but it already produces different file sizes:
* Raw ES6+ + gzip: 387Kb
* Closure Compiler ES5 + gzip: 297Kb (76.7%)
* Closure Compiler ES6 + gzip: 294Kb (76.0%)
* Babel + Terser ES5 + gzip: 287Kb (74.1%)
* Babel + Terser ES6 + gzip: 303Kb (78.3%)

Size difference between them is insignificant. ES5 Babel + Terser is smaller than Closure Compiler ES5, for around ~2.5%. While ES6 version of Babel + Terser is bigger than Closure Compiler ES6 by around ~2.3%.

Size wise, both paths produce good results, but I assume that Closure Compiler will lead for ES6, as Babael + Terser sometimes transpiles valid ES6 code to ES5.

<ins>My only concern, is why Babel + Terser is transpiling some of valid ES6 features into ES5, when instructed to use ES6 target.</ins>

willeastcott:
@Maksims Ha, yes, sorry, I get confused with all the JavaScript language naming. You're right - I have corrected my original post.

> Issues:
> default parameters - are converted to ES5 format

Fixed with [this commit](https://github.com/playcanvas/engine/pull/2641/commits/4fbf4a43198469673fbd3b702701385ebbd455dc).

> use of arguments leads to unoptimized (unable to compile to machine code) code in most JS engines.

I'm not seeing this. With the commit above, I'm getting the following (after beautifying):

```javascript
class Ray {
    constructor(e = new Vec3, t = new Vec3(0, 0, -1)) {
        this.origin = e, this.direction = t
    }
    set(e, t) {
        return this.origin.copy(e), this.direction.copy(t), this
    }
}
```

> it seem to not mangling classes like Closure does, which leads to higher file size

Yeah, I noticed that. I'm not sure why but I don't think it's making a significant difference.

willeastcott:
Oh, that the `let` + `const` thing is fixed also by that commit.

src/core/color.js
```diff
@@ -88,15 +88,14 @@ Object.assign(Color.prototype, {
      * @param {number} [a] - The value for the alpha (0-1), defaults to 1.
      * @returns {pc.Color} Self for chaining.
      */
-    set: function (r, g, b, a) {
+    set(r, g, b, a) {
         this.r = r;
         this.g = g;
         this.b = b;
         this.a = (a === undefined) ? 1 : a;
```
mvaligursky, line 95: use default parameter value?

src/math/mat4.js
```diff
@@ -4,6 +4,11 @@ import { Vec3 } from './vec3.js';
 import { Vec4 } from './vec4.js';
 
 var _halfSize = new Vec2();
+var x = new Vec3();
```
mvaligursky, line 7: just curious if these could be inside class now as static or similar?

mvaligursky: also, will these variables with names "x" "y" "z" create possible conflicts in the rest of the class?

src/math/vec2.js
```diff
@@ -8,17 +8,45 @@
...
+ *
+ * // Get y
+ * var y = vec.y;
+ *
+ * // Set y
+ * vec.y = 0;
+ */
+class Vec2 {
+    constructor(x = 0, y = 0) {
+        if (x && x.length === 2) {
```
mvaligursky, line 41: can be changed to 
`if (x.length === 2) {`

src/math/mat4.js
willeastcott: We could do. I think I'd probably put them on the `Vec3` class as `Vec3._temp1`, `Vec3._temp2`, `Vec3._temp3` maybe. But I personally don't have a big problem with these global to the module. It's kind of nice because the app code cannot access them (and therefore misuse them), which isn't true of they're on the class.

src/math/vec3.js
```diff
@@ -9,19 +9,58 @@
...
+ *
+ * // Get z
+ * var z = vec.z;
+ *
+ * // Set z
+ * vec.z = 0;
+ */
+class Vec3 {
+    constructor(x = 0, y = 0, z = 0) {
+        if (x && x.length === 3) {
```
mvaligursky, line 53: can be
`if (x.length === 3) {`

src/core/color.js
willeastcott: Well spotted. Fixed.

src/math/vec4.js
```diff
@@ -10,21 +10,77 @@
...
+ *
+ * // Get w
+ * var w = vec.w;
+ *
+ * // Set w
+ * vec.w = 0;
+ */
+class Vec4 {
+    constructor(x = 0, y = 0, z = 0, w = 0) {
+        if (x && x.length === 4) {
```
mvaligursky, line 71: same here, "x &&" part can be removed

willeastcott: Fixed

src/math/vec3.js
willeastcott: Fixed

src/math/vec2.js
willeastcott: Fixed

src/math/mat4.js
willeastcott: And no, conflicts aren't possible. Rollup renames variables that conflicts when building the bundled library. It's pretty cool!

mvaligursky: I mean more like conflict with a local variable inside function of Vec3. Few functions have "var x". If I don't have this var line would x reference the global x?

willeastcott: Yes.

src/shape/bounding-box.js
```diff
@@ -75,29 +74,28 @@ Object.assign(BoundingBox.prototype, {
...
      * @function
      * @name pc.BoundingBox#copy
      * @description Copies the contents of a source AABB.
      * @param {pc.BoundingBox} src - The AABB to copy from.
      */
-    copy: function (src) {
+    copy(src) {
         this.center.copy(src.center);
         this.halfExtents.copy(src.halfExtents);
-        this.type = src.type;
```
mvaligursky, line 89: should be ok to remove, I don't see references to it .. but curious what it was?

src/shape/oriented-box.js
```diff
@@ -19,17 +19,16 @@ var tmpMat4 = new Mat4();
...
+        this.halfExtents = halfExtents;
 
-    worldTransform = worldTransform || tmpMat4.setIdentity();
-    this._modelTransform = worldTransform.clone().invert();
+        this._modelTransform = worldTransform.clone().invert();
 
-    this._worldTransform = worldTransform.clone(); // temp - currently only used in the worldTransform accessor, see future PR for more use
-    this._aabb = new BoundingBox(new Vec3(), this.halfExtents);
-}
+        this._worldTransform = worldTransform.clone(); // temp - currently only used in the worldTransform accessor, see future PR for more use
```
mvaligursky, line 28: should probably remove the comment, seem obsolete / irrelevant

mvaligursky (approved):
looking great!

src/shape/bounding-box.js
willeastcott: So this used to be a string type property on each shape instance that specified what it was. But it was removed years ago. This bug was actually pointed out to me by the Typescript compiler but since that PR is unlikely to be merged, I'm adding the fix here.

src/shape/oriented-box.js
willeastcott: Removed

PR #2642, opened 2020-12-26
ES6+ (ECMA2015+) support in source, with compile target to ES5 and ES6.

Maksims:
Based on previous conversations to <ins>support new ECMA features in JavaScript</ins>, there is a motivation to enable modern JavaScript features, to list a few: `class, async/await, let, const, ...rest parameters, default parameters, anonymous functions, and many more`.

This will enable community to write more manageable and performant code, taking advantages of modern browsers.

But there is a need to support older browsers and platforms, as well as vast community of existing projects. So <ins>this effort should respect backwards compatibility</ins>, and should not break any public APIs.

## Supporting both, ES5 and ES6
```html
<script type='module' src='playcanvas.es6.min.js'></script>
<script nomodule src='playcanvas.min.js'></script>
```
This pretty neat snippet, will serve ES5 version (playcanvas.min.js) to every platform that does not support module syntax. And for modern JavaScript, it will load ES6 version, but not ES5.

## Build still produces ES5 version by default, with ES6 version alongside
So this PR is non breaking!

## Using Google Closure Compiler with: `language_in` and `language_out`
So no new tools been introduced here.

## Approach to migrating codebase

This effort, should be gradual, and migrate engine slowly, with good amount of research into modern JavaScript features that do transpile to ES5 well, and do not generate vasts amounts of polyfills. So our coding standards should be updated accordingly with whitelisted ES6+ features. This PR includes such list, and currently it is:

* `import`/`export` handled by build scripts for bundling <ins>this is actually already in use across repository</ins>
* `let`, `const` instead of `var`
* `for of` loop
* `class` instead of `prototype`
* function default parameters

Also in this PR I've migrated /src/shape/* files, to comply with new coding standards: classes, let, const, function default parameters. And it is compiled to ES5 without reasonable excess.

### Here is an example of Ray code compiled and minified for both targets:

ES6 (prettified):
```javascript
class Qd {
    constructor(a=new z(0,0,0), b=new z(0,0,-1)) {
        this.origin = a;
        this.direction = b
    }
    set(a, b) {
        this.origin.copy(a);
        this.direction.copy(b);
        return this
    }
}
```

ES5 (prettified):
```javascript
var nd = function(a, b) {
    a = void 0 === a ? new z(0,0,0) : a;
    b = void 0 === b ? new z(0,0,-1) : b;
    this.origin = a;
    this.direction = b
};
nd.prototype.set = function(a, b) {
    this.origin.copy(a);
    this.direction.copy(b);
    return this
};
```

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -96,11 +96,11 @@ foo();
 bar(1, 2);
 ```
 
-### Use spaces between [ ] and { } brackets
+### Use spaces between [ ] and { } brackets, unless it is empty
 ```javascript
-let a = { };
+let a = {};
 let b = { key: 'value' };
-let c = [ ];
+let c = [];
 let d = [ 32, 64 ];
 ```
 
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -22,7 +22,7 @@ Try to keep PR focused on a single feature, small PR's are easier to review and 
 
 ## General
 
-These coding standards are based on the [Google JavaScript Coding Standards](https://google.github.io/styleguide/javascriptguide.xml). If something is not defined here, use this guide as a backup.
+Out coding standards are based on the [Google JavaScript Coding Standards](https://google.github.io/styleguide/javascriptguide.xml) which is ES5 (ECMA2009), with more specifics bellow. Also we have a whitelist of modern JavaScript features (ES6+), explicitly listed below.
 
 ### Keep it simple
 
@@ -31,6 +31,13 @@ Simple code is always better. Modular (horizontal dependencies) code is easier t
 ### Use International/American English spelling
 
 For example, use "Initialize" instead of "Initialise", and "color" instead of "colour".
+
+### Whitelisted ES6+ features:
+
+* `let`, `const` instead of `var`
+* `class` instead of `prototype`
+* `import`/`export` handled by build scripts for bundling
+* function default parameters
 
 ### Opening braces should be on the same line as the statement
 
@@ -68,16 +75,16 @@ On save, set your text editor to remove trailing spaces and ensure there is an e
 ### Use spaces between operators
 
 ```javascript
-var foo = 16 + 32 / 4;
+let foo = 16 + 32 / 4;
 
-for (var i = 0, len = list.length; i < len; i++) {
+for (let i = 0, len = list.length; i < len; i++) {
     // ...
 }
 ```
 
 ### Leave a space after the function keyword for anonymous functions
 ```javascript
-var fn = function () {
+let fn = function () {
 
 };
 ```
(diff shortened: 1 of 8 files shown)


willeastcott:
Interesting to have two options to choose from now.

A couple of observations:

1. From a glance, seems that Closure Compiler has no support for static class properties that Babel can handle (like [this](https://github.com/playcanvas/engine/blob/8f7598fc07076509abcb492c1684f7752c2f250e/src/math/vec3.js#L503) for example). It seems to lag behind Babel in terms of features. I don't know if there are other language features that Closure Compiler doesn't support (yet).
2. Terser seems to have become the best choice for minification and obfuscation. [Terser](https://www.npmjs.com/package/terser) has over 100x the number of weekly downloads on NPM compared to [Closure Compiler](https://www.npmjs.com/package/google-closure-compiler). Some results from taking current `playcanvas.js`:
  * `playcanvas.js`: 1984KB
  * Uglify (gzipped): 283KB
  * Terser (gzipped): 284KB
  * Closure Compiler (gzipped): 291KB
Slightly odd that Terser output is slightly larger than Uglify. But it's still 7KB (>2%) smaller than Closure Compiler.

Maksims:
Good observations. Based on this, alternative option to try would be Babel + Terser Vs Closure.
I will experiment with that.

willeastcott:
If you want to play with Babel, just sync my branch here:

https://github.com/playcanvas/engine/pull/2641

kungfooman:
From PR: `for (let i = 0, len = list.length; i < len; i++) {`

If you want to ban/discriminate `var` and force people to use `let`/`const` properly, then this should become:

```js
const list = [1, 2, 3];
const len = list.length;
for (let i = 0; i < len; i++) {
    const value = list[i];
    console.log("value is", value);
}
```

But that feels so overly long/ugly. Did you consider to add `for/const/of` to whitelisted ES6 features?

```js
const list = [1, 2, 3];
for (const value of list) {
    console.log("value is", value);
}
```

Maksims:
@kungfooman thank you for noticing it. I've just checked the quality of code compiled for ES5 and ES6 from `for of` loops, and it is very minimal, so I've added to coding standards this as well. Thank you.

willeastcott:
About this:

```html
<script type='module' src='playcanvas.es6.min.js'></script>
<script nomodule src='playcanvas.min.js'></script>
```

I don't see how this would work unless a PlayCanvas application was fully bundled with rollup or webpack or whatever. You wouldn't be able to just put that at the top of a PlayCanvas example (e.g. [this one](https://github.com/playcanvas/engine/blob/master/examples/graphics/model-box.html)). You'd need to put the import statement somewhere. And a script tag with the `module` attribute loads asynchronously, whereas the `nomodule` element loads synchronously.

Maksims:
> About this:
> 
> ```
> <script type='module' src='playcanvas.es6.min.js'></script>
> <script nomodule src='playcanvas.min.js'></script>
> ```
> 
> I don't see how this would work unless a PlayCanvas application was fully bundled with rollup or webpack or whatever. You wouldn't be able to just put that at the top of a PlayCanvas example (e.g. [this one](https://github.com/playcanvas/engine/blob/master/examples/graphics/model-box.html)). You'd need to put the import statement somewhere. And a script tag with the `type='module'` attribute loads asynchronously, whereas the `nomodule` element loads synchronously.

You are right, this multi-script approach wont work for current examples, as they have to change slightly too. This dual appoach if for those who want to have fallback to es5 engine. They have to handle their code compatibility with es5 as well. If they use modern way (module), then they would need to import and use modules.

If they want to use only widest option, then they use es5 engine as they use it now, nothing changes here.

And if they want to use only modern way, then they do something like:
```html
<script type='module'>
    import './playcanvas.es6.min.js';

    var canvas = document.getElementById("application-canvas");
    var app = new pc.Application(canvas);
    app.start();

    // ...
</script>
```

There is also a "hybrod" option possible, where you would write your application logic in ES5 manner, while loading one or the other engine depending on support.


And here we are stepping into different topic: what ES standards developers should write their code. They can always choose to use modern or ES5 conservative features, and before this PR, engine ensures both paths are valid.
With this PR, we provide more options to developers, where they can choose to support or not, by that benefitting with slightly more to modern specs engine. But this is again, different topic, as this PR is focusing on engine developers, not engine users, without reducing engine users experience.

kungfooman:
> Please notice, that `for of` loop is about twice slower than classic `for(;;)` loop, so for long loops with frequent iterations, it is better to use `for(;;)` unless it is considered as insignificant micro optimisation.

I just figured that Firefox `for of` is even 2/3 slower. I found an existing benchmark and added some other examples: https://jsben.ch/V51If

It kinda flips with multiple iterations, because `var`/`let`/`const` seem to produce the same results:

Chrome result:

image

(funny thing is that the "potenially" slowest one using a normal `var` without cached `length` won)

Another run in Chrome:

image

(now the const version won, which I would always expect because `const` may produce some more optimisations, but it doesn't)

The mess in Firefox, completely unoptimized `for of`:

image

Another run in Firefox:

image

I guess the lessons are:
 - `for of` should not be used
 - `for const of` seems to be reliable slower than `for var of`  in Chrome 87
 - `let`/`const` doesn't seems to optimize anything
 - caching `list.length` doesn't seems to optimize anything
 - yet future JS engines will optimize JS differently

Maksims:
**Thank you for a good test!**

I've modified a bit your test, adding some load for each tick, and Chrome has insignificant difference in performance. Firefox does a bit different:
image
And results are random, and fluctuating. So I would assume, with any load inside of the loop, the difference in porformance of the loop code is not significant when there is any load inside. So this is definitely can be considered as micro optimisation. And `for of` - looks neat as a code, definitely worth using.

V8 and other JS Engines, today do a great work on optimizing code behind the scenes, and they run it sometimes very close to native speeds. So we probably should not try to go crazy here, unless there is a real reason for it.

Maksims:
@willeastcott here is another reason for `<script type="module">` approach. Basically, there are many features been added in recent years to JavaScript. Since ES6, changes been made to improve language rapidly, and they all have various availability.

`nomodule` and `type="module"` approach, gives us clear distinction between es5 and es6, but we don't know if there is es7 or es8 features.
That is why closure compiler is set up to compile to two targets: ES5 and ES6. Although source of the engine can be written even with ES8 features, we just as mentioned before: have to ensure it compiles into reasonable alternative.

PR #2644, opened 2020-12-29
Graphics files convered to classes

mvaligursky:


src/scene/batching/skin-batch-instance.js
```diff
@@ -1,27 +1,26 @@
...
 
-    this.device = device;
-    this.rootNode = rootNode;
+        super();
 
-    // Unique bones per clone
-    this.bones = nodes;
-}
+        var numBones = nodes.length;
+        SkinInstance.prototype.init.call(this, device, numBones);
```
willeastcott, line 11: Does `this.init(device, numBones);` not work?

src/scene/batching/skin-batch-instance.js
```diff
@@ -46,8 +45,8 @@ Object.assign(SkinBatchInstance.prototype, {
             mp[base + 11] = pe[14];
         }
 
-        SkinInstance.prototype.uploadBones.call(this, this.device);
+        super.uploadBones(this.device);
```
willeastcott, line 48: Can't this just be `this.uploadBones(this.device);`, with the behavior being to look up the class hierarchy to the superclass automatically?

src/scene/immediate.js
```diff
@@ -12,50 +12,53 @@ import { MeshInstance } from '../scene/mesh-instance.js';
...
-    this.lineBatches = [];
-    this.layers = [];
-    this.layerToBatch = {};
-    this.quadMesh = null;
-    this.cubeLocalPos = null;
-    this.cubeWorldPos = null;
-    this.identityGraphNode = new GraphNode();
-}
+class ImmediateData {
+
```
willeastcott, line 16: Just as a reminder, I didn't insert this newline in the classes I've migrated. We should be consistent. I just followed the convention used on MDN:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes

src/scene/batching/skin-batch-instance.js
Maksims: It should be `super`, otherwise it will start lookup from `this`, which will point to same methods.

src/scene/batching/skin-batch-instance.js
Maksims: Should be `super.init(device, numBones)`;

src/scene/batching/skin-batch-instance.js
willeastcott: But it won't find `uploadBones` on this, so will then look for the function on the superclass.

Maksims: Should do, you are right. Weird it used `prototype...call` before, as being inherited, it should've worked with `this.uploadBones` as well.

src/scene/batching/skin-batch-instance.js
mvaligursky: changed to this.init. No need for super.init and this does not implement init.

src/scene/batching/skin-batch-instance.js
mvaligursky: changed to this. uploadBones. No need for super.init and this does not implement uploadBones.

willeastcott (approved):

willeastcott:
Approving assuming comments are addressed. 😄

PR #2647, opened 2020-12-29
Refactor script reserved name code

willeastcott:
This PR removes circular dependencies from the script code. Currently, the engine stores two additional properties on the `createScript` function which looks pretty horrible. I've made one private to `script.js` and the other has been made a static class property of `ScriptAttributes`.

It also simplifies the code to generates the look-up objects for the reserved names by leveraging the `Set` object.

Since the Editor relies on the previous code, I've added support for backwards compatibility, but that can be removed some time after the Editor codebase has been updated.


yaustar (approved):
LGTM

PR #2650, opened 2020-12-30
RefCountedObject base class, used on Mesh and Morph

mvaligursky:
- extracted and improved refCount functionality from the Mesh class into standalone class
- Mesh/MeshInstance refcounting refactored to use it
- Morph class uses refcounting as well - allowing correct release of memory when Morph is shared by multiple MorphInstances

### DIFF
--- a/src/core/ref-count.js
+++ b/src/core/ref-count.js
@@ -1,0 +1,34 @@
+// base class for reference counted objects
+class RefCount {
+    constructor() {
+
+        // counter
+        this._refCount = 0;
+    }
+
+    // inrements the counter
+    incRefCount() {
+        this._refCount++;
+    }
+
+    // decrements the counter. When the value reaches zero, destroy is called
+    decRefCount() {
+        if (--this._refCount == 0) {
+            this.destroy();
+        }
+    }
+
+    // returns current count
+    getRefCount() {
+        return this._refCount;
+    }
+
+    // default version of destroy functionality
+    destroy() {
+        // #ifdef DEBUG
+        console.error("Class extended from the RefCount class needs to have a destroy function implemented.");
+        // #endif
+    }
+}
+
+export { RefCount };
--- a/src/core/ref-count.js
+++ b/src/core/ref-count.js
@@ -1,7 +1,6 @@
 // base class for reference counted objects
 class RefCount {
     constructor() {
-
         // counter
         this._refCount = 0;
     }
--- a/src/core/ref-count.js
+++ b/src/core/ref-count.js
@@ -13,7 +13,7 @@ class RefCount {
 
     // decrements the counter. When the value reaches zero, destroy is called
     decRefCount() {
-        if (--this._refCount == 0) {
+        if (--this._refCount === 0) {
             this.destroy();
         }
     }
--- a/src/scene/mesh.js
+++ b/src/scene/mesh.js
@@ -1,3 +1,4 @@
+import { RefCount } from '../core/ref-count.js';
 import { Vec3 } from '../math/vec3.js';
 
 import { BoundingBox } from '../shape/bounding-box.js';
@@ -175,9 +176,9 @@ class GeometryVertexStream {
  * @property {pc.Skin} [skin] The skin data (if any) that drives skinned mesh animations for this mesh.
  * @property {pc.Morph} [morph] The morph data (if any) that drives morph target animations for this mesh.
  */
-class Mesh {
+class Mesh extends RefCount {
     constructor(graphicsDevice) {
-        this._refCount = 0;
+        super();
         this.id = id++;
         this.device = graphicsDevice || Application.getApplication().graphicsDevice;
         this.vertexBuffer = null;
@@ -204,18 +205,6 @@ class Mesh {
 
     set aabb(aabb) {
         this._aabb = aabb;
-    }
-
-    get refCount() {
-        return this._refCount;
-    }
-
-    incReference() {
-        this._refCount++;
-    }
-
-    decReference() {
-        this._refCount--;
     }
 
     /**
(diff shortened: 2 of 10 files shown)


src/core/ref-count.js
```diff
@@ -0,0 +1,33 @@
+// base class for reference counted objects
+class RefCount {
+    constructor() {
+        // counter
+        this._refCount = 0;
+    }
+
+    // inrements the counter
+    incRefCount() {
```
willeastcott, line 9:
```suggestion
    increment() {
```

src/core/ref-count.js
```diff
@@ -0,0 +1,33 @@
...
+        this._refCount++;
+    }
+
+    // decrements the counter. When the value reaches zero, destroy is called
+    decRefCount() {
```
willeastcott, line 14:
```suggestion
    decrement() {
```

src/core/ref-count.js
```diff
@@ -0,0 +1,33 @@
...
+        if (--this._refCount == 0) {
```
willeastcott, line 15:
```suggestion
	        if (--this._refCount === 0) {
```

mvaligursky: fixed, thanks!

willeastcott (approved):

src/core/ref-count.js
Maksims, line 9: Isn't it too much information in name of the methods?
`refCount.incRefCount()`?
Probably much more clean and "webby" would be: `refCount.increment()`.

src/scene/mesh.js
```diff
@@ -175,9 +176,9 @@ class GeometryVertexStream {
  * @property {pc.Skin} [skin] The skin data (if any) that drives skinned mesh animations for this mesh.
  * @property {pc.Morph} [morph] The morph data (if any) that drives morph target animations for this mesh.
  */
-class Mesh {
+class Mesh extends RefCount {
```
Maksims, line 179: This feels as not an "ideal" use of inheritance, as both those things are very different. It feels as reference counter, should be just a property, not a super class of Mesh.

Maksims: Or, if it serves as "destroyable object", then it probably should be something like: "DestroyableResource", with reference counting inside, then it feels as good inheritable class.

src/core/ref-count.js
willeastcott: I would totally agree if this was a stand-alone class, but since it's been written as a base class, `increment` and `decrement` would be non-obvious for somebody reviewing subclass code that calls those functions.

mvaligursky: you call it as `mesh.incRefCount();` and not `refCount.incRefCount();`
Considering this is a base class, `mesh.increment();` would be too generic.

Maksims: Made a comment bellow, regarding it's purpose related to a class name.

src/scene/mesh.js
willeastcott: Yeah, I wondered something similar about the naming. e.g. 'RefCountedObject' or something.

Maksims: Sorry for offtopic, but it reminded me those days where we would spend hours arguing on how to call things 😆 
Seems like it never gets better with time. 😋

willeastcott: I think it's easy to see these 'arguments' (or discussions) as slightly petty. But I think they're incredibly valuable for building a consistent and intuitive API. I think we all have valuable opinions to share to help us make collectively great decisions.

src/core/ref-count.js
mvaligursky: I called them incRef and decRef in the past, maybe that'd be a good middle ground

src/scene/mesh.js
mvaligursky: That's the intention behind it, but I didn't want all destroyable resources to have to be ref counted. In typed languages world (c#, typescript) I'd have base class with ref counting, and enforce IDestroyable and ILostContext functions using interfaces.

I also considered hierarchy with top level being Resource, which would have Destroy and LostContext functions, next level would be RefCountedResource which would add ref counter. Then engine classes could extend either of these.

At the moment recently I added loseContext to bunch of classes to handle that case, but I'd really like an interface in JS. Maybe TS?

src/scene/sprite.js
```diff
@@ -258,11 +258,8 @@ Sprite.prototype.destroy = function () {
     // destroy old meshes
     for (i = 0, len = this._meshes.length; i < len; i++) {
         var mesh = this._meshes[i];
-        if (!mesh) continue;
-
-        mesh.vertexBuffer.destroy();
-        for (var j = 0, len2 = mesh.indexBuffer.length; j < len2; j++) {
-            mesh.indexBuffer[j].destroy();
```
willeastcott, line 265: Is this begging for a `for..of` loop? 😄

mvaligursky: yes!

mvaligursky:
I look at using mixins instead of base class for this, but I agree with most of negative impact mentioned here:
http://raganwald.com/2016/07/16/why-are-mixins-considered-harmful.html

slimbuck:
Bit late to the party, but I wonder whether using explicit object references in the engine might ultimately be cleaner?

With something like:
```javascript
class ObjectRef {
    constructor(target) {
        this._target = null;
        if (target) {
            this.set(target);
        }
    }

    set(target) {
        if (target === this._target) {
            return;
        }
        if (this._target !== null && --this._target._refCount === 0) {
            this._target.destroy();
        }
        this._target = target;
        if (this._target) {
            ++this._target._refCount;
        }
    }

    get() {
        return this._target;
    }

    clear() {
        this.set(null);
    }
};
```

Then `morph` class would contain a `this.morphRef` and `meshInstance` would contain a `this.meshRef`.

The advantage being `set()` functionality (i.e. testing for same targets etc - which is crucial) is only implemented once (instead of at every call site) and we have no horrible class inheritance.

The disadvantage is call sites must do `this._meshRef.get()` which isn't as neat as `this._mesh`. (Although being forced to interface with a reference class also makes the relationship between the two objects explicit and clear).

PR #2652, opened 2020-12-30
Build single targets via `npm run build:es5` etc.

kungfooman:
`npm run build` is building everything at once, which takes like half a minute on my machine and it would be nice to only build whatever is needed.

For example, in order to test code changes in the example browser, this is enough: `npm run build:es5`

This would also make it easier for me to merge master for the AssemblyScript PR, since I am adding even more targets (32bit, 64bit, ASJS) and the `export default [{...}, {...}, ...];` map became pretty long.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,10 @@
   },
   "scripts": {
     "build": "rollup -c",
+    "build:es5": "rollup -c --environment target:es5",
+    "build:es6": "rollup -c --environment target:es6",
+    "build:debug": "rollup -c --environment target:debug",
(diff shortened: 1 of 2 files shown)


Maksims:
Just to mention, you know you can use `npm run build -- -w` for live rebuilds when code is changed. Which builds faster.
But I aggree, more specific build options, as well as `watch` option for specific target would be neat for development purposes.

willeastcott (approved):

willeastcott:
Nice contribution - thanks @kungfooman!

PR #2656, opened 2020-12-31
Migrate most of graphics src to ES6

willeastcott:
Migrate the majority of the `src/graphics` to ES6 (mainly classes).

Left out from this PR are:

* `GraphicsDevice` - because it inherits from `EventHandler`
* `VertexFormat` - `VertexFormat.defaultInstancingFormat` might need to be refactored
* `PostEffect` - we should check that a prototype-based post effect subclassed from `pc.PostEffect` still works with an ES6 class base class.
* `program-lib` folder - to avoid creating conflicts in yet to be merged PRs

### DIFF
--- a/src/graphics/graphics.js
+++ b/src/graphics/graphics.js
@@ -4,14 +4,14 @@
  * @type {number}
  * @description Ignores the integer part of texture coordinates, using only the fractional part.
  */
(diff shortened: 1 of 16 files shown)


mvaligursky (approved):
looks great.

PR #2663, opened 2021-01-01
Split anim code to separate classes and convert to ES6

### LINKED ISSUE #2589: Separate Anim file into separate files for each class it contains
This would make it easier to navigate and understand the code.

https://github.com/playcanvas/engine/blob/master/src/anim/anim.js

current content:

`export { AnimBinder, AnimClip, AnimCurve, AnimData, AnimEvaluator, AnimSnapshot, AnimTarget, AnimTrack, DefaultAnimBinder };`

willeastcott:
* Move old animation system to `src\animation`
* Split `src\anim\anim.js` into 1 file per class
* Migrate anim code to ES6 classes

Fixes #2589


src/anim/anim-target.js
```diff
@@ -0,0 +1,41 @@
+/**
+ * @private
+ * @callback pc.AnimSetter
+ * @description Callback function that the {@link pc.AnimEvaluator} uses to set final animation values.
+ * These callbacks are stored in {@link pc.AnimTarget} instances which are constructed by an
+ * {@link pc.AnimBinder}.
+ * @param {number[]} value - updated animation value.
+ */
```
mvaligursky, line 8: is there some better place for this one?

mvaligursky (approved):
nice one, so much better.

src/anim/anim-target.js
willeastcott: It could arguably go in `src\callbacks.js`. But we can consider that in a follow up PR.

PR #2673, opened 2021-01-03
Export default for ES6 build

kungfooman:
Currently only this works:

```js
import * as pc from "./playcanvas.mjs";
```

But usually people expect this to work:

```js
import pc from "./playcanvas.mjs";
```

This is adding all export statements again, but as `export default`:

image

So the file size grows, but a HTTP server with gzip is compressing/zipping that all away (because it is equal to the previous string)

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -91,14 +91,14 @@ function exportDefault() {
     return {
         /**
          * @param {string} code - Input code.
-         * @param {RenderedChunk} chunk - See https://github.com/rollup/rollup/blob/master/src/rollup/types.d.ts
+         * @param {import('./node_modules/rollup/dist/rollup').RenderedChunk} chunk - See https://github.com/rollup/rollup/blob/master/src/rollup/types.d.ts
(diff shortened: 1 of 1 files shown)


willeastcott:
I'm not sure I see the problem here. The module build is new, so the examples and docs haven't been updated yet to talk about it. But every example and doc will have:

```
import * as pc from "./playcanvas.mjs";
```

(...or whatever we choose to call the module which still hasn't really been decided). If this was a trivial change (e.g. a flag in Rollup, say, and it didn't really add much to the module file), I'd be more into the idea of doing this. But it just looks weird from a code point of view and sure, gzip will minimize the bloat, but it just looks wasteful still. Sorry not to sound more positive. For now, I'd rather focus on getting the module properly named and exposed in the repo, properly documented and integrated into the examples. Is that OK? I don't want to sound ungrateful because I get what you're trying to do here. Happy to revisit this down the road.

willeastcott:
Closing for now - but let's remember to pick this back up in a few weeks.

PR #2675, opened 2021-01-04
[FIX] Fixing use of pc. in light.js

mvaligursky:


willeastcott (approved):

src/scene/light.js
```diff
@@ -326,15 +332,15 @@ class Light {
...
                 height: 64,
                 format: format,
-                addressU: pc.ADDRESS_CLAMP_TO_EDGE,
-                addressV: pc.ADDRESS_CLAMP_TO_EDGE,
-                type: pc.TEXTURETYPE_RGBE,
-                magFilter: pc.FILTER_LINEAR,
-                minFilter: pc.FILTER_NEAREST,
+                addressU: ADDRESS_CLAMP_TO_EDGE,
+                addressV: ADDRESS_CLAMP_TO_EDGE,
+                type: TEXTURETYPE_DEFAULT,
```
willeastcott, line 341: This is changing more than removing the `pc.`. What's the justification here?

mvaligursky: A bug I introduced by copy & paste when writing this code.

willeastcott: OK, so it's correct now? Cool.

mvaligursky: yes it's just normal RGBA texture without a special meaning for the alpha.

PR #2681, opened 2021-01-06
LayerComposition and Forward-renderer refactoring to use RenderAction

mvaligursky:
the main reason behind this refactoring is for LayerComposition to build a list of RenderActions, which is an order in which { camera / layer / renderTarget } combinations are rendered. This takes place only when LayerComposition is dirty.

This avoids forward-renderer having to do this each frame, and opens the doors to changing the way layers are organized in following PRs (by just changing RenderAction order without having to change forward-renderer)

### DIFF
--- a/src/scene/layer-composition.js
+++ b/src/scene/layer-composition.js
@@ -6,6 +6,22 @@ import {
     LIGHTTYPE_DIRECTIONAL, LIGHTTYPE_OMNI, LIGHTTYPE_SPOT
 } from './constants.js';
 
+// class representing an entry in the final order of rendering of cameras and layers in the engine
+// this is populated at runtime based on LayerComposition
+class RenderAction {
+    constructor() {
+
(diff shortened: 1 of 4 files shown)


src/scene/layer-composition.js
```diff
@@ -6,6 +6,22 @@ import {
...
+        // index into a layer stored in LayerComposition.layerList
+        this.layerIndex = 0;
+
+        // index into a camera array of the layer, stored in Layer.cameras
+        this.cameraIndex = 0;
+
+        // render target this render action renders to (taken from either camera or layer)
+        this.renderTarget = null;
+    }
+}
```
willeastcott, line 23: Should it go in it's own module perhaps?

mvaligursky: done, as it will get bigger as well in the future

willeastcott (approved):

PR #2683, opened 2021-01-06
Additional engine ES6 classes

mvaligursky:
layer
forward-renderer
particle-system

### DIFF
--- a/src/math/mat4.js
+++ b/src/math/mat4.js
@@ -1270,6 +1270,24 @@ class Mat4 {
         return t;
     }
 
+    // converts this matrix to supplied Mat3 format
+    toMat3(m3) {
+        var src = this.data;
+        var dst = m3.data;
+
+        dst[0] = src[0];
+        dst[1] = src[1];
+        dst[2] = src[2];
+
+        dst[3] = src[4];
+        dst[4] = src[5];
+        dst[5] = src[6];
+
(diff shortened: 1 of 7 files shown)


src/math/mat4.js
```diff
@@ -1270,6 +1270,24 @@ class Mat4 {
...
+        dst[2] = src[2];
+
+        dst[3] = src[4];
+        dst[4] = src[5];
+        dst[5] = src[6];
+
+        dst[6] = src[8];
+        dst[7] = src[9];
+        dst[8] = src[10];
+    }
```
willeastcott, line 1289: Not sure this quite matches the naming convention of:

```
pc.Thing#setFromOtherThing
```

e.g. `pc.Quat#setFromMat4`

I mean, we also have `pc.Quat#toString` say. Ugh, I hate these sorts of consistency problems....

mvaligursky: ok, I'll change it to Mat3.setFrom(mat4) .. that seems more consistent?
probably easier to read as well, as class that its called on is the target.

willeastcott: Yeah, I think I prefer that.

mvaligursky: and done. looks better.

willeastcott (approved):

PR #2698, opened 2021-01-11
animation component currentTime setter function

gameatp:
animation component currentTime setter function has bug


willeastcott (approved):

willeastcott:
Thanks so much! 😄

Issue #2695, opened 2021-01-08
RefCountedObject has broken a UI example


willeastcott:
Specifically this one:

http://playcanvas.github.io/#user-interface/various.html

Should look as follows:

![image](https://user-images.githubusercontent.com/697563/104072330-f3491f80-5202-11eb-8994-a25e30690a18.png)

Now looks as follows:

![image](https://user-images.githubusercontent.com/697563/104072367-01973b80-5203-11eb-8f18-12af38265f42.png)

Print following errors:

![image](https://user-images.githubusercontent.com/697563/104072389-0d82fd80-5203-11eb-8afc-1e8a496cf4be.png)

This is because the mesh of one or more elements has been destroyed, but is still being rendered by the ForwardRenderer.

PR #2700, opened 2021-01-11
Change to refcounting of meshes to avoid behaviour change

mvaligursky:
Updates to changes introduced here: https://github.com/playcanvas/engine/pull/2650

- reference counting only handles counting, but not destroying of the object when that reaches zero - responsibility of that is left to the owning classes
- both Mesh/MeshInstance and Morph/MorphInstance were updating to do follow this
- small clean up to mesh-morph engine example

Fixes https://github.com/playcanvas/engine/issues/2695


src/scene/mesh-instance.js
```diff
@@ -157,11 +157,7 @@ class MeshInstance {
 
     set mesh(mesh) {
 
-        if (mesh === this._mesh)
-            return;
-
```
willeastcott, line 162: Are you sure removing this correct? I could do:

```
meshInstance.mesh = mesh;
meshInstance.mesh = mesh;
```

...and mesh will have a reference count of 2, no?

src/core/ref-counted-object.js
```diff
@@ -1,8 +1,11 @@
 // base class for reference counted objects
 class RefCountedObject {
-    constructor() {
+    constructor(autoDestroy = true) {
```
willeastcott, line 3: Would an option be to update the element system to respect a destroy on the ref count reaching 0?

src/scene/mesh-instance.js
mvaligursky: I removed it as original code didn't have this, I added it last week.
But I'm adding it back now that you mentioned it. In general, when you call meshInstance.mesh = ..., first old mesh is decReferenced and then new one is incReferenced. It code was auto releasing mesh, then without this line, the mesh would first be dereferenced and destroyed before adding reference again.
Even though it's not needed now as meshes are not auto destroyed, it's cleaner to have it there.

src/core/ref-counted-object.js
mvaligursky: I considered this, but auto destroying meshes change could still cause errors in existing projects, that internally keep meshes around for some reason without incReferencing them - they'd end up keeping mesh that has been auto destroyed by the engine.

willeastcott: So why ref count meshes at all? Isn't the whole point to automate destruction?

mvaligursky: Well the way it's been done (and I reverted to that original solution) is that meshes are refCounted, but what happens with them when ref count reaches zero depends on the owner of the mesh - and they don't always get destroyed, but they get reused.

We can change it and fix the engine to be consistent, but what about external code? I'd be happy with this option if we're happy to take the risk.

willeastcott: I would probably prefer we made this nice and consistent. If an app is doing something 'naughty', perhaps we can at least print helpful messages? For example, if you try to render a destroyed mesh, we can console log something informative.

slimbuck: Might be better to pass in an optional destruction function (where default implementation calls `.destroy`). And IMO still better to wrap this in https://github.com/playcanvas/engine/pull/2650#issuecomment-754009364

mvaligursky: Thinking about it more, perhaps refCounting class should never destroy object. Basically keep the engine as it was - refCounting only counts .. but the owner code decides what to do with it.
The common problematic case is that user creates few meshes, and stores them. They have refCount 0. Then the user wants to assign them to MeshInstance for rendering. This increases ref count to 1. When the user then removes it from meshInstance, it would drop to 0 and get destroyed. And so users reference would become invalid.
I think I'll remove auto destroy completely, and change the new user of refCounting (Mprph / MorphInstance) to match Mesh / MeshInstance in a way it destroy Morph as needed instead of depending on it to self-destruct.

slimbuck: In which case might be worth checking whether a base class is needed at all.

mvaligursky: the base class makes all classes to use the same interface. I like that.
When adding it, I considering adding this functionality to each class using mixins (https://javascript.info/mixins) but I find the actual base class more explicit.

mvaligursky: actually, maybe I'll just use mixin, it's probably suitable for this.

willeastcott (approved):
Approving, but also get @slimbuck's approval before merge since he had some opinions around this.

Maksims:
Being it just one counter variable, isn't it overkill to use separate file for it and whole object, instead of just a variable with a number?

mvaligursky:
> Being it just one counter variable, isn't it overkill to use separate file for it and whole object, instead of just a variable with a number?

Yes and no. We've discussed this with @slimbuck as well. I prefer to hide internal class implementations in the engine behind interface / functions. I don't like external classes doing mesh._refCount++; and similar, as that makes it harder to change functionality. Wrapping these in functions is a way to go. It's a lot easier to debug as well .. I added logging to add/dec functions and could track what's going on. Having multiple places where variable is directly changes makes it harder too. Keeping it in a separate "class" makes it easier to consistently reuse on multiple classes. And build upon.

But after discussion with @slimbuck , I'll go back to previous commit as far as structure goes, and use class for it instead of mixing, as we agree that's more explicit and clear.

If we go to TypeScript at some point, I'd use an Interface, with the implementation inside the class (Mesh, Morph). But ES6 does not have that, and this is as close as I can get.

slimbuck (approved):

Maksims:
@mvaligursky Chech out his: https://javascript.info/mixins

mvaligursky:
> @mvaligursky Chech out his: https://javascript.info/mixins

yes I've tried this, and we decided against it for now, see https://github.com/playcanvas/engine/pull/2700/commits/8fe5fa26ec814a2a1769156fac516a48bded41c0

PR #2703, opened 2021-01-11
onBeforeRemove in animation component doesn't handle array of pc.Assets

yaustar:
Fixes: https://forum.playcanvas.com/t/after-loading-animation-from-glb-app-destroy-errors/16348/6

Test project here: https://playcanvas.com/editor/scene/1067584 (press 1 to destroy the app which unloads the assets and destroys the scene)

### DIFF
--- a/src/framework/components/animation/component.js
+++ b/src/framework/components/animation/component.js
@@ -458,7 +458,12 @@ class AnimationComponent extends Component {
 
     onBeforeRemove() {
         for (var i = 0; i < this.assets.length; i++) {
-            var asset = this.system.app.assets.get(this.assets[i]);
+            var assetId = this.assets[i];
+            if (typeof assetId !== 'number') {
+                assetId = assetId.id;
+            }
+
+            var asset = this.system.app.assets.get(assetId);
             if (!asset) continue;
 
             asset.off('change', this.onAssetChanged, this);
--- a/src/framework/components/animation/component.js
+++ b/src/framework/components/animation/component.js
@@ -458,6 +458,8 @@ class AnimationComponent extends Component {
 
     onBeforeRemove() {
         for (var i = 0; i < this.assets.length; i++) {
+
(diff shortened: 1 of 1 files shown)


src/framework/components/animation/component.js
```diff
@@ -458,7 +458,14 @@ class AnimationComponent extends Component {
 
     onBeforeRemove() {
         for (var i = 0; i < this.assets.length; i++) {
-            var asset = this.system.app.assets.get(this.assets[i]);
+
+            // this.assets can be an array of pc.Assets or an array of numbers (assetIds)
+            var assetId = this.assets[i];
+            if (typeof assetId !== 'number') {
```
willeastcott, line 464: I would do `        if (value instanceof Asset) {` here instead. Kinda like:

https://github.com/playcanvas/engine/blob/master/src/framework/components/animation/component.js#L394

This seems to be the way the check is done throughout the framework (seach for the string `instanceof Asset`).

src/framework/components/animation/component.js
```diff
@@ -458,7 +458,14 @@ class AnimationComponent extends Component {
...
         for (var i = 0; i < this.assets.length; i++) {
-            var asset = this.system.app.assets.get(this.assets[i]);
+
+            // this.assets can be an array of pc.Assets or an array of numbers (assetIds)
+            var assetId = this.assets[i];
+            if (typeof assetId !== 'number') {
+                assetId = assetId.id;
+            }
+
+            var asset = this.system.app.assets.get(assetId);
```
willeastcott, line 468: I would do:

```
            var asset = this.assets[i];
            if (typeof asset ===  'number') {
                asset = this.system.app.assets.get(asset);
            }
```

yaustar: Makes sense, changed

willeastcott (approved):

Issue #2705, opened 2021-01-12
When a screen has many elements, moving elements out of the screen is slow.


szevvy:
We have a project which has 7,000+ elements in a screen, some of which we pool (and have reasonably deep hierarchies of their own).  When we return entities to the pooler, we move say 20 entities each of which has perhaps 10 entities as children.  As the pooler entity has no screen, each of these entities, when reparented, calls `_unbindScreen`, which in turns calls `screen.off` for the following events:

`set:resolution`
`set:referenceresolution`
`set:scaleblend`
`set:screenspace`
`remove`

*Each* of these off calls iterates over an array that contains 7000 items (one for every element in the screen).  For each element in that list, it checks if the callback matches the one passed into `off` (which it does, every time, as every element uses the same callback for each of these events), and then checks to see if the scope matches.

All in all, moving 200 elements causes 14,000,000 comparisons.

As a result, moving these items in the hierarchy causes significant lag - and the more elements you have in the screen (even if they are disabled), the worse it gets.

Thanks!

PR #2707, opened 2021-01-12
Optimize Element / Screen Events

vkalpias:
Fixes https://github.com/playcanvas/engine/issues/2705.

This PR changes the Element component to no longer subscribe to Screen events for various Screen property changes. Instead each Element component is stored in an internal set in the Screen component, and the Screen component calls the Element functions directly.

This is a lot faster and uses less memory in the expense of using some private Element functions from the Screen component.

### DIFF
--- a/src/framework/components/element/component.js
+++ b/src/framework/components/element/component.js
@@ -465,19 +465,11 @@ class ElementComponent extends Component {
     }
 
     _bindScreen(screen) {
-        screen.on('set:resolution', this._onScreenResize, this);
-        screen.on('set:referenceresolution', this._onScreenResize, this);
-        screen.on('set:scaleblend', this._onScreenResize, this);
-        screen.on('set:screenspace', this._onScreenSpaceChange, this);
-        screen.on('remove', this._onScreenRemove, this);
+        screen._bindElement(this);
     }
 
     _unbindScreen(screen) {
-        screen.off('set:resolution', this._onScreenResize, this);
-        screen.off('set:referenceresolution', this._onScreenResize, this);
-        screen.off('set:scaleblend', this._onScreenResize, this);
-        screen.off('set:screenspace', this._onScreenSpaceChange, this);
-        screen.off('remove', this._onScreenRemove, this);
+        screen._unbindElement(this);
     }
 
     _updateScreen(screen) {
--- a/src/framework/components/element/component.js
+++ b/src/framework/components/element/component.js
@@ -465,6 +465,13 @@ class ElementComponent extends Component {
     }
 
     _bindScreen(screen) {
+        // bind the Element to the Screen
+        // We used to subscribe to Screen events here however
+        // that was very slow when there are thousands of Elements.
+        // When the time comes to unbind the Element from the Screen
+        // finding the event callbacks to remove takes a considerable amount of 
+        // time. So instead the Screen stores the Element component and calls
+        // its functions directly.
         screen._bindElement(this);
     }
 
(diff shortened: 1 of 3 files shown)


src/framework/components/element/component.js
```diff
@@ -465,19 +465,11 @@ class ElementComponent extends Component {
     }
 
     _bindScreen(screen) {
-        screen.on('set:resolution', this._onScreenResize, this);
-        screen.on('set:referenceresolution', this._onScreenResize, this);
-        screen.on('set:scaleblend', this._onScreenResize, this);
-        screen.on('set:screenspace', this._onScreenSpaceChange, this);
-        screen.on('remove', this._onScreenRemove, this);
+        screen._bindElement(this);
```
slimbuck, line 468: might be worth adding a comment here explaining the optimisation.

src/framework/components/screen/component.js
```diff
@@ -136,12 +138,29 @@ class ScreenComponent extends Component {
         }
     }
 
+    _bindElement(element) {
+        this._elements.push(element);
+    }
+
+    _unbindElement(element) {
+        const index = this._elements.indexOf(element);
```
mvaligursky, line 146: could this_elements be a Set instead of an array to avoid this possibly costly indexOf operation?

mvaligursky (approved):
approved with a question

src/framework/components/element/component.js
vkalpias: Done

src/framework/components/screen/component.js
Maksims: I would assume, iterations are more common through this set, than modifications of it. So it is a question of reduced performance on add/remove or iteration.

mvaligursky: Set uses object pointer, so no hashing.

slimbuck: for users with pooled elements we could also provide a `screen.removeElements()` function removes all elements without requiring all the `.indexOf` calls.

slimbuck: related: https://stackoverflow.com/questions/39007637/javascript-set-vs-array-performance

vkalpias: This is for my own reference but according to this https://v8.dev/blog/hash-code hash codes for Objects are just random numbers. So I guess that's how Set's hash any object reference internally..

slimbuck (approved):

src/framework/components/screen/component.js
```diff
@@ -139,24 +139,19 @@ class ScreenComponent extends Component {
...
 
     onRemove() {
         this.system.app.graphicsDevice.off("resizecanvas", this._onResize, this);
         this.fire('remove');
 
-        for (let i = 0; i < this._elements.length; i++) {
-            this._elements[i]._onScreenRemove();
-        }
-        this._elements.length = 0;
+        this._elements.forEach(element => element._onScreenRemove());
```
Maksims, line 153: Before we use non ES5 feature, we need to ensure it is well compiled to ES5 by Babel, and then added to [CONTRIBUTING.md](https://github.com/playcanvas/engine/blob/master/.github/CONTRIBUTING.md).

vkalpias: Arrow functions are converted properly by Babel and already used in other parts of the engine in the conversion to modern JavaScript.

vkalpias: I'll add an issue to add that to CONTRIBUTING.md

willeastcott (approved):

willeastcott (approved):

PR #2719, opened 2021-01-13
fix light shape property causing incorrect light mapping

### LINKED ISSUE #2716: Sponza looks different with candidate 1.38.0 release
With 1.37.1:

image

https://playcanvas.com/editor/scene/423508

With latest master branch:

image

https://playcanvas.com/editor/scene/423508?use_local_engine=https://code.playcanvas.com/playcanvas-latest.dbg.js

Interestingly enough, there were already errors in the JS console for this scene. So there might be more than one problem in play here.

raytranuk:
Fixes #2716

this fixes light map problems with unmigrated lights in editor

### DIFF
--- a/src/scene/light.js
+++ b/src/scene/light.js
@@ -452,25 +452,26 @@ class Light {
     }
 
     get shape() {
-        return this._shape ? this._shape : LIGHTSHAPE_PUNCTUAL;
+        return this._shape;
     }
 
     set shape(value) {
-        if (this._shape === value)
+        var sanitizedValue = value ? value : LIGHTSHAPE_PUNCTUAL;
+
+        if (this._shape === sanitizedValue)
             return;
 
-        if (value !== LIGHTSHAPE_PUNCTUAL) {
+        if (sanitizedValue !== LIGHTSHAPE_PUNCTUAL) {
             this.uploadAreaLightLUTs();
         }
 
-        this._shape = value ? value : LIGHTSHAPE_PUNCTUAL;
+        this._shape = sanitizedValue;
         this._destroyShadowMap();
         this.updateKey();
 
         var stype = this._shadowType;
         this._shadowType = null;
         this.shadowType = stype; // refresh shadow type; switching shape and back may change it
-
     }
 
     get shadowType() {
--- a/src/scene/light.js
+++ b/src/scene/light.js
@@ -455,17 +455,15 @@ class Light {
         return this._shape;
     }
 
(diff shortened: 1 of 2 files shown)


slimbuck (approved):

src/scene/light.js
```diff
@@ -452,7 +452,7 @@ class Light {
     }
 
     get shape() {
-        return this._shape;
+        return this._shape ? this._shape : LIGHTSHAPE_PUNCTUAL;
```
willeastcott, line 455: Surely this is unnecessary if `_shape` is always guaranteed to be set to a valid value.

mvaligursky:
how about Light.updateKey() - it uses _shape as well

src/scene/light.js
raytranuk: I can test - but it doesn't seem that in the editor _shape will use class defaults if not present in the component - which is why editor migration code is needed

src/scene/light.js
```diff
@@ -463,7 +463,7 @@ class Light {
             this.uploadAreaLightLUTs();
         }
 
-        this._shape = value;
+        this._shape = value ? value : LIGHTSHAPE_PUNCTUAL;
```
willeastcott, line 466: Is this to guard against value being `null` or `undefined`? If so, won't it be undesirable to call `uploadAreaLightLUTs` above? Where are 'bad' values being passed to this function anyway?

src/scene/light.js
raytranuk: tested and removed the redundant code in the get accessor

src/scene/light.js
raytranuk: now using a sanitizedValue - I'm not sure where the bad values are coming from - I had assumed the _shape property was set once to the property default on construction - and left alone if not used - I can investigate further.

src/scene/light.js
```diff
@@ -456,21 +456,22 @@ class Light {
     }
 
     set shape(value) {
```
mvaligursky, line 458: just use default value

`set shape(value = LIGHTSHAPE_PUNCTUAL) {`

raytranuk: done

src/scene/light.js
raytranuk: used default param for sanitizing value - also removed incorrect assignment in initializeComponentData()

mvaligursky (approved):
looks good now

PR #2721, opened 2021-01-13
Scope space cleanup

mvaligursky:
- removed unused pc.ScopeSpace#getSubSpace
- moved function to clear texture from device to the class
- converted to use Map


mvaligursky:
note - I won't this merge now before the release.

Issue #2722, opened 2021-01-13
GLB parser uses old MorphTarget constructor signature


willeastcott:
See:

https://github.com/playcanvas/engine/blob/master/src/resources/parser/glb-parser.js#L799

And:

https://github.com/playcanvas/engine/blob/master/src/scene/morph-target.js#L26

So the engine can sometimes print that message in debug.

PR #2725, opened 2021-01-13
Updated glb-parser to use current Morph/MorphTarget API

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/2722

### DIFF
--- a/src/resources/parser/glb-parser.js
+++ b/src/resources/parser/glb-parser.js
@@ -796,10 +796,10 @@ var createMesh = function (device, gltfMesh, accessors, bufferViews, callback, d
                     options.name = targets.length.toString(10);
                 }
 
-                targets.push(new MorphTarget(device, options));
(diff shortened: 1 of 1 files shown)


src/resources/parser/glb-parser.js
```diff
@@ -796,10 +796,10 @@ var createMesh = function (device, gltfMesh, accessors, bufferViews, callback, d
                     options.name = targets.length.toString(10);
                 }
 
-                targets.push(new MorphTarget(device, options));
+                targets.push(new MorphTarget(options));
             });
 
-            mesh.morph = new Morph(targets);
+            mesh.morph = new Morph(targets, device);
```
willeastcott, line 802: Maybe I mentioned this at the time, but just for the record, the rest of the engine API _always_ takes the device as the first parameter if it's required. So this is non-conforming. But approving anyway.

willeastcott (approved):

PR #2731, opened 2021-01-14
Area light in editor engine fixes/changes

raytranuk:
Engine changes needed to work with editor changes to expose area lights:
- no public APIs changed
- area light luts are now in a versioned binary format asset, that can is set on a project setting for editor projects, and for engine only projects can be loaded and set on the app
- engine example adjusted to use binary asset
- also some other small fixes to correspond with editor PR

image


mvaligursky:
What's the size of zipped up binary file vs zipped up .js file from betore?

Maksims:
This engine example is so much better showcasing area lights.
If you adjust VSM shadow to lower resolution with higher blur, it will look softer. And probably some various colors for light sources will do a nice color mixings on illuminated parts.
Long slim area light sources - are good to showcase area lights, like ceiling lights.

raytranuk:
> What's the size of zipped up binary file vs zipped up .js file from before?

binary zips to 102kb and js zips to 109kb on my machine - so pretty close when zipped.

mvaligursky:
And one more question (or two):
- have you stripped out those unused zeros from lut2?
- how do we re-generate this binary file if we need to do changes in the future .. is there some tool you wrote that we can use?

raytranuk:
> This engine example is so much better showcasing area lights.

Nice and easy to set up in the editor :)

> If you adjust VSM shadow to lower resolution with higher blur, it will look softer. And probably some various colors for light sources will do a nice color mixings on illuminated parts.

I wasn't able to get good blurry shadow results using VSMs - I just got thinner shadows with more leaking - maybe something that would help is negative bias? I will investigate further.

> Long slim area light sources - are good to showcase area lights, like ceiling lights.

I'll try adding add some to my test scenes.

> Also, if you add outside walls, then shadows will not leak light in corners: https://developer.playcanvas.com/en/user-manual/graphics/lighting/lightmapping/#other-tips

Updated the scene remove light leaking and updated the image

raytranuk:
> And one more question (or two):
> 
> * have you stripped out those unused zeros from lut2?

Haven't yet - didn't want to change your texture gen code until got other things working. 

> * how do we re-generate this binary file if we need to do changes in the future .. is there some tool you wrote that we can use?

I wrote a small script - I'll include it in this PR.

mvaligursky:
Should we perhaps rename luts.bin to something like area-lights-luts.bin?

src/framework/application.js
```diff
@@ -1672,6 +1697,136 @@ Object.assign(Application.prototype, {
...
+     */
+    setAreaLightData: function (asset) {
+        if (asset) {
+            var self = this;
+            asset.ready(function (asset) {
+                self._uploadAreaLightLUTs(asset);
+            });
+            this.assets.load(asset);
+        }
+        // TODO: decide what to do in this case
```
mvaligursky, line 1827: I would say log an error to console

src/framework/application.js
```diff
@@ -457,6 +458,21 @@ function Application(canvas, options) {
         id: LAYERID_WORLD
     });
 
+    // placeholder for area light luts
```
mvaligursky, line 461: I'd vote for moving this to a function as well

src/framework/application.js
```diff
@@ -457,6 +458,21 @@ function Application(canvas, options) {
...
+        format: PIXELFORMAT_R8_G8_B8_A8
+    });
+    placeholderLutTex.name = 'placeholder';
+
+    var pixels = placeholderLutTex.lock();
+    pixels.fill(0);
+    placeholderLutTex.unlock();
+
+    this.graphicsDevice.scope.resolve('areaLightsLutTex1').setValue(placeholderLutTex);
+    this.graphicsDevice.scope.resolve('areaLightsLutTex2').setValue(placeholderLutTex);
```
mvaligursky, line 474: I'd vote for moving this to a function as well

src/framework/application.js
```diff
@@ -1089,6 +1105,15 @@ Object.assign(Application.prototype, {
...
 
+        if (props.areaLightData) {
+            asset = this.assets.get(props.areaLightData);
+            if (asset) {
+                this.setAreaLightData(asset);
+            } else {
+                this.assets.once('add:' + props.areaLightData, this.setAreaLightData, this);
+            }
+        }
+
```
mvaligursky, line 1116: and even this should be a function to keep it cleaner

src/scene/light.js
```diff
@@ -328,111 +328,7 @@ class Light {
         this.key = key;
     }
 
-    // creates LUT texture used by area lights
```
mvaligursky, line 331: by removing these, perhaps there are some imports that can be removed from the start of this file?

mvaligursky:
I added few small comments, looks good otherwise!

src/framework/application.js
Maksims: In Debug mode.

src/framework/application.js
willeastcott: This doesn't seem right to me. The `Application` is just an abstraction of a PlayCanvas app. It shouldn't know bout setting of uniforms etc. Seems like the data should be prepared+set in the `ForwardRenderer`. We might have another renderer one day that can't use the LUT data in the same way.

src/framework/application.js
raytranuk: added a warning

src/framework/application.js
raytranuk: moved to a function

src/scene/light.js
raytranuk: unused imports are removed (none are now flagged by lint)

src/framework/application.js
raytranuk: hmm - yes - I will change it so that project settings and asset handling is in Application and placeholder and real LUT textures are initialized and owned by the renderer.

raytranuk: I moved texture and uniform setting from Application into ForwardRenderer - I noticed there are some other places in application.js that perhaps should be refactored in this way too - so will add an issue to deal with this in another PR.

src/scene/forward-renderer.js
```diff
@@ -839,6 +844,135 @@ Object.assign(ForwardRenderer.prototype, {
...
+    _createAreaLightPlaceholderLuts: function () {
+        var placeholderLutTex =  new Texture(this.device, {
+            width: 2,
+            height: 2,
+            format: PIXELFORMAT_R8_G8_B8_A8
+        });
+        placeholderLutTex.name = 'placeholder';
+
+        var pixels = placeholderLutTex.lock();
+        pixels.fill(0);
```
willeastcott, line 857: Not available in IE11: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/fill#browser_compatibility

Maksims: Will Babel add polyfill for this automatically?

willeastcott: No. This:

```
class Mat4 {
    constructor() {
        var data = new Float32Array(16);
        data.fill(0);
        // Create an identity matrix. Note that a new Float32Array has all elements set
        // to zero by default, so we only need to set the relevant elements to one.
        data[0] = data[5] = data[10] = data[15] = 1;
        this.data = data;
    }
```

Becomes:

```
	var Mat4 = function () {
		function Mat4() {
			var data = new Float32Array(16);
			data.fill(0);
			data[0] = data[5] = data[10] = data[15] = 1;
			this.data = data;
		}
```

raytranuk: I replaced the `fill()` call with a nested for loop (copied from the material handler) - which should work with ie11.

src/framework/application.js
```diff
@@ -1672,6 +1686,26 @@ Object.assign(Application.prototype, {
...
+    setAreaLightData: function (asset) {
+        if (asset) {
+            var renderer = this.renderer;
+            asset.ready(function (asset) {
+                renderer._uploadAreaLightLuts(asset.resource);
+            });
+            this.assets.load(asset);
+        }
+        // #ifdef DEBUG
+        console.warn("setAreaLightData: asset is not valid");
```
willeastcott, line 1705: Won't this always print in DEBUG?

mvaligursky:
I've just realized the problem we had when the float LUT texture was the preferred option and it didn't work on iOS. We use float texture together with linear filtering, and so in addition to extTextureFloat we also need to test extTextureFloatLinear - could you please make this change when this._areaLightLutFormat is initialized. And perhaps change it to this.areaLightLutFormat (remove _) to match other capabilities .. like this.textureFloatRenderable or this.supportsMorphTargetTexturesCore and similar. @raytranuk

examples/assets/scripts/utils/area-light-lut-bin-gen.js
```diff
@@ -6,8 +6,25 @@
...
...0.992444, 0.0120254, 0, 0.0292711, 0.991028, 0.0123314, 0, 0.036417, 0.98803, 0.0124954, 0, 0.0442295, 0.984816, 0.0126538, 0, 0.0526815, 0.981399, 0.0128537, 0, 0.0617492, 0.977085, 0.0129694, 0, 0.0714114, 0.972154, 0.013091, 0, 0.0816495, 0.966617, 0.0131166, 0, 0.0924472, 0.960628, 0.0131583, 0, 0.10379, 0.953295, 0.0131094, 0, 0.115665, 0.94575, 0.0129966, 0, 0.128062, 0.937654, 0.0128796, 0, 0.140972, 0.927716, 0.0126477, 0, 0.154387, 0.917932, 0.0123889, 0, 0.168301, 0.907719, 0.012131, 0, 0.182709, 0.89584, 0.0118013, 0, 0.197608, 0.883526, 0.0114145, 0, 0.212994, 0.870301, 0.0110075, 0, 0.228867, 0.856272, 0.0106019, 0, 0.245227, 0.842251, 0.0101938, 0, 0.262074, 0.826466, 0.00973254, 0, 0.279412, 0.810859, 0.0092846, 0, 0.297244, 0.795051, 0.00883304, 0, 0.315575, 0.780053, 0.00840272, 0, 0.334412, 0.76575, 0.00796438, 0, 0.35376, 0.752298, 0.00752526, 0, 0.373631, 0.739153, 0.00711486, 0, 0.394034, 0.725514, 0.00670361, 0, 0.414983, 0.711473, 0.00632656, 0, 0.436491, 0.696936, 0.00595206, 0, 0.458575, 0.682126, 0.00559191, 0, 0.481253, 0.667027, 0.00525362, 0, 0.504547, 0.651875, 0.00493805, 0, 0.528481, 0.636463, 0.00462848, 0, 0.553081, 0.620641, 0.00433936, 0, 0.57837
```
mvaligursky, line 28: it's little bit strange to save version as float

src/framework/application.js
raytranuk: yes - whoops - will fix

examples/assets/scripts/utils/area-light-lut-bin-gen.js
raytranuk: I can change to integer - but maybe 2 integers, one for major and one for minor version?

raytranuk:
> I've just realized the problem we had when the float LUT texture was the preferred option and it didn't work on iOS. We use float texture together with linear filtering, and so in addition to extTextureFloat we also need to test extTextureFloatLinear - could you please make this change when this._areaLightLutFormat is initialized. And perhaps change it to this.areaLightLutFormat (remove _) to match other capabilities .. like this.textureFloatRenderable or this.supportsMorphTargetTexturesCore and similar. @raytranuk

I've added `extTextureFloatLinear` checks - I'm getting some strange results on my iPhone - I will investigate further.

examples/assets/scripts/utils/area-light-lut-bin-gen.js
raytranuk: done

src/framework/application.js
raytranuk: done

raytranuk:
> I've added `extTextureFloatLinear` checks - I'm getting some strange results on my iPhone - I will investigate further.

After upgrading to iOS 14.4 it works as expected.

willeastcott (approved):

examples/graphics/area-lights.html
```diff
@@ -186,6 +183,9 @@
 
             var miniStats = new pcx.MiniStats(app);
 
+            // set the loaded area light LUT data
+            app.setAreaLightData(assetManifest[4].asset);
```
mvaligursky, line 187: I wonder if this should be called setAreaLightLuts perhaps? Data is too generic.

src/scene/forward-renderer.js
```diff
@@ -823,6 +828,143 @@ class ForwardRenderer {
...
+        });
+        placeholderLutTex.name = 'placeholder';
+
+        var pixels = placeholderLutTex.lock();
+        for (var i = 0; i < 4; i++) {
+            for (var c = 0; c < 4; c++) {
+                pixels[i * 4 + c] = 0;
+            }
+        }
+        placeholderLutTex.unlock();
```
mvaligursky, line 846: We could make this simpler and create a single pixel texture. Maybe even create this function in a way we can reuse it, as I think you copied it from elsewhere? It seems not intuitive to create 2x2 texture and then do 4x4 loop .. even though the result is correct.

copied from Editor, would need to be cleaned up
```
    var createSolidTex = function (name, r, g, b, a) {
        var result = new pc.Texture(app.graphicsDevice, { width: 1, height: 1, format: pc.PIXELFORMAT_R8_G8_B8_A8 });
        result.name = name;
        var pixels = result.lock();
        pixels.set(new Uint8Array([r, g, b, a]));
        result.unlock();
        return result;
    };

```

src/framework/application.js
```diff
@@ -1667,6 +1681,27 @@ class Application extends EventHandler {
...
     }
 
+    /**
+     * @function
+     * @private
+     * @name pc.Application#setAreaLightData
+     * @description Sets the area light LUT asset for this app.
+     * @param {pc.Asset} asset - Asset of type `binary` to be set.
+     */
+    setAreaLightData(asset) {
```
mvaligursky, line 1691: renaming this to setAreaLightLuts would match renderer._uploadAreaLightLuts better too.

mvaligursky (approved):
approved with 2 suggested changes

examples/graphics/area-lights.html
raytranuk: sure - I can change this

src/scene/forward-renderer.js
raytranuk: I think this is a good idea - I would prefer that it be in another PR that would unify the other places that uses something similar - something that I also think could be useful is to be able to specify a default/placeholder texture that isn't a solid color and is easily visually identifiable as a missing texture - this can be very useful for debugging.

mvaligursky: sounds good!

src/framework/application.js
raytranuk: agree - and done

examples/graphics/area-lights.html
raytranuk: done

src/scene/forward-renderer.js
raytranuk: placeholder issue for further discussion here: https://github.com/playcanvas/engine/issues/2794

src/scene/forward-renderer.js
```diff
@@ -823,6 +828,143 @@ class ForwardRenderer {
...
+    _uploadAreaLightLuts(resource) {
+
+        var versions = new Int16Array(resource, 32768 * 4, 2);
+
+        this._areaLightLuts = {
+            data1: new Float32Array(resource, 0, 16384),
+            data2: new Float32Array(resource, 16384 * 4, 16384),
+            majorVersion: versions[0],
+            minorVersion: versions[1]
+        };
```
mvaligursky, line 862: one more comment .. is there a need to store these luts here at all on this? When they get uploaded to textures, there is no need to keep cp copy I assume?

src/scene/forward-renderer.js
```diff
@@ -823,6 +828,143 @@ class ForwardRenderer {
...
+            }
+
+            return ret;
+        }
+
+        // create texture if not already
+        var luts = this._areaLightLuts;
+        if (luts && !luts.ready) {
+            if (luts.majorVersion !== 0 || luts.minorVersion !== 1) {
+                console.warn(`areaLightData version: ${luts.version} is not yet supported!`);
```
mvaligursky, line 922: in this case we should stop processing data below, as it might crash ..

src/scene/forward-renderer.js
raytranuk: hmm - these are views of the asset resource - so it isn't a copy - but probably would prevent the any cleanup of the asset resource - which would be nice to save CPU memory - I will remove this.

src/scene/forward-renderer.js
raytranuk: makes sense - I'll make this more robust

src/scene/forward-renderer.js
mvaligursky: new Float32Array creates a copy of the array

src/scene/forward-renderer.js
mvaligursky: also I'm not sure ${luts.version} exists?

raytranuk: yeah that's a typo :(

raytranuk:
> I've just realized this .. the version should be the first two int16s in the file, not the last? How can you change the file when the version is at fixed offset from start? It would be it impossible to make the file smaller.

I should have realized this - sorry about that - now versions are correctly at the beginning of the binary (like pretty much every other binary format out there...)

src/scene/forward-renderer.js
raytranuk: fixed

src/scene/forward-renderer.js
raytranuk: fixed

PR #2733, opened 2021-01-14
Add video texture example

### LINKED ISSUE #2615: Add video texture example
There is currently no 'engine-only' example illustrating how to implement video textures. This is a tricky topic because of all of the restrictions browser put in place to prevent auto-play of videos. Safari in particular is very restrictive and has become more so over the years. I'm currently unable to get a video to play without an initial user interaction, even if there's not audio track and the video has the `muted` attribute. We should have a tried and tested solution that we knows works as expected across all browsers. And if some browsers have restrictions that can't be circumvented, we should document that.

LeXXik:
Fixes #2615

I had an old sample similar to the open issue, so it was easy to convert. Auto-play tested and works with Android and Windows browsers. Don't have and iPhone (yet) to test it.


mvaligursky:
related to https://github.com/playcanvas/editor/issues/265

@yaustar did some testing. Conclusion: "Doesn't work on Safari desktop nor iOS. No errors in console"

yaustar:
Had a closer look and it could be non issue as it's dependent on how the server is configured (see https://github.com/playcanvas/editor/issues/265#issuecomment-789777652)

On a local python server, this doesn't work but may be fine once it is on github.io.

examples/graphics/video-texture.html
```diff
@@ -0,0 +1,137 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>PlayCanvas Video Texture</title>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0">
+    <link rel="icon" type="image/png" href="../playcanvas-favicon.png" />
+    <script src="../lib/playcanvas.js"></script>
+    <script src="../lib/playcanvas-extras.js"></script>
```
yaustar, line 9: This should refer to the build directory in the repo.
```
    <script src="../../build/playcanvas.js"></script>
    <script src="../../build/playcanvas-extras.js"></script>
```

examples/graphics/video-texture.html
```diff
@@ -0,0 +1,137 @@
...
+        videoTexture.minFilter = pc.FILTER_LINEAR;
+        videoTexture.magFilter = pc.FILTER_LINEAR;
+        videoTexture.addressU = pc.ADDRESS_CLAMP_TO_EDGE;
+        videoTexture.addressV = pc.ADDRESS_CLAMP_TO_EDGE;
+
+        // Grab our HTML element with the video
+        var video = document.getElementById('vid');
+        video.addEventListener('canplay', function (e) {
+            videoTexture.setSource(video);
+        });
```
yaustar, line 82: This will need to be changed slightly to work on iOS. See PR here: https://github.com/playcanvas/developer.playcanvas.com/pull/240

LeXXik: I added it to the asset load callback. If I understood correctly from MDN, it resets the state of the media element and prepares it for playback.

yaustar: Ah right, I see it now. You shouldn't need to load the video asset as the video tag will stream it directly to the player

LeXXik: Right, but I am loading a TV model asset, which is a GLB file containing a mesh that will display the video texture.

yaustar: Gotcha, sounds good!

yaustar:
The reason we haven't merged this yet is we are undecided on the video, not 100% sold on the Bunny vid :)

LeXXik:
No rush :) If you have something I could swap it with, please let me know.

willeastcott:
On reflection, I'm fine with the bunny vid. If necessary, we can switch it later. Let's merge this next week when we put 1.40.0 together.

LeXXik:
@yaustar, you are also good with After Effects (or what did you use to create the Templates promotional video). Perhaps the initial intro part could be cut out and used as a video piece.

yaustar:
I was thinking about that, we actually have a new intro that is in progress that we can change to at a later date :) 

Let's stick with the Bunny vid for now and perhaps change it to something else at a later date

yaustar (approved):

PR #2738, opened 2021-01-15
[AnimComponent] AnimBinder resolver fallback

ellthompson:
If the path to a node provided by an AnimCurve can't be found in the scene, look for the node in the current entities graph.


willeastcott (approved):

src/anim/binder/default-anim-binder.js
```diff
@@ -17,6 +17,17 @@ class DefaultAnimBinder {
 
         if (!graph) return;
 
+        var nodes = { };
+        // cache node names so we can quickly resolve animation paths
+        var flatten = function (node) {
+            nodes[node.name] = node;
+            for (var i = 0; i < node.children.length; ++i) {
+                flatten(node.children[i]);
```
mvaligursky, line 25: this could perhaps log a DEBUG only warning when two nodes has the same name. Or perhaps this could store count or similar way to identify this, and print warning in the code below where this flat array is used to resolve nodes where a node that is looked up this way has multiple nodes

PR #2747, opened 2021-01-18
Remove unused Asset import from i18n

vkalpias:
Fixes https://github.com/playcanvas/engine/issues/2741


willeastcott:
But it is used:

https://github.com/playcanvas/engine/blob/master/src/i18n/i18n.js#L522

vkalpias:
> But it is used:
> 
> https://github.com/playcanvas/engine/blob/master/src/i18n/i18n.js#L522

Indeed

PR #2750, opened 2021-01-18
[Fix] Fix to issue caused by refactoring, causing incorrect call to onPreCull

mvaligursky:


src/scene/forward-renderer.js
```diff
@@ -2852,7 +2852,7 @@ class ForwardRenderer {
             if (!visible.done) {
 
                 if (layer.onPreCull) {
-                    layer.onPreCull(j);
+                    layer.onPreCull(cameraPass);
```
willeastcott, line 2855: I'm guessing this is the only line that constitutes the fix. Is it worth reverting the two other `let` -> `var` edits, which seem unrelated? I'd rather not start peppering the source with the occasional `let` but instead, thoughtfully adopt `let` (and `const`) everywhere (even if only one module at a time).

mvaligursky: that is the only line, but using let in the loop let me make it local variable instead declaring up front, making similar mistakes little bit less likely.

To be honest, all the new / refactored code I'm adding is using const and let, so we will have a mix of the styles for a while, but that should make per module conversion easier in the future? I'd prefer not to have to use var everywhere till whole module gets converted.

willeastcott (approved):

Issue #2651, opened 2020-12-30
Stop engine classes accessing Application.getApplication() to get the Device.


mvaligursky:
Investigate solution for how to replace calls to Application.getApplication() which is used by engine classes to access the Device if none is provided to the class, as it creates circular references.

We can add Device parameter to constructor of those classes, but that changes the API.

mvaligursky:
In some of these cases, the Application is accessed in those classes to get the Device. Many classes take Device in constructor, but some did not initially, and to avoid having to change API, the device is access using the Application. 

two solutions come to my mind:

1. Break the API and pass in the Device. Could break a lot of client code. example: Light & Mesh classes needs Device, but do not get it in constructor.

2. Application.makeTick sets up static variable Application._currentApplication to allow getApplication to work. In a code matching updates of Application._currentApplication, we could perhaps update static variable Device._currentDevice. Then graphics classes could use that instead of Application to access the Device, removing the need to Import the Application, and this would break the circular reference.

PR #2755, opened 2021-01-19
Reduce circular dependencies to Application

vkalpias:
Separated getApplication from the application.js file so that other files do not have to depend on application.js. Added setApplication in the new globals.js file to set the current application.

Reduces circular dependencies to about 6.

Fixes https://github.com/playcanvas/engine/issues/2651


mvaligursky:
Maybe you could do light.js as well?

willeastcott (approved):
Ugh. Not a big fan of this since it doesn't really address the root problem but approving to make Rollup happier.

PR #2756, opened 2021-01-19
LayerComposition refactor to simplify rendering to texture

mvaligursky:
Refactoring of the way LayerComposition builds up a list of cameras / layers to be rendered:
- The overall rendering order is driven by cameras, and not layers. All cameras are sorted by priority, and each camera renders all layers assigned to it in the order specified by layer ordering. When rendering to textures, this makes the engine to switch to each render target exactly one time per frame, instead of switching between render targets when walking layers, which was costing us fill rate heavy restore and resolve operations each time.
- This simplified the way rendering to texture can be set up, as seen in PR with examples for this: https://github.com/playcanvas/engine/pull/2757

**API changes**
- Layer.renderTarget is deprecated (still supported)
- Camera.RenderTarget is made public as preferred way to set up rendering to texture
- LayerComposition constructor takes optional name parameter, useful for debugging of multiple compositions

### DIFF
--- a/src/framework/components/camera/component.js
+++ b/src/framework/components/camera/component.js
@@ -73,6 +73,8 @@ import { PostEffectQueue } from './post-effect-queue.js';
  * the Y-axis). Used for {@link pc.PROJECTION_ORTHOGRAPHIC} cameras only. Defaults to 10.
  * @property {number} priority Controls the order in which cameras are rendered. Cameras
  * with smaller values for priority are rendered first. Defaults to 0.
+ * @property {pc.RenderTarget} renderTarget Render target to which rendering of the cameras
+ * is performed. If not set, it will render simply to the screen.
  * @property {pc.Vec4} rect Controls where on the screen the camera will be rendered in
  * normalized screen coordinates. Defaults to [0, 0, 1, 1].
  * @property {pc.Vec4} scissorRect Clips all pixels which are not in the rectangle.
@@ -176,12 +178,9 @@ class CameraComponent extends Component {
     set priority(newValue) {
         this._priority = newValue;
 
-        var layers = this.layers;
-        for (var i = 0; i < layers.length; i++) {
-            var layer = this.system.app.scene.layers.getLayerById(layers[i]);
-            if (!layer) continue;
-            layer._sortCameras();
-        }
+        // layer composition needs to update order
+        let layerComp = this.system.app.scene.layers;
+        layerComp._dirtyCameras = true;
     }
 
     /**
@@ -235,8 +234,9 @@ class CameraComponent extends Component {
         var layers = this.layers;
         for (var i = 0; i < layers.length; i++) {
             var layer = this.system.app.scene.layers.getLayerById(layers[i]);
-            if (!layer) continue;
-            layer.addCamera(this);
+            if (layer) {
+                layer.addCamera(this);
+            }
         }
     }
 
@@ -244,8 +244,9 @@ class CameraComponent extends Component {
         var layers = this.layers;
         for (var i = 0; i < layers.length; i++) {
             var layer = this.system.app.scene.layers.getLayerById(layers[i]);
-            if (!layer) continue;
-            layer.removeCamera(this);
+            if (layer) {
+                layer.removeCamera(this);
+            }
         }
     }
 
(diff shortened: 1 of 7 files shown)


src/framework/components/camera/component.js
```diff
@@ -176,12 +178,9 @@ class CameraComponent extends Component {
...
         this._priority = newValue;
 
-        var layers = this.layers;
-        for (var i = 0; i < layers.length; i++) {
-            var layer = this.system.app.scene.layers.getLayerById(layers[i]);
-            if (!layer) continue;
-            layer._sortCameras();
-        }
+        // layer composition needs to update order
+        let layerComp = this.system.app.scene.layers;
```
slimbuck, line 182: tiny point, but no need for layerComp variable.

slimbuck (approved):
Love it! Approving, but please get more approvals before merge.

src/framework/components/camera/component.js
mvaligursky: I agree, but I added it to make the code more readable as it's not clear from naming that scene.layers is LayerComposition.

Maksims:
Just thinking through API change, here is case scenario of how I've used previously Layers for doing some effects:

Refraction: Have Refraction Layer, to which some entities were added (to be refracted), then this layer renders to texture. Which is then available for all the remaining layers to be rendered. So it was fairly simple to setup and manage, and easy to do shader stuff with it.

How it will be now? Will it be required to create layer with Refraction as before, but then render it in separate camera, which is set to render to texture? Sounds like extra unnecessary step to make in this case.

mvaligursky:
@Maksims You can still use old API, but if you render multiple layers into the refraction, you'd get better performance to use extra camera. So:
Create Refraction camera, assign render target to it, and set it to render World layer (no new layer needed). This is the same as you'd render reflection texture .. create a camera that is plane reflected, and use that to render to texture.

Maksims:
But this will not benefit from depth test, as geometry that is rendered in Refraction layer, already rendered. In two cameras, they will have own depth buffer, right?
I've actually used same approach for reflection as well, where I've flipped camera using events of before and after layer rendering. Instead of separate camera. It was simple small piece of code, no new entities, nothing else.

mvaligursky:
If you set up both cameras to use the same render target (or null for framebuffer), they'd render to the same target and will benefit from depth being already rendered by the first camera, so that should be no problem.

Maksims:
> If you set up both cameras to use the same render target (or null for framebuffer), they'd render to the same target and will benefit from depth being already rendered by the first camera, so that should be no problem.

Only refraction/reflection render should be in texture, other renders into back buffer.

mvaligursky:
I'm not entirely sure what scenario you are talking about. Are you thinking about using single depth buffer when rendering into framebuffer and then use it again when rendering to refraction texture? If so, I'm not sure that was supported before? I looked at After The Flood few days ago, and that seems to have separate depth passes for both reflection, refraction and main cameras. Could you please give me more detailed example of how the rendering of those reflection / refraction layers you were doing to see how this could map to this new recommended way to set up rendering to textures.

This is how I would expect general water rendering to be setup:

- we'd have 2 layers for geometry object, World and Refraction.
- each geometry object would be in exactly one of these layers
- there will be RefractionCamera, rendering to Refraction RT, and only objects on Refraction Layer
- similarly, ReflectionCamera will render to Reflection RT, objects in World Layer (and maybe Refraction layer as well if they're not clean cut by the water level)
- main Camera will then render World and again maybe Refraction layer, using both Reflection and Refraction textures as inputs.
- all these 3 RTs (Reflection, Refraction, framebuffer) would use separate RT (so separate color and depth buffers), and can do depth prepass into depth if needed

src/scene/layer-composition.js
```diff
@@ -24,9 +25,11 @@ import { RenderAction } from './render-action.js';
  */
 // Composition can hold only 2 sublayers of each layer
 class LayerComposition extends EventHandler {
-    constructor() {
+    constructor(name = "Untitled") {
```
willeastcott, line 28: So this is the default name for `GraphNode`s and `Material`s too. But other classes like `Texture` default to `null`. I kind of which I'd chosen `null` or empty string for `GraphNode` and `Material` TBH. Or at least enforced a case - probably all lower case. Kinda too late to rename the default for `GraphNode` and `Material` now though.... 😢

willeastcott (approved):

PR #2757, opened 2021-01-19
Added render-to-cubemap, refactored render-to-texture engine examples

### LINKED ISSUE #254: Create a mechanism to dynamically render to cubemaps
Say I want to dynamically update a cubemap. It would be good if I could create six camera that render to offscreen buffers that are the 6 faces of a cubemap and then use that cubemap in the same frame. I guess the only thing the engine is missing for this is the ability to specify which mesh instances are rendered by which cameras? So layers?

mvaligursky:
- new engine example RenderToCubemap, which shows rendering of the scene dynamically to a cubemap. Part of this is separate script component cubemapRenderer which can be used in Editor as well
cubemap small
Example (when engine is released): http://playcanvas.github.io/#graphics/render-to-cubemap.html

- RenderToTexture example updated to use skybox instead of plain clean color to better test layers setup for rendering to texture.
Screen Shot 2020-11-03 at 11 39 11 AM

Fixes #254


examples/assets/scripts/utils/cubemap-renderer.js
```diff
@@ -0,0 +1,106 @@
...
+    camera.enabled = false;
+
+    // Create cubemap render target with specified resolution and mipmap generation
+    this.cubeMap = new pc.Texture(this.app.graphicsDevice, {
+        width: this.resolution,
+        height: this.resolution,
+        format: pc.PIXELFORMAT_R8_G8_B8_A8,
+        cubemap: true,
+        mipmaps: this.mipmaps,
+        autoMipmap: this.mipmaps,
```
willeastcott, line 51: Isn't `autoMipmap` deprecated/internal/ignored now?

examples/graphics/render-to-texture.html
```diff
@@ -69,86 +77,112 @@
             format: pc.PIXELFORMAT_R8_G8_B8,
             autoMipmap: true,
```
willeastcott, line 78: Isn't `autoMipmap` deprecated/internal/ignored now?

examples/assets/scripts/utils/cubemap-renderer.js
```diff
@@ -0,0 +1,106 @@
...
+            priority: camera.priority,
+
+            // copy other camera properties
+            clearColor: camera.clearColor,
+            clearColorBuffer: camera.clearColorBuffer,
+            clearDepthBuffer: camera.clearDepthBuffer,
+            clearStencilBuffer: camera.clearStencilBuffer,
+            farClip: camera.farClip,
+            nearClip: camera.nearClip,
+            frustumCulling: camera.frustumCulling
```
willeastcott, line 94: Should this always be `true` maybe?

examples/assets/scripts/utils/cubemap-renderer.js
```diff
@@ -0,0 +1,106 @@
...
+            // priority
+            priority: camera.priority,
+
+            // copy other camera properties
+            clearColor: camera.clearColor,
+            clearColorBuffer: camera.clearColorBuffer,
+            clearDepthBuffer: camera.clearDepthBuffer,
+            clearStencilBuffer: camera.clearStencilBuffer,
+            farClip: camera.farClip,
+            nearClip: camera.nearClip,
```
willeastcott, line 93: Might the user want a different set of clip planes vs the main camera?

examples/assets/scripts/utils/cubemap-renderer.js
```diff
@@ -0,0 +1,106 @@
...
+    description: 'Resolution of one side of a cubemap. Use power of 2 resolution if you wish to use Mipmaps.',
+    type: 'number',
+    default: 64
+});
+
+CubemapRenderer.attributes.add('mipmaps', {
+    title: 'Mipmaps',
+    description: 'If set to true, mipmaps will be allocated and autogenerated.',
+    type: 'boolean',
+    default: false
```
willeastcott, line 21: Maybe default to `true`?

examples/assets/scripts/utils/cubemap-renderer.js
```diff
@@ -0,0 +1,106 @@
...
+//   Note: the layers should contain all layers visible by cybemap camera.
+// - to use generated cube map, you can access it like this using script:
+//   material.cubeMap = entity.script.cubemapRenderer.cubeMap;
+
+var CubemapRenderer = pc.createScript('cubemapRenderer');
+
+CubemapRenderer.attributes.add('resolution', {
+    title: 'Resolution',
+    description: 'Resolution of one side of a cubemap. Use power of 2 resolution if you wish to use Mipmaps.',
+    type: 'number',
```
willeastcott, line 13: I'd consider making this an enum with:
* 16x16
* 32x32
* 64x64
* 128x128
* 256x256
* 512x512
* 1024x1024
* 2048x2048

examples/assets/scripts/utils/cubemap-renderer.js
```diff
@@ -0,0 +1,106 @@
...
+        });
+
+        // create a child entity with the camera for this face
+        var e = new pc.Entity("CubeMapCamera_" + i);
+        e.addComponent('camera', {
+            aspectRatio: 1,
+            fov: 90,
+
+            // cubemap will render all layers as setup on Entity's camera
+            layers: camera.layers.slice(),
```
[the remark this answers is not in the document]
willeastcott: You don't. 😆  It's redundant.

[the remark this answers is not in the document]
willeastcott: See here for why: https://github.com/playcanvas/engine/blob/master/src/scene/camera.js#L223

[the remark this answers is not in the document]
mvaligursky: here's the reason: https://github.com/playcanvas/engine/blob/6d37be6b40660648a5cee2939fe66cc973204da2/src/framework/components/camera/component.js#L104-L107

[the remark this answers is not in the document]
willeastcott: Haha, so I think those docs aren't quite right. At least, not according to the behavior of the code.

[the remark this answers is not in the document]
willeastcott: It should say that `To update layers, don't get and edit the array. Always set the layers property to trigger an update.`

examples/assets/scripts/utils/cubemap-renderer.js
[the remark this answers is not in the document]
mvaligursky: you're right, it's the same as mipmaps. deleting the line.

examples/assets/scripts/utils/cubemap-renderer.js
mvaligursky: sounds good

examples/assets/scripts/utils/cubemap-renderer.js
mvaligursky: It feels like this would limit the options for no good reason.

willeastcott: Well, the good reason is that it prevents users from setting `100` or `8192`, say.

willeastcott: An alternative is you could:
1. snap to the closest power of two
2. clamp between min and max cubemap resolutions supported by the graphics device

mvaligursky: setting 100 is perfectly valid though? And setting it to out of range number should be responsibility of the Texture to take care of and clamp to the platform limits I think, as this class simply creates a Texture.
I tried 100 and that works fine. I also tried 15000 and that ended up with out of video memory error :)

willeastcott: Ahhh, my mistake. I thought cubemaps had to be power of two, even if not mipmapped. But yeah, I guess that's wrong.

But on a device that only supports, say, 512 max cubemap resolution, if you set it to 2048, the engine doesn't just handle that for you, does it?

mvaligursky: I added handling of maxTextureSize .. but perhaps we should handle this in Texture's constructor as well?

willeastcott: Possibly. But resolving this for now.

examples/assets/scripts/utils/cubemap-renderer.js
```diff
@@ -0,0 +1,108 @@
+// How to use in the Editor:
+// - create entity with Camera component - position of the entity defines where the cubemap is rendered from
+//   and properties of the Camera are used to render cubemap (adjust near / far distance, clearing, layers and other properties)
+//   Note: the layers should contain all layers visible by cybemap camera.
```
willeastcott, line 4:
```suggestion
//   Note: the layers should contain all layers visible by cubemap camera.
```

examples/assets/scripts/utils/cubemap-renderer.js
```diff
@@ -0,0 +1,108 @@
...
+    if (!camera) {
+        console.error("CubemapRenderer component requires Camera component to be created on the Entity.");
+        return;
+    }
+
+    // disable camera component, as it's used only as a source of properties
+    camera.enabled = false;
+
+    // limit maximum texture size
+    var resolution = Math.min(this.resolution, this.app.graphicsDevice.maxTextureSize);
```
willeastcott, line 45:
```suggestion
    var resolution = Math.min(this.resolution, this.app.graphicsDevice.maxCubeMapSize);
```

examples/assets/scripts/utils/cubemap-renderer.js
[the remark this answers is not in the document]
willeastcott: Soooo, are you gonna remove the redundant `slice`? We can tighten up the docs for `pc.CameraComponent#layers` in a separate PR.

examples/graphics/render-to-cubemap.html
```diff
@@ -0,0 +1,230 @@
...
+        var light = new pc.Entity();
+        light.addComponent("light", {
+            type: "directional",
+            color: new pc.Color(1, 1, 1),
+            range: 40,
+            castShadows: true,
+            layers: [worldLayer.id],
+            shadowBias: 0.2,
+            normalOffsetBias: 0.05,
+            shadowDistance: 25,
```
willeastcott, line 207:
```suggestion
            shadowDistance: 25
```

examples/graphics/render-to-texture.html
```diff
@@ -69,86 +77,112 @@
...
+
+            // set the priority of textureCamera to lower number than the priority of the main camera (which is at default 0)
+            // to make it rendered first each frame
+            priority: -1
         });
-        // add sphere at the position of tv camera to see it in the world
-        tvCamera.addComponent("model", {
+
+        // this camera renders into texture target
+        textureCamera.camera.renderTarget = renderTarget;
```
willeastcott, line 121: Can this be supplied in the `addComponent` options?

examples/assets/scripts/utils/cubemap-renderer.js
mvaligursky: thanks, I didn't even know about this one!

examples/assets/scripts/utils/cubemap-renderer.js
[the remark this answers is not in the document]
mvaligursky: I'm not sure the slice is redundant.
We have source camera with Layers, which is an array
and I'm assigning it to Layers of another camera on this line
without slice, both cameras will have a single instance of the array .. so if we change its one camera will change its content, it will change for the other camera as well
as this is code of setter of layer:

```
set layers(newValue) {
        this._camera.layers = newValue;
}
```

so I'd prefer to use slice to have two separate arrays.

but maybe a better solution would be to slice array inside set layers.

examples/graphics/render-to-texture.html
mvaligursky: done, thanks.

examples/assets/scripts/utils/cubemap-renderer.js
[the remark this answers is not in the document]
willeastcott: No, scroll up to my earlier comments on this where I link to the line of code that show `Camera` doing its own slice of the supplied array.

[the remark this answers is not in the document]
mvaligursky: nice one, you looked one layer deeper than me. Removed the slice now.

examples/assets/scripts/utils/cubemap-renderer.js
```diff
@@ -0,0 +1,108 @@
...
+            clearColorBuffer: camera.clearColorBuffer,
+            clearDepthBuffer: camera.clearDepthBuffer,
+            clearStencilBuffer: camera.clearStencilBuffer,
+            farClip: camera.farClip,
+            nearClip: camera.nearClip,
+            frustumCulling: camera.frustumCulling
+        });
+
+        // this camera renders into texture target
+        e.camera.camera.renderTarget = renderTarget;
```
willeastcott, line 100: Shouldn't this be `e.camera.renderTarget`? Also, can you support setting the render target in the `addComponent` options?

examples/graphics/render-to-cubemap.html
```diff
@@ -0,0 +1,230 @@
...
+            fov: 60,
+            layers: [worldLayer.id, excludedLayer.id, skyboxLayer.id]
+        });
+        app.root.addChild(camera);
+
+        // Create an Entity with a directional light component
+        var light = new pc.Entity();
+        light.addComponent("light", {
+            type: "directional",
+            color: new pc.Color(1, 1, 1),
```
willeastcott, line 201:
```suggestion
            color: pc.Color.WHITE,
```

examples/graphics/render-to-texture.html
```diff
@@ -69,86 +77,112 @@
...
         });
-        app.root.addChild(tvCamera);
+        app.root.addChild(textureCamera);
 
-        // Create an Entity with a point light component and add it to both layers
+        // Create an Entity with a point light component and add it to world layer (and so used by both cameras)
         var light = new pc.Entity();
         light.addComponent("light", {
             type: "point",
             color: new pc.Color(1, 1, 1),
```
willeastcott, line 133:
```suggestion
            color: pc.Color.WHITE,
```

examples/assets/scripts/utils/cubemap-renderer.js
mvaligursky: yes and yes, done, thanks

examples/graphics/render-to-cubemap.html
mvaligursky: changed it to YELLOW to give more happy feel ;)

willeastcott (approved):

PR #2764, opened 2021-01-20
Remove "pc." from API docs

vkalpias:
Goes together with this PR https://github.com/playcanvas/playcanvas-jsdoc-template/pull/26


willeastcott (approved):

willeastcott:
I would actually prefer that we fixed the .d.ts generation _without_ aliasing any class names. I'm convinced we can hack around this. I don't want to have to make snap API decisions just to get this merged.

vkalpias:
Use this PR instead https://github.com/playcanvas/engine/pull/2876

Issue #2753, opened 2021-01-19
Small improvements to RenderComponent


mvaligursky:
This works:

```
node.addComponent('render');
node.render.meshInstances = [ meshInstance ];
```

but this does not
```
node.addComponent('render', {
    meshInstances: [ meshInstance ]
});
```

and perhaps even try to remove the need to pass node (entity) to MeshInstance Constructor. So instead of

```
var entity = new pc.Entity();
var meshInstance = new pc.MeshInstance(entity, mesh, material);
entity.addComponent('render');
entity.render.meshInstances = [ meshInstance ];
entity.setLocalPosition(cursor, 0, 0);
this.entity.addChild(entity);
```

this could be used:
```
var meshInstance = new pc.MeshInstance(mesh, material);
var entity = new pc.Entity();
entity.addComponent('render', {
    meshInstances: [ meshInstance ]
});
entity.setLocalPosition(cursor, 0, 0);
this.entity.addChild(entity);
```

mvaligursky:
currently we have this to enable wireframe
```
render.generateWireframe();
for (var i = 0; i < render.meshInstances.length; i++) {
   render.meshInstances[i].renderStyle = pc.RENDERSTYLE_WIREFRAME;
}
```

we could have **RenderComponent#renderStyle** to hide this

PR #2765, opened 2021-01-20
New interface for RenderComponent to handle wireframe and point render style

mvaligursky:
API changes:

```javascript
RenderComponent#generateWireframe // removed this private interface
RenderComponent#renderStyle       // new property accepting RENDERSTYLE_XXX constants.
```

When different render styles exist on the Mesh, they get automatically updated when mesh updates - so that procedurally updated geometry can render in wireless without user having to manually update it.

addresses part of https://github.com/playcanvas/engine/issues/2753

https://user-images.githubusercontent.com/59932779/105339697-e51cdb00-5bd4-11eb-98db-f5c0163870af.mov

Screen Shot 2021-01-20 at 6 35 14 PM

Screen Shot 2021-01-20 at 6 34 49 PM

### DIFF
--- a/src/scene/mesh-instance.js
+++ b/src/scene/mesh-instance.js
@@ -151,6 +151,27 @@ class MeshInstance {
         this.flipFaces = false;
     }
 
+    static _meshSet = new Set();
+
+    // generates wireframes for an array of mesh instances
+    static _prepareRenderStyleForArray(meshInstances, renderStyle) {
+
+        for (let i = 0; i < meshInstances.length; i++) {
+
+            // switch mesh instance to the requested style
+            meshInstances[i].renderStyle = renderStyle;
+
+            // process all unique meshes
+            let mesh = meshInstances[i].mesh;
+            if (!MeshInstance._meshSet.has(mesh)) {
+                MeshInstance._meshSet.add(mesh);
+                mesh.prepareRenderState(renderStyle);
+            }
+        }
+
+        MeshInstance._meshSet.clear();
+    }
+
     get mesh() {
         return this._mesh;
     }
--- a/src/scene/mesh-instance.js
+++ b/src/scene/mesh-instance.js
@@ -15,6 +15,8 @@ import {
 var _tmpAabb = new BoundingBox();
 var _tempBoneAabb = new BoundingBox();
 var _tempSphere = new BoundingSphere();
+var _meshSet = new Set();
+
 
 // internal data structure used to store data used by hardware instancing
 class InstancingData {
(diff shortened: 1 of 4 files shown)


src/scene/model.js
```diff
@@ -188,24 +188,7 @@ class Model {
...
-            }
-        }
-
-        for (i = 0; i < meshes.length; ++i) {
-            mesh = meshes[i];
-            if (!mesh.primitive[RENDERSTYLE_WIREFRAME]) {
-                mesh.generateWireframe();
-            }
-        }
+        MeshInstance._prepareRenderStyleForArray(this.meshInstances, RENDERSTYLE_WIREFRAME);
```
willeastcott, line 191: Does this function need to be public any more now that we have `renderStyle` property?

mvaligursky: I only added new renderStyle property to render component, but have not changes API on model component .. so this should stay?

willeastcott: LOL, sorry, I thought this was on the RenderComponent for some reason. Please ignore.

src/scene/mesh-instance.js
```diff
@@ -151,6 +151,27 @@ class MeshInstance {
         this.flipFaces = false;
     }
 
+    static _meshSet = new Set();
```
willeastcott, line 154: Any reason not to make this a module global property? That way, it's truly private.

mvaligursky: good point, I guess from other languages I'm used to keeping stuff inside classes.
I'll try and see if syntax with # works? https://stackoverflow.com/questions/22156326/private-properties-in-javascript-es6-classes

Maksims: # syntax for private properties is very new in ECMA. It needs to be investigated how it compiles to both targets by Babel.

mvaligursky: this is what babel generates for both es5 and module build:

```
// helpers

var id$3 = 0;

function _classPrivateFieldLooseKey(name) {
  return "__private_" + id$3++ + "_" + name;
}

function _classPrivateFieldLooseBase(receiver, privateKey) {
  if (!Object.prototype.hasOwnProperty.call(receiver, privateKey)) {
    throw new TypeError("attempted to use private field on non-instance");
  }

  return receiver;
}

// variable declaration
var _meshSet = _classPrivateFieldLooseKey("meshSet");

// variable use
_classPrivateFieldLooseBase(MeshInstance, _meshSet)[_meshSet].add(mesh);

```

So that does not feel ideal for performance critical code because of the extra hasOwnProperty call on each access to it to validate access. For less performance critical code, this should be fine?

willeastcott: Yeah. A bit ugly though. Personally, I would just make it a module global. 😆

mvaligursky: agreed and done.

willeastcott (approved):

PR #2772, opened 2021-01-22
[Animation] Binding updates and bug fixes

### LINKED ISSUE #2751: Since 1.38.0, Anim state graph transitions duration are not respected. Transitions are instant
Example project: https://playcanvas.com/editor/scene/995061

There's a blend between animations on engine 1.37.1 https://launch.playcanvas.com/995061?debug=true&use_local_engine=https://code.playcanvas.com/playcanvas-1.37.1.js

But are instant on 1.38.3: https://launch.playcanvas.com/995061?debug=true

### LINKED ISSUE #2771: [AnimComponent] Support rebinding animation curves
There will be times when a user may want to swap out an entities model for a new one while maintaining the current animation state of that entity. This would require that current animation curves in the AnimEvaluator are rebound to the new hierarchy. 

A typical use case for this would be a character customisation screen where an entities animation must be maintained while it's model is constantly updated.

### LINKED ISSUE #2786: Issue with assignAnimation method of AnimController class
`assignAnimation` of `AnimController` passes `path` array to `state.addAnimation` instead of an actual path string:

https://github.com/playcanvas/engine/blob/cb4d0ddb59da9616e8d54d0a1007cebcf35cff5a/src/anim/controller/anim-controller.js#L417

It should probably pass `path[0]` instead.

This causses an issue when trying to override animation for a specific state as this condition will never be met:

https://github.com/playcanvas/engine/blob/6fe0240780b50b0a7fc5c9b1f8927e2718dc1a2a/src/anim/controller/anim-state.js#L45

P.S. This seems like a situation where using Typescript would prove useful :)

### LINKED ISSUE #2788: pc.AnimComponent#removeNodeAnimations wrong jsdoc
Name tag should be `removeNodeAnimations` instead of  `removeStateAnimations`:

https://github.com/playcanvas/engine/blob/cb4d0ddb59da9616e8d54d0a1007cebcf35cff5a/src/framework/components/anim/component.js#L217

ellthompson:
- Adds the rebind function to the anim component and component layers to allow for any currently playing animations to rebind using current entity properties.
- Uses absolute scene paths when storing AnimTargets in the AnimEvaluator to ensure the correct blending and assignment of animations even when animation curve paths mismatch the scene structure.
- Fixes jsdocs for the AnimComponent
- Fixes the AnimState.addAnimation function to overwrite existing animations when the same path is used

**New API:**
```javascript
// Rebind any animations in the layer to the currently present components and model of the anim components entity
pc.AnimComponentLayer#rebind: function ();
// Rebind all of the components layers
pc.AnimComponent#rebind: function ();
```

Fixes #2771 
Fixes #2751
Fixes #2788
Fixes #2786

### DIFF
--- a/src/anim/binder/anim-binder.js
+++ b/src/anim/binder/anim-binder.js
@@ -51,41 +51,22 @@ class AnimBinder {
      * @private
      * @static
      * @function
-     * @name pc.AnimPropertyLocator#encode
+     * @name pc.AnimBinder#encode
      * @description Converts a locator array into its string version
-     * @param {Array} locator - The property location in the scene defined as an array
+     * @param {string|Array} entityPath - The entity location in the scene defined as an array or string path
+     * @param {string} component - The component of the entity the property is located under
+     * @param {string|Array} propertyPath - The property location in the entity defined as an array or string path
      * @returns {string} The locator encoded as a string
      * @example
      * // returns 'spotLight/light/color.r'
      * encode({ entityPath: ['spotLight'], component: 'light', propertyPath: ['color', 'r'] });
      */
-    static encode(locator) {
-        return AnimBinder.joinPath([
-            AnimBinder.joinPath(Array.isArray(locator.entityPath) ? locator.entityPath : [locator.entityPath]),
-            locator.component,
-            AnimBinder.joinPath(Array.isArray(locator.propertyPath) ? locator.propertyPath : [locator.propertyPath])
-        ], '/');
-    }
-
-    /**
-     * @private
-     * @static
-     * @function
-     * @name pc.AnimPropertyLocator#decode
-     * @description Converts a locator string into its array version
-     * @param {Array} locator - The property location in the scene defined as a string
-     * @returns {Array} - The locator decoded into an array
-     * @example
-     * // returns { entityPath: ['spotLight'], component: 'light', propertyPath: ['color', 'r'] }
-     * decode('spotLight/light/color.r');
-     */
-    static decode(locator) {
-        var locatorSections = AnimBinder.splitPath(locator, '/');
-        return {
-            entityPath: AnimBinder.splitPath(locatorSections[0]),
-            component: locatorSections[1],
-            propertyPath: AnimBinder.splitPath(locatorSections[2])
-        };
+    static encode(entityPath, component, propertyPath) {
+        return `${
+            Array.isArray(entityPath) ? AnimBinder.joinPath(entityPath) : entityPath
+        }/${component}/${
+            Array.isArray(propertyPath) ? AnimBinder.joinPath(propertyPath) : propertyPath
+        }`;
     }
 
     /**
--- a/src/anim/binder/anim-binder.js
+++ b/src/anim/binder/anim-binder.js
@@ -59,13 +59,13 @@ class AnimBinder {
      * @returns {string} The locator encoded as a string
      * @example
      * // returns 'spotLight/light/color.r'
-     * encode({ entityPath: ['spotLight'], component: 'light', propertyPath: ['color', 'r'] });
+     * encode(['spotLight'], 'light', ['color', 'r']);
      */
     static encode(entityPath, component, propertyPath) {
         return `${
-            Array.isArray(entityPath) ? AnimBinder.joinPath(entityPath) : entityPath
+            Array.isArray(entityPath) ? entityPath.join('/') : entityPath
         }/${component}/${
-            Array.isArray(propertyPath) ? AnimBinder.joinPath(propertyPath) : propertyPath
+            Array.isArray(propertyPath) ? propertyPath.join('/') : propertyPath
         }`;
     }
 
--- a/src/anim/binder/default-anim-binder.js
+++ b/src/anim/binder/default-anim-binder.js
@@ -128,8 +128,11 @@ class DefaultAnimBinder {
         };
     }
 
-    findNode(graph, path) {
-        var node = graph.findByPath(path.entityPath);
+    findNode(path) {
+        var node;
+        if (this.graph) {
+            node = this.graph.findByPath(path.entityPath);
+        }
         if (!node) {
             node = this.nodes[path.entityPath[path.entityPath.length - 1] || ""];
 
@@ -154,7 +157,7 @@ class DefaultAnimBinder {
         var target = this.targetCache[encodedPath];
         if (target) return target;
 
-        var node = this.findNode(this.graph, path);
+        var node = this.findNode(path);
         if (!node) {
             return null;
         }
--- a/src/anim/binder/default-anim-binder.js
+++ b/src/anim/binder/default-anim-binder.js
@@ -184,8 +184,7 @@ class DefaultAnimBinder {
         if (path.component !== 'graph')
             return;
 
-        var entityPathArr = AnimBinder.splitPath(path.entityPath, '/');
-        var node = this.nodes[entityPathArr[entityPathArr.length - 1] || ""];
+        var node = this.nodes[path.entityPath[path.entityPath.length - 1] || ""];
 
         this.nodeCounts[node.path]--;
         if (this.nodeCounts[node.path] === 0) {
--- a/src/anim/binder/default-anim-binder.js
+++ b/src/anim/binder/default-anim-binder.js
@@ -56,7 +56,12 @@ class DefaultAnimBinder {
                 var func = function (value) {
                     object.set.apply(object, value);
                 };
-                return new AnimTarget(func, 'vector', 3);
+                var targetIdentifier = AnimBinder.encode({
+                    entityPath: node.path,
+                    component: 'entity',
+                    propertyPath: 'localPosition'
+                });
+                return new AnimTarget(func, 'vector', 3, targetIdentifier);
             },
 
             'localRotation': function (node) {
@@ -64,7 +69,12 @@ class DefaultAnimBinder {
                 var func = function (value) {
                     object.set.apply(object, value);
                 };
-                return new AnimTarget(func, 'quaternion', 4);
+                var targetIdentifier = AnimBinder.encode({
+                    entityPath: node.path,
+                    component: 'entity',
+                    propertyPath: 'localRotation'
+                });
+                return new AnimTarget(func, 'quaternion', 4, targetIdentifier);
             },
 
             'localScale': function (node) {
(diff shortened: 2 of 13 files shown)


src/anim/binder/default-anim-binder.js
```diff
@@ -56,23 +56,38 @@ class DefaultAnimBinder {
                 var func = function (value) {
                     object.set.apply(object, value);
                 };
-                return new AnimTarget(func, 'vector', 3);
+                var targetIdentifier = AnimBinder.encode({
+                    entityPath: node.path,
+                    component: 'entity',
+                    propertyPath: 'localPosition'
+                });
+                return new AnimTarget(func, 'vector', 3, targetIdentifier);
```
mvaligursky, line 64: these get repeated bellow few times .. perhaps wrap it in a function?

src/anim/evaluator/anim-evaluator.js
```diff
@@ -238,6 +236,16 @@ class AnimEvaluator {
         return null;
     }
 
+    rebind() {
+        this._binder.rebind();
+        this._targets = {};
+        var clips = [...this.clips];
```
mvaligursky, line 242: please make sure this transpiles to something acceptable in es5 .. I don't think we've tested it yet.

src/framework/components/anim/component-binder.js
```diff
@@ -87,16 +100,37 @@ class AnimComponentBinder extends DefaultAnimBinder {
                 if (!propertyComponent) {
                     var entityPath = AnimBinder.splitPath(propertyLocation.entityPath[0], '/');
                     propertyComponent = this.nodes[entityPath[entityPath.length - 1] || ""];
+
+                    // #ifdef DEBUG
+                    var fallbackGraphPath = (entityPath[entityPath.length - 1] || "") + '/' + propertyLocation.propertyPath;
+                    if (this._visitedFallbackGraphPaths[fallbackGraphPath]) {
+                        console.warn('Multiple nodes with the path ' + fallbackGraphPath + ' are present in the entities graph which may result in the incorrect binding of animations');
```
mvaligursky, line 107: if you can easily print the name of the entity the anim component is on, that would be useful as well

src/anim/evaluator/anim-evaluator.js
ellthompson: Yep i've tested this using `rollup -c --watch --environment target:es5`

src/framework/components/anim/component-binder.js
ellthompson: Sure i'll add that in

src/anim/binder/default-anim-binder.js
ellthompson: 👍

src/anim/binder/default-anim-binder.js
```diff
@@ -124,28 +128,48 @@ class DefaultAnimBinder {
...
+        return node;
+    }
+
+    static createAnimTarget(func, type, valueCount, node, propertyPath, componentType) {
+        var targetPath = AnimBinder.encode(node.path, componentType ? componentType : 'entity', propertyPath);
+        return new AnimTarget(func, type, valueCount, targetPath);
+    }
+
+    resolve(path) {
+        var target = this.targetCache[AnimBinder.encode(path.entityPath, path.component, path.propertyPath)];
```
slimbuck, line 153: you can cache the `encode` result here and reuse it down below.

src/anim/binder/anim-binder.js
```diff
@@ -51,41 +51,22 @@ class AnimBinder {
...
-    static decode(locator) {
-        var locatorSections = AnimBinder.splitPath(locator, '/');
-        return {
-            entityPath: AnimBinder.splitPath(locatorSections[0]),
-            component: locatorSections[1],
-            propertyPath: AnimBinder.splitPath(locatorSections[2])
-        };
+    static encode(entityPath, component, propertyPath) {
+        return `${
+            Array.isArray(entityPath) ? AnimBinder.joinPath(entityPath) : entityPath
```
slimbuck, line 66: we no longer have need to decode paths that were previously encoded (right?). The purpose of this encode function is to create a unique key for the given path. In this  case we don't need to be careful with escaping characters etc. So presumably `AnimBinder.joinPath` (with fancy regex etc) can be removed and this call can be replaced with a simple `array.join('/')`?

src/framework/components/anim/component-binder.js
```diff
@@ -67,36 +67,50 @@ class AnimComponentBinder extends DefaultAnimBinder {
     }
 
     resolve(path) {
-        var propertyLocation = AnimBinder.decode(path);
-
-        var entity = this._getEntityFromHierarchy(propertyLocation.entityPath);
-
-        if (!entity)
-            return null;
+        var target = this.targetCache[AnimBinder.encode(path.entityPath, path.component, path.propertyPath)];
```
slimbuck, line 70: ditto here, cache `encode` result here and use it in one of the cases below

src/anim/binder/default-anim-binder.js
```diff
@@ -124,28 +128,48 @@ class DefaultAnimBinder {
         };
     }
 
-    resolve(path) {
-        var propertyLocation = AnimBinder.decode(path);
-
-        var node = this.graph.findByPath(propertyLocation.entityPath[0]);
+    findNode(graph, path) {
+        var node = graph.findByPath(path.entityPath.join('/'));
```
slimbuck, line 132: it might be worth adding a `findByPath` function to graph-node which takes the array and uses it directly instead of joining into a string here (which won't handle escapes properly) just to split in the function. @willeastcott would you be ok adding a function to `GraphNode` or so we keep such a function here in `default-anim-binder`?

ellthompson: I've updated the graph.findByPath function to support arrays too.

src/framework/components/anim/component-binder.js
ellthompson: 👍

src/anim/binder/anim-binder.js
ellthompson: Cheers, updated 👍

src/framework/components/anim/component-binder.js
```diff
@@ -67,36 +67,51 @@ class AnimComponentBinder extends DefaultAnimBinder {
...
             case 'graph':
-                if (entity.model && entity.model.model && entity.model.model.graph) {
-                    propertyComponent = entity.model.model.graph.findByPath(propertyLocation.entityPath[0]);
-                }
-                if (!propertyComponent) {
-                    var entityPath = AnimBinder.splitPath(propertyLocation.entityPath[0], '/');
-                    propertyComponent = this.nodes[entityPath[entityPath.length - 1] || ""];
-                }
+                entity = this.animComponent.entity;
+                var graph = entity.model && entity.model.model && entity.model.model.graph;
```
willeastcott, line 90: You could potentially use optional chaining here:

```javascript
                var graph = entity.model?.model?.graph;
```

src/framework/components/anim/component-binder.js
```diff
@@ -67,36 +67,51 @@ class AnimComponentBinder extends DefaultAnimBinder {
...
+                propertyComponent = this.findNode(graph, path);
```
willeastcott, line 91: Should you be checking that `graph` is valid before doing a `findNode` op?

src/framework/components/anim/component-binder.js
```diff
@@ -240,10 +255,31 @@ class AnimComponentBinder extends DefaultAnimBinder {
...
+        return new AnimTarget(setter, animDataType, animDataComponents, targetPath);
+    }
+
+    rebind() {
+        this.targetCache = {};
+        // #ifdef DEBUG
+        this.visitedFallbackGraphPaths = {};
+        // #endif
+        var entity = this.animComponent && this.animComponent.entity;
+        var graph = entity.model && entity.model.model && entity.model.model.graph;
```
willeastcott, line 270: Again, you could potentially use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here:

```javascript
                var graph = entity.model?.model?.graph;
```

willeastcott: Although I'm a bit confused about the logic here. The code checks there's an `animComponent` property without assuming it's existence. Which indicates maybe `entity` could end up as `undefined`. But `entity` is assumed to be defined on the following line.

willeastcott (approved):
Approving with a few minor comments.

src/framework/components/anim/component-binder.js
mvaligursky: I see this handles entity.model ... and I hope there's handling of entity.render as well for the new render component?

mvaligursky:
Please make sure render component is supported in addition to a model component.

ellthompson:
> Please make sure render component is supported in addition to a model component.

I've tested this and it's working.

src/framework/components/anim/component-binder.js
ellthompson: I've removed this now. The findNode function now uses the graph class property.

src/framework/components/anim/component-binder.js
ellthompson: I've made this more generic so it'll use the entity as the base of the graph when no model component is present, tested and working 👍

src/framework/components/anim/component-binder.js
ellthompson: Yeah I didn't need to check animComponent here as it's always always available to the class. I've updated this to use conditional chaining too.

src/framework/components/anim/component-binder.js
```diff
@@ -240,10 +253,32 @@ class AnimComponentBinder extends DefaultAnimBinder {
...
+    }
+
+    rebind() {
+        this.targetCache = {};
+        // #ifdef DEBUG
+        this.visitedFallbackGraphPaths = {};
+        // #endif
+
+        if (this.animComponent.entity.model?.model?.graph) {
+            this.graph = this.animComponent.entity.model?.model?.graph;
```
mvaligursky, line 269: just double checking that this works for render component as well .. is this.graph set to this.entity or similar somewhere?

mvaligursky (approved):
approving with one comment

src/framework/components/anim/component-binder.js
ellthompson: Yep, if a model component isn't present it'll always use the current entity as the root of the graph.

Issue #2372, opened 2020-08-21
ReferenceError: window is not defined


yushimatenjin:
### Description

Using Node.js I get a window is not defined when importing node modules from playcanvas.

I updated the version of PlayCanvas from 1.32.3 to 1.34.0, and now I get this error.

#### Versions

- Node.js : v14.8.0
- playcanvas: v1.34.0
- macOS Catalina 10.15.5

### Steps to Reproduce

1. npm install playcanvas
2. import playcanvas package

**Error**

```jsx
> const playcanvas = require("playcanvas")
Uncaught ReferenceError: window is not defined
at /Users/yushimatenjin/workspace/playcanvas/node_modules/playcanvas/build/playcanvas.js:405:2
at /Users/yushimatenjin/workspace/playcanvas/node_modules/playcanvas/build/playcanvas.js:440:3
at /Users/yushimatenjin/workspace/playcanvas/node_modules/playcanvas/build/playcanvas.js:7:65
```

**Error line**

playcanvas.js:405:2

```jsx
window.setupVertexArrayObject = function(gl) {...
```

marklundin:
This looks like it's the polyfill for `OES_vertex_array_object` is this something that can be modified so that it doesn't add to the window global, or is that actually part of the spec?

willeastcott:
I believe PlayCanvas no longer has a problem running in Node.js. Therefore, I am closing this issue. If you believe a problem still remains @yushimatenjin, please reopen the issue.

PR #2777, opened 2021-01-26
[FIX] npm package not working with Node.js

yushimatenjin:
#### Fixed this issue. 
ReferenceError: window is not defined #2372

This pull request fixes a problem with Node.js that was not working.
Added environment detection when window is used.

#### Versions
- Node.js : v14.15.0
- playcanvas: v1.38.4
- macOS Catalina 10.15.5

##### Before
An error occurs when I try to use PlayCanvas with Node.js.

##### After

### DIFF
--- a/src/graphics/graphics-device.js
+++ b/src/graphics/graphics-device.js
@@ -321,8 +321,9 @@ class GraphicsDevice extends EventHandler {
         this._tempMacChromeBlitFramebufferWorkaround = isMac && isChrome && !options.alpha;
 
         // init polyfill for VAOs
-        window.setupVertexArrayObject(gl);
-
+        if (typeof window !== 'undefined') {
+            window.setupVertexArrayObject(gl);
+        }
         canvas.addEventListener("webglcontextlost", this._contextLostHandler, false);
         canvas.addEventListener("webglcontextrestored", this._contextRestoredHandler, false);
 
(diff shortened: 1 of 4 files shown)


kungfooman:
Just tested and it works with both ES5 and ES6:

ES5 REPL:
```js
const pc = require("./build/playcanvas.js");
```

ES6 REPL:
```js
import("./build/playcanvas.mjs").then(pc=>global.pc=pc)
```

But there are errors when it comes to actually using "useful" classes:

```js
> ent = new pc.Entity()
Uncaught Error: Couldn't find current application
    at new Entity (D:\web\playcanvas-engine-review\build\playcanvas.js:26205:12)
> app = new pc.Application()
Powered by PlayCanvas 1.39.0-dev 401ede6f2
Uncaught TypeError: Cannot read property 'id' of undefined
    at new Application (D:\web\playcanvas-engine-review\build\playcanvas.js:60863:37)
```

What and what shouldn't work? Writing and importing a polyfill for canvas/webgl (which is like 30 lines or so) might be easier while not having to test for `window` everywhere... it just exists and returns fixed/legit numbers and does nothing otherwise.

yushimatenjin:
@kungfooman 

I'm so sorry! 
I just built master and noticed that the problem does not occur at all.

I was only looking at npm packages 
https://www.npmjs.com/package/playcanvas

PR #2782, opened 2021-01-26
Don't add post effects to the pc namespace

willeastcott:
* There is no reason for the post effects to be added to the `pc` namespace.
* Removed `precision` property from attributes (default script attribute has been increased since the effects were written).
* Outline post effect wasn't parsing correctly in the Editor (because of malformed attribute specification)
* Gracefully handle no blend map being set on the `BlendEffect`


mvaligursky (approved):
Seems like a good change. Do we need to fix the example using it or would that just work?

willeastcott:
Ordinarily, it wouldn't need updating - but the sample is written in an incredibly weird way. So I'll rewrite/fix it.

mvaligursky:
Looks good!

PR #2798, opened 2021-01-29
Add follow camera script

willeastcott:
Add a simple 'follow camera' script - useful for tracking things like a moving vehicle, for example.

follow3

Try it out here:

https://playcanv.as/p/CxgnAp22

### DIFF
--- a/scripts/camera/follow-camera.js
+++ b/scripts/camera/follow-camera.js
@@ -17,7 +17,7 @@ FollowCamera.attributes.add('lerpAmount', {
     max: 1,
     default: 0.1,
     title: 'Lerp Amount',
-    description: 'The amount to lerp towards the target transform every frame'
+    description: 'The amount to lerp the camera towards its desired position every frame based on a 60fps frame rate. Lerping is frame rate independent though and will be correct for other frame rates.'
 });
 
 // initialize code called once per entity
@@ -56,12 +56,13 @@ FollowCamera.prototype.update = function(dt) {
         this.updateTargetPosition();
 
         // Lerp the current camera position to where we want it to be
-        this.currentPos.lerp(this.currentPos, this.targetPos, this.lerpAmount);
+        // Note that the lerping is framerate independent
+        this.currentPos.lerp(this.currentPos, this.targetPos, this.lerpAmount * dt * 60);
 
         // Set the camera's position
         this.entity.setPosition(this.currentPos);
 
-        // Look at the point relative to the target
+        // Look at the target entity from the new position
         this.entity.lookAt(this.target.getPosition());
     }
 };
--- a/scripts/camera/follow-camera.js
+++ b/scripts/camera/follow-camera.js
@@ -1,0 +1,67 @@
+var FollowCamera = pc.createScript('followCamera');
+
+FollowCamera.attributes.add('target', {
+    type: 'entity',
+    title: 'Target',
+    description: 'The target entity to follow'
+});
+FollowCamera.attributes.add('cameraOffset', {
+    type: 'vec3',
+    default: [0, 5, -10],
+    title: 'Camera Offset',
+    description: 'The local space offset with respect to the target entity coordinate system'
+});
+FollowCamera.attributes.add('lerpAmount', {
+    type: 'number',
+    min: 0,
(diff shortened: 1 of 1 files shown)


scripts/camera/follow-camera.js
```diff
@@ -0,0 +1,67 @@
...
+};
+
+// update code called every frame
+FollowCamera.prototype.update = function(dt) {
+    if (this.target) {
+        // Calculate where we want the camera to be
+        this.updateTargetPosition();
+
+        // Lerp the current camera position to where we want it to be
+        this.currentPos.lerp(this.currentPos, this.targetPos, this.lerpAmount);
```
slimbuck, line 59: might be good to make this frame rate independent?

willeastcott: Great point. Fixed!

slimbuck (approved):

scripts/camera/follow-camera.js
```diff
@@ -56,12 +56,13 @@ FollowCamera.prototype.update = function(dt) {
         this.updateTargetPosition();
 
         // Lerp the current camera position to where we want it to be
-        this.currentPos.lerp(this.currentPos, this.targetPos, this.lerpAmount);
+        // Note that the lerping is framerate independent
+        this.currentPos.lerp(this.currentPos, this.targetPos, this.lerpAmount * dt * 60);
```
LeXXik, line 60: Hmm, does multiplying by fixed scalar 60 necessary? Considering the variable delta time, it should scale it by constant value regardless of how long it took to render the last frame. That scale value should only affect how quickly the interpolation gets to `this.targetPos`. On low frame rates and high frame rates - it simply scales it proportionally. What frame rate depends on and which you already covered is - `dt` multiplication. The afterwards scalar should be only affecting the rate at which the angle reaches the target and making. Making it 60 can be misleading, suggesting you should multiply the lerping step by 60 in order to get a proper interpolation for 60 FPS.

slimbuck: Actually, this isn't quite frame rate independent because lerpAmount is a percentage. I believe you need something like:
```javascript
this.currentPos.lerp(this.currentPos, this.targetPos, 1.0 - Math.pow(this.lerpAmount, dt));
```
(untested, ymmv).

willeastcott: Think of it this way. Let's say I did this instead:

```
this.currentPos.lerp(this.currentPos, this.targetPos, this.lerpAmount);
```

If I set `lerpAmount` to 0.1, I'm saying "move 10% of the way from the current position to the target position" this frame. The problem is that the lerp animation will reach the target faster as the frame rate increases. So on slow devices, with slower frame rates, the camera will appear to start lagging behind the target.

We need to make the lerp animation the same independent of frame rate. So we need to give `lerpAmount` a different meaning. Let's define it as:

```
The amount to lerp the camera towards its desired position every frame based on a 60fps frame rate.
```

At 60fps, a frame takes 16.66ms. So now, if `lerpAmount` is 0.1, we're saying that we want the camera to animation 10% of the way towards the target in 16.66ms. `dt` is the time in seconds since the last frame. If you're running at 60fps, `dt` will be 1/60 (or 0.01666). So when `dt` is 1/60, we have to multiply it by 60 to get `this.lerpAmount * 1/60 * 60` which equals `this.lerpAmount * 1` which equals `this.lerpAmount`.

Now, if the frame rate is halved at 30fps, `dt` will be 1/30, or 0.0333. So we get `this.lerpAmount * 1/30 * 60` which equals `this.lerpAmount * 2`. So as the frame rate halves, the `lerpAmount` doubles. This makes the lerping frame rate independent.

Does that make sense?

slimbuck: The problem is after the first lerp of 10%, the next frame will lerp 10% of the remaining 90% (not 10% of the original 100%) which is why the power is needed. http://www.rorydriscoll.com/2016/03/07/frame-rate-independent-damping-using-lerp/

willeastcott: Oh right. Yeah, that's probably even better. 😆  (Sorry, didn't see your original reply coz you responded while I was typing).

PR #2831, opened 2021-02-04
Improvements to RenderComponent API

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/2753

1. API change. Constructor of MeshInstance has changed from `constructor(node, mesh, material)` to `constructor(mesh, material, node = null)` - node is optional for when used with a RenderComponent. Note that this change does not break compatibility and the constructor handles previous case as well.
2. MeshInstances can be supplied to RenderComponent using initializer.
3. MeshInstance.renderStyle works now

### DIFF
--- a/src/framework/components/render/system.js
+++ b/src/framework/components/render/system.js
@@ -12,6 +12,7 @@ const _schema = [
 // order matters here
 const _properties = [
     'material',
+    'meshInstances',
     'asset',
     'materialAssets',
     'castShadows',
@@ -75,6 +76,9 @@ class RenderComponentSystem extends ComponentSystem {
         for (i = 0; i < _properties.length; i++) {
             data[_properties[i]] = entity.render[_properties[i]];
         }
+
+        // we cannot copy mesh instances, delete them and component recreates them properly
+        delete data.meshInstances;
 
         var component = this.addComponent(clone, data);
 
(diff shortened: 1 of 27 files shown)


src/framework/components/render/system.js
```diff
@@ -12,6 +12,7 @@ const _schema = [
 // order matters here
 const _properties = [
     'material',
+    'meshInstances',
```
willeastcott, line 15: Should `renderStyle` be here too?

willeastcott (approved):

src/framework/components/render/system.js
mvaligursky: done now, add to where it needs to be for cloning to work, thanks!

src/scene/mesh-instance.js
```diff
@@ -48,7 +50,7 @@ class Command {
  * @classdesc An instance of a {@link pc.Mesh}. A single mesh can be referenced by many
  * mesh instances that can have different transforms and materials.
  * @description Create a new mesh instance.
- * @param {pc.GraphNode} node - The graph node defining the transform for this instance.
+ * @param {pc.GraphNode} [node] - The graph node defining the transform for this instance. This parameter is optional when used with {@link pc.RenderComponent} and will use the node the component is attached to.
```
mvaligursky, line 53: this is failing .. need to find different solution.
"Error: build/playcanvas.d.ts(13658,42): error TS1016: A required parameter cannot follow an optional parameter."

mvaligursky: done

willeastcott:
LGTM!

willeastcott (approved):

PR #2836, opened 2021-02-04
Handling non-uniform scaling when batching

mvaligursky:
- also the code no longer adds function to matrix instance it uses

### DIFF
--- a/src/math/mat3.js
+++ b/src/math/mat3.js
@@ -227,6 +227,31 @@ class Mat3 {
     }
 
     /**
+     * @function
+     * @name pc.Mat3#transform
+     * @description Transforms a 3-dimensional vector by a 3x3 matrix.
+     * @param {pc.Vec3} vec - The 3-dimensional vector to be transformed.
+     * @param {pc.Vec3} [res] - An optional 3-dimensional vector to receive the result of the transformation.
+     * @returns {pc.Vec3} The input vector v transformed by the current instance.
+     */
+    transform(vec, res = new Vec3()) {
+        var x, y, z, m;
+
+        m = this.data;
+
+        x = vec.x;
+        y = vec.y;
+        z = vec.z;
+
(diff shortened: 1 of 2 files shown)


slimbuck (approved):

src/math/mat3.js
```diff
@@ -226,6 +226,31 @@ class Mat3 {
...
 
+    /**
+     * @function
+     * @name pc.Mat3#transform
+     * @description Transforms a 3-dimensional vector by a 3x3 matrix.
+     * @param {pc.Vec3} vec - The 3-dimensional vector to be transformed.
+     * @param {pc.Vec3} [res] - An optional 3-dimensional vector to receive the result of the transformation.
+     * @returns {pc.Vec3} The input vector v transformed by the current instance.
+     */
+    transform(vec, res = new Vec3()) {
```
willeastcott, line 237: I'm not keen on the naming here. We already have:
https://github.com/playcanvas/engine/blob/master/src/math/mat4.js#L430

willeastcott: By the way, I'm not saying `transform` has to become `transformVector`. I mean, we could massage the `Mat4` API. I just want consistency/guessability here.

mvaligursky: my thinking was that Mat4 has transformPoint and transformVector
for Mat3 those two are the same as there's no translation .. so I just called it .transform. All that seems ok to me.

I guess I'm less keen on Mat4.transformVec4 - but I not sure what else .. it's not used by the engine.

Maksims: `transformVector` or `transformPoint` - seems reasonable and consistent. From non-math friendly developer, it might be non descriptive to have: `mat.transform(what?)`.

mvaligursky: sounds good guys, see here https://github.com/playcanvas/engine/pull/2837

willeastcott: Cool, thanks!

PR #2860, opened 2021-02-11
Keyboard - remove presses when window is blurred or becomes hidden

### LINKED ISSUE #2855: app.keyboard.isPressed() remains true after switching Tab and coming back to same tab
if any key is pressed and the tab is switched, Then the key remains pressed after coming back to the same tab if it's pressed up before coming to this tab.

The first person movement project has this same issue. Here is a link to the project :
https://playcanvas.com/editor/scene/440123

Steps to reproduce:
1- Press any control Key e.g W/UP
2- Change Tab while keeping that key pressed.
3- After changing tab, press up that key and come back to the build (fps) tab.

Maksims:
Fixes #2855

### DIFF
--- a/src/input/keyboard.js
+++ b/src/input/keyboard.js
@@ -20,7 +20,7 @@ function makeKeyboardEvent(event) {
  * @param {string|number} s - Either a character code or the key character.
  * @returns {number} The character code.
  */
-function toKeyCode(s){
+function toKeyCode(s) {
     if (typeof s === "string") {
         return s.toUpperCase().charCodeAt(0);
     }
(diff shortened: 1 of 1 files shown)


src/input/keyboard.js
```diff
@@ -226,6 +232,17 @@ class Keyboard extends EventHandler {
...
+    _handleVisibilityChange() {
+        if (document.visibilityState === 'hidden') {
+            this._handleWindowBlur();
+        }
+    }
+
+    _handleWindowBlur() {
+        this._keymap = {};
+        this._lastmap = {};
+    }
```
yaustar, line 244: Will we need to fire keyup events so that the application can handle keys being 'released' by window blur?

Maksims: I thought about it. But we won't be able to do that without faking Event object.
In the web, keydown > keyup - is not a guaranteed flow, so we probably should not change or try to mitigate it, as there will be a lot of edge cases.

This PR only focuses on static access to key states, not events. That would need to be a separate PR.

yaustar: Right, I see what you mean

yaustar (approved):

PR #2878, opened 2021-02-17
[FIX] Several fixes reported by checkJs

willeastcott:
After enabling `checkJs` in VS Code, the following were fixed:

* Removed several uses of `pc.` in the codebase.
* Various missing tags, typos and type errors in JSDocs.
* Incorrect number of args for `Math.floor`.
* Removed `BoundingSphere#compute` which has been completely broken (and private) for years.
* Removed occurrences of non-existent and unused property `Texture#_prefilteredMips`.
* Switched context lost/restore handlers to arrow functions.
* Add missing imports.
* Fixes `npm run tsd` on Windows. You can't insert semi-colons in `package.json` scripts. You need to use `&&` instead.

Note that this is only a small fraction of errors/warnings reported by `checkJs`. Hopefully, this is already a testament to the value of the work to enable it.

### DIFF
--- a/src/graphics/texture.js
+++ b/src/graphics/texture.js
@@ -607,7 +607,7 @@ class Texture {
 
                 if (format === PIXELFORMAT_PVRTC_2BPP_RGB_1 ||
                     format === PIXELFORMAT_PVRTC_2BPP_RGBA_1) {
-                    blockWidth = Math.floor(blockWidth / 2, 1);
+                    blockWidth = Math.floor(blockWidth / 2);
                 }
 
                 result += blockWidth * blockHeight * blockDepth * blockSize;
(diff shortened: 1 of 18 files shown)


src/graphics/texture.js
```diff
@@ -607,7 +607,7 @@ class Texture {
 
                 if (format === PIXELFORMAT_PVRTC_2BPP_RGB_1 ||
                     format === PIXELFORMAT_PVRTC_2BPP_RGBA_1) {
-                    blockWidth = Math.floor(blockWidth / 2, 1);
+                    blockWidth = Math.floor(blockWidth / 2);
```
willeastcott, line 610: @slimbuck This looks super-odd to me. I wonder what the original intention of the section parameter of 1 to `Math.floor` was here. Can you cast an eye and verify the logic is OK?

slimbuck: Bug! I think this should be:
```javascript
blockWidth = Math.max(Math.floor(blockWidth / 2), 1);
```

willeastcott: Thanks. Fixed!

src/shape/bounding-sphere.js
```diff
@@ -28,49 +26,6 @@ class BoundingSphere {
         return lenSq < r * r;
     }
 
-    compute(vertices) {
```
slimbuck, line 31: did you intend to remove this?

slimbuck (approved):

src/shape/bounding-sphere.js
willeastcott: Yes. It's literally never worked because it calls a non-existent function of the Vec3 class. Rather than fix the function, since we know nobody is using it, I'd rather delete it to give us a slimmer engine.

slimbuck: :+1:

ellthompson (approved):

Issue #2848, opened 2021-02-10
Post processing behaviour is different with stacked cameras since 1.39.0


yaustar:
https://playcanvas.com/project/763845/overview/overlaycamera-posteffect-problem

This has 3 cameras set up, 

Camera1 with priority 0 rendering all layers with post processing bloom effect
Camera3 with priority 0 rendering all but the ball layer and in the top half of the viewport
Camera2 with priority 1 rendering the ball layer

The desired result is the bottom view (engine version 1.38.4) with all cameras rendered and composited
The top view just shows Camera1 with the post effect
![image](https://user-images.githubusercontent.com/16639049/107520502-639fe380-6ba9-11eb-9604-87a4604eed83.png)

Without the post process effect, both engines render as desired
![image](https://user-images.githubusercontent.com/16639049/107520576-79ada400-6ba9-11eb-9870-775daad2934b.png)

PR #2882, opened 2021-02-17
Refactor of post-effects to integrate better with RenderActions and LayerComposition

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/2848

This PR refactors the way post-processing is integrated. Originally post-processing would create new layer and adjust internals of `LayerComposition` to integrate in it (which would partially break when composition changed at a later stage). New implementation gives `LayerComposition` ownership of the integration and this updates as needed when layers / cameras are changed.
`LayerComposition` now also handles camera stacking better - it decides what render targets are used and when they get cleared, this removed the need of forward-renderered to track this. 

**new public API**:
- `CameraComponent#disablePostEffectsLayer` - defaults to `LAYERID_UI` to stop postprocessing on UI layer, can be set to other layer ID or undefined to postprocess all layers of the camera
- RenderTarget constructor now takes optional `options.name` for debugging purposes

**new private API**:
- `[boolean] LayerComposition#logRenderActions` - list of `RenderAction`s get logged out in debug mode when it's updated. Example output:

Screen Shot 2021-02-18 at 12 08 02 PM

**Other changes:**
- non public WIP class post-effect-pass.js has been removed as unused and no longer functional without further work.
- depth buffer texture for the scene, and also post-processing color buffers don’t use mipmaps
- post-effects.html example - added text on the screen showing which post effects are enabled, and also option to include UI in postprocessing
Screen Shot 2021-02-18 at 12 15 17 PM


src/framework/components/camera/component.js
```diff
@@ -107,6 +107,9 @@ import { PostEffectQueue } from './post-effect-queue.js';
  * camera should belong. Don't push/pop/splice or modify this array, if you want to
  * change it, set a new one instead. Defaults to [LAYERID_WORLD, LAYERID_DEPTH,
  * LAYERID_SKYBOX, LAYERID_UI, LAYERID_IMMEDIATE].
+ * @property {number} stopPostprocessingLayer Layer ID of layer which stops postprocessing of the camera.
```
willeastcott, line 110: I'm not keen on `stopPostprocessingLayer ` as a name. Trying to think of something better...

mvaligursky: It's not my favorite name, but the best I could come up. Alternatives anybody?

raytranuk: Maybe a flag to say whether it's included, with UI layer defaulted as not included? Not sure if that would cause other problems or if we would have to add warnings if the user sets the layer order such that an included layer was after the UI layer?

mvaligursky: ideally I want to specify which layer is last for postprocessing, not just a bool to say to stop on UI
example: you can postprocess environment, but not characters and UI.

src/scene/layer-composition.js
```diff
@@ -439,6 +488,62 @@ class LayerComposition extends EventHandler {
...
+                continue;
+            }
+
+            // render it to render target
+            ra.renderTarget = renderTarget;
+        }
+    }
+
+    // logs render action and their properties
+    _logRenderActions() {
```
willeastcott, line 518: I can see how this might be useful to develop this feature. But do we really want to execute this code even in a debug build? The debug build is really for logging warnings and errors.

src/scene/layer-composition.js
```diff
@@ -439,6 +488,62 @@ class LayerComposition extends EventHandler {
         return result;
     }
 
+    // executes when post-processing camera's render actions were created to propage rendering to
+    // render targets to previous camera as needed
+    propagateRenderTarget(startIndex, renderTarget) {
+        // let lastFixed = null;
+        // let lastFixedClearColor = false;
```
willeastcott, line 495: Delete commented out code?

src/framework/components/camera/component.js
willeastcott: * `stopPostLayer`
* `stopPostFxLayer`
* `stopPostEffectsLayer`
* `stopPostprocessLayer`
* `stopPostprocessingLayer`
* `disablePostLayer`
* `disablePostFxLayer`
* `disablePostEffectsLayer`
* `disablePostprocessLayer`
* `disablePostprocessingLayer`

@slimbuck Any opinions?

src/scene/layer-composition.js
mvaligursky: how did I miss those. Thanks!

src/framework/components/camera/component.js
slimbuck: `postProcessStopLayer` would probably make most sense to me, but I guess we want the verb to go first?

mvaligursky: maybe this:
disablePostEffectsLayer

the post effect base class is PostEffect, so that matches

willeastcott: Agreed!

mvaligursky: and done.

mvaligursky:
I'm just adding one small final refactor, give me 10 minutes.

mvaligursky:
And done now, ready for final review.

willeastcott (approved):

slimbuck (approved):
Awesome work!

PR #2885, opened 2021-02-17
Asset unload event moved to public API

yaustar:
Proposal to move the event to public API

### DIFF
--- a/src/asset/asset.js
+++ b/src/asset/asset.js
@@ -112,6 +112,13 @@ class Asset extends EventHandler {
 
     /**
      * @event
+     * @name Asset#unload
+     * @description Fired just before the asset unloads the resource.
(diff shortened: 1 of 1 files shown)


src/asset/asset.js
```diff
@@ -110,6 +110,13 @@ class Asset extends EventHandler {
      * @param {Asset} asset - The asset that was loaded.
      */
 
+    /**
+     * @event
+     * @name Asset#unload
+     * @description Fired just before the asset unloads the resource.
```
willeastcott, line 116: Is it worth giving a bit more color relating to when you'd use it?

yaustar:
Added an example, was a little difficult to think of a concert example 😅

willeastcott (approved):

PR #2886, opened 2021-02-17
AnimComponent fixes

ellthompson:
- BlendTree class performance refactor
- AnimComponent now correctly handles any assignment of animations to non user created states
- Reduced the excess debug logs produced by the DefaultAnimBinder class


src/anim/controller/anim-blend-tree.js
```diff
@@ -143,25 +149,26 @@ class AnimBlendTree extends AnimNode {
...
-                        pj = this._children[j].point.clone();
+                        pj = this._children[j].point;
+                        var pjLength = this._children[j].pointLength;
                         var pipAngle = getAngleRad(pi, p);
                         var pipjAngle = getAngleRad(pi, pj);
-                        pipj = new Vec2((pj.length() - pi.length()) / ((pj.length() + pi.length()) / 2), pipjAngle * 2.0);
-                        pip = new Vec2((p.length() - pi.length()) / ((pj.length() + pi.length()) / 2), pipAngle * 2.0);
-                        result = clamp(1.0 - Math.abs((pip.clone().dot(pipj) / pipj.lengthSq())), 0.0, 1.0);
+                        pipj = new Vec2((pjLength - piLength) / ((pjLength + piLength) / 2), pipjAngle * 2.0);
+                        pip = new Vec2((pLength - piLength) / ((pjLength + piLength) / 2), pipAngle * 2.0);
```
willeastcott, line 170: I'm not saying you have to fix this now, but you really want to get to the point where your code aggressively minimizes dynamic allocations. These two `Vec2`s could be preallocated and use `Vec2#set` instead.

ellthompson: I've refactored the BlendTree class again which removes the creation of new Vec2's here.

src/anim/binder/default-anim-binder.js
```diff
@@ -138,10 +138,14 @@ class DefaultAnimBinder {
 
             // #ifdef DEBUG
             var fallbackGraphPath = AnimBinder.encode(path.entityPath[path.entityPath.length - 1] || "", 'graph', path.propertyPath);
-            if (this.visitedFallbackGraphPaths[fallbackGraphPath]) {
+            if (this.visitedFallbackGraphPaths[fallbackGraphPath] === 1) {
                 console.warn('Anim Binder: Multiple animation curves with the path ' + fallbackGraphPath + ' are present in the ' + this.graph.path + ' graph which may result in the incorrect binding of animations');
             }
-            this.visitedFallbackGraphPaths[fallbackGraphPath] = true;
+            if (!Number.isFinite(this.visitedFallbackGraphPaths[fallbackGraphPath])) {
```
mvaligursky, line 144: this feels odd .. I assume visitedFallbackGraphPaths is private, so why do you need to test for invalid numbers? Should the code store valid numbers only instead?

src/anim/controller/anim-blend-tree.js
```diff
@@ -30,19 +12,19 @@ function clamp(num, min, max) {
...
-    constructor(state, parent, name, point, type, parameters, children, findParameter) {
+    constructor(state, parent, name, point, parameters, children, findParameter) {
         super(state, parent, name, point);
-        this._type = type;
         this._parameters = parameters;
         this._parameterValues = null;
         this._children = [];
         this._findParameter = findParameter;
+        this._p = new Vec2();
+        this._pip = new Vec2();
```
mvaligursky, line 27: if these are optimizations to avoid allocations, perhaps they should be static, or global.

src/anim/controller/anim-blend-tree.js
```diff
@@ -204,6 +94,176 @@ class AnimBlendTree extends AnimNode {
...
+        return inclusive ? num >= min && num <= max : num > min && num < max;
+    }
+
+    static getAngleRad(a, b) {
+        return Math.atan2(a.x * b.y - a.y * b.x, a.x * b.x + a.y * b.y);
+    }
+
+    static clamp(num, min, max) {
+        return num <= min ? min : num >= max ? max : num;
+    }
```
mvaligursky, line 110: perhaps use existing math.clamp

src/anim/controller/anim-blend-tree.js
```diff
@@ -204,6 +94,176 @@ class AnimBlendTree extends AnimNode {
...
+
+    static between(num, a, b, inclusive) {
+        var min = Math.min(a, b),
+            max = Math.max(a, b);
+        return inclusive ? num >= min && num <= max : num > min && num < max;
+    }
+
+    static getAngleRad(a, b) {
+        return Math.atan2(a.x * b.y - a.y * b.x, a.x * b.x + a.y * b.y);
+    }
```
mvaligursky, line 106: if these are generic functions, they could be added to math class

src/anim/controller/anim-blend-tree.js
```diff
@@ -204,6 +94,176 @@ class AnimBlendTree extends AnimNode {
...
+}
+
+/**
+ * @private
+ * @class
+ * @name AnimBlendTree1D
+ * @classdesc An AnimBlendTree that calculates it's weights using a 1D algorithm
+ * @description Create a new BlendTree1D.
+ */
+class AnimBlendTree1D extends AnimBlendTree {
```
mvaligursky, line 121: move this into separate file please

src/anim/controller/anim-blend-tree.js
```diff
@@ -204,6 +94,176 @@ class AnimBlendTree extends AnimNode {
...
+}
+
+/**
+ * @private
+ * @class
+ * @name AnimBlendTreeCartesian2D
+ * @classdesc An AnimBlendTree that calculates it's weights using a 2D catesian algorithm
+ * @description Create a new BlendTree1D.
+ */
+class AnimBlendTreeCartesian2D extends AnimBlendTree {
```
mvaligursky, line 151: move this into separate file please

src/anim/controller/anim-blend-tree.js
```diff
@@ -204,6 +94,176 @@ class AnimBlendTree extends AnimNode {
...
+ * @name AnimBlendTree1D
+ * @classdesc An AnimBlendTree that calculates it's weights using a 1D algorithm
+ * @description Create a new BlendTree1D.
+ */
+class AnimBlendTree1D extends AnimBlendTree {
+    calculateWeights() {
+        var i;
+        var parameterValue = this._findParameter(this._parameters[0]).value;
+        if (this.parametersEqual([parameterValue])) return;
+        this._parameterValues = [parameterValue];
```
mvaligursky, line 126: this allocates two arrays, ideally we do without any allocations.

src/anim/controller/anim-blend-tree.js
```diff
@@ -204,6 +94,176 @@ class AnimBlendTree extends AnimNode {
...
+        this._children[0].weight = 0.0;
+        for (i = 0; i < this._children.length - 1; i++) {
+            var c1 = this._children[i];
+            var c2 = this._children[i + 1];
+            if (AnimBlendTree.between(parameterValue, c1.point, c2.point, true)) {
```
mvaligursky, line 131: this feels like children should be stored in sorted order in the array .. just double checking that's not missing

src/anim/controller/anim-blend-tree.js
```diff
@@ -204,6 +94,176 @@ class AnimBlendTree extends AnimNode {
...
+                var child2Distance = Math.abs(c1.point - c2.point);
+                var parameterDistance = Math.abs(c1.point - parameterValue);
+                var weight = (child2Distance - parameterDistance) / child2Distance;
```
mvaligursky, line 134: I know it's not expected, but user can set both child at the same point (maybe while testing stuff in editor) and we get divide by zero here

src/anim/controller/anim-blend-tree.js
```diff
@@ -84,112 +66,20 @@ class AnimBlendTree extends AnimNode {
...
-                }
-                for (i = 0; i < this._children.length; i++) {
-                    this._children[i].weight = clamp(this._parameterValues[i], 0.0, Number.MAX_VALUE) / weightSum;
-                }
-                break;
+    currentParameterValues() {
+        return this._parameters.map(function (param) {
+            return this._findParameter(param).value;
+        }.bind(this));
+    }
```
mvaligursky, line 73: each time this gets called, and array is allocated .. I think class should store single array and this function should just update it.

src/anim/controller/anim-blend-tree.js
```diff
@@ -204,6 +94,176 @@ class AnimBlendTree extends AnimNode {
...
+        var i, j, pi, parameterValues, minj, result, weightSum;
+        parameterValues = this.currentParameterValues();
+        if (this.parametersEqual(parameterValues)) return;
+        this._parameterValues = parameterValues;
+        this._p.set(this._parameterValues);
+        weightSum = 0.0;
+        for (i = 0; i < this._children.length; i++) {
+            pi = this._children[i].point;
+            minj = Number.MAX_VALUE;
+            for (j = 0; j < this._children.length; j++) {
```
mvaligursky, line 171: currently all pairs of children get evaluate twice here .. for example pairs [1,2] and [2,1] .. can this be changed to only evaluate one time?
At least the pointDistanceCache function should create pointKey with sorted numbers, to avoid doubling these.

src/anim/controller/anim-blend-tree.js
```diff
@@ -204,6 +94,176 @@ class AnimBlendTree extends AnimNode {
...
+/**
+ * @private
+ * @class
+ * @name AnimBlendTreeCartesian2D
+ * @classdesc An AnimBlendTree that calculates it's weights using a 2D catesian algorithm
+ * @description Create a new BlendTree1D.
+ */
+class AnimBlendTreeCartesian2D extends AnimBlendTree {
+    pointDistanceCache(i, j) {
+        if (!this._pointCache) this._pointCache = {};
```
mvaligursky, line 153: does the cache need to be cleaned sometimes, when position of children changes?

src/anim/controller/anim-blend-tree.js
```diff
@@ -204,6 +94,176 @@ class AnimBlendTree extends AnimNode {
...
+        if (!this._pointCache[pointKey]) {
+            this._pointCache[pointKey] = this._children[j].point.clone().sub(this._children[i].point);
+        }
+        return this._pointCache[pointKey];
+    }
+
+    calculateWeights() {
+        var i, j, pi, parameterValues, minj, result, weightSum;
+        parameterValues = this.currentParameterValues();
+        if (this.parametersEqual(parameterValues)) return;
```
mvaligursky, line 164: I would probably combine these two lines to a single function. It would take this_parameterValues as parameter, it would update it .. and inside the same loop it would check each value for equal and return true or false.

src/anim/controller/anim-blend-tree.js
```diff
@@ -204,6 +94,176 @@ class AnimBlendTree extends AnimNode {
...
+ */
+class AnimBlendTreeDirect extends AnimBlendTree {
+    calculateWeights() {
+        var i, parameterValues, weightSum;
+        parameterValues = this.currentParameterValues();
+        if (this.parametersEqual(parameterValues)) return;
+        this._parameterValues = parameterValues;
+        weightSum = 0.0;
+        for (i = 0; i < this._children.length; i++) {
+            weightSum += AnimBlendTree.clamp(this._parameterValues[i], 0.0, Number.MAX_VALUE);
```
mvaligursky, line 261: instead of clamp, just use max(value, 0) - clamping high range to MAX_VALUE is not useful

src/anim/controller/anim-controller.js
```diff
@@ -420,6 +420,9 @@ class AnimController {
     }
 
     removeNodeAnimations(nodeName) {
+        if ([ANIM_STATE_START, ANIM_STATE_END, ANIM_STATE_ANY].includes(nodeName)) {
```
mvaligursky, line 423: it's likely this
`[ANIM_STATE_START, ANIM_STATE_END, ANIM_STATE_ANY]`
allocates array each time the function runs. Maybe should be stored in some const somewhere.

src/framework/components/anim/component.js
```diff
@@ -105,6 +105,7 @@ class AnimComponent extends Component {
             var layer = this.data.layers[i];
             for (var j = 0; j < layer.states.length; j++) {
                 var stateName = layer.states[j];
+                if ([ANIM_STATE_START, ANIM_STATE_END, ANIM_STATE_ANY].includes(stateName)) continue;
```
mvaligursky, line 108: this should be const array stored elsewhere

src/anim/controller/anim-controller.js
willeastcott: Actually, I think there's another problem here. `Array#includes` is not supported in IE11. And remember, Babel just transpiles language features. I don't think it polyfills standard library functions like this.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes
I would maybe just swap for:
```javascript
    if (nodeName === ANIM_STATE_START ||
        nodeName === ANIM_STATE_END ||
        nodeName ===  ANIM_STATE_ANY) {
```

src/framework/components/anim/component.js
willeastcott: Same follow up comment about `includes` as above.

src/anim/binder/default-anim-binder.js
ellthompson: It only stores non negative numbers but there may not be an entry for a given path in the dictionary yet.

src/anim/controller/anim-blend-tree.js
ellthompson: Sure, i'll set them as static properties.

src/anim/controller/anim-blend-tree.js
ellthompson: 👍

src/anim/controller/anim-blend-tree.js
ellthompson: I've moved between to the math class and the angle radians function to Vec2 as it operates specifically on them.

src/anim/controller/anim-blend-tree.js
ellthompson: Sure, i've moved them all to separate files now.

src/anim/controller/anim-blend-tree.js
ellthompson: 👍

src/anim/controller/anim-blend-tree.js
ellthompson: Done 👍

src/anim/controller/anim-blend-tree.js
ellthompson: For this calculation pipj != pjpi so we have to calculate all combinations.

src/anim/controller/anim-blend-tree.js
ellthompson: These points are only set during construction.

src/anim/controller/anim-blend-tree.js
ellthompson: Makes sense, i've updated this.

src/anim/controller/anim-blend-tree.js
ellthompson: Done 👍

src/anim/controller/anim-controller.js
ellthompson: Sure, i've set this as a constant now.

src/framework/components/anim/component.js
ellthompson: Done 👍

src/anim/controller/anim-blend-tree.js
ellthompson: Done 👍

src/anim/controller/anim-blend-tree.js
willeastcott: Con for static: they'll be visible/accessible to devs poking at our structures. Typically we declare these types of vars at global module scope.

https://github.com/playcanvas/engine/blob/master/src/scene/graph-node.js#L9-L23

But I'm not _too_ precious about it.

src/anim/controller/anim-node.js
```diff
@@ -17,7 +17,14 @@ class AnimNode {
         this._state = state;
         this._parent = parent;
         this._name = name;
-        this._point = Array.isArray(point) ? new Vec2(point[0], point[1]) : point;
+        if (Array.isArray(point)) {
+            this._point = new Vec2(point[0], point[1]);
+            this._pointLength = this._point.length();
+        } else {
+            this._point = point;
+            this._pointLength = point;
```
willeastcott, line 25: Should this be:

```javascript
            this._pointLength = point.length();
```

ellthompson: Cheers, fixed!

src/anim/controller/anim-node.js
```diff
@@ -17,7 +17,14 @@ class AnimNode {
...
         this._parent = parent;
         this._name = name;
-        this._point = Array.isArray(point) ? new Vec2(point[0], point[1]) : point;
+        if (Array.isArray(point)) {
+            this._point = new Vec2(point[0], point[1]);
+            this._pointLength = this._point.length();
+        } else {
+            this._point = point;
+            this._pointLength = point.length();
+        }
```
willeastcott, line 26:
```suggestion
        if (Array.isArray(point)) {
            this._point = new Vec2(point[0], point[1]);
        } else {
            this._point = point;
        }
        this._pointLength = this._point.length();
```

src/anim/controller/anim-blend-tree-1d.js
```diff
@@ -0,0 +1,32 @@
+import { math } from '../../math/math.js';
+
+import { AnimBlendTree } from './anim-blend-tree.js';
+
+/**
+ * @private
+ * @class
+ * @name AnimBlendTree1D
+ * @classdesc An AnimBlendTree that calculates it's weights using a 1D algorithm
```
willeastcott, line 9:
```suggestion
 * @classdesc An AnimBlendTree that calculates its weights using a 1D algorithm
```

src/anim/controller/anim-blend-tree-2d-cartesian.js
```diff
@@ -0,0 +1,51 @@
+import { Vec2 } from '../../math/vec2.js';
+import { math } from '../../math/math.js';
+
+import { AnimBlendTree } from './anim-blend-tree.js';
+
+/**
+ * @private
+ * @class
+ * @name AnimBlendTreeCartesian2D
+ * @classdesc An AnimBlendTree that calculates it's weights using a 2D catesian algorithm
```
willeastcott, line 10:
```suggestion
	 * @classdesc An AnimBlendTree that calculates its weights using a 2D Cartesian algorithm
```

src/anim/controller/anim-blend-tree-2d-directional.js
```diff
@@ -0,0 +1,57 @@
+import { Vec2 } from '../../math/vec2.js';
+import { math } from '../../math/math';
+
+import { AnimBlendTree } from './anim-blend-tree.js';
+
+/**
+ * @private
+ * @class
+ * @name AnimBlendTreeDirectional2D
+ * @classdesc An AnimBlendTree that calculates it's weights using a 2D directional algorithm
```
willeastcott, line 10:
```suggestion
 * @classdesc An AnimBlendTree that calculates its weights using a 2D directional algorithm
```

src/framework/components/anim/component.js
```diff
@@ -87,7 +87,7 @@ class AnimComponent extends Component {
             var layerName = layer.name;
             for (var j = 0; j < layer.states.length; j++) {
                 var stateName = layer.states[j];
-                if (stateName !== ANIM_STATE_START && stateName !== ANIM_STATE_END && stateName !== ANIM_STATE_ANY) {
+                if (!ANIM_CONTROL_STATES.includes(stateName)) {
```
willeastcott, line 90: Not sure what happened to my previous comment, but `Array#includes` doesn't exist in IE11. Have you checked the Babel output to see if it's IE11-compatible?

src/framework/components/anim/component.js
```diff
@@ -105,6 +105,7 @@ class AnimComponent extends Component {
             var layer = this.data.layers[i];
             for (var j = 0; j < layer.states.length; j++) {
                 var stateName = layer.states[j];
+                if (ANIM_CONTROL_STATES.includes(stateName)) continue;
```
willeastcott, line 108: Not sure what happened to my previous comment, but `Array#includes` doesn't exist in IE11. Have you checked the Babel output to see if it's IE11-compatible?

src/math/math.js
```diff
@@ -329,7 +329,24 @@ var math = {
             bits += m & 1;
             return bits;
         };
-    }())
+    }()),
+
+
+    /**
+     * @function
```
willeastcott, line 336: I'm tempted to keep this private for now. We can always make it public later.

src/math/vec2.js
```diff
@@ -429,6 +429,18 @@ class Vec2 {
         return '[' + this.x + ', ' + this.y + ']';
     }
 
+    /**
+     * @function
```
willeastcott, line 433: Same, let's keep this private for now and just use this function for internal purposes.

src/anim/controller/anim-controller.js
```diff
@@ -420,6 +420,9 @@ class AnimController {
     }
 
     removeNodeAnimations(nodeName) {
+        if (ANIM_CONTROL_STATES.includes(nodeName)) {
```
willeastcott, line 423: `Array#includes` doesn't exist in IE11. Have you checked the Babel output to see if it's IE11-compatible?

src/anim/controller/anim-state.js
```diff
@@ -75,12 +105,12 @@ class AnimState {
...
 
     get nodeCount() {
-        if (!this._blendTree || !(this._blendTree.constructor === AnimBlendTree)) return 1;
+        if (!this._blendTree || (this._blendTree.constructor === AnimNode)) return 1;
         return this._blendTree.getNodeCount();
     }
 
     get playable() {
-        return (this.name === ANIM_STATE_START || this.name === ANIM_STATE_END || this.name === ANIM_STATE_ANY || this.animations.length === this.nodeCount);
+        return (ANIM_CONTROL_STATES.includes(this.name) || this.animations.length === this.nodeCount);
```
willeastcott, line 113: `Array#includes` doesn't exist in IE11. Have you checked the Babel output to see if it's IE11-compatible?

src/math/vec2.js
```diff
@@ -429,6 +429,18 @@ class Vec2 {
         return '[' + this.x + ', ' + this.y + ']';
     }
 
+    /**
+     * @function
+     * @name Vec2#angleRad
+     * @description Calculates the angle between two Vec2's in radians.
+     * @param {Vec2} lhs - The first vector operand for the calculation.
+     * @param {Vec2} rhs - The second vector operand for the calculation.
+     * @returns {number} the calculated angle
```
willeastcott, line 438:
```suggestion
     * @returns {number} The calculated angle in radians.
```

src/anim/binder/default-anim-binder.js
mvaligursky: there must be a way to just test if this property exists? Number.isFinite sounds like it tests more than needed and would cost more.

ellthompson: It's because 0 is a falsy value so I wouldn't want to just test if a number exists.

src/anim/controller/anim-blend-tree.js
ellthompson: This is now testing against two points with the same value.

ellthompson:
> ah actually, this is 1d .. so those are floats right.

Yep points are floats for 1D blend trees.

willeastcott (approved):

willeastcott:
That was a great team review process - really good to see this PR polished to a high degree! 😄  LGTM now!

src/anim/controller/anim-blend-tree-2d-cartesian.js
```diff
@@ -0,0 +1,51 @@
...
+        if (!this._pointCache[pointKey]) {
+            this._pointCache[pointKey] = this._children[j].point.clone().sub(this._children[i].point);
+        }
+        return this._pointCache[pointKey];
+    }
+
+    calculateWeights() {
+        if (this.updateParameterValues()) return;
+        var i, j, pi, pipj, minj, result, weightSum;
+        AnimBlendTreeCartesian2D._p.set(...this._parameterValues);
```
mvaligursky, line 30: please double check what this looks like in ES5, we have not tested this yet.

mvaligursky (approved):
I added one comment, but approving otherwise.
In the future though, please use let and const for new code instead of var.

src/anim/controller/anim-blend-tree-2d-cartesian.js
ellthompson: This is converted to a call to the apply method in es5.

PR #2893, opened 2021-02-19
Assets fix

### LINKED ISSUE #2892: asset.getFileUrl() throws TypeError
Texture type (more complicated because of variants):

```js
asset = new pc.Asset("Test", "texture", {
        url: "test.jpg"
    });
asset.getFileUrl()
```

HTML:

```js
asset = new pc.Asset("Test", "html", {
        url: "test.html"
    });
asset.getFileUrl()
```

Result for both:

image

It starts to work as soon the asset is added to registry:

```js
asset = new pc.Asset("Test", "texture", {
        url: "test.jpg"
    });
app.assets.add(asset)
asset.getFileUrl()
```

image

So this is not a big deal, yet maybe it would be good to not throw errors by accessing undefined variables? Some people will run into this and wonder what goes on (at least I did 😅).

slimbuck:
Fixes #2892


src/asset/asset.js
```diff
@@ -192,29 +193,30 @@ class Asset extends EventHandler {
...
-
-                // if the variant does not exist but the asset is in a bundle
-                // and the bundle contain assets with this variant then return the default
-                // file for the asset
-                if (app.enableBundles) {
-                    var bundles = app.bundles.listBundlesForAsset(this);
-                    if (! bundles) continue;
+            var app = this.registry?._loader?._app || getApplication();
+            var device = app?.graphicsDevice;
+            if (device) {
```
yaustar, line 198: We should return null or the original URL if no device is found?

yaustar: Oh wait, it does. Couldn't see it in the diff 😅

yaustar (approved):
Looks fine. This pattern is done elsewhere so it's inline with the rest of the engine

mvaligursky (approved):

PR #2896, opened 2021-02-21
[Fix] Converted few functions to ES6 format

mvaligursky:


willeastcott:
I think I already fixed all these. 😆

mvaligursky:
yep we did them at the same time I think :)

PR #2898, opened 2021-02-22
[AnimComponent] Sync BlendTree animation durations

ellthompson:
The animation durations for all children of a BlendTree can now be synchronised. This ensures multiple blended animations that represent similar motions (walk / run) will animate smoothly even when they have differing animation durations. The playback speed for these synced animations is based on the mean of all of the animation durations, weighted by each animations blend weight.

This can be enabled by setting `syncAnimations` to true when creating a blend tree.


mvaligursky:
- I probably would not mind some other name than syncDurations .. syncProgress, or syncAnimations or similar maybe?
- I don't see a default value for this in this PR .. but ideally this defaults to true

src/anim/controller/anim-blend-tree.js
```diff
@@ -66,6 +71,10 @@ class AnimBlendTree extends AnimNode {
         return paramsEqual;
     }
 
+    getNodeWeightedDuration(i) {
+        return this._children[i].animTrack.duration / this._children[i].speedMultiplier * this._children[i].weight;
```
mvaligursky, line 75: store this in a variable 
`this._children[i]`

src/anim/controller/anim-blend-tree-1d.js
```diff
@@ -35,6 +40,11 @@ class AnimBlendTree1D extends AnimBlendTree {
                 c2.weight = 0.0;
             }
         }
+        if (this._syncDurations) {
+            for (i = 0; i < this._children.length; i++) {
+                this._children[i].weightedSpeed = meanDuration;
```
mvaligursky, line 45: it's probably a naming again, but `weightedSpeed = meanDuration` seems to assign duration to speed? Also, 'mean' is an average, but in this case it seems to be a sum of durations?

ellthompson: This confusion must come from the fact that part of the calculations for the mean duration are occurring within the property getters/setters for weightedSpeed and weightedDuration. I can move those calculations out of there.

src/anim/controller/anim-blend-tree.js
ellthompson: Sure

ellthompson:
> * I probably would not mind some other name than syncDurations .. syncProgress, or syncAnimations or similar maybe?
> * I don't see a default value for this in this PR .. but ideally this defaults to true
I've went with syncAnimations and set the default to true.

src/anim/controller/anim-blend-tree-2d-cartesian.js
```diff
@@ -39,11 +41,18 @@ class AnimBlendTreeCartesian2D extends AnimBlendTree {
                 result = math.clamp(1.0 - (AnimBlendTreeCartesian2D._pip.dot(pipj) / pipj.lengthSq()), 0.0, 1.0);
                 if (result < minj) minj = result;
             }
-            this._children[i].weight = minj;
+            child.weight = minj;
```
mvaligursky, line 44: comment on 2 lines above this one, not part of this PR directly but just noticed it.

```
AnimBlendTreeCartesian2D._pip = AnimBlendTreeCartesian2D._p.clone().sub(pi);
result = math.clamp(1.0 - (AnimBlendTreeCartesian2D._pip.dot(pipj) / pipj.lengthSq()), 0.0, 1.0);
```
you have those two static variables to avoid allocations, _pip and _p, and here you clone _p and assign it to _pip. It seems this should be written without cloning and simply by using those static variables.

mvaligursky:
Personally I would not mind some comments, especially in those various 2D blend trees, explaining the concept and math behind them

ellthompson:
> Personally I would not mind some comments, especially in those various 2D blend trees, explaining the concept and math behind them

I'm not sure the concept could be explained concisely in comments here. Perhaps I could link to the paper that was used as a reference? http://runevision.com/thesis/rune_skovbo_johansen_thesis.pdf

src/anim/controller/anim-blend-tree-2d-cartesian.js
ellthompson: I've changed this to use set rather than clone. Also in the process I noticed that I was mistakenly calculating _pip inside the second nested for loop. I've moved it out now :).

src/anim/controller/anim-blend-tree.js
```diff
@@ -48,6 +50,10 @@ class AnimBlendTree extends AnimNode {
         return this._parent ? this._parent.weight * this._weight : this._weight;
     }
 
+    get syncAnimations() {
+        return this._syncAnimations;
+    }
+
     getChild(name) {
```
mvaligursky, line 57: it probably is not expected to happen .. but should there be `return null;` outside the loop?

src/anim/controller/anim-blend-tree.js
```diff
@@ -12,16 +12,18 @@ import { AnimNode } from './anim-node.js';
  * @param {number|Vec2} point - The coordinate/vector thats used to determine the weight of this node when it's part of a {@link AnimBlendTree}.
  * @param {string[]} parameters - The anim component parameters which are used to calculate the current weights of the blend trees children.
  * @param {object[]} children - The child nodes that this blend tree should create. Can either be of type {@link AnimNode} or {@link BlendTree}.
+ * @param {boolean} syncAnimations - If true, the playback duration of each blended animation will be synchonised.
```
mvaligursky, line 15: "playback duration" should be probably changed

mvaligursky (approved):
approved with two minor comments

src/anim/controller/anim-blend-tree.js
mvaligursky: this comment is for ` getChild(name) {` function, not syncAnimations as displayed here .. in diff this is displayed more correctly.

PR #2899, opened 2021-02-22
[Fix] Updated handling of ovrride clear settings on the layer

mvaligursky:
Fixed changes in https://github.com/playcanvas/engine/pull/2882
This fixes the gizmos rendering correctly in Editor (depth buffer was not cleared and so gizmos were depth tested instead of sitting on top)

- Layer. overrideClear and Layer.clearColor are removed as unused, confusing and very unlikely useful
- LayerComposition honors clear flags set on layer in addition to camera

### DIFF
--- a/src/scene/forward-renderer.js
+++ b/src/scene/forward-renderer.js
@@ -3066,8 +3066,8 @@ class ForwardRenderer {
 
             if (camera) {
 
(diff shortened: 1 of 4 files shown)


willeastcott (approved):
A few lint warnings still to fix but otherwise LGTM!

PR #2900, opened 2021-02-22
Added standard material sample of texture based alpha testing

mvaligursky:
Added this to standard material API doc:

Screen Shot 2021-02-23 at 10 43 12 AM

### DIFF
--- a/src/scene/materials/standard-material.js
+++ b/src/scene/materials/standard-material.js
@@ -328,10 +328,10 @@ class Chunks {
  * // Create a new Standard material
  * var material = new pc.StandardMaterial();
  *
- * // assign a texture to diffuse slot
+ * // Assign a texture to the diffuse slot
  * material.diffuseMap = texture;
  *
- * // use the alpha channel of the texture for alpha testing with a reference value of 0.5
+ * // Use the alpha channel of the texture for alpha testing with a reference value of 0.5
  * material.opacityMap = texture;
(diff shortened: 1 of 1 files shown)


raytranuk (approved):
Looks useful

src/scene/materials/standard-material.js
```diff
@@ -324,6 +324,20 @@ class Chunks {
...
+ * @example
+ * // Create a new Standard material
+ * var material = new pc.StandardMaterial();
+ *
+ * // assign a texture to diffuse slot
+ * material.diffuseMap = texture;
+ *
+ * // use the alpha channel of the texture for alpha testing with a reference value of 0.5
+ * material.opacityMap = texture;
+ * material.opacityMapChannel = "a";
```
willeastcott, line 336: [This line](https://github.com/playcanvas/engine/blob/master/src/scene/materials/standard-material.js#L1134) seems to indicate that this _should_ be the default. Although perhaps there's a bug.

willeastcott: Ideally, the default should be documented [here](https://github.com/playcanvas/engine/blob/master/src/scene/materials/standard-material.js#L167).

willeastcott: Confirmed:

image

willeastcott (approved):

PR #2903, opened 2021-02-23
[Fix] when no projection is set for 2d texture, equirect is assumed

mvaligursky:
Fixes viewer loading equirect envmaps


willeastcott:
Shouldn't all `Texture` instances have a valid value for `projection`?

mvaligursky:
Well spotted Will. This is the proper fix.

willeastcott (approved):

PR #2909, opened 2021-02-24
Add mesh-based 3D text support

### LINKED ISSUE #2689: Add a 3D text mesh geometry creator
This feature is a request that were asked multiple times in the forum. 

See my last post about it here: [3d text geometry generator](https://forum.playcanvas.com/t/3d-text-geometry-generator/16690)

### Description

The need for this feature is to be able to generate a 3D text geometry constructed by providing a string of text and other parameters like:

- the font;
- the color
- the space between letter;
- the alignment;
- the font weight (can be mixed with font);

### Examples

- Here an example from Three.js : https://threejs.org/examples/#webgl_geometry_text
- Here the documentation from Three.js: https://threejs.org/docs/#api/en/geometries/TextGeometry

Other engines can also do it:

- Unreal: https://docs.unrealengine.com/en-US/Basics/Actors/3DText/index.html
- Unity: https://docs.unity3d.com/Manual/class-TextMesh.html

### Image

image

willeastcott:
Introducing TextMesh, an easy-to-use script for creating 3D text in your PlayCanvas application.

https://user-images.githubusercontent.com/697563/108997550-9fea3e00-7697-11eb-8e5c-35f27d85f2e7.mp4

You can even apply custom shaders to the generated text:

textshaders

Fixes #2689


scripts/textmesh/text-mesh.js
```diff
@@ -0,0 +1,395 @@
...
+    title: 'Font',
+    description: 'TTF file used as the basis for this 3D text'
+});
+TextMesh.attributes.add('text', {
+    type: 'string',
+    title: 'Text',
+    description: 'The text string to render as a 3D mesh'
+});
+TextMesh.attributes.add('alignment', {
+    type: 'number',
```
kungfooman, line 114: This would be nice for engine-only use, otherwise it throws lots of errors:

```suggestion
    type: 'number',
    default: 0,
```

scripts/textmesh/text-mesh.js
```diff
@@ -0,0 +1,395 @@
...
+    description: 'Depth of the extrusion applied to the text'
+});
+TextMesh.attributes.add('maxCurveSteps', {
+    type: 'number',
+    default: 10,
+    title: 'Max Curve Steps',
+    description: 'Maximum number of divisions applied to bezier based path in a font outline'
+});
+TextMesh.attributes.add('renderStyle', {
+    type: 'number',
```
kungfooman, line 157: This would be nice for engine-only use, otherwise it throws lots of errors:

```suggestion
    type: 'number',
    default: 0,
```

kungfooman:
The script only acccepts materials as assets, however I would like to generate random colors for lots of characters individually. Is there a way to create the script with a non-asset material?

```js
        var text;
        var n;
        var i;
        var letter;
        var material;
        var entity;
        text = "looks good to me";
        n = text.length;
        letters = Array(n);
        for (i=0; i<n; i++) {
            var material = new pc.Material()
            material.diffuse = new pc.Color(Math.random(), Math.random(), Math.random());
            material.update();
            letter = text[i];
            entity = new pc.Entity(`letter ${letter}`);
            entity.addComponent("script");
            entity.script.create("textMesh", {
                attributes: {
                    text: letter,
                    font: assetFont,
                    //alignment: 0,
                    //renderStyle: 0,
                    depth: 0.5,
                    material //: assetMaterial
                }
            });
            entity.translate(i * 0.77, 0, 0);
            app.root.addChild(entity);
        }
```

willeastcott:
@kungfooman This is really a first version and I'm sure it won't offer an API that suits all use cases to begin with. But if you have suggestions for improving the API, let me know. I'm not sure how I'd handle a material per letter. Maybe you would set a material or material asset as normal and then use [pc.MeshInstance#setParameter](https://developer.playcanvas.com/en/api/pc.MeshInstance.html#setParameter) to override the diffuse color on a per-mesh instance basis.

kungfooman:
I am happy with the API 😅 

Is it okay that some normals are `NaN`?

image

I traced it down to this:

image

Basically `isNaN(0 * Infinity) == true` inside `src/scene/procedural.js#calculateNormals`

willeastcott:
Nice detective work. 😄  Any thoughts on a fix? Since this appears to be an issue with the engine rather than the script, I don't think it should block merging.

mvaligursky:
From the screenshot it seems maybe there is a vertex that is not part of any triangle .. and so it's normal cannot be evaluated.
If that's the case, we could either simply set normal to 0,0,0 as it's not being used. But perhaps remove those vertices from the mesh could be a better option.

kungfooman:
I just tried to come up with a fix but I am running in even more strange behaviour. Why is a straight letter producing quadratic bezier curves?

image

This is what I would expect (I commented out `Q` in `TextMesh#parseCommands`), which decreases the mesh size by a factor of nearly 10:

image

I tried it with this font: https://github.com/ocornut/imgui/blob/master/misc/fonts/Cousine-Regular.ttf

So either the font is strange or opentype lol

Otherwise entire PR looks super useful to me and the issues can be just that (further PR's to address specific issues)

Maksims:
> The script only acccepts materials as assets, however I would like to generate random colors for lots of characters individually. Is there a way to create the script with a non-asset material?

You could add color data to vertex buffer per character, and standart material supports colors from vertex data: https://developer.playcanvas.com/en/api/pc.StandardMaterial.html#diffuseVertexColor

kungfooman:
I am not sure yet, just started to print some values in 3d and came across this:

image

Maybe just need to delete the last/same point in each poly

scripts/textmesh/text-mesh.js
```diff
@@ -0,0 +1,397 @@
...
+const p = new pc.Vec2();
+const p1 = new pc.Vec2();
+const p2 = new pc.Vec2();
+
+const vTemp1 = new pc.Vec2();
+const vTemp2 = new pc.Vec2();
+const vTemp3 = new pc.Vec2();
+const vTemp4 = new pc.Vec2();
+const vTemp5 = new pc.Vec2();
+const vTemp6 = new pc.Vec2();
```
yaustar, line 15: Should these variables be scoped so they don't pollute the global space?

mvaligursky (approved):
looks great!

scripts/textmesh/text-mesh.js
willeastcott: I can make them static on the `Polygon` class but that'd just make the code a fair bit more verbose. I know it's not ideal, but let's keep it like this for now. At some point in the future, it'd be good to move the `Polygon` API into the engine and then this won't be a problem.

willeastcott:
So yes, @kungfooman, the fact that generated polygons use the same first and last vertex is wrong. I'll fix in a future PR, because we need to release the engine today. Thanks for the investigation on that - very useful!! 😄

PR #2914, opened 2021-02-25
Basis texture load should error if the Basis module isn't present

yaustar:
Edit: Tue 2 Mar 2021

PR updated after discussion with @slimbuck and @mvaligursky.

The Basis texture loader should error out the load if Basis module is not found/loaded. This allows the asset loader to progress rather than waiting on a basis transcode that won't call the callback.

The app will then able to progress pass the preloading screen, just with missing textures.

image

---

Draft PR to discuss implementation:

The check here is inefficient and also adds a dependency on basis to the asset which I'm not too keen on. Especially if we are thinking of doing tree shaking/specialised builds.

Ideally, I would like to do the logic in `isBasisAvailable` once on pc.Application creation and store it as a flag there instead. Also not keen on getting the extension each time and wonder if that should be cached in the asset/file object?

### DIFF
--- a/src/resources/basis.js
+++ b/src/resources/basis.js
@@ -482,8 +482,8 @@ function basisDownloadFromConfig(callback) {
             // #endif
             return false;
         }
-        return true;
     }
+    return true;
 }
 
 // render thread worker manager
--- a/src/resources/basis.js
+++ b/src/resources/basis.js
@@ -456,6 +456,7 @@ function basisSetDownloadConfig(glueUrl, wasmUrl, fallbackUrl) {
 }
 
 // search for wasm module in the global config and initialize basis
+// returns true if it can find the module
 function basisDownloadFromConfig(callback) {
     if (downloadConfig) {
         // config was user-specified
@@ -475,10 +476,13 @@ function basisDownloadFromConfig(callback) {
                           urlBase + wasmModule.wasmUrl,
                           urlBase + wasmModule.fallbackUrl,
                           callback);
+
+            return true;
         } else {
             // #ifdef DEBUG
             console.warn("WARNING: unable to load basis wasm module - no config was specified");
             // #endif
+            return false;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


Maksims:
Could basis use same code logic as the other variants? Like with compressed textures?
So when basis library is loaded, it would flip a flag, and all assets would load basis. So no basis knowledge by assets code is required.

yaustar:
The problem is that the basis module is only loaded after the first basis texture load is requested 😅

yaustar:
Updated first post

Maksims:
Based on first post update.
If basis transcoder is not in present. Then it should load original texture (non-variant), instead of missing asset.
Basically, same logic as with legacy compressed textures.

yaustar:
In this case it is slightly different as with legacy compression, it falls back to PNG/jpg if the hardware does support the formats.

With Basis, it would transcode to the best format that is supported.

Therefore, not having basis module loaded when trying to load a basis texture is a project setup error, not a hardware dependency/fallback issue.

@slimbuck can explain more on the reasoning :)

Maksims:
What about platforms that do not support Basis? I guess there are some, that do not support WASM. So this means on them no textures will be loaded?

mvaligursky:
We have a JS fallback for when basis wasm cannot be used.

Maksims:
So am I right to assume: when Basis is removed in Editor it will lead to missing textures?
In order to fix that, developer will have to go through all textures, and remove basis compression from them manually?

mvaligursky:
I suspect that is current solution, but @slimbuck can confirm.

src/resources/parser/texture/basis.js
```diff
@@ -42,7 +42,11 @@ class BasisParser {
                         // remove the swizzled flag from the asset
                         asset.file.variants.basis.opt &= ~8;
                     }
-                    basisTranscode(url.load, result, callback, { unswizzleGGGR: unswizzleGGGR });
+                    var basisModuleFound = basisTranscode(url.load, result, callback, { unswizzleGGGR: unswizzleGGGR });
+
+                    if (!basisModuleFound) {
+                        callback('WARNING: Basis module not found. ' + asset.name + ' texture will not be loaded');
```
Maksims, line 48: This will be missleading, as often textures are called something like: `diffuse.jpg`, although it did not try loading actual jpg, but its `.basis` variant.

yaustar: Would `Texture asset 'diffuse.jpg' will not be loaded` be more clear or should be give the path to the basis texture `files/diffuse.basis`?

Maksims: `WARNING: Basis module not found. Asset "' + asset.name + '" basis texture variant will not be loaded.`

yaustar:
Updated warning message

src/resources/basis.js
```diff
@@ -475,11 +476,15 @@ function basisDownloadFromConfig(callback) {
...
-            // #endif
+
+            return true;
         }
+
+        // #ifdef DEBUG
+        console.warn("WARNING: unable to load basis wasm module - no config was specified");
+        // #endif
+
+        return false;
```
slimbuck, line 487: The first block isn't returning true so I'd suggest:
```suggestion
        } else {
            // #ifdef DEBUG
            console.warn("WARNING: unable to load basis wasm module - no config was specified");
            // #endif
            return false;
        }
        return true;
```

src/resources/basis.js
```diff
@@ -479,7 +480,9 @@ function basisDownloadFromConfig(callback) {
             // #ifdef DEBUG
             console.warn("WARNING: unable to load basis wasm module - no config was specified");
             // #endif
+            return false;
         }
+        return true;
     }
```
slimbuck, line 486: Sorry, the return needs to be outside:
```suggestion
    }
    return true;
```

slimbuck (approved):

Issue #2744, opened 2021-01-16
Mesh collider fails to see a collision


LeXXik:
Forum post: https://forum.playcanvas.com/t/adding-mesh-collider-though-script/16677/14
Repro project: https://playcanvas.com/project/755286/overview/collider-test

For some reason, when using an entity set from Editor as a collider (mesh collision), it works fine. However, adding the collider through script fails.

How to repro:
1. The scene has "Lever" entity with 2 scripts `fromEditor` and `fromScript`, where one is using collider from script, the other from editor.
2. Enable `fromEditor` script to use a collider set in the Editor (works).
3. Disable `fromEditor` and enable `fromScript` - collision fails.

On a side note, I am using yaustar's booklet for drawing physical shapes at run-time. The booklet doesn't work, e.g. trying to change opacity of the physical shape will crash, stating that it cannot find meshInstances of undefined. Perhaps related? If not, let me know and I will create a new issue.

yaustar:
There's an error in the documentation (or code depending on the POV 😅 ) `asset` needs an id, not pc.Asset

Fixed fork: https://playcanvas.com/project/768847/overview/collider-test-fork

https://user-images.githubusercontent.com/16639049/109224580-0e242300-77b4-11eb-8457-85365e1809f1.mp4

See PR: https://github.com/playcanvas/engine/pull/2915

yaustar:
Side note: The debug physics visualiser doesn't support mesh collision 😅 It's due an update the next time I need to use it :)

yaustar:
@LeXXik Turns out you forgot to add the trigger script to the from script script file 😅 

```
        entity.addComponent('script');
        entity.script.create('trigger');
```

Fixed version here: https://playcanvas.com/editor/scene/1099325

PR #2915, opened 2021-02-25
Docs fix: Fixed type on pc.Collision.asset as it can be a number or pc.Asset

yaustar:
Related to: https://github.com/playcanvas/engine/issues/2744

### DIFF
--- a/src/framework/components/collision/component.js
+++ b/src/framework/components/collision/component.js
@@ -46,7 +46,7 @@ import { Component } from '../component.js';
  * collision volume's length is aligned. 0 for X, 1 for Y and 2 for Z. Defaults to 1 (Y-axis).
  * @property {number} height The total height of the capsule, cylinder or cone-shaped collision
  * volume from tip to tip. Defaults to 2.
- * @property {Asset} asset The asset id for the model of the mesh collision volume. Defaults to null.
+ * @property {Asset} number The asset id for the model of the mesh collision volume. Defaults to null.
  * @property {Model} model The model that is added to the scene graph for the mesh collision
  * volume.
  */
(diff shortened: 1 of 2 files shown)


yaustar:
We could change the code in system to accept an pc.Asset type though

src/framework/components/collision/component.js
```diff
@@ -46,8 +46,7 @@ import { Component } from '../component.js';
  * collision volume's length is aligned. 0 for X, 1 for Y and 2 for Z. Defaults to 1 (Y-axis).
  * @property {number} height The total height of the capsule, cylinder or cone-shaped collision
  * volume from tip to tip. Defaults to 2.
- * @property {Asset} asset The asset for the model of the mesh collision volume - can also be
- * an asset id. Defaults to null.
+ * @property {Asset} number The asset id for the model of the mesh collision volume. Defaults to null.
```
willeastcott, line 49: Well, first thing is that I think you meant:
```javascript
 * @property {number} asset The asset id for the model of the mesh collision volume. Defaults to null.
```
But as it happens, I think the code should be changed to take an asset. I believe that's more consistent with the model component:

https://github.com/playcanvas/engine/blob/master/src/framework/components/model/component.js#L549-L551

vkalpias: Usually components check if the value is a pc.Asset or an id and accept both.

yaustar:
Updated PR to allow user of pc.Asset  as well as an id.

yaustar:
Turns out the component already handles asset as an pc.Asset or number but both examples of this 'bug' are wrong in a different way (trigger script not added, wrong rotation).

Just a simple doc fix now 😅

willeastcott (approved):

Issue #2801, opened 2021-01-30
Decouple physics engine


LeXXik:
Currently, Ammo is deeply integrated into rigidbody and collision components, making it an integral part of the PC engine. 
I would like to propose decoupling the physics engine into own backend provider.
```
backends/ammo
```
There would be no changes to the current components API. However, instead of processing the physics inside the component, a component would call appropriate method of the currently selected backend. For now, it would be Ammo. 

For example, when a user does `rigidbody.teleport(pos, rot)`, the component would call `backend.teleport(pos, rot)`. Most of the general purpose physics engines do their simulations through iterating simulation steps, so their features are similar. PC component can abstract away the inner workings of the one or another backend. If some physics engine doesn't support a feature, it is easy to notify a user.

This makes it trivial to add support and switching to another general purpose physics engine:
```
backends/physx
backends/rapier
backends/...
```
It also makes it easier for the games to implement and add custom physics engine - it would be simply a matter of adding another backend. All backends could have the same API signature, which would make PC agnostic about which backend is currently handling the physics simulation.

mvaligursky:
Agreed, we've discussed this internally as well. Also it'd be good to perhaps investigate an option of running the physics engine on a worker thread as an option.

willeastcott:
I would love to do this - it's on our roadmap. But it's a big job.

LeXXik:
I have some experience placing Ammo and Rapier in a worker. The method has both pros and cons. 

The pros are obviously in the non-blocking nature of the parallel thread. For example, in Ammo there was a need to do a 1000 step forward-simulation. This introduces a tiny game freeze, when executed in the main thread, but has no such issues, when running in a worker. In general, though, you don't need to have forward simulations, or can cut them into smaller chunks. As such there is little benefit in a worker otherwise, unless you are fighting for 1-2ms in your frame budget for something else. A higher gain would be in using the backend that supports SIMD operations via Wasm. Even running in main thread.

The cons is that you lose a direct access to the state queries of the simulated bodies. For example, it becomes less trivial to check if the body is sleeping. You have to issue a message to the worker, which will do the query in the physics world and report back - that can easily take multiple frames, not mentioning simulation steps. Since the worker is running in own context, you also have no access to the objects there, so using custom callbacks becomes problematic, but can be solved through serializing a function into a string and sending to the worker, which deserializes it and executes as a callback within its own context.

insof:
Any news about this feature?
We need it!

[...]

PR #2931, opened 2021-03-02
collision system fix

LeXXik:
Currently the collision component system implementation removes a rigidbody from the dynamics world:
image

It does so by calling `removeBody()` method of the rigidbody component system directly. It then calls `disableSimulation()` on the component, which again calls `removeBody()` inside of it (if simulation is enabled), which makes it a double request to the dynamics world to remove the same body.

I removed the first call to the `removeBody()`, assuming that if during the next call of `disableSimulation()` the simulation is disabled, then the body is not in the world. If I understand correctly, the only way currently to disable a simulation is by also removing the body from the world. As such, if the simulation is disabled, we can assume the body is not there and does not need to be removed.

### DIFF
--- a/src/framework/components/collision/system.js
+++ b/src/framework/components/collision/system.js
@@ -169,7 +169,6 @@ class CollisionSystemImpl {
     remove(entity, data) {
         var app = this.system.app;
         if (entity.rigidbody && entity.rigidbody.body) {
-            app.systems.rigidbody.removeBody(entity.rigidbody.body);
             entity.rigidbody.disableSimulation();
         }
 


willeastcott (approved):

willeastcott:
Thanks. Well spotted. How did you find this? Presumably it doesn't cause a crash.

LeXXik:
@willeastcott yeah, its silent, so no errors thrown by Ammo. I am working on the PR related to #2801 and noticed that Rapier backend was complaining it was trying to remove a body that was not in the world.

PR #2933, opened 2021-03-02
eulers fix

### LINKED ISSUE #2911: Invalid Euler angles on zero scale
The way Euler angles are calculated, produces an invalid result, if the scale is set to zero. Repro:

```js
entity.setLocalScale(0, 0, 0);
entity.setEulerAngles(10, 10, 10);
    
console.log(entity.getEulerAngles()); // Vec3 { x: NaN, y: NaN, z: 0 }
```

The use case:
We have some items spawn in the world in a little animation, which scales them from zero to normal size.

LeXXik:
Fixes #2911 

If an entity scale is 0, the `getEulerAngles()` returns a vector with invalid (NaN) components. This change adds a check and sets eulers to 0 if scale is 0 on X or Y axis.

```js
entity.setLocalScale(0, 0, 0);
entity.setEulerAngles(10, 10, 10);
    
console.log(entity.getEulerAngles()); // Vec3 { x: NaN, y: NaN, z: 0 }
```

### DIFF
--- a/src/math/mat4.js
+++ b/src/math/mat4.js
@@ -1225,6 +1225,9 @@ class Mat4 {
         sy = scale.y;
         sz = scale.z;
 
+        if (sx === 0 || sy === 0)
+            return eulers.set(0, 0, 0);
+
         m = this.data;
 
         y = Math.asin(-m[2] / sx);
--- a/src/math/mat4.js
+++ b/src/math/mat4.js
@@ -1225,7 +1225,7 @@ class Mat4 {
         sy = scale.y;
         sz = scale.z;
 
-        if (sx === 0 || sy === 0)
+        if (sx === 0 || sy === 0 || sz === 0)
             return eulers.set(0, 0, 0);
 
         m = this.data;


LeXXik:
To be honest, I don't really like this fix because of an additional check. Perhaps it would be better to allow division by zero, but add an `isNaN()` check in debug build and print a warning.

src/math/mat4.js
```diff
@@ -1225,6 +1225,9 @@ class Mat4 {
         sy = scale.y;
         sz = scale.z;
 
+        if (sx === 0 || sy === 0)
```
willeastcott, line 1228: Don't you also have to check whether `z` is `0`?

willeastcott:
I don't think doing two checks against zero is a big performance concern, to be honest.

src/math/mat4.js
LeXXik: Yes, I do, will update.

LeXXik:
Well, it adds additional ALU for all valid reads to check an edge case. I will leave it up to your discretion.

willeastcott (approved):

willeastcott:
This is great. Thanks for the fix!

Issue #2518, opened 2020-11-02
Forward Renderer error when using first layer


LeXXik:
Steps to reproduce:

1. Add an opaque layer to the very top of the layers list (before `World`) and give it a name, e.g. "Custom"
2. Use script that adds a camera with this layer:

```
    var layer = this.app.scene.layers.getLayerByName('Custom');
    var cameraEntity = new pc.Entity();
    cameraEntity.addComponent('camera', {
        clearColor: new pc.Color(0.0, 0.0, 0.0, 0.0),
        layers: [layer.id],
    });    
    this.app.root.addChild(cameraEntity);
```

Forward Renderer throws an error:
![image](https://user-images.githubusercontent.com/5677782/97871189-c49f5c00-1d1c-11eb-8ea3-eb101f01e58b.png)
![image](https://user-images.githubusercontent.com/5677782/97871285-f284a080-1d1c-11eb-89ad-2331255cb43b.png)
`cameraPass` is 0 in this case, resulting in `settings` being `undefined`.

mvaligursky:
This is on my radar to fix already, but I don't have a fix yet. 
It's related to a layer which uses shadow map from directional light used before the shadow map is rendered for that frame.

devcem:
I came here to raise a bug ticket for that issue. Adding Custom layer on directional light solves issue, ALSO on engine code checking "settings" variable on that line and if it's not exist setting the "pass" variable to 0 solves issue.

Here is the line :
https://github.com/playcanvas/engine/blob/0342db03a2e8214a43167f94a0fb82f3f8b6a00e/src/scene/forward-renderer.js#L1369

Here is the screenshot for change :
<img width="500" alt="Screen Shot 2020-11-06 at 18 04 03" src="https://user-images.githubusercontent.com/7432641/98381156-8d9db300-205a-11eb-882c-4418b1e9edbf.png">

PR #2936, opened 2021-03-04
Shadow culling and rendering refactor

mvaligursky:
Shadow culling and rendering refactoring in order to simplify the data / code and fix existing issues with directional shadow map rendering and prepare for cascaded shadow map and shadow atlases.

- directional shadow maps render before the camera that needs it, not after the camera's render target was set active and cleared, to avoid expensive resolve to memory on tiled architectures.
- changes to lights engine example to include all 3 lights types with their shadows, key toggles per light type
- partial support for cascaded shadow maps (culling & rendering to shadow map, but not frustum distribution / shadow map use)
- new private API for cascades: **Light.numCascades**

Screen Shot 2021-03-04 at 3 45 47 PM

Screen Shot 2021-03-04 at 6 08 38 PM

Fixes https://github.com/playcanvas/engine/issues/2518

### DIFF
--- a/src/scene/forward-renderer.js
+++ b/src/scene/forward-renderer.js
@@ -1527,7 +1527,7 @@ class ForwardRenderer {
         var shadowMapStartTime = now();
         // #endif
 
-        var i, j, light, shadowShader, type, shadowCam, shadowCamNode, pass, passes, shadowType, smode;
+        var i, j, light, shadowShader, type, shadowCam, shadowCamNode, passes, shadowType, smode;
         var meshInstance, mesh, material;
         var style;
         var passFlag = 1 << SHADER_SHADOW;
--- a/src/scene/forward-renderer.js
+++ b/src/scene/forward-renderer.js
@@ -2887,12 +2887,12 @@ class ForwardRenderer {
         let globalLightCounter = -1;
         for (let i = 0; i < comp._lights.length; i++) {
             let light = comp._lights[i];
+            const casters = comp._lightShadowCasters[i];
 
             if (light._type === LIGHTTYPE_DIRECTIONAL) {
                 // directional light
                 globalLightCounter++;
                 if (light.castShadows && light.enabled && light.shadowUpdateMode !== SHADOWUPDATE_NONE) {
-                    const casters = comp._lightShadowCasters[i];
                     let cameras = comp._globalLightCameras[globalLightCounter];
                     for (let j = 0; j < cameras.length; j++) {
                         this.cullDirectionalShadowmap(light, casters, cameras[j].camera, comp._globalLightCameraIds[globalLightCounter][j]);
@@ -2901,7 +2901,6 @@ class ForwardRenderer {
             } else {
                 // local light - omni or spot
                 if (light.visibleThisFrame && light.castShadows && light.enabled && light.shadowUpdateMode !== SHADOWUPDATE_NONE) {
-                    const casters = comp._lightShadowCasters[i];
                     this.cullLocalShadowmap(light, casters);
                 }
             }
--- a/src/scene/forward-renderer.js
+++ b/src/scene/forward-renderer.js
@@ -568,7 +568,7 @@ class ForwardRenderer {
     getShadowCamera(device, light) {
         var shadowCam = light._shadowCamera;
 
-        if (shadowCam === null) {
+        if (!shadowCam) {
             shadowCam = light._shadowCamera = createShadowCamera(device, light._shadowType, light._type);
             createShadowBuffer(device, light);
         } else {
@@ -1524,7 +1524,6 @@ class ForwardRenderer {
         // #endif
 
         var i, j, light, shadowShader, type, shadowCam, shadowCamNode, pass, passes, shadowType, smode;
-        var numInstances;
         var meshInstance, mesh, material;
         var style;
         var settings;
@@ -1649,7 +1648,7 @@ class ForwardRenderer {
                     smode = shadowType + type * numShadowModes;
 
                     // Render
-                    for (j = 0, numInstances = visibleLength; j < numInstances; j++) {
+                    for (j = 0; j < visibleLength; j++) {
                         meshInstance = visibleList[j];
                         mesh = meshInstance.mesh;
                         material = meshInstance.material;
@@ -2603,14 +2602,13 @@ class ForwardRenderer {
     }
 
     cullLocalShadowmap(light, drawCalls) {
-        var i, type, shadowCam, shadowCamNode, passes, pass, numInstances, meshInstance, visibleList, vlen, visible;
+        var i, type, passes, pass, numInstances, meshInstance, visibleList, vlen, visible;
         var lightNode;
         type = light._type;
         if (type === LIGHTTYPE_DIRECTIONAL) return;
         light.visibleThisFrame = true; // force light visibility if function was manually called
 
-        shadowCam = this.getShadowCamera(this.device, light);
-
+        let shadowCam = this.getShadowCamera(this.device, light);
         shadowCam.projection = PROJECTION_PERSPECTIVE;
         shadowCam.nearClip = light.attenuationEnd / 1000;
         shadowCam.farClip = light.attenuationEnd;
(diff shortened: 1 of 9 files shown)


src/scene/forward-renderer.js
```diff
@@ -3106,18 +3089,16 @@ class ForwardRenderer {
...
                 // Set the not very clever global variable which is only useful when there's just one camera
                 this.scene._activeCamera = camera.camera;
 
                 // Set camera shader constants, viewport, scissor, render target
                 this.setCamera(camera.camera, renderAction.renderTarget);
 
-                // #ifdef PROFILER
-                draws = this._forwardDrawCalls;
-                // #endif
+                const draws = this._forwardDrawCalls;
```
willeastcott, line 3101: What's the thinking behind accumulating draw calls in all builds instead of just PROFILER build?

src/scene/light.js
```diff
@@ -319,6 +387,8 @@ class Light {
         }
 
         if (key !== this.key && this._scene !== null) {
+            // TODO: most of changes to the key should not invalidate composition,
```
willeastcott, line 390:
```suggestion
// TODO: most of the changes to the key should not invalidate composition,
```

src/scene/layer-composition.js
```diff
@@ -466,31 +317,139 @@ class LayerComposition extends EventHandler {
...
+                    const castersList = casters.list;
+
+                    // add unique meshes from the layer to casters
+                    const meshInstances = layer.shadowCasters;
+                    for (let k = 0; k < meshInstances.length; k++) {
+                        const meshInstance = meshInstances[k];
+
+                        if (!castersSet.has(meshInstance)) {
+                            castersSet.add(meshInstance);
+                            castersList.push(meshInstance);
```
willeastcott, line 369: If you've taken the trouble to define a class to abstract/hide the Set+Array nature of the data structure, should you move `add` etc into `ShadowCasters` too?

src/scene/forward-renderer.js
mvaligursky: they already get accumulated in all builds inside the actual drawing code .. so this just adds them on top level, which runs only few times a frame .. so low cost benefit for the cost we already pay for. Users were asking for some stats to be available in all builds.

src/scene/layer-composition.js
mvaligursky: good suggestion and done!
I'm planning on expanding that class later when I implement shadow atlas to avoid re-rendering all shadow maps when casters don't change .. so this is a good start.

src/scene/forward-renderer.js
willeastcott: OK, sounds good.

willeastcott (approved):

src/scene/layer.js
```diff
@@ -55,20 +55,35 @@ class InstanceList {
...
+
+        // mark them as not processed yet
+        this.visibleOpaque[index].done = false;
+        this.visibleTransparent[index].done = false;
+    }
+
+    // delete entry for a camera with specified index
+    delete(index) {
+        if (index < this.visibleOpaque.length) {
+            this.visibleOpaque.splice(index, 1);
```
slimbuck, line 83: not sure how this list is used, but just wanted to check that splicing the array (and so changing the indexes of the following elements) is ok?

src/scene/light.js
```diff
@@ -95,13 +132,33 @@ class Light {
...
     }
 
     destroy() {
         this._destroyShadowMap();
+        this._renderData = null;
+    }
+
+    // returns LightRenderData with matching camera and face
+    getRenderData(camera, face) {
+
```
slimbuck, line 149: don't think we want the empty line here?

src/scene/light.js
```diff
@@ -156,6 +216,14 @@ class Light {
         return clone;
     }
 
+    get numCascades() {
+        return this.cascades.length;
```
slimbuck, line 220: Doesn't look like the constructor creates this member? (and the clone function doesn't either).

Also, should it be named `_cascades`?

src/scene/layer-composition.js
```diff
@@ -51,26 +52,21 @@ class LayerComposition extends EventHandler {
...
 
-        // all unique lights from all layers, stored both as an array, and also map for fast search (key is light, values is its index in _lights array)
+        // an array of all unique lights from all layers
         this._lights = [];
+
+        // a map of Light to index in _lights for fast lookup
         this._lightsMap = new Map();
 
-        // each light in _lights has entry here at the same index, storing an array and also a set of shadow casters for it
+        // each entry in _lights has entry of type ShadowCasters here at the same index, storing shadow casters for the light
```
slimbuck, line 61: wish we had proper typing info :(

src/scene/layer-composition.js
```diff
@@ -466,31 +317,132 @@ class LayerComposition extends EventHandler {
...
-                    if (index < 0) {
-                        // #ifdef DEBUG
-                        console.warn("Can't find _globalLightCameras[l][i] in cameras");
-                        // #endif
-                        continue;
+        return result;
+    }
+
+    updateShadowCasters() {
+
```
slimbuck, line 330: I'm really not sure about this blank line policy.

I don't like the idea of the engine having inconsistent formatting - some code having (seemingly random) blank lines and other code not.

And I don't favour changing the entire engine to having blank lines.

src/scene/layer-composition.js
mvaligursky: every day.
But there is stuff we can do to define types for private variables and similar .. but need to wait till we switch our backend code editor to be visual code based so that it can handle it
https://medium.com/@trukrs/type-safe-javascript-with-jsdoc-7a2a63209b76

src/scene/layer.js
mvaligursky: yep, these are just caches of culled mesh instances, and get updated every frame .. so this is just removing entry from cache to release memory, but no index depends on this

src/scene/light.js
mvaligursky: both constructor and clone call setter on numCascades, which creates this member, so it is created.

Regarding the underscore, I think the rule here is that if there's a setter and getter for properties, the private variable name starts with _. If there isn't, it does not .. so from outside when you access them on light, you never use _.

src/scene/layer-composition.js
mvaligursky: I like empty line before a comment line .. so this is what I did here. But perhaps we need to have a discussion on this.

src/scene/light.js
slimbuck: Ok you're relying on the setter to create the member. Of course this works, but I much prefer constructors to explicitly 'declare' their members (even setting them to null) for the sake of documentation and future maintenance. Also adding comments in the constructor (as you have done elsewhere) is extremely useful self-documentation. As it stands, I would suggest that looking at this code, it's much less obvious that light has a member called 'cascades'.

Regarding naming, the underscore is meant to indicate whether the property is internal or public (not whether it has an accessor) - same with function names. It can be a useful convention, though I don't think we follow it at all consistently in the engine.

willeastcott: Yeah, this threw me a little bit too when I reviewed. I had to hunt for how it was defined, and while I was able to, it wasn't immediately obvious.

mvaligursky: I'll add the member. Just reading up on performance, it seems it's enough to assign null to a member in constructor to get performance benefit, and it does not need to be assigned consistent type. Note that I can find any conclusive evidence for this. What I was trying to do is to make sure an array is assigned to the member, instead of null and then array. But likely this makes no difference.

mvaligursky: and done.

Maksims: V8 can optimize objects when it is known for their properties in the constructor. If a number of properties change dynamically for an object, then this optimization might not occur, or it will go through a re-optimization process.
So attaching `null` in constructor - is already good, indeed.

mvaligursky: yep. In this case nothing has changed as constructor was calling a setter that was setting it to correct value .. so this is just for clarity when reading the code.
In general this is similar to few places where part of constructor is moved to some init function and called from constructor .. I would not initialize members to null before calling that init function.

slimbuck (approved):

PR #2488, opened 2020-10-22
RenderComponent - new component to replace ModelComponent

mvaligursky:
Superseeds #2345

Related to #2063

Engine examples are in separate PR: https://github.com/playcanvas/engine/pull/2595

Implementation of new RenderComponent, which is functionally similar to ModelComponent, and main reason for its development is to expose Entity hierarchy of GLB files to engine and eventually Editor. Glb file will be loaded as a hierarchy of entities, and RenderComponents would be added on any of these Entities to represent renderable meshes.

This is implemented as @private functionality for the moment, and the API is subject to change.

New API:
- ContainerResource.instantiateModelEntity - creates an entity and adds ModelComponent for model stores as ContainerResource.model.  

- ContainerResource.instantiateRenderEntity - this is the new functionality, and it builds hierarchy of Entities and adds RenderComponents for relevant entities representing meshes assigned to them.

RenderComponent API (similar to ModelComponent API)
- **.meshInstances** - array of meshesInstances this component renders
- **.hide, .show** - allows attached instances to change visibility
- **.generateWireframe** - generates index buffer used to wireframe rendering
- **.castShadows**
- **.receiveShadows**
- **.castShadowsLightmap**
- **.lightmapped, .lightmapSizeMultiplier**
- **.isStatic**
- **.batchGroupId** Assign meshes to a specific batch group
- **.material** The material that will be used to render the meshes (not used on models of type 'asset').

Example:

```
entity.addComponent('render', {
                type: "sphere",
                material: someMaterial,
                castShadows: false,
                castShadowsLightmap: true,
                lightmapped: true
            });
```
 
Other changes:
- Anim system supports Entity hierarchies and RenderComponents (for animation of Entity transforms, and RenderComponent morph weights)
- Batching - supports batching of meshes rendered using RenderComponent
- lightmapping - supports lightmapping of meshes rendered using RenderComponent

PR #2943, opened 2021-03-05
Making RenderComponent related API public

mvaligursky:
RenderComponent API that is different to ModelComponent:
- **ModelComponent.mapping** - not exposed on RenderComponent, override materials are stored directly in MeshInstances
- **ModelComponent.model** - not exposed on RenderComponent, as there is no model and only MeshInstances
- **RenderComponent.renderStyle** - new API to change solid / line / point rendering
- **RenderComponent.rootBone** - new API to expose animation root bone, as render component can be on any child of the hierarchy

new API in the screenshots:

Screen Shot 2021-03-05 at 11 42 07 AM

Screen Shot 2021-03-05 at 11 42 36 AM

Screen Shot 2021-03-05 at 11 55 18 AM

original PR adding RenderComponent: https://github.com/playcanvas/engine/pull/2488

### DIFF
--- a/src/resources/container.js
+++ b/src/resources/container.js
@@ -53,7 +53,7 @@ class ContainerResource {
     /**
      * @function
      * @name ContainerResource#instantiateRenderEntity
-     * @description Instantiates an entity with a render components.
+     * @description Instantiates an entity with a render component.
      * @param {object} [options] - The initialization data for the render component type {@link RenderComponent}.
      * @returns {Entity} A hierarachy of entities with render components on entities containing renderable geometry.
      * @example
--- a/src/resources/container.js
+++ b/src/resources/container.js
@@ -29,12 +29,42 @@ class ContainerResource {
         this.registry = null;
     }
 
+    /**
+     * @function
+     * @name ContainerResource#instantiateModelEntity
+     * @description Instantiates an entity with a model component.
+     * @param {object} [options] - The initialization data for the model component type {@link ModelComponent}.
+     * @returns {Entity} A single entity with a model component. Model component internally contains a hierarchy based on {@link GraphNode}.
+     * @example
+     * // load a glb file and instantiate an entity with a model component based on it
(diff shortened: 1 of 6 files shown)


Maksims:
What is the difference for developer between show/hide and enabled property?
Also, there seems to be no way to find out if it is shown/hidden?
Will it be possible to add only new APIs info compared to ModelComponent? As it will be replacement of ModelComponent.

mvaligursky:
I added the list of differences in the description, good idea.

show/hide/enabled is the same as on model component. Show/hide is fast as it only culls out the mesh but does not rebuild LayerComposition - so good for temporary things. But it does not have getter as it'd need to remove either array (per mesh instance), or yes/no/some result.

Maksims:
> But it does not have getter as it'd need to remove either array (per mesh instance), or yes/no/some result.

Would it be fine to keep a flag within components data, which would reflect its show/hide state?

mvaligursky:
> > But it does not have getter as it'd need to remove either array (per mesh instance), or yes/no/some result.
> 
> Would it be fine to keep a flag within components data, which would reflect its show/hide state?

meshInstance.visible is public as well, so some mesh instances from the array could be in different state

src/resources/container.js
```diff
@@ -29,12 +29,42 @@ class ContainerResource {
...
     instantiateModelEntity(options) {
         var entity = new Entity();
         entity.addComponent("model", Object.assign( { type: "asset", asset: this.model }, options));
         return entity;
     }
 
+    /**
+     * @function
+     * @name ContainerResource#instantiateRenderEntity
+     * @description Instantiates an entity with a render components.
```
slimbuck, line 56:
```suggestion
     * @description Instantiates an entity with a render component.
```

slimbuck (approved):
Approving with suggestion.

willeastcott (approved):

PR #2950, opened 2021-03-08
[Fix] Fix to clone functionality of Entity with RenderComponent

mvaligursky:
clone now works, before it would not clone meshInstances

### DIFF
--- a/src/framework/components/render/system.js
+++ b/src/framework/components/render/system.js
@@ -82,14 +82,14 @@ class RenderComponentSystem extends ComponentSystem {
         delete data.meshInstances;
 
         // clone component
-        var component = this.addComponent(clone, data);
+        const component = this.addComponent(clone, data);
 
         // clone mesh instances
-        const meshes = entity.render.meshInstances.map(mi => mi.mesh);
+        const srcMeshInstances = entity.render.meshInstances;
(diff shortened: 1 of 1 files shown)


src/framework/components/render/system.js
```diff
@@ -71,28 +71,27 @@ class RenderComponentSystem extends ComponentSystem {
...
-            var meshInstances = entity.render.meshInstances;
-            var meshInstancesClone = component.meshInstances;
-            for (i = 0; i < meshInstances.length && i < meshInstancesClone.length; i++) {
-                meshInstancesClone[i].mask = meshInstances[i].mask;
-                meshInstancesClone[i].material = meshInstances[i].material;
-                meshInstancesClone[i].layer = meshInstances[i].layer;
-                meshInstancesClone[i].receiveShadow = meshInstances[i].receiveShadow;
-            }
+        // clone mesh instances
+        const meshes = entity.render.meshInstances.map(mi => mi.mesh);
```
slimbuck, line 88: tiny point: you could be using srcMeshInstances here.

slimbuck (approved):
Approving with comment

src/framework/components/render/system.js
mvaligursky: done, cheers!

PR #2956, opened 2021-03-11
Added hdr parser.

figo2264:
Fixes #

I've added HdrParser from playcanvas viewer to the original engine.

Just wonder why this good feature is not included.

### DIFF
--- a/src/resources/parser/texture/hdr.js
+++ b/src/resources/parser/texture/hdr.js
@@ -1,0 +1,235 @@
+import { http } from '../../../net/http.js';
+
+
+function Reader(data) {
+    this.view = new DataView(data);
+    this.offset = 0;
+}
+
+Object.assign(Reader.prototype, {
+    readLine: function () {
+        var view = this.view;
+        var result = "";
+        while (true) {
+            if (this.offset >= view.byteLength) {
+                break;
+            }
+
+            var c = String.fromCharCode(this.readUint8());
+            if (c === '\n') {
+                break;
+            }
+            result += c;
+        }
+        return result;
+    },
+
+    readUint8: function () {
+        return this.view.getUint8(this.offset++);
+    },
+
+    readUint8s: function (result) {
+        var view = this.view;
+        for (var i = 0; i < result.length; ++i) {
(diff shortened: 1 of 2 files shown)


src/resources/texture.js
```diff
@@ -12,6 +12,7 @@ import { BasisParser } from './parser/texture/basis.js';
 import { ImgParser } from './parser/texture/img.js';
 import { KtxParser } from './parser/texture/ktx.js';
 import { LegacyDdsParser } from './parser/texture/legacy-dds.js';
+import { HdrParser} from "./parser/texture/hdr";
```
yaustar, line 15:
```suggestion
import { HdrParser } from "./parser/texture/hdr";
```

yaustar: Lint fix

willeastcott:
@slimbuck will review this on Monday. 😄

slimbuck (changes requested):
Hi @figo2264 and thanks for this PR!

@willeastcott, just checking you're happy for us to make this HDR parser part of the engine? It would mean we can remove the two versions we have (in the viewer and engine examples), but will increase engine size.

src/resources/parser/texture/hdr.js
```diff
@@ -0,0 +1,234 @@
+import { http } from '../../../net/http.js';
+
+
+function Reader(data) {
```
willeastcott, line 4: Can you make this an ES6 class please?

willeastcott:
I'm OK adding this to the engine if you are @slimbuck. It's not huge and it is being used by the viewer. In future, I'm hoping tree shaking will make it less relevant how big the engine is anyway.

Maksims:
Will it be possible to add support for HDR textures in Editor with help of this PR?

willeastcott:
@Maksims That's actually non-trivial, because currently, HDR images are auto-converted to RGBM PNGs on import into the Editor. So we'd need to add support for HDR import options that allow them to be treated as a runtime format.

slimbuck:
Closing this PR in favour of #2973, which makes a few additional changes.

Issue #2951, opened 2021-03-09
pc.createMesh doesn't use device.extUintElement?


Glidias:
I noticed createMesh function only creates UInt16 array and doesn't go for a big indices Uint32 array format if it's supported.

mvaligursky:
Hi @Glidias - thanks for reporting it, we'll try and address it in the future.
In the meantime, please use the Mesh API to create your meshes which automatically uses 32bit indices if needed. See more details here: https://developer.playcanvas.com/en/api/pc.Mesh.html

And also examples here (called Mesh***) http://playcanvas.github.io/#graphics/mesh-decals.html

PR #2966, opened 2021-03-15
pc.createMesh reimplemented using Mesh API

mvaligursky:
- allows for 32bit index buffer
- sets up blend weights and blend indices correctly to 4 components instead of 2 

Fixes https://github.com/playcanvas/engine/issues/2951


slimbuck (approved):
this is a beautiful thing

PR #2969, opened 2021-03-15
Asset list loader should only complete when all assets in list are loaded

### LINKED ISSUE #2967: Should AssetListLoader take into account assets that are currently loading
Forum post: https://forum.playcanvas.com/t/scene-loading-issues/17839/9?u=yaustar

Code line: https://github.com/playcanvas/engine/blob/38649760a3afd22161d5bdae59f1e25abce4cbeb/src/asset/asset-list-loader.js#L87

It looks like the asset list loader done callback can fire early if an asset in the list is already being loaded as the total count is based on assets that aren't loading or loaded.

This would mean that the callback could fire before all the assets in the list are ready.

Example: https://playcanvas.com/editor/scene/1109443 (throttle network to slower speeds in devtools

image

Looking at the timestamps, of the load callbacks, they should be next to each other rather than 1.5 secs apart.

yaustar:
Fixes #2967

### DIFF
--- a/src/asset/asset-list-loader.js
+++ b/src/asset/asset-list-loader.js
@@ -83,9 +83,11 @@ class AssetListLoader extends EventHandler {
         for (i = 0; i < l; i++) {
             asset = this._assets[i];
 
-            if (!asset.loading && !asset.loaded) {
-                this._registry.load(asset);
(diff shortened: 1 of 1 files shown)


yaustar:
AssetListLoader is not used in the engine, tests or examples

Issue #2596, opened 2020-12-07
Investigate and implement switching to fp16 precision as default on mobile platforms


raytranuk:
Assess workload and measure performance gains and quality.

Inspired by this excellent presentation: https://community.arm.com/cfs-file/__key/communityserver-blogs-components-weblogfiles/00-00-00-20-66/5_2D00_mmg2020_2D00_filament_2D00_romain.pdf

![image](https://user-images.githubusercontent.com/11779961/101367661-49424480-389e-11eb-8777-92ec14abdf18.png)

raytranuk:
This was investigated - on low end devices that are GPU bound doing a global switch resulted in 20-25% increase in performance - and some severe artifacts. More experimentation showed that most the found artifacts can be dealt with by marking some values as `highp`. Further experimentation has been paused and after discussion, we plan to:
1) Mark only things we know are safe to `medp` when a #define is specified
2) Only perform need clamping when the #define is specified
3) Switch the #define on by default on mobile devices
4) Test thoroughly on a range mobile devices

PR #2972, opened 2021-03-16
Shader precision optimizations

raytranuk:
WIP 
- final naming of the macros to be decided
- visual results so far:

Before:
image

After:
image

### DIFF
--- a/src/graphics/program-lib/chunks/TBNObjectSpace.frag
+++ b/src/graphics/program-lib/chunks/TBNObjectSpace.frag
@@ -1,11 +1,11 @@
 void getTBN() {
 
-    vec3 B = cross(dVertexNormalW, vObjectSpaceUpW);
-    vec3 T = cross(dVertexNormalW, B);
+    MEDP vec3 B = cross(dVertexNormalW, vObjectSpaceUpW);
+    MEDP vec3 T = cross(dVertexNormalW, B);
 
     if (dot(B,B)==0.0) // deal with case when vObjectSpaceUpW dVertexNormalW are parallel
     {
-        float major=max(max(dVertexNormalW.x, dVertexNormalW.y),dVertexNormalW.z);
+        MEDP float major=max(max(dVertexNormalW.x, dVertexNormalW.y),dVertexNormalW.z);
 
         if (dVertexNormalW.x==major)
         {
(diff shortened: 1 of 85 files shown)


willeastcott:
Wouldn't it be easier to use default precision qualifiers at the top of the shader, and then only explicitly override the default where something breaks?

raytranuk:
Using default precision was the approach taken in the initial investigation: https://github.com/playcanvas/engine/issues/2596 - however, after discussion in the graphics sync it was agreed (I think) that it would be better to opt-in shader chunks - as it was felt that some precision problems will only be visible after extensive testing and there was a danger of things slipping through if applied globally.

willeastcott:
I think I missed that discussion. If it was me, I would make use of the default qualifiers and then fix specific vars as required. Kinda like we do for GPU skinning precision.

Keeps the code cleaner, leaner, more readable and more maintainable. I literally don't think I've seen any GLSL in my life that specifies a precision for every single possible thing.

mvaligursky:
Yes we discussed this and decided it would be easier to leave fragment shaders on the default high precision, and locally drop parts of the code that we know does not need it to medium precision. Going the other way would created a lot more issues that would be hard to solve. This way we have better control over it and can progressively optimize other chunks as needed.

mvaligursky:
In a way this is similar to the way HLSL handles this .. they expose float and half types (including float2, 3, 4 and half2, 3, 4) .. So this is the define we discussed as possible option today - that would keep the code clean, we'd just `#define half mediump float` and similar

mvaligursky:
looking at PR .. I'd suggest we change line 
`MMEDP vec2 uv = $UV;`
to
`half2 uv = $UV;`

slimbuck:
Agree @mvaligursky, that would clean the code up nicely. Though we might have to be careful that the expanded expression 'mediump float' does not cause unexpected errors.

raytranuk:
The defines currently used in this WIP draft PR are definitely temporary - the grouping is to help me debug and test - I kinda mentioned in the graphics sync I was leaning towards `hfloat` `hvec2` - but `half` and `half2` is also a possibility - though less consistent with GLSL? We could also consider aliasing `vec2` for `float2` if we wanted to go full on HLSL?

raytranuk:
And @slimbuck is right - precision qualifiers can only be used in certain situations - which makes things less straight forward...

mvaligursky:
yes .. when I add some shader code here, I always do float4 instead of vec4 .. and then fix it. So I'm definitely keen on hlsl styling :)
so I guess there's code like this to handle:
`half2 uv = half2(data.x, 0.0);`
I assume the right hand side cannot use precision qualifier? So we use just vec2 there?
Are you aware of some other expctions @raytranuk ? or do we use half when declaring variables, and vec the rest of the time?

PR #2976, opened 2021-03-17
Removed no longer used area lights LUT

mvaligursky:
- binary version is used now: https://github.com/playcanvas/engine/blob/master/examples/assets/binary/area-light-luts.bin
- src data and binary generator: https://github.com/playcanvas/engine/blob/master/examples/assets/scripts/utils/area-light-lut-bin-gen.js


raytranuk (approved):
I didn't realise I didn't delete it

mvaligursky:
I think you did but it was cloned to different folder before that was merged ;)

PR #2978, opened 2021-03-17
Simplified loading of multiple assets in engine examples

mvaligursky:
- few examples use shared asset-loader.js script
- asset are referenced using name instead of array index

Thanks to @ellthompson for help on this.


examples/graphics/area-lights.html
```diff
@@ -36,9 +37,9 @@
             material.useMetalness = true;
 
             if (assetManifest) {
```
ellthompson, line 39: Will this always be true?

mvaligursky: the function is called with null in some cases, where dont want to use loaded textures

ellthompson (approved):

PR #2980, opened 2021-03-17
Allow CJK characters to be treated like words when laying out lines of text

### LINKED ISSUE #2963: Some languages don't text wrap because they don't utilise spaces in the language (eg chinese)
eg
image

We need to add some exceptions in place for some languages to allow them to wrap even if there are no spaces. This makes text unreadable when the text is resized to fit.

raytranuk:
Fixes #2963

### DIFF
--- a/src/framework/components/element/text-element.js
+++ b/src/framework/components/element/text-element.js
@@ -46,7 +46,7 @@ class MeshInfo {
 const LINE_BREAK_CHAR = /^[\r\n]$/;
 const WHITESPACE_CHAR = /^[ \t]$/;
 const WORD_BOUNDARY_CHAR = /^[ \t\-]$/;
-const CJK_CHAR = /^[\u4E00-\u9FFF]*$/;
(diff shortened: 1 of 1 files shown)


src/framework/components/element/text-element.js
```diff
@@ -46,6 +46,7 @@ class MeshInfo {
 const LINE_BREAK_CHAR = /^[\r\n]$/;
 const WHITESPACE_CHAR = /^[ \t]$/;
 const WORD_BOUNDARY_CHAR = /^[ \t\-]$/;
+const CJK_CHAR = /^[\u4E00-\u9FFF]*$/;
```
vkalpias, line 49: Do you need the star in the regex?

yaustar (approved):

src/framework/components/element/text-element.js
raytranuk: removed

vkalpias (approved):

PR #2984, opened 2021-03-18
print a console error if a char is missing in the font

yak32:
this PR prints a message about missing char in a debug mode

image

### DIFF
--- a/src/framework/components/element/text-element.js
+++ b/src/framework/components/element/text-element.js
@@ -733,6 +733,17 @@ class TextElement {
                                 break;
                             }
                         }
+
+                        // #ifdef DEBUG
+                        if (!json.missingChars) {
+                            json.missingChars = new Set();
+                        }
+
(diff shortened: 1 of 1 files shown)


src/framework/components/element/text-element.js
```diff
@@ -733,6 +733,17 @@ class TextElement {
...
                             }
                         }
+
+                        // #ifdef DEBUG
+                        if (!json.missingChars) {
+                            json.missingChars = new Set();
+                        }
+
+                        if (!json.missingChars.has(char)) {
+                            console.error("Character ", char, "is missing in the font");
```
slimbuck, line 743: This should probably be:
```suggestion
                            console.warn("Character '", char, "' is missing from the font");
```

yak32: thanks, i updated the pr

willeastcott (approved):

PR #2986, opened 2021-03-19
AnimStateGraph: Add JSDoc types

kungfooman:
Currently there are no types in `playcanvas.d.ts`:

image

This PR implements all the types the class `AnimStateGraph` accepts:

image

This enables Intellisense for e.g. PC Viewer:

image

Fixing ESLint is not so easy, because I am running into some VSCode/TypeScript issue or so: https://github.com/eslint/eslint/discussions/14235

### DIFF
--- a/src/anim/state-graph/anim-state-graph.js
+++ b/src/anim/state-graph/anim-state-graph.js
@@ -18,7 +18,7 @@
  * @property {string} to
  * @property {number} [time]
  * @property {number} [priority]
- * @property {AnimStateGraph_DataTransitionCondition[]} conditions
+ * @property {AnimStateGraph_DataTransitionCondition[]} [conditions]
  * @property {number} [exitTime]
  * @property {number} [transitionOffset]
  * @property {string} [interruptionSource]
@@ -40,8 +40,8 @@
 /**
  * @typedef {object} AnimStateGraph_DataFromObjectLayer
  * @property {string} name
- * @property {number[]} states
- * @property {number[]} transitions
+ * @property {number[] | AnimStateGraph_DataState[]} states
+ * @property {number[] | AnimStateGraph_DataTransition} transitions
  */
 
 /**
--- a/src/anim/state-graph/anim-state-graph.js
+++ b/src/anim/state-graph/anim-state-graph.js
@@ -1,5 +1,5 @@
 /**
- * @typedef {Object} AnimStateGraph_DataState
+ * @typedef {object} AnimStateGraph_DataState
  * @property {string} name
  * @property {number} [speed]
  * @property {boolean} [loop]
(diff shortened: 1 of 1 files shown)


Maksims:
How does this affects docs generation?
Also, this is something we haven't done before, and looks like specifically for TypeScript, not for everyone else.

kungfooman:
> How does this affects docs generation?

Good point, it currently produces a 404, which should be fixed to produce a proper object description (maybe inline):

image

> Also, this is something we haven't done before, and looks like specifically for TypeScript, not for everyone else.

I mainly did this for `playcanvas-viewer`, which is written in TypeScript, to get rid of `// @ts-ignore` cases. It also helps JavaScript developers in Visual Studio Code or any other editor that supports JavaScript IntelliSense.

vmwxiong:
> How does this affects docs generation?
> Also, this is something we haven't done before, and looks like specifically for TypeScript, not for everyone else.

Eavesdropping in, but I'd love to see support for explicit TS definition contributions somehow. Been doing a bunch of work to get proper types for some things, but not sure how I could possibly contribute the work to the project at all: https://gist.github.com/vmwxiong/eabc60834b32801f5a38de99f4acad80

kungfooman:
> Eavesdropping in, but I'd love to see support for explicit TS definition contributions somehow. 

The generated `playcanvas.d.ts` is kinda as "proper" as it gets, only that generating it through `JSDoc` feels a bit convoluted. The custom doc generation from it doesn't make it more straight forward either, like fixing something in this repo needs another fix in https://github.com/playcanvas/playcanvas-jsdoc-template e.g.

For more "advanced" typing like yours it would make sense to append it directly to `playcanvas.d.ts` after its generation, so it's shared for anyone who installed PlayCanvas through `npm`. Putting extra typing clues in too many places could make it overly confusing though, so one should try to make use of JSDoc whenever possible (and keep `playcanvas-jsdoc-template` synched to display types in an informative/easy/understandable way)

hixio-mh (approved):
#
Reviews and change

kungfooman:
Any is good enough, we shouldn't overcomplicate types like this.

vmwxiong:
`any` is extremely painful to with with as it blows up attempts at building type-safe code, and you need to write and use convoluted type guards every time you want to use something typed with an `any` value.

Disappointed to see the project not moving to support type safe definitions.

willeastcott:
@vmwxiong I think you missed this conversation: https://github.com/playcanvas/playcanvas-viewer/pull/98#issuecomment-966316720

The plan is to make `AnimStateGraph` internal. So this is no longer an issue.

You may have also noticed this week that I have submitted several issues on the repo about improving the Typescript definitions. So saying that 'the project is not moving to support type safe definitions' is not fair or accurate, IMHO. The only time it is difficult to solve typing definitions problems is when it negatively impacts the quality of the API reference manual. Otherwise, everybody is excited to be improving the quality of the typings.

vmwxiong:
Ah, thanks for the clarification, apologies for jumping to conclusions

Issue #2393, opened 2020-08-28
Texture based morphing + WebGL 2 doesn't work on iOS 14 beta


mvaligursky:
float framebuffer extension is not support on WebGl 2 experimental release (EXT_color_buffer_float), even though it is supported on WebGl 1 device.

This will likely be fixed when the extension support is added, but creating this issue for visibility.

mvaligursky:
Support for this is being implemented.
See:
- https://bugs.webkit.org/show_bug.cgi?id=216010
- https://github.com/KhronosGroup/WebGL/issues/3093
- https://github.com/KhronosGroup/WebGL/pull/3146

willeastcott:
Just thought I would mention that I tried this in iOS 14.3 and it's still a problem (with WebGL 2 enabled - still not on by default).

mvaligursky:
Could you please check in webglreport if EXT_color_buffer_half_float is exposed in webgl2 on your device now @willeastcott ? It was exposed in WebGl1 before, but it's expected to come to WebGl2 now as well.

willeastcott:
It's not exposed in iOS 14.3. I'll try updating to iOS 14.4 Beta 2...

willeastcott:
...it's not present in iOS 14.4 Beta 2 either. 😢

mvaligursky:
EXT_color_buffer_half_float is present in iOS 14.5 public beta, and so this can be implemented now for future webgl2 on iOS.

PR #2988, opened 2021-03-19
Support for EXT_color_buffer_half_float - texture morph targets on iOS WebGl2

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/2393

### DIFF
--- a/src/graphics/graphics-device.js
+++ b/src/graphics/graphics-device.js
@@ -652,11 +652,9 @@ class GraphicsDevice extends EventHandler {
         }
 
         // two extensions allow us to render to half float buffers
-        this.textureHalfFloatRenderable = false;
         if (this.extColorBufferHalfFloat) {
             this.textureHalfFloatRenderable = !!this.extColorBufferHalfFloat;
-        }
-        if (!this.textureHalfFloatRenderable && this.extTextureHalfFloat) {
+        } else if (this.extTextureHalfFloat) {
             if (this.webgl2) {
                 // EXT_color_buffer_float should affect both float and halffloat formats
                 this.textureHalfFloatRenderable = !!this.extColorBufferFloat;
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

src/graphics/graphics-device.js
```diff
@@ -650,16 +650,20 @@ class GraphicsDevice extends EventHandler {
         } else {
             this.textureFloatRenderable = false;
         }
-        if (this.extTextureHalfFloat) {
+
```
raytranuk, line 653:
```suggestion
        // two extensions allow us to render to half float buffers
        if (this.extColorBufferHalfFloat) {
            this.textureHalfFloatRenderable = !!this.extColorBufferHalfFloat;
        } else if (this.extTextureHalfFloat) {
            if (this.webgl2) {
                // EXT_color_buffer_float should affect both float and halffloat formats
                this.textureHalfFloatRenderable = !!this.extColorBufferFloat;
            } else {
                // Manual render check for half float
                this.textureHalfFloatRenderable = testRenderable(gl, this.extTextureHalfFloat.HALF_FLOAT_OES);
            }
        } else {
            this.textureFloatRenderable = false;
        }
```

raytranuk: NB. Not sure if the suggestion above will correctly merge into the PR

raytranuk: And its just a suggestion - as I find it a bit more readable.

willeastcott: I think you meant to multi-select the lines of code this suggestion applies to.

mvaligursky: yep, good point. I had those first two conditions the other way before, and this could not work .. but now that I swapped them, good suggestion.

PR #2990, opened 2021-03-19
Clustered lights - Version 1 (No Specular, No shadows, No area lights)

mvaligursky:
This is an initial implementation of clustered lighting for local lights without shadows (omni and spot, with punctual shape - so no area lights). There will be further improvements later to allow these lights to support shadows by rendering them into a shadow atlas. Note that in this implementation, the clustering is done in the World Space instead in more commonly used Camera Space to improve performance, and allow the clustering structure to be used by multiple cameras (for example rendering to texture). It also allows us in the future to avoid rebuilds of this structure if lights don't change).

The difference compared to the current local lights implementation is that while current implementation uses all dynamic lights on all objects in the scene, clustered implementation places these lights into a 3d grid structure with specified size, and when a pixel is rendered, lights it needs are fetched from this 3d grid. Thus it works the best if the overlap of lights is not large. Example being a large level with many local lights with small range in various places inside the level. The test case (see screenshot) replicates this. The level is represented by a large cylinder and many other meshes (spheres and cones), which all get lit by the lights. Both omni and spot lights are separated with low density to minimize amount of lights a pixel needs to use.

_There is a limitation at the moment in that only diffuse part of lighting is evaluated for clustered lights, and specular is not. This will be addressed in the future._

**Changed API:**
constructor of LayerComposition now takes additional parameter, which is a GraphicsDevice. Not passing this parameter is ok for an application using single instance of Application.

**New API (all private at this point, as it might be subject to change):**
- **WorldClusters.enabled** - enables clustered lighting (disabled by default). Needs to be called before rendering, and cannot be changed after.
- **LayerComposition.clusteredLightingCells** - specifies Vec3 with number of cells of the clustering 3d grid
- **LayerComposition.clusteredLightingMaxLights** - maximum number of lights a cell can stored. Ideally with small overlap of lights this could be set to a small number, for example 8.

**Example of API use:**

```
            // enabled clustered lighting. This is a temporary API and will change in the future
            pc.WorldClusters.enabled = true;

            // adjust default clusterered lighting parameters to handle many lights:
            // 1) subdivide space with lights into this many cells:
            app.scene.layers.clusteredLightingCells = new pc.Vec3(12, 16, 12);

            // 2) and allow this many lights per cell:
            app.scene.layers.clusteredLightingMaxLights = 48;

```
**Other changes:**
- some refactor to layer composition in how the shadow casters are updated
	- mostly non functional change, apart from processing each layer only one time (slight optimization)
- optimized light management in the layer by using a Set of lights
- refactored ShadowCasters into LightCompositionData (mostly rename)

**Performance test**
(on very old Huawei phone and MacBook Pro). Note that this depends on light overlap, grid size and similar and are only representative or "common" expected scenario. This also confirmed minimal negative impact when zero or small number of lights are used.
Also note that API allows to move the cost both ways between CPU and GPU. When you use large number of cells (clusteredLightingCells), CPU update is more costly, but GPU is cheaper as light overlap is more precise. On the other hand it small number of cells is used, CPU update is fast, but GPU needs to use many lights for each pixel.

huawei 0 lights
- clustered: 0.5ms
- normal: 0.5ms

PC 0 lights
- clustered: 0.2ms
- normal: 0.2ms

huawei 5 lights
- clustered: 2.1ms
- normal: 2.8ms

huawei 50 lights
- clustered: 14.6ms
- normal: 36ms

PC 50 lights
- clustered: 3ms
- normal: 32ms

**Example** (link valid when released)
http://playcanvas.github.io/#graphics/clustered-lighting.html
This uses clustered lighting, with a grid size of 12, 16, 12, with the maximum number of lights per cell being 48, and rendering 30 point lights and 16 spot lights. In addition there is one shadowing directional light as well.

Screen Shot 2021-04-19 at 2 12 13 PM

References:
https://newq.net/publications/more/s2015-many-lights-course


Maksims:
> PC 50 lights
> 
> * clustered: 32ms
> * normal: 3ms

I believe there is a typo here.

Regarding API, it might be non intuitive to alter clustering information on: `app.scene.layers`.
Also, is it scene related, or global per app?

If global per app, then some interface on app would be reasonable. If it is per scene, then interface on scene.

Also, it being world space, limits the ability for large environments, that can go far from the origin. There are many scenarios where localized, or moved option would be best.
Also, ability to enable clustered lighting or disable (go to non clustered) per meshes.

mvaligursky:
> I believe there is a typo here.

There is no typo there, those are the values I had in a first specific test with very small overlap on lights and large number of cells. So those 50 or so rendered meshes were using 4 lights per pixel on average, vs 50 in existing implementation, thus major speed win.

> Regarding API, it might be non intuitive to alter clustering information on: app.scene.layers.

I thought about this before, and at the moment the existing public API lets you create the LayerComposition and render it .. so the settings need to be here instead of the Scene I believe. Even thought the API to use LayerComposition (by forward renderer) is private. But that can change when we release that API publicly, at the moment clustered lighting API is private.
At the moment the API to enable clustered lighting is global, and density / num lights per cell is per LayerComposition.

> Also, it being world space, limits the ability for large environments, that can go far from the origin.

it's world space, in current implementation, only lights that are visible in the frame are added to the clustering grid, and the world space bounding box defines the area covered by the cell. So in a way the cell structure in the world space moves with the camera and visible lights, and so it should work for large environments.

> Also, ability to enable clustered lighting or disable (go to non clustered) per meshes.

The hope is that if clustered lighting works well, and when shadows support is implemented, this will become the only local light solution in the engine and previous solution will be removed.

mvaligursky:
Ah I see the typo .. it's the other way around. Fixing now :)

src/scene/world-clusters.js
```diff
@@ -0,0 +1,653 @@
...
+        // increase slightly as compression needs value < 1
+        this._maxAttenuation = maxAttenuation + epsilon;
+        this._maxColorValue = maxColorValue + epsilon;
+    }
+
+    updateClusters(gammaCorrection) {
+
+        // clear clusters
+        this.counts.fill(0);
+        this.clusters.fill(0);
```
willeastcott, line 582: TypedArray#fill is not available in IE11:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/fill#browser_compatibility

willeastcott: Have you tested the clustered lighting engine example in IE11?

examples/graphics/clustered-lighting.html
```diff
@@ -0,0 +1,235 @@
...
+        var pointLightList = [];
+        var spotLightList = [];
+        var dirLight = null;
+
+        function run() {
+
+            app.start();
+
+            // enabled clustered lighting. This is a temporary API and will change in the future
+            pc.WorldClusters.enabled = true;
```
willeastcott, line 52: I'd rather not expose the `WorldClusters` class. Maybe have `LayerComposition.clusteredLightingEnabled` instead?

src/graphics/program-lib/programs/standard.js
```diff
@@ -1391,6 +1445,120 @@ var standard = {
...
+    
+                    // spot light
+                    if (lightType > 0.5) {
+                        dAtten *= getSpotEffect(lightDir, innerConeAngleCos, outerConeAngleCos);
+                    }
+    
+                    dAtten *= getLightDiffuse();
+                    dDiffuseLight += dAtten * lightColor;
+                }
+            }
```
willeastcott, line 1558: Could this be a chunk?

src/scene/world-clusters.js
```diff
@@ -0,0 +1,653 @@
...
+
+        if (!this._cells.equals(tempVec3)) {
+            this._cells.copy(tempVec3);
+            this._cellsLimit.copy(tempVec3).sub(Vec3.ONE);
+            this._cellsDirty = true;
+        }
+    }
+
+    createTexture(width, height, format) {
+        var tex = new Texture(this.device, {
```
willeastcott, line 201:
```suggestion
        const tex = new Texture(this.device, {
```

willeastcott:
This PR is an amazing achievement! I have given it a quick review pass. I suggest @slimbuck and @raytranuk also review this one before it is merged since it's so significant.

src/graphics/program-lib/programs/standard.js
raytranuk: Apart from `${clusterTextureFormat}` it does seem suitable for putting into a chunk - this differs significantly from the old lighting loop which was generated mostly in javascript.

slimbuck:
Why is specular missing?

mvaligursky:
> Why is specular missing?

the shader part is not implemented. I suspect it might just be few lines though.

src/scene/world-clusters.js
```diff
@@ -0,0 +1,663 @@
...
+        this.max = new Vec3();
+    }
+}
+
+class WorldClusters {
+    // format for high precision light texture - float
+    static FORMAT_FLOAT = 0;
+
+    // format for high precision light texture - half
+    static FORMAT_HALF = 1;
```
slimbuck, line 57: looks like this is unused?

src/scene/world-clusters.js
```diff
@@ -0,0 +1,663 @@
...
+        } else {
+            WorldClusters.lightTextureFormat = WorldClusters.FORMAT_8BIT;
+        }
+    }
+
+    constructor(device, cells, maxCellLightCount) {
+        this.device = device;
+        this.name = "Untitled";
+
+        // number of time a warning was reported
```
slimbuck, line 82:
```suggestion
        // number of times a warning was reported
```

src/scene/world-clusters.js
```diff
@@ -0,0 +1,663 @@
...
+        // internal list of lights (of type ClusterLight)
+        this._usedLights = [];
+
+        // light 0 is always reserverd for 'no light' index
+        this._usedLights.push(new ClusterLight());
+
+        // register shader uniforms
+        this.registerUniforms(device);
+
+        // alocate textures to store lights
```
slimbuck, line 116:
```suggestion
        // allocate textures to store lights
```

src/scene/world-clusters.js
```diff
@@ -0,0 +1,663 @@
...
+            this.lightsTexture8.destroy();
+            this.lightsTexture8 = null;
+        }
+
+        if (this.lightsTextureFloat) {
+            this.lightsTextureFloat.destroy();
+            this.lightsTextureFloat = null;
+        }
+
+        this.releaseCluxterTexture();
```
slimbuck, line 133:
```suggestion
        this.releaseClusterTexture();
```

src/scene/world-clusters.js
```diff
@@ -0,0 +1,663 @@
...
+    }
+
+    releaseCluxterTexture() {
```
slimbuck, line 136:
```suggestion
    releaseClusterTexture() {
```

src/scene/world-clusters.js
mvaligursky: good idea, will do .. I've just added a polyfill for this one https://github.com/playcanvas/engine/pull/3103

src/scene/world-clusters.js
```diff
@@ -0,0 +1,663 @@
...
+
+                    // store light properties
+                    clusteredLight.light = light;
+                    light.getBoundingBox(tempBox);
+                    clusteredLight.min.copy(tempBox.getMin());
+                    clusteredLight.max.copy(tempBox.getMax());
+
+                    lightIndex++;
+                } else {
+                    console.warn("Clustered lighting: more than " + (this.maxLights - 1) + " lights in the frame, ignoring some.");
```
slimbuck, line 517: should the console.warn be wrapped in _DEBUG?

src/scene/world-clusters.js
```diff
@@ -0,0 +1,663 @@
...
+            maxAttenuation = Math.max(light.attenuationEnd, maxAttenuation);
+
+            const color = gammaCorrection ? light._linearFinalColor : light._finalColor;
+            maxColorValue = Math.max(color[0], maxColorValue);
+            maxColorValue = Math.max(color[1], maxColorValue);
+            maxColorValue = Math.max(color[2], maxColorValue);
+        }
+
+        // increase slightly as compression needs value < 1
+        this._maxAttenuation = maxAttenuation + epsilon;
```
slimbuck, line 577: wouldn't also storing mins be beneficial?

slimbuck (commented):
Lovely work @mvaligursky! Added some small comments.

TBH I think it would be a huge win to include specular in first version.

src/scene/world-clusters.js
mvaligursky: it works in IE11 now, thanks!

examples/graphics/clustered-lighting.html
mvaligursky: done, thanks

src/scene/world-clusters.js
mvaligursky: yep, removed for now, can be re-added later when half precision implementation is done

src/scene/world-clusters.js
mvaligursky: it would, but I'm not sure the benefit would be worth extra add instruction per light .. but I'll keep it in mind should precision would need to be improved.

src/scene/world-clusters.js
mvaligursky: most of them are in the engine, so that's consistent

src/graphics/program-lib/programs/standard.js
mvaligursky: done this part. I guess I was hesitant to do it as it's not finished (all light types, specular, other light properties) .. and it felt like committing it to a chunk would make it harder to change as people could have overwritten it. But that is pretty unlikely.

src/graphics/program-lib/programs/standard.js
```diff
@@ -452,6 +453,59 @@ var standard = {
         return code;
     },
 
+    _addClusteredLighting: function () {
+        return `
```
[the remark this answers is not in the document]
mvaligursky: moved to chunk too

mvaligursky:
Specular part looks more complex and should be done separately. There are multiple options there for existing lights, including Blinn, Phong, GGX, different handling for area lights, clear coat and similar.

src/scene/forward-renderer.js
```diff
@@ -51,6 +52,8 @@ import { GraphNode } from './graph-node.js';
 import { Material } from './materials/material.js';
 import { Mesh } from './mesh.js';
 import { MeshInstance } from './mesh-instance.js';
+import { WorldClusters } from './world-clusters.js';
```
willeastcott, line 55: Delete unreferenced import.

mvaligursky: and done! cheers.

src/scene/world-clusters.js
slimbuck: yes most (all?) console logs are wrapped in _DEBUG, but this one isn't?

slimbuck (approved):

raytranuk (approved):
excellent stuff

Issue #2958, opened 2021-03-11
Mesh collision example


LeXXik:
Would be nice to add a mesh collision example to engine only and User Manual pages.

Issue #2944, opened 2021-03-05
Allow Collider component to accept Render as a source of mesh, in addition to Model


mvaligursky:
- CollisionComponent accepts .model, needs to accept .render as well
- the asset could be then either of these
- CreatePhysicalShape needs to change to handle both (see screenshot) - it would just loop over mesh instances of model or render and do the same with both
- fix editor/public/js/editor/gizmo/gizmo-collision.js to handle the presence of a renderAsset

![Screen Shot 2021-03-05 at 4 02 33 PM](https://user-images.githubusercontent.com/59932779/110143064-bd718300-7dce-11eb-98fe-4c2f7fe66a5b.png)

LeXXik:
Hmm, does it mean the `CollisionComponentData` would get a new property `.render`?

mvaligursky:
also perhaps implement example using render:
https://github.com/playcanvas/engine/issues/2958

PR #2995, opened 2021-03-23
Collision component support of Meshes specified using Render instead of Model

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/2944
Fixes https://github.com/playcanvas/engine/issues/2958

**Public API**:

- Container exposes array of Render classes, instead of Meshes directly, and this was made public:
**Container.renders** - Array of assets of renders in the GLB container
- **Collision.renderAsset** - render asset can be specified in addition to existing Collision.asset accepting a model asset.

**Other changes:**
- collider uses mesh API to access indices and so supports 32bit indicies / large meshes now

**Example:**
- extended falling-shapes.html example to use torus mesh collider, and changed it to use RenderComponent instead of ModelComponent

Screen Shot 2021-03-23 at 3 50 21 PM


raytranuk (commented):
minor comment: the use of the name `renders` is is bit misleading - I wonder if there is a better name?

mvaligursky:
> minor comment: the use of the name `renders` is is bit misleading - I wonder if there is a better name?

Maybe with possibly lot of changes? It's RenderComponent, that uses Render class, and array of those is called renders.
Equivalent to LightComponent, Light class, lights array.
Not sure this is worth changing now?

willeastcott (approved):

PR #2998, opened 2021-03-24
Directional Blend Trees speed fix

ellthompson:
When using a directional blend tree with synchronised animation durations, the blend weight of each animation clip's speed should be normalised using the total weight of all clips.

### DIFF
--- a/src/anim/controller/anim-blend-tree-2d-directional.js
+++ b/src/anim/controller/anim-blend-tree-2d-directional.js
@@ -58,7 +58,7 @@ class AnimBlendTreeDirectional2D extends AnimBlendTree {
             const child = this._children[i];
             child.weight = child._weight / weightSum;
             if (this._syncAnimations) {
-                child.weightedSpeed = child.animTrack.duration / child.absoluteSpeed / weightedDurationSum;
+                child.weightedSpeed = child.animTrack.duration / child.absoluteSpeed / weightedDurationSum * weightSum;
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/anim/controller/anim-blend-tree-2d-directional.js
```diff
@@ -58,7 +58,7 @@ class AnimBlendTreeDirectional2D extends AnimBlendTree {
             const child = this._children[i];
             child.weight = child._weight / weightSum;
             if (this._syncAnimations) {
-                child.weightedSpeed = child.animTrack.duration / child.absoluteSpeed / weightedDurationSum;
+                child.weightedSpeed = child.animTrack.duration / child.absoluteSpeed / weightedDurationSum * weightSum;
```
willeastcott, line 61: Is it just me or is the operator precedence a bit scary here? Anybody prefer some brackets?

slimbuck: I was thinking the same: two divides... no brackets... whaaaa?! :D

mvaligursky: yeah. instead of `a / b / c` I prefer `a / (b * c)`

ellthompson: I've split this onto two lines for a bit a clarity as to what's being calculated.

willeastcott: Great. But I would still put the brackets around the first div in case readers forget that division takes precedence over multiplication.

ellthompson: Done

willeastcott (approved):

PR #2999, opened 2021-03-24
Normalize test

slimbuck:
TextElement normalises text in order to remain consistent with bidi results (introduced in https://github.com/playcanvas/engine/pull/2979), but unfortunately IE doesn't support string.normalize().

So test for presence of string.normalize() before invoking.

### DIFF
--- a/src/framework/components/element/text-element.js
+++ b/src/framework/components/element/text-element.js
@@ -287,7 +287,13 @@ class TextElement {
         if (text === undefined) text = this._text;
 
         // get the list of symbols
-        this._symbols = string.getSymbols(text.normalize('NFC'));
+        if (text.normalize) {
+            // NOTE: we must normalize text here in order to be consistent with the number of
+            // symbols returned from the bidi algorithm. If we don't, then it's possible bidi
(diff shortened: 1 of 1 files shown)


mvaligursky (approved):

willeastcott (approved):

yaustar:
Considering this is a crash bug, should we use a polyfill instead as the edge case of Hindi will crash IE11 since we can't normalise the string?

willeastcott:
@yaustar Since devs will hit this so incredibly rarely, I suggest they add a polyfill themselves rather than us adding a few extra KB to the engine.

LeXXik:
This should be mentioned in the docs then. Captain's out.

slimbuck:
Not quite @yaustar. The crash bug was due to incorrect handling of RTL text. However RTL text wasn't supported in IE anyway. This fix means that text elements work in IE again, though not with RTL text.

As @willeastcott says, if devs want RTL to work in IE they can provide a polyfill (in which case BIDI and this change will work).

As it stands, the crash we're talking about never happened in IE because BIDI isn't supported there.

PR #3002, opened 2021-03-24
[Fix] Reproject frag shader is compatible with IE11

mvaligursky:
fixes error: Initializer for const variable must initialize to a constant value

### DIFF
--- a/src/graphics/reproject-texture.js
+++ b/src/graphics/reproject-texture.js
@@ -76,6 +76,7 @@ function reprojectTexture(device, source, target, specularPower = 1, numSamples 
         "#define SOURCE_FUNC " + sourceFunc + "\n" +
         "#define TARGET_FUNC " + targetFunc + "\n" +
         "#define NUM_SAMPLES " + numSamples + "\n\n" +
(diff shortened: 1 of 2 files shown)


src/graphics/reproject-texture.js
```diff
@@ -76,6 +76,7 @@ function reprojectTexture(device, source, target, specularPower = 1, numSamples
         "#define SOURCE_FUNC " + sourceFunc + "\n" +
         "#define TARGET_FUNC " + targetFunc + "\n" +
         "#define NUM_SAMPLES " + numSamples + "\n\n" +
+        "#define NUM_SAMPLES_SQRT " + Math.sqrt(numSamples).toFixed(1) + "\n\n" +
```
slimbuck, line 79:
```suggestion
        "#define NUM_SAMPLES_SQRT " + Math.round(Math.sqrt(numSamples)).toFixed(1) + "\n\n" +
```

slimbuck: original code didn't do this, but probably should have

willeastcott (approved):

PR #3009, opened 2021-03-25
fix: type argument for attributes had regressed from a union of strings to plain string

TheJonRobinson:
We noticed the type of `ScriptAttributes.add` parameter `args.type` regressed from a union type to a plain string, hence losing the nicer type-checking. Found the original and merged back here.

### DIFF
--- a/src/script/script-attributes.js
+++ b/src/script/script-attributes.js
@@ -175,7 +175,7 @@ class ScriptAttributes {
      * @description Add Attribute.
      * @param {string} name - Name of an attribute.
      * @param {object} args - Object with Arguments for an attribute.
-     * @param {string} args.type - Type of an attribute value. Can be one of "boolean", "number", "string", "json", "asset", "entity", "rgb", "rgba", "vec2", "vec3", "vec4" or "curve".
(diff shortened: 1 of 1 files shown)


willeastcott:
Thanks for this, @TheJonRobinson. I just checked the Git Blame for that line and it doesn't seem to be a regression (unless I'm missing something?).

I'm assigning this to @vkalpias for confirmation that this won't negatively impact the autocompletion support in the code editor.

TheJonRobinson:
@willeastcott Looks to be sneaking in in this mega-commit: https://github.com/playcanvas/engine/pull/2876

willeastcott:
Ah yes, you are correct! So that change was also made by @vkalpias so he'll be the perfect person to explain the reason behind this when he's back on Monday. 😄

vkalpias:
Am I missing something? The type is supposed to be a string with one the values mentioned in the comments.

willeastcott:
If the parameter is limited to a set of specific string values, you can put them as literals where the type should be. The Typescript compiler can then error if you pass a string literal to a function that is not in the valid set.

vkalpias (approved):
Oh those were all strings, thought they were saying type can be a boolean, number etc. Approving

willeastcott (approved):

willeastcott:
This is now deployed in 1.40.2 @TheJonRobinson! 😄

PR #3014, opened 2021-03-26
Use correct texture filtering defaults

slimbuck:
This PR fixes gltf sampler logic by using sane defaults when sampler members aren't present. Previously, when members of the sampler weren't present in the file mipmapping was being disabled.

### DIFF
--- a/src/resources/parser/glb-parser.js
+++ b/src/resources/parser/glb-parser.js
@@ -1542,7 +1542,7 @@ var createResources = function (device, gltf, bufferViews, textureAssets, option
 };
 
 var applySampler = function (texture, gltfSampler) {
-    var defaultSampler = {
+    var sampler = {
         magFilter: 9729,
         minFilter: 9987,
         wrapS: 10497,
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

src/resources/parser/glb-parser.js
```diff
@@ -1571,11 +1571,11 @@ var applySampler = function (texture, gltfSampler) {
     };
 
     if (texture) {
-        gltfSampler = gltfSampler || defaultSampler;
-        texture.minFilter = getFilter(gltfSampler.minFilter);
-        texture.magFilter = getFilter(gltfSampler.magFilter);
-        texture.addressU = getWrap(gltfSampler.wrapS);
-        texture.addressV = getWrap(gltfSampler.wrapT);
+        Object.assign(sampler, gltfSampler || { });
+        texture.minFilter = getFilter(sampler.minFilter);
```
mvaligursky, line 1575: It'd be probably nicer to change getFilter and getWrap to take a default value to return when undefined is passed in. Something like:
`texture.minFilter = getFilter(sampler.minFilter, FILTER_LINEAR);`
This would be more readable than the `minFilter: 9987,`
Or let it return undefined and supply engine default here
`texture.minFilter = getFilter(sampler.minFilter) || FILTER_LINEAR;`

slimbuck: Agree: https://github.com/playcanvas/engine/pull/3015

PR #3015, opened 2021-03-26
Prefer clearer code

slimbuck:
Change based on suggestion by @mvaligursky in https://github.com/playcanvas/engine/pull/3014#discussion_r602249234


mvaligursky (approved):
nice!

Issue #2508, opened 2020-10-29
Create engine example to render lines


mvaligursky:
- using mesh API
- show depth test on / off, alpha on / off at least

Issue #2069, opened 2020-05-10
pc.Application#renderLine fails when rendering to world layer


willeastcott:
The API reference has the following code example

```
// Render a 1-unit long white line into the world layer
var start = new pc.Vec3(0, 0, 0);
var end = new pc.Vec3(1, 0, 0);
var color = new pc.Color(1, 1, 1);
var worldLayer = app.scene.layers.getLayerById(pc.LAYERID_WORLD);
app.renderLine(start, end, color, {
    layer: worldLayer
});
```

However, if you do this, all mesh instances in the world layer stop rendering and you just see the line.

To repro:

1. Create a new, blank project in the Editor.
2. Create a script with the following code:

```
var Line = pc.createScript('line');

// initialize code called once per entity
Line.prototype.initialize = function() {
    this.vecA = new pc.Vec3(0, 0, 0);
    this.vecB = new pc.Vec3(3, 3, 3);
    this.color = new pc.Color(1, 0, 1);
};

// update code called every frame
Line.prototype.update = function(dt) {
    var worldLayer = this.app.scene.layers.getLayerById(pc.LAYERID_WORLD);
    this.app.renderLine(this.vecA, this.vecB, this.color, {
        layer: worldLayer
    });
};
```

3. Launch the scene with the script enabled and you'll just see the line.
4. Launch the scene again with the script disabled, and you'll just see the box + plane.

mvaligursky:
I had a quick look, and it seems the engine's immediate rendering removes all MeshInstances from a layer, not only the ones added by immediate rendering. I think it was designed to only use pc.LAYERID_IMMEDIATE exclusively, and no other layer.

https://github.com/playcanvas/engine/blob/a6741dabeda5262b4b3c5019314481e0c9c2beec/src/scene/immediate.js#L379-L385

This needs to be fixed / refactored.

willeastcott:
Yeah, it's a bit of a mess. Even the function signature is pretty horrible since it has a number of variations and different parameters can have varying types.

PR #3019, opened 2021-03-26
Refactoring, fixes and improvements of the line rendering

mvaligursky:
- refactored line rendering
  - fixed issue with the system simply wiping all mesh instances on a layer a line is rendered to
  - proper handling of depthTest (each submitted line batch can specify it instead of first batch settings this for all batches)
  - mesh instance used to render lines is not added/removed from the layer each frame causing expensive LayerComposition update, and uses .visible flag instead
  - moved functionality to its own folder / separated classes to individual files / renamed
- added engine example to show / test the functionality

**New Public API:**
- **drawLine**(start, end, color = Color.WHITE, depthTest = true, layer = ..)
    Draws a single line using a single color, positions specified as Vec3
- **drawLines**(positions, colors, depthTest = true, layer = ..)
    Draws multiple lines, positions specified using Vec3
- **drawLineArrays**(positions, colors, depthTest = true, layer ..)
    Draws multiple lines, positions and colors specified as flat arrays of numbers for performance

**New Private API (subject to change):**
- **drawWireSphere**(center, radius, color = Color.WHITE, segments = 20, depthTest = true, layer = .. )
    Draws a wireframe sphere
- **drawWireAlignedBox**(minPoint, maxPoint, color = Color.WHITE, depthTest = true, layer = ..)
    Draws a wireframe AABB

**Deprecated Public API:**
- **Application.renderLine**
- **Application.renderLines**

Fixes https://github.com/playcanvas/engine/issues/2508
Fixes https://github.com/playcanvas/engine/issues/2069
Discussion: https://github.com/playcanvas/engine/discussions/3073

Screenshot 2021-10-05 at 13 36 37


src/scene/immediate/immediate.js
```diff
@@ -231,19 +233,36 @@ class Immediate {
         batch.addLines(this.cubePositions, color);
     }
 
-    renderWireSphere(center, radius, color, options) {
+    renderWireBoundingBox(boundingBox, color, options) {
+        const min = boundingBox.getMin();
+        const max = boundingBox.getMax();
+        const cubeWorldPos = [
+            new Vec3(min.x, min.y, min.z), new Vec3(min.x, max.y, min.z), new Vec3(max.x, max.y, min.z), new Vec3(max.x, min.y, min.z),
```
Maksims, line 240: This does allocations on each render, which would be best to be avoided.

mvaligursky: yep, was planning to share the array with the function above .. all maybe remove it and share array with the spere and build flat float arrays

mvaligursky: and done, thanks!

mvaligursky:
Please discuss here on the API changes
https://github.com/playcanvas/engine/discussions/3073

LeXXik:
Perhaps a rotated cone or cylinder could be added to show how `renderWireCube` is used.

slimbuck:
For complex debug shapes (sphere, cylinder, capsule etc) it's quite limiting not being able to specify an arbitrary transform (especially for physics viz).

In the past I've have taken the approach of building meshes for each shape and then instanced them with matrices (instead of building unique vertices per instance). This has ended up being more efficient overall and quite useful in the more complex debug scenarios.

Wondering how hard this approach would be here? Could we easily add it later (API-wise).

mvaligursky:
I had a version of wired box rendering using a matrix, but at the end decided to remove it .. the problem was that it was rendering a unit sized box, and using world space matrix of an entity with a mesh wasn't working well as it was not taking into account the size of the AABB of the mesh. 

We can add something like this if needed in the future, but I thing generating lines as current implementation is faster due to single render call per layer, instead of render call per primitive.

src/scene/immediate/immediate-batch.js
```diff
@@ -0,0 +1,105 @@
...
+    onPreRender() {
+        if (this.positions.length > 0) {
+
+            // update mesh vertices
+            this.mesh.setPositions(this.positions);
+            this.mesh.setColors(this.colors);
+            this.mesh.update(PRIMITIVE_LINES, false);
+
+            if (!this.meshInstance) {
+                identityGraphNode.worldTransform = Mat4.IDENTITY;
```
yaustar, line 83: Are we okay with this being a reference to an object or should this be a copy operation?

eg `identityGraphNode.worldTransform.copy(Mat4.IDENTITY)`

yaustar (approved):
I can't see any obvious issues. May want to wait for a more graphics orientated person?

examples/src/examples/graphics/lines.tsx
```diff
@@ -0,0 +1,198 @@
...
+                const nextEntity = meshes[(i + 1) % meshes.length];
+                app.drawLine(entity.getPosition(), nextEntity.getPosition(), pc.Color.MAGENTA);
+
+                // store positions and colors of lines connecting objects to a center point
+                grayLinePositions.push(entity.getPosition(), new pc.Vec3(0, 10, 0));
+                grayLineColors.push(pc.Color.GRAY, pc.Color.GRAY);
+            }
+
+            // render all gray lines
+            app.drawLines(grayLinePositions, grayLineColors);
```
yaustar, line 190: Where is the `drawlines` function? I can't see it in the PR

src/scene/immediate/immediate-batch.js
mvaligursky: I'm ok with this, it never changes.

examples/src/examples/graphics/lines.tsx
mvaligursky: application.js line 1739

yaustar: Ah, GitHub collapsed the diff 😓

slimbuck (approved):
Looking great! Please wait for sign off from @willeastcott before merging, since this PR changes API!

willeastcott:
A few comments from me. This feels like some kind of debug rendering utility library. It's quite opinionated, since there are many ways to render a wire sphere, for example. So my gut feeling is that the shape rendering belongs in the `scripts` folder and 'low-level' line rendering belongs in the core `playcanvas.js`. I can anticipate devs saying 'I want an oriented box function' and 'I want to be able to render line strips' ad infinitum. So that's my concern. But obviously, this is a nice step forward in general so I'm not sure I'd want to block it. 😄

mvaligursky:
I can remove the box / sphere rendering, or at least make them private, and adjust the example to not use them if you think that's a better way for now? I understand your concern, but on the other hand I think basic debug functionality like this is very useful to have easily available.

willeastcott:
What's @slimbuck's thoughts on this?

slimbuck:
From a code structure point of view, I agree that splitting the functionality into low-level immediate mode rendering and higher-level shape drawing would be great. Though this is probably a change we can make at any point.

As for where the high-level shape drawing should live, I'm not exactly sure. If we put them in engine scripts/extras it will help with engine bloat, but then editor users will find it much harder to leverage. Also, using the scripts in es6 applications becomes very difficult because our scripts are not modularised.

I guess the ultimate solution is tree-shaking the engine, but that's a whole other discussion.

mvaligursky:
Based on an offline discussion - the agreement was to make drawWireSphere and drawWireAlignedBox private functions, allowing us to figure out a long term strategy before exposing them in some form.

willeastcott (approved):

PR #3020, opened 2021-03-27
Migrate core, math and shape source to use let and const

willeastcott:
Migrates entire core, math and shape codebase from `var` to use `let` and `const`. Note that, for now, API ref docs still use `var`.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -10,22 +10,22 @@
  * @name _typeLookup
  * @description Create look up table for types.
  */
-var _typeLookup = function () {
-    var result = { };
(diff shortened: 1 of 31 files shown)


slimbuck (approved):
me likey! any idea whether static analysis/lint will tell us if anything is assigning to a const var?

willeastcott:
My process is:

1. For a set of files, check that all instances of `var` can be validly changed to `let`.
2. Search and replace `var` to `let`.
3. Hack our lint rules to enable 'prefer-const' which detect `let` variables and auto-fixes them to be `const`.

PR #3021, opened 2021-03-27
Update all calls of vector scale to mulScalar

willeastcott:
This PR updates all calls to `Vec2/3/4#scale` to `Vec2/3/4#mulScalar`. This is in response to a recent update to the math API.


slimbuck (approved):

Maksims:
Tbh, not a fond of this naming change. `.scale` is easy to remember and simple to write, and a very common method to use.

willeastcott:
`scale` is definitely nice and clear and succinct. In isolation, I think there's no question that it is a better function name.

However, when looking at the vector APIs holistically, we now have very nice and predictable symmetry betweeen `add`, `sub`,`mul` and `div`. And I've left `scale` defined for people who want to continue using it.

PR #3036, opened 2021-03-29
Make arrow function parentheses consistent

willeastcott:
Arrow functions can optionally omit parentheses when they have exactly one parameter. In all other cases the parameter(s) must be wrapped in parentheses. This rule enforces the consistent use of parentheses in arrow functions.

Here, we ensure parentheses are used everywhere.


Maksims:
Definitely worth a specific section in https://github.com/playcanvas/engine/blob/master/.github/CONTRIBUTING.md to describe more on the recommended use of arrow functions, and where not to use, also on parentheses.

mvaligursky (approved):

PR #3044, opened 2021-03-29
Prefer triple to double equals

willeastcott:
According to [ESLint](https://eslint.org/docs/rules/eqeqeq):

> It is considered good practice to use the type-safe equality operators === and !== instead of their regular counterparts == and !=.
>
> The reason for this is that == and != do type coercion which follows the rather obscure Abstract Equality Comparison Algorithm. For instance, the following statements are all considered true:
>
> [] == false
> [] == ![]
> 3 == "03"
>
>If one of those occurs in an innocent-looking statement such as a == b the actual problem is very difficult to spot.

This PR does not apply this principle to all occurrences of `==` and `!=`. It leaves those testing against `null`, where sometimes the use was intentional. I still believe those should be reworked, but for now, this PR just deals with the 'safe' instances.

### DIFF
--- a/src/resources/parser/glb-parser.js
+++ b/src/resources/parser/glb-parser.js
@@ -677,7 +677,7 @@ var createMesh = function (device, gltfMesh, accessors, bufferViews, callback, d
 
                         // indices
                         var numFaces = outputGeometry.num_faces();
-                        if (geometryType == decoderModule.TRIANGULAR_MESH) {
+                        if (geometryType === decoderModule.TRIANGULAR_MESH) {
                             var bit32 = outputGeometry.num_points() > 65535;
                             numIndices = numFaces * 3;
                             var dataSize = numIndices * (bit32 ? 4 : 2);
@@ -971,7 +971,7 @@ var createMaterial = function (gltfMaterial, textures, disableFlipV) {
             color = specData.diffuseFactor;
             // Convert from linear space to sRGB space
             material.diffuse.set(Math.pow(color[0], 1 / 2.2), Math.pow(color[1], 1 / 2.2), Math.pow(color[2], 1 / 2.2));
-            material.opacity = (color[3] != null) ? color[3] : 1;
+            material.opacity = color[3];
         } else {
             material.diffuse.set(1, 1, 1);
             material.opacity = 1;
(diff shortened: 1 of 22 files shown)


src/resources/parser/glb-parser.js
```diff
@@ -971,7 +971,7 @@ var createMaterial = function (gltfMaterial, textures, disableFlipV) {
             color = specData.diffuseFactor;
             // Convert from linear space to sRGB space
             material.diffuse.set(Math.pow(color[0], 1 / 2.2), Math.pow(color[1], 1 / 2.2), Math.pow(color[2], 1 / 2.2));
-            material.opacity = (color[3] != null) ? color[3] : 1;
+            material.opacity = color[3];
```
slimbuck, line 974: was this change intentional?

slimbuck (commented):
This PR has the potential to break existing code. Are we ok with that?

src/resources/parser/glb-parser.js
willeastcott: Yes. The spec states that `diffuseFactor` is a 4-element array. I don't see a circumstance where the fourth element can be `null` or `undefined`. Am I missing something?

willeastcott:
> This PR has the potential to break existing code. Are we ok with that?
Every PR has the potential to break existing code. You have to point out where it actually does break code. I believe it doesn't. 😄

src/resources/parser/glb-parser.js
willeastcott: See here: https://kcoley.github.io/glTF/extensions/2.0/Khronos/KHR_materials_pbrSpecularGlossiness/#specular---glossiness

slimbuck: Makes sense.

willeastcott:
Just thinking about it, unease and swapping `==` with `===` is a great argument for not using `==`! 😆

slimbuck (commented):
I was thinking about https://github.com/playcanvas/engine/pull/3044/files#diff-b8e35fe9f6d2040346e21c7bf17fbdd77c192d84ca3714046bb4a690abc108fcR209 - where users could currently be passing in a string value instead of number. Code that currently skips a string set - say newValue='0' and oldValue=0 - will no longer be ignored.

This is a pretty contrived example though so probably worth the risk.

slimbuck (approved):

PR #3048, opened 2021-03-30
Fix for crash: unbind (texture) asset references when material is destroyed

### LINKED ISSUE #2982: Texture loaded for unloaded material
There might be a bug in the following use case:
If the same textures are used on different materials
and one of the materials is unloaded and then one of the shared textures is loaded
then the bound _onTextureLoad() method is called for the unloaded material causing an error.
eg “TypeError: Cannot set property ‘emissiveMap’ of undefined”

Here is the forum post https://forum.playcanvas.com/t/unloading-and-reloading-materials/18838?u=kulodo133
There is an example project there.

Here are some more details on the issue.  I think this is what is happening.

When the material’s asset references are created these callbacks are set:

https://github.com/playcanvas/engine/blob/master/src/resources/material.js#L238

When the material is unloaded the callbacks for the assetReferences aren’t removed:

https://github.com/playcanvas/engine/blob/master/src/resources/material.js#L135

So when a texture from the assetReference is subsequently loaded the callback is called for the now unloaded material asset:

https://github.com/playcanvas/engine/blob/master/src/resources/material.js#L163

and this causes the error as the material asset is undefined.

https://github.com/playcanvas/engine/blob/master/src/resources/material.js#L143

So you probably want to either remove the callbacks, which is what I did in my hack workaround ( but then they need to be recreated later when the material is reloaded) or check that the material is not undefined in these callbacks. 

There is another related engine bug.

https://github.com/playcanvas/engine/issues/2260

raytranuk:
Fixes #2982

### DIFF
--- a/src/scene/materials/standard-material.js
+++ b/src/scene/materials/standard-material.js
@@ -738,7 +738,7 @@ class StandardMaterial extends Material {
         for (const asset in this._assetReferences) {
             this._assetReferences[asset]._unbind();
         }
-        this._assetReferences = {};
+        this._assetReferences = null;
         this._validator = null;
 
         super.destroy();
--- a/src/scene/materials/standard-material.js
+++ b/src/scene/materials/standard-material.js
@@ -738,7 +738,7 @@ class StandardMaterial extends Material {
         for (const asset in this._assetReferences) {
             this._assetReferences[asset]._unbind();
         }
-        this._assetReferences = null;
+        this._assetReferences = {};
         this._validator = null;
 
         super.destroy();
--- a/src/scene/materials/standard-material.js
+++ b/src/scene/materials/standard-material.js
@@ -727,6 +727,22 @@ class StandardMaterial extends Material {
 
         this.dirtyShader = false;
     }
+
+    /**
+     * @function
+     * @name StandardMaterial#destroy
+     * @description Removes this material from the scene and possibly frees up memory from its shaders (if there are no other materials using it).
+     */
+    destroy() {
+        // unbind (texture) asset references
+        for (const asset in this._assetReferences) {
+            this._assetReferences[asset]._unbind();
+        }
+        this._assetReferences = null;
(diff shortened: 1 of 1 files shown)


src/scene/materials/standard-material.js
```diff
@@ -727,6 +727,22 @@ class StandardMaterial extends Material {
 
         this.dirtyShader = false;
     }
+
+    /**
+     * @function
+     * @name StandardMaterial#destroy
+     * @description Removes this material from the scene and possibly frees up memory from its shaders (if there are no other materials using it).
+     */
+    destroy() {
```
mvaligursky, line 736: base class Material contains destroy() .. should this call it?

src/scene/materials/standard-material.js
```diff
@@ -727,6 +727,22 @@ class StandardMaterial extends Material {
...
+        // unbind (texture) asset references
+        for (const asset in this._assetReferences) {
+            this._assetReferences[asset]._unbind();
+        }
+        this._assetReferences = null;
```
yaustar, line 741: Wondering it is safer to make _assetReferences into the state that was in in the constructor (ie `_assetReferences = {}`) like we do in the Material.destroy function

willeastcott: But if you call `destroy`, the object is done. Kaput. Finished. Are you expecting it to be used again? I mean, also, that would allocate an object for no decent reason.

raytranuk: reverted - if there is expects a valid object it throw an exception - which is good so we can fix that code (hopefully shouldn't be)

willeastcott (approved):

raytranuk:
>Approving - but for extra points on this, a unit test never hurts. But that shouldn't block merging this, IMHO.

I'll merge now - and write a unit test today (after finishing the other issue I'm looking at) :)

PR #3050, opened 2021-03-31
Improve CJK character line breaking

### LINKED ISSUE #2963: Some languages don't text wrap because they don't utilise spaces in the language (eg chinese)

raytranuk:
- now deals with non-alphanumeric characters like punctuation.
- supports Korean and Japanese unicode ranges
- supports zero width space for Thai
- added some small kana and CJK punctuation characters that do not allow line break

Fixes #2963

### DIFF
--- a/src/framework/components/element/text-element.js
+++ b/src/framework/components/element/text-element.js
@@ -56,6 +56,7 @@ const ALPHANUMERIC_CHAR = /^[a-z0-9]$/i;
 // AC00—D7AF Hangul Syllables
 // D7B0—D7FF Hangul Jamo Extended-B
 const CJK_CHAR = /^[\u1100-\u11ff]|[\u3000-\u9fff]|[\ua960-\ua97f]|[\uac00-\ud7ff]$/;
+const NO_LINE_BREAK_CJK_CHAR = /^[〕〉》」』】〙〗〟ヽヾーァィゥェォッャュョヮヵヶぁぃぅぇぉっゃゅょゎゕゖㇰㇱㇲㇳㇴㇵㇶㇷㇸㇹㇺㇻㇼㇽㇾㇿ々〻]$/;
 
 // unicode bidi control characters https://en.wikipedia.org/wiki/Unicode_control_characters
 const CONTROL_CHARS = [
@@ -911,14 +912,9 @@ class TextElement {
                     _xMinusTrailingWhitespace = _x;
                 }
 
-                // char is space, tab, or dash
-                const isWordBoundary = WORD_BOUNDARY_CHAR.test(char);
-                // char is CJK character boundary
-                const isCJKBoundary = CJK_CHAR.test(char) && ((nextchar !== null) && (WORD_BOUNDARY_CHAR.test(nextchar) || ALPHANUMERIC_CHAR.test(nextchar)));
-                // next character is CJK character
-                const isNextCJK = (nextchar !== null) && CJK_CHAR.test(nextchar);
-
-                if (isWordBoundary || isCJKBoundary || isNextCJK) {
+                if ((WORD_BOUNDARY_CHAR.test(char)) || // char is space, tab, or dash
+                    (CJK_CHAR.test(char) && ((nextchar !== null) && !NO_LINE_BREAK_CJK_CHAR.test(nextchar) && (WORD_BOUNDARY_CHAR.test(nextchar) || ALPHANUMERIC_CHAR.test(nextchar)))) || // char is a CJK character and next character is a CJK boundary
+                    ((nextchar !== null) && CJK_CHAR.test(nextchar) && !NO_LINE_BREAK_CJK_CHAR.test(nextchar))) { // next character is a CJK character that can be a whole word
                     numWordsThisLine++;
                     wordStartX = _xMinusTrailingWhitespace;
                     wordStartIndex = i + 1;
--- a/src/framework/components/element/text-element.js
+++ b/src/framework/components/element/text-element.js
@@ -43,25 +43,23 @@ class MeshInfo {
     }
 }
 
+const LINE_BREAK_CHAR = /^[\r\n]$/;
+const WHITESPACE_CHAR = /^[ \t]$/;
+const WORD_BOUNDARY_CHAR = /^[ \t\-]|[\u200b]$/; // NB \u200b is zero width space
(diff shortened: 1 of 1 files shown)


src/framework/components/element/text-element.js
```diff
@@ -902,7 +904,8 @@ class TextElement {
 
                 var isWordBoundary = WORD_BOUNDARY_CHAR.test(char);
                 var isCJK = CJK_CHAR.test(char);
-                if (isWordBoundary || isCJK) { // char is space, tab, or dash OR CJK
+                var isNextCJK = nextchar ? CJK_CHAR.test(nextchar) : false;
```
slimbuck, line 907: should this test be `(nextchar !== null) ? ...` ?

raytranuk: yes - I'll change it

raytranuk: done - thanks!

src/framework/components/element/text-element.js
```diff
@@ -902,7 +904,8 @@ class TextElement {
 
                 var isWordBoundary = WORD_BOUNDARY_CHAR.test(char);
                 var isCJK = CJK_CHAR.test(char);
-                if (isWordBoundary || isCJK) { // char is space, tab, or dash OR CJK
+                var isNextCJK = (nextchar !== null) ? CJK_CHAR.test(nextchar) : false;
```
slimbuck, line 907:
```suggestion
                const isNextCJK = (nextchar !== null) && CJK_CHAR.test(nextchar);
```

and make the previous two variables const too!

src/framework/components/element/text-element.js
```diff
@@ -900,9 +911,14 @@ class TextElement {
                     _xMinusTrailingWhitespace = _x;
                 }
 
-                var isWordBoundary = WORD_BOUNDARY_CHAR.test(char);
-                var isCJK = CJK_CHAR.test(char);
-                if (isWordBoundary || isCJK) { // char is space, tab, or dash OR CJK
+                // char is space, tab, or dash
+                const isWordBoundary = WORD_BOUNDARY_CHAR.test(char);
+                // char is CJK character boundary
+                const isCJKBoundary = CJK_CHAR.test(char) && (((nextchar !== null) && (WORD_BOUNDARY_CHAR.test(nextchar)) || ALPHANUMERIC_CHAR.test(nextchar)));
```
slimbuck, line 917: I think the brackets should be like so:
```suggestion
                const isCJKBoundary = CJK_CHAR.test(char) && ((nextchar !== null) && (WORD_BOUNDARY_CHAR.test(nextchar) || ALPHANUMERIC_CHAR.test(nextchar)));
```

raytranuk: thanks :)

src/framework/components/element/text-element.js
```diff
@@ -900,9 +912,9 @@ class TextElement {
                     _xMinusTrailingWhitespace = _x;
                 }
 
-                var isWordBoundary = WORD_BOUNDARY_CHAR.test(char);
-                var isCJK = CJK_CHAR.test(char);
-                if (isWordBoundary || isCJK) { // char is space, tab, or dash OR CJK
+                if ((WORD_BOUNDARY_CHAR.test(char)) || // char is space, tab, or dash
+                    (CJK_CHAR.test(char) && ((nextchar !== null) && !NO_LINE_BREAK_CJK_CHAR.test(nextchar) && (WORD_BOUNDARY_CHAR.test(nextchar) || ALPHANUMERIC_CHAR.test(nextchar)))) || // char is a CJK character and next character is a CJK boundary
+                    ((nextchar !== null) && CJK_CHAR.test(nextchar) && !NO_LINE_BREAK_CJK_CHAR.test(nextchar))) { // next character is a CJK character that can be a whole word
```
slimbuck, line 917: How would you feel about this?

```suggestion
            // char is space, tab, or dash
            const isWordBoundary = () => {
                return WORD_BOUNDARY_CHAR.test(char);
            };

            const isValidNextChar = () => {
                return (nextchar !== null) && !NO_LINE_BREAK_CJK_CHAR.test(nextchar);
            };

            // char is a CJK character and next character is a CJK boundary
            const isNextCJKBoundary = () => {
                return CJK_CHAR.test(char) && (WORD_BOUNDARY_CHAR.test(nextchar) || ALPHANUMERIC_CHAR.test(nextchar));
            };

            // next character is a CJK character that can be a whole word
            const isNextCJKWholeWord = () => {
                return CJK_CHAR.test(nextchar);
            };

            if (isWordBoundary() || (isValidNextChar() && (isNextCJKBoundary() || isNextCJKWholeWord()))) {
```

slimbuck: (NOTE: the proposed code above is untested!)

raytranuk: I like it - it is more readable and I think the logic is equivalent - I'll commit it and just double check with my test case - thanks!

raytranuk: I tested it and it seems to work - I also moved the function declarations out of the loop (as lint thinks its unsafe)

slimbuck (approved):
Nice job, approving! Pls wait for approval from someone else on the team before merging.

src/framework/components/element/text-element.js
```diff
@@ -900,9 +931,7 @@ class TextElement {
                     _xMinusTrailingWhitespace = _x;
                 }
 
-                var isWordBoundary = WORD_BOUNDARY_CHAR.test(char);
-                var isCJK = CJK_CHAR.test(char);
-                if (isWordBoundary || isCJK) { // char is space, tab, or dash OR CJK
+                if (isWordBoundary() || (isValidNextChar() && (isNextCJKBoundary() || isNextCJKWholeWord()))) {
```
willeastcott, line 934: I find this `if` statement way easier to read now - nice. But personally, I would define the 4 functions at module scope near the top instead of inside _updateMeshes (where they'll be allocated on every call). They'd just need the char passing as an argument.

raytranuk: makes sense - moved functions to module scope

willeastcott (approved):

PR #3060, opened 2021-04-01
WebXR Input Source: linear velocity

Maksims:
Linear Velocity has been added for handheld devices in WebXR specs: https://pr-preview.s3.amazonaws.com/cabanier/webxr/pull/1182.html#dom-xrpose-linearvelocity

This also been implemented in the latest Oculus Browser: https://twitter.com/rcabanier/status/1377459759876505606

This PR implements input source linear velocity and also emulates it with the same logic if it is not available/supported.
So if the input source has a grip space - we guarantee to provide linear velocity.

### New APIs:
```javascript
// pc.XrInputSource
inputSource.getLinearVelocity() // Vec3 with world space linear velosity in units per second, or null if input source is not handheld
```

### DIFF
--- a/src/xr/xr-input-source.js
+++ b/src/xr/xr-input-source.js
@@ -54,19 +54,30 @@ class XrInputSource extends EventHandler {
 
         this._ray = new Ray();
         this._rayLocal = new Ray();
-        this._grip = false;
+        this._gripSpace = xrInputSource.gripSpace;
+        this._targetRaySpace = xrInputSource.targetRaySpace;
         this._hand = null;
+        this._velocitiesAvailable = false;
 
         if (xrInputSource.hand)
             this._hand = new XrHand(this);
 
-        this._localTransform = null;
-        this._worldTransform = null;
         this._position = new Vec3();
         this._rotation = new Quat();
+        this._dirtyLocal = true;
         this._localPosition = null;
         this._localRotation = null;
-        this._dirtyLocal = true;
+        this._localTransform = null;
+        this._worldTransform = null;
+
+        if (this._gripSpace) {
+            this._localPositionLast = new Vec3();
+            this._localPosition = new Vec3();
+            this._localRotation = new Quat();
+            this._localTransform = new Mat4();
+            this._worldTransform = new Mat4();
+            this._linearVelocity = new Vec3();
+        }
 
         this._selecting = false;
         this._squeezing = false;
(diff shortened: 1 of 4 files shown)


LeXXik:
How about returning a zero vector if it is not available? This would match the behavior we get when trying to read the velocity of a kinematic or static rigidbody in physics.

Maksims:
> How about returning a zero vector if it is not available? This would match the behavior we get when trying to read the velocity of a kinematic or static rigidbody in physics.

There are input sources that have no such concept as position, for example: touch in AR session, or voice commands input source, or gaze style input source. So we should not provide linear velocity for such types, just like we don't provide position and rotation data for such input sources. So it stays consistent for XR input source concepts.

src/xr/xr-manager.js
```diff
@@ -426,7 +426,7 @@ class XrManager extends EventHandler {
         });
     }
 
-    update(frame) {
+    update(frame, dt) {
```
mvaligursky, line 429: is the _dt_ used / passed in?

src/xr/xr-input-source.js
```diff
@@ -185,26 +201,34 @@ class XrInputSource extends EventHandler {
...
+                var gripPose = frame.getPose(this._gripSpace, this._manager._referenceSpace);
                 if (gripPose) {
-                    if (! this._grip) {
-                        this._grip = true;
+                    this._grip = true;
+                    this._dirtyLocal = true;
 
-                        this._localTransform = new Mat4();
-                        this._worldTransform = new Mat4();
+                    const now = Date.now();
```
mvaligursky, line 210: to be more consistent with the engine, perhaps you should use
```
import { now } from  '../core/time.js';
const now = now();
```

mvaligursky (approved):

src/xr/xr-input-source.js
```diff
@@ -299,6 +325,19 @@ class XrInputSource extends EventHandler {
         return this._localRotation;
     }
 
+    /**
+     * @function
+     * @name XrInputSource#getLinearVelocity
+     * @description Get the linear velocity (units per second) of input source if it is handheld ({@link XrInputSource#grip} is true). Otherwise it will return null.
+     * @returns {Vec3|null} The world space linear velocity of handheld input source.
+     */
+    getLinearVelocity() {
```
willeastcott, line 334: We already have `RigidBodyComponent#linearVelocity`. Should this be a property too?

Maksims: Fair enough, but whole XR APIs uses function notation. Also it is only a get function, there is no way to set linear velocity as it is sourced from underlying APIs.

willeastcott: Then it would just be a getter property. No requirement to have a setter. Functions are fine for `getPosition`/`setPosition` etc to mirror the function on `GraphNode` but generally, we should use properties where it makes sense.

Maksims: If we want to use properties, then all other methods would need to be changed for consistency reasons. Preferring a consistent design for WebXR APIs.

willeastcott: The WebXR APIs shouldn't define their own API style. There should be an engine-wide style. If the WebXR APIs don't adhere to that, we can gradually migrate and alias to something that is consistent with the overall engine.

Maksims: I've used `get*` style which is in line with `Entity` and other APIs. And as mentioned before, this is not settable property like in rigidbody's case.
This particular method can be changed to match very specific case of rigidbody.
Then should we change: `getOrigin` and `getDirection` of `XrInputSource` to be similar to `Ray`, but again mentioning: origin and direction of input source cannot be set by developer, and provided by an API. Also they do call a function inside to update transform, just like `GraphNode.getPosition` does.

And regarding `getPosition` and `getRotation`, should we change them to simply: `position` and `rotation`?

To be honest, it is already not consistent within engine, but it is very clear that if it is a property - then it can be set. In XR case, those methods cannot be set, they only provide data.

willeastcott: There are a large number of gettable but not settable properties in the API. Just check out `GraphNode` for example:

https://developer.playcanvas.com/en/api/pc.GraphNode.html#children

Read-only properties are 100% OK.

`getPosition` and `getRotation` could be aliased to `position` and `rotation` maybe. I think it's a little complicated and needs some serious thought.

TBH, we should probably improve `Ray` and have the `XrInputSource` provide a `Ray` instead of separate `origin` and `direction`.

Maksims: We can definitely start a discussion and do WebXR-wide refactor, with such changes.

Maksims:
Would love to get a second look at this PR, so it can be merged/resolved.

willeastcott (approved):
LGTM! Thanks @Maksims!

Issue #2941, opened 2021-03-05
Issue applying post processing on cameras rendering to renderTarget


mvaligursky:
forum thead: https://forum.playcanvas.com/t/apply-bloom-to-certian-materials/18455/12
it seems the postprocessed render target does not get copied to camera's render target or similar.

possibly related:
https://github.com/playcanvas/engine/pull/2939
https://github.com/playcanvas/engine/issues/2940

PR #3069, opened 2021-04-07
Post-effects can be applied on camera with render target

mvaligursky:
Note that render target cannot use antialiasing, as post-effects don't support it at the moment.

Fixes https://github.com/playcanvas/engine/issues/2941

### DIFF
--- a/examples/graphics/render-to-texture.html
+++ b/examples/graphics/render-to-texture.html
@@ -6,6 +6,7 @@
     <meta name="viewport" content="width=device-width, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0">
     <link rel="icon" type="image/png" href="../playcanvas-favicon.png" />
     <script src="../../build/playcanvas.js"></script>
+    <script src="../assets/scripts/asset-loader.js"></script>
     <script src="../../build/playcanvas-extras.js"></script>
     <style>
         body { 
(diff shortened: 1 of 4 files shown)


willeastcott:
I think my only concern here is that the render to texture example is being polluted a bit with another feature demonstration. Maybe if, by default, the post effect is not applied and you can enabled it by pressing a key/button. Or you can add a PCUI control like the animation examples. But devs might look at it and wonder what they're looking at because the 'TV' is now showing the same scene as  the main camera.

willeastcott:
BTW, instead of hue-saturation, how about this kind of effect: 

https://www.shadertoy.com/view/XtK3W3

😄

raytranuk:
>BTW, instead of hue-saturation, how about this kind of effect:

>https://www.shadertoy.com/view/XtK3W3

+1 on this - would make the example also show developers what cool possibilities using rendering to texture could have with post processes.

raytranuk (approved):
Code to allow post processes to render to offscreen render targets looks good - I added a comment on the changes to the engine example

mvaligursky:
I removed changes to the example to keep it simple, and moved the example changes to separate PR to be made nicer demo of rendering to texture later.

willeastcott (approved):

willeastcott:
I think that's a good decision, @mvaligursky.

PR #3071, opened 2021-04-08
Remove dependency of Light on Application

mvaligursky:
- Light gets device in constructor and stores it locally to avoid dependency on the Application class
- as program library stringifies options that contain array of lights, which now stores device, device class implement toJSON to disable serialization
- these changes caused DefaultAnimBinder to fail type checks, and so it's dependency on Entity was removed as well, which removed 2 circular dependencies

### DIFF
--- a/src/graphics/graphics-device.js
+++ b/src/graphics/graphics-device.js
@@ -697,6 +697,11 @@ class GraphicsDevice extends EventHandler {
         }
     }
 
+    // don't stringify GraphicsDevice to JSON by JSON.stringify
+    toJSON(key) {
+        return undefined;
+    }
+
     // #if _DEBUG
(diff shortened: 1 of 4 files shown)


ellthompson (approved):

src/graphics/graphics-device.js
```diff
@@ -697,6 +697,11 @@ class GraphicsDevice extends EventHandler {
         }
     }
 
+    // don't stringify GraphicsDevice to JSON by JSON.stringify
+    toJSON(key) {
+        return undefined;
+    }
```
willeastcott, line 703: Why is this needed in this PR? Seems unrelated.

willeastcott (approved):

PR #3074, opened 2021-04-09
fix material asset unload and reload

### LINKED ISSUE #3061: _assignTexture fault
function, MaterialHandler.prototype._assignTexture, changes materialAsset.data[Texture parameter name];
that will lead to wrong destroied Texture assigned to Material , if I unload the Texture asset and Material asset then reload them.

raytranuk:
Fixes #3061

This PR:
- adds a texture and cubemap `unload` event callback to clean up texture references when a texture is unloaded
- removes a not-so-robust setting of asset id to loaded resource - which was causing problems if a texture asset is unloaded and reloaded
- fixes and tidies up a calls to `_validate()` in the material json parser.

### DIFF
--- a/src/resources/material.js
+++ b/src/resources/material.js
@@ -144,10 +144,8 @@ class MaterialHandler {
         materialAsset.resource[parameterName] = texture;
     }
 
-    // assign a placeholder texture while waiting for one to load
-    // placeholder textures do not replace the data[parameterName] value
-    // in the asset.data thus preserving the final asset id until it is loaded
-    _getPlaceholderTexture(parameterName, materialAsset) {
+    // returns the correct placeholder texture for the texture parameter
+    _getPlaceholderTexture(parameterName) {
         // create placeholder textures on-demand
         if (!this._placeholderTextures) {
             this._createPlaceholders();
@@ -159,6 +157,7 @@ class MaterialHandler {
         return texture;
     }
 
+    // assign a placeholder texture while waiting for one to load
     _assignPlaceholderTexture(parameterName, materialAsset) {
 
         materialAsset.resource[parameterName] = this._getPlaceholderTexture(parameterName, materialAsset);
@@ -173,21 +172,11 @@ class MaterialHandler {
         this._assets.load(textureAsset);
     }
 
-    _onTextureRemove(parameterName, materialAsset, textureAsset) {
+    _onTextureRemoveOrUnload(parameterName, materialAsset, textureAsset) {
         const material = materialAsset.resource;
         if (material) {
-            if (material[parameterName] === textureAsset.resource) {
-                this._assignTexture(parameterName, materialAsset, null);
-                material.update();
-            }
-        }
-    }
-
-    _onTextureUnload(parameterName, materialAsset, textureAsset) {
-        const material = materialAsset.resource;
-        if (material) {
-            if (material[parameterName] === textureAsset.resource) {
-                this._assignTexture(parameterName, materialAsset, null);
+            if (materialAsset.resource[parameterName] === textureAsset.resource) {
+                this._assignPlaceholderTexture(parameterName, materialAsset);
                 material.update();
             }
         }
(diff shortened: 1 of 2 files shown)


slimbuck:
@raytranuk could you please give an overview of the changes you've made and why in the PR description?

src/resources/material.js
```diff
@@ -140,14 +140,14 @@ class MaterialHandler {
...
-        materialAsset.data[parameterName] = texture;
+        // NB removed swapping out asset id for resource here
         materialAsset.resource[parameterName] = texture;
     }
 
     // assign a placeholder texture while waiting for one to load
     // placeholder textures do not replace the data[parameterName] value
     // in the asset.data thus preserving the final asset id until it is loaded
-    _assignPlaceholderTexture(parameterName, materialAsset) {
+    _getPlaceholderTexture(parameterName, materialAsset) {
```
slimbuck, line 150: seems like this comment is no longer valid? please update it with latest usage

src/resources/material.js
```diff
@@ -156,7 +156,12 @@ class MaterialHandler {
         const placeholder = PLACEHOLDER_MAP[parameterName];
         const texture = this._placeholderTextures[placeholder];
 
-        materialAsset.resource[parameterName] = texture;
+        return texture;
+    }
+
+    _assignPlaceholderTexture(parameterName, materialAsset) {
```
slimbuck, line 162: descriptive usage comment could be useful here

src/resources/parser/material/json-standard-material.js
```diff
@@ -167,8 +167,9 @@ class JsonStandardMaterialParser {
         if (!this._validator) {
             this._validator = new StandardMaterialValidator();
         }
-        this._validator.validate(data);
-
+        if (!data.validated) {
```
slimbuck, line 170: looks like `_validate` is completely unnecessary and should be removed.

it is being called from `parse()` which immediately after invoking `validate` does `initialize()`, which handles validation anyway.

src/resources/material.js
```diff
@@ -178,8 +183,18 @@ class MaterialHandler {
         }
     }
 
+    _onTextureUnload(parameterName, materialAsset, textureAsset) {
```
slimbuck, line 186: mustn't you also handle cubemap unload?

src/resources/material.js
```diff
@@ -178,8 +183,18 @@ class MaterialHandler {
...
+        const material = materialAsset.resource;
+        if (material) {
+            if (material[parameterName] === textureAsset.resource) {
+                this._assignTexture(parameterName, materialAsset, null);
```
slimbuck, line 190: should this revert to placeholder texture or is null correct?

src/resources/material.js
raytranuk: I'll refactor the comment

src/resources/material.js
raytranuk: I'll refactor the comment

src/resources/parser/material/json-standard-material.js
raytranuk: I left it in because I thought there may be code paths that would require it? There is a similar check for the `validated` property for probably similar reasons. If you are certain there are no code paths that would require the validation I can remove it.

src/resources/material.js
raytranuk: I wasn't certain - but the test case didn't show any errors - so decided not to change too much  - there is a null check and assigning of placeholders in the bind function so I assumed placeholders get assigned in that code path.

src/resources/material.js
raytranuk: ah - yes - I will add _onCubeMapUnload

src/resources/material.js
raytranuk: Found a case that would require placeholders - so now replace null with placeholders

src/resources/material.js
raytranuk: added - NB there is no existing placeholder system for cubemaps - I didn't add one in this PR - but added an issue to investigate whether they are needed: https://github.com/playcanvas/engine/issues/3076

src/resources/parser/material/json-standard-material.js
raytranuk: I tidied up the code and removed the duplication with a call to _validate (currently required in the editor thumbnail renderer).

src/resources/material.js
raytranuk: done

src/resources/material.js
raytranuk: done

slimbuck (approved):

PR #3078, opened 2021-04-11
Added typing information for `pc.Asset`'s `type` property

TheJonRobinson:
Enhancement: Add typing info for the `type` of `pc.Asset`s

In addition to the documented types, 'template' is now a valid string too from what I see when looking at a template asset's `type` value.

---

Update - possibly this is a bit more nuanced than I first thought; there were a few types (sprite, template, container) that were not documented in the text. Poking a bit into the editor revealed a longer list than I realised:
```
0: "animation"
1: "animstategraph"
2: "audio"
3: "bundle"
4: "binary"
5: "container"
6: "cubemap"
7: "css"
8: "font"
9: "folder"
10: "json"
11: "html"
12: "material"
13: "model"
14: "render"
15: "scene"
16: "script"
17: "shader"
18: "sprite"
19: "template"
20: "text"
21: "texture"
22: "textureAtlas"
23: "wasm"
```
I'm not sure which of these Playcanvas would deem internal/undocumented though. So this PR contains ones we've been using in code when the type was just a `string`. Perhaps you would prefer to have this as a plain string, though it is useful for type-safety to know we're using the right values...?

---

### DIFF
--- a/src/asset/asset.js
+++ b/src/asset/asset.js
@@ -37,7 +37,7 @@ const VARIANT_DEFAULT_PRIORITY = ['pvr', 'dxt', 'etc2', 'etc1', 'basis'];
  * See the {@link AssetRegistry} for details on loading resources from assets.
  * @description Create a new Asset record. Generally, Assets are created in the loading process and you won't need to create them by hand.
  * @param {string} name - A non-unique but human-readable name which can be later used to retrieve the asset.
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

willeastcott:
Thanks!

TheJonRobinson:
No worries @willeastcott - just wanted to highlight that the list is probably incomplete from the 20+ that seem to be present in the editor, but wasn't sure if there was a more formal subset? Anyway, this gives us the types we're using today 👍

PR #3094, opened 2021-04-15
Support glTF morph target animations which control multiple mesh instances

### LINKED ISSUE #3085: Morph targets not updating multiple MeshInstances on the same node
[blendshapes_trial.zip](https://github.com/playcanvas/engine/files/6298409/blendshapes_trial.1.zip)

This glb contains data for two nodes, each with two MeshInstances that have morph targets animating them. However, only one MeshInstance per node actually animates.

ellthompson:
There are cases when a glTF animation contains individual animation channels which should control multiple mesh primitives. Therefore when binding our morph target animations to a node, we should bind all mesh instances on the given node which have the same name as our animated node.

Fixes #3085 

### DIFF
--- a/src/anim/binder/default-anim-binder.js
+++ b/src/anim/binder/default-anim-binder.js
@@ -136,7 +136,6 @@ class DefaultAnimBinder {
         }
         if (!node) {
             node = this.nodes[path.entityPath[path.entityPath.length - 1] || ""];
-            console.log(path, node);
 
             // #if _DEBUG
             var fallbackGraphPath = AnimBinder.encode(path.entityPath[path.entityPath.length - 1] || "", 'graph', path.propertyPath);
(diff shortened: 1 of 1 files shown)


slimbuck (approved):

willeastcott (approved):

src/anim/binder/default-anim-binder.js
```diff
@@ -82,20 +82,21 @@ class DefaultAnimBinder {
             'weights': function (node) {
                 var meshInstances = findMeshInstances(node);
                 if (meshInstances) {
-                    var morphInstance;
+                    var morphInstances = [];
                     for (var i = 0; i < meshInstances.length; ++i) {
                         if (meshInstances[i].node.name === node.name && meshInstances[i].morphInstance) {
```
mvaligursky, line 87: is this part needed, considering we find mesh instances from that node?
`meshInstances[i].node.name === node.name`

mvaligursky: maybe that's for the model component actually.

mvaligursky (approved):

src/anim/binder/default-anim-binder.js
ellthompson: Yep this is used for the model component.

PR #3097, opened 2021-04-16
Implementation of set.equals

mvaligursky:
extracted from https://github.com/playcanvas/engine/pull/2990


willeastcott:
I can't find this function in the JavaScript standard library:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

If `Set#isEqual` is not in the JS language spec, then this is not a polyfill.

mvaligursky:
it's a polyfill from the future when JS implements this function ;)

do we have some better place to add these? Perhaps we could create a new folder for these. Or just add it to "/core"?

willeastcott:
My feeling is that if we are not polyfilling a legitimately existing JS language feature/function, we shouldn't screw with standard library classes, poking in our own stuff. You _could_ stick a utility function in `core\set-utils.js` or something? But yeah, operate _on_ set instances rather than inject stuff into `Set`.

src/index.js
```diff
@@ -17,6 +17,7 @@ export { events } from './core/events.js';
 export { guid } from './core/guid.js';
 export { path } from './core/path.js';
 export { platform } from './core/platform.js';
+export { set } from './core/set-utils.js';
```
willeastcott, line 20: This shouldn't be exported to the pc namespace - it's internal.

tests/core/test_set_utils.js
```diff
@@ -0,0 +1,13 @@
...
+        var set2 = new Set([2, 1, 3]);
+        var set3 = new Set([2, 1, 6]);
+        var set4 = new Set([1, 2]);
+
+        expect(pc.set.isEqual(set1, set2)).to.equal(true);
+        expect(pc.set.isEqual(set1, set3)).to.equal(false);
+        expect(pc.set.isEqual(set1, set4)).to.equal(false);
+    });
+
+});
```
willeastcott, line 13: If this is internal, I don't think you can unit test it. Although you can unit test any public function that uses it of course. Might have to delete this.

Maksims:
Feels like it could be better named, consistent with Vec3 for example: `equals`.

willeastcott (approved):

PR #3099, opened 2021-04-19
Removed empty constructor to remove lint warning

mvaligursky:


willeastcott:
So, to be honest, we should probably just delete this whole class. As far as I can tell, no component data classes subclass from this. And it really has no value. It seems 100% legacy to me.

PR #3103, opened 2021-04-20
Array.fill and TypedArray.fill polyfills

mvaligursky:
We got burnt on this being missing few times, and it likely gives us enough performance benefit to have the polyfill vs using the loop to work around it.

### DIFF
--- a/src/polyfill/array-fill.js
+++ b/src/polyfill/array-fill.js
@@ -1,0 +1,72 @@
+if (!Array.prototype.fill) {
+    Object.defineProperty(Array.prototype, 'fill', {
+        value: function(value) {
+  
+            // Steps 1-2.
+            if (this == null) {
+                throw new TypeError('this is null or not defined');
+            }
+    
+            var O = Object(this);
+    
+            // Steps 3-5.
+            var len = O.length >>> 0;
+    
+            // Steps 6-7.
+            var start = arguments[1];
+            var relativeStart = start >> 0;
+    
+            // Step 8.
+            var k = relativeStart < 0 ? Math.max(len + relativeStart, 0) : Math.min(relativeStart, len);
+    
+            // Steps 9-10.
+            var end = arguments[2];
+            var relativeEnd = end === undefined ? len : end >> 0;
+    
+            // Step 11.
+            var finalValue = relativeEnd < 0 ? Math.max(len + relativeEnd, 0) : Math.min(relativeEnd, len);
+    
+            // Step 12.
+            while (k < finalValue) {
+                O[k] = value;
+                k++;
+            }
+    
+            // Step 13.
+            return O;
+        }
+    });
+}
+
+// implementation for typed arrays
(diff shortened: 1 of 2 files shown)


src/polyfill/array-fill.js
```diff
@@ -0,0 +1,72 @@
...
+    
+            // Step 13.
+            return O;
+        }
+    });
+}
+
+// implementation for typed arrays
+if (!Int8Array.prototype.fill) {
+    Int8Array.prototype.fill = Array.prototype.fill;
```
slimbuck, line 43: does this work correctly if `Array.fill` implementation was provided (not polyfilled)? might be worth splitting out the function and using it for all polyfills (including Array).

src/polyfill/array-fill.js
```diff
@@ -0,0 +1,72 @@
+if (!Array.prototype.fill) {
```
willeastcott, line 1:
```suggestion
// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill#polyfill
if (!Array.prototype.fill) {
```

willeastcott: (I like to link back to where polyfills are sourced generally...)

willeastcott (approved):

src/polyfill/array-fill.js
willeastcott: Good point.

mvaligursky: I would not expect a problem with this? But we only have IE to test on and that is missing both. I tested on IE and this works, and is suggested polyfill (see links I added)

src/polyfill/array-fill.js
mvaligursky: done

src/polyfill/array-fill.js
slimbuck: I understand your test worked.

However this code suggests it's possible that `Array.fill` is provided by the browser while `Int8Array.fill` isn't. Under such a circumstance this code very well may fail.

It is simple enough to make this code not fail and I don't really understand why you don't do it?

Maksims:
I do understand this polyfill comes from MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill

But it definitely is a bit too verbose, with anti-patterns used:
1. Using of `arguments` leads to function deoptimisations on V8 and other engines. It is best to pre-define arguments in a function, and use them.
2. `Object(o)` - leads to allocations?

mvaligursky:
@Maksims that's true, but this is just a polyfill for IE, not a generic code executing on all browsers.

src/polyfill/array-fill.js
willeastcott: Also, currently a bit misleading because I would expect `array-fill.js` to just polyfill `Array#fill`. I'd put the rest in `typedarray-fill.js`.

Issue #3093, opened 2021-04-14
pc.Picker doesn't work with post effects


yaustar:
Thread: https://forum.playcanvas.com/t/pc-picker-not-working-with-post-effects/8761/3?u=yaustar

https://playcanvas.com/editor/scene/1131354

Steps to see the problem

1 Run the app
2 Click on the nearest plane - it should disappear
3 Stop the app
4 Select the Camera
5 Turn on the “scanlines” script
6 Run the app
7 Click on the nearest plane - the plane behind disappears.

PR #3116, opened 2021-04-23
Picker refactoring

mvaligursky:
Refactored picker class, with these larger changes:
- ignores post effects, Fixes https://github.com/playcanvas/engine/issues/3093
- engine example - uses bloom and skydome for additional visual quality and functional testing
- refactored to reflect the fact the camera stores render target instead of layer now
 
**API change**:
- **Picker.prepare** 
    - does not support passing render target as a way to filter layers
    - allows passing in an array of layers (and also handles a single layer as before)

Screen Shot 2021-04-23 at 10 48 46 AM

### DIFF
--- a/examples/graphics/area-picker.html
+++ b/examples/graphics/area-picker.html
@@ -5,7 +5,8 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0">
     <link rel="icon" type="image/png" href="../playcanvas-favicon.png" />
-    <script src="../../build/playcanvas.js"></script>
+    <script src="../../build/playcanvas.dbg.js"></script>
+    <script src="../assets/scripts/asset-loader.js"></script>
     <script src="../../build/playcanvas-extras.js"></script>
     <style>
         body { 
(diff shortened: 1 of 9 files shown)


examples/graphics/area-picker.html
```diff
@@ -5,7 +5,8 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0">
     <link rel="icon" type="image/png" href="../playcanvas-favicon.png" />
-    <script src="../../build/playcanvas.js"></script>
+    <script src="../../build/playcanvas.dbg.js"></script>
```
willeastcott, line 8: Revert this line.

mvaligursky: done :)

willeastcott (approved):

src/scene/picker.js
```diff
@@ -163,14 +232,11 @@ class Picker {
      * in any way, {@link Picker#prepare} does not need to be called again.
      * @param {CameraComponent} camera - The camera component used to render the scene.
      * @param {Scene} scene - The scene containing the pickable mesh instances.
-     * @param {Layer|RenderTarget} [arg] - Layer or RenderTarget from which objects will be picked.
-     * If not supplied, all layers rendering to backbuffer before this layer will be used.
+     * @param {Layer[]} [layers] - Layers from which objects will be picked. If not supplied, all layers the provided camera renders will be used.
```
willeastcott, line 235:
```suggestion
    * @param {Layer[]} [layers] - Layers from which objects will be picked. If not supplied, all layers of the specified camera will be used.
```

PR #3117, opened 2021-04-23
Modify src/scene/scene.js to verify if `this.defaultMaterial` is valid before destroying

kprimo:
modify src/scene/scene.js to verify if `this.defaultMaterial` is valid before destroying (to solve the problem that deleting the last app in multiple apps will cause an exception)

Fixes #

### DIFF
--- a/src/scene/scene.js
+++ b/src/scene/scene.js
@@ -152,8 +152,10 @@ Scene.prototype.constructor = Scene;
 
 Scene.prototype.destroy = function () {
     this.root = null;
-    this.defaultMaterial.destroy();
-    this.defaultMaterial = null;
+    if (this.defaultMaterial){
+        this.defaultMaterial.destroy();
+        this.defaultMaterial = null;
+    }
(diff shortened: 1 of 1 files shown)


yaustar:
The only place that `this.defaultMaterial` is set to null is in that destroy function which implies that Scene.destroy is being called twice. 🤔

mvaligursky:
I agree with @yaustar - the null check should not be needed here, as it simply indicates the issue is elsewhere.

mvaligursky:
perhaps it'd be good to put some marker on that default material when it gets constructed, for example
```
this.defaultMaterial.hackId = Math.random();
console.log("created: " + this.defaultMaterial.hackId);
`;

and log it out as well from material's destroy function when it gets destroyed.
This could show us where it's getting destroyed from multiple times.

mvaligursky:
This is no longer applicable due to refactoring - the scene no longer owns nor destroys the defaultMaterial

PR #3118, opened 2021-04-23
Implementation of utility function instance.of

mvaligursky:
A replacement of 'instanceof' functionality which does not require the type of the object to be imported, avoiding (circular) dependencies.

Note: this is not mean to be a generic replacement for `instanceof`, even though it can be completely used as a replacement. The main reason is to allow some backwards compatibility code which imports modules otherwise not needed to be fixed, in order to remove circular dependency.

Instead of currently often used code:

```
import { GraphicsDevice } from './graphics-device.js';
if (options instanceof GraphicsDevice) {
}
```
this can be used to remove import of the GraphicsDevice.
```
import { instance } from '../core/instance-utils.js';
if (instance.of(options, "GraphicsDevice")) {
}
```

Other changes:
- applied this functionality to Render Target
- replaced some calls to constructor.name by calls to instance.name which handles IE11 fallback.

Test code (passes on IE11 as well)
```
        var dm = new pc.DepthMaterial();
        console.log("DepthMat: " + pc.instance.of(dm, "DepthMaterial"));
        console.log("Mat: " + pc.instance.of(dm, "Material"));

        console.log("app: " + pc.instance.of(app, "Application"));

        var x = 5;
        console.log("x: " + pc.instance.of(x, "Application"));
        console.log("null: " + pc.instance.of(null, "Application"));
        console.log("undefined: " + pc.instance.of(undefined, "Application"));

```

output:
Screen Shot 2021-04-23 at 12 50 55 PM

When this is approved, multiple dependencies will be removed in following PR.

### DIFF
--- a/src/core/instance-utils.js
+++ b/src/core/instance-utils.js
@@ -1,0 +1,43 @@
+const regex = new RegExp(/^\s*function\s*(\S*)\s*\(/);
+
+const instance = {
+
+    // helper function to get a class name of the object
+    name: function (object) {
+
+        // modern browsers have name property
+        if (object.constructor.name) {
+            return object.constructor.name;
+        }
+
+        // IE11 fallback
+        return object.constructor.toString().match(regex)[1];
+    },
+
+    // helper function to find out if the specified object is an instance of the specified class name
+    // example use: const isInstance = instance.of(app, "Application");
+    of: function (object, className) {
+
+        // make sure object is an object and not null of a number of similar
+        if (typeof object === 'object' && object !== null) {
+
+            let proto = Object.getPrototypeOf(object);
+            for (;;) {
+                if (proto === null) {
+                    return false;
+                }
+
+                if (this.name(proto) === className) {
+                    return true;
+                }
+
+                // trace up the chain
+                proto = Object.getPrototypeOf(proto);
+            }
+        }
+
+        return false;
+    }
+};
+
+export { instance };
--- a/src/framework/components/anim/component.js
+++ b/src/framework/components/anim/component.js
@@ -1,4 +1,5 @@
 import { Asset } from '../../../asset/asset.js';
+import { instance } from '../../../core/instance-utils.js';
 
 import { AnimEvaluator } from '../../../anim/evaluator/anim-evaluator.js';
 import { AnimController } from '../../../anim/controller/anim-controller.js';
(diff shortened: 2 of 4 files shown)


willeastcott:
I'm not a big fan of this. The problem here is that _for the purposes of Rollup_, you are removing the cyclic dependencies. But architecturally, the cycles still exist in the code. It's just that the lack of explicit typing in JavaScript allows you to avoid the import. If we ever wanted to migrate the Typescript (or add JSDoc everywhere to get better type-checking in VS Code), the cycles would be back (for the most part).

I would kind of prefer us to think about more elegant ways to solve this problem. I mentioned yesterday a solution where objects like `VertexBuffer`, `Texture`, `Light`, `Camera`, etc just hold data and properties, and it is the `ForwardRenderer` that leverages the `GraphicsDevice` to prepare and dispatch the data.

mvaligursky:
I don't believe with this solution the dependency stays in the code, and this removes it correctly.

In JS the dependency on a module is created when you construct an instance of the class from it, for example "new GraphicsDevice()", or when you need class type for something like "object instanceof GraphicsDevice". At the point js parser reads this code, it needs to have processed GraphicsDevice to be able to understand this.

All import dependencies do in JS world is that they allow the bundler to order the modules (or loaded to load them in order) to make sure the types are known when needed. Lack of import of some type does not imply you cannot use the instance of the class in the code.

Maksims:
The use of string instead of an actual prototype does feel hacky, also, it does not address prototype inheritance.

mvaligursky:
It does handle inheritance @Maksims by walking up the chain, the same way `instanceof` does, and also the tests shown in the description confirm it.

It is hacky, but it solve the hacky code base. Let me explain on the render-target.js file from this PR.
The original code would take the GraphicsDevice as an argument to the constructor. At some point this was considered incorrect, and so was removed. The handle backwards compatibility, the code checks `if (options instanceof GraphicsDevice) {` to handle it. In ideal world we don't check backwards compatibility and so won't have to check it, removing the dependency that way.  Otherwise if we have to handle backwards compatibility in these cases, we can never fix circular dependencies.

Maksims:
If RenderTarget constructor is the only reason for such `instanceof`, then there is better approach to it, also avoiding use of `arguments`:
```js
if (arguments[1] || arguments[2]) {
    // old constructor
    this._colorBuffer = arguments[1];
    options = arguments[2];
} else {
    // new constructor
    this._colorBuffer = options.colorBuffer;
}
```

willeastcott:
I suggest closing this @mvaligursky. Do you agree?

PR #2166, opened 2020-06-16
Port PlayCanvas to ES6 modules and rollup.js

willeastcott:
This PR ports the PlayCanvas codebase to ES6 module format and adopts [rollup.js](https://rollupjs.org/) to build the engine.

This brings the following benefits;
* We can easily see and eliminate circular dependencies in the codebase.
* Developers can leverage tree-shaking, and build versions of the engine without certain modules.
* We can trivially build the engine as an ES6 module.
* We can easily leverage any of the many rollup plugins available.
* The `pc` namespace has been dropped from the internal engine code.
* Removes around ~1000 lines of build configuration related scripts/configs. Rollup config is ~115 lines.

Other miscellaneous changes:
* Engine now builds to `build`, not `build/output`
* The `Object.assign(pc, {});` pattern has been eliminated.
* Function expressions have been eliminated. So we do `function MyClass() {}` instead of `var MyClass = function () {};` now.
* `pc.MiniStats` has been moved to a new `pcx` namespace (so `new pcx.MiniStats`)

Stats:
Minified GZIPd Engine 1.28.3 (Release): 266KB
Minified GZIPd Engine 1.29.0-dev (Release): 262KB

Task List:
- [x] Build extras library
- [x] Add the copyright notice
- [x] Process the GLSL files and add them to the library
- [x] Repair engine to previous state
- [x] Add deprecated API back
- [x] Build debug/release/profile versions

Note that there are alternative PRs that implement building with:

* Parcel https://github.com/playcanvas/engine/pull/2106 (by @kungfooman)
* ESBuild https://github.com/playcanvas/engine/pull/2115 (by @hxhieu)

This PR has been chosen over these alternatives.

Fixes #2103
Closes #2106 
Closes #2115

I confirm I have signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #3122, opened 2021-04-26
[FIX] Restore properties to support profiler

willeastcott:
Restores external access to properties that the launch page's Profiler relies upon to limit draw call rendering. This bug was introduced back in June 2020 by [this PR](https://github.com/playcanvas/engine/pull/2166).

Partial fix for https://github.com/playcanvas/editor/issues/103

### DIFF
--- a/src/scene/forward-renderer.js
+++ b/src/scene/forward-renderer.js
@@ -513,6 +513,7 @@ class ForwardRenderer {
         this._createAreaLightPlaceholderLuts();
     }
 
+    // Static properties used by the Profiler in the Editor's Launch Page
     static skipRenderCamera = null;
 
     static _skipRenderCounter = 0;
--- a/src/scene/forward-renderer.js
+++ b/src/scene/forward-renderer.js
@@ -114,9 +114,6 @@ const maxBlurSize = 25;
 var keyA, keyB;
 
 var _autoInstanceBuffer = null;
-var skipRenderCamera = null;
-var _skipRenderCounter = 0;
-var skipRenderAfter = 0;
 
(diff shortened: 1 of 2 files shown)


src/scene/forward-renderer.js
```diff
@@ -516,6 +513,13 @@ class ForwardRenderer {
         this._createAreaLightPlaceholderLuts();
     }
 
+    // Static properties used by the Profiler in the Editor's Launch Page
+    static skipRenderCamera = null;
```
slimbuck, line 517: should these defines also be placed in `#if PROFILER` ?

mvaligursky: it's probably simpler to leave them out so the editor can just access them with any engine build.

mvaligursky (approved):

src/scene/forward-renderer.js
willeastcott: Done

willeastcott: Oh. Just saw @mvaligursky disagreement with your suggestion @slimbuck. I went ahead and made them PROFILER build only. TBH, I don't think it's a big deal. This implementation is actually pretty nasty for a number of reasons - rather than refactor, I wanted to do the minimum to repair the original functionality. But in general:

1. There shouldn't be Editor-specific functionality in the engine. I'd rather define a general-purpose public API (with better naming!) for limiting draws calls, or early-outing from rendering a layer composition.
2. These properties are static and would break if two `Application` instances are active on a page.
3. There seems to be layer-specific draw call limiting code in there too and I'm not sure if it's currently being used. Needs investigation.

PR #3124, opened 2021-04-26
Remove gl canvas dependency resize

### LINKED ISSUE #3114: Refactor gl.canvas to use clientWidth/Height instead of relying on external calls to resizeCanvas
From WebGL anti patterns entry 4: https://webglfundamentals.org/webgl/lessons/webgl-anti-patterns.html

Example code:
```
function resizeCanvasToDisplaySize() {
  var width = gl.canvas.clientWidth;
  var height = gl.canvas.clientHeight;
  if (gl.canvas.width != width ||
      gl.canvas.height != height) {
     gl.canvas.width = width;
     gl.canvas.height = height;
  }
}

function render() {
   resizeCanvasToDisplaySize();
   drawStuff();
   requestAnimationFrame(render);
}
render();
```

This will also remove some of the reliance on DOM calls in the engine that we shouldn't really be using.

Also, this should help fix some of the rendering stretching bugs we have on the Samsung browser and iOS browser.

yaustar:
Fixes #3114 

First step in making the engine more flexible to use in different page layouts where the canvas DOM changes size. It also means that the canvas rendering resolution is no longer dependent on calling resizeCanvas manually from window resize events.

There was a comment in the ticket about doing the check every 200ms instead of every frame. I've measured performance on an older Samsung S7 and compared stable to my branch and the difference doesn't show up on the profiler.

Ideally, I would like to keep it every frame as it looks lot better when the Canvas DOM is being resized and doesn't suffer from a stutter delay due to only being updated every Xms

For comparison, Babylon JS does this too every frame: https://github.com/BabylonJS/Babylon.js/blob/5e6321d887637877d8b28b417410abbbeb651c6e/src/Engines/Extensions/engine.views.ts#L145

### DIFF
--- a/src/framework/application.js
+++ b/src/framework/application.js
@@ -2317,12 +2317,12 @@ var makeTick = function (_app) {
             if (application._resolutionMode === RESOLUTION_AUTO) {
                 // Check if the canvas DOM has changed size
                 const canvas = application.graphicsDevice.canvas;
-                const pixelRatio = application.graphicsDevice.maxPixelRatio;
+                const pixelRatio = Math.min(application.graphicsDevice.maxPixelRatio, window.devicePixelRatio);
 
                 const clientWidth = canvas.clientWidth;
                 const clientHeight = canvas.clientHeight;
 
-                if (canvas.width !== (clientWidth * pixelRatio) || canvas.height !== (clientHeight * pixelRatio)) {
+                if (canvas.width !== Math.floor(clientWidth * pixelRatio) || canvas.height !== Math.floor(clientHeight * pixelRatio)) {
                     application.graphicsDevice.resizeCanvas(clientWidth, clientHeight);
                 }
             }
--- a/src/framework/application.js
+++ b/src/framework/application.js
@@ -421,8 +421,6 @@ class Application extends EventHandler {
         this.autoRender = true;
         this.renderNextFrame = false;
 
-        this.timeBetweenCanvasSizeCheck = 0.2;
-
         // enable if you want entity type script attributes
         // to not be re-mapped when an entity is cloned
         this.useLegacyScriptAttributeCloning = script.legacy;
@@ -2260,7 +2258,6 @@ var _frameEndData = {};
 var makeTick = function (_app) {
     var application = _app;
     var frameRequest;
-    var timeToNextCanvasSizeCheck = 0;
 
     return function (timestamp, frame) {
         if (!application.graphicsDevice)
@@ -2278,11 +2275,9 @@ var makeTick = function (_app) {
 
         var currentTime = application._processTimestamp(timestamp) || now();
         var ms = currentTime - (application._time || currentTime);
-        var realDt = ms / 1000.0;
-
-        var scaledDt = realDt;
-        scaledDt = math.clamp(scaledDt, 0, application.maxDeltaTime);
-        scaledDt *= application.timeScale;
+        var dt = ms / 1000.0;
+        dt = math.clamp(dt, 0, application.maxDeltaTime);
+        dt *= application.timeScale;
 
         application._time = currentTime;
 
@@ -2298,7 +2293,7 @@ var makeTick = function (_app) {
         if (application.graphicsDevice.contextLost)
             return;
 
-        application._fillFrameStatsBasic(currentTime, scaledDt, ms);
+        application._fillFrameStatsBasic(currentTime, dt, ms);
 
         // #if _PROFILER
         application._fillFrameStats();
@@ -2313,30 +2308,23 @@ var makeTick = function (_app) {
             application.graphicsDevice.defaultFramebuffer = null;
         }
 
-        application.update(scaledDt);
+        application.update(dt);
 
         application.fire("framerender");
 
         if (application.autoRender || application.renderNextFrame) {
             // In AUTO mode the resolution is changed to match the canvas size
             if (application._resolutionMode === RESOLUTION_AUTO) {
-                timeToNextCanvasSizeCheck -= realDt;
-                if (timeToNextCanvasSizeCheck <= 0) {
-                    // Check if the canvas DOM has changed size
-                    const canvas = application.graphicsDevice.canvas;
-                    const pixelRatio = application.graphicsDevice.maxPixelRatio;
+                // Check if the canvas DOM has changed size
+                const canvas = application.graphicsDevice.canvas;
+                const pixelRatio = application.graphicsDevice.maxPixelRatio;
 
-                    const clientWidth = canvas.clientWidth;
-                    const clientHeight = canvas.clientHeight;
+                const clientWidth = canvas.clientWidth;
+                const clientHeight = canvas.clientHeight;
 
-                    if (canvas.width !== (clientWidth * pixelRatio) || canvas.height !== (clientHeight * pixelRatio)) {
-                        application.graphicsDevice.resizeCanvas(clientWidth, clientHeight);
-                    }
-
-                    timeToNextCanvasSizeCheck = application.timeBetweenCanvasSizeCheck;
+                if (canvas.width !== (clientWidth * pixelRatio) || canvas.height !== (clientHeight * pixelRatio)) {
+                    application.graphicsDevice.resizeCanvas(clientWidth, clientHeight);
                 }
-            } else {
-                timeToNextCanvasSizeCheck = 0;
             }
 
             application.render();
--- a/src/framework/application.js
+++ b/src/framework/application.js
@@ -2317,8 +2317,13 @@ var makeTick = function (_app) {
             if (application._resolutionMode === RESOLUTION_AUTO) {
                 // Check if the canvas DOM has changed size
                 const canvas = application.graphicsDevice.canvas;
-                if (canvas.width !== canvas.clientWidth || canvas.height !== canvas.clientHeight) {
-                    application.graphicsDevice.resizeCanvas(canvas.clientWidth, canvas.clientHeight);
+                const pixelRatio = application.graphicsDevice.maxPixelRatio;
+
+                const clientWidth = canvas.clientWidth;
+                const clientHeight = canvas.clientHeight;
+
+                if (canvas.width !== (clientWidth * pixelRatio) || canvas.height !== (clientHeight * pixelRatio)) {
+                    application.graphicsDevice.resizeCanvas(clientWidth, clientHeight);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


mvaligursky:
This looks good to me. Are there any implications for the user with this change, or are no changes to user code nor their project behavior expected?

src/framework/application.js
```diff
@@ -2318,6 +2313,15 @@ var makeTick = function (_app) {
         application.fire("framerender");
 
         if (application.autoRender || application.renderNextFrame) {
+            // In AUTO mode the resolution is changed to match the canvas size
+            if (application._resolutionMode === RESOLUTION_AUTO) {
+                // Check if the canvas DOM has changed size
+                const canvas = application.graphicsDevice.canvas;
+                if (canvas.width !== canvas.clientWidth || canvas.height !== canvas.clientHeight) {
```
willeastcott, line 2320: So you're comparing canvas resolution (which accounts for devicePixelRatio) against the client width and height (which doesn't). Won't this be different every frame if devicePixelRatio is not 1?

yaustar: Good question, I will take a look

yaustar:
> This looks good to me. Are there any implications for the user with this change, or are no changes to user code nor their project behavior expected?

There shouldn't be a need for users to change their code. What happens prior to this change was that the resizeCanvas on graphics device is indirectly called from application.resizeCanvas that is listening to the window resize event. eg: https://github.com/playcanvas/playcanvas.github.io/blob/master/examples/animation/blend.html#L33

This meant that the canvas width/height was only changed when the window is a different size, not the canvas DOM

Maksims:
On the empty project, doing a check on every frame will not have implications. But projects with medium/large DOM usage will have more significant penalties, as calling a check can lead to reflow and style recalculation, which can lead to a longer layout.

Changing resolution is a rare case, it is not breaking UX to have a small delay. But checking every frame can have a negative impact.

yaustar:
> But projects with medium/large DOM usage will have more significant penalties, as calling a check can lead to reflow and style recalculation, which can lead to a longer layout.

Do we have a project or example where I can benchmark/test this against? I could in theory expose the property so that it defaults to one way or another and be changed by the developer if need be?

mvaligursky:
Would just reading a property on the canvas have such an impact? In the worst case, would it not just execute the same code it would have to anyways in the same frame?

src/framework/application.js
yaustar: Good spot, fixed in a later commit

yaustar:
Added the check to be every 200ms which I'd rather not have but at least it's exposed (as a private property) if a developer wants it more/less often for a specific reason.

Video of it being at different times (200ms, then 0, then 100)

https://user-images.githubusercontent.com/16639049/116104620-80c45880-a6a8-11eb-8d6b-e330397f1647.mp4

willeastcott:
@Maksims The engine can already cause a reflow every single frame. See:

https://github.com/playcanvas/engine/blob/master/src/framework/application.js#L1308

Which is defined here:

https://github.com/playcanvas/engine/blob/master/src/graphics/graphics-device.js#L1169-L1171

Maksims:
> Added the check to be every 200ms which I'd rather not have but at least it's exposed (as a private property) if a developer wants it more/less often for a specific reason.
> 
> Video of it being at different times (200ms, then 0, then 100)
> 
>  Kapture.2021-04-26.at.15.55.41.mp4

You can call reflow on the event in addition to the timer, which eliminates all common cases where resolution changes are reported correctly after the window `resize` event.

But, as @willeastcott mentioned - reflow is already happening, so this seems like does not matter anymore. I was assuming there was no force reflows done by the engine before.

malDuffin:
Will this solution solve this problem?

https://forum.playcanvas.com/t/different-speeds-and-crash-on-ios-with-new-model-viewer-starter-kit-scene/8978

yaustar:
As we are doing the reflow potentially every frame, I will revert the timer commit.

@malDuffin The stretched canvas issue? I will try to reproduce this locally again, I'm not sure if that's still an issue with the later versions of iOS.

yaustar:
@malDuffin I can't reproduce the iOS canvas stretching issue on either my iPhone 5S or iPhone XR. Do you still encounter this issue?

yaustar:
Removed the canvas size check on a timer

src/framework/application.js
```diff
@@ -2318,6 +2313,20 @@ var makeTick = function (_app) {
...
+            // In AUTO mode the resolution is changed to match the canvas size
+            if (application._resolutionMode === RESOLUTION_AUTO) {
+                // Check if the canvas DOM has changed size
+                const canvas = application.graphicsDevice.canvas;
+                const pixelRatio = application.graphicsDevice.maxPixelRatio;
+
+                const clientWidth = canvas.clientWidth;
+                const clientHeight = canvas.clientHeight;
+
+                if (canvas.width !== (clientWidth * pixelRatio) || canvas.height !== (clientHeight * pixelRatio)) {
```
mvaligursky, line 2325: I wonder if some rounding should be used here? I assume the pixelRatio is float and so when you multiply clientWidth by it, you need to round it to whole number? Strange that even resizeCanvas function does not do that. can canvas.width be set to fractional number?

yaustar: Code from WebGL fundamentals have the following:
```
  /**
   * Resize a canvas to match the size its displayed.
   * @param {HTMLCanvasElement} canvas The canvas to resize.
   * @param {number} [multiplier] amount to multiply by.
   *    Pass in window.devicePixelRatio for native pixels.
   * @return {boolean} true if the canvas was resized.
   * @memberOf module:webgl-utils
   */
  function resizeCanvasToDisplaySize(canvas, multiplier) {
    multiplier = multiplier || 1;
    const width  = canvas.clientWidth  * multiplier | 0;
    const height = canvas.clientHeight * multiplier | 0;
    if (canvas.width !== width ||  canvas.height !== height) {
      canvas.width  = width;
      canvas.height = height;
      return true;
    }
    return false;
  }
```

I can't find anything yet about whether width/height can be non-integers

yaustar: Interesting that:
```
canvas.clientWidth  * multiplier | 0
```
The `| 0` rounds to a whole number

yaustar: I will need to do more investigation as there looks like there can be edge cases to deal with. This will not go out with 1.41.0 and moving to draft while I figure out what is the best route here 😅

Maksims: clientWidth and clientHeight should return int, but getBoundingClientRect - returns object with floats. And this should be taken into an account in such context.

In this case, you definitely want to `floor` this: `clientWidth * pixelRatio`, and then compare to `canvas.width`, as setting float to `canvas.width`, it will `floor` it for you.

malDuffin:
It's great that the issue I mentioned earlier is resolved.

As this is to do with someone adjusting the width/height of the canvas, this could be another fix that is related. It keeps the main object in focus, based on a specific aspect ratio.  I used it for a game I was working on, it was essential for maximising the space available for content - especially on a mobile device.

    var requiredAspectRatio = 0.76;
    
    pc.Mat4.prototype.setPerspective = function(fov, aspect, znear, zfar, fovIsHorizontal) 
    {
        var xmax, ymax;
        if (!fovIsHorizontal) 
        {
            ymax = znear * Math.tan(fov * Math.PI / 360);
            xmax = ymax * aspect;
        } 
        else
        {                
            xmax = znear * Math.tan(fov * Math.PI / 360);
            xmax *= requiredAspectRatio;
            ymax = xmax / aspect;
        }

        return this.setFrustum(-xmax, xmax, -ymax, ymax, znear, zfar);
    };

Maksims:
Is it the same as switching between horizontal and vertical FoV based on `aspectRation > 1` ?

malDuffin:
> Is it the same as switching between horizontal and vertical FoV based on `aspectRation > 1` ?
> This is a common case indeed, but definitely would be affecting existing applications/projects, such breaking changes are best avoided at all costs.

Agreed - with "requiredAspectRatio = 1", nothing would change ( so no breaking changes ).  

If someone did want this functionality, they change the requiredAspectRatio value to suit their app.  The actual naming of the variables etc may not be correct.

Like I said, the code above is something that worked great for me - having that functionality exposed would a useful feature for many other devs making games where their content isn't square, and they want to maximise their content on-screen with various different screen sizes and orientations.

yaustar:
I'm thinking it would be possible to do as script (maybe as part of a tutorial project?) rather than something that is built directly into the engine.

src/framework/application.js
yaustar: I've explicitly floored both the check and the assignment on the canvas (just in case some browsers round or something). :)

mvaligursky (approved):
seems ok to me now.

src/framework/application.js
```diff
@@ -2318,6 +2313,20 @@ var makeTick = function (_app) {
...
+            // In AUTO mode the resolution is changed to match the canvas size
+            if (application._resolutionMode === RESOLUTION_AUTO) {
+                // Check if the canvas DOM has changed size
+                const canvas = application.graphicsDevice.canvas;
+                const pixelRatio = Math.min(application.graphicsDevice.maxPixelRatio, window.devicePixelRatio);
+
+                const clientWidth = canvas.clientWidth;
+                const clientHeight = canvas.clientHeight;
+
+                if (canvas.width !== Math.floor(clientWidth * pixelRatio) || canvas.height !== Math.floor(clientHeight * pixelRatio)) {
```
slimbuck, line 2325: shouldn't this test live in resizeCanvas instead of here? Calculating pixelRatio and doing all the floor stuff twice ain't great.

yaustar: I didn't want add this logic in resizeCanvas as it's existing API just in case some users were using the function directly.

I could add a new function that would be along the lines of 'checkAndResizeCanvas` or have a utility function in graphicDevice that has the logic for calculating the width/height?

slimbuck: well actually the function already has the test? (see https://github.com/playcanvas/engine/pull/3124/files#diff-f7773f4bc0740b6db3da2ac63c30e595a6076b3f8d6699f5c0de9c8628ae7d57R3501). I don't think you need to do any calculation or testing here.

slimbuck: So I'd suggest just making it something like:
```javascript
            // In AUTO mode the resolution is changed to match the canvas size
            if (application._resolutionMode === RESOLUTION_AUTO) {
                const canvas = application.graphicsDevice.canvas;
                application.graphicsDevice.resizeCanvas(canvas.clientWidth, canvas.clientHeight);
            }
```

yaustar: Completely makes sense now looking back on it 😅 

Made the changes.

PR #3128, opened 2021-04-27
AnimComponent events

### LINKED ISSUE #334: Animation events
Would be good to fire events when animations start, end, etc.

### LINKED ISSUE #2816: AnimComponent events
Would be great to have `anim` component fire events when animations start, end, loop, etc. as well as when current state changes or updates.

```JavaScript
entity.anim.on('loop', ...
```

https://github.com/playcanvas/engine/issues/334 v2 :)

ellthompson:
Adds support for timed events in the `pc.AnimTrack` class. These events can be listened to on the anim component the tracks are played in.

Usage example:

```javascript
const animationTrack = asset.resource;
animationTrack.events = new pc.AnimEvents([
    {
        name: 'midpoint event',
        time: idleTrack.duration / 2
    }
]);

entity.anim.assignAnimation('Initial State', animationTrack);

entity.anim.on('midpoint event', (event) => {
    console.log(event.name, event.track);
});
```

Fixes #334 
Fixes #2816 


Maksims:
Is there a way to add/remove events with already assigned animations?

ellthompson:
> Is there a way to add/remove events with already assigned animations?

The track is passed to the event so it's possible to update the tracks events array once an event has been consumed.
```javascript
entity.anim.on('midpoint event', (event) => {
    console.log(event.name, event.track);
    event.track.events = [];
});
```

Maksims:
So, in order to add new event, it would be something like this, and only in a callback?:
```js
event.track.push({
    time: 10,
    name: 'someEvent'
});
```
This seems not intuitive and "raw" tbh.

Perhaps something like, would look more like an API?:
```js
const track = entity.anim.getAnimation('Walk')
track.addEvent(5, 'step');
track.addEvent(10, 'step');
```

ellthompson:
> So, in order to add new event, it would be something like this, and only in a callback?:
> 
> ```js
> event.track.push({
>     time: 10,
>     name: 'someEvent'
> });
> ```
> 
> This seems not intuitive and "raw" tbh.
> 
> Perhaps something like, would look more like an API?:
> 
> ```js
> const track = entity.anim.getAnimation('Walk')
> track.addEvent(5, 'step');
> track.addEvent(10, 'step');
> ```

Can you give some use cases for this? I'm expecting these events to be tied to certain poses in an animation, so I had imagined they'd be set up at load time rather than during the game loop. If you have events that you may need in the future but not at the start, you can set them in the track but not listen to them on the anim component.

Maksims:
I believe there can be cases for a developer wanting to add/remove events during application life, but out of my head, it would be mostly if they want to build some internal tool for animations.

Preparing them in advice - definitely a most common use case. It would be great to have a dedicated animation section in User Manual, for describing all different workflow and user stories, that can help to understand a system and common workflows.

ellthompson:
I've added the `pc.AnimEvents` class which should provide some structure to the process of creating events and it'll show up in the API reference now.

I've also created a section for Animation in the user manual, which I'll update with a section for animation events once the editor side of this feature is in.

src/anim/evaluator/anim-events.js
```diff
@@ -0,0 +1,27 @@
...
+ *     {
+ *         name: 'my_event',
+ *         time: 1.3 // given in seconds
+ *     }
+ * ]);
+ * animTrack.events = events;
+ */
+class AnimEvents {
+    constructor(events) {
+        events.sort((a, b) => a.time - b.time);
```
mvaligursky, line 18: I wonder if you should shallow copy the array and sort internal copy, instead of user provided array

src/anim/evaluator/anim-events.js
```diff
@@ -0,0 +1,27 @@
...
+ * @class
+ * @name AnimEvents
+ * @classdesc AnimEvents stores a sorted array of animation events which should fire sequentially during the playback of an pc.AnimTrack.
+ * @description Create a new collection of animation events.
+ * - @param {object[]}  events- An array of animation events.
+ * @example
+ * const events = new pc.AnimEvents([
+ *     {
+ *         name: 'my_event',
+ *         time: 1.3 // given in seconds
```
mvaligursky, line 11: maybe add some additional properties here to demonstrate user can add whatever he needs. Only the _time_ is the required one I assume?

src/anim/evaluator/anim-track.js
```diff
@@ -39,6 +41,14 @@ class AnimTrack {
         return this._curves;
     }
 
+    get events() {
+        return this._animEvents?.events || [];
```
mvaligursky, line 45: when value is undefined or null, it's better to return that instead of creating it at that point. Alternatively, initialize it to [] in the constructor (default value for the parameter)

src/anim/evaluator/anim-clip.js
```diff
@@ -25,6 +25,12 @@ class AnimClip {
         this._loop = loop;              // whether to loop
         this._blendWeight = 1.0;        // blend weight 0..1
         this._blendOrder = 0.0;         // blend order relative to other clips
+        this._eventHandler = eventHandler;
+        this._eventCursor = 0;
+        // move the event cursor to an event that should fire after the starting time
+        while (this._track.events[this._eventCursor]?.time < this.time) {
```
mvaligursky, line 31: not too keen on using `?.` here. For out of array access this returns _undefined_ which you then compare to time. It works and returns false, but the strongly typed part of my mind does not like it :)

src/anim/evaluator/anim-clip.js
```diff
@@ -83,13 +89,52 @@ class AnimClip {
...
+
     _update(deltaTime) {
         if (this._playing) {
             var time = this._time;
             var duration = this._track.duration;
             var speed = this._speed;
             var loop = this._loop;
 
+            // check for events that should fire during this frame
+            this.activeEventsForFrame(time, time + speed * deltaTime)
```
mvaligursky, line 133: evaluate `time + speed * deltaTime` only one time

src/anim/evaluator/anim-clip.js
```diff
@@ -83,13 +89,52 @@ class AnimClip {
...
+
+    set eventCursor(value) {
+        this._eventCursor = value;
+    }
+
+    activeEventsForFrame(frameStartTime, frameEndTime) {
+        if (frameStartTime === 0) {
+            this.eventCursor = 0;
+        }
+        var events = [];
```
mvaligursky, line 104: instead of this function allocating an array each time so that the caller call fire events, it should fire them directly and avoid the allocations.

mvaligursky: alternatively, have a global array that you reuse for this reason

src/anim/evaluator/anim-clip.js
```diff
@@ -83,13 +89,52 @@ class AnimClip {
...
+    get eventCursor() {
+        return this._eventCursor;
+    }
+
+    set eventCursor(value) {
+        this._eventCursor = value;
+    }
+
+    activeEventsForFrame(frameStartTime, frameEndTime) {
+        if (frameStartTime === 0) {
```
mvaligursky, line 101: what if the animation does not start from time 0? This won't be reset to start and won't trigger events.

src/anim/evaluator/anim-clip.js
```diff
@@ -83,13 +89,52 @@ class AnimClip {
...
+            this.eventCursor = 0;
+        }
+        var events = [];
+        var clippedFrameDuration;
+        // if this frame overlaps with the end of the track, we should clip off the end of the frame time then check that clipped time later
+        if (frameEndTime > this.track.duration) {
+            clippedFrameDuration = frameEndTime - this.track.duration;
```
mvaligursky, line 108: this should handle a case when the time wraps around more than once .. which it can for short animations and long dt. I think this should have a unit test written for all these edge cases.

src/anim/controller/anim-controller.js
```diff
@@ -355,14 +356,27 @@ class AnimController {
         this._currTransitionTime = 0;
         this._transitionInterruptionSource = transition.interruptionSource;
 
-        var hasTransitionOffset = transition.transitionOffset && transition.transitionOffset > 0.0 && transition.transitionOffset < 1.0;
+
         var activeState = this.activeState;
+        var hasTransitionOffset = transition.transitionOffset && transition.transitionOffset > 0.0 && transition.transitionOffset < 1.0;
```
mvaligursky, line 361: is this change related to the PR?

src/anim/evaluator/anim-clip.js
ellthompson: The AnimClip's constructor moves the event cursor to the correct position in the cases when the time is greater than 0.

src/anim/evaluator/anim-clip.js
ellthompson: This function is called recursively until all of the current frames duration is consumed.

src/anim/evaluator/anim-clip.js
mvaligursky: so what's the purpose of this code then? I think you should just do 
this.eventCursor = 0;
at the place you call it with 0, as I assume that's the case it handles.

src/anim/evaluator/anim-clip.js
mvaligursky: ah yeah I missed that .. that seems ok

src/anim/evaluator/anim-events.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * @class
+ * @name AnimEvents
+ * @classdesc AnimEvents stores a sorted array of animation events which should fire sequentially during the playback of an pc.AnimTrack.
+ * @description Create a new collection of animation events.
+ * - @param {object[]}  events- An array of animation events.
```
mvaligursky, line 6: why two '-' on this line?

src/anim/evaluator/anim-events.js
```diff
@@ -0,0 +1,29 @@
...
+ * @example
+ * const events = new pc.AnimEvents([
+ *     {
+ *         name: 'my_event',
+ *         time: 1.3, // given in seconds
```
mvaligursky, line 11: tiny thing, but list time as first property, as that is required .. all the rest is optional I think, including the name. Make it clearer somehow maybe.

ellthompson: The name isn't optional as it's used to fire the event. I can make this clearer though.

src/anim/evaluator/anim-track.js
```diff
@@ -9,14 +11,16 @@
  * @param {AnimData[]} inputs - List of curve key data.
  * @param {AnimData[]} outputs - List of curve value data.
  * @param {AnimCurve[]} curves - The list of curves.
+ * @param {AnimEvents} animEvents - A sequence of animation events.
  */
 class AnimTrack {
-    constructor(name, duration, inputs, outputs, curves) {
+    constructor(name, duration, inputs, outputs, curves, animEvents) {
```
mvaligursky, line 17: use `animEvents = new AnimEvents([])` here

mvaligursky (approved):
approving with some minor comments

PR #3131, opened 2021-04-28
Fix recent change in Oculus Browser regarding Hands

Maksims:
Where exiting session still triggers one frame update but with `hidden` session state, which throws an exception in that case.


willeastcott (approved):

willeastcott:
Thank you! 🤗

Issue #3130, opened 2021-04-28
FRAMEBUFFER_INCOMPLETE_ATTACHMENT with particle system that uses depth softening value 0.1


yaustar:
Project: https://playcanvas.com/project/789935/overview/particle-framebuffer_incomplete

Create a particle effect with a Depth Softening value of 0.1
Launch project
Resize the window

Error will come up: `ERROR: FRAMEBUFFER_INCOMPLETE_ATTACHMENT`

PR #3132, opened 2021-04-29
Depth rendering refactor

mvaligursky:
Fixes: https://github.com/playcanvas/engine/issues/3130
Fixes: engine example post-effects on WebGl1 when window resizes.

Changes:
- functionality extracted from Application into a DepthLayer class
- refactored for improved compatibility with LayerComposition - when resizing, the render target is not dropped, but resized by reallocating texture of the same render target 
- fix to post-processing which was expecting event on rect change on the camera which was never called
- slight refactor to render target destroy functions to support resize of render targets


src/framework/components/camera/component.js
```diff
@@ -5,6 +5,7 @@ import { Component } from '../component.js';
 
 import { PostEffectQueue } from './post-effect-queue.js';
 
+// note: when this list modified, copy() function needs to be adjusted
```
willeastcott, line 8:
```suggestion
// note: when this list is modified, the copy() function needs to be adjusted
```

src/framework/components/camera/post-effect-queue.js
```diff
@@ -353,6 +349,7 @@ class PostEffectQueue {
         if (this.resizeTimeout)
             return;
 
+        // Note: this should be reviewed, as this would make postprocessing incorrect for few frames till reize takes place
```
willeastcott, line 352:
```suggestion
        // Note: this should be reviewed, as this would make postprocessing incorrect for a few frames 
        // until the resize takes place
```

src/framework/depth-layer.js
```diff
@@ -0,0 +1,261 @@
...
+    LAYERID_DEPTH, LAYERID_WORLD,
+    SHADER_DEPTH
+} from '../scene/constants.js';
+
+import { Layer } from '../scene/layer.js';
+
+// Internal class abstracting the access to the depth texture of the scene.
+// For webgl 2 devices, actual depth buffer is copying to a texture at specified time during camera rendering
+// for webgl 1 devices, the scene's depth is rendered to a separate RGBA texture
+class DepthLayer {
```
slimbuck, line 21: wondering about this naming. it's not a layer?

mvaligursky: yeah it's not a Layer .. it's equivalent to a Picker or post PostEffectQueue, but internally all it does it creates a special Layer with name "Depth"

src/framework/depth-layer.js
```diff
@@ -0,0 +1,261 @@
...
+
+import {
+    LAYERID_DEPTH, LAYERID_WORLD,
+    SHADER_DEPTH
+} from '../scene/constants.js';
+
+import { Layer } from '../scene/layer.js';
+
+// Internal class abstracting the access to the depth texture of the scene.
+// For webgl 2 devices, actual depth buffer is copying to a texture at specified time during camera rendering
```
willeastcott, line 19:
```suggestion
// For webgl 2 devices, actual depth buffer is copyied to a texture at specified time during camera rendering
```

src/framework/depth-layer.js
willeastcott: `DepthManager`?

willeastcott:
Imagine this code was Typescript. `application.js` would import `depth-layer.js` and `depth-layer.js` would need to import `application.js`. This is what I have alluded to in previous discussions. JavaScript can sometimes 'hide' circular dependencies because  we can often avoid having to directly reference class names in code. If we are ever forced to more explicitly define types in the future, this could cause us headaches. BTW, if we wanted VS Code JS type checking, we'd need to import `application.js` so JSDoc could resolve `Application`.

mvaligursky:
> BTW, if we wanted VS Code JS type checking, we'd need to import `application.js` in `depth-layer.js` so JSDoc could resolve `Application`.

If the module is only imported for type checking, this does not create a circular dependency as Babel would strip out those imports.
https://levelup.gitconnected.com/improving-babel-support-for-typescript-with-type-only-imports-28cb209d9460
_Babel is going to remove (elide) any import declarations that are only used as types._

willeastcott:
Rollup hits the source files first and _then_ Babel. And Rollup would complain about the circular dependencies.

But the point about Typescript still stands. You'd need to add:

```typescript
    application: Application;
```

...to `depth-layer.ts`.

src/framework/depth-layer.js
mvaligursky: I considered that very name .. but it implies a depth manager does a lot more than just copy out a depth buffer.
Maybe CameraDepthBuffer or CameraDepth?

mvaligursky: or SceneDepth?

mvaligursky:
> ...to `depth-layer.ts`. And therefore and import for `Application`.

This needs a bigger discussion .. as its not simple. What is the solution for Vec3 importing Mat4 and Mat4 importing Vec3 as an example? (not sure it's an actual case, but likely yes - for type checking anyways).

willeastcott:
> This needs a bigger discussion
Agreed.

> What is the solution for Vec3 importing Mat4 and Mat4 importing Vec3 as an example?
There are important questions to answer. Static functions in a third module that imports both `Vec3` and `Mat4` could be one?

willeastcott (approved):

src/framework/depth-layer.js
slimbuck: I'd be happy with SceneDepth.

mvaligursky: done, good name.

Issue #3136, opened 2021-04-29
OBJ loader gives errors when loading monkey head example


yaustar:
https://playcanvas.com/editor/scene/1141433

Errors:
![image](https://user-images.githubusercontent.com/16639049/116573065-67b7e380-a904-11eb-86bd-282313828940.png)

mvaligursky:
yeah it looks like something fails in the parser somewhere .. mesh creation gets non-matching streams of data and as of few weeks ago we use new Mesh API to create those meshes and it complains unlike the old one

<img width="625" alt="Screenshot 2021-04-29 at 17 33 41" src="https://user-images.githubusercontent.com/59932779/116586497-6370d380-a8ce-11eb-9ff3-fd32dc060b6e.png">

yaustar:
Link to OBJ loader from the repo https://github.com/playcanvas/engine/blob/master/scripts/parsers/obj-model.js

PR #3138, opened 2021-04-29
Fix to obj format parser when there are no uvs

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/3136


yaustar (approved):
LGTM :)

PR #3151, opened 2021-05-07
Remove gl resize

### LINKED ISSUE #3114: Refactor gl.canvas to use clientWidth/Height instead of relying on external calls to resizeCanvas

yaustar:
Fixes #3114

Tested against model-outline, WebXR VR and AR projects as well as other private projects that previously ran into issues on the release of 1.41.0

Edit Sun 9 May 2021: Tested against water reflection demo: https://launch.playcanvas.com/1144354?debug=true ([forum thread](https://forum.playcanvas.com/t/solved-how-to-determine-location-entity-name-that-used-deleted-texture/19938/))


Maksims:
Not a big fan of `updateGlCanvasSize` naming. `updateCanvasSize` would do as good?

yaustar:
Actually, I take that back. `updateCanvasSize` does make sense after sitting on it for a bit.

mvaligursky (approved):

yaustar:
Planning a little ahead here. This function will need to be called when the resize and orientation events are fired for some older projects to works as they depend on the canvas resolution being correct after a resize event.

src/graphics/graphics-device.js
```diff
@@ -3496,8 +3497,8 @@ class GraphicsDevice extends EventHandler {
         this._height = height;
 
         var ratio = Math.min(this._maxPixelRatio, window.devicePixelRatio);
-        width *= ratio;
-        height *= ratio;
+        width = Math.floor(width * ratio);
+        height = Math.floor(height * ratio);
 
         if (this.canvas.width === width && this.canvas.height === height)
             return;
```
mvaligursky, line 3504: while you're here, maybe flip the logic and get rid of the return.

mvaligursky (approved):

src/graphics/graphics-device.js
yaustar: Done!

mvaligursky: perfect.

PR #3155, opened 2021-05-11
Shadow mapping refactoring

mvaligursky:
- refactoring and clean up of shadow rendering code in preparation for cascaded shadow changes
- when castShadow is disabled, shadow map is released
- lightmapper - shadow maps are shared between lights instead of allocated / destroyed per light
- render actions - fix to only keep list of shadow casting directional lights, not all directional lights
- added spot light to lights-baked example
- lights example is set up to use VSM directional shadow to test all code paths


src/scene/light.js
```diff
@@ -317,54 +362,16 @@ class Light {
...
-
     updateShadow() {
         if (this.shadowUpdateMode !== SHADOWUPDATE_REALTIME) {
             this.shadowUpdateMode = SHADOWUPDATE_THISFRAME;
         }
     }
 
     layersDirty() {
-        this._scene.layers._dirtyLights = true;
+        if (this?._scene?.layers) {
```
willeastcott, line 372: I don't think you need the first optional chaining operator here since `this` should be valid.

willeastcott (approved):

willeastcott:
Approved with some minor observations.

src/scene/light.js
mvaligursky: cheers, agreed and done.

PR #3158, opened 2021-05-12
Small miscellaneous fixes

slimbuck:
This PR combines the following changes:
- destroy render target when prefiltering
- support single cubemap PMREM case on webgl 2.0 devices
- small texture code cleanup
- add note on incorrect mipmap generation behaviour for cubemaps

### DIFF
--- a/src/graphics/graphics-device.js
+++ b/src/graphics/graphics-device.js
@@ -1555,6 +1555,8 @@ class GraphicsDevice extends EventHandler {
             // If the active render target is auto-mipmapped, generate its mip chain
             var colorBuffer = target._colorBuffer;
             if (colorBuffer && colorBuffer._glTexture && colorBuffer.mipmaps && (colorBuffer.pot || this.webgl2)) {
+                // FIXME: if colorBuffer is a cubemap currently we're re-generating mipmaps after
+                // updating each face!
                 this.activeTexture(this.maxCombinedTextures - 1);
                 this.bindTexture(colorBuffer);
                 gl.generateMipmap(colorBuffer._glTarget);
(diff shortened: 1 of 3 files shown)


mvaligursky (approved):
looks good!

src/graphics/graphics-device.js
```diff
@@ -2191,16 +2193,11 @@ class GraphicsDevice extends EventHandler {
...
             if (texture === this.grabPassTexture) {
-                processed = this.updateGrabPass();
-
-                processed = true;
-            }
-
-            if (!processed && (texture._needsUpload || texture._needsMipmapsUpload)) {
+                // grab framebuffer to be used as a texture - this returns false when not supported for current render pass
+                // (for example when rendering to shadow map), in which case previous content is used
+                this.updateGrabPass();
```
willeastcott, line 2199: If you no longer require the return value from `updateGrabPass`, is it worth removing it from the code?

slimbuck: We could do, but since the return value expresses something meaningful, I'd probably be inclined to keep it (even though the value isn't used at the call site).

Happy to remove it though if you feel strongly about it.

willeastcott: Nope, I don't feel strongly about it.

mvaligursky: actually, I think you should not have made this change .. I missed there was a return value there when approving. Have you tried running engine grab pass example with the change?

slimbuck: I chose to submit this code change because of the confusing logic:
```javascript
                processed = this.updateGrabPass();
                processed = true;
```

The return value was being ignored anyway and this PR doesn't change execution at all.

mvaligursky: that example might not reproduce the issue .. but try changing this in it to true as well
```
        glass.model.castShadows = false;
        glass.model.receiveShadows = false;

```

slimbuck: What I mean is I've made no change to this logic, only made the code more readable.

Of course we'll test, but do you see a change in the logic?

mvaligursky: I see .. somehow a bad merge got submitted. The line 
`processed = true;`
needs to be removed

slimbuck: OIC interesting. That's the PR.. I'll fix it, but will need some help testing! https://github.com/playcanvas/engine/commit/9c594e4707bdb3e9b72e404d1b98244062dd8e3a#diff-1733a306b13044b918b7a6bb04de67f86043de3b6821cfef27eccbd94e934f7aR2170

mvaligursky: will do when submitted. I already tested the repro case I mentioned .. it fails, and the code needs to change back with that processed=true line removed.

slimbuck: done

mvaligursky: nice, looks good, thanks!

PR #3162, opened 2021-05-13
Further small shadow rendering refactor + fix

mvaligursky:
- refactor around shadow matrix storage
- fix for cookie on a spot light without shadows enabled

### DIFF
--- a/src/scene/renderer/forward-renderer.js
+++ b/src/scene/renderer/forward-renderer.js
@@ -622,7 +622,7 @@ class ForwardRenderer {
                     directional._normalOffsetBias;
 
                 this.lightShadowMapId[cnt].setValue(lightRenderData.shadowBuffer);
-                this.lightShadowMatrixId[cnt].setValue(directional._shadowMatrix.data);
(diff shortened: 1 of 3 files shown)


src/scene/renderer/forward-renderer.js
```diff
@@ -768,9 +772,12 @@ class ForwardRenderer {
 
                 shadowCamView.setTRS(cookieNode.getPosition(), cookieNode.getRotation(), Vec3.ONE).invert();
                 shadowCamViewProj.mul2(cookieCam.projectionMatrix, shadowCamView);
-                spot._shadowMatrix.mul2(ShadowRenderer.scaleShiftMatrix, shadowCamViewProj);
+
+                if (!spot._cookieMatrix) spot._cookieMatrix = new Mat4();
```
willeastcott, line 776: Could you just hide this in the `Light` class? So in the getter of `Light#cookieMatrix` check if it's `null` and create it if necessary?

mvaligursky: absolutely! done.

willeastcott (approved):

PR #3164, opened 2021-05-16
Upgrade RigidBodyComponent to current component structure

willeastcott:
Upgrades the `RigidBodyComponent` to the current recommended component format. This moves component data properties into the core `RigidBodyComponent`. This makes property access faster and the code easier to read/debug.

This PR also upgrades the code to ES6 (mainly `var` to `let` and `const`...and default parameters).

The physics render script was (naughtily) accessing the non-API `RigidBodyComponent#data` property so that's been fixed too.

### DIFF
--- a/src/framework/components/rigid-body/component.js
+++ b/src/framework/components/rigid-body/component.js
@@ -20,9 +20,36 @@ var ammoVec1, ammoVec2, ammoQuat, ammoOrigin;
  * @name RigidBodyComponent
  * @augments Component
  * @classdesc The rigidbody component, when combined with a {@link CollisionComponent}, allows your
- * entities to be simulated using realistic physics.
- * A rigidbody component will fall under gravity and collide with other rigid bodies. Using scripts, you
- * can apply forces and impulses to rigid bodies.
+ * entities to be simulated using realistic physics. A rigidbody component will fall under gravity and
+ * collide with other rigid bodies. Using scripts, you can apply forces and impulses to rigid bodies.
+ *
+ * You should never need to use the RigidBodyComponent constructor. To add an RigidBodyComponent to a
+ * {@link Entity}, use {@link Entity#addComponent}:
+ *
+ * ~~~javascript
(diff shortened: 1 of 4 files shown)


src/framework/components/rigid-body/component.js
```diff
@@ -72,99 +72,263 @@ class RigidBodyComponent extends Component {
...
-        var body = this.body;
-        if (body && this.type === BODYTYPE_DYNAMIC) {
-            body.activate();
+    get mass() {
+        return this._mass;
+    }
 
-            ammoVec1.setValue(av.x, av.y, av.z);
-            body.setAngularVelocity(ammoVec1);
+    set mass(mass) {
```
LeXXik, line 252: Thinking out loud, if a mass for a dynamic body would be set to zero here, then the component system would still treat it as a dynamic body, while Ammo will change it to static. Perhaps, we can update the type here as well, which would call disableSimulation, which in turn would remove the dynamic object from the system._dynamic list.

willeastcott: I've wondered about stuff like that too. So, a couple of things:

1. This PR shouldn't change any existing functionality. The physics engine should behave exactly the same as before (but with better structure and performance). If we changed things so that `this.entity.rigidbody.mass = 0;` switched the body's type to static, that could potentially change the behavior of some peoples' apps.
2. I'm not a fan of implicitly changing the value of one component's property by setting another. Feels too 'magical' to me (to use a word @Maksims likes to use).

LeXXik: Yes, that makes sense. This is simply the default Ammo behavior when one updates the mass of a rigidbody. Thought it was worth mentioning.
https://github.com/kripken/ammo.js/blob/99d0ec0b1e26d7ccc13e013caba8e8a5c98d953b/bullet/src/BulletDynamics/Dynamics/btRigidBody.cpp#L231-L241

ellthompson (approved):

PR #3165, opened 2021-05-16
Add scene option to disable skybox

slimbuck:
It is currently not possible to set PMREM lighting on the scene without also enabling the skybox.

This PR adds an option to disable skybox.

Also fixes some incorrect state propagation.

### DIFF
--- a/src/scene/scene.js
+++ b/src/scene/scene.js
@@ -114,6 +114,7 @@ class Scene extends EventHandler {
 
         this._skyboxPrefiltered = [null, null, null, null, null, null];
 
+        this._skyboxEnabled = true;
         this._firstUpdateSkybox = true;
         this._skyboxCubeMap = null;
         this.skyboxModel = null;
@@ -130,9 +131,6 @@ class Scene extends EventHandler {
         this.lightmapSizeMultiplier = 1;
         this.lightmapMaxResolution = 2048;
         this.lightmapMode = BAKE_COLORDIR;
-
-        // fallback to prefilter textures for the skybox
-        this.skyboxPrefilterFallback = true;
 
         this._stats = {
             meshInstances: 0,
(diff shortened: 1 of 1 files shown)


src/scene/scene.js
```diff
@@ -200,14 +201,22 @@ class Scene extends EventHandler {
         }
     }
 
+    get skyboxEnabled() {
```
willeastcott, line 204: I'm not sure about the naming here. If a user does `scene.skyboxEnabled = false;`, they might expect it to have the same effect as having no skybox at all. But this just disables _rendering_ of the skybox - it doesn't disable the skybox as a source for IBL, right?

slimbuck: Yeah it's unfortunate that scene conflates skybox and lighting data. In the scene, skybox means both the background image as well as the IBL data. Ideally the scene would have a 'skybox' and 'lighting data' and they should be two separate things.

Re naming, 'renderSkybox' sounds to me like it's going to render a skybox :) Can't really think of a better name though.

Maksims:
Disabling Skybox Layer - does the same?

mvaligursky:
I don't like that this is a setting on a scene. With rendering to texture getting common, people might want to not render skybox to textures for example, but want to render it to only maybe one camera. Or the other way around. So ideally this is a camera setting instead.

Currently we expose clearColorBuffer, clearDepthBuffer and clearStencilBuffer, we could add 'clearToSkybox' maybe?

Alternatively, which is probably my preference: we already have a Skybox layer LAYERID_SKYBOX. The skybox mesh instance is in this layer, and then it's up to the user to add or remove this layer from the camera. This is what I used in render to texture / render to cubemap examples. I either add or not skybox layers to main / render to texture cameras.

slimbuck:
Yup agreed guys - thanks! I'm closing this.

PR #3167, opened 2021-05-17
Updated reprojectTexture

slimbuck:
Added the option of limiting reprojectTexture to a single face of a cubemap.

Also moved the options into an option argument as follows. 

The function now has the following signature:
``` javascript
pc.reprojectTexture(source, target, options = { });
```

Which replaces the old style:
```javascript
pc.reprojectTexture(device, source, target, specularPower = 1, numSamples = 1024);
```

### DIFF
--- a/src/graphics/reproject-texture.js
+++ b/src/graphics/reproject-texture.js
@@ -62,7 +62,7 @@ function getProjectionName(projection) {
 function reprojectTexture(source, target, options = {}) {
     // maintain backwards compatibility with previous function signature
     // reprojectTexture(device, source, target, specularPower = 1, numSamples = 1024)
-    if (source.constructor === GraphicsDevice) {
+    if (source instanceof GraphicsDevice) {
         source = arguments[1];
         target = arguments[2];
         options = {
--- a/src/graphics/reproject-texture.js
+++ b/src/graphics/reproject-texture.js
@@ -48,15 +48,23 @@ function getProjectionName(projection) {
  * function can read and write textures with pixel data in RGBE, RGBM, linear and sRGB formats. When
  * specularPower is specified it will perform a phong-weighted convolution of the source (for generating
  * a gloss maps).
- * @param {GraphicsDevice} device - The graphics device.
  * @param {Texture} source - The source texture.
  * @param {Texture} target - The target texture.
- * @param {number} [specularPower] - Optional specular power. When specular power is specified,
+ * @param {object} [options] - The options object.
+ * @param {GraphicsDevice} [options.device] - Optional graphics device (default is the source texture device).
+ * @param {number} [options.specularPower] - Optional specular power. When specular power is specified,
  * the source is convolved by a phong-weighted kernel raised to the specified power. Otherwise
  * the function performs a standard resample.
- * @param {number} [numSamples] - Optional number of samples (default is 1024).
+ * @param {number} [options.numSamples] - Optional number of samples (default is 1024).
+ * @param {number} [options.face] - Optional cubemap face to update (default is update all faces).
  */
-function reprojectTexture(device, source, target, specularPower = 1, numSamples = 1024) {
+function reprojectTexture(source, target, options = {}) {
+    // extract options
+    const device = options.device || source.device;
+    const specularPower = options.hasOwnProperty('specularPower') ? options.specularPower : 1;
+    const numSamples = options.hasOwnProperty('numSamples') ? options.numSamples : 1024;
+    const face = options.hasOwnProperty('face') ? options.face : null;
+
     const processFunc = (specularPower === 1) ? 'reproject' : 'prefilter';
     const decodeFunc = "decode" + getCoding(source);
     const encodeFunc = "encode" + getCoding(target);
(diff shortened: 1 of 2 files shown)


mvaligursky:
- It seems good change, but I would maintain backwards compatibility (worth few lines of code) as I know multiple people are already using this API.
- please list API changes in the description

src/graphics/reproject-texture.js
```diff
@@ -48,15 +49,38 @@ function getProjectionName(projection) {
...
+            specularPower: arguments[3] === undefined ? 1 : arguments[3],
+            numSamples: arguments[4] === undefined ? 1024 : arguments[4]
+        };
+        // #if _DEBUG
+        console.warn('DEPRECATED: please use the updated pc.reprojectTexture API.');
+        // #endif
+    }
+
+    // extract options
+    const device = options.device || source.device;
```
willeastcott, line 79: I don't see how it's possible for `source.device` to be invalid.

slimbuck: Correct. Question is whether it's necessary to keep supporting user provided device?

slimbuck:
@mvaligursky I've added backwards compatible code, but TBH I had hoped it's not necessary. I'd much rather we don't have code like this littered about the engine. (Though of course I'd also rather we have a better API going forward).

src/graphics/reproject-texture.js
mvaligursky: probably remove options.device completely .. texture always has a valid device.

willeastcott:
This is one of the few cases where I'm 99% sure nobody's using this function (yet).

src/graphics/reproject-texture.js
slimbuck: I'll remove :+1:

src/graphics/reproject-texture.js
```diff
@@ -48,15 +49,38 @@ function getProjectionName(projection) {
...
  * the function performs a standard resample.
- * @param {number} [numSamples] - Optional number of samples (default is 1024).
+ * @param {number} [options.numSamples] - Optional number of samples (default is 1024).
+ * @param {number} [options.face] - Optional cubemap face to update (default is update all faces).
  */
-function reprojectTexture(device, source, target, specularPower = 1, numSamples = 1024) {
+function reprojectTexture(source, target, options = {}) {
+    // maintain backwards compatibility with previous function signature
+    // reprojectTexture(device, source, target, specularPower = 1, numSamples = 1024)
+    if (source.constructor === GraphicsDevice) {
```
mvaligursky, line 65: I'm not very keen on this .. it will break in the future when we have base device class and webgl/webgpu implementation. That's a reason why I created https://github.com/playcanvas/engine/pull/3118

mvaligursky: and it also forces GraphicsDevice to be imported.

mvaligursky: you might use more correct instanceof in that case

slimbuck: yep, done

src/graphics/reproject-texture.js
```diff
@@ -48,15 +49,37 @@ function getProjectionName(projection) {
...
+        // #if _DEBUG
+        console.warn('DEPRECATED: please use the updated pc.reprojectTexture API.');
+        // #endif
+    }
+
+    // extract options
+    const device = source.device;
+    const specularPower = options.hasOwnProperty('specularPower') ? options.specularPower : 1;
+    const numSamples = options.hasOwnProperty('numSamples') ? options.numSamples : 1024;
+    const face = options.hasOwnProperty('face') ? options.face : null;
```
mvaligursky, line 81: I think it'd be better to keep face of the same type (int) .. so assign -1 to it instead of null.

mvaligursky (approved):
approving with one small comment.

mvaligursky:
also maybe add deprecated warning .. so that later we can remove the support for it.

slimbuck:
@mvaligursky something like this maybe? https://github.com/playcanvas/engine/pull/3167/files#diff-4999255070c170714a51934441b92884e5fbc1fd12351ad47cf5154a75850ed8R73 :+1:

src/graphics/reproject-texture.js
slimbuck: What advantage do you see using the same type? The value 'null' communicates 'no face' better than the value '-1' IMO.

mvaligursky: it does, but if one day codebase moves to TS, what type do you give to the variable? I believe JIT compilation can optimise better if type does not change.

src/graphics/reproject-texture.js
```diff
@@ -48,15 +49,37 @@ function getProjectionName(projection) {
  * function can read and write textures with pixel data in RGBE, RGBM, linear and sRGB formats. When
  * specularPower is specified it will perform a phong-weighted convolution of the source (for generating
  * a gloss maps).
- * @param {GraphicsDevice} device - The graphics device.
  * @param {Texture} source - The source texture.
  * @param {Texture} target - The target texture.
- * @param {number} [specularPower] - Optional specular power. When specular power is specified,
+ * @param {object} [options] - The options object.
+ * @param {GraphicsDevice} [options.device] - Optional graphics device (default is the source texture device).
```
willeastcott, line 55: I think you meant to delete this line?

willeastcott (approved):

willeastcott:
Approved with one comment.

src/graphics/reproject-texture.js
slimbuck: yes! thanks

src/graphics/reproject-texture.js
slimbuck: I would do the same (or similar) in TS tbh: https://stackoverflow.com/questions/17220114/how-to-declare-a-type-as-nullable-in-typescript

willeastcott:
Let me make a minor comment about engine architecture. Today, we pass the `GraphicsDevice` into many object constructors (e.g. `Texture`, `VertexBuffer`, `IndexBuffer`, etc). I've talked before about an alternative whereby these classes just hold data/values and the WebGL resources are created/managed in the `GraphicsDevice` and `ForwardRenderer`. This would simplify many classes and remove lots of cyclic dependencies (IMHO). Removing the `GraphicsDevice` parameter from this function is a (small) barrier to us going down that path. It's not enough for me to want to hold up this PR. Just sayin'. 😄

slimbuck:
In that case you'd need the device to perform the rendering?

willeastcott:
Yeah.

PR #3169, opened 2021-05-17
Create skybox layer enabled

slimbuck:
This PR enables the skybox layer at construction time and removes a previously introduced workaround, see #1838.

The skybox layer is currently constructed disabled and is then enabled by the scene during the first invocation of `_updateSkybox`.

This can result in the user attempting to disable the skybox layer only to have it re-enabled on first render.

@guycalledfrank @Maksims: can you remember why create the layer disabled in the first place?

### DIFF
--- a/src/framework/application.js
+++ b/src/framework/application.js
@@ -473,7 +473,7 @@ class Application extends EventHandler {
         this.defaultLayerDepth = this.sceneDepth.layer;
 
         this.defaultLayerSkybox = new Layer({
-            enabled: false,
+            enabled: true,
             name: "Skybox",
             id: LAYERID_SKYBOX,
             opaqueSortMode: SORTMODE_NONE
--- a/src/scene/scene.js
+++ b/src/scene/scene.js
@@ -114,7 +114,6 @@ class Scene extends EventHandler {
 
         this._skyboxPrefiltered = [null, null, null, null, null, null];
 
-        this._firstUpdateSkybox = true;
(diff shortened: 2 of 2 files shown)


slimbuck:
The workaround being removed was introduced in #1838

mvaligursky:
Have you tested this with the Editor? (just double checking). I think when we load Editor project, we don't use the default layers but they get created based on some data file.

Maksims:
Also, please check how it affects cubemap loading, especially scenarios where faces are set to not preload, while prefiltered data is preloading, and mip is selected for skybox.

slimbuck:
Thanks @mvaligursky and @Maksims. I'll definitely fully test this change and try what you suggest.

Just thinking about the possible impact and wondering if I am missing anything: I don't expect behavior to change because now even with the layer enabled at the start there is no skybox (since there is no mesh). When the scene creates the mesh, the skybox will appear. But this matches previous behavior, since the layer was enabled when the mesh was created.

The only difference could be if developers are creating their own meshes in the skybox layer and depend somehow on the scene to enable it.

Maksims:
Also, please check if it affects the scenario where cubemap is not preloaded, will it first create a mesh and render it with default material, and only then apply cubemap?

Currently it likely to not render skybox at all, until cubemap is loaded. With this change this will stay the same, or will it render white skybox first, and once loaded cubemap swap texture to it?

willeastcott (approved):

PR #3170, opened 2021-05-17
Fix curve evaluator crash

LeXXik:
When the `while` loop reaches the end of the array, the application crashes, trying to evaluate property 0 of an element that does not exist:
`TypeError: Cannot read property '0' of undefined`

The fix adds a check that the element exists, before accessing its property 0.

Fixes #

### DIFF
--- a/src/math/curve-evaluator.js
+++ b/src/math/curve-evaluator.js
@@ -78,7 +78,7 @@ class CurveEvaluator {
                 // be more efficient to perform a binary search here instead. Which is
                 // straight forward thanks to the sorted list of knots).
                 let index = 0;
-                while (time >= keys[index + 1][0]) {
+                while (keys[index + 1] && time >= keys[index + 1][0]) {
                     index++;
                 }
                 this._left = keys[index][0];


slimbuck:
Hi @LeXXik,

Thanks for this PR. I might be missing something, but that error should not be happening. Perhaps the knot times are not sorted correctly or the data is invalid? Are you able to provide a repo so I can take look?

Thanks!

LeXXik:
This doesn't happen always, since I assume we get into other `if` blocks on most of occasions. I will see if I can reproduce it to check the timings.

LeXXik:
I believe the cause is due to the time value having some fraction over 1.0 when the percent is calculated (passed/total). In such scenarios, a time value could be something like 1.001261522547516, while the keys are ranged from 0 to 1, allowing the loop to reach the end of it, but not stopping and trying to read element that does not exist. 

I will clamp it to 1 in the user script as a fix.

slimbuck:
I would expect line https://github.com/playcanvas/engine/pull/3170/files#diff-69afb7d9b81d326efc760fee037416c0b538049dc1edce60455fac4a0b6e33f0R67 to catch that case.

slimbuck:
...oops link doesn't work properly. I meant line 67.

LeXXik:
True that. I will try to catch it to peek the values.

LeXXik:
I was able to capture the crash. The issue was caused by the way the percent was calculated, where the duration was a function of distance in some animation context. Once distance happened to be zero, the percent would evaluate to NaN (time / duration), which is then passed to the curve evaluator, which then evaluates the time value to Infinity.

slimbuck:
Oh yes interesting. @ellthompson could perhaps look into this further?

ellthompson:
Could the first if statement in the function also capture non finite time values?

slimbuck:
I think there are two questions:
- should the animation system be generating NaN times?
- does CurveEvaluator expect NaN as valid inputs?

LeXXik:
The user script was not correct to generate a NaN value. Thanks to CurveEvaluator we were able to catch the bug :) We placed guards against division by zero now, so the user script should work properly even without this patch.

yaustar:
@LeXXik Do you think this should still be merged in?

LeXXik:
After giving it a thought, NaN is not a valid input and it breaks something inside an engine as a result. I believe this falls under a more broad issue of input validation, that already has a separate ticket about. I am closing this as not a valid PR. Feel free to re-open if you consider otherwise.

PR #3177, opened 2021-05-20
Post Process Pixelate

leonidaspir:
A simple pixelate post effect that allows you to set the pixel size. The Post Effects example has been updated (the effect is initially disabled).

image

### DIFF
--- a/examples/graphics/post-effects.html
+++ b/examples/graphics/post-effects.html
@@ -70,7 +70,11 @@
             godRays: {
                 type: "script",
                 url: "../../scripts/posteffects/posteffect-godrays.js"
-            }                                 
+            },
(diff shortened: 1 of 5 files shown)


LeXXik:
You are on fire, Leonidas! :D 
Hmm, should we find another way to put them all together, seeing we are out of numbers?

leonidaspir:
> Hmm, should we find another way to put them all together, seeing we are out of numbers?

Hmm, good point, I've just figured out the same, I am adding another one and that's it with the numbers 😇 I will see if I think of a quick fix on this.

LeXXik:
Nah, just go with KEY_Q, KEY_W, etc :)

PR #3180, opened 2021-05-20
Remove support for shadow coordinates generation in VS 

mvaligursky:
- a single directional could have it’s shadow coordinates evaluated in the VS instead of PS as all the other lights. This is now getting removed as unneeded complexity, before cascaded shadow lights are finalized as they don’t support this.


slimbuck:
wow best PR ever

slimbuck (approved):

willeastcott:
Hey @guycalledfrank - any concerns about this one (just in case we aren't considering all the consequences)?

Maksims:
I believe there was a need to calculate shadow coordinates in VS to ensure that texels wont drift with camera movement.

Directional light shadow is linked to normal view camera, so shadowmap camera moves together with main camera, but this leads to drifting texels, so with small camera movements you can see how PCF smoothing will constantly move around. I remember @guycalledfrank implemented a way to ensure texels are "sticky", so it seems like camera is always moving exactly by a texel size.

Hopefully, I've explained it well :D

mvaligursky:
The shadow map texel snapping is done elsewhere, where the shadow camera view matrix is evaluated. So shadows don't flicker even when coordinates are computed on a pixel level. So I don't think that's the reason.

willeastcott (approved):

guycalledfrank:
Shadow coords in VS were not related to texel snapping, it's separate. It was done simply for performance. When we were working on Seemore, every single thing executed in PS had noticeable cost, even computing shadow coords.

PR #3182, opened 2021-05-20
Post effects example new effects

leonidaspir:
The post effects example has been updated with the following:
- Orbit camera enabled, mainly to allow for testing the motion blur effect.
- A second point light has been added for the god rays effect.
- Key bindings for the new effects and the Boke Debug View mode.

All new or updated effects are included in this PR so the example can function, but should be reviewed on their separate PR.

Screenshot 2021-05-20 215636


franmts:
Please include on next release, super useful! Thanks Leonidas!

Maksims:
Would be nice to make bokeh more fixed on one point, instead of linked to light position.
As it has very small range, and mostly shows statue blurred, I was confused first time "is it working correctly".

PR #3188, opened 2021-05-24
Remove unused vertex shader globals

slimbuck:
Tested against engine examples.


mvaligursky (approved):
looks good, those seems to be used only in frags.

willeastcott (approved):

PR #3201, opened 2021-05-28
Faster GUID generator

munrocket:
Benchmark: https://codepen.io/munsocket/pen/QWyLKMB?editors=1010

I confirm I have read the [contributing guidelines](https://github.com/playcanvas/engine/blob/master/.github/CONTRIBUTING.md) and not signed [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform)

### DIFF
--- a/src/core/guid.js
+++ b/src/core/guid.js
@@ -1,3 +1,9 @@
+// Look up table for GUIDs
+const LUT = [];
+for (let i = 0; i < 256; i++) {
+    LUT[i] = (i < 16 ? '0' : '') + i.toString(16);
+}
+
 /**
  * @name guid
  * @namespace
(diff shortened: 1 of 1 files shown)


yaustar:
Thanks for this @munsocket! Where is the algorithm from/what is it called for us to check against please?

munsocket:
The algorithm is pretty common, we a creating look up table with alphabet and than iterating through it randomly.

It’s based on this stackoverflow https://stackoverflow.com/questions/105034/how-to-create-a-guid-uuid
But with fixed range.

Maksims:
For performance, please run tests many times to warm up the V8.

munsocket:
Already created variable count, if it will be bigger than some number it will be wamed up. Not sure that this code will be “hot” in PlayCanvas for v8 enough to be optimized, only if somebody creating 3d objects in loop or in big scenes. But in general regex solutions is slower.

src/core/guid.js
```diff
@@ -1,6 +1,11 @@
+// Look up table for GUIDs
+const LUT = [];
+for (let i = 0; i < 256; i++) {
+    LUT[i] = (i < 16 ? '0' : '') + i.toString(16);
+}
+
 /**
  * @name guid
- * @namespace
```
willeastcott, line 3: I'm curious why this PR removes this JSDoc tag.

munsocket: Ah, so this is a namespace. Fixed.

willeastcott:
I just ran the benchmark and I saw this:

image

So there's tjs, new and new2. Which one is this PR implementing. The second has a `NaN` in it (from my run of the benchmark, despite being the fastest. And new2 doesn't print the GUID.

munsocket:
Sorry for not being clear, I am tested different approaches (crypto.getRandomValues and time based).
Here more clear benchmark that show previous and new GUID generators
https://codepen.io/munsocket/pen/QWyLKMB




I am also tested collision detection from built in PRNGs, if you are interested.
https://codepen.io/munsocket/pen/dyGoqpw

willeastcott:
Cool, this is great @munsocket, thanks! You might have missed [this link](https://docs.google.com/a/playcanvas.com/forms/d/e/1FAIpQLSfG9XHIgdapCBkvwxbXow592Y5NFPEKD5TM-TyWOmbaCbLeHA/viewform) at the top - would you be able to complete that so I can merge, please? Thanks!

mvaligursky:
> PS: Seems that I found and error, better to stick to Jeff Ward MIT version that is used in Three.js too.

Just to confirm .. has this error been fixed in your following commit? Or is this still TODO.

munrocket:
Yes fixed. Your contribution license is really weird because it’s asking my phone number and county.  I am trying to keep anonymity and not planning to share this information, so PR can be closed.

kungfooman:
Can't we just close it then?

1) It adds more code and it looks more complicated
2) I've never seen this in any profiling session as a bottleneck

PR #3211, opened 2021-06-03
Add renderAsset property to particlesystem component

vkalpias:


willeastcott (approved):

src/scene/render.js
```diff
@@ -1,5 +1,13 @@
 import { EventHandler } from '../core/event-handler.js';
 
+/**
+ * @class
+ * @name Render
+ * @augments EventHandler
+ * @classdesc A render contains an array of meshes that come from a GLB model. The render is the resource of a Render Asset.
```
mvaligursky, line 7:
```suggestion
 * @classdesc A render contains an array of meshes that are referenced by a single hierarchy node in a GLB model, and are accessible using {@link ContainerResource#renders} property. The render is the resource of a Render Asset.
```

mvaligursky (approved):

src/framework/components/particle-system/component.js
```diff
@@ -477,12 +480,6 @@ class ParticleSystemComponent extends Component {
             asset = assets.get(newValue);
             if (asset) {
                 this._bindMeshAsset(asset);
-
-                if (asset.resource) {
```
vkalpias, line 481: Removed this bit because bindMeshAsset does it too

mvaligursky:
looks good to me.

PR #3217, opened 2021-06-07
Removed 'if not null' documentation as it cannot be null in Particle Component

yaustar:
Fixes https://github.com/playcanvas/developer.playcanvas.com/issues/238

### DIFF
--- a/src/framework/components/particle-system/component.js
+++ b/src/framework/components/particle-system/component.js
@@ -171,18 +171,18 @@ var depthLayer;
  *
  * @property {Vec3} particleNormal (Only for PARTICLEORIENTATION_WORLD and PARTICLEORIENTATION_EMITTER) The exception of extents of a local space bounding box within which particles are not spawned. Aligned to the center of EmitterExtents.
  * @property {CurveSet} localVelocityGraph Velocity relative to emitter over lifetime.
- * @property {CurveSet} localVelocityGraph2 If not null, particles pick random values between localVelocityGraph and localVelocityGraph2.
+ * @property {CurveSet} localVelocityGraph2 Particles pick random values between localVelocityGraph and localVelocityGraph2.
  * @property {CurveSet} velocityGraph World-space velocity over lifetime.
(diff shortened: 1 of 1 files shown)


willeastcott:
Wait, are you saying that if I add a `particlesystem` component to an entity and I set `localVelocityGraph`, and then query `localVelocityGraph2`, does it return `localVelocityGraph` or `null`...or something else? This is a bit confusing.

yaustar:
Here is the original thread: https://forum.playcanvas.com/t/change-local-velocity-of-particle-system/17961/

The current documentation implies that localVelocityGraph2 was optional but it isn't the case. Both localVelocityGraph and localVelocityGraph2 need to be set for a particle system to behave correctly.

In the Editor, if randomise is unticked for the graph, it will set both localVelocityGraph and localVelocityGraph2 to be the same graph.

mvaligursky:
I looked at engine example particles-snow, and when I comment out these two lines:

```
                velocityGraph2: velocityCurve2,
                rotationSpeedGraph2: rotCurve2,
```

the example still runs, but the randomisation is gone. That suggests that these parameters can be null.

yaustar:
Doesn't seem to work when velocityGraph2 is explicitly set to null after creation though: https://playcanvas.com/editor/scene/1093460

```
var ParticleVelocity = pc.createScript('particleVelocity');

ParticleVelocity.prototype.initialize = function() {
    var self = this;
    setTimeout(function() {
        console.log(self.entity.particlesystem.localVelocityGraph);
        self.entity.particlesystem.localVelocityGraph = new pc.CurveSet([

            [0, 0],


            [0, 0],


            [0, 0]

        ]);
        
        self.entity.particlesystem.localVelocityGraph2 = null;
        
        self.entity.particlesystem.reset();
        console.log(self.entity.particlesystem.localVelocityGraph);
    }, 1000);
};

```

mvaligursky:
We need to check if this is a bug or expected behaviour.

PR #3220, opened 2021-06-08
Add custom bounding box properties to model / render components

vkalpias:
**API CHANGE**
- **aabb** property of RenderComponent and ModelComponent has been renamed to **customAabb**, to allow aabb property to be used for different purpose in the future.
- This also fixes the issue with this customAabb property usage. Before this PR, it was incorrectly handled as if specified in world space. This is correctly being handled as in local space now. 

This PR improves / fixes functionality introduced here: https://github.com/playcanvas/engine/pull/2435

This PR also adds support to specify custom AABB in addComponent for model and render component. This is to allow customAabb to be exposed in the Editor by separate PR.

### DIFF
--- a/src/deprecated.js
+++ b/src/deprecated.js
@@ -99,6 +99,7 @@ import { ComponentSystem } from './framework/components/system.js';
 import { Entity } from './framework/entity.js';
 import { LightComponent } from './framework/components/light/component.js';
 import { ModelComponent } from './framework/components/model/component.js';
+import { RenderComponent } from './framework/components/render/component.js';
 import {
     BODYFLAG_KINEMATIC_OBJECT, BODYFLAG_NORESPONSE_OBJECT, BODYFLAG_STATIC_OBJECT,
     BODYSTATE_ACTIVE_TAG, BODYSTATE_DISABLE_DEACTIVATION, BODYSTATE_DISABLE_SIMULATION, BODYSTATE_ISLAND_SLEEPING, BODYSTATE_WANTS_DEACTIVATION,
@@ -995,6 +996,34 @@ ModelComponent.prototype.setVisible = function (visible) {
     // #endif
     this.enabled = visible;
 };
+
+Object.defineProperty(ModelComponent.prototype, "aabb", {
+    get: function () {
+        // #if _DEBUG
+        console.error('DEPRECATED: pc.ModelComponent#aabb is deprecated. Use pc.ModelComponent#customAabb instead - which expects local space AABB instead of a world space AABB.');
+        // #endif
+        return null;
+    },
+    set: function (type) {
+        // #if _DEBUG
+        console.error('DEPRECATED: pc.ModelComponent#aabb is deprecated. Use pc.ModelComponent#customAabb instead - which expects local space AABB instead of a world space AABB.');
+        // #endif
+    }
+});
+
+Object.defineProperty(RenderComponent.prototype, "aabb", {
+    get: function () {
+        // #if _DEBUG
+        console.error('DEPRECATED: pc.RenderComponent#aabb is deprecated. Use pc.RenderComponent#customAabb instead - which expects local space AABB instead of a world space AABB.');
+        // #endif
+        return null;
+    },
+    set: function (type) {
+        // #if _DEBUG
+        console.error('DEPRECATED: pc.RenderComponent#aabb is deprecated. Use pc.RenderComponent#customAabb instead - which expects local space AABB instead of a world space AABB.');
+        // #endif
+    }
+});
 
 Object.defineProperty(RigidBodyComponent.prototype, "bodyType", {
     get: function () {
--- a/src/framework/components/model/component.js
+++ b/src/framework/components/model/component.js
@@ -494,10 +494,12 @@ class ModelComponent extends Component {
         this._aabb = value;
 
         // set it on meshInstances
-        var mi = this._model.meshInstances;
-        if (mi) {
-            for (var i = 0; i < mi.length; i++) {
-                mi[i].setOverrideAabb(this._aabb);
+        if (this._model) {
+            var mi = this._model.meshInstances;
+            if (mi) {
+                for (var i = 0; i < mi.length; i++) {
+                    mi[i].setOverrideAabb(this._aabb);
+                }
             }
         }
     }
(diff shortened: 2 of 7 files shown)


willeastcott:
Very good point, @Maksims. If you use the `aabb` setter, this should tell the engine that you are overriding the auto-calculated bounding box, IMHO. I guess my only concern would be around what you'd need to do to switch back from a custom AABB to an auto-calculated one. In that case, maybe a separate boolean property would be the way to go. Or setting `aabb` to `null`.

But yeah, generally principle should be to avoid bloating the API if we already have a nice and obvious `aabb` property.

Maksims:
Regarding local/world, what happens internally - is ok. I'm more concerned about the public API:
`meshInstance.aabb` - world space AABB, very useful and I've used it often.
`model.aabb` - named same, but would be local space?
The issue is the hidden distinction between them, but the same variable name. It is not obvious from the API that they behave differently. I would expect to `model.aabb` - to be world space.
This also is similar to naming of functions: `getPosition` and `getLocalPosition` - **local** is explicit.

What could be done, is providing custom local AABB - if that is present, then AABB will be calculated by the engine using it. If it is not provided, then the engine will auto-calculate.
Also, it would be great to always have `model.aabb` available, and calculated by the engine automatically or based on provided `localAabb`.

`model.localAabb` - null or BoundingBox.

mvaligursky:
> What could be done, is providing custom local AABB - if that is present, then AABB will be calculated by the engine using it. If it is not provided, then the engine will auto-calculate.

This is exactly how it's done.

Regarding the naming .. we use **local** for position, but not for **aabb**. Example:
- mesh has aabb property, and that is in the local space
- meshInstance has aabb property, and that is in the world space

So I'd be happy to keep it called aabb.

Maksims:
> * mesh has aabb property, and that is in the local space

Mesh is independent of Scene Graph.

> * meshInstance has aabb property, and that is in the world space

Mesh Instance is always dependant on Scene Graph and belongs to an Entity.

ModelComponent/RenderComponent is always dependant on Scene Graph and belongs to an Entity.

That is why within context of Component of Mesh Instance, it is assumed to be world space, not local space.


### I propose a small change to this PR:
Instead of two new properties: `aabbCenter` and `aabbHalfExtents` to add one property `localAabb`.
And keep `aabb` world space, just like for mesh instances. As their main purpose is culling and picking, they are almost always used in world space.

This also will avoid any API breaking changes, only additions.

mvaligursky:
This PR does not introduce aabbCenter and aabbHalfExtents properties - these are only private storage space in the scene format for the aabb which is an existing property. It's done this way due to Editor not having a native support for BoundingBox in inspector, and so internally it uses two Vec3 members. These are only used to create a component, and never after.

Before this PR, I added aabb property on model and render component recently, and unfortunately an unintended bug crept in and this was stored in the world space. While testing it inside the Editor just now this was discovered, and was corrected to be in local space as intended.

We can rename .aabb to .localAabb (which would be inconsistent with the Mesh.aabb - which is a public interface and is used by the users, so I would prefer to not do it), but we won't be keeping both, at least not part of this PR.

Maksims:
> We can rename .aabb to .localAabb (which would be inconsistent with the Mesh.aabb - which is a public interface and is used by the users, so I would prefer to not do it), but we won't be keeping both, at least not part of this PR.

meshInstance.aabb - is a world space. And in `scene graph context` (same as render/model component) it makes sense.

Basically, here is a common user story:
Task: need to implement picking of an object.
Implementation (pseudo):
```js
for(meshInstance in modelComponent) {
    if (meshInstance.aabb.intersectsRay(ray)) {
        intersected();
        break;
    }
}
```
Sometimes it is like so:
```js
for(meshInstance in modelComponent) {
    if (first meshInstance) {
        aabb.copy(meshInstance.aabb);
    } else {
        aabb.add(meshInstance.aabb);
    }
}
if (aabb.intersectsRay(ray)) {
    intersected();
}
```

If model/render component AABB would be same as mesh instance, in world space, it would be easy to do:
```js
if (entity.render.aabb.intersectsRay(ray)) {
    intersected();
}
```
But if it is local, then need to do this:
```js
aabb.setFromTransformedAabb(entity.render.aabb, entity.getWorldTransform());
if (aabb.intersectsRay(ray) {
    intersected();
}
```

Basically, my question: if aabb is used for culling (which is world space), then why model/render component aabb should be local space? What is the motivation behind it?

mvaligursky:
The goal of this PR (and original PR which I linked in the description) is not to simplify picking of meshes by using bounding boxes. For this, a different API which is picking specific should be designed. For that case though I would probably transform the ray to local space and use local aabb instead of evaluating world space aabb. But that's a separate issue.

This PR is an optimisation for characters (and other meshes that modify vertex positions - either on CPU or in the vertex shader) to allow engine to avoid the expensive evaluation of local aabb for that mesh, by specifying oversized local aabb.

Before, for (even invisible) skinned characters, the engine would need to evaluate all bones, and then evaluate world space aabb of a mesh instance. This was expensive (see stats in the original PR). Now, when oversized local aabb is specified by the user, we don't evaluate bones nor the local space aabb based on those bones, and simply transform local aabb to world space for culling (which is exactly what noon-skinned meshes do). This cuts off lots of expensive computation for off screen characters - nor bones nor bone aabb is evaluated.

For this to work, the oversized aabb needs to be specified on local space .. so that the character can move (by moving entity) and engine can update its world space aabb for culling. If this was specified in world space, a user would additionally need to write a script to update this aabb as character moves. 

I'm open to renaming component.aabb to .localAabb, or even overrideAabb, but initial PR introduced .aabb already.

willeastcott:
OK, sounds reasonable @mvaligursky.

As for API/naming, what about:

* `aabb` - gettable world space AABB of model or render component [read only]
* `customAabb` - gettable/settable local space AABB of model or render component

Maksims:
Great reasoning, I thought that was the case.
### This is similar to:
mesh local aabb > mesh instance world aabb
### In our case:
component custom/local aabb > component world aabb

So engine can store local aabb or use custom, and re-calculate world aabb when transform or aabb is dirty.

mvaligursky:
great, I'll rename component.aabb to .customAabb, thanks for the discussion.

PR #3223, opened 2021-06-09
[FIX] Image element crash on assigning new layer composition to scene

willeastcott:
When the Editor disconnects and then reconnects, I suspect that the scene's layer composition is reassigned (maybe in `Application#_parseApplicationProperties`). This triggers the `onLayersChanged` handler for an image element in the scene. The code tries to access the model incorrectly. This PR specifies the correct property for the model.


mvaligursky (approved):
it seems to match other parts of the file, looks good.

PR #3230, opened 2021-06-11
Move app destroy to frame end

### LINKED ISSUE #3228: Calling app.destroy() in the middle of a frameUpdate causes crashes
Revisiting this PR: https://github.com/playcanvas/engine/pull/2703 based on this forum thread https://forum.playcanvas.com/t/after-loading-animation-from-glb-app-destroy-errors/16348/

What I didn't realise when I did that PR is the makeTick function checks for application.graphicsDevice is null at the start of the tick.

However, if the app.destroy() is called in a scriptType, we are pass that point and can cause crashes where system updates or the rest of the tick is trying to access something that is destroyed.

The workaround at the moment is to wrap the app.destroy in a setTimeout of 0ms so it does it at the start of the next frame

```
LoadGlb.prototype.update = function (dt) {
    if (this.app.keyboard.wasPressed(pc.KEY_1)) {
        setTimeout(function(){
            this.app.destroy();        
        }.bind(this), 0);
    }
};
```

The question I have is whether we should instead set a flag when app.destroy is called so the engine can actually destroy/cleanup at the start of the frame and stop updating?

eg
```
var makeTick = function (_app) {
    var application = _app;
    var frameRequest;

    return function (timestamp, frame) {
        if (application._destroy) {
            application.onDestroy();
            return;
        }
```

yaustar:
Fixes #3228

I'm really unsure about this as it is a kludge. I was thinking it might be better off as extra documentation for the function where app.destroy should not be called in an app.update.


mvaligursky:
It seems like a valid request to be able to shut down the application from within the update, and so I think the solution is fine.

src/framework/application.js
```diff
@@ -408,6 +408,8 @@ class Application extends EventHandler {
 
         app = this;
 
+        this._destroyThisFrame = false;
```
mvaligursky, line 411: I would perhaps rename this to `_destroyRequested`

src/framework/application.js
```diff
@@ -2162,6 +2170,7 @@ var makeTick = function (_app) {
         application._fillFrameStats();
         // #endif
 
+        this._inFrameUpdate = true;
```
mvaligursky, line 2173: have this variable in the constructor as well please, to not introduce variables to class during running.

yaustar:
Made changes from review and here's the test project: https://playcanvas.com/editor/scene/1067584

mvaligursky (approved):

Issue #3232, opened 2021-06-11
App destroy with a project using Anim Component causes error messages


yaustar:
Forum: https://forum.playcanvas.com/t/after-loading-animation-from-glb-app-destroy-errors/16348/14
Project: Press Q to quit: https://playcanvas.com/editor/scene/1175882

PR #3237, opened 2021-06-15
[AnimComponent] Refactor and API update

ellthompson:
This PR updates the anim component API to make it more user friendly for engine only users. It replicates the functionality of the animation component which we can now be deprecated.

### Updated functions

```javascript
     /**
     * @function
     * @name AnimComponent#assignAnimation
     * @description Associates an animation with a state in the loaded state graph. If all states are linked and the {@link AnimComponent#activate} value was set to true then the component will begin playing.
     * If no state graph is loaded, a default state graph will be created with a single state based on the provided nodeName parameter.
     * @param {string} nodeName - The name of the state node that this animation should be associated with.
     * @param {object} animTrack - The animation track that will be assigned to this state and played whenever this state is active.
     * @param {string} [layerName] - The name of the anim component layer to update. If omitted the default layer is used. If no state graph has been previously loaded this parameter is ignored.
     * @param {number} [speed] - Update the speed of the state you are assigning an animation to.
     * @param {boolean} [loop] - Update the loop property of the state you are assigning an animation to.
     */
     pc.AnimComponent#assignAnimation(nodeName, animTrack, layerName, speed, loop);

     /**
     * @function
     * @name AnimComponentLayer#assignAnimation
     * @description Assigns an animation track to a state in the current graph. If a state for the given nodeName doesn't exist, it will be created. If all states nodes are linked and the {@link AnimComponent#activate} value was set to true then the component will begin playing.
     * @param {string} nodeName - The name of the node that this animation should be associated with.
     * @param {object} animTrack - The animation track that will be assigned to this state and played whenever this state is active.
     * @param {number} [speed] - Update the speed of the state you are assigning an animation to.
     * @param {boolean} [loop] - Update the loop property of the state you are assigning an animation to.
     */
    pc.AnimComponentLayer#assignAnimation(nodeName, animTrack, speed, loop);
```

### New functions
```javascript
    /**
     * @name AnimComponent#addLayer
     * @returns {AnimComponentLayer} - The created anim component layer
     * @description Adds a new anim component layer to the anim component.
     * @param {string} layerName - The name of the layer to create.
     */
    addLayer(layerName);

    /**
     * @function
     * @name AnimComponentLayer#transition
     * @description Transition to any state in the current layers graph. Transitions can be instant or take an optional blend time.
     * @param {string} to - The state that this transition will transition to.
     * @param {number} [time=0] - The duration of the transition in seconds.
     * @param {number} [transitionOffset=null] - If provided, the destination state will begin playing its animation at this time. Given in normalised time, based on the states duration & must be between 0 and 1.
     */
    pc.AnimComponentLayer#transition(to, time = 0, transitionOffset = null);
```

### Usage

```javascript
this.entity.addComponent('anim', { activate: true });
this.entity.anim.assignAnimation('Idle', idleTrack);
this.entity.anim.assignAnimation('Wave', waveTrack);

setTimeout(() => {
    this.entity.anim.baseLayer.transition('Wave');
}, 2000);

setTimeout(() => {
    this.entity.anim.baseLayer.transition('Idle');
}, 5000);
```

### Other updates
- Fixes an issue when destroying an entity with an anim component causes the app to crash https://github.com/playcanvas/engine/issues/3232
- Fixes issues with cloning entities with anim components
- Refactors the anim component to the new style of component that doesn't rely on the data class
- Includes an engine example for animation events which makes use of the new assignAnimation API:

https://user-images.githubusercontent.com/1721533/122932333-8e732300-d365-11eb-8263-54eb3c8c8760.mov


mvaligursky:
A question on these two new parameters on 
`pc.AnimComponent#assignAnimation(nodeName, animTrack, layerName, speed, loop);`
- just double checking - those parameters won't override values on a shared class between instances, but only do it for this specific anim component instance? Meaning if I have two characters using the same state graph, only the one I set it to will change?

mvaligursky:
Also, please add example screenshot / link (release path)

src/anim/controller/anim-controller.js
```diff
@@ -217,6 +209,38 @@ class AnimController {
         return transitions;
     }
 
+    _transitionHasConditionsMet(transition) {
+        var conditionsMet = true;
+        var i;
```
mvaligursky, line 214: use let / const in the new code / convert existing when you can

mvaligursky (approved):
apart from a question, this looks good to me.

src/framework/components/anim/component-layer.js
```diff
@@ -60,18 +62,20 @@ class AnimComponentLayer {
     /**
      * @function
      * @name AnimComponentLayer#assignAnimation
-     * @description Associates an animation with a state node in the loaded state graph. If all states nodes are linked and the {@link AnimComponent#activate} value was set to true then the component will begin playing.
+     * @description Assigns an animation track to a state in the current graph. If a state for the given nodeName doesn't exist, it will be created. If all states nodes are linked and the {@link AnimComponent#activate} value was set to true then the component will begin playing.
      * @param {string} nodeName - The name of the node that this animation should be associated with.
      * @param {object} animTrack - The animation track that will be assigned to this state and played whenever this state is active.
+     * @param {number} [speed] - Update the speed of the state you are assigning an animation to.
+     * @param {boolean} [loop] - Update the loop property of the state you are assigning an animation to.
```
willeastcott, line 69: If these are optional, state what their defaults are.

src/framework/components/anim/component-layer.js
```diff
@@ -89,6 +93,23 @@ class AnimComponentLayer {
         }
     }
 
+    /**
+     * @function
+     * @name AnimComponentLayer#transition
+     * @description Transition to any state in the current layers graph. Transitions can be instant or take an optional blend time.
+     * @param {string} to - The state that this transition will transition to.
+     * @param {number} [time=0] - The duration of the transition in seconds.
+     * @param {number} [transitionOffset=null] - If provided, the destination state will begin playing its animation at this time. Given in normalised time, based on the states duration & must be between 0 and 1.
```
willeastcott, line 102: I have a funny feeling that our JSDoc output ignores the default parameter values. Might wanna check that. Generally, the standard across our JSDoc comments is to say something like `Defaults to 0.`.

willeastcott (approved):

src/framework/components/anim/component-layer.js
ellthompson: > A question on these two new parameters on
> `pc.AnimComponent#assignAnimation(nodeName, animTrack, layerName, speed, loop);`
> 
> * just double checking - those parameters won't override values on a shared class between instances, but only do it for this specific anim component instance? Meaning if I have two characters using the same state graph, only the one I set it to will change?

Yep that's right.

ellthompson:
> A question on these two new parameters on
> `pc.AnimComponent#assignAnimation(nodeName, animTrack, layerName, speed, loop);`
> 
> * just double checking - those parameters won't override values on a shared class between instances, but only do it for this specific anim component instance? Meaning if I have two characters using the same state graph, only the one I set it to will change?

Yep that's right.

src/anim/controller/anim-controller.js
```diff
@@ -94,8 +94,8 @@ class AnimController {
     }
 
     get playable() {
-        var playable = true;
-        var i;
+        let playable = true;
+        let i;
```
Maksims, line 98: If `let i` is defined outside of a for-loop, then it is in outside scope, so re-defining it will lead to an exception.
Please use this common within engine style
`for(let i = 0; ...`

Maksims: With let/const usage, variable hoisting is not a problem anymore, that's why, please don't define variables for loop counters in a header, but use local variables in loops.

ellthompson: Updated, thanks!

Maksims: Please, take a look at the whole PR, there are a lot of the same cases.

src/anim/controller/anim-transition.js
```diff
@@ -20,15 +18,14 @@ import {
  * @param {string} interruptionSource - Defines whether another transition can interrupt this one and which of the current or previous states transitions can do so. One of ANIM_INTERRUPTION_*.
  */
 class AnimTransition {
-    constructor(controller, from, to, time, priority, conditions, exitTime, transitionOffset, interruptionSource = ANIM_INTERRUPTION_NONE) {
-        this._controller = controller;
+    constructor({ from, to, time = 0, priority = 0, conditions = [], exitTime = null, transitionOffset = null, interruptionSource = ANIM_INTERRUPTION_NONE }) {
```
yaustar, line 21: As this is private, I'm less worried about this but the optional params defaults would need to be documented. Although saying that, would IDEs show the defaults?

src/framework/components/anim/system.js
```diff
@@ -35,16 +32,26 @@ class AnimComponentSystem extends ComponentSystem {
...
 
     initializeComponentData(component, data, properties) {
-        properties = ['activate', 'enabled', 'speed', 'playing'];
-        super.initializeComponentData(component, data, properties);
-        if (data.stateGraphAsset) {
-            component.stateGraphAsset = data.stateGraphAsset;
+        properties = ['activate', 'speed', 'playing'];
+        super.initializeComponentData(component, data, _schema);
+        Object.keys(data).forEach((key) => {
+            if (['animationAssets', 'stateGraph', 'layers'].includes(key)) return;
```
yaustar, line 38: Does the array here get allocated each loop? 

`['animationAssets', 'stateGraph', 'layers']`

yaustar (approved):
Minor comments

src/anim/controller/anim-transition.js
ellthompson: I wouldn't make this class public. Do IDE's show private jsdoc comments?

yaustar: They do when you start calling the function IIRC. Does depend on the IDE though. This is super minor though

yaustar: image

src/framework/components/anim/system.js
ellthompson: I tend to keep short lists in line as it makes them readable to me but I can lift it out if you have a concern.

yaustar: I think it be good to pull it out if it's allocated each iteration

src/anim/controller/anim-transition.js
ellthompson: Ok, I can add these in 👍

src/framework/components/anim/system.js
ellthompson: Done 👍

pavle-goloskokovic:
Are these docs removed intentionally or by mistake?

https://github.com/playcanvas/engine/blob/a6521aa4155723f7c06113d931324d0c570c7236/src/framework/components/anim/component.js#L387-L391

https://github.com/playcanvas/engine/blob/a6521aa4155723f7c06113d931324d0c570c7236/src/framework/components/anim/component.js#L441-L446

And is it possible to make `animationAssets` public so it appears in ts defs?

PR #3241, opened 2021-06-16
Exposed additional public functions for mesh morphing

mvaligursky:
- viewer uses these functions, and they are useful in general when accessing for example morph target names on a loaded glb.

**new public API**:
- MeshInstance.morphInstance
- MorphInstance.meshInstance
- MorphInstance.morph
- Morph.targets

**deprecated API**:
- Morph.getTarget(index) - this is replaced by new Morph.targets.

### DIFF
--- a/src/deprecated.js
+++ b/src/deprecated.js
@@ -507,7 +507,7 @@ export var scene = {
 };
 
 Morph.prototype.getTarget = function (index) {
-    return this._targets[index];
+    return this.targets[index];
 };
 
 GraphNode.prototype._dirtify = function (local) {
--- a/src/deprecated.js
+++ b/src/deprecated.js
@@ -507,6 +507,10 @@ export var scene = {
 };
 
 Morph.prototype.getTarget = function (index) {
+    // #if _DEBUG
+    console.warn('DEPRECATED: pc.Morph#getTarget is deprecated. Use pc.Morph#targets instead.');
+    // #endif
+
     return this.targets[index];
 };
 
--- a/src/deprecated.js
+++ b/src/deprecated.js
@@ -59,6 +59,7 @@ import { ForwardRenderer } from './scene/renderer/forward-renderer.js';
 import { GraphNode } from './scene/graph-node.js';
 import { Material } from './scene/materials/material.js';
 import { Mesh } from './scene/mesh.js';
+import { Morph } from './scene/morph.js';
 import { MeshInstance, Command } from './scene/mesh-instance.js';
 import { Model } from './scene/model.js';
 import { ParticleEmitter } from './scene/particle-system/particle-emitter.js';
@@ -503,6 +504,10 @@ export var scene = {
     Scene: Scene,
     Skin: Skin,
     SkinInstance: SkinInstance
+};
+
+Morph.prototype.getTarget = function (index) {
+    return this._targets[index];
 };
 
 GraphNode.prototype._dirtify = function (local) {
--- a/src/scene/morph.js
+++ b/src/scene/morph.js
@@ -222,6 +222,12 @@ class Morph extends RefCountedObject {
         this._targets.length = 0;
     }
 
+    /**
+     * @readonly
+     * @name Morph#targets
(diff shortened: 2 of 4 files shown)


src/scene/morph.js
```diff
@@ -221,6 +221,16 @@ class Morph extends RefCountedObject {
...
 
+    /**
+     * @readonly
+     * @name Morph#numTargets
+     * @type {number}
+     * @description Read-only property that returns number of targets of this morph. Access targets using {@link Morph#getTarget}.
+     */
+    get numTargets() {
+        return this._targets ? this._targets.length : 0;
+    }
```
willeastcott, line 232: We currently have `Morph#getTarget`. Any reason why we can't just have:

```javascript
get targets() {
    return this._targets;
}
```

Then, you could do:

```javascript
const numTargets = morph.targets.length;
```

mvaligursky: sounds good to me .. it changes the existing API but nobody would have used it without the 'count' property anyways .. so I've made the change. I'll update description as well.

willeastcott:
How about adding `getTarget` to `deprecated.js`? It's just `return this.targets[i];`

src/scene/morph.js
```diff
@@ -221,15 +222,8 @@ class Morph extends RefCountedObject {
...
-    /**
-     * @function
-     * @name Morph#getTarget
-     * @description Gets the morph target by index.
-     * @param {number} index - An index of morph target.
-     * @returns {MorphTarget} A morph target object.
-     */
-    getTarget(index) {
-        return this._targets[index];
+    get targets() {
```
willeastcott, line 225:
```suggestion
    /**
     * @readonly
     * @name Morph#targets
     * @type {MorphTarget[]}
     * @description The array of morph targets.
     */
    get targets() {
```

src/scene/morph.js
```diff
@@ -16,6 +16,7 @@ import { BUFFER_STATIC, TYPE_FLOAT32, SEMANTIC_ATTR15, ADDRESS_CLAMP_TO_EDGE, FI
  * @param {MorphTarget[]} targets - A list of morph targets.
  * @param {GraphicsDevice} graphicsDevice - The graphics device used to manage this morph target. If it is not provided, a device is obtained
  * from the {@link Application}.
+ * @property {MorphTarget} targets Gets an array of morph targets.
```
willeastcott, line 19: This line can be removed if you accept my JSDoc change for `targets`. I made this change because it's the only way to make `targets` as read only in the JSDoc output. See [`GraphNode#children`](https://developer.playcanvas.com/en/api/pc.GraphNode.html#children) for an example.

src/deprecated.js
```diff
@@ -505,6 +506,10 @@ export var scene = {
     SkinInstance: SkinInstance
 };
 
+Morph.prototype.getTarget = function (index) {
+    return this._targets[index];
```
willeastcott, line 510:
```suggestion
return this.targets[index];
```

willeastcott: So your code here is fine and it works. However, I've tweaked it so that the deprecated function implementation relies on the public API for `Morph` rather than directly access a private property. I've used this approach throughout `deprecated.js`.

src/deprecated.js
```diff
@@ -505,6 +506,10 @@ export var scene = {
     SkinInstance: SkinInstance
 };
 
+Morph.prototype.getTarget = function (index) {
```
willeastcott, line 509:
```suggestion
Morph.prototype.getTarget = function (index) {
    // #if _DEBUG
    console.warn('DEPRECATED: pc.Morph#getTarget is deprecated. Use pc.Morph#targets instead.');
    // #endif
```

src/scene/morph.js
mvaligursky: yep, I like this, done.

src/deprecated.js
mvaligursky: great, thanks for the changes!

willeastcott (approved):

Issue #3236, opened 2021-06-14
[.WebGL-0x47a8ea00]RENDER WARNING: there is no texture bound to the unit 0 - After Shadowmap refactor


yaustar:
Thread: https://forum.playcanvas.com/t/v1-42-gl-errors-in-console-v1-41-0-is-ok/20670/26
Project: https://playcanvas.com/editor/scene/1177100

Commit that introduced errors: https://github.com/playcanvas/engine/commit/5cd4bf6d69cd1dcc75ea4d6fafe275364a0ea9f1

Error log:
![image](https://user-images.githubusercontent.com/16639049/121906774-58112480-cd23-11eb-9591-8ada8688047d.png)

yaustar:
Looks like it is related to the position of the light to the camera. If the light is behind the camera, the error occurs. Seems to be fine when in front or within the frustum:

![image](https://user-images.githubusercontent.com/16639049/121918056-9a3f6380-cd2d-11eb-9c41-bc6c319124c0.png)

PR #3242, opened 2021-06-16
[Fix] Fixes warning for missing shadow map when light is not in the frustum

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/3236


yaustar (approved):
LGTM

PR #3245, opened 2021-06-17
findAvailableLocale refactor - Removed the static function from the API and added a member function

yaustar:
Fixes https://forum.playcanvas.com/t/playcanvas-api-function-not-working/20738

I can't think why this was public as static function as the only way to use it would be to access private member properties on i18n.

Added a member version of the function that checks against loaded translations.

### DIFF
--- a/src/i18n/i18n.js
+++ b/src/i18n/i18n.js
@@ -42,17 +42,37 @@ class I18n extends EventHandler {
     }
 
     /**
+     * @private
      * @static
      * @function
      * @name I18n.findAvailableLocale
      * @description Returns the first available locale based on the desired locale specified. First
      * tries to find the desired locale and then tries to find an alternative locale based on the language.
-     * @param {string} desiredLocale - The desired locale e.g. En-US.
+     * @param {string} desiredLocale - The desired locale e.g. en-US.
      * @param {object} availableLocales - A dictionary where each key is an available locale.
      * @returns {string} The locale found or if no locale is available returns the default en-US locale.
+     * @example
+     * // With a defined dictionary of locales
+     * var availableLocales = { en: 'en-US', fr: 'fr-FR' };
+     * var locale = pc.I18n.getText('en-US', availableLocales);
+     * // returns 'en'
      */
     static findAvailableLocale(desiredLocale, availableLocales) {
         return findAvailableLocale(desiredLocale, availableLocales);
+    }
+
+    /**
+     * @function
+     * @name I18n#findAvailableLocale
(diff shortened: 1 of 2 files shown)


src/i18n/i18n.js
```diff
@@ -42,19 +42,39 @@ class I18n extends EventHandler {
...
+     * @name I18n#findAvailableLocale
+     * @description Returns the first available locale based on the desired locale specified. First
+     * tries to find the desired locale and then tries to find an alternative locale based on the language.
+     * @param {string} desiredLocale - The desired locale e.g. en-US.
+     * @returns {string} The locale found or if no locale is available returns the default en-US locale.
+     * @example
+     * var locale = this.app.i18n.getText('en-US');
+     */
+    findAvailableLocale(desiredLocale) {
+        return findAvailableLocale(desiredLocale, this._translations);
```
mvaligursky, line 76: so when you call `findAvailableLocale` which function ends up getting executed? I think you should call it as 
`I18n.findAvailableLocale` as that's how you call a static function.

mvaligursky: ah, ignore me .. the module exports global function of this name and that's what is getting called.

mvaligursky (approved):

yaustar:
@vkalpias Can you have a quick look over this please? It looks like there is similar logic between https://github.com/playcanvas/engine/blob/master/src/i18n/utils.js#L40 and https://github.com/playcanvas/engine/blob/master/src/i18n/i18n.js#L357 that they should be merged in some way

vkalpias:
You should add some tests for this in `tests/i18n/test_i18n.js`.

yaustar:
@vkalpias Added tests :)

tests/i18n/test_i18n.js
```diff
@@ -772,4 +772,24 @@ describe('I18n tests', function () {
 
         expect(asset.getLocalizedAssetId('zh-SG')).to.equal(2);
     });
+
+    it('findAvailableLocale() should find locale if translations have been provided for it', function () {
+        addText('no-IT', 'key', 'norwegian');
+        expect(app.i18n.findAvailableLocale('no-IT')).to.equal('no-IT');
+    });
+
+    it('findAvailableLocale() should fallback to en-US if translations have not been provide for the desired locale', function () {
```
vkalpias, line 781:
```suggestion
    it('findAvailableLocale() should fallback to en-US if translations have not been provided for the desired locale', function () {
```

vkalpias (approved):

Issue #3252, opened 2021-06-21
Implement support for physically based light intensity units


mvaligursky:
The engine is currently not using physically based units to specify light intensity, which would allow an easy way to realistically set up lights in the scene.

some references:
- filament (lighting section) https://google.github.io/filament/Filament.md.html
- frostbite, light units section https://seblagarde.files.wordpress.com/2015/07/course_notes_moving_frostbite_to_pbr_v32.pdf
- unreal engine https://docs.unrealengine.com/4.26/en-US/BuildingWorlds/LightingAndShadows/PhysicalLightUnits/
- unity engine https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@6.9/manual/Physical-Light-Units.html

This would be useful when implementing support for punctual lights https://github.com/playcanvas/engine/issues/2061 which specifies light intensities in physical units.

Note that Blender does not correctly export light intensities into gltf format, as it exports them in watts instead of expected lumens. https://github.com/KhronosGroup/glTF-Blender-IO/issues/564

MAG-AdrianMeredith:
Bump!  It might not be physically based but does anyone know what the units currently are?

slimbuck:
Is this something @GSterbrant might want to investigate?

GSterbrant:
It's actually a bit complicated, since it's different for different types of lights. Filament describes it very well here: https://google.github.io/filament/Filament.md.html#lighting/units

Of course this puts even more emphasis on the importance of it. Using 1 for the sun intensity doesn't really translate to using 1 as intensity for a spotlight.

MAG-AdrianMeredith:
This might help https://docs.google.com/spreadsheets/d/1Ce9XCC2Ub9eVjAQdYbYdLvlkDBJ_0GtAyRbSF0KJWzE/edit#gid=0

GSterbrant:
We'd probably a flag like that too, to not break old projects...

mvaligursky:
closing as its fixed in https://github.com/playcanvas/engine/pull/4618

PR #3253, opened 2021-06-21
Support for KHR_lights_punctual extension

mvaligursky:
- Added support for KHR_lights_punctual extension https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_lights_punctual
- Note of the limitation: https://github.com/playcanvas/engine/issues/3252 causing light intensities to be clamped to 0..1 range
- Note that lights are supported only when creating hierarchy with render components using **instantiateRenderEntity**, but not when using model components using **instantiateModelEntity**
- The instance of the model is no longer created when a glb file is loaded, and is created and cached on the fly only when needed
- render component hierarchy is being created from parsed glb file directly with no dependency on the Model being created (and so if user uses Renders, the Model does not get created at all)
- moved SkinInstanceCache to separate file to share it between render component and glb creating one.
- moved GlbParser.CreateModel to ContainerResource.

Fixes https://github.com/playcanvas/engine/issues/2061

**Example** on how to create lights from loaded glb container (the same as creating hierarchy with renders):
`const lightsEntity = assets.lights.resource.instantiateRenderEntity();`

lights can be processed this way:
```
const lights = lightsEntity.findComponents("light");
lights.forEach((light) => {
    light.enabled = false;
});
```

Thanks to @godiagonal for initial implementation here: https://github.com/playcanvas/engine/pull/2385

### DIFF
--- a/src/scene/skin-instance-cache.js
+++ b/src/scene/skin-instance-cache.js
@@ -25,7 +25,7 @@ class SkinInstanceCache {
         SkinInstanceCache._skinInstanceCache.forEach(function (array, rootBone) {
             console.log(rootBone.name + ': Array(' + array.length + ")");
             for (let i = 0; i < array.length; i++) {
-                console.log("  " + i + ": RefCount " + array[i].getRefCount());
+                console.log(`  ${i}: RefCount ${array[i].getRefCount()}`);
             }
         });
     }
--- a/src/scene/skin-instance-cache.js
+++ b/src/scene/skin-instance-cache.js
@@ -1,0 +1,136 @@
+import { RefCountedObject } from '../core/ref-counted-object.js';
+import { SkinInstance } from './skin-instance.js';
+
+// Class used as an entry in the ref-counted skin instance cache
+class SkinInstanceCachedObject extends RefCountedObject {
+    constructor(skin, skinInstance) {
+        super();
+        this.skin = skin;
+        this.skinInstance = skinInstance;
+    }
+}
+
+// Pure static class, implementing the cache of skin instances used by render component.
+class SkinInstanceCache {
+    // map of SkinInstances allowing those to be shared between
+    // (specifically a single glb with multiple render components)
+    // It maps a rootBone to an array of SkinInstanceCachedObject
+    // this allows us to find if a skin instance already exists for a rootbone, and a specific skin
+    static _skinInstanceCache = new Map();
+
+    // #if _DEBUG
+    // function that logs out the state of the skin instances cache
+    static logCachedSkinInstances() {
+        console.log("CachedSkinInstances");
+        SkinInstanceCache._skinInstanceCache.forEach(function (array, rootBone) {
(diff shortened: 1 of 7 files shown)


src/scene/skin-instance-cache.js
```diff
@@ -0,0 +1,136 @@
...
+    // It maps a rootBone to an array of SkinInstanceCachedObject
+    // this allows us to find if a skin instance already exists for a rootbone, and a specific skin
+    static _skinInstanceCache = new Map();
+
+    // #if _DEBUG
+    // function that logs out the state of the skin instances cache
+    static logCachedSkinInstances() {
+        console.log("CachedSkinInstances");
+        SkinInstanceCache._skinInstanceCache.forEach(function (array, rootBone) {
+            console.log(rootBone.name + ': Array(' + array.length + ")");
```
willeastcott, line 26:
```suggestion
            console.log(`${rootBone.name}: Array(${array.length})`);
````

File: src/scene/skin-instance-cache.js [hunk 2]
```diff
@@ -0,0 +1,136 @@
...
+            for (let i = 0; i < array.length; i++) {
+                console.log("  " + i + ": RefCount " + array[i].getRefCount());
```
willeastcott, line 28:
```suggestion
                console.log(` ${i}: RefCount ${array[i].getRefCount()}`);
```

src/scene/skin-instance.js
```diff
@@ -76,6 +76,29 @@ class SkinInstance {
         this._rootBone = rootBone;
     }
 
+    // resolved skin bones to a hierarchy with the rootBone at its root.
+    // entity parameter specifies the entity used if the bone match is not found in the hierarhy - usually the entity the render component is attached to
```
willeastcott, line 80:
```suggestion
    // entity parameter specifies the entity used if the bone match is not found in the hierarchy - usually the entity the render component is attached to
```

src/scene/skin-instance.js
```diff
@@ -76,6 +76,29 @@ class SkinInstance {
...
+
+        // Resolve bone IDs to actual graph nodes of the hierarchy
+        const skin = this.skin;
+        const  bones = [];
+        for (let j = 0; j < skin.boneNames.length; j++) {
+            var boneName = skin.boneNames[j];
+            var bone = rootBone.findByName(boneName);
+
+            if (!bone) {
+                console.error("Failed to find bone [" + boneName + "] in the entity hierarchy, RenderComponent on " + entity.name + ", rootBone: " + rootBone.entity.name);
```
willeastcott, line 93:
```suggestion
#if _DEBUG
                console.error(`Failed to find bone [${boneName}] in the entity hierarchy, RenderComponent on ${entity.name}, rootBone: ${rootBone.entity.name}`);
#endif
```

willeastcott (approved):

slimbuck (approved):

Issue #1727, opened 2019-10-31
Animation Blending ~ Multiple Animations


ChristianTucker:
It's really a shame to not see this in the engine as it's been an issue that people have struggled with for years while doing research on the forum. I'm currently working on a game in which the player needs to be able to move and attack at the same time. Unfortunately, this means that my only options are to 

1: Get animations that walk while attacking.
2: Allow my character to slide around during the attacking animation.

I also have animations where player's can play instruments while walking around, but it's the same story, they just slide about the ground as you can't play multiple animations at once.

This causes for a really, really clunky experience inside of our games, as my options are to lock movement while an attack is happening, or have a visual "bug".

FeaturedSpace:
@ChristianTucker As you probably found my forum posts on the subject, there was a solution script that was found, which I will link here. It basically creates a separate skeleton for upper-body animations and lower body animations in-engine. This would help solve your problem.

[Here's the script](https://github.com/FeaturedSpace/playcanvas-animator/blob/master/animatorControl.js)

I've been working on maintaining it and making it user-friendly so it doesn't require too much knowledge of the inner workings of the engine in order to operate. 

Let me know if you have any issues.

cobaki:
The script works great, but the upper body animations are not blending. Any idea how I could achieve this?

FeaturedSpace:
You'd have to customize the script to fit your needs. Luckily, I have done this before and you can check out my project, MAGE, and take a look at how I adapted this script.

https://playcanvas.com/project/573916/overview/mage

cobaki:
Great project, thanks. Not sure if I'm missing something, but I don't see the upper body animation (for example reloading) to blend onto the base animation (idle/walking). When i hit "reload" it jumps to the first frame, plays and then jumps back to the underlying animation.

FeaturedSpace:
Are you talking about MAGE or your own project?

I assume your own. Again, I'm not quite sure if you've mapped the bones correctly. Make sure all of the bones in the upper body are in the upper body skeleton array. And vice versa in the lower body skeleton array.

Again, I'm here to help if you can throw some code out or share the project with me (if it's private) and I can take a closer look.

Hopefully when the new animation system comes in, per bone(s) animations are implemented. I might make a pull-request adding this into the AnimationComponent if I had help from someone else more knowledgeable about the internals of the engine.

slimbuck:
Hi @cobaki, @FeaturedSpace,

[...]

PR #3264, opened 2021-06-24
[AnimComponent] Layer masks

ellthompson:
Fixes https://github.com/playcanvas/engine/issues/1727


src/anim/binder/default-anim-binder.js
```diff
@@ -128,6 +130,30 @@ class DefaultAnimBinder {
     }
 
     findNode(path) {
+        if (this._mask) {
```
mvaligursky, line 133: use let / const please ;)

src/anim/evaluator/anim-evaluator.js
```diff
@@ -1,3 +1,130 @@
...
+        this._layerCounter = value;
+    }
+
+    get value() {
+        return this._value;
+    }
+
+    set value(value) {
+        this._value = value;
+    }
```
mvaligursky, line 70: there's no need for these setters and getters, as they don't do anything extra to just having a property with the same name.

src/anim/evaluator/anim-evaluator.js
```diff
@@ -1,3 +1,130 @@
+import { Quat } from '../../math/quat.js';
+import { Vec3 } from '../../math/vec3.js';
+import { ANIM_LAYER_OVERWRITE } from '../controller/constants.js';
+
+class AnimTarget {
```
mvaligursky, line 5: just curious - why is AnimTarget class in anim-evaluator.js? I see in current version we already have anim-target.js in the same folder as well. Is it another AnimTarget class? Maybe pick a different name in that case.

src/anim/evaluator/anim-evaluator.js
```diff
@@ -1,3 +1,130 @@
...
+        }
+    }
+
+    get valueData() {
+        switch (this._valueType) {
+            case (AnimTarget.TYPE_QUAT): {
+                return [this._value.x, this._value.y, this._value.z, this._value.w];
+            }
+            case (AnimTarget.TYPE_VEC3): {
+                return this._value.data;
```
mvaligursky, line 117: do you access Vec3.data here? That'd log warnings on each access.

src/anim/evaluator/anim-evaluator.js
```diff
@@ -1,3 +1,130 @@
...
+                this._value.add(this._currentValue.mulScalar(this.weight(index)));
+                break;
+            }
+        }
+    }
+
+    get valueData() {
+        switch (this._valueType) {
+            case (AnimTarget.TYPE_QUAT): {
+                return [this._value.x, this._value.y, this._value.z, this._value.w];
```
mvaligursky, line 114: any way to avoid array creation each time?

src/anim/evaluator/anim-evaluator.js
```diff
@@ -149,6 +276,19 @@ class AnimEvaluator {
                     }
 
                     targets[resolved.targetPath] = target;
+                    if (this._binder.animComponent) {
+                        if (!this._binder.animComponent.targets[resolved.targetPath]) {
+                            let type;
+                            if (resolved.targetPath.indexOf('localRotation') === -1) {
```
mvaligursky, line 282: does the targetPath store the entity names path as well? If yes, having an entity called "localRotation" would not work well with this.

src/anim/evaluator/anim-evaluator.js
```diff
@@ -316,7 +459,19 @@ class AnimEvaluator {
         for (var path in targets) {
             if (targets.hasOwnProperty(path)) {
                 var target = targets[path];
-                target.target.func(target.value);
+                if (this._binder.animComponent) {
```
mvaligursky, line 462: I would not mind some comments in this block

src/framework/components/anim/component-binder.js
```diff
@@ -15,9 +15,20 @@ const c  = new Color();
 const q  = new Quat();
 
 class AnimComponentBinder extends DefaultAnimBinder {
-    constructor(animComponent, graph) {
+    constructor(animComponent, graph, layerName, mask, index) {
         super(graph);
         this.animComponent = animComponent;
+        this._mask = mask;
+        this._layerName = layerName;
+        this._index = index;
```
mvaligursky, line 23: either add comment or makes less generic variable than "index"

src/framework/components/anim/component-layer.js
```diff
@@ -10,12 +11,16 @@ import { AnimTransition } from '../../../anim/controller/anim-transition.js';
...
  * @param {AnimComponent} component - The component that this layer is a member of.
+ * @param {number} weight - The weight of this layer. Defaults to 1.
  */
 class AnimComponentLayer {
-    constructor(name, controller, component) {
+    constructor(name, controller, component, index, weight = 1, blendType = ANIM_LAYER_OVERWRITE) {
         this._name = name;
         this._controller = controller;
         this._component = component;
+        this._index = index;
```
mvaligursky, line 21: again, please use less generic name .. layerIndex or something.

mvaligursky: also, I'm not sure it's being used?

src/framework/components/anim/component.js
```diff
@@ -6,12 +6,14 @@ import { AnimController } from '../../../anim/controller/anim-controller.js';
...
 
 import {
-    ANIM_PARAMETER_BOOLEAN, ANIM_PARAMETER_FLOAT, ANIM_PARAMETER_INTEGER, ANIM_PARAMETER_TRIGGER, ANIM_CONTROL_STATES
+    ANIM_PARAMETER_BOOLEAN, ANIM_PARAMETER_FLOAT, ANIM_PARAMETER_INTEGER, ANIM_PARAMETER_TRIGGER, ANIM_CONTROL_STATES, ANIM_LAYER_OVERWRITE
 } from '../../../anim/controller/constants.js';
 import { AnimComponentBinder } from './component-binder.js';
 import { AnimComponentLayer } from './component-layer.js';
 import { AnimStateGraph } from '../../../anim/state-graph/anim-state-graph.js';
 import { AnimEvents } from '../../../anim/evaluator/anim-events.js';
+import { Quat } from '../../../math/quat.js';
```
mvaligursky, line 15: remove these unused imports

src/framework/components/anim/component-layer.js
```diff
@@ -10,12 +11,16 @@ import { AnimTransition } from '../../../anim/controller/anim-transition.js';
  * @param {string} name - The name of the layer.
  * @param {object} controller - The controller to manage this layers animations.
  * @param {AnimComponent} component - The component that this layer is a member of.
+ * @param {number} weight - The weight of this layer. Defaults to 1.
  */
 class AnimComponentLayer {
-    constructor(name, controller, component) {
+    constructor(name, controller, component, index, weight = 1, blendType = ANIM_LAYER_OVERWRITE) {
```
mvaligursky, line 17: It's ok to take blendType as constructor parameter, but we need a setter for it for the case the user needs to change it later.

ellthompson: Changing the blendType at runtime would require updating the layer mask contained in each AnimTargetValue as its based on the unique combination of blendTypes for all layers associated with that AnimTargetValue. I'm not sure this is worth adding in at this point.

mvaligursky: fair enough

src/anim/binder/default-anim-binder.js
```diff
@@ -131,38 +131,40 @@ class DefaultAnimBinder {
...
-                    }
-                    break;
-                }
-                if (maskItem.children) {
-                    if (entityPath.indexOf(maskItem.path) !== -1) {
-                        shouldAnimateNode = true;
-                        break;
-                    }
-                } else if (maskItem.path === entityPath) {
+            let shouldAnimateNode = false;
```
mvaligursky, line 134: this whole code block inside `if (this._mask)` does not seem very efficient. It feels like some pre-computation should be done to make it a lot less intensive operation. Especially those joins and slices.

src/anim/evaluator/anim-evaluator.js
```diff
@@ -456,18 +335,27 @@ class AnimEvaluator {
 
         // apply result to anim targets
         var targets = this._targets;
+        const targetValue = new Array(4);
         for (var path in targets) {
             if (targets.hasOwnProperty(path)) {
                 var target = targets[path];
-                if (this._binder.animComponent) {
+                // if this evaluator is associated with an anim component then we should blend the result of this evaluator with all other anim layer's evaluators
+                if (this._binder.animComponent && target.target.isTransform) {
```
mvaligursky, line 343: you could have a local variable binder as you access this multiple times
`this._binder`

src/anim/evaluator/anim-target-value.js
```diff
@@ -0,0 +1,69 @@
+import { Quat } from '../../math/quat.js';
+import { Vec3 } from '../../math/vec3.js';
+import { ANIM_LAYER_OVERWRITE } from '../controller/constants.js';
+
+class AnimTargetValue {
```
mvaligursky, line 5: please add a comment about the purpose of this class

src/anim/evaluator/anim-target-value.js
```diff
@@ -0,0 +1,69 @@
...
+        if (this.valueType === AnimTargetValue.TYPE_QUAT) {
+            this.value = new Quat();
+            this._currentValue = new Quat();
+        } else {
+            this.value = new Vec3();
+            this._currentValue = new Vec3();
+        }
+    }
+
+    weight(index) {
```
mvaligursky, line 24: you should rename it to     `getWeight(index)`

src/anim/evaluator/anim-target-value.js
```diff
@@ -0,0 +1,69 @@
...
+        if (this._component.dirtyWeights) this.updateWeights();
```
mvaligursky, line 25: is dirtyWeights getting set to false anywhere? I don't see it in the component .. but maybe elsewhere it does. Just making sure.

src/anim/evaluator/anim-target.js
```diff
@@ -41,6 +41,12 @@ class AnimTarget {
     get targetPath() {
         return this._targetPath;
     }
+
+    get isTransform() {
+        return (this._targetPath.substring(this._targetPath.length - 13) === 'localRotation') ||
+        (this._targetPath.substring(this._targetPath.length - 13) === 'localPosition') ||
+        (this._targetPath.substring(this._targetPath.length - 10) === 'localScale');
+    }
```
mvaligursky, line 49: you should evaluate this in the constructor and store it in this.isTransform property, instead of possibly executing it many times.

mvaligursky (approved):

PR #3271, opened 2021-06-28
Clustered lighting does not use lights with intensity of zero

mvaligursky:
- small optimization to avoid having to enable / disable light which has higher cost. Intensity can instead be set to 0.

### DIFF
--- a/src/scene/world-clusters.js
+++ b/src/scene/world-clusters.js
@@ -485,7 +485,7 @@ class WorldClusters {
 
             // use enabled and visible lights
             const light = lights[i];
-            if (light.enabled && light.type !== LIGHTTYPE_DIRECTIONAL && light.visibleThisFrame) {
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

Maksims:
To match the existing engine pattern, property such as `visible` could be added (same as of meshInstance).

mvaligursky:
Hi @Maksims - thanks, I considered exactly the same thing, but decided against it for now:
- the change I did is valid optimisation anyways for lights with zero intensity
- in the future where clustered lights are the solution in the engine, disabling the light should be cheap, so no need for another interface
- if I added it now, only clustered lights would take advantage of it .. but not existing lighting solution .. and that would be inconsistent. I could implement changes for current solution (maybe internally set intensity of 0), but that's code that hopefully will be deprecated in the future.

PR #3176, opened 2021-05-20
Post Process God Rays

leonidaspir:
It introduces a post effect that can simulate light shafts (god rays) from a world point. The screen position is derived from the position of the referenced entity (it doesn't require a light component). Multiple instances of the effect can be added. The Post Effects example has been updated.

image

I confirm I have read the [contributing guidelines](https://github.com/playcanvas/engine/blob/master/.github/CONTRIBUTING.md) and signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #3272, opened 2021-06-28
Godrays post-effect

mvaligursky:
Introduced in https://github.com/playcanvas/engine/pull/3176 by @leonidaspir - thank you!

This PR uses the effects from the original PR, and integrates it to the new engine examples to demonstrate its use.

Note that it does not seem to work in WebGl1 - this needs further investigation.

Screenshot 2021-06-25 at 14 37 31


leonidaspir:
By the way, if you reduce the size of the render target, you can get for free (and for better performance) blurred rays:

```
        var colorBuffer = new pc.Texture(graphicsDevice, {
            format: pc.PIXELFORMAT_R8_G8_B8_A8,
            width: width / 4,
            height: height / 4,
            mipmaps: false
        });
```
It may worth to expose that as an attribute.

image

mvaligursky:
> It may worth to expose that as an attribute.

Good idea .. feel free to PR it at some point.

PR #3275, opened 2021-06-29
[Examples Browser] Debug mode

ellthompson:
This PR implements a debug mode for each example. These can be accessed via the `/#/debug/misc/hello-world`. In this mode an example is executed directly from the example classes `example` function rather than through a call to `eval` which occurs when using the normal app. This means debugging can provide a full stack trace with correct line numbers. However, debug mode doesn't support in browser editing of examples.

A directory of all debug example links can be found here at `/debug-directory`.

There are currently a few issues that have cropped up in debug mode which will need to be ironed out. However these affect only a small portion of the examples:
- Tween example doesn't animate (not to do with external script loading as the draco example is fine).
- Events example doesn't animate
- Model asset example doesn't load the statue

Note that this PR shouldn't impact the normal examples browser single page app.


mvaligursky (approved):
Looks good. Can you please update the readme as well in how to use it.

PR #3277, opened 2021-06-30
Basis worker fixes

### LINKED ISSUE #3191: Basis Compression: allow user to prioritize VRAM over quality
Currently with Basis the "most desirable gpu texture compression format given the device's capabilities" is defined in the following function.

[https://github.com/playcanvas/engine/blob/master/src/resources/basis.js#L268](https://github.com/playcanvas/engine/blob/master/src/resources/basis.js#L268)

However, there are cases where a developer needs to choose the format that consumes the least VRAM rather than the format with the best quality.  

It would be most flexible if the user could choose "per texture",  via the texture inspector basis settings, whether to prioritize VRAM or quality.

Screenshot 2021-05-30 at 15 27 03

Here is the forum thread

[https://forum.playcanvas.com/t/basis-compression-ios-android-issue/20317](https://forum.playcanvas.com/t/basis-compression-ios-android-issue/20317)

slimbuck:
Fixes #3191

This PR:
- simplifies the basis API and fixes a number of issues around non-POT textures
- uses more sophisticated logic for choosing runtime texture compression format and makes this user-configurable.
- adds support for multiple transcode workers with the option of having the workers be 'eager' or not (but defaults to 1 worker).
- transcodes to uncompressed data if the dimensions are invalid for the current platform

The following API functions were removed:
- `basisDownload`
- `basisSetDownloadConfig`
- `basisDownloadFromConfig`
- `basisTargetFormat`

Instead, there is a single initialize function, which is now documented as public API:
- `basisInitialize`

The call to perform transcode remains unchanged (this is internal, not public API):
- `basisTranscode`

In order to customise the texture formats chosen at runtime, supply `config.rgbPriority` and `config.rgbaPriority` to `basisInitialize`. The default priorities are as follows:
``` javascript
const defaultRgbPriority = ['etc1', 'etc2', 'astc', 'dxt', 'pvr', 'atc'];
const defaultRgbaPriority = ['astc', 'dxt', 'etc2', 'pvr', 'atc'];
```
Note however that before consulting these priorities, the transcoder will first attempt to match with the basis file compression format. For example, if the basis file is stored in etc1 and the graphics device supports etc1, then this will be chosen (similarly with astc).

A followup PR will address the remaining issue of non-POT textures on webgl1 devices.


mvaligursky:
> uses more sophisticated logic for choosing runtime texture compression format and makes this user-configurable

Could you please add details on how is this configurable to the description.

slimbuck:
A few additional notes on this PR:
- the worker source has been split into its own file.
- instead of concatenating worker source with basis source to construct the web worker, we download the basis script as a blob and load it in the worker using loadScript. This should result in less main thread overhead at startup.
- the graphics device is now defaulted to 1 byte unpack alignment
- an eager worker always requests more work, whereas a non-eager worker only requests a new job once the current job is finished. a set of non-eager workers will have better work balance over multiple workers compared to eager, but there is wasted time between the end of one job and start of the next.

slimbuck:
> > uses more sophisticated logic for choosing runtime texture compression format and makes this user-configurable
> 
> Could you please add details on how is this configurable to the description.

Done

src/resources/basis.js
```diff
@@ -1,518 +1,311 @@
...
-    };
-
-    // Initialize the web worker's BASIS interface.
-    // basisModule is the optional wasm binary.
-    const workerInit = function (basisModule) {
-        const instantiateWasmFunc = function (imports, successCallback) {
-            WebAssembly.instantiate(basisModule, imports)
-                .then(function (result) {
-                    successCallback(result);
+        const fetchPromise = fetch(config.wasmUrl);
```
willeastcott, line 68: Just wanted to flag that the updated code will call `fetch` if WASM is supported whereas before it was just if compile streaming is supported. I think it's OK though since `fetch` _should_ be supported everywhere where WASM is supported.

src/resources/basis.js
```diff
@@ -1,518 +1,311 @@
...
+            WebAssembly.compileStreaming(fetchPromise)
+                .then((module_) => {
+                    if (basisCode) {
+                        sendResponse(basisCode, module_);
+                    } else {
+                        module = module_;
+                    }
+                })
+                .catch((err) => {
+                    console.warn('compileStreaming() failed for ' + config.wasmUrl + '(' + err + '), falling back to arraybuffer download.');
```
willeastcott, line 97: Wrap in ```// #if _DEBUG```?

src/resources/basis.js
```diff
@@ -1,518 +1,311 @@
...
-                }
-            });
+// global state
+const queue = new BasisQueue();
+let initializing = false;
+
+/**
+ * @name basisInitialize
+ * @function
+ * @description Initialize the basis transcode worker.
```
willeastcott, line 229:
```suggestion
 * @description Initialize the Basis transcode worker.
```

src/resources/basis.js
```diff
@@ -1,518 +1,311 @@
...
+ * @param {object} [config] - The basis configuration.
```
willeastcott, line 230:
```suggestion
 * @param {object} [config] - The Basis configuration.
```

src/resources/basis.js
```diff
@@ -1,518 +1,311 @@
...
+ * @param {string} [config.glueUrl] - URL of glue script.
+ * @param {string} [config.wasmUrl] - URL of the wasm module.
+ * @param {string} [config.fallbackUrl] - URL of the fallback script to use when wasm modules aren't supported.
+ * @param {number} [config.numWorkers] - Number of transcode workers to create (default is 1).
+ * @param {boolean} [config.eagerWorkers] - True if workers should be eager (enabled when numWorkers is 1).
+ * @param {string[]} [config.rgbPriority] - Array of texture compression formats in priority order for textures without alpha.
+ * The supported compressed formats are: 'astc', 'atc', 'dxt', 'etc1', 'etc2', 'pvr'.
+ * @param {string[]} [config.rgbaPriority] - Array of texture compression formats in priority order for textures with alpha.Math
```
willeastcott, line 238:
```suggestion
* @param {string[]} [config.rgbaPriority] - Array of texture compression formats in priority order for textures with alpha.
```

src/resources/basis.js
willeastcott:
```suggestion
                    console.warn(`compileStreaming() failed for ${config.wasmUrl} (${err}), falling back to arraybuffer download.`);
```

willeastcott (approved):

src/resources/basis.js
```diff
@@ -1,518 +1,320 @@
...
+ * @description Initialize the Basis transcode worker.
+ * @param {object} [config] - The Basis configuration.
+ * @param {string} [config.glueUrl] - URL of glue script.
+ * @param {string} [config.wasmUrl] - URL of the wasm module.
+ * @param {string} [config.fallbackUrl] - URL of the fallback script to use when wasm modules aren't supported.
+ * @param {number} [config.numWorkers] - Number of workers to use for transcoding (default is 1). While it is
+ * possible to improve transcode performance using multiple workers, this will likely depend on the runtime
+ * platform. For example, desktop will likely benefit from more workers compared to mobile. Also
+ * keep in mind that it takes time to initialize workers and increasing this value could impact application
+ * startup time. Make sure to test your application performance on all target platforms when changing this parameter.
```
willeastcott, line 240: Is there a maximum number we allow here? I can see from the code we don't clamp and upper bound. Should we?

slimbuck: I'm not aware of any limit imposed by the underlying system and don't particularly see a risk leaving it unbounded, but I am also happy to place an upper limit if you think it prudent. What number would you suggest?

willeastcott: I'm not sure how I feel about a limit. But if I did impose one, maybe 16?

slimbuck: Added limit of 16.

mvaligursky (approved):

Kulodo:
Hi guys


> In order to customise the texture formats chosen at runtime, supply `config.rgbPriority` and `config.rgbaPriority` to `basisInitialize`. The default priorities are as follows:
> 
> ```js
> const defaultRgbPriority = ['etc1', 'etc2', 'astc', 'dxt', 'pvr', 'atc'];
> const defaultRgbaPriority = ['astc', 'dxt', 'etc2', 'pvr', 'atc'];
> ```
> 

Can I just check that basisInitialize() is called once for whole project?
Where in the app startup process should I call basisInitialize() if I need to?

> Note however that before consulting these priorities, the transcoder will first attempt to match with the basis file compression format. For example, if the basis file is stored in etc1 and the graphics device supports etc1, then this will be chosen (similarly with astc).

Does this refer to the Basis Mode ( ETC/ASTC ) setting in the editor texture inspector?

slimbuck:
Hi @Kulodo,

> Can I just check that basisInitialize() is called once for whole project?
> Where in the app startup process should I call basisInitialize() if I need to?
Yes correct. This is called once at application startup. Currently `basisInitialize` must be called after the graphics device has been created, but I am busy making a change so `basisInitialize` can be called any time at startup.

> Does this refer to the Basis Mode ( ETC/ASTC ) setting in the editor texture inspector?
Yes, correct. Basis can store the texture in either of these compressed formats. Using the matching format at runtime means the transcode step is a simple copy operation. Since no conversion is taking place, it is fast and doesn't suffer any quality degradation.

PR #3279, opened 2021-06-30
Improved error reporting for missing / incorrect shader

mvaligursky:
- when a shader (vertex or fragment) is not specified, we print error to console, and avoid crash when reporting the line number later.
- when shader failed to compile, we skip over rendering meshes that use it to avoid flood of unrelated errors.

based on an issue reported here https://forum.playcanvas.com/t/declaration-variables-in-shaders/20906/3

### DIFF
--- a/src/graphics/graphics-device.js
+++ b/src/graphics/graphics-device.js
@@ -2361,6 +2361,8 @@ class GraphicsDevice extends EventHandler {
         let sampler, samplerValue, texture, numTextures; // Samplers
         let uniform, scopeId, uniformVersion, programVersion; // Uniforms
         const shader = this.shader;
+        if (!shader)
+            return;
         const samplers = shader.samplers;
         const uniforms = shader.uniforms;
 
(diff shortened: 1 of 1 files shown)


Maksims:
Awesome!
Also, would love to have better reporting of error code, instead of the whole shader, just perhaps 3 lines before and 3 lines after with line numbers - would be great. Otherwise, it is a lot of scrolling.

Also, please take into account a scenario where material chunks are re-applied and the shader is queued for recompile, which can lead to fixing or error (again). This workflow is really useful when writing shaders with live-reload by subscribing to the asset `load` event.

willeastcott (approved):

mvaligursky:
Good suggestion @Maksims - I get annoying by those long shaders in the errors as well. I'll create a separate issue to address.

The recompiling part should work - I believe you fixed it many months ago, and that should still work.

mvaligursky:
@Maksims - issue created https://github.com/playcanvas/engine/issues/3280

PR #3293, opened 2021-07-05
[FIX] Handle inf on windows correctly

slimbuck:
Fixes an issue where reflective materials were rendering incorrectly on windows and linux.

The cause was a difference in the result of a shader multiplication when one of the operands are infinity. On mac and mobile `0.0 * inf == 0`, but on windows and linux not (presumably infinity propagates or result is nan).

The fix is to test for this case upfront.

### DIFF
--- a/src/graphics/program-lib/chunks/TBNderivative.frag
+++ b/src/graphics/program-lib/chunks/TBNderivative.frag
@@ -15,6 +15,7 @@ void getTBN() {
     vec3 B = dp2perp * duv1.y + dp1perp * duv2.y;
 
     // construct a scale-invariant frame
-    float invmax = 1.0 / sqrt( max( dot(T,T), dot(B,B) ) );
+    float denom = max( dot(T,T), dot(B,B) );
+    float invmax = (denom == 0.0) ? 0.0 : 1.0 / sqrt( denom );
     dTBN = mat3( T * invmax, B * invmax, dVertexNormalW );
 }


src/graphics/program-lib/chunks/TBNderivative.frag
```diff
@@ -15,6 +15,7 @@ void getTBN() {
     vec3 B = dp2perp * duv1.y + dp1perp * duv2.y;
 
     // construct a scale-invariant frame
-    float invmax = 1.0 / sqrt( max( dot(T,T), dot(B,B) ) );
+    float denom = max( dot(T,T), dot(B,B) );
+    float invmax = (denom == 0.0) ? 0.0 : 1.0 / sqrt( denom );
```
mvaligursky, line 19: I wonder if there's a workaround without branching?

yaustar, line 19: Does adding an 'if statement' affect performance in significant way here?

yaustar (approved):

mvaligursky (approved):

src/graphics/program-lib/chunks/TBNderivative.frag
slimbuck: I don't think the branch is an issue. The graphics card will calculate both sides and choose based on comparison (not actually perform a branch and jump). So the overhead introduced here is the comparison. If you can think of a better way though, I'm all ears.

mvaligursky: yep I agree with this, should be fine.

slimbuck: I wouldn't expect a measurable difference. Check [this](https://github.com/playcanvas/engine/pull/3293#discussion_r663878611) answer.

Issue #3294, opened 2021-06-19
Drag and drop a new texture on a diffuse material don't update scene objects


alamboley:
Hello,

I've several models in my scene sharing the same material.

When I select the material, in the assets, and drag (mouse down) a new texture on the diffuse image I preview the change. However when I drop (mouse up) the texture is changed but the game shows the previous rendering. I've to press the delete button and affect the texture to work as expected.

Same problem in fact using the selection/pen tool directly on the material for the texture :/

yaustar:
Commit that caused the bug: https://github.com/playcanvas/engine/pull/3074

yaustar:
Done some investigation:

Pre #3074:

```
                 material._assignTexture = function _assignTexture(parameterName, materialAsset, texture) {
			materialAsset.data[parameterName] = texture;
			materialAsset.resource[parameterName] = texture;
		};
```
When a texture was assigned to a material, `materialAsset.data[parameterName]` is set to the texture resource. This was 'bad' because `materialAsset.data[parameterName]` is also the assetId before _assignTexture was called.

However, when `_bindAndAssignAssets` is called, it was dependent on this by checking if `data[name]` was of type Texture. If it wasn't (which when we change textures in the Editor), it would assign the texture via `_assignTexture`
```
                     _proto._bindAndAssignAssets = function _bindAndAssignAssets(materialAsset, assets) {
			var data = this._parser.migrate(materialAsset.data);

			var material = materialAsset.resource;
			var pathMapping = data.mappingFormat === "path";
			var TEXTURES = standardMaterialTextureParameters;
			var i, name, assetReference;

			for (i = 0; i < TEXTURES.length; i++) {
				name = TEXTURES[i];
				assetReference = material._assetReferences[name];

				if (data[name] && !(data[name] instanceof Texture)) {
```

We now need to have some way of checking if the material has changed textures without reintroducing this behaviour.

PR #3296, opened 2021-07-05
Bind assets to material fix - Fixes an issue that shows in Editor when changing textures

yaustar:
Fixes https://github.com/playcanvas/engine/issues/3294

Context of fix: https://github.com/playcanvas/engine/issues/3294#issuecomment-874154144

Tested against the original issue that the commit fixed: https://github.com/playcanvas/engine/issues/3061

As far as I can tell, the data is not validated the first time `_bindAndAssignAssets` is called with new data (such as those coming from the Editor) so we can use this flag to know if this is the first time we have binded this data.

Part of me is not keen on using this flag as if we change the order of this function to validate the data first, this issue will reappear.

@vkalpias Can you see any issues/edge cases with this?

### DIFF
--- a/src/resources/material.js
+++ b/src/resources/material.js
@@ -238,7 +238,14 @@ class MaterialHandler {
 
             // data[name] contains an asset id for a texture
             // if we have an asset id and nothing is assigned to the texture resource or the placeholder texture is assigned
-            if (data[name] && (!materialAsset.resource[name] || materialAsset.resource[name] === this._getPlaceholderTexture(name, materialAsset))) {
+            // or the data has changed
+            const dataAssetId = data[name];
+
+            const materialTexture = material[name];
+            const dataValidated = data.validated;
+            const isPlaceHolderTexture = materialTexture === this._getPlaceholderTexture(name, materialAsset);
+
+            if (dataAssetId && (!materialTexture || !dataValidated || isPlaceHolderTexture)) {
                 if (!assetReference) {
                     assetReference = new AssetReference(name, materialAsset, assets, {
                         load: this._onTextureLoad,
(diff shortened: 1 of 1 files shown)


src/resources/material.js
```diff
@@ -238,7 +238,14 @@ class MaterialHandler {
 
             // data[name] contains an asset id for a texture
             // if we have an asset id and nothing is assigned to the texture resource or the placeholder texture is assigned
-            if (data[name] && (!materialAsset.resource[name] || materialAsset.resource[name] === this._getPlaceholderTexture(name, materialAsset))) {
+            // or the data has changed
+            const dataAssetId = data[name];
+
+            const materialTexture = material[name];
```
slimbuck, line 244: shouldn't this be `... = material.resource[name]`?

yaustar: Line 226 `material = materialAsset.resource`

The old code was `materialAsset.resource[name]`

slimbuck: 👍🏻

src/resources/material.js
```diff
@@ -238,7 +238,14 @@ class MaterialHandler {
...
+            const dataValidated = data.validated;
+            const isPlaceHolderTexture = materialTexture === this._getPlaceholderTexture(name, materialAsset);
```
slimbuck, line 246: I'm being a little bit pedantic here, but would suggest placing the two lines that deal with `materialTexture` together.
```suggestion
            const isPlaceHolderTexture = materialTexture === this._getPlaceholderTexture(name, materialAsset);
            const dataValidated = data.validated;
```

mvaligursky (approved):

PR #1645, opened 2019-08-07
fix: app.destroy do not destroy all textures

scarletsky:
When calling `app.destroy`, it only destroy textures in `app.assets`. But not all textures are assets, many textures are created by user via `new pc.Texture` directly.

Here is the starter kit:

image

You can see vram is not freed after you call `app.destroy`.

This PR is trying to solve that.

I confirm I have signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #3302, opened 2021-07-06
Destroy all lightmaps when the Application is destroyed

mvaligursky:
Fixes https://github.com/playcanvas/engine/pull/1645

When the application is destroyed, it destroys all lightmaps stored in the cache. This should not make any difference, but it's nice to destroy internally allocated engine resources.

Note that the issue mentions shadow maps not getting released - this has been addressed about a month ago when refactoring shadow rendering.

### DIFF
--- a/src/core/ref-counted-cache.js
+++ b/src/core/ref-counted-cache.js
@@ -8,6 +8,15 @@ class RefCountedCache {
         this.cache = new Map();
     }
 
+    // destroy all stored objects
+    destroy() {
+        console.log("RefCountedCache destroy");
+        this.cache.forEach((refCount, lightmap) => {
+            lightmap.destroy();
+        });
+        this.cache.clear();
+    }
+
     // add object reference to the cache
     incRef(object) {
         const refCount = (this.cache.get(object) || 0) + 1;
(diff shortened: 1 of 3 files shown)


src/core/ref-counted-cache.js
```diff
@@ -8,6 +8,15 @@ class RefCountedCache {
         this.cache = new Map();
     }
 
+    // destroy all stored objects
+    destroy() {
+        console.log("RefCountedCache destroy");
```
willeastcott, line 13: Delete this comment?

src/core/ref-counted-cache.js
```diff
@@ -8,6 +8,15 @@ class RefCountedCache {
...
+        this.cache.forEach((refCount, lightmap) => {
```
willeastcott, line 14: This is a general purpose utility class, right? Yet it now explicitly mentions `lightmap`s. Should we rename `lightmap` to `cacheItem` or something?

src/core/ref-counted-cache.js
mvaligursky: done

src/core/ref-counted-cache.js
mvaligursky: thanks for paying attention, not sure how I missed it.

willeastcott (approved):

PR #3323, opened 2021-07-09
asset.load should only be true if there was no network error

yaustar:
Bug: If there was a network error when loading an asset, `asset.loaded` is set to true. This seems odd as there is no resource so is it truly loaded?

I'm a bit worried about this change affecting existing projects but looking at the engine, there are a lot of assumptions that if `loaded` is true, then there is also a `resource`.

For example, here if `loaded` is true, it prevents the user from loading the asset again:

https://github.com/playcanvas/engine/blob/master/src/asset/asset-registry.js#L338

### DIFF
--- a/src/asset/asset-registry.js
+++ b/src/asset/asset-registry.js
@@ -362,7 +362,6 @@ class AssetRegistry extends EventHandler {
 
         // load has completed on the resource
         var _loaded = function (err, resource, extra) {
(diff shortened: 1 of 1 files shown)


yaustar:
Closed as `loaded` is deemed to be when the network request has finished regardless if a resource is available

Issue #3313, opened 2021-07-08
Anim transition of state to itself via ANY state does not activate


yaustar:
Repro: https://playcanvas.com/editor/scene/1191255

![image](https://user-images.githubusercontent.com/16639049/124913175-b9f14100-dfe6-11eb-9b29-237ca0ac5ccb.png)

1. Press 1 to set the trigger for the idle state transition
2. Note the animation doesn't restart nor the trigger consumed (see console)
3. Press 2 to set the trigger for the jump state transition
4. Note that it plays for a frame and then the transition to idle is activated as the trigger was still set from step 1.

Expectation: The animation should restart going from idle -> idle

yaustar:
We had to revert the fix as it broke existing state graphs such as the third person example: https://developer.playcanvas.com/en/tutorials/third-person-controller/

yaustar:
Closing this as now there is API to set the trigger for one frame only which means that it is no longer required for the graph to consume the event https://developer.playcanvas.com/en/api/pc.AnimComponent.html#setTrigger

Issue #3270, opened 2021-06-25
'One shot' trigger parameters for anim state graph?


yaustar:
At the moment, trigger parameters are only consumed if there is a transition valid for it.

There's been a few users that wish for a 'one shot' that is only valid for that update and then consumed at the end of the frame.

Issue #3268, opened 2021-06-25
Triggers on anim components to work on all layers for a single frame?


yaustar:
This is still up for discussion. At the moment, the trigger parameter is consumed by the first layer it is valid for. There was a suggestion from @mvaligursky that the trigger parameter could be valid for the whole frame update and then unset at the end of the frame.

PR #3329, opened 2021-07-12
AnimComponent transition updates

ellthompson:
This PR addresses a few requests that had been made in relation to anim component transitions.

- Transitions can now be made from the active state to itself (or from the any state to the active state). #3313
- Trigger parameters are now reset at the end of the animation update rather than when they're consumed by a transition. This allows multiple layers to use a single trigger before it is reset. #3268
- Updates the `AnimComponent#setTrigger` to support a `singleFrame` boolean parameter. If set to true the set trigger will be reset at the end of the next animation update. #3270

### DIFF
--- a/src/framework/components/anim/component.js
+++ b/src/framework/components/anim/component.js
@@ -647,7 +647,6 @@ class AnimComponent extends Component {
      * @description Sets the value of a trigger parameter that was defined in the animation components state graph to true.
      * @param {string} name - The name of the parameter to set.
      * @param {boolean} [singleFrame] - If true, this trigger will be set back to false at the end of the animation update. Defaults to false.
-
      */
     setTrigger(name, singleFrame = false) {
         this.setParameterValue(name, ANIM_PARAMETER_TRIGGER, true);
--- a/src/framework/components/anim/component.js
+++ b/src/framework/components/anim/component.js
@@ -646,9 +646,13 @@ class AnimComponent extends Component {
      * @name AnimComponent#setTrigger
      * @description Sets the value of a trigger parameter that was defined in the animation components state graph to true.
      * @param {string} name - The name of the parameter to set.
+     * @param {boolean} singleFrame - If true, this trigger will be set back to false at the end of the animation update. Defaults to false.
      */
-    setTrigger(name) {
+    setTrigger(name, singleFrame = false) {
         this.setParameterValue(name, ANIM_PARAMETER_TRIGGER, true);
+        if (singleFrame) {
+            this._consumedTriggers[name] = true;
+        }
     }
 
     /**
--- a/src/framework/components/anim/component.js
+++ b/src/framework/components/anim/component.js
@@ -646,7 +646,8 @@ class AnimComponent extends Component {
      * @name AnimComponent#setTrigger
      * @description Sets the value of a trigger parameter that was defined in the animation components state graph to true.
      * @param {string} name - The name of the parameter to set.
-     * @param {boolean} singleFrame - If true, this trigger will be set back to false at the end of the animation update. Defaults to false.
+     * @param {boolean} [singleFrame] - If true, this trigger will be set back to false at the end of the animation update. Defaults to false.
+
      */
     setTrigger(name, singleFrame = false) {
         this.setParameterValue(name, ANIM_PARAMETER_TRIGGER, true);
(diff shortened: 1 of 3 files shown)


src/framework/components/anim/system.js
```diff
@@ -58,17 +58,21 @@ class AnimComponentSystem extends ComponentSystem {
...
+                const component = components[id].entity.anim;
+                const componentData = component.data;
 
                 if (componentData.enabled && component.entity.enabled && component.playing) {
-                    for (var i = 0; i < component.layers.length; i++) {
+                    for (let i = 0; i < component.layers.length; i++) {
                         component.layers[i].update(dt * component.speed);
                     }
+                    for (let i = 0; i < Object.keys(component.consumedTriggers).length; i++) {
+                        component.parameters[Object.keys(component.consumedTriggers)[i]].value = false;
```
slimbuck, line 73:
```suggestion
                    const keys = Object.keys(component.consumedTriggers);
                    for (let i = 0; i < keys.length; i++) {
                        component.parameters[keys[i]].value = false;
```

src/framework/components/anim/component.js
```diff
@@ -636,9 +646,13 @@ class AnimComponent extends Component {
      * @name AnimComponent#setTrigger
      * @description Sets the value of a trigger parameter that was defined in the animation components state graph to true.
      * @param {string} name - The name of the parameter to set.
+     * @param {boolean} singleFrame - If true, this trigger will be set back to false at the end of the animation update. Defaults to false.
```
yaustar, line 649: JSdocs should show this is optional?
```suggestion
     * @param {boolean} [singleFrame] - If true, this trigger will be set back to false at the end of the animation update. Defaults to false.

```

ellthompson: Good catch, cheers

src/framework/components/anim/component.js
```diff
@@ -213,6 +214,14 @@ class AnimComponent extends Component {
...
     }
 
+    get consumedTriggers() {
+        return this._consumedTriggers;
+    }
+
+    set consumedTriggers(value) {
+        this._consumedTriggers = value;
+    }
+
```
mvaligursky, line 224: it seems the setter and getter does not do anything extra in this case, so I'd suggest to just have this.consumerTriggers = {}; in constructor and remove these. They can be always added when there's a need for them - you'd just rename the member in constructor by prefixing it with '_' , and implement setter / getter.

src/framework/components/anim/system.js
```diff
@@ -58,17 +58,22 @@ class AnimComponentSystem extends ComponentSystem {
...
                 if (componentData.enabled && component.entity.enabled && component.playing) {
-                    for (var i = 0; i < component.layers.length; i++) {
+                    for (let i = 0; i < component.layers.length; i++) {
                         component.layers[i].update(dt * component.speed);
                     }
+                    const keys = Object.keys(component.consumedTriggers);
+                    for (let i = 0; i < keys.length; i++) {
+                        component.parameters[keys[i]].value = false;
+                        component.consumedTriggers = {};
+                    }
```
mvaligursky, line 76: I think this block of code should be inside AnimComponent.update or so instead of here. Probably including this part of the condition `component.playing`.

Also, considering using perhaps a Set for consumedTriggers, so that you can simply clear() it after update, instead of allocating new object.

src/framework/components/anim/system.js
```diff
@@ -58,17 +58,22 @@ class AnimComponentSystem extends ComponentSystem {
...
 
                 if (componentData.enabled && component.entity.enabled && component.playing) {
-                    for (var i = 0; i < component.layers.length; i++) {
+                    for (let i = 0; i < component.layers.length; i++) {
                         component.layers[i].update(dt * component.speed);
                     }
+                    const keys = Object.keys(component.consumedTriggers);
+                    for (let i = 0; i < keys.length; i++) {
+                        component.parameters[keys[i]].value = false;
+                        component.consumedTriggers = {};
```
mvaligursky, line 75: also, should this line be inside the loop?

src/framework/components/anim/component.js
```diff
@@ -228,7 +237,8 @@ class AnimComponent extends Component {
             transitions,
             this._parameters,
             this._activate,
-            this
+            this,
+            this._resetTrigger.bind(this)
```
mvaligursky, line 241: if you implement consumedTriggers as a set, I'd simply pass the set to controller here to add to it as needed, instead of doing this callback.

mvaligursky (approved):
this looks good now, well done.

Issue #3310, opened 2021-07-07
[Examples browser] Commented out code can break the example


mvaligursky:
Two cases here:

1. When the comment // starts on the column 0 of the line it's ignored and the commented out code executes. If there is a space in front of the comment, all is good.

2. adding line like this:
`      // sampleCubemapColors(device: pc.GraphicsDevice): void {`
in front of example line breaks the example. This is due to ': void' being there used to find the example function.

<img width="605" alt="Screenshot 2021-07-07 at 15 47 43" src="https://user-images.githubusercontent.com/59932779/124780657-a1eec400-def7-11eb-8a9a-2e115cf61922.png">

Issue #3303, opened 2021-07-06
[Examples Browser] Gracefully handle incorrect example path


mvaligursky:
At the moment, when incorrect path is used, we get exceptions. Ideally this could be handled gracefully and perhaps display the hello-world example.

example: http://playcanvas.github.io/#/misc/bug

![Screenshot 2021-07-06 at 15 59 09](https://user-images.githubusercontent.com/59932779/124622548-0fcfb880-de30-11eb-9956-b70368797938.png)

Issue #3297, opened 2021-07-05
[Examples Browser] Toggle stats on/off should not reload the example


mvaligursky:
it should simply enable / disable mini-stats.

Issue #3287, opened 2021-07-02
[Examples Browser] Examples tab collapses when filter is used on mobile devices


LeXXik:
Forum: https://forum.playcanvas.com/t/bug-in-playcanvas-examples-github-pages/20943

When Filter field is used on mobile devices in Examples Browser, the Examples tab collapses, making it difficult to filter items.

PR #3333, opened 2021-07-14
[Examples Browser] Updates & fixes

ellthompson:
This PR contains a bunch of tweaks and fixes to the examples browser project. There were too many small changes to justify individual PR's so i've grouped them here.

### Updates
- Bump playcanvas dependancy to 1.44.2 playcanvas/engine@32b7a1c
- Support performance and debug builds of playcanvas in examples (defined using the static ENGINE property). This enables the mini stats in the clustered lighting example playcanvas/engine@2a61f07 (#3254)
- Show mini stats toggle no longer requires an example rebuild to switch playcanvas/engine@6b1b3ac (#3297)
- All non example paths now direct to the hello world example playcanvas/engine@70179a1 (#3303)
- When switching thumbnail sizes in the side bar, the position of the top most nav item is maintained playcanvas/engine@5c5a94d
- Opening the examples browser using a specific path to an example will scroll that example's nav item into view in the side bar playcanvas/engine@fd6f7b4

### Fixes
- Filter out code editor warnings when validating browser edits of examples. Warnings were previously blocking rebuilds. playcanvas/engine@1f593a1
- Fixes an issue where including extra functions in an Example class could break the example build playcanvas/engine@4ebe5e1 (#3310)
- Fixes an issue where comments in an example that start at line position 0 break the build playcanvas/engine@e87521d (#3310)
- The code view panel was hidden on mobile for examples without controls. Fixed in: playcanvas/engine@1eb6fe8
- The input of text in the filter text input no longer closes the side bar on mobile playcanvas/engine@c1b4904 (#3287)

### DIFF
--- a/examples/package-lock.json
+++ b/examples/package-lock.json
@@ -7108,9 +7108,9 @@
       }
     },
     "playcanvas": {
-      "version": "1.43.1",
-      "resolved": "https://registry.npmjs.org/playcanvas/-/playcanvas-1.43.1.tgz",
-      "integrity": "sha512-x7muwlT8XiEyXffaltKeisuyRlWBztSskteNok9MfiXnw7dvaRGjGDoao1pS56lXcEI6CiUm+ddbHJpxiTw/Rw==",
+      "version": "1.44.2",
+      "resolved": "https://registry.npmjs.org/playcanvas/-/playcanvas-1.44.2.tgz",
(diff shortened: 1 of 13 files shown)


mvaligursky:
Some great changes!

ellthompson:
> why are you stripping anything other than spaces from the start?

I'm stripping the first 4 columns from the example function as it begins at column 4 in each examples .tsx file. When these functions are displayed in the browser editor they should start at column 0.

ellthompson:
The fix will allow comments like this:
image
to appear as follows:
image

Issue #3322, opened 2021-07-09
Orient images correctly


slimbuck:
The PlayCanvas engine and backend considers textures as originating at the bottom left corner. This is in line with OpenGL (and modelling tools like Blender, Maya) which maps the UV coordinate (0, 0) to the bottom left corner of the image (i.e. the first pixel data encountered in texture memory is located at the bottom left of the image).

Inexplicably however WebGL considers the texture origin to be top left instead, see https://www.khronos.org/registry/webgl/specs/1.0/index.html#TEXIMAGE2D_HTML - “The first pixel transferred from the source to the WebGL implementation corresponds to the upper left corner of the source”.

More background on glTF choice of top-left can be found at https://github.com/KhronosGroup/glTF/issues/1021.

We must update the engine to use top-left texture origin in order to be in line with the industry and add better support for shared assets like glTF.

In order to accomplish this, we must make the following changes:
- refrain from flipping images at load time (png, jpg etc)
- refrain from flipping glTF V coordinates at load time (which currently compensates for the flipped images)
- flip PlayCanvas json model texture V coordinates at load time
- generate correctly oriented V coordinates for all dynamically-generated texture coordinates, this includes:
    - built-in shapes (sphere, cylinder, plane etc)
    - 2d elements (image element, text element, sprite normal, 9-sliced, 9-sliced-tiled)
- modify the PlayCanvas backend to stop flipping textures at compress time
    - this will require users to re-compress their project textures once the updated engine is released

willeastcott:
It's worth scanning the WebGPU spec to see how it treats textures. Decisions we take now maybe be relevant to our future WebGPU support.

PR #3335, opened 2021-07-15
[FIX] Adopt top-left texture origin

slimbuck:
Fixes: #3322 

This PR changes the orientation of textures in the engine from bottom-left to top-left, bringing it inline with WebGL and glTF.

Specifically, the following changes are made:
- stop flipping png/jpg textures vertically at load time
- stop inverting glTF texture coordinates at load time
- start inverting JSON model texture coordinates at load time
- invert dynamically-generated texture coordinates for:
    - text
    - images
    - particles
    - sprites
- invert texture coordinates of built-in shapes
- update the basis texture engine example so it's not ugliest of all

Screenshot 2021-07-15 at 11 09 09

NOTE: the following shader chunks have been changed in this PR:
- `particle.frag`
- `particle_cpu.vert`
- `particle_normalMap.frag`
- `startNineSliced.frag`
- `startNiceSlicedTiled.frag`

Users who have custom versions of these chunks will have to update these accordingly.


Maksims:
This change has the potential to break a lot of advanced users projects who render their own textures and use them in shaders.
Could this change be added to respect backwards compatibility, and have a project default setting to use the bottom-left coordinate system. And for the new projects, flipping the switch?

So it won't break any existing projects and ensure a smooth transition, which can be done by the developer by choice.

src/resources/parser/texture/img.js
```diff
@@ -18,7 +18,9 @@ class ImgParser {
         // by default don't try cross-origin, because some browsers send different cookies (e.g. safari) if this is set.
         this.crossOrigin = registry.prefix ? 'anonymous' : null;
         this.maxRetries = 0;
-        // disable ImageBitmap
+        // As of today (9 Jul 2021) ImageBitmap only works on Chrome:
```
mvaligursky, line 21: please create an issue for this in the engine repo (if not already) so that we can reevaluate at a later state.

mvaligursky (approved):
looks great to me, glad to have this sorted out!

slimbuck:
Hi @Maksims ,

Yes you're right. This change has the potential to impact projects, especially ones that dynamically generate texture coordinates or have custom chunks generating UVs. Fortunately in most cases the fix is simple, place `uv.y = 1.0 - uv.y;` at the end.

Keeping backwards compatibility would be difficult in the engine (think different chunks etc) and would leave us having to maintain quite a complicated global option. Over time we'd expect to see bugs creep in and ultimately we decided that making a clean break is the best approach here.

Also, with this change in place we can finally support new features that weren't possible before (glTF texture transform, glTF quantising, gltf compressed textures). If we were to keep backwards compatibility those features would also be compromised.

Along with this change we have also updated the backend to stop flipping compressed textures. When this PR goes live, users will get an alert in the editor saying textures need recompressing. A single click will recompress all textures in the project.

This is a big change which might be a little painful, but we are preparing to make the transition hopefully as smooth as possible!

src/resources/parser/texture/img.js
slimbuck: done 👍🏻

https://github.com/playcanvas/engine/issues/3336

willeastcott (approved):

PR #3337, opened 2021-07-15
[Examples Browser] Controls for the post effects / lights examples

ellthompson:
This PR moves control of the post effects and lights examples from the keyboard to the control panel. The control panel has been updated to support scrolling and will no longer extend past 100% of the screen. Also made a few other minor visual tweaks to the control panel.

image

image

### DIFF
--- a/examples/src/app/control-panel.tsx
+++ b/examples/src/app/control-panel.tsx
@@ -9,7 +9,7 @@ const ControlPanel = (props: any) => {
     const [state, setState] = useState({
         showParameters: !!props.controls,
         showCode: !props.controls,
-        collapsed: document.body.offsetWidth < 601
+        collapsed: window.top.innerWidth < 601
     });
     const beforeMount = (monaco: any) => {
         monaco.languages.typescript.typescriptDefaults.addExtraLib(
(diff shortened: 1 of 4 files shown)


mvaligursky:
fantastic. Could we keep keyboard shortcuts in place as well somehow for both examples? That is really fast way to do when working on lights / shadows and similar.

mvaligursky:
And just double checking, does it work (fit on the screen) on mobile?

ellthompson:
> fantastic. Could we keep keyboard shortcuts in place as well somehow for both examples? That is really fast way to do when working on lights / shadows and similar.

With the text elements on screen too or just the keyboard inputs? It looked pretty cluttered with that text.

ellthompson:
> And just double checking, does it work (fit on the screen) on mobile?

Tested and working on mobile. Large collections of controls will be scrollable.

mvaligursky:
> With the text elements on screen too or just the keyboard inputs?

Maybe change the header names somehow? instead of "Bloom", have "Bloom [Key_5]" or something like that.

ellthompson:
> > With the text elements on screen too or just the keyboard inputs?
> 
> Maybe change the header names somehow? instead of "Bloom", have "Bloom [Key_5]" or something like that.

This works 👍  Added they key controls back in with updated header texts in the control panel.

mvaligursky (approved):

PR #3346, opened 2021-07-20
RTT texture flip flag

slimbuck:
In #3335 we inverted V coordinates of 2d elements and built-in primitives to compensate for -not- flipping images at load time.

This change caused dynamically-rendered (RTT) textures to appear upside down.

This PR adds a flag to a render target `flipY` which will cause it to be rendered upside down. Enabling this flag on render targets destined for UI rendering will result in images correctly orientated.

### DIFF
--- a/src/graphics/render-target.js
+++ b/src/graphics/render-target.js
@@ -32,6 +32,8 @@ const defaultOptions = {
  * * {@link CUBEFACE_POSZ}
  * * {@link CUBEFACE_NEGZ}
  *
+ * @param {boolean} [options.flipY] - Render the image flipped in Y, useful when the render target will be used as a texture in UI rendering.
+ *
  * Defaults to {@link CUBEFACE_POSX}.
  * @example
  * // Create a 512x512x24-bit render target with a depth buffer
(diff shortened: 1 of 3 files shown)


src/graphics/render-target.js
```diff
@@ -32,6 +32,8 @@ const defaultOptions = {
  * * {@link CUBEFACE_POSZ}
  * * {@link CUBEFACE_NEGZ}
  *
+ * @param {boolean} [options.flipY] - Render the image flipped in Y, useful when the render target will be used as a texture in UI rendering.
```
mvaligursky, line 35: should it mentioned "texture in UI rendering"? It seems like it's useful for using it as a texture in general.

mvaligursky: Also, specify what it defaults to please.

mvaligursky (approved):

src/graphics/render-target.js
slimbuck: nice one, done. thanks

willeastcott (approved):

PR #3352, opened 2021-07-21
[FIX] Check _flipFaces on camera not camera component

slimbuck:
Error introduced in #3346

### DIFF
--- a/src/scene/renderer/forward-renderer.js
+++ b/src/scene/renderer/forward-renderer.js
@@ -2299,7 +2299,7 @@ class ForwardRenderer {
 
                 // enable flip faces if either the camera has _flipFaces enabled or the render target
                 // has flipY enabled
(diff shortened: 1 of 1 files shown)


yaustar:
Should we rename `camera` to `cameraComponent`?

yaustar (approved):
LGTM. Comment previously provided 'Should we rename `camera` to `cameraComponent`'

slimbuck:
> Should we rename `camera` to `cameraComponent`?

Maybe some other time :)

yaustar:
> > Should we rename `camera` to `cameraComponent`?
> 
> Maybe some other time :)

Would we hit the same issue next time though? It's only because I double check potential tutorial projects that may be affected by render textures that we found this?

willeastcott:
We should add that project to the test plan. 😄

slimbuck:
> We should add that project to the test plan. 😄

Done :)

PR #3356, opened 2021-07-22
[Fix] Drc loader example uses loaded colors correctly

mvaligursky:
colors are loaded as RGB, need to be converted to RGBA (fixes engine warning in debug mode too).
Result - points now render with color correctly.

Screenshot 2021-07-22 at 09 44 28

### DIFF
--- a/examples/src/examples/loaders/drc.tsx
+++ b/examples/src/examples/loaders/drc.tsx
@@ -64,11 +64,22 @@ class LoadersGlExample extends Example {
             // @ts-ignore: cannot find CORE and DRACO
             const modelData = await CORE.load(url, DRACO.DracoLoader);
 
+            // loaded colors only contain RGB, convert it to an array of RGBA with alpha of 255
+            const srcColors = modelData.attributes.COLOR_0.value;
+            const numVertices = srcColors.length / modelData.attributes.COLOR_0.size;
+            const colors32 = new Uint8Array(numVertices * 4);
+            for (let i = 0; i < numVertices; i++) {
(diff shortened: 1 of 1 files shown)


examples/src/examples/loaders/drc.tsx
```diff
@@ -64,11 +64,22 @@ class LoadersGlExample extends Example {
...
+            const srcColors = modelData.attributes.COLOR_0.value;
+            const numVertices = srcColors.length / modelData.attributes.COLOR_0.size;
+            const colors32 = new Uint8Array(numVertices * 4);
+            for (let i = 0; i < numVertices; i++) {
+                colors32[i * 4 + 0] = srcColors[i * 3 + 0];
+                colors32[i * 4 + 1] = srcColors[i * 3 + 1];
+                colors32[i * 4 + 2] = srcColors[i * 3 + 2];
+                colors32[i * 4 + 3] = 255;
+            }
+
```
willeastcott, line 77: It seems that the [glTF spec also allows for RGB vertex colors](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#meshes). Should the engine just be able to use them directly without the load time conversion?

willeastcott (approved):

examples/src/examples/loaders/drc.tsx
mvaligursky: mesh API allows / supports RGB in float format, but 8bit format needs to be RGBA .. webgl vertex attribute cannot be smaller than 32bit.

PR #3363, opened 2021-07-23
Invert TBN, glTF texture transform

slimbuck:
This change follows on from #3335 

Since we inverted V texture coordinates, it seems we must flip the texture basis too.

Before:
Screenshot 2021-07-23 at 16 38 47

After:
Screenshot 2021-07-23 at 16 38 13

Flipping the TBN basis requires more investigation, as I don't believe it should be necessary.

This PR also addresses inverted glTF texture transform:

Before:
Screenshot 2021-07-23 at 16 39 03

After:
Screenshot 2021-07-23 at 16 39 36


mvaligursky (approved):
Nice one!

PR #3366, opened 2021-07-25
Add glTF exporter

willeastcott:
Implements a glTF exporter. Currently supports:

- [x] Meshes
- [x] Materials
- [ ] Textures
- [ ] Lights
- [x] Cameras
- [x] GLB export
- [ ] glTF (JSON) export

API is:

```javascript
new pcx.GltfExporter().build(entity).then((arrayBuffer) => {
    const blob = new Blob([arrayBuffer], { type: 'application/octet-stream' });

    // download the blob here
}).catch(console.error);
```

also exposed a function on the engine's BoundingBox to evaluate min and max for an array of vertices:
```
// this allows to easily compute min and max, instead of getting those from mesh.aabb, as due to
// storing center + extents, that is not precise enough and we get gltf validation warnings.
BoundingBox.computeMinMax()
```

New engine example **GltfExport** converts source scene created from multiple glbs:
Screenshot 2022-09-28 at 11 55 49

into a single glb (limited material & texture support)
Screenshot 2022-09-28 at 11 55 34


kungfooman:
I try to get this to work by using the `download` function from here: https://github.com/playcanvas/playcanvas-gltf/blob/a96d6b307fb5c8c3a3b6c1a926e0788bc1932128/src/export-gltf.js#L48

```js
function download(buffers) {
  var element = document.createElement('a');
  var blob = new Blob(buffers, { type: "octet/stream" });
  element.href = URL.createObjectURL(blob);
  element.download = 'scene.glb';
  element.style.display = 'none';
  document.body.appendChild(element);
  element.click();
  document.body.removeChild(element);
}
// Export as scene.glb
const exporter = new GltfExporter();
const arrayBuffer = exporter.buildGlb(entity):
download([arrayBuffer]);
```

I can download the generated `scene.glb` and the hierarchy seems correct, but when I load it into PlayCanvas Viewer, it just breaks something (e.g. skybox turns off) and I can't see any models:

image

Am I doing something wrong or does it not work for meshes yet?

Generated file: [scene.zip](https://github.com/playcanvas/engine/files/6997866/scene.zip)

scripts/exporters/gltf-exporter.js
```diff
@@ -0,0 +1,441 @@
...
+                if (entity.camera && entity.camera.enabled) {
+                    node.camera = resources.cameras.indexOf(entity.camera);
+                }
+
+                if (entity.render && entity.render.enabled) {
+                    entity.render.meshInstances.forEach((meshInstance) => {
+                        node.mesh = resources.meshInstances.indexOf(meshInstance);
+                    });
+                }
+
```
kungfooman, line 239:
```suggestion

                if (entity.model && entity.model.enabled) {
                    entity.model.meshInstances.forEach((meshInstance) => {
                        node.mesh = resources.meshInstances.indexOf(meshInstance);
                    });
                }
```

Write the JSON mesh id for MeshInstances inside a ModelComponent

scripts/exporters/gltf-exporter.js
```diff
@@ -0,0 +1,441 @@
...
+                    json.cameras.push({
+                        type: "perspective",
+                        perspective: {
+                            yfov: fov * Math.PI / 180,
+                            znear: nearClip,
+                            zfar: farClip
+                        }
+                    });
+                }
+            });
```
kungfooman, line 158: A `resources.cameras.forEach` combined with `push` in every case is basically a `resources.cameras.map`, which may give that code more intention (probably subjective, I just see it as less manual array plumbing)

kungfooman (changes requested):

scripts/exporters/gltf-exporter.js
willeastcott: Good point, yeah, I don't disagree.

scripts/exporters/gltf-exporter.js
```diff
@@ -0,0 +1,446 @@
...
+                    entity.render.meshInstances.forEach((meshInstance) => {
+                        node.mesh = resources.meshInstances.indexOf(meshInstance);
+                    });
+                }
+
+                if (entity.model && entity.model.enabled) {
+                    entity.model.meshInstances.forEach((meshInstance) => {
+                        node.mesh = resources.meshInstances.indexOf(meshInstance);
+                    });
+                }
```
kungfooman, line 244:
```suggestion
```

scripts/exporters/gltf-exporter.js
```diff
@@ -0,0 +1,446 @@
...
+                    entity.children.forEach((child) => {
+                        node.children.push(resources.entities.indexOf(child));
+                    });
+                }
+
+                json.nodes.push(node);
+            });
+        }
+    }
+
```
kungfooman, line 257:
```suggestion
    assignMeshIDs(resources, json) {
        var meshInstances = resources.meshInstances;
        // Assign meshes to JSON nodes based on meshInstance.node
        meshInstances.forEach(meshInstance => {
            const entityIndex = resources.entities.indexOf(meshInstance.node);
            if (entityIndex != -1) {
                json.nodes[entityIndex].mesh = meshInstances.indexOf(meshInstance);
            } else {
                console.warn('GltfExporter#assignMeshIDs> meshInstance referring to unexported json node');
            }
        });
    }
```

scripts/exporters/gltf-exporter.js
```diff
@@ -0,0 +1,446 @@
...
+                    ]
+                }
+            ],
+            scene: 0
+        };
+
+        this.writeBuffers(resources, json);
+        this.writeBufferViews(resources, json);
+        this.writeCameras(resources, json);
+        this.writeNodes(resources, json);
```
kungfooman, line 364:
```suggestion
        this.writeNodes(resources, json);
        this.assignMeshIDs(resources, json);
```

kungfooman (changes requested):
One entity can have multiple meshInstances and each meshInstance.node (GraphNode) belongs to its own JSON glTF node. So first all JSON nodes need to be written and then the `resources.meshInstances` can be iterated to assign the mesh ID's to the proper JSON nodes:

scripts/exporters/gltf-exporter.js
```diff
@@ -0,0 +1,446 @@
...
+const ELEMENT_ARRAY_BUFFER = 34963;
+
+const BYTE = 5120;
+const UNSIGNED_BYTE = 5121;
+const SHORT = 5122;
+const UNSIGNED_SHORT = 5123;
+const UNSIGNED_INT = 5125;
+const FLOAT = 5126;
+
+function roundToNextMultiple(x, multiple = 4) {
```
kungfooman, line 11: Maybe just use `pc.math.roundUp`?

kungfooman:
I reworked the code to deinterleave the vertex buffer data and tested it on SeeMore (as non-trivial test case):

The glTF can be opened in Blender (for some reason there are still errors that prevent full loading): 
image

Or in PlayCanvas viewer:

image

SeeMore demo: https://playcanv.as/p/MflWvdTW/

SeeMore.glb file: [Seemore_1629624802680_7.zip](https://github.com/playcanvas/engine/files/7027090/Seemore_1629624802680_7.zip)

JavaScript code for quick testing: https://github.com/KILLTUBE/gltf/blob/master/src/gltf-exporter.js (just pick frame in devtools and copy&paste)

It would be nice to get to the point of a fully working and error free SeeMore glTF/glb file, currently I can't even get an error report from KhronosGroup/glTF-Validator (running out of memory)

scripts/exporters/gltf-exporter.js
willeastcott: Good point. Done.

scripts/exporters/gltf-exporter.js
willeastcott: OK, I've updated most `forEach`+`push` combinations to use `map` instead. Nice suggestion.

willeastcott:
Cool to see you experimenting with this. I was wondering why in your version you are de-interleaving the vertex buffer data. In my version, I'm trying to leave the vertex buffer data untouched. Ideally, it will handle both cases. Interleaved vertex buffer data is generally considered to be more efficient.

kungfooman:
Thanks! I de-interleaved the vertex buffer mostly because I just wanted it to work and then I am mostly using playcanvas-gltf which still has problems parsing interleaved data: https://github.com/playcanvas/playcanvas-gltf/issues/7

Once it generated a GLB file it can just be loaded into Blender and then Blender can do the interleaving and draco compression if necessary. I also found the stride logic somewhat confusing, so I just wanted to make it easier to understand

tanaydimriepigraph:
Hi Wil,
Any rough estimates on when we could have a working version of the gltf-exporter (the mesh and the materials)
Sorry, it's just that there is something in our project we can REALLY use this for but I don't have the required skills to help debug this to be able to contribute.

yaustar:
@tanaydimriepigraph Unfortunately, Will is backlogged on some internal stuff and is not able to recommence work on this just yet. He is eager to get back to it though.

tanaydimriepigraph:
@yaustar Thank you for the updates. Will keep an eye out for this!!

scripts/exporters/gltf-exporter.js
CynthiaXJia: Since there is still only one node for each entity, this will result in each meshInstance in an entity overwriting `json.nodes[entityIndex].mesh`, right? Should we be checking for meshInstances in models/renders in `writeNodes` and creating a new json node for each?

mvaligursky (approved):

src/shape/bounding-box.js
```diff
@@ -346,18 +346,20 @@ class BoundingBox {
...
      * @param {number[]|Float32Array} vertices - The vertices used to compute the new size for the
      * AABB.
+     * @param {Vec3} min - Computed min value.
+     * @param {Vec3} max - Computed max value.
      * @param {number} [numVerts] - Number of vertices to use from the beginning of vertices array.
      * All vertices are used if not specified.
      */
-    compute(vertices, numVerts) {
-        numVerts = numVerts === undefined ? vertices.length / 3 : numVerts;
+    static computeMinMax(vertices, min = new Vec3(), max = new Vec3(), numVerts = vertices.length / 3) {
```
willeastcott, line 358: Why does this create new vecs if none are supplied. They don't get returned so it seems a bit pointless?

mvaligursky: good point. fixing, thanks!

slimbuck (approved):

scripts/exporters/gltf-exporter.js
mvaligursky: This will be addressed in follow up PRs

scripts/exporters/gltf-exporter.js
mvaligursky: This will be addressed in follow up PRs

extras/exporters/gltf-exporter.js
```diff
@@ -0,0 +1,465 @@
...
+        case pc.SEMANTIC_TEXCOORD0: return 'TEXCOORD_0';
+        case pc.SEMANTIC_TEXCOORD1: return 'TEXCOORD_1';
+        case pc.SEMANTIC_TEXCOORD2: return 'TEXCOORD_2';
+        case pc.SEMANTIC_TEXCOORD3: return 'TEXCOORD_3';
+        case pc.SEMANTIC_TEXCOORD4: return 'TEXCOORD_4';
+        case pc.SEMANTIC_TEXCOORD5: return 'TEXCOORD_5';
+        case pc.SEMANTIC_TEXCOORD6: return 'TEXCOORD_6';
+        case pc.SEMANTIC_TEXCOORD7: return 'TEXCOORD_7';
+    }
+};
```
kungfooman, line 53: In `glb-parser.js` there is:

```js
const gltfToEngineSemanticMap = {
    'POSITION': SEMANTIC_POSITION,
    'NORMAL': SEMANTIC_NORMAL,
    'TANGENT': SEMANTIC_TANGENT,
    'COLOR_0': SEMANTIC_COLOR,
    'JOINTS_0': SEMANTIC_BLENDINDICES,
    'WEIGHTS_0': SEMANTIC_BLENDWEIGHT,
    'TEXCOORD_0': SEMANTIC_TEXCOORD0,
    'TEXCOORD_1': SEMANTIC_TEXCOORD1,
    'TEXCOORD_2': SEMANTIC_TEXCOORD2,
    'TEXCOORD_3': SEMANTIC_TEXCOORD3,
    'TEXCOORD_4': SEMANTIC_TEXCOORD4,
    'TEXCOORD_5': SEMANTIC_TEXCOORD5,
    'TEXCOORD_6': SEMANTIC_TEXCOORD6,
    'TEXCOORD_7': SEMANTIC_TEXCOORD7
};
```

I think an object in itself is shorter, but since its basically the same object (just inverted), I started to think these variables could be exported and then just automatically inverted, for example:

```js
function objectInvert(obj) {
    const ret = {};
    for (const key in obj) {
        const val = obj[key];
        ret[val] = key;
    }
    return ret;
}
export const engineToGltfSemanticMap = objectInvert(gltfToEngineSemanticMap);
```
(but maybe this is just overcomplicating things)

mvaligursky: I started with the object initially .. but the engine needs to be loaded before the pcx (extras) .. and so a static time access to pc constants is not possible here - and so I added a function instead, which executes after the engine has been loaded.
When we have full tree-shaking, and make this a proper module, we'd definitely do something else.

scripts/exporters/gltf-exporter.js
mvaligursky: this is addressed here now: https://github.com/playcanvas/engine/pull/4683

PR #3380, opened 2021-08-05
Support ktx2 supercompressed textures

slimbuck:
This PR:
- adds support for KTX2 basis compressed textures
- a followup PR will add support for non-supercompressed KTX2 formats
- updates the engine basis modules to include ktx2 support (code is taken from basisu version 1.15)
- removes the unused `dds` parser and renames `dds-legacy` to `dds`


willeastcott:
The Basis runtime seems to be updated here. Want to add details of that to the PR description?

slimbuck:
I mention:

> * updates the engine basis modules to ones with ktx2 load support included

What other details would be useful?

willeastcott:
I mean version info. Did you update to 1.15?

src/resources/basis-worker.js
```diff
@@ -157,8 +157,98 @@ function BasisWorker() {
...
+
+        const compressedFormat = isCompressedFormat(basisFormat);
+
+        return {
+            format: basisToEngineMapping(basisFormat, options.deviceDetails),
+            width: compressedFormat ? ((width + 3) & ~3) : width,
+            height: compressedFormat ? ((height + 3) & ~3) : height,
+            levels: levelData,
+            cubemap: false,
+            mipmaps: true,
```
willeastcott, line 243: Bit confused about this. So KTX2 won't contain non-mipmapped texture data?

src/resources/parser/texture/ktx2.js
```diff
@@ -0,0 +1,202 @@
+import { http } from '../../../net/http.js';
+import { basisTranscode } from '../../basis.js';
+import { Texture } from '../../../graphics/texture.js';
+import { ADDRESS_CLAMP_TO_EDGE, ADDRESS_REPEAT, TEXHINT_ASSET } from '../../../graphics/constants.js';
+
+// helper class for organized reading of memory
+class ReadStream {
```
willeastcott, line 7: Should this go in the core folder? Any scope for utilizing in any other modules?

willeastcott (approved):

willeastcott:
Little question about the two DDS parsers. Are we supremely confident that the parser that's been removed is genuinely not used anywhere?

src/resources/basis-worker.js
slimbuck: Ah yes nice catch. This mipmap flag is actually ignored (https://github.com/slimbuck/engine/blob/ktx2-support/src/resources/parser/texture/ktx2.js#L97) and shouldn't been here. I've removed it.

src/resources/parser/texture/ktx2.js
slimbuck: Yes I made a similar class for hdr loading and would like to extract this and have both use the same interface. Just didn't want to overcomplicate this PR.

slimbuck:
> I mean version info. Did you update to 1.15?

Oh I see. The version 1.15 actually refers to the version of the basisu tool. I don't believe it applies to the transcoder. It would be great to have versioning there though. I might request that in an issue on the repo.

src/resources/parser/texture/dds.js
```diff
@@ -5,88 +5,17 @@ import {
...
-fccToFormat[FCC_DXT1] = PIXELFORMAT_DXT1;
-fccToFormat[FCC_DXT5] = PIXELFORMAT_DXT5;
-fccToFormat[FCC_ETC1] = PIXELFORMAT_ETC1;
-fccToFormat[FCC_PVRTC_2BPP_RGB_1] = PIXELFORMAT_PVRTC_2BPP_RGB_1;
-fccToFormat[FCC_PVRTC_2BPP_RGBA_1] = PIXELFORMAT_PVRTC_2BPP_RGBA_1;
-fccToFormat[FCC_PVRTC_4BPP_RGB_1] = PIXELFORMAT_PVRTC_4BPP_RGB_1;
-fccToFormat[FCC_PVRTC_4BPP_RGBA_1] = PIXELFORMAT_PVRTC_4BPP_RGBA_1;
-
 /**
- * @private
```
willeastcott, line 85: Just wondering why the private tag is being removed here. `DdsParser` isn't exported as public API. Does this class now appear in the API referecne docs?

PR #3381, opened 2021-08-08
Support browserless environment

slimbuck:
This PR introduces changes to make running in browserless environment easier.

New API:
- added the following members to the global `pc.platform` object:
    - `environment` - either `browser` or `node`
    - `global` - the global object to use in the current environment: `window` on browser and `global` on node
    - `browser` - convenience boolean, true if the runtime environment is browser

In addition, this PR:
- uses the new API to wrap necessary code
- updated image parsers to support loading file contents directly. this is required in node and speeds up gltf parsing when textures are embedded (the engine previously would convert arraybuffer->blob->arraybuffer).
- adds an internal convenience function `Asset.fetchArrayBuffer` and uses it in image load handlers
- simplifies handling of asset variants and extracts `asset.file` into an engine private `AssetFile` class
- fixes HDR file orientation bug
- add `ReadStream` internal class and use it for both HDR and KTX2 parsers


src/asset/asset-file.js
```diff
@@ -0,0 +1,31 @@
...
+ * @classdesc Wraps a source of asset data.
+ */
+class AssetFile {
+    constructor(url, filename, hash, size, opt, contents) {
+        this.url = url || '';
+        this.filename = filename || '';
+        this.hash = hash || null;
+        this.size = size || null;
+        this.opt = opt || null;
+        this.contents = contents || null;
```
willeastcott, line 14: Maybe consider default parameters here?

src/asset/asset-file.js
```diff
@@ -0,0 +1,31 @@
...
+            this.hash === other.hash &&
+            this.size === other.size &&
+            this.opt === other.opt &&
+            this.contents === other.contents;
+    }
+}
+
+export {
+    AssetFile
+};
```
willeastcott, line 31: Super-minor stylistic quibble but all other modules declare export on one line.

willeastcott:
```suggestion
export { AssetFile };
```

src/core/platform.js
```diff
@@ -62,6 +65,34 @@ if (typeof navigator !== 'undefined') {
...
+     * the global object when running in nodejs.
+     */
+    global: (environment === 'browser') ? window : global,
+
+    /**
+     * @static
+     * @readonly
+     * @type {boolean}
+     * @name platform.isBrowser
+     * @description Convenience boolean indicating whether we're running the browser.
```
willeastcott, line 92:
```suggestion
* @description Convenience boolean indicating whether we're running in the browser.
```

src/core/read-stream.js
```diff
@@ -0,0 +1,83 @@
...
+            }
+            result += c;
+        }
+        return result;
+    }
+}
+
+export {
+    ReadStream
+};
```
willeastcott, line 83: export { ReadStream };

src/polyfill/OESVertexArrayObject.js
```diff
@@ -1,5 +1,7 @@
 // https://github.com/KhronosGroup/WebGL/blob/master/sdk/demos/google/resources/OESVertexArrayObject.js
 
+import { platform } from '../core/platform.js';
```
willeastcott, line 3: I'm not too keen on this. Polyfills shouldn't have any dependency on engine code. Could `setupVertexArrayObject` just be called from engine code if we're running in the browser?

src/resources/parser/texture/basis.js
```diff
@@ -20,33 +18,28 @@ class BasisParser {
...
+            const basisModuleFound = basisTranscode(
+                device,
+                url.load,
+                data,
+                callback,
+                { isGGGR: (asset?.file?.variants?.basis?.opt & 8) !== 0 }
+            );
+
+            if (!basisModuleFound) {
+                callback('Basis module not found. Asset "' + asset.name + '" basis texture variant will not be loaded.');
```
willeastcott, line 32:
```suggestion
                callback(`Basis module not found. Asset ${asset.name} basis texture variant will not be loaded.`);
```

src/resources/parser/texture/hdr.js
```diff
@@ -1,54 +1,13 @@
-import { http } from '../../../net/http.js';
+import { Asset } from '../../../asset/asset.js';
+import { Texture } from '../../../graphics/texture.js';
+import { ReadStream } from '../../../core/read-stream.js';
```
willeastcott, line 3: Hmm, I'm thinking we might need to write up some rules about ordering imports. When I originally ported the engine codebase to ES6 modules, the ordering was based on the root folders in src and something like:

core
math
net
graphics
scene
resources
asset
framework

(I've left a few out but you get the idea)

Within each folder group, I just ordered alphabetically. To me, this felt easier to spot circular dependency problems, where a low layer (e.g. graphics) might import from a higher layer (e.g. scene).

And I'd typically put a blank line between each group.

Look at `application.js` for a big example.

In that world, you'd have:

```javascript
import { ReadStream } from '../../../core/read-stream.js';

import { Texture } from '../../../graphics/texture.js';

import { Asset } from '../../../asset/asset.js';
```

willeastcott (approved):
Approving with a bunch of comments.

src/asset/asset-file.js
slimbuck: That wouldn't have the same result. The call site specifies all the constructor's arguments, but some of them may be missing/undefined.

src/polyfill/OESVertexArrayObject.js
slimbuck: Since the engine depends on this polyfill now, we'll need it on both browser and headless. The problem is the function must be registered _somewhere_ for it to be called by the graphics device. The only alternative I can think of is converting this polyfill to a module and importing the function into graphice-device instead?

src/resources/parser/texture/hdr.js
slimbuck: I've updated the order and spacing.

src/polyfill/OESVertexArrayObject.js
willeastcott: Yeah, I think I could live with that. I mean, this doesn't feel like a true polyfill anyway. Maybe this is the different between a shim and a polyfill...

src/asset/asset-file.js
willeastcott: I'm not sure I follow, can you give me an example where it would break?
Also, if you passed in 0 for hash or size (yeah, I'm not sure that exactly makes sense but hey)...you'd get null for both, not 0. Not sure what opt is though or what type it is - but same could apply there.

src/polyfill/OESVertexArrayObject.js
slimbuck: done

src/asset/asset-file.js
slimbuck: I'm not sure if it's better either, but changed hash/size/opt to keep 0's.

For an example, https://github.com/playcanvas/engine/pull/3381/files#diff-59cbf44184755e9db171fb7873b5b92241e4bc05bffd75d001d9e6f738c38700R363 can pass in an undefined argument, which this code deals with, but default arguments wouldn't.

vkalpias (approved):

src/asset/asset-file.js
willeastcott: OK, cool, thanks.

src/asset/asset.js
```diff
@@ -487,6 +436,35 @@ class Asset extends EventHandler {
...
         }
     }
+
+    /**
+     * @private
+     * @function
+     * @name Asset#fetchArrayBuffer
+     * @description Helper function to resolve asset file data and return the contents as an
+     * ArrayBuffer. If the asset file contents are present, that is returned. Otherwise the file
+     * data is be downloaded via http.
```
mvaligursky, line 446: .. the file data is be downloaded ..

src/core/read-stream.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    align(bytes) {
+        this.offset = (this.offset + bytes - 1) & (~(bytes - 1));
+    }
+
+    _inc(amount) {
+        this.offset += amount;
+        return this.offset - amount;
+    }
```
mvaligursky, line 34: not the cleanest implementation and its usage bellow :)

slimbuck: I don't love it either, but how would you suggest improving it?

src/scene/materials/standard-material.js
```diff
@@ -664,39 +661,47 @@ class StandardMaterial extends Material {
                          prefilteredCubeMap8 &&
                          prefilteredCubeMap4;
 
-            if (useDp && allMips) {
+            // no basic extension? likely slow device, attempt tp use dual paraboloid map with spherical harmonic ambient
```
mvaligursky, line 664: tp => to

PR #3391, opened 2021-08-13
StandardMaterial code cleanup

slimbuck:
StandardMaterial has evolved over time to include dead code and hard to follow logic.

This PR updates StandardMaterial to be (hopefully) simpler and easier to maintain.

This PR should introduce no functional change to StandardMaterial, leaving the public interface untouched.


src/scene/materials/standard-material-options-builder.js
```diff
@@ -308,42 +308,34 @@ StandardMaterialOptionsBuilder.prototype._collectLights = function (lType, light
     }
 };
 
+const arraysEqual = (a, b) => {
```
mvaligursky, line 311: see set-utils.js. Maybe we need array-utils.js for this one.

src/scene/materials/standard-material.js
```diff
@@ -382,48 +367,34 @@ class StandardMaterial extends Material {
...
     _clearParameters() {
-        const props = this._propsSet;
-        for (let i = 0; i < props.length; i++) {
-            delete this.parameters[props[i]];
-        }
-        this._propsSet = [];
+        _propsSet.forEach((p) => {
+            delete this.parameters[p];
+        });
+        _propsSet = [];
```
mvaligursky, line 397: maybe just set .length to 0 instead of allocating new array

src/scene/materials/standard-material.js
```diff
@@ -356,21 +349,13 @@ class StandardMaterial extends Material {
...
 
     reset() {
-        for (let i = 0; i < _propsSerial.length; i++) {
-            const defVal = _propsSerialDefaultVal[i];
-            this[_propsSerial[i]] = defVal ? (defVal.clone ? defVal.clone() : defVal) : defVal;
-        }
-        for (let i = 0; i < _propsInternalNull.length; i++) {
-            this[_propsInternalNull[i]] = null;
+        // set default values
+        for (const [name, prop] of Object.entries(_props)) {
```
mvaligursky, line 353: might need a polyfill for IE https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

src/scene/materials/standard-material.js
```diff
@@ -382,48 +367,34 @@ class StandardMaterial extends Material {
...
-                }
-            }
-        }
+        this._cloneInternal(clone);
 
-        return clone;
-    }
+        // set properties
+        Object.keys(_props).forEach((k) => {
+            clone[k] = this[k];
```
mvaligursky, line 375: no need to clone these?

src/scene/materials/standard-material.js
slimbuck: damn, nice catch. I keep expecting babel will sort out this stuff for us. @willeastcott, do you know if it's possible to get babel to handle this for us automatically?

src/scene/materials/standard-material.js
slimbuck: The property set implementation copies values for aggregate types (vec, color etc).

mvaligursky (approved):
looks great! Approving with few comments.

src/scene/materials/standard-material.js
slimbuck: yep good call

src/scene/materials/standard-material.js
slimbuck: using keys.forEach instead

src/scene/materials/standard-material-options-builder.js
slimbuck: TBH I'd love to add a general 'deepCompare' function which handles comparing objects, arrays and simple data types.

willeastcott (approved):

src/scene/materials/standard-material.js
```diff
@@ -356,21 +348,13 @@ class StandardMaterial extends Material {
...
-        for (let i = 0; i < _propsInternalNull.length; i++) {
-            this[_propsInternalNull[i]] = null;
-        }
-        for (let i = 0; i < _propsInternalVec3.length; i++) {
-            this[_propsInternalVec3[i]] = new Float32Array(3);
-        }
-
-        this._chunks = new Chunks();
+        // set default values
+        Object.keys(_props).forEach((name) => {
```
Maksims, line 352: Use of forEach leads to function/scope allocations, once compiled to ES5.
It is way slower than using for loops, and for code that runs on each frame, forEach to be avoided.

Issue #2563, opened 2020-11-17
Add support for glTF extension: KHR_texture_transform


raytranuk:
Add support for glTF extension: https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_texture_transform

Currently failing related tests on: https://github.com/cx20/gltf-test:

![image](https://user-images.githubusercontent.com/11779961/99427238-12ab8680-28fd-11eb-9ad7-f74cc2fe3fd5.png)
![image](https://user-images.githubusercontent.com/11779961/99427346-340c7280-28fd-11eb-8cf8-ceb77f6ef684.png)

Priority of this may change if users request / require this extension to be supported.

slimbuck:
Note that we already support the scale and offset attributes of this extension https://github.com/playcanvas/engine/blob/master/src/resources/parser/glb-parser.js#L923.

cx20:
I tested it on the latest version of PlayCanvas and noticed that the results have changed.
Is this an expected change?

|Library|TextureTransformTest.gltf result|
|:----:|:----:|
|[PlayCanvas v1.44.2](https://rawcdn.githack.com/cx20/gltf-test/ca7c5c1f4be263d67179d25f80c092127eac61d2/examples/playcanvas/index.html?category=tutorialModels&model=TextureTransformTest&scale=1&type=glTF)|[![image](https://user-images.githubusercontent.com/1833979/126215771-ab28d00c-c274-480a-ae31-4cf7ebaf0acf.png)](https://rawcdn.githack.com/cx20/gltf-test/ca7c5c1f4be263d67179d25f80c092127eac61d2/examples/playcanvas/index.html?category=tutorialModels&model=TextureTransformTest&scale=1&type=glTF)|
|[PlayCanvas v1.45.0](https://rawcdn.githack.com/cx20/gltf-test/0d6cee5b5faecf8a2ff546bb6446ab1268ab632c/examples/playcanvas/index.html?category=tutorialModels&model=TextureTransformTest&scale=1&type=glTF)|[![image](https://user-images.githubusercontent.com/1833979/126215806-007da6e6-bbe5-4c13-a9fa-2631e7ae9a83.png)](https://rawcdn.githack.com/cx20/gltf-test/0d6cee5b5faecf8a2ff546bb6446ab1268ab632c/examples/playcanvas/index.html?category=tutorialModels&model=TextureTransformTest&scale=1&type=glTF)|

cx20:
[...]

PR #3394, opened 2021-08-16
Support KHR_texture_transform

slimbuck:
This PR:
- adds `Mat3.mul2` function
- adds a `rotation` property to all the maps supported in `StandardMaterial`
- uses this to implement the full KHR_texture_transform gltf extension

Old:
chrome

New:
chrome

Fixes https://github.com/playcanvas/engine/issues/2563

### DIFF
--- a/src/graphics/program-lib/programs/standard.js
+++ b/src/graphics/program-lib/programs/standard.js
@@ -236,10 +236,9 @@ var standard = {
 
     _setMapTransform: function (codes, name, id, uv) {
         const varName = `texture_${name}MapTransform`;
+        const checkId = id + uv * 100;
 
         codes[0] += `uniform mat3 ${varName};\n`;
-
-        var checkId = id + uv * 100;
         if (!codes[3][checkId]) {
             codes[1] += `varying vec2 vUV${uv}_${id};\n`;
             codes[2] += `   vUV${uv}_${id} = (${varName} * vec3(uv${uv}, 1)).xy;\n`;
--- a/src/graphics/program-lib/programs/standard.js
+++ b/src/graphics/program-lib/programs/standard.js
@@ -238,10 +238,12 @@ var standard = {
         const varName = `texture_${name}MapTransform`;
         const checkId = id + uv * 100;
 
-        codes[0] += `uniform mat3 ${varName};\n`;
+        // upload a 3x2 matrix and manually perform the multiplication
+        codes[0] += `uniform vec3 ${varName}0;\n`;
+        codes[0] += `uniform vec3 ${varName}1;\n`;
         if (!codes[3][checkId]) {
             codes[1] += `varying vec2 vUV${uv}_${id};\n`;
-            codes[2] += `   vUV${uv}_${id} = (${varName} * vec3(uv${uv}, 1)).xy;\n`;
+            codes[2] += `   vUV${uv}_${id} = vec2(dot(uv${uv}, ${varName}0.xy) + ${varName}0.z, dot(uv${uv}, ${varName}1.xy) + ${varName}1.z).xy;\n`;
             codes[3][checkId] = true;
         }
         return codes;
--- a/src/graphics/program-lib/programs/standard.js
+++ b/src/graphics/program-lib/programs/standard.js
@@ -243,7 +243,7 @@ var standard = {
         codes[0] += `uniform vec3 ${varName}1;\n`;
         if (!codes[3][checkId]) {
             codes[1] += `varying vec2 vUV${uv}_${id};\n`;
-            codes[2] += `   vUV${uv}_${id} = vec2(dot(uv${uv}, ${varName}0.xy) + ${varName}0.z, dot(uv${uv}, ${varName}1.xy) + ${varName}1.z).xy;\n`;
+            codes[2] += `   vUV${uv}_${id} = vec2(dot(vec3(uv${uv}, 1), ${varName}0), dot(vec3(uv${uv}, 1), ${varName}1));\n`;
             codes[3][checkId] = true;
         }
         return codes;
--- a/src/graphics/program-lib/programs/standard.js
+++ b/src/graphics/program-lib/programs/standard.js
@@ -235,12 +235,14 @@ var standard = {
     },
 
     _setMapTransform: function (codes, name, id, uv) {
-        codes[0] += "uniform vec4 texture_" + name + "MapTransform;\n";
+        const varName = `texture_${name}MapTransform`;
+
+        codes[0] += `uniform mat3 ${varName};\n`;
 
         var checkId = id + uv * 100;
         if (!codes[3][checkId]) {
-            codes[1] += "varying vec2 vUV" + uv + "_" + id + ";\n";
-            codes[2] += "   vUV" + uv + "_" + id + " = uv" + uv + " * texture_" + name + "MapTransform.xy + texture_" + name + "MapTransform.zw;\n";
+            codes[1] += `varying vec2 vUV${uv}_${id};\n`;
+            codes[2] += `   vUV${uv}_${id} = (${varName} * vec3(uv${uv}, 1)).xy;\n`;
             codes[3][checkId] = true;
         }
         return codes;
--- a/src/math/mat3.js
+++ b/src/math/mat3.js
@@ -135,56 +135,6 @@ class Mat3 {
 
     /**
      * @function
-     * @name Mat3#mul2
-     * @description Multiplies the specified 3x3 matrices together and stores the result in
-     * the current instance.
-     * @param {Mat3} lhs - The 3x3 matrix used as the first multiplicand of the operation.
-     * @param {Mat3} rhs - The 3x3 matrix used as the second multiplicand of the operation.
-     * @returns {Mat3} Self for chaining.
-     */
-    mul2(lhs, rhs) {
-        const a = lhs.data;
-        const b = rhs.data;
-        const r = this.data;
-
-        const a00 = a[0];
-        const a01 = a[1];
-        const a02 = a[2];
-        const a10 = a[3];
-        const a11 = a[4];
-        const a12 = a[5];
-        const a20 = a[6];
-        const a21 = a[7];
-        const a22 = a[8];
-
-        let b0, b1, b2;
-
-        b0 = b[0];
-        b1 = b[1];
-        b2 = b[2];
-        r[0] = a00 * b0 + a10 * b1 + a20 * b2;
-        r[1] = a01 * b0 + a11 * b1 + a21 * b2;
-        r[2] = a02 * b0 + a12 * b1 + a22 * b2;
-
-        b0 = b[3];
-        b1 = b[4];
-        b2 = b[5];
-        r[3] = a00 * b0 + a10 * b1 + a20 * b2;
-        r[4] = a01 * b0 + a11 * b1 + a21 * b2;
-        r[5] = a02 * b0 + a12 * b1 + a22 * b2;
-
-        b0 = b[6];
-        b1 = b[7];
-        b2 = b[8];
-        r[6] = a00 * b0 + a10 * b1 + a20 * b2;
-        r[7] = a01 * b0 + a11 * b1 + a21 * b2;
-        r[8] = a02 * b0 + a12 * b1 + a22 * b2;
-
-        return this;
-    }
-
-    /**
-     * @function
      * @name Mat3#setIdentity
      * @description Sets the matrix to the identity matrix.
      * @returns {Mat3} Self for chaining.
--- a/src/math/mat3.js
+++ b/src/math/mat3.js
@@ -135,6 +135,56 @@ class Mat3 {
 
     /**
      * @function
+     * @name Mat3#mul2
+     * @description Multiplies the specified 3x3 matrices together and stores the result in
+     * the current instance.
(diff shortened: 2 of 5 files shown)


src/resources/parser/glb-parser.js
```diff
@@ -958,28 +955,21 @@ const createMaterial = function (gltfMaterial, textures, disableFlipV) {
...
-        if (extensions) {
-            const textureTransformData = extensions.KHR_texture_transform;
-            if (textureTransformData) {
-                if (textureTransformData.scale) {
-                    scale = textureTransformData.scale;
-                }
-                if (textureTransformData.offset) {
-                    offset = textureTransformData.offset;
-                }
+        const textureTransform = source?.extensions?.KHR_texture_transform;
```
willeastcott, line 958: You don't need the first ? because `source` is valid.

slimbuck: thanks!

willeastcott:
From an Editor perspective, I'm assuming we'll just add another number entry for rotation.

slimbuck:
> This is interesting. I assumed we would have EITHER:
> 
> ```js
> StandardMaterial#diffuseMapTiling
> StandardMaterial#diffuseMapOffset
> StandardMaterial#diffuseMapRotation
> ```
> 
> OR:
> 
> ```js
> StandardMaterial#diffuseMapTransform
> StandardMaterial#diffuseMapTiling (deprecated)
> StandardMaterial#diffuseMapOffset (deprecated)
> ```
> 
> With my preference going to the former option.

We can do it this way. It would mean we don't have to allocate a 3x3 matrix for each map. It will also mean that even though we are using a general 3x3 texture transform, as a user you can't do anything fancy with it.

Another slightly messy detail is that the exiting offset/tiling value is relative to the old bottom/left texture origin and a conversion is necessary between gltf offset/tiling and ours.

willeastcott:
In that case, we can alias. e.g.:

```
diffuseMapTiling -> diffuseMapScale
diffuseMapOffset -> diffuseMapTranslation 
```

And add `diffuseMapRotation` of course.

Any flipping can be done in the deprecated getters/setters.

src/graphics/program-lib/programs/standard.js
```diff
@@ -235,12 +235,13 @@ var standard = {
     },
 
     _setMapTransform: function (codes, name, id, uv) {
-        codes[0] += "uniform vec4 texture_" + name + "MapTransform;\n";
+        const varName = `texture_${name}MapTransform`;
+        const checkId = id + uv * 100;
 
-        var checkId = id + uv * 100;
+        codes[0] += `uniform mat3 ${varName};\n`;
```
mvaligursky, line 241: I'd either try to store this as Vec3[2], or even simpler as two Vec3s, without an array.

src/resources/parser/glb-parser.js
```diff
@@ -958,28 +955,18 @@ const createMaterial = function (gltfMaterial, textures, disableFlipV) {
             }
         }
 
-        let scale = uvONE;
-        let offset = uvZERO;
+        const textureTransform = source.extensions?.KHR_texture_transform;
+        if (textureTransform) {
+            const offset = textureTransform.offset || [0, 0];
+            const scale = textureTransform.scale || [1, 1];
+            const rotation = -(textureTransform.rotation || 0);
```
mvaligursky, line 962: the minus does not need to be applied on default 0.

src/resources/parser/glb-parser.js
```diff
@@ -958,28 +955,18 @@ const createMaterial = function (gltfMaterial, textures, disableFlipV) {
...
-            const textureTransformData = extensions.KHR_texture_transform;
-            if (textureTransformData) {
-                if (textureTransformData.scale) {
-                    scale = textureTransformData.scale;
-                }
-                if (textureTransformData.offset) {
-                    offset = textureTransformData.offset;
-                }
+            for (map = 0; map < maps.length; ++map) {
+                material[`${maps[map]}MapTiling`] = new Vec2(scale[0], scale[1]);
```
mvaligursky, line 965: I think you mentioned the material just copies values to its internal Vec2's and similar. Maybe we should not new all these Vec2's here, but create Vec2's above where arrays of their values are created?

```
const scale = textureTransform.scale ? new Vec2(textureTransform.scale.x, textureTransform.scale.y) : Vec2.ONE;
...
material[`${maps[map]}MapTiling`] = scale;
```

src/scene/materials/standard-material.js
```diff
@@ -47,6 +47,7 @@ const _propsSet = [];
  * @property {number} diffuseMapUv Main (primary) diffuse map UV channel.
  * @property {Vec2} diffuseMapTiling Controls the 2D tiling of the main (primary) diffuse map.
  * @property {Vec2} diffuseMapOffset Controls the 2D offset of the main (primary) diffuse map. Each component is between 0 and 1.
+ * @property {number} diffuseMapRotation Controls the 2D rotation of the main (primary) diffuse map.
```
mvaligursky, line 50: It should specify degrees vs radians as a unit.

mvaligursky: ideally it would be degrees I suspect, to be user friendly and consistent with angles in Quat.

src/scene/materials/standard-material.js
```diff
@@ -768,19 +788,25 @@ function _defineTex2D(name, uv, channels, defChannel, vertexColor, detailMode) {
...
             return null;
         }
 
-        const uniform = material._allocUniform(mapTransform, () => new Float32Array(4));
-        uniform[0] = tiling.x;
-        uniform[1] = tiling.y;
-        uniform[2] = offset.x;
-        uniform[3] = 1.0 - tiling.y - offset.y;
+        const uniform = material._allocUniform(mapTransform, () => new Float32Array(9));
+        uniform.set([
```
mvaligursky, line 805: don't create temp array here

mvaligursky: also, evaluate sin and cos only one time instead of twice

mvaligursky: The other option would be to store 5 uniforms. 4 as currently for offset and scale, and a single for rotation. And generate sin/cos and all that on the GPU. But that's probably not worth the effort.

src/math/mat3.js
```diff
@@ -133,6 +133,56 @@ class Mat3 {
...
+        b0 = b[6];
+        b1 = b[7];
+        b2 = b[8];
+        r[6] = a00 * b0 + a10 * b1 + a20 * b2;
+        r[7] = a01 * b0 + a11 * b1 + a21 * b2;
+        r[8] = a02 * b0 + a12 * b1 + a22 * b2;
+
+        return this;
+    }
+
```
willeastcott, line 185: Is `Mat3#mul2` actually needed by this PR? I'm not against adding it. But if this PR doesn't require it, I'd remove it for simplicity.

src/scene/materials/standard-material.js
```diff
@@ -768,19 +788,25 @@ function _defineTex2D(name, uv, channels, defChannel, vertexColor, detailMode) {
...
+            Math.cos(rotation) * tiling.x, Math.sin(rotation) * tiling.x, 0,
+            -Math.sin(rotation) * tiling.y, Math.cos(rotation) * tiling.y, 0,
+            offset.x, 1.0 - tiling.y - offset.y, 1
```
mvaligursky, line 808: I'm not sure this math is correct. You need to build scale and rotation matrix and multiply them together, which is different than multiply only matching elements of the mates. But I agree it'd be good to write optimized version. Perhaps we need Mat3.setTRS, similarly to Mat4.setTRS

src/scene/materials/standard-material.js
willeastcott: I'm not sure why `set` for matrix classes was written to take an array as a parameter. It should just take the individual elements IMHO, just like for vectors. And a `setFromArray` function could be added for the case where somebody wants to pass data that way.

mvaligursky: I think this set is Float32Array.set, which takes array. We should set individual entries of the array without the extra temp array.

src/graphics/program-lib/programs/standard.js
slimbuck: I'm not convinced optimising strictly for number of float uniforms is an overall win (at the cost of additional CPU work, for example), but I've done it that way for now.

src/math/mat3.js
slimbuck: no, I've removed it.

src/scene/materials/standard-material.js
slimbuck: The math is correct.

src/scene/materials/standard-material.js
slimbuck: UI should be concerned with presenting human-readable values to user, not the internal database. I've gone with degrees, but I'm not convinced that's the right call. (For example glTF stores radians). @willeastcott any opinion?

slimbuck: Yes I see quaternion `getEulerAngles` returns degrees. This is not great. The last line of the function does `.mulScalar(math.RAD_TO_DEG)` - a completely unnecessary operation if the returned values are used to perform more math (which would require another `mulScalar(DEG_TO_RAD)` before being useful).

So I guess in the name of consistency we should keep degrees here.

willeastcott: All other APIs in the engine use degrees. So, like it or not, for consistency, we should go with degrees. If we want to adopt radians across the engine, that's another conversation.

src/graphics/program-lib/programs/standard.js
```diff
@@ -235,12 +235,15 @@ var standard = {
...
 
-        var checkId = id + uv * 100;
+        // upload a 3x2 matrix and manually perform the multiplication
+        codes[0] += `uniform vec3 ${varName}0;\n`;
+        codes[0] += `uniform vec3 ${varName}1;\n`;
         if (!codes[3][checkId]) {
-            codes[1] += "varying vec2 vUV" + uv + "_" + id + ";\n";
-            codes[2] += "   vUV" + uv + "_" + id + " = uv" + uv + " * texture_" + name + "MapTransform.xy + texture_" + name + "MapTransform.zw;\n";
+            codes[1] += `varying vec2 vUV${uv}_${id};\n`;
+            codes[2] += `   vUV${uv}_${id} = vec2(dot(uv${uv}, ${varName}0.xy) + ${varName}0.z, dot(uv${uv}, ${varName}1.xy) + ${varName}1.z).xy;\n`;
```
mvaligursky, line 246: it would probably save an instruction to use xyz in dot 3 and avoid the addition - like you had before with the matrix.
so instead of 
`dot(uv, var.xy) + var.z` you should do
`dot(vec3(uv, 1.0), var);`

willeastcott (approved):

src/resources/parser/glb-parser.js
```diff
@@ -958,27 +955,20 @@ const createMaterial = function (gltfMaterial, textures, disableFlipV) {
...
-                if (textureTransformData.scale) {
-                    scale = textureTransformData.scale;
-                }
-                if (textureTransformData.offset) {
-                    offset = textureTransformData.offset;
-                }
-            }
-        }
+            const tilingVec = new Vec2(scale[0], scale[1]);
+            const offsetVec = new Vec2(offset[0], 1.0 - scale[1] - offset[1]);
```
mvaligursky, line 965: I would combine these together. There's no need to create arrays to later convert them to Vec2. Go to Vec2 directly.

mvaligursky (approved):
two small comments, but looks great otherwise.

src/graphics/program-lib/programs/standard.js
mvaligursky: actually, not sure it will save anything at all. As vec3(uv, 1.0) will likely generate two move instructions, being the same as your code at the end.

src/resources/parser/glb-parser.js
slimbuck: I wouldn't get too obsessed with extreme optimisation of loading code that is called maybe twice during the loading of a glTF file. Being code concise is much more important in this case.

Please offer a code suggestion though, perhaps I'm missing something.

mvaligursky: Agreed. I offered a code before for this section. Something like:
`const tilingVec = textureTransform.offset ? new Vec2(textureTransform.offset[0], textureTransform.offset[1]) : Vec2.ZERO;`

slimbuck: And for offset?

mvaligursky: hmm, same thing?
`const offsetVec = textureTransform.offset ? new Vec2(textureTransform.offset.x, 1.0 - textureTransform.offset[1] - tilingVec.y) : Vec2.ONE;`

src/graphics/program-lib/programs/standard.js
slimbuck: I don't think there is an instruction difference, but the resulting code is slightly more readable so I've made that change, thanks.

src/resources/parser/glb-parser.js
slimbuck: I'm using `Vec2.ZERO` and `Vec2.ONE` instead of inline arrays, which I think gives us both what we want.

mvaligursky:
Nice, all good now, merge away :)

PR #3408, opened 2021-08-17
Fix enabled flag on camea and light component

slimbuck:
This PR:
- propagates the enabled flag on the camera component when cloning
- set the disabled flag on lights created during gltf load

### DIFF
--- a/src/framework/components/camera/system.js
+++ b/src/framework/components/camera/system.js
@@ -107,6 +107,7 @@ class CameraComponentSystem extends ComponentSystem {
             clearDepthBuffer: c.clearDepthBuffer,
             clearStencilBuffer: c.clearStencilBuffer,
             cullFaces: c.cullFaces,
+            enabled: c.enabled,
             farClip: c.farClip,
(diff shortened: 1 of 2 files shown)


willeastcott (approved):

src/framework/components/camera/system.js
```diff
@@ -107,6 +107,7 @@ class CameraComponentSystem extends ComponentSystem {
             clearDepthBuffer: c.clearDepthBuffer,
             clearStencilBuffer: c.clearStencilBuffer,
             cullFaces: c.cullFaces,
+            enabled: c.enabled,
```
willeastcott, line 110: We probably need to do a full audit of all components to ensure they all correctly clone the `enabled` property. It looks like the `AnimComponent` doesn't clone it, for example.

mvaligursky (approved):

src/framework/components/camera/system.js
slimbuck: https://github.com/playcanvas/engine/issues/3413

PR #3414, opened 2021-08-18
WIP - handling of morph targets and animations stored as quantized in GLB

mvaligursky:
- animation seems to work
- morph targets don't seem to work. The idea here is to convert morph target data to float format and apply normalization during load. The data is later stored in vertex data as floats (only very old platforms), or stored in half / float textures (most platforms), so in most cases we need float inputs anyways.

spec: https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_mesh_quantization/README.md

references: gltf-cample-viewer: https://github.com/KhronosGroup/glTF-Sample-Viewer/tree/master/source/gltf

### DIFF
--- a/src/resources/parser/glb-parser.js
+++ b/src/resources/parser/glb-parser.js
@@ -110,6 +110,16 @@ const getComponentType = function (componentType) {
     }
 };
 
+const getComponentNormalizationScale = function (componentType) {
+    switch (componentType) {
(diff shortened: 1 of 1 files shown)


mvaligursky:
closing in favor of https://github.com/playcanvas/engine/pull/3431

PR #3417, opened 2021-08-19
Invoke equals on internal property

slimbuck:
Due to changes in #3391 a forum user encountered the following:

https://forum.playcanvas.com/t/error-on-previously-working-material-offset-tiling-manipulation/21521

That call should have worked and this PR addresses the issue by calling the `.equals` function on the internal object instead of the one passed in.


mvaligursky (approved):
feels like this should not do anything? :)

Issue #1930, opened 2020-03-23
Weird texture .exr .hdr artifacts on IOS device


TrongHieu90:
The IOS test device is Iphone 8.

### Description
The texture shows some errornous banding artifacts on .exr lightmap texture. It is normal on Android device (Samsung s9) as well as desktop.

Tested on all popular browser (Firefox, Chrome, Safari on Iphone8), (Firefox, Chrome on Samsung S9)

It is normal on MAC pc.

It also happens with the official Orange Room project, which uses .hdr file for lightmap texture.

![5af6e00afc6502b42cb4aa04bac421eb4a6512e7_2_1035x582](https://user-images.githubusercontent.com/20443445/77332015-11b55b80-6d54-11ea-9e73-72ecb2095646.jpeg)

Topic on the forum with other user confirmation: 
https://forum.playcanvas.com/t/weird-texture-artifacts-when-open-with-ios-iphone-8/12374

### Steps to Reproduce

1. Use .exr, hdr file for lightmap texture
2. View the texture in IOS device

slimbuck:
I haven’t been able to repro on iPhone 7:

![54F1D4DF-C423-4FB6-8E42-8AE7273DADF2](https://user-images.githubusercontent.com/11276292/82357994-c3dc6d00-99fd-11ea-81bf-ac9d52d52bb5.png)

(Using the link in the forum post)

slimbuck:
(Though the orange room example does have lightmap banding on the same device)

willeastcott:
It's fine on iPhone XS with iOS 13.4.1 too:

![Image from iOS (34)](https://user-images.githubusercontent.com/697563/82374724-5dfcdf00-9a17-11ea-9c0b-324899816384.png)

TrongHieu90:
@slimbuck , @willeastcott 
Its fine now because I fixed it (by using a HDR plugin of Photoshop to compress the 32 bit.exr to 8 bit png file)

I found out that Iphone does not support WebGL 2 (on all of the browsers), which will lead to this issue. Even with dithering setting when baking the lightmap does not fix it.

The orange room sample by playcanvas will display this issue the clearest.

slimbuck:
[...]

PR #3419, opened 2021-08-20
Png loader example

slimbuck:
This PR adds to the engine an example texture parser for loading PNGs manually and adds an accompanying example.

Manual loading of PNG is useful under iOS which still tramples on texture alpha (see #1930).

The example is completely static, just demonstrating a lightmap without banding and probably needs more work before merging:

Screenshot 2021-08-20 at 16 57 56


examples/src/examples/loaders/png.tsx
```diff
@@ -0,0 +1,54 @@
+import * as pc from 'playcanvas/build/playcanvas.js';
+import Example from '../../app/example';
+
+class OBJExample extends Example {
```
mvaligursky, line 4: rename the class

yaustar:
For editor only projects, this is only possible when self hosting as the developer would have to hook in the parser between the application creation and config load function call in `__start__.js`

yaustar:
Example of adding the upng loader to a self hosted solution:

Add `upng.js` to the project folder:
image

Include it in the HTML
image

Add the parser to the texture loader in `__start__.js`
```
try {
        app = new pc.Application(canvas, {
            elementInput: devices.elementInput,
            keyboard: devices.keyboard,
            mouse: devices.mouse,
            gamepads: devices.gamepads,
            touch: devices.touch,
            graphicsDeviceOptions: window.CONTEXT_OPTIONS,
            assetPrefix: window.ASSET_PREFIX || "",
            scriptPrefix: window.SCRIPT_PREFIX || "",
            scriptsOrder: window.SCRIPTS || []
        });

        app.loader.getHandler("texture").parsers.png = new PngParser(app.graphicsDevice);
    }
```

scripts/parsers/upng.js
```diff
@@ -0,0 +1,449 @@
...
+    this._device = device;
+}
+
+Object.assign(PngParser.prototype, {
+    load: function (url, callback, asset) {
+        pc.Asset.fetchArrayBuffer(url.load, callback, asset, this.maxRetries);
+    },
+
+    open: function (url, data, device) {
+        const img = UPNG.decode(data);
```
willeastcott, line 439: Can you just do this JS decode if the rgbm flag is set? And use the browser PNG loader otherwise?

yaustar:
Just thought of a way to patch it into the engine: https://playcanvas.com/project/870259/overview/added-upng

This adds the upng to the project and an after engine patch on the addHandler function that changes the PNG parser

Issue #2802, opened 2021-01-30
Blendshapes (morphs) exported from third-party .glb exporters are not importing correctly


GrimOrk:
I had exported one .fbx file with different .glb exporters: Playcanvas converter (through the Editor), FBX2glTF (https://github.com/facebookincubator/FBX2glTF), and Blender 2.82.
The only one working correctly is by the PlayCanvas exporter. Blendshapes from all other files are importing, but they are blank, do not affect the mesh.

The structure of those files is different. 
GLB by PC converter:
`{
			"type": "VEC3",
			"componentType": 5126,
			"count": 1996,
			"sparse": {
				"count": 166,
				"values": {
					"bufferView": 68
				},
				"indices": {
					"bufferView": 66,
					"componentType": 5125
				}
			}
		},`

GLB by FBX2glTF:
`{
"componentType": 5126,
"type": "VEC3",
"count": 1996,
"bufferView": 12,
"byteOffset": 0,
"min": [
-0.0135115785524249,
-0.0042192074470222,
-0.0074337050318718
],
"max": [
0.0135115832090378,
0.0130863953381777,
0.00405951496213675
],
"name": "Face.M_F00_000_00_Fcl_ALL_Fun-3854"
},`

Files to reproduce:
[PCConverter](https://vrbaka.site/files/models/from_playcanv.glb)
[FBX2glTF](https://vrbaka.site/files/models/fbx2gltf_test.glb)

mvaligursky:
relevant forum post
https://forum.playcanvas.com/t/suggestion-fbx-to-glb-offline-converter/17216/4

slimbuck:
The FBX2glTF blendshapes seem to work in https://gltf.insimo.com/ but not https://gltf-viewer.donmccurdy.com/ or https://sandbox.babylonjs.com/. The PlayCanvas model produces validation errors we should address:
![Screenshot 2021-02-01 at 09 42 01](https://user-images.githubusercontent.com/11276292/106441302-c985c000-6471-11eb-977a-bea4c003bf0d.png)

GrimOrk:
The FBX2glTF blendshpes are working in https://gltf-viewer.donmccurdy.com/ But because of inverted normals it is hard to see it. Try to rotate the camera around the face if you want to check it. At https://sandbox.babylonjs.com/ blendshapes are blank indeed.

mvaligursky:
@GrimOrk - are you able to test your glb files with the latest engine / viewer?

[...]

Issue #2630, opened 2020-12-18
Support EXT_meshopt_compression GLB extension


mvaligursky:
extension specification:
https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Vendor/EXT_meshopt_compression

library
https://github.com/zeux/meshoptimizer

this should work alongside already supported draco compression to allow engine to load both compressed mesh formats.

mvaligursky:
I've done some investigation, here's some findings:

1. We have implementation of **EXT_meshopt_compression** in the viewer: 
https://github.com/playcanvas/playcanvas-viewer/blob/72e23121ea4d83f60a32d83d13abf3b474be649b/src/viewer.ts#L612-L637

2. we have partial implementation of **KHR_texture_transform**, see here: #2563

3. we do not have implementation of **KHR_mesh_quantization**, see here: #2636

4. Size and performance. I compared uncompressed, draco compressed and meshopt compressed glbs, and the results are very promising. 
Size: draco is smallest, with meshopt being close second, uncompressed being distant third.
Loading speed: draco is slowest (5x time slower than uncompressed), meshopt is fastest (usually about 20% faster than uncompressed).

Table shows loading times (in ms, first line) and size (second line) for test glbs.

| mesh    |		draco 	|		meshOpt 		|		uncompressed |
| --- | --- |--- |--- |
| bed 	|	655			|		85 				|		105 |
|		|	2.1MB		|		3.1MB			|		24MB |
| bitmoji	|	78 (no morph) |	113 (broken morph)	|	150 |
|		|	3MB		|	0.2MB 				|		4MB |
| ChocoBunny 	 | 37		|	6					|			16 |
|		|	48k			|	91k					|		465k |
| Lion	|	70			|	23					|		26 |
|		|	1.5MB		|	1.6MB				|		2.9M |
| Iris		|      36			|	8					|		19 |
|		|	105k		|	54k					|		450k |
| Heart	|	7			|	6					|		6 |
|		|	4.7k			|	5.6k					|		15.4k |

What needs to be done to fully support this:

1. move EXT_meshopt_compression support into the engine directly. The integration code from viewer is small. 
The decoder itself (https://github.com/playcanvas/playcanvas-viewer/blob/master/lib/meshopt_decoder.js) is 26kb uncompressed, this can be optionally added to project when user requires this extension.

[...]

PR #3431, opened 2021-08-30
Fixes for loading quantized glTF data

### LINKED ISSUE #2636: Implement support for KHR_mesh_quantization extension for GLB
Specification: https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_mesh_quantization

This extension is needed to handle all cases of EXT_meshopt_compression (#2630), especially a case with morph targets, where morph attributes are specified in quantized format.

slimbuck:
Fixes #2636 #2630 #2802 (though not able to verify this one)

The glTF parser was not always handling bufferview `byteStride` correctly, especially for quantized animation data and quantized morph target data.

In summary, this PR:
- unpacks strided bufferView data except when handling interleaved vertex buffer data
- dequantizes data for animation curves and morph target normals

### DIFF
--- a/src/resources/parser/glb-parser.js
+++ b/src/resources/parser/glb-parser.js
@@ -207,6 +207,44 @@ const getAccessorData = function (gltfAccessor, bufferViews) {
     return result;
 };
 
+// unpack an array of (quantized) data to Float32Array
+const unpackDataFloat32 = function (srcTypedData, normalized) {
+    const getConvFunc = () => {
+        if (normalized) {
+            // see https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_mesh_quantization#encoding-quantized-data
+            if (srcTypedData instanceof Uint8Array) {
+                return (x) => x / 255.0;
+            } else if (srcTypedData instanceof Int8Array) {
+                return (x) => Math.max(x / 127.0, -1.0);
+            } else if (srcTypedData instanceof Int16Array) {
+                return (x) => Math.max(x / 32767.0, -1.0);
+            } else if (srcTypedData instanceof Uint16Array) {
+                return (x) => x / 65535.0;
+            }
+            // #if _DEBUG
+            console.warn("Unsupported quantized data encountered.");
+            // #endif
+        }
+        return (x) => x;
+    };
+
+    const len = srcTypedData.length;
+    const result = new Float32Array(len);
+    const convFunc = getConvFunc();
+
+    for (let i = 0; i < len; ++i) {
+        result[i] = convFunc(srcTypedData[i]);
+    }
+
+    return result;
+};
+
+// get accessor data in Float32 format
+const getAccessorDataFloat32 = function (gltfAccessor, bufferViews) {
+    const data = getAccessorData(gltfAccessor, bufferViews);
+    return (data instanceof Float32Array) ? data : unpackDataFloat32(data, gltfAccessor.normalized);
+};
+
 const getPrimitiveType = function (primitive) {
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

src/resources/parser/glb-parser.js
```diff
@@ -816,7 +871,7 @@ const createMesh = function (device, gltfMesh, accessors, bufferViews, callback,
 
                     if (target.hasOwnProperty('POSITION')) {
                         accessor = accessors[target.POSITION];
-                        options.deltaPositions = getAccessorData(accessor, bufferViews);
+                        options.deltaPositions = getAccessorData(accessor, bufferViews, true);
```
mvaligursky, line 874: morph positions need to be float format as well, it should use getAccessorDataFloat32

src/resources/parser/glb-parser.js
```diff
@@ -147,13 +147,47 @@ const gltfToEngineSemanticMap = {
...
+                return (x) => x / 255.0;
+            } else if (srcTypedData instanceof Int8Array) {
+                return (x) => Math.max(x / 127.0, -1.0);
+            } else if (srcTypedData instanceof Int16Array) {
+                return (x) => Math.max(x / 32767.0, -1.0);
+            } else if (srcTypedData instanceof Uint16Array) {
+                return (x) => x / 65535.0;
+            }
+            // #if _DEBUG
+            console.warn("Unsupported quantized data encountered.");
```
mvaligursky, line 165: this warning would be printed per vertex, that's excessive.

src/resources/parser/glb-parser.js
slimbuck: I actually found that quantized positions worked fine under texture-based morphing, but not vertex based. I was going to submit a ticket to address this, but thinking about it a little more I suspect normalized quantized positions probably won't work, even with texture based. In that case probably converting to float32 is also necessary.

So yeah, unfortunately we may have to dequantize everything upfront for now. Implementing quantized/normlized support in all mprphing render paths will be a big job (and maintenance nightmare).

src/resources/parser/glb-parser.js
```diff
@@ -816,7 +871,7 @@ const createMesh = function (device, gltfMesh, accessors, bufferViews, callback,
 
                     if (target.hasOwnProperty('POSITION')) {
                         accessor = accessors[target.POSITION];
-                        options.deltaPositions = getAccessorData(accessor, bufferViews);
+                        options.deltaPositions = getAccessorData(accessor, bufferViews, true);
                         options.deltaPositionsType = getComponentType(accessor.componentType);
                         if (accessor.hasOwnProperty('min') && accessor.hasOwnProperty('max')) {
                             options.aabb = new BoundingBox();
```
mvaligursky, line 877: when looking at other engines, I've noticed they dequantize min and max bounds of both mesh and morph targets - perhaps that should be added as well.

slimbuck: thanks good point. done.

slimbuck:
I have also added code to correctly calculate bone bounds when position data is normalized. This addresses the issue in https://github.com/playcanvas/playcanvas-viewer/issues/125.

slimbuck:
> maybe we need a shared function that does this in both cases?

thought the same, but wasn't sure where such a function would go. any suggestions?

mvaligursky:
> > maybe we need a shared function that does this in both cases?
> 
> thought the same, but wasn't sure where such a function would go. any suggestions?

maybe VertexFormat, or VertexAccessor?

mvaligursky (approved):

PR #3433, opened 2021-08-30
Fix copy-paste in matrix4 tests

olkotov:
I think this is "copypasta" 😅.

Fixes #


yaustar:
What's more worrying here is that the test passes no matter what value is used 🤔

Issue #3421, opened 2021-08-23
render component mesh instances not removed from the layers when reparenting


mvaligursky:
- if I use instantiateRenderEntity(), then removing the parent entity from the tree results in the model still being rendered (even though the entity is removed as child of the parent)
- this works as expected when I use instantiateModelEntity()
- seems like the meshInstances aren't being removed from the scene

PR #3436, opened 2021-08-31
When entity is reparented, whole child hierarchy is notified

mvaligursky:
When a hierarchy of entities was parented to an entity, only the root node of the hierarchy was notified about this, using 'insert' or 'remove' event. 
This PR changes this to notify all children of reparented hierarchy.

This allows model and render components that are part of the reparented hierarchy to correctly add or remove their MeshInstances from layers.

Fixes https://github.com/playcanvas/engine/issues/3421

### DIFF
--- a/src/scene/graph-node.js
+++ b/src/scene/graph-node.js
@@ -1099,6 +1099,17 @@ class GraphNode extends EventHandler {
     }
     // #endif
 
+    // fires an event on all children of the node
+    _fireOnHierarchy(name) {
+        if (this.fire) {
+            this.fire(name, this);
+        }
+
+        for (let i = 0; i < this._children.length; i++) {
+            this._children[i]._fireOnHierarchy(name);
+        }
+    }
+
     _onInsertChild(node) {
         node._parent = this;
 
@@ -1124,8 +1135,8 @@ class GraphNode extends EventHandler {
         if (this._frozen)
             node._unfreezeParentToRoot();
 
-        // alert an entity that it has been inserted
-        if (node.fire) node.fire('insert', this);
+        // alert an entity hierarchy that it has been inserted
+        node._fireOnHierarchy('insert');
 
         // alert the parent that it has had a child inserted
         if (this.fire) this.fire('childinsert', node);
@@ -1164,8 +1175,8 @@ class GraphNode extends EventHandler {
                 // Clear parent
                 child._parent = null;
 
-                // alert child that it has been removed
-                if (child.fire) child.fire('remove', this);
+                // alert children that they has been removed
+                child._fireOnHierarchy('remove');
 
                 // alert the parent that it has had a child removed
                 if (this.fire) this.fire('childremove', child);
--- a/src/scene/graph-node.js
+++ b/src/scene/graph-node.js
@@ -1101,10 +1101,7 @@ class GraphNode extends EventHandler {
 
     // fires an event on all children of the node
     _fireOnHierarchy(name) {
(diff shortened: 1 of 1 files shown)


src/scene/graph-node.js
```diff
@@ -1124,8 +1135,8 @@ class GraphNode extends EventHandler {
         if (this._frozen)
             node._unfreezeParentToRoot();
 
-        // alert an entity that it has been inserted
-        if (node.fire) node.fire('insert', this);
+        // alert an entity hierarchy that it has been inserted
+        node._fireOnHierarchy('insert');
```
slimbuck, line 1139: the insert event will now have a different this argument. is it intentional?

src/scene/graph-node.js
```diff
@@ -1099,6 +1099,17 @@ class GraphNode extends EventHandler {
     }
     // #endif
 
+    // fires an event on all children of the node
+    _fireOnHierarchy(name) {
+        if (this.fire) {
+            this.fire(name, this);
```
slimbuck, line 1105: the this argument is the wrong this. should be passed in as param.

mvaligursky: it's not being used on the receiving side, so it's not even clear what it should be - could pass nothing at all.

slimbuck: are the insert and remove events documented or internal only? if they're internal then better to remove. if they're documented and public then you'll want to send the parent element as argument.

mvaligursky: removed as it's private and unused.

src/scene/graph-node.js
```diff
@@ -1099,6 +1099,17 @@ class GraphNode extends EventHandler {
     }
     // #endif
 
+    // fires an event on all children of the node
+    _fireOnHierarchy(name) {
+        if (this.fire) {
```
slimbuck, line 1104: Any idea why there is a check for 'fire' here? GraphNode inherits from EventHandler which defines fire function and can't really see when that would be missing.

mvaligursky: good point and no idea. I removed it and all seems to still work :)

slimbuck (approved):

willeastcott:
Do we want this to go out with 1.46.2 or 1.47.0?

mvaligursky:
1.46.2

willeastcott:
Yeah, but it also touches some pretty fundamental code. I'm a bit concerned it could have either performance impact or introduce some kind of behavioral change.

slimbuck:
I agree with @willeastcott actually. This is a pretty serious change and I'd also vote for merge with next major engine release.

mvaligursky:
Regarding the performance impact, there should not be large. We build hierarchies from top to bottom .. so when an entity is added to parent, it does not have further children. So this code would mostly trigger only when a hierarchy (a character) is added to the scene .. in which case this event would trigger one time per entity and no more. And it's only handled by model and render component. I'll do some perf testing tomorrow, but I expect this should be the case.

mvaligursky:
I added timing to engine example hierarchy which builds hierarchy with 5460 entities + materials, and I cannot measure any performance difference at all - in both cases the times are in the range of 700-750ms.

willeastcott:
That's good - my worry levels have subsided substantially.

willeastcott:
Wouldn't we want to distinguish the root node/entity being reparented vs a descendent somehow? I can imagine an application wanting to be able to discern the two.

PR #3437, opened 2021-08-31
Converted few classes in scene folder to use let/const

mvaligursky:


src/scene/morph-instance.js
```diff
@@ -238,18 +239,18 @@ class MorphInstance {
...
-        var blending = false;
-        var count = this._activeTargets.length;
-        for (var i = 0; i < count; i++) {
-            var activeTarget = this._activeTargets[i];
-            var tex = activeTarget.target[srcTextureName];
+        let usedCount = 0;
+        let blending = false;
+        const count = this._activeTargets.length;
+        for (let i = 0; i < count; i++) {
+            const activeTarget = this._activeTargets[i];
```
willeastcott, line 252: Could this be:

```
for (const activeTarget of this._activeTargets) {
```

willeastcott (approved):

src/scene/morph-instance.js
mvaligursky: Could be. I still use `count` later, so that is needed. but in general old school for is faster, so I tend to keep it.

willeastcott: OK, fair enough. I tend to use `for ... in` for init/occasional stuff....and vanilla `for` for per-frame stuff.

PR #3441, opened 2021-09-01
Revert "Prevent shapes creation on disabled entities (#3432)"

slimbuck:
This reverts commit 4b0760b3db5af9f635d28e4d8c8781fefd896b62.

Sorry @LeXXik this change breaks the Seemore demo https://playcanvas.com/project/612100/overview/seemore-2019 so reverting for this engine release.

Will be good to try get this optimisation in without altering behavior.

### DIFF
--- a/src/framework/components/collision/system.js
+++ b/src/framework/components/collision/system.js
@@ -64,7 +64,7 @@ class CollisionSystemImpl {
         var entity = component.entity;
         var data = component.data;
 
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

LeXXik:
I see now. At the moment the collision component assumes the collision data was processed when the entity was parsed, so it doesn't call recreate physical shapes, unless it was an asset. I have a custom collision component, which is different from original, so I missed it. I did a new patch, which also works with Seemore. I will do some more tests before pushing.

PR #2838, opened 2021-02-05
Lightmapper improvements - soft shadow & Ambient Occlusion support

mvaligursky:
This PR implements improvements to existing Lighmapping functionality:
- soft directional light baking - specified by angle and number of samples
- ambient occlusion baking, for both single color as well as cubemap based, sphere and hemisphere based. Controlled by number of samples. Separate contrast and brightness control over the AO.
- bilateral filtering for lightmaps - allows smooth AO and soft shadows with smaller number of samples.

### NEW API
**Light Component:**
- _bakeNumSamples_ - umber of samples used to bake the light
- _bakeArea_ - the penumbra angle in degrees

**Scene:**
- _lightmapFilterEnabled_ - Enables bilateral filter on runtime baked color lightmaps
- _lightmapFilterRange_ and _lightmapFilterSmoothness_ - range and spatial parameters of the bilateral filter 
- _ambientBake_ - If enabled the ambient lighting will be baked into lightmaps, including Ambient Occlusion
- _ambientBakeNumSamples_ - number of samples used to bake the ambient light
- _ambientBakeSpherePart_ - part of the sphere used to bake AO, from sphere cap to hemisphere to full sphere
- _ambientBakeOcclusionContrast_ and _ambientBakeOcclusionBrightness_ - contrast and brightness of AO

Other changes:
- split Lightmapper into multiple files 
- extracted lightmap cache from MeshInstance to standalone class
- orbit camera script updated to handle render component as a focus point

engine example (when engine is released)
http://playcanvas.github.io/#/graphics/lights-baked-ao

Possible further improvements: https://github.com/playcanvas/engine/issues/3430

engine example with all features enabled:
Screenshot 2021-09-01 at 14 23 51

new features disabled - no soft shadows, no AO
Screenshot 2021-09-01 at 14 24 18

constant ambient color (AO baking disabled)
Screenshot 2021-09-01 at 14 24 55

constant ambient color (AO baking enabled)
Screenshot 2021-09-01 at 14 25 06

soft shadows and ambient occlusion
Screenshot 2021-08-27 at 08 55 07

before-after screenshot:
Screenshot 2021-08-27 at 16 01 37

PR #3449, opened 2021-09-02
Added render component support to Orbit Camera

### LINKED ISSUE #3448: Orbit Camera target error
There an issue on the orbit camera script i just discovered while i was trying to fix a camera position error.

### Description

When you give to the script a target to look at, if the model you gave is a template with only renders, it will not work because in the orbit camera script, the target system require the mesh instances of the model to work. So, to fix my problem, i had to import a model with a mesh, make it transparent and use it as the target. Not so clean.

The following screens shows the configuration i needed to make it work.

image
image

yaustar:
Fixes #3448

Taken from the updated Orbit Camera project: https://developer.playcanvas.com/en/tutorials/orbit-camera/

### DIFF
--- a/scripts/camera/orbit-camera.js
+++ b/scripts/camera/orbit-camera.js
@@ -3,10 +3,10 @@
 ////////////////////////////////////////////////////////////////////////////////
 var OrbitCamera = pc.createScript('orbitCamera');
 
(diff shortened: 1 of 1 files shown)


mvaligursky:
I have render support included here
https://github.com/playcanvas/engine/pull/2838

yaustar:
Closed as already done in light mapper PR

PR #3454, opened 2021-09-03
[AnimComponent] Fix for transitions

ellthompson:
The anim controller now uses the currently active state as the new previous state in all state transitions, irregardless of whether the transition was from the active state, previous state or ANY state. In the instance where the controller is forcing a new state rather than transitioning between states, the previousState is cleared.

### DIFF
--- a/src/anim/controller/anim-controller.js
+++ b/src/anim/controller/anim-controller.js
@@ -320,7 +320,7 @@ class AnimController {
         let state;
         let animation;
         let clip;
-        // If transition.from is set, tranisition from the active state irregardless of the transition.from value (this could be the previous, active or ANY states).
+        // If transition.from is set, transition from the active state irregardless of the transition.from value (this could be the previous, active or ANY states).
         // Otherwise the previousState is cleared.
         this.previousState = transition.from ? this.activeStateName : null;
         this.activeState = transition.to;
--- a/src/anim/controller/anim-controller.js
+++ b/src/anim/controller/anim-controller.js
@@ -323,11 +323,6 @@ class AnimController {
         // If transition.from is set, tranisition from the active state irregardless of the transition.from value (this could be the previous, active or ANY states).
         // Otherwise the previousState is cleared.
         this.previousState = (transition.from ? this.activeStateName : null);
-        if (transition.from === ANIM_STATE_ANY) {
-            this.previousState = this.activeStateName;
-        } else {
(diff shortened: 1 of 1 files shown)


src/anim/controller/anim-controller.js
```diff
@@ -320,7 +320,13 @@ class AnimController {
         let state;
         let animation;
         let clip;
-        this.previousState = transition.from;
+        // if the transition is coming from the ANY state, use the active state as the previous state to maintain the record of currently plaing animations.
```
yaustar, line 323:
```suggestion
        // If the transition is coming from the ANY state, use the active state as the previous state to maintain the record of currently playing animations.
```

yaustar (approved):

src/anim/controller/anim-controller.js
```diff
@@ -320,7 +320,13 @@ class AnimController {
...
         let animation;
         let clip;
-        this.previousState = transition.from;
+        // If the transition is coming from the ANY state, use the active state as the previous state to maintain the record of currently playing animations.
+        // Otherwise use the transitions 'from' state
+        if (transition.from === ANIM_STATE_ANY) {
+            this.previousState = this.activeStateName;
+        } else {
+            this.previousState = transition.from;
+        }
```
willeastcott, line 329:
```suggestion
        this.previousState = (transition.from === ANIM_STATE_ANY) ? this.activeStateName : transition.from;
```

willeastcott: Feel free to turn this suggestion down. 😄

willeastcott (approved):

ellthompson:
Updated this fix as I noticed another related bug. This issue was arising both when transitioning from the ANY state and when an already active transition was interrupted by a transition associated with the previous state.

src/anim/controller/anim-controller.js
```diff
@@ -320,7 +320,9 @@ class AnimController {
         let state;
         let animation;
         let clip;
-        this.previousState = transition.from;
+        // If transition.from is set, tranisition from the active state irregardless of the transition.from value (this could be the previous, active or ANY states).
```
willeastcott, line 323:
```suggestion
        // If transition.from is set, transition from the active state irregardless of the transition.from value (this could be the previous, active or ANY states).
```

willeastcott (approved):

PR #3455, opened 2021-09-05
Track active StandardMaterial parameters correctly

### LINKED ISSUE #3452: the cubemap parameters of material are missing
the cubemap parameters of material (texture_prefilteredCubeMap128, texture_prefilteredCubeMap64, ... texture_prefilteredCubeMap4) are missing

1. Because I call the material.update function in the script update

2. pc.Material.prototype.update will set dirty flag of material to true.

3. if material.dirty is true, then pc.StandardMaterial.prototype.updateUniforms will be called and clears material.parameters first.

4. Then calling pc.StandardMaterial.prototype.updateLightingUniforms is required to reset the parameters of the cube map.

5. However, updateLightingUniforms is only called by pc.StandardMaterial.prototype.updateshader, 
   but the shader is not updated because it is determined that the shader of the material has not been changed. 
   pc.StandardMaterial.prototype.updateShader is not called, and the parameters of the cubemap are not set, they are all undefined.

slimbuck:
Fixes #3452

This PR updates the method of tracking active parameters in `StandardMaterial` as the previous method was just plain wrong.

Lighting-specific parameters are now tracked and cleared separately from the rest (since they get updated at different times).

### DIFF
--- a/src/scene/materials/standard-material.js
+++ b/src/scene/materials/standard-material.js
@@ -27,7 +27,7 @@ const _props = {};
 const _uniforms = {};
 
 // temporary set of params
-let _params = { };
+let _params = new Set();
 
 /**
  * @class
@@ -358,8 +358,8 @@ class StandardMaterial extends Material {
         this._assetReferences = {};
         this._validator = null;
 
-        this._activeParams = null;
-        this._activeLightingParams = null;
+        this._activeParams = new Set();
+        this._activeLightingParams = new Set();
 
         this.shaderOptBuilder = new StandardMaterialOptionsBuilder();
 
@@ -403,7 +403,7 @@ class StandardMaterial extends Material {
     }
 
     _setParameter(name, value) {
-        _params[name] = true;
+        _params.add(name);
         this.setParameter(name, value);
     }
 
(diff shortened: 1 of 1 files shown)


src/scene/materials/standard-material.js
```diff
@@ -26,8 +26,8 @@ const _props = {};
 // special uniform functions on a standard material
 const _uniforms = {};
 
-// temporary helper array
-const _propsSet = [];
+// temporary set of params
+let _params = { };
```
mvaligursky, line 30: personally I'd use a `new Set()` as that would not require to store `true` as a value as currently nor use `hasOwnProperty`

mvaligursky (approved):

src/scene/materials/standard-material.js
willeastcott: Agreed. And you can use `Set#clear` to empty it instead of creating new objects.

willeastcott: @mvaligursky Since @slimbuck is away for a while, do you want to have a stab at making this change? Then we can merge...

slimbuck:
Thanks @willeastcott, @mvaligursky changes made: using sets and clear.

willeastcott (approved):

willeastcott:
Great stuff - thanks @slimbuck!! 😄

PR #3456, opened 2021-09-06
Make version logging at startup optional

fiso:
See title

Fixes #

Yes!

### DIFF
--- a/src/framework/application.js
+++ b/src/framework/application.js
@@ -175,6 +175,7 @@ var _deprecationWarning = false;
  * @param {string} [options.assetPrefix] - Prefix to apply to asset urls before loading.
  * @param {object} [options.graphicsDeviceOptions] - Options object that is passed into the {@link GraphicsDevice} constructor.
  * @param {string[]} [options.scriptsOrder] - Scripts in order of loading first.
+ * @param {boolean} [options.muteVersionLogging] - Disables version number logging at startup.
  * @example
  * // Engine-only example: create the application manually
  * var app = new pc.Application(canvas, options);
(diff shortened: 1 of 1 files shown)


willeastcott:
@yaustar No, I don't think it's needed due to the MIT license. But I personally do find it incredibly useful when checking if new WebGL apps I find are PlayCanvas. I'm kinda open to this....I'll give it some thought.

fiso:
> @yaustar No, I don't think it's needed due to the MIT license. But I personally do find it incredibly useful when checking if new WebGL apps I find are PlayCanvas. And to check what version engine they're on (since then I can see quickly if an issue is already fixed in a newer engine release). But I'm kinda open to this....I'll give it some thought.

If that's an important use-case, you could solve that by exposing some object on the window which can be checked by the curious. Forcing console logging in a production build is something I would say should be avoided.

For reference, here is the long-running discussion from when the author of Three.js after a lot of back and forth decided to remove the mandatory logging done by that library: https://github.com/mrdoob/three.js/pull/5835

LeXXik:
As a consumer, I do like that version output in the console. I would prefer it to stay there and perhaps add an explicit option to turn it off for those who needs it to be off for their reasons.

fiso:
> As a consumer, I do like that version output in the console. I would prefer it to stay there and perhaps add an explicit option to turn it off for those who needs it to be off for their reasons.

Yeah and this PR does exactly that. Out of the box nothing changes, but _if_ you have a preference for a clean console this gives you a nice way to disable the informational log message without resorting to money patching.

yaustar:
Closing this in favour of https://github.com/playcanvas/engine/pull/3508

Issue #3110, opened 2021-04-21
Changing tiling on material with clear coat will make the material white


yaustar:
See video
https://user-images.githubusercontent.com/16639049/115580136-ba6b1d00-a2be-11eb-86f4-ed38805a900e.mp4

mvaligursky:
this is the shader error we're getting, when tiling is enabled and clear coat normal map texture set up on a new material
[clear-coat-tiling-error.txt](https://github.com/playcanvas/engine/files/6607952/clear-coat-tiling-error.txt)

willeastcott:
So yeah, this line looks wrong, no?

```
21:	varying vec2 vUV0_1;
```

Presumably it should be:

```
21:	varying vec2 vUv0;
```

mvaligursky:
Perhaps, but this is generated in code, not a chunk. I suspect the story is more complicated though, as we can have some uv coordinates transformed individually or not at all per texture slot, so multiple texture coordinates need to be supported. Would need deeper investigation.

PR #3458, opened 2021-09-07
[Fix] Use correct UV transform for tiled normal map of clear coat

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/3110


slimbuck (approved):

src/graphics/program-lib/programs/standard.js
```diff
@@ -1132,7 +1132,8 @@ var standard = {
 
                 if (!options.hasTangents) {
                     // TODO: generalize to support each normalmap input (normalMap, normalDetailMap, clearCoatNormalMap) indenpendently
```
willeastcott, line 1134: Does this comment need updating now? It spells independently wrong too! 😆

PR #3463, opened 2021-09-08
Fix for cloned anim and render components

### LINKED ISSUE #3393: Bug cloning or setting materials for an entity that has skinning and was created with instantiateRenderEntity()
Hi, I'm loading a glb file at runtime, but when I clone my entity to have more than one copy the second will be completelly bugged. The same problem happens when I create a new material with my custom shader and set it to a meshInstance. Actually the bug is all about losing the skinning in both of these situations. Unfortunately I can't post a screenshot because I can't share the model.

To reproduce:
1 - load a glb file 
2 - clone it and add to the scene
3 - the meshes won't be affected by the bones anymore

With materials:
1 - load the glb file
2 - iterate it's mesh instances setting a newly created material with a custom shader
3 - the meshes won't be affected by the bones anymore

### LINKED ISSUE #3426: Instantiating a template with an anim component that has Root Bone set doesn't play animations
Project: https://playcanvas.com/editor/scene/1218229
Forum thread: https://forum.playcanvas.com/t/solved-animcomponent-not-work-with-templates/21578

Press space to instantiate a new instance of a template.

It looks like the root bone reference in the new instance is referencing a non existing entity (getByGuid with the referenced GUID returns null).

I'm not sure if this is an Anim Component issue or a Template one.

ellthompson:
Both the render and anim component contain a rootBone property. When either of these components were cloned, the rootBone entity wasn't being updated with the cloned rootBone entity in the new cloned hierarchy. This fixes both issues by implementing the `resolveDuplicatedEntityReferenceProperties` method in both the anim and render components, much like the current script component method.

Fixes #3426 
Fixes #3393

Test project for the fix: https://playcanvas.com/project/808975/

### DIFF
--- a/src/framework/components/render/component.js
+++ b/src/framework/components/render/component.js
@@ -112,7 +112,9 @@ class RenderComponent extends Component {
     _onRootBoneChanged() {
         // remove existing skin instances and create new ones, connected to new root bone
         this._clearSkinInstances();
-        this._cloneSkinInstances();
+        if (this.enabled && this.entity.enabled) {
+            this._cloneSkinInstances();
+        }
     }
 
     destroyMeshInstances() {
@@ -199,12 +201,9 @@ class RenderComponent extends Component {
         var app = this.system.app;
         var scene = app.scene;
 
-        if (this._initialRootBone) {
-            this.rootBone = this._initialRootBone;
-            this._initialRootBone = null;
-        }
-
         this._rootBone.onParentComponentEnable();
+
+        this._cloneSkinInstances();
 
         scene.on("set:layers", this.onLayersChanged, this);
         if (scene.layers) {
@@ -757,9 +756,8 @@ class RenderComponent extends Component {
     }
 
     resolveDuplicatedEntityReferenceProperties(oldRender, duplicatedIdsMap) {
-        const oldRootBoneId = oldRender.rootBone || (oldRender.initialRootBone && oldRender.initialRootBone.getGuid());
-        if (oldRootBoneId && duplicatedIdsMap[oldRootBoneId]) {
-            this.rootBone = duplicatedIdsMap[oldRootBoneId];
+        if (oldRender.rootBone && duplicatedIdsMap[oldRender.rootBone]) {
+            this.rootBone = duplicatedIdsMap[oldRender.rootBone];
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/framework/components/render/component.js
```diff
@@ -198,6 +199,11 @@ class RenderComponent extends Component {
         var app = this.system.app;
         var scene = app.scene;
 
+        if (this._initialRootBone) {
+            this.rootBone = this._initialRootBone;
+            this._initialRootBone = null;
+        }
```
mvaligursky, line 205: This feels odd. Why the need for initial root bone? Anim component does not have it. It does not feel like a proper fix to me.

src/resources/container.js
```diff
@@ -175,7 +175,8 @@ class ContainerResource {
             if (attachedMi) {
                 entity.addComponent("render", Object.assign({
                     type: "asset",
-                    meshInstances: attachedMi
+                    meshInstances: attachedMi,
+                    rootBone: root
```
mvaligursky, line 179: and this line can be perhaps removed now as well?

src/framework/components/render/component.js
ellthompson: Updated based on our discussion. The rootBone property is now set along with other initial component properties. However the cloning of skin instances which used to occur on the setting of the root bone has now been deferred to when the component is enabled.

src/framework/components/render/component.js
```diff
@@ -75,6 +75,7 @@ class RenderComponent extends Component {
         this._area = null;
 
         // the entity that represents the root bone if this render component has skinned meshes
+        this._initialRootBone = null;
```
mvaligursky, line 78: and this should be gone too I think.

src/resources/container.js
ellthompson: From my testing it looks like the rootBone has to be set on the entity you're cloning, otherwise the skin of the clone breaks.

mvaligursky: ok that seems ok .. look at SkinInstanceCache.createCachedSkinedInstance which is called 10 line bellow, we effectively use root as a rootBone as well.

mvaligursky (approved):
This all looks great to me, fantastic work, thanks for doing it!

PR #3472, opened 2021-09-10
Forward renderer - cleanup, separated static-meshes functionality

mvaligursky:
- separated functionality dealing with static meshes getting split up by lights into separate file (static-meshes)
- removed invalid/obsolete call to prepareStaticMeshes by the Application (completely incorrect parameters resulting to no-op)
- hiding ForwardRenderer's constructor from jsdocs
- initialized few missing members in ForwardRenderer to avoid Visual Studio giving us errors
- converted var to let/const (apart from static-meshes.js)

### DIFF
--- a/src/scene/renderer/forward-renderer.js
+++ b/src/scene/renderer/forward-renderer.js
@@ -87,9 +87,8 @@ const _tempMaterialSet = new Set();
  * @param {Scene} A - scene for the rendering.
  */
 class ForwardRenderer {
-    constructor(graphicsDevice, scene) {
+    constructor(graphicsDevice) {
         this.device = graphicsDevice;
-        this.scene = scene;
 
         this._shadowDrawCalls = 0;
         this._forwardDrawCalls = 0;
--- a/src/scene/renderer/forward-renderer.js
+++ b/src/scene/renderer/forward-renderer.js
@@ -4,7 +4,6 @@ import { Mat3 } from '../../math/mat3.js';
 import { Mat4 } from '../../math/mat4.js';
 import { Vec3 } from '../../math/vec3.js';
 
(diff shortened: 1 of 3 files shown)


src/scene/renderer/forward-renderer.js
```diff
@@ -32,69 +29,73 @@ import {
...
  * @class
  * @name ForwardRenderer
  * @classdesc The forward renderer render scene objects.
  * @description Creates a new forward renderer object.
+ * @hideconstructor
  * @param {GraphicsDevice} graphicsDevice - The graphics device used by the renderer.
+ * @param {Scene} A - scene for the rendering.
  */
 class ForwardRenderer {
-    constructor(graphicsDevice) {
```
willeastcott, line 91: So this is interesting to me. Before, I guess the API suggested you could do:

```
const sceneA = new pc.Scene();
const sceneB = new pc.Scene();

renderer.render(sceneA);
renderer.render(sceneB);
```

But this explicitly creates a 1-2-1 relationship between the scene and the renderer. Would be great to get your thoughts on this.

mvaligursky: It's a valid point @willeastcott , and so I've reverted this one change. At some point in the future it would certainly be nice to support multiple scenes this way. At the moment this would not work, as the scene is referenced in few other places  like batch manager, lightmapper - these would be easy to fix, but also by Materials and Lights, which mark the layer composition dirty when they change - this could be harder to fix.

willeastcott: Yep, I did suspect that the rendering of multiple scenes had broken over time. But yeah, I think it's a good call to roll this change back. Nice.

willeastcott (approved):

PR #3473, opened 2021-09-10
Converted more classes in scene folder to use let and const

mvaligursky:
- conversions of var -> let / const
- split one large function in batch-manager to two

### DIFF
--- a/src/scene/graph-node.js
+++ b/src/scene/graph-node.js
@@ -6,21 +6,21 @@ import { Mat4 } from '../math/mat4.js';
 import { Quat } from '../math/quat.js';
 import { Vec3 } from '../math/vec3.js';
 
-var scaleCompensatePosTransform = new Mat4();
-var scaleCompensatePos = new Vec3();
-var scaleCompensateRot = new Quat();
-var scaleCompensateRot2 = new Quat();
-var scaleCompensateScale = new Vec3();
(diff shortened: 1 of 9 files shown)


src/scene/graph-node.js
```diff
@@ -6,21 +6,21 @@ import { Mat4 } from '../math/mat4.js';
...
+const scaleCompensateScaleForParent = new Vec3();
+const tmpMat4 = new Mat4();
+const tmpQuat = new Quat();
+const position = new Vec3();
+const invParentWtm = new Mat4();
+const rotation = new Quat();
+const invParentRot = new Quat();
+const matrix = new Mat4();
+const target = new Vec3();
+const up = new Vec3();
```
yaustar, line 23: Shouldn't these technically be `let` as they will change value?

yaustar (approved):
LGTM with comment

src/scene/graph-node.js
mvaligursky: it's not c++ or c# type of const.
In JS const is if you don't assign anything else to it. So you can even have `const array = [];` and add and remove elements from it. As long as you don't do `array = null;` or similar

yaustar: Oh, I see. JS is weird 😅

PR #3474, opened 2021-09-10
Optimize collision component load time

LeXXik:
Optimizes application load time, when there are many disabled entities in a scene with a collision component.

Before (200 disabled entities):

New Project (14)

After:

New Project (17)

Tested with See More and a local physics test bed.

### DIFF
--- a/src/framework/components/collision/component.js
+++ b/src/framework/components/collision/component.js
@@ -315,6 +315,9 @@ class CollisionComponent extends Component {
             if (asset && (!asset.resource || !this.data.shape)) {
                 this.system.recreatePhysicalShapes(this);
                 return;
+            } else if (! this.data.initialized && ! this.data.shape) {
+                this.system.recreatePhysicalShapes(this);
+                return;
             }
         }
 
--- a/src/framework/components/collision/component.js
+++ b/src/framework/components/collision/component.js
@@ -322,7 +322,7 @@ class CollisionComponent extends Component {
 
         if (_recreatePhysicalShapes) {
             this.system.recreatePhysicalShapes(this);
-            return;            
+            return;
         }
 
         if (this.entity.rigidbody) {
--- a/src/framework/components/collision/component.js
+++ b/src/framework/components/collision/component.js
@@ -308,17 +308,21 @@ class CollisionComponent extends Component {
     }
 
     onEnable() {
+        let _recreatePhysicalShapes = false;
         if (this.data.type === 'mesh' && (this.data.asset || this.data.renderAsset) && this.data.initialized) {
             var asset = this.system.app.assets.get(this.data.asset || this.data.renderAsset);
             // recreate the collision shape if the model asset is not loaded
             // or the shape does not exist
             if (asset && (!asset.resource || !this.data.shape)) {
-                this.system.recreatePhysicalShapes(this);
-                return;
-            } else if (! this.data.initialized && ! this.data.shape) {
-                this.system.recreatePhysicalShapes(this);
-                return;
+                _recreatePhysicalShapes = true;
             }
+        } else if (! this.data.initialized && ! this.data.shape) {
+            _recreatePhysicalShapes = true;
+        }
+
+        if (_recreatePhysicalShapes) {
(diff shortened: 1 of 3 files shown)


src/framework/components/collision/component.js
```diff
@@ -315,6 +315,9 @@ class CollisionComponent extends Component {
             if (asset && (!asset.resource || !this.data.shape)) {
                 this.system.recreatePhysicalShapes(this);
                 return;
+            } else if (! this.data.initialized && ! this.data.shape) {
```
mvaligursky, line 318: the code both ifs execute is the same. ideally we refactor the condition and have the code only one time.

src/framework/components/collision/system.js
```diff
@@ -49,8 +49,12 @@ class CollisionSystemImpl {
 
     // Called after the call to system.super.initializeComponentData is made
     afterInitialize(component, data) {
-        this.recreatePhysicalShapes(component);
-        component.data.initialized = true;
+        const entity = component.entity;
```
mvaligursky, line 52: I'd just inline this in the line bellow.

src/framework/components/collision/component.js
LeXXik: Right and it is in the wrong place. Let me refactor.

Maksims:
Will this lead to postponed CPU-intensive operations on enabling?
Basically, does this eliminates unnecessary work in general, or it shifts it from before enabled to after enabled?

LeXXik:
Both. If the entity is never enabled, then it eliminates the unnecessary work. If it does get enabled at some point, then it processes when that happens. If it is enabled from the beginning, then no changes.

Maksims:
> Both. If the entity is never enabled, then it eliminates the unnecessary work. If it does get enabled at some point, then it processes when that happens. If it is enabled from the beginning, then no changes.

This leads to a problem: CPU stalls when entities are enabled, similar to particles case, which enforces developers to enable and disable entities on load, to ensure they are fully initialized without further CPU stalls.

LeXXik:
Yes, you care correct. The stall does take place when the entity gets processed. We can't really avoid it, as CPU has to generate the collision shape, unless we move it to worker, but the main thread would still be affected. 

However, think of a scene with many deep branches, where only one branch is active at a time. At the moment, the stall will be quite significant on game start, specially on mobile devices. Currently a developer, have no way to mitigate it, except accepting it, because the engine desided to process the whole scene in one go. This PR at least gives us a control on when that stall should take place and how large it should be.

I do take enabling/disabling the particles system as a good practice, rather than an enforced hurdle we have to accept. Speaking of this made me even think if we should have a "preload" property for entity, which marks it to be fully loaded and initialized on app start, similar to assets. It would remove the need in enabling/disabling mechanic.

Maksims:
Using a template for chunks that not necessary are required, or using multiple scenes - is a good practice to reduce the complexity of hierarchy based on application logic.

Please, provide a way for developers to enforce collision initialization beforehand (on loading), or by a function call. To allow developers to create a smooth experience where they can afford loading time stall.
Also if any assets are loading, while JS is blocked - this loading continues, so it is efficient to stall the JS thread during assets loading.

Freezes during running application - is really bad for user experience, and must be avoided as much as possible.
Enabling/disabling things - feels a workaround, not a nice solution, a better solution is required, also for particles as it has been an issue in many projects.

LeXXik:
The issue here is that a default Ammo build has about 64Mb of memory for itself. When the engine processes collision components of a complete scene, it can potentially crash, because it would run out of memory before the next step comes where it removes the disabled objects. I do agree though that a better solution is needed, instead of enabling/disabling mechanic. What are the options here? We can also mark this PR as draft, until a better alternative is found. I can simply keep a monkey patch in my project until then.

LeXXik:
Closing this one. Whether the collision components are parsed at start (used or not) or at use time seems to be an opinionated preferring of a developer. Forcing one way or the other is equally bad. A better system would allow a user to switch it via some sort of a setting, but that would also need to consider other component types.

Issue #2581, opened 2020-11-30
Scroll view component not scrollable with mousewheel.


CMRDX:
https://github.com/playcanvas/engine/blob/0038e42bc41a06e70a9b404cc07a6c72df8c90dc/src/framework/components/scroll-view/component.js#L80

Judging by this comment this was an intended feature at some point, that got maybe forgotten or overlooked. It would be great to have it added to the engine. In the meantime I will have to come up with a different solution. Any tips are greatly appreciated :)

CMRDX:
Found a simple workaround for now.
I register to the mouseWheel event and set a new scroll value like that:
`this.scrollViewComponentEntity.scrollview._onSetScroll(0, newScrollValue);`

It still requires some tweaking for it to look smooth, but it does the job.

Issue #1332, opened 2018-08-31
Remove dataStore and ComponentData


daredevildave:
Older components use the dataStore which requires functions to access data and adds overhead when modifying fields.

We should remove all ComponentData classes

Pheater:
What would be the best way to add a readonly variable to an existing component if not in dataStore?

willeastcott:
How about:
```
    Object.defineProperty(SomeComponent.prototype, "someProperty", {
        get: function () {
            return this._someProperty;
        }
    });
```

willeastcott:
Just for reference, this is the current state of old format (uses component data) vs new format component (stores data in component):

Old
* Anim
* Animation
* AudioSource
* Button
* Collision
* Light
* ParticleSystem
* Render
* ScriptLegacy

New
* AudioListener
* Camera
* Element
* Joint
* LayoutChild
* LayoutGroup
* Model
* RigidBody
* Screen
* Script
* ScrollView
* Scrollbar
* Sound
* Sprite
* Zone

mvaligursky:
how come model is new and render is called? I thought they were about the same. What did I miss.

willeastcott:
Somebody added `rootBone` to render.

https://github.com/playcanvas/engine/blob/master/src/framework/components/render/data.js#L4

willeastcott:
This work is now complete. Closing.

PR #3478, opened 2021-09-13
Use mouse scroll wheel events to navigate content in scroll-view UI component.

jpauloruschel:
Fixes https://github.com/playcanvas/engine/issues/2581.

This PR adds a listener to the scroll view UI component for mouse wheel events (both X and Y directions), and updates the scroll value accordingly. Also adds additional properties to control whether this feature is enabled at all, and the sensitivity.

### DIFF
--- a/examples/src/examples/user-interface/scroll-view.tsx
+++ b/examples/src/examples/user-interface/scroll-view.tsx
@@ -192,6 +192,8 @@ class ScrollViewExample extends Example {
             bounceAmount: 0.1,
             contentEntity: content,
             friction: 0.05,
+            useMouseWheel: true,
+            mouseWheelSensitivity: new pc.Vec2(1, 1),
             horizontal: true,
             horizontalScrollbarEntity: horizontalScrollbar,
             horizontalScrollbarVisibility: pc.SCROLLBAR_VISIBILITY_SHOW_WHEN_REQUIRED,
--- a/examples/src/examples/user-interface/scroll-view.tsx
+++ b/examples/src/examples/user-interface/scroll-view.tsx
@@ -192,8 +192,6 @@ class ScrollViewExample extends Example {
             bounceAmount: 0.1,
             contentEntity: content,
             friction: 0.05,
-            useMouseWheel: true,
-            mouseWheelSensitivity: new pc.Vec2(1, 1),
             horizontal: true,
             horizontalScrollbarEntity: horizontalScrollbar,
             horizontalScrollbarVisibility: pc.SCROLLBAR_VISIBILITY_SHOW_WHEN_REQUIRED,
--- a/examples/src/examples/user-interface/scroll-view.tsx
+++ b/examples/src/examples/user-interface/scroll-view.tsx
@@ -130,8 +130,9 @@ class ScrollViewExample extends Example {
             lineHeight: 36,
             pivot: new pc.Vec2(0, 1),
             text: "This is a scroll view control. You can scroll the content by dragging the vertical " +
-                    "or horizontal scroll bars or by dragging the content itself. Notice the elastic " +
-                    "bounce if you drag the content beyond the limits of the scroll view.",
+                    "or horizontal scroll bars, by dragging the content itself, by using the mouse wheel, or " +
+                    "by using a trackpad. Notice the elastic bounce if you drag the content beyond the " +
+                    "limits of the scroll view.",
             type: pc.ELEMENTTYPE_TEXT,
             width: 400,
             wrapLines: true
@@ -191,6 +192,8 @@ class ScrollViewExample extends Example {
             bounceAmount: 0.1,
             contentEntity: content,
             friction: 0.05,
+            useMouseWheel: true,
+            mouseWheelSensitivity: new pc.Vec2(1, 1),
             horizontal: true,
             horizontalScrollbarEntity: horizontalScrollbar,
             horizontalScrollbarVisibility: pc.SCROLLBAR_VISIBILITY_SHOW_WHEN_REQUIRED,
--- a/examples/src/examples/user-interface/scroll-view.tsx
+++ b/examples/src/examples/user-interface/scroll-view.tsx
@@ -193,7 +193,7 @@ class ScrollViewExample extends Example {
             contentEntity: content,
             friction: 0.05,
             useMouseWheel: true,
-            mouseWheelSensitivity: new pc.Vec2(1, 1),
+            mouseWheelSensitivity: pc.Vec2.ONE,
             horizontal: true,
             horizontalScrollbarEntity: horizontalScrollbar,
             horizontalScrollbarVisibility: pc.SCROLLBAR_VISIBILITY_SHOW_WHEN_REQUIRED,
--- a/src/framework/components/scroll-view/component.js
+++ b/src/framework/components/scroll-view/component.js
@@ -10,6 +10,7 @@ import { ElementDragHelper } from '../element/element-drag-helper.js';
 
 import { SCROLL_MODE_BOUNCE, SCROLL_MODE_CLAMP, SCROLL_MODE_INFINITE, SCROLLBAR_VISIBILITY_SHOW_ALWAYS, SCROLLBAR_VISIBILITY_SHOW_WHEN_REQUIRED } from './constants.js';
 import { Component } from '../component.js';
+import { EVENT_MOUSEWHEEL } from '../../../input/constants.js';
 
 var _tempScrollValue = new Vec2();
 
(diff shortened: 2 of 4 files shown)


src/framework/components/scroll-view/component.js
```diff
@@ -32,6 +33,8 @@ var _tempScrollValue = new Vec2();
  *
  * @property {number} bounceAmount Controls how far the content should move before bouncing back.
  * @property {number} friction Controls how freely the content should move if thrown, i.e. By flicking on a phone or by flinging the scroll wheel on a mouse. A value of 1 means that content will stop immediately; 0 means that content will continue moving forever (or until the bounds of the content are reached, depending on the scrollMode).
+ * @property {boolean} useMouseWheel Whether to use mouse wheel for scrolling (horizontally and vertically) when mouse is within bounds.
+ * @property {vec2} mouseWheelSensitivity Mouse wheel horizontal and vertical sensitivity. Only used if useMouseWheel is set.
```
mvaligursky, line 37: is this commonly exposed as two numbers, or just one that applies to both?

src/framework/components/scroll-view/component.js
```diff
@@ -482,6 +484,9 @@ class ScrollViewComponent extends Component {
 
                 this._setScrollFromContentPosition(position);
             }
+
+            this._velocity.x *= (1 - this.friction);
+            this._velocity.y *= (1 - this.friction);
```
jpauloruschel, line 489: Note: moving the velocity update to *after* it's used this frame otherwise SCROLL_MODE_BOUNCE doesn't work properly when `friction = 1`.

src/framework/components/scroll-view/data.js
```diff
@@ -1,6 +1,10 @@
+import { Vec2 } from '../../../math/vec2.js';
+
 class ScrollViewComponentData {
     constructor() {
         this.enabled = true;
+        this.useMouseWheel = true;
```
jpauloruschel, line 6: Open for discussion if this should be default `true` or not.

src/framework/components/scroll-view/component.js
jpauloruschel: I think support for 'X' direction was not widely used so most systems only use the y direction. I see no reason why to not add support for X-direction, and it would make sense to have separate controls. In case a developer wants the Y control, but not X, they can set the X sensitivity to 0, for instance. And if they wish to add sensitivity controls in their game's Settings, I believe it is best to have control over both if needed.

src/framework/components/scroll-view/data.js
```diff
@@ -1,6 +1,10 @@
...
+        this.mouseWheelSensitivity = new Vec2(1, 1);
```
willeastcott, line 7: Check out this issue:

https://github.com/playcanvas/engine/issues/1332

In other words, new component properties should live on the component, not the soon-to-be removed `ComponentData` classes.

src/framework/components/scroll-view/component.js
```diff
@@ -32,6 +33,8 @@ var _tempScrollValue = new Vec2();
  *
  * @property {number} bounceAmount Controls how far the content should move before bouncing back.
  * @property {number} friction Controls how freely the content should move if thrown, i.e. By flicking on a phone or by flinging the scroll wheel on a mouse. A value of 1 means that content will stop immediately; 0 means that content will continue moving forever (or until the bounds of the content are reached, depending on the scrollMode).
+ * @property {boolean} useMouseWheel Whether to use mouse wheel for scrolling (horizontally and vertically).
+ * @property {Vec2} mouseWheelSensitivity Mouse wheel horizontal and vertical sensitivity. Only used if useMouseWheel is set.
```
willeastcott, line 37: What is the range here?

examples/src/examples/user-interface/scroll-view.tsx
```diff
@@ -191,6 +192,8 @@ class ScrollViewExample extends Example {
             bounceAmount: 0.1,
             contentEntity: content,
             friction: 0.05,
+            useMouseWheel: true,
+            mouseWheelSensitivity: new pc.Vec2(1, 1),
```
willeastcott, line 196:
```suggestion
            mouseWheelSensitivity: pc.Vec2.ONE,
```

src/framework/components/scroll-view/data.js
jpauloruschel: Not sure I understand this. I've added the new properties to `const _schema_`, which, following the code in `initializeComponentData()`, is the unified location of the component's properties. It looked to me, then, that `ScrollViewComponentData` is for storing the 'default' values for those properties.

Do you then mean that I don't need to put anything under `ScrollViewComponentData`, and instead set the default values directly in the component's constructor?

src/framework/components/scroll-view/component.js
jpauloruschel: No particular range here. Developers can set to 0 if they want to disable one of the directions, [0..1] if they want slower than default, or [1..Inf] if they want faster. For UI-heavy games (strategy, 4X, etc), this would most likely be an in-game setting.

src/framework/components/scroll-view/data.js
willeastcott: Yeah, pretty much. Look at how I ported the RigidBody component from the old format to the new one:

https://github.com/playcanvas/engine/pull/3164

src/framework/components/scroll-view/component.js
willeastcott: Might be worth adding some of this info. e.g. "Setting a direction to 0 will disable mouse wheel scrolling in that direction. 1 is a default sensitivity that is considered to feel good. The values can be set higher or lower than 1 to tune the sensitivity. Set to [1, 1] by default.".

jpauloruschel: Nice, I like it!!

src/framework/components/scroll-view/system.js
```diff
@@ -50,6 +55,12 @@ class ScrollViewComponentSystem extends ComponentSystem {
         if (data.dragThreshold === undefined) {
             data.dragThreshold = DEFAULT_DRAG_THRESHOLD;
         }
+        if (data.useMouseWheel === undefined) {
+            data.useMouseWheel = true;
+        }
+        if (data.mouseWheelSensitivity === undefined) {
+            data.mouseWheelSensitivity = SCROLLBAR_MOUSE_WHEEL_SENSITIVITY;
```
willeastcott, line 62: Got a slight concern here. So if `data.mouseWheelSensitivity` is set to the _reference_ to `Vec2.ONE`, if `data.mouseWheelSensitivity` with any other value, the actual value of the `Vec2.ONE` constant will change. Might be safer to do the following further up:

```
const SCROLLBAR_MOUSE_WHEEL_SENSITIVITY = new pc.Vec2(1, 1);
```

willeastcott (approved):
Approving with one final comment.... 😄

src/framework/components/scroll-view/system.js
yaustar: Or make mouseWheelSensitivity a new pc.Vec2 and copy the constant rather than referencing it?

jpauloruschel: Ahh good callout! Will send another commit for this.

src/framework/components/scroll-view/data.js
willeastcott: I'd say that's 100% reasonable for it to default to true. 😄

willeastcott:
Looking good now! 😄 Let's get @mvaligursky 's review too and merge on Monday! 🎉

src/framework/components/scroll-view/component.js
```diff
@@ -580,6 +582,22 @@ class ScrollViewComponent extends Component {
...
+    _onMouseWheel(event) {
+        if (this.useMouseWheel) {
+            const wheelEvent = event.event;
+
+            // wheelEvent's delta variables are screen space, so they need to be normalized first
+            const normalizedDeltaX = (wheelEvent.deltaX / this._contentReference.entity.element.calculatedWidth) * this.mouseWheelSensitivity.x;
+            const normalizedDeltaY = (wheelEvent.deltaY / this._contentReference.entity.element.calculatedHeight) * this.mouseWheelSensitivity.y;
+
+            // update scroll positions, clamping to [0, maxScrollValue] to always prevent over-shooting
+            var scrollX = math.clamp(this._scroll.x + normalizedDeltaX, 0, this._getMaxScrollValue(ORIENTATION_HORIZONTAL));
```
mvaligursky, line 594: var -> const?

mvaligursky (approved):
looks good!

Issue #3427, opened 2021-08-25
UI elements events on the world layer are triggered in the 'wrong' order


yaustar:
When UI elements are placed in the world on the world layer, the hover/mouse enter event seems to trigger in a hierarchy order rather than the render order.

Here, the buttons go transparent if they are hovered and it can be seen that the mouse is on the yellow element but the green one behind is going transparent.

https://user-images.githubusercontent.com/16639049/130779431-6aef269b-b926-4c53-980a-c16b4dd74a70.mp4

Scene: https://playcanvas.com/editor/scene/1218512
Forum thread: https://forum.playcanvas.com/t/image-element-input-z-order/21586/9

PR #3482, opened 2021-09-14
Trigger events on world UI elements in the correct order.

jpauloruschel:
Fixes https://github.com/playcanvas/engine/issues/3427

This PR extends the algorithm `intersectLineQuad` as per the Real-Time Collision Detection book by Christer Ericson (`5.3.5 Intersecting Line Against Quadrilateral`, page 189) to instead of just returning if the line intersected with a quad, return the square distance to the intersection point or -1 if there's no intersection. The algorithm uses square distance (rather than distance) because the 'exact' distance doesn't really matter: the value is only used to 'sort' the elements. This saves computation by avoiding an extra `sqrt`.

Then, for 3D elements, instead of immediately returning the first element that intersected with the ray, return the one that was closest to the ray's origin. For 2D elements, the logic is unchanged: return the first one (after layer and hierarchy sorting) that hit.

### DIFF
--- a/src/input/element-input.js
+++ b/src/input/element-input.js
@@ -1026,8 +1026,9 @@ class ElementInput {
     _checkElement(ray, element, screen) {
         // ensure click is contained by any mask first
         if (element.maskedBy) {
-            var result = this._checkElement(ray, element.maskedBy.element, screen);
-            if (result < 0) return -1;
+            if (this._checkElement(ray, element.maskedBy.element, screen) < 0) {
+                return -1;
+            }
         }
 
         var scale;
--- a/src/input/element-input.js
+++ b/src/input/element-input.js
@@ -29,6 +29,11 @@ var _pc = new Vec3();
 var _pd = new Vec3();
 var _m = new Vec3();
 var _sct = new Vec3();
+var _sdpr1 = new Vec3();
+var _sdpr2 = new Vec3();
+var _sdprQ = new Vec3();
+var _sdprD = new Vec3();
+var _sdprTemp = new Vec3();
 var _accumulatedScale = new Vec2();
 var _paddingTop = new Vec3();
 var _paddingBottom = new Vec3();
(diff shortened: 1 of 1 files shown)


src/input/element-input.js
```diff
@@ -993,6 +1034,28 @@ class ElementInput {
         return false;
     }
 
+    _checkElementSqrDistanceToBounds(ray, element, screen) {
+        // ensure click is contained by any mask first
+        if (element.maskedBy) {
+            return this._checkElementSqrDistanceToBounds(ray, element.maskedBy.element, screen);
+        }
+
+        var scale;
```
mvaligursky, line 1043: use 'let' and 'const' instead of 'var' please

src/input/element-input.js
```diff
@@ -46,6 +51,40 @@ function scalarTriple(p1, p2, p3) {
     return _sct.cross(p1, p2).dot(p3);
 }
 
+// calculates the distance from a point to a 3D rectangle defined by its corners
```
mvaligursky, line 54: would it be more precise to use the ray, intersect it with the rectangle, get exact intersection point and return the distance to that point, instead of using corners?
I'm not entirely sure what this function does though .. what is 'distance from a point to a rectangle defined by its corners'? Is it a perpendicular distance or something?

mvaligursky:
Also, please add details of the change as the PR description.

src/input/element-input.js
jpauloruschel: Oh! I believe I completely misunderstood the meaning of `element.worldCorners`. I thought it was the bounding box, so the code checks the closest point within the bounding box to another point. The algorithm was taken from the book Real-Time Collision Detection (same as other algorithms in this file), Chapter 5.

Thanks for the callout! I'll be able to simply this a lot and re-use the current implementation of intersectLineQuad.

src/input/element-input.js
```diff
@@ -811,6 +838,7 @@ class ElementInput {
 
     _getTargetElement(camera, x, y) {
         var result = null;
+        var closestDistance = Infinity;
```
mvaligursky, line 841: no 'var' please :)

src/input/element-input.js
```diff
@@ -844,9 +872,10 @@ class ElementInput {
                 ray = ray3d;
             }
 
-            if (ray && this._checkElement(ray, element, screen)) {
+            var d = this._checkElement(ray, element, screen);
+            if (ray && d > 0 && d < closestDistance) {
                 result = element;
-                break;
+                closestDistance = d;
```
mvaligursky, line 878: I have a feeling this code changes the 2d path and not only 3d path I thought should return the nearest one. The 2d path should return the first one that it hits?

jpauloruschel: Fixed in next commit.

src/input/element-input.js
```diff
@@ -842,11 +877,13 @@ class ElementInput {
                     }
                 }
                 ray = ray3d;
-            }
 
-            if (ray && this._checkElement(ray, element, screen)) {
-                result = element;
-                break;
+                // continue looking through all elements, only storing the closest one
+                const currentDistance = this._checkElement(ray, element, screen);
```
slimbuck, line 882: this line needs to be wrapped in an if (ray)

src/input/element-input.js
```diff
@@ -867,7 +904,7 @@ class ElementInput {
             var element = this._elements[i];
 
             if (! element.screen || ! element.screen.screen.screenSpace) {
-                if (this._checkElement(rayA, element, false)) {
+                if (this._checkElement(rayA, element, false) > 0) {
```
slimbuck, line 907: If failure is defined as `value < 0` then success should be `value >= 0`.

(Make sure it's consistent throughout).

src/input/element-input.js
```diff
@@ -842,11 +877,13 @@ class ElementInput {
...
                 }
                 ray = ray3d;
-            }
 
-            if (ray && this._checkElement(ray, element, screen)) {
-                result = element;
-                break;
+                // continue looking through all elements, only storing the closest one
+                const currentDistance = this._checkElement(ray, element, screen);
+                if (ray && currentDistance > 0 && currentDistance < closestDistance3d) {
```
slimbuck, line 883: >= 0, see related comment.

src/input/element-input.js
```diff
@@ -833,6 +861,13 @@ class ElementInput {
                 }
                 ray = rayScreen;
                 screen = true;
+
+                // break on the first element that hit
+                const hit = this._checkElement(ray, element, screen) > 0;
```
slimbuck, line 866: >= 0

src/input/element-input.js
jpauloruschel: Good catch - and actually since the logic for calling `_checkElement` was moved to inside the ifs, the variable `ray` is no longer needed, and the code can directly use either `ray3d` or `rayScreen`. Fixed on next commit.

slimbuck (approved):
This is great. Thanks for cleaning up the existing code too!

Issue #3492, opened 2021-09-17
PostEffect with needsDepthBuffer = true - is leaking VRAM on device resize.


Maksims:
When a canvas is resized, if post effect with `needsDepthBuffer = true` is used, it is leaking VRAM (not releasing previous depth buffers).

### Steps to Reproduce

Example is Model Viewer project, with SSAO post effect from [here](https://github.com/playcanvas/engine/blob/master/scripts/posteffects/posteffect-ssao.js).

https://playcanvas.com/project/831974/overview/needsdepthbuffer-vram-leak

1. Enable profiler for Launcher
2. Launch provided project
3. Resize window

### Issue:

Check `Texture Other` under `VRAM` panel, memory usage is constantly growing on each resize.

Disabling `needsDepthBuffer` in `ssao.js` eliviates an issue, but obviously, effect is not working then.

mvaligursky:
Thanks for the report @Maksims , got it sorted out. Out in the next engine release.

Issue #3488, opened 2021-09-16
playcanvas.github.io doesn't run on iOS 8 (iPhone 6) due to use of ES6


yaustar:
Example error: http://playcanvas.github.io/#/graphics/hardware-instancing
![image](https://user-images.githubusercontent.com/16639049/133636001-0f42c713-fa2d-46b3-a79d-8139753f798d.png)

willeastcott:
I suspect this is related to the use of the spread operator in object literals:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#browser_compatibility

mvaligursky:
at the moment engine itself does not support iOS 8 - see https://github.com/playcanvas/engine/issues/3658

willeastcott:
Even iPhone 4S supported iOS 9.3.6. iOS 8 itself is now 10 years old. Closing...

PR #3493, opened 2021-09-17
Improve SSAO script

Maksims:
Supports resizing of graphics device (usefull on mobile when changing orientation).
Also implements Down Sampling - to improve performance on lower-end by allowing to reduce texture resolution.
Support for device pixel ratio, so on mobile with either setting effect looks better.
Additionally, converts shader code from an array of strings to template literals, for easier readability and editing.

Updating this script, discovered VRAM leak bug: https://github.com/playcanvas/engine/issues/3492

### DIFF
--- a/scripts/posteffects/posteffect-ssao.js
+++ b/scripts/posteffects/posteffect-ssao.js
@@ -9,364 +9,406 @@
  * @description Creates new instance of the post effect.
  * @augments PostEffect
  * @param {GraphicsDevice} graphicsDevice - The graphics device of the application.
- * @param {any} ssaoScript - The script using the effect.
+ * @param {any} script - The script using the effect.
  */
-function SSAOEffect(graphicsDevice, ssaoScript) {
+function SSAOEffect(graphicsDevice, script) {
     pc.PostEffect.call(this, graphicsDevice);
 
-    this.ssaoScript = ssaoScript;
+    this.script = script;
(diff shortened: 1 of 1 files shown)


mvaligursky:
Great changes!

> converts shader code from an array of strings to template literals

this won't work on browsers without ES6, babel works its magic only on the engine code.

Maksims:
> this won't work on browsers without ES6, babel works its magic only on the engine code.

Is `/scripts/*` folder exception to modern JS?
Examples are using TypeScript.
And this https://github.com/playcanvas/engine/blob/master/scripts/textmesh/text-mesh.js is using post ES5.

mvaligursky:
people include scripts in the Editor projects directly and those don't get transpiled to ES5 at the moment unfortunatelly.
examples at the moment use build process .. but we still need to transpile them, see https://github.com/playcanvas/engine/issues/3488

Maksims:
I'll update the code for ES5 compatibility then.

mvaligursky:
> I'll update the code for ES5 compatibility then.

Perhaps ES5 version can be built .. see https://github.com/playcanvas/engine/pull/3655

willeastcott:
The problem is that scripts aren't modules.

mvaligursky:
Closing as this has been merged in as part of https://github.com/playcanvas/engine/pull/4339
Thanks @Maksims !

Issue #3212, opened 2021-06-03
iOS audio context suspended on visibility change


Kulodo:
See Forum post [https://forum.playcanvas.com/t/ios-audio-context-suspended-on-visibility-change/20524](https://forum.playcanvas.com/t/ios-audio-context-suspended-on-visibility-change/20524)

Here is an example that demonstrates the issue.

[https://playcanv.as/p/OerFeKXH/](url)
[https://playcanvas.com/editor/scene/1168873](url)

- load this PC app in iOS Safari. Tap button few times to hear button clicks.
- switch to a new tab eg google
- switch back to PC app.
- Tap button again. **Now there is no sound.**
Mac chrome, Mac Safari, and Android Chrome, do not have this problem

It seems that the audio context is suspended when it should be active.

The following does seem to work around the issue.

```
this.app.systems.sound.manager.on('resume', function() {
    if (this.app.systems.sound.manager.context.state !== "running") {
        this.app.systems.sound.manager.context.resume();
    }        
}, this);
```

Issue #2383, opened 2020-08-24
pc.platform.mobile returns false on iPad OS


yaustar:
See thread here: https://forum.playcanvas.com/t/bug-ipad-platform-detection-issue/14515

Now iPadOS is asking for the desktop version of the site which means that it is now harder to detect if an iPad is used.

The hacky workaround is to check if it's on an Apple platform and check max touch points as seen here: https://developer.apple.com/forums/thread/119186

iXperiential:
I wouldn't rely only on touch points as there is the edge case that some people have a touch screen on their Mac. I know it's unlikely but just in case combining with the pixel ratio does the trick:
    if (window.devicePixelRatio == 2 && navigator.maxTouchPoints > 0) {
        os = 'ipad';
    }
    else {
        os = 'Mac';
    }

PR #3499, opened 2021-09-20
Properly resume AudioContext on iOS platform.

jpauloruschel:
Fixes https://github.com/playcanvas/engine/issues/3212

This PR makes sure that the `AudioContext` is properly resumed on the iOS platform when context is switched by using the [AudioContext#resume API](https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/resume). It returns a future, so the callback is only triggered once `AudioContext` is actually resumed. For other platforms, or if `AudioContext` is not being used, the logic is unchanged.

According to https://developer.mozilla.org/en-US/docs/Web/API/BaseAudioContext/state , iOS Safari sets the state to `interrupted`.

### DIFF
--- a/src/sound/manager.js
+++ b/src/sound/manager.js
@@ -81,8 +81,15 @@ class SoundManager extends EventHandler {
     }
 
     resume() {
-        this.suspended = false;
-        this.fire('resume');
+        const resumeFunction = function () {
+            this.suspended = false;
+            this.fire('resume');
+        }.bind(this);
+
+        if (platform.ios && (hasAudioContext() || this._forceWebAudioApi))
+            this.context.resume().then(resumeFunction);
+        else
+            resumeFunction();
     }
 
     destroy() {
--- a/src/sound/manager.js
+++ b/src/sound/manager.js
@@ -81,14 +81,18 @@ class SoundManager extends EventHandler {
     }
 
     resume() {
-        const resumeFunction = function () {
+        const resumeFunction = () => {
             this.suspended = false;
             this.fire('resume');
-        }.bind(this);
+        };
 
-        if (platform.ios && (hasAudioContext() || this._forceWebAudioApi))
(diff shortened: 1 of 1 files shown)


src/sound/manager.js
```diff
@@ -81,8 +81,15 @@ class SoundManager extends EventHandler {
     }
 
     resume() {
-        this.suspended = false;
-        this.fire('resume');
+        const resumeFunction = function () {
+            this.suspended = false;
+            this.fire('resume');
+        }.bind(this);
```
slimbuck, line 87:
```suggestion
        const resumeFunction = () => {
            this.suspended = false;
            this.fire('resume');
        }
```

willeastcott, line 87:
```suggestion
        const resumeFunction = () => {
            this.suspended = false;
            this.fire('resume');
        };
```

willeastcott: I'm assuming the above code suggestion works OK. But using the arrow function should avoid the need to use `bind`.

slimbuck: Beat ya! :D

willeastcott: Yeah, except you deleted the trailing semi-colon, @slimbuck! 😆

jpauloruschel:
> I have a minor concern here. There another bug you might want to check out: #2383
> So it _seems_ that `pc.platform.ios` returns `false` on iPad at the moment? So your fix here might now work as expected on iPad.

Calling `AudioContext.resume()` works on all platforms - it is only the case that on iOS it is required. To be safe, I can remove the `platform.ios` entirely and always call `AudioContext.resume` when it is being used. What do you think?

willeastcott:
Oh interesting. Yeah, that sounds like a good solution to me.

willeastcott (approved):

vkalpias (approved):

PR #3508, opened 2021-09-22
Removed PlayCanvas version logging

yaustar:
The version and revision are already exported as pc.version and pc.revision in core.js

We have decided to remove the logging completely rather than have an option. If users/developers want to add it to the console, they will have to do so explicitly in the app as:

```
console.log("Powered by PlayCanvas " + pc.version + " " + pc.revision);
```

Fixes #3456

### DIFF
--- a/src/framework/application.js
+++ b/src/framework/application.js
@@ -1,4 +1,3 @@
-import { version, revision } from '../core/core.js';
 import { platform } from '../core/platform.js';
 import { now } from '../core/time.js';
 import { path } from '../core/path.js';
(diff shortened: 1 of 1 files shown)


src/framework/application.js
```diff
@@ -400,8 +400,6 @@ class Application extends EventHandler {
     constructor(canvas, options = {}) {
         super();
 
-        console.log("Powered by PlayCanvas " + version + " " + revision);
```
mvaligursky, line 403: remove the import on top as well

mvaligursky (approved):

src/framework/application.js
yaustar: Done, thanks!

willeastcott (approved):

PR #3517, opened 2021-09-26
More var to let and const

willeastcott:
More removal of the `var` keyword, largely focused on the framework sourcebase.


src/framework/application.js
```diff
@@ -1748,11 +1739,11 @@ class Application extends EventHandler {
      * });
      */
     renderLine(start, end, color) {
-        var endColor = color;
-        var options;
+        let endColor = color;
```
mvaligursky, line 1742: these render functions are all now moved to deprecated in my other PR.

willeastcott: So what needs to be done here?

mvaligursky: nothing for now I think .. we'll resolve to conflict when we get to it.

willeastcott: ok

mvaligursky (approved):

PR #3523, opened 2021-09-27
Better shader compilation error reporting

### LINKED ISSUE #3280: Better shader compilation error reporting
When a shader fails to compile, a whole shader is printed with errors at the end. Ideally we could only print lines around the error and the error message. Maybe attach complete shader after that as well.

Maksims:
Instead of printing the whole shader and error at the end, now it prints error first, and then 6 lines before, error line, and 6 lines of code after, that way it does not take the whole console with huge shader logs.

Fixes #3280 

Screenshot:
image

### DIFF
--- a/src/graphics/graphics-device.js
+++ b/src/graphics/graphics-device.js
@@ -3358,18 +3358,32 @@ class GraphicsDevice extends EventHandler {
         }
     }
 
-    _addLineNumbers(src) {
+    _addLineNumbers(src, infoLog) {
         if (!src)
             return "";
 
-        const lines = src.split("\n");
+        const lines = src.split('\n');
+        let code = '';
+        let from = 0;
+        let to = lines.length;
+        let errorLine = 0;
 
-        // Chrome reports shader errors on lines indexed from 1
-        for (let i = 0, len = lines.length; i < len; i++) {
-            lines[i] = (i + 1) + ":\t" + lines[i];
+        // if error is in the code, only show nearby lines instead of whole shader code
+        if (infoLog && infoLog.startsWith('ERROR:')) {
+            const match = infoLog.match(/^ERROR:\s([0-9]+):([0-9]+):/);
+            if (match) {
+                errorLine = parseInt(match[2], 10);
+                from = Math.max(0, errorLine - 6);
+                to = Math.min(lines.length, errorLine + 5);
+            }
         }
 
-        return lines.join("\n");
+        // Chrome reports shader errors on lines indexed from 1
+        for (let i = from; i < to; i++) {
+            code += (i + 1) + ":\t" + lines[i] + '\n';
+        }
+
+        return code;
     }
 
     postLink(shader) {
@@ -3391,11 +3405,13 @@ class GraphicsDevice extends EventHandler {
 
         // Check for errors
         if (!gl.getShaderParameter(glVertexShader, gl.COMPILE_STATUS)) {
-            console.error("Failed to compile vertex shader:\n\n" + this._addLineNumbers(definition.vshader) + "\n\n" + gl.getShaderInfoLog(glVertexShader));
+            const infoLog = gl.getShaderInfoLog(glVertexShader);
+            console.error("Failed to compile vertex shader:\n\n" + infoLog + "\n" + this._addLineNumbers(definition.vshader, infoLog));
             return false;
         }
         if (!gl.getShaderParameter(glFragmentShader, gl.COMPILE_STATUS)) {
-            console.error("Failed to compile fragment shader:\n\n" + this._addLineNumbers(definition.fshader) + "\n\n" + gl.getShaderInfoLog(glFragmentShader));
+            const infoLog = gl.getShaderInfoLog(glFragmentShader);
+            console.error("Failed to compile fragment shader:\n\n" + infoLog + "\n" + this._addLineNumbers(definition.fshader, infoLog));
             return false;
         }
         if (!gl.getProgramParameter(glProgram, gl.LINK_STATUS)) {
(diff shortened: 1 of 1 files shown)


src/graphics/graphics-device.js
```diff
@@ -3391,11 +3405,13 @@ class GraphicsDevice extends EventHandler {
 
         // Check for errors
         if (!gl.getShaderParameter(glVertexShader, gl.COMPILE_STATUS)) {
-            console.error("Failed to compile vertex shader:\n\n" + this._addLineNumbers(definition.vshader) + "\n\n" + gl.getShaderInfoLog(glVertexShader));
+            const infoLog = gl.getShaderInfoLog(glVertexShader);
+            console.error("Failed to compile vertex shader:\n\n" + infoLog + "\n" + this._addLineNumbers(definition.vshader, infoLog));
```
willeastcott, line 3409:
```suggestion
            const errorCode = this._addLineNumbers(definition.vshader, infoLog));
            console.error(`Failed to compile vertex shader:\n\n${infoLog}\n${errorCode}`);
```

src/graphics/graphics-device.js
```diff
@@ -3391,11 +3405,13 @@ class GraphicsDevice extends EventHandler {
...
             return false;
         }
         if (!gl.getShaderParameter(glFragmentShader, gl.COMPILE_STATUS)) {
-            console.error("Failed to compile fragment shader:\n\n" + this._addLineNumbers(definition.fshader) + "\n\n" + gl.getShaderInfoLog(glFragmentShader));
+            const infoLog = gl.getShaderInfoLog(glFragmentShader);
+            console.error("Failed to compile fragment shader:\n\n" + infoLog + "\n" + this._addLineNumbers(definition.fshader, infoLog));
```
willeastcott, line 3414:
```suggestion
            const errorCode = this._addLineNumbers(definition.fshader, infoLog));
            console.error(`Failed to compile fragment shader:\n\n${infoLog}\n${errorCode}`);
```

willeastcott:
Pretty cool - I like this. 😄

mvaligursky:
Do we still have some way to see the whole shader? I often look at the whole shader to see why the error or what other parts of the code are there / missing.

mvaligursky:
And another question - does it work in other browsers as well?

Maksims:
> And another question - does it work in other browsers as well?

Tested on Chome, Microsoft Edge, Firefox, works identically.

> Do we still have some way to see the whole shader? I often look at the whole shader to see why the error or what other parts of the code are there / missing.

Console log is very hard to use for actual code browsing. Most workflow where this error happens is when developer is writing shaders, for chunks or custom for whole material. So they have context of a whole shader then.

Reporting of what chunk or material this shader was related to - is out of the scope of this change, but would be useful indeed.
Nonetheless I often have to do something like:
```js
console.log(pc.app.root.findByName('Model').render.meshInstances[0].material.shader.definition);
```
And copy `vshader`/`fshader` from there, or even `shader._variants`.

slimbuck:
Quite often shaders don't break during development, but rather during testing when we're preparing to release a new version of the engine. I also think it's important we can easily get to the whole shader when it breaks.

mvaligursky:
We could add whole shader to the console this way:

```
        const longstring = `
            // Create an entity with a camera component
            const camera = new pc.Entity();
            type: "spot",
            color: color,
        `;

        console.log("error in shader: ", { shader: longstring });
```

This add this to the console:
Screenshot 2021-09-27 at 16 13 35

and you can click on that little triangle in front of 'shader' to see the whole string.

Maksims:
> We could add whole shader to the console this way:
> 
> ```
>         const longstring = `
>             // Create an entity with a camera component
>             const camera = new pc.Entity();
>             type: "spot",
>             color: color,
>         `;
> 
>         console.log("error in shader: ", { shader: longstring });
> ```
> 
> This add this to the console:
> Screenshot 2021-09-27 at 16 13 35
> 
> and you can click on that little triangle in front of 'shader' to see the whole string.

I will experiment on how different consoles show passed objects, but good idea!
Also important to notice, is how Editor logger outputs it.

Maksims:
Added more detailed object with error information, but only for DEBUG build, because retaining Shader pointer or large strings in console - leads to memory leaking (as they won't be GC'ed).

image

src/graphics/graphics-device.js
```diff
@@ -3391,11 +3409,25 @@ class GraphicsDevice extends EventHandler {
...
         if (!gl.getShaderParameter(glVertexShader, gl.COMPILE_STATUS)) {
-            console.error("Failed to compile vertex shader:\n\n" + this._addLineNumbers(definition.vshader) + "\n\n" + gl.getShaderInfoLog(glVertexShader));
+            const infoLog = gl.getShaderInfoLog(glVertexShader);
+            const [code, error] = this._processError(definition.vshader, infoLog);
+            // #if _DEBUG
+            error.shader = shader;
+            console.error(`Failed to compile vertex shader:\n\n${infoLog}\n${code}`, error);
+            // #else
+            console.error(`Failed to compile vertex shader:\n\n${infoLog}\n${code}`);
+            // #endif
```
mvaligursky, line 3419: Could you also wrap this block in a function so that we don't have this pretty much identical code twice please.

mvaligursky: perhaps even 'getShaderParameter' line

mvaligursky (approved):
This looks awesome!

mvaligursky:
This is working really great, thanks for doing it @Maksims !
Saved me lots of annoying scrolling and clicking in the last 2 days for sure.

Maksims:
It's funny how we lived with it for years, and it only took a couple of minutes to make this improvement.
Quality of life - this should be a label in issues ;)

PR #3547, opened 2021-10-06
[Fix] Fix to texture based morph targets to occasionally morph incorrect vertex

mvaligursky:
- fix involves adding a small fraction to the number to garantee it does not get converted to number one smaller. Typical workaround for these situations.


willeastcott (approved):

src/scene/morph.js
```diff
@@ -9,6 +9,9 @@ import { getApplication } from '../framework/globals.js';
 
 import { BUFFER_STATIC, TYPE_FLOAT32, SEMANTIC_ATTR15, ADDRESS_CLAMP_TO_EDGE, FILTER_NEAREST, PIXELFORMAT_RGBA16F, PIXELFORMAT_RGB32F } from '../graphics/constants.js';
 
+// value added to floats which are used as ints on the shader side to avoid values being rounded to one less occasionally
+const _floatRounding = 0.2;
```
willeastcott, line 13: Just curious, but why 0.2 specifically?

Issue #3461, opened 2021-09-07
Application.update() updates all scripts in all apps even if fired only on one application


OlegGedzjuns:
### Description
When creating multiple applications, calling the update method on any app will update all scripts on all applications.
Issue possibly is related to ComponentSystem.

When we are calling app.update() we are triggering static ComponentSystem.update().
https://github.com/playcanvas/engine/blob/df4bb2aae6ff70f37feb9f89f37e85c3d61b90fc/src/framework/application.js#L1165

ComponentSystem update calls all updates on components in all apps.
https://github.com/playcanvas/engine/blob/df4bb2aae6ff70f37feb9f89f37e85c3d61b90fc/src/framework/components/system.js#L47

* URL to a simple, reproducible test case that illustrates the problem clearly
https://github.com/OlegGedzjuns/playcanvas-application-update
* The platform(s)/browser(s) where the issue is seen
Node.js environment

mvaligursky:
@vkalpias - any thoughts on this?

PR #3557, opened 2021-10-11
Make ComponentSystem and Default Material, Application scope

OlegGedzjuns:
Currently, ComponentSystem is global and orchestrates systems globally. This makes multi-application usage somewhat difficult, especially when applications have to be added/destroyed constantly and have various update rates.

Also, default material was global, this lead to an issue when in multi-application case we destroy one application, this lead to default material destruction which then broke other applications that are still running.

This change improves the usage of an engine in a render-less server-side context.
As well as the client-side cases with multiple applications on a page.

There are no behavior changes.
Please test on your various projects with this build.

Fixes: https://github.com/playcanvas/engine/issues/3461


willeastcott:
What's your verdict on this PR @vkalpias? I'd like it to go out with the next minor release if possible?

alamboley:
Hello, running into an issue when upgrade from 1.47 to 1.48 or later:
`pc.ComponentSystem.initialize is not a function` seems to be related to this PR.

```js
assetsToLoad.forEach(function(assetToLoad) {
        assetToLoad.ready(function(asset) {
            count++;
            loadingElement.width = 200 * (count / assetsToLoad.length);
            if (count == assetsToLoad.length) {
                this.app.scenes.loadScene(sceneToLoad.url, function (err2, scene2) {
                    if (!err2) {
                        oldHierarchy.destroy();
                        pc.ComponentSystem.initialize(scene2.root);
                        pc.ComponentSystem.postInitialize(scene2.root);
                    } else
                        console.error(err2);
                }.bind(this));
            }
        }.bind(this));
        this.app.assets.load(assetToLoad);
    }.bind(this));
}
```

How this code should be updated?

mvaligursky:
I believe this is what you need to call:
```
     // before callback:
     const self = this;

      // inside callback
      self.app.systems.fire('initialize', scene2.root);
      self.app.systems.fire('postInitialize', scene2.root);

```

willeastcott:
Actually, you don't need to cache `this` in `self` because the function has `.bind(this)`.

alamboley:
Thanks for the code @mvaligursky! I'm a bit worried firing an event is less intuitive than calling a method, and hard to have code completion in editor with that kind of thing... but it works fine!

Maksims:
Those events are actually not public, so use them with caution.

willeastcott:
Yeah, while that code change may work, this isn't public API. This looks to me like there's either a bug or a hole in the API. Can anyone who knows more than me comment?

yaustar:
loadScene used to be the only way to load a scene hierarchy and settings with one HTTP request. It was also exposed for engine only uses as it was the only way to load the first scene IIRC.

In the API docs, the function is flagged as an internal function: https://developer.playcanvas.com/api/pc.SceneRegistry.html#loadScene

Now that it is possible to preload the scene data, it is possible (a bit verbose) to load both the scene hierarchy and settings via one HTTP call and using the public API 

```
var self = this;
this.scene = this.app.scenes.find(this.sceneName);

// Load the scene data via HTTP
this.app.scenes.loadSceneData(this.scene, function (err, sceneItem) {
    if (err) {
        console.error(err);
    } else {
        // Get a reference to the scene's root object
        var oldHierarchy = self.app.root.findByName('Root');
        oldHierarchy.destroy();

        // Load both the hierarchy and scene settings
        self.app.scenes.loadSceneHierarchy(self.scene, function() {});
        self.app.scenes.loadSceneSettings(self.scene, function() {});

        // Unload the data
        self.app.scenes.unloadSceneData(self.scene);
    }
});
```

PR #3578, opened 2021-10-14
Rearanged main clustered lighting loop to avoid unrolling and improve compile time

mvaligursky:
- in the original call where lighting was called manually 4 times, the compilation time on my MacBookPro per simple shader using clustered lighting was around 180ms.
- with changed code, with just a single call to lighting, the compilation time dropped to around 70ms.


slimbuck:
And no change in runtime perf?

PR #3593, opened 2021-10-19
Anim layer mask updates

ellthompson:
Addresses some issues with the Anim layer masks feature

- Quaternion blending was not working as expected when any of the blended quaternion's directions we're opposite that of the identity quaternion. This has been remedied by blending these quaternions directly, rather than through a slerp function which always interpolates two quaternions using the shortest arc.
- The component layer assignAnimation function now creates a transition between the START state and the assigned animation state if it's the first assignment in that layer. This aligns this functions behaviour with that of the component's assignAnimation function.
- Creare an assignMask function to the component layer which allows masks to be swapped out while an anim component is running. Also gives an example of the expected format of a mask.
- The AnimComponent#addLayer function now correctly returns the created layer.

### DIFF
--- a/src/framework/components/anim/component-layer.js
+++ b/src/framework/components/anim/component-layer.js
@@ -63,10 +63,6 @@ class AnimComponentLayer {
         this._controller.update(dt);
     }
 
-    assignMask(mask) {
-        this._controller.assignMask(mask);
-    }
-
     /**
      * @function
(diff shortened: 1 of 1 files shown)


ellthompson:
Order in this instance is just the current length of the layers array. I've moved this assignment to the inside of the _addLayer function to make things clearer.

PR #3606, opened 2021-10-22
Type information improvement in Forward-renderer

mvaligursky:
Added JSDOC type information for some members, which allows intellisense and also shows errors if APIs on those type are called incorrectly. I'm keen to go forward with these kind of things, as it makes code a lot safer to refactor.

Relevant discussion: https://github.com/playcanvas/engine/discussions/3358
(@vkalpias is using this in new Editor API as well)

Also made boneTexture local variable instead of global


willeastcott (dismissed):

willeastcott:
Superseded by #3725 which successfully leverages `jsdoc-tsimport-plugin`.

PR #3609, opened 2021-10-23
Update JSDoc examples to use ES6

willeastcott:
This PR updates the JSDoc example code to use:

* `let` and `const` instead of `var`
* Arrow functions

Note that this does mean that some example code will be incompatible with IE11. According to [StatCounter](https://gs.statcounter.com/browser-market-share), IE (all versions, not just IE11) accounts for 0.57% of the browser market now.


willeastcott:
Closing because merge conflicts are too extensive.

PR #3633, opened 2021-10-28
Add prefiltering helper functions

slimbuck:
This PR adds a few helper functions for generating prefiltered lighting data:
```javascript
generateSkyboxCubemap(source, size);            // generate a skybox cubemap in the correct pixel format
generatePrefilteredCubemaps(source, mipmaps);   // generate an array of 6 prefiltered cubemaps
generatePrefilteredCubemap(source);             // generate a cubemap with prefiltered lighting mipmaps
```

They are all marked as private for now, but will be made public once the API has settled.

### DIFF
--- a/src/graphics/prefilter.js
+++ b/src/graphics/prefilter.js
@@ -11,7 +11,7 @@ class Prefilter {
      * @name isValidSkyboxCubemap
      * @description Returns true if the texture is valid for use as a skybox.
      * @param {Texture} texture - The texture to test.
-     * @returns {boolean} True if the texture is valid for use as a cubemap otherwize false.
+     * @returns {boolean} True if the texture is valid for use as a cubemap otherwise false.
      */
     static isValidSkyboxCubemap(texture) {
         return texture && texture.cubemap && (texture.type === TEXTURETYPE_DEFAULT || texture.type === TEXTURETYPE_RGBM);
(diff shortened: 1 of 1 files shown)


mvaligursky:
Personally I've created few pure static classes lately as well, and I'm pretty keen on it - it makes code base feel more unified to me. Function is written this way:

`static doSomething(param);`

instead of 
`doSomething: function (param)`

so we avoid having to have two types of function definitions, one for functions in classes, one for in objects.

src/index.js
```diff
@@ -59,6 +59,7 @@ export { reprojectTexture } from './graphics/reproject-texture.js';
 export { programlib } from './graphics/program-lib/program-lib.js';
 export { shaderChunks } from './graphics/program-lib/chunks/chunks.js';
 export { GraphicsDevice } from './graphics/graphics-device.js';
+export { Prefilter } from './graphics/prefilter.js';
```
mvaligursky, line 62: this could be alphabetically sorted.

Maksims:
> Personally I've created few pure static classes lately as well, and I'm pretty keen on it - it makes code base feel more unified to me. Function is written this way:
> 
> `static doSomething(param);`
> 
> instead of `doSomething: function (param)`
> 
> so we avoid having to have two types of function definitions, one for functions in classes, one for in objects.

Sorry, but I'm talking about public API design, not how it is written.

src/index.js
willeastcott: It kind of already is if you do 'a..z' then 'A..Z'. 😄

zishiwu123:
My bad. I meant to reference #3663 but accidentally typed 3633 in my PR.

PR #3636, opened 2021-10-28
Simplify skybox rotation

slimbuck:
This PR changes the rendering of the skybox:
- apply skybox rotation in the vertex shader instead of pixel shader
- refrain from generating a shader variant when rotation is enabled
- remove unused skybox chunks


mvaligursky:
This looks great!

PR #3653, opened 2021-11-01
Fix undefined positions bug

vmwxiong:
Was running into this error:
`TypeError: Cannot read properties of undefined (reading '0')`

Took a look at the code and pretty sure that this `return` on line 399 is incorrect. Not sure if it should be a `break` or a `return []`, but returning `undefined` seems like an error.

### DIFF
--- a/src/framework/components/layout-group/layout-calculator.js
+++ b/src/framework/components/layout-group/layout-calculator.js
@@ -396,7 +396,8 @@ function createCalculator(orientation) {
             const line = lines[lineIndex];
 
             if (line.length === 0) {
-                break;
+                positionsAllLines.push([]);
+                continue;
             }
 
             const positionsThisLine = [];
--- a/src/framework/components/layout-group/layout-calculator.js
+++ b/src/framework/components/layout-group/layout-calculator.js
@@ -396,7 +396,7 @@ function createCalculator(orientation) {
             const line = lines[lineIndex];
 
             if (line.length === 0) {
-                return;
+                break;
             }
 
             const positionsThisLine = [];


willeastcott:
Well, for sure, there's a bug there because, yes, the function is returning the wrong type. As for whether the `return` needs to be swapped with a `break`, I'm not sure. Definitely needs some investigation. Do you have an easy way to reproduce this bug?

vmwxiong:
I'm not 100% sure, but I believe it happens if you create a layout group
with only a single a layout child that is set to exclude itself from the
layout

vmwxiong:
The workaround for this at the moment is to add an element entity with 0 width and height to act as a placeholder so that this array is never totally empty

jpauloruschel:
Hi and thank you for the contribution!

This piece of code is calculating the positions of all elements per `line` (*visually, a row or column*). Lines 83-85 (https://github.com/playcanvas/engine/blob/dev/src/framework/components/layout-group/layout-calculator.js#L83) figure out the elements in each `line`, the sizes if those elements, and the positions. When there are no elements in a group, the code still generates a single empty `line`.

All those functions assume that the sizes of those arrays are the same, such that: for any `x`, `lines[x]`, `sizes[x]` and `positions[x]` all reference the same visual row. 

In order to maintain that consistency, I think it's best to do something like this instead:

```
if (line.length === 0) {
    positionsAllLines.push([]);
    continue;
}
```

vmwxiong:
Makes sense! Updated to the usggestion.

willeastcott (approved):

jpauloruschel (approved):

PR #3655, opened 2021-11-02
Add vox loading script

slimbuck:
This PR adds support for vox files as an engine script.

output

The script is written as an es6 module in `/scripts/parsers/vox-parser.mjs`. The rollup config has been updated to generate an es5 version located at `/scripts/parsers/vox-parser.js`.

In order to get external container scripts to work, the following changes to container resource were necessary:
- rename `ContainerResource` to `GlbContainerResource` (since it's glb container specific)
- add a base interface `ContainerResource` to document the API
- refrain from exporting `ContainerResource`, since it's not needed
- allow `ContainerHandler` to support multiple handlers for different file formats, much like the `TextureHandler`

NOTE:  we should consider whether it's necessary to export all the handler interfaces from the engine. Exporting an interface should only be needed if users must create new instances of the class or inherit from it. In many cases, I think the resource handlers don't need exporting.


src/resources/container.js
```diff
@@ -70,17 +23,15 @@ class ContainerResource {
...
+        return null;
     }
 
     /**
      * @function
      * @name ContainerResource#instantiateRenderEntity
      * @description Instantiates an entity with a render component.
      * @param {object} [options] - The initialization data for the render component type {@link RenderComponent}.
-     * @returns {Entity} A hierarchy of entities with render components on entities containing renderable geometry.
+     * @returns {Entity} A hierarachy of entities with render components on entities containing renderable geometry.
```
willeastcott, line 34: Spelling regression?

src/resources/container.js
```diff
@@ -343,9 +61,9 @@ class ContainerResource {
  * @implements {ResourceHandler}
  * @classdesc Loads files that contain multiple resources. For example glTF files can contain
  * textures, models and animations.
- * The asset options object can be used to pass load time callbacks for handling the various resources
- * at different stages of loading. The table below lists the resource types and the corresponding
- * supported process functions.
+ * For glTF Files, the asset options object can be used to pass load time callbacks for handling
```
willeastcott, line 64:
```suggestion
 * For glTF files, the asset options object can be used to pass load time callbacks for handling
```

src/resources/parser/glb-container-resource.js
```diff
@@ -0,0 +1,313 @@
...
+// Container resource returned by the GlbParser. Implements the ContainerResource interface.
+class GlbContainerResource {
+    constructor(data, asset, assets, defaultMaterial) {
+        const createAsset = function (type, resource, index) {
+            const subAsset = GlbContainerResource.createAsset(asset.name, type, resource, index);
+            assets.add(subAsset);
+            return subAsset;
+        };
+
+        let i;
```
willeastcott, line 18: Declare in each `for` loop. That's generally how the rest of the engine codebase declares `for` loops now.

scripts/parsers/vox-parser.mjs
```diff
@@ -0,0 +1,485 @@
...
+
+        if (!voxFrame) {
+            return null;
+        }
+
+        const flattened = voxFrame.flattened;
+
+        const positions = [];
+        const normals = [];
+        const colours = [];
```
willeastcott, line 225: US spelling please.

scripts/parsers/vox-parser.mjs
```diff
@@ -0,0 +1,485 @@
...
+        this.timer = 0;
+        this.fps = 10;
+    }
+
+    update(dt) {
+        if (this.playing) {
+            this.timer += dt;
+        }
+
+        const meshInstances = this.entity.render?.meshInstances || this.entity.model?.meshInstances;
```
willeastcott, line 377: Doesn't the Vox system rely solely on render components now that the model component is legacy?

scripts/parsers/vox-parser.mjs
slimbuck: shaking fist emoji!! ... but ok :)

src/resources/container.js
slimbuck: done, thanks

src/resources/container.js
slimbuck: done, thanks

src/resources/parser/glb-container-resource.js
slimbuck: done

willeastcott (approved):

scripts/parsers/vox-parser.mjs
slimbuck: I haven't implemented `VoxContainerResource.instantiateModelEntity()`, but really should since it's part of the public API. Perhaps we can document that model support only exists for GLB container resources instead?

willeastcott: Fine by me.

mvaligursky (approved):
looks really good. Two questions:
- we should try and not build the script each time we build the engine .. impact is small for now, but can grow
- could the rollup config be set up the way we can easily add more scripts? Current solution seems to support just one.

mvaligursky:
Also, maybe we could have an engine example with this?

slimbuck:
Added watch:target scripts to speed up development and recompiling.

slimbuck:
> o, maybe we could have an engine example with th

Will add an engine example in a followup PR.

Issue #3652, opened 2021-11-01
Support glTF extension: KHR_materials_ior


mvaligursky:
https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_ior/README.md

The dielectric BRDF of the metallic-roughness material in glTF uses a fixed value of 1.5 for the index of refraction. This is a good fit for many plastics and glass, but not for other materials like water or asphalt, sapphire or diamond. This extensions allows users to set the index of refraction to a certain value.

PR #3656, opened 2021-11-03
Grabpass functionality refactor and mipmap support for webgl2

mvaligursky:
- Refactored GrabPass class out of the GraphicsDevice
- added support for mipmap generation to allow future support for glass roughness use case (webgl2 only for now)
- updated example to test it

Screenshot 2021-11-03 at 18 39 42
 
related to https://github.com/playcanvas/engine/issues/3652
further improvements: https://github.com/playcanvas/engine/issues/3657


src/graphics/grab-pass.js
```diff
@@ -0,0 +1,158 @@
...
+import { Texture } from "./texture.js";
+
+// A class used by WebGl device to handle the capture of the current framebuffer into a texture
+// to be used by following draw calls to implement refraction
+class GrabPass {
+    constructor(device, useAlpha) {
+        this.device = device;
+        this.useAlpha = useAlpha;
+
+        // on webgl1 we cannot mipmap not power of two texture, so use additional power of two texture so disable it for now
```
willeastcott, line 12: This comment doesn't make sense when I read it back. Can you revise it please?

src/graphics/grab-pass.js
```diff
@@ -0,0 +1,158 @@
...
+            this.texture = null;
+        }
+    }
+
+    create() {
+
+        if (!this.texture) {
+
+            const texture = new Texture(this.device, {
+                format: this.useAlpha === false ? PIXELFORMAT_R8_G8_B8 : PIXELFORMAT_R8_G8_B8_A8,
```
willeastcott, line 39:
```suggestion
                format: this.useAlpha ? PIXELFORMAT_R8_G8_B8_A8 : PIXELFORMAT_R8_G8_B8,
```

src/graphics/grab-pass.js
```diff
@@ -0,0 +1,158 @@
...
+        // #if _DEBUG
+        device.popMarker();
+        // #endif
+
+        return true;
+    }
+
+    generateMipmaps() {
+
+        // TODO: imlement support for webgl1, which requires the texture to be a power of two, by first downscaling
```
willeastcott, line 137:
```suggestion
        // TODO: implement support for webgl1, which requires the texture to be a power of two, by first downscaling
```

willeastcott (approved):

Maksims:
Sorry, but am I missing something, but it is very unclear from example how to ensure grabpass between either layers or before some mesh rendering?

mvaligursky:
> Sorry, but am I missing something, but it is very unclear from example how to ensure grabpass between either layers or before some mesh rendering?

This is the answer, no change to this in the PR: https://github.com/playcanvas/engine/blob/06a237a458577d71cf654eaae09ff3bec0fbbf78/examples/src/examples/graphics/grab-pass.tsx#L29-L31

It's not ideal, I agree .. and this will change when we implement refraction support directly to the material (glTF extension)

src/graphics/grab-pass.js
mvaligursky: done, thanks.

Maksims:
> It's not ideal, I agree .. and this will change when we implement refraction support directly to the material (glTF extension)

Oh, I see. It detects if it is used, and then does copy buffer.
Looking forward to seeing a nice public API, that will allow to control the order and flow by the developer. I'm actually in need of this on one project as well, for a more efficient refraction workflow. Got the depth data, but not the color data.

It's worth thinking of API together with API for Depth data. As they both are color buffers, and one nice API that will allow controlling either "and/or" of them - would be great.

PR #3659, opened 2021-11-03
Refactor some single iteration loops

willeastcott:
Address all errors from ESLint's [`no-unreachable-loop`](https://eslint.org/docs/rules/no-unreachable-loop) rule. Using `Object.keys` is, in principle, slower than using a `for..in` loop, but it's more concise and easier to read, plus the instances refactored here are in non-performance critical code.

This PR also uses more ES6 features in the now legacy `AnimationComponent` (arrow functions and defaults arguments).

### DIFF
--- a/src/framework/components/animation/component.js
+++ b/src/framework/components/animation/component.js
@@ -356,8 +356,10 @@ class AnimationComponent extends Component {
 
         if (!data.currAnim && data.activate && data.enabled && this.entity.enabled) {
             // Set the first loaded animation as the current
-            const animName = Object.keys(data.animations)[0];
-            this.play(animName);
+            const animations = Object.keys(data.animations);
+            if (animations.length > 0) {
+                this.play(animations[0]);
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/framework/components/animation/component.js
```diff
@@ -49,7 +49,7 @@ class AnimationComponent extends Component {
      * @param {number} [blendTime] - The time in seconds to blend from the current
      * animation state to the start of the animation being set.
      */
-    play(name, blendTime) {
+    play(name, blendTime = 0) {
```
yaustar, line 52: Change JSDocs to say what the default value should be

willeastcott: Done.

mvaligursky (approved):
looks good

Maksims:
If `Object.keys` are to be run on every or every other frame (frequently), it can lead to significant memory allocations if the object provided is large. Which then can lead to GC stalls.

Even if it is to be executed rarely, e.g. on asset load/parse, it can lead to a longer GC cycle later. Which is always to be avoided, as from such small pieces it piles up.

Also, readability is arguable, as it is more of a taste.

willeastcott:
None of the uses of `Object.keys` here are from the main application loop. I'd argue the refactor in the I18n.js in particular is far easier to read.

Issue #3658, opened 2021-11-03
Engine no longer supports iOS 8


yaustar:
On startup, iOS 8 gives the following error on the default scene setup 

Forum: https://forum.playcanvas.com/t/iphone4s-revisited/22666

From the debug build

```
8a96-51-148-130-218.ngrok.io/:4 Viewport argument key "viewport-fit" not recognized and ignored.
playcanvas-stable.dbg.js:33551 TextDecoder not supported - pc.Untar module will not work
playcanvas-stable.dbg.js:70501 Powered by PlayCanvas 1.49.0 eb609bf4e
playcanvas-stable.dbg.js:14675 WebGL: INVALID_OPERATION: texImage2D: type HALF_FLOAT_OES but ArrayBufferView is not NULL
playcanvas-stable.dbg.js:14679 Above error related to HALF_FLOAT_OES can be ignored, it was triggered by testing half float texture support
playcanvas-stable.dbg.js:6718 DEPRECATED: pc.Application#loadScene is deprecated. Use pc.Application#scenes and pc.SceneRegistry#loadScene instead.
__start__.js:203 Error while loading scene JSON 1265092.json: undefined is not a function (evaluating 'Array.from(o)')
_unsupportedIterableToArray@https://8a96-51-148-130-218.ngrok.io/playcanvas-stable.dbg.js:1438:53
_createForOfIteratorHelperLoose@https://8a96-51-148-130-218.ngrok.io/playcanvas-stable.dbg.js:1454:61
_processParameters@https://8a96-51-148-130-218.ngrok.io/playcanvas-stable.dbg.js:13618:56
updateUniforms@https://8a96-51-148-130-218.ngrok.io/playcanvas-stable.dbg.js:13764:27
submitCasters@https://8a96-51-148-130-218.ngrok.io/playcanvas-stable.dbg.js:22613:29
render@https://8a96-51-148-130-218.ngrok.io/playcanvas-stable.dbg.js:22667:24
renderShadows@https://8a96-51-148-130-218.ngrok.io/playcanvas-stable.dbg.js:23955:32
renderComposition@https://8a96-51-148-130-218.ngrok.io/playcanvas-stable.dbg.js:24665:24
render@https://8a96-51-148-130-218.ngrok.io/playcanvas-stable.dbg.js:71155:35
https://8a96-51-148-130-218.ngrok.io/playcanvas-stable.dbg.js:71770:23
[...]

PR #3660, opened 2021-11-04
[Fix] Fixed looping over Set to be compatible with old browsers

mvaligursky:
Uses Set.foreach to loop over it, instead of less supported for .. in

- Fixes https://github.com/playcanvas/engine/issues/3658
- also removes 2 lint warnings

### DIFF
--- a/src/scene/composition/layer-composition.js
+++ b/src/scene/composition/layer-composition.js
@@ -112,7 +112,9 @@ class LayerComposition extends EventHandler {
         }
 
         // all other clusters
-        this._worldClusters.forEach((cluster) => cluster.destroy());
+        this._worldClusters.forEach((cluster) => {
+            cluster.destroy();
+        });
         this._worldClusters = null;
     }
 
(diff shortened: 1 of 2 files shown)


slimbuck (approved):

Maksims:
It's worth updating https://github.com/playcanvas/engine/blob/dev/.github/CONTRIBUTING.md regarding this particular case.

willeastcott:
Wow, lots of that doc should be deleted now. For example, arrow functions should be used now so using `self = this` no longer applies. A lot of the formatting recommendations are enforced by ESLint, so again, a bit redundant. We should probably start recommending private class member/function usage too.

Maksims:
> Wow, lots of that doc should be deleted now. For example, arrow functions should be used now so using `self = this` no longer applies. A lot of the formatting recommendations are enforced by ESLint, so again, a bit redundant. We should probably start recommending private class member/function usage too.

Yep, worth updating that doc frequently, as it is the central place of communicating about norms and styles.

Regarding private static methods/properties on class - be careful, as it is only available in iOS 14.1. So if it is used in examples code - it can lead people to use it in production code.

willeastcott:
Yeah, I was just thinking of using it in core engine source, which would then be transformed by Babel of course.

Issue #3200, opened 2021-05-27
ES6+ features to be well tested and allowed listed in CONTRIBUTING.md before using


Maksims:
It is important to keep engine code well maintained, and still support all major platforms, the engine promises backwards compatibility.

## Risks of adding new ES6+ features to codebase:
1. Lack of platform support for ES6+ engine version users.
2. Bad transcompilation path to ES5 leading to large polyfills and/or filesize growth.

## Benefits:
1. Easier to read code.
2. More trivial solutions, and easier to maintain.
3. Faster execution for the ES6+ version as browser engines evolve.

## Whitelisted features:
- [x] `let`
- [x] `const`
- [x] `for of`
- [x] arrow functions
- [x] classes
- [x] default parameters
- [x] modules
- [x] optional chaining
- [x] template literals

## Awaiting adding to [CONTRIBUTING.md](https://github.com/playcanvas/engine/blob/master/.github/CONTRIBUTING.md):
- [x] `Map`
- [x] `Set`

## Currently used, and untested:
- [ ] `static` on `class`
*only recently been added to Safari, leading to a large portion of module engine version users not supporting it*
- [ ] `padEnd`
- [ ] spread syntax
- [ ] `Promise`

## Considerations:
- [ ] `**` exponentiation operator
- [ ] `padStart`
- [ ] destructing
- [ ] rest parameters
- [ ] custom iterator functions
- [ ] `Symbol.iterator`
- [ ] `Symbol`
- [ ] Subclassable build-ins (e.g. Class extends Array)
- [ ] `async`, `await`
- [ ] `Reflect`
- [ ] `Object.values`
- [ ] `Object.entries`
- [ ] RegExp Named Group Captures `?`

## Features not to use:

kungfooman:
Object destructuring looks good to me too, only tested this case tho.

How it is now: https://github.com/playcanvas/engine/blob/6db49b7cb172e9476046d074711156e08d0ba754/src/input/element-input.js#L638-L641

How it could look:

```js
const {element, camera, x, y} = touchInfo;
```

The ES5 transcompilation looks like:

```js
var element = touchInfo.element,
    camera = touchInfo.camera,
    x = touchInfo.x,
    y = touchInfo.y;
```

How ES5 transcompilation looks with current code:

```js
var element = touchInfo.element;
var camera = touchInfo.camera;
var x = touchInfo.x;
var y = touchInfo.y;
```

(result is less code in ES5 as well as in ES6)

[...]

PR #3662, opened 2021-11-04
Set and Map are officially permitted in the engine, with limitations

mvaligursky:
related to https://github.com/playcanvas/engine/issues/3200


willeastcott (approved):

willeastcott:
Maybe caveat that Set iteration is no allowed?

mvaligursky:
> Maybe caveat that Set iteration is no allowed? And Map too if that applies there too?

yep just did.

Maksims:
Wandering if babel has an extension to convert `for..of` for Set/Map to ES5 versions.

Issue #3664, opened 2021-11-04
RenderComponent#material returns incorrect material


yaustar:
Engine version 1.47.2

API docs: https://developer.playcanvas.com/en/api/pc.RenderComponent.html#material

This implies that on primitives, this should return the material applied to the render component. However, it returns the Default Material:

Example project: https://playcanvas.com/project/847669/overview/render-material-bug

Launch and open the console, it should say: 'This is the new material' but it says 'Default Material'.

Related thread: https://forum.playcanvas.com/t/material-update-not-working/22710

PR #3665, opened 2021-11-04
[Fix] RenderComponent.material returns assigned material for primitives

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/3664

RenderComponent has material property, used to access a single material in case of non-asset render. This was up  not being set during material loading.

### DIFF
--- a/src/framework/components/render/component.js
+++ b/src/framework/components/render/component.js
@@ -387,22 +387,32 @@ class RenderComponent extends Component {
         }
     }
 
+    _updateMainMaterial(index, material) {
+        // first material for primitives can be accessed using material property, so set it up
+        if (index === 0) {
+            this.material = material;
+        }
+    }
+
     _onMaterialLoad(index, component, asset) {
         if (this._meshInstances[index]) {
             this._meshInstances[index].material = asset.resource;
         }
+        this._updateMainMaterial(index, asset.resource);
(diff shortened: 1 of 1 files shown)


src/framework/components/render/component.js
```diff
@@ -682,7 +692,8 @@ class RenderComponent extends Component {
         });
     }
 
-    set materialAssets(value = []) {
+    set materialAssets(value) {
+        value = value || [];
```
slimbuck, line 696: was this change intentional?

willeastcott: Yeah, I thought this looked odd too. If there is some reason why default parameters shouldn't be used in a setter, I would at least recommend a comment because otherwise, somebody will just switch it back. :smile:

mvaligursky: to me based on looking online it seems ES6 does not support default parameters in setters.
Our lint flagged it as an issue.

mvaligursky: I tested it and the default value for setter works. When I assign undefined to it, it uses the default value instead. It seems the warning is only for TS, not for JS - can it be globally hidden?
Screenshot 2021-11-05 at 09 12 47

slimbuck (approved):

PR #3677, opened 2021-11-09
[Fix] Glb parser handles a scene without nodes

mvaligursky:
This fixes a glb loading issue mentioned here:
https://forum.playcanvas.com/t/morph-target-object-upload-as-an-fbx-wont-morph/8742/22


slimbuck (commented):
is this handling an empty gltf file, or can the file contain other assets without a single node?

willeastcott (approved):

mvaligursky:
> is this handling an empty gltf file, or can the file contain other assets without a single node?

We already handled empty file before. 
This handles a case where there are two or more scenes, and some don't have any nodes.

PR #3679, opened 2021-11-09
AnimComponent layer mask path fix

ellthompson:
AnimComponent mask paths that come from the editor will need to sometimes contain encoded characters. When loading a mask into an AnimComponent from data, these characters should be decoded.

### DIFF
--- a/src/framework/components/anim/system.js
+++ b/src/framework/components/anim/system.js
@@ -59,9 +59,10 @@ class AnimComponentSystem extends ComponentSystem {
         if (data.masks) {
             Object.keys(data.masks).forEach((key) => {
                 if (component.layers[key]) {
+                    const maskData = data.masks[key].mask;
                     const mask = {};
-                    Object.keys(data.masks[key].mask).forEach((maskKey) => {
-                        mask[decodeURI(maskKey)] = data.masks[key].mask[maskKey];
+                    Object.keys(maskData).forEach((maskKey) => {
+                        mask[decodeURI(maskKey)] = maskData[maskKey];
                     });
                     component.layers[key].assignMask(mask);
                 }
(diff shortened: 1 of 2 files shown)


src/framework/components/anim/system.js
```diff
@@ -59,7 +59,12 @@ class AnimComponentSystem extends ComponentSystem {
         if (data.masks) {
             Object.keys(data.masks).forEach((key) => {
                 if (component.layers[key]) {
-                    component.layers[key].assignMask(data.masks[key].mask);
+                    const mask = data.masks[key].mask;
+                    Object.keys(mask).forEach((maskKey) => {
+                        mask[decodeURI(maskKey)] = mask[maskKey];
+                        delete mask[maskKey];
```
slimbuck, line 65: what if decodeURI(maskKey) === maskKey ?

ellthompson: Good catch. I've fixed this case.

src/framework/components/anim/system.js
```diff
@@ -59,7 +59,11 @@ class AnimComponentSystem extends ComponentSystem {
         if (data.masks) {
             Object.keys(data.masks).forEach((key) => {
                 if (component.layers[key]) {
-                    component.layers[key].assignMask(data.masks[key].mask);
+                    const mask = {};
+                    Object.keys(data.masks[key].mask).forEach((maskKey) => {
```
slimbuck, line 63: it might be worth caching `data.masks[key].mask` in a variable since it's used in a loop below, but since this is initialisation code might not be worth it.

slimbuck (approved):

src/framework/components/anim/system.js
ellthompson: 👍 done

PR #3680, opened 2021-11-09
Re-added entity references init event

### LINKED ISSUE #3678: Button tints do not work if the scene is loaded at runtime from 1.48.0+
Project: https://playcanvas.com/project/849148/overview/inactive-button-on-scene-change

Repro steps:
1. Load Scene 'Start' https://playcanvas.com/editor/scene/1268903 which will load scene 1268887 at runtime
2. Launch
3. Try to hover or click on the button. Hover should tint the button RED, clicking make it GREEN.

For comparison, loading 1268887 directly/first works fine: https://launch.playcanvas.com/1268887?debug=true

I've tracked the issue down this PR: https://github.com/playcanvas/engine/pull/3557

Digging a bit further shows that the imageEntity entity reference in button.js doesn't resolve so that the Entity is always null and therefore doesn't apply the tint. I can't see anything obvious in the PR that would cause this though.

yaustar:
Fixes #3678

In the PR #3557, there was an event fired in the ComponentSystem.postInitialize function that fired `postinitialize` on itself (note the lowercase `i`)

Event being fired: https://github.com/playcanvas/engine/pull/3557/files#diff-db3ec5f0a5ac5e79ff9192096e8673e0a195e05d91f965608ef630065760e86aL42

Event being registered: https://github.com/playcanvas/engine/pull/3557/files#diff-146cc76c99359e108357b24fb2494cffbaa3b187b07bb83c23f12d6b726fd474L180

This PR adds an event that needs to be called after `postInitialize` event on the systems object in the application.

The event name is a bit long but open to suggestions

### DIFF
--- a/src/framework/application.js
+++ b/src/framework/application.js
@@ -1103,6 +1103,7 @@ class Application extends EventHandler {
         this.fire('initialize');
 
         this.systems.fire('postInitialize', this.root);
+        this.systems.fire('initializeEntityReferences', this.root);
         this.fire('postinitialize');
 
         this.tick();
--- a/src/framework/application.js
+++ b/src/framework/application.js
@@ -1103,7 +1103,7 @@ class Application extends EventHandler {
         this.fire('initialize');
 
         this.systems.fire('postInitialize', this.root);
-        this.systems.fire('initializeEntityReferences', this.root);
+        this.systems.fire('postPostInitialize', this.root);
         this.fire('postinitialize');
 
         this.tick();
--- a/src/framework/scene-registry.js
+++ b/src/framework/scene-registry.js
@@ -269,6 +269,7 @@ class SceneRegistry {
                 // initialize components
                 self._app.systems.fire('initialize', entity);
                 self._app.systems.fire('postInitialize', entity);
+                self._app.systems.fire('initializeEntityReferences', entity);
 
                 if (callback) callback(err, entity);
             };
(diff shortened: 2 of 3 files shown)


src/framework/scene-registry.js
```diff
@@ -269,6 +269,7 @@ class SceneRegistry {
                 // initialize components
                 self._app.systems.fire('initialize', entity);
                 self._app.systems.fire('postInitialize', entity);
+                self._app.systems.fire('initializeEntityReferences', entity);
```
slimbuck, line 272: Don't think you need this new event.

Just call `self._app.fire('postinitialize');`here instead of the new one. (Since that's what entity-reference is listening for).

yaustar: We could. It does feel weird firing an app level event every time a user loads a scene at runtime, especially when this wasn't the behaviour before PR #3557

I can't find anything in the engine listening to the app initialize and postinitialize events but feel it's much safer to have a separate event like it did before PR #3557

src/framework/application.js
```diff
@@ -1103,6 +1103,7 @@ class Application extends EventHandler {
         this.fire('initialize');
 
         this.systems.fire('postInitialize', this.root);
+        this.systems.fire('initializeEntityReferences', this.root);
         this.fire('postinitialize');
```
yaustar, line 1107: We should remove this event as the engine nor editor listens for it. Cautious of removing this on a patch release

slimbuck (approved):

PR #3682, opened 2021-11-09
Expose collision point applied impulse.

LeXXik:
Collision point applied impulse is a total impulse applied by Ammo's collision solver during the last internal substep. It basically describes how hard two objects collide by a single scalar value. It can be useful for spawning some special effects, based on how hard one object hits another.

### DIFF
--- a/src/framework/components/rigid-body/system.js
+++ b/src/framework/components/rigid-body/system.js
@@ -54,6 +54,7 @@ class SingleContactResult {
         if (arguments.length === 0) {
             this.a = null;
             this.b = null;
+            this.impulse = 0;
             this.localPointA = new Vec3();
             this.localPointB = new Vec3();
             this.pointA = new Vec3();
@@ -62,6 +63,7 @@ class SingleContactResult {
         } else {
             this.a = a;
             this.b = b;
+            this.impulse = contactPoint.impulse;
             this.localPointA = contactPoint.localPoint;
             this.localPointB = contactPoint.localPointOther;
             this.pointA = contactPoint.point;
@@ -86,14 +88,16 @@ class SingleContactResult {
  * @property {Vec3} point The point on the entity where the contact occurred, in world space.
  * @property {Vec3} pointOther The point on the other entity where the contact occurred, in world space.
  * @property {Vec3} normal The normal vector of the contact on the other entity, in world space.
+ * @property {Vec3} impulse It is the total accumulated impulse applied by the constraint solver during the last substep. Describes how hard two objects collide.
  */
 class ContactPoint {
-    constructor(localPoint = new Vec3(), localPointOther = new Vec3(), point = new Vec3(), pointOther = new Vec3(), normal = new Vec3()) {
+    constructor(localPoint = new Vec3(), localPointOther = new Vec3(), point = new Vec3(), pointOther = new Vec3(), normal = new Vec3(), impulse = 0) {
         this.localPoint = localPoint;
         this.localPointOther = localPointOther;
         this.point = point;
         this.pointOther = pointOther;
         this.normal = normal;
+        this.impulse = impulse;
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/framework/components/rigid-body/system.js
```diff
@@ -86,14 +88,16 @@ class SingleContactResult {
  * @property {Vec3} point The point on the entity where the contact occurred, in world space.
  * @property {Vec3} pointOther The point on the other entity where the contact occurred, in world space.
  * @property {Vec3} normal The normal vector of the contact on the other entity, in world space.
+ * @property {Vec3} impulse It is the total accumulated impulse applied by the constraint solver during the last substep. Describes how hard two objects collide.
```
willeastcott, line 91:
```suggestion
* @property {number} impulse The total accumulated impulse applied by the constraint solver during the last substep. Describes how hard two objects collide.
```

src/framework/components/rigid-body/system.js
```diff
@@ -86,14 +88,16 @@ class SingleContactResult {
  * @property {Vec3} point The point on the entity where the contact occurred, in world space.
  * @property {Vec3} pointOther The point on the other entity where the contact occurred, in world space.
  * @property {Vec3} normal The normal vector of the contact on the other entity, in world space.
+ * @property {number} impulse The total accumulated impulse applied by the constraint solver during the last substep. Describes how hard two objects collide.
```
willeastcott, line 91: This seems like quite an abstract concept. If I'm reading this value, how would I interpret it to do something useful with it. I mean, what kind of range can I expect? What are the units?

LeXXik: Within a single simulation step, Bullet engine accumulates the impulse over the substeps and then writes that total value into contact point of the collision manifold. The value will depend on the object mass and the speed it is moving, so should be tried against actual application to find the suitable values.

Here is an example with a 1 kg ball:
https://playcanvas.com/editor/scene/1269287

A 10 kg ball, for example, would produce the same values 10 times larger.

LeXXik: In one of the projects, I am using this to spawn some dust and sound effects, when a vehicle lands too hard on a surface. I did have to test some values and smash the car a few times to find a suitable value for it. I'm afraid you won't be able to pick a single value that is usable across different objects, unless they share the same physical properties.

willeastcott: OK, fair enough. I mean, it's fine - I was just curious.

src/framework/components/rigid-body/system.js
```diff
@@ -439,6 +443,7 @@ class RigidBodyComponentSystem extends ComponentSystem {
         contact.point.set(positionWorldOnA.x(), positionWorldOnA.y(), positionWorldOnA.z());
         contact.pointOther.set(positionWorldOnB.x(), positionWorldOnB.y(), positionWorldOnB.z());
         contact.normal.set(normalWorldOnB.x(), normalWorldOnB.y(), normalWorldOnB.z());
+        contact.impulse = contactPoint.getAppliedImpulse();
```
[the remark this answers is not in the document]
LeXXik: Not sure exactly when it was added, but it was available in 2008.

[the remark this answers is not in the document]
LeXXik: Ah, hold on. It was in Bullet in 2008. Let me check Ammo.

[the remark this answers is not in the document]
LeXXik: It was added in 2015:
https://github.com/kripken/ammo.js/commit/4950ad921211356010ec1fecef5371ac5d82d04e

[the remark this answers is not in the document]
willeastcott: Nice, there I believe it has always been available. 😄

willeastcott (approved):

willeastcott:
Thanks a lot of this! Nice contribution. 🙏

Issue #3683, opened 2021-11-09
Material assets/resources don't load again after being unloaded since version 1.48.0+


yaustar:
Project: https://playcanvas.com/project/849163/overview/error-when-adding-template

Steps to repro:

1. Launch Scene 'Start'
2. Press Space to load the Untitled scene with the Ambulance
3. Press Space again to load the scene 'Start' with the cube
4. Press 1 to unload assets with the tag 'Maze1'
5. Press Space to load the Untitled scene with the Ambulance
6. Crash with the following callstack

```
playcanvas.dbg.js?version=1.49.1:20557 Uncaught TypeError: Cannot read properties of null (reading 'alphaToCoverage')
    at MeshInstance.updateKey (playcanvas.dbg.js?version=1.49.1:20557)
    at MeshInstance.set (playcanvas.dbg.js?version=1.49.1:20828)
    at new MeshInstance (playcanvas.dbg.js?version=1.49.1:20465)
    at Model.clone (playcanvas.dbg.js?version=1.49.1:28839)
    at ModelComponent._onModelAssetLoad (playcanvas.dbg.js?version=1.49.1:59592)
    at ModelComponent._bindModelAsset (playcanvas.dbg.js?version=1.49.1:59569)
    at ModelComponent.set (playcanvas.dbg.js?version=1.49.1:59713)
    at ModelComponentSystem.initializeComponentData (playcanvas.dbg.js?version=1.49.1:60091)
    at ModelComponentSystem.addComponent (playcanvas.dbg.js?version=1.49.1:43636)
    at SceneParser._openComponentData (playcanvas.dbg.js?version=1.49.1:35086)
```

On engine version 1.47.2, the scene loads fine both times

https://launch.playcanvas.com/1268937?debug=true&use_local_engine=https://code.playcanvas.com/playcanvas-1.47.2.js

Breaking commit since 1.47.2 is https://github.com/playcanvas/engine/pull/3557

yaustar:
Functionality to change the meshInstance material back to default was removed here: https://github.com/playcanvas/engine/pull/3557/files#diff-9ae4b2931c405acfbabe05b21475b6c7f96ddebc38cc36d934a1f10405646abfL438

yaustar:
Super stuck on this one, either the meshInstance or the material needs a reference to the default material in scene.js. But I can't find an elegent way to get a reference to it so considering some sort of app ref counting global.

willeastcott:
[...]

PR #3685, opened 2021-11-10
Change to the ownership of defaultMaterial to allow easier access to it

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/3683

Change:
- defaultMaterial is created and owned DefaultMaterial class, instead of by the Scene. DefaultMaterial stores one defaultMaterial instance per device. This gets us closer to allow multiple Scenes, while still sharing single defaultMaterial, which some behaviours depend on
- this fixes the issue where upon Material getting destroyed, some mesh instances would have null materials. Now they find a defaultMaterial from the device and assign it to the meshInstance.

### DIFF
--- a/src/framework/application.js
+++ b/src/framework/application.js
@@ -23,7 +23,7 @@ import { GraphicsDevice } from '../graphics/graphics-device.js';
 
 import {
     LAYERID_DEPTH, LAYERID_IMMEDIATE, LAYERID_SKYBOX, LAYERID_UI, LAYERID_WORLD,
-    SORTMODE_NONE, SORTMODE_MANUAL
+    SORTMODE_NONE, SORTMODE_MANUAL, SPECULAR_BLINN
 } from '../scene/constants.js';
 import { BatchManager } from '../scene/batching/batch-manager.js';
 import { ForwardRenderer } from '../scene/renderer/forward-renderer.js';
@@ -118,6 +118,7 @@ import {
     getApplication,
     setApplication
 } from './globals.js';
+import { StandardMaterial } from '../scene/materials/standard-material.js';
 
 // Mini-object used to measure progress of loading sets
 class Progress {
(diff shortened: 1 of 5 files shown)


slimbuck:
We don't want graphics device to store a material for the same reason we don't want graphics device to store an application!

How about material itself store the map of device->defaultMaterial instead?

mvaligursky:
> We don't want graphics device to store a material for the same reason we don't want graphics device to store an application!
> 
> How about material itself store the map of device->defaultMaterial instead?

I agree as mention on the issue ticket.

Ok I'll move it to Material - it does not import pretty much anything, so no big advantage to having a separate class just for this.

src/framework/application.js
```diff
@@ -23,7 +23,7 @@ import { GraphicsDevice } from '../graphics/graphics-device.js';
 
 import {
     LAYERID_DEPTH, LAYERID_IMMEDIATE, LAYERID_SKYBOX, LAYERID_UI, LAYERID_WORLD,
-    SORTMODE_NONE, SORTMODE_MANUAL
+    SORTMODE_NONE, SORTMODE_MANUAL, SPECULAR_BLINN
```
yaustar, line 26: Was this meant to be added?

Maksims:
`DefaultMaterial`, is very similar to StandardMaterial semantically. But they are very different things and is not meant to be instantiated.
If it is a namespace, it would be better to be similar to other APIs in the engine that uses namespace convention.

`new DefaultMaterial()` - definitely won't work.

mvaligursky:
`DefaultMaterial` is private class that is not exported to pc namespace .. so the users cannot even try to instantiate it.

But I agree on similar name .. any better suggestions? But the rest of the engine uses name defaultMaterial for it .. so maybe it should just match. Maybe DefaultMaterialCache.

src/scene/scene.js
```diff
@@ -103,9 +106,11 @@ import { StandardMaterial } from './materials/standard-material.js';
  * child to the Application root entity.
  */
 class Scene extends EventHandler {
-    constructor() {
+    constructor(graphicsDevice) {
         super();
 
+        this.device = graphicsDevice || getApplication().graphicsDevice;
```
slimbuck, line 112: Presumably we want to deprecate `scene.defaultMaterial` in favour of using 
`DefaultMaterial.get()` directly? If that's the case, then we probably don't want to store device here (and change the constructor), but instead use `getApplication().graphicsDevice()` directly in the getter and flag the function as deprecated. And we want to change engine code to use `DefaultMaterial.get()`.

slimbuck (approved):

willeastcott (approved):

PR #3693, opened 2021-11-11
Refactor JSDoc comments for ES6 classes

willeastcott:
When the engine's JSDoc comments were originally written, the code was 100% ES5. Now we have migrated to ES6 classes. JSDoc makes recommendations for documenting ES6 classes:

https://jsdoc.app/howto-es2015-classes.html

This PR updates the core, math and shape APIs to this format. JSDoc is smart enough to deduce many tags from the code without having to specify them explicitly in the JSDoc comment blocks. For example: `name`, `function` and `static`. You also do not have to explicitly include the `description` and `classdesc` tags. You mainly just need the type information, which can be used to generate the Typescript definitions and helps VS Code point out code errors.

Each ES6 class now has a JSDoc comment before it which is interpreted as the `classdesc` (class description/overview). We can now more easily expand on these overview sections and write more useful docs.

The only code change introduced here is the use of field declarations:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#field_declarations

For example, see `Color#r` etc.

One final thing: I have limited columns in all updated source to 100. I think this makes the docs (and code) more readable.

I have verified that both the generated API reference manual and the Typescript definitions are still valid/unchanged.

Once this update is agreed upon, we can roll it out to the rest of the sourcebase.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -1,14 +1,15 @@
 /**
+ * Root namespace for the PlayCanvas Engine.
+ *
  * @name pc
  * @namespace
- * @description Root namespace for the PlayCanvas Engine.
  */
 
 /**
+ * Create look up table for types.
+ *
+ * @returns {object} The type lookup table.
  * @private
- * @function
- * @name _typeLookup
- * @description Create look up table for types.
  */
 const _typeLookup = function () {
     const result = { };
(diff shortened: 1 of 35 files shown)


mvaligursky (approved):
I did not check every line, but this looks really good.

willeastcott:
Superseded by #3725.

PR #3696, opened 2021-11-15
Follow camera lerp function is framerate independent

yaustar:
The old code was effectively doing:

x * dt * 60

Where dt is commonly 1/60 which results to x

However, if the frame rate lower than 60, it starts scaling the lerpAmount to go beyond 1 and cause issues.

Fixes https://forum.playcanvas.com/t/camera-shake-problem/22801

### DIFF
--- a/scripts/camera/follow-camera.js
+++ b/scripts/camera/follow-camera.js
@@ -15,7 +15,7 @@ FollowCamera.attributes.add('lerpAmount', {
     type: 'number',
     min: 0,
     max: 1,
-    default: 0.1,
+    default: 0.99,
     title: 'Lerp Amount',
     description: 'The amount to lerp the camera towards its desired position every frame based on a 60fps frame rate. Lerping is frame rate independent though and will be correct for other frame rates.'
 });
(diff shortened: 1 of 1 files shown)


scripts/camera/follow-camera.js
```diff
@@ -15,7 +15,7 @@ FollowCamera.attributes.add('lerpAmount', {
     type: 'number',
     min: 0,
     max: 1,
-    default: 0.1,
+    default: 0.99,
```
willeastcott, line 18: Are you sure this should be 0.99? The attribute description says:

> The amount to lerp the camera towards its desired position every frame

But if the camera lerps 0.99 towards the target after the first frame, you're pretty much already at your destination.

yaustar: It doesn't lerp 99% on the first frame as it's multiplied by a power of dt:

```
this.currentPos.lerp(this.currentPos, this.targetPos, 1 - Math.pow(1 - this.lerpAmount, dt));
```

willeastcott: Cool, but then isn't the attribute description text misleading?

yaustar: That description should change though, that said 🤔

scripts/camera/follow-camera.js
```diff
@@ -15,7 +15,7 @@ FollowCamera.attributes.add('lerpAmount', {
...
     title: 'Lerp Amount',
     description: 'The amount to lerp the camera towards its desired position every frame based on a 60fps frame rate. Lerping is frame rate independent though and will be correct for other frame rates.'
```
yaustar, line 20:
```suggestion
    description: 'The amount to lerp the camera towards its desired position over time. Adjust the value until it feels right for the app. Lerping is frame rate independent though and will be correct for other frame rates.'
```

scripts/camera/follow-camera.js
yaustar: Added a description change suggestion :)

willeastcott: I'm gonna let @slimbuck approve this one - my maths sucks.

slimbuck (approved):

PR #3697, opened 2021-11-15
Layer mask blending fix

ellthompson:
The PR updates the anim layer mask blending to be consistent with that used by the anim evaluator for animation transitions and blend trees.

### DIFF
--- a/src/anim/evaluator/anim-target-value.js
+++ b/src/anim/evaluator/anim-target-value.js
@@ -1,5 +1,3 @@
-import { Quat } from '../../math/quat.js';
-import { Vec3 } from '../../math/vec3.js';
 import { ANIM_LAYER_OVERWRITE } from '../controller/constants.js';
 import { AnimEvaluator } from '../evaluator/anim-evaluator.js';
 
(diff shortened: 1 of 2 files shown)


src/anim/evaluator/anim-target-value.js
```diff
@@ -63,35 +52,15 @@ class AnimTargetValue {
     }
 
     updateValue(index, value) {
+        // always reset the value of the target when the counter is 0
         if (this.counter === 0) {
-            this.value.set(0, 0, 0, 1);
+            this.value = [0, 0, 0, 1];
```
yaustar, line 57: Does it have to be a new array? Could we reuse the same array?

yaustar (approved):
Approved with comments

src/anim/evaluator/anim-target-value.js
mvaligursky: can we not allocate an array here each time

mvaligursky (approved):

mvaligursky:
can you please replace `this.value.fill(0, 0, 3).fill(1, 3);` by 4 assignments

PR #3703, opened 2021-11-15
Updated description on follow camera script

yaustar:
Forgot to commit suggestion on #3696 

### DIFF
--- a/scripts/camera/follow-camera.js
+++ b/scripts/camera/follow-camera.js
@@ -17,7 +17,7 @@ FollowCamera.attributes.add('lerpAmount', {
     max: 1,
     default: 0.99,
     title: 'Lerp Amount',
-    description: 'The amount to lerp the camera towards its desired position over time. The closer it is to 1, the faster the camera will move. Lerping is frame rate independent though and will be correct for other frame rates.'
+    description: 'The amount to lerp the camera towards its desired position over time. The closer it is to 1, the faster the camera will move. Lerping is frame rate independent and will be correct for every frame rate.'
 });
 
 // initialize code called once per entity
--- a/scripts/camera/follow-camera.js
+++ b/scripts/camera/follow-camera.js
@@ -17,7 +17,7 @@ FollowCamera.attributes.add('lerpAmount', {
     max: 1,
     default: 0.99,
     title: 'Lerp Amount',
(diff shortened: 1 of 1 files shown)


scripts/camera/follow-camera.js
```diff
@@ -17,7 +17,7 @@ FollowCamera.attributes.add('lerpAmount', {
     max: 1,
     default: 0.99,
     title: 'Lerp Amount',
-    description: 'The amount to lerp the camera towards its desired position every frame based on a 60fps frame rate. Lerping is frame rate independent though and will be correct for other frame rates.'
+    description: 'The amount to lerp the camera towards its desired position over time. Adjust the value until it feels right for the app. Lerping is frame rate independent though and will be correct for other frame rates.'
```
mvaligursky, line 20: personally, the original text feels more like something you can work with and understand, than the new text?

yaustar: The old text is no longer correct. The lerp amount is ultimately a 'fudge' number that is changed until it 'feels' right.

mvaligursky: yeah I like that .. gives them some useful info

yaustar:
```suggestion
    description: 'The amount to lerp the camera towards its desired position over time. The closer it is to 1, the faster the camera will move. Lerping is frame rate independent though and will be correct for other frame rates.'
```
```suggestion
    description: 'The amount to lerp the camera towards its desired position over time. Adjust the value until it feels right for the app. Lerping is frame rate independent though and will be correct for other frame rates.'
```

yaustar:
Updated the description :)

scripts/camera/follow-camera.js
```diff
@@ -17,7 +17,7 @@ FollowCamera.attributes.add('lerpAmount', {
     max: 1,
     default: 0.99,
     title: 'Lerp Amount',
-    description: 'The amount to lerp the camera towards its desired position every frame based on a 60fps frame rate. Lerping is frame rate independent though and will be correct for other frame rates.'
+    description: 'The amount to lerp the camera towards its desired position over time. The closer it is to 1, the faster the camera will move. Lerping is frame rate independent though and will be correct for other frame rates.'
```
mvaligursky, line 20:
```suggestion
    description: 'The amount to lerp the camera towards its desired position over time. The closer it is to 1, the faster the camera will move. Lerping is frame rate independent and will be correct for every frame rate.'
```

mvaligursky (approved):

PR #3704, opened 2021-11-16
Small tweaks to clusered lighting to improve compatibility and performance

mvaligursky:
- updating cluster fragment code to GL ES 2.0 to compile on all devices (no const arrays)
- defaulting clustered area lights and cookies to false to improve compile and runtime cost across devices
- disabling clustered area lights on webgl1 android devices due to their poor performance / ability to compile and run at all

### DIFF
--- a/examples/src/examples/graphics/clustered-omni-shadows.tsx
+++ b/examples/src/examples/graphics/clustered-omni-shadows.tsx
@@ -47,7 +47,7 @@ class ClusteredShadowsOmniExample extends Example {
         // @ts-ignore engine-tsd
         app.scene.layers.clusteredLightingShadowsEnabled = true;
 
-        // enable clustered cookies (it's enabled by default as well)
+        // enable clustered cookies
         // @ts-ignore engine-tsd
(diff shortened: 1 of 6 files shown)


willeastcott (approved):

Maksims:
Should it be a developer call to change rendering if it underperforms, instead of engine?
On high end Android device, couple of area lights with clustering might work.

I believe engine should be as consistent as possible between platforms.

mvaligursky:
I assume you responded to `disabling clustered area lights on webgl1 android devices`?

On few old android devices I tried, it simply does not run .. shader fails to compile because too many temporary variables, or just takes 30 seconds to compile, or fails to compile and no reason given. This is basically the same as unsupported from my point of view. So just blocking this. On Webgl2 I don't limit anything, so all high end devices should and do work fine.
Note that by disabling area lights there, the lights are still used as punctual, so you don't have dark scene .. but don't have those nice square reflection and softer lighting .. but otherwise it's visually reasonably similar.

Issue #3477, opened 2021-09-13
Text Element in 3D space with outline has rendering artifacts at a distance


yaustar:
Forum: https://forum.playcanvas.com/t/transparency-bug-on-text/21863
Project: https://playcanvas.com/project/830335/overview/3d-ui-text-outline-bug

Text Element must have outlines for this issue to occur:
Related: https://github.com/playcanvas/engine/issues/3087

![image](https://user-images.githubusercontent.com/16639049/133063454-dc4704af-c596-478e-9c7f-610ad9612856.png)

sakidev:
This also happens when the render resolution is pretty low, example with one of my games:

706x444:
![image](https://user-images.githubusercontent.com/49696473/133990653-7aa8e5b5-e4f9-409c-8459-6bcf82d130b1.png)

1239x801:
![image](https://user-images.githubusercontent.com/49696473/133990671-d8ef8c28-6e28-4247-8ec8-2fd109f29b30.png)

Issue #3087, opened 2021-04-13
Text elements in World space aren't transparent from certain angles and distances


yaustar:
https://playcanvas.com/editor/scene/1130143

Zoom out at an angle and the 'background' of each glyph starts to be more visible. Related to https://forum.playcanvas.com/t/text-node-visible-through-meshes/19445/6 ?

![image](https://user-images.githubusercontent.com/16639049/114540766-d67f1680-9c4d-11eb-9f50-93b8b68098b6.png)

yak32:
ok, we use MIPMAP_LINEAR for texture filtering, but don't generate mipmaps for texture atlases

yak32:
using FILTER_LINEAR and using discard in the fragment shader fixes the issue: 
![image](https://user-images.githubusercontent.com/7876371/121219152-4f2fd700-c87b-11eb-967b-efbcb4705271.png)

yaustar:
Another developer with a similar issue: https://forum.playcanvas.com/t/3d-text-blur-problem/22877

![image](https://user-images.githubusercontent.com/16639049/141974808-1a34d39e-3104-4a78-95dc-c5a658dd0b5d.png)

PR #3707, opened 2021-11-16
Update to msdf font rendering shader to lower the smoothing

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/3087
Fixes https://github.com/playcanvas/engine/issues/3477

The lowering of the smoothing has an effect of slightly sharper / more aliased text, but removes the dark blur around the edges in the distance / at sharp angles.

Images after:
Screenshot 2021-11-16 at 13 02 50
Screenshot 2021-11-16 at 13 02 33
Screenshot 2021-11-16 at 13 58 10

Note that this does not improve / change the readability of the text when very small .. this is disadvantage of msdf technology.

### DIFF
--- a/src/graphics/program-lib/chunks/msdf.frag
+++ b/src/graphics/program-lib/chunks/msdf.frag
@@ -34,16 +34,20 @@ vec4 applyMsdf(vec4 color) {
     float sigDist = median(tsample.r, tsample.g, tsample.b);
     float sigDistShdw = median(ssample.r, ssample.g, ssample.b);
 
+    // smoothing limit - smaller value makes for sharper but more aliased text, especially on angles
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

willeastcott:
This looks like a good improvement. Do/can we leverage anisotropic filtering to make SDF text crisper at an oblique angle?

mvaligursky:
> This looks like a good improvement. Do/can we leverage anisotropic filtering to make SDF text crisper at an oblique angle?

Not really .. for anisotropic filtering you need mipmaps which don't work with SDF.

yaustar:
That's a great improvement:
image

image

While not perfect, much better than having solid blocks!

PR #3716, opened 2021-11-18
Calculate compressed mipmap sizes correctly

slimbuck:
While investigating https://forum.playcanvas.com/t/problem-concerning-none-pot-textures/22862/7 I noticed the engine specifies mipmap sizes incorrectly for the deep mipmaps of some odd-sized textures.

With this PR we fix that, however the resulting odd-sized textures are still not renderable. (Presumably because DXT compression doesn't support the odd size anyway, but still investigating).

### DIFF
--- a/src/graphics/graphics-device.js
+++ b/src/graphics/graphics-device.js
@@ -1942,8 +1942,8 @@ class GraphicsDevice extends EventHandler {
                             gl.TEXTURE_2D,
                             mipLevel,
                             texture._glInternalFormat,
-                            Math.max(texture._width * resMult, 1),
-                            Math.max(texture._height * resMult, 1),
+                            Math.max((Math.floor((texture._width * resMult)) + 3) & ~3, 1),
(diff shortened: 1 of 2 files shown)


willeastcott (approved):

willeastcott:
I have no idea if this has any relevance - but I'm dropping it here just in case: https://github.com/BinomialLLC/crunch/issues/40

bunnybones1:
Though I'm using WebGL via three.js, I think we have the same underlying issue with DXT (or mipmapped textures in general):
WebGL seems to not like mipmapped textures unless they have all levels up to 1x1.
Even if you have a nonsquare but power of two resolution like 64 *16, the mips should be:
32 * 8
16 * 4
8 * 2
4 * 1
2 * 1
1  * 1

slimbuck:
Hi headsup @willeastcott and @bunnybones1, thanks for the headsup.

From the textures I've tested basis seems to generate mips down to 1x1, but I'll definitely keep and eye our for this.

PR #3725, opened 2021-11-21
Switch to TypeScript compiler to generate declarations

willeastcott:
This PR switches from [`tsd-jsdoc`](https://github.com/englercj/tsd-jsdoc) to `tsc` to generate TypeScript declarations for the engine. The motivations are:

* The TypeScript compiler has supported generating TypeScript declarations files from JavaScript source [since version 3.7.0](https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html). `tsd-jsdoc`, on the other hand, has not been updated since April 2020 (presumably because the TypeScript compiler now provides this functionality). Switching means we can drop the dependency on `tsd-jsdoc` (and since we already depend on the TypeScript compiler, there are no dependencies to add). We can be confident that the TypeScript compiler, being an official and well maintained tool, will generate good quality declarations.
* The engine's JSDoc comments are still in ES5 format (which are extremely verbose and do not map to the actual code very well). We can't reformat them as [ES6 style comments](https://jsdoc.app/howto-es2015-classes.html) because bugs in tsd-jsdoc mean that incorrect declarations are generated. `tsc` can handle ES6 based JSDoc comments beautifully. An example of the new JSDoc format is as follows:

```
/**
 * A class description. Give a detailed overview of the class here.
 */
class MyClass {
    /**
     * Description of the constructor
     */
    constructor() {}

    /**
     * Description of some function
     *
     * @param {string} msg - Some string
     */
    printMessage(msg) {
        console.log(msg);
    }

    /**
     * Description of some property
     *
     * @type {number}
     */
    get pi() {
        return Math.PI;
    }
}
```

Notice the dramatic reduction in the use of JSDoc tags because the doc comments are now contextual to the code.

One difference is that, instead of generating a single `playcanvas.d.ts` file, `tsc` outputs a mirror of the source folder/module structure. Each source module gets a corresponding `.d.ts` file. `tsc` can handle this just fine though. So instead of outputting `build/playcanvas.d.ts`, this PR currently outputs to `types/index.d.ts` with all other `.d.ts` files under the `types` folder too. I personally find this easier to work with actually, when checking for type correctness etc. Much easier to browse the files in VS Code. The `types` folder is added to the `npm` package on publish.

@kungfooman and @aidinabedi (plus anyone else that cares about getting high quality TypeScript declarations) - what do you think?

@kungfooman I have removed the `callbacks` namespace for the callbacks declared in the engine source. Each callback is now defined in the module that it most pertains to.

Even when I imported the callbacks.js module into JS files that reference one of the definitions, the generated declarations just have:

```
declare namespace callbacks {
    type HandleEvent = () => any;
    type LoadAsset = () => any;
    type FilterAsset = () => any;
    type AssetReady = () => any;
    type ConfigureApp = () => any;
    type PreloadApp = () => any;
    type LoadSceneData = () => any;
    type UnloadSceneData = () => any;
    type LoadHierarchy = () => any;
    type LoadSettings = () => any;
    type LoadScene = () => any;
    type CalculateMatrix = () => any;
    type CalculateSortDistance = () => any;
    type VrCamera = () => any;
    type CreateScript = () => any;
    type CreateScreen = () => any;
    type LockMouse = () => any;
    type HttpResponse = () => any;
    type ResourceHandler = () => any;
    type ResourceLoader = () => any;
    type AddParser = () => any;
    type FindNode = () => any;
    type ForEach = () => any;
    type UpdateShader = () => any;
    type VrDisplay = () => any;
    type VrFrame = () => any;
    type XrError = () => any;
    type XrHitTestStart = () => any;
}
```

With the removed namespace, I get things like:

```
/**
 * Callback used by {@link GraphNodefind } and {@link GraphNodefindOne } to search through a graph
 * node and all of its descendants.
 */
export type findNodeCallback = (node: GraphNode) => boolean;
```

Which seems much better. The only problem is that the generated API reference manual doesn't include the definitions for the new callbacks, but I'm sure we can fix that. Otherwise, the JSDoc output looks more or less equivalent.


kungfooman:
Nice work and I'm glad you are working on improving the type support (its not just for the "user devs", it can also help engine devs).

I had no chance to look at e.g. the `callbacks` issue yet. I just know they are treated special and I looked into that some days ago, so I may just find where it bugs tomorrow, e.g.: 

https://github.com/playcanvas/playcanvas-jsdoc-template/blob/966152c7427a94ae89433001478032ccaac5f1a0/publish.js#L288-L290

I just started to test around a bit, do you have an uncommited `tsconfig.json`? `tsc` requires it or it just bails out, so I just fitted the paths from another `tsconfig.json` for testing:

```ts
{
	"compilerOptions": {
		"target" : "esnext",
		"module" : "esnext",
		"strict" : false,
		"rootDir": "./",
		"outDir" : "./build/declarations/",
		"esModuleInterop": true,
		"allowJs": true,
		"checkJs": true,
		"moduleResolution": "node",
		"emitDeclarationOnly": true,
		"declaration": true
	},
	"include": [
		"./src/**/*"
	]
}
```

(for reproducibility, I would rather have the exact same files)

willeastcott:
Ah, yes, sorry. I have checked in the `tsconfig.json` now!

willeastcott:
@kungfooman I've just committed a pretty big change that ensure that constructors are explicitly documented. This allows `tsc` to figure out the types of constructor parameters. So, as far as I can tell, the generated declarations should now be at least as good as what we had before.

There's still a problem with `standard.js` though. If I 'flatten' the functions in the file to global/module scope, the error goes away. Kinda weird.

yaustar:
Side note: We should definitely test the code editor intellisense as part of the engine release/editor engine update release

willeastcott:
@yaustar Yeah, for sure. This is a huge change and although it should bring some nice improvements, some regressions may occur unfortunately. Let's be vigilant.

aidinabedi:
@willeastcott Awesome work. It would be superb if the type information actually came from the code rather then separately maintained comments.

Last, I worked on the custom jsdoc template (i.e. API reference manual), I remember that the callbacks needed to be handled as a special case (otherwise the link generator would generate broken links for them). Interestingly this was not an issue with the default jsdoc template, so the custom template does something fundamentally out of the ordinary (which should be looked at). Anyway, because I wanted to avoid taking apart the entire template, rewriting it, and creating a huge PR, I opted to fix the symptom (i.e. broken links for callbacks). Unfortunately, since there wasn't any metadata in the generator to differentiate callbacks the only way was to rely on their type-name. So I put them in a namespace and [hackishly identify them using that namespace](https://github.com/playcanvas/playcanvas-jsdoc-template/blob/master/publish.js#L288). Not use if this is related to the issue you're experiencing with them not even showing up. But I would recommend at some point simplifying the template to pure css/html cosmetics rather then custom generation logic, maybe basing it off the default.

willeastcott:
Thanks for that context, @aidinabedi - that's super-helpful. So the JSDoc template we use for the API reference manual was generated from the default JSDoc template many, many years ago. So basically, it hasn't kept track with the development of the JSDoc tool over the years. We need to figure out how to modernize it, basically. Either start again with today's default template and modify it again to match the site appearance, or just fold features of the current default template into the current PlayCanvas one.

Maksims:
I've recently been using JSDoc to generate docs, the same way as used here. And they are very-very different.
Probably rewriting from clean - would produce the best results, and shouldn't be actually a big work. Although there are few custom things that might be a bit challenging to implement using default templates and generators.

kungfooman:
I am comparing the reference manual output and one issue I found is e.g. in `AudioListenerComponent`:

https://developer.playcanvas.com/en/api/pc.AudioListenerComponent.html

The old manual has the constructor arguments from `@param` above the class, but in this PR its just copied into the class, making it the arguments of the next method, when there is no constructor:

image

An ugly quick-fix with redundant code (useless constructor):

```js
    /**
     * Create a new AudioListenerComponent instance.
     *
     * @param {AudioListenerComponentSystem} system - The ComponentSystem that created this Component.
     * @param {Entity} entity - The Entity that this Component is attached to.
     */

    constructor(system, entity) {
        super(system, entity);
    }
```

Ideally the constructor arguments would come from the class that is extended (e.g. `Component`) and the correct types come from type inference.

willeastcott:
Oh, well spotted, @kungfooman!! I have checked in a fix.

kungfooman:
You can fix the declaration emit error by removing all superfluous function keywords: `: function ` in `src/graphics/program-lib/programs/standard.js`.

In short comparison:

image

(just tested `npm run build:es5` and its the same output)

willeastcott:
Oh wow, I didn't know the `: function` was redundant there. Interesting! I'll take another look.

willeastcott:
@kungfooman I managed to fixed the errors relating to `generateKey` and `createShaderDefinition` but just adding `/** @type {Function} */` above them.

I think one of the only problems remaining is that, when I run `npm run test:tsd` locally, it runs without error. But in the GitHub Action, I see:

https://github.com/playcanvas/engine/runs/4301476138?check_suite_focus=true

And I have no idea why it errors there.

kungfooman:
> @kungfooman I managed to fixed the errors relating to `generateKey` and `createShaderDefinition` but just adding `/** @type {Function} */` above them.

Why is it better to add more JSDoc to fix this problem than removing redundant function keywords? In the initial PR explanation you considered the "extreme verbosity" as pitfall of ES5, just to add more redudant JSDoc in this case (imho the less code the better, as long it doesn't turn into code golfing).

In the last commits (like https://github.com/playcanvas/engine/pull/3725/commits/e1b493870f10c124dc32f2a92f1f395b1a70fe01) you added `@private` to the class constructors and it prevents type inference in cases like `src\anim\binder\default-anim-binder.js`:

image

Without `@private`, the type inference works again and the correct types show up in the generated `d.ts` files instead of `any`:

image

Same issue with other constructors in e.g. `anim/controller/*.js`

> I think one of the only problems remaining is that, when I run `npm run test:tsd` locally, it runs without error. But in the GitHub Action, I see:

Thats strange, I get the same errors on Windows running node v15.0.1

willeastcott:
@kungfooman So there's a few reasons:

1. I'm trying to get this PR in a mergeable state without making code changes. Or at least, as few code changes as possible.
2. That `standard.js` code is a horrible mess and the team is currently planning a huge refactor of it. So I'd rather not touch it.
3. Removing the `: function` there would mix object shorthand with regular object notation which, stylistically, is inconsistent with the rest of the engine codebase. And as you may know, I **love** ensuring consistency. 😄 

So I think I'd rather get this PR 'working' for:

1. Building the engine.
2. Building the declarations.
3. Building the docs.

And then, when it's merged, we can refine further.

Thanks for the heads up on the `AnimBinder` stuff. I'll take a look.

willeastcott:
@mvaligursky I haven't given up on that approach. I'm kind of just playing here. I'll have another go at using `@typedef` and see if that could work too.

willeastcott:
And by the way, syntax highlighting in VS Code clearly shows any imports that are unused.

willeastcott:
@mvaligursky To follow up....you asked:

> could these be handled just by a typedef import like I tried here?
> https://github.com/playcanvas/engine/pull/3606/files
> 
> Importing them with lint disabled might make it hard to know which are really needed? Or does that still work somehow.

So if you do replace:

/* eslint-disable no-unused-vars */
import { GraphicsDevice } from './graphics-device.js';
/* eslint-enable no-unused-vars */

With this:

```
/** @typedef {import("./graphics-device.js").GraphicsDevice} GraphicsDevice */
```

...it does indeed work for generating the `.d.ts` files. However, JSDoc then errors as follows:

```
ERROR: Unable to parse a tag's type expression for source file C:\dev\enginetsd\src\graphics\index-buffer.js in line 6 with tag title "typedef" and text "{import("./graphics-device.js").GraphicsDevice} GraphicsDevice": Invalid type expression "import("./graphics-device.js").GraphicsDevice": Expected "!", "=", "?", "[]", "|", or end of input but "(" found.
```

I don't see a way around that unfortunately. I'm open to suggestions.

willeastcott:
It feels like I'm getting *very* close to getting this PR working. Currently:

* Engine builds fine.
* Engine lints fine.
* JSDocs are generating fine.
* One type of error remains when testing the TS declarations:

```
types/framework/components/script/component.d.ts(78,30): error TS2304: Cannot find name 'Class'.
types/framework/components/script/component.d.ts(88,30): error TS2304: Cannot find name 'Class'.
types/framework/components/script/component.d.ts(109,33): error TS2304: Cannot find name 'Class'.
types/framework/components/script/component.d.ts(124,34): error TS2304: Cannot find name 'Class'.
types/framework/components/script/component.d.ts(156,31): error TS2304: Cannot find name 'Class'.
types/script/script-attributes.d.ts(14,29): error TS2304: Cannot find name 'Class'.
types/script/script-attributes.d.ts(15,17): error TS2304: Cannot find name 'Class'.
types/script/script-registry.d.ts(37,17): error TS2304: Cannot find name 'Class'.
types/script/script-registry.d.ts(47,33): error TS2304: Cannot find name 'Class'.
types/script/script-registry.d.ts(57,24): error TS2304: Cannot find name 'Class'.
types/script/script-registry.d.ts(69,30): error TS2304: Cannot find name 'Class'.
types/script/script-registry.d.ts(81,19): error TS2304: Cannot find name 'Class'.
types/script/script.d.ts(32,64): error TS2304: Cannot find name 'Class'.
types/script/script.d.ts(71,40): error TS2304: Cannot find name 'Class'.
```

The engine's JSDoc comments use this feature of tsd-jsdoc:

https://github.com/englercj/tsd-jsdoc#supported-non-standard-features

For example, for `pc.createScript`:

```
 * @returns {Class<ScriptType>} A class type (constructor function) that inherits {@link ScriptType},
 * which the developer is meant to further extend by adding attributes and prototype methods.
```

So while tsd-jsdoc can handle that type, tsc can not. And I'm not sure what to do about that. I'm open to suggestions!

kungfooman:
Its a bit early and I haven't tested this. There is a type alias usually called `ConstructorOf`, but lets call it `Class` for this issue:

```ts
export type Class<C> = { new (...args: any[]): C; };
```

This type could be added to the `d.ts` files, which makes sure that TypeScript can infer the correct type information.

Another idea is to just replace `Class<ScriptType>` with `typeof ScriptType` in all `d.ts` files.

willeastcott:
@kungfooman Thanks for the suggestions. That really helped. I think [I've done what you suggested](https://github.com/playcanvas/engine/pull/3725/commits/bc5d55270d70a9ba136cb1ae20ef541b79e1f451) and now, everything seems to work!

willeastcott:
I feel this is probably now 'good enough' to merge. There are some remaining problems:

* Some classes that procedurally create properties do now list those properties in the type declarations. e.g. StandardMaterial. The `@property` tags don't seem to help here.
* The callback links in the API ref docs are now dead links. We probably need to update the JSDoc template separately.
* Some types can be improved/tightened up.

But I think we can use this as a starting point and improve after merging. Note that I have updated all of the XR related API docs to ES6 format to show what the 'final destination' is here.

kungfooman:
I just tested how it works on my end and I don't have much too add, you did a great job.

I would merge it aswell, instead of waiting and risking it to go like the TypeScript PR. There is nothing that can't be fixed later in additional smaller/more-focused PR's.

willeastcott:
Thanks for testing the PR, @kungfooman - I couldn't have done it without your help! 😄

kungfooman:
> * The callback links in the API ref docs are now dead links. We probably need to update the JSDoc template separately.

I found a couple of other dead links, which used to work. They can be found with this RegEx: `<a href="pc\.html">(?!pc)` (negative lookahead to ignore correct `pc` links, still finds constants tho)

Results e.g.:

image

Not sure if this is only on my end, because I tested to rewrite some templates, but I don't recall messing with links 😅

slimbuck (approved):
Looks good, checked every line.*



* may or may not be true.

PR #3728, opened 2021-11-22
Small improvements to clustered lighting

mvaligursky:
- VS for shadow pass when clustered lights are enabled does not generate unused normal
- There seems to be a precision issue in some members of a struct in GLSL compared to storing it outside of the struct. Fixed one case, where on Pixel 3A XL spot lights shadows were flickering on and off by moving the shadow matrix outside of the struct. Related to https://twitter.com/garrettkjohnson/status/1453168806558265350

### DIFF
--- a/src/graphics/program-lib/programs/standard.js
+++ b/src/graphics/program-lib/programs/standard.js
@@ -588,9 +588,9 @@ const standard = {
         if (options.cubeMap) options.sphereMap = null; // cubeMaps have higher priority
         if (options.dpAtlas) options.prefilteredCubemap = null; // dp has even higher priority
         if (!options.useSpecular) options.specularMap = options.glossMap = null;
-        const needsNormal = lighting || reflections || options.ambientSH || options.prefilteredCubemap || options.heightMap || options.enableGGXSpecular ||
-                            LayerComposition.clusteredLightingEnabled || options.clearCoatNormalMap;
         const shadowPass = options.pass >= SHADER_SHADOW && options.pass <= 17;
+        const needsNormal = lighting || reflections || options.ambientSH || options.prefilteredCubemap || options.heightMap || options.enableGGXSpecular ||
+                            (LayerComposition.clusteredLightingEnabled && !shadowPass) || options.clearCoatNormalMap;
 
         this.options = options;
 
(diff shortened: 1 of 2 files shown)


willeastcott (approved):

src/graphics/program-lib/programs/standard.js
```diff
@@ -588,9 +588,9 @@ const standard = {
         if (options.cubeMap) options.sphereMap = null; // cubeMaps have higher priority
         if (options.dpAtlas) options.prefilteredCubemap = null; // dp has even higher priority
         if (!options.useSpecular) options.specularMap = options.glossMap = null;
-        const needsNormal = lighting || reflections || options.ambientSH || options.prefilteredCubemap || options.heightMap || options.enableGGXSpecular ||
-                            LayerComposition.clusteredLightingEnabled || options.clearCoatNormalMap;
         const shadowPass = options.pass >= SHADER_SHADOW && options.pass <= 17;
+        const needsNormal = lighting || reflections || options.ambientSH || options.prefilteredCubemap || options.heightMap || options.enableGGXSpecular ||
+                            (LayerComposition.clusteredLightingEnabled && !shadowPass) || options.clearCoatNormalMap;
```
slimbuck, line 593: @mvaligursky does `!shadowPass` only apply to clustered? surely this applies everywhere?

mvaligursky: this bit applies only to clustered lighting, and that's what I changed. Have I missed something?

`(LayerComposition.clusteredLightingEnabled && !shadowPass)`

What it says is that the shader needs normal in case its clustered shader (in addition to all existing conditions) but not rendering to shadow map.

slimbuck: Sorry what I mean is: do we ever need normals in a shadow pass? Seems to me, whether rendering clustered or not we probably don't want normals in a shadow pass!

mvaligursky: ah I see what you mean. Normally shadow passes don't have normals, this is gated by 
`let lighting = options.lights.length > 0;`

so the addition of `(LayerComposition.clusteredLightingEnabled && !shadowPass)` is more about adding normals when clustered lighting needs it, rather than removing it for other cases - as when clustered lights are on, the options.lights.length is 0 and so normals would be missing.

PR #3729, opened 2021-11-22
Don't allow clustered area lights if less than 8 texture units

mvaligursky:
- this targets older devices (iPhone XR and also iPhone 11) running iOS < 15, so WebGL1 only - only 8 texture units are allowed.

### DIFF
--- a/src/graphics/graphics-device.js
+++ b/src/graphics/graphics-device.js
@@ -913,9 +913,13 @@ class GraphicsDevice extends EventHandler {
         this.samples = gl.getParameter(gl.SAMPLES);
         this.maxSamples = this.webgl2 ? gl.getParameter(gl.MAX_SAMPLES) : 1;
 
-        // Don't allow area lights on old android devices, they often fail to compile the shader,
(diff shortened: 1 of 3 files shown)


willeastcott (approved):

yaustar:
Should we log a warning in console if someone does try to do this? (debug only)

mvaligursky:
probably not really .. we scale graphics depending on the device features all over the place .. I think that's the purpose of the engine, to make it simply run with similar visuals on all devices

yaustar:
If they can't use clustered area lights, what does it fallback on?

mvaligursky:
it falls back to clustered punctual lights, which looks similar. This is an existing fallback, I just added one more case to trigger it.

Maksims:
If there is a public API (flags, events) to know when it happens?

mvaligursky:
Not at the moment, but could be added if needed. At the moment this is at a discretion of the engine, similarly how PCF5 shadow filtering drops to PCF3 if not supported, or many morph targets drop to 4 if texture sampling is not supported and more.

PR #3731, opened 2021-11-22
Fallback to uncompressed with invalid texture dimensions

slimbuck:
This PR changes the handling of basis compressed textures:
- fall back to uncompressed format when the compressed dimensions are invalid on the target format
- set width and height correctly when uploading strange sizes
- initialise `extTextureHalfFloatLinear` correctly for webgl 1 & 2


yaustar:
Not for this PR specifically, but thinking if there was some way we could have logging or recording of what formats get used with basis to help with debugging and optimisation.

Quite a lot of this is silent under the hood (?) and we could do with something like logRenderActions that the the layer composition has to help with knowing which textures formats are being used be a use of conditions like dimensions or platform.

src/resources/basis-worker.js
```diff
@@ -186,11 +219,8 @@ function BasisWorker() {
             // select output format based on supported formats
             basisFormat = hasAlpha ? alphaMapping[format] : opaqueMapping[format];
 
-            // transcode to uncompressed format if the texture is PVRTC or webgl1, and has invalid dimensions
-            const isPVRTC = (basisFormat === BASIS_FORMAT.cTFPVRTC1_4_RGB || basisFormat === BASIS_FORMAT.cTFPVRTC1_4_RGBA);
-            const notPOT = ((width & (width - 1)) !== 0) || ((height & (height - 1)) !== 0);
-            const notSquare = (width !== height);
-            if ((isPVRTC && (notPOT || notSquare)) || (!options.deviceDetails.webgl2 && notPOT)) {
+            // if image dimensions don't work on target target, fall back to uncompressed
```
willeastcott, line 222: "target target"? I'm guessing you have one target too many? 😄

src/resources/basis-worker.js
```diff
@@ -275,11 +305,8 @@ function BasisWorker() {
...
```
willeastcott, line 308: Double-double-targets!

src/resources/basis-worker.js
```diff
@@ -151,6 +155,35 @@ function BasisWorker() {
...
+            case BASIS_FORMAT.cTFPVRTC1_4_RGBA:
+                return isPOT(width, height) && ((width === height) || webgl2);
+            // astc
+            case BASIS_FORMAT.cTFASTC_4x4:
+                return true;
+            // atc
+            case BASIS_FORMAT.cTFATC_RGB:
+            case BASIS_FORMAT.cTFATC_RGBA_INTERPOLATED_ALPHA:
+                // TODO: remove atc support? looks like it's been removed from the webgl spec, see
+                // https://www.khronos.org/registry/webgl/extensions/rejected/WEBGL_compressed_texture_atc/
```
willeastcott, line 182: This is really interesting - I had no idea that the ATC extension had been rejected. I wouldn't have a problem with removing it from the engine.

willeastcott (approved):

src/resources/basis-worker.js
slimbuck: thx :)

src/resources/basis-worker.js
slimbuck: It means we can remove ATC support from the basis wasm module. Not sure how much size that'll save us, but apparently 'every little helps'.

slimbuck:
> Not for this PR specifically, but thinking if there was some way we could have logging or recording of what formats get used with basis to help with debugging and optimisation.
> 
> Quite a lot of this is silent under the hood (?) and we could do with something like logRenderActions that the the layer composition has to help with knowing which textures formats are being used be a use of conditions like dimensions or platform.

Could do. We must also update the editor PVR message to include the DXT restrictions.

PR #3745, opened 2021-11-26
Implement ggx and lambert importance sampling

slimbuck:
### Summary

This PR:
- adds a new method for generating prefiltered environment lighting data
    - GGX for reflections
    - lambert for ambient
    - both using importance sampling
- rename the helper functions wrapper `Prefilter` -> `EnvLighting` (but keep the API private)

### Background

Up to now our pre-generated environment reflections have had a number of shortcomings:
- use of phong lobe instead of more accurate alternatives (like ggx)
- inaccurate ambient term - we were using the 4x4 reflection cubemap as ambient term
- lack of importance sampling means lots of samples required, but still can have blocky artifacts
- attempting to prefilter environments with high dynamic range resulted in noise and other errors

Screenshot 2021-11-26 at 15 14 53
(generated using 4096 samples)

### New stuff

This PR starts to address some of these issues by adding support for GGX reflections and lambert phong. Both are generated using importance sampling, so higher quality is achieved using fewer samples. (This is especially important when generating the prefiltered data at application startup).

Screenshot 2021-11-26 at 15 23 23
(generated using 1024 samples)

Also notice the bright spots now bleed into each other compared to before. The is a result of performing the convolution on HDR data instead of 8bit RGBM.

### Todo

This PR only adds support for generating the data. A followup PR will integrate the new approach into the lighting pipeline and deprecate the existing approaches.


src/graphics/env-lighting.js
```diff
@@ -0,0 +1,397 @@
...
+                numSamples: options?.numSamples || 2048,
+                distribution: 'lambert'
+            });
+            copyMipmap(result, level);
+            level.destroy();
+        }
+
+        // set filtering while bypassing engine
+        device.setTexture(result, 0);
+        device.gl.texParameteri(result._glTarget, device.gl.TEXTURE_MAG_FILTER, device.gl.LINEAR);
```
mvaligursky, line 360: do these get restored to previous values?

slimbuck: Not sure what you mean. The texture shouldn't be touched by the engine again since we're not invalidating texture flags. It's the reason I set the internal`._mipmaps` member here instead of `.mipmaps`.

We actually don't want the engine to re-update this texture at all. If it did, it would call gl.generateMipmaps() since there is no CPU-side level data. Thereby killing all the mipmaps we just generated!

src/graphics/program-lib/chunks/reproject.frag
```diff
@@ -264,14 +373,47 @@ vec4 prefilter() {
...
+        vec2 uv = rnd(i, NUM_SAMPLES);
         vec3 dir = vecSpace * hemisphereSamplePhong(uv, specularPower());
-        result += DECODE_FUNC(SOURCE_FUNC(dir));
+        result += DECODE_FUNC(SOURCE_FUNC(dir, 0.0));
     }
 
     return ENCODE_FUNC(result / float(NUM_SAMPLES));
 }
 
+vec4 prefilterGGX() {
```
mvaligursky, line 384: please list external references if any

src/graphics/reproject-texture.js
```diff
@@ -78,8 +82,16 @@ function reprojectTexture(source, target, options = {}) {
     const specularPower = options.hasOwnProperty('specularPower') ? options.specularPower : 1;
     const numSamples = options.hasOwnProperty('numSamples') ? options.numSamples : 1024;
```
mvaligursky, line 83: should we default to less samples now?

src/graphics/env-lighting.js
mvaligursky: ah ok, all good .. I assumed it changes some state on the device, but it only changes the texture.

src/graphics/program-lib/chunks/reproject.frag
slimbuck: yeah good point, thanks.

src/graphics/reproject-texture.js
slimbuck: Don't think so. 1024 samples is a good choice for some pathological cases. Also, this numSamples still applies to phong specular, which isn't importance sampled.

src/graphics/env-lighting.js
```diff
@@ -0,0 +1,397 @@
...
+
+// generate mipmaps for the given target texture
+// target is either 2d equirect or cubemap with mipmaps = false
+const generateMipmaps = (target) => {
+    const device = target.device;
+    const gl = device.gl;
+    const numLevels = calcLevels(target.width, target.height);
+
+    let prevLevel = null;
+    for (let i = 1; i < numLevels; ++i) {
```
mvaligursky, line 171: this could be worth splitting into 3 loops .. first to allocate textures, then to render all, and last to delete resources. Otherwise it's possible a texture creation will need to block CPU and wait for previous rendering to finish.

src/graphics/env-lighting.js
```diff
@@ -0,0 +1,397 @@
...
+
+// helper functions to support prefiltering lighting data
+class EnvLighting {
+    /**
+     * @private
+     * @function
+     * @name generateSkyboxCubemap
+     * @description Generate a skybox cubemap in the correct pixel format from the source texture.
+     * @param {Texture} source - The source texture. This is either a 2d texture in equirect format or a cubemap.
+     * @param {number} size - Size of the resulting texture. Specify 0 for automatic sizing.
```
mvaligursky, line 218: why not simply don't pass anything for automatic sizing

src/graphics/env-lighting.js
```diff
@@ -0,0 +1,397 @@
...
+            numSamples: 1024
+        });
+
+        return result;
+    }
+
+    /**
+     * @private
+     * @function
+     * @name generateLightingSource
```
mvaligursky, line 251: perhaps this could do with a better name .. we pass in source (parameter) and it generates lighting source. Those are too similar?

src/graphics/env-lighting.js
```diff
@@ -0,0 +1,397 @@
...
+            type: format === PIXELFORMAT_R8_G8_B8_A8 ? TEXTURETYPE_RGBM : TEXTURETYPE_DEFAULT,
+            addressU: ADDRESS_CLAMP_TO_EDGE,
+            addressV: ADDRESS_CLAMP_TO_EDGE,
+            fixCubemapSeams: fixCubemapSeams,
+            mipmaps: false
+        });
+
+        // copy into top level
+        reprojectTexture(source, result, {
+            numSamples: source.mipmaps ? 1 : 1024
```
mvaligursky, line 275: should 1024 be exposed as a parameter?

src/graphics/env-lighting.js
```diff
@@ -0,0 +1,397 @@
...
+        reprojectTexture(source, result, {
+            numSamples: 1
+        });
+
+        // generate convolved mipmaps
+        const levels = calcLevels(result.width, result.height);
+        for (let i = 1; i < levels; ++i) {
+            const level = createCubemap(device, Math.max(1, result.width >> i), format);
+            reprojectTexture(source, level, {
+                numSamples: options?.numSamples || 1024,
```
mvaligursky, line 310: should we default to less than 1024?

src/graphics/env-lighting.js
```diff
@@ -0,0 +1,397 @@
...
+
+        reprojectTexture(source, result, {
+            numSamples: options?.numSamples || 2048,
+            distribution: 'lambert'
+        });
+
+        const levels = calcLevels(result.width, result.height);
+        for (let i = 1; i < levels; ++i) {
+            const level = createCubemap(device, Math.max(1, result.width >> i), format);
+            reprojectTexture(source, level, {
```
mvaligursky, line 350: even these loops for perf reason should probably be 3 loops .. create textures, reproject, then release all.

src/graphics/program-lib/chunks/reproject.frag
```diff
@@ -135,8 +143,29 @@ vec4 sampleCubemap(vec2 sph) {
     return sampleCubemap(fromSpherical(sph));
 }
 
-vec3 getDirectionEquirect() {
-    return fromSpherical((vec2(vUv0.x, 1.0 - vUv0.y) * 2.0 - 1.0) * vec2(PI, PI * 0.5));
+vec4 sampleEquirect(vec2 sph, float mipLevel) {
+    vec2 uv = sph / vec2(PI * 2.0, PI) + 0.5;
+#if SUPPORTS_TEXLOD == 1
```
mvaligursky, line 148: do you need '== 1' part? should just be #if defined(SUPPORTS_TEXLOD)

src/graphics/env-lighting.js
slimbuck: The parameter is defaulted. I'll update the docs to tag it optional.

src/graphics/program-lib/chunks/reproject.frag
mvaligursky: or similar to this: `#ifdef GL2`

src/graphics/env-lighting.js
slimbuck: This function generates the texture that'll be used for generating the reflection and ambient maps.

The texture must have mipmaps and have a filterable pixel format.

Any idea of a better name?

src/graphics/env-lighting.js
slimbuck: I think not. We're only resizing a texture here, not convolving or filtering.

We _could_ add the option, but I think would probably confuse users.

src/graphics/reproject-texture.js
```diff
@@ -98,11 +110,12 @@ function reprojectTexture(source, target, options = {}) {
         "#define SOURCE_FUNC " + sourceFunc + "\n" +
         "#define TARGET_FUNC " + targetFunc + "\n" +
         "#define NUM_SAMPLES " + numSamples + "\n" +
-        "#define NUM_SAMPLES_SQRT " + Math.round(Math.sqrt(numSamples)).toFixed(1) + "\n\n" +
+        "#define NUM_SAMPLES_SQRT " + Math.round(Math.sqrt(numSamples)).toFixed(1) + "\n" +
```
mvaligursky, line 113: this block should be converted to a string literal template

src/graphics/reproject-texture.js
```diff
@@ -98,11 +110,12 @@ function reprojectTexture(source, target, options = {}) {
...
+        "#define SUPPORTS_TEXLOD " + (device.extTextureLod ? '1' : '0') + '\n\n' +
         shaderChunks.reprojectPS,
         processFunc + decodeFunc + encodeFunc + sourceFunc + targetFunc,
         null,
-        device.webgl2 ? "" : "#extension GL_OES_standard_derivatives: enable\n"
+        device.webgl2 ? "" : "#extension GL_OES_standard_derivatives: enable\n" + (device.extTextureLod ? "#extension GL_EXT_shader_texture_lod: enable\n" : "")
```
mvaligursky, line 118: and this should probably be moved to some function on a device .. that would return the string to append, as it's likely used in other places?

src/graphics/env-lighting.js
slimbuck: Think 1024 is a good default. Users doing dynamically generated stuff might want to go lower and trade quality, but some HDR environments need 1024.

That said, we can definitely test and tweak as we go.

src/graphics/reproject-texture.js
slimbuck: Agree. I thought about it, but was too lazy at the time :)

src/graphics/program-lib/chunks/reproject.frag
slimbuck: I always define SUPPORTS_TEXLOD, but set it to 0 or 1 depending on the device support.

I like defining all the defines (because it acts to document all the options), but it does make the use of it a little less readable.

src/graphics/env-lighting.js
mvaligursky: should it then be based on the resolution of the source?

src/graphics/env-lighting.js
slimbuck: I considered adding a texture cache. Perhaps that with another function to clear the cache?

src/graphics/env-lighting.js
mvaligursky: not sure .. I'm warming up to this name already :)

src/graphics/env-lighting.js
slimbuck: Maybe :) though we're possibly projecting between equirect (or any other projection) and cubemap (and in future visa-versa) making a calculation quite complex.

I've never noticed a need for more samples, but if you feel it's important I can make it an option.

(Though I personally would rather wait till someone finds it necessary before adding).

src/graphics/program-lib/chunks/reproject.frag
mvaligursky: It's not consistent with the rest of the engine .. I'd prefer consistency here.

src/graphics/env-lighting.js
mvaligursky: yep .. or just a local array? Or is it likely we'd get some texture re-use from the cache?

src/graphics/reproject-texture.js
slimbuck: yeah agree, though probably not something I'd want to do in this PR.

src/graphics/env-lighting.js
mvaligursky: If this can be done on load time / every frame, it'd be useful to have an option to speed it up

slimbuck: In dynamic case source should have mipmaps so samples == 1.

mvaligursky: Maybe just make these a private static members of the class .. so if somebody needs to change them, they're accessible?

src/graphics/program-lib/chunks/reproject.frag
slimbuck: done, thanks.

src/graphics/reproject-texture.js
slimbuck: done

Maksims:
This has some visual differences.
Definitely worth having a holy-grail (raytraced alternatives?) reference and compare visually three options: holy-grail, new, old

slimbuck:
> This has some visual differences. Definitely worth having a holy-grail (raytraced, rendered offline in other tools) reference and compare visually three options: holy-grail, new, old.
> 
> The new version looks washed out, and less detailed :( Clearly, something is not right especially on nonblurred versions, where it lost contrast.
> 
> Also, what is the prefiltering time difference between those methods?

Try not reach too many conclusions from a single screenshot @Maksims. These changes are about updating our lighting to be more correct and aligning with industry standard, not simply an exercise in taste.

We've been comparing the lighting updates against three.js, babylon and primarily against the official glTF viewer https://github.com/KhronosGroup/glTF-Sample-Viewer. Some of the code is taken from that implementation.

The 'detail' you prefer is just blotchiness due to undersampling - i.e. it's not a loss of detail. Also note that we're not removing the old lighting path just yet, so you will be able to continue using it if you want.

The 'loss' of contrast in the screenshot is because the new prefilter no longer clamps input lighting to RGBM range before convolving. Again, this is not a loss of detail, just more accurate. (We're also using a wider GGX lobe for reflections instead of phong).

I will post performance comparisons soon.

src/graphics/env-lighting.js
slimbuck: Actually if we want to support per-frame updating of these then we must have the option of passing in the texture & render target instead of having them allocated each frame.

This will require changes to API.

mvaligursky: lets support that I think

slimbuck:
> ideally you'd add half pixel offset to u as well, so + 0.5 / 1024.0

Strange, this was an intermediate piece of code. Pls check latest.

src/math/random.js
```diff
@@ -71,6 +71,16 @@ const random = {
         point.x = Math.cos(theta) * radius;
         point.y = y;
         point.z = Math.sin(theta) * radius;
+    },
+
+    radicalInverse: function (i) {
```
mvaligursky, line 76: please add a description of what it does / returns.

slimbuck:
Along with this work we are also busy re-implementing environment lighting to be faster and simpler.

Closing this PR in favour of one that will include the changes to engine lighting.

PR #3746, opened 2021-11-26
Add SHADOWPASS #define to shader

Maksims:
Similar to https://github.com/playcanvas/engine/pull/2454
This PR adds `#define SHADOWPASS` to the vertex shader, so it can be used for branching logic when writing custom shader chunks. So the developer has the ability to avoid some unwanted shader code in the shadow pass shader variant.

Usage example: Seethrough effect, when you want to offset geometry when it obstructs visibility of a character, but don't want to lose original shadows.

This change has no effect on the existing project.

### DIFF
--- a/src/graphics/program-lib/programs/standard.js
+++ b/src/graphics/program-lib/programs/standard.js
@@ -414,6 +414,17 @@ const standard = {
         }
     },
 
+    _getPassDefineString: function (pass) {
+        if (pass === SHADER_PICK) {
+            return '#define PICK_PASS\n';
+        } else if (pass === SHADER_DEPTH) {
+            return '#define DEPTH_PASS\n';
+        } else if (pass >= SHADER_SHADOW && pass <= 17) {
+            return '#define SHADOW_PASS\n';
+        }
+        return '';
+    },
+
     _vsAddTransformCode: function (code, device, chunks, options) {
         code += chunks.transformVS;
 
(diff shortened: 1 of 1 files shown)


mvaligursky:
It'd be good to make it bit more generic. We have these in the file:

```
        if (options.pass === SHADER_PICK) {
        } else if (options.pass === SHADER_DEPTH) {
        } else if (shadowPass) {
```

so ideally they all declare the appropriate define, for both vertex and fragment shader.

Maksims:
> It'd be good to make it bit more generic. We have these in the file:
> 
> ```
>         if (options.pass === SHADER_PICK) {
>         } else if (options.pass === SHADER_DEPTH) {
>         } else if (shadowPass) {
> ```
> 
> so ideally they all declare the appropriate define, for both vertex and fragment shader.

Good suggestion.
Implemented it.

src/graphics/program-lib/programs/standard.js
```diff
@@ -655,6 +655,13 @@ const standard = {
             chunks = customChunks;
         }
 
+        if (options.pass === SHADER_PICK) {
+            code += '#define PICKPASS\n';
```
slimbuck, line 659: Can we name these with snake case? i.e. PICK_PASS

mvaligursky: I'd also create a function that converts options.pass to a string .. and just call code += getPassDefine(options.pass);

mvaligursky (approved):
Nice, thanks!

Issue #3702, opened 2021-11-15
Loading GLBs at runtime only loads singular mesh piece


RandomPandaStudios:
1. I have a imported morph-target GLB in the PlayCanvas Editor
2. The GLB is comprised of two mesh pieces (`Character` + `Gun`)
3. When I use the PlayCanvas demo for loading glTF/GLB models at runtime, it only loads the `Character` mesh piece.
https://developer.playcanvas.com/en/tutorials/loading-gltf-glbs/

PlayCanvas Project Demo:
https://playcanvas.com/project/851279/overview/assaultcube-model-viewer

Raw Model (It's also in the Editor project):
[AC_PlayerModel_Test.zip](https://github.com/playcanvas/engine/files/7541927/AC_PlayerModel_Test.zip)

## Model (Tested locally) in Microsoft 3D Viewer:
Shows character + gun.
https://cdn.discordapp.com/attachments/408617316415307778/909914319800643594/unknown.png

## Model (Tested online) in BabylonJS Viewer:
Shows character + gun.
https://cdn.discordapp.com/attachments/408617316415307778/909921310166110258/unknown.png

## Model (Tested online) in PlayCanvas Editor:
Shows only character and no gun.
https://cdn.discordapp.com/attachments/408617316415307778/909914365568909332/unknown.png

yaustar:
Are the discord links viewable if people aren't on the server?

RandomPandaStudios:
> Are the discord links viewable if people aren't on the server?

Yes they are visible to those who do not have Discord or are not on the PlayCanvas Discord. The media files are public.

yaustar:
Odd that this model in particular is causing issues. In the GLTF test, we have multiple morph targets animating here successfully: https://cx20.github.io/gltf-test/examples/playcanvas/index.html?category=tutorialModels&model=MorphStressTest&scale=2.0&type=glTF

mvaligursky:
I had a quick look .. we load one node which has two meshes on it, so this matches other viewers.
But only one of them gets rendered .. which suggests the gun is getting most likely frustum culled. This probably means that it's placed incorrectly (outside of the frustum) or something similar.

yaustar:
Martin is bang on the money, after disabling the cull on the meshinstances, the gun appears

![Uploading image.png…]()

PR #3750, opened 2021-11-29
[Fix] Corrected aabb computation for morphed meshes

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/3702

Fixes the aabb estimation for non-skinned meshes with morphing - it was incorrectly estimated, and aabb of mesh instance could be modified the way the mesh would no longer be visible as it would get culled.

This brings it to parity with skinned morphed meshes.

**Behaviour change**: This represents average case, where most morph targets expand the mesh within the same area. It does not represent the stacked worst case scenario where all morphs could be enabled at the same time, as this can result in a very large aabb. In cases like this, the users should specify customAabb for Model/Render component.

### DIFF
--- a/src/scene/morph.js
+++ b/src/scene/morph.js
@@ -260,8 +260,8 @@ class Morph extends RefCountedObject {
             max.max(targetAabb.getMax());
         }
 
-        this.aabb = new BoundingBox(new Vec3(0, 0, 0), new Vec3(0, 0, 0));
-        this.aabb._expand(min, max);
+        this.aabb = new BoundingBox();
+        this.aabb.setMinMax(min, max);
     }
 
     // creates texture. Used to create both source morph target data, as well as render target used to morph these into, positions and normals
--- a/src/scene/morph.js
+++ b/src/scene/morph.js
@@ -250,7 +250,7 @@ class Morph extends RefCountedObject {
 
         // calculate min and max expansion size
         // Note: This represents average case, where most morph targets expand the mesh within the same area. It does not
-        // represents the stacked worst case scenario where all morphs could be enabled at the same time, as this can result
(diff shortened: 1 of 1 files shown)


src/scene/morph.js
```diff
@@ -248,13 +248,20 @@ class Morph extends RefCountedObject {
 
     _calculateAabb() {
 
-        this.aabb = new BoundingBox(new Vec3(0, 0, 0), new Vec3(0, 0, 0));
-
-        // calc bounding box of the relative change this morph can add
+        // calculate min and max expansion size
+        // Note: This represents average case, where most morph targets expand the mesh within the same area. It does not
+        // represents the stacked worst case scenario where all morphs could be enabled at the same time, as this can result
```
yaustar, line 253:
```suggestion
        // represent the stacked worst case scenario where all morphs could be enabled at the same time, as this can result
```

src/scene/morph.js
```diff
@@ -248,13 +248,20 @@ class Morph extends RefCountedObject {
...
         for (let i = 0; i < this._targets.length; i++) {
-            const target = this._targets[i];
-            this.aabb._expand(target.aabb.getMin(), target.aabb.getMax());
+            const targetAabb = this._targets[i].aabb;
+            min.min(targetAabb.getMin());
+            max.max(targetAabb.getMax());
         }
+
+        this.aabb = new BoundingBox(new Vec3(0, 0, 0), new Vec3(0, 0, 0));
+        this.aabb._expand(min, max);
```
slimbuck, line 264: probably just call setMinMax directly?

mvaligursky: good idea!

slimbuck (approved):

Issue #3742, opened 2021-11-26
Remove special handling of dynamic render targets


slimbuck:
We currently optionally flip vectors in X when sampling dynamically generated cubemaps vs loaded cubemaps.

This results in flags around the engine and shader permutations which must take this option into account. It also means that dynamic cubemaps have a mirror orientation compared to loaded cubemaps. Further processing these cubemaps results in mirrored results.

Now that `RenderTarget` supports `flipY` option we can stop doing this and generate dynamic cubemaps to match texture-based cubemaps.

In order to do this we must:
- update https://github.com/playcanvas/engine/blob/dev/scripts/utils/cubemap-renderer.js to:
    - enable flipY option on render targets
    - update orientations to the following:
        - (0, 90, 0), (0, -90, 0), (-90, 0, 180), (90, 0, 180), (0, 180, 0), (0, 0, 0)
- ensure dynamic shadows work correctly (either by also orienting them correctly or ensuring the above changes don't break rendering)

Then we can:
- remove `useRightHandedCubeMap` flags from standard and skybox materials and chunks

mvaligursky:
We probably need to handle this well:
- omni cookies cubemap texture flipping .. as those can be loaded or realtime rendered (lights example uses loaded cookie texture)
- we'd need to update LightCamera.pointLightRotations to match those rotations .. so that shadows and cookies use the same cubemap rotations - as this is used by lighting, where the same math is shared for shadows and cookies. This is especially true for clustered lights (clustered-omni-shadows example). And if this is done, getCubemapFaceCoordinates in clusteredLightUtils need to be adjusted as well to handle flipped dir.x.

mvaligursky:
Closing — this was essentially implemented by #3751 (Remove left handed cubemap sampling, Dec 2021) shortly after filing, and the remaining checklist items have now been verified against the current codebase (2026):

**Done in #3751:**
- `scripts/utils/cubemap-renderer.js` uses exactly the proposed face orientations — `(0, 90, 0), (0, -90, 0), (-90, 0, 180), (90, 0, 180), (0, 180, 0), (0, 0, 0)` — and enables the render target flip (today expressed as `origin: RENDERTARGET_ORIGIN_TOP`, since #9101 replaced `flipY` with the `origin` option).
- The `useRightHandedCubeMap` flags and the associated shader permutations are fully removed from materials and chunks.

**Shadows / cookies (the open verification item from the comment above) — verified as a closed, self-consistent system:**
[...]

PR #3751, opened 2021-11-29
Remove left handed cubemap sampling

slimbuck:
Part fix for #3742.

Remove the shader permutations and options for left-handed cubemap. We can do this thanks to render target flipY option. Now loaded cubemaps and dynamically rendered cubemaps can have the same layout.

Clustered lighting keeps the existing layout for now. But since this is completely isolated code we can change this in future without breaking things.

When rendering into cubemap faces dynamically, the camera must now render faces with the following orientations:
```
        new pc.Quat().setFromEulerAngles(0, 90, 0),
        new pc.Quat().setFromEulerAngles(0, -90, 0),
        new pc.Quat().setFromEulerAngles(-90, 0, 180),
        new pc.Quat().setFromEulerAngles(90, 0, 180),
        new pc.Quat().setFromEulerAngles(0, 180, 0),
        new pc.Quat().setFromEulerAngles(0, 0, 0)
```
and render target must have `flipY` set to true.


mvaligursky (approved):
Looks great to me. Please add more details to the description related to the changed camera rotations - as I've seen people using a copy of those in scripts - we need to make them aware of this change.

PR #3754, opened 2021-11-30
Internal Debug logging functionality

mvaligursky:
- new internal "Debug" class to represent debugging functionality. At the moment it implements two static functions:
	- assert: when condition fails, an error is logged
	- deprecated: logs out deprecated message
- removed DeprecatedLog class - its functionality is now part of Debug class
- Debug class (including calls to its functions) is compiled out in non debug versions of the engine.

- converted few remaining deprecated messages (picker.js and few more) to use it

- added asserts to Shape constructor parameters to log error when frozen object is passed in, as those constructors keep the parameter objects around and modify them, creating harder to find issues.
	relevant PR - see the problem this is catching: https://github.com/playcanvas/engine/pull/2088


Maksims:
Would it be beneficial to have a more descriptive assertion message, so the regular developer will be able to share console error, and it will be apparent from error message where the error happened, instead of a generic one, which requires to click through to Sources panel in order to find exactly where it happened.

Also, in case of two errors from the same function, developers will have to look at two places: where they call the constructor (in case of AABB for example), and then where exactly error happened, so they can understand which exact argument is causing an issue. But more descriptive assertion errors would identify that straight away.

Maksims:
And on slightly related, but different note regarding BoundingBox, would it be better?:

```js
constructor(center = pc.Vec3.ZERO, halfExtents = pc.Vec3.HALF) {
    this.center = new pc.Vec3(center);
    this.halfExtents = new pc.Vec3(halfExtents);
    this._min = new Vec3();
    this._max = new Vec3();
}
```

So there is no problem with passing frozen vectors, also it avoids the dangers of developers passing their vectors: `new pc.BoundingBox(entity.getPosition())`

mvaligursky:
> Would it be beneficial to have a more descriptive assertion message

Makes sense .. I considered it .. but will do it now, thanks.

> And on slightly related, but different note regarding BoundingBox, would it be better?

This is the solution I've done a year ago in the linked PR .. unfortunately it seems people and possibly even the engine depends on current behaviour.

src/scene/procedural.js
```diff
@@ -592,7 +592,7 @@ function _createConeData(baseRadius, peakRadius, height, heightSegments, capSegm
 function createCylinder(device, opts) {
     // #if _DEBUG
     if (opts && opts.hasOwnProperty('baseRadius') && !opts.hasOwnProperty('radius')) {
-        DeprecatedLog.log('DEPRECATED: "baseRadius" in arguments, use "radius" instead');
+        Debug.deprecated('"baseRadius" in arguments, use "radius" instead');
```
willeastcott, line 595: Can you lose the surrounding `// #if _DEBUG` here?

src/shape/bounding-box.js
```diff
@@ -18,6 +19,10 @@ const tmpVecE = new Vec3();
  */
 class BoundingBox {
     constructor(center = new Vec3(), halfExtents = new Vec3(0.5, 0.5, 0.5)) {
+
+        Debug.assert(!Object.isFrozen(center), `The constructor of 'BoundingBox' does not accept a constant (frozen) object as a 'center' parameter`);
+        Debug.assert(!Object.isFrozen(center), `The constructor of 'BoundingBox' does not accept a constant (frozen) object as a 'halfExtents' parameter`);
```
willeastcott, line 24:
```suggestion
        Debug.assert(!Object.isFrozen(halfExtents), `The constructor of 'BoundingBox' does not accept a constant (frozen) object as a 'halfExtents' parameter`);
````

willeastcott (approved):
Approving, but I can't help feeling that we should properly fix the shape constructors to do the 'right thing'. But obviously, not in this PR.

src/scene/procedural.js
mvaligursky: I could, but then the condition (with an empty statement) would make it to the release build.

PR #3761, opened 2021-12-01
Always fire element set:color and set:opacity event when color and opacity is set

### LINKED ISSUE #3760: Default opacity not set on the button if the current tint opacity is the same as the opacity set on the element
Repro: https://playcanvas.com/editor/scene/1285120

Click on the red button that sets the white button opacity to 0.5
Clicking on the white button should set the default opacity to 1, but it doesn't because the time when it is set, the tint opacity is at 1 and the code optimises this out.

See
https://github.com/playcanvas/engine/blob/dev/src/framework/components/element/image-element.js#L908

https://user-images.githubusercontent.com/16639049/144252731-51bf8da2-c562-492a-8acd-72b84117252b.mp4

yaustar:
Fixes #3760

PR to branch "release 1.50.0" as I would like to get this out in the next patch

(Also remember to merge release to dev)

### DIFF
--- a/src/framework/components/element/image-element.js
+++ b/src/framework/components/element/image-element.js
@@ -1139,10 +1139,10 @@ class ImageElement {
             } else {
                 this.sprite = null;
             }
+        }
 
-            if (this._element) {
-                this._element.fire('set:spriteAsset', _id);
-            }
+        if (this._element) {
+            this._element.fire('set:spriteAsset', _id);
         }
     }
 
@@ -1207,9 +1207,9 @@ class ImageElement {
             this._spriteFrame = value;
         }
 
-        if (this._spriteFrame === oldValue) return;
-
-        this._updateSprite();
+        if (this._spriteFrame !== oldValue) {
+            this._updateSprite();
+        }
 
         if (this._element) {
             this._element.fire('set:spriteFrame', value);
--- a/src/framework/components/element/image-element.js
+++ b/src/framework/components/element/image-element.js
@@ -883,18 +883,16 @@ class ImageElement {
         }
         // #endif
 
-        if (this._color.r === r && this._color.g === g && this._color.b === b) {
-            return;
+        if (this._color.r !== r || this._color.g !== g || this._color.b !== b) {
+            this._color.r = r;
+            this._color.g = g;
+            this._color.b = b;
+
+            this._colorUniform[0] = r;
+            this._colorUniform[1] = g;
+            this._colorUniform[2] = b;
+            this._renderable.setParameter('material_emissive', this._colorUniform);
         }
-
-        this._color.r = r;
-        this._color.g = g;
-        this._color.b = b;
-
-        this._colorUniform[0] = r;
-        this._colorUniform[1] = g;
-        this._colorUniform[2] = b;
-        this._renderable.setParameter('material_emissive', this._colorUniform);
 
         if (this._element) {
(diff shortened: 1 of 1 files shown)


yaustar:
set:color and set:opacity are currently only listened for in the button component where it sets the default state color and opacity.

jpauloruschel (commented):
There are 2 other `this._element.fire()` on image-elements: `set:spriteAsset` and `set:spriteFrame`. The logic for the color and opacity will be different than for these other 2 (which are still only firing if something changed).

For the sake of consistency in event triggering, I think its best to also change the logic for those events too.

yaustar:
Good point, I will make the change

slimbuck:
Not sure I understand why there is a need to fire these events when the value doesn't change?

yaustar:
> Not sure I understand why there is a need to fire these events when the value doesn't change?

The button stores the RGBA for the default state of the button. If the events are not fired, then the default state RGBA of the button will not change.

In the example project from the bug report, the expected result is for the white button to have an opacity of 1 when clicked on after the red button is clicked. 

However, as the opacity event is not fired from the element, the button component still has the old default state opacity set at 0.5 when the mouse moves off the button.

yaustar:
@jpauloruschel Updated to fire the sprite events

slimbuck:
I think the issue you're seeing goes a little deeper.

The problem seems to stem from the button element overriding that same color and opacity [here](https://github.com/playcanvas/engine/blob/dev/src/framework/components/button/component.js#L446) and even has a flag to detect it is doing so [here](https://github.com/playcanvas/engine/blob/dev/src/framework/components/button/component.js#L205).

The event doesn't fire when you set opacity to 1 because the button component has already set it to 1.

I don't have any suggestions for a fix, but I don't think always firing these events is the answer.

yaustar:
> The event doesn't fire when you set opacity to 1 because the button component has already set it to 1.

What's the reason for not firing the events? Do we only want to fire the events if the value has changed? The events are not public API and for internal use only in the engine and currently only listened by the button component.

mvaligursky:
Is this related to those shader lines for opacity and color getting not included in the shader, so when we set them later, they don't "work"?

yaustar:
It's also consistent with some other parts of the code base where they are fired regardless if the value has changed or not:

Element set Width: https://github.com/playcanvas/engine/blob/dev/src/framework/components/element/component.js#L842
Texture atlas set texture: https://github.com/playcanvas/engine/blob/dev/src/scene/texture-atlas.js#L102
Texture atlas set frames: https://github.com/playcanvas/engine/blob/dev/src/scene/texture-atlas.js#L111

yaustar:
> Is this related to those shader lines for opacity and color getting not included in the shader, so when we set them later, they don't "work"?

I don't believe so as it's more to do with the button getting notified that the default state (be it a sprite frame, colour or opacity) has changed when its current state is the same as the value being set.

slimbuck:
Perhaps it's not too bad given only imageElement handles that event...

slimbuck (approved):

jpauloruschel (approved):

jpauloruschel:
I don't see a way to improve this without adding some kind of "default value" for Button (rather than relying on the Element's `color` property), which is not something that'd be easy to add. Also as you mentioned, Steven, this is the logic used for many other components and properties.

PR #3763, opened 2021-12-02
Implementation of the final API for clustered lighting, shadowType support

mvaligursky:
Clustered lighting API has been updated to its final form.

**New API** (still private)
```
        app.scene.clusteredLightingEnabled = true;

        app.scene.lighting.cells = new pc.Vec3(30, 2, 30);
        app.scene.lighting.maxLightsPerCell = 20;

        app.scene.lighting.shadowsEnabled = true;
        app.scene.lighting.shadowType = pc.SHADOW_PCF1;
        app.scene.lighting.shadowAtlasResolution = 2048;

        app.scene.lighting.cookiesEnabled = true;
        app.scene.lighting.cookieAtlasResolution = 1500;

        app.scene.lighting.areaLightsEnabled = true;
```
**Old API** (removed)
```
        pc.LayerComposition.clusteredLightingEnabled = true;
        app.scene.layers.clusteredLightingCells = new pc.Vec3(16, 12, 16);
        app.scene.layers.clusteredLightingMaxLights = 80;
        app.scene.layers.clusteredLightingShadowsEnabled = true;
        app.scene.layers.clusteredLightingCookiesEnabled = true;
        app.scene.layers.clusteredLightingAreaLightsEnabled = true;
```

**Other changes**
- The Scene object now accepts GraphicsDevice as a parameter. This constructor has been made private as well, as the engine does not support multiple Scenes at all.
- Added controls to Clustered spot and omni shadow examples

Clustered shadow filtering: support for PCF1, 3 and 5 filtering for clustered shadows (PCF5 is WebGl2 only, falls backs to PCF3 on WebGl1). Details about added PCF1:
- on webgl1 this is a single sample with point filtering (linear filtering would need 4 samples)
- on webgl2 this is a single sample with linear depth filtering (hardware depth filter)

Screenshot 2021-12-06 at 15 06 30
Screenshot 2021-12-06 at 15 06 14


Maksims:
If there is one scene per app, so one clustered lighting per app, would it be better to have such API (less depth): `app.clusteredLighting.enabled` and similar.
Also easier on documentation, can be separately found in own section instead of scene.

mvaligursky:
The longer team goal is to make multi Scene supported. Scene would represent its set up - global lighting, fog, clustered lighting and similar. But due to many shortcuts taken in how this was put together, this is not possible at the moment, and so there is no reason to expose Scene constructor. But I don't want to close this door.

willeastcott:
My knee-jerk reaction is that the property names are quite verbose. So presumably we're heading for a world where it's not really 'clustered lighting' any more but just 'lighting' (because we plan to dump the current lighting system).

So, I'm thinking more along the lines of:

```
        app.scene.lightingCells = new pc.Vec3(30, 2, 30);
        app.scene.maxLights = 20;
        app.scene.shadowsEnabled = true;
        app.scene.cookiesEnabled = true;
        app.scene.areaLightsEnabled = true;
        app.scene.shadowAtlasResolution = 2048;
        app.scene.cookieAtlasResolution = 1500;
```

I've not included `app.scene.clusteredLightingEnabled` above because presumably, that will go away when clustered lighting is the only option?

Also, keep in mind that the Editor codebase does call the `Scene` constructor in at least a couple of places.

mvaligursky:
great suggestions @willeastcott - I've done just that, but additionally renamed `lightingCells` to `lightCells`.

src/scene/scene.js
```diff
@@ -176,6 +183,95 @@ class Scene extends EventHandler {
...
+        this.worldClustersParams.cells.copy(value);
+    }
+
+    get maxLights() {
+        return this.worldClustersParams.maxLights;
+    }
+
+    set maxLights(value) {
+        this.worldClustersParams.maxLights = value;
+    }
```
willeastcott, line 214: I wonder whether `maxLightsPerCell` might be better here, because otherwise, devs will probably assume it's per scene (since it's a property on the `Scene`). Also, is the valid range 0 to 255? If so, maybe clamp it?

mvaligursky: but it's similar to other properties, for example `shadowsEnabled` - this only affects clustered lights (omni and spot), but directional lights, and also non-clustered lights are not affected.

That's why I left the 'clustered' part of the name there. At some point the 'clustered' part won't be needed if we deprecate previous lighting system, but this will still work only for omni and spot (local lights) and not directional.

Maybe we use the `local` prefix? Not that it's very clear name .. it's just what engine uses internally. Or `nonDirectional`. Or maybe something related to the fact they are positioned (not just a direction)?

willeastcott:
This PR introduces `Scene#shadowType` but we already have `LightComponent#shadowType`. How do we manage this API change?

mvaligursky:
> This PR introduces `Scene#shadowType` but we already have `LightComponent#shadowType`. How do we manage this API change?

Clustered lighting ignores shadowType on the LightComponent, and all cluster lights use the same shadow filtering. If at some point we switch to clustered lighting as a main lighting system, only directional lights will have shadowType exposed on the component. Similarly to shadow resolution - this is automatically assigned from the atlas.

We could possibly do the same with light attenuation - this is per light, as it's reasonably cheap .. but personally I would make it global for cluster lights to save all dynamic branches we can.

willeastcott:
Do we prefer `enableXXX` or `xxxEnabled`?

I can find a few precedents for `enableXXX`. For example:

* [`StandardMaterial#enableGGXSpecular`](https://developer.playcanvas.com/api/pc.StandardMaterial.html#enableGGXSpecular)
* [`ElementComponent#enableMarkup`](https://developer.playcanvas.com/api/pc.ElementComponent.html#enableMarkup)
* `Application#enableBundles` (currently private)

src/scene/scene.js
willeastcott: That's true. I'm not too keen on adding `local` to any property names. @slimbuck + @Maksims - do you two have a preference here? Would be cool to get your thoughts.

mvaligursky:
Good question.

I like `enabled` vs 'enable' as that matches the Component.enabled which is commonly used. In a way, its used in a similar way to for example `entity.camera.enabled' or `scripts.bloom.enabled`

We also have lightmapFilterEnabled.

But I don't mind either way.

src/scene/scene.js
mvaligursky: Originally I was thinking about simply exposing an instance of WorldClustersParams on the scene, perhaps as `scene.Lighting` or `scene.clusteredLighting`
https://github.com/playcanvas/engine/blob/ee00eca51c048fa91fad4aeaf1c9f1f5a36e599f/src/scene/lighting/world-clusters-params.js 

so then the properties would be
`scene.lighting.shadowEnabled = ..`

but then decided against it as the rest of the Scene properties are exposed directly .. even thought I'd probably prefer those to be grouped as well under `skybox`, `ambient`, 'lightmap' or similar.

slimbuck: I agree 'maxLights' on the scene would sound misleading.

'maxLightsPerCell' could work, though a name that is less implementation specific would be even better.

src/scene/constants.js
```diff
@@ -287,9 +287,25 @@ export const SHADOW_VSM32 = 3;
...
+ */
+export const SHADOW_PCF1 = 5;
 
 // non-public: number of supported depth shadow modes
-export const SHADOW_COUNT = 5;
+export const SHADOW_COUNT = 6;
+
+// map of engine SHADOW__*** to a string representation
+export const shadowTypeToString = {};
+shadowTypeToString[SHADOW_PCF3] = "PCF3";
```
slimbuck, line 303: why not just define SHADOW_PCF1 = 'PCF1' ?

src/scene/scene.js
slimbuck: We could expose the clustered lighting parameters as `scene.clusteredLightingParams.maxLights` etc and then down the line deprecate that name and expose it as `scene.lightingParams` instead.

src/scene/scene.js
```diff
@@ -147,6 +150,10 @@ class Scene extends EventHandler {
         this._lightmapFilterRange = 10;
         this._lightmapFilterSmoothness = 0.2;
 
+        // clustered lighting
+        this.worldClustersParams = new WorldClustersParams();
```
slimbuck, line 154: Not `clusteredLightingParams`?

src/scene/scene.js
```diff
@@ -176,6 +183,108 @@ class Scene extends EventHandler {
...
+    }
+
+    get areaLightsEnabled() {
+        return this.worldClustersParams.areaLightsEnabled;
+    }
+
+    set areaLightsEnabled(value) {
+
+        // ignore if not supported
+        if (this.device.supportsAreaLights) {
```
slimbuck, line 265: Seems a waste to store device on the scene just for this. Could whoever uses `supportsAreaLights` flag also perform the device check?

src/scene/scene.js
mvaligursky: Ideally we avoid creating an API now we know we need to deprecate - I'd much rather create one we can keep. I think we can simply use scene.lightingParams .. and make it clear in the documentation that these affect omni and spot lights.

src/scene/constants.js
mvaligursky: I assume changing 
`export const SHADOW_PCF3 = 0;`
to
`export const SHADOW_PCF3 = 'PCF1';
would break compatibility - perhaps even saved data in the Editor database? i'm not sure it's worth the risk. Even some engine code does arithmetic math on these, for example:
`        this._isVsm = value >= SHADOW_VSM8 && value <= SHADOW_VSM32;`

src/scene/scene.js
mvaligursky: We have `WorldClusters` and `WorldClustersParams' classes. I usually keep the name of the instance of the classes matching, especially if those are private members.

src/scene/scene.js
mvaligursky: What is the problem with storing the Device there?

I don't like to move the responsibility of this to the user. They want area lights. It's our problem we don't support them on all devices and need to block it on some? Similarly to for example PCF5 - user requests it, we use PCF3 if not supported. And many other examples.

mvaligursky:
the naming agreed to in a discussion:
```
scene.lighting
	.cells
	.maxLightsPerCell
	.shadowsEnabled
	.shadowType = pc.SHADOW_PCF1;
	.shadowAtlasResolution = 2048;

	.cookiesEnabled = true;
	.cookieAtlasResolution = 1500;

	.areaLightsEnabled = true;
```

slimbuck:
> Do we prefer `enableXXX` or `xxxEnabled`?
> 
> I can find a few precedents for `enableXXX`. For example:
> 
> * [`StandardMaterial#enableGGXSpecular`](https://developer.playcanvas.com/api/pc.StandardMaterial.html#enableGGXSpecular)
> * [`ElementComponent#enableMarkup`](https://developer.playcanvas.com/api/pc.ElementComponent.html#enableMarkup)
> * `Application#enableBundles` (currently private)

To me `xxxEnabled` sounds informative, not something I can set. Since this is lighting API for setting config, I'd probably prefer `enableXxx` form. From a quick look around the engine I see use of both forms, but probably more enableXxx at this point.

mvaligursky:
to me, `xxxEnabled` matches with the style of `entity.enabled` and `component.enabled`.
`enableXXX` sounds like a name of a function to me, not a getter & setter.
@willeastcott any thoughts on this?

willeastcott:
OK, @mvaligursky, while I'm still a little torn on this, I'll go with xxxEnabled.

willeastcott (approved):

slimbuck (approved):

PR #3770, opened 2021-12-03
[AnimComponent] First frame bug fix

### LINKED ISSUE #3533: Problem when speed of anim component is 0 at start
### Description
In my own project my character is not visible (probably out of camera view) when the speed of the anim component is set to 0. In the third person tutorial the character is visible but on a different position and in T-pose. I use this project to show the problem, after a second I set the speed back to 1 by script. 

Probably these results are to be expected because a speed of 0 doesn't start the animation at all, but personally I would see the animation start first and then use the speed of 0.

https://launch.playcanvas.com/1242536?debug=true

### Anim speed of 0:
image
image

### Anim speed of 1:
image
image

ellthompson:
The first update of the anim component created a divide by 0 error, which caused the output values of all animations during the initial frame to contain incorrect values.

Fixes #3533 

### DIFF
--- a/src/anim/controller/anim-controller.js
+++ b/src/anim/controller/anim-controller.js
@@ -521,7 +521,7 @@ class AnimController {
         if (this._isTransitioning) {
             this._currTransitionTime += dt;
             if (this._currTransitionTime <= this._totalTransitionTime) {
-                const interpolatedTime = this._totalTransitionTime ? this._currTransitionTime / this._totalTransitionTime : 1;
+                const interpolatedTime = this._totalTransitionTime === 0 ? this._currTransitionTime / this._totalTransitionTime : 1;
                 // while transitioning, set all previous state animations to be weighted by (1.0 - interpolationTime).
                 for (let i = 0; i < this._transitionPreviousStates.length; i++) {
                     state = this._findState(this._transitionPreviousStates[i].name);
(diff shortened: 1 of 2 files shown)


mvaligursky (approved):

src/anim/controller/anim-controller.js
```diff
@@ -521,7 +522,7 @@ class AnimController {
         if (this._isTransitioning) {
             this._currTransitionTime += dt;
             if (this._currTransitionTime <= this._totalTransitionTime) {
-                const interpolatedTime = this._currTransitionTime / this._totalTransitionTime;
+                const interpolatedTime = this._totalTransitionTime ? this._currTransitionTime / this._totalTransitionTime : 1;
```
yaustar, line 524: Should we explicitly check against 0 here?

src/framework/components/anim/component.js
```diff
@@ -344,6 +344,7 @@ class AnimComponent extends Component {
             this._addLayer.bind(this)({ ...layer });
         }
         this.setupAnimationAssets();
+        console.log('here');
```
yaustar: Remove debugging code :)

yaustar (approved):
Approved with comments

yaustar:
Cool, I will need to update this tutorial! https://developer.playcanvas.com/en/tutorials/world-to-ui-screen-space/

PR #3772, opened 2021-12-05
Add linear and angular offset to collision component

### LINKED ISSUE #698: Allow offsetting of collision component primitives
Currently, you have to child a model to a collision component-enabled entity if you want to handle an offset. This is actually quite a pain and non-obvious for new users. We should consider supporting a positional (and rotational?) offset in the collision component.

LeXXik:
Fixes #698

Adds an ability to set local linear and angular offsets for a collision mesh.

https://user-images.githubusercontent.com/5677782/144762092-4faf8195-bfd8-41b5-b41e-e73b41f980d9.mp4


willeastcott:
Great! Thanks for this @LeXXik. Some comments:

1. Why has `package-lock.json` changed in this PR?
2. `_updateDynamic` and `_getEntityTransform` are very sensitive to performance changes. I'm wondering if it would be useful to have a boolean on the collision component `isOffset` which can be used as a `if` statement test in those two functions. This would allow the fast path to run for what I suspect will be the majority of cases. You could update `isOffset` on every set of `linearOffset` and `angularOffset`.

src/framework/components/collision/component.js
```diff
@@ -123,6 +125,18 @@ class CollisionComponent extends Component {
...
+        if (this.data.initialized) {
+            this.system.recreatePhysicalShapes(this);
+        }
+    }
+
+    onSetAngularOffset(name, oldValue, newValue) {
+        if (this.data.initialized) {
+            this.system.recreatePhysicalShapes(this);
+        }
+    }
```
willeastcott, line 138: These could be the same function. 😄  (e.g. rename to `onSetOffset`).

LeXXik: Good point. My original functions had other stuff there. After removing them, they no longer have a difference. Will use a single method.

LeXXik:
1. No clue why it was included into the commit, I thought it would be ignored. Had to check before pushing. Will restore the original version.
2. Good points. Will incorporate.

src/framework/components/collision/component.js
```diff
@@ -123,6 +126,13 @@ class CollisionComponent extends Component {
         }
     }
 
+    onSetOffset(name, oldValue, newValue) {
+        this._isOffset = true;
```
willeastcott, line 130:
```suggestion
        this._isOffset = !this.data._linearOffset.equals(pc.Vec3.ZERO) || !this.data._angularOffset.equals(pc.Quat.IDENTITY);
```

willeastcott: I obviously haven't tested the suggestion above - but something like that handles the case where any offset is removed.

yaustar:
Is there a need to recreate the physics shape? Can the offsets be applied to an existing shape when changed?

src/framework/components/collision/component.js
willeastcott: Hey @LeXXik - what do you think about my comments on this line. If the offsets are reset to zero, _isOffset needs to be set to false.

willeastcott: Probably the component data? I'm not sure. The `CollisionComponent` is one of the few remaining components that still uses the `ComponentData` block.

LeXXik:
Hey, sorry, didn't have a chance to get to back to this. 

@willeastcott  So, in my system I don't use the `isOffset` flag, since the overhead is negligible. Every bit matters though :) I agree with your suggestion.

@yaustar With the current way rigidbody and component system is organized it is not possible. Any sort of transforms, like moving a body is a responsibility of a rigidbody, which defines the body physical properties like mass and friction, as well as its transform in the physics world. Collision component should only be responsible for the physical shape of the body and the intersections check during phases. In my system the offset is handled by the rigidbody component. This also makes it easier to simply move the body, rather than generating a new shape. 

In order for the shape to be moved, collision component would need to instruct rigidbody component to do so. Firstly, I didn't want to make collision component be aware of rigidbody component (even though rigidbody is aware of collision in some places). I could move the offset to the rigidbody, as I did in my system. However the engine has a special method of creating triggers, where the rigidbody component of the trigger is not exposed in the Editor. If I move the offset to the rigidbody component, I would also have to add duplicate field to collision component, so the trigger volume could also be offset.

Nevertheless, I think it is acceptable to generate a new shape here, since it only happens, when the user changes the offset values in the Editor.

src/framework/components/collision/component.js
LeXXik: I think we should check the `newValue` on zero equality, which is a Vec3 for both properties.

LeXXik: I will need to create a test to verify this.

LeXXik:
How does this variant look? I've added a private proeprty `_angularOffset` to the collision component, to store the quaternion. The data block would keep Vec3 containing the eulers, that could be exposed to Editor.

yaustar:
@willeastcott ?

mvaligursky:
This has unit tests failing at the moment.

LeXXik:
Hmm, I would hold off with merging this. Currently following PRs missing:
- Refactor collision component system to ES6 class to match other systems.
- Add new tests to cover the default functionality and the added offsets one.

LeXXik:
This depends on #4144. Once merged, I will update this one.

LeXXik:
Closing in favor #4865

PR #3777, opened 2021-12-07
Engine is using Debug class to log warnings / errors to avoid #ifdef _DEBUG

mvaligursky:
replacing calls inside _DEBUG by static functions of the Debug class

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -118,7 +118,7 @@ const moduleOptions = {
 };
 
 const stripOptions = {
-    functions: ['Debug.assert', 'Debug.deprecated', 'Debug.warn', 'Debug.error']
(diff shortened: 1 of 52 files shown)


willeastcott:
Just wondering. Why have the indirection via a `Debug` API when you can use the Rollup strip plugin to strip console.* functions directly?

willeastcott:
Is it reasonable to imagine a world where a production build PlayCanvas app would never output to the console? Read the first paragraph here:

https://eslint.org/docs/rules/no-console

mvaligursky:
Yep, that's possible.

My hope is that at some point we send all console logs using some logging class, so we can surface them in some Debug log panel inside the Editor.

But perhaps we should have a discussion about how to best approach this.

willeastcott:
Fair enough. Approving for now then.

willeastcott (approved):

PR #3780, opened 2021-12-07
Make the AnimTrack class public

### LINKED ISSUE #3684: AnimComponentLayer#assignAnimation takes AnimTrack argument which is not public API
Check out the function `AnimComponentLayer#assignAnimation`:

https://github.com/playcanvas/engine/blob/dev/src/framework/components/anim/component-layer.js#L87-L113

It take an `AnimTrack` object as a parameter. However, `AnimTrack` is not currently public API. The JSDoc documents the parameter as a generic `object`. You can't click through to see what an `AnimTrack` actually is in the API ref:

https://developer.playcanvas.com/api/pc.AnimComponentLayer.html#assignAnimation

It seems that loading a `Container` resource from a GLB has `asset.resource.animations` set to an array of `AnimTrack`s. So it seems this should be public? Is there a reason it's been kept private?

ellthompson:
Instances of the `AnimTrack` class are made available to users as part of the container asset resource and are used to assign animations in the anim component. It should be made public so users can understand what the `pc.AnimComponentLayer#assignAnimation` animTrack property actually is.

Fixes #3684 


willeastcott:
This PR has been around for a while so some conflicts have arisen in the meantime. Essentially you need to do two things:

1. Adopt the new, succinct JSDoc block format for getters/setters.
2. Remove the newly adopted `@ignore` tag from the classes you would like to expose to the API reference manual (which has replaced the use of the @private tag in #3918).

src/anim/evaluator/anim-data.js
```diff
@@ -1,7 +1,6 @@
 /**
  * Wraps a set of data used in animation.
  *
```
willeastcott, line 3: This line can go too.

src/anim/evaluator/anim-track.js
```diff
@@ -6,7 +6,6 @@ import { AnimEvents } from './anim-events.js';
 /**
  * AnimTrack contains a set of curve data which can be used to animate a set of target nodes.
  *
```
willeastcott, line 8: This line can go too.

src/anim/evaluator/anim-track.js
```diff
@@ -28,26 +27,57 @@ class AnimTrack {
         this._animEvents = animEvents;
     }
 
+    /**
+     * Returns the name of the AnimTrack.
```
willeastcott, line 31: Can you change Returns to Gets? Since it's a getter (property), it probably fits better. Same applies to the other getters in this PR.

src/anim/evaluator/anim-track.js
```diff
@@ -28,26 +27,57 @@ class AnimTrack {
...
+     *
+     * @type {AnimCurve[]}
+     */
     get curves() {
         return this._curves;
     }
 
+
+    /**
+     * Returns the animation events that will fire during the playback of this anim track.
```
willeastcott, line 77: Just for consistency, pick `anim track` or 'AnimTrack' in this module.

src/anim/evaluator/anim-track.js
willeastcott: It could just be `The name of the AnimTrack.` I guess.....but prepending 'Gets' hints that it's read-only.

src/anim/evaluator/anim-data.js
```diff
@@ -18,7 +16,7 @@ class AnimData {
     }
 
     /**
-     * The number of components that make up an element.
+     * Gets the number of components that make up an element.
```
mvaligursky, line 19: AnimData does not have very good explanation at what it contains .. and so I wonder if this should be public at all? Perhaps the properties of the AnimTrack that return AnimData should be private? Or do we expect user to do anything with the data.

src/anim/evaluator/anim-curve.js
```diff
@@ -25,18 +23,38 @@ class AnimCurve {
...
+     */
     get paths() {
         return this._paths;
     }
 
+    /**
+     * Returns the index of the AnimTrack input which contains the key data for this curve.
+     *
+     * @type {number}
+     */
```
mvaligursky, line 39: perhaps input and output property should be private?

src/anim/evaluator/anim-data.js
ellthompson: I've made the AnimCurve and AnimData classes private and hidden the constructor for the AnimTrack class as the goal of this PR was to display the AnimTrack class correctly in the assignAnimation method rather than provide public API for its internals.

willeastcott (approved):

mvaligursky (approved):

PR #3781, opened 2021-12-07
Update the AnimStateGraph class documentation

### LINKED ISSUE #3674: AnimStateGraph constructor is documented as having zero arguments
https://github.com/playcanvas/engine/blob/dev/src/anim/state-graph/anim-state-graph.js#L1-L7

If we intend for this to be public API, we should document the argument that it takes.

ellthompson:
The AnimStateGraph class was documented as having no arguments in it's constructor. This PR hides the constructor from the jsdocs and gives an example of how to instantiate the class.

Fixes #3674 


mvaligursky:
Do we need to expose the constructor at all?
In the engine examples, the state graph is created from json data this way:
`        modelEntity.anim.loadStateGraph(animStateGraphData);`

Do we need to expose this other way as well?
Perhaps we do, just making sure it's something users would use.

ellthompson:
Editor users that load their state graphs in scripts will get an instance of this class as the asset resource. I was thinking of keeping it public for that specific use case. I'd still want to be moving engine only users away from interacting with this class though.

yaustar:
A little off topic, maybe we should flag functions that are used by the Editor but aren't public API to help with things like this?

mvaligursky:
It's ok to keep the class public for the reason as mentioned, but do we need to have the constructor public? That's what I was asking. You could use @hideconstructor on it to make constructor hidden.

ellthompson:
I didn't know about that jsdoc property. I'll add that, remove the @param property and update the @example to be more editor specific.

mvaligursky (approved):
looks good. please update the PR description to match the final changes.

src/anim/state-graph/anim-state-graph.js
```diff
@@ -1,6 +1,10 @@
 /**
- * Creates an AnimStateGraph asset resource from a blob of JSON data that represents an anim state
- * graph.
+ * Creates an AnimStateGraph asset resource from a blob of JSON data that represents an anim state graph. Can be loaded into an anim component using the {@link AnimComponent#loadStateGraph} method
+ *
+ * @hideconstructor
+ * @example
+ * const animStateGraph = asset.resource;
+ * entity.anim.loadStateGraph(animStateGraph);
```
willeastcott, line 7: This should go on the constructor JSDoc block.

PR #3782, opened 2021-12-07
Expand the public Anim Component API

### LINKED ISSUE #3688: [API Enhacement Request] Expose additional information to the AnimComponent and AnimComponentLayer APIs
Hi PlayCanvas team,

Working on a flipbook animation feature for our particular use case we think we've found some limitations in the current `AnimComponent` APIs. The reason for these requests are explained in [this](https://forum.playcanvas.com/t/is-there-an-easy-way-to-get-the-name-of-the-animation-track-asset-associated-with-an-anim-state/22805) post on the PlayCanvas Forum.

It would be useful if the public API would expose methods to retrieve all the layers added to an `AnimComponent` not just the `baseLayer`.
Also on the `AnimComponentLayer` it would be great if we can also get methods to allow retrieval of the associated animation assets and animation tracks for each state without having to rely on internal functionality that might change.
One more thing that I think it might be useful is to have automatic animation events (`START, END`) for each animation with information about the animation (track name, asset name, etc…whatever might be useful).

I hope you consider this request as I believe it may be useful for other developers as well.
Thank you!

ellthompson:
- Makes the `speed`, `layers`, `activate`, `animationAssets` and `playing` properties on the `AnimComponent` class public.
- Makes the `assignMask` function of the `AnimComponentLayer` class private and replaces it with a public `mask` property. This PR retains the `assignMask` function so that it doesn't contain a breaking change. Updates the layer mask example to use the new API.

Fixes #3688


src/framework/components/anim/component-layer.js
```diff
@@ -87,6 +88,27 @@ class AnimComponentLayer {
         this._mask = mask;
```
mvaligursky, line 88: you should add Debug.deprecated message to this now private function, suggesting people to use the .mask instead.

src/framework/components/anim/component.js
```diff
@@ -109,10 +106,27 @@ class AnimComponent extends Component {
...
         this.loadAnimationAssets();
         // assign the previous layer masks then rebind all anim targets
-        this.layers.forEach((layer, i) => layer.assignMask(prevMasks[i]));
+        this.layers.forEach((layer, i) => {
+            layer.mask = prevMasks[i];
+        });
         this.rebind();
     }
 
+    /**
```
mvaligursky, line 115: I wonder if the public API should be `getAnimationAsset(state)` that would return the asset id of the requested state? This would be more friendly to use, and would hide the implementation detail, allowing us to change it easier if needed.

src/framework/components/anim/component.js
```diff
@@ -130,6 +149,11 @@ class AnimComponent extends Component {
         this._speed = value;
     }
 
+    /**
+     * @name AnimComponent#activate
+     * @type {boolean}
+     * @description If set to true, the anim component will automatically begin playing when all animations have been assigned to its state graph.
```
mvaligursky, line 155: the description sounds more like for a setter, not getter. Should be 'if returns true ...' instead of 'if set to true' ..' ?

src/framework/components/anim/component.js
```diff
@@ -138,6 +162,11 @@ class AnimComponent extends Component {
         this._activate = value;
     }
 
+    /**
+     * @name AnimComponent#playing
+     * @type {boolean}
+     * @description Whether the anim component should be playing.
```
mvaligursky, line 168: again, sounds like a setter.

src/framework/components/anim/component-layer.js
ellthompson: Will do

src/framework/components/anim/component.js
ellthompson: 👍  This new function should live on the AnimComponentLayer then as you need both the layer name and state name to retrieve an asset.

src/framework/components/anim/component.js
ellthompson: This property isn't read only like some of the others.

src/framework/components/anim/component.js
ellthompson: Same as above.

src/framework/components/anim/component.js
mvaligursky: ah I see .. the diff didn't show the setter only getter .. all good!

src/framework/components/anim/component-layer.js
```diff
@@ -221,8 +243,10 @@ class AnimComponentLayer {
      *     // include the hip of the current model but not all of its children
      *     "path/to/hip": true
      * });
+     * @ignore
      */
     assignMask(mask) {
+        Debug.deprecated('The pc.AnimComponentLayer#assignMask function is now deprecated. Assign masks to the pc.AnimComponentLayer#mask property instead.');
```
mvaligursky, line 249: the `set mask` internally calls this, and would print the deprecated warning?

mvaligursky: ah sorry it does not, it calls it on a controller directly

mvaligursky (approved):
seems fine.

src/framework/components/anim/component-layer.js
```diff
@@ -166,6 +166,28 @@ class AnimComponentLayer {
         return this._blendType;
     }
 
+    /**
+     *
```
willeastcott, line 170: Delete this line

src/framework/components/anim/component-layer.js
```diff
@@ -274,6 +298,16 @@ class AnimComponentLayer {
         }
     }
 
+    /**
+     *  Returns the asset that is associated with the given state.
+     *
+     * @param {string} stateName - The name of the state to get the asset for.
+     * @type {pc.Asset}
```
willeastcott, line 305: You don't need the `@type` tag for functions. Just `@param` and `@returns`.

willeastcott: So just to be clear, this should be a `@returns` tag instead. 😄

PR #3783, opened 2021-12-07
Refactor environment lighting

slimbuck:
### Overview

This PR implements a new method of rendering environment lighting.

This update doesn't change the existing lighting workflow for users: environment maps are prefiltered in the editor and stored in a cubemap with increasing roughness encoded in its mipmaps.

### Old method

Up to now we have used the prefiltered cubemaps differently at runtime depending on device capabilities:
- if the device doesn't support textureLod:
    - project the prefiltered cubemaps to dual-paraboloid and render reflections from these
    - generate a spherical harmonic from reflection data (on CPU) and use this for ambient
- otherwise, if num supported texture units < 16 (i.e. iOS/webgl1 devices):
    - render from single cubemap with roughness pyramid using `textureLod`
- otherwise (i.e. most newer android devices):
    - sample and blend 6 individual cubemaps for reflections, use blurry reflections for ambient

Having multiple render paths has a number of downsides:
- maintenance burden and difficult to test all paths
- visuals look different depending on hardware support
- customising lighting chunks requires overriding multiple chunks, error prone
- with the introduction of clustered lighting, some scenarios fail due to limited texture units
- the 6 cubemap version was unneccesarily heavy given the modest improvement in filtering quality

### New method

For these reasons we introduce a rendering approach that works on all devices.

At runtime, we load the pre-generated cubemap as before and then:
- Pack the reflection data onto a single, mipmapless, 2d texture using equirectangular projection
- Generate and store an ambient term in equirectangular projection on the same texture

With this approach we have:
- one rendering path on all devices
- environment reflections and accurate ambient using only a single texture unit
- predicable cost in terms of texture units and performance on all devices
- visuals look the same everywhere

An example environment atlas:

envAtlas

Notice increasing roughness cascade to the left and the additional lambert (ambient) term at center.

### Chunk changes

The following shader chunks have been removed from the engine:
```
ambientPrefilteredCube.frag
ambientPrefilteredCubeLod.frag
prefilterCubemap.frag
reflectionDpAtlas.frag
reflectionPrefilteredCube.frag
reflectionPrefilteredCubeLod.frag
dpAtlasQuad.frag
genParaboloid.frag
```

They have been replaced with the following chunks:
```
ambientEnvAtlas.frag
reflectionEnvAtlas.frag
skyboxEnvAtlas.frag
```

### API changes

The following properties have been deprecated:
```javascript
pc.Scene.skyboxPrefiltered128
pc.Scene.skyboxPrefiltered64
pc.Scene.skyboxPrefiltered32
pc.Scene.skyboxPrefiltered16
pc.Scene.skyboxPrefiltered8
pc.Scene.skyboxPrefiltered4
```

The following API replaces it:
```javascript
// Set of 6 prefiltered cubemaps.
pc.Scene.prefilteredCubemaps

// The environment atlas texture
pc.Scene.envAtlas
```

### Environment prefiltering updates

The method for calculating prefiltered envionment lighting has also been rewritten:
- calculate convolution from HDR data where possible, instead of first clamping to 8-bit rgbm
- use industry standard GGX falloff instead of phong
- use importance sampling with resulting quality improvements
- very fast, probably runtime speed - 1024 samples takes sub-millisecond on macbook
- compatibility has been maintained with the old phong approach, but this will likely be removed


willeastcott:
I think this is one of the most beautiful PRs I've ever seen. 😍 

A few comments:

1. Would `skyboxEnv.frag` be better than `skyboxEnvAtlas.frag`? More consistent with the other two and the `Atlas` part is really an implementation detail, no?
2. So from an end-user's perspective, what's the value of having `pc.Scene#envAltas` as public API?

Also, I just did a Spector.js capture of Seemore from the initial room, looking in the direction of the other rooms.

image

You can see that, for the frame, there are 441 WebGL calls made. For the draw call in view, it sets **10 texture units**! I'd love to know what that command trace looks like with this PR.

willeastcott (approved):

Maksims:
Also wandering how seams are handled and how it looks when box projected compared to previous approach.
And how it looks on a sphere at the top and bottom.
Great simplification! 🙏

slimbuck:
> 1. Would `skyboxEnv.frag` be better than `skyboxEnvAtlas.frag`? More consistent with the other two and the `Atlas` part is really an implementation detail, no?

Yes done, thanks!

> 2. So from an end-user's perspective, what's the value of having `pc.Scene#envAtlas` as public API? How do we expect devs to make use of this in their apps?

The viewer will set this value directly when user drags in environment hdr. In future I hope to add an env asset type which will also set this directly. The atlas itself can be generated using `pc.EnvLighting` which for now is marked as private, but will be made public once its API has settled.

> You can see that, for the frame, there are 441 WebGL calls made. For the draw call shown here in the UI, it sets **10 texture units**! I'd love to know what that command trace looks like with this PR.

Will do some comparisons with Seemore today!

slimbuck:
> Also wandering how seams are handled and how it looks when box projected compared to previous approach. And how it looks on a sphere at the top and bottom. Great simplification! 🙏

It's not obvious from looking at it, but each equirect on the atlas has a single pixel border. The runtime samples from inside the border, so there are no seams. (The seams are done [here](https://github.com/slimbuck/engine/blob/env-lighting-integration/src/graphics/program-lib/chunks/decode.frag#L33) at runtime).

I've been testing this PR with Seemore (which as you know does a bunch of box projection and per-material overrides) as well as the BMW demo and they look identical using the new method.

slimbuck:
> I've been testing this PR with Seemore (which as you know does a bunch of box projection and per-material overrides) as well as the BMW demo and they look identical using the new method.

Sorry, identical is too strong a word. What I mean is, it looks 'nearly identical' to the 1-cubemap method (iOS hardware). The 6-cubemap method has better filtering than envAtlas. I would like to investigate adding the same filtering as 6-cubemap to envAtlas to be used on higher-end hardware.

src/scene/scene.js
```diff
@@ -214,19 +221,21 @@ class Scene extends EventHandler {
     }
 
     set skybox(value) {
-        this._skyboxCubeMap = value;
-        this._resetSkyboxModel();
-        this.updateShaders = true;
+        if (value !== this._skyboxCubeMap) {
+            this._skyboxCubeMap = value;
+            this._resetSkyboxModel();
+        }
```
mvaligursky, line 227: should updateShaders be set to true if for example we replace current cubemap by no cubemap? (value is null)

slimbuck: yes: `_resetSkyboxModel` now sets `this.updateShaders = true;`.

src/graphics/program-lib/chunks/decode.frag
```diff
@@ -0,0 +1,42 @@
...
+
+vec3 decodeRGBE(vec4 raw) {
+    if (raw.a == 0.0) {
+        return vec3(0.0, 0.0, 0.0);
+    } else {
+        return raw.xyz * pow(2.0, raw.w * 255.0 - 128.0);
+    }
+}
+
+float PI = 3.141592653589793;
```
mvaligursky, line 22: make it const please

src/graphics/program-lib/chunks/decode.frag
```diff
@@ -0,0 +1,42 @@
...
+}
+
+vec2 toSphericalUv(vec3 dir) {
+    vec2 uv = toSpherical(dir) / vec2(PI * 2.0, PI) + 0.5;
+    return vec2(uv.x, 1.0 - uv.y);
+}
+
+const float seamSize = 1.0 / 512.0;
+
+vec2 mapUv(vec2 uv, vec4 rect) {
```
mvaligursky, line 35: few comments here would be nice

src/graphics/program-lib/programs/standard.js
```diff
@@ -326,6 +326,14 @@ const standard = {
         return code;
     },
 
+    _decodeFunc: function (textureFormat) {
+        return {
```
mvaligursky, line 330: this could be constructed one time instead of inside each function call. Or perhaps a switch should be used.

src/graphics/reproject-texture.js
```diff
@@ -40,7 +44,322 @@ function getProjectionName(projection) {
         case TEXTUREPROJECTION_EQUIRECT: return "Equirect";
         case TEXTUREPROJECTION_OCTAHEDRAL: return "Octahedral";
     }
+};
+
+// pack a 32bit floating point value into RGBA8
+const packFloat32ToRGBA8 = (value, array, offset) => {
```
mvaligursky, line 50: this should perhaps be moved to FloatPacking. And, please describe what the value range is: [0,1]

src/graphics/reproject-texture.js
```diff
@@ -40,7 +44,322 @@ function getProjectionName(projection) {
...
+// function kept because it's useful for debugging
+// vec3 hemisphereSampleUniform(vec2 uv) {
+//     float phi = uv.y * 2.0 * PI;
+//     float cosTheta = 1.0 - uv.x;
+//     float sinTheta = sqrt(1.0 - cosTheta * cosTheta);
+//     return vec3(cos(phi) * sinTheta, sin(phi) * sinTheta, cosTheta);
+// }
+
+// generate a vector on the hemisphere with phong reflection distribution
+const hemisphereSamplePhong = (dstVec, x, y, specularPower) => {
```
mvaligursky, line 88: I wonder if these should go to random.js?

mvaligursky: also, if there are some references / papers these are based on, it'd be great to list.

src/graphics/texture.js
```diff
@@ -970,6 +968,23 @@ class Texture {
...
+
+        if (this.type === TEXTURETYPE_RGBE) {
+            return 'rgbe';
+        }
+
+        if (this.format === PIXELFORMAT_RGBA16F || this.format === PIXELFORMAT_RGBA32F) {
+            return 'linear';
+        }
+
+        return 'srgb';
```
mvaligursky, line 986: this is not correct in general case I think .. linear vs sRGB should be a separate flag / getter.

src/scene/materials/standard-material.js
```diff
@@ -490,7 +486,7 @@ class StandardMaterial extends Material {
         if (!this.emissiveMap || this.emissiveTint) {
             this._setParameter('material_emissive', getUniform('emissive'));
         }
-        if (this.emissiveMap) {
+        if (this.emissiveIntensity !== 1) {
```
mvaligursky, line 489: does this mean that when the user changes emissiveIntensity from 1 to other value, the shader needs to be recompiled?

src/graphics/texture.js
slimbuck: I know what you mean, but encoding is only ever applied when dealing with textures used for lighting. For example normal maps are never 'decoded'.

src/scene/materials/standard-material.js
slimbuck: yep :( there are a bunch of such micro-optimisations in the shaders.

src/graphics/reproject-texture.js
slimbuck: Yes it should probably go there, but there are questions around naming and the fact it overlaps with `float2Bytes`. For this reason I'd prefer to move it in a different PR.

src/scene/materials/standard-material.js
```diff
@@ -1063,6 +979,46 @@ function _defineMaterialProps() {
     _defineAlias("glossVertexColor", "glossMapVertexColor");
     _defineAlias("opacityVertexColor", "opacityMapVertexColor");
     _defineAlias("lightVertexColor", "lightMapVertexColor");
+
+    // prefiltered cubemaps property
+    const getterFunc = function () {
```
mvaligursky, line 984: give these some less generic name please.

mvaligursky (approved):
Really fantastic effort! This will make things a lot easier and better.
Approving with just few small comments.

src/graphics/texture.js
mvaligursky: maybe it could be renamed to `lightingEncoding()`?

willeastcott: I think I'm more partial to `encoding`, TBH.

src/graphics/program-lib/chunks/decode.frag
slimbuck: done

src/graphics/program-lib/chunks/decode.frag
slimbuck: done, with slight code update.

src/graphics/program-lib/programs/standard.js
slimbuck: done

src/graphics/reproject-texture.js
slimbuck: I would prefer to consider this in followup PR?

src/scene/materials/standard-material.js
slimbuck: these are just function-local variables? don't think they need bigger name.

mvaligursky: sure, those get minified anyways, but the name helps to read the code.

slimbuck: Added a comment to each to make it clear :)

Maksims:
With the new method, blending between cubemaps is not smooth anymore.
Use this example: https://playcanvas.github.io/#/graphics/box-reflection set Shininness to 79 and then move to 80, there is a significant change in shininess.
This bump happens on every step: 79 > 80, 59 > 60, 39 > 40, 19 > 20.

Most likely it is not prefiltered data related, but the way lerping value between two levels is calculated in the shader.

Testable here:

slimbuck:
> With the new method, blending between cubemaps is not smooth anymore. Use this example: https://playcanvas.github.io/#/graphics/box-reflection set Shininness to 79 and then move to 80, there is a significant change in shininess. This bump happens on every step: 79 > 80, 59 > 60, 39 > 40, 19 > 20.
> 
> Most likely it is not prefiltered data related, but the way lerping value between two levels is calculated in the shader.

Oh wow, that looks so wrong. Will investigate, thanks!

PR #3787, opened 2021-12-08
Resolve animation nodes correctly and uniquify sibling node names

slimbuck:
Fixes #https://github.com/playcanvas/playcanvas-viewer/issues/135

We've had a bug in the gltf loader which prepends the name of the first node in the file to the animation path. This results in totally incorrect animation paths when the root node is not first in the list.

We are also unable to resolve paths in the tree when sibling nodes have the same name.

This PR does the following:
- directly build the path from the node, including root name
- uniquify sibling node names
- attempt tree traversal to resolve node names, otherwise fall back to previous methods.

This PR appears to resolve a number of existing problem gltf files.

### DIFF
--- a/src/resources/parser/glb-parser.js
+++ b/src/resources/parser/glb-parser.js
@@ -1356,6 +1356,15 @@ const createAnimation = function (gltfAnimation, animationIndex, gltfAccessors, 
         'weights': 'weights'
     };
 
+    const constructNodePath = (node) => {
+        const path = [];
+        while (node) {
+            path.splice(0, 0, node.name);
+            node = node.parent;
+        }
+        return path;
+    };
+
     // convert anim channels
     for (i = 0; i < gltfAnimation.channels.length; ++i) {
         const channel = gltfAnimation.channels[i];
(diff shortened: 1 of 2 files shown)


slimbuck:
Also appears to fix https://github.com/playcanvas/playcanvas-viewer/issues/133

ellthompson (approved):

src/resources/parser/glb-parser.js
```diff
@@ -1356,14 +1355,23 @@ const createAnimation = function (gltfAnimation, animationIndex, gltfAccessors,
         'weights': 'weights'
     };
 
+    const constructNodePath = (node) => {
+        const path = [];
+        while (node) {
+            path.splice(0, 0, node.name);
```
mvaligursky, line 1361: why not using unshift? That's easier to read

slimbuck: Ahh yes I did a quick search when I wrote this code and missed unshift. Thanks!

mvaligursky (approved):
looks good

PR #3790, opened 2021-12-09
Remove circular dependencies created by handling of deprecated constructors

mvaligursky:
added member functions to GraphicsDevice and Application that allow us to detect if the class is an instance of those. This avoids having to do 'instanceof Application'


mvaligursky:
closing in favour of https://github.com/playcanvas/engine/pull/4736

PR #3800, opened 2021-12-10
Box reflection example

mvaligursky:
dynamic cubemap rendering -> prefiltered lighting atlas -> used on material engine example, to demonstrate feature developed here https://github.com/playcanvas/engine/pull/3783
also requires this to be merged: https://github.com/playcanvas/engine/pull/3818

Screenshot 2021-12-14 at 13 18 12

Screenshot 2021-12-14 at 13 17 50


slimbuck (approved):

LeXXik:
Love this! A quick question, shouldn't there be a reflection in the ceiling of the top lamps?
image

mvaligursky:
there's no gap between those purple beams and the ceiling, so they don't show. But you can see them on the top of the sphere.

LeXXik:
Probably just a screenshot angle then. Gives a feeling that the beam is not part of the ceiling. Will see the setup, when the example is live.

PR #3804, opened 2021-12-10
Refactor of Frustum.containsSphere

mvaligursky:
Refactor of Frustum.containsSphere was returning values 0,1,2 but the result was only tested as if bool. Also, MeshInstance culling that is using it was then returning false, 0, 1, 2 values, which are different types. Element custom culling functions also return boolean.

### DIFF
--- a/src/shape/frustum.js
+++ b/src/shape/frustum.js
@@ -142,17 +142,12 @@ class Frustum {
      * @function
      * @name Frustum#containsSphere
      * @description Tests whether a bounding sphere intersects the frustum. If the sphere is outside the frustum,
-     * zero is returned. If the sphere intersects the frustum, 1 is returned. If the sphere is completely inside
-     * the frustum, 2 is returned. Note that a sphere touching a frustum plane from the outside is considered to
-     * be outside the frustum.
+     * false is returned. Otherwse true is returned. Note that a sphere touching a frustum plane from the outside
+     * is considered to be outside the frustum.
      * @param {BoundingSphere} sphere - The sphere to test.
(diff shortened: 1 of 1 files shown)


Maksims:
Is this a potentially breaking change to a public API?

slimbuck:
Being able to differentiate between in and intersection can help speed up hierarchical traversal (in means all children are also in). Even though this isn't currently used, it could be helpful in future?

mvaligursky:
Yes it changes the API .. but I very much doubt anybody uses the distinction between fully in and partially in. Even the engine does not use it .. I never did in my life.

If we in the future need this, which I doubt, we can always add it in in a consistent way.

Other option is - we keep the public API as is, but create new internal function. But I'm pretty sure this is not necessary.

Maksims:
> Yes it changes the API .. but I very much doubt anybody uses the distinction between fully in and partially in. Even the engine does not use it .. I never did in my life.
> 
> If we in the future need this, which I doubt, we can always add it in in a consistent way.
> 
> Other option is - we keep the public API as is, but create new internal function. But I'm pretty sure this is not necessary.

Well, previously there was always backwards compatibility guaranteed if API is public. Unless it was no other way (like Scripting 2.0, Sound API, etc.)
Some power user could be using that method, and there is no simple way to say: "nobody uses it".

slimbuck:
What advantage do see changing this API @mvaligursky? Not sure I see it right now.

willeastcott:
Seems there is some resistance to merging this, @mvaligursky. Perhaps it's worth closing it for now.

PR #3817, opened 2021-12-14
Fixed circular call for image element 'set:color' and 'set:opac…

Sarletor:
Fixed circular call for image element 'set:color' and 'set:opacity' events

Fixes #3814 

I confirm I have read the [contributing guidelines](https://github.com/playcanvas/engine/blob/master/.github/CONTRIBUTING.md) and signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #3821, opened 2021-12-14
Added same sprite check in Button component to break circular dependency

### LINKED ISSUE #3814: Freezes on UI loading and on any change after update from 1.50.0 to 1.50.1
The project contains a lot of UI elements, when the project is loaded and starts to initialize, everything freezes up to 30 seconds.
After the project is initialized, enabling any UI element causes the same freeze for up to 30 seconds.

The issue is connected with this commit https://github.com/playcanvas/engine/commit/6c3d7dd66c4d5875fdf8538ecdbf0488fc14777c
Building engine before this commit fixes issue.

yaustar:
Fixes #3814

Breaks cyclic call when there are multiple button components referring to the same image component

Between this PR and #3761, the check on whether the same sprite/color is being set on the element component has moved from the element component to the button component. The button component is now responsible for breaking the cyclic call.

Thank you @Sarletor for the fix on the color in PR #3817

### DIFF
--- a/src/framework/components/button/component.js
+++ b/src/framework/components/button/component.js
@@ -449,39 +449,26 @@ class ButtonComponent extends Component {
     }
 
     _applyTintImmediately(tintColor) {
-        if (!tintColor || !this._imageReference.hasComponent('element'))
-            return;
-
-        const color3 = toColor3(tintColor);
-
-        this._isApplyingTint = true;
-
-        if (!color3.equals(this._imageReference.entity.element.color))
-            this._imageReference.entity.element.color = color3;
-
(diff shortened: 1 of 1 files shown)


yaustar:
Test scenes: 

Sprite change: https://playcanvas.com/editor/scene/1294476
Tint change: https://playcanvas.com/editor/scene/1294360

Also tested against the original bug: https://playcanvas.com/editor/scene/1285120

yaustar:
There's a question here whether we should allow multiple button components to use/reference the same image element.

jpauloruschel (approved):

PR #3823, opened 2021-12-15
Fix clearcoat normals shader gen

slimbuck:
This PR fixes an invalid shader generation for [this](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/TextureTransformMultiTest) gltf test model.

The shader code for calculating tangent basis for material "ClearcoatNormalTest1Mat" was accessing an incorrect set of uvs.

### DIFF
--- a/src/graphics/program-lib/programs/standard.js
+++ b/src/graphics/program-lib/programs/standard.js
@@ -1022,16 +1022,16 @@ const standard = {
         }
 
         if (needsNormal) {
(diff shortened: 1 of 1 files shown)


slimbuck:
NOTE: the gltf test model still does not render correctly. However loading it no longer results in shader compilation error.

willeastcott (approved):

PR #3829, opened 2021-12-16
Tweak env shiny reflection filtering

slimbuck:
This PR tweaks shiny environment reflection filtering as follows:
- restrict shiny mip to level 6 as lower level mips introduced artifacts in some cases
- apply -1.0 bias to shiny mipmap calc as default mip level calc is too conservative
- shuffle shader logic to always sample the atlas 3 times
    - this should benefit older hardware which executes all shader branches

### DIFF
--- a/src/graphics/program-lib/chunks/reflectionEnv.frag
+++ b/src/graphics/program-lib/chunks/reflectionEnv.frag
@@ -17,27 +17,35 @@ float shinyMipLevel(vec2 uv) {
     // calculate min of both sets of dF to handle discontinuity at the azim edge
     float maxd = min(max(dot(dx, dx), dot(dy, dy)), max(dot(dx2, dx2), dot(dy2, dy2)));
 
-    return clamp(0.5 * log2(maxd), 0.0, 10.0);
+    return clamp(0.5 * log2(maxd) - 1.0, 0.0, 6.0);
 }
 
 vec3 calcReflection(vec3 tReflDirW, float tGlossiness) {
     vec3 dir = cubeMapProject(tReflDirW) * vec3(-1.0, 1.0, 1.0);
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

src/graphics/program-lib/chunks/reflectionEnv.frag
```diff
@@ -17,27 +17,35 @@ float shinyMipLevel(vec2 uv) {
...
-        vec3 linearB = $DECODE(texture2D(texture_envAtlas, mapMip(uv, ilevel2 + 1.0)));
-        linear0 = mix(linearA, linearB, level2 - ilevel2);
+        uv0 = mapMip(uv, ilevel2);
+        uv1 = mapMip(uv, ilevel2 + 1.0);
+        weight = level2 - ilevel2;
     } else {
-        linear0 = $DECODE(texture2D(texture_envAtlas, mapRoughnessUv(uv, ilevel)));
+        // accessing rough reflection - just sample the same part twice
+        uv0 = uv1 = mapRoughnessUv(uv, ilevel + 1.0);
+        weight = 0.0;
```
mvaligursky, line 43: you could set uv1 to a constant (0,0) perhaps, so that a single pixel is always used instead of loading more data to the cache.

mvaligursky (approved):
nice one!

src/graphics/program-lib/chunks/reflectionEnv.frag
slimbuck: if uv0 === uv1 the card will sample the same pixel twice?

mvaligursky: ah right, what was I thinking .. I somehow believed this still runs through some more math.

PR #3830, opened 2021-12-16
Improvements to internal implementation of debug line / mesh rendering

mvaligursky:
- ownership of the instance of Immediate has been moved from the Application to the Scene - which allows debug rendering functionality to be used internally by the engine on the Scene level, not only on the Framework level.
- this does not change any functionality nor public API
- changed the internal implementation if debug rendering - it no longer adds mesh instances (for both debug lines and debug meshes) to layers, which triggers layer composition update, but directly injects them to a visible lists after the culling. This also allows those to be submitted for rendering from inside the frame, after the culling.
- new private Material.transparent property to speed up layercomposition update (it only needs true / false, instead of a blend mode)

### DIFF
--- a/src/scene/immediate/immediate.js
+++ b/src/scene/immediate/immediate.js
@@ -16,7 +16,7 @@ import { ImmediateBatches } from '../immediate/immediate-batches.js';
 const tempPoints = [];
 
 class Immediate {
-    constructor(device, app) {
+    constructor(device) {
         this.device = device;
         this.quadMesh = null;
         this.textureShader = null;
@@ -39,10 +39,6 @@ class Immediate {
 
         // map of meshes instances added to a layer. The key is layer, the value is an array of mesh instances
         this.layerMeshInstances = new Map();
-
-        // connect to rendering
-        app.on('prerender', this.onPreRender, this);
-        app.on('postrender', this.onPostRender, this);
     }
 
     // creates material for line rendering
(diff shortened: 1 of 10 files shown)


src/scene/scene.js
```diff
@@ -231,6 +236,26 @@ class Scene extends EventHandler {
         this.off();
     }
 
+    // returns the default layer used by the immediate drawing functions
+    getDefaultDrawLayer() {
```
willeastcott, line 240: Any reason why this isn't a getter?

src/scene/immediate/immediate.js
```diff
@@ -234,57 +229,47 @@ class Immediate {
...
         graphNode._dirtyWorld = graphNode._dirtyNormal = false;
 
         return graphNode;
     }
 
-    // called before the frame is rendered, prepares data for rendering
-    onPreRender() {
+    // called before the layer is getting rendered, this updates debug rendering for the layer
+    // This is called just before the layer is rendered to allow lines to be added from inside
+    // the frame getting rendered
```
willeastcott, line 241: This comment probably needs to be tightened up a bit. Says the same thing twice.

willeastcott (approved):
Approving with a couple of minor comments.

src/scene/scene.js
mvaligursky: good idea and now it is, thank you.

src/scene/immediate/immediate.js
mvaligursky: done.

PR #3836, opened 2021-12-17
Make components on Entity class visible to TS Compiler

willeastcott:
At the moment, the TypeScript compiler `tsc` is unable to discern the components that can be available on an `Entity` instance. So the following generates an error in TS:

```
this.entity.camera.nearClip = 1;
```

This PR ensures the component properties are present and initialized to `undefined` so that the TS compiler takes them into account. So the main difference here is that all component properties will exist now, but will be set to `undefined` if not actually added to the entity. This should not cause any functional changes to any projects, as far as I can tell.

### DIFF
--- a/src/framework/entity.js
+++ b/src/framework/entity.js
@@ -6,6 +6,26 @@ import { GraphNode } from '../scene/graph-node.js';
 import { Application } from './application.js';
 
 /** @typedef {import('./components/component.js').Component} Component */
+/** @typedef {import('./components/anim/component.js').AnimComponent} AnimComponent */
+/** @typedef {import('./components/animation/component.js').AnimationComponent} AnimationComponent */
+/** @typedef {import('./components/audio-listener/component.js').AudioListenerComponent} AudioListenerComponent */
+/** @typedef {import('./components/button/component.js').ButtonComponent} ButtonComponent */
+/** @typedef {import('./components/camera/component.js').CameraComponent} CameraComponent */
+/** @typedef {import('./components/collision/component.js').CollisionComponent} CollisionComponent */
+/** @typedef {import('./components/element/component.js').ElementComponent} ElementComponent */
+/** @typedef {import('./components/layout-child/component.js').LayoutChildComponent} LayoutChildComponent */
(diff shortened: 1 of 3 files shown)


src/framework/entity.js
```diff
@@ -20,29 +40,149 @@ import { Application } from './application.js';
...
- * @property {ParticleSystemComponent} [particlesystem] Gets the {@link ParticleSystemComponent} attached to this entity. [read only]
- * @property {RenderComponent} [render] Gets the {@link RenderComponent} attached to this entity. [read only]
- * @property {RigidBodyComponent} [rigidbody] Gets the {@link RigidBodyComponent} attached to this entity. [read only]
- * @property {ScreenComponent} [screen] Gets the {@link ScreenComponent} attached to this entity. [read only]
- * @property {ScriptComponent} [script] Gets the {@link ScriptComponent} attached to this entity. [read only]
- * @property {ScrollbarComponent} [scrollbar] Gets the {@link ScrollbarComponent} attached to this entity. [read only]
- * @property {ScrollViewComponent} [scrollview] Gets the {@link ScrollViewComponent} attached to this entity. [read only]
- * @property {SoundComponent} [sound] Gets the {@link SoundComponent} attached to this entity. [read only]
- * @property {SpriteComponent} [sprite] Gets the {@link SpriteComponent} attached to this entity. [read only]
  * @augments GraphNode
```
mvaligursky, line 43: do we still need augments tag?

mvaligursky (approved):
fantastic solution!

src/framework/entity.js
willeastcott: Last time I checked, yeah, we do - either for JSDoc (for showing all the inherited stuff) or for the TS declarations. I can't remember which. But I will keep investigating if we can dump that somehow. Just not in this PR. 😄

willeastcott:
Thanks @mvaligursky.

I think this shows PR highlights that having typed instances of each component on the `Entity` class was probably a bad idea originally. Maybe this would have been better:

```
this.entity.get('camera').nearClip = 1;
```

Anyway, this PR does the necessary job for now.

PR #3838, opened 2021-12-19
Ensure setter appears before getter

willeastcott:
This PR ensures that all getter/setter pairs first declare the setter and then the getter. Otherwise, the TypeScript compiler `tsc` will output `.d.ts` files with the getter before the docs for the property. This is mainly for readability.

All getter/setter pairs have been moved up to directly follow the constructor.

It also properly documents the instance properties of all math classes, which in turn fixes many TS declaration types across the API.

### DIFF
--- a/extras/mini-stats/mini-stats.js
+++ b/extras/mini-stats/mini-stats.js
@@ -180,32 +180,28 @@ class MiniStats {
         };
     }
 
-    get activeSizeIndex() {
-        return this._activeSizeIndex;
-    }
-
     set activeSizeIndex(value) {
         this._activeSizeIndex = value;
         this.gspacing = this.sizes[value].spacing;
         this.resize(this.sizes[value].width, this.sizes[value].height, this.sizes[value].graphs);
     }
 
-    get opacity() {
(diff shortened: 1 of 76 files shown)


slimbuck:
I'm a little concerned that longer term rules like this will become a pain/burden if we're not automatically enforcing them (setters before getters and always after constructor). And tbh I'm a little uneasy having documentation requirements dictate code structure/layout.

willeastcott:
@slimbuck So we already enable this rule:

https://eslint.org/docs/rules/grouped-accessor-pairs

But I'm going to tighten is up by also enabling this option:

https://eslint.org/docs/rules/grouped-accessor-pairs#setbeforeget

So we can indeed enforce this.

As for placing getters/setters directly after the constructor, that's just a consistency thing I think we should agree on as a team. Now that we have embraced ES6 classes everywhere, our class layout should be part of our coding standards. When you're working across the codebase, it's quite jarring to sometimes find getters/setters and the top and sometimes and the bottom and sometimes dotted around a class.

slimbuck (approved):

PR #3839, opened 2021-12-20
[Examples Browser] playcanvas-observer 1.3.0

ellthompson:
Updates the playcanvas-observer dependency to version 1.3.0. Updates instances of the Observer class to match its type definition.


examples/src/app/example-iframe.tsx
```diff
@@ -213,7 +213,7 @@ const ExampleIframe = (props: ExampleIframeProps) => {
         return false;
     };
 
-    const observer = new Observer({});
+    const observer = new Observer({}, {});
```
willeastcott, line 216: Whoops, sorry but this is a mistake in the Observer JSDoc:

https://github.com/playcanvas/playcanvas-observer/blob/main/src/observer.js#L11

The options object should have square brackets around it to denote that it's optional.

willeastcott: It's arguable that the first parameter should be marked as optional too, so you could be able to do:

```
const observer = new Observer();
```

@vkalpias what was the intention for the original API?

ellthompson: Ahh yeah we should update the playcanvas-observer lib again in that case. I'll wait for @vkalpias answer.

vkalpias: We should be able to make both arguments optional

willeastcott: OK, I've attempted to fix this here: https://github.com/playcanvas/playcanvas-observer/pull/9

ellthompson:
I've updated this with playcanvas-observer v1.3.1

willeastcott (approved):

PR #3848, opened 2021-12-27
Remove Model from ParticleSystemComponent

### LINKED ISSUE #746: The particle system uses the same entity as its graph node
In the particle system component in onEnable we do `this.psys.graph = this.entity;` instead of creating a new pc.GraphNode and assigning that as the graph (and then that new graph node would be a child of the Entity like the model component does). This means that when onRemove is called in ParticleSystemComponentSystem we do `entity.removeChild(data.model.getGraph());` but here `entity === data.model.getGraph()` so essentially this calls `entity.removeChild(entity)` which is kinda wrong.. 

The tricky bit fixing this is making sure that we remove the new pc.GraphNode when we re-create the emitter.

willeastcott:
This PR removes the unnecessary `Model` that holds the emitter's mesh instance in the `ParticleSystemComponent`.

Fixes #746

### DIFF
--- a/src/framework/components/particle-system/component.js
+++ b/src/framework/components/particle-system/component.js
@@ -303,36 +303,36 @@ class ParticleSystemComponent extends Component {
     }
 
     addModelToLayers() {
-        if (!this.data.model) return;
+        if (!this.emitter) return;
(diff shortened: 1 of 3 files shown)


slimbuck (approved):
nice!

PR #3849, opened 2021-12-29
Modernize unit tests

willeastcott:
This PR modernizes the engine's unit tests. It would be great for those interested to contribute opinions/feedback.

Unit tests can now be run in Node.js (`npm run test`) instead of in Chrome (which was driven using Karma). For the moment, the Karma-run tests are still available (`npm run test:karma`) but the intention would be to delete them if we decide this is a superior implementation and sufficient coverage has been achieved. At the moment, this PR adds about 547 tests and maybe provides 90% of the coverage of the Karma-based tests.

Benefits:

* The unit tests are now incredibly easy to debug in VS Code. Just set a breakpoint and run the `npm run test` script in VS Code.
* All unit tests are now ES6 modules (`.mjs` files). Each test module maps to a corresponding module in the engine source tree (so `test/math/vec3.test.mjs` will test `src/math/vec3.js` for example). Each test module imports the minimal set of source modules in order to execute (so a built engine is no longer required which means faster iteration).
* Unlike Karma, there is almost zero configuration for Mocha. It 'just works'.
* Since they're running in Node, we can code tests in ESNext. Note that Mocha advises [against using arrow functions](https://mochajs.org/#arrow-functions) though.
* Performance for running the tests should be far faster when running in Node. Current timings:
  * Node tests: 2 seconds
  * Karma tests: 29 seconds
* We can unit test private API just as easily as public API now.
* Code coverage is enabled (courtesy of [C8](https://www.npmjs.com/package/c8)) via `npm run test:coverage`. Output looks as follows:

image

Using C8, I was easily able to verify that I had reached 100% coverage for several modules (e.g. `Vec3`). It should even be possible to set up a GitHub Action to require tests reach some chosen level of code coverage (or maintain 100% once that milestone is reached).

Other things of note:
* It is not currently possible to add a watch mode to the unit tests due to [this issue in Mocha](https://github.com/mochajs/mocha/issues/4374).
* The tests need a web server running. [`serve`](https://www.npmjs.com/package/serve) seemed to offer an elegant solution for this, so I have replace `http-server` with it.
* The Karma-based tests used [`sinon`](https://sinonjs.org/) to for spies, stubs and mocks. I personally find it really hard to work with and maintain, so I've avoided using it. Some `sinon` based tests have not (yet) been migrated.


Maksims:
Would be great to do "visual" tests also, where some complex scenario would be rendered, and compared to previous version.
If there is enough deviation, then it would warn and require attention.

kungfooman:
> Would be great to do "visual" tests also, where some complex scenario would be rendered, and compared to previous version. If there is enough deviation, then it would warn and require attention.

Do you think about images only or even video comparison for testing if the animation system blends and morphs correctly e.g.? Maybe not entire comparison videos, but more like "snapshots at choosen times" for animations and particle simulations.

Then by nature the particle system has random elements too,  multiple lines like: `this.seed = Math.random();`. So `Math.random()` would also require a deterministic overwrite.

willeastcott:
> Would be great to do "visual" tests also, where some complex scenario would be rendered, and compared to previous version.
If there is enough deviation, then it would warn and require attention.

We have a separate repo (written by @slimbuck) that does visual testing. It's currently private though. Hopefully, we can open source that in the new year. But for now, this PR is just about regular Mocha-based unit testing.

test/framework/entity.test.mjs
```diff
@@ -116,6 +117,51 @@ describe('Entity', function () {
 
     });
 
+    describe('#clone', function () {
+
+        it('clones an entity', function () {
+            const entity = new Entity('Test');
+            Object.keys(components).forEach((name) => {
+                entity.addComponent(name);
+            });
```
kungfooman, line 126: Isn't this just an idiomatic for-in loop?

```suggestion
            for (const name in components) {
                entity.addComponent(name);
            }
```

willeastcott: Haha, very good point @kungfooman. Thanks for pointing this out. I have updated the code based on your suggestion.

willeastcott:
OK, although this PR was never intended to implement all unit tests, I think it's now in a state where it can be merged. So if somebody can review, I'd appreciate it.

slimbuck:
Cool, but can we have some tests to test the tests?

willeastcott:
> Cool, but can we have some tests to test the tests?

You're unit testing my patience, @slimbuck! 😆

slimbuck:
Looks like some of the new tests come from the old set (i18n for example). Is the plan that we slowly migrate engine tests from `engine/tests` to `engine/test`? Shouldn't we remove the old ones in `/tests` as we migrate them to `/test`?

slimbuck:
BTW being able to unit test private API is huge win!

willeastcott:
For the moment, I've left everything in `tests` untouched. I can either:

1. Delete duplicated stuff in `tests` in this PR.
2. Delete duplicated stuff in `tests` in a follow up PR.

What would you prefer?

slimbuck:
I'd probably do it here and set-the-precedent, but not too fussed either way.

slimbuck (approved):
Whoppingly good stuff!

mvaligursky (approved):

PR #3853, opened 2021-12-30
Refactor anim controller conditions code

willeastcott:
Fixes this:

image

This refactor makes the function a bit easier to read too.

### DIFF
--- a/src/anim/controller/anim-controller.js
+++ b/src/anim/controller/anim-controller.js
@@ -249,10 +249,10 @@ class AnimController {
                     if (!(parameter.value <= condition.value)) return false;
                     break;
                 case ANIM_EQUAL_TO:
-                    if (parameter.value !== condition.value) return false;
+                    if (!(parameter.value === condition.value)) return false;
                     break;
                 case ANIM_NOT_EQUAL_TO:
-                    if (parameter.value === condition.value) return false;
+                    if (!(parameter.value !== condition.value)) return false;
                     break;
             }
         }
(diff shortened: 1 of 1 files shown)


src/anim/controller/anim-controller.js
```diff
@@ -231,34 +231,32 @@ class AnimController {
...
-        for (let i = 0; i < transition.conditions.length; i++) {
-            const condition = transition.conditions[i];
+        const conditions = transition.conditions;
+        for (let i = 0; i < conditions.length; i++) {
+            const condition = conditions[i];
             const parameter = this.findParameter(condition.parameterName);
             switch (condition.predicate) {
                 case ANIM_GREATER_THAN:
-                    conditionsMet = conditionsMet && parameter.value > condition.value;
+                    if (parameter.value <= condition.value) return false;
```
slimbuck, line 240: might be worth putting `condition.value` before `parameter.value` so operators match case statements?

willeastcott: I don't think so because the `if` condition must be the negation of the predicate in order to return false (saying the conditions have not been met).

slimbuck: How about (for greater than) ?
```javascript
if (!(parameter.value > condition.value)) return false;
```

Seems like having to mentally negate the conditions to make sense of the code is a little unforgiving!

slimbuck (approved):

src/anim/controller/anim-controller.js
willeastcott: Updated for all except the equals based conditions. What do you think?

willeastcott: I mean, I *could* use logical not on the equals ones too. Anyway, let me know what you prefer.

willeastcott: I've gone ahead and also used logical not outside the equals based conditions. Up to you. 😄

slimbuck: Looks good to me!

PR #3869, opened 2022-01-04
Small updates to documentation text

mvaligursky:


src/core/indexed-list.js
```diff
@@ -1,6 +1,5 @@
 /**
- * A ordered list-type data structure that can provide item look up by key, but also return return
- * a list.
+ * A ordered list-type data structure that can provide item look up by key, but also return a list.
```
yaustar, line 2:
```suggestion
 * A ordered list-type data structure that can provide item look up by key and can also return a list.
```

yaustar (approved):
LGTM with comments

PR #3873, opened 2022-01-05
Debug rendering of world clusters (clustered lighting)

mvaligursky:
- new WorldClustersDebug class, available in the debug build of the engine, responsible for rendering cluster cells
- fix to bounding box evaluation of the spot light. It was incorrectly using scale of its worldMatrix to scale the bounds, making lights less likely to be culled, and also occupying large area in clusters.

**New API (private for now)**
- **app.scene.lighting.debugLayer** - allows to enable debug rendering of clusters on a layer, or disable it by setting to undefined

**Updated public API**
- **BoundingBox.setFromTransformedAabb** - added optional parameter allowing the scale to be ignored when aabb is transformed by a matrix

Screenshot 2022-01-05 at 11 44 15

### DIFF
--- a/examples/src/examples/graphics/clustered-lighting.tsx
+++ b/examples/src/examples/graphics/clustered-lighting.tsx
@@ -10,6 +10,7 @@ class ClusteredLightingExample extends Example {
 
     load() {
         return <>
+            <AssetLoader name='script' type='script' url='static/scripts/camera/orbit-camera.js' />
             <AssetLoader name='normal' type='texture' url='static/assets/textures/normal-map.png' />
         </>;
     }
@@ -153,7 +154,7 @@ class ClusteredLightingExample extends Example {
             color: pc.Color.WHITE,
             intensity: 0.2,
             range: 300,
-            shadowDistance: 300,
+            shadowDistance: 600,
             castShadows: true,
             shadowBias: 0.2,
             normalOffsetBias: 0.05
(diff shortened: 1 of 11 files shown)


willeastcott:
Fancy adding a unit test for your update to `BoundingBox`? 😄

mvaligursky:
> Fancy adding a unit test for your update to `BoundingBox`? 😄

Probably not as part of this PR considering we don't yet have any tests for any shapes.

examples/src/examples/graphics/clustered-lighting.tsx
```diff
@@ -163,12 +164,25 @@ class ClusteredLightingExample extends Example {
...
-            clearColor: new pc.Color(0.2, 0.2, 0.2),
+            clearColor: new pc.Color(0.05, 0.05, 0.05),
             farClip: 500,
             nearClip: 0.1
         });
-        camera.setLocalPosition(120, 120, 120);
+        camera.setLocalPosition(140, 140, 140);
         camera.lookAt(new pc.Vec3(0, 40, 0));
+
+        // add orbit camera script with a mouse and a touch support
```
willeastcott, line 174:
```suggestion
            // add orbit camera script with mouse and touch support
```

examples/src/examples/graphics/clustered-spot-shadows.tsx
```diff
@@ -220,16 +226,38 @@ class ClusteredSpotShadowsExample extends Example {
...
             clearColor: new pc.Color(0.2, 0.2, 0.2),
-            farClip: 1000,
-            nearClip: 0.1
+            farClip: 2000,
+            nearClip: 1
         });
         app.root.addChild(camera);
+        camera.setLocalPosition(300 * Math.sin(0), 150, 300 * Math.cos(0));
+
+        // add orbit camera script with a mouse and a touch support
```
willeastcott, line 235:
```suggestion
        // add orbit camera script with mouse and touch support
```

src/scene/lighting/world-clusters-debug.js
```diff
@@ -0,0 +1,189 @@
...
+        renderCellLines(cells.x, cells.z, boundsMin.x, cellDelta.x, boundsMin.z, cellDelta.z, boundsMin.y, boundsMax.y, 0);
+        renderCellLines(cells.x, cells.y, boundsMin.x, cellDelta.x, boundsMin.y, cellDelta.y, boundsMin.z, boundsMax.z, 1);
+        renderCellLines(cells.y, cells.z, boundsMin.y, cellDelta.y, boundsMin.z, cellDelta.z, boundsMin.x, boundsMax.x, 2);
+
+        // render grid lines
+        if (gridPositions.length) {
+
+            // update colors only when needed
+            const numVerts = gridPositions.length / 3;
+            if (numVerts != gridColors.length / 4) {
```
willeastcott, line 62:
```suggestion
            if (numVerts != gridColors.length / 4) {
```

willeastcott (approved):

willeastcott:
Can we consider only including all this code in a `DEBUG` build?

mvaligursky:
> Can we consider only including all this code in a `DEBUG` build?

Yep, that is done by stripping the static function (and the whole static class) from non debug builds, the same way as Debug.log and all those are stripped.

Issue #3824, opened 2021-12-15
Morph targets are removed when the last entity using the asset in the scene is destroyed


yaustar:
Forum: https://forum.playcanvas.com/t/bug-morph-targets-get-lost-when-entity-is-destroyed/23371
Project: https://playcanvas.com/project/861695/overview/morph-target-bug

Press 1 to delete the entity
Press 2 to create the entity again

Notice that when creating the entity again, it is not animating

PR #3874, opened 2022-01-05
[Fix] Improved handling of the ownership of Morph by Mesh

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/3824

Improvement to ownership of morph by the Mesh - it correctly handles ref-counting, similarly to how MeshInstance handles ownership of the Mesh, or the way MorphInstance handles Morph.

### DIFF
--- a/src/scene/mesh.js
+++ b/src/scene/mesh.js
@@ -227,12 +227,7 @@ class Mesh extends RefCountedObject {
          */
         this.skin = null;
 
-        /**
-         * The morph data (if any) that drives morph target animations for this mesh.
-         *
-         * @type {Morph|null}
-         */
-        this.morph = null;
+        this._morph = null;
         this._geometryData = null;
 
         // AABB for object space mesh vertices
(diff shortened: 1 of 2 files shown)


src/scene/mesh.js
```diff
@@ -260,6 +279,19 @@ class Mesh extends RefCountedObject {
...
     destroy() {
+
+        const morph = this.morph;
+        if (morph) {
+
+            // this decreases ref count on the morph
+            this.morph = null;
+
+            // destroy morph
+            if (morph.getRefCount() < 1) {
```
slimbuck, line 290: couldn't `RefCountedObject` do this automatically?

slimbuck (approved):

src/scene/mesh.js
mvaligursky: That was my initial implementation a while back .. and for some forgotten reason it wasn't ideal so I removed it - we had some PR discussions around this as well I think.

Issue #3813, opened 2021-12-13
Should render target samples option be capped to graphicsDevice.maxSamples?


yaustar:
Looks like the render target doesn't cap the samples to the maximum that the device supports, is this a problem? 

Graphics Device https://github.com/playcanvas/engine/blob/d1ed79490ab300886409e43c34f55d37eb0c326e/src/graphics/graphics-device.js#L1326

Render target: https://github.com/playcanvas/engine/blob/dev/src/graphics/render-target.js#L111

SlimeQ:
It's a pretty opaque problem to figure out. I ended up with this configuration

`this.renderTarget = new pc.RenderTarget({
            colorBuffer: colorBuffer,
            depth: true,
            flipY: true,
            samples: pc.app.graphicsDevice.maxSamples
        });`
        
As you've noted on discord though, pc.app.graphicsDevice is a private API property and subject to change. Probably would be nice to have something more solid.

PR #3875, opened 2022-01-05
Render target limits number of samples to supported maximum

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/3813

If more than supported number of samples are requested, the value is clamped to valid range when the framebuffer is created.

### DIFF
--- a/examples/src/examples/graphics/render-to-texture.tsx
+++ b/examples/src/examples/graphics/render-to-texture.tsx
@@ -70,7 +70,8 @@ class RenderToTextureExample extends Example {
             colorBuffer: texture,
             depth: true,
             // @ts-ignore
-            flipY: true
+            flipY: true,
+            samples: 2
         });
 
         // create a layer for object that do not render into texture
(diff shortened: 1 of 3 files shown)


slimbuck:
does [this](https://github.com/playcanvas/engine/blob/dev/src/graphics/render-target.js#L252) also need special handling?

willeastcott:
Why not just reinstate GraphicsDevice as the first constructor parameter (it used to be that - I'm not sure why it changed). And then just read `maxSamples` there. No other class gets/sets the device after creation. And pretty much most other graphics classes take the graphics device as the first param.

mvaligursky:
I agree @willeastcott. More work, but the result will be better.

mvaligursky:
Actually, maybe instead of passing the device to RenderTarget constructor, I'll just get it from either colorBuffer or depthBuffer, one of which must be set.

willeastcott:
Yeah, you could do that too I guess.

willeastcott (approved):
Lovely. Much better.

Issue #3884, opened 2022-01-07
Refactor internal class VisibleInstanceList


mvaligursky:
The class constains an array, with its own length, and also a separate member length, and those are not in sync. We need to remove the additional .length and simply keep array's length consistent with its content, as having two numbers that do not much is very error-prone.

```
class VisibleInstanceList {
    constructor() {
        this.list = [];
        this.length = 0;
        this.done = false;
    }
}
```

mvaligursky:
Fixed in #5536 by removing the class / using replacement solution.

PR #3885, opened 2022-01-07
[Fix] Fix the way Immediate class was using visible list

mvaligursky:
related to https://github.com/playcanvas/engine/issues/3884

Fix app.drawMesh rendering (also used by app.drawTexture)


slimbuck:
Not really sure what this PR is fixing?

mvaligursky:
it uses visibleList.length length which is valid, and does not depend on visibleList.list.length, which is not. In my test case, the array had 0 meshinstances (visibleList.length was 0), but the array itself had the length of 4 with meshes from the previous frames.

But the real fix / refactoring to this mess will come from different PR which refactors VisibleLIst.

PR #3890, opened 2022-01-10
Fix shift out of range

willeastcott:
Fixes:

image

While the code in `bytesToInt32` works fine, LGTM complains, but also, it requires fairly deep understanding of how numbers are represented in JavaScript. This PR makes the code far easier for anybody to understand. And yes, I know it will be slower, but for readability purposes (plus fixing that LGTM error), I'm happy to update it.

The PR also improves related unit tests as well.

### DIFF
--- a/src/math/math.js
+++ b/src/math/math.js
@@ -115,7 +115,7 @@ const math = {
 
         // Why ((r << 24)>>>0)?
         // << operator uses signed 32 bit numbers, so 128<<24 is negative.
-        // >>> used unsigned so >>>32 converts back to an unsigned.
+        // >>> used unsigned so >>>0 converts back to an unsigned.
         // See http://stackoverflow.com/questions/1908492/unsigned-integer-in-javascript
         return ((r << 24) | (g << 16) | (b << 8) | a) >>> 0;
     },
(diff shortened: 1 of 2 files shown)


mvaligursky (approved):

slimbuck (commented):
This implementation is now quite different to bytesToInt24, intToBytes32 and intToBytes24 which all use bit operators. If the existing function works and is fast, is there really a need to change it? I don't see users particularly needing to understand or maintain the code?

willeastcott:
How would you propose fixing the LGTM error then?

slimbuck:
I would make it >>> 0? (which I believe does the same thing, but haven't read the stack overflow article)

willeastcott:
I'll give it a try. Happy to adopt that if it works.

slimbuck:
TBH I'm not really sure how `>>> 32` can possibly give any other result than 0.

willeastcott:
Hence why I don't like this code.

willeastcott:
It's not the bot shifting I don't like. It's the triple shift operator.

slimbuck:
I've just checked in the debugger and '>>> 32' does the same as '>>> 0'. Perhaps there is some legacy reason to use 32 over 0? Does using 0 calm LGTM?

willeastcott:
I don't know how to check LGTM without actually merging. I did try to get it installed in VS Code but didn't succeed. Anyway, I've updated to `>>> 0` and the unit tests all pass. I'd say it's safe.

PR #3895, opened 2022-01-12
Smart atlas slot distribution for clustered light shadows and cookies

mvaligursky:
This implements custom allocation strategy support for atlas used by shadows and cookies of cluster lights. This allows atlas to have slots of equal sizes (default allocation), or it can be split into slots of different sizes (for example 3 sizes), and the system assigned slots to lights based on their screen space size, which in effect is an LOD system allowing nearby (larger) lights to use higher resolution slots. When there is more lights than slots, the smallest screen area lights do not have slot allocated and simply don't use shadows / cookies.

The new (for now private) API:

**LightingParams.atlasSplit** = null;

Details:
```
// null: per frame split atlas into equally sized squares for each shadowmap / cookie needed
// array: fixed slot allocation - the first number specifies top subdivision of the atlas, following numbers
//            split next level (2 levels only)
```
Examples
```
// automatic - split atlas each frame to give all required lights an equal size
null,

// 7 shadows: split atlas to 2x2 (first number), and split created quarters to 1x1, 1x1, 2x2, 1x1
[2, 1, 1, 2, 1],

// 12 shadows: split atlas to 3x3 (first number), and split one of the created parts to 2x2
[3, 2],

// 16 shadows: split atlas to 4x4
[4]
```

This additionally adds support for SHADOWUPDATE_THISFRAME shadow update mode. This allows shadows to only render into the atlas one time, or when their slot allocation changes.

The slot to light assignment logic is following:
- internally the atlas slots are sorted in the descending order (done when atlas split changes)
- every frame all visible lights are sorted by their screen space size (this handles all cameras where lights are visible using max value)
- all lights in this order get a slot size from the slot list in the same order. Care is taken to not reassign slot if the size of it is the same and only index changes.

Example with all the new controls exposed to test all scenarios (see above mentioned split to create 12 slots)
Also note lights away from the camera use lower resolution shadows.
Screenshot 2022-01-12 at 10 45 56

### DIFF
--- a/examples/src/examples/graphics/clustered-spot-shadows.tsx
+++ b/examples/src/examples/graphics/clustered-spot-shadows.tsx
@@ -31,6 +31,7 @@ class ClusteredSpotShadowsExample extends Example {
         return <>
             <AssetLoader name='script' type='script' url='static/scripts/camera/orbit-camera.js' />
             <AssetLoader name="channels" type="texture" url="static/assets/textures/channels.png" />
+            <AssetLoader name="heart" type="texture" url="static/assets/textures/heart.png" />
             <AssetLoader name='normal' type='texture' url='static/assets/textures/normal-map.png' />
             <AssetLoader name='cubemap' type='cubemap' url='static/assets/cubemaps/helipad.dds' data={{ type: pc.TEXTURETYPE_RGBM }}/>
         </>;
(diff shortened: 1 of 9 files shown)


Maksims:
Should the default value be null, instead of undefined?
null - is explicit definition,. Undefined - is rarely used as defaults in engine.

mvaligursky:
> Should the default value be null, instead of undefined?

Fair point, done now, thanks!

slimbuck (approved):

PR #3902, opened 2022-01-17
Lightmapper is compatible with clustered light mode

mvaligursky:
- Lightmapper can use clustered lights when baking now - it internally allocates clustered lights structure which is updated with lights during baking.

- Clustered lighting has been improved to support flags on the light: affectLightmapped, affectRealtime.

- Renamed light mask constant to match the light properties (those are all private)
Screenshot 2022-01-17 at 17 49 14

### DIFF
--- a/src/scene/constants.js
+++ b/src/scene/constants.js
@@ -617,9 +617,10 @@ export const SHADOWUPDATE_REALTIME = 2;
 export const SORTKEY_FORWARD = 0;
 export const SORTKEY_DEPTH = 1;
 
-export const MASK_DYNAMIC = 1;
-export const MASK_BAKED = 2;
-export const MASK_LIGHTMAP = 4;
+// flags used on the mask property of the Light, and also on mask property of the MeshInstance
+export const MASK_AFFECT_DYNAMIC = 1;
+export const MASK_AFFECT_LIGHTMAPPED = 2;
+export const MASK_BAKE = 4;
 
 /**
  * Render shaded materials with gamma correction and tonemapping.
--- a/src/scene/lightmapper/lightmapper.js
+++ b/src/scene/lightmapper/lightmapper.js
@@ -162,14 +162,26 @@ class Lightmapper {
         // create light cluster structure
         if (this.scene.clusteredLightingEnabled) {
 
+            // create light params, and base most parameters on the lighting params of the scene
             const lightingParams = new LightingParams(device.supportsAreaLights, device.maxTextureSize, () => {});
             this.lightingParams = lightingParams;
 
+            const srcParams = this.scene.lighting;
+            lightingParams.shadowsEnabled = srcParams.shadowsEnabled;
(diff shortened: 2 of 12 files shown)


Maksims:
If there is a way to automatically figure out number of lights per cluster and number of clusters when baking for lightmaps, to avoid hitting a limits with too dense/high number of lights while baking lightmaps?

mvaligursky:
Lights are getting baked one at a time as before, so the clustered structure is set to support up to 4 lights are. time (single texture read, minimum supported) .. 

I considered baking more lights at a time (all that could fit their shadows into the atlas), but this has not been done (yet).

src/scene/lightmapper/lightmapper.js
```diff
@@ -447,7 +472,8 @@ class Lightmapper {
                         meshInstance._shaderDefs |= shaderDefs;
                     }
 
-                    meshInstance.mask = MASK_BAKED;
+                    // only lights that affect lightmapped objects are used on this mesh now that is baked
```
willeastcott, line 475: `now that *it* is baked`?

src/scene/constants.js
```diff
@@ -617,9 +617,10 @@ export const SHADOWUPDATE_REALTIME = 2;
 export const SORTKEY_FORWARD = 0;
 export const SORTKEY_DEPTH = 1;
 
-export const MASK_DYNAMIC = 1;
-export const MASK_BAKED = 2;
-export const MASK_LIGHTMAP = 4;
+// flags used on the mask property of the Light, and also on mask property of the MeshInstance
+export const MASK_AFFECT_DYNAMIC = 1;
+export const MASK_AFFECT_LIGHTMAPPED = 2;
```
willeastcott, line 622: Is the engine stadnard `ENUM_TYPE_ENUM_VALUE` or `ENUMTYPE_ENUMVALUE`? While I can find some of the former, the latter seems way more prevalent.

src/scene/lightmapper/lightmapper.js
mvaligursky: done, thanks

src/scene/constants.js
mvaligursky: personally I prefer the more readable version with the extra _

willeastcott: OK, so are you advocating migrating the engine's enum constants to the more liberal use of underscore?

mvaligursky: I guess so .. looking at it again, it seems more than 50% already use underscores for longer constants .. but I'm not sure it's worth adjusting the rest, especially public constants?

willeastcott: :shrug: Just pointing out this is definitely an area of inconsistency in the engine right now.

willeastcott: TBH, I wish we just used literal strings for most constants.

willeastcott (approved):

PR #3908, opened 2022-01-18
Simplify LayoutGroupComponent initialization

willeastcott:
Logic in `LayoutGroupComponentSystem#initializeComponentData` was overly complex. I suspect the author was trying extremely hard to avoid any allocations. However, in any given project, not many layoutgroup components are normally created, and these days, vectors are very quick to allocate (because they're no longer based on `Float32Array`s). Anyway, this PR simplifies it hugely.

Fixes:

image

The PR also gets rid of programmatic creation of properties on `LayoutGroupComponent` and uses regular ES6 getters/setters instead.

BREAKING CHANGE: The component no longer stores the reference to set vectors. Instead, it copies vectors into ones owned by the component. Before, this would fire two reflow events:

```
const padding = new pc.Vec4(1, 1, 1, 1);
entity.layoutgroup.padding = padding;
entity.layoutgroup.padding = padding;
```

Now it just reflows after the first set, and not the second.

You also used to be able to do:

```
const padding = new pc.Vec4(1, 1, 1, 1);
entity.layoutgroup.padding = padding;
padding.set(2, 2, 2, 2);
entity.layoutgroup.wrap = !entity.layoutgroup.wrap; // force a reflow by changing another property - padding also changes
```

Now, the setting of the `padding` variable above will be ignored.

### DIFF
--- a/src/framework/components/layout-group/system.js
+++ b/src/framework/components/layout-group/system.js
@@ -48,40 +48,13 @@ class LayoutGroupComponentSystem extends ComponentSystem {
         if (data.reverseX !== undefined) component.reverseX = data.reverseX;
         if (data.reverseY !== undefined) component.reverseY = data.reverseY;
         if (data.alignment !== undefined) {
-            if (data.alignment instanceof Vec2) {
-                component.alignment.copy(data.alignment);
-            } else {
-                component.alignment.set(data.alignment[0], data.alignment[1]);
-            }
-
-            /* eslint-disable no-self-assign */
-            // force update
-            component.alignment = component.alignment;
-            /* eslint-enable no-self-assign */
+            component.alignment = Array.isArray(data.alignment) ? new Vec2(data.alignment) : data.alignment;
         }
         if (data.padding !== undefined) {
-            if (data.padding instanceof Vec4) {
-                component.padding.copy(data.padding);
-            } else {
-                component.padding.set(data.padding[0], data.padding[1], data.padding[2], data.padding[3]);
-            }
-
-            /* eslint-disable no-self-assign */
-            // force update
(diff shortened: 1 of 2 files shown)


src/framework/components/layout-group/system.js
```diff
@@ -48,40 +48,13 @@ class LayoutGroupComponentSystem extends ComponentSystem {
...
-                component.alignment.copy(data.alignment);
-            } else {
-                component.alignment.set(data.alignment[0], data.alignment[1]);
-            }
-
-            /* eslint-disable no-self-assign */
-            // force update
-            component.alignment = component.alignment;
-            /* eslint-enable no-self-assign */
+            component.alignment = Array.isArray(data.alignment) ? new Vec2(data.alignment) : data.alignment;
```
mvaligursky, line 51: looks good, but should it maybe create new Vec2 in both cases (even when it already is Vec2), to avoid directly referencing data.alignment? Or does it get copied in .alignment?

mvaligursky (approved):

src/framework/components/layout-group/system.js
willeastcott: It gets copied in the setter, so it's fine. It's how all component properties work actually.

willeastcott: I take that back. The component takes a reference. Wow, that's wrong. I'll need to consider this a bit further.

willeastcott: OK, I've updated the component to take copies of vectors instead of storing a reference to what's set. I think the PR's good now.

mvaligursky (approved):
that looks great!

PR #3910, opened 2022-01-19
LayerComposition's constructor no longer requires device as a parameter

mvaligursky:
LayerComposition was recently updated to accept GraphicsDevice parameter, to handle clustered lighting. This is causing a problem when using in Editor, as its LayerCompositions are created before the engine even starts.

This updates the engine to not pass the device to constructor, but pass it to _update function when required.

Even though this changes the API, the old version works harmlessly and device parameter is simply ignored.

There's also a small null check of material, which was removed recently but is occasionally causing null access in the Editor.

### DIFF
--- a/src/framework/application.js
+++ b/src/framework/application.js
@@ -421,7 +421,7 @@ class Application extends EventHandler {
             passThrough: true
         });
 
-        const defaultLayerComposition = new LayerComposition(this.graphicsDevice, "default");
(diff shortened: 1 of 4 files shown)


willeastcott:
This is handy because the unit test module for `LayerComposition` doesn't pass the graphics device anyway.

willeastcott (approved):

PR #3914, opened 2022-01-21
Fix XR JSDoc type tags

willeastcott:
Add missing curly braces to `@type` tag for XR constants.

### DIFF
--- a/src/xr/constants.js
+++ b/src/xr/constants.js
@@ -2,7 +2,7 @@
  * Inline - always available type of session. It has limited features availability and is rendered
  * into HTML element.
  *
- * @type string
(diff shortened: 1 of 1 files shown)


yaustar:
Shouldn't linting or some other check warn us about this?

yaustar (approved):

willeastcott:
@yaustar Good question! 🏆  So it _would_ do if we had the [`valid-types`](https://github.com/gajus/eslint-plugin-jsdoc#valid-types) rule enabled. We have a lot of errors at the moment preventing us from turning it on though. This was actually the first step to enabling it. 😄

PR #3918, opened 2022-01-21
Adopt @ignore JSDoc tag

willeastcott:
Apply the JSDoc [`@ignore` tag](https://jsdoc.app/tags-ignore.html) where it is appropriate. This tag essentially says 'do not include this class/function/property in the docs'. This is better than using `@private` which has a negative impact on Intellisense in VS Code, for example.


mvaligursky (approved):
that's great.

PR #3920, opened 2022-01-23
Improve various JSDoc blocks and unit tests

willeastcott:
* Complete all JSDoc blocks for `graph-node.js` (for better IntelliSense and TS declarations)
* Refactor `GraphNode#findByTag` to just use the public API of `Tags`
* Optimize `GraphNode#removeChild`
* Improve `GraphNode` unit tests (and remove old Karma-based tests)
* Refactor `Channel` and `Channel3d` classes and JSDoc blocks.
* Improve curve related and float packing JSDoc blocks.


src/math/curve-evaluator.js
```diff
@@ -1,7 +1,20 @@
...
+ * A class for evaluating a curve at a specific time.
+ *
+ * @ignore
+ */
 class CurveEvaluator {
+    /**
+     * Create a new CurveEvaluator instance.
+     *
+     * @param {Curve} curve - The curve to evaluate.
+     * @param {number} time - The initial time to evaluate the curve at. Defualts to 0.
```
kungfooman, line 16: *looked up the Irish meaning* 😅 
```suggestion
     * @param {number} time - The initial time to evaluate the curve at. Defaults to 0.
```

willeastcott: Well spotted - thanks! 🙏

src/math/float-packing.js
```diff
@@ -5,16 +5,25 @@ const oneDiv255 = 1 / 255;
 const floatView = new Float32Array(1);
 const int32View = new Int32Array(floatView.buffer);
 
-// utility static class providing functionality to pack float values to various storage representations
+/**
+ * Utility static class providing functionality to pack float values to various storage
+ * representations.
+ *
+ * @ignore
```
mvaligursky, line 12: just curious if we need ignore tag on modules that are not exported from index.js?

mvaligursky (approved):

src/math/float-packing.js
willeastcott: We do, yes, because JSDoc doesn't know about import/export statements and can't discern public from internal API. You have to explicitly use the @ignore everywhere where it's necessary.

willeastcott: As time passes, it seems JSDoc is getting smarter though and leveraging the code to minimize the amount of tag usage. Today, you still need to use `@augments`/`@extends` but I suspect they make that unnecessary at some point: https://github.com/jsdoc/jsdoc/issues/1843

Issue #3915, opened 2022-01-21
Allow access to private Ammo objects in RigidbodyComponent


LeXXik:
At the moment, `RigidbodyComponent` creates private Ammo vectors when the component is initialized:

https://github.com/playcanvas/engine/blob/ec094c438ba9611de924592039871871a996d6c3/src/framework/components/rigid-body/component.js#L18-L20
https://github.com/playcanvas/engine/blob/ec094c438ba9611de924592039871871a996d6c3/src/framework/components/rigid-body/component.js#L68-L75

However, in projects, where Ammo is only one of possible physics engines, Ammo may load later with other project assets. In that case, at the time the `RigidbodyComponent` is initialized, Ammo is not available, so the private variables are not set. When the time comes for Ammo to load, there is no way to set those variables and the component breaks. Consider exposing them via `RigidbodyComponentSystem` or making them as static properties on the component.

LeXXik:
Same thing for the ammo variables in the `RigidbodyComponentSystem`. For now, a workaround is to remove the existing instance of `RigidComponentSystem` via application system registry and the existing rigidbody components on entities, load Ammo, create a new rigidbody component system instance and add to registry. Once added, re-add the rigidbody components to the entities that had it removed and recreate physical shapes. This will use the new rigidbody component instance, which has Ammo variables instantiated.

mvaligursky:
The ones in `RigidbodyComponentSystem` are called inside `onLibraryLoaded` when Ammo is loaded .. would this not get called when Ammo is loaded later?

LeXXik:
Yes, you are correct. It is the component only then.

mvaligursky:
I have a fix (PR shortly), but temporary workaround is probably just to create temporary entity, add RigidBody component on it (which initializes those temps) and then destroy it. There's no need to recreate this component on all entities.

LeXXik:
Thank you, @mvaligursky, much appreciated!

PR #3929, opened 2022-01-26
Internal Ammo variables are initialized when Ammo loads

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/3915 by creating those internal temp variables when Ammo is loaded, and not when each RigidBody component is created, allowing components to be created even when Ammo is not loaded yet.


slimbuck:
Can you take a look at this one @willeastcott?

willeastcott (approved):
LGTM

PR #3936, opened 2022-01-27
 An error occurs when the canvas is destroyed and created

tizkzks96:
Continued from the forum thread below.

There is a phenomenon that the render is broken when the app is destroyed and created without refreshing the page.

To solve this, we added a part to initialize samplesTexCache .

I haven't checked if there are any side effects, so please check and merge them or make additional corrections.

thank you.

Fixes #

### DIFF
--- a/src/framework/application.js
+++ b/src/framework/application.js
@@ -2006,6 +2006,7 @@ class Application extends EventHandler {
         if (getApplication() === this) {
             setApplication(null);
         }
+        samplesTexCache = {};
     }
 
     /**


tizkzks96:
Continued from the forum thread below.
https://forum.playcanvas.com/t/an-error-occurs-when-the-canvas-is-destroyed-and-created/23901

yaustar:
Thata interesting, I wouldn't have thought that variable would be accessible from here 🤔

https://github.com/playcanvas/engine/blob/543380a0d69fd973490ca0ffbddac526a1dad881/src/graphics/reproject-texture.js#L313

mvaligursky:
We have a proper fix in development now, thanks for the report and the contribution!
Closing in favour of https://github.com/playcanvas/engine/pull/3947

Issue #3714, opened 2021-11-18
Improve gamma handling in the engine


mvaligursky:
In most cases source textures storing color information are in gamma space. Editor handles this correctly by defaulting projects to use GAMMA, which brings textures to linear space for lighting, but engine only projects default to `_gammaCorrection = GAMMA_NONE` in the Scene. We should consider switching it to GAMMA. Perhaps even go as far as deprecate this setting and always use GAMMA. The GAMMA_NONE is likely based on old school rendering, where lighting was done in sRGB space for performance reason to avoid the cost of the pow instructions. This should be no longer required.

This is the difference this flag makes to the default shader. Texture sampling is not gamma corrected, and also color output is not converted to gamma space.

<img width="1412" alt="Screenshot 2021-11-18 at 10 51 11" src="https://user-images.githubusercontent.com/59932779/142403948-e07b7a8e-4bf8-4048-8a0f-b0826cd94510.png">

This risk with this change is the impact to existing engine only projects that are using GAMMA_NONE. Changing them to GAMMA would make them visually different. Perhaps we can only switch the default, but keep the option in for now.

Maksims:
Sometimes, for specific things, disabling gamma on material is desired. This will not be affected?

mvaligursky:
Is it desired on a standard material, or a material with a custom shader?
Standard material does lighting .. and that should always be done in correct space. So I would say we can remove support for non-gamma.
Other materials (user written shader) can do what they need.

Maksims:
> Is it desired on a standard material, or a material with a custom shader? Standard material does lighting .. and that should always be done in correct space. So I would say we can remove support for non-gamma from standard material. Other materials (user written shader) can do what they need.

Standard shader.
Here is common example where it is useful to render texture in original pixel color as possible: plane with plank shadow bellow a vehicle, with unlight material and multiply blend. If gamma or anything else affects it, then greyscale will not be from pure white to pure black, and will affect rendering of a plank shadow.
No custom shader here.

mvaligursky:
related: https://github.com/playcanvas/engine/pull/3939

PR #3939, opened 2022-01-27
[BREAKING] Scene.gammaCorrection defaults to GAMMA_SRGB instead of GAMMA_NONE

mvaligursky:
- breaking change for engine only users if they have not changed the default value, and use lighting: intensity of lights & skydome might need to be adjusted, alternatively GAMMA_NONE can be set.
- engine examples were adjusted to visually look good with the change if needed

Fixes https://github.com/playcanvas/engine/issues/3714


willeastcott (approved):
Great improvement

PR #3941, opened 2022-01-27
Use Rollup to bundle TypeScript declarations

willeastcott:
Use Rollup to bundle the engine's TypeScript declarations. The main benefit of doing this is that only publicly exported APIs are included in the TypeScript declarations. Furthermore, this means that `1.52.0` will be 100% backwards compatible with `1.51.0` (with the types still being found at `build/playcanvas.d.ts`).

As an aside, this PR stops extras and the VoxParser from being built for individual targets. Instead, if you want to build these things, you must now just run a vanilla `npm run build`.


types.mjs
```diff
@@ -0,0 +1,23 @@
...
+const regex = /Class<(.*?)>/g;
+const paths = [
+    './types/framework/components/script/component.d.ts',
+    './types/script/script-attributes.d.ts',
+    './types/script/script-registry.d.ts',
+    './types/script/script.d.ts'
+];
+
+paths.forEach(path => {
+    let dts = (fs.readFileSync(path, 'utf8')).toString();
```
kungfooman, line 13: Is the `(...).toString()` here and below really needed? I've never seen it returning a `Buffer` when specifying `utf8`

ellthompson (approved):

types.mjs
willeastcott: Ha, very cool. I love it when you point these things out to me! Fixed.

PR #3947, opened 2022-01-28
Reproject texture cache fix

slimbuck:
This PR introduces a samples cache separate from a per-device texture cache for prefiltering data. A per-device texture cache is required since multiple devices can be created in a single application.

This PR addresses [this](https://forum.playcanvas.com/t/an-error-occurs-when-the-canvas-is-destroyed-and-created/23901/15) forum post.


src/graphics/reproject-texture.js
```diff
@@ -281,74 +306,94 @@ const generateGGXSamples = (numSamples, specularPower, sourceTotalPixels) => {
...
+        });
+    }
+
+    clear() {
+        this.cache.clear();
     }
+}
+
+// cache of samples. we store these seperately from textures since multiple devices can use the same
+// set of samples.
```
mvaligursky, line 363: I wonder if the complication is worth the benefit here? If I understand correctly, this cache saves raw data, and the texture cache saves the same data stored in the texture format per device? Maybe we only need the texture cache, and if there's a miss, then regenerate raw data? It's optimising the uncommon case of multiple devices.

src/graphics/graphics-device.js
```diff
@@ -768,6 +768,8 @@ class GraphicsDevice extends EventHandler {
     destroy() {
         const gl = this.gl;
 
+        this.fire('destroy');
```
mvaligursky, line 771: is the cache handing this event? I'm not sure I see the code (in this PR)

src/graphics/reproject-texture.js
slimbuck: Generating the data isn't free and the additional complexity is pretty minimal, maybe 4 lines of code?

The original issue being addressed will benefit from caching the data too, so I think it's probably worth it.

src/graphics/graphics-device.js
slimbuck: No it isn't. I thought I'd need it but didn't in the end. I meant to ask in the PR summary whether it might be worth keeping the event anyway to address device destroy in future. Also happy to remove until we need it. cc @willeastcott

src/graphics/reproject-texture.js
mvaligursky: fair enough

src/graphics/graphics-device.js
mvaligursky: so how do textures get destroyed / removed from cache when the device is destroyed?

slimbuck: they don't. I'll update it to clear the cache when device is destroyed.

src/graphics/reproject-texture.js
```diff
@@ -281,74 +306,97 @@ const generateGGXSamples = (numSamples, specularPower, sourceTotalPixels) => {
...
     }
 
-    const samples = generateLambertSamples(numSamples, sourceTotalPixels);
-    const samplesTex = packSamplesTex(device, key, samples);
-    samplesTexCache[key] = samplesTex;
-    return samplesTex;
+    get(device, key, missFunc) {
+        return this.cache.get(device, () => {
+            device.on('destroy', () => {
+                this.cache.delete(device);
```
mvaligursky, line 352: should this loop over textures in the cache and call destroy on them too?

slimbuck: I wondered that too, but don't think it's necessary. Calling `destroy()` on the texture is necessary to clear up the graphics device-side objects, but that will happen anyway since the device is busy being destroyed. The other reason to invoke destroy on texture is to clear references to the underlying `levels` data. In our case this isn't necessary either since it's cached.

That said, perhaps we should destroy them anyway? On the one hand I don't like adding unnecessary code 'just in case', but on the other hand it might be good to "set a precedent".

What do you think?

mvaligursky: I prefer to destroy all resources like this if possible, as this makes tracking application side leaks easier - for example by printing the list of textures left over in the device.destroy

src/graphics/reproject-texture.js
```diff
@@ -78,6 +78,31 @@ const packFloat32ToRGBA8 = (value, array, offset) => {
...
+    const h = Math.ceil(numSamples / w);
+    const data = new Uint8Array(w * h * 4);
+
+    // normalize float data and pack into rgba8
+    let off = 0;
+    for (let i = 0; i < numSamples; ++i) {
+        packFloat32ToRGBA8(samples[i * 4 + 0] * 0.5 + 0.5, data, off + 0);
+        packFloat32ToRGBA8(samples[i * 4 + 1] * 0.5 + 0.5, data, off + 4);
+        packFloat32ToRGBA8(samples[i * 4 + 2] * 0.5 + 0.5, data, off + 8);
+        packFloat32ToRGBA8(samples[i * 4 + 3] / 8, data, off + 12);
```
kungfooman, line 95: Removing four multiplications and the `/ 8` division each iteration (but I didn't test actual performance difference nor the code itself):

```suggestion
    const n = numSamples * 4;
    let i = 0;
    while (i < n) {
        packFloat32ToRGBA8(samples[i++] * 0.5 + 0.5, data, off + 0);
        packFloat32ToRGBA8(samples[i++] * 0.5 + 0.5, data, off + 4);
        packFloat32ToRGBA8(samples[i++] * 0.5 + 0.5, data, off + 8);
        packFloat32ToRGBA8(samples[i++] * 0.125, data, off + 12);
```

src/graphics/reproject-texture.js
```diff
@@ -281,74 +306,97 @@ const generateGGXSamples = (numSamples, specularPower, sourceTotalPixels) => {
...
-        height: h,
+        width: packedSamples.width,
+        height: packedSamples.height,
         mipmaps: false,
         minFilter: FILTER_NEAREST,
         magFilter: FILTER_NEAREST,
-        levels: [data]
+        levels: [packedSamples.data]
     });
 };
```
kungfooman, line 320: Not sure if its worth the change, just an example of "property shorthand notation" and "object deconstructuring" to read/write less code:
```js
    const {width, height, data} = packSamples(samples);
    return new Texture(device, {
        name,
        width,
        height,
        mipmaps: false,
        minFilter: FILTER_NEAREST,
        magFilter: FILTER_NEAREST,
        levels: [data]
    });
```

src/graphics/reproject-texture.js
slimbuck: thanks for the suggestion @kungfooman. you make good points ab performance. in this case i'm probably more concerned with readability than execution speed, so would prefer to keep it a little verbose.

mvaligursky (approved):

PR #3949, opened 2022-01-28
Lite version of the Application to allow for tree-shaking (WIP)

mvaligursky:
This is exploratory WIP.

We used to have an Application class which initialised everything and imported lots of modules.
Now the code of this has been moved to the ApplicationLite class, and imports / component system registrations removed. The Application class extends ApplicationLite, and additionally it imports and registers all those component systems.

The goal here is to have an ApplicationLite class, which can be instantiated without systems. Then the systems can be imported / registered manually by the user code, allowing unused component systems to be stripped out by their build system.


slimbuck:
How would you feel about `ApplicationBase`?

mvaligursky:
> How would you feel about `ApplicationBase`?

I don't mind any name really. I've also considered ApplicationCore.

mvaligursky:
I've tested this against the Viewer repo. This by default includes playcanvas minified build, and uses Application. 
I've removed the minified build, and imported only required engine modules for testing. The generated build js file is 5MB (non-minified, non-zipped).
When I changed this to include ApplicationLite, and only created few required components, the size dropped to 4MB. This is a pretty good start with more things we can remove.

Screenshot 2022-01-31 at 11 42 26

willeastcott:
This is pretty cool actually. Nice. As for naming:

`ApplicationLite` -> `App`
`Application` -> keep but deprecate and hide?

Then port all examples to `App`. And then port Editor to `App`.

mvaligursky:
The only problem I see at the moment is scripts, and I don't have a solution for that.

Viewer uses camera-orbit script, and it works as long as all symbols is uses from the engine are already imported into the viewer app. To get them few additional ones to import, I added this the viewer.tsx:

```
import { EVENT_TOUCHSTART, EVENT_TOUCHEND, EVENT_TOUCHCANCEL, 
    EVENT_TOUCHMOVE, MOUSEBUTTON_LEFT, MOUSEBUTTON_MIDDLE, MOUSEBUTTON_RIGHT
} from '../../playcanvas/engine/src/input/constants.js';
import { createScript } from '../../playcanvas/engine/src/script/script.js';
```

and then had to use them so the imports are not ignored:

```
console.log(
            createScript,
            EVENT_TOUCHSTART, EVENT_TOUCHEND, EVENT_TOUCHCANCEL, EVENT_TOUCHMOVE,
            MOUSEBUTTON_LEFT, MOUSEBUTTON_MIDDLE, MOUSEBUTTON_RIGHT
        );
```

So I think we need to find some better solution. Like convert scripts to actual modules or something. Any ideas?

willeastcott:
Yeah, that's a pretty huge problem (at least, that's how it feels on the surface). It would be cool if scripts were modules, sure. I wonder what @Maksims thinks about this subject... 🤔

mvaligursky:
Maybe we could add a support for 'ModuleScript'. It would be more or less normal script, but with imports on top, and export and wrapped in a class. Engine users could simply use this - they would import it as a module and build their app. The class would extend some base class we have that registers it and adds attributes and all that as currently.

In Editor, when we parse the script, we would generate the current script format from it as well, and Editor would use that in launcher / even in published builds. Eventually, the user could opt into a custom engine build, and using the module versions of scripts to get three-shaking benefits - this would be only allowed for project with all script in module format.

mvaligursky:
Note: The Application class has changed a lot on dev, and this branch will need work / be recreated.

mvaligursky:
due to large conflits due to recent refactoring, closing this in favour of https://github.com/playcanvas/engine/pull/4082

Issue #3926, opened 2022-01-26
Source map generation no longer works


yaustar:
https://github.com/playcanvas/engine#generate-source-maps

Source maps look like this for the dbg version

![image](https://user-images.githubusercontent.com/16639049/151144337-6ec7ecfd-e7f9-44d7-b374-51d35666ce57.png)

PR #2010, opened 2020-04-23
Removed copyright if sourcemaps are needed

yaustar:
Fixes off by 4 lines in source map

I confirm I have signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #3952, opened 2022-01-28
Fix source maps

kungfooman:
Fixes https://github.com/playcanvas/engine/issues/3926

The mapping in `spacesToTabs()` is preventing the source maps to work out (for whatever reason), but I figured that `null` works just as well. The rollup API/docs don't really help me understand what is going on though: 

**Source Code Transformations** in https://github.com/rollup/rollup/blob/master/docs/05-plugin-development.md


willeastcott (approved):

willeastcott:
Awesome, thanks!! @yaustar - this is something you've played with a bit. I've assigned to you for final approval and merging.

yaustar:
Thanks @kungfooman! 

Debugging playcanvas.js looks fine but when trying to add breakpoints for playcanvas.dbg.js, it's a few lines off as seen in the video below

https://user-images.githubusercontent.com/16639049/151827274-e1e3c756-1cc8-447f-afec-bece8d0a25e0.mp4

kungfooman:
Maybe the `// #if ...` preprocessor plugin is mixing lines up?

yaustar:
I was wondering that. It did cause issues pre rollup where I could only get source files working on DBG versions of the engine.

I also had to do this nasty hack to get it to work too 😅 https://github.com/playcanvas/engine/pull/2010

kungfooman:
> I also had to do this nasty hack to get it to work too 😅 #2010

😅  I would try this option for `jscc`, but I don't have the time for that now: 

image

Just keeping the lines as comments seems better than trying to fight to make different source map systems "cooperate"... these source map line offset errors are all over the place in the JS bundler world.

Comments will be removed/minified anyway later?

yaustar:
@willeastcott , would this affect the engine build?

willeastcott:
I don't mind about lines in the debug engine I _guess_. But I would mind in the release, minified release or module build. Not sure about the profile build....maybe that one would be OK too.

kungfooman:
Okay, I think the result is good news, `keepLines` works better than I first expected.

First, example video that it actually works:

https://user-images.githubusercontent.com/5236548/151937072-8589367c-5a9b-4ba8-8572-15f9c088ffb7.mp4

And a picture from the `profiler` build source map:

image

The breakpoint works flawlessly and the "kept lines" refer to the source map here... you can see the `#if _DEBUG` is "kept/grayed out" and NOT the lines in the actual source code:

image

I think with this info the `keepLines` is also good for all other build targets, but I will let you guys decide on that.

yaustar:
I'm still having offset issues when adding breakpoints in the dbg.js where it is off by one in some very weird cases 🤔

https://user-images.githubusercontent.com/16639049/151968393-aa8b7f8e-79a9-4762-ae1b-41896b8d91f0.mp4

kungfooman:
Thank you for testing @yaustar !

Can you check in Firefox too? For me it works in Firefox and Chrome has this "go to end of previous function instead of clicked function" issue:

https://user-images.githubusercontent.com/5236548/151990261-1907c5ec-655b-43ba-9a14-3f978bfacf12.mp4

So I am not sure if this is still a problem with the source map or a Chrome quirk. 

Also tried source-map-visualizer which looks good to me, one issue I can see is that function names don't seem to be "mapped": http://sokra.github.io/source-map-visualization/#custom

https://user-images.githubusercontent.com/5236548/151993733-425e0a27-c767-4fc7-84d9-9bc0cce60aaa.mp4

But I don't know all the stuff that `rollup` does under the hood and how simple it is to get the one-off-error in Chrome fixed. Firefox seems to take the right choice in case of uncertainty, while Chrome ends up at the previous function.

yaustar:
Oops. Sorry, missed the last update. Will check tomorrow

yaustar:
Checked with Firefox, works so much better! I'm happy to have this merged and raise a second ticket about off by one issue

yaustar:
Can you confirm that you have no more commits incoming :) ?

yaustar:
Oh @kungfooman, can you remove build tests folder please?

willeastcott (approved):

yaustar:
Fantastic, thank you very much for your work @kungfooman !

PR #3962, opened 2022-01-30
Enable IntelliSense for WebGL API

willeastcott:
This PR correctly sets the type of the `gl` property of the `GraphicsDevice` which suddenly enables IntelliSense when working with the WebGL API:

intellisense

It also:
* Ensures `createShaderFromCode` is not passed `null` from `reprojectTexture`.
* Simplifies `GraphicsDevice#initializeExtensions`.

### DIFF
--- a/src/graphics/graphics-device.js
+++ b/src/graphics/graphics-device.js
@@ -892,7 +892,7 @@ class GraphicsDevice extends EventHandler {
 
         const supportedExtensions = gl.getSupportedExtensions();
 
-        const getExtension = () => {
+        const getExtension = function () {
             for (let i = 0; i < arguments.length; i++) {
                 if (supportedExtensions.indexOf(arguments[i]) !== -1) {
                     return gl.getExtension(arguments[i]);
--- a/src/graphics/graphics-device.js
+++ b/src/graphics/graphics-device.js
@@ -227,6 +227,7 @@ class GraphicsDevice extends EventHandler {
      * The WebGL context managed by this graphics device.
      *
      * @type {WebGLRenderingContext}
+     * @ignore
      */
     gl;
 
--- a/src/graphics/graphics-device.js
+++ b/src/graphics/graphics-device.js
@@ -266,7 +266,7 @@ class GraphicsDevice extends EventHandler {
      */
     precision;
 
-     /**
+    /**
      * The scope namespace for shader attributes and variables.
      *
      * @type {ScopeSpace}
--- a/src/graphics/graphics-device.js
+++ b/src/graphics/graphics-device.js
@@ -300,6 +300,7 @@ class GraphicsDevice extends EventHandler {
      * being used.
      *
      * @type {boolean}
+     * @ignore
      */
     webgl2;
 
--- a/src/graphics/graphics-device.js
+++ b/src/graphics/graphics-device.js
@@ -224,9 +224,12 @@ class GraphicsDevice extends EventHandler {
     canvas;
 
     /**
-     * The WebGL context managed by this graphics device.
+     * The WebGL context managed by this graphics device. The type could also technically be
+     * `WebGLRenderingContext` if WebGL 2.0 is not available. But in order for IntelliSense to be
+     * able to function for all WebGL calls in the codebase, we specify `WebGL2RenderingContext`
+     * here instead.
      *
-     * @type {WebGLRenderingContext}
+     * @type {WebGL2RenderingContext}
      * @ignore
      */
     gl;
--- a/src/graphics/graphics-device.js
+++ b/src/graphics/graphics-device.js
@@ -224,12 +224,18 @@ class GraphicsDevice extends EventHandler {
     canvas;
 
     /**
-     * The highest shader precision supported by this graphics device. Can be 'hiphp', 'mediump' or
-     * 'lowp'.
+     * The WebGL context managed by this graphics device.
      *
-     * @type {string}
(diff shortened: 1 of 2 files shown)


src/graphics/graphics-device.js
```diff
@@ -224,12 +224,18 @@ class GraphicsDevice extends EventHandler {
     canvas;
 
     /**
-     * The highest shader precision supported by this graphics device. Can be 'hiphp', 'mediump' or
-     * 'lowp'.
+     * The WebGL context managed by this graphics device.
      *
-     * @type {string}
+     * @type {WebGLRenderingContext}
```
mvaligursky, line 229: could we mark it with `@ignore`? Ideally we don't expose it publicly I think, as I'm hoping to make most of public interface here private due to WebGPU.

kungfooman, line 229: Do you intent to not include `WebGL2RenderingContext` constants? Whatever is decided upon, it would be a bit tricky to remember what was decided upon. Maybe instead of checking `if (this.webgl2)` there could be another context variable called `this.gl1` and then it can have the correct type and the if-check just becomes `if (this.gl1)` e.g.

src/graphics/graphics-device.js
```diff
@@ -1242,15 +1268,26 @@ class GraphicsDevice extends EventHandler {
...
         if (this.activeFramebuffer !== fb) {
-            this.gl.bindFramebuffer(this.gl.FRAMEBUFFER, fb);
+            const gl = this.gl;
+            gl.bindFramebuffer(gl.FRAMEBUFFER, fb);
             this.activeFramebuffer = fb;
         }
     }
 
+    /**
+     * Checks the completeness status of the currnently bound WebGLFramebuffer object.
```
kungfooman, line 1286:
```suggestion
     * Checks the completeness status of the currently bound WebGLFramebuffer object.
```

src/graphics/graphics-device.js
willeastcott: Thanks!

willeastcott: So I had a play with this last night by setting the type to `WebGLRenderingContext|WebGL2RenderingContext ` and that didn't work. What you suggest (maintaining two context properties) would be a fairly significant change so, for now, I thought it would be useful to just enable 90% of the WebGL calls to work with IntelliSense. Supporting WebGL2 can be handled in a follow up PR if we decide on a good path forwards there.

willeastcott: Sure, no problem. Does that mean you also want to hide `webgl2`?

mvaligursky: Perhaps even `webgl2`, even though I don't mind this one, it's just a bool in this case, so does not expose direct access to context

willeastcott: The `webgl2` property is currently used in the transform feedback and grab pass examples. It feels to me like there should be a better way to test for the presence of transform feedback than by testing that property. I'm less sure about the shader in the grab pass example. Anyway, for now, I'll hide that too again.

mvaligursky: fair decision - perhaps capture some of it as a comment in the code please.

src/graphics/graphics-device.js
```diff
@@ -253,11 +260,12 @@ class GraphicsDevice extends EventHandler {
     maxVolumeSize;
 
     /**
-     * The maximum supported texture anisotropy setting.
+     * The highest shader precision supported by this graphics device. Can be 'hiphp', 'mediump' or
+     * 'lowp'.
      *
-     * @type {number}
+     * @type {string}
```
mvaligursky, line 266: if this one wasn't public, perhaps hide this as well .. at least till we find some common interface between webgl/webgpu (if needed there)

src/graphics/graphics-device.js
willeastcott: Done. And I've just switched the type to `WebGL2RenderingContext` which allows all WebGL 2.0 calls to work with IntelliSense too. See comment for justification.

src/graphics/graphics-device.js
willeastcott: It was public. See here:

https://developer.playcanvas.com/api/pc.GraphicsDevice.html#precision

mvaligursky (approved):

src/graphics/graphics-device.js
kungfooman: > So I had a play with this last night by setting the type to `WebGLRenderingContext|WebGL2RenderingContext ` and that didn't work.

You need an intersection via `WebGLRenderingContext & WebGL2RenderingContext`

https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#intersection-types

kungfooman: > WebGL 2 is a superset of WebGL1 so `WebGL2RenderingContext` should be the same as `WebGLRenderingContext & WebGL2RenderingContext` right?

Yea, thats what I also think, they stick to "extending" by only adding more optional arguments to all kinds of functions, e.g. for `readPixels`:

https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts#L16131-L16133

https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts#L16983

At least I didn't see one function so far in `WebGLRenderingContext` that doesn't exist in some form (with optional arguments often) in `WebGL2RenderingContext` too, so as it is right now, its all good

PR #3970, opened 2022-02-01
Examples Browser fixes

### LINKED ISSUE #3958: Two copies of `wasm-loader.js`
There are currently two copies of `wasm-loader.js` in the `examples` folder:

* `examples/lib/wasm-loader.js`
* `examples/src/wasm-loader.js`

Are both needed?

ellthompson:
This PR readies the examples browser for release alongside the 1.52.0 engine.

Changes include:
- The `pc` type definitions are now displaying correctly in the code editor in both javascript and typescript modes.
- Removed a duplicate of the wasm loader file (Fixes #3958).
- All examples now correctly respond the resize events (Fixes ##3959).
- Updated the process for local development of examples and included new documentation for this in the README file.
- Code editing in the browser now correctly works in TypeScript mode.
- Fixed an issue that prevented the mini-stats example from loading.
- Updated all paths to static files to ensure parity when loading an example in both the main application and as an individual iframe.


examples/scripts/iframe/index.mustache
```diff
@@ -60,6 +62,23 @@
                 if (window.top._showMiniStats !== undefined) miniStats.enabled = window.top._showMiniStats;
             });
 
+            var canvasContainerElement = canvas.parentElement;
+            canvas.setAttribute('width', window.innerWidth + 'px');
+            canvas.setAttribute('height', window.innerHeight + 'px');
+            var resizeTimeout = null;
+            pc.app.setCanvasResolution(pc.RESOLUTION_AUTO);
```
willeastcott, line 69:
```suggestion
            app.setCanvasResolution(pc.RESOLUTION_AUTO);
```

examples/src/app/code-editor.tsx
```diff
@@ -37,7 +37,11 @@ const CodeEditor = (props: CodeEditorProps) => {
 
     const beforeMount = (monaco: any) => {
         monaco.languages.typescript.typescriptDefaults.addExtraLib(
-            playcanvasTypeDefs,
+            playcanvasTypeDefs + ' export as namespace pc;',
```
willeastcott, line 40: Doesn't this suggest there's something missing from the `playcanvas.d.ts` file? `export as namespace pc` used to be in the old types but I thought maybe it was unneeded. I can add it back if necessary?

ellthompson: Yeah that'd be good as this used to be present.

examples/src/app/helpers/formatters.mjs
```diff
@@ -34,7 +34,11 @@ const getTypeScriptFunctionFromText = (text) => {
 };
 
 const getInnerFunctionText = (text) => {
-    text = text.substring(text.indexOf("{") + 1);
+    if (text.includes(': void {')) {
+        text = text.substring(text.indexOf(": void {") + 8);
```
mvaligursky, line 38: the `+8` bit here does not fill me with confidence. Perhaps this should be refactored to be more self-explanatory, or a comment should be added?

ellthompson: Yeah it's definitely brittle, unfortunately I've found no great way to template these example files so I can extract various parts in a less brittle way. I'll try and update this to be more self explanatory though.

examples/src/app/code-editor.tsx
ellthompson: I've updated the rollup config as suggested.

willeastcott (approved):

PR #3982, opened 2022-02-04
Extracted webgl implementation of Vertex and Index buffer into separate modules

mvaligursky:
Related to: https://github.com/playcanvas/engine/issues/3986

This is part of refactoring plan, where all webgl specific functionality will move to graphics/webgl, allowing the WebGPU implementation.

- first step is to refactor IndexBuffer and VertexBuffer, and the shared buffer functionality has moved to WebglBuffer (which will also be used to implement UniformBuffer in the future)
- interfaces are private and are subject to change

### DIFF
--- a/src/graphics/graphics-device.js
+++ b/src/graphics/graphics-device.js
@@ -3,6 +3,8 @@ import { EventHandler } from '../core/event-handler.js';
 import { now } from '../core/time.js';
 import { Debug } from '../core/debug.js';
 import { platform } from '../core/platform.js';
+import { WebglVertexBuffer } from './webgl/webgl-vertex-buffer.js';
+import { WebglIndexBuffer } from './webgl/webgl-index-buffer.js';
 
 import {
     ADDRESS_CLAMP_TO_EDGE,
@@ -820,6 +822,16 @@ class GraphicsDevice extends EventHandler {
         this.scope = null;
         this.canvas = null;
         this.gl = null;
+    }
+
+    // provide webgl implementation for the vertex buffer
+    implementVertexBuffer(vertexBuffer) {
+        return new WebglVertexBuffer(vertexBuffer);
+    }
+
+    // provide webgl implementation for the index buffer
+    implementIndexBuffer(indexBuffer) {
+        return new WebglIndexBuffer(indexBuffer);
     }
 
     // don't stringify GraphicsDevice to JSON by JSON.stringify
@@ -2350,7 +2362,7 @@ class GraphicsDevice extends EventHandler {
 
                 // bind buffer
                 const vertexBuffer = vertexBuffers[i];
-                gl.bindBuffer(gl.ARRAY_BUFFER, vertexBuffer.bufferId);
+                gl.bindBuffer(gl.ARRAY_BUFFER, vertexBuffer.impl.bufferId);
 
                 // for each attribute
                 const elements = vertexBuffer.format.elements;
@@ -2400,10 +2412,10 @@ class GraphicsDevice extends EventHandler {
             // single VB keeps its VAO
             const vertexBuffer = this.vertexBuffers[0];
             Debug.assert(vertexBuffer.device === this, "The VertexBuffer was not created using current GraphicsDevice");
-            if (!vertexBuffer._vao) {
-                vertexBuffer._vao = this.createVertexArray(this.vertexBuffers);
+            if (!vertexBuffer.impl.vao) {
+                vertexBuffer.impl.vao = this.createVertexArray(this.vertexBuffers);
             }
-            vao = vertexBuffer._vao;
+            vao = vertexBuffer.impl.vao;
         } else {
             // obtain temporary VAO for multiple vertex buffers
             vao = this.createVertexArray(this.vertexBuffers);
@@ -2421,7 +2433,7 @@ class GraphicsDevice extends EventHandler {
         // Set the active index buffer object
         // Note: we don't cache this state and set it only when it changes, as VAO captures last bind buffer in it
         // and so we don't know what VAO sets it to.
-        const bufferId = this.indexBuffer ? this.indexBuffer.bufferId : null;
+        const bufferId = this.indexBuffer ? this.indexBuffer.impl.bufferId : null;
         gl.bindBuffer(gl.ELEMENT_ARRAY_BUFFER, bufferId);
     }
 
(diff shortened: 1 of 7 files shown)


src/graphics/graphics-device.js
```diff
@@ -822,6 +824,16 @@ class GraphicsDevice extends EventHandler {
         this.gl = null;
     }
 
+    // provide webgl implementation for the vertex buffer
+    implementVertexBuffer(vertexBuffer) {
```
slimbuck, line 828: Not a big deal since we can change it any time, but I find these names confusing.

How about `createVertexBufferImpl`?

src/graphics/graphics-device.js
```diff
@@ -3,6 +3,8 @@ import { EventHandler } from '../core/event-handler.js';
 import { now } from '../core/time.js';
 import { Debug } from '../core/debug.js';
 import { platform } from '../core/platform.js';
+import { WebglVertexBuffer } from './webgl/webgl-vertex-buffer.js';
```
slimbuck, line 6: So in future this will go into `WebglGraphicsDevice.js`?

slimbuck (commented):
Could you perhaps give us a summary of the planned class structure we'll end up with?

src/graphics/graphics-device.js
mvaligursky: done .. I was originally considering those names too

src/graphics/graphics-device.js
mvaligursky: yes

src/graphics/index-buffer.js
```diff
@@ -89,25 +77,21 @@ class IndexBuffer {
...
-        if (this.bufferId) {
-            const gl = this.device.gl;
-            gl.deleteBuffer(this.bufferId);
-            this.device._vram.ib -= this.storage.byteLength;
-            this.bufferId = null;
-
-            if (this.device.indexBuffer === this) {
-                this.device.indexBuffer = null;
-            }
+        if (this.device.indexBuffer === this) {
```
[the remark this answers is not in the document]
slimbuck: oic that's devices active index buffer, ignore me.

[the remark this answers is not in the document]
mvaligursky: it only has one active IB. This is original code, untouched see line 98 before.

src/graphics/webgl/webgl-buffer.js
```diff
@@ -0,0 +1,52 @@
+import { BUFFER_DYNAMIC, BUFFER_GPUDYNAMIC, BUFFER_STATIC, BUFFER_STREAM } from '../constants.js';
+
+/**
+ * A WebGl implementation of the Buffer.
```
willeastcott, line 4: I suspect that this will appear in the API reference manual unless you add an `@ignore` tag.

willeastcott: And WebGL has a capital 'L'. 😄

mvaligursky: fixed!, thansk

src/graphics/webgl/webgl-index-buffer.js
```diff
@@ -0,0 +1,29 @@
+import { INDEXFORMAT_UINT8, INDEXFORMAT_UINT16, INDEXFORMAT_UINT32 } from '../constants.js';
+import { WebglBuffer } from "./webgl-buffer.js";
+
+/**
+ * A WebGl implementation of the IndexBuffer.
+ */
```
willeastcott, line 6: Same here. Add @ignore and it's `WebGL`.

src/graphics/webgl/webgl-vertex-buffer.js
```diff
@@ -0,0 +1,31 @@
+import { WebglBuffer } from "./webgl-buffer.js";
+
+/**
+ * A WebGl implementation of the VertexBuffer.
+ */
```
willeastcott, line 5:
```suggestion
/**
 * A WebGL implementation of the VertexBuffer.
 *
 * @ignore
 */
```

src/graphics/webgl/webgl-index-buffer.js
willeastcott, line 8:
```suggestion
/**
 * A WebGL implementation of the IndexBuffer.
 *
 * @ignore
 */
```

willeastcott (approved):

willeastcott:
I'm OK enough with this to approve it. But I still have a few thoughts swirling in my head:

1. What can we do to address class cyclic dependencies. Right now the design reinforces this cycle between the Device and the Buffer classes. For example, the WebGL buffer must modify state on the device when it's destroyed. Could that be an event? Or could the global state of the current bound buffer be managed in the WebGLVerterBuffer class?
2. I'm _still_ not sure how we plan to ship a WebGPU flavor to devs. Do we pursue the 'all in one' approach or have two engine versions?
3. Are we concerned about lots of inheritance on performance?
4. I'm a bit undecided on naming. The WebGL spec itself already defined names like `WebGLTexture` and `WebGLBuffer`. I do like keeping the `WebGL` style of capitalization. But the API itself uses some of these names already and we'll probably cause problems if we reuse them. Yeah, just pointing out I'm still thinking about that. Obviously, these classes are private so not a huge deal.

mvaligursky:
> I'm OK enough with this to approve it. But I still have a few thoughts swirling in my head

1. I don't see a problem with classes on the same level in the architecture (in this case inside graphics) calling each other - this is a common practise in js, c#, c++ - it does not create a circular dependency, as only type for type-checking needs to be imported.

2. I see two main directions here:
A) (ideal) We work out this PR (https://github.com/playcanvas/engine/pull/3949) plus follow up PRs, related to creating a lite version of the application. This would allow the device creation to be moved from the Application class itself, into a user code. And so the user could import only WebglGraphicsDevice, construct an instance and assign it to the ApplicationLite. Or they create only Webpu version, or both based on support of the device. And the tree-shaking will only add those modules to the user app during the build.
B) (backup) We ship multiple builds of the engine. Most likely "webgl" and "webgl + webgpu" flavour initially.

3. Only slightly, but I thought about it as well. Most of the functionality requiring calls to various parts of the inheritance chain would be limited to the resource creation / updates, which is not frequent. The hot path, where the resource is used for rendering would usually only access properties directly in the Impl classes. But we'll need to keep an eye on this for sure.

4. We could also adopt some shorter names, perhaps:
- WglTexture and WgpuTexture, or only GlTexture and GpuTexture, even though this one is likely confusing.
- Or change the order ... so TextureWebGl, TextureWebGpu - I'm pretty keen to this one.

PR #3992, opened 2022-02-07
Switch box-reflection example to refresh envmap every 10 frames

mvaligursky:
To improve out of the box performance on mobile devices


slimbuck:
Just thinking... would it make more sense for the update frequency to be controlled 'once per second' instead of per number of frames?

yaustar (approved):

Issue #3986, opened 2022-02-04
WebGPU Support


mvaligursky:
This is the root ticket of WebGPU Implementation, and will be updated regularly to keep it up to date with new issues to work on and the work done. 

For live preview, visit the url on up to date Chrome: https://playcanvas.github.io/

https://user-images.githubusercontent.com/59932779/203333981-3d77fc58-5e5a-4b00-add1-8ab5a3991c84.mov

https://user-images.githubusercontent.com/59932779/205290862-22eff1fd-53d9-4336-815e-5f7311b923d5.mov

https://user-images.githubusercontent.com/59932779/205290893-555888b4-74d2-4b6d-a963-1e0a86daea12.mov

https://user-images.githubusercontent.com/59932779/208122279-d85d1c22-48b3-4db8-b9d8-9872d111c739.mov

https://user-images.githubusercontent.com/59932779/205909226-d3d90aa0-c118-49f7-be0e-b56576a5235c.mov

https://user-images.githubusercontent.com/59932779/213508864-71ba2fb6-7c59-425e-bc3d-db94c2cd6e65.mov

https://user-images.githubusercontent.com/59932779/214621963-bc1b6bd5-8775-40f7-a339-6612b59fd5c2.mov

https://user-images.githubusercontent.com/59932779/217569603-95fc0dbd-d86f-4ccd-badf-42f9873ab3a7.mov

https://user-images.githubusercontent.com/59932779/222208562-efa32552-9da7-4a3d-a27f-f67fc1e91812.mov

https://github.com/playcanvas/engine/assets/59932779/b56291dc-513f-49f6-bcbc-50b0a24ec3cc

**Class layout overview**
- graphics folder will have two subfolders created for respective implementations: `/webgl`, `/webgpu`
- existing `GraphicsDevice` class will be platform independent, and `WebglGraphicsDevice` and `WebgpuGraphicsDevice` will extend it to implement platform specific implementations.
-  other graphic classes, such as VertexBuffer, IndexBuffer and similar will be split into 3 classes as well, and this is the example structure:

```
// base platform independent class
class VertexBuffer {
    impl = null;   // platform dependent implementation (WebglVertexBuffer or WebgpuVertexBuffer)
};
```

**Refactorings**
The engine will undergo extensive refactoring, to more natively and with better performance support WebGPU API.

Issues:

- General
#4261
#2254
#4281
#4271 
#5689 
#5688
#5357
#5690
#2974
#5692

PRs:

- General
#3982
[...]

PR #3993, opened 2022-02-07
WebGL related part GraphicsDevice is moving to a separate class

mvaligursky:
Related to https://github.com/playcanvas/engine/issues/3986

A base GraphicsDevice class is going to be platform independent, and WeblGraphicsDevice will contain WebGL specific implementation. In this first step, the whole implementation was simply moved to WeblGraphicsDevice.


slimbuck:
Have you tried using `git mv graphics-device.js webgl/blah` ? Might keep better history this way.

mvaligursky:
> Have you tried using `git mv graphics-device.js webgl/blah` ? Might keep better history this way.

I did, `git mv` and few different ways. Some of them work if you only move / rename file. But as soon as you modify it, the operation changes to delete/create pair. The suggestion online is to move / rename in a separate PR to the modify. But we cannot move files without modifying .. at least import paths.
I even created a test PR with two commits, first to move, second to change .. 
https://github.com/playcanvas/engine/pull/4007
when you inspect commits, there's perfect, the one with changes only shows small changes. But when you look at changed files for whole PR, the story is different :(

willeastcott (approved):

mvaligursky:
I'm also trying to split it into two separate PRs .. maybe that's a better way. Here: #4009 
I moved graphics-device.js to webgl subfolder .. and it's seen as move, even when I fixed the import paths. So this might preserve the history better. So maybe I do this as a first step, and then do follow up refactor after #4009 is merged.

willeastcott (approved):

PR #3994, opened 2022-02-07
Implement/use defineProtoFunc for better polyfills

kungfooman:
Some polyfill functions are just added to e.g. `String.prototype.endsWith` without using `Object.defineProperty` and this causes `for-in` loops to show "funny" results:

E.g. Internet Explorer results before:

image

image

This PR makes it less code + more polyfills (`String#join` on all typed arrays)

### DIFF
--- a/src/polyfill/defineProtoFunc.js
+++ b/src/polyfill/defineProtoFunc.js
@@ -1,0 +1,16 @@
+/**
+ * 
+ * @param {ObjectConstructor} cls 
+ * @param {string} name 
+ * @param {Function} func 
+ */
+export function defineProtoFunc(cls, name, func) {
+  if (!cls.prototype[name]) {
+      Object.defineProperty(cls.prototype, name, {
+          value: func,
+          configurable: true,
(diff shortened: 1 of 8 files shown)


src/polyfill/defineProtoFunc.js
```diff
@@ -0,0 +1,16 @@
+/**
+ *
```
mvaligursky, line 2: We should add a description and `@ignore` tag to exclude it from generated docs

src/polyfill/typedarray-fill.js
```diff
@@ -1,32 +1,17 @@
...
+    Int8Array,
+    Uint8Array,
+    Uint8ClampedArray,
+    Int16Array,
+    Uint16Array,
+    Int32Array,
+    Uint32Array,
+    Float32Array
+];
+for (var typedArray of typedArrays) {
```
mvaligursky, line 13: var -> const

mvaligursky (approved):
apart from two small comments, this looks great.

PR #3998, opened 2022-02-08
Ambient lighting fix

slimbuck:
This PR fixes an occasional ambient lighting error. In some cases we were attempting to convert an unnormalised direction vector to spherical coordinates. This resulted in us invoking `asin(value)` with `|value| > 1.0` resulting in an undefined value.

This PR normalises the direction vector before conversion.

Before:
Screenshot 2022-02-08 at 12 10 01

After:
Screenshot 2022-02-08 at 12 10 59


willeastcott (approved):

yaustar:
Would this go into 1.51.7 release? Or wait for 1.52.2?

slimbuck:
Likely v1.51.7 release.

PR #4001, opened 2022-02-08
Fix error when a non image element is assigned to imageEntity on button.

rob5300:
This fixes an exception when a button is first enabled when the `imageEntity` is assigned an entity with an element, but is not an image element type.
`element.color` is null which causes an error in both `applyTint` functions.

### DIFF
--- a/src/framework/components/button/component.js
+++ b/src/framework/components/button/component.js
@@ -456,7 +456,7 @@ class ButtonComponent extends Component {
     }
 
     _applyTintImmediately(tintColor) {
-        if (!tintColor || !this._imageReference.hasComponent('element'))
+        if (!tintColor || !this._imageReference.hasComponent('element') || this._imageReference.entity.element.type != ELEMENTTYPE_IMAGE)
             return;
 
         const color3 = toColor3(tintColor);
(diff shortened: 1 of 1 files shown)


src/framework/components/button/component.js
```diff
@@ -456,7 +456,7 @@ class ButtonComponent extends Component {
     }
 
     _applyTintImmediately(tintColor) {
-        if (!tintColor || !this._imageReference.hasComponent('element'))
+        if (!tintColor || !this._imageReference.hasComponent('element') || this._imageReference.entity.element.type != ELEMENTTYPE_IMAGE)
```
jpauloruschel, line 459: Sorry for the delay in reviewing this, and thanks for this suggestion! There's a small problem with this approach which is that this would break the current behaviour of allowing you to link a `text` element to a button. I think you can change it to: ` || this._imageReference.entity.element.type === ELEMENTTYPE_GROUP` so that we keep that behaviour.

rob5300: I wasn't aware that was possible, have made this change.

jpauloruschel (approved):

PR #4002, opened 2022-02-09
Improve ModelComponent typings

willeastcott:
Improve `ModelComponent` typings.

### DIFF
--- a/src/framework/components/model/component.js
+++ b/src/framework/components/model/component.js
@@ -12,6 +12,8 @@ import { Asset } from '../../../asset/asset.js';
 
 import { Component } from '../component.js';
 
+/** @typedef {import('../../../scene/composition/layer-composition.js').LayerComposition} LayerComposition */
+/** @typedef {import('../../../scene/materials/material.js').Material} Material */
 /** @typedef {import('../../../shape/bounding-box.js').BoundingBox} BoundingBox */
 /** @typedef {import('../../entity.js').Entity} Entity */
 /** @typedef {import('./system.js').ModelComponentSystem} ModelComponentSystem */
@@ -54,6 +56,7 @@ class ModelComponent extends Component {
      * @private
      */
     _castShadows = true;
+
     /**
      * @type {boolean}
      * @private
@@ -65,6 +68,7 @@ class ModelComponent extends Component {
      * @private
      */
     _materialAsset = null;
+
     /**
      * @type {Material}
      * @private
@@ -76,16 +80,19 @@ class ModelComponent extends Component {
      * @private
      */
     _castShadowsLightmap = true;
+
     /**
      * @type {boolean}
      * @private
      */
     _lightmapped = false;
+
     /**
      * @type {number}
      * @private
      */
     _lightmapSizeMultiplier = 1;
+
     /**
      * @type {boolean}
      * @private
@@ -97,6 +104,7 @@ class ModelComponent extends Component {
      * @private
      */
     _layers = [LAYERID_WORLD]; // assign to the default world layer
+
     /**
      * @type {number}
      * @private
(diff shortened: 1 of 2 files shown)


src/framework/components/model/component.js
```diff
@@ -24,65 +26,121 @@ import { Component } from '../component.js';
...
+     */
+    _lightmapped = false;
 
-        // #if _DEBUG
-        this._batchGroup = null;
-        // #endif
+    /**
+     * @type {number}
+     * @private
+     */
```
mvaligursky, line 93: I just wonder if these could be done on small number of lines? Perhaps
```
/** @type { number } */
/** @private */
```

src/framework/components/model/component.js
```diff
@@ -24,65 +26,121 @@ import { Component } from '../component.js';
...
+    _clonedModel = false;
+
+    // #if _DEBUG
+    _batchGroup = null;
+    // #endif
+
+    /**
+     * Create a new ModelComponent instance.
+     *
+     * @param {ModelComponentSystem} system - The ComponentSystem that created this Component.
```
mvaligursky, line 137: should constructor in these be public?

mvaligursky (approved):

src/framework/components/model/component.js
```diff
@@ -24,65 +26,121 @@ import { Component } from '../component.js';
...
 
-        this._layers = [LAYERID_WORLD]; // assign to the default world layer
-        this._batchGroupId = -1;
+    /* eslint-disable jsdoc/check-types */
+    /**
+     * @type {Object.<string, number>}
+     * @private
+     */
+    _mapping = {};
+    /* eslint-enable jsdoc/check-types */
```
kungfooman, line 52:
```suggestion
    /**
     * @type {Record<string, number>}
     * @private
     */
    _mapping = {};
```

It takes two little changes in `node_modules\@playcanvas\jsdoc-template\publish.js` to make this work without extra steps to ignore `eslint` and the resulting types would also look less verbose/strange:

```ts
    /**
     * A dictionary that holds material overrides for each mesh instance. Only applies to model
     * components of type 'asset'. The mapping contains pairs of mesh instance index - material
     * asset id.
     *
     * @type {Record<string, number>}
     */
    set mapping(arg: Record<string, number>);
    get mapping(): Record<string, number>;
```

vs.

```ts
    /**
     * A dictionary that holds material overrides for each mesh instance. Only applies to model
     * components of type 'asset'. The mapping contains pairs of mesh instance index - material
     * asset id.
     *
     * @type {Object.<string, number>}
     */
    set mapping(arg: {
        [x: string]: number;
    });
    get mapping(): {
        [x: string]: number;
    };
```

In `node_modules\@playcanvas\jsdoc-template\publish.js` you can replace this:

```js
    // Check for key-value (aka dictionary or index signature) types
    match = /^Object.<string,\s*(.*)>$/i.exec(name);
    if (match) {
        name = match[1];
```

with:

```js
    // Check for key-value (aka dictionary or index signature) types
    match = /^(Object|Record).<string,\s*(.*)>$/i.exec(name);
    if (match) {
        name = match[2]; // changed from [1] to [2] because extra group
```

src/framework/components/model/component.js
willeastcott: Unfortunately not. That's actually the minimum number of lines that will work for 2 tags. But the good news is that when we adopt private members/functions (`#`), it will collapse to a single line. I'm hoping we do that relatively soon.

src/framework/components/model/component.js
willeastcott: I was thinking about this already. We should add `@hideconstructor` for all components. But I think it's best to do that in a dedicated follow up PR, since this one is all about the `ModelComponent`.

src/framework/components/model/component.js
willeastcott: This is very interesting. However, there are actually several instances of this across the engine codebase now, so rather than just fix this one up here, I think I will open an issue on this and we can address is separately. Thanks for the tip!!

PR #4004, opened 2022-02-09
[FIX] RenderComponent#clone should respect enabled property

### LINKED ISSUE #3888: Clone() doesn't respect the disabling of a render component
If you clone an entity with a disabled render component, this component is enabled after cloning. Disabling the component by script after cloning is a temporary solution.

willeastcott:
Fixes #3888


mvaligursky (approved):
it's good, but ideally we address all components in a consistent way

PR #4010, opened 2022-02-10
Initial implementation of WebglGraphicsDevice class

mvaligursky:


slimbuck:
Awesome that you managed to keep history in tact!

With this change `GraphicsDevice` types & docs will go missing. What's the plan with that?

mvaligursky:
> With this change `GraphicsDevice` types & docs will go missing. What's the plan with that?

Good point, and I'm addressing this on a separate branch / PR already, which should be ready very soon.

slimbuck (approved):
Approving, but pls wait for @willeastcott to also approve before merging!

test/graphics/graphics-device.test.mjs
```diff
@@ -1,4 +1,4 @@
-import { GraphicsDevice } from '../../src/graphics/webgl/graphics-device.js';
```
willeastcott, line 1: The correct thing to do here is `git mv` this test into `test/graphics/webgl/webgl-graphic-device.mjs`.

willeastcott (approved):
Approving with comments. Note that I'm nervous about the next engine path release - this is a pretty fundamental change.

test/graphics/graphics-device.test.mjs
mvaligursky: nice one, done, thanks

PR #4009, opened 2022-02-10
Move GraphicsDevice to webgl subfolder

mvaligursky:
This PR moves GraphicsDevice class from /graphics to /graphics/webgl folder, and fixes import paths for this to work. This is a first step of #3993 where further refactoring is done on order to preserve the file move history, as that's quite important file to have a history for.

When this is merged, the rest of #3993 changes will be done in follow up PR.

PR #4013, opened 2022-02-10
GraphicsDevice properties refactor

mvaligursky:
This PR moves some properties which need to be public, as well as some shared code, from `WebglGraphicsDevice` back to `GraphicsDevice`. This is a follow up for #4009 and #4010.

**API Changes** - which are expected to have no impact to existing projects:

Following properties and functions are no longer exposed as public in GraphicsDevice class, as they have no use case:
```
updateBegin	
updateEnd	
clear	
copyRenderTarget
draw	

getBlending	
getDepthTest	
getDepthWrite	
getRenderTarget	
setBlendEquation	
setBlendEquationSeparate	
setBlendFunction	
setBlendFunctionSeparate	
setBlending	
setColorWrite	
setCullMode	
setDepthFunc	
setDepthTest	
setDepthWrite	
setIndexBuffer	
setRenderTarget	
setScissor	
setShader	
setStencilFunc	
setStencilFuncBack	
setStencilFuncFront	
setStencilOperation	
setStencilOperationBack	
setStencilOperationFront	
setStencilTest	
setVertexBuffer	
setViewport	
```

Additionally, `VertexFormat.update` has been deprecated, as well as `VertexFormat.elements` changed into readonly, as the engine does not support those being modified.

### DIFF
--- a/src/graphics/graphics-device.js
+++ b/src/graphics/graphics-device.js
@@ -1,4 +1,16 @@
 import { EventHandler } from '../core/event-handler.js';
+import { platform } from '../core/platform.js';
+
+import { ScopeSpace } from './scope-space.js';
+import { programlib } from './program-lib/program-lib.js';
+import { ProgramLibrary } from './program-library.js';
+
+import {
+    PRIMITIVE_POINTS, PRIMITIVE_TRIFAN
+} from './constants.js';
+import { Debug } from '../core/debug.js';
+
+const EVENT_RESIZE = 'resizecanvas';
 
 /**
  * The graphics device manages the underlying graphics context. It is responsible for submitting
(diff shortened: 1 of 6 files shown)


willeastcott (approved):

src/graphics/vertex-format.js
```diff
@@ -186,37 +190,23 @@ class VertexFormat {
...
-
-        VertexFormat._defaultInstancingFormat = new VertexFormat(graphicsDevice, formatDesc);
-    }
-
     /**
      * The {@link VertexFormat} used to store matrices of type {@link Mat4} for hardware instancing.
      *
      * @type {VertexFormat}
      */
     static get defaultInstancingFormat() {
```
willeastcott, line 198: Why is the `get` needed here, BTW? I know it was like that before, but was curious.

mvaligursky: I'm not sure I understand what you're asking? The get is there to make it a property getter and not a function, so it can be called like this: `pc.VertexFormat.defaultInstancingFormat`

src/graphics/graphics-device.js
```diff
@@ -9,10 +23,288 @@ import { EventHandler } from '../core/event-handler.js';
  * @augments EventHandler
  */
 class GraphicsDevice extends EventHandler {
+    /**
+     * The canvas DOM element that provides the underlying WebGL context used by the graphics device.
+     *
+     * @type {HTMLCanvasElement}
+     */
+    canvas;
```
slimbuck, line 31: I'm wondering how this code gets executed (and also converted to es5).

This line is declaring a member and could also be initialising it, like `canvas = null;`. But the constructor does `this.canvas = canvas;`. So does the code on this line run after the constructor's call to `super();`, but before the rest of the constructor code?

src/graphics/vertex-format.js
willeastcott: Oh right, of course. Duh. 😆

src/graphics/graphics-device.js
mvaligursky: correct, when you step over this in the debugger, those things execute first (and will be set to undefined), before constructor.

mvaligursky: this is what it looks like in ES5
Screenshot 2022-02-11 at 09 48 45

slimbuck: Interesting.

So could you do `canvas = canvas;` here, instead of doing it in the constructor?

In a class as large as this it probably makes little difference, but in general we don't want to be initialising members twice.

slimbuck: So I guess in general we want to declare (and optionally initialise) variables here or in the constructor, but not both?

mvaligursky: I doubt that would make any perf difference .. but perhaps it'd be useful to test it in a class such as Vec3?

slimbuck: I really don't like the idea instance members are initialised twice in our constructors, it's gross.

However reading [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#field_declarations), they say private versions of these (which we hope to use in future) must be declared this way:

```
Note: Private fields can only be declared up-front in a field declaration.
```

So it seems we ultimately won't have a choice and hopefully compiler will optimise away the redundant set.

slimbuck (approved):

Issue #3995, opened 2022-02-07
TypeScript declarations issues with v1.52.1 


pavle-goloskokovic:
TS defs of the current version of the engine are noticeably worse and less accurate than of the previous version.
Major concern is type inaccuracy and great number of `any` types used all over. Here are some concrete problems:

- `ModelComponent` `asset` property should accept `Asset` values.

``` 
/**
* The asset for the model (only applies to models of type 'asset') can also be an asset id.
*
* @type {Asset|number}
*/
set asset(arg: number);
```

- `ScriptType` methods are defined as properties, causing issues when extending and overriding methods:

```
TS2425: Class 'ScriptType' defines instance member property 'update', but extended class defines it as instance member function.
TS2425: Class 'ScriptType' defines instance member property 'initialize', but extended class defines it as instance member function.
```

defs:
```
update: (dt: number) => any;
initialize: () => any;
```
also notice return types `any` where they should be `void`.

- Classes have odd names and aliases:

```
type Asset$7 = Asset$6;
type AssetRegistry$a = AssetRegistry$c;
type Application$h = Application$7;
type ResourceHandler$g = ResourceHandler$h;

declare class Material$2
...
```

- `Material` class `clone()` method is declared to return `Material`, causing type mismatch when called on subclass such as `StandardMaterial`. This was not previously an issue.

- Most private properties are not declared as private, e.g. `Entity` class:

```
readonly sprite: SpriteComponent;
_batchHandle: any;
c: {};
_app: Application$7;
_guid: string;
_destroying: boolean;
_template: boolean;
```

- `ContainerHandler` methods' attributes are all defined as `any`, previously defined with correct types.

```
_getUrlWithoutParams(url: any): any;
_getParser(url: any): any;
load(url: any, callback: any, asset: any): void;
open(url: any, data: any, asset: any): any;
patch(asset: any, assets: any): void;
```

- Callbacks' types are not exported:

```
type resourceHandlerCallback$1 = (err: string | null, response?: any) => any;
```

- `ResourceHandler` `patch()` method is not declared as optional.

- `TS2339: Property 'stop' does not exist on type 'SoundInstance$1'.`

willeastcott:
This is great. Thanks for compiling this list @pavle-goloskokovic. I'll be taking a look at fixing all this sometime during this week.

willeastcott:
A little update:

* The `ModelComponent` types have been improved: https://github.com/playcanvas/engine/pull/4002
[...]

PR #4016, opened 2022-02-12
Material#clone: return polymorphic this type

kungfooman:
Fixes one issue in https://github.com/playcanvas/engine/issues/3995

Before you had to write:

```js
const a = new StandardMaterial;
const b = a.clone() as StandardMaterial;
```

Now its just:

```js
const a = new StandardMaterial;
const b = a.clone();
```

And the type is correctly inferred:

image

Because the type system can infer the correct type from `this`: https://www.typescriptlang.org/docs/handbook/advanced-types.html#polymorphic-this-types

JSDoc doesn't support `this` as type yet, but I made a PR for that too: https://github.com/playcanvas/playcanvas-jsdoc-template/pull/38

### DIFF
--- a/src/scene/materials/material.js
+++ b/src/scene/materials/material.js
@@ -338,7 +338,7 @@ class Material {
     /**
      * Clone a material.
      *
-     * @returns {Material} A newly cloned material.
+     * @returns {this} A newly cloned material.
      */
     clone() {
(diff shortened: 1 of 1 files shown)


willeastcott (approved):
Great, thanks! 🙏

PR #4019, opened 2022-02-14
More IntelliSense improvements

willeastcott:
Further improve IntelliSense support in the codebase.

* Make more use of public class fields.
* Add missing JSDoc comments to some functions.
* Completely remove `Entity#_batchHandle` which has been unused in the codebase for 10 years.

### DIFF
--- a/src/framework/components/rigid-body/system.js
+++ b/src/framework/components/rigid-body/system.js
@@ -155,12 +155,53 @@ const _schema = ['enabled'];
  * valid if 3D Physics is enabled in your application. You can enable this in the application
  * settings for your project.
  *
- * @property {Vec3} gravity The world space vector representing global gravity in the physics
- * simulation. Defaults to [0, -9.81, 0] which is an approximation of the gravitational force on
- * Earth.
  * @augments ComponentSystem
  */
 class RigidBodyComponentSystem extends ComponentSystem {
+    /**
+     * @type {number}
+     * @ignore
+     */
+    maxSubSteps = 10;
+
+    /**
+     * @type {number}
+     * @ignore
+     */
+    fixedTimeStep = 1 / 60;
+
+    /**
+     * The world space vector representing global gravity in the physics simulation. Defaults to
+     * [0, -9.81, 0] which is an approximation of the gravitational force on Earth.
+     *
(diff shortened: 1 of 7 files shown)


src/vr/vr-manager.js
```diff
@@ -25,14 +41,12 @@ class VrManager extends EventHandler {
...
 
-        this.isSupported = VrManager.isSupported;
-
         this._index = {};
-        this.displays = [];
-        this.display = null; // primary display (usually the first in list)
 
         this._app = app;
 
+        this.isSupported = VrManager.isSupported;
```
kungfooman, line 48: Why not initialize it over the constructor aswell? Would save this extra line

src/framework/components/rigid-body/system.js
```diff
@@ -181,16 +222,6 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
-        this.gravity = new Vec3(0, -9.81, 0);
-
-        // Arrays of pc.RigidBodyComponents filtered on body type
-        this._dynamic = [];
-        this._kinematic = [];
-        this._triggers = [];
-        this._compounds = [];
-
         this.collisions = {};
         this.frameCollisions = {};
```
kungfooman, line 226: Should these two objects also move up or wait for Record type?

mvaligursky (approved):

src/framework/components/rigid-body/system.js
willeastcott: They could go up, yes, but the value is another object type and since it's an internal property, I thought I'd leave it for now.

src/vr/vr-manager.js
willeastcott: Because ESLint emits an 'use before defined' error if you do that.

kungfooman: Oh, thank you, I didn't realize its a static and normal property of the same class at the same time.

willeastcott: Yeah, I have no idea why it was written like that originally. The WebVR stuff is deprecated now and could be deleted from the codebase at some point, so I don't want to mess with this code too much.

PR #4025, opened 2022-02-15
Add app to entity on SceneParser.parse

### LINKED ISSUE #4024: Single template instantiating - multiple apps issue
### Description

In general, the issue is that when a template instantiated entity (scriptType, etc) is attached to the first created app

1. Create app 1, template handler will create new Template with attached app 1 https://github.com/playcanvas/engine/blob/bb5aa65caa880ab546ce9de456ef5bc195533f74/src/resources/template.js#L34
2. Instantiate template. All okay, entity attached to app 1, as needed
3. Create app 2, instantiate the same template. Entity attached to app 1. This is the first issue.

As a temp solution we are using this override in our project:
```
pc.Template.prototype.instantiate = function instantiate(app) {
    if (app) {
        this._app = app;
    }

    this._parseTemplate();

    return this._templateRoot.clone();
};
```

It would be better if, when creating templates, it was attached to the app from which they are created but without the need to pass the app argument.

4. Destroy app 2, try to create a template in app 1 again
5. Error appears: Couldn't find current application. Because when we instantiate a template, SceneParser.parse uses global [currentApplication ](https://github.com/playcanvas/engine/blob/dev/src/framework/globals.js)property. Which is [set on application create](https://github.com/playcanvas/engine/blob/e5cdb3134334d173c69a423af060334c7fbbef16/src/framework/application.js#L225).

Sarletor:
Fixes #4024

### DIFF
--- a/src/resources/parser/scene.js
+++ b/src/resources/parser/scene.js
@@ -49,7 +49,7 @@ class SceneParser {
     }
 
     _createEntity(data, compressed) {
-        const entity = new Entity();
+        const entity = new Entity(this._app);
 
         entity.name = data.name;
         entity.setGuid(data.resource_id);


src/resources/parser/scene.js
```diff
@@ -49,7 +49,7 @@ class SceneParser {
     }
 
     _createEntity(data, compressed) {
-        const entity = new Entity();
+        const entity = new Entity(this._app);
```
mvaligursky, line 52: ideally we call constructor correctly, which is `    constructor(name, app)`
the Entity constructor class handles app first, but parent class GraphNode does not seem to and would attach application as a name I think.

Sarletor: Fixed

PR #4028, opened 2022-02-16
DeviceResourceCache to store per device shared resources

mvaligursky:


src/scene/particle-system/particle-emitter.js
```diff
@@ -382,12 +381,11 @@ class ParticleEmitter {
         this.rebuild();
     }
 
-    static DEFAULT_PARAM_TEXTURE = null;
-
-    static staticInit(device) {
+    get defaultParamTexture() {
+        Debug.assert(this.graphicsDevice);
+        const deviceResoruceCache = DeviceResourceCache.get(this.graphicsDevice);
+        if (!deviceResoruceCache.particleEmitterDefaultParamTexture) {
```
kungfooman, line 387:
```suggestion
        const deviceResourceCache = DeviceResourceCache.get(this.graphicsDevice);
        if (!deviceResourceCache.particleEmitterDefaultParamTexture) {
```

src/scene/particle-system/particle-emitter.js
```diff
@@ -403,18 +401,13 @@ class ParticleEmitter {
                     dtex[p * 4 + 3] = c;
                 }
             }
-            ParticleEmitter.DEFAULT_PARAM_TEXTURE = _createTexture(device, resolution, resolution, dtex, PIXELFORMAT_R8_G8_B8_A8, 1.0, true);
-            ParticleEmitter.DEFAULT_PARAM_TEXTURE.minFilter = FILTER_LINEAR;
-            ParticleEmitter.DEFAULT_PARAM_TEXTURE.magFilter = FILTER_LINEAR;
+            const texture = _createTexture(this.graphicsDevice, resolution, resolution, dtex, PIXELFORMAT_R8_G8_B8_A8, 1.0, true);
+            texture.minFilter = FILTER_LINEAR;
+            texture.magFilter = FILTER_LINEAR;
+            deviceResoruceCache.particleEmitterDefaultParamTexture = texture;
```
kungfooman, line 407:
```suggestion
            deviceResourceCache.particleEmitterDefaultParamTexture = texture;
```

src/scene/particle-system/particle-emitter.js
```diff
@@ -403,18 +401,13 @@ class ParticleEmitter {
...
         }
-    }
 
-    static staticDestroy() {
-
-        if (ParticleEmitter.DEFAULT_PARAM_TEXTURE) {
-            ParticleEmitter.DEFAULT_PARAM_TEXTURE.destroy();
-            ParticleEmitter.DEFAULT_PARAM_TEXTURE = null;
-        }
+        return deviceResoruceCache.particleEmitterDefaultParamTexture;
```
kungfooman, line 410:
```suggestion
        return deviceResourceCache.particleEmitterDefaultParamTexture;
```

src/graphics/reproject-texture.js
```diff
@@ -332,72 +339,46 @@ class SimpleCache {
...
+    if (!cache.reprojectTextureCache) {
+        cache.reprojectTextureCache = new SimpleCache();
     }
-
-    clear() {
-        this.cache.clear();
+    if (!cache.reprojectTextureCache.map.get(key)) {
+        const texture = createSamplesTex(device, key, samplesCache.get(key, () => {
+            return getSamplesFnc();
+        }));
```
kungfooman, line 358: Looks like an empty lambda to me that is just adding overhead/code:

```suggestion
        const texture = createSamplesTex(device, key, samplesCache.get(key, getSamplesFnc));
```

src/graphics/device-resource-cache.js
```diff
@@ -0,0 +1,87 @@
...
+
+        if (this.postEffectQuadVB) {
+            this.postEffectQuadVB.destroy();
+            this.postEffectQuadVB = null;
+        }
+
+        if (this.reprojectTextureCache) {
+            this.reprojectTextureCache.destroy();
+            this.reprojectTextureCache = null;
+        }
```
kungfooman, line 59: Not a huge change, just saves some potential lines using `?`:
```suggestion
            this.defaultMaterial?.destroy();
            this.defaultMaterial = null;
            
            this.particleEmitterDefaultParamTexture?.destroy();
            this.particleEmitterDefaultParamTexture = null;
            
            this.postEffectQuadVB?.destroy();
            this.postEffectQuadVB = null;
            
            this.reprojectTextureCache?.destroy();
            this.reprojectTextureCache = null;
```

src/graphics/reproject-texture.js
```diff
@@ -332,72 +339,46 @@ class SimpleCache {
...
-            const cache = new SimpleCache();
-            device.on('destroy', () => {
-                cache.map.forEach((value, key) => {
-                    value.destroy();
-                });
-                this.cache.map.delete(device);
-            });
-            return cache;
-        }).get(key, missFunc);
+const getCachedTexture = (device, key, getSamplesFnc) => {
```
kungfooman, line 350: This function accesses `cache.reprojectTextureCache.map` three times (internals of `SimpleCache`), it should probably be a method of `SimpleCache` itself then, which will also shorten the code while using less object property lookups

mvaligursky: done, thanks!

src/graphics/reproject-texture.js
mvaligursky: done too, thanks!

src/graphics/simple-post-effect.js
```diff
@@ -10,14 +11,36 @@ import { DebugGraphics } from './debug-graphics.js';
...
 // Draws shaded full-screen quad in a single call
-let _postEffectQuadVB = null;
 const _postEffectQuadDraw = {
     type: PRIMITIVE_TRISTRIP,
     base: 0,
     count: 4,
     indexed: false
 };
 
+// device cache storing a quad vertex buffer
```
willeastcott, line 21: Convert to JSDoc block with an `@ignore`.

src/scene/particle-system/particle-emitter.js
```diff
@@ -213,6 +214,37 @@ function divGraphFrom2Curves(curve1, curve2, outUMax) {
     return sub;
 }
 
+// device cache storing a texture, used as a default texture parameter
+class ParticleEmitterDeviceCache extends DeviceCache {
+    /** @type { Texture } */
```
willeastcott, line 219: Rework as described in previous comment.

src/graphics/simple-post-effect.js
```diff
@@ -10,14 +11,36 @@ import { DebugGraphics } from './debug-graphics.js';
...
+class PostEffectDeviceCache extends DeviceCache {
+    /** @type { VertexBuffer } */
```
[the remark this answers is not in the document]
mvaligursky: removed the incorrect line with the type, thanks!

src/graphics/device-cache.js
```diff
@@ -0,0 +1,50 @@
+/** @typedef {import('./graphics-device.js').GraphicsDevice} GraphicsDevice */
+
+/**
+ * A cache storing shared resources associated with a device. The resources are removed
+ * from the cache when the device is destroyed.
+ *
+ * @ignore
+ */
+class DeviceCache {
```
slimbuck, line 9: Sorry I'm not crazy about this structure. Took me a while to figure out that `DeviceCache` instances contain a `data` member -and- a `map` of device -> 'DeviceCache' instance. So in some cases the map is used and data unused, in other cases visa versa.

I'd much prefer two interfaces: one for the device cache containing a map of devices and another for the thing that the device cache map stores.

slimbuck: In fact, you could remove the data member and just store the entry directly.

mvaligursky: great suggestion, and I've done just that, thanks!

src/framework/application.js
```diff
@@ -750,7 +748,7 @@ class Application extends EventHandler {
         const material = new StandardMaterial();
         material.name = "Default Material";
         material.shadingModel = SPECULAR_BLINN;
-        DefaultMaterial.add(this.graphicsDevice, material);
+        setDefaultMaterial(this.graphicsDevice, material);
```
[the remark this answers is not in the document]
mvaligursky: it's done this way to avoid circular dependency few months ago, so I don't want to add it back in again.

src/graphics/reproject-texture.js
```diff
@@ -324,6 +325,12 @@ const createSamplesTex = (device, name, samples) => {
 class SimpleCache {
     map = new Map();
 
+    destroy() {
+        this.map.forEach((value, key) => {
+            value.destroy();
```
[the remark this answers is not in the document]
mvaligursky: done. I thought about it before, but as the cache does not get destroyed I left it as it was .. but it's definitely cleaner this way, thanks

src/framework/application.js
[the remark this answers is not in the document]
slimbuck: Oh I see, `Material` itself accesses `getDefaultMaterial()`. I think that's probably the dependency that should be broken.

Instead of material setting the default material on `MeshInstance` itself (in `Material.destroy`), Material can call some function on `MeshInstance` to do the work, like say `onMaterialDestroyed()` or similar.

This would mean `MeshInstance` has dependency on `getDefaultMaterial()` instead of `Material`. Then `getDefaultMaterial()` can implement the cache miss handler and `Application` can be freed from default materials _and_ we can get rid of `setDefaultMaterial()` entirely.

src/graphics/device-cache.js
```diff
@@ -0,0 +1,39 @@
...
+    get(device, onCreate) {
+
+        if (!this._cache.has(device)) {
+            this._cache.set(device, onCreate());
+
+            // when the device is destroyed, destroy and remove its entry
+            device.on('destroy', () => {
+                this._cache.get(device)?.destroy();
+                this._cache.delete(device);
+            });
```
slimbuck, line 32: Agree it's not ideal to access _cache twice, but it's unnecessary:

```suggestion
            const result = onCreate();
            this._cache.set(device, result);

            // when the device is destroyed, destroy and remove its entry
            device.on('destroy', () => {
                this._cache.get(device)?.destroy();
                this._cache.delete(device);
            });
            return result;
```

src/framework/application.js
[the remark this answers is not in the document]
slimbuck: I think this is probably the right PR to restructure the material dependencies, but also happy to do if after if you feel strongly about it.

src/graphics/device-cache.js
mvaligursky: I'm not talking about this once off case .. I'm taking about the normal case where cache contains the value already.
first we need to do 'cache.has'
and then do 'cache.get'

before we only did cache.get and compared to null .. (which limited storage of null)

src/framework/application.js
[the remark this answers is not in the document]
mvaligursky: I'd prefer separate PR. This PR is focused on creating a device cache for the few use cases, not refactoring material dependencies.

src/graphics/device-cache.js
slimbuck: yes true. if perf were an issue you could rather do:
```
const value = this._cache.get(device);
if (!value && !this._cache.has(device)) {
    this._cache.set(blah);
    etc
}
return value;

slimbuck (approved):

Issue #2465, opened 2020-10-06
Masking of 3D scroll views doesn't work on Mobile VR


yaustar:
https://playcanvas.com/project/725268/overview/scrollview-in-vr-bug

On desktop browser
![image](https://user-images.githubusercontent.com/16639049/95181591-66ce3180-07bb-11eb-83c6-a20b5727323e.png)

On Mobile VR
![Screenshot_20201006-100152_Chrome](https://user-images.githubusercontent.com/16639049/95182553-b4976980-07bc-11eb-9085-edb92855931e.jpg)

Maksims:
And how it looks not in VR?

yaustar:
It looks like the desktop browser version

mvaligursky:
likely related https://github.com/playcanvas/engine/issues/3917

Issue #3031, opened 2021-03-29
WebXR AR experience doesn't look like it depth tests


yaustar:
Android Chrome

Test project: https://playcanvas.com/editor/scene/1120517 (when in AR, look for a floor, dinosaurs should appear as you move around the room)

<img width="1293" alt="Screenshot 2021-03-29 at 09 32 52" src="https://user-images.githubusercontent.com/16639049/112809875-692a8d80-9072-11eb-84e0-52ee87051fa8.png">

https://user-images.githubusercontent.com/16639049/112809993-85c6c580-9072-11eb-85d9-36d0085c122e.mp4

The feet/legs can be seen through the body and the circle should be below the dinosaur too.

Sometimes the issue goes away after moving around the room or looking from certain angles.

The dinosaur models can be looked at separately in this scene https://playcanvas.com/editor/scene/1120526

yaustar:
Also tested against the earliest version that still works with this AR experience: https://t.co/YUcFNikNpB?amp=1

mvaligursky:
@Maksims - any idea about this one?

Maksims:
> @Maksims - any idea about this one?

Could this be related?
https://github.com/playcanvas/engine/blob/master/src/framework/application.js#L2302

XRSession provides its own framebuffer, so rendering in XR session happens not into backbuffer, but a provided framebuffer. I actually not sure why it is happening.

mvaligursky:
Just looking at the documentation and code, wondering if this is related:

- @Maksims pointed out we use frame.session.renderState.baseLayer.framebuffer as a framebuffer
- the BaseLayer api https://developer.mozilla.org/en-US/docs/Web/API/XRWebGLLayer/XRWebGLLayer specifies that we can pass depth property, which we don't but it defaults to true anyways .. so perhaps not needed?
- this bit here https://developer.mozilla.org/en-US/docs/Web/API/XRWebGLLayer/framebuffer talks about opaque framebuffer, but mentions: _Opaque framebuffers initialized with the depth property set to false will not have a depth buffer and will rely on the coordinates alone to determine distance._ As we don't sppecify false, this should be ok? It seems like the behaviour seems consistent with depth being false though.

mvaligursky:
possibly related as well (missing stencil buffer on the Quest)
https://forum.playcanvas.com/t/stencil-buffer-on-oculus-quest/21471/3

mvaligursky:
[...]

PR #4038, opened 2022-02-21
XR session requests stencil buffer

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/2465

When XR session starts, a stencil buffer is requested (which defaults to false).
Also requesting depth and alpha, even though those default to true https://developer.mozilla.org/en-US/docs/Web/API/XRWebGLLayer/XRWebGLLayer, but based on https://github.com/playcanvas/engine/issues/3031 it seems perhaps depth buffer is not getting created by default on all platforms.


willeastcott (approved):
Nice fix. 👍

PR #4039, opened 2022-02-22
instantiateRenderEntity sets up render assets on render components

mvaligursky:
- When an Entity hierarchy was created using GlbContainer.instantiateRenderEntity, the render component didn't have .asset property set up. This is now fixed, and those properties point to a render assets.
- Additionally, a Render class is now public, as it is a type of the resource accessible using render component's asset property.

### DIFF
--- a/src/resources/parser/glb-container-resource.js
+++ b/src/resources/parser/glb-container-resource.js
@@ -99,7 +99,7 @@ class GlbContainerResource {
         };
 
         // helper function to recursively clone a hierarchy of GraphNodes to Entities
-        const cloneHierarchy = function (root, node, glb) {
+        const cloneHierarchy = (root, node, glb) => {
 
             const entity = new Entity();
             node._cloneInternal(entity);
@@ -109,6 +109,7 @@ class GlbContainerResource {
 
             // find all components needed for this node
             let attachedMi = null;
+            let renderAssset = null;
             for (let i = 0; i < glb.nodes.length; i++) {
                 const glbNode = glb.nodes[i];
                 if (glbNode === node) {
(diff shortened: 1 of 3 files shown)


src/resources/parser/glb-container-resource.js
```diff
@@ -109,6 +109,7 @@ class GlbContainerResource {
 
             // find all components needed for this node
             let attachedMi = null;
+            let renderAssset = null;
```
slimbuck, line 112: renderAsset

mvaligursky: hah :)

willeastcott (approved):

willeastcott:
Wait, what's the type of the resource of a render asset? e.g. `entity.render.asset`?

mvaligursky:
> Wait, what's the type of the resource of a render asset? e.g. `entity.render.asset`?

the code I added assigns Asset type to it created here:
https://github.com/playcanvas/engine/blob/d5bef0eed6b3d1f1c8970b6d33e089139d39773e/src/resources/parser/glb-container-resource.js#L58-L65

But internally this is stored as an id on the component, and render.asset returns asset.id

willeastcott:
I meant, is the asset's resource a `Render`? Just wondering, because that type is not public API.

mvaligursky:
> I meant, is the asset's resource a `Render`? Just wondering, because that type is not public API.

Fair point. Making the Render class public.

willeastcott:
Are developers expected to create `Render` objects themselves? Or do anything with them? Is the `Render` API in 'final' form?

mvaligursky:
I'm hiding the constructor as well .. as there's not a big use case for its public use.
Otherwise only meshes properties is exposed .. and I think that's useful to have.

slimbuck:
Do we need to make `Render` class public API?

Looks like instances of `Render` are created by the glb-parser at load time and then used by the render component internally (via `asset.resource`).

Even though the `Render` class seems simple an innocuous enough,  it's not actually possible to set a `Render` instance directly on the `RenderComponent` anyway. So seems like we can leave `Render private?

willeastcott:
I'd probably agree with that assessment. Making `Render` public is not necessary to solve the original motivation for this PR.

slimbuck (approved):

PR #4040, opened 2022-02-22
AnimController activate state progress fix

### LINKED ISSUE #3837: Anim transition Exit Time is being affected by source state speed
Exit Time is described as "a normalized value of the source states duration". This means it should refer to the same frame regardless of source animation speed but that's not the case. Using speed = 2 and Exit Time = 1 will result in animation being frozen on the last frame for half the duration.

ellthompson:
When animating properties with a graph, the activate state's progress should be affected by the active state's speed property.

Fixes #3837

### DIFF
--- a/src/anim/controller/anim-controller.js
+++ b/src/anim/controller/anim-controller.js
@@ -187,7 +187,7 @@ class AnimController {
 
         const activeClip = this._animEvaluator.findClip(this.activeStateAnimations[0].name);
         if (activeClip) {
-            return time / activeClip.track.duration;
+            return time / activeClip.track.duration * this.activeState.speed;
         }
 
         return null;


src/anim/controller/anim-controller.js
```diff
@@ -187,7 +187,7 @@ class AnimController {
 
         const activeClip = this._animEvaluator.findClip(this.activeStateAnimations[0].name);
         if (activeClip) {
-            return time / activeClip.track.duration;
+            return time / activeClip.track.duration * this.activeState.speed;
```
yaustar, line 190: Potential divide by 0 error. 

Also, what happens when (if) speed is negative?

mvaligursky, line 190: do we need to handle speed of 0?

yaustar: Shouldn't the duration be longer the slower the speed is? ie, should this be:

```
return time / (activeClip.track.duration / this.activeState.speed);
```

PR #4045, opened 2022-02-23
Recompile basis transcoder 1.16.2 with emsdk 3.1.5

slimbuck:
This PR updates the basis wasm module with the latest version and compiled with latest emscripten compiler.

Sizes are slightly smaller by estimate 5%.


willeastcott:
GZIPped size comparison for `basis.wasm.wasm`:

* New: 268KB
* Old: 288KB

This is a 6.68435% decrease. Really great!

willeastcott (approved):

bkotsopoulossc:
What compiler / linker flags? If size is important you might be able to do better without sacrificing too much perf with -Oz

slimbuck:
Hi @bkotsopoulossc,

We use a slightly modified cmake file [here](https://github.com/playcanvas/basis_universal/compare/master...pc-master). We're using -O3, but could definitely try -Oz. Thanks!

PR #4047, opened 2022-02-24
[Fix] Removed children are no longer marked as enabled in hierarchy

### LINKED ISSUE #2464: Adding an entity to a unparented entity will cause the child to render
Project: https://playcanvas.com/editor/scene/1002949 (press 1)

Repo steps:
1. Unparent an Entity
2. Clone an Entity
3. Add the clone to the unparented Entity
4. The clone will render

Related issues: https://github.com/playcanvas/engine/issues/2183

mvaligursky:
Fixes #2464

When a child GraphNode is removed from the parent, the removed hierarchy nodes have _enabledInHierarchy set to false. Without this, when any node is added to this removed hierarchy, it's believed to be active.

### DIFF
--- a/src/scene/graph-node.js
+++ b/src/scene/graph-node.js
@@ -1304,6 +1304,12 @@ class GraphNode extends EventHandler {
         // Clear parent
         child._parent = null;
 
+        // notify the child hierarchy it has been removed from the parent,
+        // which marks them as not enabled in hierarchy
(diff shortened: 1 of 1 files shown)


willeastcott (approved):
Slightly scared about weird edge cases/side effects here, but in principle, seems good.

PR #4048, opened 2022-02-24
Add application bake settings

slimbuck:
Add support for bake parameters to the application render settings.

### DIFF
--- a/src/scene/scene.js
+++ b/src/scene/scene.js
@@ -623,7 +623,7 @@ class Scene extends EventHandler {
         const apply = (name) => {
             if (render.hasOwnProperty(name)) {
                 this[name] = render[name];
-            };
+            }
         };
 
         // bake
--- a/src/scene/scene.js
+++ b/src/scene/scene.js
@@ -596,25 +596,41 @@ class Scene extends EventHandler {
     }
 
     applySettings(settings) {
-        // settings
-        this._gravity.set(settings.physics.gravity[0], settings.physics.gravity[1], settings.physics.gravity[2]);
-        this.ambientLight.set(settings.render.global_ambient[0], settings.render.global_ambient[1], settings.render.global_ambient[2]);
-        this._fog = settings.render.fog;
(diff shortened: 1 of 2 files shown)


src/scene/scene.js
```diff
@@ -596,27 +596,43 @@ class Scene extends EventHandler {
...
+        this.lightmapMode = render.lightmapMode;
+        this.exposure = render.exposure;
+        this._skyboxIntensity = render.skyboxIntensity === undefined ? 1 : render.skyboxIntensity;
+        this._skyboxMip = render.skyboxMip === undefined ? 0 : render.skyboxMip;
+
+        if (render.skyboxRotation) {
+            this._skyboxRotation.setFromEulerAngles(render.skyboxRotation[0], render.skyboxRotation[1], render.skyboxRotation[2]);
         }
 
+        const apply = (name) => {
```
mvaligursky, line 623: maybe call this `applyRender`? And apply to the rest of this function.

src/scene/scene.js
```diff
@@ -596,27 +596,43 @@ class Scene extends EventHandler {
...
+            }
+        };
+
+        // bake
+        apply('ambientBake');
+        apply('ambientBakeNumSamples');
+        apply('ambientBakeSpherePart');
+        apply('ambientBakeOcclusionBrightness');
+        apply('ambientBakeOcclusionContrast');
+
```
willeastcott, line 635: Not that you have to address this now, but I just want to point out that I'm not keen on `applySettings` being on the `Scene` class. This is really just a utility function for when serialized scene settings are loaded in. So I'd put this wherever that happens. What do other people think?

slimbuck: I agree. This is actually a detail of the application (whence this function is called), not the scene.

src/scene/scene.js
slimbuck: I thought about that, but actually only `skyboxIntensity` and `skyboxMip` work the same way, so figured I'd leave it.

PR #4049, opened 2022-02-24
Small refactor to move temp array to where its used from the Application

mvaligursky:
the list is only used in one function of the entity, so moved it there. It’s not used in the Editor either.

### DIFF
--- a/src/framework/entity.js
+++ b/src/framework/entity.js
@@ -27,6 +27,9 @@ import { Application } from './application.js';
 /** @typedef {import('./components/sound/component.js').SoundComponent} SoundComponent */
 /** @typedef {import('./components/sprite/component.js').SpriteComponent} SpriteComponent */
 
+/** @type {GraphNode[]} */
+const _enableList = [];
+
 /**
  * The Entity is the core primitive of a PlayCanvas game. Generally speaking an object in your game
  * will consist of an {@link Entity}, and a set of {@link Component}s which are managed by their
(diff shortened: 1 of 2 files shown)


src/framework/entity.js
```diff
@@ -27,6 +27,9 @@ import { Application } from './application.js';
 /** @typedef {import('./components/sound/component.js').SoundComponent} SoundComponent */
 /** @typedef {import('./components/sprite/component.js').SpriteComponent} SpriteComponent */
 
+/** @type {GraphNode[]} */
```
willeastcott, line 30: You'll need an @ignore here or this symbol will actually appear in the API reference (in the `pc.` namespace).

src/framework/entity.js
```diff
@@ -465,11 +468,11 @@ class Entity extends GraphNode {
 
         if (enableFirst) {
             // do not cache the length here, as enableList may be added to during loop
-            for (let i = 0; i < this._app._enableList.length; i++) {
-                this._app._enableList[i]._onHierarchyStatePostChanged();
+            for (let i = 0; i < _enableList.length; i++) {
+                _enableList[i]._onHierarchyStatePostChanged();
```
willeastcott, line 472: Presumably this won't work in the multi-application case?

mvaligursky: I thought about it, and I think it should? This is only used inside a single function, those apps don't run in parallel, but in serial.

src/framework/entity.js
mvaligursky: done. I assumed those are exposed in docs only if there's a description, not only a type.

willeastcott (approved):

willeastcott:
I love little refactors like this. Nice job.

PR #4052, opened 2022-02-25
Deprecate DepthMaterial

### LINKED ISSUE #3301: Deprecate DepthMaterial as it's not needed anymore
- the file should be either deprecated or even removed
https://github.com/playcanvas/engine/blob/master/src/scene/materials/depth-material.js

mvaligursky:
Fixes #3301, deprecates DepthMaterial which has not been needed for a long time.


src/deprecated/deprecated.js
```diff
@@ -465,6 +464,19 @@ Object.defineProperties(Texture.prototype, {
...
 // SCENE
 
+class DepthMaterial extends Material {
+    updateShader(device) {
+        Debug.deprecated("pc.DepthMaterial is deprecated.");
+        const options = {
+            skin: !!this.meshInstances[0].skinInstance
+        };
+        const library = device.getProgramLibrary();
+        this.shader = library.getProgram('depth', options);
```
willeastcott, line 474: Does this even work? Where is the 'depth' program registered?

willeastcott: image

mvaligursky: you're right, none of that exists anymore. I'll delete DepthMaterial completely.

willeastcott (approved):

PR #4054, opened 2022-02-25
[Fix] Entity.enable=true does not activate enties not connected to scene hierarchy

### LINKED ISSUE #2183: Cloning a disabled entity, then enabling the clone renders without being added to scene graph
Project: https://playcanvas.com/project/696174/overview/cloned-entities-with-no-parents

Cloning a disabled entity and enabling the clone renders the cloned entity with it having a parent

mvaligursky:
Fixes #2183

When the entity.enabled is set to true, only enable the subtree if the parent is enabled (which also means enabled in hierarchy)

### DIFF
--- a/src/scene/graph-node.js
+++ b/src/scene/graph-node.js
@@ -260,8 +260,11 @@ class GraphNode extends EventHandler {
         if (this._enabled !== enabled) {
             this._enabled = enabled;
 
-            if (this._parent && this._parent.enabled)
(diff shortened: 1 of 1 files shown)


willeastcott:
Just wanted to check you still intend this to be marked as a draft PR.

mvaligursky:
> Just wanted to check you still intend this to be marked as a draft PR.

Not anymore ;) Cheers!

slimbuck (approved):
needs thorough testing!

PR #4057, opened 2022-02-26
anim-controller: fix transitions sorting / minimize code duplication

kungfooman:
A sort function doesn't act on true/false, so this never sorted anything (copy/paste test for F12/DevTools):

```js
transitions = [
  {priority: 10},
  {priority: 50},
  {priority: 1},
  {priority: 2},
  {priority: 100},
  {priority: 3},
];
transitions.sort(function (a, b) {
    return a.priority < b.priority;
});
console.log(transitions.map(_ => _.priority));
```

Result: `[10, 50, 1, 2, 100, 3]` (its fixable by replacing `<` with `-`)

And then priority sorting is used in multiple places, so this PR reduces the code size/repetition of the same sorting function by just reusing functions (imo saving time and hopefully less bugs)

### DIFF
--- a/src/core/sort.js
+++ b/src/core/sort.js
@@ -9,7 +9,7 @@ export const cmpPriority = (a, b) => a.priority - b.priority;
 /**
  * @param {Array<{priority: number}>} arr - Array to be sorted in place where each element contains
  * an object with at least a priority property.
- * @returns {Array<{priority: number}>} In place sorted array
+ * @returns {Array<{priority: number}>} In place sorted array.
  * @ignore
  */
 export const sortPriority = arr => arr.sort(cmpPriority);
--- a/src/core/sort.js
+++ b/src/core/sort.js
@@ -2,11 +2,13 @@
  * @param {{priority: number}} a - First object with priority property.
  * @param {{priority: number}} b - Second object with priority property.
  * @returns {number} A number indicating the relative position.
+ * @ignore
  */
 export const cmpPriority = (a, b) => a.priority - b.priority;
 
 /**
  * @param {Array<{priority: number}>} arr - Array to be sorted in place where each element contains an object with at least a priority property.
  * @returns {Array<{priority: number}>} In place sorted array
+ * @ignore
  */
 export const sortPriority = arr => arr.sort(cmpPriority);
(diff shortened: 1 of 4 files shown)


src/core/sort.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * @param {{priority: number}} a
+ * @param {{priority: number}} b
+ */
+
```
willeastcott, line 5: I think you need to remove this blank line or VS Code can't associate the code with the JSDoc block.

src/core/sort.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * @param {{priority: number}} a - First object with priority property.
+ * @param {{priority: number}} b - Second object with priority property.
+ * @returns {number} A number indicating the relative position.
+ */
```
willeastcott, line 5: To prevent this from appearing in the API reference manual, you'll need to add `@ignore`.

src/core/sort.js
```diff
@@ -0,0 +1,12 @@
...
+export const cmpPriority = (a, b) => a.priority - b.priority;
+
+/**
+ * @param {Array<{priority: number}>} arr - Array to be sorted in place where each element contains an object with at least a priority property.
+ * @returns {Array<{priority: number}>} In place sorted array
+ */
```
willeastcott, line 11: And another `@ignore`.

src/core/sort.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * @param {{priority: number}} a - First object with priority property.
+ * @param {{priority: number}} b - Second object with priority property.
+ * @returns {number} A number indicating the relative position.
+ * @ignore
+ */
+export const cmpPriority = (a, b) => a.priority - b.priority;
```
willeastcott, line 7: Why does `cmpPriority` need to be exported. Should it be internal to this module?

willeastcott:
LGTM. However, @mvaligursky and @ellthompson should approve before merging since this affects both animation and graphics subsystems.

mvaligursky:
I looks great, but I would not mind some simple test of this in the test/core folder

ellthompson (approved):

src/core/sort.js
kungfooman: I just like to keep things public for potential reuse, but technically it can be private (since its not being used elsewhere)

mvaligursky: I'll merge this and hide cmpPriority in separate PR.

PR #4060, opened 2022-02-28
Remove unused auto instancing code

erikdubbelboer:
Code doesn't do anything and was left from the past.

Removing the `enableAutoInstancing` getter and setter could be a breaking change for users still calling these methods (even though they don't do anything).

### DIFF
--- a/src/framework/application.js
+++ b/src/framework/application.js
@@ -1344,7 +1344,6 @@ class Application extends EventHandler {
         stats.sortTime = this.renderer._sortTime;
         stats.skinTime = this.renderer._skinTime;
         stats.morphTime = this.renderer._morphTime;
-        stats.instancingTime = this.renderer._instancingTime;
         stats.lightClusters = this.renderer._lightClusters;
         stats.lightClustersTime = this.renderer._lightClustersTime;
         stats.otherPrimitives = 0;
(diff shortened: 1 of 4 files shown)


mvaligursky:
I looked at this few times before, but always decided to keep this in, as I'm keen to implement auto instancing / auto batching at some point. Auto batching would be particularly useful for UI rendering.

erikdubbelboer:
Isn't it better to just remove it now and in the future have a single PR which adds all the code required for this? It will probably work differently anyways.

willeastcott:
I'm inclined to agree. I mean, instancing is not down to be worked on in the next few months while we focus on WebGPU etc. If this stuff genuinely doesn't currently work and isn't being used anywhere, I'd say just remove it. But just my humble opinion.

mvaligursky:
I'm fine with removing it. 
With this changed code now, `drawInstance` functions now all returns 0 I believe, we should stop them from returning anything, and clean up how they get called as well to not use the return value.

Otherwise it looks good .. I was worried Editor would need to be updated to hide those stats from it, but those are not exposed anyways.

erikdubbelboer:
Good point, I'll make those changes tomorrow.

erikdubbelboer:
It's fixed.

mvaligursky (approved):
Looks great to me, nicely done!

Issue #3977, opened 2022-02-03
iOS 15 sound does not resume when woken from sleep


yaustar:
Test Project: https://playcanvas.com/project/878048/overview/ios-15-sound-resume-issue

- Set the autolock in settings to be 1m
- Clear Safari Cache
- Open https://playcanv.as/e/p/o6Ew3BtU/
- Touch the screen to start playing sound
- Wait to phone autolocks
- Wait 20s
- Unlock phone
- Audio does not resume, even after touching the device

(Note, may take a few tries)

There's a script that is printing out the status of the context. When it locks, it has interrupted state and when it is unlocked, it prints 'running' which is odd. Do we need to create a new Audio Context because the old one is corrupted?

yaustar:
Risen in priority as a vendor cannot pass a TRC with this bug

mvaligursky:
sounds similar to graphics engine handling context lost and restored events, where we need to re-create all graphics resources. Perhaps something similar is needed for sound.

jpauloruschel:
Reopening due to issue re-appearing: https://forum.playcanvas.com/t/ios-audio-not-resuming-after-standby/26408

jpauloruschel:
I've been investigating this issue and here are my findings:
- Issue occurs consistently 100% of the time on iPhone 13 Safari *and* Chrome;
- Issue never occurs on any non-mobile device;
- When issue occurs, WebAudio's `AudioContext` has a `running` state and successfully returns all `.resume()` requests;
- There are 2 ways to programmatically determine the `AudioContext` is actually broken:
    - attempt to play any sound (which will succeed), but will actually *never* finish (`onended` event will never be fired)
    - query `AudioContext.currentTime`, which will have the same value forever

I believe the issue has something to do with PlayCanvas not correctly verifying if the `AudioContext` is unlocked before being used (which should not be done), leading to an undefined internal state in the browser's implementation. Currently still working on a way of fixing this.

PR #4062, opened 2022-02-28
Fix AudioContext not being properly resumed on iOS 15.

jpauloruschel:
Fixes https://github.com/playcanvas/engine/issues/3977

Fix issues on the `AudioContext` not being properly resumed, especially on iOS 15. This is done by explicitly calling `.resume()` on the `AudioContext` when required. Extra checks were added to make sure `resume` is only called when the window is being re-focused, and the state is going from `suspended` or `interrupted` back to `running` - otherwise exceptions are thrown. Additionally, if for any reason the context can't be resumed successfully, the Auto-Play event listeners are re-bound so that the audio can be resumed later.

Documentation:
- https://developer.mozilla.org/en-US/docs/Web/API/AudioContext
- https://developer.mozilla.org/en-US/docs/Web/API/BaseAudioContext/state
- https://developer.mozilla.org/en-US/docs/Web/API/BaseAudioContext/statechange_event
- https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/resume
- https://developers.google.com/web/updates/2018/11/web-audio-autoplay

As far as I could test, this new code behaves well with all of our current target devices and browsers, but we could also consider using external libraries for this kind of stuff, such as the ones below (pending analysis, of course).
- https://github.com/chrisguttandin/standardized-audio-context
- https://howlerjs.com/

New test scenarios in Safari iOS 15:
- Launch game with audio, change Tabs (audio stops), change back to game (audio resumes)
- Launch game with audio, close Safari (audio stops), wait some time (from immediate to around 40 seconds), open Safari (audio resumes)
- Launch game with audio, wait for 30 second auto-lock (audio stops), wait some time (from immediate to around 40 seconds), unlock phone (audio resumes)
- Launch game with audio, close browser, lock phone. Go back several minutes later: requires interaction due to auto-play policy.

### DIFF
--- a/src/sound/manager.js
+++ b/src/sound/manager.js
@@ -32,6 +32,11 @@ class SoundManager extends EventHandler {
          */
         this._context = null;
         /**
+         * @type {string} the current state of the underlying AudioContext
+         * @private
+         */
+        this._state = 'not created';
+        /**
          * @type {boolean}
          * @private
          */
@@ -116,6 +121,19 @@ class SoundManager extends EventHandler {
                 } else if (typeof webkitAudioContext !== 'undefined') {
                     this._context = new webkitAudioContext();
                 }
+                this._state = this._context.state;
+
+                // When the browser window looses focus (i.e. switching tab, hiding the app on Mobile, etc), the AudioContext state
+                // will be set to 'interrupted' (on iOS Safari) or 'suspended' (on other browsers), and 'resume' must be expliclty called.
+                const self = this;
+                this._context.onstatechange = function () {
+                    // _context.state will have the state its transitioning to, while _state contains the current state
+                    if ((self._state === 'suspended' || self._state === 'interrupted') && self._context.state === 'running') {
+                        // explicitly call .resume() when going from suspended or interrupted to running
+                        self._context.resume();
+                    }
+                    self._state = self._context.state;
+                };
             }
         }
 
@@ -133,11 +151,11 @@ class SoundManager extends EventHandler {
             this.fire('resume');
         };
 
-        // On iOS safari, switching tab or minimizing the browser will set the AudioContext state as 'interrupted'.
-        // On other browsers, AudioContext state will be set to 'suspended'.
-        // In those situations, the .resume() API must be called explicitly
-        if ((hasAudioContext() || this._forceWebAudioApi) &&
-            (this.context.state === 'interrupted' || this.context.state === 'suspended')) {
+        // When the browser window looses focus (i.e. switching tab, hiding the app on Mobile, etc), the AudioContext state
+        // will be set to 'interrupted' (on iOS Safari) or 'suspended' (on other browsers), and 'resume' must be called.
+        // If the AudioContext has not yet been resumed, call for a resume.
+        if ((hasAudioContext() || this._forceWebAudioApi) && this._context &&
+            (this._state === 'interrupted' || this._state === 'suspended')) {
             this.context.resume().then(resumeFunction);
         } else
             resumeFunction();
--- a/src/sound/manager.js
+++ b/src/sound/manager.js
@@ -55,7 +55,9 @@ class SoundManager extends EventHandler {
         this._forceWebAudioApi = options.forceWebAudioApi;
 
         this._resumeContext = null;
+        this._resumeContextAttached = false;
         this._unlock = null;
+        this._unlockAttached = false;
 
         if (hasAudioContext() || this._forceWebAudioApi) {
             this._addAudioContextUserInteractionListeners();
@@ -139,13 +141,13 @@ class SoundManager extends EventHandler {
     }
 
     destroy() {
-        if (this._resumeContext && !!this._resumeContextIsAttached) {
+        if (this._resumeContext && this._resumeContextAttached) {
             USER_INPUT_EVENTS.forEach((eventName) => {
                 window.removeEventListener(eventName, this._resumeContext);
             });
         }
 
-        if (this._unlock && !!this._unlockAttached) {
+        if (this._unlock && this._unlockAttached) {
             window.removeEventListener('touchend', this._unlock);
         }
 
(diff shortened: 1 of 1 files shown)


src/sound/manager.js
```diff
@@ -31,6 +31,11 @@ class SoundManager extends EventHandler {
          * @private
          */
         this._context = null;
+        /**
+         * @type {string} the current state of the underlying AudioContext
```
willeastcott, line 35:
```suggestion
         * The current state of the underlying AudioContext.
         *
         * @type {string}
```

src/sound/manager.js
```diff
@@ -116,6 +121,19 @@ class SoundManager extends EventHandler {
                 } else if (typeof webkitAudioContext !== 'undefined') {
                     this._context = new webkitAudioContext();
                 }
+                this._state = this._context.state;
+
+                // When the browser window looses focus (i.e. switching tab, hiding the app on Mobile, etc), the AudioContext state
```
willeastcott, line 126:
```suggestion
            // When the browser window loses focus (i.e. switching tab, hiding the app on mobile, etc), the AudioContext state

File: src/sound/manager.js [hunk 3]
```diff
@@ -133,11 +151,11 @@ class SoundManager extends EventHandler {
             this.fire('resume');
         };
 
-        // On iOS safari, switching tab or minimizing the browser will set the AudioContext state as 'interrupted'.
-        // On other browsers, AudioContext state will be set to 'suspended'.
-        // In those situations, the .resume() API must be called explicitly
-        if ((hasAudioContext() || this._forceWebAudioApi) &&
-            (this.context.state === 'interrupted' || this.context.state === 'suspended')) {
+        // When the browser window looses focus (i.e. switching tab, hiding the app on Mobile, etc), the AudioContext state
```
willeastcott, line 154:
```suggestion
            // When the browser window loses focus (i.e. switching tab, hiding the app on mobile, etc), the AudioContext state
```

src/sound/manager.js
```diff
@@ -116,6 +121,19 @@ class SoundManager extends EventHandler {
                 } else if (typeof webkitAudioContext !== 'undefined') {
                     this._context = new webkitAudioContext();
                 }
+                this._state = this._context.state;
+
+                // When the browser window looses focus (i.e. switching tab, hiding the app on Mobile, etc), the AudioContext state
+                // will be set to 'interrupted' (on iOS Safari) or 'suspended' (on other browsers), and 'resume' must be expliclty called.
+                const self = this;
+                this._context.onstatechange = function () {
```
willeastcott, line 129: Maybe use an arrow function to avoid having to cache `this` in `self`?

willeastcott (approved):
Really cool to see this fixed! 👏

src/sound/manager.js
```diff
@@ -31,6 +31,13 @@ class SoundManager extends EventHandler {
          * @private
          */
         this._context = null;
+        /**
+         * The current state of the underlying AudioContext.
+         *
+         * @type {string}
+         * @private
+         */
+        this._state = 'not created';
```
mvaligursky, line 40: please create some constant states just before `class SoundManager` line, something along:
```
const STATE_NOTCREATED = "not created";
...
```

and use those in the code to avoid possibility of typo.

src/sound/manager.js
```diff
@@ -220,6 +205,95 @@ class SoundManager extends EventHandler {
...
+    }
+
+    /**
+     * Add the necessary Window EventListeners for resuming the AudioContext to comply with auto-play policies.
+     * For more info, https://developers.google.com/web/updates/2018/11/web-audio-autoplay.
+     *
+     * @private
+     */
+    _addAudioContextUserInteractionListeners() {
+        const userInputEventNames = [
```
mvaligursky, line 241: should `destroy` function be updated to remove all these events?

mvaligursky:
Please test this on IE11 as well .. as that could behave differently too (if not already)

src/sound/manager.js
```diff
@@ -220,6 +216,90 @@ class SoundManager extends EventHandler {
...
+    _addAudioContextUserInteractionListeners() {
+        // resume AudioContext on user interaction because of autoplay policy
+        if (!this._resumeContext) {
+            this._resumeContext = () => {
+                if (!this.context || this.context.state === 'running') {
+                    USER_INPUT_EVENTS.forEach((eventName) => {
+                        window.removeEventListener(eventName, this._resumeContext);
+                    });
+
+                    this._resumeContextIsAttached = false;
```
mvaligursky, line 260: ideally _resumeContextIsAttached would be initialized in constructor as well

src/sound/manager.js
```diff
@@ -128,28 +129,23 @@ class SoundManager extends EventHandler {
...
+        if (this.context && (this._state === CONTEXT_STATE_INTERRUPTED || this._state === CONTEXT_STATE_SUSPENDED)) {
+            this._safelyResumeContext();
+        }
     }
 
     destroy() {
-        if (this._resumeContext) {
-            window.removeEventListener('mousedown', this._resumeContext);
-            window.removeEventListener('touchend', this._resumeContext);
+        if (this._resumeContext && !!this._resumeContextIsAttached) {
```
mvaligursky, line 142: no need for !!

src/sound/manager.js
```diff
@@ -128,28 +129,23 @@ class SoundManager extends EventHandler {
...
+            USER_INPUT_EVENTS.forEach((eventName) => {
+                window.removeEventListener(eventName, this._resumeContext);
+            });
         }
 
-        if (this._unlock) {
+        if (this._unlock && !!this._unlockAttached) {
```
mvaligursky, line 148: same here, you don't need `!!`
and also add it to constructor please

mvaligursky (approved):
Good work. Anybody else wants to have a look?

src/sound/manager.js
```diff
@@ -220,6 +218,90 @@ class SoundManager extends EventHandler {
...
+     *
+     * @private
+     */
+    _addAudioContextUserInteractionListeners() {
+        // resume AudioContext on user interaction because of autoplay policy
+        if (!this._resumeContext) {
+            this._resumeContext = () => {
+                if (!this.context || this.context.state === 'running') {
+                    USER_INPUT_EVENTS.forEach((eventName) => {
+                        window.removeEventListener(eventName, this._resumeContext);
```
slimbuck, line 259: inception. i like.

slimbuck (approved):
l approve of this!

PR #4064, opened 2022-02-28
[Breaking][Anim Component] Layer blending update

ellthompson:
This PR adds the option to toggle the normalisation of layer weights on or off.

If weight normalisation is turned on the algorithm functionality is the same as before:
- Weights of all layers sum up to 1 (normalised by total weight)
- Overwrite layers set all previous layer weights to 0
- The resulting animation is a linear combination of all layer values and their weights

If weight normalisation is turned off the following effects take place:
- Weight values are not normalised by their total weight
- Layer weights are clamped between the 0 and 1 ranges
- Additive layers add their values on top of previous layer values based on their weight value
- Overwrite layers replace a portion of previous layer values based on their weight value
- Overwrite layers no longer set the weights of previous layers to 0

The default setting going forward will be not to normalize weights. As the anim component will no longer use the previous blending algorithm as the default, this PR becomes a breaking change. If users wish to use the old blending algorithm, they will be required to set the `normalizeWeights` property to `true`.

This PR creates new public API:
```javascript

/**
 * Blend from the current weight value to the provided weight value over a given amount of time.
 *
 * @param {number} weight - The new weight value to blend to.
 * @param {number} time - The duration of the blend in seconds.
 */
pc.AnimComponentLayer#blendToWeight

/**
 * If true the animation component will normalize the weights of its layers by their sum total.
 *
 * @type {boolean}
 */
pc.AnimComponent#normalizeWeights
```


mvaligursky:
> The default setting going forward will be not to normalize weights. As the anim component will no longer use the previous blending algorithm as the default, this PR becomes a breaking change. If users wish to use the old blending algorithm, they will be required to set the `normalizeWeights` property to `true`.

I wonder if we could handle this as non-breaking change? If normalizeWeights is undefined (existing assets), we should assume it's true. For new assets this will be initialized to false.

src/anim/evaluator/anim-target-value.js
```diff
@@ -28,22 +41,36 @@ class AnimTargetValue {
...
+        return this._component.layers[index].blendType;
     }
 
     setMask(index, value) {
         this.mask[index] = value;
-        if (this._component.layers[index].blendType === ANIM_LAYER_OVERWRITE) {
-            this.mask = this.mask.fill(0, 0, index);
+        if (this._normalizeWeights) {
+            if (this._component.layers[index].blendType === ANIM_LAYER_OVERWRITE) {
+                this.mask = this.mask.fill(0, 0, index);
```
mvaligursky, line 70: this should probably execute in !normalizeWeights case a well, if the weight of the layer is 1?

src/anim/evaluator/anim-target-value.js
```diff
@@ -58,17 +85,37 @@ class AnimTargetValue {
     updateValue(index, value) {
         // always reset the value of the target when the counter is 0
         if (this.counter === 0) {
-            this.value[0] = 0;
-            this.value[1] = 0;
-            this.value[2] = 0;
-            this.value[3] = 1;
+            AnimEvaluator._set(this.value, [0, 0, 0, 1], this.valueType);
```
mvaligursky, line 88: don't construct the constant array here, have it once outside of the class

src/anim/evaluator/anim-target-value.js
```diff
@@ -58,17 +85,37 @@ class AnimTargetValue {
...
+            if (this.valueType === AnimTargetValue.TYPE_QUAT) {
+                // current value
+                const v = AnimTargetValue.q1.set(...this.value);
+                // additive value
+                const aV = AnimTargetValue.q2.set(...this.baseValue).invert().mul(AnimTargetValue.q3.set(...value));
+                // scale additive value by it's weight
+                aV.slerp(Quat.IDENTITY, aV, this.getWeight(index));
+                // add the additive value onto the current value then set it to the targets value
+                v.mul(aV);
+                AnimEvaluator._set(this.value, [v.x, v.y, v.z, v.w], this.valueType);
```
mvaligursky, line 104: any good way to implement this without array allocation? Maybe have preallocate array and just set values to it.

src/anim/evaluator/anim-target-value.js
```diff
@@ -58,17 +85,37 @@ class AnimTargetValue {
...
-            this.value[3] = 1;
+            AnimEvaluator._set(this.value, [0, 0, 0, 1], this.valueType);
+            if (!this._normalizeWeights) {
+                AnimEvaluator._blend(this.value, this.baseValue, 1, this.valueType);
+            }
         }
         if (!this.mask[index]) return;
-        if (this.counter === 0) {
-            AnimEvaluator._set(this.value, value, this.valueType);
+        if (this._layerBlendType(index) === ANIM_LAYER_ADDITIVE && !this._normalizeWeights) {
```
mvaligursky, line 94: this whole block of code is performance critical path, it'd be good to make it very optimized, it does not seem to be at the moment. Even `...` likely allocates array under the hood.

src/framework/components/anim/component-layer.js
```diff
@@ -199,6 +208,19 @@ class AnimComponentLayer {
         return this._mask;
     }
 
+    /**
+     * Whether this layers weight should be normalized by the total weight of all layers or calculated separately.
+     *
+     * @type {boolean}
+     */
+    set normalizedWeight(value) {
+        this._normalizedWeight = value;
```
mvaligursky, line 217: looking at `set blendType`, it feels like maybe `rebind()` should be called from here too in some case?

src/framework/components/anim/component-layer.js
```diff
@@ -233,9 +255,35 @@ class AnimComponentLayer {
...
+        if (this._blendTime) {
+            if (this._blendTimeElapsed < this._blendTime) {
+                this.weight = math.lerp(this._startingWeight, this._targetWeight, this._blendTimeElapsed / this._blendTime);
+                this._blendTimeElapsed += dt;
+            } else {
+                this.weight = this._targetWeight;
+                this._blendTime = null;
+                this._blendTimeElapsed = 0;
+                this._startingWeight = null;
+                this._targetWeight = null;
```
mvaligursky, line 267: personally, if something is a number, I only store numbers to it and not null.
I suspect when type changes, the function needs recompile under the hood to run optimized.
I'd suggest to use numbers .. 0 would work just fine.

src/anim/evaluator/anim-target-value.js
ellthompson: The mask is used by the normalised weight algorithm to removed these layers from the equation independent of their current weight value. Though I've added a check in the updateValue function to check whether a layer weight is 0 though, in which case it is also skipped.

src/anim/evaluator/anim-target-value.js
ellthompson: Done 👍

src/anim/evaluator/anim-target-value.js
ellthompson: I've moved this to a static array.

src/anim/evaluator/anim-target-value.js
ellthompson: I've removed the spread operations here, it's now just setting individual values from the various arrays.

src/framework/components/anim/component-layer.js
ellthompson: Sorry this was an old function. It's been moved to the component class and it resets the state graph after being set.

src/framework/components/anim/component-layer.js
ellthompson: Done 👍

src/anim/evaluator/anim-target-value.js
mvaligursky: I understand. But what I was wondering is if there is a case where some layers should be 'ignored' in case normalizeWeights is false. Is there some setup in this case, where all underlaying layers are ignored? For example if the layer is 'overwrite' type (not additive) and its weight is 1? Or is there nothing like that when normalizeWeights is false.

src/anim/evaluator/anim-target-value.js
```diff
@@ -58,17 +91,48 @@ class AnimTargetValue {
...
+                v.mul(aV);
+                AnimTargetValue.quatArr[0] = v.x;
+                AnimTargetValue.quatArr[1] = v.y;
+                AnimTargetValue.quatArr[2] = v.z;
+                AnimTargetValue.quatArr[3] = v.w;
+                AnimEvaluator._set(this.value, AnimTargetValue.quatArr, this.valueType);
+            } else {
+                AnimTargetValue.v1.set(value[0], value[1], value[2]);
+                AnimTargetValue.v2.set(this.baseValue[0], this.baseValue[1], this.baseValue[2]);
+                const addValue = AnimTargetValue.v1.sub(AnimTargetValue.v2);
```
mvaligursky, line 120: you copy those values to v1 and v2 here just to run sub. I'd probably just do (with having some static addValue Vec3)
`addValue.set(value[0] - this.baseValue[0], value[1] - this.baseValue[1], value[2] - this.baseValue[2]);

mvaligursky: and then you copy it to `AnimTargetValue.vecArr[0]`
maybe just sub straight to it?
```
AnimTargetValue.vecArr[0] = value[0] - this.baseValue[0];
...
```

src/anim/evaluator/anim-target-value.js
ellthompson: Nope the mask setup is only used for the normalizeWeights algorithm, theres nothing like that used when normalizeWeights is false.

src/anim/evaluator/anim-target-value.js
ellthompson: Sure, I'll update this.

mvaligursky (approved):
Looks good now, I'm approving it.

But before the merge, I would just want to double check this is a breaking change and cannot be done as non-breaking. I think you explained it cannot, but please consider it one more time.

PR #4072, opened 2022-03-01
Capitalize callback types

willeastcott:
This PR capitalizes the types of callbacks in the engine, which feels more consistent with the rest of the API.


slimbuck (approved):
Might this break existing ts applications which declare these types? If so, might be worth adding typedefs somewhere to the old names.

willeastcott:
@slimbuck I don't think so because callbacks had different names in 1.51 anyway (namespaced as `callbacks.*`).

Issue #4066, opened 2022-02-28
Box-reflection examples has screen reflection missing


mvaligursky:
For some reason, the reflection of the movie is gone, locally and on prod.
![Screenshot 2022-02-28 at 16 48 16](https://user-images.githubusercontent.com/59932779/156023827-15ae5a40-116c-4b92-9ece-64f15bdf1d62.png)

Based on quick capture, the reflection is rendered to cubemap correctly:
![Screenshot 2022-02-28 at 16 49 21](https://user-images.githubusercontent.com/59932779/156023898-1a8910da-c9ab-4bc3-87c9-b1025f840a75.png)

Follow up code is harder to debug as it's in a format (RGBM?) that spector does not display. Needs deeper investingation to pinpoint the issue. I tried going to history of engine changes and didn't find it issue in the last 3 weeks, must be older.

mvaligursky:
update on this. On my macOS Monterey 12.2.1, I get:
- not working in Chrome  98.0.4758.102 
- working in Forefox 97.0.1 
- not working in Safari 15.3 
- working in Safari Technology Preview 15.4

slimbuck:
Is it related to engine version at all?

mvaligursky:
I don't believe so, but local as well as older I tried behave the same way.

mvaligursky:
It can still be an issue in our code though and it's getting exposed by update in browser possibly?

mvaligursky:
the reflection is missing in Chrome Canary 101.0.4916.0 as well

PR #4074, opened 2022-03-01
Changes to make BoxReflection example work more reliably

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/4066

- changed the default format of the depth buffer on webgl2 platform to 32bit, from 16bit. In case MSAA is used, this format now matches the format depth was getting resolved to, potentially avoiding format conversion
- adjusted camera near/far distance for when box-reflection example renders into cubemap, to avoid hitting precision issues on 16bit depth buffer (on webgl1 only with this change)

### DIFF
--- a/src/graphics/webgl/webgl-graphics-device.js
+++ b/src/graphics/webgl/webgl-graphics-device.js
@@ -1270,7 +1270,8 @@ class WebglGraphicsDevice extends GraphicsDevice {
                     gl.renderbufferStorage(gl.RENDERBUFFER, gl.DEPTH_STENCIL, target.width, target.height);
                     gl.framebufferRenderbuffer(gl.FRAMEBUFFER, gl.DEPTH_STENCIL_ATTACHMENT, gl.RENDERBUFFER, target._glDepthBuffer);
                 } else {
-                    gl.renderbufferStorage(gl.RENDERBUFFER, gl.DEPTH_COMPONENT16, target.width, target.height);
+                    const depthFormat = this.webgl2 ? gl.DEPTH_COMPONENT32F : gl.DEPTH_COMPONENT16;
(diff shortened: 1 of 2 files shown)


src/graphics/webgl/webgl-graphics-device.js
```diff
@@ -1270,7 +1270,8 @@ class WebglGraphicsDevice extends GraphicsDevice {
                     gl.renderbufferStorage(gl.RENDERBUFFER, gl.DEPTH_STENCIL, target.width, target.height);
                     gl.framebufferRenderbuffer(gl.FRAMEBUFFER, gl.DEPTH_STENCIL_ATTACHMENT, gl.RENDERBUFFER, target._glDepthBuffer);
                 } else {
-                    gl.renderbufferStorage(gl.RENDERBUFFER, gl.DEPTH_COMPONENT16, target.width, target.height);
+                    const depthFormat = this.webgl2 ? gl.DEPTH_COMPONENT32F : gl.DEPTH_COMPONENT16;
```
slimbuck, line 1273: wouldn't gl.DEPTH_COMPONENT24 for webgl2 match the usual (backbuffer) case better?

slimbuck: slightly concerned about unnecessary perf cost using fp32

mvaligursky: And also, even 24 bit buffer would internally be stored as 32bit, so there is no size / bandwidth win either.

slimbuck: I would assume float slower than int, especially on mobile. Yes you're right bandwidth is the same between 24 and 32bit, but float could still be slower than int on mobile.

slimbuck (approved):

PR #4075, opened 2022-03-01
[AnimComponent] Fix for layer masks in cloned components

### LINKED ISSUE #4031: Instantiating a template with a layer mask in Anim doesn't apply the mask
Forum post: https://forum.playcanvas.com/t/entities-instantiated-from-templates-are-missing-their-anim-layer-masks/24330

Example project: https://playcanvas.com/project/883797/overview/layer-mask-clone-bug

Only happens when calling instantiate from code. Creating a new instance of the template in the Editor is fine.

ellthompson:
If an anim component is getting cloned and it is animating from the components own entity, we should be mapping all layer mask paths to start from the cloned entity rather than the old entity.

Fixes #4031 

### DIFF
--- a/src/framework/components/anim/system.js
+++ b/src/framework/components/anim/system.js
@@ -116,6 +116,7 @@ class AnimComponentSystem extends ComponentSystem {
             layers: entity.anim.layers,
             layerIndices: entity.anim.layerIndices,
             parameters: entity.anim.parameters,
+            masks
         };
         return this.addComponent(clone, data);
     }
--- a/src/framework/components/anim/system.js
+++ b/src/framework/components/anim/system.js
@@ -89,6 +89,22 @@ class AnimComponentSystem extends ComponentSystem {
     }
 
     cloneComponent(entity, clone) {
+        let masks;
+        // If the component doesn't specify a rootBone, any layer mask hierarchy should be updated from the old entity to the cloned entity.
+        if (!entity.anim.rootBone) {
+            masks = {};
+            entity.anim.layers.forEach((layer, i) => {
+                if (layer.mask) {
+                    const mask = {};
+                    Object.keys(layer.mask).forEach((path) => {
+                        // The base of all mask paths should be mapped from the previous entity to the cloned entity
+                        const clonePath = path.replace(path.split('/')[0], clone.name);
+                        mask[clonePath] = layer.mask[path];
+                    });
+                    masks[i] = { mask };
+                }
+            });
+        }
         const data = {
             stateGraphAsset: entity.anim.stateGraphAsset,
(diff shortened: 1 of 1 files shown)


src/framework/components/anim/system.js
```diff
@@ -89,6 +89,22 @@ class AnimComponentSystem extends ComponentSystem {
     }
 
     cloneComponent(entity, clone) {
+        let masks;
```
mvaligursky, line 92: looks good, but this not take place even when the rootBone is specified? It can point to the root of the hierarchy in which case it's the same case?
I also wonder if we should in debug mode validate those paths against the hierarchy? Or does it take place when binding takes place? To make sure it's compatible skeleton.

ellthompson: Yeah, I was thinking that the rootBone would be null in that case (it is on the editor side). That should probably be replicated in the engine logic (assigning the components entity as the rootBone should result in a null rootBone).

The debugging would need to occur in the binder.

mvaligursky: I think it would not be null if the user assigned root entity to it?

ellthompson: Perhaps we should null the rootBone property if the components entity is assigned to it? To mirror the editor logic of entity assignments? Or we do a check for the component's entity as the rootBone here.

mvaligursky: I'd prefer we don't change the rootBone assignment .. if user assigns something, it should stay assigned, even if it's the same as the default case.

ellthompson: Ok i've updated this check to include the current entity.

src/framework/components/anim/system.js
```diff
@@ -89,6 +89,22 @@ class AnimComponentSystem extends ComponentSystem {
...
+        // If the component animaites from the components entity, any layer mask hierarchy should be updated from the old entity to the cloned entity.
+        if (!entity.anim.rootBone || entity.anim.rootBone === entity) {
+            masks = {};
+            entity.anim.layers.forEach((layer, i) => {
+                if (layer.mask) {
+                    const mask = {};
+                    Object.keys(layer.mask).forEach((path) => {
+                        // The base of all mask paths should be mapped from the previous entity to the cloned entity
+                        const clonePath = path.replace(path.split('/')[0], clone.name);
```
mvaligursky, line 101: could you also improve this line to make sure it only renames the first segment of the path (that's what needs to happen right?) .. for example replacing `root` by `clone` in `/root/rooter/something' would not give us expected result.

mvaligursky (approved):

Issue #1996, opened 2020-04-14
UI button component bug when removing from list


LeXXik:
I have an element group in 2D Screen. Each child has a button, that allows it to remove itself from the list. I simply destroy the list and re-populate it with the new list, excluding the removed element.

The problem happens when the `.destroy()` method is called. I have made [a sample project](https://playcanvas.com/editor/scene/904854) to demonstrate the issue.

yaustar:
Error when clicking on the button:

```
playcanvas.dbg.js?version=1.51.4:43575 Uncaught TypeError: Cannot read properties of null (reading 'active')
    at ButtonComponent.get (playcanvas.dbg.js?version=1.51.4:43575)
    at ButtonComponent._determineVisualState (playcanvas.dbg.js?version=1.51.4:49399)
    at ButtonComponent._updateVisualState (playcanvas.dbg.js?version=1.51.4:49329)
    at ButtonComponent._onMouseUp (playcanvas.dbg.js?version=1.51.4:49237)
    at ElementComponent.fire (playcanvas.dbg.js?version=1.51.4:758)
    at ElementInput._fireEvent (playcanvas.dbg.js?version=1.51.4:75357)
    at ElementInput._onElementMouseEvent (playcanvas.dbg.js?version=1.51.4:75219)
    at ElementInput._handleUp (playcanvas.dbg.js?version=1.51.4:75044)
```

jpauloruschel:
As a separate issue I came across while testing the fix I explained in my earlier comment, I noticed that the `Button`s and `Element`s were not responding properly to all events: they were responding to `click`, but not to `hover`, so they were interactable but the "hover tint" animation was not working. The quick fix for that is to  set `enabled = true` *after* it is added to the hierarchy:

```
// instead of:
  ...
  entry.enabled = true;
  this.group.addChild(entry);
}

// do this:
  ...
  this.group.addChild(entry);
  entry.enabled = true;
}
```

I think this needs a new investigation.

LeXXik:
That was a good investigation! Thank you for looking into it, @jpauloruschel !

jpauloruschel:
About the `enabled` and `addChild` order issue:

## The Issue

[...]

PR #4081, opened 2022-03-02
Fix EntityReference not finding the entity if the parent was not added to the scene yet.

jpauloruschel:
Fixes https://github.com/playcanvas/engine/issues/1996

This PR fixes an issue in the `EntityReference` class, which is used for managing references between entities like, for instance, a `Button`'s `imageEntity` (https://developer.playcanvas.com/api/pc.ButtonComponent.html#imageEntity): if the parent entity (the one that contains the `EntityReference`) is not in the scene graph (not a descendant to the App's `root`), and the reference is made via a `guid` (string), the reference is never found.

The fix is simply to always check against the parent entity's root - which, would be either be `app.root`, or the root of the detached graph.

This issue is especially noticeable when cloning entities or templates which contain such references internally.

### DIFF
--- a/src/framework/utils/entity-reference.js
+++ b/src/framework/utils/entity-reference.js
@@ -266,11 +266,12 @@ class EntityReference extends EventHandler {
             nextEntityGuid = nextEntity.getGuid();
             this._parentComponent.data[this._entityPropertyName] = nextEntityGuid;
         } else {
-            const appRoot = this._parentComponent.system.app.root;
-            const isOnSceneGraph = this._parentComponent.entity.isDescendantOf(appRoot);
+            // If value is the GUID, look for the entity from the root of the parent entity.
+            // We use entity.root rather than app.root because the parent entity may be detached
+            // from the scene, and the value is referencing an entity within this detached graph.
+            const root = this._parentComponent.entity.root;
 
-            const searchRoot = isOnSceneGraph ? appRoot : this._parentComponent.entity;
-            nextEntity = nextEntityGuid ? searchRoot.findByGuid(nextEntityGuid) : null;
+            nextEntity = nextEntityGuid ? root.findByGuid(nextEntityGuid) : null;
         }
 
         const hasChanged = this._entity !== nextEntity;
--- a/src/framework/utils/entity-reference.js
+++ b/src/framework/utils/entity-reference.js
@@ -266,10 +266,11 @@ class EntityReference extends EventHandler {
             nextEntityGuid = nextEntity.getGuid();
             this._parentComponent.data[this._entityPropertyName] = nextEntityGuid;
         } else {
-            const root = this._parentComponent.system.app.root;
-            const isOnSceneGraph = this._parentComponent.entity.isDescendantOf(root);
+            const appRoot = this._parentComponent.system.app.root;
+            const isOnSceneGraph = this._parentComponent.entity.isDescendantOf(appRoot);
 
-            nextEntity = (isOnSceneGraph && nextEntityGuid) ? root.findByGuid(nextEntityGuid) : null;
+            const searchRoot = isOnSceneGraph ? appRoot : this._parentComponent.entity;
+            nextEntity = nextEntityGuid ? searchRoot.findByGuid(nextEntityGuid) : null;
         }
 
         const hasChanged = this._entity !== nextEntity;
--- a/src/framework/utils/entity-reference.js
+++ b/src/framework/utils/entity-reference.js
@@ -265,16 +265,14 @@ class EntityReference extends EventHandler {
             nextEntity = nextEntityGuid;
             nextEntityGuid = nextEntity.getGuid();
             this._parentComponent.data[this._entityPropertyName] = nextEntityGuid;
-        } else {
-            if (nextEntityGuid) {
-                // If value is the GUID, look for the entity from the root of the parent entity.
-                // We use both entity.root and app.root because the parent entity may be detached
-                // from the scene, and the value could be referencing an entity within this detached graph.
-                const sceneRoot = this._parentComponent.system.app.root;
-                const entityRoot = this._parentComponent.entity.root;
+        } else if (nextEntityGuid) {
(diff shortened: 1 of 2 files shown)


willeastcott (approved):
Nice. 😄

src/framework/utils/entity-reference.js
```diff
@@ -266,10 +266,11 @@ class EntityReference extends EventHandler {
             nextEntityGuid = nextEntity.getGuid();
             this._parentComponent.data[this._entityPropertyName] = nextEntityGuid;
         } else {
-            const root = this._parentComponent.system.app.root;
-            const isOnSceneGraph = this._parentComponent.entity.isDescendantOf(root);
+            const appRoot = this._parentComponent.system.app.root;
+            const isOnSceneGraph = this._parentComponent.entity.isDescendantOf(appRoot);
 
-            nextEntity = (isOnSceneGraph && nextEntityGuid) ? root.findByGuid(nextEntityGuid) : null;
+            const searchRoot = isOnSceneGraph ? appRoot : this._parentComponent.entity;
```
mvaligursky, line 272: I wonder if `this._parentComponent.entity` could even be `this._parentComponent.entity.root` - to search on whole disconnected tree, not only entity's subtree.

mvaligursky (approved):

mvaligursky:
Now that you understand the code, feel free to add some comments to it, at least the changed function.

src/framework/utils/entity-reference.js
```diff
@@ -266,10 +266,12 @@ class EntityReference extends EventHandler {
...
         } else {
-            const root = this._parentComponent.system.app.root;
-            const isOnSceneGraph = this._parentComponent.entity.isDescendantOf(root);
+            // If value is the GUID, look for the entity from the root of the parent entity.
+            // We use entity.root rather than app.root because the parent entity may be detached
+            // from the scene, and the value is referencing an entity within this detached graph.
+            const root = this._parentComponent.entity.root;
 
-            nextEntity = (isOnSceneGraph && nextEntityGuid) ? root.findByGuid(nextEntityGuid) : null;
+            nextEntity = nextEntityGuid ? root.findByGuid(nextEntityGuid) : null;
```
mvaligursky, line 274: much cleaner. How about something like this, to search in both scene and entity trees, to increase likelihood of finding it?

```
let nextEntity = null;
if (nextEntityGuid instanceof Entity) {
    // if value is set to a Entity itself replace value with the GUID
    nextEntity = nextEntityGuid;
    nextEntityGuid = nextEntity.getGuid();
    this._parentComponent.data[this._entityPropertyName] = nextEntityGuid;
} else if (nextEntityGuid) {
    const sceneRoot = this._parentComponent.system.app.root;
    const entityRoot = this._parentComponent.entity.root;

    nextEntity = entityRoot.findByGuid(nextEntityGuid) || sceneRoot.findByGuid(nextEntityGuid);
}
```

src/framework/utils/entity-reference.js
```diff
@@ -266,12 +266,17 @@ class EntityReference extends EventHandler {
...
+            if (nextEntityGuid) {
+                // If value is the GUID, look for the entity from the root of the parent entity.
+                // We use both entity.root and app.root because the parent entity may be detached
+                // from the scene, and the value could be referencing an entity within this detached graph.
+                const sceneRoot = this._parentComponent.system.app.root;
+                const entityRoot = this._parentComponent.entity.root;
+
+                nextEntity = entityRoot.findByGuid(nextEntityGuid) || sceneRoot.findByGuid(nextEntityGuid);
+            } else {
+                nextEntity = null;
```
mvaligursky, line 278: just initialize to null when declaring .. no need for else

mvaligursky (approved):
looks great, done.

src/framework/utils/entity-reference.js
jpauloruschel: Oh yea, that's better!

src/framework/utils/entity-reference.js
```diff
@@ -258,18 +258,23 @@ class EntityReference extends EventHandler {
...
         let nextEntityGuid = this._parentComponent.data[this._entityPropertyName];
-        let nextEntity;
+        let nextEntity = null;
 
         if (nextEntityGuid instanceof Entity) {
             // if value is set to a Entity itself replace value with the GUID
             nextEntity = nextEntityGuid;
             nextEntityGuid = nextEntity.getGuid();
             this._parentComponent.data[this._entityPropertyName] = nextEntityGuid;
         } else {
```
mvaligursky, line 268: tiny, but make it 

`} else if (nextEntityGuid) {`

mvaligursky: instead of 

```
} else {
    if (nextEntityGuid) {
```

mvaligursky (approved):

src/framework/utils/entity-reference.js
```diff
@@ -235,15 +235,11 @@ class EntityReference extends EventHandler {
...
-     * to be correctly resolved when {@link Entity#clone} is called.
+     * to be correctly resolved when the referenced entity was not in the same graph as the parent.
      *
      * @private
      */
     onParentComponentEnable() {
-        // When an entity is cloned via the JS API, we won't be able to resolve the
-        // entity reference until the cloned entity has been added to the scene graph.
-        // We can detect this by waiting for the parent component to be enabled, in the
-        // specific case where we haven't yet been able to resolve an entity reference.
```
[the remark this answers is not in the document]
mvaligursky: still partially applies. We can resolve entities in scene graph, or inside a tree the component is on, but not on other subtrees that will later get added (maybe even templates?)
ideally we run this entity-reference resolve in a separate pass after the scene has been assembled, but that's separate work.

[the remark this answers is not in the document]
jpauloruschel: The specific comment was about cloning via the JS API, which with the new changes is now going to work. The only case this wouldn't work is if they are in completely separate graphs, and I added this case to the comment on the function above.

PR #4082, opened 2022-03-04
Lite version of the Application to allow for tree-shaking (WIP)

mvaligursky:


LeXXik:
It would be great, if we would have access to components selection from the Editor. The builder would then assemble an application based on that configuration.

mvaligursky:
> It would be great, if we would have access to components selection from the Editor. The builder would then assemble an application based on that configuration.

Yes, in the longer term the plan is to certainly allow tree shaking in the Editor, likely in a form of some builder. The complexity we need to solve after this PR is the scripts - as those are not modules (for now anyways), and so we cannot easily find out what needs to be imported to the application.

willeastcott (approved):

slimbuck (approved):

Issue #2856, opened 2021-02-11
Implement support for WEBGL_depth_texture extension


mvaligursky:
At the moment on WebGl1 platforms, we re-render World layer to a Depth texture to give post-processing and similar give access to scene's depth.

WebGl1 implements mentioned WEBGL_depth_texture extension: https://www.khronos.org/registry/webgl/extensions/WEBGL_depth_texture/

and based on this, we can pretty much take it for granted on all devices:
https://webgl2fundamentals.org/webgl/lessons/webgl1-to-webgl2.html

This could be a considerable performance gain in some cases.

Note - it's not supported by IE11 .. we can keep the old path for this, or we can perhaps disable any post-effects that require depth.

Maksims:
> or we can perhaps disable any post-effects that require depth.

This would make development and debugging for users a difficulty. Not all users have access to IE11, also in engine, pretty much all features have a fallback, so this very likely shouldn't be an exception. If fallback has a slow path - that is different story, but at least it works as expected.

mvaligursky:
I just checked on iOS 14.5, and WEBGL_depth_texture is supported there - we should definitely use it there.

mvaligursky:
This is less relevant with the release of iOS 15, which supports WebGl2 and hence the extension is no longer required.

willeastcott:
Let's close then...

PR #4088, opened 2022-03-06
Add shader textureBias

slimbuck:
This PR:
- adds a global `textureBias` uniform to control biased texture reads across standard material textures
- enables `WEBGL_depth_texture` extension on webgl1 devices (used when creating texture-based render targets)
- updates example package-lock file to version 2

These changes aid with the implementation of a high quality, multi-frame rendering solution.

Questions:
1. the biased version of texture2D is supported out-of-the-box on webgl 1.0, but is there any overhead calling this version of the function when bias is 0? (I expect probably not, but how do we check?)

Answers:
1. likely no performance impact


src/graphics/program-lib/chunks/base.frag.js
```diff
@@ -3,6 +3,8 @@ uniform vec3 view_position;
 
 uniform vec3 light_globalAmbient;
 
+uniform float globalTextureBias;
```
willeastcott, line 6: Wondering whether this should just be `textureBias`.

src/graphics/grab-pass.js
```diff
@@ -8,26 +8,33 @@ import { Texture } from "./texture.js";
...
         this.texture = null;
         this.renderTarget = null;
         this.textureId = null;
+
+        const handler = () => {
+            this.destroy();
+        };
+
+        device.once('destroy', handler);
+        device.on('devicelost', handler);
```
willeastcott, line 37: Do you need to `off` these somewhere?

willeastcott:
> we don't want `GrabPass` to become official public API yet: do we flag the class as @Private or @ignore?

Use @ignore.

@private is just for class fields and methods and not classes themselves.

src/graphics/program-lib/chunks/base.frag.js
slimbuck: Since texture bias could be specified per texture2D call I thought it was useful to highlight the fact this bias is used globally on all standard material textures.

willeastcott: The shader is agnostic as to whether the `textureBias` value is updated once per scene or once per draw call though, right? To the shader, it's just a 'texture bias'.

willeastcott:
Did you mean to update pacakge-lock.json in this PR?

src/graphics/webgl/webgl-graphics-device.js
```diff
@@ -791,6 +791,7 @@ class WebglGraphicsDevice extends GraphicsDevice {
             }
             this.extColorBufferFloat = null;
             this.extDisjointTimerQuery = null;
+            this.extDepthTexture = gl.getExtension('WEBGL_depth_texture');
```
willeastcott, line 794: Can you give some background why this PR needs this extension?

src/graphics/program-lib/chunks/base.frag.js
slimbuck: Right, but from the application perspective setting 'globalTextureBias' sets it for all the textures, globally. While naming it 'textureBias' wouldn't be wrong, I think there's value in making the fact this is a global uniform explicit. If I was reading code with a 'textureBias' variable, I'd assume there were one of these per texture.

slimbuck:
> Did you mean to update pacakge-lock.json in this PR?

Yes! but forgot to add to the PR description.

src/graphics/program-lib/chunks/base.frag.js
slimbuck: @mvaligursky do you have an opinion either way?

slimbuck: and @willeastcott if you feel strongly ab it, I'm ok to go with 'textureBias'.

willeastcott: I shall let @mvaligursky tie break.

src/graphics/webgl/webgl-graphics-device.js
slimbuck: added to the PR description, thanks.

src/graphics/program-lib/chunks/base.frag.js
mvaligursky: I agree with @willeastcott here. The shader just uses the uniform, that the engine sets up once per frame or once per draw call - shader knows no difference.

slimbuck: noooooooooooo

slimbuck: done

willeastcott (approved):
Awesome. Maybe edit the PR title tho.

mvaligursky:
related to https://github.com/playcanvas/engine/issues/2856

src/graphics/webgl/webgl-graphics-device.js
```diff
@@ -1004,9 +1009,6 @@ class WebglGraphicsDevice extends GraphicsDevice {
             shader.loseContext();
         }
 
-        // grab pass
-        this.grabPass.destroy();
-
```
mvaligursky, line 1009: should this be removed?

src/graphics/webgl/webgl-graphics-device.js
```diff
@@ -1045,8 +1047,6 @@ class WebglGraphicsDevice extends GraphicsDevice {
         for (const buffer of this.buffers) {
             buffer.unlock();
         }
-
-        this.grabPass.create();
```
mvaligursky, line 1049: should this be removed?

src/graphics/webgl/webgl-graphics-device.js
```diff
@@ -568,6 +568,9 @@ class WebglGraphicsDevice extends GraphicsDevice {
         }
 
         this.constantTexSource = this.scope.resolve("source");
+        this.textureBias = this.scope.resolve("textureBias");
+
+        this.textureBias.setValue(0.0);
```
mvaligursky, line 573: I think we should have a public API for this. Ideally the viewer code you test it on would not do
`this.scope.resolve("textureBias").SetValue(...);`

willeastcott: This transform feedback example already uses this API. This uniform management is _long_ overdue a redesign.

mvaligursky: That's true, but transform feedback example is accessing this API to set uniforms in its own shader.
But here we're adding new functionality to the engine that the Viewer wants to use .. and that should be different API, allowing us to even rename the variable later if needed.  Maybe `Scene.standardMaterialTextureBias' or a shorter version of it? Something we publicly document.

willeastcott: I'd prefer something like `Scene#textureBias` and then potentially add `Texture#bias` which would take precedence.

slimbuck: Could be a scene setting... though I would prefer creating and exposing public API in followup PR.

willeastcott: Fine by me.

mvaligursky (approved):

PR #4098, opened 2022-03-09
DepthFunc per material

### LINKED ISSUE #4093: Expose depth compare functionality
To give developers full control, these depth compare functionality should be exposed:

```
     * - {@link FUNC_NEVER}: don't draw
     * - {@link FUNC_LESS}: draw if new depth < depth buffer
     * - {@link FUNC_EQUAL}: draw if new depth == depth buffer
     * - {@link FUNC_LESSEQUAL}: draw if new depth <= depth buffer
     * - {@link FUNC_GREATER}: draw if new depth > depth buffer
     * - {@link FUNC_NOTEQUAL}: draw if new depth != depth buffer
     * - {@link FUNC_GREATEREQUAL}: draw if new depth >= depth buffer
     * - {@link FUNC_ALWAYS}: always draw
```

at the moment only this is exposed:

https://github.com/playcanvas/engine/blob/d95908bbafd5b3a3e79cde3171942a1b364a9ec6/src/scene/renderer/forward-renderer.js#L1376-L1382

MrMabulous:
This change adds the material property 'depthFunc' that can be set to
any of the following values in order to control what comparison function
is used for the depth test during rendering:

FUNC_NEVER: don't draw
FUNC_LESS: draw if new depth < depth buffer
FUNC_EQUAL: draw if new depth == depth buffer
FUNC_LESSEQUAL: draw if new depth <= depth buffer
FUNC_GREATER: draw if new depth > depth buffer
FUNC_NOTEQUAL: draw if new depth != depth buffer
FUNC_GREATEREQUAL: draw if new depth >= depth buffer
FUNC_ALWAYS: always draw

previously depthFunc was always set to FUNC_LESSEQUAL, which is thus the
new default value for this property.

Fixes #4093


Maksims:
Personally, I would name it something more specific, as by reading it out of context e.g. "func less" - does not speak enough for itself.
`DEPTHFUNC_LESS` - would work better to give an idea of a constant out of context.

mvaligursky:
> DEPTHFUNC_LESS ...

I agree with you, but these constant names are existing constants in the engine. The PR only allows them to be set on material, but does not create any additional constant names.

PR #4099, opened 2022-03-09
Removed line from readme about preprocessor being ignored during source map generation

michael-graves-dojo:
@yaustar Follow up to #3952. Now that source maps are fixed this line in the readme doesn't seem accurate anymore.

Also I'm not sure where the publishing to code.playcanvas.com happens at? I looked through the github workflow and didn't notice anything. But it would be amazing to start publishing the map files there to match the specific versions of the engine.


willeastcott (approved):

yaustar:
> Also I'm not sure where the publishing to code.playcanvas.com happens at?

That's on our end, we have a separate job that uploads files during a release/build but thanks for looking into it.

Good catch on the japanese readme too

michael-graves-dojo:
> That's on our end, we have a separate job that uploads files during a release/build but thanks for looking into it.

@yaustar Ah gotcha, makes sense. Any possibility of getting that updated soon? We pull our branches from there currently and it would be really nice to pull the map files at the same time to help with debugging errors.

michael-graves-dojo:
Sorry, should have explained better. We are running an engine only build and so when we go to publish we are pulling playcanvas from code.playcanvas.com.

For example for our production build we pull the min.js from a specific version to have a reproducible setup:
https://code.playcanvas.com/playcanvas-1.51.7.min.js

For dev builds we pull the .dbg version instead

It would be really nice if we could also download the maps that match those versions, for example:
https://code.playcanvas.com/playcanvas-1.51.7.min.js.map

yaustar:
Ah, I see 🤔 Let me ask the team. It isn't something we had scheduled yet

yaustar:
@michael-graves-dojo I've scheduled this as part of our next regular 'blitz' week that we have. This will be in mid April, however, if we find spare bandwidth before then, we will try to fit this in

michael-graves-dojo:
@yaustar Awesome, thank you so much! Really appreciate it

PR #4104, opened 2022-03-10
[FIX] Don't use bitwise op to set boolean

willeastcott:
Interesting bug. The code actually works but it's just plain wrong. Bitwise `|=` is used to set a boolean. This means that `hasChanged` ends up being set to `0` or `1` instead of `true` or `false`. The original developer meant to use `||=` but frankly, I think my mini-rework here is easier on the eye.


mvaligursky (approved):
you could even inline those two lines inside the if directly, allowing the second part to not execute at all if first one returns true. But that would be less clean.

PR #4113, opened 2022-03-14
fix: Entity.Clone() should have return type Entity

GWattsJP:
Fixes #
fix: AnimationComponent assets should be of type `Asset[]|number[]`
fix: AnimationComponent animations should be of type `{[string]: Animation}`
fix: ScreenComponent.screen should be of type Entity
fix: ScriptComponent.scripts should be of type `ScriptType[]`
fix: MeshInstance.material should be of type Material

### DIFF
--- a/src/framework/components/script/component.js
+++ b/src/framework/components/script/component.js
@@ -73,7 +73,7 @@ class ScriptComponent extends Component {
      * An array of all script instances attached to an entity. This array is read-only and should
      * not be modified by developer.
      *
-     * @type {ScriptType[]}
+     * @param {ScriptType[]} value
      */
     set scripts(value) {
         this._scriptsData = value;
--- a/src/framework/entity.js
+++ b/src/framework/entity.js
@@ -587,7 +587,7 @@ class Entity extends GraphNode {
      * Create a deep copy of the Entity. Duplicate the full Entity hierarchy, with all Components
      * and all descendants. Note, this Entity is not in the hierarchy and must be added manually.
      *
-     * @returns {GraphNode} A new Entity which is a deep copy of the original.
+     * @returns {Entity} A new Entity which is a deep copy of the original.
      * @example
      * var e = this.entity.clone();
      *
--- a/src/scene/mesh-instance.js
+++ b/src/scene/mesh-instance.js
@@ -124,7 +124,7 @@ class MeshInstance {
          */
         this.node = node;           // The node that defines the transform of the mesh instance
         this._mesh = mesh;          // The mesh that this instance renders
-        mesh.incRefCount();
+        mesh.incRefCount(); 
         this.material = material;   // The material with which to render this instance
 
         this._shaderDefs = MASK_AFFECT_DYNAMIC << 16; // 2 byte toggles, 2 bytes light mask; Default value is no toggles and mask = pc.MASK_AFFECT_DYNAMIC
@@ -365,7 +365,7 @@ class MeshInstance {
     /**
      * The material used by this mesh instance.
      *
-     * @type {Material}
+     * @param {Material} material
      */
     set material(material) {
         for (let i = 0; i < this._shader.length; i++) {
--- a/types-fixup.mjs
+++ b/types-fixup.mjs
@@ -49,7 +49,8 @@ fs.writeFileSync(path, dts);
 
 const animationComponentProps = [
     ['activate', 'boolean'],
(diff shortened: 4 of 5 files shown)


src/framework/components/script/component.js
```diff
@@ -73,7 +73,7 @@ class ScriptComponent extends Component {
      * An array of all script instances attached to an entity. This array is read-only and should
      * not be modified by developer.
      *
-     * @type {ScriptType[]}
+     * @param {ScriptType[]} value
```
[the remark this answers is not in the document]
willeastcott: See: https://github.com/jsdoc/jsdoc/issues/973#issuecomment-90274714

src/scene/mesh-instance.js
```diff
@@ -124,7 +124,7 @@ class MeshInstance {
          */
         this.node = node;           // The node that defines the transform of the mesh instance
         this._mesh = mesh;          // The mesh that this instance renders
-        mesh.incRefCount();
+        mesh.incRefCount();
```
willeastcott, line 127:
```suggestion
        mesh.incRefCount();
```

src/scene/mesh-instance.js
```diff
@@ -365,7 +365,7 @@ class MeshInstance {
     /**
      * The material used by this mesh instance.
      *
-     * @type {Material}
+     * @param {Material} material
```
willeastcott, line 368: Again, this is a setter so you need to use `@type`.

src/framework/components/script/component.js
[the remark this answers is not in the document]
GWattsJP: With it set to @type the resulting type in the playcanvas.d.ts file goes back to 'any', it doesn't seem to pick it up as a property correctly.

Probably some other way to fix this, I'll have another go.

types-fixup.mjs
```diff
@@ -49,7 +49,8 @@ fs.writeFileSync(path, dts);
 
 const animationComponentProps = [
     ['activate', 'boolean'],
-    ['assets', 'any[]'],
+    ['assets', 'Asset[]|number[]'],
```
GWattsJP, line 52: This line seems to work but...

types-fixup.mjs
```diff
@@ -49,7 +49,8 @@ fs.writeFileSync(path, dts);
...
+    ['animations', '{[string]: Animation}'],
```
GWattsJP, line 53: This line doesn't, I end up with two definitions of the 'animations' property in the types file.

src/framework/entity.js
```diff
@@ -587,7 +587,7 @@ class Entity extends GraphNode {
      * Create a deep copy of the Entity. Duplicate the full Entity hierarchy, with all Components
      * and all descendants. Note, this Entity is not in the hierarchy and must be added manually.
      *
-     * @returns {GraphNode} A new Entity which is a deep copy of the original.
+     * @returns {Entity} A new Entity which is a deep copy of the original.
```
kungfooman, line 590:
```suggestion
     * @returns {this} A new Entity which is a deep copy of the original.
```

Reasoning by example:

```js
class UserEntity extends pc.Entity {
  userprop = 123;
  clone() {
    const ret = super.clone();
    ret.userprop = this.userprop;
    return ret;
  }
}
const a = new UserEntity();
const b = a.clone(); // Should be UserEntity and not Entity
```

The `Entity` method `_cloneRecursively(duplicatedIdsMap)` is also wrong, you need to change it to `new this.constructor` instead of `new Entity` (just like in `Material` class).

GWattsJP: All I'm changing in my pull is the return type to be Entity instead of GraphNode in order that the playcanvas types are correct.

I don't know if you can specify {this} in the typescript type annotations?

kungfooman: > I don't know if you can specify {this} in the typescript type annotations?

Sure, I tried to explain it here: https://github.com/playcanvas/engine/pull/4016

As soon as you try to extend from `Entity`, it will fail (as the example shall illustrate), but I know this fixes probably your use case.

willeastcott:
I've gone ahead and addressed three of the problems you highlight here in this PR: #4122

willeastcott:
The `Entity#clone` and `AnimationComponent#animations` issues are still outstanding (they're not so obvious).

willeastcott:
Closing sine #4122 and #4143 now address the fixes in this PR.

Issue #4115, opened 2022-03-14
Engine 1.52.3 - Error when exiting from VR


PeppeDK:
As you can see in the screenshot, i get this error when im going out from VR mode.
But everything works well if i move back to the engine version (1.51.7).

![err](https://user-images.githubusercontent.com/16405820/158192935-c3e9cda3-e564-4f0d-a9dc-ebac24472459.PNG)

Steps to reproduce:
- Create a new project with the "VR Starter Kit" (make sure the engine version is the latest)
- Run
- Press VR button
- Exit from VR by pressing "esc" button

And the error will appear.

slimbuck:
Error appears to originate from https://github.com/playcanvas/engine/blob/dev/src/xr/xr-input.js#L199

slimbuck:
Think it's a bug in https://github.com/playcanvas/engine/pull/4015:
`const session = this.manager.session;`
should be:
`const session = this.session;`
@willeastcott ?

slimbuck:
Referring to https://github.com/playcanvas/engine/pull/4015/files#diff-9aaa29403900068fbb88865bc7de283f2852676cf87574620ae4a7d57e024b03L180

PR #4121, opened 2022-03-15
[Fix] setting xr managed references to null after cleanup

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/4115 by cleaning references after end event, not before.


slimbuck:
We had planned on moving just _session to below the fire, seems a little less risky?

willeastcott (approved):

PR #4126, opened 2022-03-16
Cleaned up material references in render component # onRemove

### LINKED ISSUE #4124: Memory leak: RenderComponent with material asset keeps stale references on pc.app.assets._callbacks
### Description

When an entity in the scene, or an instantiated template, or a cloned entity, has a RenderComponent with a material, which is an asset, PlayCanvas adds some internal callbacks to `pc.app.assets._callbacks`.
For example, if I have an entity in the scene, with a RenderComponent, with the material with id `73244573` (**see picture 1**), various events are added to `pc.app.assets._callbacks`. That is, `add:73244573`, `load:73244573`, `remove:73244573` and `unload:73244573`. These callbacks contain a reference to both the asset, but also the RenderComponent (**see picture 2**). The problem is related to that if I call `entity.destroy()` on that entity, this callback still exists, which retains the reference to the RenderComponent, which is supposed to be destroyed. For example, if I clone the entity, and then immediately destroy it, say, 10.000 times (**see picture 3**), then the the `pc.app.assets._callbacks`-object retains references to 10.000 different destroyed RenderComponents. (**see picture 4**)
This seems like unintentional behaviour, and in the worst case, each of the 10.000 retained RenderComponents could have different unique materials which can't be garbage collected, because they initially had the same material asset.

### Steps to Reproduce

1. In a PlayCanvas project, set up a scene with an entity, with a RenderComponent
2. Make sure the entity is disabled
3. Create a material asset
4. Assign the material to the RenderComponent
5. Launch the game, and use the dev console to inspect `pc.app.assets._callbacks`. Notice that none of the callbacks have more than a couple of listeners.
6. In the editor, create a script in the scene, which has an attribute reference to the entity
7. In a 10.000-iteration for loop:
a. clone the entity `const clone = this.entityTemplate.clone();`
b. add the entity to the hierarchy `this.app.root.addChild(clone);`
c. set the entity to be enabled `clone.enabled = true;`
d. destroy the entity `clone.destroy();`
8. Launch the game, and inspect `pc.app.assets_callbacks` again. Notice that some of the events now have more than 10.000 listeners.

#### Picture 1
image
#### Picture 2
image
#### Picture 3
image
#### Picture 4
image

yaustar:
Fixes #4124

### DIFF
--- a/src/framework/components/render/component.js
+++ b/src/framework/components/render/component.js
@@ -666,6 +666,10 @@ class RenderComponent extends Component {
         this.asset = null;
         this.materialAsset = null;
 
+        for (let i = 0; i < this._materialReferences.length; i++) {
+            this._materialReferences[i].id = null;
+        }
+
         this.entity.off('remove', this.onRemoveChild, this);
         this.entity.off('insert', this.onInsertChild, this);
     }
--- a/src/framework/components/render/component.js
+++ b/src/framework/components/render/component.js
@@ -666,6 +666,8 @@ class RenderComponent extends Component {
         this.asset = null;
         this.materialAsset = null;
 
+        this._assetReference.id = null;
+
         for (let i = 0; i < this._materialReferences.length; i++) {
             this._materialReferences[i].id = null;
         }


mvaligursky (approved):

mvaligursky (approved):

yaustar:
Request re-review please @mvaligursky 

The EntityReference doesn't need to be manually cleaned up as it listens for the remove event on the component and cleans itself up. 

This does mean we have an inconsistency between EntityReference and AssetReference where one needs to be cleanup when the owner is destroy and the other manages itself.

mvaligursky (approved):

PR #4139, opened 2022-03-17
Add Vec3 Support for setFromEulerAngles() pc.Quat function

zishiwu123:
Fixes #3633

**Discussion**
I thought of two ways to instantiate `ex`, `ey`, and `ez` given the condition where `ex instanceof Vec3` is true:
1. Prematurely optimize by setting `ex = ex.x` so we don't have to make a clone of `ex`. A rule of thumb in software development is that "premature optimization is the root of all evil". However, since this is a game engine where optimization matters, I think it is okay to bend the rules. If reviewer says otherwise, I will make changes based on their suggestions.
```js
if (ex instanceof Vec3) {
    ez = ex.z;
    ey = ex.y;
    ex = ex.x;
}
```

2. Clone `ex` and use the cloned `Vec3` to instantiate `ex`, `ey`, and `ez`. Depending on the person, this might be easier to understand that option 1. But there is a small performance penalty because this creates an extra `Vec3` object.
```js
if (ex instanceof Vec3) {
    const v = ex.clone();
    ex = v.x;
    ey = v.y;
    ez = v.z;
}
```

### DIFF
--- a/src/math/quat.js
+++ b/src/math/quat.js
@@ -372,15 +372,28 @@ class Quat {
     /**
      * Sets a quaternion from Euler angles specified in XYZ order.
      *
-     * @param {number} ex - Angle to rotate around X axis in degrees.
+     * @param {number|Vec3} ex - Angle to rotate around X axis in degrees. If ex is a Vec3,
+     * the Vec3 will be used to populate the first 3 components.
      * @param {number} ey - Angle to rotate around Y axis in degrees.
      * @param {number} ez - Angle to rotate around Z axis in degrees.
      * @returns {Quat} Self for chaining.
      * @example
+     * // Create a quaternion from 3 euler angles
      * var q = new pc.Quat();
      * q.setFromEulerAngles(45, 90, 180);
+     *
+     * // Create the same quaternion from a vector containing the same 3 euler angles
+     * var v = new Vec3(45, 90, 180);
+     * var r = new pc.Quat();
+     * r.setFromEulerAngles(v);
      */
     setFromEulerAngles(ex, ey, ez) {
+        if (ex instanceof Vec3) {
+            ez = ex.z;
+            ey = ex.y;
+            ex = ex.x;
+        }
+
         const halfToRad = 0.5 * math.DEG_TO_RAD;
         ex *= halfToRad;
         ey *= halfToRad;
--- a/test/math/quat.test.mjs
+++ b/test/math/quat.test.mjs
@@ -486,16 +486,23 @@ describe('Quat', function () {
             it('sets the quaternion from ' + x + '°, ' + y + '°, ' + z + '°', function () {
                 const q1 = new Quat();
                 const q2 = new Quat();
+                const q3 = new Quat();
                 const m = new Mat4();
 
                 q1.setFromEulerAngles(x, y, z);
                 m.setFromEulerAngles(x, y, z);
                 q2.setFromMat4(m);
+                q3.setFromEulerAngles(new Vec3(x, y, z));
 
                 expect(q1.x).to.be.closeTo(q2.x, 0.0001);
                 expect(q1.y).to.be.closeTo(q2.y, 0.0001);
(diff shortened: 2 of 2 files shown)


Maksims:
No need to clone (it will lead to memory allocations, which then lead to minor/major GC stalls).
Code like this I believe is fine too, which eliminates the potential issue of someone changing the order of lines.

```javascript
if (ex instanceof Vec3) {
    const vec = ex;
    ex = vec.x;
    ey = vec.y;
    ez = vec.z;
}
```

zishiwu123:
Good idea @Maksims . Compound values are assigned by reference in JavaScript so it would not create a new copy of the object. Plus it's safer because the order in which ex, ey, and ez are assigned doesn't matter in your idea.

kungfooman:
Maybe simplify these four cases in `GraphNode` with this PR then:

https://github.com/playcanvas/engine/blob/main/src/scene/graph-node.js#L918

https://github.com/playcanvas/engine/blob/main/src/scene/graph-node.js#L1142

https://github.com/playcanvas/engine/blob/main/src/scene/graph-node.js#L1552

https://github.com/playcanvas/engine/blob/main/src/scene/graph-node.js#L1591

kungfooman:
Maybe simplify these four cases in `GraphNode` with this PR then:

https://github.com/playcanvas/engine/blob/main/src/scene/graph-node.js#L918

https://github.com/playcanvas/engine/blob/main/src/scene/graph-node.js#L1142

https://github.com/playcanvas/engine/blob/main/src/scene/graph-node.js#L1552

https://github.com/playcanvas/engine/blob/main/src/scene/graph-node.js#L1591

zishiwu123:
@kungfooman Good idea. I pushed another commit in this PR based on your suggestion.

kungfooman:
> @kungfooman Good idea. I pushed another commit in this PR based on your suggestion.

Why still check if its a `Vec3`, can't you just simplify the code completely now (removing all those `instanceof` lines)? By just always passing `x, y, z` to `Quat#setFromEulerAngles` and let only that method check if its a `Vec3`

Not sure if I'm missing something, I didn't test it. Good job!

src/math/quat.js
```diff
@@ -372,15 +372,29 @@ class Quat {
     /**
      * Sets a quaternion from Euler angles specified in XYZ order.
      *
-     * @param {number} ex - Angle to rotate around X axis in degrees.
+     * @param {number|Vec3} ex - Angle to rotate around X axis in degrees. If ex is a Vec3,
+     * the Vec3 will be used to populate the first 3 components.
      * @param {number} ey - Angle to rotate around Y axis in degrees.
      * @param {number} ez - Angle to rotate around Z axis in degrees.
```
willeastcott, line 378: You now need to mark `ey` and `ez` as optional.

zishiwu123: I noticed many optional parameters in functions such as https://github.com/playcanvas/engine/blob/main/src/resources/container.js#L158 are indicated by:
1. wrapped by square brackets in their name
2. having the word "optional" in the beginning of their comments

The functions in `graph-node.js` that I made changes to have indicator 1 but not indicator 2 for their optional parameters. Should I indicator 2 to ensure consistency? (in addition to marking `ey` and `ez`)

https://github.com/playcanvas/engine/blob/7e5ddc0e93cf3f054d0547a61b3d6af46c035eea/src/scene/graph-node.js#L906-L907

https://github.com/playcanvas/engine/blob/7e5ddc0e93cf3f054d0547a61b3d6af46c035eea/src/scene/graph-node.js#L1130-L1131

https://github.com/playcanvas/engine/blob/7e5ddc0e93cf3f054d0547a61b3d6af46c035eea/src/scene/graph-node.js#L1540-L1541

https://github.com/playcanvas/engine/blob/7e5ddc0e93cf3f054d0547a61b3d6af46c035eea/src/scene/graph-node.js#L1579-L1580

PR #4143, opened 2022-03-19
Use `this.constructor`, fix JSDoc types, add unit tests for clone

kungfooman:
Many classes can't be properly extended and cloned, so this is just fixing a bunch of these issues (math classes + GraphNode/Entity).

This allows developers e.g. to differentiate between points and vectors. Adding two vectors/directions creates another direction, but adding a point and a vector creates another point:

```js
class Point3 extends pc.Vec3 {}
a = new Point3;
b = a.clone();
b instanceof Point3; // `true` now
```

One case where it would be interesting to distinguish these differences is e.g. `Mat4#transformPoint` and `Mat4#transformVector` and this PR allows developers to write their own math abstractions using `instanceof Point3` e.g.

Another use case is to just add some special properties/functionality to entities:

```js
class UserEntity extends pc.Entity {
  userprop = 123;
  clone() {
    const ret = super.clone();
    ret.userprop = this.userprop;
    return ret;
  }
}
const a = new UserEntity();
const b = a.clone(); // This isn't an Entity anymore, but a UserEntity
```


willeastcott:
Interesting PR! Looks good. Except, let's look at mat4.js, for example:

```
    clone() {
        // @ts-ignore
        return new this.constructor().copy(this);
    }
```

We've never added `@ts-ignore` to the engine source before and I'd rather not do that. An alternative in this instance would be to do:

```
    clone() {
        /** @type {*} */
        const cstr = this.constructor;
        return new cstr().copy(this);
    }
```

kungfooman:
> We've never added `@ts-ignore` to the engine source before and I'd rather not do that. An alternative in this instance would be to do:

It would lose type information in `Curve#clone` and `CurveSet#clone`, but I like your idea for all other files:

image

Rather keep `@ts-ignore` for IntelliSense in these two cases? Kinda a TypeScript joke that such normal JavaScript patterns are still causing "errors". Reported since 7 years by now: https://github.com/Microsoft/TypeScript/issues/2081

willeastcott:
`Curve` and `CurveSet` should probably have a `copy` function, and could therefore clone like everything else.

willeastcott:
Hey @kungfooman - what's the plan here. It'd be cool to get this merge. Should the `Curve` and `CurveSet` changes be backed out for now?

kungfooman:
> Should the `Curve` and `CurveSet` changes be backed out for now?

Yea, that would be nice. I like your idea of adding `copy` methods and doing that properly requires a bit more work/thinking and a fresh PR would make it less confusing, since copying is a bit more complicated than cloning when it comes to internal arrays.

For example copying a `CurveSet` with 120 curves into a `CurveSet` with 80 curves would ideally:

 - `copy` first 80 `other` curves into existing curves
 - `push`/`clone` the last 40 `other` curves

This should reduce Garbage Collection slowdowns, if one were to use `copy` very often per frame (like some kind of dynamic/procedural animation system).

And since the `copy` method suddenly takes things like GC into account, maybe it would even be better to just keep the original/straight-forward `clone` method too, because it doesn't need to care about array resizing and less `if`'s usually makes things faster. 🤔 😅

(but I didn't test any of this yet, ideas/thoughts very welcome)

willeastcott (approved):

Issue #3638, opened 2021-10-30
Render asset as a collision mesh when container is not loaded


LeXXik:
We have a framework that loads a level, stored as a template. That level contains the level meshes, imported as a hierarchy (amazing feature, by the way, cudos!). They are then rigged with collision and rigidbody components, scripts, etc. Each level is loaded and instantiated via script.

On one of the levels I tried to add a piece that is a dynamic body with a render asset as a mesh collision. However, the app crashes, as soon as the level is loaded:

![image](https://user-images.githubusercontent.com/5677782/139555388-8715f95b-5294-46c9-b267-7b691a84e948.png)

![image](https://user-images.githubusercontent.com/5677782/139555427-68c49ca0-2a5a-4970-9786-f371aab81bba.png)

I tried to isolate it and create a blank project with that piece only in a template. However, the template is loaded fine and `_meshes` property is not null. I tried both instantiating the template via script and just add it in Editor. 

I am a bit stuck, and looking for advice on the conditions when `_meshes` property could be `null`, while `renderAsset` is not and having a correct reference id. After all the shape is recreated when the asset is loaded:

![image](https://user-images.githubusercontent.com/5677782/139555640-01800ca5-a5bd-4fef-8297-0d0e90f9a999.png)
![image](https://user-images.githubusercontent.com/5677782/139555556-82a9764c-d35b-439d-8e62-a4a3063ca135.png)

yaustar:
Dumb question from me: is the container asset that the render assets are referring to loaded yet?

LeXXik:
It's not a dumb question. This is what the issue is indicating. However, I thought that the shape is recreated on asset loaded callback. If the asset is loaded, then I assume mesh is available? I will add an explicit load to test it.

yaustar:
At a guess without investigating, the render asset is just a bit JSON that references the GLB container asset. So the load completion of the render asset is not dependent on waiting for the GLB container. Much like the sprite and texture atlas relationship.

LeXXik:
I did make sure now the framework loads the asset first. Still errors out. Actually this step is not needed, since collision system does load it automatically, if it finds a render asset is used:
![image](https://user-images.githubusercontent.com/5677782/139588481-f068c221-1403-4be4-af63-e01fbdb9e338.png)

However, later on, in `asset.ready` the callback is returned as soon as `asset.resource` property is found. Regardless of what that property contains:

[...]

PR #4144, opened 2022-03-19
Collision Component System update

LeXXik:
Updates Collision Component System to be in line with the other components.

- Moved shape implementations into own files.


LeXXik:
@willeastcott could you please advice on where the typescript gets duplicate declarations on the collision component from?

Maksims:
Through engine class naming, we rarely use shortening, such as: `CollisionSystemImpl`, also, because this is not instantiated by the developer, it is worth using the full name: `CollisionSystemImplementation`.

willeastcott:
@Maksims It was `CollisionSystemImpl` before right? This PR is just moving stuff around (plus reducing dependence on the `Data` class). I could also mention `Vec*`, `Mat*`, `Quat`. `Anim*`, etc. But yeah, as a private, internal class, I don't think it really matters too much in this instance.

LeXXik:
@Maksims I simply kept the original name, but this could be a good time to change it.

LeXXik:
Added a few tests for the Collision Component. There are currently 2 that are failing due to:
1. #3638
2. A compound trigger is failing to get created. Is it supported?

yaustar:
@willeastcott Is compound collision shapes for triggers supported? 

@LeXXik Can 1. be solved by preloading the GLB container first?

LeXXik:
Preloading it would work around the actual issue and pass the test case. I can remove the test case here, and it could be added to #3638 when a solution is found.

yaustar:
> Preloading it would work around the actual issue and pass the test case. 

I think that would be fine for this PR

LeXXik:
- Updated the mesh trigger test case. It was actually loading correctly, I simply didn't assign a correct render asset to the `renderAsset` field. Added an additional check on `btTriangleMesh` use.
- Removed compound trigger test case, as it is not supported by the engine at the moment.

LeXXik:
@willeastcott could you please let me know the status of this PR review? Are there some engine features missing that you wish to see before this PR can be merged/reviewed?

LeXXik:
Closing as not planned.

Issue #3967, opened 2022-01-31
Particle rendering Issue with Samsung Exynos CPU on Android 12


CMRDX:
After recieveing bugreports we've discovered that some of our games are not rendering at all on Samsung devices that are using a variant of the Exynos CPU. 
We specifically tested on the Samsung Galaxy S10e and S10+ running Android 12. The problem could not be reproduced on the S10+ with Android 11. 
This issue might be related to: 
* https://github.com/playcanvas/engine/issues/3460

It can be reproduced when opening the sample project I referenced in the linked issue. When we ran the games with WebGL 1 we did not encounter the bug. Remote debugging on the S10e leat to the following error log: 
[playcanvas.redox-interactive.com-1643626936209_PARTICLES.log](https://github.com/playcanvas/engine/files/7971706/playcanvas.redox-interactive.com-1643626936209_PARTICLES.log)

Ps.: [Forum Post](https://forum.playcanvas.com/t/rendering-issue-with-samsung-exynos-cpu-on-android-12/24116)

CMRDX:
@yaustar  it affects the whole game this time, not just the particle effects.

mvaligursky:
Is there some repro project (ideally a simple one) we could reproduce this?

CMRDX:
https://playcanvas.com/editor/project/828090 
This is the project we used to test the particle behaviour in the previously mentioned issue. It can be used to reproduce this problem as well.

yaustar:
While the particles are rendering incorrectly, the scene is still rendering for me. Is there a way we can repro the issue where the game is not rendering at all?

yaustar:
With the scene you linked above, we don't get the shader warnings/errors on device:

![image](https://user-images.githubusercontent.com/16639049/151954238-0124d006-dc46-4f2b-98b7-cbe4dca58027.png)

yaustar:
We are requesting a device with this spec that we can access via our device lab

CMRDX:
As a reference, these are the results we got on the sample project. (Samsung Galaxy S10+ Android 12)

![Screenshot_20220201-114944_Chrome](https://user-images.githubusercontent.com/75259672/151956055-6465fac2-f569-4b8b-90d1-3d845dff49c7.jpg)
![Screenshot_20220201-114901_Chrome](https://user-images.githubusercontent.com/75259672/151956061-90675518-7830-4b3d-80b8-36c4232b4090.jpg)

yaustar:
[...]

Issue #3460, opened 2021-09-07
Particle system broken on Samsung Galaxy S10


CMRDX:
We are currently experiencing that the PlayCanvas particle system behaves incorrectly on specific Samsung devices, namely Galaxy S10, Galaxy S10+ and Galaxy S10e with Android 11 and Chrome 93.0.4577.62

The corresponding Forum post can be found here (including videos and a link to our test project):
https://forum.playcanvas.com/t/particle-system-broken-on-samsung-galaxy-s10/21787

mvaligursky:
Possibly related to https://github.com/playcanvas/engine/issues/3415

yaustar:
FYI: Cannot reproduce on a Galaxy Fold 3

yaustar:
I couldn't reproduce this on the Browserstack S10+

Wondering if it's specific to the Exynos CPU and not the Snapdragon variant.

@CMRDX Are you devices the Snapdragon or the Exynos variant?

CMRDX:
The device seen in the Video (S10e), as well as the S10+ is featuring a Exynos 9820 2.73GHz. 
![rn_image_picker_lib_temp_d859749c-cb0a-459d-b8d2-f43db297eb33](https://user-images.githubusercontent.com/75259672/132355714-4f2512ac-4f3a-4564-865d-eb66ac631ce5.jpg)

yaustar:
That may explain why my Fold 3 doesn't have the issue, it's using a Snapdragon 🤔

leonidaspir:
I've tried it on a Samsung G10 with the same CPU and I am getting the same issue on the project provided on the forum.

yaustar:
@CMRDX or @leonidaspir Can you try running this with WebGL 1 forced, ie: https://launch.playcanvas.com/1225034?debug=true&webgl1=true

leonidaspir:
Works as expected when running with WebGL1.

CMRDX:
@yaustar We got the same result as @leonidaspir. Seems to work as expected when enforcing WebGL 1.

yaustar:
Thanks, that helps with extra information/debugging.

We are working on getting a device to replicate the issue.

yaustar:
Tried on multiple Exynos CPUs https://github.com/playcanvas/engine/issues/3415#issuecomment-921687853

No joy in reproducing the issue :(

marcusx2:
[...]

Issue #3415, opened 2021-08-18
Particles renders incorrect on Samsung M31


Sarletor:
### Description

Particles renders incorrect on:
**Samsung M31
Android 10.0
CPU - Samsung Exynos 9611** 

Should be this: 
![image](https://user-images.githubusercontent.com/37019196/129930315-8d76cb3d-d4c0-4be0-acbb-ebf122de1c1b.png)
On Samsung M31:
![image](https://user-images.githubusercontent.com/37019196/129930428-c1c14782-e272-43b0-b73f-55e94eecc02b.png)
![image](https://user-images.githubusercontent.com/37019196/129930516-0dca55b2-402a-4dc2-89f5-142a75a1e35f.png)

Project: https://playcanvas.com/editor/project/823129

mvaligursky:
Does it support WebGL2?

Sarletor:
Yes

Sarletor:
![image](https://user-images.githubusercontent.com/37019196/130095731-ce087c19-7b16-4ba4-b974-9b456c701179.png)

yaustar:
I've been to the office and unfortunately all the Samsung devices that we have are Snapdragon CPUs and don't showcase the issue :( (S10, S10+ and S7)

dexterdeluxe88:
> I've been to the office and unfortunately all the Samsung devices that we have are Snapdragon CPUs and don't showcase the issue :( (S10, S10+ and S7)

@yaustar thanks for checking at the office. What's your proposed solution for the given problem then? Maybe I can send you a device - its overseas, but possible.

yaustar:
@dexterdeluxe88 Trying to see if our remote device lab can source a device despite being in the US. I'm also going to have a look if I can find an exact model number and source one in the UK

yaustar:
Just to get an idea of urgency, is this a blocking issue?

yaustar:
Testing project: https://playcanv.as/e/p/RWgZ7YlT/ on a range of Exynos devices I've found

Not reproducible on:
SM-A730F - Exynos 7885 (Android 7.1.1)
SM-G935F - Exynos 8890 (Android 7)
SM-J260M - Exynos 7570
SM-A520W - Exynos 7880
GT-I9500 - Exynos 5410

yaustar:
The S10 has a Exynos 9820 and the M31 has a Exynos 9611 so maybe this is super specific to the 9 series CPU?

yaustar:
[...]

PR #4147, opened 2022-03-21
Workaround for Samsung Exynos particle rendering/crash

yaustar:
Added supportsGpuParticles flag to graphics device which checks against the Samsung model name and ARM chipset (exynos)

A search on GSM Arena for 'SM-' shows only Samsung devices so I'm reasonable confident there will be minimal false negatives where CPU is used for particles on non-Samsung devices.

https://www.gsmarena.com/results.php3?sQuickSearch=yes&sName=sm-

Worksaround:
https://github.com/playcanvas/engine/issues/3415
https://github.com/playcanvas/engine/issues/3460
https://github.com/playcanvas/engine/issues/3967

### DIFF
--- a/src/graphics/webgl/webgl-graphics-device.js
+++ b/src/graphics/webgl/webgl-graphics-device.js
@@ -749,6 +749,7 @@ class WebglGraphicsDevice extends GraphicsDevice {
      */
     initializeExtensions() {
         const gl = this.gl;
+        const userAgent = typeof navigator !== 'undefined' ? navigator.userAgent : "";
 
         const supportedExtensions = gl.getSupportedExtensions();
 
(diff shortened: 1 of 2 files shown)


src/graphics/webgl/webgl-graphics-device.js
```diff
@@ -823,7 +824,18 @@ class WebglGraphicsDevice extends GraphicsDevice {
...
+        this.unmaskedVendor = ext ? gl.getParameter(ext.UNMASKED_VENDOR_WEBGL) : '';
+
         this.supportsInstancing = !!this.extInstancing;
+
+        // Check if we support GPU particles. At the moment, Samsung devices with Exynos (ARM) either crash or render
+        // incorrectly when using GPU for particles. See:
+        // https://github.com/playcanvas/engine/issues/3967
+        // https://github.com/playcanvas/engine/issues/3415
+        const samsungModelRegex = /SM-[a-zA-Z0-9]+\)/;
+        this.supportsGpuParticles = !(this.unmaskedVendor === 'ARM' && userAgent.match(samsungModelRegex));
```
yaustar, line 838: Should this be moved to `initializeCapabilities` instead? I placed it here because I saw `this.supportsInstancing` being done here but looking at it again, it makes more sense to move to `initializeCapabilities` call instead?

mvaligursky: yes, I assumed that's where it should go too.

yaustar: Moved!

mvaligursky (approved):

PR #4148, opened 2022-03-22
Add retry option for glue and fallback downloading methods

### LINKED ISSUE #4078: Basis glue and fallback downloading methods aren't retrying
See line: https://github.com/playcanvas/engine/blob/dev/src/resources/basis.js#L53

Where it uses http.get, the option for retries is not passed in compared to other resource loaders. eg https://github.com/playcanvas/engine/blob/dev/src/resources/template.js#L20

zishiwu123:
Fixes #4078

Adds `maxRetries`. Consistent with: https://github.com/playcanvas/engine/blob/main/src/resources/scene-utils.js#L14.

### DIFF
--- a/src/resources/basis.js
+++ b/src/resources/basis.js
@@ -18,7 +18,7 @@ const getCompressionFormats = (device) => {
 };
 
 // download basis code and compile the wasm module for use in workers
-const prepareWorkerModules = (config, callback) => {
+const prepareWorkerModules = (config, maxRetries, callback) => {
     const getWorkerBlob = () => {
         const code = '(' + BasisWorker.toString() + ')()\n\n';
         return new Blob([code], { type: 'application/javascript' });
@@ -45,12 +45,18 @@ const prepareWorkerModules = (config, callback) => {
         });
     };
 
+    const options = {
+        responseType: 'blob',
+        retry: maxRetries > 0,
+        maxRetries: maxRetries
+    };
+
     if (config.glueUrl && config.wasmUrl && wasmSupported()) {
         let basisCode = null;
         let module = null;
 
         // download glue script
-        http.get(config.glueUrl, { responseType: 'blob' }, (err, response) => {
+        http.get(config.glueUrl, options, (err, response) => {
             if (err) {
                 callback(err);
             } else {
@@ -98,7 +104,7 @@ const prepareWorkerModules = (config, callback) => {
             compileManual();
         }
     } else {
-        http.get(config.fallbackUrl, { responseType: 'blob' }, (err, response) => {
+        http.get(config.fallbackUrl, options, (err, response) => {
             if (err) {
                 callback(err, null);
             } else {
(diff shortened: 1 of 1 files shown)


willeastcott:
What about this line:

https://github.com/playcanvas/engine/blob/5c481172523186023c18caa264c5e5473c4045ca/src/resources/basis.js#L302

zishiwu123:
> What about this line:
> 
> https://github.com/playcanvas/engine/blob/5c481172523186023c18caa264c5e5473c4045ca/src/resources/basis.js#L302

Good point. However the `basisInitialize()` function that calls `prepareWorkerModules()` currently does not have a `maxRetries` parameter. We could create a parameter with a default value like this, but I'm not sure if that's the right way to go.
https://github.com/playcanvas/engine/blob/5c481172523186023c18caa264c5e5473c4045ca/src/resources/loader.js#L277

willeastcott:
Possibly. @slimbuck - what do you think?

willeastcott:
I'd suggest making `maxRetries` a property of the config object passed into `basisInitialize` (defaulting to 5 maybe?). But yeah, I'll leave it to @slimbuck to provide some thoughts...

slimbuck:
> I'd suggest making `maxRetries` a property of the config object passed into `basisInitialize` (defaulting to 5 maybe?). But yeah, I'll leave it to @slimbuck to provide some thoughts...

I agree. We want `maxRetries` added to the basis config. In `prepareWorkerModules`, if `maxRetries` is missing, default to 5.

slimbuck:
Thanks for this @zishiwu123 it's awesome!

src/resources/basis.js
```diff
@@ -292,6 +300,7 @@ function basisInitialize(config) {
 
         config.rgbPriority = config.rgbPriority || defaultRgbPriority;
         config.rgbaPriority = config.rgbaPriority || defaultRgbaPriority;
+        config.maxRetries = config.maxRetries || defaultMaxRetries;
```
slimbuck, line 303:
```suggestion
        config.maxRetries = config.hasOwnProperty('maxRetries') ? config.maxRetries : defaultMaxRetries;
```

Since 0 is a valid maxRetries.

src/resources/basis.js
```diff
@@ -250,6 +257,7 @@ let initializing = false;
  * @param {string[]} [config.rgbaPriority] - Array of texture compression formats in priority order
  * for textures with alpha. The supported compressed formats are: 'astc', 'atc', 'dxt', 'etc1',
  * 'etc2', 'pvr'.
+ * @param {number} [config.maxRetries] - Number of http load retry attempts.
```
slimbuck, line 260:
```suggestion
 * @param {number} [config.maxRetries] - Number of http load retry attempts. Defaults is 5.
```

zishiwu123:
@slimbuck Good catch pointing out that `config.maxRetries = 0` is a valid value but the expression `(config.maxRetries || defaultMaxRetries)` would have evaluated to false and used `defaultMaxRetries` when it shouldn't have.

willeastcott (approved):

slimbuck (approved):
🥳 nice one!

willeastcott:
Seconded. 😄 Thanks for the contribution @zishiwu123!

PR #3068, opened 2021-04-07
WebXR Plane Detection

Maksims:
Implementation of an experimental draft spec: https://immersive-web.github.io/real-world-geometry/plane-detection.html

It is available in Chrome for Android 89, with enabled chrome://flags#webxr-incubations flag.

Plane detection provides information about real-world surfaces estimations in form of planes.
Each plane can have its position, rotation and a list of edge points making a complex polygon.

### New APIs:
```js
// pc.XrManager
app.xr.planeDetection // interface to access plane detection
app.xr.start(camera, ..., ..., {
    planeDetection: true
}); // new option to enable plane detection

// pc.XrPlaneDetection
planeDetection.supported // true if plane detection is supported
planeDetection.available // true if plane detection is available
planeDetection.planes // a list of pc.XrPlane's

planeDetection.on('available', () => { }); // fired when plane detection becomes available, due to spec design it is not possible to know if it is available before a first XR session update
planeDetection.on('unavailable', () => { }); // fired when plane detection becomes unavailable, for example when session is ended
planeDetection.on('add', (plane) => { }); // fired when a new plane is added
planeDetection.on('remove', (plane) => { }); // fired when a plane is removed

// pc.XrPlane
plane.id // numerical id of a plane
plane.orientation // 'horizontal' or 'vertical' orientation of a plane, or null if any other orientation
plane.points // a list of edge points of a plane. Their coords are in local space to a plane
plane.getPosition() // get world position of a plane
plane.getRotation() // get world rotation of a plane

plane.once('remove', () => { }); // fired when plane is removed
plane.on('change', () => { }); // fired when plane orientation or points have been changed
```

### Test project:
https://playcanvas.com/project/782753/overview/webxr-plane-detection
This project builds mesh, updates it when the plane is changed and draw lines for each detected plane.

### Test build:
https://playcanv.as/p/f2ESRGge/
Ensure you have chrome://flags#webxr-incubations enabled

### Video:
https://twitter.com/mrmaxm/status/1379802592163864581

thumb

I confirm I have signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #4149, opened 2022-03-22
Public AssetListLoader & examples update

ellthompson:
This PR brings back the AssetListLoader class which allows an array of assets to be loaded before the 'ready' callback is called. This class is made public and includes unit tests.

The examples browser now makes use of the AssetListLoader for asset loading in all examples. The examples browser therefore no longer needs to hoist the application instantiation in each example in order to preload example assets in.

New Public API:
```javascript
// Create a new AssetListLoader using a list of assets to load and the asset registry used to load and manage them.
pc.AssetListLoader.constructor(assetList: pc.Asset[], assetRegistry: pc.AssetRegistry): void

// Destroys the AssetListLoader.
pc.AssetListLoader.destroy(): void

// Start loading asset list, call done() when all assets have loaded or failed to load.
pc.AssetListLoader.load(done: Function, scope: ?object): void

// Sets a callback which will be called when all assets in the list have been loaded.
pc.AssetListLoader.ready(done: Function, scope: ?object): void
```


package.json
```diff
@@ -115,7 +115,7 @@
     "docs": "jsdoc -c conf-api.json",
     "lint": "eslint --ext .js,.mjs extras scripts src test rollup.config.js",
     "serve": "serve build -l 51000",
-    "test": "mocha --recursive --require test/fixtures.mjs",
+    "test": "mocha --recursive --require test/fixtures.mjs \"test/asset/asset-list-loader.test.mjs\"",
```
willeastcott, line 118: I'm guessing you need to back out this change?

src/asset/asset-list-loader.js
```diff
@@ -0,0 +1,214 @@
...
+import { Asset } from './asset.js';
+
+/**
+ * @class
+ * @name AssetListLoader
+ * @augments EventHandler
+ * @classdesc Used to load a group of assets and fires a callback when all assets are loaded.
+ * @param {Asset[]|number[]} assetList - An array of {@link Asset} objects to load or an array of Asset IDs to load.
+ * @param {AssetRegistry} assetRegistry - The application's asset registry.
+ */
```
willeastcott, line 12: This is the 'old-style' JSDoc comment block. Look at other classes to see how we do this now. 😄

src/asset/asset-list-loader.js
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    /**
+     * @function
+     * @name AssetListLoader#load
+     * @description Start loading asset list, call done() when all assets have loaded or failed to load.
+     * @param {Function} done - Callback called when all assets in the list are loaded. Passed (err, failed) where err is the undefined if no errors are encountered and failed contains a list of assets that failed to load.
+     * @param {object} [scope] - Scope to use when calling callback.
+     *
+     */
```
willeastcott, line 75: Again, this is 'old-style'. Please update to the awesome new ES6-style.

src/asset/asset-list-loader.js
```diff
@@ -0,0 +1,214 @@
...
+    load(done, scope) {
+        var i = 0;
+        var l = this._assets.length;
+        var asset;
```
willeastcott, line 79: We don't use `var` any more in engine source (except JSDoc comments - although even that should go soon).

src/asset/asset-list-loader.js
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    /**
+     * @function
+     * @name AssetListLoader#ready
+     * @param {Function} done - Callback called when all assets in the list are loaded.
+     * @param {object} [scope] - Scope to use when calling callback.
+     */
+    ready(done, scope) {
+        scope = scope || this;
```
willeastcott, line 117: Default parameter please.

src/asset/asset-list-loader.js
```diff
@@ -0,0 +1,214 @@
...
+            if (this._callback) {
+                this._callback.call(this._scope);
+            }
+            this.fire("load", this._assets);
+        }
+    }
+
+    // called when an (any) asset is loaded
+    _onLoad(asset) {
+        var self = this;
```
willeastcott, line 150: This wouldn't be needed if the `setTimeout` below used an arrow function.

src/asset/asset-list-loader.js
```diff
@@ -0,0 +1,214 @@
...
+            // asset load events
+            setTimeout(function () {
+                self._loadingComplete(self._failed);
+            }, 0);
+        }
+    }
+
+    // called when an asset fails to load
+    _onError(err, asset) {
+        var self = this;
```
willeastcott, line 170: This wouldn't be needed if the `setTimeout` below used an arrow function.

src/asset/asset-list-loader.js
willeastcott: There are several other instances in this file. Please migrate them all. There is an ESLint rule that prevents use of `var` which I want to turn on but we need to upgrade the JSDoc occurrences first.

src/asset/asset-list-loader.js
```diff
@@ -0,0 +1,214 @@
...
+                if (asset) {
+                    this._assets.push(asset);
+                } else {
+                    this._waitForAsset(assetList[i]);
+                }
+            });
+        }
+    }
+
+    destroy() {
```
willeastcott, line 53: Should this be public API?

src/asset/asset.js
```diff
@@ -70,7 +70,7 @@ class Asset extends EventHandler {
      * field or null if hash was set (e.g from using {@link AssetRegistry#loadFromUrl}).
      * @param {ArrayBuffer} [file.contents] - Optional file contents. This is faster than wrapping
      * the data in a (base64 encoded) blob. Currently only used by container assets.
-     * @param {object} [data] - JSON object with additional data about the asset. (e.g. for texture
+     * @param {object|string} [data] - JSON object or string with additional data about the asset. (e.g. for texture
```
willeastcott, line 73: Note that all JSDoc comments fit in 100 columns now (I formatted the entire codebase that way). Does this line now go over that?

test/asset/asset-list-loader.test.mjs
```diff
@@ -0,0 +1,255 @@
...
+        });
+        it('adds the supplied registry to any assets that do not have one', function () {
+            const assets = [
+                new Asset('model', 'container', { url: `${assetPath}test.glb` })
+            ];
+            expect(assets[0].registry).to.be.equal(null);
+            const assetListLoader = new AssetListLoader([assets[0], assets[0]], app.assets);
+            expect(assetListLoader._assets[0].registry).to.be.equal(app.assets);
+        });
+    });
```
willeastcott, line 59: Minor point but the whitespace formatting around `describe`/`it` blocks doesn't match the other tests. Please try to match the formatting elsewhere.

package-lock.json
```diff
@@ -41,9 +41,9 @@
       }
     },
     "node_modules/@ampproject/remapping": {
-      "version": "2.1.1",
-      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.1.1.tgz",
-      "integrity": "sha512-Aolwjd7HSC2PyY0fDj/wA/EimQT4HfEnFYNp5s9CQlrdhyvWTtvZ5YzrUPu6R6/1jKiUlxu8bUhkdSnKHNAHMA==",
+      "version": "2.1.2",
+      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.1.2.tgz",
+      "integrity": "sha512-hoyByceqwKirw7w3Z7gnIIZC3Wx3J484Y3L/cMpXFbr7d9ZQj2mODrirNzcJa+SM3UlpWXYvKV4RlRpFXlWgXg==",
```
willeastcott, line 46: Since the `package.json` NPM dependencies have not changed, I'm curious why the `package-lock.json` has been updated.

willeastcott:
Looks good to me, in general. I've left a bunch of comments. However, should we take a moment to review the `AssetListLoader` API? Do we like the naming? Is the public interface properly established? Do the function signatures make sense (e.g. the constructor taking the asset registry as the second param instead of the first)? Should this functionality just be a function on the `AssetRegistry` directly instead?

Maksims:
Would it be possible to write a new API in [similar format](https://github.com/playcanvas/engine/pull/3068) in the first post of this PR?

examples/scripts/example-data.mjs
```diff
@@ -33,6 +33,9 @@ fs.readdirSync(`${MAIN_DIR}/src/examples/`).forEach(function (category) {
         exampleData[category][example].javaScriptFunction = Prettier.format(Babel.transform(exampleData[category][example].typeScriptFunction, { retainLines: true, filename: `transformedScript.tsx`, presets: ["typescript"] }).code, { parser: BabelParser.parse, tabWidth: 4 });
         exampleData[category][example].nameSlug = example;
         exampleData[category][example].categorySlug = category;
+        const files = formatters.retrieveStaticObject(exampleFileText, 'FILES');
+        // eslint-disable-next-line no-eval
+        if (files) exampleData[category][example].files = eval('(' + files + ')');
```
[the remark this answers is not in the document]
willeastcott: Yeah, this is the kind of thing that lgtm.com often complains about. Just generally, @ellthompson, it might be worth reviewing the kinds of warnings/errors reporting there regarding the examples browser:
https://lgtm.com/projects/g/playcanvas/engine/alerts/?mode=list (not necessarily for this PR)

package-lock.json
ellthompson: This is probably due to me updating node locally to v16.14.0. Shall I back out this change?

examples/scripts/example-data.mjs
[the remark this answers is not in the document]
ellthompson: This code is never executed on the client, only during the built step in a node environment.

ellthompson:
> Would it be possible to write a new API in [similar format](https://github.com/playcanvas/engine/pull/3068) in the first post of this PR?

Sure, I'e added this in.

package.json
ellthompson: Yep, done 👍

test/asset/asset-list-loader.test.mjs
ellthompson: Done 👍

src/asset/asset-list-loader.js
ellthompson: I'm not sure why this change isn't showing up but it's been added: https://github.com/playcanvas/engine/pull/4149/commits/fb1b9b088b656cf815626df121582a83896815ff

mvaligursky:
```
AssetListLoader.load(done: Function, scope: ?object): void
AssetListLoader.ready(done: Function, scope: ?object): void
```

I just wonder if the `ready` function is needed? It does not seem to expose anything additional to the `load` function. Is this being used by the examples?

ellthompson:
> ```
> AssetListLoader.load(done: Function, scope: ?object): void
> AssetListLoader.ready(done: Function, scope: ?object): void
> ```
> 
> I just wonder if the `ready` function is needed? It does not seem to expose anything additional to the `load` function. Is this being used by the examples?

It's not used by the examples but it allows other parts of an application to subscribe to the loading of these assets without triggering the load itself.

src/asset/asset-list-loader.js
```diff
@@ -0,0 +1,215 @@
...
+        this._loaded = false;
+        this._count = 0; // running count of successfully loaded assets
+        this._failed = []; // list of assets that failed to load
+
+        this._waitingAssets = [];
+
+        if (assetList.length && assetList[0] instanceof Asset) {
+            assetList.forEach((asset, i, array) => {
+                // filter out duplicates
+                if (array.indexOf(asset) !== i) {
```
mvaligursky, line 36: isn't this always true?

mvaligursky: ah I see .. it tests if this was in the array already.

examples/src/examples/graphics/area-lights.tsx
```diff
@@ -1,196 +1,197 @@
...
+        const assets = {
+            'color': new pc.Asset('color', 'texture', { url: '/static/assets/textures/seaside-rocks01-color.jpg' }),
+            'normal': new pc.Asset('normal', 'texture', { url: '/static/assets/textures/seaside-rocks01-normal.jpg' }),
+            'gloss': new pc.Asset('gloss', 'texture', { url: '/static/assets/textures/seaside-rocks01-gloss.jpg' }),
+            'statue': new pc.Asset('statue', 'container', { url: '/static/assets/models/statue.glb' }),
+            'luts': new pc.Asset('luts', 'binary', { url: '/static/assets/binary/area-light-luts.bin' }),
+            'helipad.dds': new pc.Asset('helipad.dds', 'cubemap', { url: '/static/assets/cubemaps/helipad.dds' }, { type: pc.TEXTURETYPE_RGBM })
+        };
+
+        const assetListLoader = new pc.AssetListLoader(Object.values(assets), app.assets);
```
mvaligursky, line 23: IE11 does not implement Object.values .. so just making sure we have some way to make this work (polyfill or transpilation)

src/asset/asset-list-loader.js
```diff
@@ -0,0 +1,215 @@
...
+            setTimeout(() => {
+                this._loadingComplete(this._failed);
+            }, 0);
+        }
+    }
+
+    // called when a expected asset is added to the asset registry
+    _onAddAsset(asset) {
+        // remove from waiting list
+        const index = this._waitingAssets.indexOf(asset);
```
mvaligursky, line 193: _waitingAssets should be a Set to avoid indexOf and splice.

src/asset/asset-list-loader.js
```diff
@@ -0,0 +1,215 @@
...
+        this._assets.push(asset);
+        const l = this._assets.length;
+        for (let i = 0; i < l; i++) {
+            asset = this._assets[i];
+            if (!asset.loaded) {
+                this._loadingAssets.add(asset);
+                this._registry.load(asset);
+            }
+        }
+    }
```
mvaligursky, line 207: it feels odd this function does not only handle the asset that was added, but loops over all assets - is that required?

src/asset/asset-list-loader.js
```diff
@@ -0,0 +1,215 @@
...
+     */
+    load(done, scope) {
+        const l = this._assets.length;
+        let asset;
+
+        this._count = 0;
+        this._failed = [];
+        this._callback = done;
+        this._scope = scope;
+        this._loaded = false;
```
mvaligursky, line 93: this block suggests the function can be called multiple times, but I'm not sure it should, it that it would work?

src/asset/asset-list-loader.js
```diff
@@ -0,0 +1,215 @@
...
+     *     new pc.Asset("texture2", "texture", { url: 'http://example.com/my/assets/here/texture2.png') })
+     * ], pc.app.assets);
+     */
+    constructor(assetList, assetRegistry) {
+        super();
+        this._assets = [];
+        this._loadingAssets = new Set();
+        this._registry = assetRegistry;
+        this._loaded = false;
+        this._count = 0; // running count of successfully loaded assets
```
mvaligursky, line 28: I think we could remove _count, and just simply remove assets when done (successfully or not) from _loadingAssets. When the set is empty, we're done.

src/asset/asset-list-loader.js
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    // called when all assets are loaded
+    _loadingComplete() {
+        if (this._loaded) return;
+        this._loaded = true;
+        this._registry.off("load", this._onLoad, this);
+        this._registry.off("error", this._onError, this);
+
+        if (this._failed && this._failed.length) {
```
mvaligursky, line 141: no need to check 'this._failed' as it's initialized to []

examples/scripts/iframe/index.mustache
```diff
@@ -117,17 +92,17 @@
...
             }
 
-            return app;
-        }
-
-        function hasBasisAssets (assets) {
-        for (let i = 0; i < assets.length; i++) {
-            if (assets[i].url && assets[i].url.includes('.basis')) {
-                return true;
+            if (location.pathname.indexOf('misc/mini-stats') === -1) {
```
mvaligursky, line 95: could we add another variable to the example to control this instead of here? We already have NAME and CATEGORY and few more. For example 
`static MINISTATS = false;`

examples/src/examples/graphics/area-lights.tsx
ellthompson: Yep this example code is transpiled.

src/asset/asset-list-loader.js
ellthompson: I've made this function return with a debug message if it's called multiple times. I've also added in a test case for multiple function calls.

src/asset/asset-list-loader.js
```diff
@@ -0,0 +1,215 @@
...
+            if (this._callback) {
+                this._callback.call(this._scope, "Failed to load some assets", this._failed);
+            }
+            this.fire("error", this._failed);
+        } else {
```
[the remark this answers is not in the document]
ellthompson: The load event isn't called when it's failed but the load functions callback is invoked directly with the failed assets. This can be used to retry the load. I've added in a test case for this.

src/asset/asset-list-loader.js
```diff
@@ -0,0 +1,195 @@
+import { EventHandler } from '../core/event-handler.js';
+
+import { Asset } from './asset.js';
+
+/**
+ * Used to load a group of assets and fires a callback when all assets are loaded.
```
mvaligursky, line 6: It'd be fantastic to have an example here in JSDocs format on how to use it .. including how to retry when some assets fail to load

src/asset/asset-list-loader.js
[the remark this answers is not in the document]
mvaligursky: That sounds good .. I added a comment to provide an example on how to use this. Should we provide some maxRetries parameter or is this somehow already available for assets loaded this way?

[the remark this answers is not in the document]
ellthompson: The AssetListLoader just acts as a wrapper for Asset loading, so any parameters should be set on the assets themselves before they're passed into the list loader.

ellthompson:
> I'm not sure it works like that .. all browsers apart from IE11 implement the function, and that is accessible in ES5 as well .. we don't need a polyfill for ES5, but for IE11. Perhaps check ES5 version of the engine built - i likely just calls this function directly.

Yeah you're right, i've included this in the list of example browser polyfills.

package-lock.json
ellthompson: I've backed this change out.

mvaligursky (approved):
approving, nicely done

PR #4153, opened 2022-03-26
[FIX] Ensure consistent function return types

willeastcott:
There are several places in the codebase where a function has inconsistent return types. These problems are spotted by the ESLint rule [`consistent-return`](https://eslint.org/docs/rules/consistent-return). This PR attempts to fix these problems with a view to enabling this rule in the future.


willeastcott:
Remaining problems that still need fixing:

image

kungfooman:
Should the build step replace `return undefined;` with `return;` again to save the bytes? Probably doesn't matter too much after zipping. This would also remove some bytes from these existing cases:

image

Another option that helps with detecting this is to enable `"strictNullChecks": true` in `jsconfig.json`, for example:

image

willeastcott:
Yeah, I think such a micro-optimization is probably not worth it. 😄  I'll take a look at the `strictNullChecks` setting - thanks!

src/framework/app-base.js
```diff
@@ -1471,11 +1471,11 @@ class Application extends EventHandler {
      * @returns {object} A object containing the values calculated to use as width and height.
      */
     resizeCanvas(width, height) {
-        if (!this._allowResize) return; // prevent resizing (e.g. if presenting in VR HMD)
+        if (!this._allowResize) return undefined; // prevent resizing (e.g. if presenting in VR HMD)
```
mvaligursky, line 1474: I think these should be null instead of undefined .. as the return type is object.

src/scene/batching/batch-manager.js
```diff
@@ -142,8 +142,8 @@ class BatchManager {
         }
 
         if (this._batchGroups[id]) {
-            Debug.error(`batch group with id ${id} already exists`);
-            return;
+            Debug.error(`Batch group with id ${id} already exists.`);
+            return undefined;
```
mvaligursky, line 146: maybe it could return `this._batchGroups[id]` so that the code would still work

mvaligursky (approved):

src/framework/app-base.js
willeastcott: Possibly. I'm mainly just trying to keep the functionality identical and satisfy ESLint.

src/scene/batching/batch-manager.js
willeastcott: Again, it'd be changing functionality. I'd rather address that separately.

PR #4154, opened 2022-03-27
Enforce single quotes and disallow `var` in unit tests

willeastcott:
* Enforce single quotes for all node-based unit tests (note that single quotes are used over double quotes in our codebase by over 2 to 1). I would like to see this standardized across the codebase.
* Migrate all occurrences of `var` to `let` and `const`.

### DIFF
--- a/test/.eslintrc.json
+++ b/test/.eslintrc.json
@@ -3,6 +3,7 @@
         "mocha": true
     },
     "rules": {
-        "no-unused-expressions": "off"
+        "no-unused-expressions": "off",
+        "quotes": ["error", "single"]
     }
 }
--- a/test/.eslintrc.json
+++ b/test/.eslintrc.json
@@ -4,6 +4,7 @@
     },
     "rules": {
         "no-unused-expressions": "off",
+        "no-var": "error",
         "quotes": ["error", "single"]
     }
 }
--- a/test/anim/evaluator/anim-evaluator.test.mjs
+++ b/test/anim/evaluator/anim-evaluator.test.mjs
@@ -19,31 +19,31 @@ describe('AnimEvaluator', function () {
         const app = new Application(canvas);
 
         // build the graph to be animated
-        var parent = new GraphNode('parent');
-        var child1 = new GraphNode('child1');
-        var child2 = new GraphNode('child2');
+        const parent = new GraphNode('parent');
+        const child1 = new GraphNode('child1');
+        const child2 = new GraphNode('child2');
 
         app.root.addChild(parent);
         parent.addChild(child1);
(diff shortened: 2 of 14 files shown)


slimbuck:
Perhaps we should standardise on ` instead?

willeastcott:
@slimbuck That's a reasonable suggestion. However, I kind of like reserving ` for template string literals. For example, say there was a long string and I saw it started with a `, I would then have to scan the whole string to find the embedded variables which may or may not be there.

slimbuck:
Yeah makes sense. ' is really similar to ` - perhaps " would be a better standard?

willeastcott:
I mainly went for single because:

* It appears twice as many times in our source as double (so we'll generate fewer changes when we standardize, plus people obviously prefer single over double if that's the statistic)
* single is one keypress, double is two. :wink:

Yeah, double is visually differentiated a bit more. But does that outweigh the above? I'm not totally sold.

slimbuck:
I would say readability of the codebase is more important than either the one-off change or double keystroke.

slimbuck:
(and tbh I think there's a big difference in readability between " and ' vs `).

willeastcott:
Actually, just remembered that we unofficially stick to Google's JS coding standards for the most part:

https://google.github.io/styleguide/jsguide.html#features-strings-use-single-quotes

So that's another key reason why you'll find way more singles than doubles right now. We just never enforced it.

mvaligursky:
Personally I tend to use " for these main reasons, but no super strong preference:
- I find it easier to read, as ' vs ` is getting less obvious
- I'm used to it from c++ and c# and it feels right. My mind is wired to use ' for a single char type only
- the plugin (or built in functionality?) that VSCode has for managing imports uses " in import statements: `import { Vec3 } from "../../math/vec3.js";
`

willeastcott:
> the plugin (or built in functionality?) that VSCode has for managing imports uses " in import statements: import { Vec3 } from "../../math/vec3.js";

If I'm following what you're talking about here, I believe VS Code is smart enough to check the quotes of other imports in the file and try to match them.

quotes

mvaligursky:
Great! Sounds like that one is a non issue then.

Maksims:
A single quote is easier to type, as no extra button press is required.
Template literals should be used only where it is required, linters in IDE can even highlight strings that use ` but do not utilize it with actual templating.

willeastcott:
Haha, yeah, I can relate to that. I have a US keyboard at home and a UK keyboard in the office!

slimbuck (approved):
We chatted internally and decided single quote is the way to go.

PR #4155, opened 2022-03-28
Partial migration to single quotes

willeastcott:
This PR performs a partial migration of the `src` and `extras` folders to standardize on single quotes (in preference to double quotes). This represents 344 occurrences of double quotes. This PR excludes `src/framework`, `src/graphics`, `src/resources` and `src/scene` for now. Instead, this migration will be done in phases to minimize the impact of anyone working on PRs in those folders.


slimbuck (approved):

kungfooman:
What do you think about doing this for JSDoc types aswell? For example:

```js
    /**
     * Add Attribute.
     *
     * @param {string} name - Name of an attribute.
     * @param {object} args - Object with Arguments for an attribute.
     * @param {("boolean"|"number"|"string"|"json"|"asset"|"entity"|"rgb"|"rgba"|"vec2"|"vec3"|"vec4"|"curve")} args.type - Type
     */
```

Could be:

```js
    /**
     * Add Attribute.
     *
     * @param {string} name - Name of an attribute.
     * @param {object} args - Object with Arguments for an attribute.
     * @param {'boolean'|'number'|'string'|'json'|'asset'|'entity'|'rgb'|'rgba'|'vec2'|'vec3'|'vec4'|'curve'} args.type - Type
     */
```

I also don't know the reason for the extra brackets around the unions, imo it would look way cleaner without them.

Simple query to find all cases in any IDE using file search: `"|"`

PR #4156, opened 2022-03-31
Preprocessor to resolve ifdefs is applied to shader programs

mvaligursky:
A new private class is added to preprocess defines and ifdefs in the shader with the goal of removing unused code. This is a first preparation step. Following steps will further modify the shader code, in order to convert uniforms to uniform blocks.

**New (private) API:**
```
/**
 * Run c-like proprocessor on the source code, and resolves the code based on the defines and ifdefs
 *
 * @param {string} source - The source code to work on.
 * @returns {string|null} Returns preprocessed source code.
 */
static Preprocessor.run(source) {
```

**Note** - Simple expression evaluation only is supported for ifdefs, handles cases:
- expression
- defined(expression)
- !defined(expression)
 
 It does not handle more complex cases, which would require more complex system:
 - defined(A) || defined(B)
And so few use cases in the shaders were converted to the supported format.

Unit test has also been added.

Additionally, Debug logging functionality has been extended with **trace** function.
**New (private) API (and also examples of use):**
```
Debug.setTrace(`Preprocessor`, true);
Debug.trace(`Preprocessor`, "message")
```

and this is displayed in the console:
`[Preprocessor]  message`


yaustar:
Does this affect the sourcemap generation?

mvaligursky:
> Does this affect the sourcemap generation?

This is only used to process shaders in memory .. so no.

willeastcott:
This is definitely something for @slimbuck to review - I'll pass.

src/core/preprocessor.js
```diff
@@ -0,0 +1,256 @@
+import { Debug } from './debug.js';
+
+// set to true to enable debug logging
+const enableLogging = false;
```
slimbuck, line 4: I love this idea, really useful. It does mean the log calls in this file remain in production code though.

Perhaps we could embrace a more general approach to trace logging?

We could add `Debug.trace` function which takes a named "stream" parameter and allow user to enable/disable streams at runtime. Calls to `Debug.trace` (and the stream enable calls) would be removed by rollup as usual.

src/core/preprocessor.js
```diff
@@ -0,0 +1,256 @@
...
+ * @ignore
+ */
+class Preprocessor {
+    /**
+     * Run c-like proprocessor on the source code, and resolves the code based on the defines and ifdefs
+     *
+     * @param {string} source - The source code to work on.
+     * @returns {string} Returns preprocessed source code.
+     */
+    static run(source) {
```
slimbuck, line 53: Shouldn't this function return success or failure?

src/core/preprocessor.js
mvaligursky: Yep I thought about it few times as well .. and now is a good time to do it. I'll add description to this PR.

src/graphics/program-lib/chunks/lit/clusteredLight.frag.js
```diff
@@ -3,6 +3,17 @@ uniform sampler2D clusterWorldTexture;
 uniform sampler2D lightsTexture8;
 uniform highp sampler2D lightsTextureFloat;
 
+// complex ifdef expression are not supported, handle it here
+// defined(CLUSTER_COOKIES) || defined(CLUSTER_SHADOWS)
+#if defined(CLUSTER_COOKIES)
+    #define CLUSTER_COOKIES_OR_SHADOWOWS
+#endif
+#if defined(CLUSTER_SHADOWS)
+    #ifndef CLUSTER_COOKIES_OR_SHADOWOWS
```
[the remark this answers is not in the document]
mvaligursky: I assumed those are needed .. but clearly not. Removed, thanks.

src/core/preprocessor.js
mvaligursky: it's little error validation taking place in this function, but I return error (null) and handle it on the other side correctly now (there was an unrelated bug in shader compilation returning incorrect type)

mvaligursky:
Updates based on comments @slimbuck 
@willeastcott - could you please look at the Debug.trace part I added.

rollup.config.js
```diff
@@ -136,6 +136,7 @@ const stripOptions = {
         'Debug.warn',
         'Debug.error',
         'Debug.log',
+        'Debug.trace',
```
slimbuck, line 139: Is possible (/does it make sense) to remove `setTrace` and `_traceChannels` too?

slimbuck (approved):
We'll probably update the preprocessor over time, but this is a reasonable first pass.

rollup.config.js
mvaligursky: _traceChannels is automatically stripped (the same as the existing other set).
setTrace is an empty function in non debug build which I'd like to keep in .. allowing the code to compile in non-debug builds as well.

slimbuck: Wouldn't the rollup plugin remove references in the code to 'Debug.setTrace' as well as the function itself?

slimbuck: Isn't that an issue for `Debug.trace` anyway?

mvaligursky: well, Debug.trace is to be used inside the engine only .. but Debug.setTrace is called from outside to enabled the tracing. That's my plan on how this should be used.

slimbuck: I see, makes sense 👍

PR #4157, opened 2022-03-31
Preallocate `Float32Array` in `calculateNormals`/`calculateTangents`

kungfooman:
This is my first attempt at optimizing these two methods, since they became a bottle neck for procedural mesh generation:

Without PR:

image

With PR:

image

I don't know if anyone relies upon returning of an `Array` instead of a `Float32Array`. Technically an `Array` is 64-bit double values and the length can be manipuliated afterwards, but does anyone need that?

I didn't test this barely at all yet and I have more ideas how it could be optimized. Would anyone want to add unit tests for normals/tangents to make sure its still correct?

### DIFF
--- a/src/scene/procedural.js
+++ b/src/scene/procedural.js
@@ -18,6 +18,31 @@ const primitiveUv1PaddingScale = 1.0 - primitiveUv1Padding * 2;
 const shapePrimitives = [];
 
 /**
+ * Normalizes the entire array in place.
+ * 
+ * @example ```js
+ *   let arr = [0,10,0, 20,20,20, 0,1,30];
+ *   normalizeArray(arr);
+ *   arr.map(_ => _.toFixed(3));
+ *   // Output: ['0.000', '1.000', '0.000', '0.577', '0.577', '0.577', '0.000', '0.033', '0.999']
+ * ```
+ * @param {number[] | Float32Array} arr
+ */
+function normalizeArray(arr) {
+    const n = arr.length;
+    const sqrt = Math.sqrt;
+    for (let i = 0; i < n; i += 3) {
+        const nx = arr[i];
+        const ny = arr[i + 1];
+        const nz = arr[i + 2];
+        const invLen = 1 / sqrt(nx * nx + ny * ny + nz * nz);
+        arr[i] *= invLen;
+        arr[i + 1] *= invLen;
+        arr[i + 2] *= invLen;
+    }
+}
+
+/**
  * Generates normal information from the specified positions and triangle indices. See
  * {@link createMesh}.
  *
@@ -32,9 +57,15 @@ const shapePrimitives = [];
 function calculateNormals(positions, indices) {
     const indicesLength = indices.length;
     const positionsLength = positions.length;
-    const p1 = new Vec3();
-    const p2 = new Vec3();
-    const p3 = new Vec3();
+    let p1x = 0;
+    let p1y = 0;
+    let p1z = 0;
+    let p2x = 0;
+    let p2y = 0;
+    let p2z = 0;
+    let p3x = 0;
+    let p3y = 0;
+    let p3z = 0;
     const p1p2 = new Vec3();
     const p1p3 = new Vec3();
     const faceNormal = new Vec3();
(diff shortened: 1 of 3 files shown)


src/scene/procedural.js
```diff
@@ -39,36 +39,31 @@ function calculateNormals(positions, indices) {
...
-        const i2 = indices[i * 3 + 1];
-        const i3 = indices[i * 3 + 2];
+        const i1 = indices[i * 3    ] * 3;
+        const i2 = indices[i * 3 + 1] * 3;
+        const i3 = indices[i * 3 + 2] * 3;
 
-        p1.set(positions[i1 * 3], positions[i1 * 3 + 1], positions[i1 * 3 + 2]);
-        p2.set(positions[i2 * 3], positions[i2 * 3 + 1], positions[i2 * 3 + 2]);
-        p3.set(positions[i3 * 3], positions[i3 * 3 + 1], positions[i3 * 3 + 2]);
+        p1.set(positions[i1], positions[i1 + 1], positions[i1 + 2]);
```
mvaligursky, line 50: I would not even set these on p1, p2 and p3, but directly evaluate subtraction of them to p1p2 and p1p2, something along
`p1p2.x = positions[i2] - positions[i1];

src/scene/procedural.js
```diff
@@ -114,20 +109,28 @@ function calculateTangents(positions, normals, uvs, indices) {
     const tan1 = new Float32Array(vertexCount * 3);
     const tan2 = new Float32Array(vertexCount * 3);
 
-    const tangents = [];
+    const tangents = new Float32Array(vertexCount * 4);
```
mvaligursky, line 112: should this be`*4`?

src/scene/procedural.js
```diff
@@ -17,192 +17,295 @@ const primitiveUv1PaddingScale = 1.0 - primitiveUv1Padding * 2;
...
- * @returns {number[]} An array of 3-dimensional vertex normals.
+ * @param {number[]|Uint16Array|Float32Array|Int8Array|Uint8Array|Int16Array|Int32Array|Uint32Array} positions - An array of 3-dimensional vertex positions.
+ * @param {number[]|Uint16Array|Float32Array|Int8Array|Uint8Array|Int16Array|Int32Array|Uint32Array} indices - An array of triangle indices.
+ * @returns {Float32Array} An array of 3-dimensional vertex normals.
  * @example
  * var normals = pc.calculateNormals(positions, indices);
  * var tangents = pc.calculateTangents(positions, normals, uvs, indices);
  * var mesh = pc.createMesh(positions, normals, tangents, uvs, indices);
  */
 function calculateNormals(positions, indices) {
```
querielo, line 57: Do we use IIFE to store temporal variables in the engine to be garbage-collector-friendly?

For example,

```js
const calculateNormals = (() => {
  const vec3Tmp = new pc.Vec3();
  const vec3Tmp2 = new pc.Vec3();
  
  const quatTmp = new pc.Quat();
  
  return function calculateNormals(positions, indices) {
    // here is the main code
    const p1p2 = vec3Tmp;
    const p1p3 = vec3Tmp2;
    // lalilulelo
   }
})()
```

kungfooman: Hey @querielo, nice and useful idea, we used to have that style for ES5 prototype style methods, but ES6 doesn't support it (sadly). For a normal function like this it still works of course.

Why do you do `const p1p2 = vec3Tmp` tho, don't you like closure variables? Or any problem with just naming the temporary variable `p1p2` in the first place?

querielo: @kungfooman

> but ES6 doesn't support it (sadly).

Could you explain what do you mean?

The next pattern works well:
```js
class Foo {
    bar = (() => {
        const vec3Tmp = new pc.Vec3();

        return function bar() {
             // work with vec3Tmp
        }
    })()
}
```

So, here we store temporal variable in closure that is created by IIFE.

Pros:
1) garbage-collection-freindly
2) `vec3Tmp` is defined as close to `bar` as possible. Less probability that somebody else reuses `vec3Tmp` and can break other function

Cons:
1) It's hard to read if you don't know the pattern.
2) A new instance of Foo create a new function **bar**.
3) `bar` is stored in instance of Foo. It cannot be overridden or extended by subclasses as usual methods. The method can be overridden in the next way:

```js
class Bar extends Foo {
    bar = (() => {
        const barFromFoo = this.bar

        return function bar() {
            barFromFoo()

            // write your extension
        }
    })()
}
```

**NOTE: If we are talking about functions (not methods) then the second and third cons are not valid.**

I think, it may be more appropriate to open a new issue to address this topic, as it seems to be unrelated to the current pull request.

kungfooman: Well, usually it was just:

```js
function Test() {
    // nothing yet
}
Test.prototype.hai = (() => {
    const oldNow = Date.now();
    return function bar() {
        console.log("Test#hai> oldNow", oldNow)
    }
})();
const test = new Test();
test.hai();
```

But the ES6 rewrite can't handle that, so the temporary variables just went into the local file scope (or into static class variables in some other cases [1]).

[1] https://github.com/playcanvas/engine/blob/main/src/framework/anim/evaluator/anim-target-value.js#L17

I basically agree with your pros/cons.

Syntax that comes closest to ES6 might be:

```js
function Test() {
    // nothing yet
}
Object.assign(Test.prototype, {
    hai: (() => {
        const oldNow = Date.now();
        return function bar() {
            console.log("Test#hai> oldNow", oldNow)
        }
    })(),
    
    bye() {
        console.log("bye");
    }
});
const test = new Test();
test.hai();
test.bye();
```

But rewriting everything just to replace some tmp vectors in local file scope via IIFE methods is kind of overkill for classes (for functions I still think it's a good idea, as you proposed).

Do you want to answer my initial two questions in previous post about closures?

mvaligursky: @kungfooman - any update on this PR or should we close it for now?

kungfooman:
@mvaligursky I could see quite a speed-up (which I needed for runtime decal generation), but ideally we would have unit tests testing generated normals and tangents, since I had to change a bit of code.

So maybe someone would add some unit tests and another team member takes the time to review it.

However if there is no interest, we can of course just close it, since I can also see downsides with this PR:

1) A `Float32Array` is not using 64-bit precision as compared to `number[]`
2) `Float32Array` cannot be resized either, so it actually removes certain earlier possibilities
3) Ideally I would want to remove the GC allocation entirely and not just make it faster (by simply using a `result` input array as usual)

mvaligursky:
Thanks for the update. I've extracted the easy bit here: https://github.com/playcanvas/engine/pull/5177
And will close this till somebody has time to work on this. It looks really promising, but as you said, it needs more work.

PR #4171, opened 2022-04-04
support different sized env atlas

slimbuck:
This PR adds a little math to support generating env atlases larger than 512.

For more context see https://forum.playcanvas.com/t/prefiltered-dds-resolution-issue/25035

### DIFF
--- a/src/graphics/env-lighting.js
+++ b/src/graphics/env-lighting.js
@@ -224,14 +224,16 @@ class EnvLighting {
 
         DebugGraphics.pushGpuMarker(device, "mipmaps");
 
+        const s = result.width / 512;
+
         // generate mipmaps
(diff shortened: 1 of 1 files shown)


mvaligursky (approved):

slimbuck:
> On a bit different topic, but related. Will it be reasonable to provide a setting for prefiltering CubeMap to different resolutions?

Hi @Maksims! I think it does make sense.

Unfortunately we still store prefiltered data in uncompressed rgba8 DDS cubemap. A 128x128 prefiltered cubemap comes in at 512k. 256x256 would take 2MB!

Ideally we'd generate and store the env-atlas directly and store as PNG. Advantage is simplicity and png compression, however some browsers still drop bits when alpha channel is used.

PR #4173, opened 2022-04-04
Line rendering - avoid spread operator on large arrays

mvaligursky:


LeXXik:
@mvaligursky I am still unable to render large arrays of lines. Not sure if the issue is the same or a new one, let me know if you want a new issue.

image

mvaligursky:
That means the number of colors does not match the number of positions. I'd double check you're passing the array of equal length to the line drawing. But could be something else, in which case a repro would be great.

LeXXik:
I will debug it on my side first then.

PR #4175, opened 2022-04-05
AppBase - additional refactor

mvaligursky:


src/framework/app-base.js
```diff
@@ -522,7 +524,7 @@ class AppBase extends EventHandler {
          *     // VR is available
          * }
          */
-        this.xr = new XrManager(this);
+        this.xr = new createOptions.xr(this);
```
willeastcott, line 527: This feels weird. Why not just pass an `XrManager` instance to the app?

src/deprecated/deprecated.js
```diff
@@ -115,6 +115,8 @@ import { basisInitialize } from '../resources/basis.js';
 import { EventHandler } from '../core/event-handler.js';
 import { Asset } from '../asset/asset.js';
 
+import { VrManager } from '../vr/vr-manager.js';
```
willeastcott, line 118: I actually think we should just strip out WebVR support completely. https://caniuse.com/webvr

src/framework/app-base.js
mvaligursky: most of the others are done the same way - we just assign the type, but not constructed object.

```
        createOptions.lightmapper = Lightmapper;
        createOptions.batchManager = BatchManager;

        createOptions.componentSystems = [
            RigidBodyComponentSystem,
            CollisionComponentSystem,
            JointComponentSystem,
           ...
```

this allows us to hide constructor / modify constructor parameters in the future without affecting user's code. And allows us to instantiate those in some fixed order we need.

src/deprecated/deprecated.js
mvaligursky: ah that bad? Ok, maybe I can do that in a separate PR? Are there any old projects that depend on it? I assume not, considering it would not work in a browser at all.

willeastcott: https://github.com/playcanvas/engine/issues/4176

willeastcott:
The naming for `AppCreateOptions` suggests to me it's used on creation (i.e. in the constructor). So what about `AppInitOptions` or just `AppOptions`?

src/framework/app-base.js
willeastcott: I think it works nicely for component systems and resource handles. But yeah, feels weird for the `XrManager` and `Lightmapper`.

willeastcott (approved):
I've left a couple of comments laying out some concerns I have. But approving for now since I don't want to delay anything. BTW, I think this is a really nice step forwards overall!!! 🚀

src/framework/app-base.js
mvaligursky: makes sense .. I felt strange about it .. especially since GraphicsDevice and SoundManager are created using new, and others not. I'll unify that soon, and use `new` on all those.

PR #4177, opened 2022-04-06
Remove WebVR API support

### LINKED ISSUE #4176: Remove WebVR API support from the engine
Today, everyone should be targeting AR/VR via the engine's WebXR support. WebVR is now only supported by Firefox on Windows:

https://caniuse.com/webvr

I believe that now is a suitable time to strip this functionality from the engine codebase. This would involve removing the files in `src/vr` and stripping the creation and management of the `VrManager` from the `Application`.

I would move a minimal set of API into `deprecated.js` that always returns false for availability of WebVR so existing apps just fail to enter VR (via WebVR) instead of crashing.

Hey @Maksims, do you see any problem with this?

mvaligursky:
Fixes #4176


willeastcott (approved):
This warms my ❤️

yaustar:
Would be worth doing at a tweet about this ahead of time so users are sort of warned ahead of time if anyone is still depending on it with the polyfill

PR #4179, opened 2022-04-07
AppBase handles missing XR correctly

mvaligursky:
small fixes to handle app.xr being null

### DIFF
--- a/src/framework/app-base.js
+++ b/src/framework/app-base.js
@@ -516,7 +516,10 @@ class AppBase extends EventHandler {
          *     // VR is available
          * }
          */
-        this.xr = new appOptions.xr(this);
+        this.xr = null;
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

src/framework/app-base.js
```diff
@@ -516,7 +516,10 @@ class AppBase extends EventHandler {
          *     // VR is available
          * }
          */
-        this.xr = new appOptions.xr(this);
+        this.xr = null;
+        if (appOptions.xr) {
+            this.xr = new appOptions.xr(this);
+        }
```
slimbuck, line 522:
```suggestion
        this.xr = appOptions.xr ? new appOptions.xr(this) : null;
```

PR #4184, opened 2022-04-08
Turn off Antialiasing on webkit 15.3 to avoid rendering issues

tizkzks96:
Update #4141 

The same issue appeared in 15.3 and added as a workaround.

### DIFF
--- a/src/graphics/webgl/webgl-graphics-device.js
+++ b/src/graphics/webgl/webgl-graphics-device.js
@@ -259,7 +259,7 @@ class WebglGraphicsDevice extends GraphicsDevice {
         // #4136 - turn off antialiasing on AppleWebKit browsers 15.4
         if (typeof navigator !== 'undefined') {
             const ua = navigator.userAgent;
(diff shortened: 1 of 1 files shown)


slimbuck:
It seems likely this multisampling issue also applies to offscreen render targets (see https://github.com/playcanvas/model-viewer/issues/165). Might be worth setting maxSamples on the device to 1 as well? @mvaligursky

mvaligursky:
Is there some repro of the issue on iOS 15.3? All related webkit tickets list this as 15.4 only.

https://bugs.webkit.org/show_bug.cgi?id=238022
https://bugs.webkit.org/show_bug.cgi?id=237906
https://bugs.webkit.org/show_bug.cgi?id=237006
https://bugs.webkit.org/show_bug.cgi?id=238196

tizkzks96:
Although we did not report any bugs, an issue occurred in 15.3 in our service.

mvaligursky:
To go ahead with disabling of antialiasing for 15.3 devices, we'd need to have an issue to investigate and confirm this is the right workaround. It's possible there is some other bug, possibly in your application, causing the issue.

PR #4188, opened 2022-04-08
enable support for retrying loading of container assets

vfernandestoptal:
While tracking down an issue with my project, I noticed that playcanvas was not automatically retrying loading of GLB files when there was a network failure, like it does for other assets like texture files for instance.

This PR updates the `ContainerHandler` class to support updating `maxRetries` and forward it to any configured parsers, including the `GlbParser`, to which it delegates the loading of the asset by default.

### DIFF
--- a/test/asset/asset-registry.test.mjs
+++ b/test/asset/asset-registry.test.mjs
@@ -3,22 +3,29 @@ import { Asset } from '../../src/asset/asset.js';
 import { AssetRegistry } from '../../src/asset/asset-registry.js';
 import { GlbContainerResource } from '../../src/resources/parser/glb-container-resource.js';
 import { ResourceLoader } from '../../src/resources/loader.js';
+import { http, Http } from '../../src/net/http.js';
 
 import { HTMLCanvasElement } from '@playcanvas/canvas-mock';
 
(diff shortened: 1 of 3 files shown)


vfernandestoptal:
hey, any update on when this might get merged and released?

test/asset/asset-registry.test.mjs
```diff
@@ -178,6 +185,15 @@ describe('AssetRegistry', function () {
             });
         });
 
+        it('supports retry loading of container assets', function (done) {
+            sinon.spy(http, 'request');
```
slimbuck, line 189: did you mean to leave sinon in here?

vfernandestoptal: yes, so it can assert on call count

willeastcott (approved):

test/asset/asset-registry.test.mjs
slimbuck: I see, thanks. @willeastcott just explained to me that sinon populates `http.request.callCount` below.

slimbuck (approved):
Thanks so much for this PR!

PR #4193, opened 2022-04-11
Envatlas mipmap fix

slimbuck:
We generate a 1-pixel border around equirect mipmaps in the envatlas and must be careful not to rely on smaller mips which are predominantly mipmap border (i.e. sizes 1x1, 2x2, 4x4).

This PR removes the last level of mipmap we were relying on from the shiny reflections, because it was causing unruly aliasing on thin sliver polygons.

Here is a closeup of a vehicle under harsh lighting conditions demonstrating the aliasing before and after:

env-aliasing


willeastcott (approved):
Perhaps the greatest PR ever developed by humankind.

PR #4197, opened 2022-04-14
Fixes some WebGL types

kungfooman:
First I just wanted to fix this line:

```js
            location = gl.getUniformLocation(glProgram, info.name);

            shaderInput = new ShaderInput(device, info.name, device.pcUniformType[info.type], location);
```

`location` is a `WebGLUniformLocation`, while `ShaderInput` just treats it as a `number`. So maybe `ShaderInput` should at some point also have its own `Webgl` wrapper? And then it escalated to get the other types right.

API: https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/getUniformLocation

### DIFF
--- a/src/graphics/webgl/webgl-shader.js
+++ b/src/graphics/webgl/webgl-shader.js
@@ -5,7 +5,7 @@ import { Preprocessor } from '../../core/preprocessor.js';
 import { ShaderInput } from '../shader-input.js';
 import { SHADERTAG_MATERIAL, semanticToLocation } from '../constants.js';
 
-/** @typedef {import('../graphics-device.js').GraphicsDevice} GraphicsDevice */
+/** @typedef {import('./webgl-graphics-device.js').WebglGraphicsDevice} WebglGraphicsDevice */
 /** @typedef {import('../shader.js').Shader} Shader */
 
 /**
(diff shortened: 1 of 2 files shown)


src/graphics/webgl/webgl-shader.js
```diff
@@ -34,6 +34,8 @@ class WebglShader {
      * @param {Shader} shader - The shader to free.
      */
     destroy(shader) {
+        /** @type {WebglGraphicsDevice} */
+        // @ts-ignore
```
willeastcott, line 38: Please don't add `@ts-ignore`s to the codebase (at least for now)

willeastcott:
After merging this, the docs build process has errors:

image

kungfooman:
Thank you for pointing that out @willeastcott 

I traced it a bit and it comes from https://www.npmjs.com/package/catharsis

Testing expression in node:

```node
> catharsis = require("catharsis");
> catharsis.parse("number", {jsdoc: true}); // working example
> catharsis.parse("[number, number]", {jsdoc: true}); // throws this error
```

Luckily the other pages generate without a problem so far.

I tried this kind of declaring a tuple too: https://stackoverflow.com/questions/64739315/how-to-document-tuple-using-jsdoc

But for some reason that throws the same error (`npm run docs`)

So I have to see how to make `catharsis` accept a tuple or just reverting that tuple.

PR #4203, opened 2022-04-20
[Anim Component] Fix for the normalizeWeights property

### LINKED ISSUE #4189: [1.53.0] Setting anim normalizeWeights to true means that the base animation layer cannot be found
Repo: https://playcanvas.com/editor/scene/1385366
Launch: https://launch.playcanvas.com/1385366?debug=true&version=1.53.0

Code:
```
KeyboardControls.prototype.initialize = function() {
    this.app.keyboard.on(pc.EVENT_KEYDOWN, this.keyDown, this);
    this.app.keyboard.on(pc.EVENT_KEYUP, this.keyUp, this);

    this.entity.anim.normalizeWeights = true;
    var locomotionLayer = this.entity.anim.findAnimationLayer('Base');
    if (!locomotionLayer) {
        console.error('Can\'t find animation layer');
    }
};
```

Trying to find the Base layer after setting normalizeWeights to true fails

ellthompson:
Previously when setting the normalizeWeights property, the state graph was reset in order to unbind any dirty targets it contained. This PR moves the target unbinding to its own function so that the normalizeWeights setter can call it directly. It also removes the resetStateGraph function which is no longer necessary.

Fixes #4189 

### DIFF
--- a/src/framework/components/anim/component.js
+++ b/src/framework/components/anim/component.js
@@ -456,7 +456,7 @@ class AnimComponent extends Component {
         const assetId = this._stateGraphAsset;
         this.removeStateGraph();
         if (assetId) {
-            const stateGraph = this.system.app.assets.get(this.stateGraphAsset).resource;
+            const stateGraph = this.system.app.assets.get(assetId).resource;
             this.loadStateGraph(stateGraph);
         }
     }
--- a/src/framework/components/anim/component.js
+++ b/src/framework/components/anim/component.js
@@ -106,7 +106,7 @@ class AnimComponent extends Component {
      */
     set normalizeWeights(value) {
         this._normalizeWeights = value;
-        this.resetStateGraph();
+        this.unbind();
     }
 
     get normalizeWeights() {
@@ -447,16 +447,15 @@ class AnimComponent extends Component {
         this._layerIndices = {};
         this._parameters = {};
         this._playing = false;
-        Object.keys(this._targets).forEach((targetKey) => {
-            this._targets[targetKey].unbind();
-        });
+        this.unbind();
         // clear all targets from previous binding
         this._targets = {};
     }
 
     resetStateGraph() {
+        const assetId = this._stateGraphAsset;
         this.removeStateGraph();
-        if (this.stateGraphAsset) {
+        if (assetId) {
             const stateGraph = this.system.app.assets.get(this.stateGraphAsset).resource;
             this.loadStateGraph(stateGraph);
         }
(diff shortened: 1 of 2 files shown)


src/framework/components/anim/component.js
```diff
@@ -447,16 +447,15 @@ class AnimComponent extends Component {
...
-        Object.keys(this._targets).forEach((targetKey) => {
-            this._targets[targetKey].unbind();
-        });
+        this.unbind();
         // clear all targets from previous binding
         this._targets = {};
     }
 
     resetStateGraph() {
+        const assetId = this._stateGraphAsset;
```
yaustar, line 456: Was this tested against the model component when it isn't playing?

src/framework/components/anim/component.js
```diff
@@ -447,16 +447,15 @@ class AnimComponent extends Component {
...
         this.removeStateGraph();
-        if (this.stateGraphAsset) {
+        if (assetId) {
             const stateGraph = this.system.app.assets.get(this.stateGraphAsset).resource;
```
yaustar, line 459:
```suggestion
            const stateGraph = this.system.app.assets.get(assetId).resource;
```
Should this be `assetId` instead?

src/framework/components/anim/component.js
```diff
@@ -475,6 +474,14 @@ class AnimComponent extends Component {
         }
     }
 
+    unbind() {
+        if (!this._normalizeWeights) {
```
yaustar, line 478: Given the function name `unbind`, it would imply it would unbind regardless of `_normalizeWeights`. Do we want the if check here or in the setter of `normalizeWeights`?

ellthompson: We'd want this check any time unbind is called (currently changing the normalizeWeights property or removing the state graph).

src/framework/components/anim/component.js
```diff
@@ -447,17 +447,16 @@ class AnimComponent extends Component {
...
     }
 
     resetStateGraph() {
+        const assetId = this._stateGraphAsset;
         this.removeStateGraph();
-        if (this.stateGraphAsset) {
-            const stateGraph = this.system.app.assets.get(this.stateGraphAsset).resource;
+        if (assetId) {
+            const stateGraph = this.system.app.assets.get(assetId).resource;
             this.loadStateGraph(stateGraph);
```
yaustar, line 460: `_stateGraphAsset` is still null after this call. Should we set this prior to the call?

yaustar (approved):
LGTM

Issue #3726, opened 2021-11-21
Have an Aspect Ratio component


marcusx2:
Something exactly like Unity's https://docs.unity3d.com/Packages/com.unity.ugui@1.0/manual/script-AspectRatioFitter.html . The use case?

It's not possible to make image elements with different anchors and pivots work under the same group. It's not possible to have 1 image anchored to the bottom, 1 to the bottom right, while having both under the same group and have it work across all screens for example.

Maybe I'm doing something wrong though. But I guess the issue lies with the fact that I can't make a group fill the whole screen. There isn't an option to do that. I can make the group fill 1280x720 like the 2d screen, and place the images inside the group. It will seem like it works, but then when you run on different screen resolutions it doesn't. If there was an option to make the width/height fill the parent container(2D screen on this case), it would solve the problem,

yaustar:
> It's not possible to make image elements with different anchors and pivots work under the same group. It's not possible to have 1 image anchored to the bottom, 1 to the bottom right, while having both under the same group and have it work across all screens for example.

Set the group elemnt anchors to be 0, 0, 1, 1 as shown here for the render texture https://developer.playcanvas.com/en/tutorials/resolution-scaling/

By doing so, it will always fill to the size of the parent.

Aspect size fitter is a valid request but does have more specialised behaviour compared to what is being asked for here

PR #4206, opened 2022-04-22
Add fitMode property on Image Elements.

jpauloruschel:
Fixes https://github.com/playcanvas/engine/issues/3726

Adds a new property to `Element` component, used only on `Image`: `fitMode`. Set how the content should be fitted and preserve the aspect ratio of the source texture or sprite. (nomenclature is borrowed from https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit):
- `stretch` (default; `pc.FITMODE_STRETCH`): Fit the content exactly to the Element's bounding box.
- `contain` (`pc.FITMODE_CONTAIN`): Fit the content within the Element's bounding box while preserving its Aspect Ratio.
- `cover` (`pc.FITMODE_COVER`): Fit the content to cover the entire Element's bounding box while preserving its Aspect Ratio.

Screenshot 2022-04-25 at 14 41 18

_the white boxes are other Image elements placed on top of the actual image and are there just to represent the actual bounds of each *Element*_

#### Extra: Cover + Masking

The `cover` setting works very well with an additional `mask` `Element` if the desired effect is to clip the content to only display the part _inside_ the bounding box:

mask

### DIFF
--- a/src/framework/components/element/constants.js
+++ b/src/framework/components/element/constants.js
@@ -22,20 +22,20 @@ export const ELEMENTTYPE_TEXT = 'text';
 /**
  * Disable Aspect Ratio content resizing.
  *
- * @type {number}
+ * @type {string}
  */
-export const ELEMENT_IMAGE_ASPECT_NONE = 0;
+export const ELEMENT_IMAGE_ASPECT_NONE = 'none';
 
 /**
  * Resize the content to fit within the Element's bounding box.
  *
- * @type {number}
+ * @type {string}
  */
-export const ELEMENT_IMAGE_ASPECT_CONTAIN = 1;
+export const ELEMENT_IMAGE_ASPECT_CONTAIN = 'contain';
 
 /**
  * Resize the content to cover the entire Element's bounding box.
  *
- * @type {number}
+ * @type {string}
  */
-export const ELEMENT_IMAGE_ASPECT_COVER = 2;
+export const ELEMENT_IMAGE_ASPECT_COVER = 'cover';
--- a/src/framework/components/element/constants.js
+++ b/src/framework/components/element/constants.js
@@ -24,7 +24,7 @@ export const ELEMENTTYPE_TEXT = 'text';
  *
  * @type {string}
  */
-export const ELEMENT_IMAGE_FIT_STRETCH = 'stretch';
+export const FITMODE_STRETCH = 'stretch';
 
 /**
  * Fit the content within the Element's bounding box while
@@ -32,7 +32,7 @@ export const ELEMENT_IMAGE_FIT_STRETCH = 'stretch';
  *
  * @type {string}
  */
-export const ELEMENT_IMAGE_FIT_CONTAIN = 'contain';
+export const FITMODE_CONTAIN = 'contain';
 
 /**
  * Fit the content to cover the entire Element's bounding box while
@@ -40,4 +40,4 @@ export const ELEMENT_IMAGE_FIT_CONTAIN = 'contain';
  *
  * @type {string}
  */
-export const ELEMENT_IMAGE_FIT_COVER = 'cover';
+export const FITMODE_COVER = 'cover';
--- a/src/framework/components/element/constants.js
+++ b/src/framework/components/element/constants.js
@@ -18,3 +18,24 @@ export const ELEMENTTYPE_IMAGE = 'image';
  * @type {string}
  */
 export const ELEMENTTYPE_TEXT = 'text';
+
+/**
+ * Disable Aspect Ratio content resizing.
+ *
+ * @type {number}
+ */
+export const ELEMENT_IMAGE_ASPECT_NONE = 0;
+
+/**
+ * Resize the content to fit within the Element's bounding box.
+ *
+ * @type {number}
+ */
+export const ELEMENT_IMAGE_ASPECT_CONTAIN = 1;
+
+/**
+ * Resize the content to cover the entire Element's bounding box.
+ *
+ * @type {number}
+ */
+export const ELEMENT_IMAGE_ASPECT_COVER = 2;
--- a/src/framework/components/element/constants.js
+++ b/src/framework/components/element/constants.js
@@ -20,22 +20,24 @@ export const ELEMENTTYPE_IMAGE = 'image';
 export const ELEMENTTYPE_TEXT = 'text';
 
 /**
- * Disable Aspect Ratio content resizing.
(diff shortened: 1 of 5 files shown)


src/framework/components/element/constants.js
```diff
@@ -18,3 +18,24 @@ export const ELEMENTTYPE_IMAGE = 'image';
  * @type {string}
  */
 export const ELEMENTTYPE_TEXT = 'text';
+
+/**
+ * Disable Aspect Ratio content resizing.
+ *
+ * @type {number}
+ */
+export const ELEMENT_IMAGE_ASPECT_NONE = 0;
```
mvaligursky, line 27: I wonder if those should be strings instead of numbers, for easier way to debug things.

jpauloruschel: I was a bit worried if there'd be any noticeable performance implications in UI-heavy scenes, considering it needs to be checked every time the mesh changes.

An alternative perhaps is to have a "public" property which is a string, but then gets converted to a number on `set`.

src/framework/components/element/image-element.js
```diff
@@ -584,21 +600,35 @@ class ImageElement {
...
+        this._targetAspectRatio = -1;
+
         if (this._sprite && this._sprite.atlas) {
+            // take mesh from sprite
             mesh = this._sprite.meshes[this.spriteFrame];
             nineSlice = this._sprite.renderMode === SPRITE_RENDERMODE_SLICED || this._sprite.renderMode === SPRITE_RENDERMODE_TILED;
+
+            // re-calculate aspect ratio from sprite frame
+            const frameData = this._sprite.atlas.frames[this._sprite.frameKeys[this._spriteFrame]];
+            if (frameData) {
```
mvaligursky, line 620: just to be sure, change to `if (frameData?.rect.w > 0) {

mvaligursky (approved):
Few small comments, but this looks great!

src/framework/components/element/constants.js
mvaligursky: There would be no difference, as constant strings are compared by pointers, so the speed is the same as numbers.

jpauloruschel: Oh nice - didn't know that! I'll change it to strings, then.

Maksims:
If `contain` is used, the out of image part, will be discarded?

mvaligursky (approved):

jpauloruschel:
> If `contain` is used, the out of image part, will be discarded?

What do you mean by 'discarded'? Both `contain` and `cover` change the vertices of the internal mesh used to render the element. So for `contain`, the part out of the image is completely empty.

jpauloruschel:
> > > If `contain` is used, the out of image part, will be discarded?
> > 
> > 
> > What do you mean by 'discarded'? Both `contain` and `cover` change the vertices of the internal mesh used to render the element. So for `contain`, the part out of the image is completely empty.
> 
> I see, just on a screenshot in PR, it showed grey areas.

Aahh right! Yea sorry those are just other image elements placed on top just to represent the actual bounds. I updated the description to try and explain that further.

src/framework/components/element/constants.js
```diff
@@ -20,22 +20,24 @@ export const ELEMENTTYPE_IMAGE = 'image';
 export const ELEMENTTYPE_TEXT = 'text';
 
 /**
- * Disable Aspect Ratio content resizing.
+ * Fit the content to the exact Element's bounding box.
  *
  * @type {string}
  */
-export const ELEMENT_IMAGE_ASPECT_NONE = 'none';
+export const ELEMENT_IMAGE_FIT_STRETCH = 'stretch';
```
willeastcott, line 27: What about `FITMODE_STRETCH`?

jpauloruschel: Yea, I like it! I was thinking of going with the `<component>_<subcomponent>_<property>_<enum value>` style of constants, but it can get quite verbose, and there's quite a few of these less-verbose types of enums. Will change.

src/framework/components/element/image-element.js
```diff
@@ -584,21 +600,35 @@ class ImageElement {
     // Gets the mesh from the sprite asset
     // if the sprite is 9-sliced or the default mesh from the
     // image element and calls _updateMesh or sets meshDirty to true
-    // if the component is currently being initialized. We need to call
-    // _updateSprite every time something related to the sprite asset changes
+    // if the component is currently being initialized. Also updates
+    // aspect ratio. /We need to call _updateSprite every time
+    // something related to the sprite asset changes
```
willeastcott, line 605:
```suggestion
    // aspect ratio. We need to call _updateSprite every time
    // something related to the sprite asset changes.
```

src/framework/components/element/constants.js
```diff
@@ -18,3 +18,26 @@ export const ELEMENTTYPE_IMAGE = 'image';
...
+ * Fit the content within the Element's bounding box while
+ * preserving its Aspect Ratio.
+ *
+ * @type {string}
+ */
+export const FITMODE_CONTAIN = 'contain';
+
+/**
+ * Fit the content to cover the entire Element's bounding box while
+ * preserving its Aspect Ratio.
```
willeastcott, line 39:
```suggestion
 * Fit the content to cover the entire Element's bounding box while preserving its aspect ratio.
```

src/framework/components/element/constants.js
```diff
@@ -18,3 +18,26 @@ export const ELEMENTTYPE_IMAGE = 'image';
...
+/**
+ * Fit the content to the exact Element's bounding box.
+ *
+ * @type {string}
+ */
+export const FITMODE_STRETCH = 'stretch';
+
+/**
+ * Fit the content within the Element's bounding box while
+ * preserving its Aspect Ratio.
```
willeastcott, line 31:
```suggestion
 * Fit the content within the Element's bounding box while preserving its aspect ratio.
```

src/framework/components/element/constants.js
```diff
@@ -18,3 +18,26 @@ export const ELEMENTTYPE_IMAGE = 'image';
  * @type {string}
  */
 export const ELEMENTTYPE_TEXT = 'text';
+
+/**
+ * Fit the content to the exact Element's bounding box.
```
willeastcott, line 23:
```suggestion
 * Fit the content exactly to the Element's bounding box.
```

willeastcott (approved):

marcusx2:
This was merged into the main branch right? Where is the fitMode property?
image

Issue #4135, opened 2022-03-16
Application destroy doesn't clear all memory


OlegGedzjuns:
After each application create/destroy cycle extra heap memory is used.

https://github.com/OlegGedzjuns/multi-application-destroy

1. npm install
2. npm run start

yaustar:
Have you find any clues of which parts of the engine aren't being cleaned up properly?

OlegGedzjuns:
No, unfortunately I did not see anything unusual in the memory profiler

mvaligursky:
I'm not sure how much these number represent memory that is not released, but if they do, there have been some improvements:

<img width="302" alt="Screenshot 2022-11-17 at 15 29 17" src="https://user-images.githubusercontent.com/59932779/202488100-38b44c8d-111a-42bb-9ada-0726168b041c.png">

mvaligursky:
@OlegGedzjuns - could you please check the status on the engine 2.1.0? There were many deallocation changes since the last update 2 years ago.

I'll close this ticket, but please re-open it when you still find some leaks in 2.1.0, thanks!

PR #4207, opened 2022-04-22
Destroy XR-DepthSensing texture when Application is destroyed

mvaligursky:
- XR-DepthSensing texture gets destroyed when Application is destroyed
- small cleanup to pass texture name in options for all engine created textures


mvaligursky:
slightly related to https://github.com/playcanvas/engine/issues/4135 (even though that one likely does not allocated any gfx resources)

slimbuck:
This is great! Though I'd love to see 'name' specified as the first option everywhere.

willeastcott (approved):

Issue #4118, opened 2022-03-15
GLB vertex colors not appearing 


yaustar:
Forum: https://forum.playcanvas.com/t/draco-glb-diffuse-emissivevertexcolor/24769

Test project with Draco and uncompressed GLBs: https://playcanvas.com/editor/scene/1363413

<img width="1273" alt="image" src="https://user-images.githubusercontent.com/16639049/158355893-89fe7608-3837-4668-a0d3-6b6476a94061.png">

Should look like this:
![image](https://user-images.githubusercontent.com/16639049/158356052-329d21c1-a387-48ba-bf5a-6cc72d64efda.png)

willeastcott:
The problem here seems to be that the GLB contains no materials. The engine just uses a 'default material' if no materials are found:

https://github.com/playcanvas/engine/blob/main/src/resources/parser/glb-container-resource.js#L81

Materials created by the `GlbParser` enable tint+vertex colors:

https://github.com/playcanvas/engine/blob/main/src/resources/parser/glb-parser.js#L1007-L1008

But the default material does not have this enabled. The `GlbContainerResource` constructor should probably be passed a default *glTF* material, rather than just a vanilla `StandardMaterial`.

willeastcott:
So there are actually two problems here. One affects the uncompressed model and one affects the Draco compressed model.

* **Uncompressed**
  * Problem: A 'default' `StandardMaterial` is applied to mesh instances when the GLB does not explicitly specify a material.
  * Solution: Set the `GlbParser`'s default material to be a default **glTF** material (as created by `createMaterial` in glb-parser.js.
* **Draco Compressed**
  * Problem: The Draco module function `normalized()` returns false for the color vertex attribute. It should return true.
  * Solution: I'm less sure what to do here. But I tried setting `normalized` to true for all Uin8 attributes and that fixed the problem.

willeastcott:
It's probably best if I hand this over to @mvaligursky now since I'm not 100% whether my solutions fully resolve the problems.

mvaligursky:
Great investigation @willeastcott !

PR #4209, opened 2022-04-22
Improved vertex color handling for GLB files

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/4118 (great investigation and solutions @willeastcott )

- instead of using DefaultMaterial for missing materials inside GLB, use GLB specific default material
- assume 8but colors in draco compressed glb format are normalized (some compressor don't set that flag correctly, I've seen it few times in the past)

### DIFF
--- a/src/resources/parser/glb-parser.js
+++ b/src/resources/parser/glb-parser.js
@@ -40,7 +40,6 @@ import { Morph } from '../../scene/morph.js';
 import { MorphTarget } from '../../scene/morph-target.js';
 import { Skin } from '../../scene/skin.js';
 import { StandardMaterial } from '../../scene/materials/standard-material.js';
-import { getDefaultMaterial } from '../../scene/materials/default-material.js';
 import { Render } from '../../scene/render.js';
 
 import { Entity } from '../../framework/entity.js';
--- a/src/resources/parser/glb-parser.js
+++ b/src/resources/parser/glb-parser.js
@@ -568,7 +568,7 @@ const createVertexBufferDraco = function (device, outputGeometry, extDraco, deco
     const numPoints = outputGeometry.num_points();
 
     // helper function to decode data stream with id to TypedArray of appropriate type
-    const extractDracoAttributeInfo = function (uniqueId) {
+    const extractDracoAttributeInfo = function (uniqueId, semantic) {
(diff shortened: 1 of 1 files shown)


src/resources/parser/glb-parser.js
```diff
@@ -2293,7 +2294,9 @@ class GlbParser {
     constructor(device, assets, maxRetries) {
         this._device = device;
         this._assets = assets;
-        this._defaultMaterial = getDefaultMaterial(device);
+        this._defaultMaterial = createMaterial({
```
slimbuck, line 2297: What is the difference between the material created here and the device default material? It looks like the device default material has had [shadingModel](https://github.com/playcanvas/engine/blob/main/src/framework/app-base.js#L648) set, but that's the standard material [default anyway](https://github.com/playcanvas/engine/blob/main/src/scene/materials/standard-material.js#L1057).

Is it missing:
```
    material.diffuseTint = true;
    material.diffuseVertexColor = true;

    material.specularTint = true;
    material.specularVertexColor = true;
```
?

mvaligursky:
yes, those options are what make the vertex colors to work

src/resources/parser/glb-parser.js
mvaligursky: yes, those options are what make the vertex colors to work

slimbuck: So why aren't you setting the options on the material here? What am I missing? :)

mvaligursky: I don't want to modify the global default material - that one is used by many other things. So we create a glb specific material here. 
I though about having a single global one .. but creating one per glb is probably a good idea, so that the use can modify it further without affecting other glbs

slimbuck: Sorry, what I mean is: you're not setting 'diffuseTint' or 'diffuseVertexColor' on this material?

mvaligursky: the createMaterial function does that, which I'm calling.

slimbuck: Yes, right that's what I missed. `Face palm`. Thanks for your patience at this time!

slimbuck (approved):

PR #4214, opened 2022-04-27
Fix tiled nineslice shader error

slimbuck:
Fixes: https://forum.playcanvas.com/t/tiled-image-and-sprite-generate-fragment-shader-error-on-new-engine-version-1-53-2/25423/3

### DIFF
--- a/src/graphics/program-lib/programs/standard.js
+++ b/src/graphics/program-lib/programs/standard.js
@@ -143,7 +143,7 @@ const standard = {
         if (isMainPass && options.nineSlicedMode === SPRITE_RENDERMODE_SLICED) {
             expression = "nineSlicedUv";
         } else if (isMainPass && options.nineSlicedMode === SPRITE_RENDERMODE_TILED) {
-            expression = "nineSlicedUv, -1000.0";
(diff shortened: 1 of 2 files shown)


src/graphics/program-lib/programs/standard.js
```diff
@@ -332,6 +332,13 @@ const standard = {
             code += chunks.baseNineSlicedTiledPS;
         }
 
+        // declare textureBias
+        if (options.nineSlicedMode === SPRITE_RENDERMODE_TILED) {
+            code += `const float textureBias = -1000.0;\n`;
+        } else {
+            code += `uniform float textureBias;\n`;
```
mvaligursky, line 339: _getUvSourceExpression can get called multiple times for a shader - will we have multiple uniform declarations?

slimbuck: no, this is `_fsAddBaseCode` function.

mvaligursky: ah right, I missed that. all good then.

mvaligursky (approved):

Issue #4228, opened 2022-05-05
Review all throws/exceptions in the graph node class


yaustar:
There are several instances in the graph node class where it throws an exception and stop loading when it's possible to recover (eg being parented twice)

We should review these and see if they should remain exceptions or be changed to a warning and attempt to recover/ignore

PR #4232, opened 2022-05-05
Sanity check on Scene data where the child entity is in the Entity data children list twice

yaustar:
Related tickets: 

https://github.com/playcanvas/engine/issues/4228
https://github.com/playcanvas/editor/issues/758

I'm not advocating for this change as it seems like it could affect runtime performance but more to show what we would need if we wanted to make the engine resilient against scene corruption from the Editor in issue: https://github.com/playcanvas/editor/issues/758

### DIFF
--- a/src/resources/parser/scene.js
+++ b/src/resources/parser/scene.js
@@ -114,9 +114,16 @@ class SceneParser {
 
         // If the data mis matches the entities created, attempt
         // to recover
-        if (len !== children.length) {
+        let entityChildrenCount = 0;
+        for (let i = 0; i < children.length; i++) {
+            if (children[i] instanceof Entity) {
+                ++entityChildrenCount;
+            }
+        }
+
+        if (len !== entityChildrenCount) {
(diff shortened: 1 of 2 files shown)


src/resources/parser/scene.js
```diff
@@ -110,6 +111,21 @@ class SceneParser {
         // Open all children and add them to the node
         len = entityData.children.length;
         const children = entity._children;
+
+        // If the data mis matches the entities created, attempt
+        // to recover
+        let entityChildrenCount = 0;
+        for (let i = 0; i < children.length; i++) {
+            if (children[i] instanceof Entity) {
```
mvaligursky, line 119: can the scene contains anything else other than the Entity? I wonder if this loop is needed at all.

yaustar: Yes, it does. I can't remember exactly what it was from memory (I think graph nodes) but the check did need to be there.

yaustar:
Closing as this isn't realistic to merge in

PR #4235, opened 2022-05-10
Support wasm module loading

slimbuck:
This PR adds standard engine support for loading and initialising emscripten-style wasm modules.

The API supports lazy loading, which is useful for tools that don't require all modules loaded at startup.

The following API is added:
```javascript
WasmModule.setConfig(moduleName, config)         // set wasm module url config
WasmModule.getInstance(moduleName, callback)     // request the wasm module instance
```

The two functions can be invoked in any order and will resolve correctly.

The Draco decompressor has been updated to use this interface.

Once this PR is merged, the other uses of wasm modules in the codebase (editor applications, examples, model-viewer) should all migrate to this interface.


src/core/module.js
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * Callback used by {@link Module#getInstance}.
+ *
+ * @callback ModuleInstanceCallback
+ * @param {string} error - If the instance fails to load this will contain a description of the error.
+ * @param {any} moduleInstance - The module instance.
+ */
+
+class Module {
```
mvaligursky, line 101: The `Module` names fees little too generic here.

mvaligursky:
I wonder what is the ideal behaviour with wasm loading for web application that create multiple instances of the engine - do we need just one instance of wasm module, or does it need to be per application.
Also, do we need to take care of their release?

src/core/module.js
slimbuck: I thought about it, but `EMSModule` didn't seem great either.

yaustar: WasmModule? Could a module be something else that isn't Wasm?

slimbuck:
> does it need to be per application.

I think it depends on the module. Draco probably can be used across-application whereas Ammo might require separate instances. If this became a requirement we could update the config accordingly.

> Also, do we need to take care of their release?

Don't believe you can directly destroy wasm modules.

yaustar:
Does this also support network retries if the module fails to load from the network?

mvaligursky:
I wonder if this should be utilized when supported. (Could be done in a separate PR)
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/instantiateStreaming

slimbuck:
> Does this also support network retries if the module fails to load from the network?

I don't think the glue code has retry logic for the wasm module, but if this was something we needed, we could manage the download & compile of the wasm module ourselves.

src/core/module.js
slimbuck: It's a subset of wasm modules - those generated by emscripten compiler.

yaustar:
> I don't think the glue code has retry logic for the wasm module, but if this was something we needed, we could manage the download & compile of the wasm module ourselves.

I'm just thinking of the issues that we have with Snapchat where network requests can fail on iOS. It's on a WASM module that doesn't have retry, it could stop loading of the game or prevent the game from working correctly.

yaustar:
> I don't -think- we need script tag retry logic (and wouldn't know how to implement such a thing). If we wanted manual retry logic we'd have to download the script manually (using http interface) instead. I'm not sure that'd be better and suspect the browser's built-in script tag handling is sufficient.

It will be something we could test if/when it becomes an issue I guess. Even if it doesn't handle retrying, given this can be lazy loaded, we could potentially mitigate the issue by loading the WASM module first and than other assets that are needed?

src/core/module.js
slimbuck: renamed to `WasmModule`

mvaligursky:
- just looking at wasm-module.js and wonder why you created those functions at the start of the file as standalone functions instead of making them static functions of the class?

- should WasmModule have some jsdocs for the class itself? I think it might need to to have api docs generated for the class, not sure

- I wonder if we should upgrade one of the engine examples to register the draco wasm and load draco glb? multi-view example would be good, as that uses vertex heavy chess-board. I can try that later as well.

slimbuck:
> * just looking at wasm-module.js and wonder why you created those functions at the start of the file as standalone functions instead of making them static functions of the class?
> * should WasmModule have some jsdocs for the class itself? I think it might need to to have api docs generated for the class, not sure
> * I wonder if we should upgrade one of the engine examples to register the draco wasm and load draco glb? multi-view example would be good, as that uses vertex heavy chess-board. I can try that later as well.

Nice! Thanks for the suggestions:
- I moved the naked functions into an `Impl` class with static functions. This way code looks a little neater and doesn't expose the internal functions to the public interface
- Added class jsdoc, pls check it
- Upgraded all the examples to use the new wasm loading and removed loading from example code

cc @ellthompson - please check changes to examples.

src/core/wasm-module.js
```diff
@@ -0,0 +1,164 @@
...
+                        onAbort: () => {
+                            callback('wasm module aborted.');
+                        }
+                    }).then((instance) => {
+                        callback(null, instance);
+                    });
+                }
+            });
+        } else {
+            callback('No supported wasm modules found.', null);
```
mvaligursky, line 56: - could this add the module name to the error message
- and perhaps even log a debug warning - to help find the problem in case the user does not handle errors correctly

slimbuck: The default error handler logs the details as you describe. Otherwise if the caller registers an error handler, it's up to them to do this.

willeastcott (approved):

PR #4236, opened 2022-05-10
Explicitly mark lone getters as readonly

### LINKED ISSUE #4229: The generated API documentation does not indicate some properties are read only
For example all properties of the RenderTarget class are read only (only have getters but not setters), but the API documentation does not mention it:

https://developer.playcanvas.com/api/pc.RenderTarget.html

I believe this was displayed as ReadOnly.

willeastcott:
Fixes #4229 

### DIFF
--- a/src/anim/evaluator/anim-curve.js
+++ b/src/anim/evaluator/anim-curve.js
@@ -29,6 +29,7 @@ class AnimCurve {
      * The list of paths which identify targets of this curve.
      *
      * @type {string[]}
+     * @readonly
      */
     get paths() {
         return this._paths;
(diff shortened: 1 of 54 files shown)


mvaligursky:
This is great! Such an easy solution!

willeastcott:
Actually, it's not that easy. The TS declarations now fail to compile with errors like:

```
build/playcanvas.d.ts(466,5): error TS1024: 'readonly' modifier can only appear on a property declaration or index signature.
```

And it seems JSDoc isn't smart enough to spot a getter without a setter and mark it as readonly implicitly. 😢

willeastcott:
Closing for now this since it breaks the TS declarations.

Issue #4238, opened 2022-05-11
Inconsistent line endings (LF vs CRLF)


liamdon:
### Description

Line endings across the codebase are inconsistent between CRLF (Windows) and LF (Unix).
95% of files use LF (Unix), and the other 5% use CRLF (30 files total). There are no cases of line endings being mixed in one file.

This is definitely a nitpick issue, but it would be nice to fix to avoid any ambiguity for future contributors.

I've opened this as an issue rather than a PR because the diff would change every line in 30 files, making it hard to review and potentially a security concern. Instead, I've added some instructions on how to fix it below.

### Instructions to fix
Install `dos2unix` util:
```
brew install dos2unix
```

Print a report of all line endings without changing anything (see this output in table below):
```
find ./src -type f -print0 | xargs -0 dos2unix -i
```

Convert all the line endings to LF:
```
find ./src -type f -print0 | xargs -0 dos2unix
```

### Full report of line ending counts

<details>
<summary>Expand table</summary>
</br>

|File|Windows (CRLF)|Unix (LF)|
|----|--------------|---------|
|./anim/constants.js|0|23|
|./anim/evaluator/anim-data.js|0|39|
|./anim/evaluator/anim-clip.js|0|200|
|./anim/evaluator/anim-target-value.js|0|133|
|./anim/evaluator/anim-evaluator.js|0|365|
|./anim/evaluator/anim-target.js|0|67|
|./anim/evaluator/anim-cache.js|0|144|
|./anim/evaluator/anim-track.js|0|116|
|./anim/evaluator/anim-curve.js|0|65|
|./anim/evaluator/anim-events.js|0|32|
|./anim/evaluator/anim-snapshot.js|0|46|
|./anim/binder/anim-binder.js|0|100|
|./anim/binder/default-anim-binder.js|0|252|
|./anim/state-graph/anim-state-graph.js|0|135|
|./anim/controller/constants.js|0|163|
|./anim/controller/anim-controller.js|0|593|
|./anim/controller/anim-node.js|0|100|
|./anim/controller/anim-transition.js|0|88|
|./anim/controller/anim-blend-tree-2d-directional.js|0|65|
|./anim/controller/anim-blend-tree-2d-cartesian.js|0|58|
|./anim/controller/anim-blend-tree-1d.js|0|69|
|./anim/controller/anim-blend-tree.js|0|105|
|./anim/controller/anim-state.js|0|165|
|./anim/controller/anim-blend-tree-direct.js|0|38|
[...]

PR #1824, opened 2020-01-17
Update to reflect that the Animation component's assets can be assets or asset ids

TheJonRobinson:
Per the documentation, the `assets` array in the Animation component can be assets or asset ids. The current docs declare assets, but we find ids in the arrays. So, change to be either so we can get type-safe results.

NB Confirmed generated docs as working fine.

I confirm I have signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #4240, opened 2022-05-11
Add .gitattributes to normalize how line endings are stored

liamdon:
Related to issue #4238

## Problem
Line endings in the repo are inconsistent between CRLF (Windows) and LF (Unix/Git).

## Solution
Add a .gitattributes file so that git will normalize files as they are checked in, while still allowing developers to use their native line endings locally. 

Issue #4238 suggests an additional cleanup task to fix existing files that were already checked in as CRLF. I am happy to open that task as a PR as well (just let me know), but I thought that the core maintainers might prefer to do so since it will be a scary diff (every line changed in 30 files).

## Context
Currently around 5% of the source files are stored as CRLF (Windows), and the other 95% are LF (Unix) (see full report in #4238).

This causes issues for developers on both platforms: 
* It can be tricky to ensure your editor retains the existing line endings for a given file, but if you change a file's line endings the resulting diff will show every line as changed. This makes it hard to review the PR! ([example](https://github.com/playcanvas/engine/pull/1824/commits/07f7ba2b8bc0c4b211cc9d0d506d28daa1b2e0b4))
* Developers might also be confused about whether they should be changing line endings deliberately for consistency. 
 For example, `mat4.js` and `vec2.js` use different line endings - should a developer change them to be consistent, or leave them as they are? What format should they use for a new file? 

By using `.gitattributes` to normalize line endings at checkin/checkout, no-one will have to think about line endings ever again 🥳

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,12 +1,12 @@
-# This file normalizes line endings stored in the repo to LF.
+# This file normalizes line endings stored in the repo to LF (Unix/Git).
 # Contributors are still able to use their native line endings locally.
 # More info here: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings
 
 # Set the default behavior, in case people don't have core.autocrlf set.
 * text=auto
 
-# Explicitly declare text files to be normalized and converted
-# to native line endings on checkout.
+# Explicitly declare text files to be normalized on checkin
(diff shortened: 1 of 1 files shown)


willeastcott:
This is great @liamdon - thanks so much. And yep, it would be a big help if you also attended to the 5% of files which use CRLF. 🙏

Issue #4090, opened 2022-03-08
Create simple API to give users access to depth buffer


mvaligursky:
At the moment, the engine by default creates a Depth layer, which is disabled. If this layer is added to a camera, and the layer is enabled, this triggers the rendering operation (render on webgl1, copying of depth buffer on webgl2) at the point the camera gets to render the Depth layer. After this point, the depth texture is accessible to following layers.

The way to enable the depth layer is this:
```
        const depthLayer = this.app.scene.layers.getLayerById(LAYERID_DEPTH);
        depthLayer.incrementCounter();
```

and to disable:
```
        depthLayer.decrementCounter();
```

We should expose a simple API to manage this, perhaps 
```
CameraComponent.renderDepth = true;
```

Perhaps some changes could be made to how the Depth layer is integrated into the camera. By default, the layer is automatically added to the default camera. And the user is responsible for adding it to other cameras (when rendering to texture). Perhaps the renderDepth API could automatically add / remove the layer for the camera as well.

It'd be also great to create an engine example showing how to use the rendered depth in some custom shader.

Issue #2953, opened 2021-03-10
Refactor depth buffer rendering


mvaligursky:
- refactor to better integrate with the engine
- implement support for WEBGL_depth_texture: https://github.com/playcanvas/engine/issues/2856
- fix: https://github.com/playcanvas/engine/issues/2932
- fix: ERROR: FRAMEBUFFER_INCOMPLETE_ATTACHMENT when resizing window and depth buffer is rendering. Repro: Post effects engine example with debug engine [this seems fixed in #4244]

Maksims:
Not sure in details how it works now, but recently developer had to do this:

1. Render World Layer into RT.
2. Do few passes to blur texture into another RT.
3. Render Glass Layer (with previously blurred texture) from the respecting actual depth of the World Layer, so it occludes correctly.

And I think it was not smooth or difficult to do such thing efficiently, avoiding extra depth pre-pass on Glass Layer, so to avoid unnecessary drawcalls.

YessicaSD:
Help!!! I have created SSAO post effect and It works fine in windows but in Mac it triggers the ERROR: FRAMEBUFFER_INCOMPLETE_ATTACHMENT (although it shows the effect)

mvaligursky:
this is our attempt to have good quality SSAO .. https://github.com/playcanvas/engine/pull/3092

YessicaSD:
I've made some changes and it doesn't crush but it seams that the depth buffer is not clearing.
I followed this project's structure -> https://playcanvas.com/project/642131/overview/depthmap-access

mvaligursky:
if you need support, please post on the forum: http://forum.playcanvas.com/
This is to post issues with the engine code for us to work on.

mvaligursky:
I tried to use Webgl1 extension `WEBGL_depth_texture` when refactoring the depth rendering in https://github.com/playcanvas/engine/pull/4244, to avoid having to re-render meshes from layers before the Depth layer.  
This can be done when the camera is rendering to a render target, as using the extension we can then copy the depth using shader into a texture (similarly as done in the Viewer: https://github.com/playcanvas/model-viewer/blob/main/src/read-depth.ts). The problem is when we're rendering straight into a default framebuffer - I don't see a way to copy out the depth from it, as it cannot be accessed as a texture and used in the shader. Specifically, I tried calling this to obtain the depth attachment (texture)
```
gl.bindFramebuffer(gl.FRAMEBUFFER, null);
[...]

PR #4244, opened 2022-05-12
Refactored the grab pass and scene depth rendering

mvaligursky:
Fixes: https://github.com/playcanvas/engine/issues/4090
Related: https://github.com/playcanvas/engine/issues/2953

This PR refactors and improved both grab pass functionality, as well as the scene depth rendering. Files grab-pass.js and scene-depth.js were refactored into scene-grab.js. The main motivation is to minimize number of resolves and buffer copies taking place, and keeping to more formal render pass implementation required for webGPU.

**Shared changes**
- previously, if the depth texture was used, a depth layer was enabled, and all cameras that had the depth layer would capture the depth, wasting performance for cameras other than the main one. This is now controlled by camera (details below)
- there are now new and consistent textures names for these two textures. Note that even though the engine uses new names, old names are still supported:
```
uDepthMap -> uSceneDepthMap
texture_grabPass -> uSceneColorMap
```

**Changes to grab-pass**
- previously, when a shader referenced grab pass texture, a rendering would be interrupted in the middle of draw call, a framebuffer copied to texture and then the texture used for the draw call. This gets very expensive when multiple materials use the texture. The commonly used solution is to do a single framebuffer copy after opaque meshes are rendered, making it available for the transparent meshes / post-processing. In our case, this executed during the depth layer, giving used the control over the time inside the frame.
- There is new public API to request a camera to capture the scene: 
```
CameraComponent.requestSceneColorMap
```

**Changes to depth pass**
- There is new public API to request a camera to capture the scene. This is automatically requested by post-effects, but needs to be manually requested for other uses (for example Soft-Particles).
```
CameraComponent.requestSceneDepthMap
```


src/framework/components/camera/component.js
```diff
@@ -144,6 +146,20 @@ class CameraComponent extends Component {
      */
     onPostRender = null;
 
+    /**
+     * A counter of requests of depth map rendering.
+     *
+     * @type {number}
```
willeastcott, line 152: Does this need a @private?

src/framework/components/camera/component.js
```diff
@@ -144,6 +146,20 @@ class CameraComponent extends Component {
...
+     */
+    _renderSceneDepthMap = 0;
+
+    /**
+     * A counter of requests of color map rendering.
+     *
+     * @type {number}
```
willeastcott, line 159: Does this need a @private tag?

src/framework/components/camera/component.js
mvaligursky: well spotted, thanks, I keep forgetting anything documented is public by default now. Done.

src/framework/components/camera/component.js
mvaligursky: and done.

src/framework/scene-grab.js
```diff
@@ -0,0 +1,363 @@
...
+
+// Internal class abstracting the access to the depth and color texture of the scene.
+// color frame buffer is copied to a texture
+// For webgl 2 devices, the depth buffer is copied to a texture
+// for webgl 1 devices, the scene's depth is rendered to a separate RGBA texture
+//
+// TODO: implement mipmapped color buffer support for WebGL 1 as well, which requires
+// the texture to be a power of two, by first downscaling the captured framebuffer
+// texture to smaller power of 2 texture, and then generate mipmaps and use it for rendering
+// TODO: or even better, implement blur filter to have smoother lower levels
```
willeastcott, line 30: Any reason not to make this a JSDoc block with an `@ignore` tag?

src/framework/scene-grab.js
```diff
@@ -0,0 +1,363 @@
...
+            this.initWebGl1();
+        }
+    }
+
+    allocateTexture(device, name, format, isDepth, mipmaps) {
+
+        // allocate texture that will store the depth
+        const texture = new Texture(device, {
+            name,
+            format: format,
```
willeastcott, line 59: Interesting - you're starting to use object shorthand....but I would use it on all properties or none. So maybe update this one to just `format`?

src/framework/scene-grab.js
mvaligursky: Good point, done.

src/framework/scene-grab.js
mvaligursky: and done as well, I'm warming up to it ;)

willeastcott (approved):

NewboO:
Hello,

How to get the old behaviour for the color buffer? In some of my projects, I'm using shaders to blend UI sprites with custom functions (similar to photoshop blend modes) so I need the current color buffer when rendering the sprite, not one captured some time before because as those sprites can overlap each other they're now cancelling each other instead of stacking.

mvaligursky:
@NewboO - Unfortunately we don't have a support for this anymore. For multiple reason, this has been removed for now. We will consider having this as an option in the future, but most likely not very immediately.

yaustar:
@NewboO, would you be able to share an example shader that no longer works? Perhaps we can provide an alternative path?

NewboO:
Here's a sample project: https://playcanvas.com/project/950042/overview/color-buffer

And here's how it looks with previous engine version: https://launch.playcanvas.com/1456173?debug=true&version=1.53.4
image

Now with current engine it gives this: https://launch.playcanvas.com/1456173?debug=true
image

This is a very basic example with only a difference blending on a single layer but I'd like to be able to do this kind of thing with more complex blending formulas on any layer.

mvaligursky:
I think a possible way to do something like this would be to create multiple render targets .. and swap rendering between them, so that you can use what was already rendered to as a texture for following rendering. Not as convenient, especially if this is in 3D with dynamically changing orders unfortunately, so more of a limited case solution I'd say.

NewboO:
An update if anyone else than me is interested in getting the old behaviour: the best workaround I found was to duplicate the SceneGrab code I need and call it in a custom Layer.onDrawCall function. It's not perfect but good enough for my use cases and it doesn't require juggling with render targets.

My sample project is updated with a dirty PoC showing both behaviours.

Issue #4246, opened 2022-05-13
When Cubemap Asset is loaded, .ready() doesn't think it is loaded


yaustar:
Example project: https://playcanvas.com/project/940344/overview/cubemap-loaded-bug

Cubemap is preloaded, script has the following code: 

```
// initialize code called once per entity
AssetReadyTest.prototype.initialize = function() {
    var readyCb = function() {
        console.log('should be called');
    };

    var asset = this.app.assets.find('Helipad', 'cubemap');
    asset.ready(readyCb);
    this.app.assets.load(asset);
};
```

If an asset is loaded when .ready is called, it should call the callback. 

https://github.com/playcanvas/engine/blob/main/src/asset/asset.js#L464

However, cubemap assets don't store the resource in `resource` but as an array in `resources` so this check in .ready is not valid for cubemaps

slimbuck:
NOTE: this is an issue for cubemaps with no faces (cubemaps that only have prefiltered lighting data) because the faces cubemap texture is stored in `resources[0]` (which is what `resource` returns).

yaustar:
> because the faces cubemap texture is stored in resources[0] (which is what resource returns).
```
    ready(callback, scope) {
        scope = scope || this;

        if (this.resource) {
            callback.call(scope, this);
        } else {
            this.once('load', function (asset) {
                callback.call(scope, asset);
            });
        }
    }
```
That's odd as if `resource` returns `resources[0]`, the ready call should work as it's checking `this.resource`

yaustar:
Looking at cubemap.update, I can't see where asset.resource is set. Should it be? https://github.com/playcanvas/engine/blob/main/src/resources/cubemap.js#L93

slimbuck:
https://github.com/playcanvas/engine/blob/main/src/resources/cubemap.js#L192

yaustar:
> https://github.com/playcanvas/engine/blob/main/src/resources/cubemap.js#L192

That's to .resources, not .resource? Isn't .resource meant to reference .resources[0]?

PR #4248, opened 2022-05-13
Fixed asset.ready() for prefiltered cubemaps

yaustar:
Fixes https://github.com/playcanvas/engine/issues/4246

### DIFF
--- a/src/asset/asset.js
+++ b/src/asset/asset.js
@@ -464,7 +464,12 @@ class Asset extends EventHandler {
     ready(callback, scope) {
         scope = scope || this;
 
-        if (this.resource) {
+        // If cubemap asset doesn't need to load the faces, then resource is null as it references
+        // resources[0] that stores the faces. We need to check if the prefiltered assets are
+        // loaded instead
+        const assetLoaded = this.resource || (this.type === "cubemap" && !this.loadFaces && this.resources.length > 0);
+
+        if (assetLoaded) {
             callback.call(scope, this);
         } else {
             this.once('load', function (asset) {
--- a/src/asset/asset.js
+++ b/src/asset/asset.js
@@ -464,12 +464,7 @@ class Asset extends EventHandler {
     ready(callback, scope) {
         scope = scope || this;
 
-        // If cubemap asset doesn't need to load the faces, then resource is null as it references
-        // resources[0] that stores the faces. We need to check if the prefiltered assets are
-        // loaded instead
-        const assetLoaded = this.resource || (this.type === "cubemap" && !this.loadFaces && this.resources.length > 0);
-
-        if (assetLoaded) {
+        if (this.loaded) {
             callback.call(scope, this);
         } else {
             this.once('load', function (asset) {


src/asset/asset.js
```diff
@@ -464,7 +464,12 @@ class Asset extends EventHandler {
     ready(callback, scope) {
         scope = scope || this;
 
-        if (this.resource) {
+        // If cubemap asset doesn't need to load the faces, then resource is null as it references
+        // resources[0] that stores the faces. We need to check if the prefiltered assets are
+        // loaded instead
+        const assetLoaded = this.resource || (this.type === "cubemap" && !this.loadFaces && this.resources.length > 0);
```
yaustar, line 470: This feels very specific but not really sure how else to check

src/asset/asset.js
```diff
@@ -464,7 +464,12 @@ class Asset extends EventHandler {
...
+
+        if (assetLoaded) {
```
slimbuck, line 472: Looks like this code was written years ago, perhaps before `loaded` flag was introduced. 

The correct code should be:
```suggestion
        if (this.loaded) {
```
This change would impact all assets though and will need thorough testing, but I think it's worth it.

yaustar: I will have another look at this as I vaguely remember there being an issue where something could be 'loaded' but the resource not be ready but can't remember the details

Maksims:
Prefiltered cubemap has a complex "loaded" state:
1. Prefiltered data can be loaded only.
2. Faces can be loaded only.
3. Both can be loaded.

And depending on the situation, they all are valid cases. This is made to prevent unnecessary face loading (when they are not used) while loading prefiltered data for PBR.

yaustar:
Had a test against the test cases that maksims listed above. In all cases, `.loaded` is true if the asset is used in the scene as skybox.

I also tested it as environment with a material and had the same results.

The case where this would potentially fail is if `loadFaces` is called on the cubemap asset after only the prefilter'ed data is loaded.

While the faces are loading, the cubemap asset loaded flag would still be true. It's an edge case that I think is fine to leave for now?

slimbuck:
> Prefiltered cubemap has a complex "loaded" state:
> 
> 1. Prefiltered data can be loaded only.
> 2. Faces can be loaded only.
> 3. Both can be loaded.
> 
> And depending on the situation, they all are valid cases. This is made to prevent unnecessary face loading (when they are not used) while loading prefiltered data for PBR.

The complexity originates from the fact that a cubemap asset contains both a cubemap and/or prefiltered lighting data. These two things should not be the same asset. We can't easily change it now, but my hope is we can add a new prefiltered asset in future (perhaps storing the 2d equirect image directly, perhaps as png) and then deprecate the magical invisible prefiltered data on the cubemap.

yaustar:
> Yeah tbh I kinda expected the cubemap asset to handle loaded correctly (since I rewrote it! 😀), but am less sure about all the other asset types. They'd need proper testing.

To be defensive, maybe we should do the following?

```
if (this.loaded || this.resource) {
```

src/asset/asset.js
```diff
@@ -464,7 +464,7 @@ class Asset extends EventHandler {
     ready(callback, scope) {
         scope = scope || this;
 
-        if (this.resource) {
+        if (this.loaded) {
```
Maksims, line 467: Please, ensure logic described above for cubemaps are still present to prevent unnecessary downloads.

yaustar: I've double tested the following:

- Not preloaded, no prefiltered cubemap, on a material environment map that is used in the scene - Only faces are loaded
- Preloaded, no prefiltered cubemap, on a material environment map that is used in the scene - Only faces are loaded
- Not preloaded, prefiltered cubemap generated, on a material environment map that is used in the scene - Only prefiltered data is loaded
- Preloaded, prefiltered cubemap generated, on a material environment map that is used in the scene - Only prefiltered data is loaded

- Not preloaded, no prefiltered cubemap, asset.load called by code - Only faces are loaded
- Preloaded, no prefiltered cubemap, asset.load called by code - Only faces are loaded
- Not preloaded, prefiltered cubemap generated, asset.load called by code - Only prefiltered data is loaded
- Preloaded, prefiltered cubemap generated, asset.load called by code - Only prefiltered data is loaded

- Not preloaded, no prefiltered cubemap, used as skybox, mip level 2 - Only faces are loaded
- Preloaded, no prefiltered cubemap, used as skybox, mip level 2 - Only faces are loaded
- Not preloaded, prefiltered cubemap generated, used as skybox, mip level 2 - Only prefiltered data is loaded
- Preloaded, prefiltered cubemap generated, used as skybox, mip level 2 - Only prefiltered data is loaded
- Not preloaded, prefiltered cubemap generated, used as skybox, mip level 1 - Only faces and prefiltered data are loaded
- Preloaded, prefiltered cubemap generated, used as skybox, mip level 1 - Only faces and prefiltered data are loaded

Have I missed any other possible test cases?

yaustar: I can't see any where in the code base where .ready() is used for cubemap texture loading internally in the engine so we should have the same behaviour as before

yaustar:
Created a project that has all the assets and tests ready() against them. 

https://playcanvas.com/project/927453/overview

slimbuck (approved):

PR #4255, opened 2022-05-17
Use script tag for engine in examples

slimbuck:
Use script tag for loading engine which makes debugging etc much easier.

Also filter out non-directories when reading category names during build.

### DIFF
--- a/examples/scripts/iframe/index.mustache
+++ b/examples/scripts/iframe/index.mustache
@@ -66,16 +66,12 @@
         function loadEngine(callback) {
             const enginePath = urlParams && urlParams.get('use_local_engine') || "{{{enginePath}}}";
 
-            fetch(enginePath)
-                .then(function(response) { return response.text() })
-                .then(function(data) {
-                    var module = {
-                        exports: {}
-                    };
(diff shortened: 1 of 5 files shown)


examples/scripts/iframe/index.mustache
```diff
@@ -66,16 +66,12 @@
...
             const enginePath = urlParams && urlParams.get('use_local_engine') || "{{{enginePath}}}";
 
-            fetch(enginePath)
-                .then(function(response) { return response.text() })
-                .then(function(data) {
-                    var module = {
-                        exports: {}
-                    };
-                    window.pc = (Function('module', 'exports', data).call(module, module, module.exports), module).exports;
-                    window.top.pc = window.pc;
```
ellthompson, line 76: Can we retain this line so that `pc` is accessible from the console when viewing the examples browser?

ellthompson (approved):
Approving with a single comment.

examples/scripts/iframe/index.mustache
slimbuck: Yeah sure.

Just fyi: if you select the example iframe in the "Javascript context" dropdown in Chrome's console tab you can also access pc there.

Issue #123, opened 2014-12-18
Convex Hulls


willeastcott:
We should add support for convex hulls.
- Either we should add a check box called 'Convex' to the collision component's 'Mesh' type. Or we create a new type called 'Convex Hull' (or 'Convex Mesh') that also takes a model asset.
- We then just need to use the btConvexHullShape API to create the shape.

The benefit is that we can have two triangle mesh based objects colliding with each other dynamically.

PR #4256, opened 2022-05-17
Update torus primitive

slimbuck:
This PR:
- adds support for torus to `getShapePrimitive`
- calculates torus area
- sets the second set of UVs on torus prim (like the others)

! Make torus great again !

### DIFF
--- a/src/scene/procedural.js
+++ b/src/scene/procedural.js
@@ -344,6 +344,7 @@ function createTorus(device, opts) {
     const options = {
         normals: normals,
         uvs: uvs,
+        uvs1: uvs,
         indices: indices
     };
 
@@ -1039,6 +1040,11 @@ function getShapePrimitive(device, type) {
             case 'sphere':
                 mesh = createSphere(device, { radius: 0.5 });
                 area = { x: Math.PI, y: Math.PI, z: Math.PI, uv: 1 };
+                break;
+
+            case 'torus':
+                mesh = createTorus(device, { tubeRadius: 0.2, ringRadius: 0.3 });
+                area = { x: Math.PI * 0.5 * 0.5 - Math.PI * 0.1 * 0.1, y: 0.4, z: 0.4, uv: 1 };
                 break;
 
             default:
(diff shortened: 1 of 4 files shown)


mvaligursky (approved):
Nice! I look forward seeing it in the Editor in the future.

src/scene/procedural.js
```diff
@@ -1041,6 +1042,11 @@ function getShapePrimitive(device, type) {
                 area = { x: Math.PI, y: Math.PI, z: Math.PI, uv: 1 };
                 break;
 
+            case 'torus':
+                mesh = createTorus(device, { tubeRadius: 0.2, ringRadius: 0.3 });
+                area = { x: Math.PI * 0.5 * 0.5 - Math.PI * 0.1 * 0.1, y: 0.4, z: 0.4, uv: 1 };
+                break;
+
```
willeastcott, line 1049: `getShapePrimitive` is only called from the `render` and `model` components. But those components don't list `torus` as a valid shape for `type`. Is your intention here to add support for it to those components?

slimbuck: oh yes you're right the docs need updating.

willeastcott:
I'm cool with this. Some notes:

1. One of the reasons torus was never exposed to the Editor is that there's no torus collision shape available. And it seemed nice to keep the visual and physics list of shapes the same.
2. If we want to add more procedural shapes in the future, where should they live? Do we want the core library bloated with torus knots and teapots....or should shapes be in some external sets of library scripts?

willeastcott (approved):

LeXXik:
Just to note, that torus could collide with other dynamic mesh, if it were created as convex hull mesh. Currently all mesh colliders use `btBvhTriangleMeshShape`, which is optimized in Bullet for static meshes. Convex hulls are more expensive, but they can collide with each other. There was an ancient issue about it: https://github.com/playcanvas/engine/issues/123 (still open miraculously)

I would vote for keeping them as external, like the post effects. This would allow add teapots and whatnot.

Issue #4271, opened 2022-05-25
Implement modular and more explicit render passes


mvaligursky:
Currently, the PlayCanvas engine does not implement explicit render passes (rendering processes with outputs going to different render targets), and rendering is driven directly by ForwardRenderer, making it pretty complex and harder to extend. This includes rendering of the main scene, depth and color grab passes, shadow maps, shadow and cookie atlases and multiple post-processing effects and similar.

We're facing multiple issues with this system currently:
- WebGPU requires the rendering to be submitted using render passes, to allow optimal performance on tiled architectures, in relation to load and store operations on the render attachments.
- WebGL can benefit from similar organisation, and use invalidateFramebuffer to save considerable memory bandwidth on tiled architectures.
- Without clear understanding of render passes, the management of resolves of multisampled attachments as well as mipmap generation (specifically for cube maps) becomes complicated, and we are wasting a lot of performance performing those when not needed.

We've identified a solution to improve this, which is the **FrameGraph**. It allows the frame rendering to be described in a graph form, where the nodes represent render passes (rendering into a render target). Edges represent dependencies of those render targets. This will allow us to solve the mentioned issues, and implement more flexible and performant rendering architecture.

References:
https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/invalidateFramebuffer
https://www.gdcvault.com/play/1024612/FrameGraph-Extensible-Rendering-Architecture-in
https://www.dropbox.com/s/89pvx9obhv6m0pk/High%20Quality,%20High%20Performance%20Graphics%20in%20Filament%20(Handout).pdf?dl=0

Adriaaaaan:
I'd just like to say, I think its awesome you're working on these kinds of tasks.  Its so easy nowadays to focus on visual flashy features but its the architecture work that really helps to deliver them in a practical, maintainable and performant manner.

mvaligursky:
I'll close this, as on a generic level this has been done already, and will open separate issues on further work in this area.

mvaligursky:
https://github.com/playcanvas/engine/issues/5686

PR #4272, opened 2022-05-25
Basic implementation of RenderPass based architecture

mvaligursky:
This is a first step in the implementation of https://github.com/playcanvas/engine/issues/4271. ForwardRenderer, instead of directly rendering the scene, performs two steps:
1. analyses the scene, and constructs a FrameGraph of rendering passes. Only layers rendered by the cameras are fully split into render passes, other rendering systems (shadows, grab passes, post-effects and similar) are inserted into the frame graph as black boxes. Those will be further broken down into actual render passes in the future.
2. renders the generated render passes

Other changes:
- making Debug.setTrace public, to allow external applications to enable log tracing
- added few trace ids for debug logging related to render passes

example of the trace output for post-effects engine example
Screenshot 2022-05-25 at 11 47 48

and here are the frame passes from the same example when postprocessing is disabled - this results in a smaller number of render passes, which in the future PRs will translate to a smaller number of multi-sampled resolves:
Screenshot 2022-05-25 at 11 49 41

### DIFF
--- a/scripts/utils/cubemap-renderer.js
+++ b/scripts/utils/cubemap-renderer.js
@@ -70,6 +70,7 @@ CubemapRenderer.prototype.initialize = function () {
 
         // render target, connected to cubemap texture face
         var renderTarget = new pc.RenderTarget({
+            name: `CubemapRenderer-Face${i}`,
             colorBuffer: this.cubeMap,
             depth: this.depth,
             face: i,
--- a/scripts/utils/cubemap-renderer.js
+++ b/scripts/utils/cubemap-renderer.js
@@ -70,7 +70,7 @@ CubemapRenderer.prototype.initialize = function () {
 
         // render target, connected to cubemap texture face
         var renderTarget = new pc.RenderTarget({
-            name: `CubemapRenderer-Face${i}`,
+            name: 'CubemapRenderer-Face' + i,
             colorBuffer: this.cubeMap,
             depth: this.depth,
             face: i,
--- a/src/core/debug.js
+++ b/src/core/debug.js
@@ -167,4 +167,23 @@ class Debug {
     }
 }
 
-export { Debug };
+/**
+ * A helper debug functionality.
+ *
+ * @ignore
+ */
+class DebugHelper {
+    /**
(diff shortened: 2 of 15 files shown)


src/core/debug.js
```diff
@@ -24,7 +44,11 @@ class Debug {
     /**
      * Enable or disable trace channel
      *
-     * @param {string} channel - Name of the trace channel
+     * @param {string} channel - Name of the trace channel. Can be
```
willeastcott, line 47:
```
     * @param {string} channel - Name of the trace channel. Can be:
```

src/scene/render-pass.js
```diff
@@ -0,0 +1,21 @@
...
+ * A render pass represents a node in the frame graph, and encapsulates a system which
+ * renders to a render target using an execution callback.
+ *
+ * @ignore
+ */
+class RenderPass {
+    /** @type {string} */
+    name;
+
+    constructor(renderTarget, execute) {
```
willeastcott, line 13: Might be a good idea to add JSDoc comments for this constructor. `execute` in particular is unexplained here.

src/scene/renderer/forward-renderer.js
```diff
@@ -81,6 +86,9 @@ const _tempMaterialSet = new Set();
  * The forward renderer renders {@link Scene}s.
  */
 class ForwardRenderer {
+    /** @type {boolean} */
```
willeastcott, line 89: Shouldn't this have `@ignore`?

willeastcott (approved):

src/scene/render-pass.js
mvaligursky: yep, done

scripts/utils/cubemap-renderer.js
```diff
@@ -70,6 +70,7 @@ CubemapRenderer.prototype.initialize = function () {
 
         // render target, connected to cubemap texture face
         var renderTarget = new pc.RenderTarget({
+            name: `CubemapRenderer-Face${i}`,
```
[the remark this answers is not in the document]
mvaligursky: great point, on it

PR #4279, opened 2022-05-29
Make touch events passive like mouse events

erikdubbelboer:
Prevent console messages like:
 ```
[Violation] Added non-passive event listener to a scroll-blocking 'touchmove' event. Consider marking event handler as 'passive' to make the page more responsive. See https://www.chromestatus.com/feature/5745543795965952
```

See also: https://github.com/playcanvas/engine/blob/9a0697c6b2cd521db510df49f90028f414006696/src/input/element-input.js#L421-L426

### DIFF
--- a/src/input/touch-device.js
+++ b/src/input/touch-device.js
@@ -1,3 +1,5 @@
+import { platform } from '../core/platform.js';
+
 import { EventHandler } from '../core/event-handler.js';
 
 import { TouchEvent } from './touch-event.js';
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

willeastcott:
The mouse module sets the third parameter of `removeEventHandler` as well as `addEventHandler`. Presumably, touch should do the same?

erikdubbelboer:
That's giving me linting errors as `passive` is not a valid option for `removeEventListener`. See: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#options
You want me to still add it?

I'm also getting this linting error in the mouse module:
Screenshot 2022-05-31 at 10 04 59

willeastcott:
Ah yeah, you're right. I just read this: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener
I'm wondering whether the best solution would be to completely drop the third parameter from the calls to `removeEventListener`.

erikdubbelboer:
I would keep the `false` there just for clarity.

willeastcott:
OK, sounds good. 😄

PR #4280, opened 2022-05-30
Add support for .opus audio files

erikdubbelboer:
These files are smaller than .mp3 and .ogg and are supported by all browsers except Safari [1].

1. https://caniuse.com/?search=opus

### DIFF
--- a/src/resources/audio.js
+++ b/src/resources/audio.js
@@ -34,15 +34,16 @@ const ie = (function () {
     return false;
 })();
 
-const toMIME = {
-    '.ogg': 'audio/ogg',
-    '.mp3': 'audio/mpeg',
-    '.wav': 'audio/x-wav',
-    '.mp4a': 'audio/mp4',
-    '.m4a': 'audio/mp4',
-    '.mp4': 'audio/mp4',
-    '.aac': 'audio/aac'
-};
+const supportedExtensions = [
+    '.ogg',
+    '.mp3',
+    '.wav',
+    '.mp4a',
+    '.m4a',
+    '.mp4',
+    '.aac',
+    '.opus'
+];
 
 /**
  * Resource handler used for loading {@link Sound} resources.
@@ -73,7 +74,7 @@ class AudioHandler {
     _isSupported(url) {
         const ext = path.getExtension(url);
 
-        if (toMIME[ext]) {
+        if (supportedExtensions.includes(ext)) {
             return true;
         }
         return false;
--- a/src/resources/audio.js
+++ b/src/resources/audio.js
@@ -74,7 +74,7 @@ class AudioHandler {
     _isSupported(url) {
         const ext = path.getExtension(url);
 
-        if (supportedExtensions.includes(ext)) {
+        if (supportedExtensions.indexOf(ext) > -1) {
             return true;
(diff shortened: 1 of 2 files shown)


src/resources/audio.js
```diff
@@ -34,15 +34,16 @@ const ie = (function () {
...
 
-const toMIME = {
-    '.ogg': 'audio/ogg',
-    '.mp3': 'audio/mpeg',
-    '.wav': 'audio/x-wav',
-    '.mp4a': 'audio/mp4',
-    '.m4a': 'audio/mp4',
-    '.mp4': 'audio/mp4',
-    '.aac': 'audio/aac'
-};
```
yaustar, line 45: What's the reason for changing to use extensions instead of MIMEs?

erikdubbelboer: The mime wasn't used at all. That's why I also renamed it to `supportedExtensions`, because it was only used to do a lookup if the extension was supported.

src/resources/audio.js
```diff
@@ -73,7 +74,7 @@ class AudioHandler {
     _isSupported(url) {
         const ext = path.getExtension(url);
 
-        if (toMIME[ext]) {
+        if (supportedExtensions.includes(ext)) {
```
willeastcott, line 77: IE11 doesn't have `Array.prototype.includes`. Maybe change to `supportExtensions.indexOf(ext) !== -1`

erikdubbelboer: Should I also fix it here? This is the only other place where Array.prototype.includes is used outside of tests and examples:
https://github.com/playcanvas/engine/blob/08b1f08622f128d1a6bd9df21b8b48a3cd17b2f3/src/framework/components/anim/system.js#L44
Or you want this in a separate pull?

That it's also used here tells me that nobody ever tests it on IE11 and it might be broken already in other places as well?

willeastcott: We do test on IE11 but only the examples browser application. So if none of the examples hit that line, the test plan would pass. I just went to check, but since the last release, a new IE11-incompatible change has been introduced: https://github.com/playcanvas/engine/issues/4287
When that's fixed, I'll check the anim-related examples.
I would leave that occurrence of `includes` out of this PR. Feel free to open a new one though.

erikdubbelboer: Ok I removed `supportedExtensions.includes` from my pull only.

src/resources/audio.js
```diff
@@ -73,7 +74,7 @@ class AudioHandler {
     _isSupported(url) {
         const ext = path.getExtension(url);
 
-        if (toMIME[ext]) {
+        if (supportedExtensions.indexOf(ext) > -1) {
```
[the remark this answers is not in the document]
erikdubbelboer: Fixed.

willeastcott:
This is great. Thanks @erikdubbelboer! 🎉

erikdubbelboer:
@willeastcott I'm not sure if something needs to happen in the editor as well to be able to support .opus files?

willeastcott:
Yes, we'd need to do some work there too. First step would be to open an issue here:

https://github.com/playcanvas/editor/issues

PR #4283, opened 2022-05-30
Fix color mix in the linear fog

querielo:
Colors has to be mixed in the linear space (not factor) similar to [fogExp.js](https://github.com/playcanvas/engine/blob/main/src/graphics/program-lib/chunks/lit/frag/fogExp.js#L10).


slimbuck:
How very strange. Just checked the history and it seems the `gammaCorrect` call has been in there from the very first version. Might break existing scenes...

slimbuck (approved):

willeastcott:
Thanks for this @querielo

PR #4285, opened 2022-05-30
Return original source if preprocessor fails

slimbuck:
It seems a little safer to return the original shader source if preprocessing fails (rather than null) since the shader may still be valid.

Debug build of the engine outputs the details needed to debug failing shader.

### DIFF
--- a/src/core/preprocessor.js
+++ b/src/core/preprocessor.js
@@ -220,7 +220,7 @@ class Preprocessor {
 
         if (error) {
             Debug.error("Failed to preprocess shader: ", { source: originalSource });
-            return null;
+            return originalSource;
         }
 
         return source;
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

mvaligursky:
It is safer for now, but when we implement uniform buffers or other features which depend on it, it won't work at all. So I would prefer we return null.

slimbuck:
The problem with returning null right now is that you only get feedback in debug build. The shader currently fails to compile without any context (since shader = null). At least a failing shader in future will print context, because there will be a shader printout. (And of course in debug build you'll also hopefully get preprocessor errors).

mvaligursky (approved):

Issue #2579, opened 2020-11-30
Update UI image and text elements to work more seamlessly with batch groups 


yaustar:
At the moment, text and image elements need to be added/removed from their batch group when updated (eg text is changed).

This should be handled in the engine if they are part of a batch group when changed and be updated when the batch group is marked as dirty

mvaligursky:
related: https://github.com/playcanvas/engine/issues/2578

yaustar:
Example project: https://playcanvas.com/editor/scene/1046807

Press Space to change the text and mark the batch group as dirty to show the issue:
![Kapture 2020-12-03 at 16 42 15](https://user-images.githubusercontent.com/16639049/101059775-8baa0f80-3586-11eb-9130-40dfb20534d7.gif)

Press R to change the text with the workaround of disabling and re-enabling the element to fix the batcher:
![Kapture 2020-12-03 at 16 43 14](https://user-images.githubusercontent.com/16639049/101059874-aa100b00-3586-11eb-951d-d7528c0830f9.gif)

PR #4289, opened 2022-05-31
Element component: mark a batch group as dirty if an attribute has been changed

querielo:
Fixes [#2579](https://github.com/playcanvas/engine/issues/2579)

MR makes batch group dirty if an element has been changed. So, a user don't have to mark a batch group as dirty manually, because it is error-prone.

### DIFF
--- a/src/framework/components/element/component.js
+++ b/src/framework/components/element/component.js
@@ -1667,10 +1667,18 @@ function _define(name) {
             return null;
         },
         set: function (value) {
-            if (this._text) {
-                this._text[name] = value;
-            } else if (this._image) {
-                this._image[name] = value;
+            if (this._text && this._text[name] !== value ||
+                this._image && this._image[name] !== value) {
+
+                if (this._text) {
+                    this._text[name] = value;
+                } else if (this._image) {
+                    this._image[name] = value;
+                }
+
+                if (this.batchGroupId !== -1 && this.system.app.batcher) {
+                    this.system.app.batcher.markGroupDirty(this.batchGroupId);
+                }
             }
         }
     });


mvaligursky:
Looks great! I'd probably add a function like this to the class

```
    _dirtyBatch() {
        if (this.batchGroupId !== -1) {
            this.system.app.batcher?.markGroupDirty(this.batchGroupId);
        }
    }
```

and then use this
```
        set: function (value) {
            if (this._text && this._text[name] !== value) {
                this._text[name] = value;
                this._dirtyBatch();
            } else if (this._image && this._image[name] !== value) {
                this._image[name] = value;
                this._dirtyBatch();
            }
        }
```

it would allow us to call _dirtyBatch from multiple places.

querielo:
> Looks great! I'd probably add a function like this to the class
> 
> ```
>     _dirtyBatch() {
>         if (this.batchGroupId !== -1) {
>             this.system.app.batcher?.markGroupDirty(this.batchGroupId);
>         }
>     }
> ```
> 
> and then use this
> 
> ```
>         set: function (value) {
>             if (this._text && this._text[name] !== value) {
>                 this._text[name] = value;
>                 this._dirtyBatch();
>             } else if (this._image && this._image[name] !== value) {
>                 this._image[name] = value;
>                 this._dirtyBatch();
>             }
>         }
> ```
> 
> it would allow us to call _dirtyBatch from multiple places.

**_Done_**

mvaligursky (approved):
looks like a great change to me!

jpauloruschel (changes requested):

src/framework/components/element/component.js
```diff
@@ -1667,10 +1673,12 @@ function _define(name) {
             return null;
         },
         set: function (value) {
-            if (this._text) {
+            if (this._text && this._text[name] !== value) {
```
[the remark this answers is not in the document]
querielo: Done: https://github.com/playcanvas/engine/pull/4289/commits/a09bb294585e4cded5a62c2e418651b7db1467a3

src/framework/components/element/component.js
```diff
@@ -1668,8 +1674,16 @@ function _define(name) {
         },
         set: function (value) {
             if (this._text) {
+                if (this._text[name] !== value) {
+                    this._dirtyBatch();
```
jpauloruschel, line 1678: Shouldn't the callback for dirtifying be *after* the value is changed?

mvaligursky: it does not matter, this just makes it dirty .. the actual value is not used in this function.
And the batch rebuilds after all scripts are done executing.

jpauloruschel: Ahh ok! All good, then. Thanks, @querielo !

PR #4292, opened 2022-06-01
Deprecated ForwardRenderer.renderComposition to keep compatibility.

mvaligursky:
.. to keep application using this functioning.

### DIFF
--- a/src/deprecated/deprecated.js
+++ b/src/deprecated/deprecated.js
@@ -548,6 +548,11 @@ Object.defineProperty(Batch.prototype, 'model', {
     }
 });
 
+ForwardRenderer.prototype.renderComposition = function (comp) {
+    Debug.deprecated('pc.ForwardRenderer#renderComposition is deprecated. Use pc.AppBase.renderComposition instead.');
+    getApplication().renderComposition(comp);
+};
+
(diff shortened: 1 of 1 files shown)


slimbuck (approved):

willeastcott:
From an architectural point of view, `AppBase` feels like a strange place to find a function like `renderComposition`. However, I might expect an app to own a scene and a renderer, where the scene can be rendered with the renderer. Thoughts?

mvaligursky:
> From an architectural point of view, `AppBase` feels like a strange place to find a function like `renderComposition`. However, I might expect an app to own a scene and a renderer, where the scene can be rendered with the renderer. Thoughts?

I agree with you, but the Scene is not currently usable as a container to render, as we only can have a single scene per application. At the moment, the RenderComposition is set up as the thing you can have multiple copies of (even this does not really work, and there are issues with it), so the current API matches this limitation. Longer term, we're hoping to allow to have multiple scenes and render those as needed.
That's one of the reasons APpBase.renderComposition is not public as well .. even though people for a lack of better way sometimes use it.

willeastcott:
If it's not public, then fine - it'd be good to refactor this in future. Approving...

Issue #3996, opened 2022-02-07
Sourcemaps debugging map seems to be off by 1 line in Chrome Devtools 


yaustar:
See comment in PR https://github.com/playcanvas/engine/pull/3952#issuecomment-1026792082

This works fine in Firefox so we wonder if this is a Chrome specific bug/behaviour and if it's something we can workaround?

PR #4293, opened 2022-06-01
Get sourcemaps working

slimbuck:
Fixes: #3996

This PR updates the engine rollup build script as follows:
- remove rollup-replace plugin for setting module version and revision, use jscc for this instead
- disable tabsToSpaces and shaderChunks processing in debug build, which appears to break sourcemaps (presumably because sourcemaps aren't returned from these functions)

With these changes in place sourcemaps work correctly in Chrome.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,5 +1,4 @@
 import babel from '@rollup/plugin-babel';
-import replace from '@rollup/plugin-replace';
 import strip from '@rollup/plugin-strip';
 import { createFilter } from '@rollup/pluginutils';
 import dts from 'rollup-plugin-dts';
@@ -26,14 +25,14 @@ function getBanner(config) {
     ].join('\n');
 }
 
-function spacesToTabs() {
+function spacesToTabs(enable) {
     const filter = createFilter([
         '**/*.js'
     ], []);
 
     return {
         transform(code, id) {
-            if (!filter(id)) return undefined;
+            if (!enable || !filter(id)) return undefined;
             return {
                 code: code.replace(/ {2}/g, '\t'),
                 map: null
@@ -42,7 +41,7 @@ function spacesToTabs() {
     };
 }
 
-function shaderChunks(removeComments) {
+function shaderChunks(enable) {
     const filter = createFilter([
         '**/*.vert.js',
         '**/*.frag.js'
@@ -50,7 +49,7 @@ function shaderChunks(removeComments) {
 
     return {
         transform(code, id) {
-            if (!filter(id)) return undefined;
+            if (!enable || !filter(id)) return undefined;
 
             code = code.replace(/\/\* glsl \*\/\`((.|\r|\n)*)\`/, (match, glsl) => {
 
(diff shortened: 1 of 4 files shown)


mvaligursky (approved):

rollup.config.js
```diff
@@ -200,22 +197,33 @@ function buildTarget(buildType, moduleFormat) {
...
-                _PROFILER: 1
+                ...sdkVersion,
+                ...{
+                    _DEBUG: 1,
+                    _PROFILER: 1
+                }
             },
-            keepLines: true
+            keepLines: true,
+            sourcemap: 'inline'
```
kungfooman, line 215: `sourcemap` is a `Boolean` and not a `string` and by default `true`, so this can be completely removed:

https://github.com/aMarCruz/jscc/blob/7f8634b9677097c0e16311caae27937ff69d7f51/index.d.ts#L86-L93

rollup.config.js
```diff
@@ -200,22 +197,33 @@ function buildTarget(buildType, moduleFormat) {
...
     const jsccOptions = {
         debug: {
             values: {
-                _DEBUG: 1,
-                _PROFILER: 1
+                ...sdkVersion,
+                ...{
+                    _DEBUG: 1,
+                    _PROFILER: 1
+                }
```
kungfooman, line 212: This looks a bit verbose/strange, why not just add the properties directly? Saves two lines and indentation.

kungfooman, line 212: This looks a bit verbose/strange, why not just add the properties directly? Saves two lines and indentation.

rollup.config.js
```diff
@@ -200,22 +197,33 @@ function buildTarget(buildType, moduleFormat) {
...
         },
         profiler: {
             values: {
-                _PROFILER: 1
-            },
-            keepLines: true
+                ...sdkVersion,
+                ...{
+                    _PROFILER: 1
+                }
```
kungfooman, line 225: Same here, this should be a simple property?

kungfooman, line 225: Same here, this should be a simple property?

kungfooman:
GitHub is acting strange for some reason, first the comments doubled and now they are still pending:

image

I will just write it once more and try to explain it here.

Code like this:

```js
o = {
  ...{
    ...{
      ...{
        ...{
          ...{
            ...{
              ...{
                ...{
                    a: 123
                }
              }
            }
          }
        }
      }
    }
  }
}
```

Is basically the same as: `o = {a: 123}` e.g.

https://github.com/playcanvas/engine/blob/bf628f23b70bddd01c247cdc3d4b809f4ef6afb5/rollup.config.js#L205-L216

Line 215: `sourcemap` is a `Boolean` and not a `string` and by default `true`, so this can be completely removed:

https://github.com/aMarCruz/jscc/blob/7f8634b9677097c0e16311caae27937ff69d7f51/index.d.ts#L86-L93

slimbuck:
Hi @kungfooman ,

Sourcemaps true will create a separate .map file. Sourcemaps inline will place them in the output module.

As for the destructuring assignment, you may have missed that _PROFILE etc is being combined with sdkVersion object. I didn't want to duplicate the version members multiple times.

This was the most straightforward method I could think of, but agree it's still rather messy.

slimbuck:
(Oh and jscc declares the variable a boolean, but says to match it to the output sourcemap value, which is inline. I chose to keep inline because it's a little more explicit and works).

kungfooman:
Hi @slimbuck thank you for your quick answer, I made a little example to demonstrate what I mean:

```js
const version = 'version 1';
const revision = 'revision 2';
const sdkVersion = {
  _CURRENT_SDK_VERSION: version,
  _CURRENT_SDK_REVISION: revision
};
const values1 = { 
  ...sdkVersion, 
  _DEBUG: 1, 
  _PROFILER: 1
}
const values2 = { 
  ...sdkVersion,
  ...{
    _DEBUG: 1, 
    _PROFILER: 1
  }
}
console.log(values1, values2);
```

`values1` and `values2` are equal and still combined with `sdkVersion`, while `values1` is using two lines less.

> Sourcemaps true will create a separate .map file. Sourcemaps inline will place them in the output module.

You are defining `inline` sourcemaps in the `jscc` options, which is the preprocessor only? There is another inline option for the `rollup` targets (`outputOptions`), which I think you refer to.

PR #4297, opened 2022-06-06
Engine example using multiple cameras / viewports

mvaligursky:
- example showing / testing multiple viewports
- needed for testing of render-pass refactor

Screenshot 2022-06-06 at 12 44 55

### DIFF
--- a/examples/assets/models/chess-board.txt
+++ b/examples/assets/models/chess-board.txt
@@ -1,0 +1,11 @@
+Model Information:
+* title:	Chess Board
+* source:	https://sketchfab.com/3d-models/chess-board-901eeeca884f4622ac37b7e8f7cb82c3
+* author:	Idmental (https://sketchfab.com/idmental.id)
+
+Model License:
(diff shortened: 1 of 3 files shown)


willeastcott:
The specular on the board surface looks a bit....flat/bright from an angle. Why is that?

mvaligursky:
> The specular on the board surface looks a bit....flat/bright from an angle. Why is that?

I'm not entirely sure .. both IBL and directional light give us this look. It also matches pretty well with other viewers. I'd say the material properties (roughness or similar) are not set up the best for this.

willeastcott (approved):

Maksims:
The environment map with a strong light source (in the texture) combined with directional light and shadows - does not go well.
To improve, reducing exposure of a cubemap - is recommended. Also matching the direction of dynamic light to the light source on a cubemap helps as well.

willeastcott:
Whoops. I just noticed the chess board GLB is 7.6MB. We should avoid checking in such big binaries where possible to keep the clone size of the repo nice and lean.

PR #4309, opened 2022-06-08
Add examples develop script

slimbuck:
This PR adds develop script so `npm run develop` launches a server and listens for changes to examples.


willeastcott (approved):

ellthompson (approved):

mvaligursky:
currently this is what I use to work with engine examples and debug engine

engine: 
`target=debug_es5 npm build watch`
examples:
```
npm run watch
npm run watch:iframes:debug
```
maybe both of these should be added to the develop?

Issue #4308, opened 2022-06-08
Shadow markup requires color attribute


slimbuck:
The markup introduced in https://github.com/playcanvas/engine/pull/4302 fails at runtime unless the `shadow` tag specifies a `color` attribute.

This is because `fromString` is being passed a null [here](https://github.com/playcanvas/engine/blob/main/src/framework/components/element/text-element.js#L514).

Ideally color would be unaffected in this case, but it could also just take the global shadow color instead.

cc @querielo

querielo:
Thanks for the report. Tried to fix it [here](https://github.com/playcanvas/engine/pull/4311).

PR #4311, opened 2022-06-09
Fix: Shadow markup requires color attribute

querielo:
Fixes [#4308](https://github.com/playcanvas/engine/issues/4308)

Changed encoding of shadow offsets/outline thickness in WebGL buffers.

Suggested: **enableMarkups** switches uniforms to attributes even if there is no tags. So, if user need to improve batching they just set **enableMarkups**.

Tests: added 2 tests according to the issue.

### DIFF
--- a/src/framework/components/element/text-element.js
+++ b/src/framework/components/element/text-element.js
@@ -342,7 +342,10 @@ class TextElement {
         if (this._enableMarkup) {
             const results = Markup.evaluate(this._symbols);
             this._symbols = results.symbols;
-            tags = results.tags;
+            // NOTE: if results.tags is null, we assign [] to increase
+            // probability of batching. So, if a user want to use as less
+            // WebGL buffers memory as possible they can just disable markups.
+            tags = results.tags || [];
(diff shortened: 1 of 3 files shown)


src/framework/components/element/text-element.js
```diff
@@ -342,7 +342,10 @@ class TextElement {
         if (this._enableMarkup) {
             const results = Markup.evaluate(this._symbols);
             this._symbols = results.symbols;
-            tags = results.tags;
+            // NOTE: if results.tags is null, we assign [] to increase
+            // probability of batching. So, if a user want to use as less
+            // WebGL buffers memory as possible they can just disable markups.
+            tags = results.tags || [];
```
querielo, line 348: It is a suggestion that increases size of buffers if enableMarkup is true, but it also increases probability of batching.

mvaligursky (approved):
looks good to me. Any thoughts @slimbuck ?

slimbuck (approved):
Thanks for this!!

PR #4319, opened 2022-06-10
Material registry

slimbuck:
This PR adds a debug-only `MaterialRegistry` class and shader events which fire when shader compilation fails.

These are used to output the impacted materials on compile failure.

The `MaterialRegistry` class is completely removed from the codebase in non-debug builds.

### DIFF
--- a/src/graphics/shader.js
+++ b/src/graphics/shader.js
@@ -1,5 +1,7 @@
 /** @typedef {import('./graphics-device.js').GraphicsDevice} GraphicsDevice */
 
+import { EventHandler } from '../core/event-handler.js';
+
 /**
  * A shader is a program that is responsible for rendering graphical primitives on a device's
  * graphics processor. The shader is generated from a shader definition. This shader definition
@@ -9,6 +11,10 @@
  * attributes specified in the vertex shader code.
  */
 class Shader {
+    // #if _DEBUG
+    static debugEvents = new EventHandler();
+    // #endif
+
     /**
      * Creates a new Shader instance.
      *
--- a/src/graphics/shader.js
+++ b/src/graphics/shader.js
@@ -1,6 +1,6 @@
-/** @typedef {import('./graphics-device.js').GraphicsDevice} GraphicsDevice */
-
 import { EventHandler } from '../core/event-handler.js';
+
+/** @typedef {import('./graphics-device.js').GraphicsDevice} GraphicsDevice */
 
 /**
  * A shader is a program that is responsible for rendering graphical primitives on a device's
--- a/src/scene/materials/material-registry.js
+++ b/src/scene/materials/material-registry.js
@@ -3,8 +3,8 @@ import { Shader } from '../../graphics/shader.js';
 class MaterialRegistry {
     constructor() {
         this.materials = new Set();
-        Shader.debugEvents.on('compileError', (shader) => this.reportEffectedMaterials(shader));
-        Shader.debugEvents.on('linkError', (shader) => this.reportEffectedMaterials(shader));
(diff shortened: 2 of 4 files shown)


src/graphics/shader.js
```diff
@@ -1,5 +1,7 @@
 /** @typedef {import('./graphics-device.js').GraphicsDevice} GraphicsDevice */
 
+import { EventHandler } from '../core/event-handler.js';
+
```
willeastcott, line 4: While this isn't mentioned in the coding standards, typedefs come after imports. We should probably document module layout...

src/scene/materials/material-registry.js
```diff
@@ -0,0 +1,55 @@
...
+            });
+        });
+
+        if (materials.size > 0) {
+            console.error(`Effected materials: `, Array.from(materials));
+        }
+    }
+}
+
+const materialRegistry = new MaterialRegistry();
```
willeastcott, line 51: So all running applications share the same registry? 😨 Are we cool with that?

mvaligursky: I'm not. If an application won't destroy some materials, they will stay in memory as the registry would keep the references.

src/graphics/shader.js
slimbuck: done, thanks

src/graphics/shader.js
```diff
@@ -9,6 +11,10 @@
  * attributes specified in the vertex shader code.
  */
 class Shader {
+    // #if _DEBUG
+    static debugEvents = new EventHandler();
```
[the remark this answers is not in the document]
slimbuck: If we did that then shader would have to know about material.

src/scene/materials/material-registry.js
slimbuck: How would shader know which application it belongs to?

mvaligursky: I think it needs to be per device and not per application, like the other device things we store using DeviceCache.

slimbuck: The issue is materials are not application or device specific.

slimbuck: Ideally the registry would be per-application, but I don't see how.

But anyway I don't think the registry hanging onto un-destroyed materials is such a big deal because:
- they're not very big (compared to texture data for example)
- this is only a potential issue in debug builds running multiple applications
- if it ever was an issue, users could just delete their materials

And actually having access to the list of undeleted materials could be extremely useful to users, something we might want to expose in future.

mvaligursky: they are valid for use only on a single device, so they are device specific. For example shader and textures they reference are all device specific, making them the same.
But which device is not known ahead of time. Perhaps at a point the shader is assigned to the material (which has a device), the material could be added to the cache, and not in the constructor? It could even be a Set on a shader directly?

slimbuck:
BTW all of this is simply to know which materials caused the invalid shader to be generated. Perhaps there is a simpler or more straight-forward approach I haven't considered?

(Though as I commented above, having a list of materials in itself could be very useful).

mvaligursky:
> Perhaps there is a simpler or more straight-forward approach I haven't considered?

I think as mentioned in the comment, the shader could just have a Set of materials, and the material could be added/removed in material.shader setter, and also removed in material destroy function?

leonidaspir:
Oh, agreed, having a list of all materials generated is super useful. I was happy to see the PR show up. We've found it would make our life much easier when working with a project wide HW instancing system.

slimbuck:
We've decided to not merge this right now, but will continue discussing how we could maintain a list of materials.

PR #4321, opened 2022-06-10
Box-reflection example uses batmobile mesh instead of a sphere

mvaligursky:
Screenshot 2022-06-10 at 14 17 55


willeastcott:
Y'know, I'm not sure I'm a fan of the....what is it....purple? (Colorblind here). Maybe white strip lights would be better?

willeastcott (approved):

PR #4327, opened 2022-06-14
Animate morph targets by name

ellthompson:
Currently, all the morph targets that animate on a given node are included in a single AnimCurve, then referenced by their indices in the curves output data. This poses problems when morph targets are deleted from either a model asset or it's related animation asset, as the indices of each morph target in the curve will no longer match up with those found in the mesh's MorphInstance.

In this PR the glb parser instead creates a single curve for each morph target, which contains the morph target's name in the curve path. When the anim binder applies the value of each curve to their bound properties, it can then reference this morph target name to set the appropriate morph target on the MorphInstance. To support this, the MorphInstance class now contains a lookup table of morph names and allows the setting of morphTarget values using this lookup. Including individual curves for each morph target has the added benefit of making debugging much simpler as their keyframe output data is now stored in sequential order with a reference to the given morph targets name.

If morph target names are not present in the glb data, the index of each morph target is used as a fallback.

### DIFF
--- a/src/anim/binder/default-anim-binder.js
+++ b/src/anim/binder/default-anim-binder.js
@@ -78,24 +78,17 @@ class DefaultAnimBinder {
                 return DefaultAnimBinder.createAnimTarget(func, 'vector', 3, node, 'localScale');
             },
 
-            'weights': function (node) {
+            'weight': function (node, weightName) {
                 const meshInstances = findMeshInstances(node);
                 if (meshInstances) {
-                    const morphInstances = [];
                     for (let i = 0; i < meshInstances.length; ++i) {
                         if (meshInstances[i].node.name === node.name && meshInstances[i].morphInstance) {
-                            morphInstances.push(meshInstances[i].morphInstance);
+                            const morphInstance = meshInstances[i].morphInstance;
+                            const func = function (value) {
+                                morphInstance.setWeight(weightName, value[0]);
+                            };
+                            return DefaultAnimBinder.createAnimTarget(func, 'number', 1, node, `weight.${weightName}`);
                         }
-                    }
-                    if (morphInstances.length > 0) {
-                        const func = function (value) {
-                            for (let i = 0; i < value.length; ++i) {
-                                for (let j = 0; j < morphInstances.length; j++) {
-                                    morphInstances[j].setWeight(i, value[i]);
-                                }
-                            }
-                        };
-                        return DefaultAnimBinder.createAnimTarget(func, 'vector', morphInstances[0].morph._targets.length, node, 'weights');
                     }
                 }
 
--- a/src/resources/parser/glb-parser.js
+++ b/src/resources/parser/glb-parser.js
@@ -1297,7 +1297,7 @@ const createMaterial = function (gltfMaterial, textures, flipV) {
 };
 
 // create the anim structure
-const createAnimation = function (gltfAnimation, animationIndex, gltfAccessors, bufferViews, nodes) {
+const createAnimation = function (gltfAnimation, animationIndex, gltfAccessors, bufferViews, nodes, meshes) {
 
     // create animation data block for the accessor
     const createAnimData = function (gltfAccessor) {
@@ -1310,13 +1310,13 @@ const createAnimation = function (gltfAnimation, animationIndex, gltfAccessors, 
         'CUBICSPLINE': INTERPOLATION_CUBIC
     };
 
+    // Input map and output maps reference data by sampler input/output key.
     const inputMap = { };
-    const inputs = [];
-
     const outputMap = { };
-    const outputs = [];
-
-    const curves = [];
+    // The curve map stores temporary curve data by sampler index. Each curves input/output value will be resolved to an inputs/outputs array index after all samplers have been processed.
+    // Curves and outputs that are deleted from their maps will not be included in the final AnimTrack
+    const curveMap = { };
+    let outputCounter = 1;
 
     let i;
 
@@ -1326,14 +1326,12 @@ const createAnimation = function (gltfAnimation, animationIndex, gltfAccessors, 
 
         // get input data
         if (!inputMap.hasOwnProperty(sampler.input)) {
-            inputMap[sampler.input] = inputs.length;
-            inputs.push(createAnimData(gltfAccessors[sampler.input]));
+            inputMap[sampler.input] = createAnimData(gltfAccessors[sampler.input]);
         }
 
         // get output data
         if (!outputMap.hasOwnProperty(sampler.output)) {
-            outputMap[sampler.output] = outputs.length;
-            outputs.push(createAnimData(gltfAccessors[sampler.output]));
+            outputMap[sampler.output] = createAnimData(gltfAccessors[sampler.output]);
         }
 
         const interpolation =
@@ -1342,11 +1340,14 @@ const createAnimation = function (gltfAnimation, animationIndex, gltfAccessors, 
                 interpMap[sampler.interpolation] : INTERPOLATION_LINEAR;
 
         // create curve
-        curves.push(new AnimCurve(
-            [],
-            inputMap[sampler.input],
-            outputMap[sampler.output],
-            interpolation));
+        const curve = {
+            paths: [],
+            input: sampler.input,
+            output: sampler.output,
+            interpolation: interpolation
+        };
+
+        curveMap[i] = curve;
     }
 
     const quatArrays = [];
@@ -1354,8 +1355,7 @@ const createAnimation = function (gltfAnimation, animationIndex, gltfAccessors, 
     const transformSchema = {
         'translation': 'localPosition',
         'rotation': 'localRotation',
-        'scale': 'localScale',
-        'weights': 'weights'
+        'scale': 'localScale'
     };
 
     const constructNodePath = (node) => {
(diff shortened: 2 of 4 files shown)


src/anim/binder/default-anim-binder.js
```diff
@@ -78,25 +78,18 @@ class DefaultAnimBinder {
...
-            'weights': function (node) {
+            'weight': function (node, weightName) {
                 const meshInstances = findMeshInstances(node);
                 if (meshInstances) {
-                    const morphInstances = [];
                     for (let i = 0; i < meshInstances.length; ++i) {
                         if (meshInstances[i].node.name === node.name && meshInstances[i].morphInstance) {
-                            morphInstances.push(meshInstances[i].morphInstance);
+                            const morphInstance = meshInstances[i].morphInstance;
+                            const func = function (value) {
```
willeastcott, line 87: Any reason not to use an arrow function here?

src/resources/parser/glb-parser.js
```diff
@@ -1367,29 +1367,103 @@ const createAnimation = function (gltfAnimation, animationIndex, gltfAccessors,
...
+                outputMap[-outputCounter] = output;
+                const morphCurve = {
+                    paths: [{
+                        entityPath: entityPath,
+                        component: 'graph',
+                        propertyPath: [`weight.${retrieveWeightName(node.name, j)}`]
+                    }],
+                    // each morph target curve input can use the same sampler.input from the channel they were all in
+                    input: curve.input,
+                    // but each morph target curve should reference it's individual output that was just created
```
willeastcott, line 1412:
```suggestion
                // but each morph target curve should reference its individual output that was just created
```

src/resources/parser/glb-parser.js
```diff
@@ -1367,29 +1367,103 @@ const createAnimation = function (gltfAnimation, animationIndex, gltfAccessors,
...
+                    output: -outputCounter,
+                    interpolation: curve.interpolation
+                };
+                outputCounter++;
+                // add the morph target curve to the curveMap
+                curveMap[`morphCurve-${i}-${j}`] = morphCurve;
+            }
+            // after all morph targets in this curve have been included in the curveMap, this curve and it's output data can be deleted
```
willeastcott, line 1420:
```suggestion
            // after all morph targets in this curve have been included in the curveMap, this curve and its output data can be deleted
```

src/resources/parser/glb-parser.js
```diff
@@ -1367,29 +1367,103 @@ const createAnimation = function (gltfAnimation, animationIndex, gltfAccessors,
...
         });
+    }
+
+    const inputs = [];
+    const outputs = [];
+    const curves = [];
+
+    // Add each input in the map to the final inputs array. The inputMap should now reference the index of input in the inputs array instead of the input itself.
+    for (i = 0; i < Object.keys(inputMap).length; i++) {
+        const inputKey = Object.keys(inputMap)[i];
```
willeastcott, line 1439: Am I right in saying that `Object.keys(inputMap)` is loop invariant?

src/resources/parser/glb-parser.js
```diff
@@ -1367,29 +1367,103 @@ const createAnimation = function (gltfAnimation, animationIndex, gltfAccessors,
...
+        const input = inputMap[inputKey];
+        inputs.push(input);
+        inputMap[inputKey] = inputs.length - 1;
```
willeastcott, line 1442: Excuse me if I'm missing something obvious here, but are you reading a value from an object, pushing the value into an array and then writing that same value back into the object with the same key?

src/scene/morph-instance.js
```diff
@@ -183,10 +186,11 @@ class MorphInstance {
     /**
      * Sets weight of the specified morph target.
      *
-     * @param {number} index - An index of morph target.
+     * @param {number|string} key - An identifier for the morph target. Either the weight index or the weight name
      * @param {number} weight - Weight.
      */
-    setWeight(index, weight) {
+    setWeight(key, weight) {
```
willeastcott, line 192: Just throwing it out there, but does it make sense to have two functions? I'd be interested to hear @mvaligursky's preference since he has experience with this API.

src/resources/parser/glb-parser.js
ellthompson: I'm moving each input into an array and storing a reference to that values position in the array in the inputMap instead. Each curve is initially created with a reference to an input in the inputMap (so that inputs can be added / removed without updating multiple references), which I then swap out for a reference to the final storage array once all inputs / outputs have been processed.

src/resources/parser/glb-parser.js
ellthompson: Afaik the result of Object.keys will always be the same as long as the object's keys haven't been updated between loops.

src/anim/binder/default-anim-binder.js
ellthompson: Nope, done.

src/scene/morph-instance.js
mvaligursky: I thought about it yesterday, and I think I'm ok with this - the API is not used directly all that much, usually I'd expect to be driven by the animation, and an extra dictionary look up is not a big deal.

src/scene/morph-instance.js
```diff
@@ -183,10 +186,11 @@ class MorphInstance {
...
+        const index = typeof key === 'string' ? this._weightLookup[key] : key;
```
mvaligursky, line 193: we need to handle a case where the map look up fails here (invalid target name)

src/resources/parser/glb-parser.js
willeastcott: So would it be reasonable to put `const keys = Object.keys(inputMap);` before the for loop?

src/resources/parser/glb-parser.js
willeastcott: Ha, ok, I totally misread the last line. Thanks for the explanation.

src/resources/parser/glb-parser.js
ellthompson: Sure, done

src/scene/morph-instance.js
```diff
@@ -48,8 +48,11 @@ class MorphInstance {
 
         // weights
         this._weights = [];
+        this._weightLookup = {};
```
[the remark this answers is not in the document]
ellthompson: Yep, switched this to a map.

src/scene/morph-instance.js
```diff
@@ -173,22 +177,40 @@ class MorphInstance {
...
      * Gets current weight of the specified morph target.
      *
-     * @param {number} index - An index of morph target.
+     * @param {string|number} key - An identifier for the morph target. Either the weight index or the weight name
      * @returns {number} Weight.
      */
-    getWeight(index) {
-        return this._weights[index];
+    getWeight(key) {
+        if (typeof key === 'string') {
```
mvaligursky, line 184: I'd create _getWeightIndex function to avoid duplicating code in set and get. Also, no need to do Number.IsInteger .. Map.get returns undefined when the key does not exist.

src/resources/parser/glb-parser.js
```diff
@@ -1367,29 +1367,107 @@ const createAnimation = function (gltfAnimation, animationIndex, gltfAccessors,
...
+    let keys;
+
+    // Add each input in the map to the final inputs array. The inputMap should now reference the index of input in the inputs array instead of the input itself.
+    keys = Object.keys(inputMap);
+    for (i = 0; i < keys.length; i++) {
+        const inputKey = keys[i];
+        const input = inputMap[inputKey];
+        inputs.push(input);
+        inputMap[inputKey] = inputs.length - 1;
+    }
```
willeastcott, line 1445: Actually, instead of using `Object.keys`, why not just do something like:

```
    for (const inputKey in inputMap) {
        inputs.push(inputMap[inputKey]);
        inputMap[inputKey] = inputs.length - 1;
    }
```

src/scene/morph-instance.js
ellthompson: Done 👍

src/scene/morph-instance.js
```diff
@@ -170,25 +174,47 @@ class MorphInstance {
...
      *
-     * @param {number} index - An index of morph target.
+     * @param {string|number} key - An identifier for the morph target. Either the weight index or the weight name
      * @returns {number} Weight.
      */
-    getWeight(index) {
+    getWeight(key) {
+        let index;
+        if (typeof key === 'string') {
+            this._getWeightIndex(key);
```
mvaligursky, line 194: should the result be assigned to index?

mvaligursky: just do `const index = (typeof key === 'string') ? this._getWeightIndex(key) : key;`

src/scene/morph-instance.js
```diff
@@ -170,25 +174,47 @@ class MorphInstance {
         return clone;
     }
 
+    _getWeightIndex(key) {
+        const index = this._weightMap.get(key);
+        if (index) {
+            return this._weights[index];
+        }
+        Debug.error('Attempting to retrieve a morph target weight using a key that doesn\'t exist.');
```
mvaligursky, line 182: make the missing key part of the message, so the user knows what is missing,

src/scene/morph-instance.js
ellthompson: 👍

src/scene/morph-instance.js
ellthompson: Done

src/scene/morph-instance.js
```diff
@@ -170,25 +174,47 @@ class MorphInstance {
...
-        this._weights[index] = weight;
-        this._dirty = true;
+    setWeight(key, weight) {
+        let index;
+        if (typeof key === 'string') {
+            index = this._getWeightIndex(key);
+        } else {
+            index = key;
+        }
+        if (Number.isFinite(this._weights[key])) {
```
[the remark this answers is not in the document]
ellthompson: Sorry this was a bug from the refactor. Weights should be retrieved using the index instead of the key. Then if we cant retrieve a number from the weights array, we shouldn't be setting a new value or flagging the MorphInstance as dirty.

mvaligursky (approved):
Looks good to me, thanks.
@slimbuck - would you mind to give the binder / parser part a quick look as well, you understand that code more.

slimbuck (approved):

Issue #4325, opened 2022-06-10
Shader pass bugs


slimbuck:
Since the number of supported shadow types was increased from 5->6 (in https://github.com/playcanvas/engine/pull/3763), some existing code mistakingly categorises shaders incorrectly.

For example see:
https://github.com/playcanvas/engine/blob/main/src/graphics/program-lib/programs/standard.js#L295 and https://github.com/playcanvas/engine/blob/main/src/graphics/program-lib/programs/standard.js#L479

We must:
- change the constants defined at https://github.com/playcanvas/engine/blob/main/src/scene/constants.js#L631
    - redefine SHADER_PICK to be value 3
    - redefine SHADER_SHADOW to be value 4
- perhaps add helper functions to make sense of it all
- use the new helpers around the codebase instead of relying on the value `17` in the codebase
- consider renaming the defines from `SHADER_SHADOW` to `SHADERPASS_SHADOW`

mvaligursky:
I've refactored the code in the linked PR. I have avoided the constant renames in this PR, as that would possibly create a larger conflict with shader refactoring open PR.

PR #4328, opened 2022-06-14
Utility private ShaderPass class to abstract some code scattered in the codebase

mvaligursky:
Related to https://github.com/playcanvas/engine/issues/4325 (does not handle the renaming part yet)

### DIFF
--- a/src/scene/shader-pass.js
+++ b/src/scene/shader-pass.js
@@ -8,7 +8,7 @@ import { SHADER_SHADOW, SHADOW_COUNT, LIGHTTYPE_COUNT } from './constants.js';
  */
 class ShaderPass {
     /**
-     * Rturns true if the shader pass is a shadow shader pass.
+     * Returns true if the shader pass is a shadow shader pass.
      *
      * @param {number} pass - The shader pass.
      * @returns {boolean} - True if the pass is a shadow shader pass.
--- a/src/scene/shader-pass.js
+++ b/src/scene/shader-pass.js
@@ -43,7 +43,7 @@ class ShaderPass {
     }
 
     /**
-     * Returns a shader pass for specified light and shadow types.
+     * Returns a shader pass for specified light and shadow type.
      *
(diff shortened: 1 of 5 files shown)


src/scene/shader-pass.js
```diff
@@ -0,0 +1,60 @@
...
+import { SHADER_SHADOW, SHADOW_COUNT, LIGHTTYPE_COUNT } from './constants.js';
+
+/**
+ * A pure static utility class, responsible for math operations on the shader pass constants.
+ *
+ * @ignore
+ */
+class ShaderPass {
+    /**
+     * Rturns true if the shader pass is a shadow shader pass.
```
GSterbrant, line 11: Returns

mvaligursky: thanks!

slimbuck:
This is so much better!

Though TBH I'm now wondering why we store the precise type of shadow generation as a parse! Seems like SHADER_PARSE should be 'shadow depth' and the shadow type should be stored separately.

This should simplify all the code that does parse-specific logic.

slimbuck:
(This is actually a question @mvaligursky - what do you think?)

GSterbrant (approved):

src/scene/shader-pass.js
```diff
@@ -0,0 +1,60 @@
...
+     */
+    static toShadowType(pass) {
+        Debug.assert(ShaderPass.isShadow(pass));
+        const shadowMode = pass - SHADER_SHADOW;
+        const lightType = Math.floor(shadowMode / SHADOW_COUNT);
+        return shadowMode - lightType * SHADOW_COUNT;
+    }
+
+    /**
+     * Returns a shader pass for specified light and shadow types.
```
willeastcott, line 46:
```suggestion
     * Returns a shader pass for specified light and shadow types.
```

slimbuck (approved):

PR #4339, opened 2022-06-17
SSAO Downscale

GSterbrant:
This PR adds downscaling to the SSAO effect, similar to https://github.com/playcanvas/engine/pull/3493 but with the addition of also performing the blur pass to a downscaled target, thus reducing the pressure on texture fetches. 

It also implements the resize and reaction to changing the downscaling attribute similar to the PR mentioned above, but with the addition of adding the downscale to the post-effects.tsx example.

### DIFF
--- a/scripts/posteffects/posteffect-ssao.js
+++ b/scripts/posteffects/posteffect-ssao.js
@@ -250,8 +250,7 @@ function SSAOEffect(graphicsDevice, ssaoScript) {
             "",
             "    vec4 inCol = vec4(1.0, 1.0, 1.0, 1.0); //texture2D( uColorBuffer,  uv );",
             "",
-            "    gl_FragColor.rgb = inCol.rgb * vec3(aoVisibility); //postProcess.color.rgb = vec3(aoVisibility, pack(origin.z));",
-            "    gl_FragColor.a = 1.0;",
+            "    gl_FragColor.r = aoVisibility; //postProcess.color.rgb = vec3(aoVisibility, pack(origin.z));",
             "}",
             "",
             "void main_old()",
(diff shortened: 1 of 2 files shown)


scripts/posteffects/posteffect-ssao.js
```diff
@@ -354,42 +352,114 @@ function SSAOEffect(graphicsDevice, ssaoScript) {
...
+            graphicsDevice.webgl2 ? "" : "#extension GL_OES_standard_derivatives: enable\n",
+            "attribute vec2 aPosition;",
+            "",
+            "varying vec2 vUv0;",
+            "",
+            "void main(void)",
+            "{",
+            "    gl_Position = vec4(aPosition, 0.0, 1.0);",
+            "    vUv0 = (aPosition.xy + 1.0) * 0.5;",
+            "}"
```
mvaligursky, line 376: it seems this vertex shader is repeated 2-3x in this file .. perhaps store it in a variable and reuse?

mvaligursky:
If you have some performance / quality comments related to this change, please add them to the description, that'd be great to see.

scripts/posteffects/posteffect-ssao.js
GSterbrant: I am actually working on adding a createVertexShader for post effects so we can reuse it everywhere. It seems this code is the same for every post effect, so it would make sense to have it declared in a single place and reused across multiple post effects. I will add it to this PR.

mvaligursky (approved):

GSterbrant:
Performance at 50% scale: ~1.6x faster than 100%.

PR #4342, opened 2022-06-20
SSAO delayed resize fix

### LINKED ISSUE #4239: SSAO / depth buffer rendering does not handle resize correctly
Repro: 
- run https://playcanvas.github.io/#/graphics/post-effects
- have debug tools in Chome opened
- quickly move the splitter between webpage and the debug tools left and right, which resizes the rendering window
- notice the one frame incorrect rendering, SSAO not aligning with the scene
Screenshot 2022-05-11 at 14 54 47

GSterbrant:
Fixes #4239 

Removed the resize timeouts. Added a general resize callback to the post effects such that they can handle resize. 

Bloom needs such a callback to adapt it's internal targets (which should save performance when viewport is small).

### DIFF
--- a/scripts/posteffects/posteffect-bloom.js
+++ b/scripts/posteffects/posteffect-bloom.js
@@ -171,12 +171,47 @@ function BloomEffect(graphicsDevice) {
         fshader: bloomCombineFrag
     });
 
+    // Effect defaults
+    this.bloomThreshold = 0.25;
+    this.blurAmount = 4;
+    this.bloomIntensity = 1.25;
+
+    // Uniforms
+    this.sampleWeights = new Float32Array(SAMPLE_COUNT);
+    this.sampleOffsets = new Float32Array(SAMPLE_COUNT * 2);
+
+    this.resize(null);
+}
+
+BloomEffect.prototype = Object.create(pc.PostEffect.prototype);
+BloomEffect.prototype.constructor = BloomEffect;
+
+BloomEffect.prototype.resize = function (target) {
+
+    var width, height;
+    if (target == null) {
+        width = this.device.width;
+        height = this.device.height;
+    } else {
+        width = target.colorBuffer.width;
+        height = target.colorBuffer.height;
+    }
+
+    if (width == this.width && height == this.height)
+        return;
+
+    if (this.targets) {
+        for (let i = 0; i < this.targets.length; i++) {
+            this.targets[i].destroyFrameBuffers();
+            this.targets[i].destroyTextureBuffers();
+            this.targets[i].destroy();
+        }
+    }
+
     // Render targets
-    var width = graphicsDevice.width;
-    var height = graphicsDevice.height;
     this.targets = [];
     for (var i = 0; i < 2; i++) {
-        var colorBuffer = new pc.Texture(graphicsDevice, {
+        var colorBuffer = new pc.Texture(this.device, {
             format: pc.PIXELFORMAT_R8_G8_B8_A8,
             width: width >> 1,
             height: height >> 1,
@@ -187,26 +222,14 @@ function BloomEffect(graphicsDevice) {
         colorBuffer.addressU = pc.ADDRESS_CLAMP_TO_EDGE;
         colorBuffer.addressV = pc.ADDRESS_CLAMP_TO_EDGE;
         colorBuffer.name = 'pe-bloom';
-        var target = new pc.RenderTarget({
+        var bloomTarget = new pc.RenderTarget({
             colorBuffer: colorBuffer,
             depth: false
         });
 
-        this.targets.push(target);
+        this.targets.push(bloomTarget);
     }
-
-    // Effect defaults
-    this.bloomThreshold = 0.25;
-    this.blurAmount = 4;
-    this.bloomIntensity = 1.25;
-
-    // Uniforms
-    this.sampleWeights = new Float32Array(SAMPLE_COUNT);
-    this.sampleOffsets = new Float32Array(SAMPLE_COUNT * 2);
 }
-
-BloomEffect.prototype = Object.create(pc.PostEffect.prototype);
-BloomEffect.prototype.constructor = BloomEffect;
 
 Object.assign(BloomEffect.prototype, {
     render: function (inputTarget, outputTarget, rect) {
--- a/scripts/posteffects/posteffect-bloom.js
+++ b/scripts/posteffects/posteffect-bloom.js
@@ -229,7 +229,7 @@ BloomEffect.prototype.resize = function (target) {
 
         this.targets.push(bloomTarget);
     }
-}
+};
 
 Object.assign(BloomEffect.prototype, {
     render: function (inputTarget, outputTarget, rect) {
--- a/scripts/posteffects/posteffect-bloom.js
+++ b/scripts/posteffects/posteffect-bloom.js
@@ -189,7 +189,7 @@ BloomEffect.prototype.constructor = BloomEffect;
 BloomEffect.prototype.resize = function (target) {
 
     var width, height;
-    if (target == null) {
+    if (target === null) {
         width = this.device.width;
         height = this.device.height;
     } else {
@@ -197,11 +197,11 @@ BloomEffect.prototype.resize = function (target) {
         height = target.colorBuffer.height;
     }
 
-    if (width == this.width && height == this.height)
+    if (width === this.width && height === this.height)
         return;
 
     if (this.targets) {
-        for (let i = 0; i < this.targets.length; i++) {
+        for (var i = 0; i < this.targets.length; i++) {
             this.targets[i].destroyFrameBuffers();
             this.targets[i].destroyTextureBuffers();
             this.targets[i].destroy();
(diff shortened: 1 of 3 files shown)


src/framework/components/camera/post-effect-queue.js
```diff
@@ -373,26 +365,10 @@ class PostEffectQueue {
...
+        this.resizeRenderTargets();
     }
 
     resizeRenderTargets() {
-        if (this.resizeTimeout) {
-            clearTimeout(this.resizeTimeout);
-            this.resizeTimeout = null;
-        }
 
         this.resizeLast = now();
```
mvaligursky, line 373: should this line be removed?

scripts/posteffects/posteffect-ssao.js
```diff
@@ -17,23 +17,25 @@ function SSAOEffect(graphicsDevice, ssaoScript) {
...
+        "",
+        "varying vec2 vUv0;",
+        "",
+        "void main(void)",
+        "{",
+        graphicsDevice.webgl2 ? "    vUv0 = vec2((gl_VertexID << 1) & 2, gl_VertexID & 2);" : "   vUv0 = (aPosition.xy + 1.0) * 0.5;",
+        graphicsDevice.webgl2 ? "    gl_Position = vec4(vUv0 * 2.0f - 1.0f, 0.0f, 1.0f);" : "    gl_Position = vec4(aPosition, 0.0, 1.0);",
+        "}"
+    ].join("\n");
+
```
slimbuck, line 33: I feel like this would be simpler as two different shaders?

src/framework/components/camera/post-effect-queue.js
```diff
@@ -404,6 +380,8 @@ class PostEffectQueue {
 
         for (let i = 0, len = effects.length; i < len; i++) {
             const fx = effects[i];
+            if (fx.effect.resize !== undefined)
+                fx.effect.resize();
```
mvaligursky, line 384: should this be called after _resizeOffscreenTarget so that the effect can use the siez of those targets for resizing?

scripts/posteffects/posteffect-ssao.js
GSterbrant: Ah, oh my. Thanks for catching this, I think this shader won't work.

src/framework/components/camera/post-effect-queue.js
GSterbrant: Is it better that they depend on the input target or the graphics device size? The input target could be quarter res but the depth target full res, for example.

scripts/posteffects/posteffect-ssao.js
mvaligursky: Or just use webgl1 version for both platforms? Not sure we win much with custom webgl2 shader

slimbuck: Yeah I have no idea why the fancy bit ops on webgl2. The headers are still different though between 1 and 2.

src/framework/components/camera/post-effect-queue.js
mvaligursky: what if we render with camera to a different size or even different aspect than device size? Device size should not be used at all.

GSterbrant: True. Okay I'll change this to be dependent on the input target.

GSterbrant: Okay, so looking back at this, every post effect is created without the input target and rely on the device size on creation. I think considering these parameters, the base size of the effect shouldn't be handled by the effect at all, but rather by the post effects queue.

So on init, we should also pass a context containing the size for the effect. I guess we can add that to the list of things we want to do when refactoring the post effects stuff 🙂 .

scripts/posteffects/posteffect-ssao.js
GSterbrant: This was a part of another piece of work I did and accidentally carried over in this PR.

src/framework/components/camera/post-effect-queue.js
GSterbrant: Ah, thanks!

src/framework/components/camera/post-effect-queue.js
```diff
@@ -404,6 +368,8 @@ class PostEffectQueue {
 
         for (let i = 0, len = effects.length; i < len; i++) {
             const fx = effects[i];
+            if (fx.effect.resize !== undefined)
```
mvaligursky, line 371:
```suggestion
            if (fx.effect.resize)
```

scripts/posteffects/posteffect-bloom.js
```diff
@@ -171,12 +171,47 @@ function BloomEffect(graphicsDevice) {
...
+    this.resize(null);
+}
+
+BloomEffect.prototype = Object.create(pc.PostEffect.prototype);
+BloomEffect.prototype.constructor = BloomEffect;
+
+BloomEffect.prototype.resize = function (target) {
+
+    var width, height;
+    if (target == null) {
```
mvaligursky, line 192:
```suggestion
    if (target === null) {
```

scripts/posteffects/posteffect-bloom.js
```diff
@@ -171,12 +171,47 @@ function BloomEffect(graphicsDevice) {
...
+        width = this.device.width;
+        height = this.device.height;
+    } else {
+        width = target.colorBuffer.width;
+        height = target.colorBuffer.height;
+    }
+
+    if (width == this.width && height == this.height)
```
mvaligursky, line 200:
```suggestion
    if (width === this.width && height === this.height)
```

scripts/posteffects/posteffect-bloom.js
```diff
@@ -171,12 +171,47 @@ function BloomEffect(graphicsDevice) {
...
+        return;
+
+    if (this.targets) {
+        for (let i = 0; i < this.targets.length; i++) {
```
mvaligursky, line 204: cannot use `let` in ES5, only `var`

scripts/posteffects/posteffect-ssao.js
```diff
@@ -400,16 +378,22 @@ function SSAOEffect(graphicsDevice, ssaoScript) {
...
+    if (target == null) {
+        width = Math.ceil(this.device.width / this.device.maxPixelRatio / this.downscale);
+        height = Math.ceil(this.device.height / this.device.maxPixelRatio / this.downscale);
+    } else {
+        width = Math.ceil(target.colorBuffer.width / this.device.maxPixelRatio / this.downscale);
+        height = Math.ceil(target.colorBuffer.height / this.device.maxPixelRatio / this.downscale);
+    }
 
     // If no change, skip resize
     if (width == this.width && height == this.height)
```
mvaligursky, line 399: could you please use `===` here as well

mvaligursky (approved):
approving with some minor comments (please fix before the merge)

lucaheft:
Also the new resize function should be added to the pc.PostEffect docs.
(https://developer.playcanvas.com/en/api/pc.PostEffect.html)

mvaligursky:
Perhaps the resize function should be called internally within the effect, and not from the engine (post-effect-queue). Inside the render call, the effect should check the sizes of its internal buffers vs the source / target buffer it has, and rescale those internal buffers as needed.

GSterbrant:
The effect itself doesn't know about the input target and has to defer to the graphics device for dimensions. The post effects queue knows about the input targets, and can therefore forward it to the effect when it's enabled/disabled. 

Wouldn't it be better if we can avoid dealing with resize related code if we're not resizing?

mvaligursky:
the render function of the effect knows input and output
`    render: function (inputTarget, outputTarget, rect)`

handling size compare is not very expensive, we handle it in few places like this. It's simpler for the user / creator of the effect.

https://github.com/playcanvas/engine/blob/24f325f078c2753ae15948450d3da4b0918932a6/src/framework/scene-grab.js#L156-L158

mvaligursky:
This way you also avoid to maintain / handle resizing while the effect / camera is disabled and similar. Each time you render, you make sure targets are the right size.

lucaheft:
I like that approach. This way the resizing can be handled on a per post-effect instance.
The post-effect-queue shouldnt handle it anyway, it's just a simple queue isn't it?

lucaheft:
Also what i noticed, in the post-effect-queue the width and height is calculated like this:
https://github.com/playcanvas/engine/blob/24f325f078c2753ae15948450d3da4b0918932a6/src/framework/components/camera/post-effect-queue.js#L364-L365
In the bloom post effect like this:
https://github.com/playcanvas/engine/blob/24f325f078c2753ae15948450d3da4b0918932a6/scripts/posteffects/posteffect-bloom.js#L193-L194
And in ssao like this:
https://github.com/playcanvas/engine/blob/24f325f078c2753ae15948450d3da4b0918932a6/scripts/posteffects/posteffect-ssao.js#L391-L392
Shouldnt those calculations all be the same? Except for the downscale in the ssao effect.

GSterbrant:
The issue of having to make this calculation at all is due to the fact that post effects are setup prior to knowing the input target. This will be fixed in the PR I am working on addressing these issues :).

The post effect queue calculates the size of the internal targets, and really the effects should just read those values. The device dimensions shouldn't be used at all.

GSterbrant:
https://github.com/playcanvas/engine/pull/4397

PR #4254, opened 2022-05-17
Extract backend from standard.js

slimbuck:
This PR contains the first round of changes toward restructuring standard material and its chunks. (For background see #4250).

#### Summary of changes:
- extract the material backend chunk-building code from `standard.js` into `lit-shader.js`
- modify chunks into strict frontend and backend
- reorder runtime chunks so all frontend code executes together (some frontend and backend chunks were previously interleaved)
- remove toksvig specular AA (see #4260)

#### Notes:
- these changes have been tested with test-server and engine examples
- a followup PR will address the material options driving `LitShader`, which is currently a monolith that includes all options for frontend and backend

#### Questions:
- is the change to `emissivePS` worth the breakage? cc @mvaligursky @Maksims - YES

#### Changes to frontend chunks

| chunk | end | changes |
| --- | --- | --- |
| `clearCoatNormalPS` | front | <ul><li>refrain from generating world CC reflection, now done on the backend instead</li><li>normalize final world space normal</li></ul> |
| `diffusePS` | front | <ul><li>Fix gamma handling relative to albedo detail</li></ul> |
| `diffuseDetailMapPS` | front | <ul><li>Gamma correct detail map before combining with base albedo</li></ul> |
| `emissivePS` | front | <ul><li>set `dEmissive` global instead of returning the value in order to bring it in line with the other frontend components</li></ul> |
| `lightmapSingleVert.js` | front | <ul><li>removed (unused)</li></ul> |
| `lightmapDirPS`, `lightmapSinglePS`| front | <ul><li>renamed the lightmap function to `getLightMap()` instead of `addLightMap()`</li><li>changed the implementation to write `dLightmap` and `dLightmapDir` global instead of updating `dDiffuseLight` and `dSpecularLight` directly</li><li>backend now handles combining lightmap in `lightmapAddPS` and `lightmapDirAddPS`</li></ul> |
| `normalMapFastPS` | front | <ul><li>removed</li></ul> |
| `normalMapPS` | front | <ul><li>added `MAPTEXTURE` #define like the other chunks</li><li>normalize final normal</li><li>when normal texture isn't defined, calculate normal from geometry normal instead</li></ul> |
| `normalDetailMapPS` | front | <ul><li>remove two (mostly) unnecessary calls to `normalise` - final normal is normalized instead</li></ul> |
| `clusteredLightPS` | back | <ul><li>remove dead code</li></ul> |
[...]

PR #4343, opened 2022-06-20
Chunk version checking

slimbuck:
Followup to #4254.

This PR adds a chunk system runtime compatibility check based on the engine version.

Applications use `StandardMaterial.chunks.APIVersion` to specify which engine version the overridden chunks implement. If any of the overridden chunks have a newer API defined in the engine (or the chunk was removed), a warning is printed to the debug console.

This helps app developers know which chunks have changed in their apps and in what version.


mvaligursky:
- Is there some public documentation to explain the user how to use the versioning?
- When we print some of these 'incorrect version' messages, could we include a link to PR with the change, or engine release notes for the version or similar, to guide them in how to fix it?

slimbuck:
As part of engine v1.55 release I think we should start documenting chunk changes. Once we've decided where and how we can update the log message. (Though I am hesitant at the idea of printing URLs, seems a bit strange...).

Perhaps we can add a `CHUNK_HISTORY.md` file to the engine repo.

willeastcott:
From a quick glance, it doesn't look like all this new functionality is restricted to a debug build. Is that right?

willeastcott:
My concern is that this is going to bloat the release build over time as we continue to bump versions. Do we really have to carry around what is essentially a debugging aid in production builds of apps?

mvaligursky:
> My concern is that this is going to bloat the release build over time as we continue to bump versions. Do we really have to carry around what is essentially a debugging aid in production builds of apps?

That's a good point. As long as the behaviour of the code with and without validation is the same (for example to always use user chunks, even when not up to date), we can drop the validation in release I think.

slimbuck:
Done. Also corrected chunk versions.

src/graphics/program-lib/programs/lit-shader.js
```diff
@@ -65,19 +66,23 @@ class LitShader {
...
-                                this.attributes[a] = builtinAttributes[a];
-                            }
+            const userChunks = options.chunks;
+
+            // #if _DEBUG
+            validateUserChunks(options.chunks);
+            // #endif
+
+            for (const chunkName in shaderChunks) {
+                if (userChunks[chunkName]) {
```
Maksims, line 76: Use `hasOwnProperty` for this check instead. Users can define an empty string as a chunk.

slimbuck: Done, thanks

mvaligursky (approved):
looks good to me.

PR #4350, opened 2022-06-21
Remove throws from GraphNode which were unnecessarily interrupting execution

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/4228

### DIFF
--- a/src/scene/graph-node.js
+++ b/src/scene/graph-node.js
@@ -1,5 +1,6 @@
 import { EventHandler } from '../core/event-handler.js';
 import { Tags } from '../core/tags.js';
+import { Debug } from '../core/debug.js';
 
 import { Mat3 } from '../math/mat3.js';
 import { Mat4 } from '../math/mat4.js';
@@ -1155,13 +1156,7 @@ class GraphNode extends EventHandler {
      * this.entity.addChild(e);
      */
     addChild(node) {
-        if (node._parent !== null)
-            throw new Error('GraphNode is already parented');
-
-        // #if _DEBUG
-        this._debugInsertChild(node);
-        // #endif
-
+        this._prepareInsertChild(node);
         this._children.push(node);
         this._onInsertChild(node);
     }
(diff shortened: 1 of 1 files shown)


yaustar (approved):
LGTM

src/scene/graph-node.js
```diff
@@ -1155,13 +1156,7 @@ class GraphNode extends EventHandler {
...
      */
     addChild(node) {
-        if (node._parent !== null)
-            throw new Error('GraphNode is already parented');
-
-        // #if _DEBUG
-        this._debugInsertChild(node);
-        // #endif
-
+        this._prepareInsertChild(node);
```
slimbuck, line 1159: This function now does `node._parent.removeChild(node);` where it didn't before.

Might that change behaviour?

(Same for `insertChild()`).

mvaligursky: That's what is needed to continue with a valid graph. Better solution than throw.

src/scene/graph-node.js
```diff
@@ -1206,30 +1196,28 @@ class GraphNode extends EventHandler {
...
+
+        Debug.assert(!node._parent, `GraphNode ${node?.name} is already parented`);
+        Debug.assert(node !== this, `GraphNode ${node?.name} cannot be a child of itself`);
+        Debug.assert(!this.isDescendantOf(node), `GraphNode ${node?.name} cannot add an ancestor as a child`);
 
-        if (this.isDescendantOf(node))
-            throw new Error('GraphNode cannot add an ancestor as a child');
+        // remove it from existing parent
+        if (node._parent)
+            node._parent.removeChild(node);
```
willeastcott, line 1219: If the new behavior is that a node is unparented if it already has a parent, perhaps this should be mentioned in the API ref docs?

mvaligursky: But this triggers a console error, so it's not expected / new behaviour. It's just a mitigation for the app to reasonably continue instead of crash / create invalid graph.

willeastcott: Any reason it _can't_ just be documented behavior and not a mitigation?

willeastcott (approved):

PR #4352, opened 2022-06-21
Adding shadow intensity control to clustered lighting.

GSterbrant:
Highjack the castShadows value to instead pass the shadow intensity. Shadow intensity 0 could just as well be castShadows: false.
Updated the spot light shadow example to allow for controlling the shadow intensity.


mvaligursky (approved):
Looks good.

Issue #3954, opened 2022-01-28
Calling `Entity#destroy` can cause invalid calls to `GraphNode#removeChild`


willeastcott:
Do:

```
const a = new pc.Entity('a',app);
const b = new pc.Entity('b',app);
a.addChild(b);
a.destroy();
```

If you set a breakpoint inside `GraphNode#removeChild`, you'll see that the engine tries to remove a child from a node that is not actually a child of that node.

PR #4354, opened 2022-06-22
[Fix] Improvement to Entity.destroy

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/3954

- when the child is removed from the array, clear the parent as well, before calling destroy, to avoid it from trying to remove it from the array again
- use pop instead of shift, which is faster, and we don't really care about the order of children destruction.

### DIFF
--- a/src/framework/entity.js
+++ b/src/framework/entity.js
@@ -552,19 +552,17 @@ class Entity extends GraphNode {
         if (this._parent)
             this._parent.removeChild(this);
 
+        // recursively destroy all children
         const children = this._children;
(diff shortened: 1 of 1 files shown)


GSterbrant (approved):

slimbuck:
Could there be logic that depends on still being parented when destroyed?

mvaligursky:
> Could there be logic that depends on still being parented when destroyed?

I would hope not, but this is hard to find out.

willeastcott (approved):

Issue #4199, opened 2022-04-15
Frustum culling and UI element issue 


Kulodo:
I have a scene with frustum culling set on the camera. If I create a UI element in the editor it is always displayed ( the yellow cog ). However, if I create a similar UI element in code ( the white cog) then the element will appear and disappear as I move the camera around with the 3 person controller. If I disable frustum culling then this issue doesn’t happen. 

Here is the forum post with links to the test project and quick work around.

[https://forum.playcanvas.com/t/frustum-culling-and-ui-element-issue/25252](https://forum.playcanvas.com/t/frustum-culling-and-ui-element-issue/25252)
<img width="300" alt="Screenshot 2022-04-15 at 20 43 45" src="https://user-images.githubusercontent.com/74154551/163629342-017c8c80-0ba2-4913-8e04-27256534b89f.png">

yaustar:
@Kulodo Do you have a link to the project that still shows the issue?

Kulodo:
I've commented out the quick fix from Leonidas now, so it is displaying the issue again.
[https://playcanvas.com/project/913583/overview/culluitest](https://playcanvas.com/project/913583/overview/culluitest)
Here is the code [https://playcanvas.com/editor/code/913583?tabs=78008457](https://playcanvas.com/editor/code/913583?tabs=78008457)

yaustar:
Changing the code from:
```
    entity.name = 'DeveloperMenuScreenEntity';
    entity.addComponent('screen');
    entity.screen.screenSpace = true;
    entity.screen.scaleMode = pc.SCALEMODE_BLEND;
    entity.screen.scaleBlend = 0.5;
    entity.screen.referenceResolution = new pc.Vec2(1280, 720);
```

to
```
    entity.name = 'DeveloperMenuScreenEntity';
    entity.addComponent('screen', {
        screenSpace: true
    });
    entity.screen.scaleMode = pc.SCALEMODE_BLEND;
    entity.screen.scaleBlend = 0.5;
    entity.screen.referenceResolution = new pc.Vec2(1280, 720);
```

Works around the issue without needing to go through the whole render layer meshes

PR #4355, opened 2022-06-22
Fix culling on screen-space screen.

jpauloruschel:
Fixes https://github.com/playcanvas/engine/issues/4199

This PR fixes the way a `Screen`'s `cull` and `screenSpace` properties are used on `ImageElement`. I confirmed that `TextElement` is working correctly.

The *before* code was not following the logic outlined in the `Screen` API documentation (https://developer.playcanvas.com/en/api/pc.ScreenComponent.html#cull)

```
cull: If true then elements inside this screen will be not be rendered when outside of the screen (only valid when screenSpace is true).
```

in that it was ignoring the `cull` property of the parent `Screen` and always culling. Additionally, the `visibleFn` (used to calculate if that element is visible for a given camera) was not being set correctly for screenSpace screens.

Before (https://playcanvas.com/project/913583/overview/culluitest):

Screenshot 2022-06-22 at 12 49 43

After (https://playcanvas.com/project/913583/overview/culluitest):

Screenshot 2022-06-22 at 12 49 12

### DIFF
--- a/src/framework/components/element/image-element.js
+++ b/src/framework/components/element/image-element.js
@@ -206,7 +206,7 @@ class ImageRenderable {
         const element = this._element;
 
         let visibleFn = null;
-        if (cull && element._isScreenCulled()) {
+        if (cull && element._isScreenSpace()) {
             visibleFn = function (camera) {
                 return element.isVisibleForCamera(camera);
             };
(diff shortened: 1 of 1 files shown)


slimbuck (approved):
nice

Issue #4358, opened 2022-06-22
Fix programlib


slimbuck:
The `programlib` object is currently included directly in `GraphicsDevice` and also exported as a public object (though not documented).

Ideally we should:
- refrain from exporting programlib publicly
- register programs in the application instead - this will aid with tree shaking

willeastcott:
Since https://github.com/playcanvas/engine/pull/4692 was merged, what's left to do here?

mvaligursky:
I'm still hoping to do a minor cleanup, but this can be closed as done.

PR #4359, opened 2022-06-22
Remove chunks from programlib

slimbuck:
Chunks should not be included in the list of supported programs.

Added #4358 for followup work.


mvaligursky (approved):
I wonder if that was some kind of old code that was left for compatibility?
Only one way to find out. Approving.

willeastcott (approved):

slimbuck:
Actually I've had a change of heart. Instead of removing the chunks from `programlib` I would rather remove the programs.

Closing this in favour of a different approach.

PR #4360, opened 2022-06-22
Fix depth linearization

### LINKED ISSUE #3411: SSAO post effect not working with Orthographic camera mode
### Description

It seems that SSAO post effect not work with orthographic camera mode.
https://playcanvas.com/project/816088/overview/screenshottest 4

image
image

### Steps to Reproduce

1.  Add Playcanvas SSAO on camera
2. Switch camera projection to Orthographic

GSterbrant:
Fixes #3411

Linearization of the depth buffer was done incorrectly, however it produced coherently variant results with a perspective projection, but not with an orthogonal one. That resulted in post effects linearizing depth producing incorrect results, such as the SSAO:

Screenshot 2022-06-22 at 16 23 51
 effect. 

With this PR, the results are more correct:

Screenshot 2022-06-22 at 16 53 48

### DIFF
--- a/src/graphics/program-lib/chunks/common/frag/screenDepth.js
+++ b/src/graphics/program-lib/chunks/common/frag/screenDepth.js
@@ -13,13 +13,13 @@ uniform mat4 matrix_view;
 
 #ifndef CAMERAPLANES
 #define CAMERAPLANES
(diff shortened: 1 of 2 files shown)


willeastcott:
Cool, great to see this fixed. As an aside, I'm wondering whether the chessboard @mvaligursky added would be a much better scene to show off post effects (particularly bokeh).

mvaligursky (approved):
nice one!

PR #4363, opened 2022-06-23
Update programlib

slimbuck:
This PR:
- registers programs in the application directly on the device instead of using the `programlib` object
- remove unused `device` object from StandardMaterialOptionsBuilder
- fix `ProgramLibrary` test with SHADER_ defines updated in #4328

### DIFF
--- a/src/framework/application.js
+++ b/src/framework/application.js
@@ -2,6 +2,11 @@ import { platform } from '../core/platform.js';
 
 import { WebglGraphicsDevice } from '../graphics/webgl/webgl-graphics-device.js';
 
+import { basic } from '../graphics/program-lib/programs/basic.js';
+import { particle } from '../graphics/program-lib/programs/particle.js';
+import { skybox } from '../graphics/program-lib/programs/skybox.js';
+import { standard } from '../graphics/program-lib/programs/standard.js';
+
 import { SoundManager } from '../sound/manager.js';
 
 import { Lightmapper } from '../scene/lightmapper/lightmapper.js';
@@ -154,7 +159,15 @@ class Application extends AppBase {
         }
         options.graphicsDeviceOptions.alpha = options.graphicsDeviceOptions.alpha || false;
 
-        return new WebglGraphicsDevice(canvas, options.graphicsDeviceOptions);
+        const device = new WebglGraphicsDevice(canvas, options.graphicsDeviceOptions);
+
+        // register shader programs
+        device.programLib.register('basic', basic);
+        device.programLib.register('particle', particle);
+        device.programLib.register('skybox', skybox);
+        device.programLib.register('standard', standard);
+
+        return device;
     }
 
     addComponentSystems(appOptions) {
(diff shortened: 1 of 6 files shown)


src/framework/application.js
```diff
@@ -154,7 +159,15 @@ class Application extends AppBase {
...
         options.graphicsDeviceOptions.alpha = options.graphicsDeviceOptions.alpha || false;
 
-        return new WebglGraphicsDevice(canvas, options.graphicsDeviceOptions);
+        const device = new WebglGraphicsDevice(canvas, options.graphicsDeviceOptions);
+
+        // register shader programs
+        device.programLib.register('basic', basic);
+        device.programLib.register('particle', particle);
+        device.programLib.register('skybox', skybox);
+        device.programLib.register('standard', standard);
```
willeastcott, line 168: So this presumably means that anyone using AppBase now also needs to explicitly register shader programs?

slimbuck: Correct. It also becomes tree-shakable.

mvaligursky:
I think a first step we should take is consider moving /program-lib folder outside of the graphics. It this it should standalone, one level up from graphics, on the same level as the /scene. Currently it is in graphics, but uses many constants and even few modules the scene. And consider how this works out after that. I'm not convinced the GraphicsDevice should own the builders library. Graphics device should be a user of generated shaders, but don't be part of the generation itself.

slimbuck:
> I think a first step we should take is consider moving /program-lib folder outside of the graphics. It this it should standalone, one level up from graphics, on the same level as the /scene. Currently it is in graphics, but uses many constants and even few modules the scene. And consider how this works out after that. I'm not convinced the GraphicsDevice should own the builders library. Graphics device should be a user of generated shaders, but don't be part of the generation itself.

Yep totally agree. Also Device owns ProgramLibrary and ProgramLibrary keeps device :(

What this PR is doing is -not- registering random shader chunks as programs (by not using programlib for this) without potentially breaking existing projects.

src/framework/application.js
mvaligursky: At the moment this PR would not allow BaseApp to be used instead of the Application (which it is currently) as those generators (basic, particle ..) are not exported and so cannot be from outside set on the device.

slimbuck: what would you suggest we do instead?

mvaligursky: Maybe those lines should be moved to AppBase till we're ready to deal with it?

slimbuck: Yep makes sense. Moved to AppBase.init. Thanks!

src/graphics/program-lib/program-lib.js
```diff
@@ -13,12 +9,7 @@ const programlib = {
     precisionCode: precisionCode,
     skinCode: skinCode,
     tonemapCode: tonemapCode,
-    versionCode: versionCode,
-
-    basic: basic,
-    particle: particle,
-    skybox: skybox,
-    standard: standard
+    versionCode: versionCode
```
mvaligursky, line 12: I think the remainder if this file should move to deprecated.js .. that's where I think it's the only place it is exposed for compatibility reasons.

willeastcott (approved):

mvaligursky (approved):

PR #4368, opened 2022-06-24
Support tuples for JSDoc

kungfooman:
Fixes https://github.com/playcanvas/engine/pull/4197#issuecomment-1164984831

Issue open since 2015: https://github.com/jsdoc/jsdoc/issues/1073
Issue open since 2019: https://github.com/jsdoc/jsdoc/issues/1703

Based on that they couldn't care enough for the last 7 years, I just fixed the `Catharsis#parse` function myself.

### DIFF
--- a/conf-api.json
+++ b/conf-api.json
@@ -1,7 +1,8 @@
 {
     "plugins": [
         "plugins/markdown",
-        "./node_modules/jsdoc-tsimport-plugin/index.js"
+        "./node_modules/jsdoc-tsimport-plugin/index.js",
+        "supportTuples"
     ],
     "recurseDepth": 10,
     "source": {
--- a/supportTuples.js
+++ b/supportTuples.js
@@ -1,0 +1,8 @@
+const catharsis = require("catharsis");
+const originalParse = catharsis.parse;
+catharsis.parse = function(str) {
+  if (str[0] == '[') {
+    return originalParse('Array<*>');
+  }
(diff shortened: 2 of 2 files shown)


willeastcott:
This is an ingenious workaround. But I'm not particularly comfortable hacking around a [missing feature in JSDoc](https://github.com/jsdoc/jsdoc/issues/1703). I have a goal to eliminate as many files and configuration from the root of the repo as possible - this keeps it simple and accessible to contributors - but this PR introduces a new root script just to fix the return type of an internal function. I just don't think it's worth it, to be honest - at least for now (I do appreciate that this can now be applied to other JSDoc blocks). On balance, for now, I'd rather kick that `@return` tag back to something vanilla JSDoc can process.

kungfooman:
I could add a folder called `src_build` which collects the small snippets (`types-*.js`) that support the build process? Then it can also have a little `README.md` giving an overview of whats going on.

I don't really feel like lagging behind in proper typing just because another project can't get their stuff together. What if the JSDoc project takes another 7 years or dies as a whole before that?

kungfooman:
I am trying to add named/unnamed tuple support into `catharsis` (which is used by `JSDoc`), so eventually it may be supported and this hacky PR can be forgotten about 😅

https://github.com/hegemonic/catharsis/pull/70

willeastcott:
Nice - that'd definitely be the ideal solution. Thanks @kungfooman!

PR #4369, opened 2022-06-24
Update comment in Grab-Pass example to match current behaviour

mvaligursky:
- comment updated

### DIFF
--- a/examples/src/examples/graphics/grab-pass.tsx
+++ b/examples/src/examples/graphics/grab-pass.tsx
@@ -26,8 +26,8 @@ class GrabPassExample {
         'shader.frag': /* glsl */`
             precision mediump float;
 
-            // use the special uSceneColorMap texture, which is a built-in texture, and which contains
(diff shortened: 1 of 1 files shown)


examples/src/examples/graphics/grab-pass.tsx
```diff
@@ -26,8 +26,8 @@ class GrabPassExample {
         'shader.frag': /* glsl */`
             precision mediump float;
 
-            // use the special uSceneColorMap texture, which is a built-in texture. Each time this texture is used
-            // for rendering, the engine will copy color framebuffer to it which represents already rendered scene
+            // use the special uSceneColorMap texture, which is a built-in texture, and which contains
+            // the copy of the color buffered at a point of capture, inside the Depth layer.
```
slimbuck, line 30:
```suggestion
            // use the special uSceneColorMap texture, which is a built-in texture containing
            // a copy of the color buffer at the point of capture, inside the Depth layer.
```

slimbuck (approved):
approving with suggestion

PR #4377, opened 2022-06-26
Use exponentiation operator instead of Math.pow

willeastcott:
JavaScript now support the [exponentiation operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Exponentiation) (`**`). This essentially does the same thing as [`Math.pow`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/pow). There is an ESLint rule called [`prefer-exponentiation-operator`](https://eslint.org/docs/latest/rules/prefer-exponentiation-operator) (which we currently have disabled) - and it makes the argument for the rule as follows:

    Introduced in ES2016, the infix exponentiation operator ** is an alternative for the standard Math.pow function.

    Infix notation is considered to be more readable and thus more preferable than the function notation.

So this PR is a suggestion to poll opinion as to whether we adopt officially `**`.

### DIFF
--- a/scripts/camera/follow-camera.js
+++ b/scripts/camera/follow-camera.js
@@ -58,7 +58,7 @@ FollowCamera.prototype.postUpdate = function (dt) {
         // Lerp the current camera position to where we want it to be
         // Note that the lerping is framerate independent
         // From: https://www.rorydriscoll.com/2016/03/07/frame-rate-independent-damping-using-lerp/
-        this.currentPos.lerp(this.currentPos, this.targetPos, 1 - Math.pow(1 - this.lerpAmount, dt));
+        this.currentPos.lerp(this.currentPos, this.targetPos, 1 - (1 - this.lerpAmount) ** dt);
 
         // Set the camera's position
         this.entity.setPosition(this.currentPos);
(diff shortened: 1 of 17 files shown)


yaustar:
I prefer `Math.pow`. `**` looks like something that is easy to mistype or mis-read

slimbuck (approved):
Not sure I prefer ** given the extra complication of operator precedence, but it's not a strong objection.

kungfooman:
Personally I like removing fluff from code, to make it as expressive in least amount of space. Never even realized that `**=`is possible, I like that one.

There are three cases of `resMult = 1 / 2 ** mipLevel;` which could become `resMult = 0.5 ** mipLevel;` to get rid of the division and its just a bit shorter (not sure if its easier to understand).

willeastcott:
@kungfooman `**` takes precedence over `/` so you can't make that substitution.

kungfooman:
@willeastcott Its a special case, you can also write `2 ** -mipLevel`.

mvaligursky:
> Not sure I prefer ** given the extra complication of operator precedence, but it's not a strong objection.

I feel the same. To me the code is harder to understand, if the brackets are missing. Perhaps we should have brackets around it (apart from a trivial code).

Exponentiation is also right-associative (unlike multiply and divide) but it's on common to have code like `2 ** 3 ** 4` so this is a small issue.

I also checked ES5 compilation, and its transpiled to `Math.pow` .. so that's great.

I'm approving, but I'm not 100% behind this.

mvaligursky (approved):

willeastcott:
This doesn't seem to be overwhelmingly superior to `Math.pow` so I think I'll close this for now.

Issue #3514, opened 2021-09-24
Be able to control the rendering order of UI screens on the same layer


yaustar:
At the moment, there doesn't seem to be a way of controlling the order of rendering of multiple UI screens. Typical usecases are things like popup windows that would overlay over the standard HUD.

Currently the workaround is to use a separate layer or have the notion of sub screens (that are image groups) on a single UI screen in the scene.

jiajasper:
@jpauloruschel @yaustar A very quick solution is to determine render order of UI on the same layer with their order on the hierarchy panel. So the one on the top of the hierarchy will be rendered on top. 

eg. hud 2d will always be overlaying world 3d screen and within hud 2d, time bar will be rendered over score

<img width="207" alt="image" src="https://user-images.githubusercontent.com/30747200/174271328-c3436842-b321-4dce-ba43-6f33d856399c.png">

yaustar:
It doesn't quite work like that for the UI.

By default, the UI elements for a UI screen are rendered in the 'downward' order of the elements in the hierarchy. So on HUD - 2D Screen where it's a UI screen, Time Bar is rendered first for the screen, then it's children and last would be Score.

This works okay for the most part as elements should be contained by the screen.

However, for screens, developers may have them in different places in the hierarchy for organisation or because they are created/loaded dynamically so ordering them is a fiddly at runtime.

What we could do is to use what the cameras do and add a priority property so they render in that order instead

yaustar:
Another idea after discussing with Jasper, we could have a ScreenGroup component where all the child screens are rendered in the order of the hierarchy depth in the graph (similar to child elements under a screen component)

yaustar:
At the moment, the order of rendering seems to be in the order that the screens were created (last one is on top) which means dynamically created screens (eg from Templates) would always be on top

jpauloruschel:
> Another idea after discussing with Jasper, we could have a ScreenGroup component where all the child screens are rendered in the order of the hierarchy depth in the graph (similar to child elements under a screen component)

I think this particular wouldn't be ideal because it would still mean users would not be able to organize their scenes as they want (as you mentioned in your previous comment).

> At the moment, the order of rendering seems to be in the order that the screens were created (last one is on top) which means dynamically created screens (eg from Templates) would always be on top

Interesting finding! I'll look into this.

I do like the idea of a `priority` attribute which is used to sort the screens before rendering.

yaustar:
[...]

PR #4380, opened 2022-06-27
Sync screens draw order on set Priority.

jpauloruschel:
Part of https://github.com/playcanvas/engine/issues/3514

This PR makes it so the raw order of UI Elements is refreshed properly when changing a `Screen`'s `priority` attribute.

### DIFF
--- a/src/framework/components/screen/component.js
+++ b/src/framework/components/screen/component.js
@@ -287,8 +287,8 @@ class ScreenComponent extends Component {
 
     /**
      * Priority determines the order in which Screen components in the same layer are rendered.
-     * Number must be an integer between 0 and 255.
-     * Priority is set into the top 8 bits of the drawOrder property in an element.
+     * Number must be an integer between 0 and 255. Priority is set into the top 8 bits of the
+     * drawOrder property in an element.
      *
      * @type {number}
      */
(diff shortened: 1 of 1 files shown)


yaustar:
We should expose `priority` as public API too :)

slimbuck (approved):

jpauloruschel:
> We should expose `priority` as public API too :)

Changed in JSDocs (so it appears in API Reference) and also changed the manual: https://github.com/playcanvas/developer.playcanvas.com/pull/411

src/framework/components/screen/component.js
```diff
@@ -286,19 +286,24 @@ class ScreenComponent extends Component {
...
     /**
-     * Priority determines the order in which screens components are rendered. Priority is set into
-     * the top 8 bits of the drawOrder property in an element.
+     * Priority determines the order in which Screen components in the same layer are rendered.
+     * Number must be an integer between 0 and 255.
+     * Priority is set into the top 8 bits of the drawOrder property in an element.
      *
      * @type {number}
-     * @private
+     * @public
```
willeastcott, line 294: You don't need this tag. It will be public by default, so you can delete it.

jpauloruschel: Good call!

willeastcott (approved):

PR #3172, opened 2021-05-18
Half angle fresnel

raytranuk:
This PR:
- is built on top of the envBRDF PR (https://github.com/playcanvas/engine/pull/3157) as it didn't make sense to revert the envBRDF changes
- switches punctual lights to use half-angle fresnel
- removes some code complexity dealing with differences to where area lights calculate fresnel

Below are some screen captures before and after this PR (Tip: its best to open each image in a separate tab and switch tabs to see the difference):

Before:
image
After:
image

Before:
image
After:
image

As requested, here is a list of shader chunks altered by this PR:
- combineClearCoat.frag 
- combineDiffuseSpecular.frag 
- combineDiffuseSpecularNoConserve.frag 
- combineDiffuseSpecularNoRefl.frag 
- combineDiffuseSpecularNoReflSeparateAmbient.frag 
- combineDiffuseSpecularOld.frag 
- fresnelSchlick.frag 
- lightSpecularAnisoGGX.frag 
- lightSpecularBlinn.frag 
- lightSpecularPhong.frag 
- ltc.frag

I confirm I have read the [contributing guidelines](https://github.com/playcanvas/engine/blob/master/.github/CONTRIBUTING.md) and signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #4386, opened 2022-06-29
Material specular color

### LINKED ISSUE #3650: Support glTF extension: KHR_materials_specular
https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_specular/README.md

This extension adds two parameters to the metallic-roughness material: specular and specularColor.

specular allows users to configure the strength of the specular reflection in the dielectric BRDF. A value of zero disables the specular reflection, resulting in a pure diffuse material. The metal BRDF is not affected by the parameter.

specularColor changes the F0 color of the specular reflection in the dielectric BRDF, allowing artists to use effects known from the specular-glossiness material (KHR_materials_pbrSpecularGlossiness) in the metallic-roughness material.

Screenshot 2021-11-01 at 15 38 32

GSterbrant:
### Description

Fixes #3650. 

image

Added support for KHR_materials_specular, which provides a specular tint (same as specularMap) as well as a specularity factor. The specularity factor isn't a substitute for metalness however, but simply a specular reflection intensity, which is applied to both fresnel and general specularity.

Decoupled the fresnel component from the specularity. This is because fresnel is dependent on each light, and in the case of reflections dependent on the surface normal. Therefore, we should calculate the fresnel factor once per each light source. Currently though, the fresnel factor only applies to reflections, but that's the same behaviour from before, and this should be addressed in a separate PR.

Metalness modulation (linear interpolation between f0 and albedo) is done in the backend as a separate shader chunk, and metalness is passed to the backend with dMetalness.

Since the specular color, f0, is dependent on the fresnel effect, and the fresnel effect is dependent on a per-light basis, the specular color can't just be applied on top of everything when combining the light together. Instead, the f0 color has to be an argument to each specularity calculation for it to be accurate.

We can also remove the dSpecularityNoFres from the area light calculation as the fresnel function is no longer combined with the f0 color.

This PR also prepares #3652 for adding index of reflection (ior) as a component to the fresnel code. 

Relates to #3172. In this issue, the reason why the view and normal vectors are used was not clear, but this PR clarifies it by exposing fresnel as a function separate from the specular color, thus allowing it to be used for each light source and environment reflections.


src/scene/materials/standard-material.js
```diff
@@ -109,6 +109,13 @@ let _params = new Set();
  * are specularTint are set, they'll be multiplied by vertex colors.
  * @property {string} specularVertexColorChannel Vertex color channels to use for specular. Can be
  * "r", "g", "b", "a", "rgb" or any swizzled combination.
+ * @property {boolean} useSpecularityFactor Use the specularity factor to weight specularity
+ * @property {number} specularityFactor The factor of specularity, exclusive to specular
+ * @property {Texture|null} specularityFactorMap The factor of specularity as a texture
+ * @property {boolean} f0Tint Multiply F0 reflection color
+ * @property {Vec3} f0 The F0 color
+ * @property {Texture|null} f0Map The map used for the F0 reflection color
+ * @property {string} f0MapChannel Color channels of the F0 color
```
willeastcott, line 118:
```suggestion
 * @property {boolean} f0Tint Multiply F0 reflection color.
 * @property {Vec3} f0 The F0 color.
 * @property {Texture|null} f0Map The map used for the F0 reflection color.
 * @property {string} f0MapChannel Color channels of the F0 color.
```

willeastcott: Maybe explain what F0 is somewhere? Also, consider adding 'Defaults to...' to each property description.

types-fixup.mjs
```diff
@@ -437,6 +437,13 @@ const standardMaterialProps = [
     ['specularTint', 'boolean'],
     ['specularVertexColor', 'boolean'],
     ['specularVertexColorChannel', 'string'],
+    ['useSpecularityFactor', 'boolean'],
```
willeastcott, line 440: Is this property needed if 'specularityFactor' is set to 0?

willeastcott:
Wow, this is awesome!! I've done a quick first pass review but I'll let @slimbuck and @mvaligursky submit approvals before me. 😄

src/graphics/program-lib/chunks/standard/frag/specular.js
```diff
@@ -8,18 +8,22 @@ uniform sampler2D texture_specularMap;
...
 
     #ifdef MAPCOLOR
-    dSpecularity *= material_specular;
+    specularity *= material_specular;
     #endif
 
     #ifdef MAPTEXTURE
-    dSpecularity *= texture2D(texture_specularMap, $UV, textureBias).$CH;
+    vec3 srgb = texture2D(texture_specularMap, $UV, textureBias).$CH;
+    specularity *=  vec3(pow(srgb.r, 2.2), pow(srgb.g, 2.2), pow(srgb.b, 2.2));
```
mvaligursky, line 20: we have gamma2_2.js which probably be used / updated to handle sRGB <-> linear

mvaligursky:
Great looking change!
Some chunks, and even their API has changed .. and so the new chunk-validation.js should be updated to reflect it.

types-fixup.mjs
GSterbrant: Yes, I used this before to select the shader, but I've changed this to only be reflected in the material options, not in the material API.

src/resources/parser/glb-parser.js
```diff
@@ -1296,6 +1266,33 @@ const createMaterial = function (gltfMaterial, textures, flipV) {
         material.diffuseVertexColor = false;
     }
 
+    // handle KHR_materials_specular absorption
+    if (gltfMaterial.hasOwnProperty('extensions') &&
+        gltfMaterial.extensions.hasOwnProperty('KHR_materials_specular')) {
```
mvaligursky, line 1271: I think we should split these extension handling code out to separate functions.

src/scene/materials/standard-material.js
```diff
@@ -109,6 +109,13 @@ let _params = new Set();
  * are specularTint are set, they'll be multiplied by vertex colors.
  * @property {string} specularVertexColorChannel Vertex color channels to use for specular. Can be
  * "r", "g", "b", "a", "rgb" or any swizzled combination.
+ * @property {boolean} useSpecularityFactor Use the specularity factor to weight specularity
+ * @property {number} specularityFactor The factor of specularity, exclusive to specular
+ * @property {Texture|null} specularityFactorMap The factor of specularity as a texture
+ * @property {boolean} f0Tint Multiply F0 reflection color.
```
mvaligursky: This is part of the public API, and maybe we need to consider more meaningful names, or at least explanations. Are these properties commonly exposed in other engines?

GSterbrant: But yes, I named it specularityFactor as to not confuse it with the specular flow, as this is actually applied when using metalness.

kungfooman:
Models that always looked good/normal look quite wrong now:

pr marble tower

GSterbrant:
@kungfooman Would it be possible to have a look at that glb for testing?

GSterbrant:
Actually, I was able to repro this and I think the problem was that fresnel wasn't calculated for lights, which is usually okay, but for directional lights this becomes a real problem.

Correct PBR would apply fresnel for local lights as well, however the old method we used for fresnel was the fresnel factor you use for the environment reflections, but applied it to all lights. If you don't mind testing now, it should work better :)

kungfooman:
> If you don't mind testing now, it should work better 🙂

Wow, thank you for the quick fix, it looks way better now! I tested and still see some problems, like the wall joints seem to light bleed now:

image

And the other side of the top tower looks still way too shiny:

image

Like the parts of the tower that theoretically should be darkest (grayish wall joints) start to light up?

GSterbrant:
Hello again @kungfooman, I have updated the fresnel function to take glossiness into account, which should fix this issue.

src/graphics/program-lib/chunks/lit/frag/fresnelSchlick.js
```diff
@@ -1,14 +1,14 @@
 export default /* glsl */`
 // Schlick's approximation
 vec3 getFresnel(float cosTheta, vec3 f0) {
-    float fresnel = pow(1.0 - abs(cosTheta), 4.0);
-    float glossSq = dGlossiness * dGlossiness;
-    return f0 + (1.0 - f0) * fresnel * glossSq;
+    float fresnel = pow(1.0 - max(cosTheta, 0.0f), 5.0);
+    float glossSq = (dGlossiness * dGlossiness);
+    return f0 + (max(vec3(glossSq), f0) - f0) * fresnel;
```
mvaligursky, line 6: I wonder if some comments here would be useful to understand what is happening here.

kungfooman:
@GSterbrant Thank you again, great work, I just tested and it looks imo better than it was now:
pr2
pr3

GSterbrant:
@kungfooman I made a bunch of other changes and tested with that model to make sure it looks consistent. Please give it another try if you have time.

I discovered some inconsistencies between clustered lights and normal lights for fresnel/specular colour, which caused over-bright specularity for clustered (which is now on by default).

kungfooman:
> I discovered some inconsistencies between clustered lights and normal lights for fresnel/specular colour

I tried to add a bunch of clustered lights from the `clustered-lighting.html` by turning it into pure JS, so I can run it via devtools:

```js
var app = pc.app;
var pointLightList = [];
var spotLightList = [];
var dirLight = null;
// enabled clustered lighting. This is a temporary API and will change in the future
// @ts-ignore engine-tsd
app.scene.clusteredLightingEnabled = true;
// adjust default clustered lighting parameters to handle many lights:
var lighting = app.scene.lighting;
// 1) subdivide space with lights into this many cells:
// @ts-ignore engine-tsd
lighting.cells = new pc.Vec3(12, 16, 12);
// 2) and allow this many lights per cell:
// @ts-ignore engine-tsd
lighting.maxLightsPerCell = 48;
// @ts-ignore engine-tsd
lighting.shadowsEnabled = false;
// create many omni lights that do not cast shadows
var count = 30;
var intensity = 1.6;
for (var i = 0; i < count; i++) {
  var color = new pc.Color(intensity * Math.random(), intensity * Math.random(), intensity * Math.random(), 1);
  var lightPoint = new pc.Entity();
  lightPoint.addComponent("light", {
    type: "omni",
    color: color,
    range: 12,
    castShadows: false,
  });
  // attach a render component with a small sphere to each light
  var _material = new pc.StandardMaterial();
  _material.emissive = color;
  _material.update();
  lightPoint.addComponent('render', {
    type: "sphere",
    material: _material,
    castShadows: true,
  });
  lightPoint.setLocalScale(5, 5, 5);
  // add it to the scene and also keep it in an array
  app.root.addChild(lightPoint);
  pointLightList.push(lightPoint);
}
// create many spot lights
count = 16;
for (var _i = 0; _i < count; _i++) {
  var _color = new pc.Color(intensity * Math.random(), intensity * Math.random(), intensity * Math.random(), 1);
  var lightSpot = new pc.Entity();
  lightSpot.addComponent("light", {
    type: "spot",
    color: _color,
    innerConeAngle: 5,
    outerConeAngle: 6 + Math.random() * 40,
    range: 25,
    castShadows: false,
  });
  // attach a render component with a small cone to each light
  material = new pc.StandardMaterial();
  material.emissive = _color;
  material.update();
  lightSpot.addComponent('render', {
    type: "cone",
    material
  });
  lightSpot.setLocalScale(5, 5, 5);
  lightSpot.setLocalPosition(100, 50, 70);
  lightSpot.lookAt(new pc.Vec3(100, 60, 70));
  app.root.addChild(lightSpot);
  spotLightList.push(lightSpot);
}
// Create a single directional light which casts shadows
dirLight = new pc.Entity();
dirLight.addComponent("light", {
  type: "directional",
  color: pc.Color.WHITE,
  intensity: 0.2,
  range: 300,
  shadowDistance: 600,
  castShadows: true,
  shadowBias: 0.2,
  normalOffsetBias: 0.05,
});
app.root.addChild(dirLight);
// Set an update function on the app's update event
var time = 0;
app.on("update", function (dt) {
  time += dt;
  // move lights along sin based waves around the cylinder
  pointLightList.forEach(function (light, i) {
    var angle = i / pointLightList.length * Math.PI * 2;
    var y = Math.sin(time * 0.5 + 7 * angle) * 30 + 70;
    light.setLocalPosition(30 * Math.sin(angle), y, 30 * Math.cos(angle));
  });
  // rotate spot lights around
  spotLightList.forEach(function (spotlight, i) {
    var angle = i / spotLightList.length * Math.PI * 2;
    spotlight.setLocalPosition(40 * Math.sin(time + angle), 5, 40 * Math.cos(time + angle));
    spotlight.lookAt(pc.Vec3.ZERO);
    spotlight.rotateLocal(90, 0, 0);
  });
  // rotate directional light
  if (dirLight) {
    dirLight.setLocalEulerAngles(25, -30 * time, 0);
  }
});
```

The problem is currently that it generates two errors for the extra directional light, which used to work:

image

I flipped `this._clusteredLightingEnabled = true;` in `class Scene` and can't see any differences, but I also didn't experiment with extra lights yet. Do you have a small snippet that adds/positions/configures the light that makes the difference for you?

What I like to do is opening two tabs and then copy over the exact camera position/rotation:

```js
a = viewer.orbitCamera.azimElevDistance.target + "";
b = viewer.orbitCamera.focalPoint.target + "";
console.log(`
viewer.orbitCamera.azimElevDistance.target.set(...${a});
viewer.orbitCamera.focalPoint.target.set(...${b});
`); // copy/paste output in other tab
```

Then just switch via CTRL+PageUp/PageDown to see the difference.

I will experiment more with this 😅

GSterbrant:
Ah, I see what this is. Local lights are in a scope in GLSL, directional lights are just copied in to the code. It should be fixed now :)

src/resources/parser/glb-parser.js
```diff
@@ -1297,6 +1269,28 @@ const createMaterial = function (gltfMaterial, textures, flipV) {
...
+
+        const specularData = gltfMaterial.extensions.KHR_materials_specular;
+        if (specularData.hasOwnProperty('specularColorTexture')) {
+            material.f0Map = textures[specularData.specularColorTexture.index];
+            material.f0MapChannel = 'rgb';
```
mvaligursky, line 1279: do we need the channel? Is there some other way this could be configured, or is rgb the only option here.

src/resources/parser/glb-parser.js
```diff
@@ -1297,6 +1269,28 @@ const createMaterial = function (gltfMaterial, textures, flipV) {
...
+        }
+        if (specularData.hasOwnProperty('specularColorFactor')) {
+            color = specularData.specularColorFactor;
+            material.f0.set(Math.pow(color[0], 1 / 2.2), Math.pow(color[1], 1 / 2.2), Math.pow(color[2], 1 / 2.2));
```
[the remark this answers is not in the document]
GSterbrant: This is the same as the specular factor from the other glTF reading of specular information, so I just assumed the same transformation was needed.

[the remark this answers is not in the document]
mvaligursky: the extension spec says this .. so I think only texture is in sRGB, but the color factor is linear
Screenshot 2022-07-06 at 12 10 23

perhaps double check the spec for the other specular information, perhaps there's a bug there.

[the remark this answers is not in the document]
GSterbrant: Strange, it's the same for KHR_materials_pbrSpecularGlossiness, but there we also gamma convert the specular factor. We also do it for diffuse with that extension and pbrMetallicRoughness.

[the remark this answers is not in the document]
mvaligursky: its possible we bring it back to linear space in the shader .. but I suspect a bug is more likely.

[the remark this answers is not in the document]
GSterbrant: Yeah I checked that, but we don't... We only gamma correct texture samples.

src/graphics/program-lib/programs/standard.js
```diff
@@ -307,13 +307,25 @@ const standard = {
 
             // specularity & glossiness
             if ((litShader.lighting && options.useSpecular) || litShader.reflections) {
-                const specularPropName = options.useMetalness ? "metalness" : "specular";
                 decl.append("vec3 dSpecularity;");
                 decl.append("float dGlossiness;");
-                code.append(this._addMap(specularPropName, specularPropName + "PS", options, litShader.chunks));
+                if (options.useMetalness) {
```
slimbuck, line 312: Think you can move this block into the the one below.

src/scene/materials/standard-material-parameters.js
```diff
@@ -45,6 +45,8 @@ const standardMaterialParameterTypes = {
     specularMapTiling: 'vec2',
     specularMapOffset: 'vec2',
     specularMapRotation: 'number',
+    specularityFactor: 'number',
+    specularityFactorMap: 'texture',
```
slimbuck, line 49: The other map members are also expected by addMap in `standard.js`:
```suggestion
    specularityFactorMap: 'texture',
    specularityFactorMapChannel: 'string',
    specularityFactorMapUv: 'number',
    specularityFactorMapTiling: 'vec2',
    specularityFactorMapOffset: 'vec2',
    specularityFactorMapRotation: 'number',
```

src/scene/materials/standard-material.js
```diff
@@ -988,9 +997,10 @@ function _defineFlag(name, defaultValue) {
 function _defineMaterialProps() {
     _defineColor('ambient', new Color(0.7, 0.7, 0.7));
     _defineColor('diffuse', new Color(1, 1, 1));
-    _defineColor('specular', new Color(0, 0, 0));
+    _defineColor('specular', new Color(1, 1, 1));
```
slimbuck, line 1000: Wondering about this change: is the meaning/usage of specular now inverse of before? will this break existing projects?

src/scene/materials/standard-material.js
```diff
@@ -1076,6 +1086,7 @@ function _defineMaterialProps() {
     _defineTex2D('diffuse', 0, 3, '', true);
     _defineTex2D('specular', 0, 3, '', true);
     _defineTex2D('emissive', 0, 3, '', true);
+    _defineTex2D('specularityFactor', 0, 1, '', false);
```
slimbuck, line 1089: Looks like `specularityFactorPS` supports vertex color:
```suggestion
    _defineTex2D('specularityFactor', 0, 1, '', true);
```
Is there no default, expected channel to find specularityFactor? (like alpha maybe?)

types-fixup.mjs
```diff
@@ -437,6 +437,12 @@ const standardMaterialProps = [
     ['specularTint', 'boolean'],
     ['specularVertexColor', 'boolean'],
     ['specularVertexColorChannel', 'string'],
+    ['specularityFactor', 'number'],
+    ['specularityFactorMap', 'Texture|null'],
+    ['specularityFactorMapChannel', 'string'],
```
slimbuck, line 442: You'll have to add the additional map fields here too (offset, tiling etc)

src/scene/materials/standard-material-parameters.js
```diff
@@ -59,7 +61,10 @@ const standardMaterialParameterTypes = {
     metalnessMapUv: 'number',
     metalnessMapTiling: 'vec2',
     metalnessMapOffset: 'vec2',
-    netalnessMapRotation: 'number',
+    metalnessMapRotation: 'number',
+    f0: "rgb",
+    f0Tint: 'boolean',
+    f0Map: 'texture',
```
slimbuck, line 67: Wondering where these are used?

src/scene/materials/standard-material.js
GSterbrant: Okay, I think this is a good point. I will revert it to 0, but set specular tint to 1 when metalness is used. The problem here was that when metalness was enabled, specularity would be set to 0 and all specular effects would disappear because metalness enables specular, so if specular tint was the default it would just swallow all specular effects.

src/scene/materials/standard-material.js
```diff
@@ -111,6 +111,9 @@ let _params = new Set();
  * are specularTint are set, they'll be multiplied by vertex colors.
  * @property {string} specularVertexColorChannel Vertex color channels to use for specular. Can be
  * "r", "g", "b", "a", "rgb" or any swizzled combination.
+ * @property {number} specularityFactor The factor of specular intensity, used to weight the fresnel and specularity. Default is 1.0.
+ * @property {Texture|null} specularityFactorMap The factor of specularity as a texture (default is null).
+ * @property {string} specularityFactorMapChannel The channel used by the specularity factor texture to sample from (default is 'a').
```
slimbuck, line 116: pls add docs here for the new map members

src/graphics/program-lib/programs/standard.js
GSterbrant: Yes!

types-fixup.mjs
```diff
@@ -437,6 +437,16 @@ const standardMaterialProps = [
...
+    ['specularityFactorMapOffset', 'Vec2'],
+    ['specularityFactorMapRotation', 'number'],
+    ['specularityFactorMapTiling', 'Vec2'],
+    ['specularityFactorMapUv', 'number'],
+    ['f0', 'Color'],
+    ['f0Map', 'Texture|null'],
+    ['f0MapChannel', 'string'],
```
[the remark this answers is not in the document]
GSterbrant: Done!

src/resources/parser/glb-parser.js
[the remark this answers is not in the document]
mvaligursky: it turns out we expect the value to be in sRGB, and when convert color based material properties to uniforms, we bring those back to linear. So bit more math than needed, but the result is as expected.

src/graphics/program-lib/programs/lit-shader.js
```diff
@@ -892,12 +900,7 @@ class LitShader {
             }
 
             if (options.fresnelModel > 0) {
-                if (options.conserveEnergy && !hasAreaLights) {
-                    // NB if there are area lights, energy conservation is done differently
-                    code += chunks.combineDiffuseSpecularPS; // this one is correct, others are old stuff
-                } else {
-                    code += chunks.combineDiffuseSpecularNoConservePS; // if you don't use environment cubemaps, you may consider this
-                }
+                code += chunks.combineDiffuseSpecularPS; // if you don't use environment cubemaps, you may consider this
```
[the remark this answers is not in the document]
mvaligursky: ah I see .. its an existing comment .. is it relevant?

mvaligursky (approved):
Looks great!

src/graphics/program-lib/programs/lit-shader.js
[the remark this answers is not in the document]
GSterbrant: I think this comment just carried over, I can put it on NoRefl as that makes more sense.

PR #4394, opened 2022-07-01
Small internal refactor of shader variants handling by renderer

mvaligursky:


slimbuck:
Why not call the function `getShaderVariant` ?

mvaligursky:
> Why not call the function `getShaderVariant` ?

Why not indeed. I'll make it change.

src/scene/renderer/forward-renderer.js
```diff
@@ -1564,6 +1565,10 @@ class ForwardRenderer {
         _tempMaterialSet.clear();
     }
 
+    /**
+     * @param {LayerComposition} comp - The layer composition to update.
+     * @param {boolean} lightsChanged - The if lights of the composition has changed.
```
willeastcott, line 1570:
```suggestion
* @param {boolean} lightsChanged - True if lights of the composition has changed.
```

src/scene/materials/standard-material.js
```diff
@@ -719,19 +729,21 @@ class StandardMaterial extends Material {
...
         }
 
         const library = device.getProgramLibrary();
-        this.shader = library.getProgram('standard', options);
+        const shader = library.getProgram('standard', options);
 
         if (!objDefs) {
             this.clearVariants();
-            this.variants[0] = this.shader;
+            this.variants[0] = shader;
```
slimbuck, line 742: any idea what the purpose of this `!objDefs` block is?

mvaligursky: as far as I can tell, it has no purpose / logic. It was getting triggered by lightmapper, but even that line served no purpose. So I've removed it all.

src/scene/materials/material.js
```diff
@@ -393,11 +404,16 @@ class Material {
...
+        // clear variants on the material
         this.variants = {};
+
+        // but also clear them from all materials that reference them
         for (let i = 0; i < this.meshInstances.length; i++) {
             const meshInstance = this.meshInstances[i];
-            for (let j = 0; j < meshInstance._shader.length; j++) {
-                meshInstance._shader[j] = null;
+            const shaders = meshInstance._shader;
+            for (let j = 0; j < shaders.length; j++) {
```
kungfooman, line 415: This looks like filling the `shaders` array with `null`'s, why not just `shaders.length = 0`, is it required to keep the length of the array for maybe performance reasons?

src/scene/renderer/forward-renderer.js
```diff
@@ -1325,26 +1314,33 @@ class ForwardRenderer {
...
                             drawCall._shader[pass] = material.variants[variantKey];
                             if (!drawCall._shader[pass]) {
-                                this.updateShader(drawCall, objDefs, null, pass, sortedLights);
+                                drawCall.updatePassShader(scene, pass, null, sortedLights);
                                 material.variants[variantKey] = drawCall._shader[pass];
                             }
                         } else {
-                            this.updateShader(drawCall, objDefs, drawCall._staticLightList, pass, sortedLights);
+                            drawCall.updatePassShader(scene, pass, drawCall._staticLightList, sortedLights);
                         }
```
mvaligursky, line 1339: Non-clarity of this blocks also still bothers be, and I'll refactor it later, probably soon as I might need it. I don't like that the caller of updateShader needs to calculate the variant key, and add the shader to material.variants. This should be completely internal ideally, the key computed from shader options generated, and not on some very few variables. I suspect that at the moment, this does not allow us to have all the variations we need, as the keys are the same for them.

src/scene/materials/basic-material.js
```diff
@@ -94,7 +104,8 @@ class BasicMaterial extends Material {
             pass: pass
         };
         const library = device.getProgramLibrary();
-        this.shader = library.getProgram('basic', options);
+        const shader = library.getProgram('basic', options);
+        return shader;
```
slimbuck, line 108:
```suggestion
        return library.getProgram('basic', options);
```

src/scene/mesh-instance.js
```diff
@@ -654,6 +655,21 @@ class MeshInstance {
...
+     * @param {Scene} scene - The scene.
+     * @param {number} pass - The render pass.
+     * @param {any} staticLightList - List of static lights.
+     * @param {any} sortedLights - Array of array of lights.
+     * @ignore
+     */
+    updatePassShader(scene, pass, staticLightList, sortedLights) {
+        const material = this.material;
+        const shader = material.getShaderVariant(this.mesh.device, scene, this._shaderDefs, staticLightList, pass, sortedLights);
+        this._shader[pass] = shader;
```
slimbuck, line 670:
```suggestion
        this._shader[pass] = this.material.getShaderVariant(this.mesh.device, scene, this._shaderDefs, staticLightList, pass, sortedLights);
```

slimbuck (commented):
TBH I can't actually see what material.shader member is used for. Can it just be removed?

mvaligursky:
> TBH I can't actually see what material.shader member is used for. Can it just be removed?

It's used to assign a user defined shader to the instance of Material class.
It's not used at all for classes derived from Material, so I added some asserts to validate it.

src/scene/particle-system/particle-emitter.js
```diff
@@ -879,8 +879,9 @@ class ParticleEmitter {
                 customFace: this.emitter.orientation !== PARTICLEORIENTATION_SCREEN
             });
             this.shader = shader;
+            return shader;
```
slimbuck, line 882: previous line can (should) be deleted?

src/scene/particle-system/particle-emitter.js
```diff
@@ -879,8 +879,9 @@ class ParticleEmitter {
...
         };
-        this.material.updateShader();
+        this.material.getShaderVariant();
```
slimbuck, line 884: presumably this call is meant to cache the shader up-front, but presumably it doesn't help because `getShaderVariant` will be called again by `ForwardRenderer` anyway. should the call be removed?

slimbuck (commented):
This is like excavation! Sorry few more questions.

src/scene/particle-system/particle-emitter.js
slimbuck: actually, programLib will cache the shader, so this call does help.

src/scene/sky.js
```diff
@@ -66,9 +66,11 @@ class Sky {
                     toneMapping: (pass === SHADER_FORWARDHDR ? TONEMAP_LINEAR : scene.toneMapping)
                 });
             }
+
+            return this.shader;
```
slimbuck, line 70: shouldn't be using `this.shader` anymore?

src/scene/particle-system/particle-emitter.js
mvaligursky: yep, I moved it outside of this function, as that's the consistent way to do it.

src/scene/sky.js
mvaligursky: I change this one as well .. but for Material instance, the shader is / should be set. But for consistency, it should be done from outside the call, not inside.

But I'll be adding assignment to internal variations shader there I think soon, as that will be needed for my next steps.

slimbuck (approved):
more of this code needs refactoring, but this is a step in the right direction. approving.

PR #4397, opened 2022-07-04
Post effect resize fixes

GSterbrant:
Resizing happens prior to rendering, only with respect to the input target, and resize is a private function, called internally by the effect.

### DIFF
--- a/scripts/posteffects/posteffect-bloom.js
+++ b/scripts/posteffects/posteffect-bloom.js
@@ -186,8 +186,8 @@ BloomEffect.prototype.constructor = BloomEffect;
 
 BloomEffect.prototype._resize = function (target) {
 
-    const width = target.colorBuffer.width;
-    const height = target.colorBuffer.height;
+    var width = target.colorBuffer.width;
+    var height = target.colorBuffer.height;
 
     if (width === this.width && height === this.height)
         return;
--- a/scripts/posteffects/posteffect-bloom.js
+++ b/scripts/posteffects/posteffect-bloom.js
@@ -186,14 +186,8 @@ BloomEffect.prototype.constructor = BloomEffect;
 
 BloomEffect.prototype._resize = function (target) {
 
-    var width, height;
-    if (target === undefined) {
-        width = this.device.width;
-        height = this.device.height;
-    } else {
-        width = target.colorBuffer.width;
-        height = target.colorBuffer.height;
-    }
+    const width = target.colorBuffer.width;
+    const height = target.colorBuffer.height;
 
     if (width === this.width && height === this.height)
         return;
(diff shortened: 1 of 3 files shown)


mvaligursky:
Perhaps also remove the resize getting called by the queue?

scripts/posteffects/posteffect-bloom.js
```diff
@@ -179,17 +179,15 @@ function BloomEffect(graphicsDevice) {
...
 
 BloomEffect.prototype = Object.create(pc.PostEffect.prototype);
 BloomEffect.prototype.constructor = BloomEffect;
 
-BloomEffect.prototype.resize = function (target) {
+BloomEffect.prototype._resize = function (target) {
 
     var width, height;
-    if (target === null) {
+    if (target === undefined) {
```
mvaligursky, line 190: I'm not sure the target can be undefined here, we're inside rendering.

scripts/posteffects/posteffect-bloom.js
```diff
@@ -179,23 +179,15 @@ function BloomEffect(graphicsDevice) {
...
 
-    var width, height;
-    if (target === null) {
-        width = this.device.width;
-        height = this.device.height;
-    } else {
-        width = target.colorBuffer.width;
-        height = target.colorBuffer.height;
-    }
+    const width = target.colorBuffer.width;
```
mvaligursky, line 189: no const .. just var .. this is ES5 still

mvaligursky (approved):

scripts/posteffects/posteffect-ssao.js
```diff
@@ -377,23 +377,15 @@ function SSAOEffect(graphicsDevice, ssaoScript) {
...
 
-    var width, height;
-    if (target === null) {
-        width = Math.ceil(this.device.width / this.device.maxPixelRatio / this.downscale);
-        height = Math.ceil(this.device.height / this.device.maxPixelRatio / this.downscale);
-    } else {
-        width = Math.ceil(target.colorBuffer.width / this.device.maxPixelRatio / this.downscale);
-        height = Math.ceil(target.colorBuffer.height / this.device.maxPixelRatio / this.downscale);
-    }
+    var width = Math.ceil(target.colorBuffer.width / this.device.maxPixelRatio / this.downscale);
```
slimbuck, line 387: not sure maxPixelRatio should factor into this?

mvaligursky: Maks added this this in his PR and I think that's correct? If pixel ration says render to half the size screen res .. we should use half the size for internal buffers, unless `colorBuffer.width` already has this scale in?

mvaligursky: Looking at the code, it seems renderTargetScale is never modified nor publicly exposed / documented variable always holding 1. I suspect we should just remove it.

GSterbrant: I am guessing this is legacy code from trying to perform dynamic resolution downscaling? Just a speculation though.

slimbuck: Using maxPixelRatio here means the SSAO downscale buffer depends on the browser's CSS scaling rules, which I think is wrong.

SSAO downscaled buffer size should be a simple multiple of render target size.

PR #4398, opened 2022-07-04
Deallocate internal resources for post effects

GSterbrant:
### Description

Moves the deallocation of the internal render targets for bloom and SSAO to its own function such that we can deallocate said buffers when we disable the effects.

### DIFF
--- a/scripts/posteffects/posteffect-bloom.js
+++ b/scripts/posteffects/posteffect-bloom.js
@@ -192,6 +192,7 @@ BloomEffect.prototype._destroy = function () {
             this.targets[i].destroy();
         }
     }
+    this.targets.length = 0;
 };
 
 BloomEffect.prototype._resize = function (target) {
@@ -206,7 +207,6 @@ BloomEffect.prototype._resize = function (target) {
     this._destroy();
 
     // Render targets
-    this.targets = [];
     var i;
     for (i = 0; i < 2; i++) {
         var colorBuffer = new pc.Texture(this.device, {
--- a/scripts/posteffects/posteffect-bloom.js
+++ b/scripts/posteffects/posteffect-bloom.js
@@ -210,6 +210,7 @@ BloomEffect.prototype._resize = function (target) {
     var i;
     for (i = 0; i < 2; i++) {
         var colorBuffer = new pc.Texture(this.device, {
+            name: "Bloom Texture" + i,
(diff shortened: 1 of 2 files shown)


scripts/posteffects/posteffect-ssao.js
```diff
@@ -382,8 +382,21 @@ function SSAOEffect(graphicsDevice, ssaoScript) {
 SSAOEffect.prototype = Object.create(pc.PostEffect.prototype);
 SSAOEffect.prototype.constructor = SSAOEffect;
 
+SSAOEffect.prototype._destroy = function () {
+    if (this.target) {
+        this.target.destroyTextureBuffers();
+        this.target.destroy();
+        this.target = null;
+
+        this.blurTarget.destroyTextureBuffers();
```
mvaligursky, line 391: maybe do a separate `if (this.blurTarget)` here?

scripts/posteffects/posteffect-bloom.js
```diff
@@ -184,27 +184,33 @@ function BloomEffect(graphicsDevice) {
 BloomEffect.prototype = Object.create(pc.PostEffect.prototype);
 BloomEffect.prototype.constructor = BloomEffect;
 
+BloomEffect.prototype._destroy = function () {
+    if (this.targets) {
+        var i;
+        for (i = 0; i < this.targets.length; i++) {
+            this.targets[i].destroyTextureBuffers();
```
slimbuck, line 191: was it intentional to no longer call `destroyFrameBuffers` ?

slimbuck: Great, now I'm wondering why destroy doesn't also do `destroyTextureBuffers()`. @mvaligursky any idea?

mvaligursky: Yes. Framebuffers we allocate internally, we deallocate them.
Color buffers are allocated by the users, they need to deallocate. Part of the reason here is that a single color buffer can be used for multiple render targets, and we dont want to ref count them to handle release properly.

GSterbrant: I had the same thought :) Probably because a render target can own the textures, but can also alias them. In the case of an aliasing situation, you would destroy the textures even if they're still in use.

slimbuck: makes sense - thanks!

slimbuck (approved):

PR #4400, opened 2022-07-04
Post Effect resize fixes

GSterbrant:
### Description

A few more adjustments to the post effect resizing code, cleaning up some old deprecated stuff to do with render target scaling as well as removing the maxPixelRatio usage from the SSAO effect.


lucaheft:
`this.width` and `this.height` in the bloom post-effect is never set, which results in the targets being resized every frame

slimbuck (approved):
Approving! (but pls also wait for @mvaligursky approval before merge).

mvaligursky (approved):

Issue #4261, opened 2022-05-20
Implement support for Uniform Buffers


mvaligursky:
Uniform Buffer Objects let you specify a bunch of uniforms from a buffer. The advantages are:
- You can manipulate all the uniforms in the buffer outside of WebGL.
- You can have different sets of uniform buffer objects.
- assign whole buffer with many uniforms with a single webgl call

Also, the WebGPU allows uniforms to only be set using uniform buffers.

Basic design:
UniformBufferFormat - describes the content of the buffer (similar to VertexFormat) - list of uniforms and their types and so on.
UniformBuffer - an actual buffer, storing uniforms in a format described by UniformBufferFormat

Web GPU specifically has API that connects multiple uniform buffers as well as other buffer objects (textures) into a binding, an d all of those can be set on a device (assigned to a bind slot) by a single call. The engine will support this as a way of setting up textures and uniform buffers as well. Design:
BindGroupFormat - describes the format as an array of uniform buffers / textures.
BindGroup - an actual instance of the BindGroupFormat, attaching actual uniform buffers and textures into the bind group. This can then be set on device using a single call.

willeastcott:
I notice we have [`UniformBuffer`](https://github.com/playcanvas/engine/blob/b232b1380564018d477da156066f7abe39ca3630/src/platform/graphics/uniform-buffer.js#L99) class in the engine now. Can this be closed yet?

MatthewDZane:
Webgl2 Uniform Buffer Objects support would be great!!

mvaligursky:
Closing this, as WebGL support is here: https://github.com/playcanvas/engine/pull/8987
and leftover work is tracked by https://github.com/playcanvas/engine/issues/6157

PR #4402, opened 2022-07-05
ShaderProcessor partially updating shader source to support uniform buffers

mvaligursky:
a progress towards https://github.com/playcanvas/engine/issues/4261

- a GLSL shader processing in order to allow WebGL shader to be transformed info a format we can transpile to WebGPU. Initially in this PR, attributes, varyings and uniforms are extracted, and transforms  into required form - which means a fixed slot allocation, as well as uniform buffers.
- ProgramLibrary has been updated to allow this processing to be executed globally on any shader, which includes shaders set by the user directly on the Material, in addition to Standard, Basic, Sky, and Particle.

Additional changes:
- small clean up to material's handling of transparency functions.

Note: The shader processing does not currently execute on WebGl, but some shader functionality refactoring is used.

### DIFF
--- a/src/graphics/bind-group-format.js
+++ b/src/graphics/bind-group-format.js
@@ -50,7 +50,7 @@ class BindGroupFormat {
         /** @type {BindTextureFormat[]} */
         this.textureFormats = textureFormats;
 
-            // maps a texture format name to a slot index
+        // maps a texture format name to a slot index
         /** @type {Map<string, number>} */
         this.textureFormatsMap = new Map();
         textureFormats.forEach((tf, i) => this.textureFormatsMap.set(tf.name, i));
--- a/src/graphics/bind-group-format.js
+++ b/src/graphics/bind-group-format.js
@@ -50,7 +50,7 @@ class BindGroupFormat {
         /** @type {BindTextureFormat[]} */
         this.textureFormats = textureFormats;
 
-        // maps a texture format name to an slot index
+            // maps a texture format name to a slot index
         /** @type {Map<string, number>} */
         this.textureFormatsMap = new Map();
         textureFormats.forEach((tf, i) => this.textureFormatsMap.set(tf.name, i));
--- a/src/graphics/bind-group-format.js
+++ b/src/graphics/bind-group-format.js
@@ -50,7 +50,7 @@ class BindGroupFormat {
         /** @type {BindTextureFormat[]} */
         this.textureFormats = textureFormats;
 
-        // maps a texture format name to an index
+        // maps a texture format name to an slot index
         /** @type {Map<string, number>} */
         this.textureFormatsMap = new Map();
         textureFormats.forEach((tf, i) => this.textureFormatsMap.set(tf.name, i));
@@ -63,6 +63,16 @@ class BindGroupFormat {
      */
     destroy() {
         this.impl.destroy();
+    }
+
+    /**
+     * Returns slot index for a texture slot name.
+     *
+     * @param {string} name - The name of the texture slot.
+     * @returns {number} - The slot index.
+     */
+    getTextureSlot(name) {
+        return this.textureFormatsMap.get(name);
     }
 
     loseContext() {
--- a/src/graphics/shader-processor-options.js
+++ b/src/graphics/shader-processor-options.js
@@ -13,7 +13,7 @@ class ShaderProcessorOptions {
     /**
      * Constructs shader processing options, used to process the shader for uniform buffer support.
      *
-     * @param {UniformBufferFormat} viewUniformFormat - Format if the uniform buffer.
+     * @param {UniformBufferFormat} viewUniformFormat - Format of the uniform buffer.
      * @param {BindGroupFormat} viewBindGroupFormat - Format of the bind group.
      */
     constructor(viewUniformFormat, viewBindGroupFormat) {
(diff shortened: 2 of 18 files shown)


src/graphics/bind-group-format.js
```diff
@@ -50,7 +50,7 @@ class BindGroupFormat {
         /** @type {BindTextureFormat[]} */
         this.textureFormats = textureFormats;
 
-        // maps a texture format name to an index
+        // maps a texture format name to an slot index
```
willeastcott, line 53:
```suggestion
            // maps a texture format name to a slot index
```

src/scene/batching/batch-manager.js
```diff
@@ -943,7 +943,7 @@ class BatchManager {
         }
 
         batch2.meshInstance.castShadow = batch.meshInstance.castShadow;
-        batch2.meshInstance._shader = batch.meshInstance._shader;
+        batch2.meshInstance._shader = batch.meshInstance._shader.slice();
```
slimbuck, line 946: wondering, does this fix a bug?

src/scene/materials/material.js
```diff
@@ -380,10 +381,39 @@ class Material {
...
+            createShaderDefinition: function (device, options) {
+                return shaderDefinition;
+            }
+        };
+
+        // temporarily register the program generator
+        const libraryModuleName = 'shader';
+        const library = device.getProgramLibrary();
+        Debug.assert(!library.isRegistered(libraryModuleName));
+        library.register(libraryModuleName, materialGenerator);
```
slimbuck, line 407: having to register then deregister is pretty messy. can the code be reshuffled to expose getProgram on the generator directly?

slimbuck (commented):
It seems like we're adding an invisible layer of processing to our shader code that will ultimately be difficult for users to debug if and when it goes wrong. I'm a little concerned that we're not really documenting the expectations or limitations of this step and it will ultimately become a magical black-box.

Are you happy with our approach of regex'ing shaders like this and processing the code in this way? Or do you expect we'll be rewrite this using different tools or ...?

src/scene/batching/batch-manager.js
mvaligursky: a possible bug yeah, not an actual one I observed. I don't think two mesh instances should share a private array. If one gets deallocated, the other should not keep their copy.

GSterbrant (approved):
Approved with a few comment/formatting comments.

src/scene/batching/batch-manager.js
mvaligursky: The before solution would save memory, as it would share the same array between two mesh instances. Now these are separate arrays. But the arrays are tiny, so I prefer the correct behaviour in all cases.

src/graphics/shader-processor-options.js
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Constructs shader processing options, used to process the shader for uniform buffer support.
+     *
+     * @param {UniformBufferFormat} viewUniformFormat - Format if the uniform buffer.
+     * @param {BindGroupFormat} viewBindGroupFormat - Format of the bind group.
+     */
+    constructor(viewUniformFormat, viewBindGroupFormat) {
+
+        // construct a sparse array
+        this.uniformFormats[BINDGROUP_VIEW] = viewUniformFormat;
```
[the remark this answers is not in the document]
mvaligursky: yes

src/graphics/shader-processor-options.js
```diff
@@ -0,0 +1,54 @@
...
+    /** @type {Array<UniformBufferFormat>}*/
+    uniformFormats = [];
+
+    /** @type {Array<BindGroupFormat>}*/
+    bindGroupFormats = [];
+
+    /**
+     * Constructs shader processing options, used to process the shader for uniform buffer support.
+     *
+     * @param {UniformBufferFormat} viewUniformFormat - Format if the uniform buffer.
```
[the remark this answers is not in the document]
mvaligursky: fixed, thanks!

src/scene/mesh-instance.js
```diff
@@ -661,11 +672,14 @@ class MeshInstance {
...
-     * @param {any} sortedLights - Array of array of lights.
+     * @param {any} sortedLights - Array of arrays of lights.
+     * @param {UniformBufferFormat} viewUniformFormat - THe format of the view uniform buffer.
+     * @param {BindGroupFormat} viewBindGroupFormat - The format of the view bind group.
      * @ignore
      */
-    updatePassShader(scene, pass, staticLightList, sortedLights) {
-        this._shader[pass] = this.material.getShaderVariant(this.mesh.device, scene, this._shaderDefs, staticLightList, pass, sortedLights);
+    updatePassShader(scene, pass, staticLightList, sortedLights, viewUniformFormat, viewBindGroupFormat) {
+        this._shader[pass] = this.material.getShaderVariant(this.mesh.device, scene, this._shaderDefs, staticLightList, pass, sortedLights,
```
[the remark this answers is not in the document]
mvaligursky: to make the line shorter. We have a lint rule and the next line when those are parameters need to be aligned with the start of the parameters of the previous line. Strange, but lint.

mvaligursky:
> Are you happy with our approach of regex'ing shaders like this and processing the code in this way? Or do you expect we'll be rewrite this using different tools or ...?

It's hard to know at this stage. Right now my focus is to transpile GLSL shaders so that I can directly execute them on WebGPU device, allowing me to develop the rest of the WebGPU rendering. This is a black box, but a pretty small one, and executes for now only on WebGPU.

There is also an option to execute this on WebGl to add support for Uniform Buffers in the future.

Even if we went ahead and rewrote all shaders in the WebGPU native shader language, due to the current lack of support of reflection for those, we'd need to do processing there as well and manually assign slots.

PR #4405, opened 2022-07-07
Encoding cleanup

slimbuck:
This PR:
- simplifies handling of texture encoding logic in the standard material
- splits out encode, decode and spherical helper chunks
- removes the hideous `texture2DRGBM` shader functions
- skybox can now render any of the supported encodings


willeastcott (approved):
Really nice refactor - cool!

src/graphics/program-lib/chunks/common/frag/envAtlas.js
```diff
@@ -0,0 +1,23 @@
+export default /* glsl */`
+// the envAtlas is fixed at 512 pixels. every equirect is generated with 1 pixel boundary.
+const float atlasSize = 512.0;
```
mvaligursky, line 3: not related to this PR, but don't we handle larger atlases now?

slimbuck: We do, but actually still layout them out as if they were 512.

mvaligursky (approved):
looks great. any problem with changed options - I suspect the old options would not be commonly used?

PR #4409, opened 2022-07-08
Refactor of glTF extension handling

GSterbrant:
### Description
Refactors the way we deal with glTF extensions, giving us a better overview of the extensions we support and moving the inline extension handling to their own functions.

### DIFF
--- a/src/resources/parser/glb-parser.js
+++ b/src/resources/parser/glb-parser.js
@@ -966,7 +966,7 @@ const extractTextureTransform = function (source, material, maps) {
     }
 };
 
-const pbrSpecGlossiness = function (data, material, textures) {
+const extensionPbrSpecGlossiness = function (data, material, textures) {
     let color, texture;
     if (data.hasOwnProperty('diffuseFactor')) {
         color = data.diffuseFactor;
@@ -1011,7 +1011,7 @@ const pbrSpecGlossiness = function (data, material, textures) {
     }
 };
 
-const clearCoat = function (data, material, textures) {
+const extensionClearCoat = function (data, material, textures) {
     if (data.hasOwnProperty('clearcoatFactor')) {
         material.clearCoat = data.clearcoatFactor * 0.25; // TODO: remove temporary workaround for replicating glTF clear-coat visuals
     } else {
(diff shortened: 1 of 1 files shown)


src/resources/parser/glb-parser.js
```diff
@@ -1201,99 +1285,24 @@ const createMaterial = function (gltfMaterial, textures, flipV) {
...
+        "KHR_materials_clearcoat": clearCoat,
+        "KHR_materials_unlit": unlit,
+        "KHR_materials_specular": metallicF0
+    };
 
-            if (clearcoatNormalTexture.hasOwnProperty('scale')) {
-                material.clearCoatBumpiness = clearcoatNormalTexture.scale;
+    // Handle extensions
+    if (gltfMaterial.hasOwnProperty('extensions')) {
+        for (var key in gltfMaterial.extensions) {
```
mvaligursky, line 1298: const instead of var here please.

src/resources/parser/glb-parser.js
```diff
@@ -1201,99 +1285,24 @@ const createMaterial = function (gltfMaterial, textures, flipV) {
...
+            const extensionFunc = extensions[key];
+            if (extensionFunc != undefined) {
```
mvaligursky: either `!==` or simply `if (extensionFunc) {`

GSterbrant: Already fixed :D

src/resources/parser/glb-parser.js
```diff
@@ -1201,99 +1285,24 @@ const createMaterial = function (gltfMaterial, textures, flipV) {
...
-            const clearcoatNormalTexture = ccData.clearcoatNormalTexture;
-            material.clearCoatNormalMap = textures[clearcoatNormalTexture.index];
-
-            extractTextureTransform(clearcoatNormalTexture, material, ['clearCoatNormal']);
+    // Provide list of supported extensions and their functions
+    const extensions = {
+        "KHR_materials_pbrSpecularGlossiness": pbrSpecGlossiness,
+        "KHR_materials_clearcoat": clearCoat,
+        "KHR_materials_unlit": unlit,
+        "KHR_materials_specular": metallicF0
```
mvaligursky, line 1293: perhaps give the function names some common naming .. for example instead of `clearCoat` have `extensionClearCoat`.

GSterbrant: Good point, done.

mvaligursky (approved):
Looks good to me. 
Next step could go even further, similar to what I've done with other large files in the engine: create a folder called `glb-parser`, and move this file into it. Then create a file per extension in the same folder.

GSterbrant:
Yeah, that could be a good plan to prepare for the potentially big library of extensions we will support.

PR #4412, opened 2022-07-08
Postprocessing engine example now uses the chess board

mvaligursky:
Screenshot 2022-07-08 at 17 19 55


willeastcott (approved):
Oooo so much prettier! 😍

PR #4414, opened 2022-07-08
Support RGBP encoding

slimbuck:
This PR adds support for a new HDR texture encoding (similar to RGBM).

Instead of storing pixel intensity in alpha, RGBP stores roughly `1.0 - max(1.0, pixel_intensity)`.

This has some desirable properties, especially when stored as PNG:
- should solve most banding issues on safari browsers which still premultiply PNG alpha (since only bright areas now lose bits)
- PNG file size is reduced by up to 20% in typical environments due to simpler alpha data in LDR areas
- resulting images are easier to manage, since much more colour data is left intact
- slightly higher quality HDR areas compared to RGBM
- pixel data is filterable (like RGBM, but unlike RGBE)

For example, given this HDR image:
hdr

RGBM colour and alpha (3.6MB):
rgbm

RGBP colour and alpha (3.1MB):
rgbp


src/graphics/constants.js
```diff
@@ -984,6 +984,13 @@ export const TEXTURETYPE_RGBM = 'rgbm';
  */
 export const TEXTURETYPE_RGBE = 'rgbe';
 
+/**
+ * Texture stores high dynamic range data in RGBP encoding.
```
willeastcott, line 988: Wondering if some detail on the format is helpful here. Same goes for RGBM and RGBE I guess.

willeastcott (approved):
This is super-cool. Any benchmark stats covering quality and size would be good though before merging, just to give us confidence this is genuinely superior.

slimbuck:
I created an RGBP version of a particularly problematic lightmap originally taken from https://playcanvas.com/project/345310/overview/orange-room.

The result running on iphone 7:


And comparison of the two lightmaps:
| encoding | size (gzipped) | lightmap |
| :-: | :-: | --- |
| RGBM | 5.8MB |  |
| RGBP | 4MB |  |

willeastcott:
Wow. I'd say that a massive win! 😮

src/graphics/env-lighting.js
```diff
@@ -10,6 +10,7 @@ import {
 import { DebugGraphics } from './debug-graphics.js';
 
 const fixCubemapSeams = true;
+const RGBA8_TYPE = TEXTURETYPE_RGBM;
```
mvaligursky, line 13: - local variables should not be all capital in `RGBA8_TYPE`
- should this default to RGBP now?

mvaligursky: ah this actually is a global constant .. ignore the first point :)

src/graphics/program-lib/chunk-utils.js
```diff
@@ -1,15 +1,28 @@
 const decodeTable = {
-    linear: 'decodeLinear',
-    srgb: 'decodeGamma',
-    rgbm: 'decodeRGBM',
-    rgbe: 'decodeRGBE'
+    'linear': 'decodeLinear',
```
mvaligursky, line 2: I think these tables should use TEXTURETYPE_RGBP and similar constants, instead of stating the strings again?
for example:
```
TEXTURETYPE_RGBP: 'decodeRGBP'
```

mvaligursky:
This looks really awesome, simple solution that solves the problem, and likely even faster to encode to!

src/graphics/program-lib/chunk-utils.js
slimbuck: So this table is actually based on texture encodings not texture types, see https://github.com/playcanvas/engine/blob/main/src/graphics/texture.js#L652

The difference between texture encoding and type is subtle and I don't like it.

Texture encoding is based on texture type and pixel format:
- texture with type 'default' and RGB8 pixels has encoding 'srgb'
- texture with type 'default' and RGB16F pixels has encoding 'linear'
- texture with type 'rgbm/rgbe/rgbp' is always encoding 'rgbm/rgbe/rgbp' no matter the pixel format.

Ideally we'd have just one enum, since it's confusing have two concepts.

(I actually added 'encoding' to texture as a helper, but it's getting used more and more).

src/graphics/constants.js
slimbuck: Agree, will add something to the engine in a future PR.

PR #4424, opened 2022-07-12
Add helper for getting shader pass type

slimbuck:
This PR adds a helper function for categorising shader passes into pass types and uses them when building shaders.

This fixes invalid shaders in editor when rendering custom passes (specifically pass 24 used for outline rendering).

This following function is added:
```javascript
ShaderPass.getType(shaderPass)
```

Which returns one of the following new constants:
```javascript
const SHADERTYPE_FORWARD = 'forward';
const SHADERTYPE_DEPTH = 'depth';
const SHADERTYPE_PICK = 'pick';
const SHADERTYPE_SHADOW = 'shadow';
```


src/graphics/constants.js
```diff
@@ -1207,3 +1207,31 @@ semanticToLocation[SEMANTIC_ATTR15] = 15;
  */
 export const CHUNKAPI_1_51 = '1.51';
 export const CHUNKAPI_1_55 = '1.55';
+
```
mvaligursky, line 1210: this should be in the scene/constants

mvaligursky: and perhaps it should be SHADERPASSTYPE_

slimbuck: > this should be in the scene/constants

Don't think so. Ideally the graphics module (lit-shader, standard etc) would base all its logic only on PASSTYPE and not know anything about SHADER_ (since SHADER_ types are defined in scene).

Unfortunately right now the function that maps between SHADER_ and PASSTYPE_ is implemented in scene, so graphics must know about it, but ideally we'd implement something in future which sorts that out.

(Probably standard-material-options-builder would pass in PASSTYPE_ instead of SHADER_).

slimbuck: > and perhaps it should be SHADERPASSTYPE_

I thought about that too, but actually think the correct name is probably RENDERPASSTYPE_. Both seem a bit long though?

slimbuck: > render pass is used for different concept

Ahh yes I forgot about your RenderPass class.

I'll name these `SHADERTYPE_` for now (since the `SHADER_` constants should probably be `SHADERPASS_`).

> I suspect I'll be moving lit-shader, standard and all that outside of the graphics

This makes sense actually, have graphics just wrap the hardware/API. I actually wonder whether having a 'materials' module might be better than lobbing it in with scene... Anyway I'll move these to scene for now.

src/scene/shader-pass.js
```diff
@@ -1,20 +1,54 @@
...
+} from './constants.js';
 
 /**
  * A pure static utility class, responsible for math operations on the shader pass constants.
  *
  * @ignore
  */
 class ShaderPass {
+    /**
+     * Returns true if the shader pass is a shadow shader pass.
```
mvaligursky, line 15: copy-paste description I think?

src/scene/shader-pass.js
```diff
@@ -1,20 +1,54 @@
...
+     *
+     * @param {number} shaderPass - The shader pass.
+     * @returns {string} - True if the pass is a shadow shader pass.
```
mvaligursky, line 18: and here

mvaligursky (approved):
approving, few comments need fixing.

PR #4429, opened 2022-07-13
Support for KHR_materials_ior

### LINKED ISSUE #3652: Support glTF extension: KHR_materials_ior

GSterbrant:
### Description

This PR adds support for controlling index of refraction (IOR) from the frontend of the shader chunks, adds support for the KHR_materials_ior extension, as well as precalculates the F0 specular factor with is dependent on IOR.

Fixes #3652 

This image illustrates, in Y diffuse color from black to red, in X refraction index from 1 to 2.
image

This illustrates the visual change on the specular when IOR is changed, as that impacts the f0 specular component. As in the previous image, in Y diffuse color ranges from black to red, in X refraction index from 1 to 2.
image

### DIFF
--- a/src/graphics/program-lib/programs/standard.js
+++ b/src/graphics/program-lib/programs/standard.js
@@ -305,13 +305,18 @@ const standard = {
             code.append(this._addMap("diffuse", "diffusePS", options, litShader.chunks));
             func.append("getAlbedo();");
 
+            if (options.refraction) {
+                decl.append("float dIor;");
+                code.append(this._addMap("refractionIndex", "iorPS", options, litShader.chunks));
+                func.append("getRefractionIndex();");
+            }
+
             // specularity & glossiness
             if ((litShader.lighting && options.useSpecular) || litShader.reflections) {
                 decl.append("vec3 dSpecularity;");
                 decl.append("float dGlossiness;");
                 if (options.useMetalness) {
                     decl.append("float dMetalness;");
-                    decl.append("float dIor;");
                     code.append(this._addMap("metalness", "metalnessPS", options, litShader.chunks));
                     func.append("getMetalness();");
                 }
--- a/src/scene/materials/standard-material-options-builder.js
+++ b/src/scene/materials/standard-material-options-builder.js
@@ -162,6 +162,7 @@ class StandardMaterialOptionsBuilder {
         options.cubeMapProjection = stdMat.cubeMapProjection;
         options.customFragmentShader = stdMat.customFragmentShader;
         options.refraction = !!stdMat.refraction;
+        options.refractionIndexTint = (stdMat.refractionIndex !== 1.5) ? 1 : 0;
         options.useMetalness = stdMat.useMetalness;
         options.enableGGXSpecular = stdMat.enableGGXSpecular;
         options.msdf = !!stdMat.msdfMap;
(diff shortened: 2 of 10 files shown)


src/scene/materials/standard-material.js
```diff
@@ -220,6 +220,7 @@ let _params = new Set();
  * indices of refraction, the one around the object and the one of its own surface. In most
  * situations outer medium is air, so outerIor will be approximately 1. Then you only need to do
  * (1.0 / surfaceIor).
+ * @property {number} f0 Defines the f0 weight as calculated using ((ior - 1) / (ior + 1))^2
```
willeastcott, line 223: Just wondering - does this really need to be public API? Would we expect somebody to set this in code considering it seems to be directly derived from the `ior (refractionIndex)` value?

willeastcott (approved):
Approving with a minor comment. This PR is a beautiful thing! 🌷

mvaligursky:
Would you have some screenshot demonstrating what it can be used for?

GSterbrant:
Added two screenshots showing both the refraction and how the specular effects change with the refraction index.

src/scene/materials/standard-material-options-builder.js
```diff
@@ -162,6 +162,7 @@ class StandardMaterialOptionsBuilder {
         options.cubeMapProjection = stdMat.cubeMapProjection;
         options.customFragmentShader = stdMat.customFragmentShader;
         options.refraction = !!stdMat.refraction;
+        options.refractionIndexTint = (stdMat.refractionIndex !== 1.5) ? 1 : 0;
```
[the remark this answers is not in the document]
willeastcott: Aha, got it, OK.

mvaligursky (approved):
looks really good to me.

src/graphics/program-lib/programs/standard.js
```diff
@@ -305,13 +305,18 @@ const standard = {
             code.append(this._addMap("diffuse", "diffusePS", options, litShader.chunks));
             func.append("getAlbedo();");
 
+            if (options.refraction) {
+                decl.append("float dIor;");
+                code.append(this._addMap("refractionIndex", "iorPS", options, litShader.chunks));
```
slimbuck, line 310: This function assumes a full map - so you need to add all the iorPS members to material - iorPSMap, iorPSUv, iorPSTiling, iorPSOffset etc and then handle them in the shader.

willeastcott: I think you mean `refractionIndexMap`, `refractionMapUv`, etc, right?

slimbuck: Yes!

Though I see from https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_ior/README.md that refraction index is a per-material constant (not texture or vertex driven).

@GSterbrant is it even necessary to have a customisation chunk for this? Seems like ior is literally just a float per material. User can set that on javascript - no need for customising chunks etc.

PR #4431, opened 2022-07-14
Reintroduce support for chunk $texture2DSAMPLE

slimbuck:
We recently removed support for `$texture2DSAMPLE(...)` macro in shaders since it's completely unnecessary ( #4405). The way to do this now is: `$DECODE(texture2D(...))`.

However, almost all overridden lightmap and emissive chunks use this macro and so this PR re-introduces the macro to remain backwards compatible.

In the process, we also shuffle `gammaCorrect` function to use the already-available `decodeXXX` functions instead of reimplementing them.


src/graphics/program-lib/chunks/common/frag/gamma2_2.js
```diff
@@ -1,40 +1,21 @@
...
 vec3 gammaCorrectOutput(vec3 color) {
-    #ifdef HDR
+#ifdef HDR
     return color;
-    #else
-    color += vec3(0.0000001);
-    return pow(color, vec3(0.45));
-    #endif
+#else
+    return pow(color + 0.0000001, vec3(1.0 / 2.2));
```
mvaligursky, line 18: doesn't need to be vec3(0.0000001) ?

src/graphics/program-lib/programs/standard.js
```diff
@@ -155,6 +155,18 @@ const standard = {
 
             if (encoding) {
                 subCode = subCode.replace(/\$DECODE/g, ChunkUtils.decodeFunc((!options.gamma && encoding === 'srgb') ? 'linear' : encoding));
+
+                // continue to support $texture2DSAMPLE
+                if (subCode.indexOf('$texture2DSAMPLE')) {
+                    const decodeTable = {
```
mvaligursky, line 161: keep this object at the global scope?

src/graphics/program-lib/programs/standard.js
```diff
@@ -371,6 +383,14 @@ const standard = {
                 code.append(this._addMap("light", lightmapChunkPropName, options, litShader.chunks, options.lightMapEncoding));
                 func.append("getLightMap();");
             }
+
+            // only add the legacy chunk if it's referenced
+            if (code.code.indexOf('texture2DSRGB') !== -1 ||
```
mvaligursky, line 388: could this be tracked differently than using .indexOf?
Likely not I assume, as any chunk can just call this directly.

mvaligursky (approved):
approving with some minor comments

willeastcott (approved):

src/graphics/program-lib/programs/standard.js
```diff
@@ -371,6 +383,14 @@ const standard = {
...
+                code.code.indexOf('texture2DRGBM') !== -1 ||
+                code.code.indexOf('texture2DRGBE') !== -1) {
+                code.prepend(litShader.chunks.textureSamplePS);
```
willeastcott, line 391: Worth a debug warning, asking people to update their code?

src/graphics/program-lib/chunks/common/frag/gamma2_2.js
slimbuck: nope this works. the float is added to vector components.

src/graphics/program-lib/programs/standard.js
slimbuck: I am convinced the compiler does not construct/allocate this object every time the function is called (but rather hoists it globally or something similar - equivalent to c++ const block). Having it declared here where it is used is much clearer for the reader.

Now just need to find some confirmation of my conviction on the webs...

src/graphics/program-lib/programs/standard.js
slimbuck: Yeah this is no nasty :`(

Alternatively since only emissive and lightmap chunks do this, we could somehow pass flags around tracking the use of these functions instead. This code is just so much simpler.

PR #4432, opened 2022-07-14
Fix lightmap dir not using halfdir

GSterbrant:
### Description

Fixes lightmapDir.js using getLightSpecular but not providing a half direction for the light. This issue was introduced in #4386.

### DIFF
--- a/src/graphics/program-lib/chunks/lit/frag/lightmapDirAdd.js
+++ b/src/graphics/program-lib/chunks/lit/frag/lightmapDirAdd.js
@@ -11,7 +11,7 @@ void addLightMap() {
 
         dDiffuseLight += dLightmap * nlight * 2.0;
     }
-
-    dSpecularLight += dLightmap * getLightSpecular();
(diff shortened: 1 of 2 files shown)


src/graphics/program-lib/chunks/lit/frag/lightmapDirAdd.js
```diff
@@ -11,7 +11,7 @@ void addLightMap() {
 
         dDiffuseLight += dLightmap * nlight * 2.0;
     }
-
-    dSpecularLight += dLightmap * getLightSpecular();
+    vec3 halfDir = normalize(normalize(-dLightDirNormW) + normalize(dViewDirW));
```
slimbuck, line 14: I believe both of these are already normalised?

GSterbrant: Yes, that is right. I think the same mistake exists in clusteredLights.js, fixing it too.

PR #4435, opened 2022-07-14
Remove redundant normalisations

GSterbrant:
### Description

Short and simple, this accidentally made it into the #4386 PR.

### DIFF
--- a/src/graphics/program-lib/chunks/lit/frag/clusteredLight.js
+++ b/src/graphics/program-lib/chunks/lit/frag/clusteredLight.js
@@ -512,7 +512,7 @@ void evaluateLight(ClusterLightData light) {
             // specular and clear coat are material settings and get included by a define based on the material
             #ifdef CLUSTER_SPECULAR
 
-                vec3 halfDir = normalize(normalize(-dLightDirNormW) + normalize(dViewDirW));
+                vec3 halfDir = normalize(-dLightDirNormW + dViewDirW);
                 
                 // specular
                 #ifdef CLUSTER_SPECULAR_FRESNEL


slimbuck (approved):
smokin hot!

kungfooman:
Is this really redundant or something else going on? The term redundant implies for me that the result is basically the same, but that's not what I observe when I test it:

```js
a = new pc.Vec3(10,20,30)
a.normalize()
b = new pc.Vec3(300,160,60)
b.normalize()
c = new pc.Vec3().add2(a, b)
c.normalize()
// Vec3 {x: 0.6313556044752356, y: 0.5545397420309195, z: 0.5421031038520173}
```

And removing the two normalisations:

```js
a = new pc.Vec3(10,20,30)
b = new pc.Vec3(300,160,60)
c = new pc.Vec3().add2(a, b)
c.normalize()
// Vec3 {x: 0.8387573601608582, y: 0.4870204026740467, z: 0.24351020133702336}
```

I just wonder what would be "more right".

willeastcott:
@kungfooman In the GLSL, `dLightDirNormW` and `dViewDirW` are already normalized. In your code, `a` and `b` are not.

kungfooman:
Thank you @willeastcott that makes sense. Just wondered if there are more such cases and found some:

https://github.com/playcanvas/engine/blob/ed0ccb055596a91f1c3e700944d1f20ebeeebed7/src/graphics/program-lib/programs/lit-shader.js#L994

https://github.com/playcanvas/engine/blob/ed0ccb055596a91f1c3e700944d1f20ebeeebed7/src/graphics/program-lib/programs/lit-shader.js#L277

And `getNormal()` seems to be always normalized: https://github.com/playcanvas/engine/blob/ed0ccb055596a91f1c3e700944d1f20ebeeebed7/src/graphics/program-lib/chunks/lit/vert/normal.js#L39

(there are four of them)

`dNormalW` seems to be always normalized:

https://github.com/playcanvas/engine/blob/ed0ccb055596a91f1c3e700944d1f20ebeeebed7/src/graphics/program-lib/chunks/lit/frag/reflDirAniso.js#L8

`dVertexNormalW` seems to be always normalized:

https://github.com/playcanvas/engine/blob/ed0ccb055596a91f1c3e700944d1f20ebeeebed7/src/graphics/program-lib/chunks/lit/frag/TBN.js#L3

https://github.com/playcanvas/engine/blob/ed0ccb055596a91f1c3e700944d1f20ebeeebed7/src/graphics/program-lib/chunks/lit/frag/TBNObjectSpace.js#L28

(not sure if its true, I am not good at GLSL, just following what makes sense to me)

willeastcott:
@slimbuck?

slimbuck:
`vNormalW` is an interpolant - the vertex shader sets the normal at each polygon vertex which is then interpolated across the surface of the polygon. When the fragment shader receives it (unless all 3 normals point in the same direction) the normal will no longer be length 1.0. (Which is why we renormalise it in the fragment shader).

We can likely remove the renormalization of `dVertexNormalW` in  TBN.js and TBHObjectSpace.js though, but I'll leave that for @GSterbrant to investigate.

PR #4436, opened 2022-07-14
Only degamma specular if we use a metalness specular color map

GSterbrant:
### Description

Specular should only apply degamma when using specular maps in the metalness workflow.

### DIFF
--- a/src/graphics/program-lib/programs/standard.js
+++ b/src/graphics/program-lib/programs/standard.js
@@ -321,6 +321,9 @@ const standard = {
                     func.append("getSpecularityFactor();");
                 }
                 if (options.useSpecularColor) {
+                    if (options.degammaSpecular) {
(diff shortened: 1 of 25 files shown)


src/graphics/program-lib/programs/standard.js
```diff
@@ -339,7 +339,7 @@ const standard = {
                     func.append("getSpecularityFactor();");
                 }
                 if (options.useSpecularColor) {
-                    code.append(this._addMap("specular", "specularPS", options, litShader.chunks));
+                    code.append(this._addMap("specular", "specularPS", options, litShader.chunks, options.specularEncoding));
```
slimbuck, line 342: I think we want this:

```suggestion
                    code.append(this._addMap("specular", "specularPS", options, litShader.chunks, options.useMetalness ? 'srgb' : 'linear'));
```

slimbuck (approved):

PR #4437, opened 2022-07-14
Adds support for refraction as a texture and KHR_materials_transmission

### LINKED ISSUE #3648: Support glTF extension: KHR_materials_transmission
https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_transmission/README.md

This extension aims to address the simplest and most common use cases for optical transparency: infinitely-thin materials with no refraction, scattering, or dispersion. Dealing exclusively with “thin” materials (i.e. materials where only the surface is considered and not the volume) allows many simplifications to be made when calculating things like refraction and absorption.

image

GSterbrant:
### Description

This PR adds support for the KHR_materials_transmission as well as provides a new texture to the standard material for refraction. Fixes #3648.

This image shows the glTF example TransmissionTest. Note that we don't deal with handling refraction of the opaque objects in the scene in this PR.
image

### DIFF
--- a/src/graphics/program-lib/chunks/lit/frag/refraction.js
+++ b/src/graphics/program-lib/chunks/lit/frag/refraction.js
@@ -1,5 +1,4 @@
 export default /* glsl */`
-uniform float material_refraction;
 uniform float material_refractionIndex;
 
 vec3 refract2(vec3 viewVec, vec3 Normal, float IOR) {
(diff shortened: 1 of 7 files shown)


src/graphics/program-lib/chunks/lit/frag/refraction.js
```diff
@@ -1,5 +1,6 @@
 export default /* glsl */`
-uniform float material_refraction;
+uniform sampler2D uSceneColorMap;
+uniform vec4 uScreenSize;
```
slimbuck, line 3: remove these?

slimbuck (approved):
approved with comment

mvaligursky:
Do we need to create a follow up ticket, or link an existing ticket?

mvaligursky (approved):

PR #4440, opened 2022-07-15
Handle shader-decoded texture alpha correctly

slimbuck:
This PR fixes a shader compile error resulting from user choosing the alpha channel of a decoded texture. (This is because the decoding step takes a vec4 and returns a vec3 so alpha doesn't exist).

Since there is never actual decoding of texture alpha (i.e. gamma textures don't correct alpha and RGBM/P/E textures use alpha channel to reconstruct RGB) we just add a passthrough for this case instead.

Also move `textureSample.js` chunk to the correct location.


mvaligursky:
I just wonder - how come textures stored in sRGB don't gamma correct the alpha channel? If I'm reading this right.

slimbuck:
> if I'm reading this right.

Yep you are: alpha is just never stored in gamma space (presumably because it's not a color value).

mvaligursky:
Ah, looking online, it certainly seems to be the case!

mvaligursky (approved):

PR #4447, opened 2022-07-15
Append Action Function

jath-git:
Fixes #


kungfooman:
Why is this minus and not minus, shouldn't it be both minus then? Then it can be shortened aswell.

https://github.com/playcanvas/engine/blob/4cbd6b8289d0386844f8639b947220eb67cd539f/src/input/controller.js#L186-L197

jath-git:
I updated the pull request by adding the appendAction function for button

PR #4449, opened 2022-07-18
Fix legacy shader chunk combine functions

slimbuck:
This PR fixes the legacy combine chunks to not factor dSpecularity twice.

Also fixes odd bits that have changed while preparing the current engine release.

### DIFF
--- a/src/resources/cubemap.js
+++ b/src/resources/cubemap.js
@@ -120,10 +120,10 @@ class CubemapHandler {
                     resources[i + 1] = new Texture(this._device, {
                         name: cubemapAsset.name + '_prelitCubemap' + (tex.width >> i),
                         cubemap: true,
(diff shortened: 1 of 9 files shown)


src/resources/cubemap.js
```diff
@@ -120,7 +120,11 @@ class CubemapHandler {
                     resources[i + 1] = new Texture(this._device, {
                         name: cubemapAsset.name + '_prelitCubemap' + (tex.width >> i),
                         cubemap: true,
-                        type: TEXTURETYPE_RGBM,
+                        // we have some old assets that were not created in RGBM format (see gangnum
```
willeastcott, line 123:
```suggestion
                // we have some old assets that were not created in RGBM format (see Gangnam
```

PR #4450, opened 2022-07-18
[FIX] Remove guard that always evaluates to true in wasmSupported

willeastcott:
Fixes the following (from [lgtm.com](https://lgtm.com/projects/g/playcanvas/engine/snapshot/207d4fbf9c0e23e3feea5f96da7614710487c3b8/files/src/core/wasm-module.js?sort=name&dir=ASC&mode=heatmap#x4c53ea2a7a033341:1)):

image

### DIFF
--- a/src/core/wasm-module.js
+++ b/src/core/wasm-module.js
@@ -3,20 +3,16 @@ class Impl {
 
     // returns true if the running host supports wasm modules (all browsers except IE)
     static wasmSupported() {
-        let supported = null;
-
-        if (supported === null) {
-            supported = (() => {
-                try {
-                    if (typeof WebAssembly === "object" && typeof WebAssembly.instantiate === "function") {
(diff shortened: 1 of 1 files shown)


willeastcott:
@slimbuck I'm not sure if the original intention was that `supported` would be a static var and the test would be only run once. But for now, I've just gotten rid of the useless `if` statement.

willeastcott:
In fact, this doesn't even need to be wrapped in a closure. So yeah, I'm guessing the intention was to run the check once and return that cached result on future calls to `wasmSupported`.

slimbuck:
> In fact, this doesn't even need to be wrapped in a closure. So yeah, I'm guessing the intention was to run the check once and return that cached result on future calls to `wasmSupported`.

Yes, exactly - will fix ASAP.

kungfooman:
Maybe a little util function would make the intent clearer:

```js
function returnCache(fn) {
  const cache = fn();
  return () => cache;
}
cachedNow = returnCache(()=>Date.now());
```

image

```js
class Test {
  static dateOfCreation = returnCache(()=>Date.now())
}
```

image

PR #4458, opened 2022-07-21
remove negative mouse action requirement

### LINKED ISSUE #4463: Positive Mouse Action
Mouse input actions should be negative. In src/input/controller.js, button is set to a positive value.
```sh
line 183: button: button
```
Opened a pull request to resolve this.

jath-git:
Fixes #4463


src/input/controller.js
```diff
@@ -177,17 +184,10 @@ class Controller {
...
-            });
-        } else {
-            this._actions[action] = [{
-                type: ACTION_MOUSE,
-                button: -button
-            }];
-        }
+        this.appendAction(action, {
+            type: ACTION_MOUSE,
+            button: -button
```
slimbuck, line 189: Interesting... so this fixes a bug where old code pushed `button` instead of `-button`?

src/input/controller.js
```diff
@@ -202,19 +202,11 @@ class Controller {
...
-                pad: pad
-            });
-        } else {
-            this._actions[action] = [{
-                type: ACTION_GAMEPAD,
-                button: button,
-                pad: pad
-            }];
-        }
+        this.appendAction(action, {
```
slimbuck, line 205: Would you mind also updating the comment above this line which is clearly wrong?

jath-git: I have updated the pull request with this fix

slimbuck (approved):
Lovely!

yaustar:
Is there an example project/problem that we can test against/look at?

jath-git:
@yaustar, I'm not too sure about specific example tests. Changing to -button was suggested by Hermann. @kungfooman could you provide any assistance?

kungfooman:
Negating buttons for "prevent clashing with keycodes" doesn't make any sense to me, the entire negating business seems to be an old artifact that is buggy now and should probably be stripped altogether.

I made a simple snippet so everyone can test for themselves, just open https://playcanvas.com/viewer and paste this in devtools:

```js
let app = pc.app;
let {canvas} = app.scene.device;
let controller = new pc.Controller(canvas, {
  mouse: pc.app.mouse
});
controller.registerMouse("attack", pc.MOUSEBUTTON_LEFT  );
controller.registerMouse("nade"  , pc.MOUSEBUTTON_MIDDLE);
controller.registerMouse("zoom"  , pc.MOUSEBUTTON_RIGHT );
app.on('update', () => {
  if (controller.isPressed('attack')) {
    console.log('attack');
  }
  if (controller.isPressed('nade')) {
    console.log('nade');
  }
  if (controller.isPressed('zoom')) {
    console.log('zoom');
  }
});
```

Should work, right? Well, no, because keycodes are negated but then actually not "negated back" here:

https://github.com/playcanvas/engine/blob/3a9e881f2dff74923076ec4497929c11fffe2c15/src/input/controller.js#L311-L315

But... attack works! Because `0 === -0`.

In order to make it work one would have to write this code with negations right now:

```js
controller.registerMouse("attack", -pc.MOUSEBUTTON_LEFT  );
controller.registerMouse("nade"  , -pc.MOUSEBUTTON_MIDDLE);
controller.registerMouse("zoom"  , -pc.MOUSEBUTTON_RIGHT );
```

I can't believe this was ever intended. Is this deprecated/legacy?

src/input/controller.js
```diff
@@ -125,6 +125,20 @@ class Controller {
...
+     *
+     * @param {string} action - The name of the action.
+     * @param {object} item - An object to add.
+     */
+    appendAction(action, item) {
+        if (this._actions[action]) {
+            this._actions[action].push(item);
+        } else {
+            this._actions[action] = [item];
+        }
```
kungfooman, line 139:
```suggestion
        this._actions[action] = this._actions[action] || [];
        this._actions[action].push(item);
```

(IMO this creates less confusion and less potential bugs like the +- issue)

src/input/controller.js
```diff
@@ -177,17 +185,10 @@ class Controller {
...
-        } else {
-            this._actions[action] = [{
-                type: ACTION_MOUSE,
-                button: -button
-            }];
-        }
+        this.appendAction(action, {
+            type: ACTION_MOUSE,
+            button: -Math.abs(button)
+        });
```
kungfooman, line 191:
```suggestion
        this.appendAction(action, {
            type: ACTION_MOUSE,
            button
        });
```

kungfooman (changes requested):
Please countercheck with example snippet I posted, hard to validate through GitHub suggestions only

slimbuck:
Looking through the history it seems the negate bug was introduced 10 years ago https://github.com/playcanvas/engine/commit/779ffc7fa46cbb1113684fb1f1a826f4ad1654f1#diff-ab4935d07ba8a4243978a03f478cff9ff45ceecef49b62cdbca39a70e2169b61L137. Notice one of the `-button`s got lost in translation. Also notice that negate was there before the action type structure was introduced.

So yes, @kungfooman you're absolutely right, we can (should!) remove the negate altogether (it is also something @GSterbrant pointed out).

Are you happy to make that change @jath-git (or accept @kungfooman 's suggestion)?

jath-git:
@slimbuck Agreed. Absolutely, I'll work on making those changes now

kungfooman:
https://github.com/playcanvas/engine/blob/1d3da63969629b60e82800c834547864e47f46c0/src/input/controller.js#L129-L135

It sounds a bit strange, every function is reusable? I also don't know what it is checking or validating. The JSDoc can be more specific:

```js
    /**
     * Helper function to push an action.
     *
     * @param {string} name - The name of the action.
     * @param {object} action - An object to add.
     * @param {ACTION_KEYBOARD|ACTION_MOUSE|ACTION_GAMEPAD} action.type - The name of the action.
     * @param {number[]} [action.keys] - Keyboard: A list of keycodes like `[pc.KEY_A, pc.KEY_ENTER]`.
     * @param {number} [action.button] - Mouse: e.g. `pc.MOUSEBUTTON_LEFT` Gamepad: E.g. `pc.PAD_FACE_1`
     * @param {number} [action.pad] - Gamepad: the index of the pad to register (use {@link PAD_1}, etc).
     */
    appendAction(name, action) {
        this._actions[name] = this._actions[name] || [];
        this._actions[name].push(action);
    }
```

Currently the signature is `appendAction(action, item)`, which feels also confusing to me. `action` feels like an object, but it is just the action name. Item is the actual action object.

slimbuck:
This is looking great - thanks so much @jath-git and @kungfooman!

@jpauloruschel would you mind taking a quick look before we merge?

jpauloruschel (approved):

PR #1080, opened 2018-02-02
More consistent maps

guycalledfrank:
- Reduces the amount of shader chunks needed for all maps;
- Supports multiplying texture * color * vertexColor (needed for GLTF)
- New property for each map: VertexColorChannel (e.g. diffuseVertexColorChannel)

PR #4460, opened 2022-07-21
Add support for KHR_materials_sheen

GSterbrant:
### Description

This PR adds support for the sheen BRDF, which is a good BRDF model for cloth and fabric materials. This PR also adds support for KHR_materials_sheen. Sheen is layered on top of clearcoat and the base layer, but is (approximately) energy conservative to the other layers. 

image


willeastcott:
I'm wondering whether the combining of the combiner chunks should be a separate PR...  What do other people think?

slimbuck:
> I'm wondering whether the combining of the combiner chunks should be a separate PR... What do other people think?

Agree. @GSterbrant perhaps you could submit a PR with changes to combine chunks first and include a description detailing the chunk changes, then followup with sheen PR?

GSterbrant:
Done, it's here: https://github.com/playcanvas/engine/pull/4461

Will revert those changes in this PR.

src/graphics/program-lib/chunks/standard/frag/sheenGloss.js
```diff
@@ -0,0 +1,27 @@
...
+
+    #ifdef MAPTEXTURE
+    sheenGlossiness *= texture2D(texture_sheenGlossinessMap, $UV, textureBias).$CH;
+    #endif
+
+    #ifdef MAPVERTEX
+    sheenGlossiness *= saturate(vVertexColor.$VC);
+    #endif
+
+    sheenGlossiness += 0.0000001;
```
slimbuck, line 25:
```suggestion
    sheenGlossiness += 0.0000001;
    
    sGlossiness = sheenGlossiness;
```

src/resources/parser/glb-parser.js
```diff
@@ -1145,6 +1144,59 @@ const extensionTransmission = function (data, material, textures) {
...
+    const sheenGlossChunk = `
+    #ifdef MAPFLOAT
+    uniform float material_sheenGlossiness;
+    #endif
+
+    #ifdef MAPTEXTURE
+    uniform sampler2D texture_sheenGlossinessMap;
+    #endif
+
+    void getSheenGlossiness() {
```
slimbuck, line 1177: any reason not to make this the default sheen glossiness chunk?

GSterbrant: glTF provides roughness, not glossiness, but I kept it as glossiness to be consistent with the rest of the chunks.

src/graphics/program-lib/chunks/standard/frag/sheenGloss.js
slimbuck: This chunk is calculating a local variable and not setting the global?

GSterbrant: Ah yes, I merged in the change but did a rebase on main and force push, which reverted it. It's fixed again.

src/resources/parser/glb-parser.js
slimbuck: Ok, but we are actually looking to -remove- all the custom gltf chunks from gltf-loader (by adding support to standard-material for the custom behaviours).

This way, in future, users will be able to create and modify gltf materials in the editor (which they wouldn't be able to do given these missing features).

So overall I'd suggest either supporting the gltf chunk in standard-material as-is or keep the consistency but expose an option to get this behaviour.

GSterbrant: Yeah, but the way glTF deals with microsurface descriptions is by using roughness, but all our UI and whatnot is designed for glossiness. 

So unless we want to change so sheen has a Sheen Roughness parameter, instead of Sheen Glossiness like the rest of the parameters, we should do this to not be confusing.

slimbuck: Could just be an 'invert' option on the gloss, ccgloss and this one?

GSterbrant: Yeah, we could have that, but ideally we would never want to invert the glossiness at all. But sure, we could definitely create a PR for having an invert flag on the glossiness in the editor.

slimbuck: Do you have any idea why gloss has the `+= 0.000001` term?

Seems this was originally added years ago in https://github.com/playcanvas/engine/pull/1080/files#diff-50e810cc6e8bd1669e9f4411172f8e6f573ba9c1e61bff5401606a09156aa187R24.

GSterbrant: I think you might divide by gloss, and this is to avoid a div by 0. If you look at sheenD, the first line does 1 / (roughness * roughness) so this is just a way to prevent that from resulting in undefined behaviour.

slimbuck (approved):

PR #4461, opened 2022-07-21
Merge the combine chunks

GSterbrant:
### Description 

The combine chunks were very inflexible and hard to use. This PR fixes that by providing a more modular approach to how to compose the final colour value. 

### Adds chunk:
combine.js

### Deletes chunks:
combineClearCoat
combineDiffuse
combineDiffuseSpecular
combineDiffuseSpecularNoRefl
combineDiffuseSpecularNoReflSeparateAmbient
combineDiffuseSpecularOld

### Chunk changes:
clusteredLight.js - Replaced the ```CLUSTER_(SPECULAR/CONSERVE_ENERGY/SPECULAR_FRESNEL/CLEARCOAT)``` with the new ```LIT_``` counterparts
end.js - Removed the ```CLEARCOAT``` define
lightSpecularX.js/lightmapDirAdd.js/ltc.js/reflectionCC.js/start.js - Replaced ```CLEARCOAT``` with ```LIT_CLEARCOAT```


willeastcott:
Shouldn't this PR be called 'Combine the combine chunks'? 😄

src/graphics/program-lib/chunks/lit/frag/combine.js
```diff
@@ -0,0 +1,20 @@
+export default /* glsl */`
+vec3 combineColor() {
+    vec3 ret = vec3(0, 0, 0);
```
willeastcott, line 3:
```suggestion
    vec3 ret = vec3(0);
```

src/graphics/program-lib/programs/lit-shader.js
```diff
@@ -874,6 +874,15 @@ class LitShader {
 
             // enable specular path in clustered chunk
             code += "#define CLUSTER_SPECULAR\n";
+            code += "#define COMBINE_SPECULAR\n";
```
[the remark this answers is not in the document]
GSterbrant: Done

src/graphics/program-lib/programs/lit-shader.js
```diff
@@ -608,17 +608,31 @@ class LitShader {
...
+
+            if (options.fresnelModel > 0) {
+                code += '#define LIT_SPECULAR_FRESNEL\n';
+            }
+
+            // enable conserve energy path in clustered chunk
+            if (options.conserveEnergy) {
+                code += "#define LIT_CONSERVE_ENERGY\n";
+            }
+        }
```
willeastcott, line 634: Just as a general point, when generating GLSL shaders, would it make sense to prepend all `#define`s to the start of the code? cc @slimbuck 
Note that I'm not suggesting we make this change in this PR!

slimbuck: Yes agree. I'd actually suggest adding a `globalDefines` variable and adding the defines to that. Then at the end of the function where the final shader is combined insert it at the right place (before this.varyings probably).

GSterbrant: Sounds like a good idea, we can provide a ```addDefine('LIT_SPECULAR_FRESNEL')``` and have that be added at shader generation time. 

I can do it for this PR if you think it would fit, and then have other defines merged over to it later.

GSterbrant: I made this change in the PR. We now inject defines on ```_fsGetBeginCode```.

slimbuck (approved):

PR #4462, opened 2022-07-21
Fix AudioContext not resuming properly.

jpauloruschel:
Fixes https://github.com/playcanvas/engine/issues/3977

## Summary

Fix `AudioContext` not resuming properly in Mobile devices. Previously, the `AudioContext` often broke when minimising / switching tabs and going back to the page. Once broken, no Audio could be played, and the only resolution was reloading the entire page. This PR addresses this by re-writing the entire `SoundManager ` suspend / resume flow.

I've added documentation to the code, but basically:
- If `AudioContext` is locked (verifiable by checking if the `AudioContext.state` is `suspended` right after creation), all `SoundInstance`s will wait until the Context is unlocked to actually start playing;
- Once unlocked, `SoundManager.resume()` (called manually or automatically on `visibilitychange`) will call `audioContext.resume()` if state was `interrupted`, and `SoundInstance`s will only resume once that call was successful;
- Once unlocked, every time the `AudioContext`'s `state` chages (`onstatechange`), an additional check will be made to resume the `AudioContext`.

## Testing

Test project using custom engine with changes: https://playcanv.as/p/nwJQKFeE/

Tested the following situations on iPhone Chrome and Safari:

- Test 1
    1. Open page
    2. Tap to start playing audio
    3. Minimize app, wait 30 seconds
    4. Back to app, audio resumes immediately
- Test 2
    1. Open page
    2. Minimize app, wait 30 seconds
    3. Back to app, audio doesn't resume (still locked)
    4. Tap to unlock, audio plays
- Test 3
    1. Open page
    2. Tap to start playing audio
    3. Minimize app, turn off screen, wait 30 seconds
    4. Turn screen back on, back to app, audio resumes immediately
   

Tested the following situations on MacBook Chrome and Safari:

- Test 1
    1. Open page
    2. Click to start playing audio
    3. Minimize app, wait 30 seconds
    4. Back to app, audio resumes immediately
- Test 2
    1. Open page
    2. Minimize app, wait 30 seconds
    3. Back to app, audio doesn't resume (still locked)
    4. Tap to unlock, audio plays

Tested the following situtation on MacBook & iPhone Chrome & Safari:

Screenshot 2022-08-01 at 14 10 18

(Available here: https://playcanv.as/p/nwJQKFeE/) Wait for tank to pass through each sphere - which performs the action in the text. This is done so there's no user interaction which would unlock the page - this way we can test what happens when context is unlocked in various stages.

### DIFF
--- a/src/sound/instance.js
+++ b/src/sound/instance.js
@@ -133,8 +133,29 @@ class SoundInstance extends EventHandler {
          */
         this._playWhenLoaded = true;
 
+        /**
+         * Set to true if a play() request was issued when the AudioContext was still suspended,
+         * and will therefore wait until it is resumed to play the audio.
+         *
+         * @type {boolean}
+         * @private
+         */
         this._waitingContextSuspension = false;
+
+        /**
+         * Set to true if a stop() request was issued while _waitingContextSuspension was also true.
+         *
+         * @type {boolean}
+         * @private
+         */
         this._stopOnWaitingContextSuspension = false;
+
+        /**
+         * Set to true if a pause() request was issued while _waitingContextSuspension was also true.
+         *
+         * @type {boolean}
+         * @private
+         */
         this._pauseOnWaitingContextSuspension = false;
 
         /**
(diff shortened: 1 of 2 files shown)


src/sound/instance.js
```diff
@@ -573,15 +598,51 @@ class SoundInstance extends EventHandler {
...
+     * @returns {boolean} True if the sound was started immediately.
      */
     play() {
         if (this._state !== STATE_STOPPED) {
             this.stop();
         }
+        // set state to playing
+        this._state = STATE_PLAYING;
+        // no need for this anymore
+        this._playWhenLoaded = false;
```
slimbuck, line 614: Looks like `slot.js` still references this member [here](https://github.com/playcanvas/engine/blob/main/src/framework/components/sound/slot.js#L167). If this member is no longer used, shouldn't it just be removed?

src/sound/instance.js
```diff
@@ -621,15 +677,18 @@ class SoundInstance extends EventHandler {
...
-            this._onManagerSuspend();
-        }
-
-        if (!this._suspendInstanceEvents)
+        if (!this._suspendInstanceEvents) {
             this._onPlay();
+        }
 
-        return true;
+        if (this._pauseOnWaitingContextSuspension) {
```
slimbuck, line 684: I am wondering about the mechanics of this bit - looks like this code could result in play & pause & stop being called on resume? Is it necessary to (possibly) call them all?

If you do need to call them all on resume, then presumably order matters? For example if the app attempted to play then pause then stop then play, presumably during resume you just want to play? (Where currently it'll play, then pause then stop the sound).

Seems like you may only need to track the current sound state and set that on resume? Sorry if I am misunderstanding everything!

src/sound/instance.js
jpauloruschel: Are you referencing the comment `no need for this anymore`? If so, these 4 lines of code were already in code, but I had to move them, and I think that the `no need for this anymore` is about no longer needing for `_playWhenLoaded` to be true. The way that property is used is weird, but I'd prefer not to change too much of that in this PR (unless you see another issue).

src/sound/instance.js
jpauloruschel: That makes total sense! Good callout, simplifies the code a lot!

slimbuck (approved):

PR #4480, opened 2022-07-26
Release 1.54 cut

peter-hutsul:
Fixes #


yaustar:
Hi Peter, can you describe what this PR is full please?

peter-hutsul:
> Hi Peter, can you describe what this PR is full please? Did you mean to make a PR for the main repo?

Hi, sorry about this. It's not finished yet.

In general i am thinking of adding few more versions without non-critical modules for v1.54 - to reduce overall playcanvas engine filesize

release-1.54-core - don't contains any non critical module
release-1.54-core-p - with particles
release-1.54-core-a - with animations
release-1.54-core-s - with sounds
release-1.54-core-pa - with particles and animations
release-1.54-core-ui - with UI
release-1.54-core-uipas - with UI, particles, animations, sounds
etc.

Or may be you can propose a better way to have multiple builds with different number of components / modules

yaustar:
Unfortunately , its unlikely that we will take on more build targets in this manner and would have  more configurable way for an end developer to specify what to include.

@mvaligursky has been looking at this but I'm not sure what their plans are long term. Personally, it would great if there was a config file or similar where I can add/remove component systems/modules from the engine to build

slimbuck:
Please see https://github.com/playcanvas/engine/pull/4082 for more context. Ultimately we'd like to make the engine tree-shakeable, even editor projects. We have a long way to go though.

mvaligursky:
We're heading towards the solution that was started here: https://github.com/playcanvas/engine/pull/4082
It allows users to build their engine based application using tree shaking. I already tested this on our model viewer project and also some standalone examples, and even as it's not a public interface yet, it's useable as is.

Two main areas we still need to solve for the full benefits are:
- scripts. Those are ES5 and not module based, and so are not compatible with tree-shaking. We might need to have a new, module based version, of scripts, or some other solution. If your project does not use ES5 scripts, this is non-issue I believe. But unfortunately, most project do use scripts.
- Editor integration - we need some way to easily configure the AppOptions (https://github.com/playcanvas/engine/blob/main/src/framework/app-options.js) in the Editor, and build the editor project including the engine, using rollup or similar, to apply the tree-shaking, and only include enabled / imported engine modules.

If in the meantime you need a smaller engine, you're welcome to do a custom build, but as it was mentioned, it's not the direction we'd like to take, and have a grid of build targets.

yaustar:
Closing the ticket as we won't be merging this but please feel free to continue the discussion :)

peter-hutsul:
> We're heading towards the solution that was started here: #4082 It allows users to build their engine based application using tree shaking. I already tested this on our model viewer project and also some standalone examples, and even as it's not a public interface yet, it's useable as is.
> 
> Two main areas we still need to solve for the full benefits are:
> 
> * scripts. Those are ES5 and not module based, and so are not compatible with tree-shaking. We might need to have a new, module based version, of scripts, or some other solution. If your project does not use ES5 scripts, this is non-issue I believe. But unfortunately, most project do use scripts.
> * Editor integration - we need some way to easily configure the AppOptions (https://github.com/playcanvas/engine/blob/main/src/framework/app-options.js) in the Editor, and build the editor project including the engine, using rollup or similar, to apply the tree-shaking, and only include enabled / imported engine modules.
> 
> If in the meantime you need a smaller engine, you're welcome to do a custom build, but as it was mentioned, it's not the direction we'd like to take, and have a grid of build targets.

Thank you.
Yes, definitely, if you manage to add module based system in the Editor - that's the only true solution.

Not sure, if this will help you, but i guess this scheme will work

**global.js**
`window.pc = {}`

**button/**
`pc.ButtonComponentSystem = ButtonComponentSystem;`

**aplication.js**
`if (pc.ButtonComponentSystem) this.ButtonComponentSystem = new pc.ButtonComponentSystem();`

PR #4485, opened 2022-07-28
Add support for KHR_materials_volume

### LINKED ISSUE #3651: Support glTF extension: KHR_materials_volume
https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_volume/README.md

By default, a glTF 2.0 material describes the scattering properties of a surface enclosing an infinitely thin volume. The surface defined by the mesh represents a thin wall. The volume extension makes it possible to turn the surface into an interface between volumes. The mesh to which the material is attached defines the boundaries of an homogeneous medium and therefore must be manifold. Volumes provide effects like refraction, absorption and scattering. Scattering is not subject of this extension.

image

GSterbrant:
Fixes #3651

Adds support for light attenuation and medium thickness to be used with refractive (light transmitting) volumes and surfaces. Also adds support for KHR_materials_volume.

Furthermore, increases the complexity of the refraction function to sample from the scene grab pass in order to have dynamic refractions. 

image

https://user-images.githubusercontent.com/107400752/184104765-34349042-c004-4765-ac2a-5012f94ed578.mp4

### **Chunk changes**
**Adds:**
- thickness (frontend) - controls thickness for refractive volumes
- refractionCube (backend) - implements old refraction behaviour
- refractionDynamic (backend) - implements new dynamic refractions using scene grab pass, and supports medium thickness

**Removes:**
- refraction (backend) - split into the two above


src/scene/materials/standard-material.js
```diff
@@ -1088,7 +1093,9 @@ function _defineMaterialProps() {
     _defineFloat('reflectivity', 1);
     _defineFloat('occludeSpecularIntensity', 1);
     _defineFloat('refraction', 0);
-    _defineFloat('refractionIndex', 1.0 / 1.5); // approx. (air ior / glass ior)
+    _defineFloat('refractionIndex', 1.5); // approx. (air ior / glass ior)
```
GSterbrant, line 1096: Ah, this one might be an issue in the future, just like how specular introduced a bunch of problems.

slimbuck (commented):
I think we must keep the old refraction chunk for use most of the time (since its much simpler and faster) and add the the new HQ refraction chunk as an option.

(Or instead of a material flag, perhaps we can base the choice of chunk on whether thickness is specified/non-zero? Since it looks like the new chunk requires some thickness).

GSterbrant:
It's more than that too. We don't refract with the environment map in this one, but use the grab pass result. But yes, I think there should be a flag to switch between volumetric refractions and the old.

GSterbrant:
The new refraction code is now behind a flag, useDynamicRefractions, which is enabled in the GLB loader by default.

slimbuck:
This looks awesome!

Please also add the chunk change to chunk-validation.js and update the developer chunk docs.

src/graphics/program-lib/chunks/lit/frag/refractionDynamic.js
```diff
@@ -0,0 +1,58 @@
+export default /* glsl */`
+uniform float material_refractionIndex;
+uniform float material_invAttenuationDistance;
+uniform vec3 material_attenuation;
+uniform mat4 matrix_viewProjection;
+uniform mat4 matrix_model;
+uniform sampler2D uSceneColorMap;
+uniform vec4 uScreenSize;
```
mvaligursky, line 8: I wonder about these 4 typically global uniforms being added to this chunk. How do we handle other chunks including those as well?

src/graphics/program-lib/chunks/lit/frag/refractionDynamic.js
```diff
@@ -0,0 +1,58 @@
...
+    // Project to texture space so we can sample it
+    vec4 projectionPoint = matrix_viewProjection * pointOfRefraction;
+    vec2 uv = projectionPoint.xy / projectionPoint.ww;
+    uv += vec2(1.0);
+    uv *= vec2(0.5);
+
+    // Use IOR and roughness to select mip
+    float iorToRoughness = (1.0 - dGlossiness) * clamp((1.0 / material_refractionIndex) * 2.0 - 2.0, 0.0, 1.0);
+    float refractionLod = log2(uScreenSize.x) * iorToRoughness;
+    vec3 refraction = textureLod(uSceneColorMap, uv, refractionLod).rgb;
```
mvaligursky, line 40: Perhaps you should use texture2DLodEXT macro instead of textureLod, which would cover Webgl2 & Webgl1 with extension. We'd likely still need a fallback for webgl1 devices without extension. Something like done here perhaps:
https://github.com/playcanvas/engine/blob/58ccb584f43d2df0c5338ce9cdf2a10a991deab6/examples/src/examples/graphics/grab-pass.tsx#L62-L67

mvaligursky: Please make sure to test on Webgl1 device.

mvaligursky:
Github supports video on the PR description .. please attach one.

src/graphics/program-lib/chunks/lit/frag/refractionCube.js
```diff
@@ -1,23 +1,18 @@
 export default /* glsl */`
 uniform float material_refractionIndex;
-
```
willeastcott, line 3: Maybe a bit overzealous in deleting newlines from this chunk?

src/scene/materials/standard-material.js
```diff
@@ -1150,6 +1161,7 @@ function _defineMaterialProps() {
     _defineFlag('occludeSpecular', SPECOCC_AO);
     _defineFlag('shadingModel', SPECULAR_BLINN);
     _defineFlag('fresnelModel', FRESNEL_SCHLICK); // NOTE: this has been made to match the default shading model (to fix a bug)
+    _defineFlag('useDynamicRefraction', false);
```
willeastcott, line 1164: Is there an argument for this to default to true?

GSterbrant: It can be a bit expensive, the dynamic refractions will also change peoples projects (for the better presumably) but might also incur a cost. 

But it's worth discussing if this could indeed be defaulted to true.

mvaligursky (approved):
Looks good to me, the suggestions were resolved, please wait for @slimbuck as well.

willeastcott (approved):

src/graphics/program-lib/chunks/chunk-validation.js
```diff
@@ -26,7 +27,9 @@ const chunkVersions = {
     lightSpecularBlinnPS: CHUNKAPI_1_55,
     lightSpecularPhongPS: CHUNKAPI_1_55,
     normalVertexPS: CHUNKAPI_1_55,
-    startPS: CHUNKAPI_1_55
+    startPS: CHUNKAPI_1_55,
+    refractionCubePS: CHUNKAPI_1_56,
+    refractionDynamicPS: CHUNKAPI_1_56
```
slimbuck, line 32: we don't have to add newly added chunks. only if we change their API will we add them here.

slimbuck (approved):

Issue #4473, opened 2022-07-25
Shadow acne artifacts in highlight engine demo


MAG-AdrianMeredith:
### Description

I'm seeing a lot of shadow acne artifacts on the highlight demo.  Can't recall what it looked like before (and therefore if theres a regression) but at the least it should probably be cleaned up.  Surely a simple scenario like this should work by default?

![image](https://user-images.githubusercontent.com/100127317/180729359-0fef277c-0d6f-43db-bb1c-5ac287e71af6.png)

mac book pro m1
chrome

slimbuck:
Thanks so much for reporting! We'll investigate. (My guess would be this is related to clustered lighting).

MAG-AdrianMeredith:
no worries unfortunately I can't recall what it looked like before but with the new default lighting switch, thats set off some alarm bells ringing

slimbuck:
I reproduced this on PROD. Adding `app.scene.clusteredLightingEnabled = false;` fixes the shadows.

slimbuck:
The clustered lighting shadows just appear lower precision than before. More shadowBias is required to remove acne on the example scene. I can't see or think of any obvious reason for this. Could be the difference of depth texture format (RGBA8 vs F32) or perhaps an additional rounding off error in clustered atlas depth pack/unpack.

Either way this will require @mvaligursky to investigate so handing over to him.

MAG-AdrianMeredith:
Should we log another issue for the underlying change?  I noticed this In my app too and have avoided shadows for this reason.  switching back to old lighting has none of those artifacts

mvaligursky:
I added a comment here: https://github.com/playcanvas/engine/issues/2780

PR #4491, opened 2022-08-01
[Fix] Increase the shadow resolution in model-outline example to avoid banding

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/4473

by default, a shadow resolution for non-clustered light is 1024, which is the resolution for the cubemap face in face of omni light. Clustered atlas res was increased to provide similar resolution, to avoid banding - as the offset constants are resolution dependent.

### DIFF
--- a/examples/src/examples/graphics/model-outline.tsx
+++ b/examples/src/examples/graphics/model-outline.tsx
@@ -23,6 +23,10 @@ class ModelOutlineExample {
 
             app.scene.ambientLight = new pc.Color(0.2, 0.2, 0.2);
 
+            // adjust default clustered lighting parameters to handle many lights
+            const lighting = app.scene.lighting;
+            lighting.shadowAtlasResolution = 4096;
+
             // helper function to create a primitive with shape type, position, scale, color and layer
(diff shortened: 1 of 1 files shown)


examples/src/examples/graphics/model-outline.tsx
```diff
@@ -23,6 +23,10 @@ class ModelOutlineExample {
 
             app.scene.ambientLight = new pc.Color(0.2, 0.2, 0.2);
 
+            // adjust default clustered lighting parameters to handle many lights
+            const lighting = app.scene.lighting;
+            lighting.shadowAtlasResolution = 4096;
```
slimbuck, line 28: all devices go to 4096?

mvaligursky: yep. internally we max it to max as well.

slimbuck: should we add this setting to the editor scene settings?

mvaligursky: yep, all clustered lighting settings will be exposed.

slimbuck (approved):

PR #4496, opened 2022-08-01
Added validation for the chunks with changed APIs

GSterbrant:
### Description

Make sure the users are notified when any of these chunks are used now with 1.55.


slimbuck (approved):
So few changes, I was expecting more!

GSterbrant:
The API changes were rather few :) And there are a bunch of new chunks that provide most of the changes.

PR #4513, opened 2022-08-05
Add HQ reflections

### LINKED ISSUE #4483: Implement HQ environment reflections
We render environment reflections using a single 2d texture atlas which contains shiny reflections, blurry reflections and ambient lighting (see #3810). This approach is resource friendly and works well for many applications.

However this approach limits the maximum sharpness possible on shiny reflections. Which is especially troublesome for our e-commerce users who require the highest possible quality on very shiny objects, like cars and jewellery.

In order to support these users, we must add an "HQ" reflections mode to standard material which will sample the original environment cubemap for sharp reflections instead of relying on the env atlas sharpness pyramid.

The resulting HQ shader will use one extra sampler (for the cubemap) and may run marginally slower, but should be a good tradeoff considering reflections can be arbitrarily detailed.

slimbuck:
Fixes #4483

This PR adds a high quality ("HQ") reflection chunk which uses the original environment cubemap for sharp reflections and the envAtlas for blurry reflections.

The HQ chunk requires an additional sampler compared to the existing solution, but otherwise performance should be comparable. 

Before:
before

After:
before

### DIFF
--- a/src/graphics/program-lib/chunks/chunks.js
+++ b/src/graphics/program-lib/chunks/chunks.js
@@ -139,6 +139,7 @@ import reflDirPS from './lit/frag/reflDir.js';
 import reflDirAnisoPS from './lit/frag/reflDirAniso.js';
 import reflectionCCPS from './lit/frag/reflectionCC.js';
 import reflectionCubePS from './lit/frag/reflectionCube.js';
+import reflectionEnvHQPS from './lit/frag/reflectionEnvHQ.js';
 import reflectionEnvPS from './lit/frag/reflectionEnv.js';
 import reflectionSpherePS from './lit/frag/reflectionSphere.js';
 import reflectionSphereLowPS from './lit/frag/reflectionSphereLow.js';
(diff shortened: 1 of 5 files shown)


willeastcott:
Just to confirm - this is just how the engine works now and you aren't expected to turn it on or toggle it etc?

willeastcott (approved):

mvaligursky:
There's the case of using faces for the skydome, but only atlas for lighting / reflections, that we cannot do without a switch I think?

slimbuck:
> There's the case of using faces for the skydome, but only atlas for lighting / reflections, that we cannot do without a switch I think?

Yes exactly.

willeastcott:
> We could add another flag to standard-material to opt-in, but I'd prefer we didn't.

I'm totally with you on that. 👍

mvaligursky (approved):

Maksims:
I'm wandering how it affects performance on XR devices such as Quest 1r mid-range mobile, as they do have difficulty on heavy shaders.

willeastcott:
@Maksims It's just one extra sampler so I seriously doubt it would make a noticeable difference. However, if you do encounter a specific performance problem, let us know and we'll take a closer look.

figo2264:
Thanks and this is what we really needed~!!!
Just wonder when this feature is available in editor~~!!!

slimbuck:
Hi @figo2264, already available with latest engine 👍🏻

PR #4517, opened 2022-08-05
Fixed issue (#3733) that prevents XR from working in certain circumstances

### LINKED ISSUE #3733: WebXR immersive VR launching issues on Desktop with Oculus AirLink
There's some oddness that I can't explain when launch PlayCanvas WebXR VR experiences on desktop via Oculus AirLink.

Forums thread: https://forum.playcanvas.com/t/recommended-setup-for-xr-development/22794

Setup:
Oculus Quest 1
Connected to Windows 10 with Oculus software using Air link
Using the Oculus OpenXR runtime
Chrome on Windows

PlayCanvas experiences only launches correctly if it's the only tab open that has a WebGL app.

If I have more than one tab (eg and Editor tab and Launch tab), the VR experience fails to load as shown below

https://user-images.githubusercontent.com/16639049/143031177-720c0cae-24cb-479c-a3b6-dbb2440810ef.mp4

However, the hello world from immersive web loads fine no matter the situation: https://immersive-web.github.io/webxr-samples/immersive-vr-session.html

Examples paths I've tried:

Two tabs open with https://immersive-web.github.io/webxr-samples/immersive-vr-session.html , starting VR in either works fine

One tab with https://immersive-web.github.io/webxr-samples/immersive-vr-session.html and another with https://developer.playcanvas.com/en/tutorials/webxr-controllerhand-models/ , starting VR in PlayCanvas fails while immersive-web works fine.

Two tabs open with https://developer.playcanvas.com/en/tutorials/webxr-controllerhand-models/ , neither succeeds entering into VR.

Clicky02:
Fixes #3733 

Fixed bug where having another WebGL instance open in the same browser stops the experience from launching properly.

The problem was solved by preventing playcanvas from rendering new frames when it cannot get the pose of the connected xr device.

It was tested with Oculus Quest 1 and 2 on both Airlink and the link cable.

This is my first PR, so if I need to make any changes let me know.

### DIFF
--- a/src/xr/xr-manager.js
+++ b/src/xr/xr-manager.js
@@ -668,7 +668,7 @@ class XrManager extends EventHandler {
 
     /**
      * @param {*} frame - XRFrame from requestAnimationFrame callback.
-     * 
+     *
      * @returns {boolean} True if update was successful, false otherwise.
      * @ignore
      */
--- a/src/xr/xr-manager.js
+++ b/src/xr/xr-manager.js
@@ -668,10 +668,12 @@ class XrManager extends EventHandler {
 
     /**
      * @param {*} frame - XRFrame from requestAnimationFrame callback.
+     * 
+     * @returns {boolean} True if update was successful, false otherwise.
      * @ignore
      */
     update(frame) {
-        if (!this._session) return;
+        if (!this._session) return false;
 
         // canvas resolution should be set on first frame availability or resolution changes
         const width = frame.session.renderState.baseLayer.framebufferWidth;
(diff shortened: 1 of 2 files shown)


willeastcott:
Hey @Maksims, any comments on this? Looks good to me...

src/xr/xr-manager.js
```diff
@@ -683,6 +685,9 @@ class XrManager extends EventHandler {
         }
 
         const pose = frame.getViewerPose(this._referenceSpace);
+
+        if (!pose) return false;
+
         const lengthNew = pose ? pose.views.length : 0;
```
mvaligursky, line 691: this line can be simplified now that pose is always valid. It could be
`const lengthNew = pose.views.length;`

mvaligursky (approved):
It looks great to me, I added one tiny improvement suggestion, but otherwise great job!

src/xr/xr-manager.js
mvaligursky, line 691:
```suggestion
        const lengthNew = pose.views.length;
```

mvaligursky: I've committed the suggested change.

PR #4527, opened 2022-08-09
Clear wasm module global

slimbuck:
Glb-parser currently fails to work with draco because it looks at window global for wasm module.

This PR changes WasmModule to clear the global module object on window.

### DIFF
--- a/src/core/wasm-module.js
+++ b/src/core/wasm-module.js
@@ -47,7 +47,13 @@ class Impl {
                 if (err) {
                     callback(err, null);
                 } else {
-                    window[moduleName]({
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

mvaligursky (approved):

yaustar:
Could this cause issues on multi engine instances apps if it clears a global?

slimbuck:
> Could this cause issues on multi engine instances apps if it clears a global?

Nope. I'd like to add support for multiple applications or wasm instances, but that won't be affected by this change.

PR #4528, opened 2022-08-10
Add support for material variants

### LINKED ISSUE #3647: Support glTF extension: KHR_materials_variants
https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_variants/README.md

This extension allows for a compact glTF representation of multiple material variants of an asset, structured to allow low-latency switching at runtime.

A typical use case is digital commerce, where a user might be presented with e.g. a pair of sneakers and the ability to switch between different colours.

image

GSterbrant:
### Description
Adds support for KHR_materials_variants as well as provides an API for changing the material. Also fixes the default material index ownership, having them live in the container instead of on the mesh. 

image

image

Fixes #3647

### API changes
- Adds getMaterialVariants to GlbContainerResource. This API fetches the list of variants present in the GLB.
- Adds applyMaterialVariant(name, entity) to GlbContainerResource. This API applies a material variant to all mesh instances in the render component of an entity. 
- Adds applyMaterialVariantInstances(name, instances) to GlbContainerResource. This API applies a material variant to a set of mesh instances, using the GlbContainerResource as the source of the variants, their mappings, and the materials.

### DIFF
--- a/src/resources/container.js
+++ b/src/resources/container.js
@@ -90,11 +90,12 @@ class ContainerResource {
 
     /**
      * Applies a material variant to a set of mesh instances. Compared to the applyMaterialVariant,
-     * this method allows for setting the variant on a specific set of mesh instances instead of the 
+     * this method allows for setting the variant on a specific set of mesh instances instead of the
      * whole entity.
      *
-     * @param {string} name - Is the name of the variant, as quered from getMaterialVariants
+     * @param {string} name - The the name of the variant, as quered from getMaterialVariants
      * @param {MeshInstance[]} instances - An array of mesh instances
+     * @example
      * // load a glb file and instantiate an entity with a render component based on it
      * app.assets.loadFromUrl("statue.glb", "container", function (err, asset) {
      *     var entity = asset.resource.instantiateRenderEntity({
@@ -109,6 +110,20 @@ class ContainerResource {
      *     }
      */
     applyMaterialVariantInstances(name, instances) {}
+
+    /**
+     * Reset variant selection on entity hierarchy to use the default material
+     *
+     * @param {Entity} entity - The entity to reset variants on
+     */
+    resetMaterialVariant(entity) {}
+
+    /**
+     * Reset variant selection on set of mesh instances to use the default material
+     *
+     * @param {MeshInstance[]} instances - An array of mesh instances to reset variants on
+     */
+    resetMaterialVariantInstances(instances) {}
 }
 
 /**
--- a/src/resources/container.js
+++ b/src/resources/container.js
@@ -60,6 +60,24 @@ class ContainerResource {
     instantiateRenderEntity(options) {
         return null;
     }
+
+    /**
+     * Queries the list of available material variants.
+     *
+     * @returns {object} A map which maps a variant name as a string to the variant index
+     * @example
+     * // load a glb file and instantiate an entity with a render component based on it
+     * app.assets.loadFromUrl("statue.glb", "container", function (err, asset) {
+     *     var entity = asset.resource.instantiateRenderEntity({
+     *         castShadows: true
+     *     });
+     *     app.root.addChild(entity);
+     *     let materialVariants = Object.keys(asset.resource.getMaterialVariants());
+     *     const variantListOptions: Array<{ v:string, t:string }> = materialVariants.map((variant: string) => ({ v: variant, t: variant }));
+     */
+    getMaterialVariants() {
+        return null;
+    }
 }
 
 /**
--- a/src/resources/container.js
+++ b/src/resources/container.js
@@ -92,7 +92,7 @@ class ContainerResource {
      *         castShadows: true
      *     });
      *     app.root.addChild(entity);
-     *     let materialVariants = Object.keys(asset.resource.getMaterialVariants());
+     *     var materialVariants = Object.keys(asset.resource.getMaterialVariants());
      *     asset.resource.applyMaterialVariant(materialVariants[0], entity);
      */
     applyMaterialVariant(name, entity) {}
@@ -109,9 +109,9 @@ class ContainerResource {
      *         castShadows: true
      *     });
      *     app.root.addChild(entity);
-     *     let materialVariants = Object.keys(asset.resource.getMaterialVariants());
+     *     var materialVariants = Object.keys(asset.resource.getMaterialVariants());
      *     const renders = entity.findComponents("render");
-     *     for (let i = 0; i < renders.length; i++) {
+     *     for (var i = 0; i < renders.length; i++) {
      *         const renderComponent = renders[i];
      *         asset.resource.applyMaterialVariantInstances(materialVariants[0], renderComponent.meshInstances);
      *     }
--- a/src/resources/container.js
+++ b/src/resources/container.js
@@ -2,6 +2,7 @@ import { path } from '../core/path.js';
 import { GlbParser } from './parser/glb-parser.js';
 
 /** @typedef {import('../framework/entity.js').Entity} Entity */
+/** @typedef {import('../scene/mesh-instance').MeshInstance} MeshInstance */
 /** @typedef {import('../framework/app-base.js').AppBase} AppBase */
 /** @typedef {import('./handler.js').ResourceHandler} ResourceHandler */
 /** @typedef {import('./handler.js').ResourceHandlerCallback} ResourceHandlerCallback */
@@ -78,6 +79,44 @@ class ContainerResource {
     getMaterialVariants() {
         return null;
     }
+
+    /**
+     * Applies a material variant to an entity
+     *
+     * @param {string} name - Is the name of the variant, as quered from getMaterialVariants
+     * @param {Entity} entity - Is the entity to which material variants will be applied
+     * @example
+     * // load a glb file and instantiate an entity with a render component based on it
+     * app.assets.loadFromUrl("statue.glb", "container", function (err, asset) {
+     *     var entity = asset.resource.instantiateRenderEntity({
+     *         castShadows: true
+     *     });
+     *     app.root.addChild(entity);
+     *     let materialVariants = Object.keys(asset.resource.getMaterialVariants());
+     *     asset.resource.applyMaterialVariant(materialVariants[0], entity);
+     */
+    applyMaterialVariant(name, entity) {}
+
+    /**
+     * Applies a material variant to a set of mesh instances. This method allows for a selective
+     * set of MeshInstances to apply a variant, instead of the entire entity
+     *
+     * @param {string} name - Is the name of the variant, as quered from getMaterialVariants
+     * @param {MeshInstance} [instances] - An array of mesh instances
+     * // load a glb file and instantiate an entity with a render component based on it
+     * app.assets.loadFromUrl("statue.glb", "container", function (err, asset) {
+     *     var entity = asset.resource.instantiateRenderEntity({
+     *         castShadows: true
+     *     });
+     *     app.root.addChild(entity);
+     *     let materialVariants = Object.keys(asset.resource.getMaterialVariants());
+     *     const renders = entity.findComponents("render");
+     *     for (let i = 0; i < renders.length; i++) {
+     *         const renderComponent = renders[i];
+     *         asset.resource.applyMaterialVariantInstances(materialVariants[0], renderComponent.meshInstances);
+     *     }
+     */
+    applyMaterialVariantInstances(name, instances) {}
 }
 
 /**
--- a/src/resources/container.js
+++ b/src/resources/container.js
@@ -74,8 +74,9 @@ class ContainerResource {
     /**
      * Applies a material variant to an entity hierarchy.
      *
-     * @param {string} name - The name of the variant, as quered from getMaterialVariants.
      * @param {Entity} entity - The entity root to which material variants will be applied
+     * @param {string} [name] - The name of the variant, as quered from getMaterialVariants,
+     * if null the variant will be reset to the default
      * @example
      * // load a glb file and instantiate an entity with a render component based on it
      * app.assets.loadFromUrl("statue.glb", "container", function (err, asset) {
@@ -86,15 +87,16 @@ class ContainerResource {
      *     var materialVariants = asset.resource.getMaterialVariants();
      *     asset.resource.applyMaterialVariant(materialVariants[0], entity);
      */
-    applyMaterialVariant(name, entity) {}
+    applyMaterialVariant(entity, name) {}
 
     /**
      * Applies a material variant to a set of mesh instances. Compared to the applyMaterialVariant,
      * this method allows for setting the variant on a specific set of mesh instances instead of the
      * whole entity.
      *
-     * @param {string} name - The the name of the variant, as quered from getMaterialVariants
      * @param {MeshInstance[]} instances - An array of mesh instances
+     * @param {string} [name] - The the name of the variant, as quered from getMaterialVariants,
+     * if null the variant will be reset to the default
      * @example
      * // load a glb file and instantiate an entity with a render component based on it
      * app.assets.loadFromUrl("statue.glb", "container", function (err, asset) {
@@ -109,21 +111,7 @@ class ContainerResource {
      *         asset.resource.applyMaterialVariantInstances(materialVariants[0], renderComponent.meshInstances);
      *     }
      */
-    applyMaterialVariantInstances(name, instances) {}
-
-    /**
-     * Reset variant selection on entity hierarchy to use the default material
-     *
-     * @param {Entity} entity - The entity to reset variants on
-     */
-    resetMaterialVariant(entity) {}
-
-    /**
-     * Reset variant selection on set of mesh instances to use the default material
-     *
-     * @param {MeshInstance[]} instances - An array of mesh instances to reset variants on
-     */
-    resetMaterialVariantInstances(instances) {}
+    applyMaterialVariantInstances(instances, name) {}
 }
 
 /**
--- a/src/resources/container.js
+++ b/src/resources/container.js
@@ -65,7 +65,7 @@ class ContainerResource {
     /**
      * Queries the list of available material variants.
      *
-     * @returns {object} A map which maps a variant name as a string to the variant index
+     * @returns {string[]} An array of variant names.
      * @example
      * // load a glb file and instantiate an entity with a render component based on it
      * app.assets.loadFromUrl("statue.glb", "container", function (err, asset) {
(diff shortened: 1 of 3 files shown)


mvaligursky:
Could you please add an overview of the new public API to the PR description.

src/resources/parser/glb-container-resource.js
```diff
@@ -190,6 +191,22 @@ class GlbContainerResource {
         return GlbContainerResource.createSceneHierarchy(sceneClones, 'Entity');
     }
 
+    // get material variants
+    getMaterialVariants() {
```
mvaligursky, line 195: if this is a public API, it needs a JSDocs documentation

src/resources/parser/glb-container-resource.js
```diff
@@ -190,6 +191,22 @@ class GlbContainerResource {
...
+        return this.data.variants;
+    }
+
+    // apply material variant to mesh instances
+    applyMaterialVariant(name, instances) {
```
mvaligursky, line 200: if this is a public API, it needs a JSDocs documentation

src/resources/parser/glb-container-resource.js
```diff
@@ -190,6 +191,22 @@ class GlbContainerResource {
...
+        const variant = this.data.variants[name];
```
mvaligursky, line 201: handle the case when the variant is not found

src/resources/parser/glb-parser.js
```diff
@@ -753,14 +756,15 @@ const createSkin = function (device, gltfSkin, accessors, bufferViews, nodes, gl
...
-const createMesh = function (device, gltfMesh, accessors, bufferViews, callback, flipV, vertexBufferDict) {
+const createMesh = function (device, gltfMesh, accessors, bufferViews, callback, flipV, vertexBufferDict, meshVariants, meshDefaultMaterials) {
     const meshes = [];
 
     gltfMesh.primitives.forEach(function (primitive) {
 
         let primitiveType, vertexBuffer, numIndices;
         let indices = null;
         let canUseMorph = true;
+        meshVariants.push({});
```
slimbuck, line 767: is this necessary?

src/resources/parser/glb-container-resource.js
slimbuck, line 195: do we need to expose this? how do you see it being used?

GSterbrant: On the user side, we need to provide some API to query the variants that exist in the GLB.

src/resources/parser/glb-parser.js
GSterbrant: Deleted

src/resources/parser/glb-container-resource.js
GSterbrant: Done

src/resources/parser/glb-container-resource.js
GSterbrant: Done

src/resources/parser/glb-container-resource.js
GSterbrant: Done

src/resources/parser/glb-parser.js
```diff
@@ -1768,6 +1780,17 @@ const createMaterials = function (gltf, textures, options, flipV) {
     });
 };
 
+const createVariants = function (gltf) {
+    if (!gltf.hasOwnProperty("extensions") || !gltf.extensions.hasOwnProperty("KHR_materials_variants"))
+        return null;
+
+    const variants = {};
+    for (let i = 0; i < gltf.extensions.KHR_materials_variants.variants.length; i++) {
```
mvaligursky, line 1788: store `gltf.extensions.KHR_materials_variants.variants` in some local variable to avoid all these lookups.

src/resources/parser/glb-parser.js
```diff
@@ -887,9 +890,18 @@ const createMesh = function (device, gltfMesh, accessors, bufferViews, callback,
...
-            mesh.materialIndex = primitive.material;
+            if (primitive.hasOwnProperty("extensions") && primitive.extensions.hasOwnProperty("KHR_materials_variants")) {
+                const variants = primitive.extensions.KHR_materials_variants;
+                const tempMapping = {};
+                variants.mappings.forEach((mapping) => {
+                    mapping.variants.forEach((variant) => {
+                        tempMapping[variant] = mapping.material;
+                    });
+                });
+                meshVariants[mesh.id] = tempMapping;
```
mvaligursky, line 901: I'd suggest meshVariants and meshDefaultMaterials to use object (with a key lookup), or a Map, instead of being an array. If the user has lots of meshes, the mesh.id will be high and allocate large arrays for no reason.

src/resources/parser/glb-container-resource.js
```diff
@@ -190,6 +191,41 @@ class GlbContainerResource {
...
 
+    // get material variants
+    getMaterialVariants() {
+        return this.data.variants;
+    }
+
+    // apply material variant to entity
+    applyMaterialVariant(name, entity) {
+        const variant = this.data.variants[name];
+        if (!variant)
```
mvaligursky, line 202: add Debug.warn to print a message that variant does not exist.

src/resources/parser/glb-container-resource.js
```diff
@@ -190,6 +191,41 @@ class GlbContainerResource {
...
+            return;
+        const renders = entity.findComponents("render");
+        for (let i = 0; i < renders.length; i++) {
+            const renderComponent = renders[i];
+            renderComponent.meshInstances.forEach((instance) => {
+                const meshVariants = this.data.meshVariants[instance.mesh.id];
+                if (meshVariants) {
+                    instance.material = this.data.materials[meshVariants[variant]];
```
mvaligursky, line 210: I'd just add line like this after this one, to make sure this holds true:
`Debug.Assert(instance.material);`

mvaligursky: also, could this function call applyMaterialVariantInstances to avoid duplication of the same code?

src/resources/container.js
```diff
@@ -60,6 +61,62 @@ class ContainerResource {
     instantiateRenderEntity(options) {
         return null;
     }
+
+    /**
+     * Queries the list of available material variants.
+     *
+     * @returns {object} A map which maps a variant name as a string to the variant index
```
mvaligursky, line 68: should the public interface simply return the array of strings? Or is the mapping from string index somehow useful to the user?

src/resources/container.js
```diff
@@ -60,6 +61,62 @@ class ContainerResource {
...
+     * @example
+     * // load a glb file and instantiate an entity with a render component based on it
+     * app.assets.loadFromUrl("statue.glb", "container", function (err, asset) {
+     *     var entity = asset.resource.instantiateRenderEntity({
+     *         castShadows: true
+     *     });
+     *     app.root.addChild(entity);
+     *     let materialVariants = Object.keys(asset.resource.getMaterialVariants());
```
mvaligursky, line 76: in the examples we so far stick to ES5, so use vars, but no let / const. Also, this should be JS and not TS, so avoid the types.

src/resources/container.js
```diff
@@ -60,6 +61,62 @@ class ContainerResource {
...
+     *     const variantListOptions: Array<{ v:string, t:string }> = materialVariants.map((variant: string) => ({ v: variant, t: variant }));
+     */
+    getMaterialVariants() {
+        return null;
+    }
+
+    /**
+     * Applies a material variant to an entity
```
mvaligursky, line 84:
```suggestion
     * Applies a material variant to an entity hierarchy.
```

src/resources/container.js
```diff
@@ -60,6 +61,62 @@ class ContainerResource {
...
+     *
+     * @param {string} name - Is the name of the variant, as quered from getMaterialVariants
```
mvaligursky, line 86:
```suggestion
     * @param {string} name - The name of the variant, as queried from getMaterialVariants.
```

src/resources/container.js
```diff
@@ -60,6 +61,62 @@ class ContainerResource {
...
+     * @param {Entity} entity - Is the entity to which material variants will be applied
```
mvaligursky, line 87:
```suggestion
     * @param {Entity} entity - The entity root to which material variants will be applied
```

src/resources/container.js
```diff
@@ -60,6 +61,62 @@ class ContainerResource {
...
+     *     asset.resource.applyMaterialVariant(materialVariants[0], entity);
+     */
+    applyMaterialVariant(name, entity) {}
+
+    /**
+     * Applies a material variant to a set of mesh instances. This method allows for a selective
+     * set of MeshInstances to apply a variant, instead of the entire entity
+     *
+     * @param {string} name - Is the name of the variant, as quered from getMaterialVariants
+     * @param {MeshInstance} [instances] - An array of mesh instances
```
mvaligursky, line 105:
```suggestion
     * @param {MeshInstance[]} instances - An array of mesh instances
```

src/resources/parser/glb-parser.js
GSterbrant: Does it become an array? I am assigning a member to a value, doesn't that just add a member to an object?

mvaligursky: you initialise them to be arrays in the code below I think?
```
    const meshVariants = [];
    const meshDefaultMaterials = [];
```
you could change those to be objects { }, but I think those are being passed to some functions that expect arrays and look them up by index, so it would not work I think.

src/resources/container.js
GSterbrant: Yeah, I think it's actually pointless to use the index, so this could just as well return strings.

src/resources/container.js
```diff
@@ -60,6 +61,62 @@ class ContainerResource {
...
+     * @example
+     * // load a glb file and instantiate an entity with a render component based on it
+     * app.assets.loadFromUrl("statue.glb", "container", function (err, asset) {
+     *     var entity = asset.resource.instantiateRenderEntity({
+     *         castShadows: true
+     *     });
+     *     app.root.addChild(entity);
+     *     var materialVariants = asset.resource.getMaterialVariants();
+     *     var variantListOptions = materialVariants.map((variant) => ({ v: variant, t: variant }));
+     */
```
mvaligursky, line 78: I'm not sure this example is needed, as it's essentially the same as for the other function.

src/resources/container.js
```diff
@@ -60,6 +61,62 @@ class ContainerResource {
...
+     * @param {string} name - The name of the variant, as quered from getMaterialVariants
+     * @param {Entity} entity - The entity root to which material variants will be applied
+     * @example
+     * // load a glb file and instantiate an entity with a render component based on it
+     * app.assets.loadFromUrl("statue.glb", "container", function (err, asset) {
+     *     var entity = asset.resource.instantiateRenderEntity({
+     *         castShadows: true
+     *     });
+     *     app.root.addChild(entity);
+     *     var materialVariants = Object.keys(asset.resource.getMaterialVariants());
```
mvaligursky, line 95: this is out of date - variants is an array directly, no need for Object.keys

src/resources/container.js
```diff
@@ -60,6 +61,62 @@ class ContainerResource {
...
+     * @param {string} name - Is the name of the variant, as quered from getMaterialVariants
+     * @param {MeshInstance[]} instances - An array of mesh instances
+     * // load a glb file and instantiate an entity with a render component based on it
+     * app.assets.loadFromUrl("statue.glb", "container", function (err, asset) {
+     *     var entity = asset.resource.instantiateRenderEntity({
+     *         castShadows: true
+     *     });
+     *     app.root.addChild(entity);
+     *     var materialVariants = Object.keys(asset.resource.getMaterialVariants());
+     *     const renders = entity.findComponents("render");
```
mvaligursky, line 113: here as well

src/resources/container.js
```diff
@@ -60,6 +61,62 @@ class ContainerResource {
...
+     *     var variantListOptions = materialVariants.map((variant) => ({ v: variant, t: variant }));
+     */
+    getMaterialVariants() {
+        return null;
+    }
+
+    /**
+     * Applies a material variant to an entity hierarchy.
+     *
+     * @param {string} name - The name of the variant, as quered from getMaterialVariants
```
mvaligursky, line 86: all these lines need a period '.' at the end.

src/resources/container.js
```diff
@@ -60,6 +61,62 @@ class ContainerResource {
...
+     *     });
+     *     app.root.addChild(entity);
+     *     var materialVariants = Object.keys(asset.resource.getMaterialVariants());
+     *     asset.resource.applyMaterialVariant(materialVariants[0], entity);
+     */
+    applyMaterialVariant(name, entity) {}
+
+    /**
+     * Applies a material variant to a set of mesh instances. This method allows for a selective
+     * set of MeshInstances to apply a variant, instead of the entire entity
```
mvaligursky, line 102: this could be more clear I think : `MeshInstances to apply a variant`

src/resources/container.js
```diff
@@ -60,6 +61,62 @@ class ContainerResource {
...
+     *
+     * @param {string} name - Is the name of the variant, as quered from getMaterialVariants
+     * @param {MeshInstance[]} instances - An array of mesh instances
+     * // load a glb file and instantiate an entity with a render component based on it
+     * app.assets.loadFromUrl("statue.glb", "container", function (err, asset) {
+     *     var entity = asset.resource.instantiateRenderEntity({
+     *         castShadows: true
+     *     });
+     *     app.root.addChild(entity);
+     *     var materialVariants = Object.keys(asset.resource.getMaterialVariants());
```
mvaligursky, line 112: out of date

src/resources/parser/glb-container-resource.js
```diff
@@ -190,6 +192,45 @@ class GlbContainerResource {
...
 
+    // get material variants
+    getMaterialVariants() {
+        return Object.keys(this.data.variants);
+    }
+
+    // apply material variant to entity
+    applyMaterialVariant(name, entity) {
+        const variant = this.data.variants[name];
+        if (variant === undefined) {
```
mvaligursky, line 203: it could be just `if (!variant)` as you had before. It's a string, so that should be ok. It's only a problem if it is a number and can be 0. Same below.

src/resources/parser/glb-container-resource.js
```diff
@@ -190,6 +192,45 @@ class GlbContainerResource {
...
+        }
+        this._applyMaterialVariant(variant, instances);
+    }
+
+    // internally apply variant to instances
+    _applyMaterialVariant(variant, instances) {
+        instances.forEach((instance) => {
+            const meshVariants = this.data.meshVariants[instance.mesh.id];
+            if (meshVariants) {
+                instance.material = this.data.materials[meshVariants[variant]];
```
mvaligursky, line 229: please add `Debug.assert(instance.material);` just after this line, to validate it's not null, as those are errors that are harder to find.

mvaligursky:
I just wonder .. are default mesh materials usually listed as some variant? If not, do we need some resetVariants(entity) function to disable variants and switch the mesh instances to the default materials? Of maybe if applyMaterialVariant is called with null / invalid name, it should apply the default materials?

src/resources/parser/glb-container-resource.js
GSterbrant: It can be 0 :)

GSterbrant:
Yeah, I added an API and documentation to reset variants to the default material.

GSterbrant:
And reverted the reset from the public API after some discussion.

src/resources/container.js
```diff
@@ -60,6 +61,57 @@ class ContainerResource {
...
+     */
+    getMaterialVariants() {
+        return null;
+    }
+
+    /**
+     * Applies a material variant to an entity hierarchy.
+     *
+     * @param {Entity} entity - The entity root to which material variants will be applied
+     * @param {string} [name] - The name of the variant, as quered from getMaterialVariants,
```
mvaligursky, line 78: quered = queried?

src/resources/container.js
```diff
@@ -60,6 +61,57 @@ class ContainerResource {
...
+     * // load a glb file and instantiate an entity with a render component based on it
+     * app.assets.loadFromUrl("statue.glb", "container", function (err, asset) {
+     *     var entity = asset.resource.instantiateRenderEntity({
+     *         castShadows: true
+     *     });
+     *     app.root.addChild(entity);
+     *     var materialVariants = asset.resource.getMaterialVariants();
+     *     const renders = entity.findComponents("render");
+     *     for (var i = 0; i < renders.length; i++) {
+     *         const renderComponent = renders[i];
```
mvaligursky, line 110:
```suggestion
     *         var renderComponent = renders[i];
```

src/resources/container.js
```diff
@@ -60,6 +61,57 @@ class ContainerResource {
...
+     *         asset.resource.applyMaterialVariantInstances(materialVariants[0], renderComponent.meshInstances);
```
mvaligursky, line 111: fix the order of parameters

src/resources/parser/glb-container-resource.js
mvaligursky: ah right, this is the index!

GSterbrant: Yep!

src/resources/parser/glb-container-resource.js
```diff
@@ -190,6 +192,70 @@ class GlbContainerResource {
...
+        }
+        const renders = entity.findComponents("render");
+        for (let i = 0; i < renders.length; i++) {
+            const renderComponent = renders[i];
+            this._applyMaterialVariant(variant, renderComponent.meshInstances);
+        }
+    }
+
+    // reset material variants on entity
+    _resetMaterialVariant(entity) {
```
mvaligursky, line 219: I think I'd get rid of these _reset functions, as its the same code as the function above apart from the material selection - just make a small change to that function to handle it and remove the code duplication?

src/resources/parser/glb-container-resource.js
```diff
@@ -190,6 +192,50 @@ class GlbContainerResource {
...
+    // internally apply variant to instances
+    _applyMaterialVariant(variant, instances) {
+        instances.forEach((instance) => {
+            if (variant === null) {
+                instance.material = this._defaultMaterial;
+            } else {
+                const meshVariants = this.data.meshVariants[instance.mesh.id];
+                if (meshVariants) {
+                    instance.material = this.data.materials[meshVariants[variant]];
+                    Debug.assert(instance.material);
```
mvaligursky, line 233: if you could move the assert outside of if/else, so that it validates both cases.

mvaligursky (approved):
approved with a single small comment.

PR #4532, opened 2022-08-11
Morph target animation fix

### LINKED ISSUE #4520: Morph target animation issues in 1.55
There's an issue related to morph target animating using names instead of indices, as in 1.54

more details: https://forum.playcanvas.com/t/assert-failed-cannot-find-morph-target-with-name-0/26952

### LINKED ISSUE #4522: Shape key / morph key bug
After the v1.55.3 update basic animations using shape keys / morph keys break, producing this runtime error:

**ASSERT FAILED:
Cannot find morph target with name: 0**

I have reproduced the error in this open project: [Example project](https://playcanvas.com/project/967373/overview/shape-key-bug)
The icososphere model ('Test_shapeKeys') has been imported from Blender as an FBX file and automatically converted to GLB in Playcanvas. The animation is all in one file and played on loop using an animation state graph. The model is supposed to deform right after doing the jump. Instead it produces multiple errors as decribed above. Both the jump and deform are driven by bones, but the former by weights and the latter by shape keys.

When deformed the icososphere is supposed to look like this:
ShapeKeyBug

While in Playcanvas looks like this:
image

Reverting to the previous version removes the error.

ellthompson:
Morph target animation curves that are named via the glb mesh's extras property should be differentiated from index based curves.

Fixes #4522
Fixes #4520

### DIFF
--- a/src/anim/binder/default-anim-binder.js
+++ b/src/anim/binder/default-anim-binder.js
@@ -79,6 +79,11 @@ class DefaultAnimBinder {
             },
 
             'weight': function (node, weightName) {
+                if (weightName.indexOf('name.') === 0) {
+                    weightName = weightName.replace('name.', '');
+                } else {
+                    weightName = Number(weightName);
+                }
                 const meshInstances = findMeshInstances(node);
                 if (meshInstances) {
                     for (let i = 0; i < meshInstances.length; ++i) {
--- a/src/resources/parser/glb-parser.js
+++ b/src/resources/parser/glb-parser.js
@@ -1463,7 +1463,7 @@ const createAnimation = function (gltfAnimation, animationIndex, gltfAccessors, 
         for (let i = 0; i < meshes.length; i++) {
             const mesh = meshes[i];
             if (mesh.name === nodeName && mesh.hasOwnProperty('extras') && mesh.extras.hasOwnProperty('targetNames') && mesh.extras.targetNames[weightIndex]) {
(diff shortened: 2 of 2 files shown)


mvaligursky:
Could you please explain what this does? you add the prefix, and later strip it.

src/anim/binder/default-anim-binder.js
```diff
@@ -79,6 +79,11 @@ class DefaultAnimBinder {
             },
 
             'weight': function (node, weightName) {
+                if (weightName.indexOf('name.') === 0) {
+                    weightName = weightName.replace('name.', '');
+                } else {
+                    weightName = Number(weightName);
```
slimbuck, line 85: worth checking for NaN here?

slimbuck (approved):

slimbuck:
(approved, but pls wait for @mvaligursky to also approve before merging)

ellthompson:
> Could you please explain what this does? you add the prefix, and later strip it.

The prefix is only added to morph target curves which are named via the mesh's extras property. This means we can differentiate them at binding time from indexed morph target curves, passing them to the setWeight function as a string instead of a number. Then those named curves will be referenced against the name lookup table, indexed curves will behave as before.

src/anim/binder/default-anim-binder.js
ellthompson: This will always be the index of the morph target as a string.

mvaligursky (approved):

PR #4544, opened 2022-08-16
Fix biasing issues with WGL1

### LINKED ISSUE #4540: Shadow artefacts with clustered spot lights running with WGL1
### Description
Shadows for spot lights appear broken at cone centre. 

image

GSterbrant:
Normal bias is not angle dependent anymore. Slope scale bias is applied to all light sources when rendering the shadow map if running WGL1.

Fixes #4540


mvaligursky (approved):
looks good

PR #4545, opened 2022-08-16
Animation resource with events fix

### LINKED ISSUE #4542: AnimEvents aren't created when animations are loaded via anim.assignAnimation
Thread: https://forum.playcanvas.com/t/seeking-with-animcomponent/27060

Project: https://playcanvas.com/project/970251/overview/rewinding-animation-events

Animation event is added on the Idle animation and the animation is added via anim.assignAnimation
An event listener is added to script to console out some text but it is never called

Looking at the engine code, AnimEvents is only created on onAnimationAssetLoaded so no events are created here.

Workaround is to add the events manually before calling anim.assignAnimation

ellthompson:
Previously events from an animation asset were assigned to each AnimTrack when an anim state graph is loaded. However it's now possible to load AnimTracks without creating a state graph (using the assignAnimation API). This meant animations loaded this way would not have any events from their asset assigned to them. This PR moves the assignment of events to the animation resource handler instead. Ensuring that events are always added to AnimTracks that are created via animation assets.

Fixes #4542 

### DIFF
--- a/src/framework/components/anim/component.js
+++ b/src/framework/components/anim/component.js
@@ -12,7 +12,6 @@ import {
 import { AnimComponentBinder } from './component-binder.js';
 import { AnimComponentLayer } from './component-layer.js';
 import { AnimStateGraph } from '../../../anim/state-graph/anim-state-graph.js';
-import { AnimEvents } from '../../../anim/evaluator/anim-events.js';
 import { Entity } from '../../entity.js';
 
 /** @typedef {import('./system.js').AnimComponentSystem} AnimComponentSystem */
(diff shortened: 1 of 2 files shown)


mvaligursky (approved):

yaustar:
Does this mean events only work on GLB assets?

ellthompson:
> Does this mean events only work on GLB assets? Did it ever work with JSON or Base64 URLs?

Events are only supported by the anim component (glb only).

yaustar:
> Events are only supported by the anim component (glb only).

An animation asset could be created with a Base64 URL that is a GLB type but not have the extension of .GLB in the path. Would that mean that events won't work in this case?

yaustar:
Checked with engine 1.55.4 using Base64 URLs. Animation and Anim Events work fine with the Anim Component and 'GLB' asset that as converted to Base64

PR #4548, opened 2022-08-17
AnimBinder path resolution update

ellthompson:
This PR updates the AnimBinder to support two types of animation curve paths.
The first is currently supported in the engine and looks something like this:
```
RootBone/RootNode/Hips/etc/etc
```
This covers the case where the entity that the anim component is attached to (RootBone), is the parent of the GLB hierarchy to be animated. The GLB hierarchy in this instance starts at RootNode.

The second type of path which is supported in this PR looks like:
```
RootBone/Hips/etc/etc
```
This will occur when the entity that the anim component is attached to (RootBone) is also the root of the GLB hierachy which is being animated. FBX's loaded using the editor's import hierarchy feature will have this structure. The GLB hierarchy in this instance starts at RootBone.

Fixes: https://github.com/playcanvas/editor/issues/602

### DIFF
--- a/src/anim/binder/default-anim-binder.js
+++ b/src/anim/binder/default-anim-binder.js
@@ -156,6 +156,10 @@ class DefaultAnimBinder {
         let node;
         if (this.graph) {
             node = this.graph.findByPath(path.entityPath);
+            // if the path is not found under the given root node, try to find the path using the root node as the base of the path instead
+            if (!node) {
+                node = this.graph.findByPath(path.entityPath.slice(1));
+            }
         }
         if (!node) {
             node = this.nodes[path.entityPath[path.entityPath.length - 1] || ""];


mvaligursky (approved):
glad it has such an easy solution!

src/anim/binder/default-anim-binder.js
```diff
@@ -156,6 +156,10 @@ class DefaultAnimBinder {
         let node;
         if (this.graph) {
             node = this.graph.findByPath(path.entityPath);
+            // if the path is not found under the given root node, try to find the path using the root node as the base of the path instead
+            if (!node) {
+                node = this.graph.findByPath(path.entityPath.slice(1));
```
slimbuck, line 161: shouldn't this be `slice(1, 1)` ?

slimbuck: oh, no, ignore me :)

slimbuck: from the PR description it looks like it's the second child that needs removing, not the root?

ellthompson: path.entityPath is an array of strings (each string being a node in the path). We want to search using index 1 to the end of the path.

PR #4549, opened 2022-08-17
Return null if no variants are present

GSterbrant:
### Description
Fixes error where Object.keys would take a null/undefined argument.

### DIFF
--- a/src/resources/parser/glb-container-resource.js
+++ b/src/resources/parser/glb-container-resource.js
@@ -194,7 +194,7 @@ class GlbContainerResource {
 
     // get material variants
     getMaterialVariants() {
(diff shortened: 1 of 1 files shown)


src/resources/parser/glb-container-resource.js
```diff
@@ -194,7 +194,7 @@ class GlbContainerResource {
 
     // get material variants
     getMaterialVariants() {
-        return Object.keys(this.data.variants);
+        return this.data.variants ? Object.keys(this.data.variants) : null;
```
slimbuck, line 197: perhaps return an empty array instead?

slimbuck (approved):

Issue #4341, opened 2022-06-17
Using multiple touches on screen with scrollbar causes the bar to move erratically 


yaustar:
Open https://launch.playcanvas.com/1449404?debug=true on a touch device
Start scrolling with one finger
Use another finger to touch and move on the screen
Notice the scroll bar jumps everywhere

https://user-images.githubusercontent.com/16639049/174276147-7423b9f0-6b6a-4329-8e55-1a4ed9ed9ed2.mp4

This is because pc.ElementDragHelper doesn't track the touch id that was used to start the drag for the element and uses the most recent changedTouches

I suspect this also affects anything we use pc.ElementDragHelper for such as scroll view

Related thread: https://forum.playcanvas.com/t/multiple-joysticks-issue/26209

Issue #4470, opened 2022-07-23
Click event never fires in multi-touch scenarios


LeXXik:
Please, evaluate if it is a bug.

Currently, the `click` event never fires on an element input, if there is already another finger touching the screen.

https://github.com/playcanvas/engine/blob/3a9e881f2dff74923076ec4497929c11fffe2c15/src/input/element-input.js#L648-L664

The `event.touches.length` will be more than zero, skipping the block.

Repro project: https://playcanvas.com/project/961849/overview
Touch the screen somewhere first, then while holding it, use another finger to tap the button.

slimbuck:
@jpauloruschel are you happy to take this one?

yaustar:
The original design is to mirror what the browser does for raw inputs. So if the browser doesn't do a click event in the same scenario, then by original design, it isn't a bug

LeXXik:
I vote for it to be considered a bug. In many games we often use two fingers simultaneously for controlling the gameplay. If the browsers do it differently, I suppose, it is because it is not common to use 2 fingers to read a web page.

yaustar:
Potentially related issue https://github.com/playcanvas/engine/issues/4341

PR #4562, opened 2022-08-18
Fix UI Element input events on multiple touches.

jpauloruschel:
Fixes https://github.com/playcanvas/engine/issues/4341 and https://github.com/playcanvas/engine/issues/4470

Issue https://github.com/playcanvas/engine/issues/4341's root cause is that the `element-drag-helper` class (used by `Scrollview` components and similar) uses Window DOM Events for handling dragging, and does not check the actual target of the event. This leads to *any* touch interaction being handled by *all* UI Elements at once. Issue https://github.com/playcanvas/engine/issues/4470 is also related to multitouch.

This PR makes the `element-drag-helper` use the Events triggered by `element-input` system (rather than Window DOM events) for handling dragging. The `element-input` system already contains all logic and checks necessary to figure out which is the correct element a given user events (like touch or mouse) should target.

The only reason `element-drag-helper` was still using Window DOM events was to handle the cases of outside-of-canvas events (like letting go of mouse or touch outside the bounds of the screen), whereas `element-input` system was not. This PR also addresses that by having `element-input` handle that use-case properly, by removing hard-coded checks that prevent such events from firing. Note that `element-input` system does also Window DOM events.

Finally, this PR also makes the `mouse` events also use `element-input` system so everything is standardised to use the same system.

## Testing

Test project with custom engine build: https://launch.playcanvas.com/1223792

### Before

Before, any touch event (to any target) was being processed by all `element-drag-helper`s, so there were flickering and unwanted movement of bars. Also, if there was any touch still active, you couldn't click the button. It would highlight, but not increase its value on click.

https://user-images.githubusercontent.com/6392953/185370766-69e91963-2463-436c-95d7-a479339f573b.MP4

### After

After, each touch finger is handled properly by its target element only, and you can click on the button.

https://user-images.githubusercontent.com/6392953/185370794-942e9b22-c84e-4580-a0fa-1fdc0491013b.MP4

### DIFF
--- a/src/framework/components/element/element-drag-helper.js
+++ b/src/framework/components/element/element-drag-helper.js
@@ -83,7 +83,6 @@ class ElementDragHelper extends EventHandler {
 
     _toggleDragListeners(onOrOff) {
         const isOn = onOrOff === 'on';
-        const addOrRemoveEventListener = isOn ? 'addEventListener' : 'removeEventListener';
 
         // Prevent multiple listeners
         if (this._hasDragListeners && isOn) {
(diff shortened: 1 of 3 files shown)


slimbuck (approved):

mvaligursky (approved):
looks great!

yaustar:
I noticed that when both scrollbars were moving, they weren't highlighted as white. Is that expected behaviour?

jpauloruschel:
> I noticed that when both scrollbars were moving, they weren't highlighted as white. Is that expected behaviour?

Yes, because my fingers moved outside the bounding boxes of the elements, but keeping them pressed in order to test and show that the original pressed element continued to receive the touch events, as expected. If you keep the fingers(or mouse) within the bounding box of the element, they continue to be highlighted.

PR #4568, opened 2022-08-19
Standard shader compatibility improvements for WebGPU

mvaligursky:
- 2d texture sampling is using texture2DBias instead of texture2D to allow macro replacement for WebGPU
- Uniform buffer handles additional uniform types - alignment, setup
- adding extensions earlier in the shader code to avoid warnings

### DIFF
--- a/src/graphics/uniform-buffer-format.js
+++ b/src/graphics/uniform-buffer-format.js
@@ -83,6 +83,7 @@ class UniformFormat {
     }
 
     // std140 rules: https://registry.khronos.org/OpenGL/specs/gl/glspec45.core.pdf#page=159
+    // TODO: this support limited subset of functionality, arrays and structs are not supported.
     calculateOffset(offset) {
 
         // Note: vec3 has the same alignment as vec4
--- a/src/graphics/uniform-buffer-format.js
+++ b/src/graphics/uniform-buffer-format.js
@@ -1,41 +1,36 @@
 import { Debug } from '../core/debug.js';
+import { math } from '../math/math.js';
 import {
     uniformTypeToName, bindGroupNames,
     UNIFORMTYPE_BOOL, UNIFORMTYPE_INT, UNIFORMTYPE_FLOAT, UNIFORMTYPE_VEC2, UNIFORMTYPE_VEC3,
     UNIFORMTYPE_VEC4, UNIFORMTYPE_IVEC2, UNIFORMTYPE_IVEC3, UNIFORMTYPE_IVEC4, UNIFORMTYPE_BVEC2,
-    UNIFORMTYPE_BVEC3, UNIFORMTYPE_BVEC4, UNIFORMTYPE_MAT4
+    UNIFORMTYPE_BVEC3, UNIFORMTYPE_BVEC4, UNIFORMTYPE_MAT4, UNIFORMTYPE_MAT2, UNIFORMTYPE_MAT3
 } from './constants.js';
 
 /** @typedef {import('./scope-id.js').ScopeId} ScopeId */
 /** @typedef {import('./uniform-buffer.js').UniformBuffer} UniformBuffer */
 /** @typedef {import('./graphics-device.js').GraphicsDevice} GraphicsDevice */
 
-// map of UNIFORMTYPE_*** to byte size
-// rules: http://ptgmedia.pearsoncmg.com/images/9780321552624/downloads/0321552628_appl.pdf
-const uniformTypeToByteSize = [];
-uniformTypeToByteSize[UNIFORMTYPE_FLOAT] = 4;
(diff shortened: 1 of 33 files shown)


willeastcott:
This seems to touch a lot of front end chunks. Are we concerned this is going to break a lot of projects that override chunks?

willeastcott (approved):
Please get 1 more approval on this one... 😄

src/graphics/uniform-buffer-format.js
```diff
@@ -1,41 +1,36 @@
...
+uniformTypeToNumElements[UNIFORMTYPE_VEC4] = 4;
+uniformTypeToNumElements[UNIFORMTYPE_INT] = 1;
+uniformTypeToNumElements[UNIFORMTYPE_IVEC2] = 2;
+uniformTypeToNumElements[UNIFORMTYPE_IVEC3] = 3;
+uniformTypeToNumElements[UNIFORMTYPE_IVEC4] = 4;
+uniformTypeToNumElements[UNIFORMTYPE_BOOL] = 1;
+uniformTypeToNumElements[UNIFORMTYPE_BVEC2] = 2;
+uniformTypeToNumElements[UNIFORMTYPE_BVEC3] = 3;
+uniformTypeToNumElements[UNIFORMTYPE_BVEC4] = 4;
+uniformTypeToNumElements[UNIFORMTYPE_MAT2] = 8;    // 2 x vec4
```
GSterbrant, line 28: Do we want to handle 2x3, 2x4, 3x2, 3x4, 4x2, 4x3 matrices as well?

src/graphics/uniform-buffer-format.js
```diff
@@ -66,14 +61,37 @@ class UniformFormat {
...
+
+        this.byteSize = count * elementSize * 4;
         Debug.assert(this.byteSize, `Unknown byte size for uniform format ${type} used for ${name}`);
     }
+
+    // std140 rules: https://registry.khronos.org/OpenGL/specs/gl/glspec45.core.pdf#page=159
+    calculateOffset(offset) {
+
+        // Note: vec3 has the same alignment as vec4
+        const alignment = this.byteSize <= 8 ? this.byteSize : 16;
```
GSterbrant, line 89: This is correct for single values, but it's incomplete for arrays and structs where it has to be run recursively. 

Perhaps note this down so we don't forget it, as alignment issues are rather hard to detect.

src/graphics/uniform-buffer-format.js
```diff
@@ -66,14 +61,37 @@ class UniformFormat {
...
+
+        // align the start offset
+        offset = math.roundUp(offset, alignment);
+        this.offset = offset / 4;
```
GSterbrant, line 93: Why the division by 4?

src/graphics/uniform-buffer-format.js
```diff
@@ -102,17 +120,19 @@ class UniformBufferFormat {
...
         // TODO: optimize uniforms ordering
 
-        let byteSize = 0;
+        let offset = 0;
         for (let i = 0; i < uniforms.length; i++) {
             const uniform = uniforms[i];
-            uniform.offset = byteSize / 4;
-            byteSize += uniform.byteSize;
+            uniform.calculateOffset(offset);
+            offset = uniform.offset * 4 + uniform.byteSize;
```
GSterbrant, line 127: And then the multiplication by 4?

GSterbrant (commented):
I would probably move the chunk changes to another PR just to keep them separate, seeing as they take up the majority of the changes in this PR.

src/graphics/uniform-buffer-format.js
GSterbrant: Alright, if so, we don't need to support them either :)

src/graphics/uniform-buffer-format.js
mvaligursky: adding comment

src/graphics/uniform-buffer-format.js
mvaligursky: in this instance, I need byte size, so I convert it. In UniformBuffer I don't multiply it when writing to the buffer.

GSterbrant: Is the uniform offset in indices and not in bytes?

mvaligursky:
> I would probably move the chunk changes to another PR just to keep them separate, seeing as they take up the majority of the changes in this PR.

It's painful to split PRs where lots of little things change to get shader compilation moving, I'm not sure its worth it.

src/graphics/uniform-buffer-format.js
mvaligursky: yep, as that's how I use it each frame .. so optimizing for that.

GSterbrant (approved):

PR #4575, opened 2022-08-22
Refactoring to get more of the standard shader compile for WebGPU

mvaligursky:
- WIP on loading glb model and rendering it on WebGPU device
- moved bone limits from webgl device to the device to share it across platforms
- updated shader processor to handle precision specification for uniforms
- updated shader processor to ignore out keywords for now, as those are handled differently for now (we only need more generic solution when we add support for MRT)

### DIFF
--- a/src/graphics/shader-processor.js
+++ b/src/graphics/shader-processor.js
@@ -27,7 +27,9 @@ class UniformLine {
     constructor(line) {
         this.line = line;
 
-        const words = this.splitToWords(line);
+        // split to words handling any number of spaces
+        const words = line.trim().split(/\s+/);
+
         this.precision = undefined;
         this.type = words[0];
         this.name = words[1];
@@ -38,15 +40,8 @@ class UniformLine {
             this.type = words[1];
             this.name = words[2];
         }
-    }
 
-    splitToWords(line) {
-        // remove any double spaces
-        return line.trim().split(/\s+/);
-    }
-
-    get isSampler() {
-        return this.type.indexOf('sampler') !== -1;
+        this.isSampler = this.type.indexOf('sampler') !== -1;
     }
 }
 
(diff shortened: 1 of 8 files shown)


src/graphics/shader-processor.js
```diff
@@ -11,14 +11,46 @@ import { BindGroupFormat, BindBufferFormat, BindTextureFormat } from './bind-gro
...
+            this.precision = words[0];
+            this.type = words[1];
+            this.name = words[2];
+        }
+    }
+
+    splitToWords(line) {
+        // remove any double spaces
+        line = line.replace(/\s+/g, ' ').trim();
+        return line.split(' ');
```
kungfooman, line 46: You could also split it right away via RegEx, I don't know if it feels easier:
```suggestion
        return line.trim().split(/\s+/);
```

src/graphics/shader-processor.js
```diff
@@ -57,7 +89,15 @@ class ShaderProcessor {
...
         // uniforms - merge vertex and fragment uniforms, and create shared uniform buffers
         const uniforms = vertexExtracted.uniforms.concat(fragmentExtracted.uniforms);
-        const uniformsData = ShaderProcessor.processUniforms(device, uniforms, shaderDefinition.processingOptions);
+
+        // parse uniform lines
+        /** @type {Array<UniformLine>} */
+        const parsedUniforms = [];
+        uniforms.forEach((line) => {
+            parsedUniforms.push(new UniformLine(line));
+        });
```
kungfooman, line 98: There is no need for a type annotation when it can be inferred through mapping, which boils it all down to one line:

```suggestion
        const parsedUniforms = uniforms.map(line => new UniformLine(line));
```

slimbuck (approved):

PR #4581, opened 2022-08-23
Export AppOptions

slimbuck:
This is needed when creating a custom application.


mvaligursky:
Actually, this is probably not needed, as it's only useful for tree-shaking, when importing the file directly.

mvaligursky (approved):

PR #4584, opened 2022-08-23
Enable ImageBitmap if interface is present

slimbuck:
Since we no longer flip images at load time in the engine (#3335) we can use ImageBitmap for loading images.

ImageBitmap performs more work off the main thread and should result in much smoother experiences for our users.

This PR was tested on a recent version of Chrome, Safari and Firefox.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   ],
   "license": "MIT",
   "main": "build/playcanvas.js",
-  "module": "build/playcanvas.mjs/index.js",
+  "module": "build/playcanvas.mjs",
   "types": "build/playcanvas.d.ts",
   "bugs": {
     "url": "https://github.com/playcanvas/engine/issues"
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   ],
   "license": "MIT",
   "main": "build/playcanvas.js",
-  "module": "build/playcanvas.mjs",
+  "module": "build/playcanvas.mjs/index.js",
   "types": "build/playcanvas.d.ts",
   "bugs": {
     "url": "https://github.com/playcanvas/engine/issues"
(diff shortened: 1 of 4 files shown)


mvaligursky:
any concerns on older browses? For example:
- Firefox doesn't support options parameter to createImageBitmap (see https://bugzilla.mozilla.org/show_bug.cgi?id=1533680)

slimbuck:
> any concerns on older browses? For example:
> 
> * Firefox doesn't support options parameter to createImageBitmap (see https://bugzilla.mozilla.org/show_bug.cgi?id=1533680)

Nope. That was an issue with 'flipY' option which we no longer need.

slimbuck:
Think I will try find an old version of firefox to test on though.

package.json
```diff
@@ -18,7 +18,7 @@
   ],
   "license": "MIT",
   "main": "build/playcanvas.js",
-  "module": "build/playcanvas.mjs",
+  "module": "build/playcanvas.mjs/index.js",
```
mvaligursky, line 21: keep this as a separate PR?

src/deprecated/deprecated.js
```diff
@@ -904,7 +904,7 @@ Object.defineProperty(XrInputSource.prototype, 'rotation', {
 
 // INPUT
 
-export const input = {
+export const input_aa = {
```
mvaligursky, line 907: you sure?

package.json
slimbuck: oops, unintended: reverted.

src/deprecated/deprecated.js
slimbuck: oops, unintended: reverted.

PR #4585, opened 2022-08-23
Physically based camera

GSterbrant:
### Description

Initial PR for physical light units #3252 by enabling the use of physically based camera properties. The default properties (f/1.0, ISO 100, 1s) could have an effect on the overall lighting in the scene, so these are behind a switch.

### DIFF
--- a/src/framework/components/camera/component.js
+++ b/src/framework/components/camera/component.js
@@ -33,7 +33,10 @@ const properties = [
     { name: 'nearClip', readonly: false },
     { name: 'orthoHeight', readonly: false },
     { name: 'projection', readonly: false },
-    { name: 'scissorRect', readonly: false }
+    { name: 'scissorRect', readonly: false },
+    { name: 'aperture', readonly: false },
+    { name: 'shutter', readonly: false },
+    { name: 'sensitivity', readonly: false }
 ];
 
 /**
@@ -121,6 +124,10 @@ const properties = [
  * both front and back faces will be rendered. Defaults to true.
  * @property {boolean} flipFaces If true the camera will invert front and back faces. Can be useful
  * for reflection rendering. Defaults to false.
+ * @property {number} aperture The aperture of the camera lens in f-stops. High values means small aperture,
+ * low values means high aperture. Default is 1.4.
+ * @property {number} shutter The shutter speed of the camera in seconds. Default is 1.
+ * @property {number} sensitivity The sensitivity in ISO. Default is 100.
  * @augments Component
  */
 class CameraComponent extends Component {
--- a/src/framework/components/camera/component.js
+++ b/src/framework/components/camera/component.js
@@ -125,7 +125,7 @@ const properties = [
  * @property {boolean} flipFaces If true the camera will invert front and back faces. Can be useful
  * for reflection rendering. Defaults to false.
  * @property {number} aperture The aperture of the camera lens in f-stops. High values means small aperture,
- * low values means high aperture. Default is 1.4.
+ * low values means high aperture. Default is 1.0.
  * @property {number} shutter The shutter speed of the camera in seconds. Default is 1.
  * @property {number} sensitivity The sensitivity in ISO. Default is 100.
  * @augments Component
--- a/src/framework/components/camera/component.js
+++ b/src/framework/components/camera/component.js
@@ -396,6 +396,45 @@ class CameraComponent extends Component {
     }
 
     /**
+     * Set camera aperture
+     *
+     * @type {number}
+     */
+    set aperture(newValue) {
+        this._camera.aperture = newValue;
+    }
+
+    get aperture() {
+        return this._camera.aperture;
+    }
+
+    /**
+     * Set camera sensitivity
+     *
+     * @type {number}
+     */
+    set sensitivity(newValue) {
+        this._camera.sensitivity = newValue;
+    }
+
+    get sensitivity() {
+        return this._camera.sensitivity;
+    }
+
+    /**
+     * Set camera shutter speed
+     *
+     * @type {number}
+     */
+    set shutter(newValue) {
+        this._camera.shutter = newValue;
+    }
+
+    get shutter() {
+        return this._camera.shutter;
+    }
+
+    /**
      * Controls where on the screen the camera will be rendered in normalized screen coordinates.
      * Defaults to [0, 0, 1, 1].
      *
@@ -701,6 +740,9 @@ class CameraComponent extends Component {
         this.priority = source.priority;
         this.renderTarget = source.renderTarget;
         this.rect = source.rect;
+        this.aperture = source.aperture;
+        this.sensitivity = source.sensitivity;
+        this.shutter = source.shutter;
     }
 }
 
--- a/src/framework/components/camera/component.js
+++ b/src/framework/components/camera/component.js
@@ -124,10 +124,6 @@ const properties = [
  * both front and back faces will be rendered. Defaults to true.
  * @property {boolean} flipFaces If true the camera will invert front and back faces. Can be useful
  * for reflection rendering. Defaults to false.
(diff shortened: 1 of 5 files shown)


src/scene/camera.js
```diff
@@ -51,6 +51,10 @@ class Camera {
         this._renderTarget = null;
         this._scissorRect = new Vec4(0, 0, 1, 1);
         this._scissorRectClear = false; // by default rect is used when clearing. this allows scissorRect to be used when clearing.
+        this._usePhysicallyBasedCamera = false;
```
mvaligursky, line 54: I wonder if this should be exposed in the LightingParameters perhaps (which is part of the Scene). If this is per Camera, what happens if one camera is physical and one not, would they be able to use the lights, where some are physical and some not? Perhaps we need this per scene setting that would apply to both lights and cameras?

GSterbrant: Ah yeah, good point. Let's move it out to scene settings instead!

src/scene/camera.js
```diff
@@ -465,6 +495,11 @@ class Camera {
         return this._projMatSkybox;
     }
 
+    getExposure() {
+        const ev100 = Math.log2((this._aperture * this._aperture) / this._shutter * 100.0 / this._sensitivity);
+        return 1.0 / (Math.pow(2.0, ev100) * 1.2);
```
kungfooman, line 500: I can find references for calculating the exposure value (`ev100`) and I wonder what `1.2` means here. Is it like a magic number to make things look right?

src/framework/components/camera/component.js
```diff
@@ -121,6 +124,10 @@ const properties = [
  * both front and back faces will be rendered. Defaults to true.
  * @property {boolean} flipFaces If true the camera will invert front and back faces. Can be useful
  * for reflection rendering. Defaults to false.
+ * @property {number} aperture The aperture of the camera lens in f-stops. High values means small aperture,
+ * low values means high aperture. Default is 1.0.
+ * @property {number} shutter The shutter speed of the camera in seconds. Default is 1.
```
slimbuck, line 129: Is 1s default so lighting looks correct?

Since the common case for rendering on monitors is 60hz, I would have expected a default shutter speed of 1/60?

GSterbrant: Yeah, having it at 1s, f/1 and 100 ISO looks like it produces the same exposure as before. We can tweak them later, but for the current lighting parameters, these settings are good.

src/scene/camera.js
GSterbrant: It's from the filament documentation: https://google.github.io/filament/Filament.md.html#imagingpipeline/physicallybasedcamera

src/scene/scene.js
```diff
@@ -150,6 +150,13 @@ class Scene extends EventHandler {
      */
     sky = null;
 
+    /**
+     * Use physically based camera parameters to describe exposure
+     *
+     * @type {boolean}
+     */
+    usePhysicallyBasedCamera = false;
```
slimbuck, line 158: shouldn't this be a camera setting?

mvaligursky: we discussed that .. if you have multiple cameras and multiple lights, they probably all should be in the either physical or not .. so a global switch perhaps makes sense?

mvaligursky: even though it could perhaps be renamed to 'physicalBasedIntensities' or something

GSterbrant: I pushed this change.

slimbuck: So lights will also have legacy and physically-based units?

GSterbrant: Yeah, we already expose the exposure setting on the camera, and that 'fits' with the old light intensities. These units will have to be enabled in parallel for it to be consistent.

willeastcott: Just a little note to say this property is pretty verbose. `physicalCameras`?

GSterbrant: It's usePhysicallyBasedUnits currently, we could call it physicalUnits I guess.

src/scene/scene.js
```diff
@@ -150,6 +150,13 @@ class Scene extends EventHandler {
      */
     sky = null;
 
+    /**
+     * Use physically based units for cameras and lights
```
willeastcott, line 154:
```suggestion
 * Use physically based units for cameras and lights.
```

src/framework/components/camera/component.js
```diff
@@ -121,6 +124,10 @@ const properties = [
...
+ * @property {number} sensitivity The sensitivity in ISO. Default is 100.
```
willeastcott, line 130: Using the `@property` tag is mostly retired now. It is sufficient to move the descriptions here to the JSDoc block for the getter/setter pairs below. See `clearStencilBuffer` for an example.

src/scene/scene.js
```diff
@@ -150,6 +150,13 @@ class Scene extends EventHandler {
      */
     sky = null;
 
+    /**
+     * Use physically based units for cameras and lights.
+     *
+     * @type {boolean}
+     */
+    physicalUnits = true;
```
slimbuck, line 158: I know this has already been renamed once, but probably `physicalLightUnits` would make more sense?

GSterbrant: It's also units for cameras, they are not light units surely?

slimbuck: yeah right... physical units just sounds like a physics engine setting! :D

mvaligursky: maybe physicallyBasedRendering

slimbuck (approved):

src/framework/components/camera/component.js
```diff
@@ -388,6 +391,45 @@ class CameraComponent extends Component {
...
+    set sensitivity(newValue) {
+        this._camera.sensitivity = newValue;
+    }
+
+    get sensitivity() {
+        return this._camera.sensitivity;
+    }
+
+    /**
+     * Set camera shutter speed in seconds, the default value is 1s. Longer shutter means more exposure.
```
mvaligursky, line 421: I'm really not sure about 1s as a default here. What are we going to do in the future when we perhaps add a motion blur? I think the default value should be something that would work reasonably well for future purposes.

src/scene/camera.js
```diff
@@ -316,6 +319,30 @@ class Camera {
...
+        return this._sensitivity;
+    }
+
+    set shutter(newValue) {
+        this._shutter = newValue;
+    }
+
+    get shutter() {
+        return this._shutter;
+    }
```
mvaligursky, line 344: Currently these functions do nothing other than se / return the value. Personally, I'd remove them, and just rename the private variables to not have underscores in them.

src/scene/renderer/forward-renderer.js
```diff
@@ -415,6 +415,12 @@ class ForwardRenderer {
         this.nearClipId.setValue(camera._nearClip);
         this.farClipId.setValue(camera._farClip);
 
+        if (this.scene.physicalUnits) {
+            this.exposureId.setValue(camera.getExposure());
```
mvaligursky, line 419: Should the scene exposure be ignored here, or should it be somehow added on top of the camera exposure? If ignored, it'd be good to update its docs in the Scene to mention its not used when physicalUnits are used.

src/scene/camera.js
GSterbrant: This was just to have it be consistent with the other parameters, like clearDepth and clearDepthBuffer, etc.

src/framework/components/camera/component.js
GSterbrant: Addressing that here:

https://github.com/playcanvas/engine/pull/4617

src/scene/renderer/forward-renderer.js
GSterbrant: Done, it's mentioned in the documentation for both the physicalUnits switch as well as the direct exposure value on the camera.

PR #4587, opened 2022-08-25
Register shader generators at the point of use instead of always

mvaligursky:
- this removes relevant imports and allows improved tree-shaking - for example if particle systems are not used at all, generator code can be excluded (needs further improvement to handle relevant shader chunks as well)


slimbuck (approved):
I would rather pass the `programs` directly into `getProgram()` instead of the program name, since registering them seems wholly unnecessary.

However @mvaligursky assures me this will happen in a follow up PR, so approving.

PR #4590, opened 2022-08-25
Avoid using null Batcher

mvaligursky:
when batcher is not created, don’t use it

### DIFF
--- a/src/framework/app-base.js
+++ b/src/framework/app-base.js
@@ -1703,7 +1703,10 @@ class AppBase extends EventHandler {
 
     /** @private */
     _firstBatch() {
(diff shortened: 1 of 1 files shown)


src/framework/app-base.js
```diff
@@ -1703,7 +1703,10 @@ class AppBase extends EventHandler {
 
     /** @private */
     _firstBatch() {
-        this.batcher.generate();
+        const batcher = this.batcher;
```
slimbuck, line 1706: `this.batcher?.generate();` ?

mvaligursky: sure

PR #4591, opened 2022-08-25
Fix grab pass with downscaled render target

### LINKED ISSUE #4580: Dynamic refractions don't support scaling
### Description

When using pixel scaling, the dynamic refractions using the grab pass look like this:

image

When they should look like this:

image

This is probably due to not taking the pixel scaling into account when sampling the grab pass in the dynamic refraction chunk.

GSterbrant:
### Description
Fixes bug when pixel scaling was used with refractive objects. Fixes #4580.

Used to look like this:

image

After fix:

image

### DIFF
--- a/src/framework/scene-grab.js
+++ b/src/framework/scene-grab.js
@@ -81,12 +81,9 @@ class SceneGrab {
     }
 
     resizeCondition(source, requestedRenderTarget, device) {
-        const width = source ? source.width : device.width;
-        const height = source ? source.height : device.height;
-        if (requestedRenderTarget) {
-            return width !== requestedRenderTarget.width || height !== requestedRenderTarget.height;
-        }
-        return width !== device.width || height !== device.height;
+        const width = source?.width || device.width;
+        const height = source?.height || device.height;
+        return width !== requestedRenderTarget?.width || height !== requestedRenderTarget?.height;
     }
 
     allocateRenderTarget(renderTarget, sourceRenderTarget, device, format, isDepth, mipmaps, isDepthUniforms) {
--- a/src/framework/scene-grab.js
+++ b/src/framework/scene-grab.js
@@ -80,11 +80,13 @@ class SceneGrab {
         });
     }
 
-    resizeCondition(originalRenderTarget, requestedRenderTarget, device) {
-        if (originalRenderTarget && requestedRenderTarget) {
-            return originalRenderTarget.width !== requestedRenderTarget.width || originalRenderTarget.height !== requestedRenderTarget.height;
+    resizeCondition(source, requestedRenderTarget, device) {
+        const width = source ? source.width : device.width;
+        const height = source ? source.height : device.height;
+        if (requestedRenderTarget) {
+            return width !== requestedRenderTarget.width || height !== requestedRenderTarget.height;
         }
-        return originalRenderTarget?.width !== device.width || originalRenderTarget?.height !== device.height;
+        return width !== device.width || height !== device.height;
     }
 
     allocateRenderTarget(renderTarget, sourceRenderTarget, device, format, isDepth, mipmaps, isDepthUniforms) {
--- a/src/framework/scene-grab.js
+++ b/src/framework/scene-grab.js
@@ -183,7 +183,7 @@ class SceneGrab {
                 }
 
                 if (camera.renderSceneDepthMap) {
-                    
+
                     // reallocate RT if needed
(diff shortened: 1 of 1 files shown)


src/framework/scene-grab.js
```diff
@@ -153,9 +153,9 @@ class SceneGrab {
                 if (camera.renderSceneColorMap) {
 
                     // allocate / resize existing RT as needed
-                    if (this.colorRenderTarget?.width !== device.width || this.colorRenderTarget?.height !== device.height) {
+                    if (this.colorRenderTarget?.width !== camera.renderTarget?.colorBuffer?.width || this.colorRenderTarget?.height !== camera.renderTarget?.colorBuffer?.height) {
```
mvaligursky, line 156: this won't work in case camera.renderTarget is null/undefined.
I'd suggest to create a compare function, which takes two render targets and a device, compares it all ... and use it 4 times in this file.

mvaligursky: test it with the grab-pass engine example and resize the window - that likely does not work

src/framework/scene-grab.js
```diff
@@ -80,13 +80,20 @@ class SceneGrab {
         });
     }
 
-    allocateRenderTarget(renderTarget, device, format, isDepth, mipmaps, isDepthUniforms) {
+    resizeCondition(originalRenderTarget, requestedRenderTarget, device) {
+        if (originalRenderTarget && requestedRenderTarget) {
+            return originalRenderTarget.width !== requestedRenderTarget.width || originalRenderTarget.height !== requestedRenderTarget.height;
```
mvaligursky, line 85: why not just have code like above
```
const width = source ? source.colorBuffer.width : device.width,
const height = source ? source.colorBuffer.height : device.height,
```
and then compare it to requestedRenderTarget?

mvaligursky: it would make it more readable

src/framework/scene-grab.js
```diff
@@ -80,13 +80,22 @@ class SceneGrab {
...
 
-    allocateRenderTarget(renderTarget, device, format, isDepth, mipmaps, isDepthUniforms) {
+    resizeCondition(source, requestedRenderTarget, device) {
+        const width = source ? source.width : device.width;
+        const height = source ? source.height : device.height;
+        if (requestedRenderTarget) {
+            return width !== requestedRenderTarget.width || height !== requestedRenderTarget.height;
+        }
+        return width !== device.width || height !== device.height;
+    }
```
mvaligursky, line 90: I'm not sure that's correct, it should be this way I think
```
    resizeCondition(resizeTarget, cameraTarget, device) {
        const width = cameraTarget?.width || device.width;
        const height = cameraTarget?.height || device.height;
        return width !== resizeTarget?.width || height !== resizeTarget?.height;
    }
```

GSterbrant: Ah, didn't know you can OR together undefined and a value. Yes that would make it even terser.

slimbuck (approved):

src/framework/scene-grab.js
kungfooman: Using object destructuring it would be even shorter:

```js
function resizeCondition(resizeTarget, cameraTarget, device) {
    const {width, height} = cameraTarget || device;
    return width !== resizeTarget?.width || height !== resizeTarget?.height;
}
```

The generated ES5 code is also shorter (354 vs 488 chars ): https://babeljs.io/repl/#?browsers=&build=&builtIns=false&corejs=3.6&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABAJwKYGcYC9UGEEAmMsCAFGpjgCoCGyA5qlADSIQ0C2qyNtDTrAqgBuMCKgCUiAN4AoRArYJ0URAHcYBKAAtEAXjaduvOoygB-AHQatugD53EQ0eOuadAbnmKIy1dtQYem1VA3YuHj4zKwCgkMQHJxExVEtY4KgvRRQmEGQkGx1EAEI9AwpsVCimK0L7R3T40vKMSuqLNMCMrwBfWVBIEiQKnHwwIiGAJnJW6lMBQwiTfhYkl0kZbwVfMBUZOtZGqB79ReN2hMdnFKzFNCg8gvddZpzKKvmOusvEI5Kyt5tT4xLohXqyIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=script&lineWrap=true&presets=env&prettier=false&targets=&version=7.18.13&externalPlugins=&assumptions=%7B%7D

(untested, maybe I miss something here)

PR #4596, opened 2022-08-30
Revert enabling image bitmap

slimbuck:
This PR reverts #4584.

After doing some testing it seems ImageBitmap is still not well supported across browsers. Details have been left in the code for next time.

It might be worth investigating more closely which browsers have concrete support for ImageBitmap and start selectively enabling the interface on those browsers.


mvaligursky:
This PR removes support for WebGPU, which it should not.

slimbuck:
> This PR removes support for WebGPU, which it should not.

Webgp.. what? :D Thanks - fixed!

mvaligursky (approved):

PR #4598, opened 2022-08-30
Test for ImageBitmap support at startup

slimbuck:
We recently disabled ImageBitmap support again after finding issues in Safari (#4596).

This PR introduces a test at startup for determining whether ImageBitmap support is solid or not.

### DIFF
--- a/src/graphics/webgl/webgl-graphics-device.js
+++ b/src/graphics/webgl/webgl-graphics-device.js
@@ -191,7 +191,7 @@ function testImageBitmap(device) {
 
             // read pixels
             const rt = new RenderTarget({ colorBuffer: texture, depth: false });
-            device.setFramebuffer(rt._glFrameBuffer);
+            device.setFramebuffer(rt.impl._glFrameBuffer);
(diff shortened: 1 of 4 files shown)


src/graphics/webgl/webgl-graphics-device.js
```diff
@@ -309,6 +355,12 @@ class WebglGraphicsDevice extends GraphicsDevice {
         this.initializeRenderState();
         this.initializeContextCaches();
 
+        // start async image bitmap test
+        this.supportsImageBitmap = null;
```
mvaligursky, line 359: should this be initialized to false instead?

mvaligursky (approved):
neat!

src/graphics/webgl/webgl-graphics-device.js
slimbuck: I thought null would be more useful since it'll evaluate to false, but also indicate the result has not yet been set.

mvaligursky: fair point

Issue #4595, opened 2022-08-26
Enabling markup on a text element adds shadow even when offset is 0, 0


yaustar:
Thread: https://forum.playcanvas.com/t/text-markup-display-issues-shadow-default/27240
Project: https://playcanvas.com/project/974166/overview/ui-font-issue

Enable/disable markup on the text element and you can see the shadow get added/removed in the view

https://user-images.githubusercontent.com/16639049/186944109-56a99b0b-59a3-4290-9b85-0fa6e4cb1cf7.mp4

yaustar:
@querielo , do you think this could related to the changes you made to the text element a while back?

PR #4600, opened 2022-08-30
Fix unpacking of msdf markup params

NewboO:
Fixes: https://github.com/playcanvas/engine/issues/4595

At the moment, unpacking function of text markup parameters isn't the exact inverse of packing, especially on shadow offset.

Simple example of the issue is to keep default settings for shadow (ie. white text with a black shadow at an offset of (0, 0)) and toggling markup without actually using any tag. It gives something like this (slightly visible on black background, obvious on white blackground):
image

This is because an offset of (0, 0) is packed as 127 + 127 * 256 = 32639 but unpacked as ((32639 % 256) / 256 * 2 - 1, (32639 / 256 / 256) * 2 - 1) = (-0.008, -0.004).

This corrected function gives the expected rendering:
image

Color unpacking error is barely visible but comes from the same kind of difference (packed by multiplying by 255, unpacked by dividing by 256).

### DIFF
--- a/src/graphics/program-lib/chunks/common/vert/msdf.js
+++ b/src/graphics/program-lib/chunks/common/vert/msdf.js
@@ -8,20 +8,23 @@ varying vec4 shadow_color;
 varying vec2 shadow_offset;
 
 void unpackMsdfParams() {
-    outline_color.rb = mod(vertex_outlineParameters.xy, 256.) / 256.;
-    outline_color.ga = vertex_outlineParameters.xy / 256. / 256.;
+    vec3 little = mod(vertex_outlineParameters, 256.);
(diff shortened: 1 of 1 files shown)


yaustar:
Is it related to this issue? https://github.com/playcanvas/engine/issues/4595

NewboO:
Indeed, forgot to check if this was reported.

yaustar:
That's an awesome find, thank you!

yaustar:
@slimbuck @mvaligursky @GSterbrant Would you be able to check the PR please?

slimbuck (approved):
Approving without testing. Thanks for this contribution!!

mvaligursky (approved):

PR #4601, opened 2022-08-31
Output treeshakeable module builds

slimbuck:
This PR changes the engine es6/esm module output to preserve modules. This results in a hierarchy of files for es6 target, one module for each interface exported by index.js (the rest is not treeshakable and is placed directly into index.js).

Preserving modules will allow applications to treeshake the engine source.

We also remove the header comment from the minified build, otherwise it becomes bloated with comments (since each individual module gets it).

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
   "main": "build/playcanvas.js",
   "module": "build/playcanvas.mjs/index.js",
   "types": "build/playcanvas.d.ts",
+  "sideEffects": false,
   "bugs": {
     "url": "https://github.com/playcanvas/engine/issues"
   },
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   ],
   "license": "MIT",
   "main": "build/playcanvas.js",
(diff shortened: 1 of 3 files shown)


mvaligursky:
Could you please expand on this a bit more?
`the rest is not treeshakable and is placed directly into index.js`
What is 'the rest'?

slimbuck:
> What is 'the rest'?
All the bits not exposed by index.js. It isn't much - looks like just the polyfills.

slimbuck:
One of the questions is around naming of the output. With this PR in place, `build/` folder contains:
```
playcanvas.js
playcanvas.mjs/
    index.js
    anim/
    ...
playcanvas.min.js
playcanvas.min.mjs/
    index.js
    anim/
    ...
...
```
I think this works ok. What do you think @willeastcott?

willeastcott:
Yes, I think this is fine. I mean, we can change what the `package.json` `module` field points to whenever we like and peoples' build processes should continue to work just fine.

willeastcott (approved):

PR #4603, opened 2022-08-31
New hierarchy find functions

NewboO:
This PR add various functions to find entities, components and script instances either through parents or children. It completes the already existing functions to find through children and can be used the same way.

List of new functions:
- GraphNode.findDescendants
- GraphNode.findDescendant
- GraphNode.forEachDescendant
- GraphNode.findAncestors
- GraphNode.findAncestor
- GraphNode.forEachAncestor
- Entity.findScript
- Entity.findScripts
- Entity.findComponentInDescendants
- Entity.findComponentsInDescendants
- Entity.findScriptInDescendants
- Entity.findScriptsInDescendants
- Entity.findComponentInAncestors
- Entity.findComponentsInAncestors
- Entity.findScriptInAncestors
- Entity.findScriptsInAncestors

It also simplifies some already existing functions in `GraphNode`.

I hope I documentated them correctly. I also added unit tests inspired by the existing ones.

### DIFF
--- a/src/scene/graph-node.js
+++ b/src/scene/graph-node.js
@@ -37,8 +37,9 @@ function _createTest(attr, value) {
 }
 
 /**
- * Callback used by {@link GraphNode#find}, {@link GraphNode#findOne}, {@link GraphNode#findAncestor}
- * and {@link GraphNode#findAncestors} to search through a graph node and all of its descendants or ascendants.
+ * Callback used by {@link GraphNode#find}, {@link GraphNode#findOne}, {@link GraphNode#findDescendants},
+ * {@link GraphNode#findDescendant}, {@link GraphNode#findAncestors} or {@link GraphNode#findAncestor}
+ * to search through a graph node and all of its descendants or ascendants.
  *
  * @callback FindNodeCallback
  * @param {GraphNode} node - The current graph node.
--- a/src/scene/graph-node.js
+++ b/src/scene/graph-node.js
@@ -539,7 +539,6 @@ class GraphNode extends EventHandler {
      */
     findOne(attr, value) {
         const len = this._children.length;
-        let result = null;
 
         if (attr instanceof Function) {
             if (attr(this))
@@ -557,7 +556,7 @@ class GraphNode extends EventHandler {
         }
 
         for (let i = 0; i < len; ++i) {
-            result = this._children[i].findOne(attr, value);
+            const result = this._children[i].findOne(attr, value);
             if (result)
                 return result;
         }
--- a/src/scene/graph-node.js
+++ b/src/scene/graph-node.js
@@ -22,12 +22,19 @@ const invParentRot = new Quat();
 const matrix = new Mat4();
 const target = new Vec3();
 const up = new Vec3();
-const checkAttr = (node, attr, value) => {
-    if (node[attr] instanceof Function)
-        return node[attr]() === value;
 
-    return node[attr] === value;
-};
+function _createTest(attr, value) {
+    if (attr instanceof Function) {
+        return attr;
+    }
+    return (node) => {
+        let x = node[attr];
+        if (x instanceof Function) {
+            x = x();
+        }
+        return x === value;
+    };
+}
 
 /**
  * Callback used by {@link GraphNode#find}, {@link GraphNode#findOne}, {@link GraphNode#findParents}
@@ -495,18 +502,12 @@ class GraphNode extends EventHandler {
      */
     find(attr, value) {
         const results = [];
+        const test = _createTest(attr, value);
 
-        if (attr instanceof Function) {
-            this.forEach((node) => {
-                if (attr(node))
-                    results.push(node);
-            });
-        } else {
-            this.forEach((node) => {
-                if (checkAttr(node, attr, value))
-                    results.push(node);
-            });
-        }
+        this.forEach((node) => {
+            if (test(node))
+                results.push(node);
+        });
 
         return results;
     }
@@ -536,17 +537,14 @@ class GraphNode extends EventHandler {
      * var node = parent.findOne('name', 'Test');
      */
     findOne(attr, value) {
+        const test = _createTest(attr, value);
         const len = this._children.length;
 
-        if (attr instanceof Function) {
-            if (attr(this))
-                return this;
-        } else if (checkAttr(this, attr, value)) {
+        if (test(this))
             return this;
-        }
 
         for (let i = 0; i < len; ++i) {
-            const result = this._children[i].findOne(attr, value);
+            const result = this._children[i].findOne(test);
             if (result)
                 return result;
         }
@@ -579,18 +577,12 @@ class GraphNode extends EventHandler {
      */
     findParents(attr, value) {
         const results = [];
+        const test = _createTest(attr, value);
 
-        if (attr instanceof Function) {
-            this.forEachParent((node) => {
-                if (attr(node))
-                    results.push(node);
-            });
-        } else {
-            this.forEachParent((node) => {
-                if (checkAttr(node, attr, value))
-                    results.push(node);
-            });
-        }
+        this.forEachParent((node) => {
+            if (test(node))
+                results.push(node);
+        });
 
         return results;
     }
@@ -620,16 +612,11 @@ class GraphNode extends EventHandler {
      * var node = parent.findOneParent('name', 'Test');
      */
     findOneParent(attr, value) {
-        let testNode, current = this;
-
-        if (attr instanceof Function) {
-            testNode = node => attr(node);
-        } else {
-            testNode = node => checkAttr(node, attr, value);
-        }
+        const test = _createTest(attr, value);
+        let current = this;
 
         while (current) {
-            if (testNode(current))
+            if (test(current))
                 return current;
 
             current = current._parent;
--- a/src/scene/graph-node.js
+++ b/src/scene/graph-node.js
@@ -622,7 +622,7 @@ class GraphNode extends EventHandler {
             if (result)
                 results.push(this);
 
-            if(this._parent) {
+            if (this._parent) {
                 results = results.concat(this._parent.findInParent(fn));
             }
         } else {
@@ -638,7 +638,7 @@ class GraphNode extends EventHandler {
                     results.push(this);
             }
 
-            if(this._parent) {
+            if (this._parent) {
                 results = results.concat(this._parent.findInParent(attr, value));
             }
         }
@@ -672,7 +672,7 @@ class GraphNode extends EventHandler {
      */
     findOneInParent(attr, value) {
         let result = null;
-        
+
         if (attr instanceof Function) {
             const fn = attr;
 
@@ -680,7 +680,7 @@ class GraphNode extends EventHandler {
             if (result)
                 return this;
 
-            if(this._parent) {
+            if (this._parent) {
                 return this._parent.findOneInParent(fn);
             }
         } else {
@@ -696,7 +696,7 @@ class GraphNode extends EventHandler {
                 }
             }
 
-            if(this._parent) {
+            if (this._parent) {
                 return this._parent.findOneInParent(attr, value);
             }
         }
--- a/src/scene/graph-node.js
+++ b/src/scene/graph-node.js
@@ -502,17 +502,15 @@ class GraphNode extends EventHandler {
             };
         } else {
             queryNode = (node) => {
-                if (node[attr]) {
-                    let testValue;
+                let testValue;
 
-                    if (node[attr] instanceof Function) {
-                        testValue = node[attr]();
-                    } else {
-                        testValue = node[attr];
-                    }
-                    if (testValue === value)
-                        results.push(node);
+                if (node[attr] instanceof Function) {
+                    testValue = node[attr]();
+                } else {
+                    testValue = node[attr];
                 }
+                if (testValue === value)
+                    results.push(node);
 
                 for (let i = 0, len = node._children.length; i < len; ++i) {
                     queryNode(node._children[i]);
@@ -555,7 +553,7 @@ class GraphNode extends EventHandler {
         if (attr instanceof Function) {
             if (attr(this))
                 return this;
-        } else if (this[attr]) {
+        } else {
             let testValue;
 
             if (this[attr] instanceof Function) {
@@ -607,9 +605,6 @@ class GraphNode extends EventHandler {
             testNode = node => attr(node);
         } else {
             testNode = (node) => {
-                if (!node[attr])
-                    return false;
-
                 if (node[attr] instanceof Function)
                     return node[attr]() === value;
 
@@ -658,9 +653,6 @@ class GraphNode extends EventHandler {
             testNode = node => attr(node);
         } else {
             testNode = (node) => {
-                if (!node[attr])
-                    return false;
-
                 if (node[attr] instanceof Function)
                     return node[attr]() === value;
 
--- a/src/scene/graph-node.js
+++ b/src/scene/graph-node.js
@@ -492,34 +492,22 @@ class GraphNode extends EventHandler {
         const len = this._children.length;
 
         if (attr instanceof Function) {
-            const fn = attr;
-
-            if (fn(this))
+            if (attr(this))
                 results.push(this);
-
-            for (let i = 0; i < len; i++) {
-                const descendants = this._children[i].find(fn);
-                if (descendants.length)
-                    results.push(...descendants);
-            }
-        } else {
+        } else if (this[attr]) {
             let testValue;
 
-            if (this[attr]) {
-                if (this[attr] instanceof Function) {
-                    testValue = this[attr]();
-                } else {
-                    testValue = this[attr];
-                }
-                if (testValue === value)
-                    results.push(this);
+            if (this[attr] instanceof Function) {
+                testValue = this[attr]();
+            } else {
+                testValue = this[attr];
             }
+            if (testValue === value)
+                results.push(this);
+        }
 
-            for (let i = 0; i < len; ++i) {
-                const descendants = this._children[i].find(attr, value);
-                if (descendants.length)
-                    results.push(...descendants);
-            }
+        for (let i = 0; i < len; ++i) {
+            results.push(...this._children[i].find(attr, value));
         }
 
         return results;
@@ -554,34 +542,24 @@ class GraphNode extends EventHandler {
         let result = null;
 
         if (attr instanceof Function) {
-            const fn = attr;
-
-            if (fn(this))
+            if (attr(this))
                 return this;
-
-            for (let i = 0; i < len; i++) {
-                result = this._children[i].findOne(fn);
-                if (result)
-                    return result;
-            }
-        } else {
+        } else if (this[attr]) {
             let testValue;
-            if (this[attr]) {
-                if (this[attr] instanceof Function) {
-                    testValue = this[attr]();
-                } else {
-                    testValue = this[attr];
-                }
-                if (testValue === value) {
-                    return this;
-                }
-            }
 
-            for (let i = 0; i < len; i++) {
-                result = this._children[i].findOne(attr, value);
-                if (result)
-                    return result;
+            if (this[attr] instanceof Function) {
+                testValue = this[attr]();
+            } else {
+                testValue = this[attr];
             }
+            if (testValue === value)
+                return this;
+        }
+
+        for (let i = 0; i < len; ++i) {
+            result = this._children[i].findOne(attr, value);
+            if (result)
+                return result;
         }
 
         return null;
@@ -614,30 +592,22 @@ class GraphNode extends EventHandler {
         const results = [];
 
         if (attr instanceof Function) {
-            const fn = attr;
-
-            if (fn(this))
+            if (attr(this))
                 results.push(this);
-
-            if (this._parent) {
-                results.push(...this._parent.findInParent(fn));
-            }
-        } else {
+        } else if (this[attr]) {
             let testValue;
 
-            if (this[attr]) {
-                if (this[attr] instanceof Function) {
-                    testValue = this[attr]();
-                } else {
-                    testValue = this[attr];
-                }
-                if (testValue === value)
-                    results.push(this);
+            if (this[attr] instanceof Function) {
+                testValue = this[attr]();
+            } else {
+                testValue = this[attr];
             }
+            if (testValue === value)
+                results.push(this);
+        }
 
-            if (this._parent) {
-                results.push(...this._parent.findInParent(attr, value));
-            }
+        if (this._parent) {
+            results.push(...this._parent.findInParent(attr, value));
         }
 
         return results;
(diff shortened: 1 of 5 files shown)


src/scene/graph-node.js
```diff
@@ -589,6 +589,121 @@ class GraphNode extends EventHandler {
...
+     * // Finds all nodes that have the name property set to 'Test'
+     * var entities = entity.findInParent('name', 'Test');
+     */
+    findInParent(attr, value) {
+        let result, results = [];
+
+        if (attr instanceof Function) {
+            const fn = attr;
+
+            result = fn(this);
```
kungfooman, line 621: `result` has more context than it requires and you could turn it into a `const`, the results array should probably also be a `const` and then simply `results.push(...this._parent.findInParent(fn))`.

(everything is arguable, I just prefer `const` and not creating new intermediate arrays, that makes it easier to argue about)

And it is called `findInParent`, but doesn't this also check the same entity aswell? It would be like referring to `>` as `>=`, which is kinda confusing?

NewboO: I agree about `result` and array spreading, this is mainly copy/pasted from current functions `find` and `findOne`. Maybe those two should be also rewritten then.

It checks the current entity because I'm used to Unity where `GetComponent`, `GetComponentInChildren` and `GetComponentInParent` are all checking the current node. It prevents the need to use a weird syntax or to create a child node only for getting all matching entities, which isn't such an unusual use case imho.
Moreover, `find` and `findOne` are also checking current entity, which can also be confusing and is harder to circumvent if you want to ignore the current entity.

yaustar:
Personal take as an end user. I think Entity.findScript and Entity.findScripts could be very useful but the *in parent don't seem as useful given that you can get the parent and then call find* functions on that?

NewboO:
@yaustar What if you don't know if the script or component you want is the direct parent or higher in the hierarchy? What if the parent doesn't have the script or component you want but a child do?

For instance, as I love having generic scripts, I have some to tint, scale, etc. elements based on a parent button state (the button component can only tint OR change sprite of a single element) thus my scripts can simply find the closest button in its ancestors (which can be itself) and I don't need to write specific scripts each time a button doesn't have the same hierarchy.

yaustar:
Does `Entity.findComponentsInParent` search all the parents? Or just the immediate parent?

NewboO:
All the parents, same as Unity though "*InParents" would be more accurate indeed.

NewboO:
On a side note, does anyone know how to fix the errors in TypeScripts declarations? I wanted to documentate `findScript[s]*` functions so the first parameter show as `string, typeof(ScriptType)` like in ScriptComponent documentation but I can't find a way to make it work:
- `typeof ScriptType` passes the check but the documentation can't be generated anymore
- `@typedef {...}` doesn't work
- `import { ScriptType } from ...` doesn't work

src/scene/graph-node.js
```diff
@@ -519,7 +518,7 @@ class GraphNode extends EventHandler {
             for (let i = 0; i < len; ++i) {
                 const descendants = this._children[i].find(attr, value);
                 if (descendants.length)
-                    results = results.concat(descendants);
+                    results.push(...descendants);
```
kungfooman, line 521: I could imagine that `descendants.length` is only checked for optimisation to not create a new results array all the time, but since `push` in the case of zero length would just do nothing, there shouldn't be any need for length checking anymore?

NewboO: That's right. I modified those parts and also made other changes such as moving parent/children recursion out of if...else blocks so they're not duplicated anymore.

NewboO:
I included the suggestion to avoid empty arrays, and also applied it to `find` function.

However, using a `this[]` type feels really off. In your example, it would indeed return a `TestNode[]` but nothing prevents you from having a hierarchy mixing different subtypes of `GraphNode`, most of them being `Entity` in real use cases, so it would be wrong to say it returns a `TestNode[]` in this other example:

```javascript
class TestNode extends GraphNode {
    // Nothing yet.
};
const rootNode = new GraphNode("hai");
const testNode = new TestNode("hai");
rootNode.addChild(testNode);
const ret = testNode.findInParent("name", "hai");
```

Moreover, that would be way beyond the initial objective of this PR because every properties and functions to get, add and remove nodes are documented as returning or expecting `GraphNode` thus they would all need to be changed accordingly.

NewboO:
Function are now pluralized as `*InParents` for clarity and are also non-recursive so the `results` parameter is now removed again.
`find` is now like `findByTag` to also remove the `results` parameter but still have a single array where matching nodes are pushed only when necessary.
I also added `findByNameInParents` to have the `findByName` equivalent.

By the way, about whether current node can be returned or not, I noticed that `findByTag` and `findByPath` are excluding it while `find`, `findOne` and `findByName` are including it. Shouldn't all those functions behave the same way? Though changing any of those might break some projects...

kungfooman:
Your `GraphNode#find` is duplicating quite some lines of code now, but using a closure is of course a good idea to not add `results = []` as default value argument.

You can just reuse `GraphNode#forEach`:

```js
    find(attr, value) {
        const results = [];
        if (attr instanceof Function) {
            this.forEach((node) => {
                if (attr(node))
                    results.push(node);
            });
        } else {
            this.forEach((node) => {
                let testValue;

                if (node[attr] instanceof Function) {
                    testValue = node[attr]();
                } else {
                    testValue = node[attr];
                }
                if (testValue === value)
                    results.push(node);
            });
        }
        return results;
    }
```

I am also contemplating more helper functions for code like this:

```js
        while (current) {
            if (current.name === name)
                return current;

            current = current._parent;
        }
```

`GraphNode#forEachParent` (inspired by `GraphNode#forEach`)
`GraphNode#findOneParent` (inspired by `GraphNode#findOne`)

Then there is also more code duplication regarding:

```js
            // ...
            if (this[attr] instanceof Function) {
                testValue = this[attr]();
            } else {
                testValue = this[attr];
            }
```

Usually its called `resolve`. But yea, I guess you didn't plan to refactor all of these other methods.

NewboO:
I added `forEachParent` and used it in `findParents` the same way `find` is now using `forEach`.

I already did `findOneParent` but it was named `findOneInParents`, I changed its name for clarity as well as `findInParents` is now `findParents` and `findByNameInParents` is now `findParentByName`. I'm wondering about doing `findParentsByTag` but I can't think of relevant examples where it could be useful.

I'm not sure I understood correctly what you meant with the duplicated code to get the test value, but I added a const function at the beginning of the file to directly check if an attribute matches a value and all four functions are now using it. Is this what you had in mind?

kungfooman:
> I'm not sure I understood correctly what you meant with the duplicated code to get the test value, but I added a const function at the beginning of the file to directly check if an attribute matches a value and all four functions are now using it. Is this what you had in mind?

Thank you for the quick implementation, yes, that is what I had in mind and you can even go a step further:

```js
/**
 * 
 * @param {Function | string} attr 
 * @param {*} value 
 * @returns {Function}
 */
function createTest(attr, value) {
    if (attr instanceof Function) {
        return attr;
    }
    return (node) => {
        let x = node[attr];
        if (x instanceof Function) {
            x = x();
        }
        return x === value;
    };
}
```

If you do not "hardcode" `node`, but return a function which takes said `node`, other code just "melts" away, because its unified into one `test` function:

For example, `GraphNode#find` you can rewrite into:

```js
    find(attr, value) {
        const results = [];
        const test = createTest(attr, value);
        this.forEach((node) => {
            if (test(node))
                results.push(node);
        });
        return results;
    }
```

That boils this method from 37 lines down to these 9 lines and makes it more reusable aswell. You can simplify/shorten the other methods with the same technique.

src/scene/graph-node.js
```diff
@@ -646,6 +682,26 @@ class GraphNode extends EventHandler {
...
+
+        while (current) {
+            if (current.name === name)
+                return current;
+
+            current = current._parent;
+        }
+
+        return null;
+    }
```
kungfooman, line 703: By not reimplementing the iteration logic everywhere, you can save quite some lines. Do you see reasons to not just do this?

```suggestion
    findParentByName(name) {
        return this.findOneParent('name', name);
    }
```

Maksims:
Those are useful, but might be overly specific and subject to memorizing due to quantity and naming.

Current naming convention uses "findBy".

In graphs, "parent" - is immediate node above, if you want to go all they way, then it is "ancestors", as opposing to "descendants".

kungfooman:
> In graphs, "parent" - is immediate node above, if you want to go all they way, then it is "ancestors", as opposing to "descendants".

Strictly I agree with you and in practice programmers prefer `parents` to refer to the ancestors of a node in an out-tree (as its called in graph theory).

For example in jQuery: https://api.jquery.com/parents/

`.parents( [selector ] )`: `Get the ancestors of each element in the current set of matched elements, optionally filtered by a selector.`

What bugs me most is that I am supposed to be my own parent in the current implementation. This is aligned with Unity, but would that make any sense in practical usage?

src/scene/graph-node.js
NewboO: I was thinking about performance though it wouldn't make such a difference for `findParentByName`, a bit more if implementing `findByName` with `findOne('name', name)` but I guess both should be changed to respect PlayCanvas coding standards (and these functions aren't supposed to be called each frame anyway).

NewboO:
Imo, what's more important than aligning with another game engine (or even with a javascript library) is both the consistency and the symmetry between finding among descendants and finding among ascendants.

In Unity, both `GetComponent[s]InChildren` and `GetComponent[s]InParent` are returning components of the current node (Unity doesn't have a lot of find functions anyway).
In jQuery, current node can't be returned in both `find` and `parents` (nor in `children` and `parent` as there are specific functions for searching through a single level).

But at the moment in PlayCanvas, as noticed already, `findByTag` and `findByPath` can't return current node while `find`, `findOne` and `findByName` can. It means for instance that `node.findByName('SomeNode')` might not return always the same node as `node.findByPath('SomeNode')`. That should be addressed whatever happens with this PR.

kungfooman:
> Also, about naming, `findByName` and `findByPath` are both returning a single node while `findByTag` returns an array, that's another inconsistency at the moment.

Good point which I would like to add to: `find` (ES6/2015) and `filter` (ES5/2009) suggest certain return values:

```js
[10, 20, 30].find(x => x >= 20); // 20
```

```js
[10, 20, 30].filter(x => x >= 20); // [20, 30]
```

In short:
 - `find` is associated with a single/the-first returned element that fits the bill
 - `filter` is what a JS dev associates with a returned array

So the outlier in your example is `GraphNode#findByTag`, which should return a single element based on established JavaScript and PlayCanvas methods.

But that would also mean that `GraphNode#find` is wrong, as it returns an array currently. From a purist point of view there shouldn't be any `findOne` method, as `find` is used for that. And the current `find` should be `filter`. `forEach` is already `forEach` in the engine and `GraphNode#map` is missing altogether.

IMO it would be nice/intuitive that `GraphNode` would follow the rest of the JS world, but it would also break a bunch of projects in the meantime...

src/scene/graph-node.js
kungfooman: Its nearly impossible to judge performance without measuring. I wouldn't be surprised if V8 optimizes the heck out of `GraphNode#findOneParent` as soon its called often enough to justify JIT hot patching which optimizes it to opcode machine code level performance.

NewboO:
> IMO it would be nice/intuitive that `GraphNode` would follow the rest of the JS world, but it would also break a bunch of projects in the meantime...

Is there really any shared convention? I mean, there are `find` and `filter` for arrays, but the DOM API defines `querySelector` and `querySelectorAll` for DOM tree, jQuery `find` returns an array-like object, and it seems each library has its own naming anyway. Also, some people might expect calling `filter` (or `map`) will return an array of filtered/mapped direct children (JS array functions aren't recursive after all, except the ES10 `flat` and `flatMap`).

Moreover, I don't really want to be the root cause of breaking lots of projects if many functions end up being renamed 😄

kungfooman:
> Also, some people might expect calling `filter` (or `map`) will return an array of filtered/mapped direct children (JS array functions aren't recursive after all, except the ES10 `flat` and `flatMap`).

You would just write `app.root.children.forEach` instead of `app.root.forEach`. The developer can just think "forEach/map/filter/find will exhaust the underlying type". If you only want the direct children... pick the array instead.

jQuery messes this simple principle up, mixing `find()`, `filter()` and `children()` and people start to wonder what the actual difference is: https://stackoverflow.com/questions/10378757/difference-between-find-and-filter-in-jquery

But yes, I agree that JavaScript API's have disagreement, its the meme of having 14 standards and you are going to make the 15th to fix them all. I am not sure what will last longer, ES6 or jQuery. jQuery is actually 9 years older and gaining more downloads on NPM month after month.

> Moreover, I don't really want to be the root cause of breaking lots of projects if many functions end up being renamed 😄

It would be another PR and people would actually need to share what their preference would be.

Possible stances I see regarding ES6 naming conventions:

1) Just keep everything as it is, mix and match conventions
2) Gently align naming with ES6 (e.g. a PR which slowly deprecates `GraphNode#findByTag` into `GraphNode#filterByTag`)
3) Hard align with ES6 and tell everyone to immediately fix their projects after update
4) Make use of OOP and just introduce `class GraphNodeES6 extends GraphNode` which follows ES6 naming conventions... every purist nagger like me can of course do that without any change in the engine aswell

And then the entire issue with "include current node in ancestors/parents or children"...

> But at the moment in PlayCanvas, as noticed already, `findByTag` and `findByPath` can't return current node while `find`, `findOne` and `findByName` can.

To make it even stranger, yes, sometimes `GraphNode#findByPath` can return the current node:

```js
const root = app.root.findByPath([]);
```

Maksims:
GraphNode != Array
There is no unified convention between standards/languages on such methods.
Also, including own node in search, with returning a single result - will lead to unwanted cases where in order to ignore own node - need to start from parent/each child. While not checking own node in filter/find methods - gives simpler choice to developer if need to check for self.

Again, GraphNode - is not an array. Its an individual node.
So `node.filter`

NewboO:
> Also, including own node in search, with returning a single result - will lead to unwanted cases where in order to ignore own node - need to start from parent/each child. While not checking own node in filter/find methods - gives simpler choice to developer if need to check for self.

Problem is `find`, `findOne` and `findByName` have been always including current node since their creation many years ago so how not to break anything? Though I agree it must be avoided as much as possible, breaking projects happened in the past so it's always an option.

However, as people never had the choice of ignoring current node, I'm now wondering about the actual usage of the different find functions and in how many cases people are relying on the fact these functions can return self node (or self component). Because maybe the behaviour can be changed without actually breaking many projects but just a few ones? For instance, most people using `findByName` are probably calling it on nodes which can't be returned because its name is never the searched one.

kungfooman:
Removing current node from `GraphNode#find` is going too far, as that behaviour is expected and relied upon.

But removing self from `GraphNode#findAncestors` makes sense to me (after all its called `GraphNode#findAncestors` and `GraphNode#find` isn't called `GraphNode#findDescendants`).

You also need to test your changes against non-trivial projects to find the special cases. SeeMore for example is doing this:

image

image

You can do so by e.g.:

```js
    findByName(name) {
        if (this.name === name) {
            console.log("findByName", this, name);
            debugger;
        }
        return this.findOne('name', name);
    }
```

NewboO:
So let's try another solution by adding a parameter to let the user choose to include self node in the search. It's true by default for `GraphNode.forEach`, `GraphNode.find`, `GraphNode.findOne`, `GraphNode.findByName`, `Entity.findComponent[s]` and `Entity.findScript[s]` and false for all the other so old functions keep their default behaviour.

However, I'm faking overloading so the second parameter of some functions can be skipped when irrelevant (or not used) and I'm not sure if it respects PlayCanvas standards.

kungfooman:
> However, I'm faking overloading so the second parameter of some functions can be skipped when irrelevant (or not used) and I'm not sure if it respects PlayCanvas standards.

I can't speak for others obviously so my personal objection is increased code complexity. I liked to simplify the methods, they are shorter and (at least for me) easier/clearer to read.

`_getIncludeSelf` and `if (typeof query[query.length - 1] === 'boolean') { includeSelf = query.pop(); }` etc. is doing the opposite now.

The initially reduced complexity is back and it is hard to understand/track and just harder than necessary.

You are basically trapping into flag argument code smell, when you just want e.g. `GraphNode#findExcludeSelf`:

- https://medium.com/@amlcurran/clean-code-the-curse-of-a-boolean-parameter-c237a830b7a3
- https://martinfowler.com/bliki/FlagArgument.html

And then the question is: how many people need that method or in which context do you require it?

yaustar:
Speaking from an end user: I think the Entity.findScript and Entity.findScripts are the most useful out of the list and I'm thinking about pulling that out in a separate PR, keep them as consistent as possible with the out find functions (which from this thread seem to be inconsistent as it is).

The find in parents/ancestors would need an engine team call about whether they should be core engine functionality. It's not something I've seen asked about on the forums but wonder how many teams/developers have written their own variant 🤔

LeXXik:
Our team made a similar one to `.findScript`, as some members prefer to access the script instance of a child entity for one reason or another. We don't have "find in parents/ancestors", as we never had a need in such.

kungfooman:
> You can also notice how the current version of `findByTag` is using flag argument, though in an anonymous function (when it actually wasn't before that change 8 months ago).

Good point, a helper function could be:

```js
    findExcludeSelf(attr, value) {
        const results = [];
        const test = _createTest(attr, value);
        this._children.forEach(child => {
            child.forEach((node) => {
                if (test(node))
                    results.push(node);
            });
        });
        return results;
    }
```

And then it just becomes:

```js
    findByTag(...query) {
        return this.findExcludeSelf(node => node.tags.has(...query));
    }
```

1) No flag arguments
2) Reusable
3) Not pushing `checkNode`/`includeSelf` into every stack frame and removing `if (checkNode ...)`
4) Keeps method signatures as they are

> And then we can also wonder why `find` and `findOne` can take either a function or an attribute name and a value, instead of having `findByAttribute` and `findOneByAttribute` for the latter? Most of the complexity comes from that weird overloading in the first place.

I agree, so why create even more overloading? 

> Anyway, I don't see how adding tons of functions for every single variant would help reduce complexity for the user. And that's worse if adding as you're suggesting functions like `map`, `some` and `every` for which one can ask the same btw: how many people need these methods or in which context are they required?

Technically you already used `GraphNode#some` without calling it so:

This code:

```js
    isDescendantOf(node) {
        return !!this.findAncestor(parent => parent === node);
    }
```

Is semantically this:

```js
    isDescendantOf(node) {
        return this.someAncestor(ancestor => ancestor === node);
    }

    someAncestor(attr, value) {
        const test = _createTest(attr, value);
        let parent = this._parent;
        while (parent) {
            if (test(parent))
                return true;
            parent = parent._parent;
        }
        return false;
    }
```

> So now I'm thinking of simply closing this PR and just keep the new functions for myself. The real usefulness relies in the new `Entity` functions I've been using in all of my projects for quite some time but it seems more important to focus on `GraphNode` details which weren't even the point of the PR.

Right, I definitely see the point for your `includeSelf` code and maybe that should be considered in a follow-up PR while this PR refactors and adds new methods and keeping the function signatures as they are (once the signatures are changed, its hard to go back).

> We don't have "find in parents/ancestors", as we never had a need in such.

@LeXXik Just regexing a bit (via `while \([a-zA-Z0-9 &&\.]+\)`) and there is e.g. this code:

https://github.com/playcanvas/engine/blob/b17180c12adabc4a24740aa92e5e25144b7b3fe8/src/framework/components/collision/system.js#L102-L113

This looks like this could be expressed way shorter via `GraphNode#findAncestor`(untested):

```js
                    const isTypeCompound = entity => entity.collision && entity.collision.type === 'compound';
                    component._compoundParent = entity.findAncestor(isTypeCompound);
```

LeXXik:
@kungfooman ah, I meant in our team's custom code, not the engine 😅

NewboO:
So, last try to make everyone happy. I considered `find`, `findOne` and `forEach` as legacy, so the only changes in them are now the refactoring.

There are now `findDescendant[s]`, `findAncestor[s]`, `forEachDescendant` and `forEachAncestor` for traversing tree except the current node, and most importantly all of the `findComponent[s]*` and `findScript[s]*` helpers in `Entity` deriving from them.

I kept `findByName` and `findByTag` refactorings but because one is including current node while the other don't, I'm not adding any descendant/ancestor variant. If anyone want to add other `GraphNode` functions, then it shoud be done in another PR to keep this one focused on the main `GraphNode` find functions and the new `Entity` component/script finding functions.

kungfooman:
Thank you for the changes @NewboO !

I had no time to test it yet, just looking it over. Two notes:

- `_createTest` is missing type information through JSDoc, a nice example is e.g.:
https://github.com/playcanvas/engine/blob/b17180c12adabc4a24740aa92e5e25144b7b3fe8/src/framework/components/element/text-element.js#L50-L71

JSDoc that should work:

```
/**
 * Helper function that handles signature overloading to receive a test function.
 *
 * @param {Function | string} attr - Attribute or lambda.
 * @param {*} [value] - Optional value in case of `attr` being a `string`
 * @returns {Function} Test function that receives a GraphNode and returns a boolean.
 * @ignore
 */
```

I would also remove the underscore, it is private by default if you don't export it in in ES6 file scope, but I've seen it with and without underscore, so it doesn't really matter I guess.

- I don't see your Contributor License Agreement:

> I confirm I have read the [contributing guidelines](https://github.com/playcanvas/engine/blob/master/.github/CONTRIBUTING.md) and signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

src/scene/graph-node.js
```diff
@@ -24,18 +24,41 @@ const target = new Vec3();
 const up = new Vec3();
 
 /**
- * Callback used by {@link GraphNode#find} and {@link GraphNode#findOne} to search through a graph
- * node and all of its descendants.
+ * Helper function that handles signature overloading to receive a test function.
+ *
+ * @param {Function|string} attr - Attribute or lambda.
+ * @param {*} [value] - Optional value in case of `attr` being a `string`
+ * @returns {Function} Test function that receives a GraphNode and returns a boolean.
```
kungfooman, line 31:
```suggestion
 * @param {FindNodeCallback|string} attr - Attribute or lambda.
 * @param {*} [value] - Optional value in case of `attr` being a `string`
 * @returns {FindNodeCallback} Test function that receives a GraphNode and returns a boolean.
```

Just realized that the callback type is already defined and we can just reuse it here to make it more explicit

mvaligursky:
We had a quick discussion, and agreed it's a great PR from many points of view, but it's so large, including the follow up discussion, that it's becoming hard to continue the work on this.

The suggestion is to break it down into small issues, addressing individual improvements / new APIs. The discussion should take place on those separate issues, and a relevant PR should be created when an agreement is reached there.

If there are some optimizations to existing done here for example, it'd be awesome to move those to a separate PR, as that would be valuable to have reviewed and merged.

NewboO:
Finally took time to break down this PR.

So 2 new PR :
#5850 for the optimization
#5851 for the most consensual new feature

And a discussion #5852 for the main debate about ancestors/descendants.

PR #4610, opened 2022-09-01
Added CC-BY-4.0 licences to asset viewer example

yaustar:


mvaligursky:
you could also do it this way, to be more in line with other models we have?
https://github.com/playcanvas/engine/blob/main/examples/assets/models/chess-board.txt

yaustar:
I would have thought we need to add it where it's being used?

mvaligursky:
Not sure. This applies and does not specifies it: https://creativecommons.org/licenses/by/4.0/
I always though it better to be stored next to the file we distribute, rather than elsewhere where its referenced.

yaustar:
I change it to be consistent to be the chess board for now. We can change it all later

yaustar:
@mvaligursky Added text files for the licenses

mvaligursky (approved):

PR #4612, opened 2022-09-02
Fix AppOption types

slimbuck:
This PR updates AppOption types to class constructors (instead of class instances), which appears to appease typescript.

VSCode shows errors on these now, not sure how to fix that.


mvaligursky:
I tried few variations of these (not this one yet), and something always does not work.

src/framework/app-options.js
```diff
@@ -71,49 +71,49 @@ class AppOptions {
...
-     * @type {SoundManager}
+     * @type {SoundManager.constructor}
      */
     soundManager;
 
     /**
      * The graphics device.
      *
-     * @type {GraphicsDevice}
+     * @type {GraphicsDevice.constructor}
```
mvaligursky, line 81: this one we pass an instance, and possibly others, not all we pass as types.

slimbuck: ahh yes of course, updated.

slimbuck:
With the latest update playcanvas.d.ts looks as follows:
```javascript
    lightmapper: any;
    /**
     * The BatchManager.
     *
     * @type {typeof BatchManager}
     */
    batchManager: any;
    /**
     * The XrManager.
     *
     * @type {typeof XrManager}
     */
    xr: any;
    /**
     * The component systems the app requires.
     *
     * @type {(typeof ComponentSystem)[]}
     */
    componentSystems: (typeof ComponentSystem)[];
    /**
     * The resource handlers the app requires.
     *
     * @type {(typeof ResourceHandler)[]}
     */
    resourceHandlers: (typeof ResourceHandler)[];

```

slimbuck:
resource handler is being recast as an interface in types-fixup.js:
```javascript
dts = dts.replace('export class ResourceHandler', 'export interface ResourceHandler');
```

@willeastcott any idea why this is done?

mvaligursky:
also, try to build API docs from jsdocs locally, I think I had issues with this in the past as well.

willeastcott:
Yep, I have a feeling the docs build process fails on those typeof keywords, IIRC.

slimbuck:
The latest version (with constructor syntax) seems to work fine when compiling typescript and also generates docs.

CI fails though.

mvaligursky:
related https://github.com/playcanvas/engine/pull/4324

PR #4617, opened 2022-09-05
Use better defaults for physically based camera parameters

GSterbrant:
### Description
This is to make sure that shutter speed and aperture won't have too big of an impact on depth of field and motion blur, given the eventual future where we might want to have those effect depend on the camera parameters.

### DIFF
--- a/src/framework/components/camera/component.js
+++ b/src/framework/components/camera/component.js
@@ -392,7 +392,7 @@ class CameraComponent extends Component {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


slimbuck (approved):

mvaligursky (approved):
looks like really sensible defaults.

iso 1000 is no problem on modern cameras .. you mostly care about aperture and shutter to control DOF and blur .. and f16 and 1/1000 sounds like a good option that matches our visuals currently (everything is sharp, and no motion blur)

PR #4618, opened 2022-09-05
Physically based light units

GSterbrant:
### Description
Enables the use of physically based light units, by having the physicalUnits enabled on the scene, and setting the luminance (not intensity) parameter on the lights. For spot lights, this also means the strength of the light is dependent on the aperture (outer cone angle), meaning a narrower angle will concentrate its energy on a smaller surface, thus making that area brighter than if it was using a wide angle. 

Also contains an example scene showing off physically based light and camera units in a typical scene with a strong sun, sky and a bunch of local lights, called Physical Light Units. 

Also fixes support for KHR_punctual_lights where we previously clamped the light intensity. 

https://user-images.githubusercontent.com/107400752/188444600-e33f3b28-ec5e-4957-9af6-b04385188f50.mp4

https://user-images.githubusercontent.com/107400752/188445670-17451c9f-35cb-452c-b53f-32888acadabf.mp4


examples/src/examples/graphics/physical-light-units.tsx
```diff
@@ -0,0 +1,303 @@
+import React from 'react';
+import * as pc from '../../../../';
+
+import { BindingTwoWay } from '@playcanvas/pcui';
+import { BooleanInput, LabelGroup, Panel, SliderInput } from '@playcanvas/pcui/react';
+import { Observer } from '@playcanvas/observer';
+class PhysicalLightUnitsExample {
+    static CATEGORY = 'Graphics';
+    static NAME = 'Physical Light Units';
```
mvaligursky, line 9: maybe call it 'Light physical' to show them next to 'Lights' example?

GSterbrant: Done.

src/scene/light.js
```diff
@@ -698,7 +714,17 @@ class Light {
         const g = color.g;
         const b = color.b;
 
-        const i = this._intensity;
+        let i = this._intensity;
+        if (this._scene?.physicalUnits) {
+            if (this._type === LIGHTTYPE_SPOT) {
+                const angleAsRadians = this._outerConeAngle * Math.PI / 180.0;
+                i = this._luminance / (2 * Math.PI * (1 - Math.cos(angleAsRadians / 2.0)));
```
querielo, line 721: I'm not sure but I think we have to use "_innerConeAngle" here.

Maybe `i = this._luminance / (2 * Math.PI * (1 - (this._innerConeAngleCos + this._outerConeAngleCos)  / 2.0)));`. 
Probably, it is wrong.

And I'm just curious why do we use `Math.cos(angleAsRadians / 2.0)` and not `Math.cos(angleAsRadians) / 2.0`

GSterbrant: Ah, so the outer cone angle controls how much energy should transmit through the light volume, such that we can convert the light in lm to lm/m^2, so it would only make sense to use the outer angle as the inner angle is only used to feather the edge of the light. 

As for the math, I would defer to the [filament documentation](https://google.github.io/filament/Filament.md.html#lighting/units/lightunitsvalidation:~:text=light%27s%20direction%20vector-,Equation,-60).

querielo: It sounds a little strange to me. Energy cannot be equal for **innerConeAngle === 0** and **innerConeAngle === outerConeAngle**. Filament evaluates this as if the spot light has uniformly distributed intensity.
Screen Shot 2022-09-07 at 00 20 24

It may be true from 0 to **innerConeAngle**. But  between **innerConeAngle** and **outerConeAngle** we have the angular falloff attenuation (`(cos(angle) - _outerConeAngleCos) / (_innerConeAngle - _outerConeAngleCos)` or something like this). So, it looks like multiplier has to use linear combination of **cos(innerConeAngle)** and **cos(outerConeAngle)**.

querielo: This is what part uses: https://github.com/mmp/pbrt-v4/blob/faac34d1a0ebd24928828fe9fa65b65f7efc5937/src/pbrt/lights.cpp#L1463

querielo: This is what part uses: https://github.com/mmp/pbrt-v4/blob/faac34d1a0ebd24928828fe9fa65b65f7efc5937/src/pbrt/lights.cpp#L1463

MAG-AdrianMeredith:
This is fantastic!  As a light unit layman I noticed that punctual lights uses candela (lm/sr) but we're using lm here as the unit type.  is that correct? what needs to be done as a user to get consistent results?

querielo (approved):

src/scene/light.js
GSterbrant: When thinking about this a bit, the inner angle basically defines a border outside of which the spot light cone is in the shape of solid angle, where no inner == outer means we have a perfect cone and inner == 0 means we have a perfect solid angle. If we take that into account, the amount of energy in the volume is dependent on the inner angle, where there is a bit of energy lost if the inner angle is less than the outer. 

In these illustrations I use a hemisphere to indicate the solid angle cutout of the spot light. I hope they make sense.

Solid angle when outer == inner
image

Solid angle when inner == 0.
image

Solid angle when inner > 0 && inner < outer
image

So we can definitely look into accounting for that energy loss, I don't think it makes a ton of difference but it's definitely worth looking into.

GSterbrant:
@MAG-AdrianMeredith We support point lights in lumen. However, you can always write a converter candela -> lumen and feed that value instead. I thought we might want to write a few converters as well, so you'd be able to basically feed whatever value you'd like and you'd get back a value with the units we are using, would that be useful?

src/scene/light.js
querielo: If we assume that the angular falloff attenuation is defined as `(cos(angle) - _outerConeAngleCos) / (_innerConeAngle - _outerConeAngleCos)`, then the complete integral should look like [this](https://www.wolframalpha.com/input?i2d=true&i=+phi+%3D+2*pi*%5C%2840%29Integrate%5BI%2C%7Bx%2Ccos%5C%2840%29a%5C%2841%29%2C1%7D%5D+%2BIntegrate%5BDivide%5B%5C%2840%29cos%5C%2840%29x%5C%2841%29+-+cos%5C%2840%29a%5C%2841%29%5C%2841%29%2C%5C%2840%29cos%5C%2840%29b%5C%2841%29+-+cos%5C%2840%29a%5C%2841%29%5C%2841%29%5D*+I%2C%7Bx%2Ccos%5C%2840%29b%5C%2841%29%2Ccos%5C%2840%29a%5C%2841%29%7D%5D%5C%2841%29)

Where
* **_a_** is `this._innerConeAngle * Math.PI / 360.0`
* **_b_** is `this._outerConeAngle * Math.PI / 360.0`,
* **the first integral** is about everything that inside the inner cone,
* **the second integral** is about everything that between the inner cone and the outer cone

Screen Shot 2022-09-10 at 03 20 25

And, probably, the final result is something like this

```js
i = this._luminance / (
    2 * Math.PI * I * (
        1 + (
            Math.sin(Math.cos(b)) - Math.sin(Math.cos(a))
        ) / (
            Math.cos(a) - Math.cos(b)
        )
    )
);
```

But, I may miss something, so it is better to check it.

querielo: I constructed two plots for two angular falloff attenuation models.

1) [Cosine attenuation](https://www.desmos.com/calculator/tw4hncxrx0?lang=ru) + [smoothstep](https://github.com/playcanvas/engine/blob/6424d6786c5f78853221b7177dc9da887fda410b/src/graphics/program-lib/chunks/lit/frag/spot.js#L4): `(cos(angle) - _outerConeAngleCos) / (_innerConeAngleCos - _outerConeAngleCos)`
Screen Shot 2022-09-10 at 15 11 37

2) [Linear attenuation](https://www.desmos.com/calculator/tzuisqed11?lang=ru)  + [smoothstep](https://github.com/playcanvas/engine/blob/6424d6786c5f78853221b7177dc9da887fda410b/src/graphics/program-lib/chunks/lit/frag/spot.js#L4): `(angle- _outerConeAngle) / (_innerConeAngle - _outerConeAngle)`
Screen Shot 2022-09-10 at 15 11 33

Dash lines show luminous intensity factor. As you can see on the second model, if we shift the left side on the right and sum area with "attenuations", then we get constant 1. And this means that combinations of two "attenuations" equal area inside innerConeAngle. So for linear attenuation we get exactly what is used in PBR book:

```js
const innerAngleAsRadians = this._innerConeAngle * Math.PI / 180.0;
const outerAngleAsRadians = this._outerConeAngle * Math.PI / 180.0;
i = this._luminance / (
    2 * Math.PI * (
        1 - Math.cos(innerAngleAsRadians) + (Math.cos(innerAngleAsRadians) -  Math.cos(outerAngleAsRadians)) / 2
    )
);
```

MAG-AdrianMeredith:
Thanks for the update, I'm just wondering what the reason is for not following gltf's definition of lights seems to be a pretty common definition for light intensities https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_lights_punctual/README.md

src/scene/light.js
querielo: @GSterbrant
I don't know if it is correct. So, [here is an issue in FIlament](https://github.com/google/filament/issues/6060). Since, they don't use the **[smoothstep](https://github.com/playcanvas/engine/blob/6424d6786c5f78853221b7177dc9da887fda410b/src/graphics/program-lib/chunks/lit/frag/spot.js#L4)** function their falloff attenuation function is linear.

GSterbrant:
@MAG-AdrianMeredith Ah, so from my understanding, glTF provides light intensities in candela because it represents luminous power per solid angle. From a user perspective, using candela makes little sense since you set the luminous power and either have a point/omni light, for which the angle is a fixed 4 PI, or use a spot light with explicit cone angles. 

However, in the glTF, the lights angles are already known, so they can pre calculate that value and therefore provide the implementor with those fixed values. I hope that makes sense 🙂 .

GSterbrant:
@querielo I implemented your suggestion, huge thanks for all the legwork there! To test it, I used a spot light with a 0 inner angle and 90 degree outer angle, which when having the same luminous power as 1/4 of a point light gave identical light intensity, which makes a lot of sense since a point light could be conceptualised as having 4 x 90 degree spot lights in terms of energy per square meter.

MAG-AdrianMeredith:
@GSterbrant So its just a terminology thing and they should line up (given similar camera setup) great thats what I thought.

MAG-AdrianMeredith:
One question, how does IBL and lightmaps fit into this equation?  I'm currently testing against this PR in a fork (I've been tasking with supporting physical light units for a project if you haven't noticed already ;) ).  Currently enabling this plunges everything into complete darkness (including the gltf_punctual_lights) and boosting the camera settings produces a strange result where it looks like only the specular component is lit.  

I'm assuming this is just because its not done yet?

p.s. great work as always everyone

GSterbrant:
@MAG-AdrianMeredith In this PR, I would suggest the lights-physical-units example which shows off a scene using the GLB with punctual lights and all that. That GLB does have very weak lights, you need to reduce the strength of the IBL, sun, spot, point and area light and turn the exposure settings up to see it. 

Would it be possible to post a picture showing off the issues you're seeing?

MAG-AdrianMeredith:
It seems to me that the default camera sensitivity is two orders of magnitude out from what you'd expect.  From what I can see a really bright bulb is around 1000 lumens.  in the default settings this wont even cast any light and the example is set to 100000 and is still really dim.

lumens-bltdirect
Light-Output-Lumens-Chart-Meant2be-Co-1

GSterbrant:
@MAG-AdrianMeredith The default camera settings are adjusted for outdoors lighting like the sun and sky. In proportion to the sun at around 100k-120k lm/m^2, a 100W light bulb is something like 1100/4*PI = 91.6 lm/m^2 given no attenuation, so it makes sense that those settings are poorly adjusted for indoor lighting. 

I found this link giving a list of approximate lux (lm/m^2) readings: https://en.wikipedia.org/wiki/Lux.

GSterbrant:
@MAG-AdrianMeredith So we will definitely need to implement automatic exposure correction to adjust for what I just mentioned. That's going to be a separate PR though 🙂, for which we created this issue: #4650.

examples/src/examples/graphics/light-physical-units.tsx
```diff
@@ -0,0 +1,303 @@
...
+            material.shininess = 80;
+            material.useMetalness = true;
+
+            material.diffuseMapTiling.set(17, 17);
+            material.normalMapTiling.set(17, 17);
+            material.glossMapTiling.set(17, 17);
+            material.update();
+
+            const plane = new pc.Entity();
+            plane.addComponent('model', {
```
mvaligursky, line 130: please change 'model' to 'render'

src/scene/light.js
```diff
@@ -393,6 +394,9 @@ class Light {
 
         this._outerConeAngle = value;
         this._outerConeAngleCos = Math.cos(value * Math.PI / 180);
+        if (this._scene?.physicalUnits) {
+            this._updateFinalColor();
+        }
```
mvaligursky, line 399: one possible issue here is if this property is set first, and then the scene is switched to use physical units. Or in general, if scene is switched to use physical units after the lights were already created. But perhaps that's not a common case, as a solution is not straightforward.

mvaligursky:
Just a note here, that the clustered lights store light colors in 2 bytes per channel (so not full floats). So for physical lights, where the color range is really large, this would mean a pretty low precision. But I suspect in reality this could be just fine I suspect.

src/scene/light.js
mvaligursky: That's a tricky one to do, the scene is created by the application and not the user. We could do automatic job with some 'version'  property on the light .. which would be compared each frame the the 'version' stored on the scene. If scene.physicalUnits changes, we'd increment the version on the scene. When culling lights during rendering, if the versions don't match, we'd evaluate final color again. But it's not very nice solution.

mvaligursky: actually. Maybe the light can have internal flag 'usesPhysicalUnits' or something similar. During rendering, inside forwardRenderer.cullLights, when the light is visible, and its usesPhysicalUnits does not match scene.physicalUnits, we'd force evaluate the light's final color.

src/scene/light.js
```diff
@@ -152,6 +153,8 @@ class Light {
         this._innerConeAngleCos = Math.cos(this._innerConeAngle * Math.PI / 180);
         this._outerConeAngleCos = Math.cos(this._outerConeAngle * Math.PI / 180);
 
+        this._usesPhysicalUnits = false;
```
mvaligursky, line 156: I'd set this to undefined, so that it's not false nor true .. otherwise the value needs to be in sync with the initial setting in the scene.

src/scene/renderer/forward-renderer.js
```diff
@@ -929,6 +929,12 @@ class ForwardRenderer {
 
             if (light.enabled) {
 
+                // make sure the lights update their final color if the scene
+                // use of physical units mismatch the lights internal flag
+                if (this.scene.physicalUnits !== light._usesPhysicalUnits) {
+                    light._updateFinalColor();
+                    light._usesPhysicalUnits = this.scene.physicalUnits;
+                }
```
mvaligursky, line 937: I would move this to a function on a Light class, for example setPhysicalUnits(boolean), and call it not from here, but few lines below from two places where we set `light.visibleThisFrame = true;` - so only for the visible lights.

src/framework/components/light/component.js
```diff
@@ -58,6 +58,7 @@ const _lightPropsDefault = [];
  * @property {Color} color The Color of the light. The alpha component of the color is ignored.
  * Defaults to white (1, 1, 1).
  * @property {number} intensity The brightness of the light. Defaults to 1.
+ * @property {number} luminance The physically based luminance. Defaults to 0.
```
mvaligursky, line 61: perhaps add a note that this is used when scene.physicalUnits is enabled.

src/scene/renderer/forward-renderer.js
```diff
@@ -576,7 +576,11 @@ class ForwardRenderer {
         this.ambientId.setValue(this.ambientColor);
 
         if (scene.sky) {
-            this.skyboxIntensityId.setValue(scene.skyboxIntensity);
+            if (scene.physicalUnits) {
+                this.skyboxIntensityId.setValue(scene.skyboxLuminance);
+            } else {
+                this.skyboxIntensityId.setValue(scene.skyboxIntensity);
+            }
```
mvaligursky, line 583:
```suggestion
            this.skyboxIntensityId.setValue(scene.physicalUnits ? scene.skyboxLuminance :  scene.skyboxIntensity);
```

src/scene/renderer/forward-renderer.js
```diff
@@ -930,12 +934,12 @@ class ForwardRenderer {
             const light = lights[i];
 
             if (light.enabled) {
-
                 // directional lights are marked visible at the start of the frame
                 if (light._type !== LIGHTTYPE_DIRECTIONAL) {
                     light.getBoundingSphere(tempSphere);
                     if (camera.frustum.containsSphere(tempSphere)) {
                         light.visibleThisFrame = true;
+                        light.usePhysicalUnits = this.scene.physicalUnits;
```
mvaligursky, line 942: add `const scenePhysicalUnits = this.scene.physicalUnits;` outside of the loop, and use it inside the loop to avoid class member lookups.

src/scene/scene.js
```diff
@@ -549,7 +550,8 @@ class Scene extends EventHandler {
     }
 
     /**
-     * Multiplier for skybox intensity. Defaults to 1.
+     * Multiplier for skybox intensity. Defaults to 1. Unused if
+     * physical units are used
```
mvaligursky, line 554:
```suggestion
     * physical units are used.
```

src/scene/scene.js
```diff
@@ -564,6 +566,23 @@ class Scene extends EventHandler {
         return this._skyboxIntensity;
     }
 
+    /**
+     * Luminance (in lm/m^2) of skybox. Defaults to 0. Only used if
+     * physical units are used
```
mvaligursky, line 571:
```suggestion
     * physical units are used.
```

mvaligursky:
I added few minor comments. And also:
- search for skyboxIntensity in app-base.js and apply the same to skyboxLuminosity
- in the scene, we have skyboxLuminosity. Do we need some physical treatment of the ambientLight, which is used when the skydome is not set up? ambientLightLuminosity?

mvaligursky:
One more suggestion. It'd be great to check if the lightmapping tech works with physical lights. Perhaps locally modify the lights-baked-a-o example to use physical lights and see if that works correctly.

src/framework/app-base.js
```diff
@@ -1528,6 +1528,7 @@ class AppBase extends EventHandler {
      * @param {number|null} [settings.render.skybox] - The asset ID of the cube map texture to be
      * used as the scene's skybox. Defaults to null.
      * @param {number} settings.render.skyboxIntensity - Multiplier for skybox intensity.
+     * @param {number} settings.render.skyboxLuminance - Lux (lm/m^2) value for skybox intensity when physical light units are enabled.
```
willeastcott, line 1531: Worth adding the default value?

src/scene/scene.js
```diff
@@ -564,6 +573,23 @@ class Scene extends EventHandler {
...
```
willeastcott, line 578:
```suggestion
     * Luminance (in lm/m^2) of skybox. Defaults to 0. Only used if physical units are used.
```

src/scene/scene.js
```diff
@@ -549,7 +557,8 @@ class Scene extends EventHandler {
     }
 
     /**
-     * Multiplier for skybox intensity. Defaults to 1.
+     * Multiplier for skybox intensity. Defaults to 1. Unused if
+     * physical units are used.
```
willeastcott, line 561:
```suggestion
     * Multiplier for skybox intensity. Defaults to 1. Unused if physical units are used.
```

willeastcott:
This is soooo cool. Can you be explicit about what new API is introduced in the PR description, please?

willeastcott (approved):

PR #4619, opened 2022-09-05
Fix sheen glossiness and colour not getting applied correctly

GSterbrant:
### Description
Fixes a bunch of issues with sheen. The first of which is that due to glossiness being roughness from glTF, we can't optimise the value out sheen glossiness tinting if the value is 1.0. 

Furthermore, it turns out the sheen textures weren't even applied, so this PR fixes that as well.

### DIFF
--- a/src/scene/materials/standard-material-options-builder.js
+++ b/src/scene/materials/standard-material-options-builder.js
@@ -173,8 +173,8 @@ class StandardMaterialOptionsBuilder {
         options.refractionIndexTint = (stdMat.refractionIndex !== 1.0 / 1.5) ? 1 : 0;
         options.thicknessTint = (stdMat.useDynamicRefraction && stdMat.thickness !== 1.0) ? 1 : 0;
         options.useMetalness = stdMat.useMetalness;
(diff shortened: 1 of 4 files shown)


willeastcott (approved):

mvaligursky (approved):

src/scene/materials/standard-material-options-builder.js
```diff
@@ -174,6 +174,7 @@ class StandardMaterialOptionsBuilder {
         options.thicknessTint = (stdMat.useDynamicRefraction && stdMat.thickness !== 1.0) ? 1 : 0;
         options.useMetalness = stdMat.useMetalness;
         options.specularEncoding = stdMat.specularEncoding === undefined ? 'linear' : stdMat.specularEncoding;
+        options.sheenEncoding = stdMat.sheenEncoding === undefined ? 'linear' : stdMat.sheenEncoding;
```
slimbuck, line 177:
```suggestion
        options.sheenEncoding = stdMat.sheenEncoding ||  'linear';
```

slimbuck (approved):

PR #4621, opened 2022-09-06
Variant bugfix

slimbuck:
If no variant name was supplied, model were being assigned default materials.


mvaligursky:
That was by design, as we need a way to assign default materials.

slimbuck:
> That was by design, as we need a way to assign default materials.

What is the use of assigning engine default materials to a model?

slimbuck:
cc @GSterbrant

slimbuck:
actually the issue is elsewhere in the variant code, closing.

PR #4626, opened 2022-09-08
Added changeScene API

yaustar:
### Updated: Tue 20 Sep 2022

This is to address the many topics for switching scene in PlayCanvas. The page and helper script I wrote doesn't seem to be enough https://developer.playcanvas.com/en/user-manual/packs/loading-scenes/

So I'm adding the most common use case to the API.

Changing scenes becomes a one liner with this change where it loads the scene data, destroys all entities under the `app.root` and adds the scene settings and hierarchy:

```
this.app.scenes.changeScene('name of scene');
```

This PR also adds support for loading scene settings/hierarchy by name, again reducing the extra boilerplate that the user needs to write.

### DIFF
--- a/src/framework/scene-registry.js
+++ b/src/framework/scene-registry.js
@@ -177,7 +177,7 @@ class SceneRegistry {
         }
 
         if (!sceneItem.url) {
-            callback("Cannot find scene when loading a scene");
+            callback("Cannot find scene to load");
             return;
         }
 
--- a/src/framework/scene-registry.js
+++ b/src/framework/scene-registry.js
@@ -291,6 +291,23 @@ class SceneRegistry {
         app._preloadScripts(data, _loaded);
     }
 
+    _loadSceneHierarchy(sceneItem, onBeforeAddHierarchy, callback) {
+        this._loadSceneData(sceneItem, false, (err, sceneItem) => {
+            if (err) {
+                if (callback) {
+                    callback(err);
+                }
+                return;
+            }
+
+            if (onBeforeAddHierarchy) {
+                onBeforeAddHierarchy(sceneItem);
+            }
+
+            this._addHierarchyToScene(sceneItem.url, sceneItem.data, callback);
+        });
+    }
+
     /**
      * Load a scene file, create and initialize the Entity hierarchy and add the hierarchy to the
      * application root Entity.
@@ -310,16 +327,7 @@ class SceneRegistry {
      * });
      */
     loadSceneHierarchy(sceneItem, callback) {
-        const self = this;
-
-        this._loadSceneData(sceneItem, false, (err, sceneItem) => {
-            if (err) {
-                if (callback) callback(err);
-                return;
-            }
-
-            self._addHierarchyToScene(sceneItem.url, sceneItem.data, callback);
-        });
+        this._loadSceneHierarchy(sceneItem, callback);
     }
 
     /**
@@ -375,17 +383,8 @@ class SceneRegistry {
      */
     changeScene(sceneItem, callback) {
         const app = this._app;
-        const self = this;
 
-        this._loadSceneData(sceneItem, false, (err, sceneItem) => {
-            if (err) {
-                if (callback) {
-                    callback(err);
-                }
-
-                return;
-            }
-
+        const onBeforeAddHierarchy = (sceneItem) => {
             // Destroy/Remove all nodes on the app.root
             const rootChildren = app.root.children;
             while (rootChildren.length > 0) {
@@ -395,8 +394,9 @@ class SceneRegistry {
             }
 
             app.applySceneSettings(sceneItem.data.settings);
-            self._addHierarchyToScene(sceneItem.url, sceneItem.data, callback);
-        });
+        };
+
+        this._loadSceneHierarchy(sceneItem, onBeforeAddHierarchy, callback);
     }
 
     /**
@@ -410,7 +410,6 @@ class SceneRegistry {
      */
     loadScene(url, callback) {
         const app = this._app;
-        const self = this;
 
         const handler = app.loader.getHandler("scene");
 
@@ -427,7 +426,7 @@ class SceneRegistry {
                     const scene = handler.open(url, data);
 
                     // Cache the data as we are loading via URL only
-                    const sceneItem = self.findByUrl(url);
+                    const sceneItem = this.findByUrl(url);
                     if (sceneItem && !sceneItem.loaded) {
                         sceneItem.data = data;
                     }
--- a/src/framework/scene-registry.js
+++ b/src/framework/scene-registry.js
@@ -312,13 +312,15 @@ class SceneRegistry {
      * });
      */
     loadSceneHierarchy(sceneItem, callback) {
+        const self = this;
+
         this._loadSceneData(sceneItem, false, function (err, sceneItem) {
             if (err) {
                 if (callback) callback(err);
                 return;
             }
 
-            this._addHierarchyToScene(sceneItem.url, sceneItem.data, callback);
+            self._addHierarchyToScene(sceneItem.url, sceneItem.data, callback);
         });
     }
 
--- a/src/framework/scene-registry.js
+++ b/src/framework/scene-registry.js
@@ -327,7 +327,7 @@ class SceneRegistry {
      * });
      */
     loadSceneHierarchy(sceneItem, callback) {
-        this._loadSceneHierarchy(sceneItem, callback);
+        this._loadSceneHierarchy(sceneItem, null, callback);
     }
 
     /**
--- a/src/framework/scene-registry.js
+++ b/src/framework/scene-registry.js
@@ -175,7 +175,7 @@ class SceneRegistry {
 
         url = sceneItem.url;
 
-        if (!sceneItem.url) {
+        if (!url) {
             callback("Cannot find scene to load");
             return;
         }
--- a/src/framework/scene-registry.js
+++ b/src/framework/scene-registry.js
@@ -367,7 +367,7 @@ class SceneRegistry {
      * @param {ChangeSceneCallback} [callback] - The function to call after loading,
      * passed (err, entity) where err is null if no errors occurred.
      * @example
-     * app.scenes.ChangeScene("Scene Name", function (err, entity) {
+     * app.scenes.changeScene("Scene Name", function (err, entity) {
      *     if (!err) {
      *         // success
      *     } else {
--- a/src/framework/scene-registry.js
+++ b/src/framework/scene-registry.js
@@ -4,7 +4,6 @@ import { Debug } from '../core/debug.js';
 import { ABSOLUTE_URL } from '../asset/constants.js';
 
 import { SceneRegistryItem } from './scene-registry-item.js';
-import { Entity } from './entity.js';
 
 /** @typedef {import('./app-base.js').AppBase} AppBase */
 
@@ -395,11 +394,8 @@ class SceneRegistry {
             const rootChildren = self._app.root.children;
             while (rootChildren.length > 0) {
                 const child = rootChildren[0];
-                if (child instanceof Entity) {
-                    child.destroy();
-                } else {
-                    child.parent.removeChild(child);
-                }
+                child.reparent(null);
+                child.destroy?.();
             }
 
             self._app.applySceneSettings(sceneItem.data.settings);
--- a/src/framework/scene-registry.js
+++ b/src/framework/scene-registry.js
@@ -382,6 +382,7 @@ class SceneRegistry {
         }
 
         const app = this._app;
+        const self = this;
 
         this._loadSceneData(sceneItem, false, function (err, sceneItem) {
             if (err) {
--- a/src/framework/scene-registry.js
+++ b/src/framework/scene-registry.js
@@ -188,14 +188,14 @@ class SceneRegistry {
             callback(null, sceneItem);
             return;
         }
-        
+
         // include asset prefix if present
         if (app.assets && app.assets.prefix && !ABSOLUTE_URL.test(url)) {
             url = path.join(app.assets.prefix, url);
         }
-        
+
         sceneItem._onLoadedCallbacks.push(callback);
-        
+
         if (!sceneItem._loading) {
             // Because we need to load scripts before we instance the hierarchy (i.e. before we create script components)
             // Split loading into load and open
@@ -264,7 +264,7 @@ class SceneRegistry {
 
     _addHierarchyToScene(url, data, callback) {
         const app = this._app;
-        
+
         // called after scripts are preloaded
         const _loaded = function () {
             // Because we need to load scripts before we instance the hierarchy (i.e. before we create script components)
--- a/src/framework/scene-registry.js
+++ b/src/framework/scene-registry.js
@@ -158,6 +158,7 @@ class SceneRegistry {
     // This allows us to retain expected behavior of loadSceneSettings and loadSceneHierarchy where they
     // don't store loaded data which may be undesired behavior with projects that have many scenes.
     _loadSceneData(sceneItem, storeInCache, callback) {
+        const app = this._app;
         // If it's a sceneItem, we want to be able to cache the data
         // that is loaded so we don't do a subsequent http requests
         // on the same scene later
@@ -186,19 +187,19 @@ class SceneRegistry {
             callback(null, sceneItem);
             return;
         }
-
-        // Because we need to load scripts before we instance the hierarchy (i.e. before we create script components)
-        // Split loading into load and open
-        const handler = this._app.loader.getHandler("hierarchy");
-
+        
         // include asset prefix if present
-        if (this._app.assets && this._app.assets.prefix && !ABSOLUTE_URL.test(url)) {
-            url = path.join(this._app.assets.prefix, url);
+        if (app.assets && app.assets.prefix && !ABSOLUTE_URL.test(url)) {
+            url = path.join(app.assets.prefix, url);
         }
-
+        
         sceneItem._onLoadedCallbacks.push(callback);
-
+        
         if (!sceneItem._loading) {
+            // Because we need to load scripts before we instance the hierarchy (i.e. before we create script components)
+            // Split loading into load and open
+            const handler = app.loader.getHandler("hierarchy");
+
             handler.load(url, function (err, data) {
                 sceneItem.data = data;
                 sceneItem._loading = false;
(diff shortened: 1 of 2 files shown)


src/framework/scene-registry.js
```diff
@@ -347,6 +358,55 @@ class SceneRegistry {
...
+            // Destroy/Remove all nodes on the app.root
+            const rootChildren = self._app.root.children;
+            while (rootChildren.length > 0) {
+                const child = rootChildren[0];
+                if (child instanceof Entity) {
+                    child.destroy();
+                } else {
+                    child.parent.removeChild(child);
+                }
+            }
```
yaustar, line 403: Is there a nicer way to do this? This feels pretty bad?

mvaligursky: this should work
```
while (rootChildren.length > 0) {
    const child = rootChildren[0];
    child.reparent(null);
    child.destroy?.();
}
```

yaustar:
Project to test this PR with : https://playcanvas.com/project/979499

src/framework/scene-registry.js
```diff
@@ -6,7 +6,6 @@ import { ABSOLUTE_URL } from '../asset/constants.js';
 import { SceneRegistryItem } from './scene-registry-item.js';
 
 /** @typedef {import('./app-base.js').AppBase} AppBase */
-/** @typedef {import('./entity.js').Entity} Entity */
```
yaustar, line 9: Note to self: This line needs to be re-added

src/framework/scene-registry.js
```diff
@@ -170,7 +177,7 @@ class SceneRegistry {
         }
 
         if (!sceneItem.url) {
-            callback("URL or SceneRegistryItem is null when loading a scene");
+            callback("Cannot find scene when loading a scene");
```
yaustar, line 180:
```suggestion
            callback("Cannot find scene to load");
```

src/framework/scene-registry.js
```diff
@@ -253,6 +261,38 @@ class SceneRegistry {
...
+            self._app.systems.fire('initialize', entity);
+            self._app.systems.fire('postInitialize', entity);
+            self._app.systems.fire('postPostInitialize', entity);
+
+            if (callback) callback(null, entity);
+        };
+
+        // load priority and referenced scripts before opening scene
+        self._app._preloadScripts(data, _loaded);
+    }
```
kungfooman, line 294:
```suggestion
    _addHierarchyToScene(url, data, callback) {
        const app = this._app;

        // Because we need to load scripts before we instance the hierarchy (i.e. before we create script components)
        // Split loading into load and open
        const handler = app.loader.getHandler("hierarchy");

        // called after scripts are preloaded
        const _loaded = function () {
            app.systems.script.preloading = true;
            const entity = handler.open(url, data);

            app.systems.script.preloading = false;

            // clear from cache because this data is modified by entity operations (e.g. destroy)
            app.loader.clearCache(url, "hierarchy");

            // add to hierarchy
            app.root.addChild(entity);

            // initialize components
            app.systems.fire('initialize', entity);
            app.systems.fire('postInitialize', entity);
            app.systems.fire('postPostInitialize', entity);

            if (callback) callback(null, entity);
        };

        // load priority and referenced scripts before opening scene
        app._preloadScripts(data, _loaded);
    }
```

src/framework/scene-registry.js
```diff
@@ -253,6 +261,38 @@ class SceneRegistry {
         }
     }
 
+    _addHierarchyToScene(url, data, callback) {
+        const self = this;
+
+        // Because we need to load scripts before we instance the hierarchy (i.e. before we create script components)
+        // Split loading into load and open
+        const handler = this._app.loader.getHandler("hierarchy");
```
kungfooman, line 269: Should `handler` be a closure variable? Seems like it should be in the inner scope of `_loaded`

src/framework/scene-registry.js
yaustar: Is that property lookup that expensive? Happy to change it but interested in the performance difference.

src/framework/scene-registry.js
yaustar: Good call, this was probably some left over code from how it used to work that should be cleaned up

src/framework/scene-registry.js
kungfooman: I didn't benchmark 😅 Probably good to reduce code-size/ascii-soup aswell

src/framework/scene-registry.js
```diff
@@ -347,6 +358,52 @@ class SceneRegistry {
...
+     * app.scenes.ChangeScene("Scene Name", function (err, entity) {
+     *     if (!err) {
+     *         // success
+     *     } else {
+     *         // error
+     *     }
+     * });
+     */
+    changeScene(sceneItem, callback) {
+        if (!(sceneItem instanceof SceneRegistryItem)) {
```
kungfooman, line 379: Somewhat shorter without negation:

```suggestion
        if (typeof sceneItem === 'string') {
```

kungfooman:
Right now in `SceneRegistry#_loadSceneData` there is this code:

```js
        let url = sceneItem;

        if (sceneItem instanceof SceneRegistryItem) {
            url = sceneItem.url;
        } else {
            sceneItem = this.findByUrl(url);
            if (!sceneItem) {
                sceneItem = new SceneRegistryItem('Untitled', url);
            }
        }
```

It could be shortened into this code:

```js
        let url = sceneItem;
        if (typeof sceneItem === 'string') {
            sceneItem = this.findByUrl(url) || this.find(url) || new SceneRegistryItem('Untitled', url);
        }
        url = sceneItem.url;
```

Then you can delete this code in `SceneRegistry#changeCode` aswell:

```js
if (!(sceneItem instanceof SceneRegistryItem)) {
    sceneItem = this.find(sceneItem);
}
```

Overall it removes lines of code and `SceneRegistry#changeCode` would also accept URL's (and `SceneRegistry#loadSceneData` would also accept scene names).

I don't know if scene names and scene URL's could conflict though (in my little testing everything worked as it should and the unit tests are also happy).

src/framework/scene-registry.js
yaustar: Done

yaustar:
> Right now in `SceneRegistry#_loadSceneData` there is this code:
> 
> ```js
>         let url = sceneItem;
> 
>         if (sceneItem instanceof SceneRegistryItem) {
>             url = sceneItem.url;
>         } else {
>             sceneItem = this.findByUrl(url);
>             if (!sceneItem) {
>                 sceneItem = new SceneRegistryItem('Untitled', url);
>             }
>         }
> ```
> 
> It could be shortened into this code:
> 
> ```js
>         let url = sceneItem;
>         if (typeof sceneItem === 'string') {
>             sceneItem = this.findByUrl(url) || this.find(url) || new SceneRegistryItem('Untitled', url);
>         }
>         url = sceneItem.url;
> ```
> 
> Then you can delete this code in `SceneRegistry#changeCode` aswell:
> 
> ```js
> if (!(sceneItem instanceof SceneRegistryItem)) {
>     sceneItem = this.find(sceneItem);
> }
> ```
> 
> Overall it removes lines of code and `SceneRegistry#changeCode` would also accept URL's (and `SceneRegistry#loadSceneData` would also accept scene names).
> 
> I don't know if scene names and scene URL's could conflict though (in my little testing everything worked as it should and the unit tests are also happy).

I like the idea that loadScene* and changeScene can all take a name, URL or SceneRegistryItem and would make the loadScene* functions more flexible.

Any opinions from the rest of the engine team?

yaustar:
I've added @kungfooman's suggestion on being able to use the scene name and added a few more tests

kungfooman:
Thank you for the changes @yaustar 

Since `SceneRegistry#changeScene` is basically `SceneRegistry#loadSceneHierarchy` with a preface, it can be rewritten as exactly that to reduce quite some lines of code (and makes future rewrites easier I think too):

```js
    changeScene(sceneItem, callback) {
        /**
         * This function is called before the new scene hierarchy is added.
         *
         * @param {SceneRegistryItem} sceneItem - The scene item
         */
        const preface = (sceneItem) => {
            const app = this._app;
            // Destroy/Remove all nodes on the app.root
            const rootChildren = app.root.children;
            while (rootChildren.length > 0) {
                const child = rootChildren[0];
                child.reparent(null);
                child.destroy?.();
            }
            app.applySceneSettings(sceneItem.data.settings);
        };
        this.loadSceneHierarchy(sceneItem, callback, preface);
    }
```

`SceneRegistry#loadSceneHierarchy` just needs to handle the preface:

```js
    loadSceneHierarchy(sceneItem, callback, preface) {
        this._loadSceneData(sceneItem, false, (err, sceneItem) => {
            if (err) {
                if (callback) {
                    callback(err);
                }
                return;
            }
            if (preface) {
                preface(sceneItem);
            }
            this._addHierarchyToScene(sceneItem.url, sceneItem.data, callback);
        });
    }
```

Using arrow functions is reducing code size aswell, you don't need to worry about `self` (arrow functions inherit `this` and its fixed/unchangeable). Babel takes care to rewrite it into ES5:

```js
	  _proto.changeScene = function changeScene(sceneItem, callback) {
	    var _this2 = this;

	    var preface = function preface(sceneItem) {
	      var app = _this2._app;
	      var rootChildren = app.root.children;

	      while (rootChildren.length > 0) {
	        var child = rootChildren[0];
	        child.reparent(null);
	        child.destroy == null ? void 0 : child.destroy();
	      }

	      app.applySceneSettings(sceneItem.data.settings);
	    };

	    this.loadSceneHierarchy(sceneItem, callback, preface);
	  };
```

yaustar:
I'm not keen on use the preface as loadSceneHierarchy is existing public API and traditionally, the callback is the last param

kungfooman:
I kinda liked the idea because it would lead to more code reduction afterwards for `SceneHierarchy#_addHierarchyToScene`, since its basically just a stub for its inner function called `_loaded`, but I see your point 👍

yaustar:
We could create a new private function that does what you suggest I suppose? That way that preface is not exposed as public API nor changes the existing?

kungfooman:
> We could create a new private function that does what you suggest I suppose?

That sounds like a good idea and I just tested and it works:

```js
    loadSceneHierarchy(sceneItem, callback) {
        this._loadSceneHierarchy(sceneItem, callback);
    }
    _loadSceneHierarchy(sceneItem, callback, preface) {
        this._loadSceneData(sceneItem, false, (err, sceneItem) => {
            if (err) {
                if (callback) {
                    callback(err);
                }
                return;
            }
            if (preface) {
                preface(sceneItem);
            }
            this._addHierarchyToScene(sceneItem.url, sceneItem.data, callback);
        });
    }
```

And then you can just use this private method in `SceneRegistry#changeScene`

yaustar:
I've made the changes to DRY but it feels off as it's adding some complexity for a few extra lines 🤔

src/framework/scene-registry.js
```diff
@@ -347,6 +359,41 @@ class SceneRegistry {
...
+     *         // success
+     *     } else {
+     *         // error
+     *     }
+     * });
+     */
+    changeScene(sceneItem, callback) {
+        const app = this._app;
+
+        const onBeforeAddHierarchy = (sceneItem) => {
```
kungfooman, line 382: This `app` closure variable also has too much context and maybe add JSDoc?

src/framework/scene-registry.js
```diff
@@ -253,89 +258,96 @@ class SceneRegistry {
...
      */
     loadSceneSettings(sceneItem, callback) {
-        const self = this;
+        const app = this._app;
 
-        this._loadSceneData(sceneItem, false, function (err, sceneItem) {
+        this._loadSceneData(sceneItem, false, (err, sceneItem) => {
             if (!err) {
-                self._app.applySceneSettings(sceneItem.data.settings);
+                app.applySceneSettings(sceneItem.data.settings);
```
kungfooman, line 350: Since you are using arrow functions now, you can just use `this` and move that closure variable into the arrow function itself. And since `app` is only used once, it doesn't really need to be declared at all

src/framework/scene-registry.js
yaustar: Is JSDoc needed here for a private function that is only used in the scope of the current class?

kungfooman: I guess its not needed, I just use them as a mind crutch and when the types are checked, it can offer some insight about what is going wrong in which position

kungfooman:
I am running into an issue loading SeeMore as non-trivial scene and for some reason it doesn't destroy the old camera:

image

(one camera rendering over the other)

I can fix it by adding this to your `onBeforeAddHierarchy()`:

```js
app.scene?._activeCamera?.node?.destroy();
```

But I don't understand it, shouldn't the camera be in `app.root` anyway or what else is required to completely destroy a scene?

yaustar:
> I am running into an issue loading SeeMore as non-trivial scene and for some reason it doesn't destroy the old camera:
> 
> image
> 
> (one camera rendering over the other)
> 
> I can fix it by adding this to your `onBeforeAddHierarchy()`:
> 
> ```js
> app.scene?._activeCamera?.node?.destroy();
> ```
> 
> But I don't understand it, shouldn't the camera be in `app.root` anyway or what else is required to completely destroy a scene?

That's old, there's no scenes in that project

yaustar:
Testing with a Seemore fork with changing scenes with the existing stable engine and the double render is there too: https://playcanvas.com/project/983477/overview/f-seemore

kungfooman:
> Edit: I just tried it with my branch and can't reproduce your issue?

Thank you for testing, I downloaded SeeMore in 2019 but I also just tested the last version and it has the same issue.

1) Open https://playcanv.as/p/MflWvdTW/
2) Get both cameras: `cams = pc.app.root.find("name", "Camera");`
3) Delete as you do in this PR:
```js
const rootChildren = pc.root.children;
while (rootChildren.length > 0) {
    const child = rootChildren[0];
    child.reparent(null);
    child.destroy?.();
}
```
4) It still renders skybox for you too?

image

The first cam seems to "survive" 🧟

yaustar:
> > Edit: I just tried it with my branch and can't reproduce your issue?
> 
> Thank you for testing, I downloaded SeeMore in 2019 but I also just tested the last version and it has the same issue.
> 
> 1. Open https://playcanv.as/p/MflWvdTW/
> 2. Get both cameras: `cams = pc.app.root.find("name", "Camera");`
> 3. Delete as you do in this PR:
> 
> ```js
> const rootChildren = pc.root.children;
> while (rootChildren.length > 0) {
>     const child = rootChildren[0];
>     child.reparent(null);
>     child.destroy?.();
> }
> ```
> 
> 4. It still renders skybox for you too?
> 
> image
> 
> The first cam seems to "survive" 🧟

I tried this with engine 1.56.0 (current stable) and loadSceneSettings + Hierarchy methods and the issue is there too.

This means it's not a new issue that this PR introduces

kungfooman:
I am also not quite happy with `onBeforeAddHierarchy` and I realized that we basically just need an event...

The first step would be to move your scene clearing code into `class Scene` (where it imo belongs). Since `Scene` already extends `EventHandler`, we have events without any changes.

New method for `src/scene/scene.js`:

```js
    clear() {
        this.fire("clear");
        const app = getApplication();
        // Destroy/Remove all nodes on the app.root
        const rootChildren = app.root.children;
        while (rootChildren.length > 0) {
            const child = rootChildren[0];
            child.reparent(null);
            child.destroy?.();
        }
        this?._activeCamera?.node?.destroy();
    }
```

(the last line with `_activeCamera` fixes the SeeMore issue/bug, until someone comes up with a better solution)

The `SceneRegistry#changeScene` turns into:

```js
    changeScene(sceneItem, callback) {
        const app = this._app;
        app.scene.once("beforeAddHierarchy", (sceneItem) => {
            app.scene.clear();
            app.applySceneSettings(sceneItem.data.settings);
        });
        this.loadSceneHierarchy(sceneItem, callback);
    }
```

`onBeforeAddHierarchy` is completely removed and you can remove that extra private method.

Finally, in `loadSceneHierarchy`, instead of:

```js
            if (onBeforeAddHierarchy) {
                onBeforeAddHierarchy(sceneItem);
            }

            const app = this._app;
```

We simply have to fire the event while passing `sceneItem`:

```js
            const app = this._app;
            app.scene.fire("beforeAddHierarchy", sceneItem);
```

mvaligursky:
I would not use `_activeCamera` at all .. I want to remove it from code base, as it's just a hack added by particle system rendering. I'm not entirely what this is trying to solve, but it should be done differently.

kungfooman:
> I would not use `_activeCamera` at all ..

Yes, I can imagine that, for testing I just needed a way to clear the entire scene, since deleting only `app.root.children` still retains one camera. I don't know why, just that it can be deleted by that `_activeCamera` line. Do you have an idea how to fix this?

yaustar:
I'm less keen to add an event into a system that hasn't used them before. It feels a bit overkill and would like to put a pin in it here for team review.

In regards to Seemore, I will create a new ticket to investigate as that's odd but not introduced by this PR. Maybe due to th use of zones or portals?

mvaligursky:
It's really strange the Camera component survives the scene hierarchy getting deleted. There must be some code re-adding it or similar?

If you have a repro, it'd be interesting to trace layer composition:

`pc.Tracing.set(pc.TRACEID_RENDER_ACTION, true);`

at the point the hierarchy is deleted, I'd expect this to print no render actions - meaning no cameras are in the scene.

src/framework/scene-registry.js
```diff
@@ -151,26 +159,24 @@ class SceneRegistry {
...
-            if (!sceneItem) {
-                sceneItem = new SceneRegistryItem('Untitled', url);
-            }
+        if (typeof sceneItem === 'string') {
+            sceneItem = this.findByUrl(url) || this.find(url) || new SceneRegistryItem('Untitled', url);
         }
 
+        url = sceneItem.url;
+
         if (!sceneItem.url) {
```
mvaligursky, line 178: this could now be `if (!url) {`

src/framework/scene-registry.js
```diff
@@ -347,6 +358,41 @@ class SceneRegistry {
...
+    /**
+     * Change to a new scene. Calling this function will load the scene data, delete all
+     * entities and graph nodes under `app.root` and load the scene settings and hierarchy.
+     *
+     * @param {SceneRegistryItem | string} sceneItem - The scene item (which can be found with
+     * {@link SceneRegistry#find}, URL of the scene file (e.g."scene_id.json") or name of the scene.
+     * @param {ChangeSceneCallback} [callback] - The function to call after loading,
+     * passed (err, entity) where err is null if no errors occurred.
+     * @example
+     * app.scenes.ChangeScene("Scene Name", function (err, entity) {
```
mvaligursky, line 370:
```suggestion
     * app.scenes.changeScene("Scene Name", function (err, entity) {
```

mvaligursky (approved):
approving with small comments

src/framework/scene-registry.js
yaustar, line 178:
```suggestion
        if (!url) {
```

PR #4634, opened 2022-09-12
Make ObjectPool#freeAll() explicit

ertugrulcetin:
Reset `this._pool` array.

### DIFF
--- a/src/core/object-pool.js
+++ b/src/core/object-pool.js
@@ -24,6 +24,7 @@ class ObjectPool {
 
     freeAll() {
         this._count = 0;
+        this._pool = [];
     }
 }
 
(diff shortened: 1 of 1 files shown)


kungfooman:
The point of a pool is to pool objects. If you just empty `ObjectPool#_pool`, the pooled objects are gone.

When the engine tries to `ObjectPool#allocate`, it will return undefined:

image

image

In which way did you test this?

ertugrulcetin:
@kungfooman I think it is correct now.

LeXXik:
@ertugrulcetin The purpose of freeing the pool objects is to mark them available for reuse, not destroying them. Once you release all in the pool, the very first object that is already allocated in the pool becomes available the next time it is requested. There would be no new allocation taking place, hence the purpose of having a pool.

If you want to destroy all allocated objects and clear the pool, I would suggest to add a new method that would explicitly clear it.

ertugrulcetin:
@LeXXik thank you for the explanation. It makes sense to me now.

PR #4636, opened 2022-09-14
Adds a texture mapping allowing us to combine sampler uniforms

GSterbrant:
### Description
All frontend chunks use the new textureMapping lookup to understand if the chunk needs to output a new uniform sampler for its texture fetches or if the same texture is already declared as a sampler elsewhere. This solves a bug where some materials just fails to compile due to overuse of texture samplers.

For example, the glTF sample [Stained Glass Lamp](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/StainedGlassLamp) would fail to compile with clustered on using over 16 samplers, with this PR only uses 10.


willeastcott (approved):
A great improvement that also actually removes code from the engine. Nice! Get at least one more review though before merging.

src/graphics/program-lib/programs/standard.js
```diff
@@ -334,29 +349,29 @@ const standard = {
             // albedo
             decl.append("vec3 dAlbedo;");
             if (options.diffuseDetail) {
-                code.append(this._addMap("diffuseDetail", "diffuseDetailMapPS", options, litShader.chunks));
+                code.append(this._addMap("diffuseDetail", "diffuseDetailMapPS", options, litShader.chunks, textureMapping, options.diffuseEncoding));
```
slimbuck, line 352: diffuseDetail doesn't necessarily have the same encoding as diffuse

slimbuck (commented):
Awesome PR @GSterbrant. I especially like that existing chunks are still supported (..right?).

Few things:
- please update chunk validation with updated API for frontend chunks
- please update chunk docs on developer site

src/scene/materials/standard-material-options-builder.js
```diff
@@ -154,8 +154,9 @@ class StandardMaterialOptionsBuilder {
         options.normalizeNormalMap = stdMat.normalizeNormalMap;
         options.ambientSH = !!stdMat.ambientSH;
         options.useSpecular = useSpecular;
-        options.emissiveEncoding = stdMat.emissiveMap ? stdMat.emissiveMap.encoding : null;
-        options.lightMapEncoding = stdMat.lightMap ? stdMat.lightMap.encoding : null;
+        options.diffuseEncoding = stdMat.diffuseMap?.encoding;
+        options.emissiveEncoding = stdMat.emissiveMap?.encoding;
+        options.lightMapEncoding = stdMat.lightMap?.encoding;
```
slimbuck, line 159: also need to add sheenEncoding?

src/graphics/program-lib/programs/standard.js
GSterbrant: Ah! Good catch, will fix that!

GSterbrant: https://github.com/playcanvas/developer.playcanvas.com/pull/439

Here's the PR for the chunk migrations.

slimbuck (approved):
Fricken awesome!

mvaligursky (approved):

PR #4637, opened 2022-09-14
Mesh picker

sri-30:
Fixes #
This PR adds a triangle class and set of bvh classes which allow for fast ray intersection lookups with mesh instances. 

These lookups can be performed with the following code:

`d = meshInstance.rayCast(ray)`

This function returns the closest hitpoint of the ray on the mesh, relative to the ray origin, or null if there is no intersection.

Two new example have also been added to demonstrate this: mesh-picker and mesh-picker-animated. mesh-picker-animated demonstrates the BVH refitting with an animated mesh.


src/math/bvh.js
```diff
@@ -0,0 +1,451 @@
+import { Vec3 } from '../math/vec3.js';
+import { BoundingBox } from '../shape/bounding-box.js';
+
+/** @typedef {import('../shape/ray.js').Ray} Ray */
+
+const INFINITY = Infinity;
+const BINS = 10;
```
kungfooman, line 7: `BINS` should be configurable or a default property and `INFINITY` just doesn't helps anything

src/math/bvh.js
```diff
@@ -0,0 +1,451 @@
...
+}
+
+/**
+ * Used to store the BVH, triangles and minimum ray intersection distance for a single mesh
+ */
+class BVHGlobal {
+    /**
+     * Create a new array to store the BVH and helpers
+     *
+     * @param {Array} [triangles] - the triangles used to build the BVH
```
kungfooman, line 57:
```suggestion
     * @param {Tri[]} [triangles] - the triangles used to build the BVH
```

src/math/bvh.js
```diff
@@ -0,0 +1,451 @@
...
+        root.leftFirst = 0;
+        root.triCount = N;
+        this.UpdateNodeBounds(rootNodeIdx);
+        this.Subdivide(rootNodeIdx);
+    }
+
+    /**
+     * Refits the BVH
+     *
+     * @param {Array} [triangles] - The triangles to refit the BVH on
```
kungfooman, line 313:
```suggestion
     * @param {Tri[]} [triangles] - The triangles to refit the BVH on
```

src/math/bvh.js
```diff
@@ -0,0 +1,451 @@
...
+     * @ignore
+     */
+    IntersectBVH(ray, nodeIdx = 0) {
+        let node = this.bvhNode[nodeIdx];
+        const stack = [];
+        let stackPtr = 0;
+        while (true) {
+            if (node.isLeaf()) {
+                for (let i = 0; i < node.triCount; i++) {
+                    const dist = this.triangles[this.triIdx[node.leftFirst + i]].intersectWithRay(ray);
```
kungfooman, line 349: You can do `const {triangles, triIdx, bvhNode} = this;` at top and prevent all these property lookups again and again and make the code shorter and faster.

src/math/bvh.js
```diff
@@ -0,0 +1,451 @@
...
+
+function vec3MinInPlace(a, b, c) {
+    c.set(Math.min(a.x, b.x), Math.min(a.y, b.y), Math.min(a.z, b.z));
+    return c;
+}
+
+function vec3MaxInPlace(a, b, c) {
+    if (c == null) {
+        c = new Vec3();
+    }
```
kungfooman, line 429: 1) This looks strange, is this supposed to work as default argument?
2) Please add JSDoc
3) Prefer `===` over `==`

src/math/bvh.js
```diff
@@ -75,23 +149,74 @@ class BVHGlobal {
...
-                const candidatePos = triangle.centroid[b];
-                const cost = this.EvaluateSAH(node, a, candidatePos);
-                if (cost < bestCost) {
-                    splitDetails.splitPos = candidatePos;
+                boundsMin = Math.min(boundsMin, triangle.centroid[b]);
+                boundsMax = Math.max(boundsMax, triangle.centroid[b]);
+            }
+            if (boundsMin === boundsMax) continue;
+
+            const BINS = 8;
```
kungfooman: Another instance of `BINS` that just confuses things

sri-30:
> Nice work, I know this is rather complicated so I wondered what resources you used.
> 
> I could find this one: https://jacco.ompf2.com/2022/04/21/how-to-build-a-bvh-part-3-quick-builds/ (looks like exact copies partly)

Thank you for your comments - I've acted on them and made the appropriate changes. Yes, it's almost entirely based on that tutorial series - I'm currently looking into other potential algorithms for building the BVH, while keeping the same base class structure.

src/shape/ray.js
```diff
@@ -44,10 +46,24 @@ class Ray {
      * @returns {Ray} Self for chaining.
      */
     set(origin, direction) {
+        this.rDx = null;
+        this.rDy = null;
+        this.rDz = null;
```
mvaligursky, line 51: what are these?

src/shape/ray.js
```diff
@@ -44,10 +46,24 @@ class Ray {
...
+    /**
+     * Transforms the ray to a new space via Matrix in-place
+     *
+     * @param {Mat4} matrix - The transformation matrix for the ray
+     */
+    transform(matrix) {
+        const newOrigin = matrix.transformPoint(this.origin);
+        const newDirection = matrix.transformVector(this.direction);
+        this.set(newOrigin, newDirection);
+    }
```
mvaligursky, line 66: This should work as well, and avoid internal allocations:

```
transform(matrix) {
        matrix.transformPoint(this.origin, this.origin);
        matrix.transformVector(this.direction, this.direction);
}
```

src/shape/bounding-box.js
```diff
@@ -421,6 +421,12 @@ class BoundingBox {
         return sq;
     }
 
+    area() {
+        const extents = this.halfExtents.clone();
+        extents.mulScalar(2);
+        return extents.x * extents.y + extents.z * extents.x + extents.y * extents.z;
+    }
+
```
mvaligursky, line 429: this does not seem to be used in this PR .. can it be deleted?

sri-30: It's used in the `findBestSplitPlane` function to evaluate surface area sizes when considering the best bin to split along; should I move the function to the bvh class since it's only used there?

src/shape/bounding-box.js
```diff
@@ -421,6 +421,12 @@ class BoundingBox {
         return sq;
     }
 
+    area() {
+        const extents = this.halfExtents.clone();
+        extents.mulScalar(2);
+        return extents.x * extents.y + extents.z * extents.x + extents.y * extents.z;
+    }
```
kungfooman, line 428: No matter where this lands, it is a getter and can be shortened aswell with less lookups:

```suggestion
    get area() {
        const {x, y, z} = this.halfExtents.clone().mulScalar(2);
        return x * y + z * x + y * z;
    }
```

(the `.area()` braces need to be removed of course too)

mvaligursky: personally, I'd prefer solution without .clone() in it, as it allocates object. We only clone is so that we can multiply it by 2 .. I think we should just factor that multiplication into the last line directly.

src/math/tri.js
```diff
@@ -0,0 +1,73 @@
...
+
+// Internal variables
+const h = new Vec3();
+const s = new Vec3();
+const q = new Vec3();
+
+/**
+ * Class for representing a triangle in 3D space
+ */
+class Tri {
```
mvaligursky, line 13: is this class needed here? It seems it's mostly / completely a copy of tri.js?

mvaligursky: actually ignore this, the review just wasn't opening the file :)

src/math/bvh.js
```diff
@@ -0,0 +1,456 @@
...
+        this.BINS = BINS;
+        this.bounds = new BoundingBox(new Vec3(), new Vec3());
+        this.triCount = 0;
+    }
+}
+
+/**
+ * Used to store the BVH, triangles and minimum ray intersection distance for a single mesh
+ */
+class BVHGlobal {
```
mvaligursky, line 51: usually we try the filename to match the main class the file exports. I'd probably rename the class to be Bvh.

src/scene/mesh.js
```diff
@@ -981,6 +997,85 @@ class Mesh extends RefCountedObject {
         };
         this.indexBuffer[RENDERSTYLE_WIREFRAME] = wireBuffer;
     }
+
+    /**
+     * Builds the array of triangles of the mesh
+     */
```
mvaligursky, line 1003: functions that are not public need to be marked with `@ignore`

src/scene/mesh.js
```diff
@@ -981,6 +997,85 @@ class Mesh extends RefCountedObject {
...
+                triangle.vertex2.copy(points[indices[j + 2]]);
+                triangle.resetEdges();
+                j += 3;
+            }
+        }
+
+    }
+
+    /**
+     * @param {Ray} ray - The ray to interesect the mesh with
```
mvaligursky, line 1052: A function description is missing, see for example getPositions

src/scene/mesh.js
```diff
@@ -981,6 +997,85 @@ class Mesh extends RefCountedObject {
...
+     * @returns {number} The distance between the ray origin and closest point of intersection
+     */
+    rayCast(ray) {
+
+        // if the BVH doesn't exist, build it
+        // TODO: rebuilding bvh when dirtyCount exceeds limit
+        // Currently bvh building is not efficient enough for this to be feasible in real-time
+        if (!this.bvh || this.dirtyCount > Infinity) {
```
mvaligursky, line 1060: what does `this.dirtyCount > Infinity` mean?

src/scene/mesh.js
```diff
@@ -981,6 +997,85 @@ class Mesh extends RefCountedObject {
...
+
+
+        // Rebuild triangles and refit the BVH if the mesh has been altered
+        if (this.dirtyBVH) {
+            this.buildTriangleArray();
+            this.bvh.refitBVH(this.triangles);
+            this.dirtyBVH = false;
+        }
+
+        this.bvh.minDist = null;
```
mvaligursky, line 1074: should this line be a part of intersectBVH function?

src/math/bvh.js
```diff
@@ -0,0 +1,456 @@
...
+        this.Subdivide(leftChildIdx);
+        this.Subdivide(rightChildIdx);
+    }
+
+    /**
+     * Builds the BVH
+     *
+     * @ignore
+     */
+    buildBVH() {
```
mvaligursky, line 291: should just be called `build`

src/math/bvh.js
```diff
@@ -0,0 +1,456 @@
...
+    }
+
+    /**
+     * Calculates the shortest distance between the origin of a ray and its intersection with the mesh
+     *
+     * @param {Ray} ray - The ray to be traced
+     * @param {number} nodeIdx - The index of the node to begin the raycast at (default is zero: the root)
+     * @ignore
+     */
+    intersectBVH(ray, nodeIdx = 0) {
```
mvaligursky, line 348: call it 'intersect'

src/math/bvh.js
```diff
@@ -0,0 +1,456 @@
...
+        ray.rDx = 1 / ray.direction.x;
+        ray.rDy = 1 / ray.direction.y;
+        ray.rDz = 1 / ray.direction.z;
+
+        let node = this.bvhNode[nodeIdx];
+        const stack = [];
```
mvaligursky, line 354: have stack as a property of this class
- created that in constructor as `this.stack = [];`
- to clear it, do `this.stack.lenght = 0;`  // this keeps it allocated, and avoids new array allocation

src/math/bvh.js
```diff
@@ -0,0 +1,456 @@
...
+/**
+ * Returns a vector with the minimum x, y and z coordinates of the two vectors passed in
+ *
+ * @param {Vec3} a - First Vector
+ * @param {Vec3} b - Second Vector
+ * @param {Vec3} [c] - The vector to output the result to - default is new Vec3()
+ * @returns {Vec3} Returns the vector result
+ * @ignore
+ */
+function vec3Min(a, b, c = new Vec3()) {
```
mvaligursky, line 437: I'd add these to Vec3 class, called min2 (we already have min). See Vec3.add vs Vec3.add2

src/math/tri.js
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Creates a new Triangle Object.
+     *
+     * @param {Vec3} vertex0 - Vertex 0 value
+     * @param {Vec3} vertex1 - Vertex 1 value
+     * @param {Vec3} vertex2 - Vertex 2 value
+     * @param {Vec3} [centroid] - Centroid value
+     */
+
+    constructor(vertex0, vertex1, vertex2, centroid) {
```
mvaligursky, line 23: I think you only call it from one place, without centroid parameter .. is it still required as a parameter to the constructor?

src/math/tri.js
```diff
@@ -0,0 +1,73 @@
...
+class Tri {
+    /**
+     * Creates a new Triangle Object.
+     *
+     * @param {Vec3} vertex0 - Vertex 0 value
+     * @param {Vec3} vertex1 - Vertex 1 value
+     * @param {Vec3} vertex2 - Vertex 2 value
+     * @param {Vec3} [centroid] - Centroid value
+     */
+
```
mvaligursky, line 22: remove the space

src/math/tri.js
```diff
@@ -0,0 +1,73 @@
...
+     * and intersection point if there is an intersection.
+     *
+     * @param {Ray} ray - Ray to interect triangle with
+     * @returns {number} Distance between ray origin and intersection
+     */
+    intersectWithRay(ray) {
+        h.cross(ray.direction, this.edge2);
+        const a = this.edge1.dot(h);
+        if (a > -0.0001 && a < 0.0001) {
+            return;
```
mvaligursky, line 52: these returns without parameter return `undefined`, but at the place you call this function, you test the result against null using != which probably converts types to make null equal to undefined.
ideally you change that test to be just `if (dist)` instead of `if (dist != null)`

src/scene/mesh-instance.js
```diff
@@ -455,6 +461,32 @@ class MeshInstance {
...
+    rayCast(ray) {
+        _worldTransformInverted.copy(this.node.getWorldTransform());
+        _worldTransformInverted.invert();
+        _transformedRay.set(ray.origin, ray.direction);
+        _transformedRay.transform(_worldTransformInverted);
+        _transformedRay.direction.normalize();
+        const dist = this._mesh.rayCast(_transformedRay);
+        if (dist != null) {
+            _direction.copy(_transformedRay.direction);
+            _direction.normalize();
```
mvaligursky, line 480: _transformedRay.direction is normalized few lines up, no need to do this again?

src/scene/mesh-instance.js
```diff
@@ -455,6 +461,32 @@ class MeshInstance {
...
 
+    /**
+     * Performs a raycast with a given ray against the mesh
+     *
+     * @param {Ray} ray - the ray to perform the intersect test with
+     * @returns {Vec3} Null if there is no intersection or the point of intersection
+     * in world space
+     */
+    rayCast(ray) {
+        _worldTransformInverted.copy(this.node.getWorldTransform());
```
mvaligursky, line 472: avoid calling this twice. Do
`const worldTransform = this.node.getWorldTransform())';
and use it at the end of the function too

src/math/bvh.js
```diff
@@ -0,0 +1,456 @@
+import { Vec3 } from '../math/vec3.js';
+import { BoundingBox } from '../shape/bounding-box.js';
+
+/** @typedef {import('../shape/ray.js').Ray} Ray */
+/** @typedef {import('../math/tri.js').Tri} Tri */
+
+class BVHNode {
```
mvaligursky, line 7: mark whole class with `@ignore`, instead of individual functions.

src/math/bvh.js
```diff
@@ -0,0 +1,456 @@
...
+    /**
+     * Create a new BVHNode
+     *
+     * @param {Vec3} [aabbMin] - The bounds of the BVH node
+     * @param {Vec3} [aabbMax] - The bounds of the BVH node
+     * @param {number} [leftFirst] - The index of the left child of the BVH node
+     * @param {number} [triCount] - The number of primitives
+     * @ignore
+     */
+    constructor(aabbMin, aabbMax, leftFirst, triCount) {
```
mvaligursky, line 17: change it to handle default params:
`constructor(aabbMin = new Vec3(), aabbMax = new Vec3(), leftFirst = null, triCount = 0) {`

src/math/bvh.js
```diff
@@ -0,0 +1,456 @@
...
+
+    /**
+     * Calculates the estimated cost of raycasting through a node
+     *
+     * @param {BVHNode} node - The node to be evaluated
+     * @returns {number} The cost of the raycast
+     * @ignore
+     */
+    calculateNodeCost(node) {
+        const extent = new Vec3();
```
mvaligursky, line 212: use some global variable to avoid the allocation each time here

src/math/bvh.js
```diff
@@ -0,0 +1,456 @@
...
+        return node.triCount * surfaceArea;
+    }
+
+    /**
+     * Subdivides the BVH on a node
+     *
+     * @param {number} nodeIdx - The index of the node to subdivide at
+     * @ignore
+     */
+    Subdivide(nodeIdx) {
```
mvaligursky, line 226: lower case function name

src/math/bvh.js
```diff
@@ -0,0 +1,456 @@
...
+        const axis = splitDetails.axis;
+
+        const splitPos = splitDetails.splitPos;
+
+        // const noSplitCost = this.calculateNodeCost(node);
+
+        // terminates tree building based on split cost
+        // if (splitCost >= noSplitCost) {
+        //     return;
+        // }
```
mvaligursky, line 245: delete if not needed

src/math/bvh.js
```diff
@@ -0,0 +1,456 @@
...
+
+        // terminates tree building when number of leafnodes is less than 5
+        if (node.triCount < 10) {
```
mvaligursky, line 248: 5 or 10?

src/math/bvh.js
```diff
@@ -0,0 +1,456 @@
...
+    }
+
+    /**
+     * Builds the BVH
+     *
+     * @ignore
+     */
+    buildBVH() {
+        const { triangles } = this;
+        const N = triangles.length;
```
mvaligursky, line 293: lower case variable names

src/math/bvh.js
```diff
@@ -0,0 +1,456 @@
...
+        });
+        const rootNodeIdx = 0;
+        this.nodesUsed = 1;
+
+        for (let i = 0; i < N; i++) {
+            triangles[i].centroid = new Vec3();
+            triangles[i].centroid.add(triangles[i].vertex0);
+            triangles[i].centroid.add(triangles[i].vertex1);
+            triangles[i].centroid.add(triangles[i].vertex2);
+            triangles[i].centroid.mulScalar(1 / 3);
```
mvaligursky, line 306: this should be a function on a triangle. And perhaps use divScalar(3); (which should be changed to do a single division only, and multiply with it)

src/shape/bounding-box.js
kungfooman: Updated to least amount of GC and float operations, can you counter-check?

mvaligursky: looks good to me.
Also, the box has 6 sides .. but we only add up areas of 3 of them .. to be really correct, the result should be *2 ?

kungfooman: > Also, the box has 6 sides .. but we only add up areas of 3 of them .. to be really correct, the result should be *2 ?

Yes, I agree, maybe its not important for the algorithm, but @sri-30 has to check it now. Otherwise I guess it should be called `halfArea` or something

sri-30: My thoughts exactly; since it's only required for a heuristic. I'll rename it to halfArea, and use the version that @kungfooman has suggested.

src/math/bvh.js
sri-30: Renamed to Bvh

src/scene/mesh.js
sri-30: Done

src/scene/mesh.js
sri-30: Added description: `Performs a rayCast on the mesh`

src/scene/mesh.js
sri-30: Moved to the intersect function

src/math/bvh.js
sri-30: Changed

src/math/bvh.js
```diff
@@ -0,0 +1,456 @@
...
+        this.Subdivide(rootNodeIdx);
+    }
+
+    /**
+     * Refits the BVH
+     *
+     * @param {Tri[]} triangles - The triangles to refit the BVH on
+     * @ignore
+     */
+    refitBVH(triangles) {
```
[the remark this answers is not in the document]
sri-30: All functions have been renamed as appropriate

src/math/bvh.js
sri-30: Done

src/math/tri.js
sri-30: The centroid parameter is no longer in the constructor, and is initialized as a Vec3. New function `calculateCentroid()` has been added to the class

src/math/bvh.js
sri-30: Added new min2 and max2 functions, and changed all lines in bvh to use those as appropriate, as well as a new temporary internal: `tempVec1`

src/math/tri.js
sri-30: Changed all null checks to `if (dist)`

src/scene/mesh-instance.js
sri-30: Removed _direction variable as it was redundant:

```js
if (dist) {
            _transformedRay.direction.mulScalar(dist);
            _transformedRay.origin.add(_transformedRay.direction);
            this.node.getWorldTransform().transformPoint(_transformedRay.origin, _transformedRay.origin);
            return _transformedRay.origin;
}
```

src/math/bvh.js
sri-30: Marked all classes in this file as `@ignore`

src/math/bvh.js
sri-30: Added

src/math/bvh.js
sri-30: Added

src/math/bvh.js
sri-30: Changed all functions to camel-case

src/math/bvh.js
sri-30: Removed

src/math/bvh.js
sri-30: Changed the comment to say threshold; should I make this configurable?

src/math/bvh.js
sri-30: Declared a constant 1/3 at the top of the file which is used in mulScalar.

src/math/bvh.js
mvaligursky: probably not at this stage, a fixed number should be ok

src/scene/mesh.js
sri-30: Added a comment to explain this - dirtyCount is the number of times the bvh has been refitted since it was last built. Since this is likely to result in it becoming outdated for animated meshes, dirtyCount can be used to rebuild the bvh as appropriate when this occurs, but rebuilding is not currently fast enough for this process to be seamless.

mvaligursky: but what does the comparison mean logically? Isn't `count ? Infinity` always false?

sri-30: Yeah I've left it like that for now so that the line isn't triggered - should I remove this feature completely?

mvaligursky: I'd suggest we remove it.
And create a new issue on the engine repo with what can be done to improve the PR (and link this PR in the issue). Mention this one bit, but also anything else / links to documents, performance improvement and so on.

mvaligursky:
also, please add some screenshots of the examples to the PR.

src/shape/bounding-box.js
```diff
@@ -421,6 +421,24 @@ class BoundingBox {
...
+     */
+    get halfArea() {
+        const { x, y, z } = this.halfExtents;
+        // Multiply halfExtents by 2 and simplify to least amount of add/mul:
+        // == (x*2)*(y*2) + (z*2)*(x*2) + (y*2)*(z*2)
+        // == 4 * x * y   + 4 * z * x   + 4 * y * z
+        // == 4 * (x * y  + z * x       + y * z)
+        // == 4 * (x * (y + z)          + y * z)
+        // This is for three perpendicular sides, but each side exists twice, so 4 turns into 8:
+        return 8 * (x * (y + z) + y * z);
```
kungfooman, line 439:
```suggestion
        return 4 * (x * (y + z) + y * z);
```

src/shape/bounding-box.js
```diff
@@ -421,6 +421,24 @@ class BoundingBox {
         return sq;
     }
 
+    /**
+     * @example
+     * const halfExtents = new pc.Vec3(2, 3, 4);
+     * const aabb = new pc.BoundingBox(undefined, halfExtents);
+     * console.log("Half surface area of bounding box is", aabb.area); // 208
```
kungfooman, line 428:
```suggestion
     * console.log("Half surface area of bounding box is", aabb.area); // 104
```

kungfooman (changes requested):
Multiplication with 8 is full area and the each-side-twice-comment can be removed now, as its only half the area

src/shape/bounding-box.js
sri-30: Added

src/math/tri.js
```diff
@@ -47,8 +47,10 @@ class Tri {
      * @returns {number} Distance between ray origin and intersection
      */
     intersectWithRay(ray) {
-        h.cross(ray.direction, this.edge2);
-        const a = this.edge1.dot(h);
+        const edge1 = this.edge1;
+        const edge2 = this.edge2;
```
kungfooman, line 51:
```suggestion
        const { edge1, edge2 } = this;
```

src/scene/mesh.js
sri-30: Done

src/math/bvh.js
```diff
@@ -0,0 +1,398 @@
...
+     *
+     * @param {BVHNode} node - The node to be evaluated
+     * @returns {number} The cost of the raycast
+     * @ignore
+     */
+    calculateNodeCost(node) {
+        // Calculate the extent of the node's aabb
+        _tempVec1.sub2(node.aabbMax, node.aabbMin);
+
+        const surfaceArea = _tempVec1.x * _tempVec1.y + _tempVec1.y * _tempVec1.z + _tempVec1.z * _tempVec1.x;
```
kungfooman, line 193: This looks like the `halfArea` calculation and should be a getter in `BVHNode`?

src/scene/mesh-instance.js
```diff
@@ -455,6 +459,29 @@ class MeshInstance {
...
+        _worldTransformInverted.invert();
+        _transformedRay.set(ray.origin, ray.direction);
```
kungfooman, line 472: This is a crutch for a missing `Ray#copy`, maybe just add that?

kungfooman:
A note about `MeshInstance#rayCast(ray)`:

Returning the result in the form of a reference to a temporary variable isn't good practise, you can check how other methods are doing it, like `transformVector(vec, res = new Vec3()) {`

Basically the user tells the method where to save the result, and if the result is not specified, allocate it. As it is right now, its impossible to save a bunch of results without `.copy()`-ing them manually (the results overwrite each other, as they are all the same reference).

Also every trace function I ever used returns more than just the end position, at least people need the surface normal aswell (e.g. to create bullet marks or place an otherwise aligned entity).

There is already a class for this which could be reused:

https://github.com/playcanvas/engine/blob/005871dc5ec721af3cd5ecb5c804e3ad98f7319b/src/framework/components/rigid-body/system.js#L22-L31

Another issue is that Ammo physics sets the standard here by having `raycastFirst(start, end)` and `raycastAll(start, end)`.

IMO that makes more sense, for example if you do mouse/camera picking, you use `camera.screenToWorld`, `nearClip` and `farClip` giving you a pair of start/end points.

Since @willeastcott worked a lot with the Ammo API, what are your thoughts on this? Can e.g. `RaycastResult` just be exported and reused? Or maybe it should be separated, to reduce bundle/treeshaking issues. Or just a new "raycast result class" altogether.

kungfooman:
I stress-tested the performance a little bit to see where it can be improved and without refitting, it boils down to:

image

And:

image

Both methods have lots of potential to be shortened/optimized by removing useless property lookups.

src/shape/ray.js
```diff
@@ -1,6 +1,8 @@
 import { Debug } from '../core/debug.js';
 import { Vec3 } from '../math/vec3.js';
 
+/** @typedef {import('../math/mat4.js').Mat4} Mat4 */
+
```
kungfooman, line 5:
```suggestion
```

src/shape/ray.js
```diff
@@ -48,6 +50,16 @@ class Ray {
         this.direction.copy(direction);
         return this;
     }
+
+    /**
+     * Transforms the ray to a new space via Matrix in-place
+     *
+     * @param {Mat4} matrix - The transformation matrix for the ray
```
kungfooman, line 57:
```suggestion
     * @param {import('../math/mat4.js').Mat4} matrix - The transformation matrix for the ray
```

kungfooman:
I reworked this locally a bit to not mix it up with Mesh and MeshInstance because not everyone requires Bvh and that makes tree-shaking tricky (and also somewhat Separation of Concerns here).

So my idea is just: Bvh and BvhInstance aswell.

So Bvh is holding the tree, but BvhInstance tells you the node for pos/rot/scale to rotate the raycasting ray around.

mvaligursky:
> I reworked this locally a bit to not mix it up with Mesh and MeshInstance because not everyone requires Bvh and that makes tree-shaking tricky (and also somewhat Separation of Concerns here).

Yep agreed, I was thinking about it few weeks ago and reached the same conclusion.

src/scene/mesh-instance.js
kungfooman: `Ray#copy` is in the engine now... just wondering, is this still planned/worked-on?

mvaligursky: it's not being worked on at the moment unfortunately, but hopefully can be picked up a finished one day.

issacclee:
Hi Guys,
Am I right to assume that when this PR is done and merged, we'll effectively get the capabilities of [three-mesh-bvh](https://github.com/gkjohnson/three-mesh-bvh) build into the Playcanvas main engine?

mvaligursky:
> .. we'll effectively get the capabilities of [three-mesh-bvh](https://github.com/gkjohnson/three-mesh-bvh) build into the Playcanvas main engine

That is partially true. It will support fast ray to mesh intersection on CPU, but not on the GPU (yet anyways). It also does not in its current form handles morphing, skinning and instancing. But for common static meshes, that is the case.

src/scene/mesh-instance.js
ErikSom: Why does the meshInstance.raycast return the intersection point and a mesh.raycast the distance? Preferably you would get either in both cases. (btw awesome work on this!)

ErikSom:
Really great work on this code, hope this gets added soon! In the meantime I created this gist for anyone to use as a quick plugin: [Playcanvas Mesh Raycast](https://gist.github.com/ErikSom/621742367628d220def12194f2158863)

PR #4643, opened 2022-09-16
pass area lights LUT directly

yak32:
Fixes #

This PR introduces a change to pass LUTs arrays for area lights directly instead of using a binary asset.
This is necessary to support related change in the editor, where these arrays will be passed automatically.
 
the previous approach is removed without a proper deprecation, as discussed in the comments, because few users using it (the feature wasn't enabled in the editor).

- [x] update Examples browser
- [x] update JSDocs


willeastcott:
Just wondering if it's worth deprecating the old API for this? Was it ever public? Did anyone actually use it? I expect not.

src/deprecated/deprecated.js
```diff
@@ -1154,6 +1157,113 @@ Application.prototype.enableVr = function () {
...
+        asset.ready((asset) => {
+            AreaLightLuts.set(device, asset.resource);
+        });
+        this.assets.load(asset);
+    } else {
+        Debug.warn("setAreaLightLuts: asset is not valid");
+    }
+};
+
+AreaLightLuts.prototype.set = function (device, resource) {
```
mvaligursky, line 1172: that's a large function being deprecated. And the refactored function uses majority of the code. I'd suggest to change this deprecated function to call the new function?

yak32:
> Just wondering if it's worth deprecating the old API for this? Was it ever public? Did anyone actually use it? I expect not.

agree probably nobody, what do you think @mvaligursky ?

mvaligursky:
Few people used it, but not a large amount.
This is the public function:
https://github.com/playcanvas/engine/blob/dbfdb82be845f57b45531d70205feef53cd3a777/src/framework/app-base.js#L1629-L1644

I guess if we simply remove the old functionality, their projects should still work, but area lights would look like normal spot lights. They'd need to go to render settings, and tick the 'support area lights' or whatever this is called. So maybe it's ok to just remove.

slimbuck:
Could you add some justification or rationale behind the change to the PR description?

What's the reason for the change?

yak32:
> Could you add some justification or rationale behind the change to the PR description?
> 
> What's the reason for the change?

did I change PR description?

src/deprecated/deprecated.js
```diff
@@ -1154,6 +1157,34 @@ Application.prototype.enableVr = function () {
     Debug.deprecated('pc.Application#enableVR is deprecated, and WebVR API is no longer supported.');
 };
 
+Application.prototype.setAreaLightLuts = function (asset) {
```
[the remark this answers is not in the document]
mvaligursky: also, please add Debug.deprecated so that people know to upgrade their code.

src/deprecated/deprecated.js
```diff
@@ -112,6 +114,7 @@ import {
 import { RigidBodyComponent } from '../framework/components/rigid-body/component.js';
 import { RigidBodyComponentSystem } from '../framework/components/rigid-body/system.js';
 import { basisInitialize } from '../resources/basis.js';
+import { set } from 'src/core/set-utils.js';
```
mvaligursky, line 117: is this being used?

src/deprecated/deprecated.js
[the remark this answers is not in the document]
yak32: thanks, changed

src/framework/app-base.js
```diff
@@ -1622,19 +1609,18 @@ class AppBase extends EventHandler {
...
-        if (asset) {
+    setAreaLightLuts(ltcMat1, ltcMat2) {
+
+        if (ltcMat1 && ltcMat2) {
             const device = this.graphicsDevice;
-            asset.ready((asset) => {
-                AreaLightLuts.set(device, asset.resource);
-            });
-            this.assets.load(asset);
+            AreaLightLuts.set(device, ltcMat1, ltcMat2);
```
mvaligursky, line 1621:
```suggestion
            AreaLightLuts.set(this.graphicsDevice, ltcMat1, ltcMat2);
```

mvaligursky (approved):
looks good to me

examples/src/examples/graphics/area-lights.tsx
```diff
@@ -16,7 +16,7 @@ class AreaLightsExample {
             'normal': new pc.Asset('normal', 'texture', { url: '/static/assets/textures/seaside-rocks01-normal.jpg' }),
             'gloss': new pc.Asset('gloss', 'texture', { url: '/static/assets/textures/seaside-rocks01-gloss.jpg' }),
             'statue': new pc.Asset('statue', 'container', { url: '/static/assets/models/statue.glb' }),
-            'luts': new pc.Asset('luts', 'binary', { url: '/static/assets/binary/area-light-luts.bin' }),
+            'luts': new pc.Asset('luts', 'json', { url: '/static/assets/binary/area-light-luts.json' }),
```
willeastcott, line 19: If this file is now a `json` file and not a `bin` file, shouldn't it be in a folder called `json` instead of `binary`?

src/framework/app-base.js
```diff
@@ -1622,19 +1609,18 @@ class AppBase extends EventHandler {
     }
 
     /**
-     * Sets the area light LUT asset for this app.
+     * Sets the area light LUT tables for this app.
      *
-     * @param {Asset} asset - LUT asset of type `binary` to be set.
+     * @param {Array.<number>} ltcMat1 - LUT table of type `array` to be set.
+     * @param {Array.<number>} ltcMat2 - LUT table of type `array` to be set.
```
willeastcott, line 1615: The `number[]` notation is more common in the engine source.

willeastcott (approved):
Approving with reasonably minor comments.

examples/src/examples/graphics/area-lights.tsx
yak32: right, changing

PR #4658, opened 2022-09-22
[Fix] Fixed Webgl’s shader handling of lost context

mvaligursky:


src/graphics/webgl/webgl-shader.js
```diff
@@ -14,15 +14,7 @@ import { SHADERTAG_MATERIAL, semanticToLocation } from '../constants.js';
...
-
-        this.uniforms = [];
-        this.samplers = [];
-        this.attributes = [];
-
-        this.glProgram = null;
-        this.glVertexShader = null;
-        this.glFragmentShader = null;
-
+        this.loseContext();
```
willeastcott, line 17: This looks weird to me. I'd rather than `loseContext` function was called something like `init` and `loseContext` called it.

mvaligursky: full agreed.

slimbuck (approved):

PR #4660, opened 2022-09-23
drawQuadWithShader fix

slimbuck:
When target matches the active render target, don't perform begin/end, since it's unnecessary.

### DIFF
--- a/src/graphics/simple-post-effect.js
+++ b/src/graphics/simple-post-effect.js
@@ -53,8 +53,10 @@ function drawQuadWithShader(device, target, shader, rect, scissorRect, useBlend 
     DebugGraphics.pushGpuMarker(device, "drawQuadWithShader");
 
     const oldRt = device.renderTarget;
-    device.setRenderTarget(target);
-    device.updateBegin();
+    if (target !== oldRt) {
+        device.setRenderTarget(target);
+        device.updateBegin();
+    }
 
     let x, y, w, h;
     let sx, sy, sw, sh;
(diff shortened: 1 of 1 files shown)


mvaligursky:
I think we can perhaps skip the begin, as the target would have that called already, but not the end .. what if this is the last end, which cannot be skipped?

slimbuck:
Well, what happens in apps that don't call drawQuadWithShader at all?

slimbuck:
Also, begin/end should not be nested and should have matching pairs. This function calls begin -> end -> being.

Two invocations of this function result in two begins.

mvaligursky:
I think this function is not really designed for render quads to a render target, but it renders a quat to the specified render target .. so it does begin and end there.
The plan is to at some point have a function which only renders quad, without specifying render target - as that would come from render pass.

mvaligursky:
begin -> end -> begin

this is not nested, as the second **begin** is called after the first **begin** was **ended**.

slimbuck:
> begin -> end -> begin
> 
> this is not nested, as the second **begin** is called after the first **begin** was **ended**.

Two invocations of drawQuadWithShader become:

begin -> end -> begin -> begin -> end -> begin

mvaligursky:
yes, that's how it was designed, which I don't like as those are duplicate calls to webgl, but those are still not nested, as the middle begin->begin set up the same render target, and so it's just a duplication (unless filtered deeper), but those are not two nested render target. After rendering the inner one, we don't go back to render to the outer one.

slimbuck:
I'm sorry, this is pretty much the definition of nested.

If the current active rt is null and I do:
```
drawQuadWithShader(rt1) 
drawQuadWithShader(rt2)
```

the result is:
```
updateBegin(rt1)
updateEnd(rt1)
updateBegin(null) <- this has no updateEnd matching updateEnd()
updateBegin(rt2)  <- this is nested inside updateBegin(null)
updateEnd(rt2)
updateBegin(null)
```

It fails both things specified here https://github.com/playcanvas/engine/blob/main/src/graphics/webgl/webgl-graphics-device.js#L1387.

The practical result (as you know) is that we're currently doing a bunch of unnecessary work.

Anyway, I agree that changing this here might cause applications issues, but we need to figure out a solution sooner rather than later.

mvaligursky:
I think nested in this context, which is not supported, is that when you call two begins and then only a single end, that you can continue rendering to the target set up with the first begin.

something like:

```
begin 1
    begin 2
    end 2
    // you cannot expect to rendering to target 1 here, as nesting is not supported
end 1
```

we have this ticket to remember to implement render passes / linking it for reference https://github.com/playcanvas/engine/issues/4271

mvaligursky:
This is my ultimate goal which I'd like to work on soonish, as needed for WebGPU: https://github.com/playcanvas/engine/issues/4843

GSterbrant (dismissed):

mvaligursky:
Lets not merge this as I'm rewriting this part at the moment. Maybe we should close it @slimbuck ?

GSterbrant:
I didn't see the conversation, just looked at the code. Yeah probably a good idea not to merge this yet.

PR #4661, opened 2022-09-23
Resolve render target before generating mips

slimbuck:
Makes sense to resolve render target before generating it's texture mipmaps.


mvaligursky (approved):
Makes sense. Please test with at least post-effect engine example to make sure that works fine.

Issue #3748, opened 2021-11-29
If a draggable element entity parent is moving while the element is dragged, it will move in local Z


yaustar:
Example project: https://playcanvas.com/editor/scene/1282724

Expected behaviour is for the element entity to stay in its current local Z position or the local Z position from when it was started to be dragged.

This issue affects scroll views/elements.

https://user-images.githubusercontent.com/16639049/143869233-a51c7974-fb4e-4563-b80c-5bd2311ef550.mp4

As a workaround, a patch can be made to the ElementDragHelper:

```
(function () {
    const OPPOSITE_AXIS = {
        x: 'y',
        y: 'x'
    };
    pc.ElementDragHelper.prototype._onMove = function (event) {
        if (this._element && this._isDragging && this.enabled && this._element.enabled && this._element.entity.enabled) {
            const currentMousePosition = this._screenToLocal(event);

            if (this._dragStartMousePosition && currentMousePosition) {
                this._deltaMousePosition.copy(currentMousePosition).sub(this._dragStartMousePosition);
                this._deltaHandlePosition.copy(this._dragStartHandlePosition).add(this._deltaMousePosition);
                const currentPosition = this._element.entity.getLocalPosition();

                if (this._axis) {
                    const constrainedAxis = OPPOSITE_AXIS[this._axis];
                    this._deltaHandlePosition[constrainedAxis] = currentPosition[constrainedAxis];
                }

                // Enforce no movement on the Z axis
                this._deltaHandlePosition.z = currentPosition.z;
                this._element.entity.setLocalPosition(this._deltaHandlePosition);
                this.fire('drag:move', this._deltaHandlePosition);
            }
        }
    };
})();
```

kungfooman:
Just tested and this issue is also fixed: https://launch.playcanvas.com/1282724?debug=true&use_local_engine=https://code.playcanvas.com/playcanvas-1.57.0.js

yaustar:
Fixed with #4666

PR #4666, opened 2022-09-23
ElementDragHelper: Fix drag NaN

### LINKED ISSUE #4635: Scrollview content disappears, Content entity recieving localPosition of NaN
While the debug console is opened and the Device toggler is set to mobile the scrollView Content localPosition sets to NaN. Scrollable content disappears. We noticed that this issue is present almost only if the Debug console is on and mobile mode is toggled. 

https://user-images.githubusercontent.com/10512039/189824028-b10012e0-5145-4603-9d9f-bf4eb56e169c.mp4

### Steps to Reproduce

1. Open https://playcanv.as/b/E9AhLZ7J/ or Fork the project https://playcanvas.com/editor/scene/1319039 and hit Launch
2. Open Dev Console and toggle the mobile view (setting the resolution to 407px x 745px or smaller) 
3. Hit Inventory / Toggle button
4. Drag the contents of the inventory ScrollView, !!! drag fast and nervous =)  so the cursor get's out of the browser's window. 
5. Check out the inventory-ui.js (lines 19-29) script for the console.trace() hook

kungfooman:
Fixes #4635 and #3748

The deltas in `ElementDragHelper#_onMove` become larger every frame, because one is using global position and the other takes local position... that goes well for a few frames, but then explodes into either extremely near 0 or Infinity. And in the case of Infinity it perpetuates itself further into NaN.

I also did some refactoring, which reduces the amount of code while using less method calls and less property lookups, which also reduces the amount of code.

It was quite some work to actually find the bug Infinity/NaN bug, so I added JSDoc along the way...

### DIFF
--- a/src/framework/components/element/component.js
+++ b/src/framework/components/element/component.js
@@ -1021,8 +1021,10 @@ class ElementComponent extends Component {
      * @private
      */
     _setPosition(x, y, z) {
-        if (!this.element.screen)
-            return Entity.prototype.setPosition.call(this, x, y, z);
+        if (!this.element.screen) {
+            Entity.prototype.setPosition.call(this, x, y, z);
+            return;
+        }
 
         if (x instanceof Vec3) {
             position.copy(x);
--- a/src/framework/components/element/component.js
+++ b/src/framework/components/element/component.js
@@ -674,27 +674,33 @@ class ElementComponent extends Component {
      * The position of the pivot of the component relative to its anchor. Each value ranges from 0
      * to 1 where [0,0] is the bottom left and [1,1] is the top right.
      *
-     * @type {Vec2}
+     * @example
+     * pc.app.root.findByName("Inventory").element.pivot = [Math.random() * 0.1, Math.random() * 0.1];
+     * @example
+     * pc.app.root.findByName("Inventory").element.pivot = new pc.Vec2(Math.random() * 0.1, Math.random() * 0.1);
+     * 
+     * @type {Vec2 | number[]}
      */
     set pivot(value) {
-        const prevX = this._pivot.x;
-        const prevY = this._pivot.y;
+        const { _pivot, _margin } = this;
+        const prevX = _pivot.x;
+        const prevY = _pivot.y;
 
         if (value instanceof Vec2) {
-            this._pivot.set(value.x, value.y);
+            _pivot.copy(value);
         } else {
-            this._pivot.set(value[0], value[1]);
+            _pivot.set(...value);
         }
 
-        const mx = this._margin.x + this._margin.z;
-        const dx = this._pivot.x - prevX;
-        this._margin.x += mx * dx;
-        this._margin.z -= mx * dx;
+        const mx = _margin.x + _margin.z;
+        const dx = _pivot.x - prevX;
+        _margin.x += mx * dx;
+        _margin.z -= mx * dx;
 
-        const my = this._margin.y + this._margin.w;
-        const dy = this._pivot.y - prevY;
-        this._margin.y += my * dy;
-        this._margin.w -= my * dy;
+        const my = _margin.y + _margin.w;
+        const dy = _pivot.y - prevY;
+        _margin.y += my * dy;
+        _margin.w -= my * dy;
 
         this._anchorDirty = true;
         this._cornersDirty = true;
@@ -706,7 +712,7 @@ class ElementComponent extends Component {
         // in order for them to update their position
         this._flagChildrenAsDirty();
 
-        this.fire('set:pivot', this._pivot);
+        this.fire('set:pivot', _pivot);
     }
 
     get pivot() {
--- a/src/framework/components/element/component.js
+++ b/src/framework/components/element/component.js
@@ -1018,7 +1018,7 @@ class ElementComponent extends Component {
      * @param {number|Vec3} x - The x coordinate or Vec3
      * @param {number} y - The y coordinate
      * @param {number} z - The z coordinate
-     * @returns {void}
+     * @private
      */
     _setPosition(x, y, z) {
         if (!this.element.screen)
@@ -1044,6 +1044,7 @@ class ElementComponent extends Component {
      * @param {number|Vec3} x - The x coordinate or Vec3
      * @param {number} y - The y coordinate
      * @param {number} z - The z coordinate
+     * @private
      */
     _setLocalPosition(x, y, z) {
         if (x instanceof Vec3) {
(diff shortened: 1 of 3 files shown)


yaustar:
Where's the actual fix as you made some code improvements along the way that makes it hard to see the actual fix 😅

src/framework/components/element/element-drag-helper.js
```diff
@@ -140,16 +140,16 @@ class ElementDragHelper extends EventHandler {
         this._determineInputPosition(event);
         this._chooseRayOriginAndDirection();
 
-        _planeOrigin.copy(this._element.entity.getPosition());
-        _planeNormal.copy(this._element.entity.forward).mulScalar(-1);
+        _plane.point.copy(this._element.entity.getLocalPosition());
```
kungfooman, line 143: @yaustar this is the actual fix 😅

yaustar: Oooh, that's interesting!

kungfooman:
@yaustar Yes, this PR also fixes your issue https://github.com/playcanvas/engine/issues/3748:

fixMovingDragNaN

src/framework/components/element/component.js
```diff
@@ -1001,6 +1012,14 @@ class ElementComponent extends Component {
         this.entity.setLocalPosition = Entity.prototype.setLocalPosition;
     }
 
+    /**
+     * Patched method for setting the position.
+     *
+     * @param {number|Vec3} x - The x coordinate or Vec3
+     * @param {number} y - The y coordinate
+     * @param {number} z - The z coordinate
+     * @returns {void}
```
willeastcott, line 1021: This line is redundant so best to remove it.

src/framework/components/element/component.js
```diff
@@ -1001,6 +1012,14 @@ class ElementComponent extends Component {
...
+     */
```
willeastcott, line 1022: This JSDoc block presumably needs a @private tag?

willeastcott: Same for JSDocs on other `_` functions in this PR.

src/framework/components/element/component.js
```diff
@@ -1563,10 +1589,15 @@ class ElementComponent extends Component {
...
 
-    // recalculates
-    // localAnchor, width, height, (local position is updated if anchors are split)
-    // assumes these properties are up to date
-    // _margin
+    /**
+     * recalculates
+     * localAnchor, width, height, (local position is updated if anchors are split)
+     * assumes these properties are up to date
+     * _margin
```
willeastcott, line 1596: Reformat assuming a max columns of 100.

src/framework/components/element/component.js
kungfooman: It triggered a linting error because of `return Entity.prototype.setPosition.call(this, x, y, z);`, so I just changed that now to make it clearer that there is really no return value.

src/framework/components/element/component.js
```diff
@@ -674,27 +679,33 @@ class ElementComponent extends Component {
...
+     * pc.app.root.findByName("Inventory").element.pivot = [Math.random() * 0.1, Math.random() * 0.1];
+     * @example
+     * pc.app.root.findByName("Inventory").element.pivot = new pc.Vec2(Math.random() * 0.1, Math.random() * 0.1);
+     *
+     * @type {Vec2 | number[]}
      */
     set pivot(value) {
-        const prevX = this._pivot.x;
-        const prevY = this._pivot.y;
+        const { _pivot, _margin } = this;
```
mvaligursky, line 690: why local variables with _ ? I'd remove it.

src/framework/components/element/component.js
```diff
@@ -1563,10 +1592,18 @@ class ElementComponent extends Component {
         this.off();
     }
 
-    // recalculates
-    // localAnchor, width, height, (local position is updated if anchors are split)
-    // assumes these properties are up to date
-    // _margin
+    /**
+     * recalculates
```
mvaligursky, line 1596: If we make this jsdocs, it'd be good to format it nicer as well (even when private)

src/framework/components/element/element-drag-helper.js
```diff
@@ -210,21 +210,21 @@ class ElementDragHelper extends EventHandler {
     }
 
     _onMove(event) {
-        if (this._element && this._isDragging && this.enabled && this._element.enabled && this._element.entity.enabled) {
+        const { _element, _deltaMousePosition, _deltaHandlePosition, _axis } = this;
+        if (_element && this._isDragging && this.enabled && _element.enabled && _element.entity.enabled) {
             const currentMousePosition = this._screenToLocal(event);
+            if (currentMousePosition) {
```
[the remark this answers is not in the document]
kungfooman: @jpauloruschel Yes, its assigned in the constructor and never deleted. Also other code depends on it being never deleted, like this (line 122): https://github.com/playcanvas/engine/blob/6c7d3dc2dae80364dfedebb2bd371a48aba976d2/src/framework/components/element/element-drag-helper.js#L112-L128

Therefore checking if a reference exists (which always exists), is a no-op.

src/framework/components/element/component.js
kungfooman: Do you mean like `const { _pivot: pivot, _margin: margin } = this;` instead?

mvaligursky: yep, that would do.

src/framework/components/element/component.js
kungfooman: Okay, I worked it out a little bit: 
image

I renamed `localAnchor` to `_localAnchor` aswell, becaues there is no getter for `localAnchor`.

src/framework/components/element/component.js
kungfooman: I just realized its also possible to just write `const { pivot, margin } = this;`, since `get pivot()` is just returning `_pivot` anyway, which would decrease the amount of code here. Is that okay with you?

mvaligursky: totally

src/framework/components/element/component.js
```diff
@@ -397,13 +397,18 @@ class ElementComponent extends Component {
...
+     *
+     * @type {Vec4 | number[]}
      */
     set anchor(value) {
         if (value instanceof Vec4) {
-            this._anchor.set(value.x, value.y, value.z, value.w);
+            this._anchor.copy(value);
         } else {
-            this._anchor.set(value[0], value[1], value[2], value[3]);
+            this._anchor.set(...value);
```
mvaligursky, line 411: one more tiny request .. I'd prefer to previous version of this line, as the changed version allocates an array internally I believe, I'd prefer to avoid that.

src/framework/components/element/component.js
```diff
@@ -674,27 +679,33 @@ class ElementComponent extends Component {
...
+        const { pivot, margin } = this;
+        const prevX = pivot.x;
+        const prevY = pivot.y;
 
         if (value instanceof Vec2) {
-            this._pivot.set(value.x, value.y);
+            pivot.copy(value);
         } else {
-            this._pivot.set(value[0], value[1]);
+            pivot.set(...value);
```
mvaligursky, line 697: and the same here.

src/framework/components/element/component.js
kungfooman: It's just using `.apply`, passing through the array.

mvaligursky: ah right, the value is already array, I missed that somehow.

mvaligursky (approved):
Looks great, thanks for the work!

src/framework/components/element/component.js
kungfooman: Same here, the ES5 version is just something like:
```js
value = [1, 2, 3];
console.log.apply(null, value);
```
Is it okay to keep? I would argue it reduces the amount of stack pollution / needless array unpacking

mvaligursky: I still believe the original code would run faster, but this is not critical path so it's ok.

jpauloruschel (approved):

PR #4669, opened 2022-09-26
Only add the shared sampler if the chunk uses the $SAMPLER macro.

GSterbrant:
### Description

Make sure chunks that don't use the $SAMPLER macro doesn't have its samplers declared in the preamble of the shader frontend.

### DIFF
--- a/src/graphics/program-lib/programs/standard.js
+++ b/src/graphics/program-lib/programs/standard.js
@@ -157,7 +157,7 @@ const standard = {
 
             subCode = subCode.replace(/\$UV/g, uv).replace(/\$CH/g, options[channelPropName]);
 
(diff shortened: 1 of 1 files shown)


src/graphics/program-lib/programs/standard.js
```diff
@@ -157,7 +157,7 @@ const standard = {
 
             subCode = subCode.replace(/\$UV/g, uv).replace(/\$CH/g, options[channelPropName]);
 
-            if (mapping) {
+            if (mapping && subCode.search(/\$SAMPLER/g)) {
```
slimbuck, line 160:
```suggestion
            if (mapping && subCode.search(/\$SAMPLER/g) !== -1) {
```

slimbuck (approved):

PR #4674, opened 2022-09-28
Fix texture sharing change triggering shader build

GSterbrant:
### Description

With the new shared sampler code, it's important we know that a shader has to be rebuilt whenever it changes a texture since that might change the way textures are being shared. This PR fixes that issue by always setting the xMapIdentifier in the options such that it becomes part of the options key. 

This PR also defers updating the material until after all textures have loaded. This is to save time not having to potentially recompiling shaders in between each texture load.

### DIFF
--- a/src/resources/material.js
+++ b/src/resources/material.js
@@ -186,8 +186,7 @@ class MaterialHandler {
 
     _onTextureLoad(parameterName, materialAsset, textureAsset) {
         this._assignTexture(parameterName, materialAsset, textureAsset.resource);
-        if (--materialAsset.textureLoadCount === 0)
-            materialAsset.resource.update();
+        materialAsset.resource.update();
     }
 
     _onTextureAdd(parameterName, materialAsset, textureAsset) {
@@ -248,8 +247,6 @@ class MaterialHandler {
         const TEXTURES = standardMaterialTextureParameters;
 
         let i, name, assetReference;
-        materialAsset.textureLoadCount = 0;
-
         // iterate through all texture parameters
         for (i = 0; i < TEXTURES.length; i++) {
             name = TEXTURES[i];
@@ -267,7 +264,6 @@ class MaterialHandler {
 
             if (dataAssetId && (!materialTexture || !dataValidated || isPlaceHolderTexture)) {
                 if (!assetReference) {
-                    materialAsset.textureLoadCount++;
                     assetReference = new AssetReference(name, materialAsset, assets, {
                         load: this._onTextureLoad,
                         add: this._onTextureAdd,
@@ -288,7 +284,6 @@ class MaterialHandler {
                 if (assetReference.asset) {
                     if (assetReference.asset.resource) {
                         // asset is already loaded
-                        materialAsset.textureLoadCount--;
                         this._assignTexture(name, materialAsset, assetReference.asset.resource);
                     } else {
                         this._assignPlaceholderTexture(name, materialAsset);
--- a/src/resources/material.js
+++ b/src/resources/material.js
@@ -186,7 +186,8 @@ class MaterialHandler {
 
     _onTextureLoad(parameterName, materialAsset, textureAsset) {
         this._assignTexture(parameterName, materialAsset, textureAsset.resource);
-        materialAsset.resource.update();
+        if (--materialAsset.textureLoadCount === 0)
+            materialAsset.resource.update();
     }
 
     _onTextureAdd(parameterName, materialAsset, textureAsset) {
@@ -247,6 +248,8 @@ class MaterialHandler {
         const TEXTURES = standardMaterialTextureParameters;
 
         let i, name, assetReference;
+        materialAsset.textureLoadCount = 0;
+
         // iterate through all texture parameters
         for (i = 0; i < TEXTURES.length; i++) {
             name = TEXTURES[i];
@@ -264,6 +267,7 @@ class MaterialHandler {
 
             if (dataAssetId && (!materialTexture || !dataValidated || isPlaceHolderTexture)) {
                 if (!assetReference) {
+                    materialAsset.textureLoadCount++;
                     assetReference = new AssetReference(name, materialAsset, assets, {
                         load: this._onTextureLoad,
                         add: this._onTextureAdd,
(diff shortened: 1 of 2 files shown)


src/scene/materials/standard-material-options-builder.js
```diff
@@ -348,7 +349,7 @@ class StandardMaterialOptionsBuilder {
                 if (stdMat[uname] === 1 && !hasUv1) allow = false;
                 if (allow) {
                     options[mname] = !!stdMat[mname];
-                    options[iname] = stdMat[mname].name;
+                    options[iname] = stdMat[mname].id;
```
mvaligursky, line 352: actually I'm not sure I like new version, or old version, as this would make a unique shader for each unique texture, causing a lot more shaders getting compiled if material simply uses a different texture.

mvaligursky: enable tracing of shader creations this way and test with / without this line, it will likely generate more shaders if I understand it correctly
`pc.Tracing.set(pc.TRACEID_SHADER_ALLOC, true);`

Maksims:
Slightly off-topic, but related to what this issue tries to address: having a small placeholder texture while the actual texture is loading could be a way to tackle many issues.

src/scene/materials/standard-material-options-builder.js
GSterbrant: I agree, I was thinking how this could be prevented, and I figured this would work:

```js
                    const mapId = stdMat[mname].id;
                    let identifier = textureMap[mapId];
                    if (identifier === undefined) {
                        textureMap[mapId] = textureMappingCounter.value;
                        identifier = textureMappingCounter.value;
                        textureMappingCounter.value++;
                    }
                    ...
                    options[iname] = identifier;

```

Where textureMap and textureMappingCounter is shared over all textures. This way, the mapping counter will be incremented deterministically for each material, so if there is an identical mapping between two changes, there is no change in the shader.

GSterbrant:
@Maksims We already use a placeholder texture in place when a texture is loading. One of the problems is that for each placeholder getting replaced with a loaded one, we do material.update() which will trigger the material to rebuild the shader. Preferably, we'd do this when all textures have been loaded.

Maksims:
> @Maksims We already use a placeholder texture in place when a texture is loading. One of the problems is that for each placeholder getting replaced with a loaded one, we do material.update() which will trigger the material to rebuild the shader. Preferably, we'd do this when all textures have been loaded.

If material shader is already built, then material.update, should only result in getting the shader from a cache. No recompilation should happen.

src/scene/materials/standard-material-options-builder.js
mvaligursky: I'm not sure I understand. It seems that you're creating a map, which maps unique texture.id into some numerical id that gets incremented for each texture this code encounters?

The problem with solving this on the options-builder side is that it might not match what the actual shader generator does I think. Some chunks (non-updated user chunks, or lightmapper chunk) are not sharing their samplers, and this won't handle it? This could cause incorrect shader being pulled from the cache, and hard to track incorrect textures bug introduced.

GSterbrant: Yeah, what this will do is to avoid accidentally generating a new shader if the texture sharing is identical but the texture ids mismatch. With this, the ids can't mismatch because we generate the identifiers in the function when going over all the supported textures in the material.

This should make sure the identifiers are deterministic if the texture changes but the sharing remains identical. 

For chunks that don't share their samplers, do we need to worry about it? As long as they define the samplers, don't use the $SAMPLER macro, and provide a texture it shouldn't be an issue?

mvaligursky: yep, ok, sounds good to me I think.

GSterbrant:
@Maksims To figure out if several front end chunks can use the same texture, thereby reducing the amount of samplers in the shader, we might need to recompile the shaders if the textures change. In fact, we already do with some, for example cubemaps and such, in order to avoid sampling from an unbound texture. 

This doesn't really force update either, it checks whether or not the options have changed and trigger a recompile if it's the case there is a difference. 

This behaviour is essentially unchanged from before, the only difference is that we don't recompile shaders per each texture set, but rather when all textures are loaded.

src/resources/material.js
```diff
@@ -186,7 +186,8 @@ class MaterialHandler {
 
     _onTextureLoad(parameterName, materialAsset, textureAsset) {
         this._assignTexture(parameterName, materialAsset, textureAsset.resource);
-        materialAsset.resource.update();
+        if (--materialAsset.textureLoadCount === 0)
```
mvaligursky, line 189: The changes in this file look good, but I don't understand it all enough to be confident. I assume _bindAndAssignAssets is called more often when in Editor / live launcher and the material is being edited (textures added / removed), so please test this well. It's tricky to maintain a count in a case like this, when things change.

src/scene/materials/standard-material-options-builder.js
```diff
@@ -347,8 +351,17 @@ class StandardMaterialOptionsBuilder {
                 if (stdMat[uname] === 0 && !hasUv0) allow = false;
                 if (stdMat[uname] === 1 && !hasUv1) allow = false;
                 if (allow) {
+
```
mvaligursky, line 354: you mentioned you wanted to add some comments .. please do so, next year it will be less clear

src/scene/materials/standard-material-options-builder.js
```diff
@@ -106,8 +106,11 @@ class StandardMaterialOptionsBuilder {
 
         options.vertexColors = false;
         this._mapXForms = [];
+
+        const textureMappingCounter = { value: 0 };
```
mvaligursky, line 110: I'd change it to `this.textureMappingCounter = 0;` to keep it simpler

src/scene/materials/standard-material-options-builder.js
```diff
@@ -347,8 +351,17 @@ class StandardMaterialOptionsBuilder {
                 if (stdMat[uname] === 0 && !hasUv0) allow = false;
                 if (stdMat[uname] === 1 && !hasUv1) allow = false;
                 if (allow) {
+
+                    const mapId = stdMat[mname].id;
+                    let identifier = textureMap[mapId];
+                    if (identifier === undefined) {
+                        textureMap[mapId] = textureMappingCounter.value;
```
mvaligursky, line 358: I'd change the names to make it clearer. `textureMap` usually means a texture map.
textureMap -> uniqueTexturesMap
identifier -> uniqueTextureId

mvaligursky (approved):
approving with some comments

src/resources/material.js
GSterbrant: I tested this in the editor with changing the sharing, adding/removing textures and reloading the project. Is there some other test that would better check if this might cause issues?

slimbuck (approved):

PR #4679, opened 2022-09-29
Small improvements to physical lights units example

mvaligursky:
Small updates:
- camera bit closer
- enabled area light, so the area light reflection is a rect shape
- sizes / intensities
- changed some HUD names to fit without clipping
- wider panel to fit large range of values

Screenshot 2022-09-29 at 12 07 44

### DIFF
--- a/examples/src/examples/graphics/light-physical-units.tsx
+++ b/examples/src/examples/graphics/light-physical-units.tsx
@@ -11,45 +11,45 @@ class LightPhysicalUnitsExample {
     controls(data: Observer) {
         return <>
             <Panel headerText='Lights'>
-                <LabelGroup text='Sun (lm/m2)'>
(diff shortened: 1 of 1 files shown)


GSterbrant (approved):

willeastcott (approved):
Nice - looks much better!

willeastcott:
One minor point though. The UI panel labels are mostly truncated. Can we fix that perhaps?

MAG-AdrianMeredith:
am I reading that right that the shutter speed is set to 1 second?  That seems like an odd value that would introduce a ton of motion blur if it were real

mvaligursky:
> am I reading that right that the shutter speed is set to 1 second? That seems like an odd value that would introduce a ton of motion blur if it were real

For cameras, the units used for Shutter Speed are 1/x seconds, so in this case its 1/1000s

PR #4683, opened 2022-09-30
[Fix] GltfExporter correctly handles multiple mesh instances per entity

mvaligursky:
- before, we would walk all mesh instances in a render component, and store their index as a mesh index of a node. The problem was that if there were multiple mesh instances, the index would get overwritten and only the last mesh would be assigned
- this PR changes the way mesh instances are collected to collect them per node. And from all of these, a single gltf mesh is created, with each mesh instance forming a primitive.
- this also makes export of model components working - as all mesh instances are attached to a root node, even though internally they store the node they need the transform from.

In the screenshot, both a sphere and a cone are added to a single render component, and both meshes are exported. Before, only the cone (second mesh) would be visible.
Screenshot 2022-09-30 at 14 08 36

### DIFF
--- a/extras/exporters/gltf-exporter.js
+++ b/extras/exporters/gltf-exporter.js
@@ -59,17 +59,18 @@ class GltfExporter {
             cameras: [],
             entities: [],
             materials: [],
-            meshInstances: []
+
+            // entry: { node, meshInstances}
+            entityMeshInstances: []
         };
 
-        const { materials, buffers } = resources;
+        const { materials, buffers, entityMeshInstances } = resources;
 
         // Collect entities
         root.forEach((entity) => {
             resources.entities.push(entity);
         });
 
-        // Collect materials
         const collectMeshInstances = (meshInstances) => {
             meshInstances.forEach((meshInstance) => {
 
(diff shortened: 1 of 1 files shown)


extras/exporters/gltf-exporter.js
```diff
@@ -101,11 +108,11 @@ class GltfExporter {
                 resources.cameras.push(entity.camera);
             }
 
-            if (entity.render) {
+            if (entity.render && entity.render.enabled) {
                 collectMeshInstances(entity.render.meshInstances);
             }
 
-            if (entity.model) {
+            if (entity.model && entity.model.enabled) {
```
kungfooman, line 115:
```suggestion
            if (entity.model && entity.model.enabled && entity.model.meshInstances) {
```

`ModelComponent#meshInstances` is of type `MeshInstance[]|null`, so this code is crashing currently:

```js
const e = new pc.Entity();
e.addComponent("model");
app.root.addChild(e);
```

kungfooman: This is an error which developers trip over all the time, maybe the type `Array|null` should be prevented in the `ModelComponent` itself? I don't really like checking everything to be an array or `null` and everything crashes when its forgotten to check.

Last time I run into this: https://github.com/playcanvas/editor/issues/735

mvaligursky: thanks, good point.

PR #4691, opened 2022-10-03
Fix GLB punctual lights with physical units

GSterbrant:
### Description

The GLB parser doesn't set the luminance parameter on the light entities it creates, resulting in no lights being visible in GLBs when physical lights are enabled. 

This PR also fixes another bug where sky light intensity can't be 0 when physical units are enabled.

### DIFF
--- a/src/resources/parser/glb-parser.js
+++ b/src/resources/parser/glb-parser.js
@@ -34,11 +34,9 @@ import {
     ASPECT_MANUAL, ASPECT_AUTO, SPECOCC_AO
 } from '../../scene/constants.js';
 
-import { lightTypes } from '../../framework/components/light/system.js';
-
 import { calculateNormals } from '../../scene/procedural.js';
 import { GraphNode } from '../../scene/graph-node.js';
-import { Light } from '../../scene/light.js';
+import { Light, lightTypes } from '../../scene/light.js';
 import { Mesh } from '../../scene/mesh.js';
 import { Morph } from '../../scene/morph.js';
 import { MorphTarget } from '../../scene/morph-target.js';
--- a/src/resources/parser/glb-parser.js
+++ b/src/resources/parser/glb-parser.js
@@ -1794,8 +1794,7 @@ const createLight = function (gltfLight, node) {
                 lightProps.luminance = gltfLight.intensity * Light.getLightUnitConversion(LIGHTTYPE_OMNI);
                 break;
             case "directional":
-                // Directional light luminance is already in lux
-                lightProps.luminance = gltfLight.intensity;
+                lightProps.luminance = gltfLight.intensity * Light.getLightUnitConversion(LIGHTTYPE_DIRECTIONAL);
                 break;
         }
     }
--- a/src/resources/parser/glb-parser.js
+++ b/src/resources/parser/glb-parser.js
@@ -30,6 +30,7 @@ import { VertexFormat } from '../../graphics/vertex-format.js';
 
 import {
     BLEND_NONE, BLEND_NORMAL, LIGHTFALLOFF_INVERSESQUARED,
+    LIGHTTYPE_DIRECTIONAL, LIGHTTYPE_OMNI, LIGHTTYPE_SPOT,
     PROJECTION_ORTHOGRAPHIC, PROJECTION_PERSPECTIVE,
     ASPECT_MANUAL, ASPECT_AUTO, SPECOCC_AO
 } from '../../scene/constants.js';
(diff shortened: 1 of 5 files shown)


src/resources/parser/glb-parser.js
```diff
@@ -1782,6 +1783,24 @@ const createLight = function (gltfLight, node) {
         lightProps.outerConeAngle = gltfLight.spot.hasOwnProperty('outerConeAngle') ? gltfLight.spot.outerConeAngle * math.RAD_TO_DEG : Math.PI / 4;
     }
 
+    // glTF stores light already in energy/area, but we need to provide the light with only the energy parameter,
```
mvaligursky, line 1786: We already have the same code to compute coefficients in the Light class. There we divide by them, here we multiply. We'll also need them in the gltf-exporter soon. And so I wonder if those should be in a place where we can share them.
Just not sure where to put them. Perhaps a static function on the Light class?

GSterbrant: That's a good point actually. Since they basically just act as multipliers, we could move them to somewhere shared. I'll do that 🙂

src/resources/parser/glb-parser.js
```diff
@@ -1774,6 +1774,7 @@ const createLight = function (gltfLight, node) {
         // - Omni and spot lights use luminous intensity in candela (lm/sr)
         // - Directional lights use illuminance in lux (lm/m2).
         // Current implementation: clapms specified intensity to 0..2 range
+        luminance: gltfLight.intensity,
```
mvaligursky, line 1777: see the line below - that's how it should be set, making sure the property exists first.

src/scene/light.js
```diff
@@ -637,6 +637,31 @@ class Light {
         return clone;
     }
 
+    /**
+     * Get conversion factor for luminance -> light specific light unit
+     * @param {number} type - The type of light
+     * @param {number} outerAngle - The outer angle of a spot light
+     * @param {number} innerAngle - The inner angle of a spot light
```
mvaligursky, line 644: Marking those params as optional
```suggestion
     * @param {number} [outerAngle] - The outer angle of a spot light
     * @param {number} [innerAngle] - The inner angle of a spot light
```

src/resources/parser/glb-parser.js
```diff
@@ -1785,20 +1785,19 @@ const createLight = function (gltfLight, node) {
...
+            case "spot":
+                lightProps.luminance = gltfLight.intensity * Light.getLightUnitConversion(LIGHTTYPE_SPOT, gltfLight.spot.outerConeAngle, gltfLight.spot.innerConeAngle);
+                break;
+            case "point":
+                lightProps.luminance = gltfLight.intensity * Light.getLightUnitConversion(LIGHTTYPE_OMNI);
+                break;
+            case "directional":
+                // Directional light luminance is already in lux
+                lightProps.luminance = gltfLight.intensity;
+                break;
```
mvaligursky, line 1798: I think this whole section could be as simple as this
```
lightProps.luminance = gltf.intensity * Light.getLightUnitConversion(lightProps.type, lightProps.innerConeAngle * math.DEG_TO_RAD, lightProps.outerConeAngle * math.DEG_TO_RAD);
```

src/scene/light.js
```diff
@@ -637,6 +637,31 @@ class Light {
         return clone;
     }
 
+    /**
+     * Get conversion factor for luminance -> light specific light unit
```
mvaligursky, line 641: all jsdocs end with '.' - function description and also parameter description.

src/scene/light.js
```diff
@@ -735,22 +760,12 @@ class Light {
...
         if (this._usePhysicalUnits) {
             switch (this._type) {
-                case LIGHTTYPE_SPOT: {
-                    const falloffEnd = Math.cos(this._outerConeAngle * Math.PI / 180.0);
-                    const falloffStart = Math.cos(this._innerConeAngle * Math.PI / 180.0);
-
-                    // https://github.com/mmp/pbrt-v4/blob/faac34d1a0ebd24928828fe9fa65b65f7efc5937/src/pbrt/lights.cpp#L1463
-                    i = this._luminance / (2 * Math.PI * ((1 - falloffStart) + (falloffStart - falloffEnd) / 2.0));
+                case LIGHTTYPE_SPOT:
+                    i = this._luminance / Light.getLightUnitConversion(this._type, this._outerConeAngle * Math.PI / 180.0, this._innerConeAngle * Math.PI / 180.0);
```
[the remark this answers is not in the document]
mvaligursky: also, considering we do this on multiple places .. perhaps getLightUnitConversion should simply accept angles in DEG and convert to RAD internally.

src/resources/parser/glb-parser.js
GSterbrant: I wanted to avoid having to do the math with converting degrees to radians and providing those arguments, but since it's loader code it's not so vital I suppose.

src/scene/light.js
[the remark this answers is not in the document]
GSterbrant: Does it matter though? The GLBs provide the angle in radians, and we provide it in degrees. So either way, you'd have to perform a conversion before passing the value to the function.

src/resources/parser/glb-parser.js
```diff
@@ -34,8 +34,11 @@ import {
     ASPECT_MANUAL, ASPECT_AUTO, SPECOCC_AO
 } from '../../scene/constants.js';
 
+import { lightTypes } from '../../framework/components/light/system.js';
```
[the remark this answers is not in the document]
GSterbrant: Moved it to scene/light.js and updated imports.

mvaligursky (approved):

PR #4692, opened 2022-10-03
Refactored ProgramLibrary to use the DeviceCache

mvaligursky:
- removed an instance of ProgramLibrary from GraphicsDevice, removing a circular references, and breaking dependency of Graphics on what should be a higher level library

deprecated function on the GraphicsDevice:
```
getProgramLibrary();
setProgramLibrary(programLib);
removeShaderFromCache(shader);
```

removed circular dependencies:
```
src/graphics/program-lib/utils.js -> src/graphics/program-library.js -> src/scene/materials/standard-material.js
 -> src/graphics/env-lighting.js -> src/graphics/reproject-texture.js -> src/graphics/program-lib/utils.js

src/graphics/program-library.js -> src/scene/materials/standard-material.js -> src/graphics/env-lighting.js
 -> src/graphics/reproject-texture.js -> src/graphics/program-library.js

src/graphics/program-library.js -> src/scene/materials/standard-material.js -> src/scene/materials/material.js
 -> src/graphics/program-library.js
```

### DIFF
--- a/src/deprecated/deprecated.js
+++ b/src/deprecated/deprecated.js
@@ -41,7 +41,8 @@ import { GraphicsDevice } from '../graphics/graphics-device.js';
 import { IndexBuffer } from '../graphics/index-buffer.js';
 import { createFullscreenQuad, drawFullscreenQuad, PostEffect } from '../graphics/post-effect.js';
 import { PostEffectQueue } from '../framework/components/camera/post-effect-queue.js';
-import { getProgramLibrary, ProgramLibrary } from '../graphics/program-library.js';
+import { ProgramLibrary } from '../graphics/program-library.js';
+import { getProgramLibrary } from '../graphics/get-program-library.js';
 import { RenderTarget } from '../graphics/render-target.js';
 import { ScopeId } from '../graphics/scope-id.js';
 import { Shader } from '../graphics/shader.js';
@@ -499,6 +500,11 @@ GraphicsDevice.prototype.getProgramLibrary = function () {
 
 GraphicsDevice.prototype.setProgramLibrary = function () {
     Debug.deprecated(`pc.GraphicsDevice#setProgramLibrary is deprecated.`);
+};
+
+GraphicsDevice.prototype.removeShaderFromCache = function(shader) {
+    Debug.deprecated(`pc.GraphicsDevice#removeShaderFromCache is deprecated.`);
+    getProgramLibrary(this).removeFromCache(shader);
 };
 
 // SCENE
--- a/src/deprecated/deprecated.js
+++ b/src/deprecated/deprecated.js
@@ -42,7 +42,7 @@ import { IndexBuffer } from '../graphics/index-buffer.js';
 import { createFullscreenQuad, drawFullscreenQuad, PostEffect } from '../graphics/post-effect.js';
 import { PostEffectQueue } from '../framework/components/camera/post-effect-queue.js';
 import { ProgramLibrary } from '../graphics/program-library.js';
-import { getProgramLibrary } from '../graphics/get-program-library.js';
+import { getProgramLibrary, setProgramLibrary } from '../graphics/get-program-library.js';
 import { RenderTarget } from '../graphics/render-target.js';
 import { ScopeId } from '../graphics/scope-id.js';
 import { Shader } from '../graphics/shader.js';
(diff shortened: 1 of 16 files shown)


slimbuck (commented):
Shouldn't the new file be called 'program-library' instead of 'get-program-library'?

mvaligursky:
> Shouldn't the new file be called 'program-library' instead of 'get-program-library'?

Yes, but we already have a file with that name. Maybe access-program-library?

slimbuck:
> > Shouldn't the new file be called 'program-library' instead of 'get-program-library'?
> 
> Yes, but we already have a file with that name. Maybe access-program-library?

Any reason not to put the new code in existing program-library file?

mvaligursky:
> > > Shouldn't the new file be called 'program-library' instead of 'get-program-library'?
> > 
> > 
> > Yes, but we already have a file with that name. Maybe access-program-library?
> 
> Any reason not to put the new code in existing program-library file?

Yes, I moved it out to remove dependency on program-library module, as that was creating those circular dependencies.
It's similar to existing default-material.

src/deprecated/deprecated.js
```diff
@@ -492,6 +493,20 @@ Object.defineProperties(Texture.prototype, {
...
 });
 
+GraphicsDevice.prototype.getProgramLibrary = function () {
+    Debug.deprecated(`pc.GraphicsDevice#getProgramLibrary is deprecated.`);
+    return getProgramLibrary(this);
+};
+
+GraphicsDevice.prototype.setProgramLibrary = function () {
+    Debug.deprecated(`pc.GraphicsDevice#setProgramLibrary is deprecated.`);
+};
```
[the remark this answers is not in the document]
mvaligursky: I don't see why not, thanks.

src/framework/app-base.js
```diff
@@ -20,6 +20,8 @@ import { http } from '../net/http.js';
 import {
     PRIMITIVE_TRIANGLES, PRIMITIVE_TRIFAN, PRIMITIVE_TRISTRIP
 } from '../graphics/constants.js';
+import { setProgramLibrary } from '../graphics/get-program-library.js';
```
willeastcott, line 23: LOL, I know this is intentional, but damn, it looks like a bug.

src/graphics/get-program-library.js
```diff
@@ -0,0 +1,37 @@
...
+ * @ignore
+ */
+function getProgramLibrary(device) {
+    const library = programLibraryDeviceCache.get(device);
+    Debug.assert(library);
+    return library;
+}
+
+/**
+ * Assigns the program library to device cache
```
willeastcott, line 24:
```suggestion
 * Assigns the program library to device cache.
```

src/graphics/get-program-library.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @param {GraphicsDevice} device - The graphics device used to own the program library.
+ * @param {ProgramLibrary} library - The instance of {@link ProgramLibrary}
```
willeastcott, line 27:
```suggestion
 * @param {ProgramLibrary} library - The instance of {@link ProgramLibrary}.
```

src/graphics/program-library.js
```diff
@@ -35,11 +34,14 @@ class ProgramLibrary {
...
 
-        const m = new StandardMaterial();
-        m.shaderOptBuilder.updateRef(
-            this._defaultStdMatOption, {}, m, null, [], SHADER_FORWARD, null);
-        m.shaderOptBuilder.updateMinRef(
-            this._defaultStdMatOptionMin, {}, m, null, [], SHADER_SHADOW, null);
+        standardMaterial.shaderOptBuilder.updateRef(
+            this._defaultStdMatOption, {}, standardMaterial, null, [], SHADER_FORWARD, null);
+        standardMaterial.shaderOptBuilder.updateMinRef(
+            this._defaultStdMatOptionMin, {}, standardMaterial, null, [], SHADER_SHADOW, null);
```
willeastcott, line 40: I feel like this stuff should be factored out of the program library code. But for now, it's fine.

willeastcott (approved):
Approving despite some peculiar naming of `get-program-library.js`.

src/graphics/program-library.js
mvaligursky: agreed

src/framework/app-base.js
mvaligursky: maybe I rename it to 'program-library-access.js'?

PR #4697, opened 2022-10-05
Fixes sheen intensity

GSterbrant:
### Description

Our analytical approach inspired by Three.js inconsistently divides by PI to balance the energy over the hemisphere. This caused a strong reduction in sheen light intensities. 

Furthermore, we also used the roughness of the sheen to sample different environment mips, but sampling the lowest mip is more correct as reflections will always be extremely rough, seeing as sheen is simulating quite big microfibers. 

Before:
model-viewer (8)

After:
model-viewer (6)

This PR also fixes an issue where clear coat was using the specular color to tint reflections. According to the glTF spec, the f0 for clear coat is always white.

### DIFF
--- a/src/framework/components/light/system.js
+++ b/src/framework/components/light/system.js
@@ -1,8 +1,8 @@
 import { Color } from '../../../math/color.js';
 import { Vec2 } from '../../../math/vec2.js';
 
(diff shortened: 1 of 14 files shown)


willeastcott (approved):
Excellent. Do you advocate including this fix in a 1.57.1? Or wait until the next minor bump?

slimbuck:
This PR is also updating clearcoat. Is that intended? You don't mention this in the description.

GSterbrant:
@slimbuck Thanks! Had forgotten about that. Added it to the comment.

mvaligursky (approved):

PR #4699, opened 2022-10-06
[Fix] GltfExporter correctly exports buffer views for non-interleaved vertex data

mvaligursky:
- Previously, a non-interleaved vertex buffer was stored using a single buffer view, which is not correct according to specs, as there is no single stride that can be specified for elements using it.
- This stores a per element buffer view for non=interleaved vertex formats, and removes last validation errors.

Additionally, handling of unsupported textures is now handled without crashing (by ignoring them), for both gltf and usdz exporters.

### DIFF
--- a/extras/exporters/gltf-exporter.js
+++ b/extras/exporters/gltf-exporter.js
@@ -508,6 +508,11 @@ class GltfExporter extends CoreExporter {
         this.writeMeshes(resources, json);
         this.convertTextures(resources.textures, json, options);
 
+        // delete unused properties
+        if (!json.images.length) delete json.images;
+        if (!json.samplers.length) delete json.samplers;
+        if (!json.textures.length) delete json.textures;
+
         return json;
     }
 
--- a/extras/exporters/gltf-exporter.js
+++ b/extras/exporters/gltf-exporter.js
@@ -184,24 +184,28 @@ class GltfExporter extends CoreExporter {
 
         resources.buffers.forEach((buffer) => {
 
+            const addBufferView = (target, byteLength, byteOffset, byteStride) => {
+
+                const bufferView =  {
+                    target: target,
+                    buffer: 0,
+                    byteLength: byteLength,
+                    byteOffset: byteOffset,
+                    byteStride: byteStride
+                };
+
+                return json.bufferViews.push(bufferView) - 1;
+            };
+
             const arrayBuffer = buffer.lock();
 
             if (buffer instanceof pc.VertexBuffer) {
 
                 const format = buffer.getFormat();
-
                 if (format.interleaved) {
 
-                    const bufferView = {
-                        target: ARRAY_BUFFER,
-                        buffer: 0,
-                        byteLength: arrayBuffer.byteLength,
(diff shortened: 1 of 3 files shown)


mvaligursky:
@kungfooman - thanks for the feedback, I fixed both issues.

extras/exporters/gltf-exporter.js
```diff
@@ -348,33 +375,32 @@ class GltfExporter extends CoreExporter {
...
                     mesh.primitives.push(primitive);
 
-                    // An accessor is a vertex attribute
-                    const writeAccessor = (element) => {
+                    // vertex buffer
+                    const vertexBuffer = meshInstance.mesh.vertexBuffer;
+                    const numVertices = vertexBuffer.getNumVertices();
+                    const vertexFormat = vertexBuffer.getFormat();
+                    const interleaved = vertexFormat.interleaved;
+                    vertexFormat.elements.forEach((element, elementIndex) => {
```
kungfooman, line 390:
```suggestion
                    // vertex buffer
                    const { vertexBuffer } = meshInstance.mesh;
                    const { numVertices, format } = vertexBuffer;
                    const { interleaved, elements } = format;
                    elements.forEach((element, elementIndex) => {
```

I guess this is kind of a taste thing, I don't know why you prefer `vertexBuffer.getNumVertices()` over `vertexBuffer.numVertices`, the `fake getter` looks longer than necessary to me. And using the properties allows to remove one line and every line looks shorter than before.

mvaligursky: numVertices is private, and as much as possible I avoid access to those from other classes.

kungfooman: Thank you, that explains it, I thought it would be `_numVertices` then.

mvaligursky: > I thought it would be `_numVertices` then.

We consider members which are not documented using JSDocs to be private, regardless of _, and as we develop the engine, we're free to change those as needed (rename / remove ..). If something has JSDocs and is exposed in public documentation, those are public and we take care to avoid their changes.

So here I prefer to use documented functions, limiting scope for broken code in the future.

and thanks for the getSource(), change to that as well.

kungfooman:
Great work as always, I tested as far as I could and it works.

Only problem I have is the texture exporting code fails in SeeMore, because of a type issue. Instead of `Texture#levels` being some kind of image, it is actually a `UInt8Array`... and then it just crashes currently.

I wanted to fix it myself, but the size of the UInt8Array just doesn't make sense to me:

image

Apparently the texture size of that DDS is 1024x1024, but the length of the `UInt8Array` is `524.288`. 1024*1024 is already `1.048.576`🤔

mvaligursky:
> Great work as always, I tested as far as I could and it works.

Thanks! Much appreciate your testing and feedback!

> Only problem I have is the texture exporting code fails in SeeMore,

I added code to ignore those for now, just to avoid the crash. Can be properly handled in a separate PR there is more work on textures / materials to be done.

slimbuck (approved):

PR #4700, opened 2022-10-07
Small refactor in the way VertexBuffer for post effects is created.

mvaligursky:
just saving few lines

### DIFF
--- a/src/graphics/post-effect.js
+++ b/src/graphics/post-effect.js
@@ -1,7 +1,6 @@
-import { CULLFACE_NONE, PRIMITIVE_TRISTRIP, SEMANTIC_POSITION, TYPE_FLOAT32 } from './constants.js';
+import { CULLFACE_NONE, PRIMITIVE_TRISTRIP, SEMANTIC_POSITION, TYPE_FLOAT32, BUFFER_STATIC } from './constants.js';
 import { VertexBuffer } from './vertex-buffer.js';
 import { VertexFormat } from './vertex-format.js';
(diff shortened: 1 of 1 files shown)


src/graphics/post-effect.js
```diff
@@ -92,13 +92,7 @@ function createFullscreenQuad(device) {
...
     // Fill the vertex buffer
     const iterator = new VertexIterator(vertexBuffer);
-    iterator.element[SEMANTIC_POSITION].set(-1.0, -1.0);
-    iterator.next();
-    iterator.element[SEMANTIC_POSITION].set(1.0, -1.0);
-    iterator.next();
-    iterator.element[SEMANTIC_POSITION].set(-1.0, 1.0);
-    iterator.next();
-    iterator.element[SEMANTIC_POSITION].set(1.0, 1.0);
+    iterator.writeData(SEMANTIC_POSITION, [-1, -1, 1, -1, -1, 1, 1, 1], 4);
```
willeastcott, line 95: Actually, you can do something like:

```
const data = new Float32Array([-1, -1, 1, -1, -1, 1, 1, 1]);
const vertexBuffer = new VertexBuffer(device, vertexFormat, 4, BUFFER_STATIC, data.buffer);
```

No need for an iterator.

willeastcott (approved):

src/graphics/post-effect.js
mvaligursky: even better! done.

PR #4702, opened 2022-10-07
Refactoring to remove dependency of WebglGraphicsDevice on shader chunks

mvaligursky:
- all chunks are inlined, and the prevision related ones are removed from chunks completely


src/graphics/webgl/webgl-graphics-device.js
```diff
@@ -44,6 +43,46 @@ import { Color } from '../../math/color.js';
...
 const invalidateAttachments = [];
 
+const _fullScreenQuadVS = /* glsl */`
+    attribute vec2 vertex_position;
+    varying vec2 vUv0;
+    void main(void)
+    {
+        gl_Position = vec4(vertex_position, 0.5, 1.0);
+        vUv0 = vertex_position.xy*0.5+0.5;
+    }
```
willeastcott, line 53: Do we wanna indent here? Arguably helps readability but no convinced it's needed.

willeastcott (approved):

src/graphics/webgl/webgl-graphics-device.js
mvaligursky: sure, let's save some space.

PR #4707, opened 2022-10-07
Moved /input folder to /platform/input

mvaligursky:
- in order to reorganize the engine folder structure to layered approach
- no other changes other than move + automated import paths fixed up


willeastcott (approved):

LeXXik:
Is there somewhere we could read about the upcoming "layered approach" and what it means to us?

mvaligursky:
> Is there somewhere we could read about the upcoming "layered approach" and what it means to us?

There is nothing new really. The engine was always build from 3 layers, where the bottom layer should not reference (import) modules from the higher levels:
- low level - core, math, graphics, audio, net
- mid level - scene, meshes, layers, lights, forward renderer
- application, components

But the folders were always a flat list, while now we'll have them organized per layer, to make reference restrictions more obvious, as otherwise incorrect references tend to sneak it easily - which makes for example a tree-shaking less beneficial.

PR #4727, opened 2022-10-10
Convert extras to modules

slimbuck:
This PR updates the extras library to es6, specifically:
- use import from 'playcanvas' instead of relying on the global pc object
- output a tree-shakable es6 module build of extras along with the es5 build

NOTE: the umd build of extras now requires the playcanvas engine (the `window.pc` object) to be loaded/ready before loading the extras script. Examples code gen was updated to work this way. (The launch page loads engine before extras anyway, so there should be no problem there).

@willeastcott can we disable lint error `import/no-unresolved` or squash the error some other way?


package.json
```diff
@@ -83,6 +83,7 @@
     "@playcanvas/eslint-config": "^1.3.0",
     "@playcanvas/jsdoc-template": "^1.0.27",
     "@rollup/plugin-babel": "^5.3.1",
+    "@rollup/plugin-node-resolve": "^15.0.0",
```
mvaligursky, line 86: so this allows us to use this locally?
` import { } from 'playcanvas';`

willeastcott (approved):

package.json
slimbuck: this was needed for rollup to find fflate and embed it in the es5 build of extras.

slimbuck: we've marked 'playcanvas' as an external module so rollup leaves it alone and the embedding app is responsible for providing it.

in the es5 build playcanvas is renamed to global 'pc' so it all works out magically.

mvaligursky (approved):
looks awesome

PR #4735, opened 2022-10-11
Moved PROJECTION_*** constants from /scene to /core/shape

mvaligursky:
.. as the Frustum references them there

### DIFF
--- a/src/core/shape/constants.js
+++ b/src/core/shape/constants.js
@@ -1,0 +1,13 @@
+/**
+ * A perspective camera projection where the frustum shape is essentially pyramidal.
+ *
+ * @type {number}
+ */
+export const PROJECTION_PERSPECTIVE = 0;
+
+ /**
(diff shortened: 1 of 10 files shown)


slimbuck (approved):

willeastcott:
I believe this is not the right approach here. IMHO, `Frustum.getPoints` should never have been added to this class. I think it's arguably better suited to being part of the `Camera` class. Possibly a getter called `Camera.frustumCorners`.

mvaligursky:
> I believe this is not the right approach here. IMHO, `Frustum.getPoints` should never have been added to this class. I think it's arguably better suited to being part of the `Camera` class. Possibly a non-API (`@ignore`) getter called `Camera#frustumCorners`.

You're totally right, I missed the frustum internally in the mentioned function even uses an instance of the camera. I'll have it done.

willeastcott:
Thanks @mvaligursky! U rock! 🙏

PR #4736, opened 2022-10-11
Cleanup of Entity constructor to remove circular dependency

mvaligursky:


willeastcott (approved):

src/framework/entity.js
```diff
@@ -278,18 +278,10 @@ class Entity extends GraphNode {
...
-        if (name instanceof AppBase) app = name;
-
-        if (!app) {
-            app = AppBase.getApplication(); // get the current application
-            if (!app) {
-                throw new Error("Couldn't find current application");
-            }
-        }
-
+        Debug.assert(app, 'Could not find current application');
```
willeastcott, line 284: Aren't you losing the deprecated constructor signature here?

mvaligursky: Well, all it did was allow the app to be passed in .. but we obtain it the other way anyways if that's missing, which is most of the cases. So there is no need to support it.

kungfooman: I never liked the deprecated constructor signature, but since it is/was used, the cases probably should be fixed. I only happen to remember this case: https://github.com/playcanvas/engine/blob/8b93e8211bf92c701e8cc902ec540827747fe373/src/framework/entity.js#L609-L612
I didn't test, but I think right now it would result in asking for `app` more than necessary (`Entity#name` and `Entity#app`). I think you can just change the line into `const clone = new this.constructor(untitled, this._app);` so the default argument logic catches here.
The old system is/was also wrong:
image
(`Entity#name` is not a string but... an app)

mvaligursky: yep agreed, always good to fix internal code to correct signatures. Done.

Issue #2011, opened 2020-04-23
Thick Lines


Maksims:
It would be great, to have engine solution for thick lines. `renderLine` - provides gpu based lines, which guarantees only 1 pixel wide lines, but multi-pixel lines, are up to platform. And uses immediate renderer, so mainly used for debugging purposes.

Thick lines - are common thing to use, and useful in various ways, not just for debugging, but in games and apps as well.
**So list of features would be:**

1. Rendered as normal model, with materials as usual.
2. Affected by light and cast shadows.
3. Can have option to face camera always, or be fixed orientation.
4. Have different endings, like `lineCap` on canvas2d.
5. Can be rendered in world, or as an element on screen.
6. Ensure there is a way to provide either material opacity map to make lines dotted.
7. Modes of UVs, with tiling or based on line length.
8. Shader chunks, just like on normal models should be changeable, so custom rendering features can be implemented.

LeXXik:
While working with lines, I noticed they are not culled. Could be added as a features too.

Maksims:
When rendering to **world** layer, with `depthTest: true`, still no culling currently?

LeXXik:
My bad, it does. Thanks for pointing, @Maksims :) You can ignore my comment.

willeastcott:
Very cool. Even if it wasn't core engine functionality, it would be awesome to implement that paper as some kind of extension/plugin/script.

mvaligursky:
This is mostly implemented in https://github.com/playcanvas/engine/pull/9090, closing. Additional features can be requested as needed, as suitable for the wide-line architecture.

PR #4747, opened 2022-10-13
Extend immediate mode line drawing with support for custom materials

heretique:
Hi PlayCanvas team,
please consider this PR that adds support for using custom materials when drawing lines in immediate mode.
It adds new methods to the `pc.Application`
```
drawLineWithMaterial
drawLinesWithMaterial
drawLineArraysWithMaterial
```
and leaves the rest of the line drawing functionality untouched. This could be extended to all debug drawing that draw lines `drawWireSphere` and `drawWireSphere` as well.

The goal of this PR is to be able to customize even further the drawing of debug lines, adding the possibility of drawing dashed lines, fading with distance, etc.
In order to achieve that we expose the following vertex attributes for the material:
 - `aPosition`: pc.Vec3 (vertex position)
 - `aTAndLength`: pc.Vec2 (x - 1D texture coordinate [0..1], y - line length. Not quite convinced about the name of this attribute)
 - `aColor`: pc.Vec4 (vertex color)

The `Lines` example is also updated to show the new capabilities and shows example of drawing dashed lines

image

### DIFF
--- a/examples/src/examples/graphics/lines.tsx
+++ b/examples/src/examples/graphics/lines.tsx
@@ -4,9 +4,44 @@ import * as pc from '../../../../';
 class LinesExample {
     static CATEGORY = 'Graphics';
     static NAME = 'Lines';
+    static FILES = {
+        'shader.vert': /* glsl */`
+attribute vec3 aPosition;
+attribute vec2 aTAndLength;
+attribute vec4 aColor;
+
+uniform mat4 matrix_model;
+uniform mat4 matrix_viewProjection;
+
+varying vec2 vTAndLength;
+varying vec4 vColor;
+
+void main(void)
+{
+    vTAndLength = aTAndLength;
+    vColor = aColor;
+    gl_Position = matrix_viewProjection * matrix_model * vec4(aPosition, 1.0);
+}`,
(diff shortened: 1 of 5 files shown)


mvaligursky:
Hi @heretique and thanks for the PR. I took a while and I kept coming back to it to think about it. It's a great PR that improves the lines quite a bit, but I'm hesitant to approve it based on:

- ultimate goal is something along this: https://github.com/playcanvas/engine/issues/2011 - where the line rendering code is more powerful, vs existing code being used for a quick debug like line rendering. Ideally it would be implemented as a module in the [extras](https://github.com/playcanvas/engine/tree/main/extras) folder, to allow this to be excluded from builds for users not needing it.
- I would really like to keep the current line rendering fast, but the length calculation of each segment makes it a lot slower than currently

heretique:
Thanks @mvaligursky , I guess it makes sense for you guys, I know it's slower with the length calculation, but I figured since these are only used for debugging purposes, it is not that problematic. We're drawing thousands of them at interactive frame rate on a desktop computer.
Anyway, thanks for spending the time reviewing it, we'll keep our own patch for it going forward.

PR #4750, opened 2022-10-14
Add missing material parameters

GSterbrant:
### Description

Turns out we have tons of missing parameters in the standard material parameters list. 

This PR also adds documentation for the iridescence, thickness, attenuation as well as missing vertex color attributes for refraction. 

Also renamed sheenGlossiness to sheenGloss to have it be more consistent with the normal gloss and clear coat gloss.


src/scene/materials/standard-material.js
```diff
@@ -1094,7 +1147,7 @@ function _defineMaterialProps() {
     _defineColor('attenuation', new Color(1, 1, 1));
     _defineFloat('emissiveIntensity', 1);
     _defineFloat('specularityFactor', 1);
-    _defineFloat('sheenGlossiness', 0);
+    _defineFloat('sheenGloss', 0);
```
mvaligursky, line 1150: so this renames the property .. what would happen to materials people have set up with the old name?

Search the codebase for RENAMED_PROPERTIES .. I wonder if that could handle it? (I have not used it before, but worth a test)

src/scene/materials/standard-material-parameters.js
```diff
@@ -172,12 +182,58 @@ const standardMaterialParameterTypes = {
...
     refraction: 'number',
+    refractionMap: 'texture',
+    refractionMapChannel: 'string',
+    refractionMapUv: 'number',
+    refractionMapTiling: 'vec2',
+    refractionMapOffset: 'vec2',
+    refractionMapMapRotation: 'number',
+    refractionTint: 'boolean',
+    refractionVertexColor: 'boolean',
+    refractionVertexColorChannel: 'string',
```
willeastcott, line 193: Seems like there's an opportunity to define map properties in a function perhaps? Most maps share the same set.

src/scene/materials/standard-material.js
GSterbrant: This parameter wasn't exposed in the editor before, so I doubt anyone has used it. But I can definitely add it to the renamed properties list just to make sure.

src/scene/materials/standard-material-parameters.js
GSterbrant: Yes, that would be the third such function :D 

But I totally agree, this is annoying, repetitive and easy to miss, which is made abundantly clear by this PR.

src/scene/materials/standard-material.js
mvaligursky: ah true, ignore in that case, there's no way the parameter would be stored in the database then.

mvaligursky: people could have used it in the engine scripts is the only place .. do we have some way to print warning on setting an incorrect parameter or something?

src/scene/materials/standard-material.js
```diff
@@ -183,6 +183,31 @@ let _params = new Set();
  * @property {number} clearCoatBumpiness The bumpiness of the clear coat layer. This value scales
  * the assigned main clear coat normal map. It should be normally between 0 (no bump mapping) and 1
  * (full bump mapping), but can be set to e.g. 2 to give even more pronounced bump effect.
+ * @property {boolean} useIridescence Enable thin-film iridescence.
+ * @property {Texture|null} iridescenceMap The per-pixel iridescence intensity. Only used when useIridescence is enabled.
```
willeastcott, line 187: I need to add this to the coding standards but all JSDoc blocks should stick to 100 columns.

src/scene/materials/standard-material.js
GSterbrant: Hmm, if it's not called that in the standard-material-parameters, it will be shown in the log as an invalid standard material parameter. But that's about it afaik.

src/scene/materials/standard-material-parameters.js
GSterbrant: I wrote a function for this and condensed these declarations to a simple function call.

src/scene/materials/standard-material.js
mvaligursky: great, nothing to do here, resolving.

src/scene/materials/standard-material.js
```diff
@@ -183,12 +183,42 @@ let _params = new Set();
...
+ * @property {number} iridescenceThicknessMapRotation Controls the 2D rotation (in degrees)
+ * of the iridescence map.
+ * @property {string} iridescenceThicknessMapChannel Color channels of the iridescence thickness
+ * map to use. Can be "r", "g", "b" or "a".
+ * @property {number} iridescenceThicknessMin The minimum thickness for the iridescence layer.
+ * Only used when an iridescence thickness map is used.
+ * @property {number} iridescenceThicknessMax The maximum thickness for the iridescence layer.
+ * Used as the 'base' thickness when no iridescence thickness map is defined.
+ * @property {number} iridescenceRefractionIndex The index of refraction of the iridescent
+ * thin-film.
```
willeastcott, line 214: Worth specifying a range, or giving guidance as to the effect of different values?

src/scene/materials/standard-material.js
```diff
@@ -183,12 +183,42 @@ let _params = new Set();
...
  * @property {boolean} useMetalness Use metalness properties instead of specular. When enabled,
  * diffuse colors also affect specular instead of the dedicated specular map. This can be used as
  * alternative to specular color to save space. With metaless == 0, the pixel is assumed to be
  * dielectric, and diffuse color is used as normal. With metaless == 1, the pixel is fully
  * metallic, and diffuse color is used as specular color instead.
- * @property {boolean} useMetalnessSpecularColor When metalness is enabled, use the specular map to apply color tint to specular reflections
+ * @property {boolean} useMetalnessSpecularColor When metalness is enabled, use the
+ * specular map to apply color tint to specular reflections
```
willeastcott, line 221:
```suggestion
 * specular map to apply color tint to specular reflections.
```

src/scene/materials/standard-material.js
```diff
@@ -227,19 +257,41 @@ let _params = new Set();
...
+ * Can be "r", "g", "b" or "a".
  * @property {number} refractionIndex Defines the index of refraction, i.e. The amount of
  * distortion. The value is calculated as (outerIor / surfaceIor), where inputs are measured
  * indices of refraction, the one around the object and the one of its own surface. In most
  * situations outer medium is air, so outerIor will be approximately 1. Then you only need to do
  * (1.0 / surfaceIor).
  * @property {boolean} useDynamicRefraction Enables higher quality refractions using the grab pass
  * instead of pre-computed cube maps for refractions.
+ * @property {number} thickness The thickness of the medium, only used when useDynamicRefraction
+ * is enabled.
```
willeastcott, line 278: So for a 1x1x1 (unit) cube, you'd set this to 1?

src/scene/materials/standard-material.js
```diff
@@ -227,19 +257,41 @@ let _params = new Set();
...
+ * @property {number} thicknessMapRotation Controls the 2D rotation (in degrees) of the thickness
+ * map.
+ * @property {string} thicknessMapChannel Color channels of the thickness map to use. Can be "r",
+ * "g", "b" or "a".
+ * @property {boolean} thicknessVertexColor Use mesh vertex colors for thickness. If
+ * thickness map is set, it will be be multiplied by vertex colors.
+ * @property {Color} attenuation The attenuation color for refractive materials, only used when
+ * useDynamicRefraction is enabled.
+ * @property {number} attenuationDistance The distance defining the absorption rate of light
+ * within the medium, only used when useDynamicRefraction is enabled.
```
willeastcott, line 294:
```suggestion
 * within the medium. Only used when useDynamicRefraction is enabled.
```

willeastcott (approved):
Big improvement - approving with minor feedback.

src/scene/materials/standard-material.js
GSterbrant: I guess we should specify which unit this is in, but yeah, it's whatever unit the scale of the model is in. 1 would mean 1 world unit, and 2 would mean 2, and so forth. I'll add unit denotations in the docs as well.

src/scene/materials/standard-material.js
GSterbrant: It's a bit tricky to make it easy to understand. The index of refraction will change the phase shift of the iridescent effect, as in how frequent the different color bands will appear on the surface. Reasonable values range from between 0 to ~7 where the upper limit represents the highest measured index of refraction of a physical surface to date.

In short, it's very technical 🙂 .

src/scene/materials/standard-material.js
GSterbrant: But yeah, for a unit sphere of size 1, it would make sense to make this 1. However, it also scales with the object using the object transform, so if you set it to 1 and scale the object up, you can expect the thickness to conform to the final transformation. So this property is reusable for objects with different scale, no need to create a material for each.

mvaligursky (approved):

src/scene/materials/standard-material.js
GSterbrant: Added documentation links to the glTF extension spec which explains how the iridescence refraction index affects color phase shift of the material.

PR #4761, opened 2022-10-19
Improvements to clustered lighting shader compilation speed on Windows (DirectX)

mvaligursky:
Shader compilation speed improvements of around 50% on Windows Platform (when using DirectX backend), when clustered lights are used, and especially with area lights and shadows. DirectX shader compiler does not allow gradient instructions inside of the loops with varying number of iterations, and attemps to unroll the loops, which in the mentioned cases takes a long time. The solution is to avoid using gradient instructions inside the loop.

### Details:
- macros to allow texture2DLodEXT texture sampling on all platforms (simulated on Webgl1 without the extension). This is used to read clustered light data without gradients, and also read area lights LUTs. (and also in one engine example)
- new textureShadow macro to sample shadow map without gradients (internally uses textureGrad, as textureLod seems to have WebGL translation issue when used with shadow samplers)

### Timings

All with area lights enabled, and PCF3 shadows, up to date Chrome browser.

Windows, original state
original area + pcf3

Windows, now
now area + pcf3

### Other timings for comparison

Windows, clustered lights disabled (using old lighting code)
windows non-clustered

Mac OS on the first run
mac first run

Mac OS on the following runs (some caching takes place?)
mac next run

Mac OS with clustered lights enabled, first run
mac non-clustered

### DIFF
--- a/src/platform/graphics/webgl/webgl-shader.js
+++ b/src/platform/graphics/webgl/webgl-shader.js
@@ -229,8 +229,8 @@ class WebglShader {
             // #if _DEBUG
 
             // log translated shaders
-            definition.translatedFrag = gl.getExtension('WEBGL_debug_shaders').getTranslatedShaderSource(this.glFragmentShader);
-            definition.translatedVert = gl.getExtension('WEBGL_debug_shaders').getTranslatedShaderSource(this.glVertexShader);
+            definition.translatedFrag = gl.getExtension('WEBGL_debug_shaders')?.getTranslatedShaderSource(this.glFragmentShader);
+            definition.translatedVert = gl.getExtension('WEBGL_debug_shaders')?.getTranslatedShaderSource(this.glVertexShader);
 
             console.error(message, definition);
             // #else
--- a/src/platform/graphics/webgl/webgl-shader.js
+++ b/src/platform/graphics/webgl/webgl-shader.js
@@ -227,6 +227,11 @@ class WebglShader {
             const message = "Failed to link shader program. Error: " + gl.getProgramInfoLog(glProgram);
 
             // #if _DEBUG
+
+            // log translated shaders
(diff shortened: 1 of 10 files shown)


src/platform/graphics/webgl/webgl-shader.js
```diff
@@ -227,6 +227,11 @@ class WebglShader {
             const message = "Failed to link shader program. Error: " + gl.getProgramInfoLog(glProgram);
 
             // #if _DEBUG
+
+            // log translated shaders
+            definition.translatedFrag = gl.getExtension('WEBGL_debug_shaders').getTranslatedShaderSource(this.glFragmentShader);
+            definition.translatedVert = gl.getExtension('WEBGL_debug_shaders').getTranslatedShaderSource(this.glVertexShader);
```
willeastcott, line 233: This _could_ return `null` right?

mvaligursky: it could, and that is fine, the error console log will have undefined/null for the field.

mvaligursky: ah sorry, I missed what you're saying, I'll fix this :)

willeastcott: No, `getTranslatedShaderSource` will through an exception, surely (called by `null`).

willeastcott (commented):
Approving. So this is really cool, but I still suspect we're going to have problems with compilation time. I was seeing a 2 minute load time in the Editor for a very basic scene. Should I expect maybe a 1 minute load time after this PR?

mvaligursky:
Yes, this PR seems to about halve the compilation time. The total time depends on the number of shaders - the test scene in this example had one shader using the lighting. By comparison, MacOS is still faster by 70% or so, but even there the shaders take time to compile.

The ultimate solution here is to allow for parallel shader compilation. Code-base was partially refactored to allow this, but there's a lot more work needed before this is done.

GSterbrant (approved):

PR #4764, opened 2022-10-19
Remove two circular dependencies from scripts

mvaligursky:


willeastcott:
Am I missing something here? Why can't those occurrences of `scriptComponentMethods` just be replaced with string literals?

mvaligursky:
> Am I missing something here? Why can't those occurrences of `scriptComponentMethods` just be replaced with string literals?

I guess this is set up like an enum .. we could turn it into constants? We could even just inline the string literals, but we'd have multiple copies of it .. not that anybody would expect those to be renamed.

so maybe I turn this scriptComponentMethods file into constants.js and have those as constants?

src/framework/script/constants.js
```diff
@@ -0,0 +1,5 @@
+export const SCRIPTMETHOD_INITIALIZE = 'initialize';
+export const SCRIPTMETHOD_POST_INITIALIZE = 'postInitialize';
```
willeastcott, line 2: Playing devil's advocate here. What about: SCRIPT_POSTINITIALIZE. More succinct.

willeastcott:
Yeah, sounds like a good plan to me, @mvaligursky.

willeastcott (approved):
Nice job! ❤️

PR #4765, opened 2022-10-19
Return a sensible value for animation component duration if no animation is playing

yaustar:
Stops exception being thrown when using duration if there is no animation playing

Returns `Number.MAX_VALUE` instead of 0 as some developers may be dividing by this number for a normalised value and this avoids divide by 0.

A bit torn on whether this is the 'right' number to return though?


mvaligursky:
I'd like returning 0 more I think, and using Debug.warn to let the user know.

yaustar:
Yeah, I'm very torn on getting infinity errors 'silently' and not seeing it until it hits something that fires an exception (usually far away in the call stack from the actual cause) and actually giving a number that makes sense common use cases.

If you still think it should be 0, I will change it but have the above concern.

mvaligursky:
I understand, but a warning should solve the problem really.

yaustar:
Updated PR to use 0 with warning

src/framework/components/animation/component.js
```diff
@@ -216,12 +217,18 @@ class AnimationComponent extends Component {
...
      */
     get duration() {
-        return this.animations[this.currAnim].duration;
+        // #if _DEBUG
+        if (!this.currAnim) {
+            Debug.warn(`No animation is playing to get a duration. Returning 0.`);
+        }
+        // #endif
+
+        return this.currAnim ? this.animations[this.currAnim].duration : 0;
```
mvaligursky, line 231: Not much shorter I guess, just trying to avoid ifdefs in general. You could also use Debug.call.
```suggestion
        if (this.currAnim) {
            return this.animations[this.currAnim].duration;
        }

        Debug.warn(`No animation is playing to get a duration. Returning 0.`);
        return 0;
```

mvaligursky (approved):

src/framework/components/animation/component.js
yaustar: Oh, I forgot you don't need ifdefs now for messages

PR #4766, opened 2022-10-19
Extraced ScriptTypes from ScriptHandler to avoid circular dependency

mvaligursky:
That looks like the last circular dependency. How do we disable them now?


willeastcott (approved):

willeastcott:
All circular dependencies GONE! 🍾 You deserve one of these: 🍪

PR #4778, opened 2022-10-21
Renamed the export function to have Async postfix

mvaligursky:
- for clarity. If agreed, I’d add this to the coding guidelines we have too.


willeastcott:
I have to admit, I don't recall seeing APIs call out in the function name that a function is async. So personally, I would avoid this.

mvaligursky:
Sounds good, closing.

PR #4781, opened 2022-10-24
Identify glb header correctly

slimbuck:
Fixes #4768

### DIFF
--- a/src/framework/parsers/glb-parser.js
+++ b/src/framework/parsers/glb-parser.js
@@ -2470,8 +2470,13 @@ const parseGlb = function (glbData, callback) {
 
 // parse the chunk of data, which can be glb or gltf
 const parseChunk = function (filename, data, callback) {
-    if ((filename && filename.toLowerCase().endsWith('.glb')) ||
-        (data[0] === 'g' && data[1] === 'l' && data[2] === 'T' && data[3] === 'F')) {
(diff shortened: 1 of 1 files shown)


src/framework/parsers/glb-parser.js
```diff
@@ -2470,8 +2470,13 @@ const parseGlb = function (glbData, callback) {
...
 const parseChunk = function (filename, data, callback) {
-    if ((filename && filename.toLowerCase().endsWith('.glb')) ||
-        (data[0] === 'g' && data[1] === 'l' && data[2] === 'T' && data[3] === 'F')) {
+    const hasGlbHeader = () => {
+        // glb format starts with 'glTF'
+        const u8 = new Uint8Array(data);
+        return u8[0] === 103 && u8[1] === 108 && u8[2] === 84 && u8[3] === 70;
+    }
+
+    if ((filename && filename.toLowerCase().endsWith('.glb')) || hasGlbHeader()) {
```
willeastcott, line 2479: Wait, so if it's got a GLB extension but the header is wrong, we still proceed anyway?

slimbuck: This code is deciding between loading a glb or gltf. If the filename is recognised or header we load the glb. Otherwise we load it as gltf (which is headerless).

willeastcott (approved):

PR #4788, opened 2022-10-25
Audio resume fix

### LINKED ISSUE #4760: Audio is blocked from playing on iOS if user starts with a swipe gesture since 1.56
Forum: https://forum.playcanvas.com/t/the-only-way-how-to-start-audio-on-ios-is-by-fast-tap-on-the-screen-no-swipe-or-touch-and-hold-anymore/27850

Using this project as the example: https://playcanvas.com/project/405821

https://launch.playcanvas.com/440098?debug=true&use_local_engine=https://code.playcanvas.com/playcanvas-1.56.0.js will start playing sound on a tap. However, if you swipe it will not play sound and subsequent taps will NOT start the sound.

https://launch.playcanvas.com/440098?debug=true&use_local_engine=https://code.playcanvas.com/playcanvas-1.55.4.js was the last working version.

This issue does not occur on Android

slimbuck:
Fixes #4760

iOS doesn't correctly resume on touchend event, but does on touchstart event.

This PR also simplifies manager suspend/resume logic a bit.

API Changes:
- removed SoundManager options object

### DIFF
--- a/src/platform/sound/manager.js
+++ b/src/platform/sound/manager.js
@@ -103,7 +103,7 @@ class SoundManager extends EventHandler {
     }
 
     suspend() {
-        if (!this._enabled) {
+        if (this._enabled) {
             this._enabled = false;
             if (this._context && this._context.state === CONTEXT_STATE_RUNNING) {
                 this._suspend();
--- a/src/platform/sound/manager.js
+++ b/src/platform/sound/manager.js
@@ -28,11 +28,8 @@ class SoundManager extends EventHandler {
     /**
      * Create a new SoundManager instance.
      *
-     * @param {object} [options] - Options options object.
-     * @param {boolean} [options.forceWebAudioApi] - Always use the Web Audio API, even if check
-     * indicates that it is not available.
      */
-    constructor(options) {
+    constructor() {
         super();
 
         /**
--- a/src/platform/sound/manager.js
+++ b/src/platform/sound/manager.js
@@ -126,7 +126,7 @@ class SoundManager extends EventHandler {
 
         if (this._context) {
             this._removeUnlockListeners();
-            this?._context?.close();
+            this._context?.close();
             this._context = null;
         }
     }
--- a/src/platform/sound/manager.js
+++ b/src/platform/sound/manager.js
@@ -184,10 +184,7 @@ class SoundManager extends EventHandler {
     // resume the sound context
     _resume() {
         // explictly resume() context, and only fire 'resume' event after context has resumed
-        console.log(`before: ${this._context.state}`);
         this._context.resume().then(() => {
-            console.log(`after: ${this._context.state}`);
-
             // Some platforms (mostly iOS) require an additional sound to be played.
             // This also performs a sanity check and verifies sounds can be played.
             const buffer = this._context.createBuffer(1, 1, this._context.sampleRate);
--- a/src/platform/sound/manager.js
+++ b/src/platform/sound/manager.js
@@ -52,7 +52,8 @@ class SoundManager extends EventHandler {
 
         this._unlockHandlerFunc = this._unlockHandler.bind(this);
 
-        this._enabled = true;
+        // user suspended audio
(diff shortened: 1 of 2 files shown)


src/platform/sound/manager.js
```diff
@@ -46,52 +43,18 @@ class SoundManager extends EventHandler {
...
-
-        /**
-         * The function callback attached to the Window events USER_INPUT_EVENTS
-         *
-         * @type {EventListenerOrEventListenerObject}
-         * @private
-         */
-        this._resumeContextCallback = null;
+        // @ts-ignore
+        this.AudioContext = window.AudioContext || window.webkitAudioContext;
```
willeastcott, line 47: You can't assume existence of `window` (e.g. when you're running in Node).

src/platform/sound/manager.js
```diff
@@ -46,52 +43,18 @@ class SoundManager extends EventHandler {
...
-        this._forceWebAudioApi = options.forceWebAudioApi;
-
-        /**
-         * The function callback attached to the Window events USER_INPUT_EVENTS
-         *
-         * @type {EventListenerOrEventListenerObject}
-         * @private
-         */
-        this._resumeContextCallback = null;
+        // @ts-ignore
```
willeastcott, line 46: I'd rather we disallowed `@ts-ignore` in the codebase. It appears nowhere else. There must be an alternative...

slimbuck: Oh right, I'm so used to working in ts these days... removed.

src/platform/sound/manager.js
```diff
@@ -125,87 +92,41 @@ class SoundManager extends EventHandler {
...
     destroy() {
-        this._removeUserInputListeners();
-
         this.fire('destroy');
 
-        if (this._context && this._context.close) {
-            this._context.close();
+        if (this._context) {
+            this._removeUnlockListeners();
+            this?._context?.close();
```
willeastcott, line 129: But isn't `this._context` guaranteed to be valid by this point?

willeastcott (approved):

willeastcott:
What about these lines:

https://github.com/playcanvas/engine/pull/4788/files#diff-bf489dbc39f80880f4921c574a985e18d6abd0c6716cd3d2907dbb96ed37e255R31-R32

That's public API:

https://developer.playcanvas.com/api/pc.SoundManager.html#SoundManager

So presumably this PR actually removes that option? I'm fine with that (to be honest, I'm not sure why anyone would want to use that option!). But if this is so, the docs should be updated.

src/platform/sound/manager.js
slimbuck: Not if AudioContext doesn't exist, like in engine tests :)

willeastcott: I'm confused. Then the `if` condition would have failed, no?

slimbuck: Oic you're referring to the first ? on this line. Yeah I can remove it. The old code tested for presence of context.close before invoking it, so the second ? should probably stay.

slimbuck: Done, thanks!

slimbuck:
> So presumably this PR actually removes that option? I'm fine with that (to be honest, I'm not sure why anyone would want to use that option!). But if this is so, the docs should be updated.

Oh yes, that code just didn't make any sense (and I assumed this was some legacy workaround). Perhaps I should investigate the origin of that option first and then decide whether to remove it entirely.

willeastcott:
Sure, but also the whole options object from the code (plus passing it in application.js) since that was the only option. 😄

slimbuck:
> Sure, but also the whole options object from the code (plus passing it in application.js) since that was the only option. 😄

Done!

willeastcott:
OK, LGTM now. Nice job! 🙌

willeastcott:
But maybe call out the API change in the PR description!

slimbuck:
Thanks for your help @willeastcott !

PR #4792, opened 2022-10-27
Options split

GSterbrant:
### Description

First pass on splitting material options between frontend and backend. In this PR, the same options are passed as both, but this will allow for stripping down the input parameters according to what's needed for frontend and backend.  

### Public API changes
The material onUpdateShader callbacks may not work if the standard material variable being changed has been moved to litOptions, in which case the callback must be adjusted to work with options split.

### DIFF
--- a/src/scene/materials/standard-material-options-builder.js
+++ b/src/scene/materials/standard-material-options-builder.js
@@ -347,11 +347,18 @@ class StandardMaterialOptionsBuilder {
         }
         options[vname] = false;
         options[vcname] = '';
-        options.litOptions[vname] = options[vname];
+        const litOptions = {
+            enabled: false,
+            transform: 0,
+            uv: 0,
+            vertexColors: false
+        };
 
         const isOpacity = p === 'opacity';
-        if (isOpacity && stdMat.blendType === BLEND_NONE && stdMat.alphaTest === 0.0 && !stdMat.alphaToCoverage)
+        if (isOpacity && stdMat.blendType === BLEND_NONE && stdMat.alphaTest === 0.0 && !stdMat.alphaToCoverage) {
+            options.litOptions[p] = litOptions;
             return;
+        }
 
         if (!minimalOptions || isOpacity) {
             if (p !== 'height' && stdMat[vname]) {
@@ -386,11 +393,12 @@ class StandardMaterialOptionsBuilder {
             }
         }
 
-        const litOptionsName = 'use' + mname[0].toUpperCase() + mname.substring(1);
-        options.litOptions[litOptionsName] = options[mname];
-        options.litOptions[tname] = options[tname];
-        options.litOptions[uname] = options[uname];
-        options.litOptions.useVertexColors = options.vertexColors;
+        litOptions.enabled = options[mname];
+        litOptions.tranform = options[tname];
+        litOptions.uv = options[uname];
+        litOptions.vertexColors = options[vname];
+        options.litOptions[p] = litOptions;
+        options.litOptions.vertexColors = options.vertexColors;
     }
 
     _collectLights(lType, lights, lightsFiltered, mask, staticLightList) {
--- a/src/scene/materials/standard-material-options-builder.js
+++ b/src/scene/materials/standard-material-options-builder.js
@@ -343,6 +343,8 @@ class StandardMaterialOptionsBuilder {
         }
         options[vname] = false;
         options[vcname] = '';
+        options.litOptions[vname] = options[vname];
+        options.litOptions[vcname] = options[vcname];
 
         const isOpacity = p === 'opacity';
         if (isOpacity && stdMat.blendType === BLEND_NONE && stdMat.alphaTest === 0.0 && !stdMat.alphaToCoverage)
@@ -386,8 +388,6 @@ class StandardMaterialOptionsBuilder {
         options.litOptions[cname] = options[cname];
         options.litOptions[tname] = options[tname];
         options.litOptions[uname] = options[uname];
-        options.litOptions[vname] = options[vname];
-        options.litOptions[vcname] = options[vcname];
         options.litOptions.vertexColors = options.vertexColors;
     }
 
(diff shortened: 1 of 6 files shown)


slimbuck:
Wouldn't it be neater for options to be the frontend options with a member to the backend options object? This way you can still pass around just one object?

mvaligursky:
I wonder if backend and frontend naming is ideal. Perhaps the backendOptions should be litOptions, as those are used by lit-shader.
As @slimbuck mentioned .. maybe have options as before, and options.lit object for lit options?
And perhaps in the lit-shader.js, those should still just be called options there, no need to rename. We'd pass in the options.lit to it.

GSterbrant:
But it's not just the lit shader, the particles and the skybox also has backend options. I can make it into a single object if that makes it better. However considering the public API has to change as well as all special shader ```getProgram``` calls, I don't see what the difference is other than having a single parameter to the function and having them dependent on each other despite us wanting to have a clear separation between them.

src/scene/materials/standard-material-options-builder.js
```diff
@@ -370,6 +385,12 @@ class StandardMaterialOptionsBuilder {
                 }
             }
         }
+
+        const litOptionsName = 'use' + mname[0].toUpperCase() + mname.substring(1);
```
slimbuck, line 389: if these map options are changing anyway, would it make sense to group each map into its own object? (A bit like the standard material should have done).

So instead of having:
options.useDiffuseMap
options.diffuseMap
options.diffuseMapTransform
options.diffuseMapUv

make it:
options.diffuseMap.enabled
options.diffuseMap.texture
options.diffuseMap.transform
options.diffuseMap.uv
etc

Though now I'm wondering why these map options are part of litOptions in the first place. Aren't these all frontend?

GSterbrant: Done. What bothers me a little still (but we can deal with that in another pass over this) is that we only need a certain subset of texture usage info in the lit shader, but this code effectively adds that for all of them.

mvaligursky:
The public API change is something we could handle better I think. Currently the user gets no indication about it, and things just won't render. Some of those flag, for example useInstancing is used often, and not warning user / not handling this in a backwards compatible way will make it hard to users to find and fix. Suggestion:
- The options class (the public API) is currently just an unnamed object here: https://github.com/playcanvas/engine/blob/9c62f9bf55b48062dc1c0591ad6b0af1c2884685/src/scene/materials/standard-material.js#L33-L41
- we should create a class for it, called let's say ShaderOptions, and move all options including documentation (https://github.com/playcanvas/engine/blob/9c62f9bf55b48062dc1c0591ad6b0af1c2884685/src/scene/materials/standard-material.js#L495-L564) to it.
- We should then properly deprecate old options that are changing in deprecated.js for this new class. Maybe when the user returns the options from the callback, we make sure he did not set options we do not use. Or have setters for those old options print a deprecated message, and assign it to where it needs to be now.

src/scene/shader-lib/program-library.js
```diff
@@ -75,9 +75,9 @@ class ProgramLibrary {
         let def = this.definitionsCache.get(key);
         if (!def) {
             let lights;
-            if (options.lights) {
-                lights = options.lights;
-                options.lights = lights.map(function (l) {
+            if (options.litOptions?.lights) {
```
mvaligursky, line 78: In majority of places you use `options.litOptions.property` but here you have `options.litOptions?.property` - are you expecting litOptions to be null/undefined here but not elsewhere? and One more time few lines below.

src/scene/shader-lib/program-library.js
```diff
@@ -87,8 +87,8 @@ class ProgramLibrary {
 
             this.storeNewProgram(name, options);
 
-            if (options.lights)
-                options.lights = lights;
+            if (options.litOptions?.lights)
```
mvaligursky, line 90: another instance here

src/scene/shader-lib/program-library.js
```diff
@@ -157,6 +157,9 @@ class ProgramLibrary {
                 if ((options.hasOwnProperty(p) && defaultMat[p] !== options[p]) || p === "pass")
                     opt[p] = options[p];
             }
+            for (const p in options.litOptions) {
```
mvaligursky, line 160: do we copy all lit options into options here?

src/scene/shader-lib/program-library.js
GSterbrant: Yeah, because program-library isn't just called by the standard material, so the lit options member is optional.

src/scene/shader-lib/programs/lit-shader.js
```diff
@@ -877,11 +877,11 @@ class LitShader {
         code += chunks.combinePS;
 
         // lightmap support
-        if (options.lightMap || options.lightVertexColor) {
-            code += (options.useSpecular && options.dirLightMap) ? chunks.lightmapDirAddPS : chunks.lightmapAddPS;
+        if (options.light?.enabled || options.light?.vertexColors) {
```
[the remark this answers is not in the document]
GSterbrant: Yeah, it's just that it follows the same pattern as all other maps, as in you have a name + Map. So to convert this to the struct Donovan suggested, the only automatic way to do that would be to use the name of the map. 

The lit options also have light**S** which is the array of all lights...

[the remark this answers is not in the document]
mvaligursky: maybe the names should contain the word Map as before? so lightMap, but it would be an object with other properties inside.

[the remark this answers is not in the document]
mvaligursky: I just looked above and that's what was suggested there too: https://github.com/playcanvas/engine/pull/4792#discussion_r1015721413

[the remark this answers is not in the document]
GSterbrant: Fixed this by selectively picking the texture options to pass to the lit options.

src/scene/materials/standard-material-options-builder.js
GSterbrant: Actually reverted this seeing as the lit options should not have all texture options but rather only needs a few flags from the frontend.

slimbuck (approved):

mvaligursky (approved):
Approving. Could you please create an issue on the remaining work that needs to be done before the next engine release (regarding deprecating and backwards compatibility)

GSterbrant:
Added a new issue for the follow-up work to this PR #4871

Issue #4793, opened 2022-10-28
Ammo corrupts Entity transform for dynamic objects


ErikSom:
Project: https://playcanvas.com/project/1001582/overview/physics-bug

### Description

(See video below)
There is 2 physics objects in the scene. One kinematic and one dynamic.
For the kinematic object PlayCanvas drives the transform of the RigidBody and you can see in all orientations my clicks position my red debug ball correctly.
However for the dynamic object, initially the clicks go fine but the moment the box falls and Ammo is driving the PlayCanvas transform it gets corrupted.

### Steps to Reproduce

https://user-images.githubusercontent.com/1382022/198521550-cf44fa9f-26ec-4b83-b670-3c402022567b.mov

ErikSom:
Its all my mistake
instead of:
`const rot = body.entity.getRotation().clone().invert();`
I used:
`const rot = body.entity.rotation.clone().invert();`

PR #4794, opened 2022-10-28
Resync dynamic rigidbodies transforms in update

ErikSom:
Fixes https://github.com/playcanvas/engine/issues/4793

### DIFF
--- a/src/framework/components/rigid-body/component.js
+++ b/src/framework/components/rigid-body/component.js
@@ -939,6 +939,9 @@ class RigidBodyComponent extends Component {
                 const q = ammoTransform.getRotation();
                 this.entity.setPosition(p.x(), p.y(), p.z());
                 this.entity.setRotation(q.x(), q.y(), q.z(), q.w());
+
+                this._getEntityTransform(ammoTransform);
+                motionState.setWorldTransform(ammoTransform);
             }
         }
     }
--- a/src/framework/components/rigid-body/component.js
+++ b/src/framework/components/rigid-body/component.js
@@ -939,9 +939,6 @@ class RigidBodyComponent extends Component {
                 const q = ammoTransform.getRotation();
                 this.entity.setPosition(p.x(), p.y(), p.z());
                 this.entity.setRotation(q.x(), q.y(), q.z(), q.w());
-
-                this._getEntityTransform(ammoTransform);
-                motionState.setWorldTransform(ammoTransform);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


ErikSom:
https://user-images.githubusercontent.com/1382022/198541883-5cf081eb-9ebc-48e5-83d8-e07a612938b8.mov

(sorry for the music)

src/framework/components/rigid-body/component.js
```diff
@@ -939,6 +939,9 @@ class RigidBodyComponent extends Component {
                 const q = ammoTransform.getRotation();
                 this.entity.setPosition(p.x(), p.y(), p.z());
                 this.entity.setRotation(q.x(), q.y(), q.z(), q.w());
+
+                this._getEntityTransform(ammoTransform);
+                motionState.setWorldTransform(ammoTransform);
```
yaustar, line 944: This looks odd to me as the ammoTransform should be the same value as what we got on line 936.

This change implies that setting the position and rotation of the entity is wrong or that _getEntityTransform is giving a different value to what we started with ammoTransform.

ErikSom: It did to me as well:
image
That also fixes it, which I think points the issue to inside of `setRotation()`.
Because `getRotation()` overwrites the rotation with the world transform.
https://github.com/playcanvas/engine/blob/bf31671060633d8e26dff833ded278fcf14d5315/src/scene/graph-node.js#L853

LeXXik: I haven't looked in details, but this looks odd as well. The motion state of a dynamic body is and should be driven by Ammo under normal update cycle. The issue must be elsewhere.

ErikSom: Not sure if it's intended, but shouldn't this refer to `this`?
https://github.com/playcanvas/engine/blob/bf31671060633d8e26dff833ded278fcf14d5315/src/scene/graph-node.js#L1120

LeXXik: It is intended, as it is using a private variable:
https://github.com/playcanvas/engine/blob/bf31671060633d8e26dff833ded278fcf14d5315/src/scene/graph-node.js#L20

ErikSom: Yes - but getRotation doesn't read from that global rotation.

ErikSom: Now I see the issue, I'm referring to `entity.rotation` and not `entity.getRotation`, but its weird as `.rotation` seems to a public property that does not accurately reads the rotation.

yaustar:
```
        /**
         * @type {Quat}
         * @private
         */
        this.rotation = new Quat();
```

It shouldn't show up on code predication as it's tagged by JS Docs as private

yaustar:
From out conversation on Discord @ErikSom, can we close the reported issue and PR?

PR #4796, opened 2022-10-30
rollup: add JSDoc, spacesToTabs fix, improved debugging builds

kungfooman:
I would like to suggest some changes I figured over time:

1) `spacesToTabs(enable)` requires `true` as argument, or it simply does nothing
2) `debugger` statements were stripped by the strip plugin, even in debug mode, which annoys and surprises me every time again. Since no function is stripped in Debug mode, I just enable the `strip` plugin for the other modes.
3) The regex in `spacesToTabs` never possibly could work, hence no output file actually contained tabs. Even if regex would work, it would still be a dangerous regex, that doesn't respect "line start spaces" only, so it would e.g. even replace spaces inside a string.
4) I added JSDoc with IntelliSense (the `jsconfig.json` changes are required for this - as far I tested it doesn't affect anything else  so far)
5) Plugins needs to specify a `name`, it is a mandatory property, otherwise Babel error messages cannot point out the name of the plugin that encountered an exception.
6) I would like to keep comments in my debug builds, they always help me to understand issues while debugging


willeastcott:
This looks good. But can you please ensure there's a newline before every JSDoc block please?

For comments in the debug build, didn't we make it so source maps are now available with engine builds now?

rollup.config.mjs
```diff
@@ -14,12 +14,24 @@ import dts from 'rollup-plugin-dts';
 import jscc from 'rollup-plugin-jscc';
 import { visualizer } from 'rollup-plugin-visualizer';
 
+/** @typedef {import('rollup'              ).RollupOptions                } RollupOptions                 */
+/** @typedef {import('rollup'              ).Plugin                       } Plugin                        */
+/** @typedef {import('rollup'              ).OutputOptions                } OutputOptions                 */
+/** @typedef {import('@rollup/plugin-babel').RollupBabelInputPluginOptions} RollupBabelInputPluginOptions */
+/** @typedef {import('@rollup/plugin-strip').RollupStripOptions           } RollupStripOptions            */
```
willeastcott, line 21: /** @typedef {import('rollup').RollupOptions} RollupOptions */
/** @typedef {import('rollup').Plugin} Plugin */
/** @typedef {import('rollup').OutputOptions} OutputOptions */
/** @typedef {import('@rollup/plugin-babel').RollupBabelInputPluginOptions} RollupBabelInputPluginOptions */
/** @typedef {import('@rollup/plugin-strip').RollupStripOptions} RollupStripOptions */

willeastcott: I know this is visually pleasing but we don't align imports/typedefs/etc like there elsewhere in the codebase, so for the purpose of standardization, I've suggested the above.

jsconfig.json
```diff
@@ -3,7 +3,8 @@
...
         "checkJs": true,
         "module": "es6",
-        "target": "es6"
+        "target": "es6",
+        "moduleResolution": "node"
     },
-    "include": ["src/**/*"]
-}
\ No newline at end of file
+    "include": ["src/**/*", "rollup.config.js"],
```
willeastcott, line 9:
```suggestion
    "include": ["src/**/*", "rollup.config.mjs"]
```

kungfooman:
> For comments in the debug build, didn't we make it so source maps are now available with engine builds now?

I just checked if sourcemaps still work and they cause no problems with comments:

image

Basically just `npm run build:sourcemaps` and then clicking on `pc.Entity.prototype.translate` in devtools.

Were you wondering about this, if it still works?

willeastcott:
No, what I meant was, why keep the comments in the debug build when you can just use source maps and see the original comments?

kungfooman:
> No, what I meant was, why keep the comments in the debug build when you can just use source maps and see the original comments?

Oh, that's what you mean, I don't really like sourcemaps for debugging. If you change/fix some code directly in the build then you won't even see your changes through the "sourcemap lense".

So what I can do right now is to just make a small fix to `rollup.config.mjs` to ignore inline debug sourcemaps and then I am free to edit the debug output files again, without sourcemaps messing around with my changes.

I realize my work flow is maybe a bit strange, but I just try to not rebuild everything all the time, since it takes quite some time.

Then the argument is, that I could also use the watch-mode, but even though it's quicker, it still takes like three seconds - for updating one file change. And then always these background tasks you need to deal with (which block my default VSCode shell aswell... ending up creating multiple ones that I need to cognize), I feel way too impatient for all that.

mvaligursky (approved):
This all looks great to me! 
Any further thoughts on this @willeastcott ?

willeastcott (approved):

willeastcott:
@mvaligursky LGTM.

mvaligursky:
Thanks for an awesome contribution @kungfooman !

Issue #4791, opened 2022-10-27
Clustered lighting ignores layers on lights and meshes


yaustar:
Project: https://playcanvas.com/project/1001340/overview/clustered-lighting-with-layers

There's a light for Layer A and B and Boxes that are also on Layer A and B

Move Box B to be under Light A and it is lit by that Light A if clustered lighting is enabled

https://user-images.githubusercontent.com/16639049/198271446-4af307d9-ff7c-4e37-8720-7c8aeac40646.mp4

With clustered lighting disabled, Box B is not lit

https://user-images.githubusercontent.com/16639049/198271496-b5c6bc50-1aa1-4653-be3d-02cfb660d7ef.mp4

Expectation: Box B should not be lit when moved under Light A as they are on different layers

PR #4799, opened 2022-10-31
[Fix] Clustered lighting correctly supports light layers

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/4791

- instead of using all lights for clusters, only a set of lights used by the layer is used.

### DIFF
--- a/src/scene/lighting/world-clusters.js
+++ b/src/scene/lighting/world-clusters.js
@@ -272,10 +272,7 @@ class WorldClusters {
         const usedLights = this._usedLights;
         let lightIndex = 1;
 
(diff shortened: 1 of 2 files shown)


willeastcott (approved):
Approving although slightly concerned about the `forEach` call.  I tend to avoid those within the render loop because `forEach` is like an order of magnitude slower than a `for` loop (he says without the supporting data to hand).

mvaligursky:
Yeah I agree, but the data is stored in a Set and forEach is the only way to iterate it. I could maintain a dirty flag and update an array if Set changes, but I suspect this could be fine for just lights doing a single loop over those per frame.

Issue #4795, opened 2022-10-28
Clustered lighting spotlight is culled (?) too early


yaustar:
Project: https://playcanvas.com/project/1001876

Video

https://user-images.githubusercontent.com/16639049/198632227-763aebb7-41ea-492a-a129-ce2c280eb467.mp4

As the camera moves closer to the cube, you can see the light stops lighting as the camera moves closer to the cube

From some testing, I can't reproduce the issue with point lights and is more apparent with low FOVs on the camera.

Bug is reported by a client

PR #4800, opened 2022-10-31
[Fix] Fix the bounding sphere calculation for spot lights

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/4795

Some old code was generating very approximate bounding sphere for spot light. This has been replaced by exact code now, to correctly handle spot light culling.

Before

https://user-images.githubusercontent.com/59932779/199060279-25d03a5b-cfbf-4180-aca6-f30ff416024c.mov

Now

https://user-images.githubusercontent.com/59932779/199060309-5597039b-330d-46ee-bfcf-020198045768.mov

### DIFF
--- a/src/scene/light.js
+++ b/src/scene/light.js
@@ -15,8 +15,6 @@ import {
 } from './constants.js';
 import { ShadowRenderer } from './renderer/shadow-renderer.js';
 
-const spotCenter = new Vec3();
-const spotEndPoint = new Vec3();
 const tmpVec = new Vec3();
 const tmpBiases = {
     bias: 0,
--- a/src/scene/light.js
+++ b/src/scene/light.js
@@ -727,11 +727,11 @@ class Light {
             tmpVec.copy(node.up);
 
             if (angle > 45) {
-                tmpVec.mulScalar(-size * cosAngle);
                 sphere.radius = size * this._outerConeAngleSin;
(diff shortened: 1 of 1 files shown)


src/scene/light.js
```diff
@@ -713,24 +718,23 @@ class Light {
...
-
-            tmpVec.copy(node.right);
-            tmpVec.mulScalar(Math.sin(angle * math.DEG_TO_RAD) * range);
-            spotEndPoint.add(tmpVec);
+            if (angle > 45) {
+                tmpVec.mulScalar(-size * cosAngle);
+                sphere.radius = size * this._outerConeAngleSin;
+            } else {
+                tmpVec.mulScalar(size / (-2 * cosAngle));
+                sphere.radius = size / (2 * cosAngle);
```
kungfooman, line 734: I don't know how performance critical this is, it just looks like the same calculation (negated), so that:

```suggestion
                const r = size / (2 * cosAngle);
                tmpVec.mulScalar(-r);
                sphere.radius = r;
```

(completely untested)

mvaligursky: yep don't see why not

willeastcott (approved):

willeastcott:
Now that's what I call a great PR description! 👏

PR #4814, opened 2022-11-03
Add grab pass toggles as component attribute

GSterbrant:
### Description

To enable grab passes for depth and color on a camera, it was required to execute ```requestSceneColorMap``` or ```requestSceneDepthMap``` on the camera component, but that wouldn't allow serialisation and loading of these attributes.


src/framework/components/camera/component.js
```diff
@@ -289,6 +289,10 @@ class CameraComponent extends Component {
         }
     }
 
+    set renderSceneColorMap(value) {
```
willeastcott, line 292: Just a suggestion....but maybe `renderColorMap` and `renderDepthMap` are a bit more succinct? The `Scene` part seems kinda superfluous.

GSterbrant: Sure, we can rename it but then we should rename the getter too, which would require an API change :)

mvaligursky: I think I added 'scene' to the name on purpose .. otherwise it's not very clear what color map the camera should render. I sounded almost like a property to enable / disable color and depth write on the camera.

GSterbrant: We could also call it something completely different, like depthGrabPass/colorGrabPass.

src/framework/components/camera/component.js
```diff
@@ -289,6 +289,10 @@ class CameraComponent extends Component {
...
+        this.requestSceneColorMap(value);
```
mvaligursky, line 293: I'm not sure it's correct this just calls requestSceneColorMap with the value.

In most cases (default), this will be false, and it will cause the internal counter to be decremented to -1.

So if something else would set it to true, the counter would be incremented to 0, and it would still not render, as it needs to be > 0.

src/framework/components/camera/component.js
```diff
@@ -289,6 +289,14 @@ class CameraComponent extends Component {
         }
     }
 
+    set renderSceneColorMap(value) {
+        if (value) {
+            this._colorRequestedBefore = true;
```
mvaligursky, line 294: please add this to constructor as well, so that we do not introduce class members during runtime.

src/framework/components/camera/component.js
```diff
@@ -289,6 +289,14 @@ class CameraComponent extends Component {
...
+            this.requestSceneColorMap(true);
+        } else if (this._colorRequestedBefore)
+            this.requestSceneColorMap(false);
```
mvaligursky, line 297: I think in the else branch we need to set _requestedBefore to false, so that multiple calls with false won't decrement counter more than once.

GSterbrant: So to be clear, we want this API to clamp the counter at 0? Because in that case, we could just use the counter to test if we should decrement or not.

mvaligursky: that's not the case. Any client can only cancel request made by themselves. They cannot cancel any request made by somebody else. We need a flag to know if we requested, and only cancel it one time.

src/framework/components/camera/component.js
```diff
@@ -283,12 +286,23 @@ class CameraComponent extends Component {
...
         this._renderSceneColorMap += enabled ? 1 : -1;
+        Debug.assert(this._renderSceneColorMap >= 0);
         const ok = this._enableDepthLayer(enabled);
         if (!ok) {
             Debug.warnOnce('CameraComponent.requestSceneColorMap was called, but the camera does not have a Depth layer, ignoring.');
         }
     }
 
+    set renderSceneColorMap(value) {
+        if (value) {
```
mvaligursky, line 297: one more change. this probably needs to be:
`if (value && !_sceneColorMapRequested) {`
to avoid multiple requests.

GSterbrant: Done

src/framework/components/camera/component.js
GSterbrant: True, we should keep track of the request cast by the component internally, doing that.

mvaligursky (approved):

Issue #4775, opened 2022-10-20
Rendering resolution is higher in WebXR AR than not


yaustar:
Project: https://playcanvas.com/project/998364/overview/webxr-ar-camera-high-dpr

When in AR, it looks like we are rendering in a higher resolution than when we are not. Device pixel ratio is disabled

The 'i' icon in the bottom left is an example of this. Expectation is that the resolution should be the same between the two

Normal
![Screenshot_20221020-155231](https://user-images.githubusercontent.com/16639049/196984009-eb19a48c-b03c-4755-a526-2de5987b43cd.png)

In AR
![Screenshot_20221020-155301](https://user-images.githubusercontent.com/16639049/196984293-0d83439e-85b6-4bf1-a8a6-881eb6e2009d.png)

Maksims:
Framebuffer is created for AR session by underlying system.
When starting XR session, it is possible to provide base layer for rendering, with frame buffer scaler! https://www.w3.org/TR/webxr/#xrwebgllayer-interface

Important to ensure how things like Depth Sensing and few others will be affected.

Also, I've avoided Layers all together from integration, as it is large and complex, and requires well thought API and integration design.

yaustar:
@Maksims I'm looking at the XR manager here: https://github.com/playcanvas/engine/blob/main/src/framework/xr/xr-manager.js#L675

And it changes the resolution of the graphics device when we first enter via the framebufferWidth/Height. 

I guess my question here is whether it should be user controllable on what the resolution scale in XR should be? Or should it be left up to the device via [getNativeFramebufferScaleFactor](https://www.w3.org/TR/webxr/#dom-xrwebgllayer-getnativeframebufferscalefactor)?

yaustar:
I was looking at `_onSessionStart` and changing the creation of the layer to:

```
        this._baseLayer = new XRWebGLLayer(session, this.app.graphicsDevice.gl, {
            alpha: true,
            depth: true,
            stencil: true,
            framebufferScaleFactor: adjustedScaleFactor
        });
```
For a test and to match the scale of resolution that is used in the non XR part of the app but the framebuffer being created is still at full resolution with `getNativeFramebufferScaleFactor` returning 1 instead of the value set above 🤔

PR #4816, opened 2022-11-03
[Breaking] WebXr framebuffer to match render resolution scale outside of XR

yaustar:
Fixes https://github.com/playcanvas/engine/issues/4775

This is potentially a breaking change as projects that don't have device pixel ratio enabled will now be rendering at a lower resolution in AR.

@Maksims mentioned potential issues with other features like Depth Sensing but I can't see how it would be affected looking at the APIs and docs.

Also fixes an issue I found where some browsers like the Oculus Quest one has a device pixel ratio of less than 1. This meant that browsing projects in the browser (non XR) was rendering at a higher resolution when device pixel ratio in Editor is disabled.


src/platform/graphics/graphics-device.js
```diff
@@ -140,7 +140,9 @@ class GraphicsDevice extends EventHandler {
         this._width = 0;
         this._height = 0;
 
-        this._maxPixelRatio = 1;
+        // Some devices window.devicePixelRatio can be less than one
+        // eg Oculus Quest 1
+        this._maxPixelRatio = Math.min(1, platform.browser ? window.devicePixelRatio : 1);
```
mvaligursky, line 145: If I understand the intent here, we don't want to allow ratio less than one. In that case we should use max instead of min?

src/framework/xr/xr-manager.js
```diff
@@ -613,10 +613,16 @@ class XrManager extends EventHandler {
         this._camera.on('set_nearClip', onClipPlanesChange);
         this._camera.on('set_farClip', onClipPlanesChange);
 
+        // A framebufferScaleFactor scale of 1 is the full resolution of the display
+        // so we need to calculate this based on devicePixelRatio of the dislay and what
+        // we've set this in the graphics device
+        const framebufferScaleFactor = (1 / window.devicePixelRatio) * this.app.graphicsDevice.maxPixelRatio;
```
mvaligursky, line 619: what if window is not defined here? I guess we assume this executes only in the browser?

also the code could be written as 
```
const framebufferScaleFactor = this.app.graphicsDevice.maxPixelRatio / window.devicePixelRatio;
```

src/platform/graphics/graphics-device.js
yaustar: Ah, maybe I should explain in the comment better. We *do* want to allow the pixel ratio to be less than 1 otherwise we are rendering at higher resolution than the browser's native resolution.

In this case, the Quest DPR is 0.8 so before this change, the resolution of the canvas is higher with device pixel ratio disabled than it is with it enabled. As with it disabled, `_maxPixelRatio` and with it enabled, it is set at `window. devicePixelRatio` which would be 0.8

src/framework/xr/xr-manager.js
yaustar: > what if window is not defined here? I guess we assume this executes only in the browser?

Given it's XR, I was assuming it would only be ran in the browser. I could do `platform.browser ? window.devicePixelRatio : 1` for safety. 

There are a number of places in the engine where we use `window.devicePixelRatio` directly so I don't know if we have a consistent practise/rules on this?

src/platform/graphics/graphics-device.js
mvaligursky:
```
this._maxPixelRatio = platform.browser ? Math.min(1, window.devicePixelRatio) : 1;
```

src/framework/xr/xr-manager.js
mvaligursky: maybe just add Debug.assert to validate.

Maksims:
Is default setting in Editor, takes dpi in account, and makes by default resolution lower?
If this is a case, then most existing applications will get affected by this change. Worth communicating it to users somehow.

yaustar:
> Is default setting in Editor, takes dpi in account, and makes by default resolution lower? 
> If this is a case, then most existing applications will get affected by this change. Worth communicating it to users somehow.

@Maksims That is correct. It will only affect WebXR AR experiences as far as I can tell. I intend to add a 'breaking' section on the release notes and also when it makes Editor current version.

Beyond that, maybe a tweet or forum post?

src/framework/xr/xr-manager.js
yaustar: Added an assert

src/platform/graphics/graphics-device.js
yaustar: Updated code on suggestion

mvaligursky (approved):

PR #4819, opened 2022-11-06
Add touchenter event, add JSDoc, small refactor

### LINKED ISSUE #4755: ElementComponent 'mousemove' event fires outside element when LMB is pressed
### Description
Event 'mousemove' is invoked on ElementComponent, with useInput set to true, when mouse is outside the element but previously LMB was pressed inside this element. Works correctly on engine version 1.55.0.

* Browser: Chrome
* Sample project: https://playcanvas.com/project/996750/overview/elementmousemove
* Sample Code:
```
const TestMouse = pc.createScript('testMouse');
TestMouse.attributes.add('status', { type: 'entity' });
TestMouse.attributes.add('target', { type: 'entity' });

TestMouse.prototype.initialize = function() {
    this.isOver = false;
    this.targetElement = this.target?.element;
    if (!this.targetElement) return;

    this.targetElement.on('mouseenter', this.onEnter, this);
    this.targetElement.on('mouseleave', this.onLeave, this);
    this.targetElement.on('mousemove', this.onMove, this);
};
TestMouse.prototype.update = function() {
    this.status.element.text = 'Over UI: ' + this.isOver;
};
TestMouse.prototype.onEnter = function() {
    this.isOver = true;
};
TestMouse.prototype.onLeave = function() {
    this.isOver = false;
};
TestMouse.prototype.onMove = function() {
    this.isOver = true;
};
```

### Steps to Reproduce

1. Go to: https://launch.playcanvas.com/1563175?debug=true
2. Press and hold LMB inside white rectangle and move mouse outside the rectangle
3. Text on top will still show that mouse is inside ui
4. Now go to: https://launch.playcanvas.com/1563175?debug=true&version=1.55.0 <- notice older version
5. Repeat step 2
6. Now event will work correctly, text on top will change to false

kungfooman:
Fixes #4755

Firing `mousemove` even when the mouse is outside the `ElementComponent` is actually correct behaviour and other elements depend on this, e.g. `ScrollViewComponent`. So the first misconception in the issue is to imply "mouse must be over element"  based on "mousemove". The events that actually give you this information are `mouseenter`/`mouseleave`.

That works just fine, but when I tested this on iPhone, I realized there isn't even a `touchenter` event yet, so this PR is adding that for the same abilities across plattforms.

I saw @jpauloruschel working with this and your last PR also changed the `mousemove` behaviour: https://github.com/playcanvas/engine/commit/ff00ed58a50f8f5633c39ae48842805183ac3244

IMO the current behaviour is exactly how it should be, otherwise any developer who can only rely on `mousemove` events trying to implement e.g. a custom scroll element simply has no events to base the scrolling behaviour on.

Maybe I am completely wrong, any thoughts/feedback?

### DIFF
--- a/src/framework/input/element-input.js
+++ b/src/framework/input/element-input.js
@@ -369,9 +369,13 @@ class ElementInput {
         this._touchmoveHandler = this._handleTouchMove.bind(this);
         this._sortHandler = this._sortElements.bind(this);
 
+        /** @type {ElementComponent[]} */
         this._elements = [];
+        /** @type {ElementComponent|null} */
         this._hoveredElement = null;
+        /** @type {ElementComponent|null} */
         this._pressedElement = null;
+        /** @type {Record<number, {x: number, y: number, camera: CameraComponent, element: ElementComponent}>} */
         this._touchedElements = {};
         this._touchesForWhichTouchLeaveHasFired = {};
         this._selectedElements = {};
@@ -547,10 +551,17 @@ class ElementInput {
         this._onElementMouseEvent('mousewheel', event);
     }
 
+    /**
+     * @param {TouchEvent} event - The TouchEvent.
+     * @returns {Record<number, {x: number, y: number, camera: CameraComponent, element: ElementComponent}>} - Object
+     * containing information about touched ElementComponent's.
+     * @private
+     */
     _determineTouchedElements(event) {
+        /** @type {Record<number, {x: number, y: number, camera: CameraComponent, element: ElementComponent}>} */
         const touchedElements = {};
         const cameras = this.app.systems.camera.cameras;
-
+        const { changedTouches } = event;
         // check cameras from last to front
         // so that elements that are drawn above others
         // receive events first
@@ -558,19 +569,20 @@ class ElementInput {
             const camera = cameras[i];
 
             let done = 0;
-            const len = event.changedTouches.length;
+            const len = changedTouches.length;
             for (let j = 0; j < len; j++) {
-                if (touchedElements[event.changedTouches[j].identifier]) {
+                const changedTouch = changedTouches[j];
+                if (touchedElements[changedTouch.identifier]) {
                     done++;
                     continue;
                 }
 
-                const coords = this._calcTouchCoords(event.changedTouches[j]);
+                const coords = this._calcTouchCoords(changedTouch);
 
                 const element = this._getTargetElementByCoords(camera, coords.x, coords.y);
                 if (element) {
                     done++;
-                    touchedElements[event.changedTouches[j].identifier] = {
+                    touchedElements[changedTouch.identifier] = {
                         element: element,
                         camera: camera,
                         x: coords.x,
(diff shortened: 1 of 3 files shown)


src/framework/input/element-input.js
```diff
@@ -369,9 +369,13 @@ class ElementInput {
         this._touchmoveHandler = this._handleTouchMove.bind(this);
         this._sortHandler = this._sortElements.bind(this);
 
+        /** @type {ElementComponent[]} */
```
willeastcott, line 372: Presumably this appears in the JSDoc output?

src/framework/input/element-input.js
```diff
@@ -369,9 +369,13 @@ class ElementInput {
...
         this._elements = [];
+        /** @type {ElementComponent|null} */
         this._hoveredElement = null;
+        /** @type {ElementComponent|null} */
         this._pressedElement = null;
+        /** @type {Record<number, {x: number, y: number, camera: CameraComponent, element: ElementComponent}>} */
```
willeastcott, line 378: I think this would be the first appearance of `Record` in the source's JSDocs. I can't find any resources documenting that (e.g. here: https://jsdoc.app/tags-type.html).

src/framework/input/element-input.js
kungfooman: Thank you for noting, I always forget `@ignore` :see_no_evil:

src/framework/input/element-input.js
kungfooman: Yea, I should definitely use `Object<..., ...>`. I am working on it + some other changes (converting to style without `@typedef`)

kungfooman:
Can anyone think of any objections to making `_clickedEntities` a definitely assigned property of `ElementInput`?

Right now in `ElementInput#constructor`:

https://github.com/playcanvas/engine/blob/b232b1380564018d477da156066f7abe39ca3630/src/framework/input/element-input.js#L378-L381

First it checks `platform.touch` and then later `AppBase` is calling `ElementInput#attachSelectEvents`:

https://github.com/playcanvas/engine/blob/b232b1380564018d477da156066f7abe39ca3630/src/framework/app-base.js#L519-L522

Which usually assigns `ElementInput#_clickedEntities` (just not in cases that a browser doesn't support XR):

https://github.com/playcanvas/engine/blob/b232b1380564018d477da156066f7abe39ca3630/src/framework/input/element-input.js#L435-L443

It would remove some lines of code and simplify the `{}|undefined` type to `{}`.

kungfooman:
I forked @mgawinKatana's project which handles now mouse/touch for testing this PR:

https://playcanvas.com/project/1007930/overview/elementmousemove-fix

https://launch.playcanvas.com/1587642

https://launch.playcanvas.com/1587642?use_local_engine=http://localhost/playcanvas-engine-pr/build/playcanvas.js

Is there a way to `use_local_engine` for `10.*.*.*` or `192.168.*.*` private networks? Then it would be testable on local touch devices connected via LAN/WLAN.

yaustar:
> Is there a way to use_local_engine for 10.*.*.* or 192.168.*.* private networks? Then it would be testable on local touch devices connected via LAN/WLAN.

I'm afraid not, no. 

The way I've been doing it is either overrides in the devtools or putting the engine file into the project and having it load after the engine.

src/framework/input/element-input.js
```diff
@@ -659,33 +682,44 @@ class ElementInput {
...
                     // events keep on firing for the same element until the touch ends, even
                     // if the touch position moves away from the element. Touchleave, on the
                     // other hand, should fire once when the touch position moves away from
                     // the element and then not re-fire again within the same touch session.
-                    this._touchesForWhichTouchLeaveHasFired[touch.identifier] = true;
+                    _touchesForWhichTouchLeaveHasFired[identifier] = true;
                 }
-
-                this._fireEvent('touchmove', new ElementTouchEvent(event, oldTouchInfo.element, oldTouchInfo.camera, coords.x, coords.y, touch));
+                // And fire touchenter if we've entered the previously left touched element again
```
yaustar, line 707: I feel like we should fire touchenter if a user has clicked outside the element and moved into it as well. Seems odd to only fire it if we started in an element, leave and then re-enter

yaustar (commented):
We also need to bubble these events to button like we do for touch leave https://github.com/playcanvas/engine/blob/main/src/framework/components/button/component.js#L297

PR #4820, opened 2022-11-06
Remove `@typedef` blocks from codebase

willeastcott:
Typedefs appear at the top of our modules and look like this:

```javascript
/** @typedef {import('../../app-base.js').AppBase} AppBase */
```

Then, throughout the JSDocs in that module, you can simply do:

```javascript
@type {AppBase}
```

This PR removes usage of the `@typedef` JSDoc tag from the codebase. So now, the code does:

```javascript
@type {import('../../app-base.js').AppBase}
```

Advantages:

1. The typedefs are actual defined types to JSDoc! Our custom JSDoc template that builds the API reference manual has had to be hacked to ignore them. Otherwise we have something like 50+ occurrences of most classes in the JSDocs. I would like to start considering a different/better JSDoc template for our API ref, but it's not easy when our current output looks bad in everything except our template. Our JSDoc comments should trivially generate a nice and valid API Reference site using any template with as few hacks as possible.
2. It's not obvious to those who are new to the source why the typedefs are there.
3. Our autogenerated TypeScript declarations file suddenly becomes much simpler/cleaner. As as an added bonus, we can probably dump the `types-undollar.mjs` script! 🎉 
4. It gives better IntelliSense (currently we get lots of [Type -> Type] popups). So this is before:

typedef

This is after:

typedefgone

Disadvantages:

1. Some modules reference typedef-based names more than once. `standard-material.js` is perhaps the worst offender here. This can result in the repetition of type imports in the JSDoc blocks. However, 90% of modules only use a typedef type once, meaning that removing typedefs actually makes the code more concise.


types-fixup.mjs
```diff
@@ -469,6 +478,12 @@ const standardMaterialProps = [
 path = './types/scene/materials/standard-material.d.ts';
 dts = fs.readFileSync(path, 'utf8');
 dts = dts.replace('reset(): void;', 'reset(): void;\n' + getDeclarations(standardMaterialProps));
+dts += `
```
mvaligursky, line 520: I would not mind in code comments for changes in this file as they are not obvious.

mvaligursky (approved):
This looks pretty good actually!

types-fixup.mjs
willeastcott: Added some comments :smile:

slimbuck (approved):

kungfooman:
> Advantages:
> 
> 1. The typedefs are actual defined types to JSDoc! Our custom JSDoc template that builds the API reference manual has had to be hacked to ignore them. Otherwise we have something like 50+ occurrences of most classes in the JSDocs. I would like to start considering a different/better JSDoc template for our API ref, but it's not easy when our current output looks bad in everything except our template. Our JSDoc comments should trivially generate a nice and valid API Reference site using any template with as few hacks as possible.

I didn't investigate too much, just wondering, because you said "has had to be hacked". Isn't it still hacked because `@implements {...}` doesn't support `import(...)`?

And wouldn't that perpetuate itself again into other JSDoc templates if they don't support `@typedef`?

willeastcott:
@kungfooman This is the hack we currently have to do to strip all of those `typedef` statements:

https://github.com/playcanvas/playcanvas-jsdoc-template/blob/main/publish.js#L449-L452

While I made this PR, I noticed that `@implements` didn't support `import` so I left that as it is for now. To be honest, I don't think the engine's half-hearted attempt to define interfaces (`TextureParser` and `ResourceHandle`) are particularly useful. Maybe they should be removed.

kungfooman:
> While I made this PR, I noticed that `@implements` didn't support `import` so I left that as it is for now. To be honest, I don't think the engine's half-hearted attempt to define interfaces (`TextureParser` and `ResourceHandle`) are particularly useful. Maybe they should be removed.

I agree, I think you can just replace `@implements {ResourceHandler}` with `extends ResourceHandler`, just that you have to add `super()` in constructor aswell then.

PR #4830, opened 2022-11-08
New createGraphicsDevice function to handle both WebGU and WebGPU device creation

mvaligursky:
### New API
- **createGraphicsDevice(canvas, options)** function, allowing async creation of device, supporting both WebGL and WebGPU devices, returning a promise, resolving to a device. The async need is there due to initialization of WebGPU being async. This function is intended as an easy way to create any compatible device, but the user will still be able to allocate a single device only should they prefer that - for example WebglGraphicsDevice and tree-shake out WebGPU implementation.

Example of how this can be used together with AppBase, to create WebGPU device is supported, otherwise fallback to WebGL device. WebGPU implementation currently depends on glslang to transform glsl shaders to wgpu shaders, and so in this case, a url to the library needs to be passed in. The library is dynamically loaded internally.
```
const gfxOptions = {
   deviceTypes: [pc.DEVICETYPE_WEBGPU, pc.DEVICETYPE_WEBGL],
   glslangUrl: '/static/lib/glslang/glslang.js'
};
pc.createGraphicsDevice(canvas, gfxOptions).then((graphicsDevice: pc.GraphicsDevice) => {
   const createOptions = new pc.AppOptions();
   createOptions.graphicsDevice = graphicsDevice;
}).catch(console.error);

```
### Additional changes
- WebgpuGraphicsDevice is built into debug only playcanvas engine, allowing for easier experimentation and development
- Shapes example has been converted to use this new functionality and AppBase, allowing easy way to run it on WebGPU
- glslang library has been moved to a new home inside lib/glslang, where other wasm libraries are stored


willeastcott:
Is this till draft?

mvaligursky:
> Is this still draft?

Not anymore, sorry, please review.

PR #4831, opened 2022-11-08
Shorten some pixel format enums

slimbuck:
This PR shortens 6 pixel format enums since IMO `PIXELFORMAT_R8_G8_B8_A8` is less readable (and typeable) than `PIXELFORMAT_RGBA8`. The shorter convention also matches more closely some of the other pixel formats, like `RGBA32F`.


willeastcott (approved):
This PR is a thing of true beauty! ❤️

src/platform/graphics/constants.js
```diff
@@ -419,42 +419,42 @@ export const PIXELFORMAT_L8 = 1;
  *
  * @type {number}
  */
-export const PIXELFORMAT_L8_A8 = 2;
+export const PIXELFORMAT_LA8 = 2;
```
mvaligursky, line 422: I think these constants should be properly deprecated? Either keep them here (in addition to new ones), or add them to deprecated, but not to gfx name space as currently.

mvaligursky: actually, I take it back, there are in deprecated in both places. All good.

Issue #4776, opened 2022-10-20
WebXR AR screen space UI input doesn't register touches correctly


yaustar:
Project: https://playcanvas.com/project/998371/overview/webxr-ar-ui-screen-space-broken

Pressing on the very far left corner (shown below) doesn't register you've pressed the button
![image](https://user-images.githubusercontent.com/16639049/196992191-54f1b477-b004-4a23-a887-fb1d0feaab1e.png)

yaustar:
I believe this is because as part of the worldToScreen function call, the camera updates the projection matrix based on the FOV, aspect ratio, etc as shown here:

https://github.com/playcanvas/engine/blob/554e659619a381b1f43ad890121cffd2daae20b5/src/scene/camera.js#L478

However, when in XR, in the renderer, the camera uses a different projection and view matrix
https://github.com/playcanvas/engine/blob/main/src/scene/renderer/forward-renderer.js#L285

Therefore, it looks like the raycasting is using the wrong matrix for transformation of the ray.

mvaligursky:
related to https://github.com/playcanvas/engine/issues/3635

yaustar:
Wondering if we could either:

1. Update the FOV, aspect ratio etc values etc back to camera in the same way 8th Wall and Zappar do it
2. Use the Input XR Sources for the ray when in XR for the element input logic instead of camera projection

Maksims:
For the UI, it should be using input source rays already.

yaustar:
> For the UI, it should be using input source rays already.

@Maksims It seems to be only use that for World Space UI, not screen space?

yaustar:
I've tried 1. in https://playcanvas.com/project/1001902/overview

```
AugmentedRealityManager.prototype._onXrTracking = function () {
    const camera = this.arCameraEntity.camera.camera;
    if (camera.xr && camera.xr.views.length) {
        // calculate frustum based on XR view
        const viewProjMat = new pc.Mat4();
        const view = camera.xr.views[0];
        //viewProjMat.mul2(view.projMat, view.viewOffMat);
        
        viewProjMat.copy(view.projMat);
        const data = viewProjMat.data;

        const fov = (2.0 * Math.atan(1.0 / data[5]) * 180.0) / Math.PI;
        const aspectRatio = data[5] / data[0];
        const farClip = data[14] / (data[10] + 1);
        const nearClip = data[14] / (data[10] - 1);

        this.arCameraEntity.camera.horizontalFov = false;
[...]

PR #4837, opened 2022-11-10
Use hardware device properties (fov) when in XR

yaustar:
Fixes https://github.com/playcanvas/engine/issues/4776

This changes the camera component getters for certain properties to use the AR camera FOV, clip planes etc instead of the components so that screenToWorld is done correctly for the UI input


src/scene/camera.js
```diff
@@ -65,13 +65,19 @@ class Camera {
         this._viewProjMatDirty = true;
 
         this.frustum = new Frustum();
+
+        // Set by XrManager
+        this._xrHorizontalFov = this._horizontalFov;
```
slimbuck, line 70: Might be neater to group these settings in an object of its own? Then names can mirror the values on the camera, so something like:
```
this.xr = {
    horizonalFov: this._horizontalFov,
    fov: this._fov,
    .... etc
};
```

yaustar: I did wonder that. Would we group the camera component counter parts so that the separation is clearer?

src/framework/xr/xr-manager.js
```diff
@@ -581,8 +581,8 @@ class XrManager extends EventHandler {
             if (this._camera) {
                 this._camera.off('set_nearClip', onClipPlanesChange);
                 this._camera.off('set_farClip', onClipPlanesChange);
-
                 this._camera.camera.xr = null;
```
mvaligursky, line 584: we should add the `xr` property on a camera, it currently does not exist. (existing issue, I know)

src/framework/xr/xr-manager.js
```diff
@@ -746,6 +747,29 @@ class XrManager extends EventHandler {
...
+            viewProjMat.copy(view.projMat);
+            const data = viewProjMat.data;
+
+            const fov = (2.0 * Math.atan(1.0 / data[5]) * 180.0) / Math.PI;
+            const aspectRatio = data[5] / data[0];
+            const farClip = data[14] / (data[10] + 1);
+            const nearClip = data[14] / (data[10] - 1);
+
+            const camera = this._camera.camera;
+            camera._xrAspectRatio = aspectRatio;
```
mvaligursky, line 764: ideally we don't access private properties, but create some function on the Camera class for this, maybe `setXrProperties` or similar.

src/scene/camera.js
```diff
@@ -187,7 +193,7 @@ class Camera {
     }
 
     get farClip() {
-        return this._farClip;
+        return (this.xr && this.xr.active) ? this._xrFarClip : this._farClip;
```
mvaligursky, line 196: `this.xr && this.xr.active` can be `this.xr?.active`

src/framework/xr/xr-manager.js
mvaligursky: internally it would also do `_projMatDirty = true;`

src/framework/xr/xr-manager.js
mvaligursky: in its setter, it should also handle setting `_projMatDirty` so we don't need to poke it in like this

src/scene/camera.js
mvaligursky: `this.xr` is used already, so be careful here. Maybe call it `this.xrProperties`
Also with moving non-xr properties to an object - there could be code directly accessing those in the engine, so maybe leave it as is for now.

yaustar:
Updated from Martin's feedback

src/framework/xr/xr-manager.js
```diff
@@ -746,6 +747,29 @@ class XrManager extends EventHandler {
...
+            const aspectRatio = data[5] / data[0];
+            const farClip = data[14] / (data[10] + 1);
+            const nearClip = data[14] / (data[10] - 1);
+
+            const camera = this._camera.camera;
+            camera.setXrProperties({
+                aspectRatio,
+                farClip,
+                fov,
+                horizontalFov,
```
slimbuck, line 768: Is this github error message correct? I also don't see where horizontalFov comes from.

src/framework/xr/xr-manager.js
```diff
@@ -581,8 +581,8 @@ class XrManager extends EventHandler {
             if (this._camera) {
                 this._camera.off('set_nearClip', onClipPlanesChange);
                 this._camera.off('set_farClip', onClipPlanesChange);
-
                 this._camera.camera.xr = null;
+                this._camera.camera._projMatDirty = true;
```
mvaligursky, line 585: is this line now still needed? the `xr` setter handles it.

mvaligursky (approved):
It looks good enough, but one day I'd want to further refactor it. Currently, when XR session is active, the XR manager modifies the camera to point to itself. The reason is to allow forward-renderer to access an array of views that the manager stores. I'd want to move the views to camera, and the XR manager would just update them. But that's a separate PR.

yaustar:
Created a ticket for the refactor/tech debt mentioned by Martin: https://github.com/playcanvas/engine/issues/5011

src/framework/xr/xr-manager.js
yaustar: Fixed in the lint commit. The original code was setting it to false directly on the camera so the github error is correct

PR #4838, opened 2022-11-11
Expose option to set lightmap pixel format

slimbuck:
This PR adds the option to change the lightmap pixel format. This is useful when generating very soft shadows using many samples.


src/scene/scene.js
```diff
@@ -137,6 +138,17 @@ class Scene extends EventHandler {
      */
     lightmapFilterEnabled = false;
 
+    /**
+     * Specify the pixel format to use for lightmaps. Can be:
```
willeastcott, line 142: Maybe add some info about why you'd modify this property?

willeastcott (approved):

src/scene/scene.js
mvaligursky: I think instead of exposing the format, we should just expose 'lightmapHDR' flag or similar. And under the hood, use the format which is available, or automatically fallback to LDR.

We also need to only pick the HDR if OES_texture_float_linear is available, otherwise float textures cannot be linearly filtered, and only point sampling is allowed.

Otherwise, the user would need to understand this, and also we'd need to make these public: extTextureHalfFloat, textureHalfFloatRenderable, extTextureHalfFloat, textureHalfFloatUpdatable

Alternatively, we could make device.lightmapHdrFormat which would return format, or undefined if no hdr format is viable of similar.

slimbuck: I thought about adding a flag, but didn't want to restrict users who may prefer one format over the other (when both are available). Is there a reason not to make the texture extensions public? They are being used all over the show.

slimbuck: Perhaps a device getter called 'lightmapHdrFormat' as you suggestion that takes a boolean indicating whether to prefer 'larger' format (a.k.a 32F) over 'smaller' (16F)?

mvaligursky: yep, I'd be happy with that. Or maybe even `lightmapFormat(precision);` where precision can be 8, 16 or 32?
in that case, I guess we could just expose `scene.lightmapPrecision` (or similar name) to be one of those?

src/scene/scene.js
```diff
@@ -137,6 +138,14 @@ class Scene extends EventHandler {
      */
     lightmapFilterEnabled = false;
 
+    /**
+     * Enables HDR lightmaps. This can result in smoother lightmaps especially when many samples
+     * are used.
```
willeastcott, line 143:
```suggestion
     * Enables HDR lightmaps. This can result in smoother lightmaps especially when many samples
     * are used. Defaults to false.
```

willeastcott (approved):

src/scene/scene.js
```diff
@@ -774,6 +783,15 @@ class Scene extends EventHandler {
             this.prefilteredCubemaps = cubemaps.slice(1);
         }
     }
+
+    /**
+     * Get the lightmap pixel format.
+     *
+     * @type {number} The pixel format.
+     */
+    get lightmapPixelFormat() {
```
mvaligursky, line 792: this function should be moved to lightmapper, no need for a public getter like this on the scene I think

mvaligursky: ah actually the options builder uses it ..

slimbuck: that would be ideal except standard-material-options-builder needs it.

the correct way to solve this is probably to make lightmapper's black texture the correct format (it's currently RGBA8) and then use the texture in options-builder to determine encoding.

are you ok with this change?

mvaligursky (approved):

src/scene/scene.js
mvaligursky: I think I prefer the way it's done currently.

PR #4839, opened 2022-11-11
Fix missing click events in mixed touch/mouse scenarios

### LINKED ISSUE #4832: Element mouse click event does not fire if the user has touch clicked on it first
Project: https://playcanvas.com/project/1006310/overview

Repro steps:

1. Launch the tab in phone emulator mode
2. Touch click on the element a few times. It will change color on each click
3. Come out of phone emulator
4. Mouse click on the element a few times. The click event is not fired

This can cause issues on devices like laptops where you have both touch and mouse controls and are likely to swap between the two.

https://user-images.githubusercontent.com/16639049/200815692-9515b56d-f800-4d89-901b-b7f65cf18d46.mp4

kungfooman:
Fixes #4832

@yaustar's issue describes how to test this already + some extra information about the 300ms delay.

I added appropriate comments to hopefully make it clear, if there are any question please just ask.

### DIFF
--- a/src/framework/input/element-input.js
+++ b/src/framework/input/element-input.js
@@ -643,7 +643,7 @@ class ElementInput {
 
                     if (!this._clickedEntities[element.entity.getGuid()]) {
                         this._fireEvent('click', new ElementTouchEvent(event, element, camera, x, y, touch));
-                        this._clickedEntities[element.entity.getGuid()] = true;
+                        this._clickedEntities[element.entity.getGuid()] = Date.now();
                     }
 
                 }
@@ -736,15 +736,23 @@ class ElementInput {
         if (eventType === 'mouseup' && this._pressedElement) {
             // click event
             if (this._pressedElement === this._hoveredElement) {
-                this._pressedElement = null;
-
-                // fire click event if it hasn't been fired already by the touchup handler
-                if (!this._clickedEntities || !this._clickedEntities[this._hoveredElement.entity.getGuid()]) {
+                // fire click event if it hasn't been fired already by the touchend handler
+                const guid = this._hoveredElement.entity.getGuid();
+                // Always fire, if there are no clicked entities
+                let fireClick = !this._clickedEntities;
+                // But if there are, we need to check how long ago touchend added a "click brake"
+                if (this._clickedEntities) {
+                    const lastTouchUp = this._clickedEntities[guid] || 0;
+                    const dt = Date.now() - lastTouchUp;
+                    fireClick = dt > 300;
+                }
+                // We do not check another time, so the worst thing that can happen is one ignored click in 300ms.
+                delete this._clickedEntities[guid];
+                if (fireClick) {
                     this._fireEvent('click', new ElementMouseEvent(event, this._hoveredElement, camera, targetX, targetY, this._lastX, this._lastY));
                 }
-            } else {
-                this._pressedElement = null;
             }
+            this._pressedElement = null;
         }
     }
 


src/framework/input/element-input.js
```diff
@@ -736,15 +736,23 @@ class ElementInput {
...
             // click event
             if (this._pressedElement === this._hoveredElement) {
-                this._pressedElement = null;
-
-                // fire click event if it hasn't been fired already by the touchup handler
-                if (!this._clickedEntities || !this._clickedEntities[this._hoveredElement.entity.getGuid()]) {
+                // fire click event if it hasn't been fired already by the touchend handler
+                const guid = this._hoveredElement.entity.getGuid();
+                // Always fire, if there are no clicked entities
+                let fireClick = !this._clickedEntities;
```
yaustar, line 742:
```suggestion
                let fireClick = this._clickedEntities.length > 0;
```
May be a personal style but if it's an array, I prefer to check it like an array rather than `!` as I see that as checking against null/undefined

src/framework/input/element-input.js
```diff
@@ -736,15 +736,23 @@ class ElementInput {
...
+                // But if there are, we need to check how long ago touchend added a "click brake"
+                if (this._clickedEntities) {
```
yaustar, line 744:
```suggestion
                if (this._clickedEntities.length > 0) {
```
Same as above

src/framework/input/element-input.js
```diff
@@ -736,15 +736,23 @@ class ElementInput {
...
+                    const lastTouchUp = this._clickedEntities[guid] || 0;
+                    const dt = Date.now() - lastTouchUp;
+                    fireClick = dt > 300;
+                }
+                // We do not check another time, so the worst thing that can happen is one ignored click in 300ms.
```
yaustar, line 749: Is 300ms dependent on device speed/performance?

src/framework/input/element-input.js
kungfooman: It is checking if `this._clickedEntities` exists first, because it's not definitely assigned in the constructor:
```js
if (platform.touch)
            this._clickedEntities = {};
```

src/framework/input/element-input.js
kungfooman: Same as above, btw it's also an `{}` and not an `[]`

src/framework/input/element-input.js
yaustar: Oops, thats' true! My bad

src/framework/input/element-input.js
kungfooman: Yea, that is the question, I tested on my devices and its either around 50ms or 270ms (newish iPhone vs oldish Android). In Chrome Touch devtools, I got 250ms (Chrome on Linux). Considering extreme lag / bad performance (lets say `mouseup` would be delayed to 400ms), then a double click could theoretically come through. I don't know if Chrome still fires in bad performance cases.

There is also this article that mentions "no delay" and 300ms depending on meta tag: https://web.dev/mobile-touchandmouse/

But "no delay" also means at least 1ms in my test, probably an internal Chrome "setTimeout" with 0ms, but it takes some time anyway.

yaustar: I'm pretty happy to have 300ms and this working over it not working at all. 

We can always tweak the figure later on or rework it. Given I'm the only one that has spotted this so far, we shouldn't have a problem :)

yaustar (approved):

PR #4851, opened 2022-11-16
ElementDragHelper#_toggleDragListeners: fix wrong `this` type

kungfooman:
So `this._element[onOrOff]` maps to either the on/off-method and the third argument can only be an optional scope. But the current code passes `false`, which basically means the scope of `this._element`. However, it is expecting `this`, so I assume this is the reason someone introduced the workaround:

```js
if (!this._handleMouseUpOrTouchEnd) {
    this._handleMouseUpOrTouchEnd = this._onMouseUpOrTouchEnd.bind(this);
}
```

Which works, but it's unidiomatic/cumbersome if one had just put in the right scope in the first place.


willeastcott:
Hmm. Well spotted, @kungfooman! Thanks for the fix. 😄

PR #4852, opened 2022-11-16
[Breaking] AnimController time in state fix

### LINKED ISSUE #3837: Anim transition Exit Time is being affected by source state speed

ellthompson:
The AnimController class doesn't factor in the speed of the current state when tracking the total time that state has been active. This means transitions with an exit time parameter will not trigger as designed (based on the normalised duration of the active state) unless that source state has a speed of 1.

This PR addresses this bug, but in doing so will change the behaviour of all transition's that use an exit time to determine when to move from a state with a modified speed. It also affects the values returned by the AnimComponentLayer#activeStateCurrentTime and AnimComponentLayer#activeStateProgress properties. When the active state has a modified speed value, these functions now return values which account for that speed.

Fixes #3837

### DIFF
--- a/src/framework/anim/controller/anim-controller.js
+++ b/src/framework/anim/controller/anim-controller.js
@@ -516,7 +516,7 @@ class AnimController {
         let animation;
         let clip;
         this._timeInStateBefore = this._timeInState;
-        this._timeInState += dt;
+        this._timeInState += dt * this.activeState.speed;
 
         // transition between states if a transition is available from the active state
         const transition = this._findTransition(this._activeStateName);


willeastcott:
I'm struggling to see how much disruption this breaking change will cause existing customers. Do you have any concerns if we roll this out? Do we need to carefully communicate this change in the next engine release? Anything else we could/should do to mitigate any disruption?

yaustar:
I don't think it would be a huge impact otherwise I would have thought I see more activity in the forums and ticket about animation speed

ellthompson:
Most state graphs will be fine so I think we just have to be clear in the release notes which specific usages of these properties may require adjustment.

mvaligursky (approved):

PR #4853, opened 2022-11-16
Expose buildHitCorners function in ElementInput 

yaustar:
Related to https://github.com/playcanvas/editor/issues/896 where I would like visualise the hit area in the Editor like so:

image

Not sure how to go about this. All the math used to workout the button hit area is in ElementInput.

For testing purposes, I've just made them static so I can access them in the Editor as private API

```
// Add the padding to the existing world corners
// We are always in world space in the Editor
const elementScale = pc.ElementInput.calculateScaleToWorld(entity.element)
hitCorners = pc.ElementInput.buildHitCorners(entity.element, hitCorners, elementScale.x, elementScale.y, elementScale.z);

renderLineCorners[0].copy(hitCorners[0]);
renderLineCorners[1].copy(hitCorners[1]);
renderLineCorners[2].copy(hitCorners[1]);
renderLineCorners[3].copy(hitCorners[2]);
renderLineCorners[4].copy(hitCorners[2]);
renderLineCorners[5].copy(hitCorners[3]);
renderLineCorners[6].copy(hitCorners[3]);
renderLineCorners[7].copy(hitCorners[0]);

app.renderLines(renderLineCorners, cornerColor, immediateRenderOptions);
```

This works of course but just feels wrong to have this as the standard.

So I'm think on doing:
1. Leaving it as is for minimal change to the engine
2. Making this public API and/or moving these to some utils namespace and making it public API
3. Making a separate, public function for this in ButtonComponent that gets the worldHitCorners

I'm leaning towards 3 but could use some input

### DIFF
--- a/src/framework/input/element-input.js
+++ b/src/framework/input/element-input.js
@@ -987,96 +987,6 @@ class ElementInput {
         return result;
     }
 
-    // In most cases the corners used for hit testing will just be the element's
-    // screen corners. However, in cases where the element has additional hit
-    // padding specified, we need to expand the screenCorners to incorporate the
-    // padding.
-    _buildHitCorners(element, screenOrWorldCorners, scaleX, scaleY, scaleZ) {
-        let hitCorners = screenOrWorldCorners;
-        const button = element.entity && element.entity.button;
-
-        if (button) {
-            const hitPadding = element.entity.button.hitPadding || ZERO_VEC4;
-
-            _paddingTop.copy(element.entity.up);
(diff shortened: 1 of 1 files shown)


mvaligursky:
I'd leave it as is (static) to be honest. Just add a comment that the function is called directly by the Editor, so we're aware when we change it.

mvaligursky (approved):

yaustar:
Thanks Martin!

PR #4854, opened 2022-11-16
Fix cross origin docs

slimbuck:
The cross origin specifier for texture assets should be a string not a boolean as the value is used directly (see https://github.com/playcanvas/engine/blob/main/src/framework/parsers/texture/img.js#L51).

### DIFF
--- a/src/framework/asset/asset.js
+++ b/src/framework/asset/asset.js
@@ -72,8 +72,8 @@ class Asset extends EventHandler {
      * materials).
      * @param {object} [options] - The asset handler options. For container options see
      * {@link ContainerHandler}.
-     * @param {boolean} [options.crossOrigin] - For use with texture resources. For
-     * browser-supported image formats only, enable cross origin.
+     * @param {string} [options.crossOrigin] - For use with browser-supported texture resources
+     * only, override the default cross origin.
      * @example
      * var asset = new pc.Asset("a texture", "texture", {
      *     url: "http://example.com/my/assets/here/texture.png"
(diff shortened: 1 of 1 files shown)


mvaligursky (approved):
Looks good, but should we perhaps not change the API and change the use of it? Not sure how likely we're to break existing code here passing the boolean.

src/framework/asset/asset.js
```diff
@@ -72,8 +72,8 @@ class Asset extends EventHandler {
      * materials).
      * @param {object} [options] - The asset handler options. For container options see
      * {@link ContainerHandler}.
-     * @param {boolean} [options.crossOrigin] - For use with texture resources. For
-     * browser-supported image formats only, enable cross origin.
+     * @param {string|null} [options.crossOrigin] - For use with browser-supported texture resources
```
willeastcott, line 75: Why `null`?

willeastcott: You could tighten this up further with `@param {'anonymous'}`

slimbuck:
> Looks good, but should we perhaps not change the API and change the use of it? Not sure how likely we're to break existing code here passing the boolean.

No we want to expose a string because the value can be 'anonymous' or 'use-credentials' or null, and it makes sense that these values override the default. (i.e. in some case you may want to null crossOrigin).

mvaligursky:
Sounds good. Are people likely to pass booleans to the function though? If they do, maybe we can handle it and convert to some string value? And log a warning.

mvaligursky:
Actually, you didn't change a function, only docs, so that will not cause any problems. All good.

willeastcott (approved):

PR #4857, opened 2022-11-17
Fix skybox rotation

slimbuck:
Fixes https://forum.playcanvas.com/t/skybox-rotation-fails-in-1-58/28413/5

Skybox rotation in the editor was not being applied to the scene.

### DIFF
--- a/src/scene/scene.js
+++ b/src/scene/scene.js
@@ -705,7 +705,7 @@ class Scene extends EventHandler {
         this._skyboxMip = render.skyboxMip === undefined ? 0 : render.skyboxMip;
 
         if (render.skyboxRotation) {
-            this._skyboxRotation.setFromEulerAngles(render.skyboxRotation[0], render.skyboxRotation[1], render.skyboxRotation[2]);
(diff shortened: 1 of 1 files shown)


mvaligursky (approved):

src/scene/scene.js
```diff
@@ -705,7 +705,7 @@ class Scene extends EventHandler {
         this._skyboxMip = render.skyboxMip === undefined ? 0 : render.skyboxMip;
 
         if (render.skyboxRotation) {
-            this._skyboxRotation.setFromEulerAngles(render.skyboxRotation[0], render.skyboxRotation[1], render.skyboxRotation[2]);
+            this.skyboxRotation = (new Quat()).setFromEulerAngles(render.skyboxRotation[0], render.skyboxRotation[1], render.skyboxRotation[2]);
```
LeXXik, line 708: Just out of curiosity, why is quat enclosed with brackets here?

slimbuck: No particular reason other than I like to be explicit in this case.

I suppose `new Quat().setFromEulerAngles()` is pretty standard, but still looks a little strange to me.

Issue #4858, opened 2022-11-17
`Camera._evaluateProjectionMatrix` calculates NaN values


kungfooman:
`Camera#_projMatSkybox` contains NaN calculations, it is reproducible by using spinning cube of README.md with an extra directional light. Just paste this in devtools to check for NaN at runtime:

```js
pc.Camera.prototype._evaluateProjectionMatrix = function () {
    if (this._projMatDirty) {
        if (this._projection === pc.PROJECTION_PERSPECTIVE) {
            this._projMat.setPerspective(this._fov, this._aspectRatio, this._nearClip, this._farClip, this._horizontalFov);
            this._projMatSkybox.copy(this._projMat);
        } else {
            const y = this._orthoHeight;
            const x = y * this._aspectRatio;
            this._projMat.setOrtho(-x, x, -y, y, this._nearClip, this._farClip);
            this._projMatSkybox.setPerspective(this._fov, this._aspectRatio, this._nearClip, this._farClip);
            for (var i=0; i<16; i++) {
                const val = this._projMatSkybox.data[i];
                if (isNaN(val)) {
                    debugger;
                }
            }
        }
        this._projMatDirty = false;
    }
}
```

It needs at least a directional light to trigger this:

```js
const light = new pc.Entity('light');
light.addComponent('light', {
    type: "directional",
    color: new pc.Color(0.8, 0.8, 0.8),
    castShadows: true,
    shadowBias: 0.05,
    normalOffsetBias: 0.05,
    shadowResolution: 2048
});
pc.app.root.addChild(light);
```

![image](https://user-images.githubusercontent.com/5236548/202414782-9e8248cb-a890-4344-b01a-87d9d303cf43.png)

mvaligursky:
@kungfooman - would you have a repro project set up for this?

kungfooman:
@mvaligursky Yep, no problem, the minimal default project is already enough: https://playcanvas.com/project/1009615/overview/test-camera-nan

Then you can just execute the snippet in devtools and the `debugger` inside the NaN detection will show you the details.

PR #4859, opened 2022-11-17
[Fix] Shadow renderer does not use nearClip of 0

mvaligursky:
Fixes: https://github.com/playcanvas/engine/issues/4858

- Shadow renderer uses ortho camera for shadow projection, for which the nearClip of 0 works fine. But Camera class itself evaluates a perspective projection for skydome (not used in this case), which cannot use clip plane of 0, causing NaN results. Those values are not used at all, but avoiding them anyways.

### DIFF
--- a/src/scene/renderer/shadow-renderer.js
+++ b/src/scene/renderer/shadow-renderer.js
@@ -352,7 +352,7 @@ class ShadowRenderer {
             // look at the center from far away to include all casters during culling
             shadowCamNode.setPosition(center);
             shadowCamNode.translateLocal(0, 0, 1000000);
-            shadowCam.nearClip = 0;
+            shadowCam.nearClip = 0.01;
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

willeastcott:
Approving but can you put something in the description that explains what this is fixing exactly?

mvaligursky:
> Approving but can you put something in the description that explains what this is fixing exactly?

Yes, on it .. wait with reviews for when invited ;)

willeastcott:
Haha, I normally fill out the description when I open the PR! 😉

mvaligursky:
Yeah, I use Tower desktop tool, and it only lets me enter a single line of description (cannot do Shift-Enter to go to next line), and so I start with the empty description, and then continue in the browser.

PR #4862, opened 2022-11-18
Fix packSamples calculating out-of-bounds

### LINKED ISSUE #4860: reproject-texture packSamples overcalculates 4x + index out of bounds
Thank you a lot for your super quick fix @mvaligursky!

The bad news is I have another issue, the good news is that afterwards this should be 4x faster. :sweat_smile: 

I will try to explain by debug session:

image

Basically there is a `samples` array of length 8192 and its length is assigned to `numSamples`. Then `numSamples` is iterated over one-by-one and `i` is multiplied by 4 to index the `samples` array again (so basically 75% of that for loop is out-of-bounds).

The same then happens with the `Uint8Array` that has a length of 32768 but is indexed up to 131072.

kungfooman:
Fixes #4860

Hi @slimbuck, you suggested `i < numSamples / 4` and then keep the `* 4`, but getting rid of all these divisions/multiplications via a simple `+` feels a bit too tempting, are you okay rewriting the for-loop like this?

To make sure the functionality didn't change, I wrote a little test (requires src/index.js changes tho to export `packSamples`):

```js
const { data } = pc.packSamples([
    0.1, 0.2, 0.3, 0.4, // row 1
    0.5, 0.6, 0.7, 0.8,
    0.1, 0.2, 0.3, 0.4, // same row 1 for control
]);
console.table([
    [...data.slice( 0, 16)],
    [...data.slice(16, 32)],
    [...data.slice(32, 48)],
]);
data;
```

Output for with and without PR is the same:

image

### DIFF
--- a/src/scene/graphics/reproject-texture.js
+++ b/src/scene/graphics/reproject-texture.js
@@ -68,11 +68,11 @@ const packSamples = (samples) => {
 
     // normalize float data and pack into rgba8
     let off = 0;
-    for (let i = 0; i < numSamples; ++i) {
-        packFloat32ToRGBA8(samples[i * 4 + 0] * 0.5 + 0.5, data, off + 0);
-        packFloat32ToRGBA8(samples[i * 4 + 1] * 0.5 + 0.5, data, off + 4);
-        packFloat32ToRGBA8(samples[i * 4 + 2] * 0.5 + 0.5, data, off + 8);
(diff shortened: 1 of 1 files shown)


slimbuck (approved):
Yep looks good thanks!

willeastcott (approved):

PR #4863, opened 2022-11-18
Clustered lighting uses texFetch instruction on WebGL2 to read data textures

mvaligursky:
- this avoids a multiply per data sampling, and avoids possible uv coordinate rounding issues
- also using dynamic count in the for loop on webgl2 to avoid additional condition to exit the loop
- requiring extension on WebGPU platform to get access to texelFetch

screenshot to see the main clustered loop difference easier (WebGL2 vs WebGL1):
Screenshot 2022-11-18 at 14 40 54

### DIFF
--- a/src/scene/shader-lib/chunks/lit/frag/clusteredLight.js
+++ b/src/scene/shader-lib/chunks/lit/frag/clusteredLight.js
@@ -142,7 +142,7 @@ vec4 decodeClusterLowRange4Vec4(vec4 d0, vec4 d1, vec4 d2, vec4 d3) {
 #ifdef GL2
 
     vec4 sampleLightsTexture8(const ClusterLightData clusterLightData, int index) {
-            return texelFetch(lightsTexture8, ivec2(index, clusterLightData.lightIndex), 0);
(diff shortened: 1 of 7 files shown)


src/scene/shader-lib/chunks/lit/frag/clusteredLight.js
```diff
@@ -135,18 +139,38 @@ vec4 decodeClusterLowRange4Vec4(vec4 d0, vec4 d1, vec4 d2, vec4 d3) {
...
+        return texelFetch(lightsTexture8, ivec2(index, clusterLightData.lightIndex), 0);
+    }
+
+    vec4 sampleLightTextureF(const ClusterLightData clusterLightData, int index) {
+        return texelFetch(lightsTextureFloat, ivec2(index, clusterLightData.lightIndex), 0);
+    }
+
+#else
+
+    uniform vec4 lightsTextureInvSize;
```
willeastcott, line 154: Would it be better to keep the uniforms in one place (at the top) and just #ifdef it there?

mvaligursky: yep done, thanks

GSterbrant (approved):

Issue #4861, opened 2022-11-18
Spotlights are producing 'light' when cones are 0 (clustered lighting)


yaustar:
Forum: https://forum.playcanvas.com/t/light-wont-work/28440
Project: https://playcanvas.com/project/1009817/overview/spotlight-00-bug

<img width="1680" alt="image" src="https://user-images.githubusercontent.com/16639049/202589231-763bcefc-96b7-4019-b24c-b7849533c78c.png">

Both spot lights have the inner cone at 0 and outer cone at 0 so shouldn't be emitting light but in the screenshot above, one light is emitting on the cube.

This only happens with clustered and only with more than 1 light.

If I move any light around, the light will get smaller/disappear if they are closer to each other

https://user-images.githubusercontent.com/16639049/202589863-600a4464-ee00-43aa-991f-154fe8b35b8a.mp4

PR #4864, opened 2022-11-18
Spot lights with zero size outer cone are filtered out from clustered lights

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/4861

It fixes the issue with spot light with 0 outer angle still rendering something in case other spot lights are nearby. When the other spot light moves, the shape gets affected - somehow we access wrong data somewhere / pixels are interpolated that should not. I tried to use textureFetch to avoid it without success so far: https://github.com/playcanvas/engine/pull/4863

### DIFF
--- a/src/scene/lighting/world-clusters.js
+++ b/src/scene/lighting/world-clusters.js
@@ -274,8 +274,8 @@ class WorldClusters {
 
         lights.forEach((light) => {
             const runtimeLight = !!(light.mask & (MASK_AFFECT_DYNAMIC | MASK_AFFECT_LIGHTMAPPED));
-            const nonZeroSpotCone = light.type !== LIGHTTYPE_SPOT || light._outerConeAngle > 0;
-            if (light.enabled && light.type !== LIGHTTYPE_DIRECTIONAL && light.visibleThisFrame && light.intensity > 0 && runtimeLight && nonZeroSpotCone) {
+            const zeroAngleSpotlight = light.type === LIGHTTYPE_SPOT && light._outerConeAngle === 0;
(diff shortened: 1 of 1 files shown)


src/scene/lighting/world-clusters.js
```diff
@@ -274,7 +274,8 @@ class WorldClusters {
 
         lights.forEach((light) => {
             const runtimeLight = !!(light.mask & (MASK_AFFECT_DYNAMIC | MASK_AFFECT_LIGHTMAPPED));
-            if (light.enabled && light.type !== LIGHTTYPE_DIRECTIONAL && light.visibleThisFrame && light.intensity > 0 && runtimeLight) {
+            const nonZeroSpotCone = light.type !== LIGHTTYPE_SPOT || light._outerConeAngle > 0;
+            if (light.enabled && light.type !== LIGHTTYPE_DIRECTIONAL && light.visibleThisFrame && light.intensity > 0 && runtimeLight && nonZeroSpotCone) {
```
willeastcott, line 278:
```suggestion
            const zeroAngleSpotlight = light.type === LIGHTTYPE_SPOT || light._outerConeAngle === 0;
            if (light.enabled && light.type !== LIGHTTYPE_DIRECTIONAL && light.visibleThisFrame && light.intensity > 0 && runtimeLight && !zeroAngleSpotlight ) {
```

willeastcott: Hope you don't mind the above suggestion - just found it easier to read. 😄

willeastcott (approved):

src/scene/lighting/world-clusters.js
mvaligursky: completely agree!

GSterbrant (approved):

PR #4865, opened 2022-11-18
Added collision component linear and angular offset

yaustar:
Fixes https://github.com/playcanvas/engine/issues/698

Full credit to @LeXXik as this is based off their PR https://github.com/playcanvas/engine/pull/3772

I've just cleaned it up a bit and made it work with the latest engine code

Adds the following public API to collision component:

```
    /**
     * @description Returns the world position for the collision shape taking into account of any offsets.
     * @returns {Vec3} The world position for the collision shape.
     */
    getShapePosition() {
```

```
    /**
     * @description Returns the world rotation for the collision shape taking into account of any offsets.
     * @returns {Quat} The world rotation for the collision.
     */
    getShapeRotation() {
        const rot = this.entity.getRotation();

        if (this._hasOffset) {
            return _quat.copy(rot).mul(this.data.angularOffset);
        }

        return rot;
    }
```

Test project with this engine already in the project as a patch: https://playcanvas.com/project/1009973/overview/f-test-offset-physics-lexxik (press space to pause time)

Offsets on the spawned cubes
image

https://user-images.githubusercontent.com/16639049/202735734-98691c73-1086-40b4-ada4-dc847d2d63ff.mp4

Tested against:
- static rigidbodies
- dynamic rigidbodies
- kinematic rigidbodies
- triggers
- compound shapes
- https://playcanvas.com/project/618829/overview/physics-constraints without using offsets to check for changing existing behaviour

Added engine example with a character using an offset collision

https://user-images.githubusercontent.com/16639049/205319422-870f0566-b5eb-4323-8c40-cf1c1e2149c3.mp4


src/framework/components/collision/system.js
```diff
@@ -772,14 +776,23 @@ class CollisionComponentSystem extends ComponentSystem {
...
-
+        const ammoQuat = new Ammo.btQuaternion();
         const transform = new Ammo.btTransform();
+
         transform.setIdentity();
         const origin = transform.getOrigin();
-        origin.setValue(pos.x, pos.y, pos.z);
+        const component = node.collision;
+
+        if (component) {
```
yaustar, line 786: We should always have a component? Is there a case where we don't?

src/framework/components/collision/system.js
```diff
@@ -772,14 +776,28 @@ class CollisionComponentSystem extends ComponentSystem {
...
-
+        const ammoQuat = new Ammo.btQuaternion();
         const transform = new Ammo.btTransform();
+
         transform.setIdentity();
         const origin = transform.getOrigin();
-        origin.setValue(pos.x, pos.y, pos.z);
+        const component = node.collision;
+
+        if (component && component._hasOffset) {
```
yaustar, line 786: Looking back on this, is this unnecessary? We should always have the component right?

LeXXik: I haven't yet looked at the code, but at least here we do need a check on component. A node may come from the collision mesh implementation, where a temporary node is used, which has no collision component.

yaustar:
Checked against mesh collisions with the vehicle project

image

src/framework/components/collision/component.js
```diff
@@ -377,6 +404,41 @@ class CollisionComponent extends Component {
...
+
+        if (this._hasOffset) {
+            const ao = this.data.angularOffset;
+            quat.copy(rot).mul(ao);
+            return quat;
+        }
+
+        return rot;
+    }
+
```
yaustar, line 441: Wondering if these should be public API as they would need to be used by the Editor for visualisation? Or just a comment that they will be used by the Editor be enough when it comes to Editor integration?

mvaligursky: It feels like it could be useful to have these public in general.

src/framework/components/collision/component.js
```diff
@@ -31,8 +37,12 @@ import { Component } from '../component.js';
  * - "sphere": A sphere-shaped collision volume.
  *
  * Defaults to "box".
- * @property {import('../../../core/math/vec3.js').Vec3} halfExtents The half-extents of the
+ * @property {Vec3} halfExtents The half-extents of the
  * box-shaped collision volume in the x, y and z axes. Defaults to [0.5, 0.5, 0.5].
+ * @property {Vec3} linearOffset The positional offset of the collision from the Entity position in local space.
```
mvaligursky, line 42: is it an 'offset of the collision' or 'offset of a collision shape' / volume or similar?

src/framework/components/collision/component.js
```diff
@@ -31,8 +37,12 @@ import { Component } from '../component.js';
...
+ * Defaults to [0, 0, 0].
+ * @property {Quat} angularOffset The rotational offset of the collision from the Entity rotation in local space.
+ * Defaults to [0, 0, 0, 1].
```
mvaligursky, line 45:
```suggestion
 * Defaults to identity.
```

src/framework/components/collision/component.js
```diff
@@ -1,7 +1,13 @@
+import { Quat } from '../../../core/math/quat.js';
+import { Vec3 } from '../../../core/math/vec3.js';
+
 import { Asset } from '../../asset/asset.js';
 
 import { Component } from '../component.js';
 
+const vec3 = new Vec3();
```
mvaligursky, line 8: typically we add _ to these, so when you read the code, you don't think it's a local variable.

src/framework/components/collision/component.js
```diff
@@ -377,6 +404,41 @@ class CollisionComponent extends Component {
...
     }
 
+
+    /**
+     * @returns {Vec3} The world position for the physics body taking into account of any offsets.
+     * @private
+     */
+    _getBodyPosition() {
+        const pos = this.entity.getPosition();
+        const rot = this.entity.getRotation();
```
mvaligursky, line 414: this line is only needed inside the if, move it there.

src/framework/components/collision/component.js
```diff
@@ -377,6 +404,41 @@ class CollisionComponent extends Component {
...
+     * @returns {Quat} The world rotation for the physics body taking into account of any angular offset.
+     * @private
+     */
+    _getBodyRotation() {
+        const rot = this.entity.getRotation();
+
+        if (this._hasOffset) {
+            const ao = this.data.angularOffset;
+            quat.copy(rot).mul(ao);
+            return quat;
```
mvaligursky, line 436:
```suggestion
            return quat.copy(rot).mul(this.data.angularOffset);
```

src/framework/components/collision/system.js
```diff
@@ -189,6 +191,8 @@ class CollisionSystemImpl {
             enabled: src.data.enabled,
             type: src.data.type,
             halfExtents: [src.data.halfExtents.x, src.data.halfExtents.y, src.data.halfExtents.z],
+            linearOffset: [src.data.linearOffset.x, src.data.linearOffset.y, src.data.linearOffset.z],
+            angularOffset: [src.data.angularOffset.x, src.data.angularOffset.y, src.data.angularOffset.z],
```
mvaligursky, line 195: should there be .w as well, or are these just 3 angles here

src/framework/components/collision/trigger.js
```diff
@@ -1,7 +1,7 @@
-let ammoVec1, ammoQuat, ammoTransform;
-
 import { BODYFLAG_NORESPONSE_OBJECT, BODYMASK_NOT_STATIC, BODYGROUP_TRIGGER, BODYSTATE_ACTIVE_TAG, BODYSTATE_DISABLE_SIMULATION } from '../rigid-body/constants.js';
 
+let ammoVec1, ammoQuat, ammoTransform;
```
mvaligursky, line 3: please add _ here as well, _ammoVec1 and so on

src/framework/components/collision/trigger.js
```diff
@@ -43,9 +43,17 @@ class Trigger {
 
             const pos = entity.getPosition();
```
mvaligursky, line 44: these are only used in the else branch, move them there

src/framework/components/collision/trigger.js
```diff
@@ -80,11 +88,20 @@ class Trigger {
     }
 
     _getEntityTransform(transform) {
+        const component = this.component;
+
         const pos = this.entity.getPosition();
```
mvaligursky, line 93: move these to the else branch

src/framework/components/rigid-body/component.js
```diff
@@ -13,6 +13,9 @@ import { Component } from '../component.js';
 // Shared math variable to avoid excessive allocation
 let ammoTransform;
 let ammoVec1, ammoVec2, ammoQuat;
+const quat1 = new Quat();
```
mvaligursky, line 16: underscores please

src/framework/components/rigid-body/component.js
```diff
@@ -854,8 +857,16 @@ class RigidBodyComponent extends Component {
         const pos = entity.getPosition();
```
mvaligursky, line 857: move to else branch

mvaligursky (approved):
approving with multiple small comments

src/framework/components/collision/trigger.js
yaustar: They are also used in the `else` statement too

mvaligursky: they seem to be ONLY used in the else statement?

yaustar: Yep, got my wires crossed 😓

src/framework/components/collision/system.js
yaustar: Nope, that should be 4 as it's a quat

src/framework/components/collision/component.js
yaustar: Shape makes more sense here

src/framework/components/collision/component.js
yaustar: Made public and change name from body to shape to be more inline with the rest of the API and terms

yaustar:
Updated from feedback. There are more changes due to the renaming of temp variables with `_` prefixes.

I've also added public API to collision component based on feedback and it will be used with the Editor visualisation at later date

```
    /**
     * @description Returns the world position for the collision shape taking into account of any offsets.
     * @returns {Vec3} The world position for the collision shape.
     */
    getShapePosition() {
```

```
    /**
     * @description Returns the world rotation for the collision shape taking into account of any offsets.
     * @returns {Quat} The world rotation for the collision.
     */
    getShapeRotation() {
        const rot = this.entity.getRotation();

        if (this._hasOffset) {
            return _quat.copy(rot).mul(this.data.angularOffset);
        }

        return rot;
    }
```

@willeastcott As it adds new public API and you are most familiar with this, can you have give a quick review on the new API please?

MAG-AdrianMeredith:
Talk about good timing, this fixes a ton of mess with our code.  Just tested on our app and it works great so many thanks again everyone.  
One curious thing I noted (I think I'm doing it right) is the option didn't seem to work when passed into the `addComponent('collision', etc)`. it only worked when set on the component after creating it, could just be me but worth double checking...

yaustar:
Good shout. I will put together an engine example as part of this PR and investigate

yaustar:
@mvaligursky @MAG-AdrianMeredith Fixed bug where the offsets weren't being applied when passed as data at the point of component creation: https://github.com/playcanvas/engine/pull/4865/commits/b888f48c7178a336968771f56f14a9b049096197

@mvaligursky Also added example:

https://user-images.githubusercontent.com/16639049/205319422-870f0566-b5eb-4323-8c40-cf1c1e2149c3.mp4

src/framework/components/collision/system.js
```diff
@@ -643,6 +645,14 @@ class CollisionComponentSystem extends ComponentSystem {
             data.halfExtents = new Vec3(data.halfExtents[0], data.halfExtents[1], data.halfExtents[2]);
         }
 
+        if (data.linearOffset && Array.isArray(data.linearOffset)) {
```
mvaligursky, line 648: don't really need this part `data.linearOffset &&` as isArray returns false for undefined or null parameter

src/framework/components/collision/system.js
```diff
@@ -643,6 +645,14 @@ class CollisionComponentSystem extends ComponentSystem {
...
+            data.linearOffset = new Vec3(data.linearOffset[0], data.linearOffset[1], data.linearOffset[2]);
```
mvaligursky, line 649: this could just be 
`new Vec3(data.linearOffset);`

examples/src/examples/physics/offset-collision.tsx
```diff
@@ -0,0 +1,217 @@
...
+                // load the state graph asset resource into the anim component
+                const characterStateLayer = modelEntity.anim.baseLayer;
+                characterStateLayer.assignAnimation('Idle', assets.idleAnim.resource.animations[0].resource);
+
+                app.root.addChild(modelEntity);
+
+                // Create an Entity with a camera component
+                const cameraEntity = new pc.Entity();
+                cameraEntity.addComponent("camera", {
+                    clearColor: new pc.Color(0.1, 0.1, 0.1)
```
mvaligursky, line 156: probably not needed as it uses skydome

mvaligursky:
Looks great, love the example!

mvaligursky (approved):

yaustar:
Per @mvaligursky request: Made the collision a sphere and rendered it by drawing a wire sphere

image

PR #4869, opened 2022-11-21
Multiple Examples updated to work on WebGPU

mvaligursky:
- switched few examples to use EnvAtlas instead of cubemap
- also switched them to use createGraphicsDevice to allow easy way to enable WebGPU 
- updated envMap setter to set up appropriate texture filtering
- sky shader gets processed for Uniform Buffers / Bind Groups

Screenshot 2022-11-21 at 14 19 07


src/scene/scene.js
```diff
@@ -396,6 +396,15 @@ class Scene extends EventHandler {
     set envAtlas(value) {
         if (value !== this._envAtlas) {
             this._envAtlas = value;
+
+            // make sure required options are set up on the texture
+            if (value) {
```
slimbuck, line 401: Actually mipmaps must also be disabled, but this should be done at load time.

mvaligursky: do we have some asset options for that do you know?

slimbuck (approved):

willeastcott:
I notice this PR switched to loading an envAtlas texture. I'm wondering whether it's 'nicer' to have the examples load a pano and reproject it at load time. It then easier for devs to experiment since panos are available all over the web.

mvaligursky:
The reprojection does not work on WebGPU, so I'm switching to this option at the moment. It's also the most optimal solution. But agreed that some examples should use other options, and perhaps even an example showcasing multiple options.

PR #4877, opened 2022-11-23
Run scenes with corrupted data

yaustar:
Related to https://github.com/playcanvas/editor/issues/758

We have existing scenes that could be corrupted by this issue and at the moment, they fail to launch leaving users confused.

This PR allows these scenes to launch with a warning

Tested against:
https://playcanvas.com/project/923035/overview/double-graphnode-parent

Current
image

With this PR
image

### DIFF
--- a/src/framework/parsers/scene.js
+++ b/src/framework/parsers/scene.js
@@ -98,29 +98,28 @@ class SceneParser {
         // Create components in order
         const systemsList = this._app.systems.list;
 
-        if (entity) {
-            let len = systemsList.length;
-            const entityData = entities[entity.getGuid()];
-            for (let i = 0; i < len; i++) {
-                const system = systemsList[i];
-                const componentData = entityData.components[system.id];
-                if (componentData) {
-                    system.addComponent(entity, componentData);
-                }
+        let len = systemsList.length;
+        const entityData = entities[entity.getGuid()];
+        for (let i = 0; i < len; i++) {
+            const system = systemsList[i];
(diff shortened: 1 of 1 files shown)


mvaligursky (approved):

slimbuck:
What can users do when they see this message? Can they fix their scenes?

yaustar:
@slimbuck Currently, the only way they can fix the issue is to do one of the methods here (the version control method is the easiest) https://forum.playcanvas.com/t/cannot-find-child-entity-of-parent-root-editor-error-cant-delete-entities/23048

slimbuck:
We should really be fixing these automatically in the editor at load time since we now know what the issue is.

yaustar:
Should it be like the pop up that we had before around the template issue so that it's a user action to fix?

slimbuck:
> Should it be like the pop up
I think we can just fix it at load time and perhaps show a message saying we have. The reason we don't need user input is we won't be modifying the scene in any way, just getting rid of invalid entities (right?).

yaustar:
In theory, yes. From a user perspective, I would prefer to have the opt-in so I can have the assurance of making a checkpoint, or being able to examine the scene before making changes

slimbuck:
I thought these problem scenes don't load correctly in the editor?

yaustar:
They do but will have Entities that you can't delete (eg https://playcanvas.com/project/1011865/overview/example-broken-project try deleting New Entity). However, you can still edit, add new entities etc.

PR #4886, opened 2022-11-24
Split generic parts of ForwardRenderer into a Renderer class

mvaligursky:
- moving parts of ForwardRenderer into its base Renderer class, to keep the size of the file reasonable, before further render pass related refactoring for shadows rendering
- in the future, this might allow us to write additional renderers (deferred for example), using the same blocks (culling, updates, uniform submission, draw calls ..)
- there were no other updates other then moving code and members, and calling those functions appropriately. Tested on all engine examples.

### DIFF
--- a/src/scene/renderer/forward-renderer.js
+++ b/src/scene/renderer/forward-renderer.js
@@ -4,8 +4,6 @@ import { Debug, DebugHelper } from '../../core/debug.js';
 import { Vec3 } from '../../core/math/vec3.js';
 import { Color } from '../../core/math/color.js';
 
-import { BoundingSphere } from '../../core/shape/bounding-sphere.js';
-
 import {
     FUNC_ALWAYS,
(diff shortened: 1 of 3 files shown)


slimbuck (approved):

src/scene/renderer/forward-renderer.js
```diff
@@ -196,26 +98,12 @@ class ForwardRenderer {
...
     destroy() {
-        this._shadowRenderer.destroy();
-        this._shadowRenderer = null;
-
-        this._cookieRenderer.destroy();
-        this._cookieRenderer = null;
-
-        this.lightTextureAtlas.destroy();
-        this.lightTextureAtlas = null;
+        super.destroy();
```
willeastcott, line 106: Isn't the whole subclassed `destroy` function redundant here?

PR #4898, opened 2022-11-30
Fix sheen glossiness being consistent throughout the engine

GSterbrant:
### Description

Sheen glossiness wasn't applied correctly from the GLB due to an inconsistency with the sheen glossiness material parameter naming.


GSterbrant:
Yep! Let me prepare a PR for that as well.

PR #4899, opened 2022-11-30
Various small improvements and validations to WebGPU rendering

mvaligursky:
- shader processor handles the same uniform coming from vertex and fragment shader by de-duplication
- validation of vertex format support on WebGPU (at some point we might need to add workaround for these when loading glbs with the format)
- grab texture sampling temporary fix, might need to create some wrapper function for this in the future

And this now works

https://user-images.githubusercontent.com/59932779/204857225-eac5b6a1-49be-4b58-a0c4-d06cef932f08.mov


slimbuck:
wow this awesome!

think I can see twinkles - mipmaps not yet supported?

slimbuck (approved):

mvaligursky:
Yep, no mipmaps .. WebGPU does not have generateMipmaps function, and that needs to be done by rendering down the mipmaps chain. Leaving that for later.

PR #4900, opened 2022-11-30
LitOptions class to hold backend options

### LINKED ISSUE #4871: Add LitOptions to Standard Material
### Description

With the splitting of the options between frontend and backend (#4792), the backend options are stowed away in an anonymous object in the options class. Because it is anonymous, it's difficult to figure out what options can be provided without digging into the engine code, so it would add a lot of value to have a class setup for these settings.

Furthermore, we also need to deprecate any parameters that were moved from the frontend options to the backend options, so that users will be informed that these are deprecated and will have no effect on the shading.

GSterbrant:
### Description

Fixes #4871. 

The backend options for the lit shader are now defined in their own class, which makes it possible to know what options are available. 

Additionally, this allows us to validate options which might have moved to the backend object, and inform developers if they are still expecting the option to be there. This primarily pertains to options set through the onUpdateShader callback.

### DIFF
--- a/src/deprecated/deprecated.js
+++ b/src/deprecated/deprecated.js
@@ -788,7 +788,7 @@ _defineAlias('opacityVertexColor', 'opacityMapVertexColor');
 _defineAlias('lightVertexColor', 'lightMapVertexColor');
 
 function _defineOption(name, newName) {
-    if (name !== 'chunks') {
+    if (name !== 'chunks' || name == 'pass') {
         Object.defineProperty(StandardMaterialOptions.prototype, name, {
             get: function () {
                 Debug.deprecated(`Getting pc.Options#${name} has been deprecated as the property has been moved to pc.Options.LitOptions#${newName || name}.`);
--- a/src/deprecated/deprecated.js
+++ b/src/deprecated/deprecated.js
@@ -71,7 +71,7 @@ import { SkinInstance } from '../scene/skin-instance.js';
 import { StandardMaterial } from '../scene/materials/standard-material.js';
 import { Batch } from '../scene/batching/batch.js';
 import { getDefaultMaterial } from '../scene/materials/default-material.js';
-import { Options } from '../scene/materials/options.js';
+import { StandardMaterialOptions } from '../scene/materials/standard-material-options.js';
 import { LitOptions } from '../scene/materials/lit-options.js';
 
 import { Animation, Key, Node } from '../scene/animation/animation.js';
@@ -787,17 +787,19 @@ _defineAlias('glossVertexColor', 'glossMapVertexColor');
 _defineAlias('opacityVertexColor', 'opacityMapVertexColor');
 _defineAlias('lightVertexColor', 'lightMapVertexColor');
 
-function _defineOption(name) {
-    Object.defineProperty(Options.prototype, name, {
+function _defineOption(name, newName) {
+    Object.defineProperty(StandardMaterialOptions.prototype, name, {
         get: function () {
-            Debug.deprecated(`Getting pc.Options#${name} has been deprecated as the property has been moved to pc.Options.LitOptions.`);
-            return null;
+            Debug.deprecated(`Getting pc.Options#${name} has been deprecated as the property has been moved to pc.Options.LitOptions#${newName || name}.`);
+            return this.litOptions[newName || name];
         },
         set: function (value) {
-            Debug.deprecated(`Setting pc.Options#${name} has been deprecated as the property has been moved to pc.Options.LitOptions.`);
+            Debug.deprecated(`Setting pc.Options#${name} has been deprecated as the property has been moved to pc.Options.LitOptions#${newName || name}.`);
+            this.litOptions[newName || name] = value;
         }
     });
 }
+_defineOption('refraction', 'useRefraction');
 
 const tempOptions = new LitOptions();
 const litOptionProperties = Object.getOwnPropertyNames(tempOptions);
--- a/src/deprecated/deprecated.js
+++ b/src/deprecated/deprecated.js
@@ -788,16 +788,18 @@ _defineAlias('opacityVertexColor', 'opacityMapVertexColor');
 _defineAlias('lightVertexColor', 'lightMapVertexColor');
 
 function _defineOption(name, newName) {
-    Object.defineProperty(StandardMaterialOptions.prototype, name, {
-        get: function () {
-            Debug.deprecated(`Getting pc.Options#${name} has been deprecated as the property has been moved to pc.Options.LitOptions#${newName || name}.`);
-            return this.litOptions[newName || name];
-        },
-        set: function (value) {
-            Debug.deprecated(`Setting pc.Options#${name} has been deprecated as the property has been moved to pc.Options.LitOptions#${newName || name}.`);
-            this.litOptions[newName || name] = value;
-        }
-    });
+    if (name !== 'chunks') {
+        Object.defineProperty(StandardMaterialOptions.prototype, name, {
+            get: function () {
+                Debug.deprecated(`Getting pc.Options#${name} has been deprecated as the property has been moved to pc.Options.LitOptions#${newName || name}.`);
+                return this.litOptions[newName || name];
+            },
+            set: function (value) {
+                Debug.deprecated(`Setting pc.Options#${name} has been deprecated as the property has been moved to pc.Options.LitOptions#${newName || name}.`);
+                this.litOptions[newName || name] = value;
+            }
+        });
+    }
 }
 _defineOption('refraction', 'useRefraction');
 
--- a/src/deprecated/deprecated.js
+++ b/src/deprecated/deprecated.js
@@ -788,7 +788,7 @@ _defineAlias('opacityVertexColor', 'opacityMapVertexColor');
 _defineAlias('lightVertexColor', 'lightMapVertexColor');
 
 function _defineOption(name, newName) {
-    if (name !== 'chunks' || name == 'pass') {
+    if (name !== 'chunks' || name !== 'pass') {
         Object.defineProperty(StandardMaterialOptions.prototype, name, {
             get: function () {
                 Debug.deprecated(`Getting pc.Options#${name} has been deprecated as the property has been moved to pc.Options.LitOptions#${newName || name}.`);
--- a/src/deprecated/deprecated.js
+++ b/src/deprecated/deprecated.js
@@ -788,7 +788,7 @@ _defineAlias('opacityVertexColor', 'opacityMapVertexColor');
 _defineAlias('lightVertexColor', 'lightMapVertexColor');
 
 function _defineOption(name, newName) {
-    if (name !== 'chunks' || name !== 'pass') {
+    if (name !== 'chunks' && name !== 'pass') {
         Object.defineProperty(StandardMaterialOptions.prototype, name, {
             get: function () {
                 Debug.deprecated(`Getting pc.Options#${name} has been deprecated as the property has been moved to pc.Options.LitOptions#${newName || name}.`);
--- a/src/deprecated/deprecated.js
+++ b/src/deprecated/deprecated.js
@@ -788,7 +788,7 @@ _defineAlias('opacityVertexColor', 'opacityMapVertexColor');
 _defineAlias('lightVertexColor', 'lightMapVertexColor');
 
 function _defineOption(name, newName) {
-    if (name !== 'chunks' && name !== 'pass') {
+    if (name !== 'chunks' && name !== '_pass') {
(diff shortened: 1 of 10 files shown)


mvaligursky:
The docs for these should probably move to appropriate classes too
https://github.com/playcanvas/engine/blob/636571ee71f5257ec6a38f717e765401e2de374b/src/scene/materials/standard-material.js#L495-L564

mvaligursky:
And also, if the option has moved to litOptions, we should generate some deprecated properties that warn user the option has moved.

GSterbrant:
I added an options class which is used to hookup deprecation callbacks, and replaced the use of an anonymous options object with this new (albeit empty) one.

src/deprecated/deprecated.js
```diff
@@ -785,6 +787,24 @@ _defineAlias('glossVertexColor', 'glossMapVertexColor');
 _defineAlias('opacityVertexColor', 'opacityMapVertexColor');
 _defineAlias('lightVertexColor', 'lightMapVertexColor');
 
+function _defineOption(name) {
+    Object.defineProperty(Options.prototype, name, {
+        get: function () {
+            Debug.deprecated(`Getting pc.Options#${name} has been deprecated as the property has been moved to pc.Options.LitOptions.`);
+            return null;
```
mvaligursky, line 794: could this return `this.LitOptions[name];` to be backwards compatible?

src/scene/materials/lit-options.js
```diff
@@ -0,0 +1,102 @@
+import {
+    BLEND_NONE, GAMMA_NONE
+} from '../constants.js';
+
+class LitOptions {
```
mvaligursky, line 5: we should add the documentation to some of those properties that is still stored in the old place .. see my other comment on this PR.

src/scene/materials/options.js
```diff
@@ -0,0 +1,7 @@
+class Options {
```
mvaligursky, line 1: we need to give it a less generic name than `Options`. Maybe `StandardOptions`?

src/scene/shader-lib/program-library.js
```diff
@@ -95,7 +95,7 @@ class ProgramLibrary {
 
             const device = this._device;
             def = generator.createShaderDefinition(device, options);
-            def.name = `${name}-pass:${options.pass}`;
+            def.name = `${name}-pass:${options.litOptions?.pass}`;
```
mvaligursky, line 98: I think .pass should stay on options as well to avoid changes in this file, as it needs to be compatible with other generators (for example basic.js) which has uses options.pass.

GSterbrant: The shader backend needs access to the pass, the front end doesn't. The backend only has the backend options, so in that case we need to duplicate the pass. How do we solve that?

src/scene/materials/lit-options.js
GSterbrant: Do you mean the properties stored in the ordinary options object?

src/scene/shader-lib/program-library.js
mvaligursky: yeah duplication should be fine I think.

src/scene/materials/options.js
GSterbrant: Done

src/deprecated/deprecated.js
```diff
@@ -785,6 +787,24 @@ _defineAlias('glossVertexColor', 'glossMapVertexColor');
...
+        },
+        set: function (value) {
+            Debug.deprecated(`Setting pc.Options#${name} has been deprecated as the property has been moved to pc.Options.LitOptions.`);
```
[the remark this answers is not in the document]
GSterbrant: Yeah, good catch, I've added that.

src/scene/materials/lit-options.js
```diff
@@ -0,0 +1,147 @@
+import {
+    BLEND_NONE, GAMMA_NONE
+} from '../constants.js';
+
+/**
+ * - pass: value of {@link Layer#shaderPass} of the Layer being rendered. Must be set to the
+ * same in {@link StandardMaterialOptions#pass}.
```
mvaligursky, line 7: Could we only have `pass` on StandardMaterialOptions, and the setter would also set it on LitOptions? Better than having to keep two in sync

src/scene/materials/lit-options.js
```diff
@@ -0,0 +1,147 @@
...
+ * for the list of possible values.
+ * - conserveEnergy: the value of {@link StandardMaterial#conserveEnergy}.
+ * - occludeSpecular: the value of {@link StandardMaterial#occludeSpecular}.
+ * - occludeDirect: the value of {@link StandardMaterial#occludeDirect}.
+ * - shadingModel: the value of {@link StandardMaterial#shadingModel}.
+ * - fresnelModel: the value of {@link StandardMaterial#fresnelModel}.
+ * - cubeMapProjection: the value of {@link StandardMaterial#cubeMapProjection}.
+ * - useMetalness: the value of {@link StandardMaterial#useMetalness}.
+ * - blendType: the value of {@link Material#blendType}.
+ * - twoSidedLighting: the value of {@link Material#twoSidedLighting}.
```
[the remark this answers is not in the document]
mvaligursky: that would make them useable in typescript as well, as they would be correctly types

src/scene/materials/standard-material-options.js
```diff
@@ -0,0 +1,44 @@
+import { LitOptions } from "./lit-options.js";
+
+/**
+ * - pass: value of {@link Layer#shaderPass} of the Layer being rendered. Must be set to the
```
mvaligursky, line 4: same here, document as properties.

src/scene/materials/standard-material.js
```diff
@@ -498,70 +498,10 @@ let _params = new Set();
...
- * - useCubeMapRotation: if cube map rotation is enabled.
- * - useInstancing: if hardware instancing compatible shader should be generated. Transform is read
- * from per-instance {@link VertexBuffer} instead of shader's uniforms.
- * - useMorphPosition: if morphing code should be generated to morph positions.
- * - useMorphNormal: if morphing code should be generated to morph normals.
- * - reflectionSource: one of "envAtlasHQ", "envAtlas", "cubeMap", "sphereMap"
- * - reflectionEncoding: one of null, "rgbm", "rgbe", "linear", "srgb"
- * - ambientSource: one of "ambientSH", "envAtlas", "constant"
- * - ambientEncoding: one of null, "rgbm", "rgbe", "linear", "srgb"
+ * assigned to the material, a reflection pass with simpler shaders and so on. These properties are
```
mvaligursky, line 501: on top of this class is also this code

Screenshot 2022-12-07 at 17 45 26

we should replace both `*` by the options type.

src/scene/materials/lit-options.js
```diff
@@ -0,0 +1,153 @@
...
+ * @property {boolean} useMorphNormal If morphing code should be generated to morph normals.
+ * @property {string} reflectionSource One of "envAtlasHQ", "envAtlas", "cubeMap", "sphereMap"
+ * @property {boolean} ambientSource One of "ambientSH", "envAtlas", "constant"
+ */
+class LitOptions {
+    constructor() {
+        this.hasTangents = false;
+        this.chunks = [];
+
+        this.pass = 0;
```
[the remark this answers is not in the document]
mvaligursky: I think you need variable this._pass
getter would return it
setter would complain
and from outside, you'd do options._pass = ... to actually assign to it

src/scene/materials/lit-options.js
```diff
@@ -0,0 +1,153 @@
...
+ * @property {number} cubeMapProjection The value of {@link StandardMaterial#cubeMapProjection}.
+ * @property {boolean} useMetalness The value of {@link StandardMaterial#useMetalness}.
+ * @property {number} blendType The value of {@link Material#blendType}.
+ * @property {boolean} twoSidedLighting The value of {@link Material#twoSidedLighting}.
+ * @property {number} occludeSpecularFloat Defines if {@link StandardMaterial#occludeSpecularIntensity} constant
+ * should affect specular occlusion.
+ * @property {boolean} alphaTest Enable alpha testing. See {@link Material#alphaTest}.
+ * @property {boolean} alphaToCoverage Enable alpha to coverage. See {@link Material#alphaToCoverage}.
+ * @property {boolean} opacityFadesSpecular Enable specular fade. See {@link Material#opacityFadesSpecular}.
+ * @property {float} ambientSH If ambient spherical harmonics are used. Ambient SH replace prefiltered cubemap
```
mvaligursky, line 29: some missing full stops at the end of the sentences.

src/scene/materials/lit-options.js
```diff
@@ -0,0 +1,153 @@
...
+        this.enableGGXSpecular = false;
+        this.fresnelModel = 0;
+        this.useRefraction = false;
+        this.useClearCoat = false;
+        this.useSheen = false;
+        this.useIridescence = false;
+        this.useMetalness = false;
+        this.useDynamicRefraction = false;
+
+        this.fog = 'none';
```
mvaligursky, line 122: is there some constant for this 'none'?

src/scene/materials/standard-material-options.js
```diff
@@ -0,0 +1,51 @@
...
+        this.specularTint = false;
+        this.metalnessTint = false;
+        this.glossTint = false;
+        this.emissiveTint = false;
+        this.opacityTint = false;
+        this.emissiveEncoding = 'linear';
+        this.lightMapEncoding = 'linear';
+        this.packedNormal = false;
+
+        this.litOptions = new LitOptions();
```
mvaligursky, line 38: this needs JSdoc so that it's exposed to documentation.
Please build the API documentation locally to make sure it all builds correctly and is linked up from StandardMaterial

src/scene/materials/standard-material-options.js
```diff
@@ -0,0 +1,51 @@
...
+ * @property {boolean} emissiveTint Defines if {@link StandardMaterial#emissive} constant should affect emission
+ * value.
+ * @property {boolean} opacityTint Defines if {@link StandardMaterial#opacity} constant should affect opacity value.
+ * @property {boolean} packedNormal If normal map contains X in RGB, Y in Alpha, and Z must be reconstructed.
+ *
+ */
+class StandardMaterialOptions {
+    constructor() {
+        this.chunks = [];
+        this.pass = 0;
```
mvaligursky, line 25: this needs to be _pass as well, as we have setter / getter.

src/scene/materials/lit-options.js
```diff
@@ -0,0 +1,153 @@
...
+        this.gamma = GAMMA_NONE;
+        this.toneMap = -1;
+        this.fixSeams = false;
+
+        this.reflectionSource = null;
+        this.reflectionEncoding = null;
+        this.ambientSource = 'constant';
```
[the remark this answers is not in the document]
GSterbrant: Only string values afaik, that's how it's done in the standard material options builder.

src/scene/materials/lit-options.js
GSterbrant: Yes! Good catch.

src/scene/materials/standard-material.js
mvaligursky: This still needs to be done too.

src/scene/materials/lit-options.js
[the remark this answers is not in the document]
GSterbrant: Fixed

src/scene/materials/lit-options.js
```diff
@@ -0,0 +1,153 @@
...
+        this.noShadow = false;
+        this.lightMaskDynamic = 0x0;
+    }
+
+    set pass(p) {
+        Debug.warn(`pc.LitOptions#pass should be set by its parent pc.StandardMaterialOptions, setting it directly has no effect.`);
+    }
+
+    get pass() {
+        Debug.warn(`pc.LitOptions#pass should be accessed by its parent pc.StandardMaterialOptions.`);
```
mvaligursky, line 148: this should be no warning and return this._pass
otherwise code using litOptions.pass always gets 0

GSterbrant: Yeah good point, this would break the lit-shader for sure.

src/scene/materials/standard-material.js
GSterbrant: Yep, fixed.

src/scene/materials/lit-options.js
GSterbrant: Fixed.

mvaligursky (approved):
Looks good to me. Please do test it with at least engine examples to make sure all is happy.

PR #4917, opened 2022-12-09
Directional shadows are rendered using RenderPass setup

mvaligursky:
- refactored shadows rendering to be more modular
- based on this, directional shadow rendering has been refactored to utilize render pass, so that it can be used on WebGPU
- improved logging
- render pass has an `after` function which executes after the render pass
- shadow cascades now use single full render target clear, instead of clearing each cascade by viewport clear - that could speed up things
- tested all combinations of Webgl1 & 2 and clustered / non-clustered shadows + shadow cascades


slimbuck (commented):
nice

slimbuck (approved):

Issue #4910, opened 2022-12-05
API to make registering events in scriptType easier


yaustar:
A common issue with developers new to PlayCanvas is ensuring that event listeners to objects like the mouse are cleaned up properly when the scriptType is destroyed (eg destroying the Entity or changing scenes)

Users either don't know and/or assume it's done automatically. And to do it correctly requires some boilerplate. eg:

```
// initialize code called once per entity
BoxPlacement.prototype.initialize = function () {
    if (this.app.touch) {
        this.app.touch.on('touchstart', this.onTouchStart, this);
    } 
    
    this.app.mouse.on('mousedown', this.onMouseDown, this);

    this.on('destroy', function () {
        if (this.app.touch) {
            this.app.touch.off('touchstart', this.onTouchStart, this);
        } 

        this.app.mouse.off('mousedown', this.onMouseDown, this);       
    }, this);
};
```

Could we introduce some API to make this easier/automatic? Such as a function that automatically removes the listener when the scriptType is destroyed? It would turn the above code into:

```
BoxPlacement.prototype.initialize = function () {
    if (this.app.touch) {
        this.listen(this.app.touch, 'touchstart', this.onTouchStart, this);
    } 
    
    this.listen(this.app.mouse, 'mousedown', this.onMouseDown, this);
};
```

And would also work with anonymous functions too

Maksims:
If event methods such as on/once would return event handler instead of `self`, it would be easier to manage events also. Similar to the way it is implemented in the Editor.

yaustar:
So it would be more like?
```
    this.mouseEventHandler = this.app.mouse.on('mousedown', this.onMouseDown, this);

    this.on('destroy', function () {
        this.mouseEventHandler.off();
    }, this);
```

querielo:
Maybe, something like this:
```js
this.handlers.mouseDown = this.app.mouse.on('mousedown', this.onMouseDown, this);
```

And execute the next code on the side of the engine while destroying a component instance:

```js
for (let i = 0; i < this.handlers.length; i++) {
  const handler = this.handlers[I]
  if (handler.off === "function") {
    handler.off()
  }
}
```

So, a user can forget about destroying and do just the first line.

kungfooman:
To be even more pedantic, a "good" script should look at least somewhat like this:

```js
class DebugGamepadFlyCamera extends pc.ScriptType {
    // ...
    initialize() {
        this.hookEvents();
        this.on('destroy', this.onDestroy   , this);
[...]

PR #4920, opened 2022-12-10
Add/implement pc.EVENT_MOUSEOUT

kungfooman:
This event is missing, so devs always had to add custom event handling in their scripts, like here:

https://github.com/playcanvas/engine/blob/b07a438362a7a7e8ea4c9353dafecc808640556a/scripts/camera/orbit-camera.js#L378-L388

This will also simplify event handling later on based on https://github.com/playcanvas/engine/issues/4910, in short `ScriptType#listen` provides:

 - Less code
 - Automatically removing events on `disable`/`destroy`
 - Automatically adding them back on `enable`
 - Harder to get it wrong

### DIFF
--- a/src/platform/input/constants.js
+++ b/src/platform/input/constants.js
@@ -53,6 +53,13 @@ export const EVENT_MOUSEUP = 'mouseup';
 export const EVENT_MOUSEWHEEL = 'mousewheel';
 
 /**
+ * Name of event fired when the mouse moves out or enters another element.
+ *
+ * @type {string}
+ */
+export const EVENT_MOUSEOUT = 'mouseout';
+
+/**
  * Name of event fired when a new touch occurs. For example, a finger is placed on the device.
  *
  * @type {string}
--- a/src/platform/input/mouse-event.js
+++ b/src/platform/input/mouse-event.js
@@ -50,7 +50,7 @@ class MouseEvent {
              * @type {number}
              */
             this.y = coords.y;
-        } else if (isMousePointerLocked()) {
+        } else if (isMousePointerLocked() || event.type === 'mouseout') {
             this.x = 0;
             this.y = 0;
         } else {
--- a/src/platform/input/mouse.js
+++ b/src/platform/input/mouse.js
@@ -1,7 +1,7 @@
 import { platform } from '../../core/platform.js';
 import { EventHandler } from '../../core/event-handler.js';
 
-import { EVENT_MOUSEDOWN, EVENT_MOUSEMOVE, EVENT_MOUSEUP, EVENT_MOUSEWHEEL } from './constants.js';
+import { EVENT_MOUSEDOWN, EVENT_MOUSEMOVE, EVENT_MOUSEOUT, EVENT_MOUSEUP, EVENT_MOUSEWHEEL } from './constants.js';
 import { isMousePointerLocked, MouseEvent } from './mouse-event.js';
 
 /**
@@ -36,6 +36,7 @@ class Mouse extends EventHandler {
         this._downHandler = this._handleDown.bind(this);
         this._moveHandler = this._handleMove.bind(this);
         this._wheelHandler = this._handleWheel.bind(this);
+        this._outHandler = this._handleOut.bind(this);
         this._contextMenuHandler = (event) => {
             event.preventDefault();
         };
@@ -98,6 +99,7 @@ class Mouse extends EventHandler {
         window.addEventListener('mouseup', this._upHandler, opts);
         window.addEventListener('mousedown', this._downHandler, opts);
         window.addEventListener('mousemove', this._moveHandler, opts);
+        window.addEventListener('mouseout', this._outHandler, opts);
         window.addEventListener('wheel', this._wheelHandler, opts);
     }
 
@@ -113,6 +115,7 @@ class Mouse extends EventHandler {
         window.removeEventListener('mouseup', this._upHandler, opts);
         window.removeEventListener('mousedown', this._downHandler, opts);
         window.removeEventListener('mousemove', this._moveHandler, opts);
+        window.removeEventListener('mouseout', this._outHandler, opts);
         window.removeEventListener('wheel', this._wheelHandler, opts);
     }
 
@@ -287,6 +290,13 @@ class Mouse extends EventHandler {
         if (!e.event) return;
 
         this.fire(EVENT_MOUSEWHEEL, e);
+    }
+
+    _handleOut(event) {
+        const e = new MouseEvent(this, event);
+        if (!e.event) return;
+
+        this.fire(EVENT_MOUSEOUT, e);
     }
 
     _getTargetCoords(event) {


src/platform/input/mouse-event.js
```diff
@@ -50,7 +50,7 @@ class MouseEvent {
              * @type {number}
              */
             this.y = coords.y;
-        } else if (isMousePointerLocked()) {
+        } else if (isMousePointerLocked() || event.type === 'mouseout') {
```
slimbuck, line 53: why is this necessary?

yaustar: On the same note, what values are coords.x/y on a mouse out event?

src/platform/input/constants.js
```diff
@@ -52,6 +52,13 @@ export const EVENT_MOUSEUP = 'mouseup';
  */
 export const EVENT_MOUSEWHEEL = 'mousewheel';
 
+/**
+ * Name of event fired when the mouse moves out or enters another element.
```
yaustar, line 56:
```suggestion
 * Name of event fired when the mouse moves out or enters another DOM element.
```
Making it clear that we aren't talking about pc Elements

src/platform/input/mouse-event.js
kungfooman: > why is this necessary?

`coords = mouse._getTargetCoords(event);` returns null and then the constructor just ends here:

https://github.com/playcanvas/engine/blob/7c582ab58e69ae21291771b586910a9e52bf4b62/src/platform/input/mouse-event.js#L56-L58

Resulting in the entire event being "uninitialized".

yaustar: Ah, I see 🤔 This seems fine to me

kungfooman:
> On the other, it would be nice to have mouse in and mouse out instead of both being on the same event?

Yea, I share your concerns here, seems like this works best:

```js
document.body.onmouseout = console.log;
```

It isn't confused between BODY/HTML, but what I don't like is that it's accessing `body` which can be undefined (coming from a `<script>` in `<head>` e.g.)

yaustar:
```
document.body.onmouseout = console.log;
```

Yeah, let's not do that as the mouse DOM element can be any DOM element too on `attach`.

Wait, is target element not used in the mouse input? 👀 We just listen the window regardless?

kungfooman:
> Wait, is target element not used in the mouse input? :eyes: We just listen the window regardless?

Right, `_target` is just used for some context menu handling and position calculations, the rest attaches events onto window:

https://github.com/playcanvas/engine/blob/c3d5ff7ca2103490f463c5a24703617493b32e57/src/platform/input/mouse.js#L91-L102

yaustar:
Can we use it to know if we are 'mousing out' or 'in'?

kungfooman:
> Can we use it to know if we are 'mousing out' or 'in'?

You mean to have e.g. a small 100x100px canvas and then trigger `mouseout` when this small area is left and not the window?

It should be possible, just that it would deviate from how e.g. the OrbitCamera script used to use it (if we speak about the same thing here).

I tested `window.onmouseenter` but that doesn't fires at all for me. Having a `pc.EVENT_MOUSEIN` should still be quite easy using the mouse move events tho (or `_target` of `pc.app.mouse`)

kungfooman:
Yea, it seems more sensible to have it on `Mouse#_target` as it will behave less surprisingly.

(I can rewrite it tomorrow if no one comes up with a counter-argument :sweat_smile:)

yaustar:
@kungfooman Is this ready for a re-review?

kungfooman:
> @kungfooman Is this ready for a re-review?

Yep, I would like that, quick intro:

```js
pc.app.mouse.on(pc.EVENT_MOUSEENTER, console.log);
pc.app.mouse.on(pc.EVENT_MOUSEOUT  , console.log);
```

yaustar:
If the window is not in focus, when the mouse enters, it fires both the ENTER and OUT events (see end of the video). Was this intended?

https://user-images.githubusercontent.com/16639049/209349691-b30fcc81-4ae5-454e-b941-5a853dd48665.mp4

kungfooman:
> If the window is not in focus, when the mouse enters, it fires both the ENTER and OUT events (see end of the video). Was this intended?

Thank you for testing! It might be a Chrome bug on Mac, I can't reproduce on Linux/Chrome:

Peek 2022-12-24 11-36

yaustar:
Definitely Chrome specific but it's difficult to ship this when it 'doesn't work' with the major browser user share 😅

LeXXik:
Yeah, I also remember having issues with `onmouseout` in the past. At least on Windows. Have you tried `onmouseleave`?

kungfooman:
I reported mouse issues in the past and it usually takes some time until issues are fixed, but imo that is the way to go - just someone needs to create a Chrome issue and keep up with it.

Issue 356090: mouseout on element triggered on mouseup after entering the element with the mouse button pressed
 -  https://bugs.chromium.org/p/chromium/issues/detail?id=356090

Sounds like they had it before, apparently "fixed", but not quite it seems.

For my sake it works good enough besides this have-another-window-over-canvas and I'm not trying to find Chrome specific hacks for that case.

> Edit: Wondering if we should only fire the event if the window is in focus?

Just for testing:

```js
setInterval(()=>console.log(document.hasFocus()), 1000)
```

But I don't see how that would work out nicely, as soon you click outside the tab you won't have any "entering" events anymore?

yaustar:
> But I don't see how that would work out nicely, as soon you click outside the tab you won't have any "entering" events anymore?

There aren't any mousemove events on the app if the tab isn't in focus either so I would argue it's consistent behaviour?

kungfooman:
> There aren't any mousemove events on the app if the tab isn't in focus either so I would argue it's consistent behaviour?

The mousemove events are always fired, focused or not:

```js
pc.app.mouse.on("mousemove", console.log);
```

E.g. for devtools in https://playcanvas.com/viewer if you want to test. Or do you mean something different?

yaustar:
Hmm, I get different behaviour on Chrome/Mac
https://user-images.githubusercontent.com/16639049/210997265-996da4bb-c4b2-45cc-b4cd-302553a1b154.mp4

kungfooman:
Funny how different the same browser is on different platforms:

https://user-images.githubusercontent.com/5236548/211007086-27078234-4184-4ecf-bbf2-016fc7620bfd.mp4

IIRC Chrome on Windows also sends mousemove even though the window is unfocused, so Mac is the oddball here.

Technically it imo makes more sense to not filter events too early, since once they are filtered, there is no way back. And for a dev it would be simple enough to just `if (!document.hasFocus()) return;` if they want that?

yaustar:
> Technically it imo makes more sense to not filter events too early, since once they are filtered, there is no way back. And for a dev it would be simple enough to just if (!document.hasFocus()) return; if they want that?

Part of me feels like the engine should be as consistent as possible in the events it fires. If a developer wants the raw events, then there's nothing stopping them registering events on the DOM

kungfooman:
> Part of me feels like the engine should be as consistent as possible in the events it fires.

So mousemove shall also be filtered now to make everything work like on a Chrome Mac?

yaustar:
> So mousemove shall also be filtered now to make everything work like on a Chrome Mac?

Ideally, it should be consistent otherwise we would end up with bug reports in the engine.

The fact that mouse move is different on a per OS basis is frustrating and truthfully, I don't know what the best way forward here is. 

We could ship this as is where the behaviour is consistent across OS's browser when the tab is in focus by the user. I think that would be okay given noone has reported about mouse move being an issue when the browser is not in focus.

yaustar:
> Thank you for testing! It might be a Chrome bug on Mac, I can't reproduce on Linux/Chrome:

It works fine if the active window is not another Chrome window. That's probably why you can't reproduce it as you are checking against VSCode

src/platform/input/constants.js
```diff
@@ -52,6 +52,20 @@ export const EVENT_MOUSEUP = 'mouseup';
  */
 export const EVENT_MOUSEWHEEL = 'mousewheel';
 
+/**
+ * Name of event fired when the mouse moves out or enters another DOM element.
```
yaustar, line 56:
```suggestion
 * Name of event fired when the mouse moves out or enters another DOM element. This can fire on entering a defocused PlayCanvas browser window so you may want to check if it is focused with {@link https://developer.mozilla.org/en-US/docs/Web/API/Document/hasFocus document.hasFocus()}
```

yaustar (approved):
Approved with comment

yaustar (approved):
Approved with comment

src/platform/input/mouse.js
```diff
@@ -107,13 +119,17 @@ class Mouse extends EventHandler {
...
         this._attached = false;
-        this._target = null;
 
         const opts = platform.passiveEvents ? { passive: false } : false;
         window.removeEventListener('mouseup', this._upHandler, opts);
         window.removeEventListener('mousedown', this._downHandler, opts);
         window.removeEventListener('mousemove', this._moveHandler, opts);
         window.removeEventListener('wheel', this._wheelHandler, opts);
+        if (this._target) {
+            this._target.removeEventListener('mousein', this._enterHandler, opts);
```
willeastcott, line 129: Shouldn't this be `mouseenter`?

src/platform/input/constants.js
yaustar: maybe 'background' is a better term?

yaustar: Updated suggested change with something with less jargon

kungfooman: Just changed it via normal commit, GitHub is having some strange issues:

image

(for my sake it sounds good enough, but you can make a suggestion if needed)

kungfooman:
> It works fine if the active window is not another Chrome window.

Did you test that on Linux? I tried with another tab of Chrome and I couldn't trigger the bug either. Also tried it with other apps like KolourPaint and it worked nicely.

So far I believe we could just point out it only happens on Chrome/Mac, which should be fixed in Chrome/Mac's OS code (until someone brings this issue to some Mac developer attention).

yaustar:
> Did you test that on Linux? I tried with another tab of Chrome and I couldn't trigger the bug either. Also tried it with other apps like KolourPaint and it worked nicely.

On have access to Mac unfortunately 😅  Either way, I don't think it's a blocker for the PR

mvaligursky:
@kungfooman - any thoughts on the last comments here?

kungfooman:
Yea, it's kinda confusing, because `mouseout` triggers when entering other elements:

mouseenter mouseout

However, good catch @willeastcott, because the `EVENT_MOUSEENTER` description should be more specific.

I will rewrite it to make it clearer, also happy about suggestions.

kungfooman:
Looking like this in VSCode:

image

PR #4924, opened 2022-12-14
WebGPU support for rendering to shadow map for Directional lights

mvaligursky:
WebGPU platform can render to a shadow map for directional light. The shadow map cannot be still used to sample from, that would be done in a follow up PR.

Part of the changes is refactoring of ShadowRendererLocal and ShadowRendererDirectional - those no longer extend the ShadowRenderer, but share a single instance of it, to limit some resource duplication.

Screenshot 2022-12-14 at 11 32 20

### DIFF
--- a/src/scene/light.js
+++ b/src/scene/light.js
@@ -75,6 +75,19 @@ class LightRenderData {
 
         // visible shadow casters
         this.visibleCasters = [];
+
+        // an array of view bind groups, single entry is used for shadows
+        /** @type {import('../platform/graphics/bind-group.js').BindGroup[]} */
+        this.viewBindGroups = [];
+    }
+
+    // releases GPU resources
+    destroy() {
+        this.viewBindGroups.forEach((bg) => {
+            bg.defaultUniformBuffer.destroy();
+            bg.destroy();
+        });
+        this.viewBindGroups.length = 0;
     }
 
     // returns shadow buffer currently attached to the shadow camera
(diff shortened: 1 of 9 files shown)


src/scene/light.js
```diff
@@ -75,6 +75,19 @@ class LightRenderData {
...
+        this.viewBindGroups = [];
+    }
+
+    // releases GPU resources
+    destroy() {
+        this.viewBindGroups.forEach((bg) => {
+            bg.defaultUniformBuffer.destroy();
+            bg.destroy();
+        });
+        this.viewBindGroups.length = 0;
```
kungfooman, line 90: Wouldn't it make sense to delete `bg.defaultUniformBuffer` in `BindGroup#destroy` instead? It feels like `BindGroup#destroy` isn't doing a full job when you still need to delete internal properties yourself first.

mvaligursky: It's a property value I assign to it, but it does not own it, as multiple bindGroups can use the same uniform buffer. If JS had some easier way to do ref counting, this would be ref counted perhaps.
It's similar to how when you create a render target, you give it color buffer, but you need to destroy it even when RT itself is destroyed.

mvaligursky: actually, a material -> texture is even more similar to this. Texture, as well as Uniform Buffer is a buffer GPU resource. Material does not own it, it does not destroy it, only uses it. Same here. In this specific instance, there is only one user of this uniform buffer, so I destroy it, but that's not a case in general.

willeastcott (approved):

PR #4925, opened 2022-12-14
Plane/Ray add copy/clone methods + prefer ownership over references

kungfooman:
We are preaching that devs should use `Vec3.FORWARD` etc. but then these two classes don't even support it.

This also adds `copy` and `clone` methods that I always expect (but then I run into script errors).

### DIFF
--- a/src/core/shape/plane.js
+++ b/src/core/shape/plane.js
@@ -9,16 +9,40 @@ const tmpVecA = new Vec3();
  */
 class Plane {
     /**
+     * The starting point of the plane.
+     *
+     * @type {Vec3}
+     */
+    point = new Vec3();
+
+    /**
+     * The normal of the plane.
+     *
+     * @type {Vec3}
+     */
+    normal = new Vec3();
+
+    /**
      * Create a new Plane instance.
      *
-     * @param {Vec3} [point] - Point position on the plane. The constructor takes a reference of
-     * this parameter.
-     * @param {Vec3} [normal] - Normal of the plane. The constructor takes a reference of this
-     * parameter.
+     * @param {Vec3} [point] - Point position on the plane. The constructor copies this parameter.
+     * @param {Vec3} [normal] - Normal of the plane. The constructor copies this parameter.
      */
-    constructor(point = new Vec3(), normal = new Vec3(0, 0, 1)) {
-        this.point = point;
-        this.normal = normal;
+    constructor(point = Vec3.ZERO, normal = Vec3.BACK) {
+        this.set(point, normal);
+    }
+
+    /**
+     * Sets point and normal to the supplied vector values.
+     *
+     * @param {Vec3} point - The starting point of the plane.
+     * @param {Vec3} normal - The normal of the plane.
+     * @returns {Plane} Self for chaining.
+     */
+    set(point, normal) {
+        this.point.copy(point);
+        this.normal.copy(normal);
+        return this;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


yaustar:
I wonder if references were done for optimisation purposes 🤔 We also have to be careful here as this pattern is done across the bounding shapes like AABB, orientated box and sphere.

kungfooman:
> I wonder if references were done for optimisation purposes

I rewrote it a little bit for comparing the ES5 transpilation.

How it looks transpiled to ES5 right now without PR:

```js
function Plane(point, normal) {
    if (point === void 0) {
      point = new Vec3();
    }
    if (normal === void 0) {
      normal = new Vec3(0, 0, 1);
    }
    this.point = point;
    this.normal = normal;
}
```

The way it looks transpiled to ES5 with PR:

```js
function Plane(point, normal) {
    this.point = Vec3.ZERO.clone();
    this.normal = Vec3.BACK.clone();
    if (point) {
        this.point.copy(point);
    }
    if (normal) {
        this.normal.copy(normal);
    }
}
```

Then I wondered how often the references are actually used (tldr: never):

image


For plane its 1x used without references:

image

I don't think this is a CPU performance drain.

Basically I would just like these Math classes to be simpler for two cases:

 - It's easy to forget to clone the input arguments and then the Plane/Ray instances are just referencing instances from completely different classes (e.g. `GraphNode#localPosition`).

- Allow the use of `Vec3.FORWARD` etc. to be in line with the rest of the engine "style decisions"

Thank you for bringing in the other classes like `BoundingBox`, imo they should rather "own" their data aswell, but it's nice to simply brainstorm this first for pros/cons.

mvaligursky:
I fully agree a copy is lot better, and it should not cause any problems for the Plane and the Ray, so I'm happy to do this.
I've tried doing this for all shapes in the past, but hit some problems where some engine code depends on references (and those are hard to find), so this never got merged. Since then I've removed some of those dependencies on references, but there might still be some.

https://github.com/playcanvas/engine/pull/2088

kungfooman:
Wow, that PR is back in ES5 :sweat_smile: 

> Since then I've removed some of those dependencies on references, but there might still be some.

Yea, there are cases a reference makes sense and it feels to me that should rather be quite a "special case". There could be a `static Plane#fromReferences` method for example? That would be as explicit as it gets to mark such cases (and easy to search over hundreds of files).

mvaligursky:
Personally, I'd be a lot more happy to not use references for the shape members at all, the copy is not that costly, and those do not tend to get constructed on the fly anyways.

src/core/shape/plane.js
```diff
@@ -8,17 +8,46 @@ const tmpVecA = new Vec3();
  * @ignore
  */
 class Plane {
+    /**
+     * The starting point of the plane.
+     *
+     * @type {Vec3}
+     */
+    point = Vec3.ZERO.clone();
```
mvaligursky, line 16:
```suggestion
    point = new Vec3();
```

src/core/shape/ray.js
```diff
@@ -1,39 +1,43 @@
...
 /**
  * An infinite ray.
  */
 class Ray {
+    /**
+     * The starting point of the ray.
+     *
+     * @type {Vec3}
+     */
+    origin = Vec3.ZERO.clone();
```
mvaligursky, line 12: we should add '@ignore' tag on the origin and direction members, to hide a public API allowing user to swap those to other instances of Vec3

mvaligursky, line 12:
```suggestion
    origin = new Vec3();
```

kungfooman: Thank you for your review @mvaligursky, I like your simplification! In this case, I am not sure, do you mean `@readonly`?

image

(it allows read-access while not just dropping it from docs)

mvaligursky: ah yeah, readonly is even better!

kungfooman: Thank you a lot for the annotation hint! It feels much better now that Intellisense points it out.

mvaligursky (approved):
Looks great, thanks for the contri!

willeastcott:
This PR generates the following error:

image

This is solved in other classes as follows:

https://github.com/playcanvas/engine/blob/e477bd13b42a9f548153fe33df84430d891cf3a9/src/core/math/vec3.js#L119-L121

kungfooman:
@willeastcott It's not really solved, just dragging TypeScript's incompetence of understanding basic OOP concepts over more lines:

image

PR #4935, opened 2022-12-19
Element Drag Helper - Screen to Local gives wrong values if parent is scaled

### LINKED ISSUE #4934: Scroll View in 3D Space does not scroll correctly
Project: https://playcanvas.com/project/1020980/overview/scroll-view-doesnt-work-in-world

Try to scroll the view with the mouse and the content will shoot off to the side

I've tracked down the issue to ElementDragHelper#_screenToLocal return a change in Z component for the position which shouldn't be possible if it's in the local space of the element

yaustar:
Fixes #4934

`_screenToLocal` was using localPosition which meant if the parent of the element is scaled and not at 0, 0, 0 (which is the case for most 3D screens where the default is 0.01 scale), the calculated intersection would be incorrect and the local Z value would change

### DIFF
--- a/src/framework/components/element/element-drag-helper.js
+++ b/src/framework/components/element/element-drag-helper.js
@@ -149,7 +149,7 @@ class ElementDragHelper extends EventHandler {
         this._determineInputPosition(event);
         this._chooseRayOriginAndDirection();
 
-        _plane.point.copy(this._element.entity.getLocalPosition());
+        _plane.point.copy(this._element.entity.getPosition());
         _plane.normal.copy(this._element.entity.forward).mulScalar(-1);
 
         const denominator = _plane.normal.dot(_ray.direction);


willeastcott (approved):

kungfooman:
Please reconsider, this should break https://github.com/playcanvas/engine/issues/3748 again

yaustar:
Confirmed it does break #3748 again which leaves us in a problem. I will investigate more

yaustar:
@kungfooman The current fix I can think of for both issues is to 0 out the local Z position from _screenToLocal

Test projects:
https://playcanvas.com/project/1020980/overview/scroll-view-doesnt-work-in-world
https://playcanvas.com/project/856038/overview/dragging-in-world-space-issue

https://github.com/playcanvas/engine/commit/c72be6a5650f65ce5bf4c8bfd9b15abdb8a8b78a

PR #4937, opened 2022-12-19
Fix regression of world space drag bug #3748 

yaustar:
Fixes https://github.com/playcanvas/engine/issues/3748

PR https://github.com/playcanvas/engine/pull/4935 reintroduces the bug #3748

This PR zeros out the local Z translation of the drag in `_screenToLocal` so that the element can only move along it's local Y and X axis

1.59

https://user-images.githubusercontent.com/16639049/208466764-9a696a48-025e-40d8-94ef-fb0a17cebd34.mp4

With this PR and #4935

https://user-images.githubusercontent.com/16639049/208466741-67a52e15-fb32-43cc-8a56-8a7a72388750.mp4

CC @kungfooman 

I confirm I have read the [contributing guidelines](https://github.com/playcanvas/engine/blob/master/.github/CONTRIBUTING.md) and signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #4940, opened 2022-12-20
[BREAKING] UI Scroll Views now work in XR

### LINKED ISSUE #3983: UI Scroll view doesn't work in VR - Can't scroll with the VR controllers
Example project: https://playcanvas.com/project/878556/overview/test-scroll-area-in-vr

Forum thread: https://forum.playcanvas.com/t/scrollbar-not-working-in-vr/24047

yaustar:
Fixes #3983

Tested in AR and VR with these projects that already have the PR engine build in them:
- https://playcanvas.com/project/1021298/overview/f-webxr-ui-interaction
- https://playcanvas.com/project/1020950/overview/vr-scrolling-elements

https://user-images.githubusercontent.com/16639049/208667064-2c011316-0f4b-4604-a2f3-5ad8b18abb80.mp4

Breaking change: The `selectend` XR input event for elements has changed to work like `touchend` where it fires the event on the element that it was originally pressed on rather than the element that it is currently on.

I believe this will affect a very small number of users as typically they would either use `click` event to know if a button has been pressed or `selectstart` if they wanted an immediately response.

### DIFF
--- a/src/framework/components/element/element-drag-helper.js
+++ b/src/framework/components/element/element-drag-helper.js
@@ -79,6 +79,7 @@ class ElementDragHelper extends EventHandler {
     _toggleLifecycleListeners(onOrOff) {
         this._element[onOrOff]('mousedown', this._onMouseDownOrTouchStart, this);
         this._element[onOrOff]('touchstart', this._onMouseDownOrTouchStart, this);
+        this._element[onOrOff]('selectstart', this._onMouseDownOrTouchStart, this);
     }
 
     /**
@@ -105,6 +106,10 @@ class ElementDragHelper extends EventHandler {
             this._element[onOrOff]('touchend', this._onMouseUpOrTouchEnd, this);
             this._element[onOrOff]('touchcancel', this._onMouseUpOrTouchEnd, this);
         }
+
+        // webxr events
+        this._element[onOrOff]('selectmove', this._onMove, this);
+        this._element[onOrOff]('selectend', this._onMouseUpOrTouchEnd, this);
 
         this._hasDragListeners = isOn;
     }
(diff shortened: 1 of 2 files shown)


mvaligursky:
If this is [Breaking], it'd be good to have some more details of the breaking change in the description.

yaustar:
@mvaligursky I started the PR but hit post to early 😅  

Breaking change added to the description

src/framework/components/element/element-drag-helper.js
```diff
@@ -140,16 +145,20 @@ class ElementDragHelper extends EventHandler {
...
-        this._chooseRayOriginAndDirection();
+        if (event.inputSource) {
+            _ray.set(event.inputSource.getOrigin(), event.inputSource.getDirection());
+        } else {
+            this._determineInputPosition(event);
+            this._chooseRayOriginAndDirection();
+        }
 
-        _plane.point.copy(this._element.entity.getLocalPosition());
+        _plane.point.copy(this._element.entity.getPosition());
```
yaustar, line 161: May need to revert or resolve a conflict as this was added in https://github.com/playcanvas/engine/pull/4935

src/framework/components/element/element-drag-helper.js
```diff
@@ -215,7 +224,7 @@ class ElementDragHelper extends EventHandler {
 
         dragScale.x = 1 / dragScale.x;
         dragScale.y = 1 / dragScale.y;
-        dragScale.z = 1 / dragScale.z;
+        dragScale.z = 0;
```
yaustar, line 227: This is currently in PR https://github.com/playcanvas/engine/pull/4937

src/framework/input/element-input.js
```diff
@@ -837,21 +837,29 @@ class ElementInput {
             if (hoveredNow) this._fireEvent('selectenter', new ElementSelectEvent(event, hoveredNow, camera, inputSource));
         }
 
+        const pressed = this._selectedPressedElements[inputSource.id];
+        if (eventType === 'selectmove' && pressed) {
+            this._fireEvent('selectmove', new ElementSelectEvent(event, pressed, camera, inputSource));
+        }
```
yaustar, line 843: This works like the touchmove event where it only fires on the element that it was originally clicked on

yaustar:
@Maksims It would be great to get your review too as you originally wrote a lot of this :)

src/framework/input/element-input.js
```diff
@@ -837,21 +837,29 @@ class ElementInput {
...
             if (hoveredNow) this._fireEvent('selectstart', new ElementSelectEvent(event, hoveredNow, camera, inputSource));
         }
 
-        const pressed = this._selectedPressedElements[inputSource.id];
         if (!inputSource.elementInput && pressed) {
             delete this._selectedPressedElements[inputSource.id];
-            if (hoveredBefore) this._fireEvent('selectend', new ElementSelectEvent(event, hoveredBefore, camera, inputSource));
+            if (hoveredBefore) {
+                this._fireEvent('selectend', new ElementSelectEvent(event, pressed, camera, inputSource));
+            }
```
yaustar, line 854: Breaking change as this changes which element `selectend` is fired on.

I could make this non-breaking by changing the way this works more like the mouse where it fires the `mouseup` event on the current element if `mousedown` was not done on an element. And if it was, `mouseup` would fire on the element that was originally clicked on.

src/framework/input/element-input.js
yaustar: Question here is if I should make this more like the mouse where `selectmove` is fired on the current element that it is over unless it was clicked on an existing element to start with, in which case, it would do the above.

src/framework/components/element/element-drag-helper.js
mvaligursky: merge that change it and resolve the comment if no longer relevant

mvaligursky (approved):

PR #4943, opened 2022-12-20
Improve performance of some path's methods

querielo:
The PR improves performance, simplify methods, don't create temporal arrays.


mvaligursky (approved):
Great improvement, thanks!

PR #4949, opened 2022-12-23
Add standard material invert for gloss channels

slimbuck:
This PR:
- Adds three new boolean properties to StandardMaterial:
    - glossInvert
    - sheenGlossInvert
    - clearCoatGlossInvert
- Fixes a bug with sheenGloss
- Deprecates some of the non-standard naming around gloss in Standard Material

The new properties result in glossiness channel being inverted in shader (thereby making them roughness channels). With this PR our materials can support GLTF materials without resorting to custom shader chunks.


src/deprecated/deprecated.js
```diff
@@ -952,7 +955,7 @@ Object.defineProperty(XrInputSource.prototype, 'rotation', {
 
 // INPUT
 
-export const input = {
+export const input_aaaa = {
```
willeastcott, line 958: Whoops

yaustar:
Side question: Does this mean we can added option to the Material Inspector in the Editor for the invert so that users can use roughness maps instead of having to do in an image editor?

src/scene/materials/standard-material.js
```diff
@@ -136,52 +136,53 @@ let _params = new Set();
  * - When anisotropy > 0, anisotropy direction aligns with the bi-normal, and specular anisotropy
  * increases as anisotropy value increases to maximum of 1.
  *
- * @property {number} clearCoat Defines intensity of clear coat layer from 0 to 1. Clear coat layer
+ * @property {number} clearCoat Defines intensity of clearcoat layer from 0 to 1. Clearcoat layer
```
willeastcott, line 139: Are you sure `clearcoat` is one word? See glTF spec: https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_clearcoat/README.md#overview

src/scene/shader-lib/chunks/standard/frag/clearCoatGloss.js
```diff
@@ -18,6 +18,10 @@ void getClearCoatGlossiness() {
     ccGlossiness *= saturate(vVertexColor.$VC);
     #endif
 
+    #ifdef MAPINVERT
+    ccGlossiness = 1.0 - ccGlossiness;
```
willeastcott, line 22: I brought this up in a discussion previously and was told that roughness is definitely **not** 1 - gloss. What's the deal here? cc @GSterbrant @mvaligursky

src/scene/materials/standard-material.js
slimbuck: I had done a quick google and found clearcoat in relatively common usage, for example Wikipedia page https://en.m.wikipedia.org/wiki/Automotive_paint. I could be wrong though!

src/scene/shader-lib/chunks/standard/frag/clearCoatGloss.js
slimbuck: This is not my invention. It was transferred from our previous set of custom chunks in glb-parser.

slimbuck:
> Side question: Does this mean we can added option to the Material Inspector in the Editor for the invert so that users can use roughness maps instead of having to do in an image editor?

Yep 👍🏻

src/deprecated/deprecated.js
slimbuck: Whoops indeed! Thanks

src/scene/shader-lib/chunks/standard/frag/clearCoatGloss.js
willeastcott: Fair enough. I did just do a bit of Googling and there are several posts saying roughness is just gloss inverted. But anyway, seems this isn't a major concern... 😄

src/framework/parsers/glb-parser.js
```diff
@@ -1181,36 +1156,7 @@ const extensionSheen = function (data, material, textures) {
         extractTextureTransform(data.sheenRoughnessTexture, material, ['sheenGlossiness']);
```
GSterbrant, line 1156: Shouldn't this instance of sheenGlossiness also be renamed to sheenGloss?

GSterbrant (changes requested):

src/framework/parsers/glb-parser.js
slimbuck: Yes you're right. I've already done this in a followup PR which includes this and other changes.

mvaligursky:
I think the meaning of the `Invert` parameters is not obvious to the users, and I wonder if we should introduce `roughness` public API, that would under the hood use the same gloss properties / slots, but also switch the internal invert parameters. My main goal here is to make the public API simpler / cleaner / more future proof.

src/scene/materials/standard-material-parameters.js
```diff
@@ -1,13 +1,18 @@
...
-    const tex = {};
-    tex[`${name}Map`] = 'texture';
-    if (channel)
-        tex[`${name}MapChannel`] = 'string';
-    tex[`${name}MapUv`] = 'number';
-    tex[`${name}MapTiling`] = 'vec2';
-    tex[`${name}MapOffset`] = 'vec2';
-    tex[`${name}MapRotation`] = 'number';
-    return tex;
+function _textureParameter(name, channel = true, vertexColor = true) {
```
GSterbrant, line 1: Still think we should change this name if we are going to introduce vertex colors. If we call it _modifierParameters and add texture as a boolean as well, this can be much more generic in my opinion.

src/scene/materials/standard-material.js
```diff
@@ -695,7 +699,7 @@ class StandardMaterial extends Material {
                 this._setParameter('material_sheen', getUniform('sheen'));
             }
             if (!this.sheenGlossMap || this.sheenGlossTint) {
-                this._setParameter('material_sheenGlossiness', this.sheenGlossiness);
+                this._setParameter('material_sheenGlossiness', this.sheenGloss);
```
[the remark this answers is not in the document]
GSterbrant: Do we have an issue for this?

src/scene/materials/standard-material.js
```diff
@@ -1129,7 +1136,8 @@ function _defineMaterialProps() {
     _defineFloat('metalness', 1);
     _defineFloat('anisotropy', 0);
     _defineFloat('clearCoat', 0);
-    _defineFloat('clearCoatGlossiness', 1);
+    _defineFloat('clearCoatGloss', 1);
+    _defineFlag('clearCoatGlossInvert', false);
```
GSterbrant, line 1140: I think the flag should be defined with the other flags. We tend to keep this separated on define type, not on it's use.

src/scene/materials/standard-material-parameters.js
slimbuck: I can change the name, but what's the point of adding a boolean that will always be true?

src/scene/materials/standard-material-options.js
```diff
@@ -20,6 +20,9 @@ import { LitOptions } from "./lit-options.js";
  * value.
  * @property {boolean} opacityTint Defines if {@link StandardMaterial#opacity} constant should affect opacity value.
  * @property {boolean} packedNormal If normal map contains X in RGB, Y in Alpha, and Z must be reconstructed.
+ * @property {boolean} glossInvert Invert the gloss channel.
```
mvaligursky, line 23: These comments could explain relation to roughness.

src/scene/materials/standard-material-parameters.js
GSterbrant: Can't we have only vertex color without texture? If so, then there is no need.

src/scene/materials/standard-material.js
slimbuck: I've moved that.

Though it would probably make more sense to order these definitions similar to the class jsdoc order (or material inspector order) - i.e. _logically_ instead of by data type. Doing so should make visually comparing the lists of data between engine/schema/parameters more natural. To me, it makes sense for all parameters affecting the diffuse component to be grouped together, then specular, gloss etc.

src/scene/materials/standard-material.js
```diff
@@ -920,9 +924,9 @@ const defineProp = (prop) => {
     return prop.defaultValue && prop.defaultValue.clone ? defineAggProp(prop) : defineValueProp(prop);
 };
 
-function _defineTex2D(name, uv, channels, defChannel, vertexColor, detailMode) {
+function _defineTex2D(name, channel = "rgb", vertexColor = true, uv = 0) {
     // store texture name
-    _matTex2D[name] = channels;
+    _matTex2D[name] = channel.length || -1;
```
mvaligursky, line 929: but strange to store -1 as texture name?

src/scene/materials/standard-material.js
slimbuck: Perhaps alphabetical order would make most sense...

src/scene/materials/standard-material.js
slimbuck: Yup agree (this is not new).

src/scene/materials/standard-material.js
```diff
@@ -955,34 +959,25 @@ function _defineTex2D(name, uv, channels, defChannel, vertexColor, detailMode) {
...
-        if (channels > 0) {
             defineProp({
                 name: `${name}VertexColorChannel`,
-                defaultValue: defChannel ? defChannel : (channels > 1 ? 'rgb' : 'g')
+                defaultValue: channel
             });
         }
     }
 
-    if (detailMode) {
```
mvaligursky, line 979: is the detail mode no longer needed?

src/scene/materials/standard-material.js
GSterbrant: I agree, but then we should change them all in my opinion 🙂

src/scene/materials/standard-material.js
mvaligursky: I see it was moved to a separate setup

src/scene/shader-lib/chunks/standard/frag/gloss.js
```diff
@@ -1,13 +1,13 @@
 export default /* glsl */`
 #ifdef MAPFLOAT
-uniform float material_shininess;
+uniform float material_glossiness;
```
mvaligursky, line 3: considering you're already renaming the uniform .. should it be called material_gloss?

src/scene/materials/standard-material.js
slimbuck: > we should change them all in my opinion
yes!

The more I think about it, the more it makes sense we reorder material members to strict alphabetical. Then we can reorder everything else too and comparing diffs becomes much more straight-forward.

I won't do that that here though, of course.

mvaligursky (approved):
looks great, but added few comments

src/scene/shader-lib/chunks/standard/frag/gloss.js
slimbuck: Currently all shaders use 'glossiness' naming. I will create a ticket to address them all separately.

GSterbrant (approved):

src/scene/materials/standard-material.js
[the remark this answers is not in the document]
slimbuck: https://github.com/playcanvas/engine/issues/4952

src/scene/materials/standard-material-parameters.js
slimbuck: Actually can we make this change in another PR?

`textureParameter` currently mirrors standard-material's `defineTex2D`, so if we change this naming we should also change that.

Since we're going to reorder members to be alphabetical as well, I'd rather we just do all that together in a separate PR.

src/scene/materials/standard-material-options.js
slimbuck: explanation was added to standard-material jsdoc

PR #4956, opened 2023-01-09
[Fix] Fix null access caused in clustered light code when no meshes use lights

### LINKED ISSUE #4950: Disabling a model Entity with Clustered lighting gives errors
Related thread: https://forum.playcanvas.com/t/solved-cannot-read-properties-of-null-reading-device/29128/

Project: https://playcanvas.com/project/1025290/overview/clustered-lighting-disable-bug

Launch project
Press 1 to disable the cube
Errors appear:

```
Uncaught TypeError: Cannot read properties of null (reading 'width')
    at LightsBuffer.addLightData (lights-buffer.js:400:61)
    at WorldClusters.updateClusters (world-clusters.js:389:31)
    at WorldClusters.update (world-clusters.js:440:14)
    at ForwardRenderer.updateClusters (forward-renderer.js:1913:29)
    at RenderPass.execute (forward-renderer.js:1969:22)
    at RenderPass.render (render-pass.js:215:14)
    at FrameGraph.render (frame-graph.js:127:29)
    at Application.renderComposition (app-base.js:1240:25)
    at Application.render (app-base.js:1228:14)
    at app-base.js:2201:29
```

mvaligursky:
Fixes #4950 

Caused by keeping a reference to released WorldClusters in case a layer no longer has meshes, instead of setting it to null.

### DIFF
--- a/src/scene/composition/layer-composition.js
+++ b/src/scene/composition/layer-composition.js
@@ -542,6 +542,8 @@ class LayerComposition extends EventHandler {
             const ra = this._renderActions[i];
             const layer = this.layerList[ra.layerIndex];
 
+            ra.lightClusters = null;
+
             // if the layer has lights used by clusters
             if (layer.hasClusteredLights) {
 


yaustar (approved):
LGTM. `lightClusters` is not a getter so there's no extra logic being ran under the hood

PR #4962, opened 2023-01-10
Rename material uniform glossiness to gloss

### LINKED ISSUE #4952: Rename shader 'glossiness' uniforms to 'gloss'
The engine now uses 'gloss' everywhere else, so it would be better for shader uniform names to also.

slimbuck:
Fixes #4952

Rename material uniforms from glossiness -> gloss and update chunk API versions.


mvaligursky (approved):
Looks good. Please update shader chunk migration doc as well.

Issue #3559, opened 2021-10-11
Investigate parallel shader compilation


mvaligursky:
Try the suggestion here:
https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#compile_shaders_and_link_programs_in_parallel

It's tempting to compile shaders and link programs serially, but many browsers can compile and link in parallel on background threads.

Adriaaaaan:
Very much looking forward to progress in this area, my use case is very dynamic in terms of loading in assets and so shader compilation pauses are very detrimental

PR #4964, opened 2023-01-10
Update to WebGL shader compilation and linking

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/3559

Situation before:
- when a shader is created (which happens at a point the mesh using it needs to be rendered), it would internally compile both vertex and fragment shaders, and then it would immediate link them and wait for the result, blocking the main thread till the shaders are compiled.

Situation now:
- The way forward renderer mesh drawing loop was refactored months ago is that we loop over the meshes in two passes. In the first loop, we just create shaders if those are missing. The second loop uses those shaders for rendering. With this PR, the shader creation was changed to only trigger their compilation. The shader linking takes place when a batch of shaders is created by forward renderer - basically after shaders for a layer are created (so one batch for shadows, one for forward pass for example). This only kicks off linking of all those new shaders, without waiting for the results, so they can link in parallel. Waiting on the result is done when the shader is set up for rendering.

What this delivers is conceptually this (2 meshes with 2 shaders):

**Before:**
```
shader1.vertex.compile
shader1.fragment.compile
shader1.link
shader1.finalize **(blocking)**
shader2.vertex.compile
shader2.fragment.compile
shader2.link
shader2.finalize **(blocking)**
```

**Now:**
```
shader1.vertex.compile
shader1.fragment.compile
shader2.vertex.compile
shader2.fragment.compile
shader1.link
shader2.link
shader1.finalize **(blocking)**
shader2.finalize **(blocking)**
```

Note that those loops execute per render layer / render pass, so shadows would do compile/link/finalize on shaders before main forward pass woud.

The actual log from multi-view engine example (the number is unique ID of the shader):
Screenshot 2023-01-10 at 17 32 34

In a larger project, I see a lot more shaders getting compiled before any linking, for example
Screenshot 2023-01-10 at 17 38 34

So in theory, this does the right thing, but in practise, on few projects I have tried, I cannot see a performance difference (using Chrome on both MacOS and Windows).

### DIFF
--- a/src/platform/graphics/device-cache.js
+++ b/src/platform/graphics/device-cache.js
@@ -43,7 +43,7 @@ class DeviceCache {
      * @param {import('./graphics-device.js').GraphicsDevice} device - The graphics device.
      */
     remove(device) {
-        this._cache.get(device)?.destroy(device);
+        this._cache.get(device)?.destroy?.(device);
         this._cache.delete(device);
     }
 }
--- a/src/platform/graphics/webgl/webgl-graphics-device.js
+++ b/src/platform/graphics/webgl/webgl-graphics-device.js
@@ -1156,6 +1156,15 @@ class WebglGraphicsDevice extends GraphicsDevice {
     }
 
     /**
+     * Called after a batch of shaders was created, to guide in their optimal preparation for rendering.
+     *
+     * @ignore
+     */
+    endShaderBatch() {
+        WebglShader.endShaderBatch(this);
+    }
+
+    /**
      * Set the active rectangle for rendering on the specified device.
      *
      * @param {number} x - The pixel space x-coordinate of the bottom left corner of the viewport.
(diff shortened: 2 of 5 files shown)


mvaligursky:
I'll follow this up with additional experiments and possibly PR to try and get shader link steps to run in parallel as well.

mvaligursky:
> Could it be that when you call link it actually triggers compilation which is not blocking, but then checking link status is blocking. So if you call link for all awaiting shaders and only then check their statuses - it might do the trick.

Yep that is exactly what I'm experimenting with now. Code is ready, but trying to add some reliable way to measure the impact.

mvaligursky:
I made link calls to run in parallel, before waiting for them by calling
```
        const linkStatus = gl.getProgramParameter(glProgram, gl.LINK_STATUS);
```
but I still do not see any improvement.

PR #4975, opened 2023-01-13
Additional engine tracing related to frame timing

mvaligursky:


src/core/constants.js
```diff
@@ -5,6 +5,13 @@
  */
 export const TRACEID_RENDER_FRAME = 'RenderFrame';
 
+/**
+ * Logs a frame times.
```
willeastcott, line 9:
```suggestion
 * Logs a frame time.
```

willeastcott (approved):

PR #4984, opened 2023-01-16
Added attributes parameter to createShaderFromCode

mvaligursky:
- updated createShaderFromCode function to accept attributes object, which is commonly used. Also, removed fragmentPreamble parameter, as it's been used only once internally, and that is no longer needed.
- changed reproject-texture shader creation. It no longer needs to enable those extensions, as this is handled inside createShaderFromCode (tested on WebGL1)
- improved shader error reporting / handling
- updated shader*** examples to use createShaderFromCode, and confirmed they all work on WebGPU now

### DIFF
--- a/src/scene/shader-lib/utils.js
+++ b/src/scene/shader-lib/utils.js
@@ -2,6 +2,7 @@ import { Shader } from '../../platform/graphics/shader.js';
 import { ShaderUtils } from '../../platform/graphics/shader-utils.js';
 import { shaderChunks } from './chunks/chunks.js';
 import { getProgramLibrary } from './get-program-library.js';
+import { Debug } from '../../core/debug.js';
 
 /**
  * Create a shader from named shader chunks.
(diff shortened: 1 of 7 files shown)


src/scene/shader-lib/utils.js
```diff
@@ -31,19 +31,20 @@ function createShader(device, vsName, fsName, useTransformFeedback = false) {
...
  * @param {boolean} [useTransformFeedback] - Whether to use transform feedback. Defaults to false.
- * @param {string} [fragmentPreamble] - An optional 'preamble' string for the fragment shader. Defaults
- * to ''.
+ * @param {Object<string, string>} [attributes] - Object detailing the mapping of vertex shader
+ * attribute names to semantics SEMANTIC_*. This enables the engine to match vertex buffer data as
+ * inputs to the shader. Defaults to undefined, which generates the default attributes.
  * @returns {Shader} The newly created shader.
  */
-function createShaderFromCode(device, vsCode, fsCode, uniqueName, useTransformFeedback = false, fragmentPreamble = '') {
+function createShaderFromCode(device, vsCode, fsCode, uniqueName, useTransformFeedback = false, attributes) {
```
slimbuck, line 39: Might be worth swapping the last two arguments around since attributes is (presumably) much more likely to get used compared to useTransformFeedback?

slimbuck (approved):

src/scene/shader-lib/utils.js
mvaligursky: I thought about it, but users are likely using the transform feedback parameter.
Maybe I can check types and warn the users.

slimbuck: OIC, thought this was internal. Probably best to leave it then.

mvaligursky: ok done, and added assert to test for incorrect calls to the function

PR #4985, opened 2023-01-16
Fix shadow map using vertex normals.

### LINKED ISSUE #4971: Remove dependency of shadow vertex shader on matrix_normal
This is a vertex shader we use to render to shadow map. 

```
uniform mat4 matrix_viewProjection;
uniform mat4 matrix_model;
uniform mat3 matrix_normal;

vec3 dPositionW;
mat4 dModelMatrix;
mat3 dNormalMatrix;

mat4 getModelMatrix() {
    return matrix_model;
}

vec4 getPosition() {

    vec4 posW = dModelMatrix * vec4(localPos, 1.0);
    dPositionW = posW.xyz;

    vec4 screenPos;
    screenPos = matrix_viewProjection * posW;

    return screenPos;
}

vec3 getWorldPosition() {
    return dPositionW;
}

vec3 getNormal() {
    dNormalMatrix = matrix_normal;

    vec3 tempNormal = vertex_normal;
    return normalize(dNormalMatrix * tempNormal);
}

void main(void) {
    gl_Position = getPosition();
   vPositionW    = getWorldPosition();
   vNormalW = getNormal();

}
```

I added some new warnings to engine (commented out for now), and it triggers this:
`Shader [Shader Id 0 standard] requires uniform [matrix_normal] which has not been set`

When WebGL translates the shader for the platform, the uniform is not stripped out as dead code, but is compiled out on the driver level.

Ideally we'd change the shader generator to not include it to avoid - as we unnecessarily handle it. On WebGPU we need to allocate space for it in the uniform buffer as well.

When this is done, the warning should be enabled, as it's very helpful to have.

```
// Debug.warnOnce(`Shader [${shader.label}] requires uniform [${uniform.scopeId.name}] which has not been set, while rendering [${DebugGraphics.toString()}]`);

```

GSterbrant:
Fixes #4971

Shadow map tries to output to wNormal which is an unused attribute. This cause a chain of event which begins with a uniform matrix_normal being unused but not stripped out of the shader code.

### DIFF
--- a/src/scene/shader-lib/programs/lit-shader.js
+++ b/src/scene/shader-lib/programs/lit-shader.js
@@ -91,6 +91,7 @@ class LitShader {
         this.shadowPass = ShaderPass.isShadow(options.pass);
         this.needsNormal = this.lighting || this.reflections || options.useSpecular || options.ambientSH || options.heightMapEnabled || options.enableGGXSpecular ||
                             (options.clusteredLightingEnabled && !this.shadowPass) || options.clearCoatNormalMapEnabled;
+        this.needsNormal = this.needsNormal && !this.shadowPass;
         this.needsSceneColor = options.useDynamicRefraction;
         this.needsScreenSize = options.useDynamicRefraction;
         this.needsTransforms = options.useDynamicRefraction;


mvaligursky (approved):
Looks good.
Can you please test this with omni a spot shadows for both WebGL1 and WebGL2 .. just making sure we don't do manual slope bias or something similar on GL1 that uses normals. Try both clustered, but especially non-clustered lighting.

GSterbrant:
Tested GL1, GL2, clustered, non clustered and all looks good. We use the derivatives on depth for the polygon offset stuff on GL1 :).

PR #4997, opened 2023-01-20
Improved documentation for Shader creation

mvaligursky:
- clarification

### DIFF
--- a/src/scene/shader-lib/utils.js
+++ b/src/scene/shader-lib/utils.js
@@ -24,13 +24,17 @@ function createShader(device, vsName, fsName, useTransformFeedback = false) {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/scene/shader-lib/utils.js
```diff
@@ -24,13 +24,17 @@ function createShader(device, vsName, fsName, useTransformFeedback = false) {
 }
 
 /**
- * Create a shader from the supplied source code.
+ * Create a shader from the supplied source code. Note that this functions adds additional shader
```
willeastcott, line 27:
```suggestion
* Create a shader from the supplied source code. Note that this function adds additional shader
```

willeastcott (approved):

PR #4999, opened 2023-01-24
Added public API notes for the container resource

### LINKED ISSUE #3872: `GlbContainerResource` is not public API
Currently, `GlbContainerResource` does not appear in the API reference manual. So we have no public docs for functions like `GlbContainerResource#instantiateRenderEntity`. This is despite the viewer and 25 engine examples using this function. And as a result, it does not appear in the TypeScript declarations.

However, I would argue that this function is not named or located (source-wise) correctly. All other resources do not have `Resource` appended to the name. e.g. `Model`, `Texture`, `Material`, etc. And all of those resources appear in the relevant subsystem folder (e.g. `graphics`, `scene`, etc). I would have thought it should be `GlbContainer` in `scene`, perhaps. That said, the `scene` layer doesn't otherwise know what a GLB is, so I'm not completely comfortable with that either.

yaustar:
Fixes #3872

Adds public API for the different asset types

### DIFF
--- a/src/framework/handlers/container.js
+++ b/src/framework/handlers/container.js
@@ -8,6 +8,10 @@ import { GlbParser } from '../parsers/glb-parser.js';
  * @interface
  * @name ContainerResource
  * @description Container for a list of animations, textures, materials, renders and a model.
+ * @property {import('../asset/asset.js').Asset[]} renders An array of the Render assets.
+ * @property {import('../asset/asset.js').Asset[]} materials An array of {@link Material} and/or {@link StandardMaterial} assets.
+ * @property {import('../asset/asset.js').Asset[]} textures An array of the {@link Texture} assets.
+ * @property {import('../asset/asset.js').Asset[]} animations An array of the {@link Animation} assets.
  */
 class ContainerResource {
     /**


src/framework/handlers/container.js
```diff
@@ -8,6 +8,10 @@ import { GlbParser } from '../parsers/glb-parser.js';
  * @interface
  * @name ContainerResource
  * @description Container for a list of animations, textures, materials, renders and a model.
+ * @property {import('../asset/asset.js').Asset[]} renders An array of the Render assets.
+ * @property {import('../asset/asset.js').Asset[]} materials An array of {@link Material} and/or {@link StandardMaterial} assets.
```
mvaligursky, line 12: glb parser always creates StandardMaterial, never Material.

mvaligursky (approved):

src/framework/handlers/container.js
yaustar, line 12:
```suggestion
 * @property {import('../asset/asset.js').Asset[]} materials An array of the {@link StandardMaterial} assets.
```

PR #5012, opened 2023-01-27
Optimize clustered lighting shaders compilation time on Windows platform

mvaligursky:
- using A8 instead of RGBA8 texture format to store light indices, which means a single texel stores a single light index, instead of 4. This allows the inner loop over 4 lights per pixel to be removed. Measured compilation speedup of 30-40% on Windows, both with and without area lights.
- small refactor of texture to allow WebGPU support for A8 texture.

### DIFF
--- a/src/platform/graphics/webgpu/webgpu-texture.js
+++ b/src/platform/graphics/webgpu/webgpu-texture.js
@@ -1,6 +1,7 @@
 import { Debug, DebugHelper } from '../../../core/debug.js';
 
 import {
+    pixelFormatByteSizes,
     ADDRESS_REPEAT, ADDRESS_CLAMP_TO_EDGE, ADDRESS_MIRRORED_REPEAT,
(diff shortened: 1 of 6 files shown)


src/platform/graphics/webgpu/webgpu-texture.js
```diff
@@ -265,13 +266,19 @@ class WebgpuTexture {
             mipLevel: 0
         };
 
-        // TODO: RGBA only for now, needs to be more generic
-        const numElementsPerPixel = 4;
+        // TODO: handle update to
```
willeastcott, line 269: Did you mean 'too'?

willeastcott (approved):
This is incredible. Awesome speedup! 🏃

src/platform/graphics/webgpu/webgpu-texture.js
mvaligursky: fixed. I was in the middle of typing the comment, jumped elsewhere to make some other change, and never finished it :)

PR #5017, opened 2023-01-30
ES6'ify GamePads + proper JSDoc typing

kungfooman:
Fixes two issues:

- Using `src/platform/input/game-pads.js` "standalone":

```js
const { GamePads } = await import("/playcanvas-engine-jsdoc/src/platform/input/game-pads.js?now=" + Date.now());
const gamePads = new GamePads();
gamePads.update();
gamePads.isPressed(0, 0);
```

Result:

image

 - Attemping to use `GamePads` in node:

```js
> pc = require("./build/playcanvas.js");
> const gamePads = new pc.GamePads()
Uncaught ReferenceError: navigator is not defined
    at new GamePads (/var/www/html/playcanvas-engine-jsdoc/build/playcanvas.js:11412:31)
```

This PR also optimizes the `for` loop `GamePads#update` a bit, the "make sure it's an array" statement doesn't depends on `j`.

Also added `Product: 09cc`, which is a PS4 DualShock controller.


src/platform/input/game-pads.js
```diff
@@ -133,28 +152,28 @@ class GamePads {
...
     }
 
+    /**
+     * @param {Gamepad} pad - The HTML5 Gamepad object.
+     * @returns {ButtonsAxes} Object defining the order of buttons and axes for given HTML5 Gamepad.
+     */
     getMap(pad) {
         for (const code in PRODUCT_CODES) {
-            if (pad.id.indexOf(code) >= 0) {
+            if (pad.id.includes(code)) {
```
willeastcott, line 173: `Array#includes` is not available for IE11, hence the use of `indexOf`.

src/core/platform.js
```diff
@@ -9,6 +9,16 @@ let gamepads = false;
 let workers = false;
 let passiveEvents = false;
 
+// If the environment doesn't support the Gamepad API, we return the same array (GC friendly).
+// Examples: node or browsers that hide USB devices to prevent fingerprinting of users.
+const dummy = [];
+const getDummy = () => dummy;
+
+/**
+ * @type {typeof navigator.getGamepads}
```
mvaligursky, line 18: JSDocs is not keen on `typeof` - I tried to use it few times already. Any workaround?

src/platform/input/game-pads.js
kungfooman: Hi @willeastcott, this isn't `Array#includes` but `String#includes` and it's polyfilled and used in other places of the engine aswell. Thank you for taking your time to review, I appreciate it a lot!

kungfooman:
> Note that I'm not too keen on JSDoc plugins in the engine source tree - I'm hoping this year we transition to TypeDoc anyway

My main motivation is just to finally achieve proper typing. It is disheartening to know that we still have incorrect data types.

The good/old `typeof` issue:

https://github.com/playcanvas/engine/blob/9e00e9344c384da25c6beef808038c4b236399f8/src/framework/app-options.js#L72-L91

I'm not sure why incorrect types are better than a little script.

Will you allow `@typedef` once we have transitioned to TypeDoc?

willeastcott:
> I'm not sure why incorrect types are better than a little script.

Well, I wouldn't say that's my view. But we already have a stack of hacks in the repo to generate the correct docs and types and I desperately want to get rid of them rather than introduce more. If we can migrate all components to the 'new' format and eliminate all uses of `Object.defineProperty` to define API, we can get rid of the types hacks and get rid of JSDoc and use TypeDoc instead with (hopefully) zero workarounds.

PR #5019, opened 2023-02-01
Fixed dynamic world gravity constant update

### LINKED ISSUE #5018: Application World Gravity being set on every update
Fixes: #5018

Hello,

I noticed that the `RigidBodyComponentSystem` class had something going on with gravity setting on its `onUpdate` function. In fact due to some weird stuff from Ammo.js, when the world gravity is set using `this.dynamicsWorld.setGravity(gravity)`, Ammo set it to something close but not exactly the given vector which leads to making this following check always true, setting a new gravity on every update instead of changing it when there is an actual real gravity change on the application.
 
 
https://github.com/playcanvas/engine/blob/6b1a93a7c2dc19f48c70c201100ffdf6e07d5358/src/framework/components/rigid-body/system.js#L868-L873
 
 
Would it be possible to make it a distance check instead? From what I saw on Ammo.js is having a precision around 0.001 or  0.0001, so a code like this should fix the issue:

```JS
const gravity = this.dynamicsWorld.getGravity();
if (this.gravity.distance({ 'x': gravity.x(), 'y': gravity.y(), 'z': gravity.z() }) > 0.001) {
    gravity.setValue(this.gravity.x, this.gravity.y, this.gravity.z);
    this.dynamicsWorld.setGravity(gravity);
}
```
 
 
I was working on a script where I need to disable the world gravity for an entity unfortunately I checked within the Bullet code and as this `setGravity` function is setting individual entity gravity to all entities it overwrites my entity's gravity every update...
 
 
Thank you very much for the consideration and your work!

MushAsterion:
Changing gravity comparison function to be distance based instead of precise value due to Ammo.js not making the gravity a precise value when assigning which was leading to the gravity being set every single update if gravity was not [ 0, 0, 0 ]. Distance check require a bit more of calculations however it's nothing compared to updating every single rigid bodies gravity on every frame not to mention that is was blocking an important Ammo.js feature (entity's gravity).

Fixes #5018 


yaustar:
Confirmed that gravity is being set each frame 😱

LeXXik:
Do we have a repro?

yaustar:
> Do we have a repro?

In my case, I've just added Ammo module and add a log point to that line where it sets the gravity

LeXXik:
I see, yes, it is a precision loss:
image

yaustar:
@LeXXik Do you think we can do a distanceSq check against Number.EPSILON or will the downcasting to 32bit be robust enough?

LeXXik:
I would suggest to downcast, it should do be fine. Using `this.gravity.x = Math.fround(value);`. Another option would be storing it in a temporary typed array `this.array.x = float32arr[0];`.

kungfooman:
`Number.EPSILON` seems a bit too small, but 0.000000000001 is good with distanceSq check.

Basically -9.8100004196167 and -9.81 are unequal and the squared distance is 0.00000000000017607817:

```js
a = new pc.Vec3(0, -9.8100004196167, 0);
b = new pc.Vec3(0, -9.81, 0);
c = new pc.Vec3().sub2(a, b).lengthSq().toFixed(20);
```

Then just removing one zero for some more tolerance, we end up with:

```js
        const lengthSq = tmpAmmoGravity.set(gravity.x(), gravity.y(), gravity.z()).sub(this.gravity).lengthSq();
        if (lengthSq > 0.000000000001)
        {
```

Which allows way less coarse gravity setting.

Issue with PR is also the introduction of the "Vec3Like" object `{x: ..., y: ..., z: ...}` which introduces GC, which of course can be circumvented using a tmp vector in top of the file:

```
const tmpAmmoGravity = new Vec3();
```

@LeXXik 

That doesn't seem to work: `Math.fround(-9.8100004196167) == -9.81`

mvaligursky:
Should we just use ammo temporary, set it on that and compare with that? something like
`tempGravity.setValue(this.gravity.x, this.gravity.y, this.gravity.z);`

kungfooman:
@LeXXik Oops, this is how it works with your suggestion:
```
Math.fround(-9.8100004196167) == Math.fround(-9.81) // true
```

slimbuck:
Oops I hadn't seen this.

I also did #5020, which is probably a better approach (and tested to work).

MushAsterion:
Thank you for your feedback. I updated the code accordantly however it seem that @slimbuck's solution on #5020 is would be better for performance.

PR #5024, opened 2023-02-01
Fixing Entities being cloned on attribute change

### LINKED ISSUE #4628: attribute setter will clone Entity, if new value setted
In class ScriptAttributes, attribute-setter will clone old value, if it is not json and has clone function.
But, entities should not be cloned, in this situation. That may waste plenty of memory and time, if I had referenced Root entity previously.

MushAsterion:
Added additional validation before cloning so entities don't get cloned when attribute is changed.

Fixes #4628 

### DIFF
--- a/src/framework/script/script-attributes.js
+++ b/src/framework/script/script-attributes.js
@@ -283,7 +283,8 @@ class ScriptAttributes {
                 let oldCopy = old;
                 // json types might have a 'clone' field in their
                 // schema so make sure it's not that
-                if (old && args.type !== 'json' && old.clone) {
+                // entities should not be cloned as well
+                if (old && args.type !== 'json' && args.type !== 'entity' && old.clone) {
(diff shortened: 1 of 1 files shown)


slimbuck (approved):
I doubt existing apps depend on the old value being a cloned entity, but we will have to test this change thoroughly.

This is a good fix, thanks so much!

MushAsterion:
> I doubt existing apps depend on the old value being a cloned entity, but we will have to test this change thoroughly.
> 
> This is a good fix, thanks so much!

Considering the entity type is still being check in the `rawToValue` function being called right after with the same old value I think it's just safer to test it at the cloning as well.

https://github.com/playcanvas/engine/blob/55e8304c1f88c8741bc08bdab833964f91e3c12b/src/framework/script/script-attributes.js#L71-L77

PR #5028, opened 2023-02-02
Add support for building the API reference with Typedoc

willeastcott:
This PR adds the ability to use [Typedoc](https://typedoc.org/) to build the API reference manual. The JSDoc build process is retained but the thinking is that we'll migrate to Typedoc in time, once the output is sufficiently good.

To build, do:

```
npm run typedocs
```

It generates this:

image

And here's an example class:

image


mvaligursky (approved):
That looks great. It'd be good to spin up an issue of what needs to be resolved that you know of.

PR #5033, opened 2023-02-02
Add field declarations for all core classes

willeastcott:
This PR adds [field declarations](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields) for all classes in the `src/core` folder. This improves the Typedocs, changing this:

image

Into this:

image

It also further aligns the code with TypeScript, where field declarations are mandatory.

### DIFF
--- a/src/core/event-handler.js
+++ b/src/core/event-handler.js
@@ -14,34 +14,31 @@
 
 /**
  * Abstract base class that implements functionality for event handling.
+ *
+ * ```javascript
+ * var obj = new EventHandlerSubclass();
+ *
+ * // subscribe to an event
+ * obj.on('hello', function (str) {
+ *     console.log('event hello is fired', str);
+ * });
+ *
+ * // fire event
+ * obj.fire('hello', 'world');
+ * ```
  */
 class EventHandler {
     /**
-     * Create a new EventHandler instance.
-     *
-     * @example
-     * var obj = new EventHandlerSubclass();
-     *
-     * // subscribe to an event
-     * obj.on('hello', function (str) {
-     *     console.log('event hello is fired', str);
-     * });
-     *
-     * // fire event
-     * obj.fire('hello', 'world');
+     * @type {object}
+     * @private
      */
-    constructor() {
-        /**
-         * @type {object}
-         * @private
-         */
-        this._callbacks = { };
-        /**
-         * @type {object}
-         * @private
-         */
-        this._callbackActive = { };
-    }
+    _callbacks = {};
+
+    /**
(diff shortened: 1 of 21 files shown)


slimbuck (approved):

kungfooman:
You are rolling through this like a champ, very nice to see the progress :+1: :dart: 

May the JSDoc shackles be forgotten at some point :sweat_smile:

willeastcott:
Wouldn't that be great? Have you inspected the Typedoc output yet? What do you think of the styling and layout compared to the JSdoc output?

kungfooman:
So far I can see some pros and cons, lets just take one example to make it specific:

image


image

I think the presentation of a factual data in TypeDoc is:
1) overly "typified", which makes it rather confusing to read
2) redundant
3) taking too many new lines, which creates a lot of "air" inbetween
4) `numVerts??: number` is simply ugly compared to `[numVerts]`
5) the addiction to `:` everywhere, creating ASCII soup that brain/eyes need to filter away

typedoc signature: `computeMinMax(vertices: number[] | Float32Array, min: Vec3, max: Vec3, numVerts??: number): void`

While in JSDoc there is a short line that looks like something you can copy straight into the source code with minimal changes:

JSDoc signature: `computeMinMax(vertices, min, max, [numVerts])`

JSDoc presents what I really want to see/copy/use, in a short line.

Another example:

image

"returns a BoundingBox Returns Returns BoundingBox" :thinking: 

image

It nearly feels like 3 or 4 times as big as it "really" needs to be. They somehow got addicted to putting that `:` everywhere, as if that helps.

What I like about TypeDoc is the link to the actual source code.

Or the search just feels better:

image

JSDoc finds nothing:

image

So my feeling is that typedoc is technically more "aware" of certain things, but the representation absolutely needs to be improved. While I like JSDoc for its simple presentation of exactly what I want to see.

Hm, I thought that typedoc supports `@typedef`'s but it doesn't :cry: 

```js
/**
 * Super typey type.
 * 
 * @typedef {object} SuperTypeyType
 * @property {number} a - The a.
 * @property {number} b - The b.
 */
```

image

Feels still better in JSDoc, even tho it's a 404 that would need to be created:

image

willeastcott:
@kungfooman This is great feedback. I get what you say about the verbosity of the Typedoc output. But I guess anyone developing against TypeScript are going to love it! 😄  

At the moment, the Typedoc output is highly experimental, and of course, it just uses the default template for now. It is no doubt customizable (although I don't know to what extent). I think I'll continue to experiment as see where it goes....

PR #5039, opened 2023-02-03
Added support for `shapeCast` and merged `raycastFirst` and `raycastAll` into `raycast`

### LINKED ISSUE #2094: Add support for spherecast
From forum thread: https://forum.playcanvas.com/t/ability-to-use-spherecast/13310/3

It would be great to add some form of sphere casting if the current Ammo,js API allows it.

MushAsterion:
- [x] **Requires to update Ammo.js to latest version.** (which is not a breaking change)
- [X] **No custom IDL for Ammo.js required.**

Deprecated `RaycastResult` for `HitResult` as it is now returned by `shapeCast`.
Added the following property for `HitResult` public API:
- `distance`: The distance at which the hit occurred from the starting point.

As shape may have lower `hitFraction` but be more distant than another hit with higher `hitFraction` (for example a side of the cylinder will be closest in distance than its other tip. (#5179)

Deprecated `raycastFirst` and `raycastAll` for a new function `raycast` defaulting to a "first" behavior, with possibility to use with option `findAll = true` to have all results.
```JS
raycast(start: Vec3, end: Vec3, options)
```

Added test and cast for shapes using `shapeCast`:
```JS
shapeCast(shape, startPosition: Vec3, startRotation: Vec3|Quat, endPosition; Vec3, options)
```

Where `shape` can either be a `Ammo.btCollisionShape` or an object defined as follows:
```JS
{
    type: "box"|"cylinder"|"capsule"|"cone"|"sphere",
    halfExtends: Vec3, // For type "box"
    radius: number, // For any type but "box"
    height: number, // For types "cylinder", "capsule" and "cone"
    axis: number // For types "cylinder", "capsule" and "cone"
}
```

And where available `options` are:
```JS
{
    sort: boolean,
    endRotation: Vec3|Quat,
    findAll: boolean, // Not available if different end transform
    filterCollisionGroup: number,
    filterCollisionMask: number,
    filterTags: string[]|string[][], // Not available if different end transform
    filterCallback: (entity: Entity) => boolean, // Not available if different end transform
    destroyShape: boolean // Forced true when shape is a configuration
}
```

Fixes #2094


src/framework/components/rigid-body/system.js
```diff
@@ -552,6 +560,240 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+     * detected, the returned array will be of length 0.
+     *
+     * @param {Ammo.btCollisionShape} shape - The Ammo.btCollisionShape to use for collision check.
+     * @param {Vec3} position - The world space position for the shape to be.
+     * @param {Vec3} rotation - The world space rotation for the shape to have.
+     *
+     * @returns {RaycastResult[]} An array of spherecast hit results (0 length if there were no hits).
+     * @private
+     */
+    _shapecast(shape, position, rotation) {
```
willeastcott, line 723: Am I right in saying the `shape` instance is not currently being destroyed anywhere?

src/framework/components/rigid-body/system.js
```diff
@@ -552,6 +560,240 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
     }
 
+    /**
+     * Perform a collision check on the world and return all entities the shape hits.
+     * It returns an array of {@link RaycastResult}, one for each hit. If no hits are
+     * detected, the returned array will be of length 0.
+     *
+     * @param {object} shape - The shape to use for collision.
+     * @param {number} shape.axis - The local space axis with which the capsule, cylinder or cone shape's length is aligned. 0 for X, 1 for Y and 2 for Z. Defaults to 1 (Y-axis).
+     * @param {Vec3} shape.halfExtends - The half-extents of the box in the x, y and z axes.
```
willeastcott, line 570: The engine tends to use `halfExtents` instead of `halfExtends`. Search the source to see what I mean.

src/framework/components/rigid-body/system.js
MushAsterion: Yes it is not destroyed. I don't have full knowledge over Ammo.js so I don't know if it will affect the btRigidBody if destroyed as the btRigidBody is never destroyed to avoid re-creating one every call (and is only created when calling it for first time).

src/framework/components/rigid-body/system.js
MushAsterion: My bad. Fixed in last commit!

src/framework/components/rigid-body/system.js
```diff
@@ -552,6 +560,240 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+     * @param {number} shape.height - The total height of the capsule, cylinder or cone from tip to tip.
+     * @param {string} shape.type - The type of shape to use. Available options are "box", "capsule", "cone", "cylinder" or "sphere". Defaults to "box".
+     * @param {number} shape.radius - The radius of the sphere, capsule, cylinder or cone.
+     * @param {Vec3} position - The world space position for the shape to be.
+     * @param {Vec3} rotation - The world space rotation for the shape to have.
+     *
+     * @returns {RaycastResult[]} An array of shapecast hit results (0 length if there were no hits).
+     */
+    shapeCast(shape, position, rotation) {
+        Debug.assert(Ammo.ConcreteContactResultCallback, 'pc.RigidBodyComponentSystem#shapeCast: Your version of ammo.js does not expose Ammo.ConcreteContactResultCallback. Update it to latest.');
```
willeastcott, line 580: You could just have this assert call once in `_shapecast` maybe?

src/framework/components/rigid-body/system.js
willeastcott: I'm not 100% sure what the best solution is here, but at the moment, the code would definitely leak the created shape on every cast. How about doing:

```
shapecastBody.setCollisionShape(null);
Ammo.destroy(shape);
```

src/framework/components/rigid-body/system.js
MushAsterion: That's what I wanted initially but the problem is that the log output the name of the function and if it outputs to the user the name of a private function or a function they didn't called they will never find where it comes from.

src/framework/components/rigid-body/system.js
```diff
@@ -552,6 +560,240 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+     * @param {object} shape - The shape to use for collision.
+     * @param {number} shape.axis - The local space axis with which the capsule, cylinder or cone shape's length is aligned. 0 for X, 1 for Y and 2 for Z. Defaults to 1 (Y-axis).
+     * @param {Vec3} shape.halfExtents - The half-extents of the box in the x, y and z axes.
+     * @param {number} shape.height - The total height of the capsule, cylinder or cone from tip to tip.
+     * @param {string} shape.type - The type of shape to use. Available options are "box", "capsule", "cone", "cylinder" or "sphere". Defaults to "box".
+     * @param {number} shape.radius - The radius of the sphere, capsule, cylinder or cone.
+     * @param {Vec3} position - The world space position for the shape to be.
+     * @param {Vec3} rotation - The world space rotation for the shape to have.
+     *
+     * @returns {RaycastResult[]} An array of shapecast hit results (0 length if there were no hits).
```
willeastcott, line 577: It's not clear to me what it means to raycast with a shape that has volume. For a ray, the point is the point of ray penetration and the normal is the normal on the surface at that point. What does it mean for a shape? And is the `Raycast` naming a bit confusing?

src/framework/components/rigid-body/system.js
MushAsterion: I just tested out and it works. Just pushed a new commit to fix it.

willeastcott: Nice, thanks.

src/framework/components/rigid-body/system.js
MushAsterion: When using `dynamicsWorld.contactTest` the callback gives only 1 point per collision. But that's true that it may confuse users. Regarding `RaycastResult` being returned it's because it has the same structure. Another solution could be to directly return an array of entities?

src/framework/components/rigid-body/system.js
```diff
@@ -552,6 +560,242 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+     */
+    shapeCast(shape, position, rotation) {
```
willeastcott, line 579: So this is a utility function that basically calls the shape-specific function. For the purpose of maintaining a nice and accessible API, should we:

1. just retain this 'all-in-one' function as public
2. just retain the shape-specific function as public
3. keep all public

MushAsterion: The advantage of "all-in-one" function is that you can actually use a JSON format for all the shape config meaning it can easily be assigned from a script attribute to edit it within the editor directly and avoid having to be only code-based. In another hand the shape-specific function allow users used to other engines to find the function easily (especially `sphereCast` which is really common).

willeastcott: Hmm, that's a great point. OK, fair enough. 😄

src/framework/components/rigid-body/system.js
mvaligursky: I'd definitely move it to `_shapecast` - the callstack would show the users where they called it from. The assert shows the whole callstack (when you expand it)

src/framework/components/rigid-body/system.js
```diff
@@ -552,6 +560,242 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+     * @param {number} axis - The local space axis with which the capsule's length is aligned. 0 for X, 1 for Y and 2 for Z. Defaults to 1 (Y-axis).
+     * @param {Vec3} position - The world space position for the capsule to be.
+     * @param {Vec3} rotation - The world space rotation for the capsule to have.
+     *
+     * @returns {RaycastResult[]} An array of capsulecast hit results (0 length if there were no hits).
+     */
+    capsuleCast(radius, height, axis, position, rotation) {
+        Debug.assert(Ammo.ConcreteContactResultCallback, 'pc.RigidBodyComponentSystem#capsuleCast: Your version of ammo.js does not expose Ammo.ConcreteContactResultCallback. Update it to latest.');
+
+        if (axis === 0) {
```
mvaligursky, line 629: I would try to keep variables of constant type .. you pass in a number, and here you assign string to it. I'd prefer you use new variable for string instead. I would probably just have an array with these 3 strings as a global array variable, and simply index into it here. And assert if the lookup is undefined.

src/framework/components/rigid-body/system.js
```diff
@@ -552,6 +560,242 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+            axis = 'btCapsuleShapeX';
+        } else if (axis === 2) {
+            axis = 'btCapsuleShapeZ';
+        } else {
+            axis = 'btCapsuleShape';
+        }
+
+        return this._shapecast(new Ammo[axis](radius, height), position, rotation);
```
mvaligursky, line 637: just double checking here that this allocated object here does not need to be somehow disposed of.

src/framework/components/rigid-body/system.js
```diff
@@ -552,6 +560,242 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+
+    /**
+     * Perform a collision check on the world and return all entities the cylinder hits.
+     * It returns an array of {@link RaycastResult}, one for each hit. If no hits are
+     * detected, the returned array will be of length 0.
+     *
+     * @param {number} radius - The radius of the cylinder.
+     * @param {number} height - The total height of the cylinder from tip to tip.
+     * @param {number} axis - The local space axis with which the cylinder's length is aligned. 0 for X, 1 for Y and 2 for Z. Defaults to 1 (Y-axis).
+     * @param {Vec3} position - The world space position for the cylinder to be.
```
mvaligursky, line 675: it seems both position and rotation is optional, so use []. For example 
```
@param {Vec3} [position] - The world ...
```

The question is - should at least position be required?

src/framework/components/rigid-body/system.js
```diff
@@ -552,6 +560,242 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+     * detected, the returned array will be of length 0.
+     *
+     * @param {Ammo.btCollisionShape} shape - The Ammo.btCollisionShape to use for collision check.
+     * @param {Vec3} position - The world space position for the shape to be.
+     * @param {Vec3} rotation - The world space rotation for the shape to have.
+     *
+     * @returns {RaycastResult[]} An array of spherecast hit results (0 length if there were no hits).
+     * @private
+     */
+    _shapecast(shape, position, rotation) {
```
mvaligursky, line 723: maybe provide defaults here so you don't have to handle them later
```
_shapecast(shape, position = Vec3.ZERO, rotation = Vec3.ZERO)
```

mvaligursky:
Nice PR! Two comments:
- does it work with the ammo version PlayCanvas uses by default? If not, we should update the version at least in the examples: https://github.com/playcanvas/engine/tree/main/examples/src/lib/ammo

- I'd love to have an example for this here in the physics section:  https://github.com/playcanvas/engine/tree/main/examples - could be done in a separate PR as well, but it'd be awesome to have a test we can run before each engine release.

src/framework/components/rigid-body/system.js
willeastcott: This is destroyed yes. See the call:

```
Ammo.destroy(shape);
```

...in `_shapecast`.

src/framework/components/rigid-body/system.js
willeastcott: Ah, yeah. I'd agree with that. Best to keep the type of a variable the same as it flows through the code.

src/framework/components/rigid-body/system.js
mvaligursky: ah yes, ok .. bit uncommon to destroy temporary object in a different function, but makes sense.

willeastcott: Yeah, I agree. But in this case, it'd make the caller functions a fair bit more verbose if the destruction was done there. So I guess it's ok...

src/framework/components/rigid-body/system.js
MushAsterion: Alright if it does then yeah it makes sense. Just added to last commit!

src/framework/components/rigid-body/system.js
MushAsterion: Thanks for the feedback, it's fixed with last commit!

src/framework/components/rigid-body/system.js
MushAsterion: That's added!

MushAsterion:
Thanks! And thank you for your review.

Regarding your questions @mvaligursky,

> * does it work with the ammo version PlayCanvas uses by default? If not, we should update the version at least in the examples: https://github.com/playcanvas/engine/tree/main/examples/src/lib/ammo

I tested it within the production version from online editor and re-checked after with engine-only to make sure it was working. Here is the project I made to do it: https://playcanvas.com/project/1034600.

> * I'd love to have an example for this here in the physics section:  https://github.com/playcanvas/engine/tree/main/examples - could be done in a separate PR as well, but it'd be awesome to have a test we can run before each engine release.

Sure, I can make something similar to the raycast example, however to be completely honest with you I'm far from being the best on creating full project.

src/framework/components/rigid-body/system.js
```diff
@@ -552,6 +560,222 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+        ammoTransform.setOrigin(ammoVec3);
+        ammoTransform.setRotation(ammoQuat);
+
+        // We only initialize the shapecast body here so we don't have an extra body unless the user uses this function
+        if (!shapecastBody) {
+            shapecastBody = this.createBody(0, shape, ammoTransform);
+        }
+
+        // Add the body to the world.
+        this.addBody(shapecastBody, BODYGROUP_TRIGGER, BODYMASK_ALL);
```
willeastcott, line 731: Check this page out:

https://andysomogyi.github.io/mechanica/bullet.html#contact-text

It says:

> The contactTest query will peform a collision test against all overlapping objects in the world, and produces the results using a callback. The query object doesn’t need to be part of the world.

Is that true?

willeastcott:
@MushAsterion I think it's OK to skip the example if you're not confident about putting one together. We can maybe open another issue requesting that...

src/framework/components/rigid-body/system.js
MushAsterion: Ah yeah my bad, I added it as I had some issue during my testings and forgot to test without. After testing it works well with the body outside of world. Updated!

src/framework/components/rigid-body/system.js
MushAsterion: Alright so yeah I just removed the RaycastResult and now it returns an array of entities instead. Does that sound better to you @willeastcott?

src/framework/components/rigid-body/system.js
```diff
@@ -552,6 +560,211 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+        } else if (shape.type === 'cone') {
+            return this.coneCast(shape.radius, shape.height, shape.axis, position, rotation);
+        } else if (shape.type === 'cylinder') {
+            return this.cylinderCast(shape.radius, shape.height, shape.axis, position, rotation);
+        } else if (shape.type === 'sphere') {
+            return this.sphereCast(shape.radius, position, rotation);
+        }
+
+        return this.boxCast(shape.halfExtents, position, rotation);
+    }
```
ertugrulcetin, line 591: switch case could be used here IMHO

ertugrulcetin:
```js
switch (shape.type) {
  case 'capsule':
    return this.capsuleCast(shape.radius, shape.height, shape.axis, position, rotation);
  case 'cone':
    return this.coneCast(shape.radius, shape.height, shape.axis, position, rotation);
  case 'cylinder':
    return this.cylinderCast(shape.radius, shape.height, shape.axis, position, rotation);
  case 'sphere':
    return this.sphereCast(shape.radius, position, rotation);
  default:
    return this.boxCast(shape.halfExtents, position, rotation);
}

```

MushAsterion: Ah yeah, after test on mesurethat.com it seems that it's slightly more performant, that's updated!

src/framework/components/rigid-body/system.js
```diff
@@ -552,6 +560,212 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+     *
+     * @param {Vec3} halfExtents - The half-extents of the box in the x, y and z axes.
+     * @param {Vec3} [position] - The world space position for the box to be.
+     * @param {Vec3} [rotation] - The world space rotation for the box to have.
+     *
+     * @returns {import('../../entity.js').Entity[]} An array of entities hit by this boxcast (0 length if there were no hits).
+     */
+    boxCast(halfExtents, position, rotation) {
+        ammoVec3.setValue(halfExtents.x, halfExtents.y, halfExtents.z);
+        return this._shapecast(new Ammo.btSphereShape(ammoVec3), position, rotation);
```
kungfooman, line 607:
```suggestion
        return this._shapecast(new Ammo.btBoxShape(ammoVec3), position, rotation);
```

MushAsterion: Nice spot, thank you!

src/framework/components/rigid-body/system.js
willeastcott: I'm not sure, @MushAsterion. It sort of depends on the meaning of the point + normal that was returned in the `RaycastResult` that was returned before. If Ammo is returning useful information there, we should pass it back to the application. Can you maybe look into that?

willeastcott:
Something else occurred to me. We have already have `raycastFirst` and `raycastAll`. Should this new API map to this? So we'd have:

* `boxCastFirst` and `boxCastAll`
* `coneCastFirst` and `coneCastAll`
* `capsuleCastFirst` and `capsuleCastAll`
* `cylinderCastFirst` and `cylinderCastAll`
* `shapeCastFirst` and `shapeCastAll`
* `sphereCastFirst` and `sphereCastAll`

willeastcott:
@MushAsterion By the way, I'm sorry if this review seems a bit overly rigorous and sometimes pedantic! Just want to make sure we get everything perfect as possible - but I'm super-excited to get this merged! 😄

src/framework/components/rigid-body/system.js
kungfooman: > If Ammo is returning useful information there, we should pass it back to the application. Can you maybe look into that?

The information is as crucial as it gets here, otherwise there is no way to tell where the collision occurred.

I agree `RaycastResult` is a little bit too specific, so either it could be renamed into `CastResult` (`CastingResult`?) or simply `class CastResult extends RaycastResult {}` e.g.

The last thing I want is just an array of entities without any kind of collision information.

willeastcott:
I have another question. I see how to specify the shape we want to cast and where to position/orient it initially. But how do you specify the direction and/or distance you want to cast the shape? Can you maybe give a code example showing, say, how you'd do a sphere cast from world space position [1, 2, 3] 10 units in the direction [1, 1, 1]?

src/framework/components/rigid-body/system.js
willeastcott: I'd probably go with `CastResult` or `HitResult`.

kungfooman:
> But how do you specify the direction and/or distance you want to cast the shape?

As far as I understand this, it just does "in place testing". I think what you are asking for could be achieved via:

```js
pc.app.systems.rigidbody.dynamicsWorld.convexSweepTest
```

src/framework/components/rigid-body/system.js
kungfooman: > I'd probably go with `CastResult` or `HitResult`

Nice, I like the sound of `HitResult` and it's even shorter :+1:

willeastcott:
> As far as I understand this, it just does "in place testing". I think what you are asking for could be achieved via:

Oh. Well, in that case, it's not casting, right? I thought casting is analogous to sweeping. I'd say names like `sphereTest` are more accurate than `sphereCast`, IMHO.

willeastcott:
I should point out that the original poster on the forum was asking for spherecasting that is analogous to raycasting. i.e. a swept sphere. And I should probably also point out that Unity's API does sweeping when using this naming. Check it out:

https://docs.unity3d.com/ScriptReference/Physics.SphereCast.html

src/framework/components/rigid-body/system.js
MushAsterion: That's added!

Regarding point + normal here are the two definitions I used:
- `HitResult#point` The point on the entity where the hit occurred, in world space. When returned by a shapecast it is the first point found to collide, it does not have to be the closest.
- `HitResult#normal` The normal vector of the hit on the entity, in world space.

MushAsterion:
That's a good point, I renamed the existing functions and am adding the proper functions for casting using sweeping then!

MushAsterion:
> @MushAsterion By the way, I'm sorry if this review seems a bit overly rigorous and sometimes pedantic! Just want to make sure we get everything perfect as possible - but I'm super-excited to get this merged! 😄

No worries, it's normal to make sure that no problems come with it as well as everything is working as expected.



> > As far as I understand this, it just does "in place testing". I think what you are asking for could be achieved via:
> 
> Oh. Well, in that case, it's not casting, right? I thought casting is analogous to sweeping. I'd say names like `sphereTest` are more accurate than `sphereCast`, IMHO.

I changed the existing functions to be "Test" and added `sphereCast` and `boxCast` to be based on starting and ending positions. However regarding other shapes it would be more complex (as it would require multiple shapes or to generate one) to add and I'm not sure if it would be needed, what do you think?

Also please note that due to how the shape casting works on Ammo the points and normal will be related to the  shape of which the middle is known by default for `shapeTest` but not `sphereCast`/`boxCast`, I'm not sure if there is an existing workaround to get first point from cast start point.

MushAsterion:
> > But how do you specify the direction and/or distance you want to cast the shape?
> 
> As far as I understand this, it just does "in place testing". I think what you are asking for could be achieved via:
> 
> ```js
> pc.app.systems.rigidbody.dynamicsWorld.convexSweepTest
> ```

Problem is that the current Ammo version doesn't expose the `btCollisionObject` on `ClosestConvexResultCallback` (for `shapeCastFirst`) and the `ConvexResultCallback` is not exposed at all (for `shapeCastAll`).

I however added `boxCast` and `sphereCast` function based on `shapeTest`. For other shapes it would require to expose more of the Bullet API...

src/framework/components/rigid-body/system.js
```diff
@@ -552,6 +600,275 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+    boxCast(halfExtents, start, end) {
+        // Sweeping
+        halfExtents.z += start.distance(end) / 2;
+
+        // Find rotation
+        shapecastRotationMatrix.setLookAt(start, end, Vec3.UP);
+        shapecastRotation.setFromMat4(shapecastRotationMatrix);
+
+        // Transform start vector to make it beween initial start and end.
+        start.add(end).divScalar(2);
```
kungfooman, line 623: Please don't change input vectors, otherwise vectors are messed up and everyone is scratching their head why the data in some other place changed :thought_balloon: :thinking: 

If I understand it right, it should be something like

1) On top of file a temp vector: `const startEndCenter = new Vec3();`
2) Instead of `start.add(end).divScalar(2);`: `startEndCenter.lerp(start, end, 0.5);`
3) Replace `start` with `startEndCenter` in `this._shapeTest` line

src/framework/components/rigid-body/system.js
```diff
@@ -552,6 +600,275 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+    sphereCast(radius, start, end) {
+        // Sweeping
+        const height = start.distance(end) + radius;
+
+        // Find rotation
+        shapecastRotationMatrix.setLookAt(start, end, Vec3.UP);
+        shapecastRotation.setFromMat4(shapecastRotationMatrix);
+
+        // Transform start vector to make it beween initial start and end.
+        start.add(end).divScalar(2);
```
kungfooman, line 648: Same issue here

src/framework/components/rigid-body/system.js
```diff
@@ -552,6 +600,275 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+        halfExtents.z += start.distance(end) / 2;
+
+        // Find rotation
+        shapecastRotationMatrix.setLookAt(start, end, Vec3.UP);
+        shapecastRotation.setFromMat4(shapecastRotationMatrix);
+
+        // Transform start vector to make it beween initial start and end.
+        start.add(end).divScalar(2);
+
+        return this._shapeTest(new Ammo.btBoxShape(halfExtents), start, shapecastRotation);
```
kungfooman, line 625: How does this work, don't you need to use `ammoVec3` instead of `halfExtents` here (`pc.Vec3` vs `Ammo.btVector3`) to create a `btBoxShape`?

src/framework/components/rigid-body/system.js
MushAsterion: Alright! That's updated

src/framework/components/rigid-body/system.js
MushAsterion: That's fixed, thank you

src/framework/components/rigid-body/system.js
MushAsterion: Oh yeah, my bad, had it locally but not on commit, that's updated!

kungfooman:
Thank you for your quick changes! I think now the only thing thats left is what @willeastcott wrote about, to achieve a unified API:

> `boxCastFirst` and `boxCastAll`

Since Ammo seems to have no proper sweeping API `boxCastFirst` needs to be implemented using a distance-squared-check of the `boxCastAll` results?

MushAsterion:
> Thank you for your quick changes! I think now the only thing thats left is what @willeastcott wrote about, to achieve a unified API:
> 
> > `boxCastFirst` and `boxCastAll`
> 
> Since Ammo seems to have no proper sweeping API `boxCastFirst` needs to be implemented using a distance-squared-check of the `boxCastAll` results?

I updated the functions to be `boxCastAll` and `sphereCastAll`, however I have no idea how Ammo found the points when performing shape casting therefore nothing proves the closest point from start will be the actual first result therefore I'm not confident in adding something that could lead in wrong results to the code. For it to be proper `boxCastFirst`/`sphereCastFirst` it will require real sweeping I think.

LeXXik:
Thank you for the PR!

Convex cast or collision tests are useful in kinematics, when the developer wants to know if the spatial variables (position/orientation) of the rigidbody are ok or we have a contact and we ought to do something. It is also often used for camera position test to prevent it from clipping through. Camera is using a sphere, which is fine, but character is usually a capsule, which we cannot do without upgrading Ammo (or unless we do some custom convex hull shape mumbo-jumbo, which we shouldn't). However, since we are offering `boxCastAll`, I would assume it is using a sweep test (from the name), and I would wonder how to orient the shape I am sweeping. Obviously we cannot change orientations with a collision test, so I would rename it to `boxTestAll` or something.

As for the first contact - testing the result distance from the start, suggested by @kungfooman, is a solid suggestion, and works fine. We use the method in production no problem. The closest contact to start is first in distance and. as a result, first in time (even if it was added to the contacts list sometime in the middle). You can add `boxTestFirst`, which does `boxTestAll` internally and simply finds the closest result from the list. Just add a comment for the future contributor that it should be changed to convex sweep test once Ammo version gets updated, as it would be more performant.

MushAsterion:
Alright thank you for your feedback.

I removed `boxCallAll` until Ammo get upgraded to include shape rotation.

Regarding the first result I don't see how would the engine know which tip my start point would be so the point be on the good side of a shape being in the middle of the cast. Do you have an example with it working on all directions?

src/framework/components/rigid-body/system.js
```diff
@@ -552,6 +601,275 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+                    ));
+                }
+            }
+        };
+
+        // Check for contacts.
+        this.dynamicsWorld.contactTest(shapeTestBody, resultCallback);
+
+        // Disable body and remove shape.
+        shapeTestBody.forceActivationState(BODYSTATE_DISABLE_DEACTIVATION);
```
LeXXik, line 838: Hmm. BODYSTATE_DISABLE_DEACTIVATION is used, because we don't have CF_NO_CONTACT_RESPONSE defined, and they both happen to equal to 4. The result will be the same, but I personally don't like a flag that was created for different purpose is used.

LeXXik:
> Regarding the first result I don't see how would the engine know which tip my start point would be so the point be on the good side of a shape being in the middle of the cast. Do you have an example with it working on all directions?

Ah, you are right. I recall now, I was adding some additional logic to filter the test. I think we can go without it, then. As long as we show all contacts, the developer will have an opportunity to select the one needed.

src/framework/components/rigid-body/system.js
LeXXik: Actually we do have it: BODYFLAG_NORESPONSE_OBJECT
Was it the flag you wanted to use? BODYSTATE_DISABLE_DEACTIVATION is used when we want to prevent the body from sleeping.

MushAsterion: I'm not sure to get it, how would a flag named BODYSTATE_DISABLE_DEACTIVATION would not be to set body state to disabled?

MushAsterion: Well I was trying to disable the shape, so I should use BODYFLAG_NORESPONSE_OBJECT?

LeXXik: Well, to be perfectly honest, I won't be able to tell for sure, unless I tinker with it myself :) Setting a state to BODYSTATE_DISABLE_DEACTIVATION would just prevent it from sleeping, it won't disable it.

You would have to test it, but I think it should go something like:
```
body.setCollisionFlags(body.getCollisionFlags() | BODYFLAG_NORESPONSE_OBJECT);
```

MushAsterion: Collision flag? Is it really useful to disable collision when body is removed from world and shape is removed?

LeXXik: Ah, I might have missed where the body is removed from the world. I thought only the shape was destroyed. If it is, then we don't need that line. It was the comment there `// Disable body and remove shape` that didn't match the code, which didn't disable the body. If the body is removed, that whole line can be removed then altogether. Actually, if the shape is removed, then there should be no collisions reported either.

MushAsterion: Interesting to know, thanks! I removed the line.

src/framework/components/rigid-body/system.js
```diff
@@ -227,6 +236,42 @@ class ContactResult {
     }
 }
 
+/**
+ * Object holding the result of a hit on an Entity.
+ */
+class HitResult {
```
willeastcott, line 242: This class looks identical to `RaycastResult` but with a different name. I would consider renaming `RaycastResult` to `HitResult` and adding the following to deprecated.js:

```
export const RaycastResult = HitResult;
```

willeastcott:
Sorry I've been a bit quiet this week - I've been travelling. Anyway, so it sounds like we would want to use this to do shape casting:

https://github.com/kripken/ammo.js/blob/main/ammo.idl#L888

You say:

> Problem is that the current Ammo version doesn't expose the btCollisionObject on ClosestConvexResultCallback (for shapeCastFirst) and the ConvexResultCallback is not exposed at all (for shapeCastAll).

Are you saying we need to grab the latest ammo.js build from here:

https://github.com/kripken/ammo.js/tree/main/builds

...replace the version currently added to the Editor? We can absolutely do that if this requires it.

My concern is that we deploy the current implementation of `sphereCastAll` that you have here, and then we update the implementation later to use `convexSweepTest` and the returned results differ somehow. I would rather we update ammo.js and have the initial implementation using `convexSweepTest` personally.

willeastcott:
Also, one other minor thing. Whenever you update any changes to public API in the PR, please update your original PR description to reflect the latest API state. It's really useful to be able to clearly see at a glance how a PR will alter the public API.

src/framework/components/rigid-body/system.js
MushAsterion: It's done!

MushAsterion:
> Are you saying we need to grab the latest ammo.js build from here:
> 
> https://github.com/kripken/ammo.js/tree/main/builds
> 
> ...replace the version currently added to the Editor? We can absolutely do that if this requires it.

I don't specially mean update the version as it will be a real breaking change as many functions changed since the version that you use. I only mean expose the property and class needed:
- `ClosestConvexResultCallback#m_hitCollisionObject` (`[Const] attribute btCollisionObject m_hitCollisionObject;`)
- `ConvexResultCallback`, problem is that the class is only meant to be an interface which means we have no real way to have all the results with the current version and from my tests changing the `addSingleResult` from `ClosestConvexResultCallback` doesn't change the results function...

> My concern is that we deploy the current implementation of `sphereCastAll` that you have here, and then we update the implementation later to use `convexSweepTest` and the returned results differ somehow. I would rather we update ammo.js and have the initial implementation using `convexSweepTest` personally.

I got the point and yes result order will probably differ as it will check for a full shape (capsule) at once instead of starting at a point and going to another.

> Also, one other minor thing. Whenever you update any changes to public API in the PR, please update your original PR description to reflect the latest API state. It's really useful to be able to clearly see at a glance how a PR will alter the public API.

It's updated, sorry for that

willeastcott:
> I don't specially mean update the version as it will be a real breaking change as many functions changed since the version that you use.

Really?? In the past, we have been able to occasionally update the version of Ammo available in the Editor to the latest build in the Ammo.js repo. It's always been backwards compatible until now. Are you completely sure that taking the latest build will cause breakage to existing projects?

MushAsterion:
Well when I was trying to use latest version on NodeJS it was showing me some errors, or maybe I did anything wrong on my end, possible as well, I'm far from being an expert... But yeah from the Bullet docs some classes doesn't exist anymore and it in fact include the `ConvexResultCallback`, unless Ammo is not matching completely the Bullet API?

MushAsterion:
Oh after looking at some other classes within the IDL file I found it's possible to expose the `addSingleResult` function from `ClosestConvexResultCallback` by adding:
`float addSingleResult([Ref] LocalConvexResult convexResult, bool normalInWorldSpace);`

This way it's possible by modifying the `idl` file only!

MushAsterion:
> Really?? In the past, we have been able to occasionally update the version of Ammo available in the Editor to the latest build in the Ammo.js repo. It's always been backwards compatible until now. Are you completely sure that taking the latest build will cause breakage to existing projects?

Ok so I was just idiot and importing it the wrong way. It's in fact all backwards compatible... 🤦

MushAsterion:
Added all `shapeCastFirst` and linked primitive shape function using `convexSweepTest` with proper assertion on `_shapeCastFirst`. All it needs now is exposure of `Ammo.ClosestConvexResultCallback.#get_m_hitCollisionObject` for it to work.

willeastcott:
> Ok so I was just idiot and importing it the wrong way. It's in fact all backwards compatible and works well with latest build... 🤦

LOL. That's great that you managed to implement 'true' shape casting. So you now have the 'First' functions defined and I have a couple of questions:

1. Did you want to also provide 'All' functions (that return all hits)?
2. From an API perspective, would it be better to have `shapeCastFirst` and `shapeCastAll` or to have `shapeCast(arg1...argN, findAll = false)`?
3. Anything else you think is important to include in this PR before it gets merged?

It feels like we're getting very close with this now! 🚀

MushAsterion:
1. It would be nice but will require a custom `idl` as original `idl` from [kripken/ammo.js](https://github.com/kripken/ammo.js) does not expose `ConvexResultCallback` constructor and `ConvexResultCallback#addSingleResult` that are required to perform a `CastAll`.
2. As you mentionned in a previous comment the `First`/`All` match the existing names for `raycastFirst`/`raycastAll` so I guess it's better this way. But for sure if we get to implement a `All` version for `sphereCast` and a `First` version for `sphereTest` it can still be used to reduce number of lines and keep the `First`/`All` functions for consistency.
3. Only the `shapeCastAll` function if we go for it from (1)

willeastcott:
Sorry to be annoying after you've made the change @MushAsterion, but `raycast` is a word, whereas `boxcast` is not. I would stick with `boxCast` etc. We _could_ alias `raycast` to `rayCast`....but I'm not entirely sure we should since, like I say, it is actually a word. I do also note that Unity used a [small 'c' for Raycast](https://docs.unity3d.com/ScriptReference/Physics.Raycast.html) and a [capital 'C' for BoxCast](https://docs.unity3d.com/ScriptReference/Physics.BoxCast.html).

MushAsterion:
Oh alright my bad. That's reverted.

MushAsterion:
Hello @willeastcott,
I'm getting back to this PR,

I added all the `shapeTestFirst` functions which work based on the first returned result, while there is no documentation on this result being the closest it still provide users with a computation that is slightly faster than the `All` counterpart (shorter on PC side, Ammo is still computing all results).

I also went ahead and tested for `shapeCastAll` however it does not work by updating the `Ammo.ClosestConvexResultCallback#addSingleResult` and as `Ammo.ConvexResultCallback` is only a virtual class there is no real class provided to handle this situation making it impossible to add.

MushAsterion:
Added support for similar options as raycast: sorting (#5179 #5181) and filtering (#5180)

src/framework/components/rigid-body/system.js
```diff
@@ -3,33 +3,39 @@ import { ObjectPool } from '../../../core/object-pool.js';
...
+import { BODYFLAG_NORESPONSE_OBJECT, BODYSTATE_ACTIVE_TAG } from './constants.js';
 import { RigidBodyComponent } from './component.js';
 import { RigidBodyComponentData } from './data.js';
 
-let ammoRayStart, ammoRayEnd;
+// Ammo.js variable for performance saving.
+let ammoRayStart, ammoRayEnd, ammoVec3, ammoQuat, ammoTransform, ammoTransform2;
+
+// RigidBody for shape tests. Permanent to save performance.
+let shapeTestBody;
```
MAG-AdrianMeredith, line 19: Any concerns with what happens if two instances of playcanvas are on the same page?  They'll share the same objects

MushAsterion: Many other files contain global variables to the module, I don't see the difference or problem here considering those variables are initialized once then every time a value is assigned to it, it is immediately used.

Do you suggest it could conflict with anything? And if so, why would it conflict here and not in any other module?

MAG-AdrianMeredith: We previously had issues with them hanging around an not being cleaned up properly.  More of a general point to be careful about module scope variables as they live forever and are shared between different instances.  I think in this case they are purely temporary variables so it should be fine but I thought it worth pointing out just in case

OlegGedzjuns:
Unfortunately tested only sphere cast
Definitely some test and example projects are needed

mvaligursky:
This PR is waiting for up to date ammo .. I wonder if somebody is keen to compile the wasm we need here?

MushAsterion:
> This PR is waiting for up to date ammo .. I wonder if somebody is keen to compile the wasm we need here?

You can simply use latest version from [kripken/ammo.js](https://github.com/kripken/ammo.js/tree/main)

OlegGedzjuns:
> > This PR is waiting for up to date ammo .. I wonder if somebody is keen to compile the wasm we need here?
> 
> You can simply use latest version from [kripken/ammo.js](https://github.com/kripken/ammo.js/tree/main)

Tested just now, seems good

MushAsterion:
Editor must be updated to latest version of Ammo.js, I was multiple times mentioned by @willeastcott about it coming but still waiting for it.

willeastcott:
@MushAsterion I just chatted to @slimbuck about this. He's busy working on a SuperSplat update at the moment but he'll deploy the latest ammo.js as soon as he's able.

src/framework/components/rigid-body/system.js
willeastcott: I wanted to unresolve this discussion and talk about it some more, @MushAsterion. I do still feel that this PR introduces a lot of additions to the API. 18 methods at the moment. And if `shapeCastAll` flavors get added, it takes it to 24. Plus the two existing raycasting functions, that'd be 26 functions for testing and casting. I think that's just too much. We _could_ just have:

```
raycast (we could deprecate raycastFirst and raycastAll - note we don't _have_ to do that in this PR)
shapeCast
shapeTest
```

The all versus first could just be an option (maybe defaulting to first).

It just feels to me that the JSON approach is interesting, but it's very niche and, in reality, the vast majority of developers won't call the API like that.

willeastcott: You could potentially also drop `shapeTest` if start and end pos/rot of `shapeCast` where the same too maybe?

MushAsterion: I agree that taking the type of result (ie "all" and "first") could be an option, but wouldn't that actually mess with IntelliSense and other suggestion systems? I mean it could be declared through a .d.ts file but I'm far from an expert in this field and not sure how it would behave at build time...

Honestly the point of the `shapeCast`/`shapeTest` with JSON was just to provide an "all-in-one" solution but that might not be for everyone as you mentioned especially since most of the API is high level, hence the function for each type separately. I tried to follow along with other exposed functions of the API and looked at some other engine to find the best solution and actually most engines only have sphere and sometimes box hence not a big deal to have them separated.

> You could potentially also drop `shapeTest` if start and end pos/rot of `shapeCast` where the same too maybe?

I'm definitely not sure on how Ammo threat such case but I feel like testing whether the values are equal to run another function could impact performance through that extra check at engine-level for a project-level issue? I mean position is always a Vec3 making it easy but rotation accepts Vec3 and Quat making it more conversion before comparing

MushAsterion: Back on making "all" and "first" an option wouldn't that confuse people that only "first" is supported for `shapeCast`?

willeastcott:
@MushAsterion We just checked our current Ammo version. It has both `convexSweepTest` and `ClosestConvexResultCallback`. Is there anything else we might be missing?

MushAsterion:
> @MushAsterion We just checked our current Ammo version. It has both `convexSweepTest` and `ClosestConvexResultCallback`. Is there anything else we might be missing?

Not too sure, actually checking from my previous edits of original post it requires exposure of:

> btCollisionObject on ClosestConvexResultCallback (for shapeCastFirst) and the ConvexResultCallback constructor (for shapeCastAll)

src/framework/components/rigid-body/system.js
MushAsterion: > I don't see how. The `playcanvas.d.ts` TypeScript declarations would specify the option and Intellisense would check it.
My concern is not about the options type but the returned value type as it returns `HitResult|null` for first values and `HitResult[]` for all. As as my knowledge of JSDoc it's not possible to document as-is and required a .d.ts override to declare a function override for any documentation to pick correctly.

> Another suggestion would be to simply not specify and end pos/rot. This would suggest that it's a test, not a cast.
That would imply setting end position/rotation as an option, do you think it would be intuitive enough? Also it would only work again on the "first" result for shape casting. Do you think documentation would be enough to guide people?

daredevildave: Just a few thoughts:

1. Does shapeTestFirst even make sense? My understanding of this is that it is a point in time test for all intersections with the shape. Which result is "first" doesn't seem intuitive, what defines the order? If this is the case, then I'd suggest just dropping First/All methods and always return an array.
2. It looks like shapeCastAll doesn't exist, is this intentional?
3. I think the distinction between raycastFirst and raycastAll is historical because the original version of Ammo didn't support all results. However, it does force the user to make the distinction because (I think) they primarily only want the first result. So I actually don't mind them being separate methods. If they are combined my preference would probably be for a raycast() method would be: a) always return an array of results b) make abort-on-first-hit the default behavior, which does have the downside of making the default return always a length 1 array, but I think provides overall consistency. You don't want people having to check if the return result is null, object or array.

MushAsterion: 1. For `shapeTestFirst` the first result is the closest to the position of the shape. Which is not always the first returned by `shapeTestAll`. Always returning an array would make it a different experience from `raycastFirst`. But if standardizing everything it could actually make sense.

2. `shapeCastAll` requires additional exposure of the Ammo.js library, probably the reason why it was not directly added to the PR. I'm actually not even sure to have managed to make it in any of my builds, but if you manage to make it feel free to contribute or share the knowledge.

3. For sure making both results an array would make everything easier on certain projects but will it be clear for everyone without deep knowledge especially when your option is set to return first result only but the function actually returns an array of length = 1|0 instead of a single object or `null` ?

daredevildave: Right, so shapeTestFirst is actually doing more work than shapeTestAll (because it gets all contacts, sorts them and then discards all but the first), which is counter-intuitive if you understand what raycastFirst/All is doing. 

Do you think that use-case is common enough to have an explicit engine method? We could just support shapeTestAll and have the user do the sort/discard if they only need the closest result.

daredevildave: And by the way, thank you for this PR! I appreciate the hard work! I found it, because I want to use shape casting in my project.

MushAsterion: `shapeTestFirst` is doing more work due to how Bullet/Ammo work and completely different from raycasting as raycasting is only a question of point while the closest point in a shape's position is not always the first to be detected since it depends on how the physics engine perform the test.

Once again the exposure of First/All methods is to match with what the engine has already implemented. I'm an external contributor so I don't want to step or change core project features. I personally find it handy for beginners to have the support for sure but as @willeastcott mentioned it's a very long list of new public methods which might also bring more confusion. I think it's a good solution to merge as much as possible to reduce public methods but I'm concerned about how easy it would be for people that are just looking to make a quick project without trying to get a deep understanding of the engine's API... I'll surely update to fit whatever they think is best anyway, the core features of the PR are already done and updating them is not a big deal

LeXXik: Good stuff! Many useful features are being added. Although, I do agree that the API is too verbose. We can simplify those, and move most of the customization stuff to options.

My vote goes for dropping -all/-first from the method name, and use an option setting. 

`shapeTestFirst` seems a bit superfluous. We can just return everything and let user decide what to do with the results. If future Ammo adds a feature to allow collecting only the first intersection (which sounds strange to me), we can add it via an option. Shape and point collision tests are usually useful for getting all colliders they intersect, otherwise a sweep test is used.

About shape target rotation during sweep. Personally, I've never met a need to use it. Maybe it is useful in robotics, where Bullet engine is popular? No idea. Most of the other other physics engines don't allow it (Rapier, PhysX, Jolt). As such, I would actually move it to options as well.

Whatever the names are, but signatures could look something like:
```
raycast(start, end, opts = {});

shapecast(shape, pos, rot, dir = Vec3.ZERO, opts = {});
```
If `dir` is a zero vector, we do shape test. Otherwise it is a non-normalized cast direction (its normal for direction, its magnitude as cast distance). Shape properties can be part of the options, e.g. radius for radius-based shapes.

About destroying a shape used in sweep. If a shape is not auto-destroyed (e.g. via some option), then there should be a method to destroy it later.

MushAsterion: I'm not sure to understand the point of changing `endPosition` into `dir` while `raycast` has a `end` param, is there any particular reason to make it different?

Something like
```JS
shapecast(shape, start, end, opts = {})
```

Where options could have something like

```JS
{
    startRotation: Vec3|Quat,
    endRotation: Vec3|Quat,
    findAll: boolean, // Defaults to false, not possible if end transform !== start transform
    // ... other options
}
```

Would be more fitting with the suggested `raycast` signature.

LeXXik: Yep, something like that. I think starting rotation is probably not optional, as it is often used for non-spheres. Target rotation can be optional.

MushAsterion: Alright so based on feedback I just rewrote the PR. Let me know if it better fits now 🙃

src/framework/components/rigid-body/system.js
```diff
@@ -504,70 +544,9 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
-        Ammo.destroy(rayCallback);
-
-        return result;
-    }
-
-    /**
-     * Raycast the world and return all entities the ray hits. It returns an array of
-     * {@link RaycastResult}, one for each hit. If no hits are detected, the returned array will be
-     * of length 0. Results are sorted by distance with closest first.
+     * Raycast the world and return the entities the ray hits. It returns an array of
```
LeXXik, line 547: Please, use a 100 symbols per limit.

src/framework/components/rigid-body/system.js
```diff
@@ -580,44 +559,49 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
      * // where hit entity has a `camera` component
-     * const hits = this.app.systems.rigidbody.raycastAll(new Vec3(0, 2, 2), new Vec3(0, -2, -2), {
-     *     filterCallback: (entity) => entity && entity.camera
+     * const hits = this.app.systems.rigidbody.raycast(new Vec3(0, 2, 2), new Vec3(0, -2, -2), {
+     *     filterCallback: (entity) => entity && entity.camera,
+     *     findAll: true
      * });
      * @example
-     * // Return all results of a raycast between 0, 2, 2 and 0, -2, -2
+     * // Return the closest result of a raycast between 0, 2, 2 and 0, -2, -2
```
LeXXik, line 585: If those are vectors, we should wrap the elements in brackets, e.g. (0, 2, 2)

LeXXik (approved):
LGTM with minor comments

src/framework/components/rigid-body/system.js
MushAsterion: Done

src/framework/components/rigid-body/system.js
MushAsterion: Done

LeXXik (approved):

LeXXik:
Can we move forward with this? If something is impeding the merge, please state it, so we can do changes.

MushAsterion:
Hey @LeXXik, as far as I'm concerned, the only missing part were examples, which I previously indicated not being an expert and leaving to other contributors. I've just updated from main and can confirm it's now free from conflicts. Not sure what are engine changes for the past 3 years and if it still fits.

src/index.js
```diff
@@ -282,7 +282,7 @@ export { RenderComponent } from './framework/components/render/component.js';
 export { RenderComponentSystem } from './framework/components/render/system.js';
 export * from './framework/components/rigid-body/constants.js';
 export { RigidBodyComponent } from './framework/components/rigid-body/component.js';
-export { RigidBodyComponentSystem, ContactPoint, ContactResult, RaycastResult, SingleContactResult } from './framework/components/rigid-body/system.js';
+export { RigidBodyComponentSystem, ContactPoint, ContactResult, HitResult, SingleContactResult } from './framework/components/rigid-body/system.js';
```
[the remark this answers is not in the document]
MushAsterion: Incorrect, Raycast result is exported from `export * from './deprecated/deprecated.js';`

src/framework/components/rigid-body/system.js
```diff
@@ -577,120 +614,68 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+    raycast(start, end, options = {}) {
+        Debug.assert(Ammo.AllHitsRayResultCallback, 'pc.RigidBodyComponentSystem#raycast: Your version of ammo.js does not expose Ammo.AllHitsRayResultCallback. Update it to latest.');
         const results = [];
 
         ammoRayStart.setValue(start.x, start.y, start.z);
         ammoRayEnd.setValue(end.x, end.y, end.z);
-        const rayCallback = new Ammo.AllHitsRayResultCallback(ammoRayStart, ammoRayEnd);
+
+        const findAll = options.findAll || options.filterTags || options.filterCallback;
+        const rayCallback = findAll ? new Ammo.AllHitsRayResultCallback(ammoRayStart, ammoRayEnd) : new Ammo.ClosestRayResultCallback(ammoRayStart, ammoRayEnd);
```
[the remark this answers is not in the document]
MushAsterion: Applied.

src/framework/components/rigid-body/system.js
```diff
@@ -701,40 +686,272 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+     * "capsule", "cone", "cylinder" or "sphere".
+     * @param {number} [shape.radius] - The radius of the sphere, capsule, cylinder or cone.
+     * @param {Vec3} startPosition - The world space position for the shape to be at start.
+     * @param {Vec3|Quat} [startRotation] - The world space rotation for the shape to have
+     * at start.
+     * @param {Vec3} [endPosition] - The world space position for the shape to be at end.
+     * @param {object} [options] - The additional options for the shape casting.
+     * @param {Vec3|Quat} [options.endRotation] - The world space rotation for the shape to have
+     * at end.
+     * @param {boolean} [options.sort] - Whether to sort raycast results based on distance with
```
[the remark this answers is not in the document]
MushAsterion: Applied

src/deprecated/deprecated.js
```diff
@@ -931,3 +932,13 @@ RigidBodyComponentSystem.prototype.setGravity = function () {
...
 };
+
+RigidBodyComponentSystem.prototype.raycastFirst = function (start, end, options) {
+    Debug.deprecated('pc.RigidBodyComponentSystem#raycastFirst is deprecated. Use pc.RigidBodyComponentSystem#raycast instead.');
+    return this.raycast(start, end, options)[0] ?? null;
+};
+
+RigidBodyComponentSystem.prototype.raycastAll = function (start, end, options = {}) {
+    Debug.deprecated('pc.RigidBodyComponentSystem#raycastAll is deprecated. Use pc.RigidBodyComponentSystem#raycast with findAll option instead.');
+    return this.raycast(start, end, Object.assign(options, { findAll: true }));
```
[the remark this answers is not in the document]
MushAsterion: Applied.

src/framework/components/rigid-body/system.js
```diff
@@ -701,40 +686,272 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+            // Destroy Ammo variables.
+            Ammo.destroy(resultCallback);
+            if (options.destroyShape) {
+                Ammo.destroy(btShape);
+            }
+
+            return results;
+        }
+
+        Debug.assert(Ammo.ConcreteContactResultCallback, 'pc.RigidBodyComponentSystem#shapeCest: Your version of ammo.js does not expose Ammo.ConcreteContactResultCallback. Update it to latest.');
```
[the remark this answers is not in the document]
MushAsterion: Applied.

src/framework/components/rigid-body/system.js
```diff
@@ -701,40 +686,272 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+                options.destroyShape = true;
+                break;
+            case 'cone':
+                btShape = createShape('Cone', shape.axis, shape.radius ?? 0.5, shape.height ?? 1);
+                options.destroyShape = true;
+                break;
+            case 'cylinder':
+                btShape = createShape('Cylinder', shape.axis, shape.radius ?? 0.5, shape.height ?? 1);
+                options.destroyShape = true;
+                break;
```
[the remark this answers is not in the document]
MushAsterion: Applied.

src/framework/components/rigid-body/system.js
```diff
@@ -701,40 +686,272 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+            case 'sphere':
+                btShape = new Ammo.btSphereShape(shape.radius ?? 0.5);
+                options.destroyShape = true;
+                break;
```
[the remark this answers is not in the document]
MushAsterion: Applied.

src/framework/components/rigid-body/system.js
```diff
@@ -701,40 +686,272 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+            case 'box':
+                ammoVec3.setValue(shape.halfExtents?.x ?? 0.5, shape.halfExtents?.y ?? 0.5, shape.halfExtents?.z ?? 0.5);
+
+                btShape = new Ammo.btBoxShape(ammoVec3);
+                options.destroyShape = true;
+                break;
+            case 'capsule':
+                btShape = createShape('Capsule', shape.axis, shape.radius ?? 0.5, shape.height ?? 1);
+                options.destroyShape = true;
+                break;
```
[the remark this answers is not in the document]
MushAsterion: Applied.

src/framework/components/rigid-body/system.js
```diff
@@ -701,40 +686,272 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+            default:
+                btShape = shape;
+                break;
+        }
+
+        ammoVec3.setValue(startPosition.x, startPosition.y, startPosition.z);
+        if (startRotation instanceof Quat) {
+            ammoQuat.setValue(startRotation.x, startRotation.y, startRotation.z, startRotation.w);
+        } else if (startRotation instanceof Vec3) {
+            ammoQuat.setEulerZYX(startRotation.z, startRotation.y, startRotation.x);
```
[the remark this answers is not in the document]
MushAsterion: Applied.

src/framework/components/rigid-body/system.js
```diff
@@ -701,40 +686,272 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+        const endRotation = options.endRotation;
+        if ((endPosition && !endPosition.equals(startPosition)) || endRotation) {
+            Debug.assert(Ammo.ClosestConvexResultCallback && Ammo.ClosestConvexResultCallback.get_m_hitCollisionObject, 'pc.RigidBodyComponentSystem#shapeCast: Your version of ammo.js does not expose Ammo.ClosestConvexResultCallback or Ammo.ClosestConvexResultCallback#get_m_hitCollisionObject. Update it to latest.');
+
+            ammoVec3.setValue(endPosition.x, endPosition.y, endPosition.z);
+            if (endRotation instanceof Quat) {
+                ammoQuat.setValue(endRotation.x, endRotation.y, endRotation.z, endRotation.w);
+            } else if (endRotation instanceof Vec3) {
+                ammoQuat.setEulerZYX(endRotation.z, endRotation.y, endRotation.x);
+            } else {
```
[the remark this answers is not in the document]
MushAsterion: Moved to options.

src/framework/components/rigid-body/system.js
```diff
@@ -701,40 +686,272 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+     * Defaults to false, forced true when shape is not a btCollisionShape.
+     * @param {boolean} [options.findAll] - Whether to return all results. When false will return
+     * only the closest result. Defaults to false. Only available if shape is not having different
+     * ending transform from starting one.
+     *
+     * @returns {HitResult[]} An array of shapeCast hit results (0 length if there were no hits).
+     */
+    shapeCast(shape, startPosition, startRotation, endPosition, options = {}) {
+        const results = [];
+
```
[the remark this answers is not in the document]
MushAsterion: Moved to options.

MushAsterion:
Actually, added an example, similar to raycast's.

ezgif-82b978c670ab3d0b

src/framework/components/rigid-body/system.js
```diff
@@ -615,126 +691,355 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
+     *             [ "carnivore", "reptile" ]
+     *         ],
+     *         filterCallback: (entity) => entity && entity.anim
+     *     }
+     * )[0];
      */
-    raycastAll(start, end, options = {}) {
-        Debug.assert(Ammo.AllHitsRayResultCallback, 'pc.RigidBodyComponentSystem#raycastAll: Your version of ammo.js does not expose Ammo.AllHitsRayResultCallback. Update it to latest.');
-
+    shapeCast(shape, startPosition, options = {}) {
```
AlexAPPi, line 855: What if I already have a Shape and want to reuse it instead of calling the constructor every time?

AlexAPPi:
I propose completely rethinking this system's concept. We're missing key features, so in my implementation, I rewrote core methods and made direct changes to Ammo.js (https://github.com/kripken/ammo.js/pull/433), enabling a universal interface for different physics engines (Ammo, Jolt, Cannon-es, etc.).

I'd also add NonAlloc methods—no memory allocation for new objects

```
// Pre-allocate reusable results array
const hitResults: HitResult[] = [
  new HitResult(),
  new HitResult(),
  new HitResult(),
  new HitResult(),
  new HitResult()
];

// Prototype
// Array size = max tests
shapeCastNonAlloc(results, shape, ....) // return hits count
raycastCastNonAlloc(results, ...) // return hits count

// Usage
const hits = shapeCastNonAlloc(results, boxShape, startPos, endPos);
for (let i = 0; i < hits; i++) {
  const hit = hitResults[i];
  // Process hit.body, hit.distance, etc.
}
```

src/framework/components/rigid-body/system.js
MushAsterion: You can provide the shape directly and add `destroy: false` as option. If no shape is provided and instead the shapes gets created based on the shape configuration as an object, it will be automatically destroyed.

MushAsterion: I've just extended with another `@example` to highlight the case. Let me know if it's any more clear!

MushAsterion:
> I rewrote core methods and made direct changes to Ammo.js ([kripken/ammo.js#433](https://github.com/kripken/ammo.js/pull/433)), enabling a universal interface for different physics engines (Ammo, Jolt, Cannon-es, etc.).

It's definitely good idea and positive for the engine, but it goes beyond the scope of this PR and would fit better having its own PR and a proper revamp label compared to enhancement here.

> I'd also add NonAlloc methods—no memory allocation for new objects

This PR is extending the existing API which already goes with new allocations, I have just kept the existing structure and focus the PR towards small changes (which is already big with 2 features). As highlight from previous comments and code behavior, Ammo doesn't natively sort the results, which would become a risk at knowing the appropriate allocation criteria.

PR #5043, opened 2023-02-05
Synchronize Dynamic RigidBody from Entity

### LINKED ISSUE #2215: Add warning when trying to setPosition on dynamic physics object
We do get a lot of posts where users are setting the position of an entity instead of using teleport for a dynamic physics object that is enabled.

Eg https://forum.playcanvas.com/t/how-to-handle-mass-cloning-of-objects-when-deep-clones-are-required/13805/6

If be great to add a warning for this.

### LINKED ISSUE #5042: Entity rendered different from Dynamic RigidBody
If a Dynamic RigidBody is disabled in simulation and a transform change is performed on entity it will display the entity with entity's transform instead of RigidBody's.

https://github.com/playcanvas/engine/blob/0da44850d353c61cffc6938f12baa72796f9cb5e/src/framework/components/rigid-body/component.js#L908-L910

Example: https://playcanvas.com/project/1035704

MushAsterion:
Fixes #2215 
Fixes #5042 

Synchronize Dynamic RigidBody transform on Entity before physics simulation to make it simpler for people modifying entity's transform directly and avoid ghost entity if Dynamic RigidBody is disabled.


LeXXik:
I am no physics specialist, but setting a transform on a dynamic body like that would end up with a body that has a moment of inertia and torques not matching the actual state of the body. Personally, I'd just disable setting a transform on a simulated entity altogether and issue a warning in the console. We can't expose Rigidbody Component to the Graph Node for that, so Graph Node should have a state or a flag that would stop it from auto-updating the transforms. Rigidbody Component then could enable that flag and take the ownership of updating them.

PR #5044, opened 2023-02-05
Extended GamePad support

MushAsterion:
Exposed more of Gamepad API:
- Added button touching
- Added gamepad connection/disconnection support
- Added gamepad retrieving from its ID/index
- Added gamepad pulse support
- Added gamepad button live-retrieving for better support from contant pooling devices (such as Mozilla)

### DIFF
--- a/src/platform/input/game-pads.js
+++ b/src/platform/input/game-pads.js
@@ -1,3 +1,5 @@
+import { EventHandler } from '../../core/event-handler.js';
+
 const MAPS = {
     DEFAULT: {
         buttons: [
@@ -81,20 +83,96 @@ const PRODUCT_CODES = {
     'Product: 0268': 'PS3'
 };
 
+let deadZone = 0.25;
+
 /**
  * Input handler for accessing GamePad input.
+ *
+ * @augments EventHandler
  */
-class GamePads {
+class GamePads extends EventHandler {
     /**
      * Create a new GamePads instance.
      */
     constructor() {
+        super();
+
+        /**
+         * Whether gamepads are supported by this device.
+         *
+         * @type {boolean}
+         */
         this.gamepadsSupported = !!navigator.getGamepads || !!navigator.webkitGetGamepads;
 
+        /**
+         * The list of current gamepads.
+         *
+         * @type {GamePad[]}
+         */
         this.current = [];
+
+        /**
+         * The list of previous button states.
+         *
+         * @type {boolean[][]}
+         */
         this.previous = [];
 
-        this.deadZone = 0.25;
+        window.addEventListener('gamepadconnected', event => {
+            const pad = new GamePad(event.gamepad, this.getMap(event.gamepad));
+
+            while(this.current.findIndex(gp => gp.index === pad.index) !== -1) {
+                this.current.splice(this.current.findIndex(gp => gp.index === pad.index), 1);
+            }
+
+            this.current.push(pad);
+            this.fire('gamepadconnected', pad);
+        });
+
+        window.addEventListener('gamepaddisconnected', event => {
+            const padIndex = this.current.findIndex(gp => gp.index === event.gamepad.index);
+
+            if (padIndex !== -1) {
+                this.fire('gamepaddisconnected', this.current[padIndex]);
+                this.current.splice(padIndex, 1);
+            }
+        });
+    }
+
+    /**
+     * Fired when a gamepad is connected.
+     *
+     * @event GamePads#gamepadconnected
+     * @param {GamePad} gamepad - The gamepad that was just connected.
+     * @example
+     * var onPadConnected = function (pad) {
+     *     // Make the gamepad pulse.
+     * };
+     * app.keyboard.on("gamepadconnected", onPadConnected, this);
+     */
+
+    /**
+     * Fired when a gamepad is disconnected.
+     *
+     * @event GamePads#gamepaddisconnected
+     * @param {GamePad} gamepad - The gamepad that was just disconnected.
+     * @example
+     * var onPadDisconnected = function (pad) {
+     *     // Pause the game.
+     * };
+     * app.keyboard.on("gamepaddisconnected", onPadDisconnected, this);
+     */
+
+    /**
+     * @type {number} Threshold for axes to return values. Must be between 0 and 1.
+     * @ignore
+     */
+    get deadZone() {
+        return deadZone;
+    }
+
+    set deadZone(value) {
+        deadZone = value;
     }
 
     /**
@@ -102,28 +180,28 @@ class GamePads {
      * `wasPressed` to work.
      */
     update() {
-        // move current buttons status into previous array
         for (let i = 0, l = this.current.length; i < l; i++) {
-            const buttons = this.current[i].pad.buttons;
-            const buttonsLen = buttons.length;
-            for (let j = 0; j < buttonsLen; j++) {
+            for (let j = 0, m = this.current[i].buttons.length; j < m; j++) {
                 if (this.previous[i] === undefined) {
                     this.previous[i] = [];
                 }
-                this.previous[i][j] = buttons[j].pressed;
+
+                this.current[i].buttons[j]._wasPressed = this.current[i].buttons[j].isPressed();
+                this.current[i].buttons[j]._wasTouched = this.current[i].buttons[j].isTouched();
+                this.previous[i][j] = this.current[i].buttons[j]._wasPressed;
             }
         }
 
-        // update current
-        this.poll(this.current);
+        this.previous.length = this.current.length;
+        this.poll();
     }
 
     /**
      * Poll for the latest data from the gamepad API.
      *
-     * @param {object[]} [pads] - An optional array used to receive the gamepads mapping. This
+     * @param {GamePad[]} [pads] - An optional array used to receive the gamepads mapping. This
      * array will be returned by this function.
-     * @returns {object[]} An array of gamepads and mappings for the model of gamepad that is
+     * @returns {GamePad[]} An array of gamepads and mappings for the model of gamepad that is
      * attached.
      * @example
      * var gamepads = new pc.GamePads();
@@ -136,15 +214,15 @@ class GamePads {
 
         if (this.gamepadsSupported) {
             const padDevices = navigator.getGamepads ? navigator.getGamepads() : navigator.webkitGetGamepads();
+
             for (let i = 0, len = padDevices.length; i < len; i++) {
                 if (padDevices[i]) {
-                    pads.push({
-                        map: this.getMap(padDevices[i]),
-                        pad: padDevices[i]
-                    });
+                    let pad = this.findByIndex(padDevices[i].index);
+                    pads.push(pad._update(padDevices[i], this.getMap(padDevices[i])));
                 }
             }
         }
+
         return pads;
     }
 
@@ -161,84 +239,381 @@ class GamePads {
     /**
      * Returns true if the button on the pad requested is pressed.
      *
-     * @param {number} index - The index of the pad to check, use constants {@link PAD_1},
-     * {@link PAD_2}, etc.
+     * @param {number} index - The index of the pad to check, use constants {@link PAD_1}, {@link PAD_2}, etc.
      * @param {number} button - The button to test, use constants {@link PAD_FACE_1}, etc.
      * @returns {boolean} True if the button is pressed.
      */
     isPressed(index, button) {
-        if (!this.current[index]) {
-            return false;
-        }
-
-        const key = this.current[index].map.buttons[button];
-        return this.current[index].pad.buttons[pc[key]].pressed;
+        return this.current[index] && this.current[index].isPressed(button);
     }
 
     /**
      * Returns true if the button was pressed since the last frame.
      *
-     * @param {number} index - The index of the pad to check, use constants {@link PAD_1},
-     * {@link PAD_2}, etc.
+     * @param {number} index - The index of the pad to check, use constants {@link PAD_1}, {@link PAD_2}, etc.
      * @param {number} button - The button to test, use constants {@link PAD_FACE_1}, etc.
      * @returns {boolean} True if the button was pressed since the last frame.
      */
     wasPressed(index, button) {
-        if (!this.current[index]) {
-            return false;
-        }
-
-        const key = this.current[index].map.buttons[button];
-        const i = pc[key];
-
-        // Previous pad buttons may not have been populated yet
-        // If this is the first time frame a pad has been detected
-        return this.current[index].pad.buttons[i].pressed && !(this.previous[index] && this.previous[index][i]);
+        return this.current[index] && this.current[index].wasPressed(button);
     }
 
     /**
      * Returns true if the button was released since the last frame.
      *
-     * @param {number} index - The index of the pad to check, use constants {@link PAD_1},
-     * {@link PAD_2}, etc.
+     * @param {number} index - The index of the pad to check, use constants {@link PAD_1}, {@link PAD_2}, etc.
      * @param {number} button - The button to test, use constants {@link PAD_FACE_1}, etc.
      * @returns {boolean} True if the button was released since the last frame.
      */
     wasReleased(index, button) {
-        if (!this.current[index]) {
-            return false;
-        }
-
-        const key = this.current[index].map.buttons[button];
-        const i = pc[key];
-
-        // Previous pad buttons may not have been populated yet
-        // If this is the first time frame a pad has been detected
-        return !this.current[index].pad.buttons[i].pressed && (this.previous[index] && this.previous[index][i]);
+        return this.current[index] && this.current[index].wasReleased(button);
     }
 
     /**
      * Get the value of one of the analogue axes of the pad.
      *
-     * @param {number} index - The index of the pad to check, use constants {@link PAD_1},
-     * {@link PAD_2}, etc.
-     * @param {number} axes - The axes to get the value of, use constants {@link PAD_L_STICK_X},
-     * etc.
+     * @param {number} index - The index of the pad to check, use constants {@link PAD_1}, {@link PAD_2}, etc.
+     * @param {number} axis - The axis to get the value of, use constants {@link PAD_L_STICK_X}, etc.
      * @returns {number} The value of the axis between -1 and 1.
      */
-    getAxis(index, axes) {
-        if (!this.current[index]) {
-            return 0;
-        }
+    getAxis(index, axis) {
+        return this.current[index] ? this.current[index].getAxis(axis) : 0;
+    }
 
-        const key = this.current[index].map.axes[axes];
-        let value = this.current[index].pad.axes[pc[key]];
+    /**
+     * Find a connected {@link GamePad} from its identifier.
+     *
+     * @param {string} id - The identifier to search for.
+     * @returns {GamePad|null} The {@link GamePad} with the matching identifier or null if no gamepad is found or the gamepad is not connected.
+     */
+    findById(id) {
+        return this.current.find(gp => gp && gp.id === id) || null;
+    }
 
-        if (Math.abs(value) < this.deadZone) {
-            value = 0;
-        }
-        return value;
+    /**
+     * Find a connected {@link GamePad} from its device index.
+     *
+     * @param {number} index - The device index to search for.
+     * @returns {GamePad|null} The {@link GamePad} with the matching device index or null if no gamepad is found or the gamepad is not connected.
+     */
+    findByIndex(index) {
+        return this.current.find(gp => gp && gp.index === index) || null;
     }
 }
 
-export { GamePads };
+/**
+ * A GamePad stores information about a gamepad from the Gamepad API.
+ */
+class GamePad {
+    /**
+     * Create a new GamePad Instance.
+     *
+     * @param {Gamepad} gamepad - The original Gamepad API gamepad.
+     * @param {object} map - The buttons and axes map.
+     * @hideconstructor
+     */
+    constructor(gamepad, map) {
+        /**
+         * The identifier for the gamepad. Its structure depends on device.
+         *
+         * @type {string}
+         */
+        this.id = gamepad.id;
+
+        /**
+         * The index for this controller. A gamepad that is disconnected and reconnected will retain the same index.
+         *
+         * @type {number}
+         */
+        this.index = gamepad.index;
+
+        /**
+         * The buttons present on the GamePad.
+         *
+         * @type {GamePadButton[]}
+         * @readonly
+         */
+        this.buttons = gamepad.buttons.map(b => new GamePadButton(b));
+
+        /**
+         * The gamepad mapping detected by the browser. Value is either "standard" or "xr-standard".
+         *
+         * @type {string}
+         */
+        this.mapping = gamepad.mapping === 'xr-standard' ? 'xr-standard' : 'standard';
+
+        /**
+         * The buttons and axes map.
+         *
+         * @type {object}
+         */
+        this.map = map;
+
+        /**
+         * The original Gamepad API gamepad.
+         *
+         * @type {Gamepad}
+         * @ignore
+         */
+        this.pad = gamepad;
+    }
+
+    /**
+     * Update the existing GamePad Instance.
+     *
+     * @param {Gamepad} gamepad - The original Gamepad API gamepad.
+     * @param {object} map - The buttons and axes map.
+     * @ignore
+     */
+    _update(gamepad, map) {
+        this.id = gamepad.id;
+        this.index = gamepad.index;
+        this.mapping = gamepad.mapping === 'xr-standard' ? 'xr-standard' : 'standard';
+        this.map = map;
+        this.pad = gamepad;
+
+        for (let i = 0, l = this.buttons.length; i < l; i++) {
+            this.buttons[i]._update(gamepad.buttons[i]);
+        }
+
+        return this;
+    }
+
+    /**
+     * @type {number[]} - The values from analog axes present on the GamePad. Values are between -1 and 1.
+     * @readonly
+     */
+    get axes() {
+        return this.pad.axes;
+    }
+
+    /**
+     * Make the gamepad vibrate.
+     *
+     * @param {number} intensity - Intensity for the vibrations, must be between 0 and 1.
+     * @param {number} duration - Duration for the vibration, in miliseconds.
+     * @param {object} [options] - Options for special vibration patern.
+     * @param {string} [options.type] - Type of patern. Available types are "dual-rumble" and "vibration". Defaults to "vibration".
+     * @param {number} [options.startDelay] - Delay before the patern starts, in miliseconds. Defaults to 0.
+     * @param {number} [options.strongMagnitude] - Intensity for strong actuators, must be between 0 and 1. Defaults to intensity.
+     * @param {number} [options.weakMagnitude] - Intensity for weak actuators, must be between 0 and 1. Defaults to intensity.
+     * @returns {Promise<boolean>} Return a Promise resulting in true if the pulse was successfully completed.
+     */
+    pulse(intensity, duration, options) {
+        const actuator = this.pad.vibrationActuator;
+
+        if (actuator) {
+            if (actuator.playEffect) {
+                if (options) {
+                    return actuator.playEffect(options.type === 'dual-rumble' ? 'dual-rumble' : 'vibration', {
+                        'duration': duration,
+                        'startDelay': options.startDelay || 0,
+                        'strongMagnitude': typeof options.strongMagnitude === 'number' ? options.strongMagnitude : intensity,
+                        'weakMagnitude':  typeof options.weakMagnitude === 'number' ? options.weakMagnitude : intensity
+                    });
+                } else {
+                    return actuator.playEffect('vibration', {
+                        'duration': duration,
+                        'startDelay': 0,
+                        'strongMagnitude': intensity,
+                        'weakMagnitude': intensity
+                    });
+                }
+            } else if (actuator.pulse) {
+                if (options && options.startDelay) {
+                    // Custom delay
+                    return new Promise(function(resolve, reject) {
+                        setTimeout(function() {
+                            actuator.pulse(intensity, duration).then(resolve).catch(reject);
+                        }, options.startDelay);
+                    });
+                } else {
+                    return actuator.pulse(intensity, duration);
+                }
+            }
+        }
+
+        return new Promise(resolve => resolve(false));
+    }
+
+    /**
+     * Return true if the gamepad is connected.
+     *
+     * @returns {boolean} Return true if the gamepad is connected.
+     */
+    isConnected() {
+        return this.pad.connected;
+    }
+
+    /**
+     * Returns true if the button is pressed.
+     *
+     * @param {number} button - The button to test, use constants {@link PAD_FACE_1}, etc.
+     * @returns {boolean} True if the button is pressed.
+     */
+    isPressed(button) {
+        return this.buttons[button] ? this.buttons[button].isPressed() : false;
+    }
+
+    /**
+     * Return true if the button was pressed since the last update.
+     *
+     * @param {number} button - The button to test, use constants {@link PAD_FACE_1}, etc.
+     * @returns {boolean} Return true if the button was pressed, false if not.
+     */
+    wasPressed(button) {
+        return this.buttons[button] ? this.buttons[button].wasPressed() : false;
+    }
+
+    /**
+     * Return true if the button was released since the last update.
+     *
+     * @param {number} button - The button to test, use constants {@link PAD_FACE_1}, etc.
+     * @returns {boolean} Return true if the button was released, false if not.
+     */
+    wasReleased(button) {
+        return this.buttons[button] ? this.buttons[button].wasReleased() : false;
+    }
+
+    /**
+     * Returns true if the button is touched.
+     *
+     * @param {number} button - The button to test, use constants {@link PAD_FACE_1}, etc.
+     * @returns {boolean} True if the button is touched.
+     */
+    isTouched(button) {
+        return this.buttons[button] ? this.buttons[button].isTouched() : false;
+    }
+
+    /**
+     * Return true if the button was touched since the last update.
+     *
+     * @param {number} button - The button to test, use constants {@link PAD_FACE_1}, etc.
+     * @returns {boolean} Return true if the button was touched, false if not.
+     */
+    wasTouched(button) {
+        return this.buttons[button] ? this.buttons[button].wasTouched() : false;
+    }
+
+    /**
+     * Returns the value of a button between 0 and 1, with 0 representing a button that is not pressed, and 1 representing a button that is fully pressed.
+     *
+     * @param {number} button - The button to retrieve, use constants {@link PAD_FACE_1}, etc.
+     * @returns {number} The value of the button between 0 and 1, with 0 representing a button that is not pressed, and 1 representing a button that is fully pressed.
+     */
+    getValue(button) {
+        return this.buttons[button] ? this.buttons[button].value : 0;
+    }
+
+    /**
+     * Get the value of one of the analogue axes of the pad.
+     *
+     * @param {number} axis - The axis to get the value of, use constants {@link PAD_L_STICK_X}, etc.
+     * @returns {number} The value of the axis between -1 and 1.
+     */
+    getAxis(axis) {
+        return this.axes[axis] && Math.abs(this.axes[axis]) > deadZone ? this.axes[axis] : 0;
+    }
+}
+
+/**
+ * A GamePadButton stores information about a button from the Gamepad API.
+ */
+class GamePadButton {
+    /**
+     * Create a new GamePadButton instance.
+     *
+     * @param {GamepadButton} button - The original Gamepad API gamepad button.
+     * @hideconstructor
+     */
+    constructor(button) {
+        /**
+         * Whether this button was pressed on last frame.
+         *
+         * @type {boolean}
+         * @ignore
+         */
+        this._wasPressed = button.pressed;
+
+        /**
+         * Whether this button was touched on last frame.
+         *
+         * @type {boolean}
+         * @ignore
+         */
+        this._wasTouched = typeof button.touched === 'boolean' ? button.touched : button.value > 0;
+
+        /**
+         * The original Gamepad API gamepad button.
+         *
+         * @type {GamepadButton}
+         * @ignore
+         */
+        this._button = button;
+    }
+
+    /**
+     * Update the existing GamePadButton Instance.
+     *
+     * @param {GamepadButton} button - The original Gamepad API gamepad button.
+     * @ignore
+     */
+    _update(button) {
+        this._button = button;
+    }
+
+    /**
+     * @type {number} - The value for the button between 0 and 1, with 0 representing a button that is not pressed, and 1 representing a button that is fully pressed.
+     * @readonly
+     */
+    get value() {
+        return this._button.value;
+    }
+
+    /**
+     * Return true if the button is currently down.
+     *
+     * @returns {boolean} Return true if the button is currently down.
+     */
+    isPressed() {
+        return this._button.pressed;
+    }
+
+    /**
+     * Return true if the button was pressed since the last update.
+     *
+     * @returns {boolean} Return true if the button was pressed, false if not.
+     */
+    wasPressed() {
+        return this._wasPressed === false && this.isPressed();
+    }
+
+    /**
+     * Return true if the button was released since the last update.
+     *
+     * @returns {boolean} Return true if the button was released, false if not.
+     */
+    wasReleased() {
+        return this._wasPressed === true && !this.isPressed();
+    }
+
+    /**
+     * Return true if the button is currently touched.
+     *
+     * @returns {boolean} Return true if the button is currently touched.
+     */
+    isTouched() {
+        return typeof this._button.touched === 'boolean' ? this._button.touched : this._button.value > 0;
+    }
+
+    /**
+     * Return true if the button was touched since the last update.
+     *
+     * @returns {boolean} Return true if the button was touched, false if not.
+     */
+    wasTouched() {
+        return this._wasTouched === false && this.isTouched();
+    }
+}
+
+export { GamePads, GamePad, GamePadButton };


src/platform/input/game-pads.js
```diff
@@ -81,49 +83,440 @@ const PRODUCT_CODES = {
...
+    get axes() {
+        return this.pad.axes;
+    }
+
+    /**
+     * Make the gamepad vibrate.
+     *
+     * @param {number} intensity - Intensity for the vibrations, must be between 0 and 1.
+     * @param {number} duration - Duration for the vibration, in miliseconds.
+     * @param {object} [options] - Options for special vibration patern.
```
willeastcott, line 277:
```suggestion
* @param {object} [options] - Options for special vibration pattern.
```

kungfooman:
1) I am getting this error:

image

I'm now sure why, but I think it's because you removed `this.poll(this.current)` and changed all kinds of other things.

2) Do we allow Promises now?

3) I would rather see this merged first: https://github.com/playcanvas/engine/pull/5017

4) MAPS doesn't work anymore as it used to work - TRI/QUAD buttons are flipped on my PS4 Dual-Shock after adding the product code:

```js
const PRODUCT_CODES = {
    'Product: 0268': 'PS3', // PS3
    'Product: 09cc': 'PS3'  // PS4 - same layout
};
```

src/platform/input/game-pads.js
```diff
@@ -74,56 +77,480 @@ const MAPS = {
...
+    set deadZone(value) {
+        deadZone = value;
+    }
+
+    /**
+     * @type {number} Threshold for axes to return values. Must be between 0 and 1.
+     * @ignore
+     */
+    get deadZone() {
+        return deadZone;
```
kungfooman, line 524: Is this an over-engineered static variable now?

MushAsterion:
My bad, I just fixed the controllers not being initialized if they were already connected. You should not have this error anymore.

Regarding 2, it's because of how the API works and I think having the information on completition is an important information especially if playing with complex effects.

Regarding 4, honestly I don't really get the point of MAPS at all especially with the variables, I just kept it for compatibility as goes for `GamePads#previous`

src/platform/input/game-pads.js
MushAsterion: As I didn't added a reference to the GamePads on the GamePad class I had to make this variable global to be able to access it. Added getter/setter for compatibility on it being GamePads variable at first

kungfooman:
> Regarding 4, honestly I don't really get the point of MAPS at all especially with the variables, I just kept it for compatibility as goes for `GamePads#previous`

I'm not sure what you mean with compatibility, because it didn't work anymore when I tested it. What I use `MAPS` for is basically to "normalize" the constants between different gamepads.

E.g. on a PS4 controller the common buttons are: X, O, TRI, SQ and the indices in HTML5 Gamepad#buttons are `[0, 1, 3, 2]`

The same `X, O, TRI, SQ` on a Logitech G29 are indexed via `[0, 2, 3, 1]`

I don't know who initially came up with `MAPS`, maybe someone can comment on it? I would like to keep this functionality.

MushAsterion:
Currently the map is being stored on the objects within `GamePads#current`, so if anyone is using it and we remove it then they might have issue with their code.

MushAsterion:
So I implemented with the MAPS, does that look better to you @kungfooman?

src/platform/input/game-pads.js
```diff
@@ -74,56 +114,487 @@ const MAPS = {
...
+                    });
+                }
+
+                return actuator.pulse(intensity, duration);
+            }
+        }
+
+        return new Promise((resolve) => {
+            resolve(false);
+        });
```
kungfooman, line 396:
```suggestion
        return Promise.resolve(false);
```

This is shorter, but the function could also be marked as `async function`. Not sure if we want the promises still, so far we stick to IE11 support. I can't test it as I'm on Linux.

Any thoughts from others who introduced Promises previously? I see `Promise` used in `src/platform/graphics/graphics-device-create.js` for example.

kungfooman:
> So I implemented with the MAPS, does that look better to you @kungfooman?

Thank you for the quick change, I just tested it and it works perfectly! :tada: 

Since this is about extended GamePad support, I would like to point out another issue. On the Logitech G29 steering wheel are the four direction buttons, marked red:

image

They evaluate to either 1 or -1, no inbetween values:

image

So they are not implemented as four buttons, but as two axes. Right now it seems to me there is no way to properly map them, because we are treating buttons/axes as inherently different/distinct. Do you maybe have any idea on how to combine it?

src/platform/input/game-pads.js
mvaligursky: @kungfooman - good question regarding the Promises, we'll have an internal discussion on this, as that's something we should really resolve.

willeastcott: @mvaligursky Always good to resolve your promises.

MushAsterion:
This should be working now @kungfooman, however I don't have this controller to test out so if you could have a look please. Simply add a variable within the map for the relevant controller:

```JS
dualButtons: [
    [],
    [],
    [],
    [],
    [ 'PAD_DOWN', 'PAD_UP' ],
    [ 'PAD_LEFT', 'PAD_RIGHT' ]
]
```

or whatever the good order is and it should be working properly!

kungfooman:
Thank you @MushAsterion :+1: I tried to get it to work, but so far I couldn't (will try more later).

**1)**

I see that you like a rather functional programming style, which right now seems to create a bunch of GC / performance considerations:

image

image

image

Since `GamePad#buttons` is a getter and is called all the time in `GamePad#update`, it makes quite an impact.

I added the `count++` to count the `Array#map` (which creates a new array all the time) and it happens 138x for two connected gamepads.

You can already reduce 138 array allocations per update to just 2 by simply reusing what you have:

```js
    update() {
        for (let i = 0, l = this.current.length; i < l; i++) {
            const gamePad = this.current[i];
            const { buttons } = gamePad;
            for (let j = 0, m = buttons.length; j < m; j++) {
                const button = buttons[j];
                button._wasPressed = button.isPressed();
                button._wasTouched = button.isTouched();
            }
        }
        this.poll();
    }
```

I like some amount of functional programming myself, while trying to be as sneaky with GC as possible. Usually you can do that by simply reusing arrays by either shrinking or growing them, depending on the circumstances.

**2)**

`MAPS#INDEXES` should also be its own `const INDEXES = ...`, that keeps the type of `MAPS` simple while removing the property lookup.

**3)**

`GamePadButton#isPressed()` should be a getter, and simply named `GamePadButton#pressed`. Same with `GamePadButton#wasPressed()`. The name "isPressed" made sense on `GamePads#isPressed` since it requires parameters, but once it's boiled down to an actual `GamePadButton`, it's simply a getter.

E.g. this is how it's used previously here aswell: https://github.com/playcanvas/engine/blob/374b55fd8d4641106beb88bb5392eabdc84e7bb3/src/framework/components/button/component.js#L81

MushAsterion:
My bad, I am not really used to optimizing...

I fixed 1 and 2 and also added a `getButton` function to retrieve only a single button instead of generating the whole array and used it within the different functions to retrieve something from a single button.

Regarding 3, I used this naming to keep it consistent with other input systems

src/platform/input/game-pads.js
```diff
@@ -1,3 +1,44 @@
...
+        'PAD_DOWN': PAD_DOWN,
+        'PAD_LEFT': PAD_LEFT,
+        'PAD_RIGHT': PAD_RIGHT,
+        'PAD_VENDOR': PAD_VENDOR,
+        'XRPAD_TRIGGER': XRPAD_TRIGGER,
+        'XRPAD_SQUEEZE': XRPAD_SQUEEZE,
+        'XRPAD_TOUCHPAD_BUTTON': XRPAD_TOUCHPAD_BUTTON,
+        'XRPAD_STICK_BUTTON': XRPAD_STICK_BUTTON,
+        'XRPAD_A': XRPAD_A,
+        'XRPAD_B': XRPAD_B
```
kungfooman, line 28:
```suggestion
        PAD_FACE_1,
        PAD_FACE_2,
        PAD_FACE_3,
        PAD_FACE_4,
        PAD_L_SHOULDER_1,
        PAD_R_SHOULDER_1,
        PAD_L_SHOULDER_2,
        PAD_R_SHOULDER_2,
        PAD_SELECT,
        PAD_START,
        PAD_L_STICK_BUTTON,
        PAD_R_STICK_BUTTON,
        PAD_UP,
        PAD_DOWN,
        PAD_LEFT,
        PAD_RIGHT,
        PAD_VENDOR,
        XRPAD_TRIGGER,
        XRPAD_SQUEEZE,
        XRPAD_TOUCHPAD_BUTTON,
        XRPAD_STICK_BUTTON,
        XRPAD_A,
        XRPAD_B
```

src/platform/input/game-pads.js
```diff
@@ -1,3 +1,44 @@
...
+    },
+    axes: {
+        'PAD_L_STICK_X': PAD_L_STICK_X,
+        'PAD_L_STICK_Y': PAD_L_STICK_Y,
+        'PAD_R_STICK_X': PAD_R_STICK_X,
+        'PAD_R_STICK_Y': PAD_R_STICK_Y,
+        'XRPAD_TOUCHPAD_X': XRPAD_TOUCHPAD_X,
+        'XRPAD_TOUCHPAD_Y': XRPAD_TOUCHPAD_Y,
+        'XRPAD_STICK_X': XRPAD_STICK_X,
+        'XRPAD_STICK_Y': XRPAD_STICK_Y
```
kungfooman, line 38:
```suggestion
        PAD_L_STICK_X,
        PAD_L_STICK_Y,
        PAD_R_STICK_X,
        PAD_R_STICK_Y,
        XRPAD_TOUCHPAD_X,
        XRPAD_TOUCHPAD_Y,
        XRPAD_STICK_X,
        XRPAD_STICK_Y
```

kungfooman (changes requested):
If the variable name equals the key name, you can simply drop it. This doesn't just remove ASCII soup, but also makes it harder to introduce bugs (I remember such a bug)

E.g. as here and a few lines below too:

https://github.com/playcanvas/engine/blob/374b55fd8d4641106beb88bb5392eabdc84e7bb3/src/framework/components/anim/system.js#L105

kungfooman:
I see some code duplication between `poll` and the gamepad `gamepadconnected`/`gamepaddisconnected` events.

Not that it's bad, but I thought previously about using these events to get rid of the constant navigator polling to see if there are new gamepads.

Why do we need to poll in `update`, if we use these events now? And wouldn't it be enough to just call `poll` in these events, to remove the code duplication?

Another issue with these two events is that you need to get rid of them aswell. E.g. calling `app.destroy()` shouldn't leave any dangling events.

About point 3)

Right now you have:

```js
                button._wasPressed = button.isPressed();
                button._wasTouched = button.isTouched();
```

It simply looks incoherent. Also you access a private variable (denoted by _) from the outside here now.

IMO `button.wasPressed = button.pressed;` is shortest and also looks most natural without abusing functions as getters/setters. What you mean is probably `Keyboard/Mouse#wasPressed(key)`, but that is only a function because its a method for a container (entire mouse or keyboad), in contrast to an individual button. After all the HTML5 Gamepad API also uses `Gamepad#pressed`.

MushAsterion:
In fact some browsers such as Chrome only poll values when calling the `getGamepads`/`webkitGetGamepads` functions which makes it impossible to have functions based on previous values if we don't poll those values every frame. And polling on constructor is made to initialize the gamepads already connected as there is no standard on whether gamepads are connected on page load or require a first action.

You're true, using `_was` for previous values might seem incoherent, that's fixed to `_previously` to avoid confusion. And I got your point, it's now getters. Regarding accessing private properties I did a bit of cleaning and made public the functions that were being accessed from outside and moved private properties update to functions within the class. Does that sound better to you?

Events are now cleaned on app destroy, gamepads module as well.

src/platform/input/game-pads.js
```diff
@@ -74,56 +115,601 @@ const MAPS = {
...
+        ]
     }
 };
 
 const PRODUCT_CODES = {
     'Product: 0268': 'PS3'
 };
 
+let deadZone = 0.25;
+const getGamepads = navigator ? navigator.getGamepads || navigator.webkitGetGamepads : undefined;
```
kungfooman, line 150: How does this work for you? `getGamepads` is now unbound and throws `Uncaught TypeError: Illegal invocation`:

image

I would still rather like it the way I did it in my PR, but you can also just bind it to navigator or rework it in another way.

MushAsterion: That's a very good question, in fact it was not, I forgot to remove this change when updating the file... Thanks for checking!

src/platform/input/game-pads.js
```diff
@@ -74,56 +115,601 @@ const MAPS = {
...
+
+/**
+ * A GamePadButton stores information about a button from the Gamepad API.
+ */
+class GamePadButton {
```
kungfooman, line 154: I think we need some discussion on the naming here, the history so far is that we copy the exact name in `window` while using JSDoc like `/** @type {globalThis.GamepadButton} */` to distinguish. I am not quite happy with both ways, because I confuse these names easily and often wonder "what is it now?".

There are multiple ways for distinguishing:
1) GamepadButton vs GamePadButton (as you have it right now)
2) globalThis.GamepadButton vs GamepadButton (as in Mouse/Keyboard/Touch classes/events)
3) GamepadButton vs e.g. ExtendedGamepadButton (or any other prefix like "Higher...", "Advanced...")

Any thoughts on this @willeastcott since you handle most of JSDoc/typings?

MushAsterion: My point on having with `GamePadButton` is that the inital class was `GamePads` so I made everything consistent from the first naming

src/platform/input/game-pads.js
```diff
@@ -135,110 +720,116 @@ class GamePads {
         }
 
         if (this.gamepadsSupported) {
-            const padDevices = navigator.getGamepads ? navigator.getGamepads() : navigator.webkitGetGamepads();
+            const padDevices = navigator.getGamepads() || navigator.webkitGetGamepads();
```
kungfooman, line 723: This will not work either, since `navigator.getGamepads` will throw in the case that only `navigator.webkitGetGamepads` exists.

kungfooman:
I still think code like:

```js
    get previous() {
        return this.current.map((c) => {
            return c.buttons.map((b) => {
                return b ? b._previouslyPressed : false;
            });
        });
    }
```

is unacceptable, generating not just one temporary array, but a temporary array for each element of a temporary array.

Just a general question, why is all the mapping here and in other methods required? IMO it should be as simple as it gets... a preallocated `boolean[][]` filled during `update`'ing.

src/platform/input/game-pads.js
```diff
@@ -74,56 +115,600 @@ const MAPS = {
...
+        return this._button.pressed;
+    }
+
+    /**
+     * Whether the button was pressed.
+     *
+     * @type {boolean}
+     */
+    get wasPressed() {
+        return this._previouslyPressed === false && this.pressed;
```
kungfooman, line 231: `=== false` is just `!`

src/platform/input/game-pads.js
```diff
@@ -74,56 +115,600 @@ const MAPS = {
...
+    }
+
+    /**
+     * Whether the button was released since the last update.
+     *
+     * @type {boolean}
+     */
+    get wasReleased() {
+        return this._previouslyPressed === true && !this.pressed;
```
kungfooman, line 240: `=== true` is just nothing since you declared the type as boolean?

src/platform/input/game-pads.js
```diff
@@ -74,56 +115,600 @@ const MAPS = {
...
+        return typeof this._button.touched === 'boolean' ? this._button.touched : this._button.value > 0;
+    }
+
+    /**
+     * Whether the button was touched since the last update.
+     *
+     * @type {boolean}
+     */
+    get wasTouched() {
+        return this._previouslyTouched === false && this.touched;
```
kungfooman, line 258: `=== false` is just `!`

kungfooman (changes requested):

src/platform/input/game-pads.js
kungfooman: It might make sense to give each GamePadButton a reference to GamePad, to receive the deadZone via `this.gamePad.deadZone`. But on the other hand, this is still not fine-grained enough.

For example clamping the steering wheel doesn't make sense, but other axes on it should be clamped. IMO there needs to be a system to enable clamping on a per-axis-base.

src/platform/input/game-pads.js
MushAsterion: That's fixed! Thanks

src/platform/input/game-pads.js
MushAsterion: That's fixed! Thanks

src/platform/input/game-pads.js
MushAsterion: That's fixed! Thanks

MushAsterion:
> Just a general question, why is all the mapping here and in other methods required? IMO it should be as simple as it gets... a preallocated `boolean[][]` filled during `update`'ing.

`previous` is only kept for users who are using it from previous versions and is not used at all in this version, therefore I don't see the reason while it should be updated on every update rather than only if it is used by the project. It would be a loss to update it while this does not represent anything for the class and the result might be wrong based on internal mapping (in fact I forgot to change it here but it should not be using `buttons` but `_buttons` to have same behavior as before, which is now fixed). If you want an array to be preallocated it's still possible to have it in a global variable outside of the class so it saves a bit, would that sound better to you?

src/platform/input/game-pads.js
kungfooman: Ideally the clamping code can be skipped/ignored entirely, if it's known that it's useless/harmful-to-precision (like for a steering wheel). As it is right now, I would need to set it to 0 to control e.g. a racing car in the full range, but then other players (lets say playing with a normal gamepad) would be affected by that setting aswell.

I don't see how a mapped axis button hurts this, you wouldn't assign the steering wheel axis as a button.

mvaligursky: A circular reference like this does not excite me at all .. it suggest it should be designed differently in most cases.

I'd suggest to step away from the code a bit and work on a a design of what this system should look like: what functionality is required and what should the public API for this to be. Ideally in comparison to what the current design is, to see where a backwards compatibility would change.

MushAsterion: To me the best option is to keep a global variable and then up to the user to make it 0 if they need more precision and use their own precision on the script. Fast, reliable and simple.

But question is if `GamePadButton`s should use `deadZone`, in current code it doesn't.

kungfooman: @MushAsterion

There is no need for a `GamePadAxis`, you can give `GamePadButton` a reference to the `GamePad` and figure out if this particular button needs mapping from an axis based on `GamePad#map`. If so, the `GamePad#pressed` will tell so and that will also make "GamePadButton#wasPressed" work for an axis button.

So the extra information a `GamePadButton` requires is if it's mapped to an axis and then to which one in the form of a number/index.

@mvaligursky In this case I would suggest to deprecate `GamePads#deadZone`, as it makes least sense on `GamePads`. It would make more sense on individual GamePad's, but then it makes most sense on individual axes.

MushAsterion: > There is no need for a `GamePadAxis`

If you want a per-axis deadZone it's the easiest way and most understandable

> you can give `GamePadButton` a reference to the `GamePad` and figure out if this particular button needs mapping from an axis based on `GamePad#map`. If so, the `GamePad#pressed` will tell so and that will also make "GamePadButton#wasPressed" work for an axis button.

In fact the `GamePadButton#wasPressed` already works for an axis button from the getter, the only moment it's not working is if you don't fetch the most recent data. But again it introduces a circular reference.

src/platform/input/game-pads.js
```diff
@@ -74,56 +115,600 @@ const MAPS = {
...
+     */
+    get buttons() {
+        return this.map.buttons ? this.map.buttons.map(b => this._getButton(b)) : [];
+    }
+
+    /**
+     * Make the gamepad vibrate.
+     *
+     * @param {number} intensity - Intensity for the vibrations, must be between 0 and 1.
+     * @param {number} duration - Duration for the vibration, in miliseconds.
```
willeastcott, line 391:
```suggestion
     * @param {number} duration - Duration for the vibration, in milliseconds.
```

src/platform/input/game-pads.js
```diff
@@ -74,56 +115,600 @@ const MAPS = {
...
+     * @param {object} [options] - Options for special vibration patern.
+     * @param {string} [options.type] - Type of patern. Available types are "dual-rumble" and "vibration". Defaults to "vibration".
+     * @param {number} [options.startDelay] - Delay before the patern starts, in miliseconds. Defaults to 0.
```
willeastcott, line 394:
```suggestion
* @param {number} [options.startDelay] - Delay before the pattern starts, in milliseconds. Defaults to 0.
```

src/platform/input/game-pads.js
```diff
@@ -74,56 +115,600 @@ const MAPS = {
...
+     * @type {GamePadButton[]} - The buttons present on the GamePad. Some buttons may be null.
+     */
+    get buttons() {
+        return this.map.buttons ? this.map.buttons.map(b => this._getButton(b)) : [];
+    }
+
+    /**
+     * Make the gamepad vibrate.
+     *
+     * @param {number} intensity - Intensity for the vibrations, must be between 0 and 1.
```
willeastcott, line 390:
```suggestion
     * @param {number} intensity - Intensity for the vibration in the range 0 to 1.
```

src/platform/input/game-pads.js
```diff
@@ -74,56 +115,600 @@ const MAPS = {
...
+     * @param {number} duration - Duration for the vibration, in miliseconds.
+     * @param {object} [options] - Options for special vibration patern.
```
willeastcott, line 392:
```suggestion
     * @param {object} [options] - Options for special vibration pattern.
```

src/platform/input/game-pads.js
```diff
@@ -74,56 +115,600 @@ const MAPS = {
...
+     * @param {string} [options.type] - Type of patern. Available types are "dual-rumble" and "vibration". Defaults to "vibration".
```
willeastcott, line 393:
```suggestion
     * @param {string} [options.type] - Type of pattern. Available types are "dual-rumble" and "vibration". Defaults to "vibration".
```

src/platform/input/game-pads.js
```diff
@@ -74,56 +115,600 @@ const MAPS = {
...
+     * @param {number} [options.startDelay] - Delay before the patern starts, in miliseconds. Defaults to 0.
+     * @param {number} [options.strongMagnitude] - Intensity for strong actuators, must be between 0 and 1. Defaults to intensity.
```
willeastcott, line 395:
```suggestion
     * @param {number} [options.strongMagnitude] - Intensity for strong actuators in the range 0 to 1. Defaults to intensity.
```

src/platform/input/game-pads.js
```diff
@@ -74,56 +115,600 @@ const MAPS = {
...
+     * @param {number} [options.weakMagnitude] - Intensity for weak actuators, must be between 0 and 1. Defaults to intensity.
```
willeastcott, line 396:
```suggestion
     * @param {number} [options.weakMagnitude] - Intensity for weak actuators in the range 0 to 1. Defaults to intensity.
```

src/platform/input/game-pads.js
```diff
@@ -74,56 +115,600 @@ const MAPS = {
...
+    wasTouched(button) {
+        const b = this.getButton(button);
+        return b ? b.wasTouched : false;
+    }
+
+    /**
+     * Returns the value of a button between 0 and 1, with 0 representing a button that is not pressed, and 1 representing a button that is fully pressed.
+     *
+     * @param {number} button - The button to retrieve, use constants {@link PAD_FACE_1}, etc.
+     * @returns {number} The value of the button between 0 and 1, with 0 representing a button that is not pressed, and 1 representing a button that is fully pressed.
```
willeastcott, line 555:
```suggestion
     * @returns {number} The value of the button between 0 and 1.
```

willeastcott: Probably not worth repeating the full function description here.

src/platform/input/game-pads.js
```diff
@@ -74,56 +115,600 @@ const MAPS = {
...
+            'XRPAD_TOUCHPAD_BUTTON',
+            'XRPAD_STICK_BUTTON',
+
+            // Face buttons
+            'XRPAD_A',
+            'XRPAD_B'
+        ],
+
+        axes: [
+            // Analogue Sticks
```
willeastcott, line 136: In this file, there are instances of `analog` and analogue`. From doing a bit of Googling, it seems `analog` is more appropriate. Can you replace all occurrences of the word with `analog` please?

src/platform/input/game-pads.js
```diff
@@ -74,56 +115,600 @@ const MAPS = {
...
+            this.fire(EVENT_GAMEPADDISCONNECTED, current[padIndex]);
+            current.splice(padIndex, 1);
+        }
     }
 
     /**
-     * Update the current and previous state of the gamepads. This must be called every frame for
-     * `wasPressed` to work.
+     * Update the previous state of the gamepads. This must be called every frame for
+     * `wasPressed` as `wasTouched` to work.
```
willeastcott, line 695: `as` -> `and`?

src/platform/input/game-pads.js
willeastcott: Fine by me.

willeastcott:
Can you provide a public API change summary in the PR description please?

Also, can you confirm that this PR is backwards-compatible....or does it introduce any breaking changes?

src/platform/input/game-pads.js
MushAsterion: It's fixed!

src/platform/input/game-pads.js
MushAsterion: It's fixed, thanks!

MushAsterion:
What do you mean by "changes"? Everything is on same format as before, there is no breaking change or change to how previous functions were working. All the changes were made in addition to this.

willeastcott:
OK, that's great. It just wasn't obvious whether this PR is backwards compatible but it seems it is. Just wanted to be absolutely clear on that. 😄

MushAsterion:
It is, even if it was really hard to keep it as it was very raw. The only behavior that may have changed is that now functions regarding buttons on the `GamePads` may not lead to same button as now it is properly mapped. However that behavior was never clearly documented to be on the button index from the API and is more beneficial the way it is now. May you want me to revert that let me know!

src/platform/input/game-pads.js
willeastcott: Can we consider this conversation resolved? Anything left to be decided here?

willeastcott:
If you believe the mapping is now correct, I think it's fine and doesn't need reverting.

willeastcott (approved):
This looks in great shape now so I'm approving. But I'll wait for at least one other approval before merging. @MushAsterion - is there anything else you want to change before this gets merged?

MushAsterion:
Everything good for me, thanks a looooooooooot to @kungfooman for being so patient and helping with all the reviews 💪

src/platform/input/game-pads.js
```diff
@@ -68,62 +109,619 @@ const MAPS = {
...
+
+        return Promise.resolve(false);
+    }
+
+    /**
+     * Return true if the gamepad is connected.
+     *
+     * @returns {boolean} Return true if the gamepad is connected.
+     */
+    isConnected() {
```
kungfooman, line 442: Should be a getter too to keep it coherent

src/platform/input/game-pads.js
```diff
@@ -136,109 +734,115 @@ class GamePads {
...
+        return pad ? pad.getAxis(axis) : 0;
+    }
 
-        const key = this.current[index].map.axes[axes];
-        let value = this.current[index].pad.axes[pc[key]];
+    /**
+     * Find a connected {@link GamePad} from its identifier.
+     *
+     * @param {string} id - The identifier to search for.
+     * @returns {GamePad} The {@link GamePad} with the matching identifier or null if no gamepad is found or the gamepad is not connected.
```
kungfooman, line 831: Return value is `GamePad | null`

src/platform/input/game-pads.js
```diff
@@ -136,109 +734,115 @@ class GamePads {
...
 
-        if (Math.abs(value) < this.deadZone) {
-            value = 0;
-        }
-        return value;
+    /**
+     * Find a connected {@link GamePad} from its device index.
+     *
+     * @param {number} index - The device index to search for.
+     * @returns {GamePad} The {@link GamePad} with the matching device index or null if no gamepad is found or the gamepad is not connected.
```
kungfooman, line 841: Return value is `GamePad | null`

src/platform/input/game-pads.js
```diff
@@ -136,109 +734,115 @@ class GamePads {
...
+     * Destroy the event listeners.
+     *
+     * @ignore
+     */
+    destroy() {
+        window.removeEventListener('gamepadconnected', this._ongamepadconnectedHandler, false);
+        window.removeEventListener('gamepaddisconnected', this._ongamepaddisconnectedHandler, false);
+    }
+
     getMap(pad) {
```
kungfooman, line 766: Just optional, but I would still like typing here for type safety

kungfooman (changes requested):
I am happy about the changes and the pace you worked on it @MushAsterion :1st_place_medal: 

@willeastcott the deadZone / axis buttons issue is not really resolved, but this PR is so full already I would rather discuss that in a new PR and it's not breaking anything.

src/platform/input/game-pads.js
```diff
@@ -147,6 +147,7 @@ const PRODUCT_CODES = {
 };
 
 let deadZone = 0.25;
+const previous = [];
```
kungfooman, line 150: I am not a big fan of these globals as they conflict with each when running multiple `pc.Application`'s with two `pc.GamePads`

IMO it should be `GamaPads#_previous = [];`

src/platform/input/game-pads.js
```diff
@@ -68,62 +109,619 @@ const MAPS = {
...
+     * @hideconstructor
+     */
+    constructor(button) {
+        /**
+         * Whether this button was pressed on last frame.
+         *
+         * @type {boolean}
+         * @ignore
+         */
+        this._previouslyPressed = button.pressed;
```
kungfooman, line 169: This looks like `this.pressed`, while...

src/platform/input/game-pads.js
```diff
@@ -68,62 +109,619 @@ const MAPS = {
...
+
+        /**
+         * Whether this button was touched on last frame.
+         *
+         * @type {boolean}
+         * @ignore
+         */
+        this._previouslyTouched = typeof button.touched === 'boolean' ? button.touched : button.value > 0;
```
kungfooman, line 177: ... this looks like `this.touched`... and both together looks like calling `this.updatePrevious()` to reduce code duplication, so this can be a constructor as simple as two lines:

```js
class GamePadButton {
    /**
     * The original Gamepad API gamepad button.
     *
     * @type {GamepadButton}
     * @ignore
     */
    _button;

    /**
     * Whether this button was pressed on last frame.
     *
     * @type {boolean}
     * @ignore
     */
    _previouslyPressed;

    /**
     * Whether this button was touched on last frame.
     *
     * @type {boolean}
     * @ignore
     */
    _previouslyTouched;

    /**
     * Create a new GamePadButton instance.
     *
     * @param {GamepadButton} button - The original Gamepad API gamepad button.
     * @hideconstructor
     */
    constructor(button) {
        this._button = button;
        this.updatePrevious();
    }
```

kungfooman (changes requested):

src/platform/input/game-pads.js
MushAsterion: It's fixed!

src/platform/input/game-pads.js
MushAsterion: It's fixed!

src/platform/input/game-pads.js
MushAsterion: It's fixed!

Maksims:
As this PR improves access and interaction with gamepads, does it apply same improvements to WebXR GamePads?

src/platform/input/game-pads.js
MushAsterion: It's changed, thanks!

MushAsterion:
It does, at least for the browsers considering WebXR Gamepads as gamepads, which in short means Mozilla as all other browsers consider them as WebXR Input Devices

src/platform/input/game-pads.js
MushAsterion: It's fixed, I used your docs, for it. Hope you won't mind!

src/platform/input/game-pads.js
MushAsterion: It's changed, thanks!

src/platform/input/game-pads.js
MushAsterion: It's changed, thanks!

kungfooman:
I will celebrate the day when PC supports typedefs :sweat_smile: :+1: 

From my side it looks as good as it gets now, good job :100:

MushAsterion:
Just noticed I forgot to add a global API for pulsing as it was the only missing, that's done now! With `GamePads#pulse(orderIndex, intensity, duration, options)` and `GamePads#pulseAll(intensity, duration, options)`

src/platform/input/game-pads.js
```diff
@@ -136,109 +751,159 @@ class GamePads {
...
-        if (!this.current[index]) {
-            return 0;
-        }
+    pulseAll(intensity, duration, options) {
+        return new Promise((resolve, reject) => {
+            Promise.all(
+                this.current.map(pad => pad.pulse(intensity, duration, options))
+            ).then(resolve).catch(reject);
+        });
+    }
```
kungfooman, line 886:
```suggestion
    pulseAll(intensity, duration, options) {
        return Promise.all(
            this.current.map(pad => pad.pulse(intensity, duration, options))
        );
    }
```

src/platform/input/game-pads.js
```diff
@@ -68,62 +109,636 @@ const MAPS = {
...
+        const actuator = this.pad.vibrationActuator;
+
+        if (actuator) {
+            if (actuator.playEffect) {
+                if (options) {
+                    return actuator.playEffect(options.type === 'dual-rumble' ? 'dual-rumble' : 'vibration', {
+                        'duration': duration,
+                        'startDelay': options.startDelay || 0,
+                        'strongMagnitude': typeof options.strongMagnitude === 'number' ? options.strongMagnitude : intensity,
+                        'weakMagnitude': typeof options.weakMagnitude === 'number' ? options.weakMagnitude : intensity
```
kungfooman, line 425:
```suggestion
                        'strongMagnitude': options.strongMagnitude ?? intensity,
                        'weakMagnitude': options.weakMagnitude ?? intensity
```
@willeastcott started to clean up the code in other places and we can do that here too

src/platform/input/game-pads.js
```diff
@@ -68,62 +109,636 @@ const MAPS = {
...
+     * @param {number} [options.weakMagnitude] - Intensity for weak actuators in the range 0 to 1. Defaults to intensity.
+     * @returns {Promise<boolean>} Return a Promise resulting in true if the pulse was successfully completed.
+     */
+    pulse(intensity, duration, options) {
+        const actuator = this.pad.vibrationActuator;
+
+        if (actuator) {
+            if (actuator.playEffect) {
+                if (options) {
+                    return actuator.playEffect(options.type === 'dual-rumble' ? 'dual-rumble' : 'vibration', {
```
kungfooman, line 421:
```suggestion
                    return actuator.playEffect(options.type ?? actuator.type, {
```
If the actuator doesn't support vibration, the promise throws, so I think it's best to stick to what type it says it supports by default?

src/platform/input/game-pads.js
```diff
@@ -68,62 +109,636 @@ const MAPS = {
...
+                        'duration': duration,
+                        'startDelay': options.startDelay || 0,
+                        'strongMagnitude': typeof options.strongMagnitude === 'number' ? options.strongMagnitude : intensity,
+                        'weakMagnitude': typeof options.weakMagnitude === 'number' ? options.weakMagnitude : intensity
+                    });
+                }
+
+                return actuator.playEffect('vibration', {
```
kungfooman, line 429:
```suggestion
                return actuator.playEffect(actuator.type, {
```

Same here, otherwise it throws if it doesn't support "vibration" (like a PS4 controller).

src/platform/input/game-pads.js
```diff
@@ -68,62 +109,636 @@ const MAPS = {
...
+                    'duration': duration,
+                    'startDelay': 0,
+                    'strongMagnitude': intensity,
+                    'weakMagnitude': intensity
```
kungfooman, line 433:
```suggestion
                    duration,
                    startDelay: 0,
                    strongMagnitude: intensity,
                    weakMagnitude: intensity
```

(removing bunch of unneeded ascii characters)

src/platform/input/game-pads.js
```diff
@@ -68,62 +109,636 @@ const MAPS = {
...
+     * Make the gamepad vibrate.
+     *
+     * @param {number} intensity - Intensity for the vibration in the range 0 to 1.
+     * @param {number} duration - Duration for the vibration in milliseconds.
+     * @param {object} [options] - Options for special vibration pattern.
+     * @param {string} [options.type] - Type of pattern. Available types are "dual-rumble" and "vibration". Defaults to "vibration".
+     * @param {number} [options.startDelay] - Delay before the pattern starts, in milliseconds. Defaults to 0.
+     * @param {number} [options.strongMagnitude] - Intensity for strong actuators in the range 0 to 1. Defaults to intensity.
+     * @param {number} [options.weakMagnitude] - Intensity for weak actuators in the range 0 to 1. Defaults to intensity.
+     * @returns {Promise<boolean>} Return a Promise resulting in true if the pulse was successfully completed.
```
kungfooman, line 413: `actuator.playEffect` on Chrome returns a string:

image 

(`"preempted"` happens if you plug the USB out while it vibrates)

I can't test Firefox, because it doesn't even give access to `vibrationActuator`.

I don't know when this was changed, at least this link says it should be a boolean, yet it isn't: https://developer.mozilla.org/en-US/docs/Web/API/GamepadHapticActuator/playEffect

MushAsterion: Nice spot! I fixed it by forcing into boolean and added support for Mozilla.

kungfooman:
You are introducing a callback hell here (which doesn't work aswell).

Modern JavaScript is designed to exactly prevent that. Please mark the function `async` and use `await` inside.

You can also make use of helper functions, like:

```js
/**
 * @param {number} ms 
 * @returns {Promise<void>}
 */
function sleep(ms) {
  return new Promise(resolve => setTimeout(resolve, ms));
}
```

You can then rewrite something like this:

```js
                        else if (actuator.pulse) {
                            if (options && options.startDelay) {
                                // Custom delay
                                return new Promise(function (resolve, reject) {
                                    setTimeout(function () {
                                        actuator.pulse(intensity, duration).then(resolve).catch(reject);
                                    }, options.startDelay);
                                });
                            }

                            return actuator.pulse(intensity, duration);
                        }
```

into something like this:

```js
                            const startDelay = options?.startDelay ?? 0;
                            await sleep(startDelay);
                            return await actuator.pulse(intensity, duration);
```

(other stuff needs to be rewritten aswell, but you can do that to practice promises)

Parts of the code also reach the 9th level of nesting, which usually reflects that something strange is going on (with lots of potential to boil it down to something simpler).

src/platform/input/game-pads.js
```diff
@@ -68,62 +109,651 @@ const MAPS = {
...
+
+            for (let i = 0, l = results.length; i < l; i++) {
+                const result = results[i];
+
+                if (result === false || typeof result === 'string' && result !== 'complete') {
+                    return false;
+                }
+            }
+
+            return !!results.length;
```
kungfooman, line 461: Can't we say that the pulse was successful if at least one actuator vibrated?

MushAsterion: That's a more positive way a thinking, that's updated!

src/platform/input/game-pads.js
```diff
@@ -68,62 +109,645 @@ const MAPS = {
...
+        if (actuators.length) {
+            const startDelay = options?.startDelay ?? 0;
+            const strongMagnitude = options?.strongMagnitude ?? intensity;
+            const weakMagnitude = options?.weakMagnitude ?? intensity;
+
+            const results = await Promise.all(
+                actuators.map(async (actuator) => {
+                    if (!actuator) {
+                        return true;
+                    }
```
kungfooman, line 437: Is this check really needed? My test on Linux/Firefox doesn't show any `[null]` array or something. And if this is really needed, I guess it should return `false` instead?

image

src/platform/input/game-pads.js
```diff
@@ -68,62 +109,645 @@ const MAPS = {
...
 };
 
 const PRODUCT_CODES = {
     'Product: 0268': 'PS3'
 };
 
+let deadZone = 0.25;
+
+/**
+ * @param {number} ms - Number of miliseconds to sleep for.
```
kungfooman, line 152:
```suggestion
 * @param {number} ms - Number of milliseconds to sleep for.
```

src/platform/input/game-pads.js
```diff
@@ -68,62 +109,645 @@ const MAPS = {
...
+        return this.map.buttons ? this.map.buttons.map(b => this._getButton(b)) : [];
+    }
+
+    /**
+     * Make the gamepad vibrate.
+     *
+     * @param {number} intensity - Intensity for the vibration in the range 0 to 1.
+     * @param {number} duration - Duration for the vibration in milliseconds.
+     * @param {object} [options] - Options for special vibration pattern.
+     * @param {string} [options.type] - Type of pattern. Available types are "dual-rumble" and "vibration". Defaults to "vibration".
```
kungfooman, line 419: Doesn't default to "vibration" anymore

src/platform/input/game-pads.js
```diff
@@ -136,109 +760,157 @@ class GamePads {
...
-        // If this is the first time frame a pad has been detected
-        return !this.current[index].pad.buttons[i].pressed && (this.previous[index] && this.previous[index][i]);
+    /**
+     * Make the gamepad vibrate.
+     *
+     * @param {number} orderIndex - The index of the pad to check, use constants {@link PAD_1}, {@link PAD_2}, etc. For gamepad index call the function from the pad.
+     * @param {number} intensity - Intensity for the vibration in the range 0 to 1.
+     * @param {number} duration - Duration for the vibration in milliseconds.
+     * @param {object} [options] - Options for special vibration pattern.
+     * @param {string} [options.type] - Type of pattern. Available types are "dual-rumble" and "vibration". Defaults to "vibration".
```
kungfooman, line 866: Doesn't default to "vibration" anymore

src/platform/input/game-pads.js
MushAsterion: There is no clear documentation on whether or not it's nullable, so I went better safe than sorry. And no if there is none it will return an error when checking if the functions exist inside.

src/platform/input/game-pads.js
MushAsterion: Fixed, thanks!

src/platform/input/game-pads.js
MushAsterion: Fixed, thanks!

kungfooman:
I would love to call this finished, so I just tested on Linux/Chrome and everything works nicely. Then I tested on Linux/Firefox and `GamePads#isPressed`, `GamePads#wasPressed` and `GamePads#wasReleased` doesn't work lol...

On Linux/Firefox it throws:

image

image

The reason for this I assume is that MAPS mapping is implying more buttons, but Linux/Firefox only has 13:

image

But 17 on Linux/Chrome (which fits the default mapping):

image

MushAsterion:
Ugh... As there is no easy documentation on this, I added the possibility for users to change the map using `GamePad#updateMap` (documented with an example).

I also added possibility for `MAPS` to support raw mapping using "`RAW_` + product" name.

`GamePad#mapping` was also changed so now it's either `"xr-standard"`, `"standard"`, `""` or `"custom"` where `"custom"` is when user changed the map at one moment. Custom maps are saved into a variable in case the gamepad disconnects.

And I updated the example of `gamepadconnected` event to add a check if map is detected by system so users don't forget to update mapping in that case.

MushAsterion:
> On Linux/Firefox it throws:
> 
> image
> 
> image

I don't know where you are taking that code from but if a button doesn't exist it should still return a value from my code.

However I added a new default, it might not fit all cases but is still some kind of safe-web in case the browser is not able to detect the relevant mapping as it shows in your screenshot (`mapping: ""`).

This new default is only used if the `getMap` function was going to return the default mapping but there are less buttons in the gamepad than in the mapping:

```JS
const defaultmap = MAPS.DEFAULT;
const map = pad.buttons.length < defaultmap.buttons.length ? MAPS.DEFAULT_DUAL : defaultmap;
map.mapping = pad.mapping;
return map;
```

kungfooman:
Sorry, I got a bit confused there... I tested different versions and then I thought I was on this PR 

I would like to suggest a little bit bigger refactor about dualButton handling, but I'm not sure if this should still go in this PR

There are also some other issues I realized while testing around, I have a bit more time today to review

src/platform/input/game-pads.js
```diff
@@ -68,62 +152,757 @@ const MAPS = {
...
+         * @type {boolean[][]}
+         * @ignore
+         */
+        this._previous = [];
+
+        this._ongamepadconnectedHandler = this._ongamepadconnected.bind(this);
+        this._ongamepaddisconnectedHandler = this._ongamepaddisconnected.bind(this);
+
+        window.addEventListener('gamepadconnected', this._ongamepadconnectedHandler, false);
+        window.addEventListener('gamepaddisconnected', this._ongamepadconnectedHandler, false);
```
kungfooman, line 778:
```suggestion
        window.addEventListener('gamepaddisconnected', this._ongamepaddisconnectedHandler, false);
```

src/platform/input/game-pads.js
```diff
@@ -68,62 +152,757 @@ const MAPS = {
...
+            current.splice(padIndex, 1);
+            padIndex = current.findIndex(gp => gp.index === pad.index);
+        }
+
+        current.push(pad);
+        this.fire(EVENT_GAMEPADCONNECTED, pad);
+    }
+
+    /**
+     * Callback function when a gamepad is connection.
```
kungfooman, line 872:
```suggestion
     * Callback function when a gamepad is disconnecting.
```

Like this maybe? Or "disconnected"?

src/platform/input/game-pads.js
```diff
@@ -1,3 +1,45 @@
+import { EventHandler } from '../../core/event-handler.js';
+import { EVENT_GAMEPADCONNECTED, EVENT_GAMEPADDISCONNECTED, PAD_FACE_1, PAD_FACE_2, PAD_FACE_3, PAD_FACE_4, PAD_L_SHOULDER_1, PAD_R_SHOULDER_1, PAD_L_SHOULDER_2, PAD_R_SHOULDER_2, PAD_SELECT, PAD_START, PAD_L_STICK_BUTTON, PAD_R_STICK_BUTTON, PAD_UP, PAD_DOWN, PAD_LEFT, PAD_RIGHT, PAD_VENDOR, XRPAD_TRIGGER, XRPAD_SQUEEZE, XRPAD_TOUCHPAD_BUTTON, XRPAD_STICK_BUTTON, XRPAD_A, XRPAD_B, PAD_L_STICK_X, PAD_L_STICK_Y, PAD_R_STICK_X, PAD_R_STICK_Y, XRPAD_TOUCHPAD_X, XRPAD_TOUCHPAD_Y, XRPAD_STICK_X, XRPAD_STICK_Y } from './constants.js';
+import { math } from '../../core/math/math';
```
kungfooman, line 3:
```suggestion
import { math } from '../../core/math/math.js';
```

Always use full path, this is how ES6 is implemented in every modern browser; everything else is just an attitude of build systems.

kungfooman:
Okay, suggesting some refactor, because:

1) Right now the dualButtons code simply doesn't work for me
2) It is producing lots of garbage
3) It is undynamic / too hardcoded

1 + 2)

Buttons are created all the time:
```js
                new GamePadButton({
                    pressed: value === 1,
                    touched: !!value,
                    value
                });
```

Access to `axes` is also implemented as a map operation: `return this.map.axes ? this.map.axes.map(a => this.pad.axes[MAPS_INDEXES.axes[a]] || 0) : [];`

It fails with:

image

Because `this.axes` only contains 4 values:

image

The gamepad actually has 8:

image

I refactored this myself some time ago, and I settled on this design:

```js
        buttonsSynthesized: {
            PAD_LEFT : gamepad => gamepad.axes[4] === -1,
            PAD_RIGHT: gamepad => gamepad.axes[4] ===  1,
            PAD_UP   : gamepad => gamepad.axes[5] === -1,
            PAD_DOWN : gamepad => gamepad.axes[5] ===  1,
        }
```

This also allows you to use other axes in a more dynamic way, for example:

```js
        buttonsSynthesized: {
            PAD_LEFT : gamepad => gamepad.axes[0] < -0.4,
            PAD_RIGHT: gamepad => gamepad.axes[0] >  0.4,
            PAD_UP   : gamepad => gamepad.axes[1] < -0.4,
            PAD_DOWN : gamepad => gamepad.axes[1] >  0.4,
        }
```

(this removes all the empty `[[], [], [], ...]` just like `indexOf finding` etc., because the mapping is as straight forward as it gets)

MushAsterion:
Ah nice spot, I was using `axes` instead of `pad.axes` which was wrong.

So I was testing out and found that GamePadButton was not working properly on constantly pooling browser (like Firefox) as previous value was pulled within same frame as current value making all `was` function working since the beginning of this frame and not previous one. Same was happening with axes. Therefore I updated the classes to save the values which also now save one update when polling on every frame.

Problem with your refractor is that is does not support previous values, I updated to be an object with information inside, hope that will sound good to you.

src/platform/input/game-pads.js
```diff
@@ -2,6 +2,17 @@ import { EventHandler } from '../../core/event-handler.js';
...
 import { math } from '../../core/math/math.js';
 
+/**
+ * Fetch Gamepads from API.
+ *
+ * @type {Function}
+ * @returns {Gamepad[]} Retrieved gamepads from the device.
+ * @ignore
+ */
+const fetchGamepads = (navigator.getGamepads || navigator.webkitGetGamepads || function () {
```
kungfooman: I worked this out in my PR which makes sure that 1) no garbage is produced by reusing `[]` while 2) supporting node, this looks like it's accessing navigator without

MushAsterion: Yeah but as your PR is still not live so I cannot use it... Once it gets merged I'll be more than happy to use it!
Regarding 2 I already corrected it right after this commit, you are probably reviewing on an older version...

kungfooman: Yea, I just realized, for some reason the entire file in "Files changed" was still in the old state, so I could comment on that line

kungfooman:
I find it unfortunate that typedefs are not supported in the engine due to JSDoc, causing avoidable errors.

Would you be willing to reconsider your opinion on this, @willeastcott?

If we properly type the existing code, Intellisense points out everything so nicely:

image

Via these typedefs:

```js
/**
 * @typedef {object} SynthesizedButton
 * @property {number} axis
 * @property {number} min
 * @property {number} max
 */

/**
 * An object defining the order of buttons and axes for a HTML5 Gamepad.
 *
 * @typedef {object} ButtonsAxes
 * @property {string[]} buttons - Order of PAD_FACE_1, PAD_FACE_2, ...
 * @property {string[]} axes - Order of PAD_L_STICK_X, PAD_L_STICK_Y, ...
 * @property {Record<string, SynthesizedButton>} [synthesizedButtons] - Extra information to turn an axis into a button
 */

/**
 * @type {Record<string, ButtonsAxes>}
 */

const MAPS = {
```

But right now, the only place which displays type errors is the debug console :see_no_evil: 

image

We are creating a back-and-forth for issues that are pretty much 100% avoidable if we simply stick to JavaScript 101.

src/platform/input/game-pads.js
```diff
@@ -68,62 +166,763 @@ const MAPS = {
...
+     */
+    _previouslyTouched;
+
+    /**
+     * The original Gamepad API gamepad button.
+     *
+     * @type {GamepadButton}
+     * @ignore
+     */
+    _button;
```
kungfooman, line 310: We can get rid of `_button` now after you fixed the Firefox polling issue? Makes the shape of the class a bit smaller + less lines

src/platform/input/game-pads.js
```diff
@@ -68,62 +166,763 @@ const MAPS = {
...
+    constructor(current, previous) {
+        if (typeof current === 'number') {
+            this.value = current;
+            this.pressed = current === 1;
+            this.touched = current > 0;
+        } else {
+            this.value = current.value;
+            this.pressed = current.pressed;
+            this.touched = current.touched ?? current.value > 0;
+            this._button = current;
```
kungfooman, line 328: Can also be deleted

src/platform/input/game-pads.js
```diff
@@ -68,62 +166,763 @@ const MAPS = {
...
+
+        /**
+         * The buttons present on the GamePad. Some buttons may be null. Order is provided by API, use GamePad#buttons instead.
+         *
+         * @type {GamePadButton[]}
+         * @ignore
+         */
+        this._buttons = gamepad.buttons.map((b) => {
+            return b ? new GamePadButton(b) : null;
+        });
```
kungfooman, line 430:
```suggestion
        this._buttons = gamepad.buttons.map((b) => {
            return new GamePadButton(b);
        });
```
Can a native controller button really be `null`? And if so, then the type would be `(GamePadButton|null)[]`.

src/platform/input/game-pads.js
MushAsterion: It's done!

src/platform/input/game-pads.js
MushAsterion: Also done!

src/platform/input/game-pads.js
MushAsterion: Yes it can be as per XR Gamepads documentation. And no the type definition in parenthesis is not supported, I already tried in previous commit but just did again and still failing...

kungfooman: Can you read it again? It says that it exactly cannot be `null`:

> Devices that lack one of the optional inputs listed in the tables above MUST preserve their place in the [buttons](https://www.w3.org/TR/gamepad/#dom-gamepad-buttons) or [axes](https://www.w3.org/TR/gamepad/#dom-gamepad-axes) array, reporting a [placeholder button](https://immersive-web.github.io/webxr-gamepads-module/#placeholder-button) or [placeholder axis](https://immersive-web.github.io/webxr-gamepads-module/#placeholder-axis), respectively. A placeholder button MUST report 0 for [value](https://www.w3.org/TR/gamepad/#dom-gamepadbutton-value), false for [pressed](https://www.w3.org/TR/gamepad/#dom-gamepadbutton-pressed), and false for [touched](https://www.w3.org/TR/gamepad/#dom-gamepadbutton-touched). A placeholder axis MUST report 0. [Placeholder buttons](https://immersive-web.github.io/webxr-gamepads-module/#placeholder-button) and [axes](https://immersive-web.github.io/webxr-gamepads-module/#placeholder-axis) MUST be omitted if they are the last element in the array or all following elements are also [placeholder buttons](https://immersive-web.github.io/webxr-gamepads-module/#placeholder-button) or [axes](https://immersive-web.github.io/webxr-gamepads-module/#placeholder-axis).

src/platform/input/game-pads.js
```diff
@@ -68,62 +166,754 @@ const MAPS = {
...
+
+    /**
+     * Return true if the button was released since the last update.
+     *
+     * @param {number} button - The button to test, use constants {@link PAD_FACE_1}, etc.
+     * @returns {boolean} Return true if the button was released, false if not.
+     */
+    wasReleased(button) {
+        const b = this.getButton(button);
+        return b && b.wasReleased;
```
kungfooman, line 706:
```suggestion
        return this.getButton(button)?.wasReleased || false;
```

Just to make it type safe, right now this is possibly returning `null`:

image

Same for isPressed/wasPressed/isTouched/wasTouched

src/platform/input/game-pads.js
MushAsterion: Oh right. I was more relying on the table with "required no" and miss that statement.

src/platform/input/game-pads.js
MushAsterion: Oh that's interesting, I was thinking that a logical operator was always returning boolean. That's fixed, thanks a lot!

kungfooman:
Just for brain storming I would like to share my contemplations that I got after rewriting everything a little bit. I wasn't happy how complicated lots of the code is, e.g. the mapping just never fully made sense to me. I kinda could figure the mappings out, but not really through understanding, but testing around "until it happens to work".

What I did now is merging `buttons` with `synthesizedButtons`:

```js
/** @type {ButtonsAxes} */
const PS4LinuxFirefox = {
    buttons: {
        [PAD_FACE_1        ]: gamepad => gamepad.buttons[ 0], // X
        [PAD_FACE_2        ]: gamepad => gamepad.buttons[ 1], // O
        [PAD_FACE_3        ]: gamepad => gamepad.buttons[ 2], // TRI
        [PAD_FACE_4        ]: gamepad => gamepad.buttons[ 3], // SQUARE
        [PAD_L_SHOULDER_1  ]: gamepad => gamepad.buttons[ 4],
        [PAD_R_SHOULDER_1  ]: gamepad => gamepad.buttons[ 5],
        [PAD_L_SHOULDER_2  ]: gamepad => gamepad.buttons[ 6],
        [PAD_R_SHOULDER_2  ]: gamepad => gamepad.buttons[ 7],
        [PAD_SELECT        ]: gamepad => gamepad.buttons[ 8], // "SHARE" on pad
        [PAD_START         ]: gamepad => gamepad.buttons[ 9], // "OPTIONS" on pad
        [PAD_VENDOR        ]: gamepad => gamepad.buttons[10],
        [PAD_L_STICK_BUTTON]: gamepad => gamepad.buttons[11],
        [PAD_R_STICK_BUTTON]: gamepad => gamepad.buttons[12],
        [PAD_LEFT          ]: gamepad => gamepad.axes[0] < -0.4 || gamepad.axes[6] === -1,
        [PAD_RIGHT         ]: gamepad => gamepad.axes[0] >  0.4 || gamepad.axes[6] ===  1,
        [PAD_UP            ]: gamepad => gamepad.axes[1] < -0.4 || gamepad.axes[7] === -1,
        [PAD_DOWN          ]: gamepad => gamepad.axes[1] >  0.4 || gamepad.axes[7] ===  1,
    },
    axes: [
        // Analog Sticks
        'PAD_L_STICK_X',
        'PAD_L_STICK_Y',
        'PAD_R_STICK_X',
        'PAD_R_STICK_Y'
    ],
}
```

I have the feeling I could unify the axes aswell, but right now I'm in a state that it works for me in Chrome/Firefox to the point of what my project requires (aka I don't need axes for it).

Since you realized/fixed the Firefox polling issue, I could also simplify the `GamePadButton` class a bit:

```js
/**
 * A GamePadButton stores information about a button from the Gamepad API.
 */
class GamePadButton {
    /**
     * @param {boolean} pressed - New `pressed` state.
     * @param {boolean} touched - New `touched` state.
     * @param {number} value - New `value` state.
     */
    update(pressed, touched, value) {
        this.wasPressed  = !this.pressed && pressed;
        this.wasReleased = this.pressed && !pressed;
        this.wasTouched = !this.touched && touched;
        this.pressed = pressed;
        this.touched = touched;
        this.value   = value;
    }

    /**
     * Whether the button is currently down.
     *
     * @type {boolean}
     */
    pressed = false;

    /**
     * Whether the button was pressed.
     *
     * @type {boolean}
     */
    wasPressed = false;

    /**
     * Whether the button was released since the last update.
     *
     * @type {boolean}
     */
    wasReleased = false;

    /**
     * Whether the button is currently touched.
     *
     * @type {boolean}
     */
    touched = false;

    /**
     * Whether the button was touched since the last update.
     *
     * @type {boolean}
     */
    wasTouched = false;

    /**
     * The value for the button between 0 and 1, with 0 representing a button that is not pressed, and 1 representing a button that is fully pressed.
     *
     * @type {number}
     */
    value = 0;
}
```

Updating the buttons from game pads works like this then, in `GamePad#update`:

```js
    /**
     * Update the existing GamePad Instance.
     *
     * @param {Gamepad} gamepad - The original Gamepad API gamepad.
     * @ignore
     */
    update(gamepad) {
        this.pad = gamepad;
        for (const b in this.map.buttons) {
            const test = this.map.buttons[b];
            const ret = test(this.pad);
            let pressed = false;
            let touched = false;
            let value = 0;
            if (ret instanceof GamepadButton) {
                pressed = ret.pressed;
                touched = ret.touched ?? ret.value > 0;
                value   = ret.value;
            } else {
                if (ret) {
                    pressed = true;
                    touched = true;
                    value = 1;
                }
            }
            this.buttons[b].update(pressed, touched, value);
        }
        return this;
    }
```

Since `GamePadButton` is now `constructor`-free, this is also easily possible:

```js
const dummyButton = Object.freeze(new GamePadButton());
```

which helps to reduce the amount of `?.` operators and `null` typings:

```js
    getButton(index) {
        return this.buttons[index] || dummyButton;
    }
    isPressed(index) {
        return this.getButton(index).pressed;
    }
    wasPressed(index) {
        return this.getButton(index).wasPressed;
    }
```

MushAsterion:
Thank you for the feedback.

So I reduced a bit `GamePadButton` and introduced a "dummy" one as placeholder. However regarding your main and first change the problem I have with it is that it is a breaking change and I don't want to introduce a breaking change. The current API exposes the map even if it is not being used.

willeastcott:
I did already approve this PR some time ago, but it has gone through some more changes since then. Are any more changes expected, or is this now considered 'done' (barring any further feedback of course)? 😄

MushAsterion:
I need a bit more changes, just realized some stuff are not working. Setting to draft and will make it ready once I'm done. Sorry for that!

kungfooman:
I tested the functionality with my own gamepad project + @yaustar's `DebugGamepadFlyCamera` and I can't see any issues and I'm happy about the direction the code evolved.

I still find the lack of `@typedef` unfortunate, but maybe someone finds time at some point to fix JSDoc (e.g. to simply display typedef information inline instead of generating 404 links).

I consider the Promise issue to be resolved and that Promises generally are accepted now (since @mvaligursky uses them too and there is no way around that anyway, since modern API's return them).

Only thing that looks a little bit too defensive to me is this:

```js
        if (this.gamepadsSupported) {
            const padDevices = fetchGamepads();

            for (let i = 0, len = padDevices.length; i < len; i++) {
                if (padDevices[i]) {
                    const pad = this.findByIndex(padDevices[i].index);

                    if (pad) {
                        pads.push(pad.update(padDevices[i]));
                    } else {
                        const nPad = new GamePad(padDevices[i], this.getMap(padDevices[i]));
                        this.current.push(nPad);
                        pads.push(nPad);
                    }
                }
            }
        }
```

It could be just:

```js
        const padDevices = fetchGamepads();
        for (let i = 0, len = padDevices.length; i < len; i++) {
            if (padDevices[i]) {
                const pad = this.findByIndex(padDevices[i].index);

                if (pad) {
                    pads.push(pad.update(padDevices[i]));
                } else {
                    const nPad = new GamePad(padDevices[i], this.getMap(padDevices[i]));
                    this.current.push(nPad);
                    pads.push(nPad);
                }
            }
        }
```

The point with the dummy array is that it works in any case (even e.g. in nodejs) and you don't need to check anymore if gamepads are supported. I would also like to see `GamePads#gamepadsSupported` deprecated in favor of the already existing `pc.platform.gamepads`, otherwise we have multiple ways of doing the same things or "sources of truth".

By now its only my nitpicking: I would prefer `getGamepads` over `fetchGamepads`, because the term `fetch` has a connotation of HTTP requests and responses for me and the original API only uses `get` aswell.

But I'm really impressed by this pull request: it took three weeks to complete and it was worth it. The work is well done and it looks great. I'm very happy with the results and the attention to detail that you put into it. Thank you for all your hard work!

MushAsterion:
It's all changed!

src/platform/input/game-pads.js
```diff
@@ -68,62 +168,763 @@ const MAPS = {
...
+    set deadZone(value) {
+        deadZone = value;
+    }
+
+    /**
+     * Threshold for axes to return values. Must be between 0 and 1.
+     *
+     * @type {number}
+     * @ignore
+     */
```
willeastcott, line 845: The JSDoc block here should be above the setter.

src/platform/input/game-pads.js
```diff
@@ -68,62 +168,763 @@ const MAPS = {
...
+
+        this._previous.length = this.current.length;
+        return this._previous;
+    }
+
+    /**
+     * Callback function when a gamepad is connecting.
+     *
+     * @param {GamepadEvent} event - The event containing the connecting gamepad.
+     * @ignore
```
willeastcott, line 880: Should this be @private instead?

src/platform/input/game-pads.js
```diff
@@ -68,62 +168,763 @@ const MAPS = {
...
-        this.poll(this.current);
+        current.push(pad);
+        this.fire(EVENT_GAMEPADCONNECTED, pad);
+    }
+
+    /**
+     * Callback function when a gamepad is disconnecting.
+     *
+     * @param {GamepadEvent} event - The event containing the disconnecting gamepad.
+     * @ignore
```
willeastcott, line 900:
```suggestion
     @private
```

src/platform/input/game-pads.js
```diff
@@ -68,62 +168,763 @@ const MAPS = {
...
+         * @type {Gamepad}
+         * @ignore
+         */
+        this.pad = gamepad;
+
+        this._compileMapping();
+    }
+
+    /**
+     * Whether the gamepad is connected
```
willeastcott, line 457:
```suggestion
     * Whether the gamepad is connected.
```

src/platform/input/game-pads.js
```diff
@@ -68,62 +168,763 @@ const MAPS = {
...
+    return map;
+}
+
+let deadZone = 0.25;
+
+/**
+ * @param {number} ms - Number of milliseconds to sleep for.
+ * @returns {Promise<void>}
+ */
+function sleep(ms) {
```
willeastcott, line 258: I'm assuming this doesn't appear in the API docs? Just wondered if an @ignore was needed...

src/platform/input/game-pads.js
```diff
@@ -134,111 +935,142 @@ class GamePads {
...
+            this.current.map(pad => pad.pulse(intensity, duration, options))
+        );
+    }
 
-        const key = this.current[index].map.axes[axes];
-        let value = this.current[index].pad.axes[pc[key]];
+    /**
+     * Find a connected {@link GamePad} from its identifier.
+     *
+     * @param {string} id - The identifier to search for.
```
willeastcott, line 1058: Worth enumerating the valid values here? It's not obvious what the ids are.

src/platform/input/game-pads.js
```diff
@@ -134,111 +935,142 @@ class GamePads {
...
+     * @returns {Promise<boolean>} Return a Promise resulting in true if the pulse was successfully completed.
+     */
+    pulse(orderIndex, intensity, duration, options) {
+        const pad = this.current[orderIndex];
+        return pad ? pad.pulse(intensity, duration, options) : Promise.resolve(false);
     }
 
     /**
-     * Get the value of one of the analogue axes of the pad.
+     * Make all gamepads vibrate.
```
willeastcott, line 1039: As far as I'm aware, vibration is not widely supported in browsers (as of today). Worth putting in a note about what happens when the underlying browser doesn't support the operation?

src/platform/input/game-pads.js
```diff
@@ -134,111 +935,142 @@ class GamePads {
...
+     */
+    getAxis(orderIndex, axis) {
+        return this.current[orderIndex]?.getAxis(axis) || 0;
+    }
 
-        // Previous pad buttons may not have been populated yet
-        // If this is the first time frame a pad has been detected
-        return !this.current[index].pad.buttons[i].pressed && (this.previous[index] && this.previous[index][i]);
+    /**
+     * Make the gamepad vibrate.
```
willeastcott, line 1022: As far as I'm aware, vibration is not widely supported in browsers (as of today). Worth putting in a note about what happens when the underlying browser doesn't support the operation?

willeastcott:
OK, @MushAsterion, I've done my final review pass. Only some minor comments to address. Once that's done (and since @kungfooman has also given a thumbs up), I shall merge! 😄

src/platform/input/game-pads.js
willeastcott: https://caniuse.com/mdn-api_gamepadhapticactuator_pulse

src/platform/input/game-pads.js
willeastcott:
```suggestion
     * @private
```

src/platform/input/game-pads.js
MushAsterion: Fixed!

src/platform/input/game-pads.js
MushAsterion: It's not in the class so it should not appear, right?

src/platform/input/game-pads.js
MushAsterion: What do you mean? Id is defined by constructor, it's the name of the gamepad so there is no real value to enumerate...

src/platform/input/game-pads.js
willeastcott: I'm 75% sure it won't appear - worth checking.

src/platform/input/game-pads.js
willeastcott: Oh right. I thought this took `PAD_0`, PAD_1` etc. Fair enough.

src/platform/input/game-pads.js
MushAsterion: In fact only Safari doesn't support it. And if it could not complete it will result in `Promise<false>` so I'm not sure that's really needed.

src/platform/input/game-pads.js
MushAsterion: Ah yeah you were right, it was displayed. That's fixed! Same for getGamepads

src/platform/input/game-pads.js
willeastcott: Oh really? But the 'caniuse' page seems to indicate it's Firefox only, no?

MushAsterion: Because you are looking at the `pulse` function which is only supported by Firefox. Other browsers are using `playEffect`.

MushAsterion: And as `pulse` name is more explicit than `playEffect` I used `pulse` as name for the function. But in the code if you look at it, it selects the proper API function.

willeastcott: Oh, ok, got it.

willeastcott:
What a fantastic contribution. Thanks @MushAsterion, you're a ⭐

slimbuck:
I love how the PR description starts with "In short...", nothing short about this PR! 😄 

Thanks so much @MushAsterion for this legendary contribution!

Now all that's left is to make a kick-ass driving game with force feedback steering wheel.

PR #5050, opened 2023-02-08
WebGPU support for int and uint vertex attributes

mvaligursky:
Shaders are now processed for WebGPU not only to handle uniform buffers, but additionally to handle different vertex formats. Specifically - till now only float based attributes were supported, but this PR adds support for int and uint attributes.

Shader attribute declaration like this:
```
attribute vec4 vertex_position;
```
gets translated to this format, if the vertex format specifies position as TYPE_INT32 for example:
```
attribute ivec4 _vertex_position;
vec4 vertex_position = vec4(_vertex_position);
```

- additionally, removed some manual workaround to handle this specifically for the skinning.
- also, uniform arrays of 1 are handled now correctly - before they were assumed to not be arrays, so this now works:
`uniform float data[1]`

This makes our bitmoji characters work, as they provide positions in `int16` format.
(Note that there is still some morphing issue with them, for a separate PR)

https://user-images.githubusercontent.com/59932779/217568394-c3c6ab3d-e306-4c4e-abea-a5b60731dc45.mov

### DIFF
--- a/src/platform/graphics/shader-processor.js
+++ b/src/platform/graphics/shader-processor.js
@@ -368,17 +368,21 @@ class ShaderProcessor {
             const words = ShaderProcessor.splitToWords(line);
             let type = words[0];
             let name = words[1];
-            let copyCode;
 
             if (shaderDefinitionAttributes.hasOwnProperty(name)) {
                 const semantic = shaderDefinitionAttributes[name];
                 const location = semanticToLocation[semantic];
+
+                Debug.assert(!usedLocations.hasOwnProperty(location),
+                             `WARNING: Two vertex attributes are mapped to the same location in a shader: ${usedLocations[location]} and ${semantic}`);
+                usedLocations[location] = semantic;
 
                 // if vertex format for this attribute is not of a float type, we need to adjust the attribute format, for example we convert
                 //      attribute vec4 vertex_position;
                 // to
                 //      attribute ivec4 _private_vertex_position;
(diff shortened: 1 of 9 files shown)


src/platform/graphics/shader-processor.js
```diff
@@ -353,24 +354,64 @@ class ShaderProcessor {
         return block;
     }
 
-    static processAttributes(attributeLines, shaderDefinitionAttributes) {
+    // extract count from type ('vec3' => 3, 'float' => 1)
+    static getTypeCount(type) {
```
slimbuck, line 358: not crazy about this function name, though probably not very important

src/platform/graphics/shader-processor.js
```diff
@@ -353,24 +354,64 @@ class ShaderProcessor {
...
+    static processAttributes(attributeLines, shaderDefinitionAttributes, processingOptions) {
         let block = '';
         const usedLocations = {};
         attributeLines.forEach((line) => {
             const words = ShaderProcessor.splitToWords(line);
-            const type = words[0];
-            const name = words[1];
+            let type = words[0];
+            let name = words[1];
+            let copyCode;
```
slimbuck, line 371: this declaration should go inside the if below?

slimbuck (approved):
approving with a few tiny comments

src/platform/graphics/shader-processor.js
mvaligursky: yeah, but its internal private, and I didnt want a very long name :)

PR #5051, opened 2023-02-08
Limit number of ‘ignored import property on standard material` warnings

mvaligursky:
to avoid problems like this:
Screenshot 2023-02-08 at 15 25 56

### DIFF
--- a/src/scene/materials/standard-material-validator.js
+++ b/src/scene/materials/standard-material-validator.js
@@ -90,7 +90,7 @@ class StandardMaterialValidator {
                 if (REMOVED[key]) {
                     delete data[key];
                 } else {
-                    Debug.warn(`Ignoring unsupported input property to standard material: ${key}`);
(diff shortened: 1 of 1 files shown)


src/scene/materials/standard-material-validator.js
```diff
@@ -90,7 +90,7 @@ class StandardMaterialValidator {
                 if (REMOVED[key]) {
                     delete data[key];
                 } else {
-                    Debug.warn(`Ignoring unsupported input property to standard material: ${key}`);
+                    Debug.warnOnce(`Ignoring unsupported input property to standard material: ${key}`);
```
slimbuck, line 93: were you going to add material name?

mvaligursky: I looked at it, but we don't have the material there handy, and I don't really want to import it just for a debug log.
Happy to close this PR as well though, now that you've fixed the logging.

slimbuck (approved):

PR #5052, opened 2023-02-08
Fix typo when setting the use AO and Lightmap vertex color flags.

GSterbrant:
Fixes bug where AO wasn't used by shader due to a typo.


slimbuck:
https://github.com/playcanvas/engine/pull/5047 :)

PR #5054, opened 2023-02-09
Further field declaration updates

willeastcott:
This PR rids the engine codebase of about 60 `@property` tags (which are not recognized by Typedoc or the TypeScript compiler when generating our type declarations).

Additional:

* There were several incorrectly typed properties in the `LitOptions` class. This PR aims to fix all of them.
* There appear to be many undocumented properties in `LitOptions`. It's not clear if that was intentional.
* This PR fixes several errors in the JSDoc version of the API reference site.
* Updates the `StencilParameters` constructor to use the `??` operator (nullish coalescing).
* Adds `StencilParameters#clone` to the public API.


src/scene/batching/batch-group.js
```diff
@@ -2,17 +2,59 @@ import { LAYERID_WORLD } from '../constants.js';
...
+    /** @private */
+    _obj = {
+        model: [],
+        element: [],
+        sprite: [],
+        render: []
+    };
+
+    /**
+     * Unique id. Can be assigned to model and element components.
```
mvaligursky, line 22: please mention the render component here too

mvaligursky (approved):

src/scene/batching/batch-group.js
willeastcott: Fixed!

PR #5060, opened 2023-02-09
Updated createGraphicsDevice with easier way to create WebGL1 device

mvaligursky:
Adjusted createGraphicsDevice function:
- WebGL2 and WebGL1 fallbacks are automatically added to the list

So the two typical scenarios:
1. called with **[] or undefined**: WebGL2 and WebGL1 are added to the array internally
2. called with **[WebGPU]** - WebGL2 and WebGL1 are added to the array internally, and are used as a falback if WebGPU is not supported.
But users can still specify these 3 constants in their preferred order.

- added `device.isWebGPU`, instead of existing `device.deviceType === DEVICETYPE_WEBGPU`, as that does not require constant import and is shorter / easier to use.

### DIFF
--- a/src/platform/graphics/graphics-device.js
+++ b/src/platform/graphics/graphics-device.js
@@ -30,12 +30,11 @@ class GraphicsDevice extends EventHandler {
     canvas;
 
     /**
-     * The graphics device type, DEVICETYPE_WEBGL or DEVICETYPE_WEBGPU.
+     * True if the deviceType is WebGPU
      *
-     * @type {string}
-     * @ignore
+     * @type {boolean}
      */
-    deviceType;
+    isWebGPU = false;
 
     /**
      * The scope namespace for shader attributes and variables.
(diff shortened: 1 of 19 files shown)


ellthompson (approved):

src/platform/graphics/graphics-device.js
```diff
@@ -30,12 +30,11 @@ class GraphicsDevice extends EventHandler {
...
     /**
-     * The graphics device type, DEVICETYPE_WEBGL or DEVICETYPE_WEBGPU.
+     * True if the deviceType is WebGPU
      *
-     * @type {string}
-     * @ignore
+     * @type {boolean}
      */
-    deviceType;
+    isWebGPU = false;
```
willeastcott, line 37: Isn't this property's naming a bit different to `WebglGraphicsDevice#webgl2`?

mvaligursky: it is, and I considered renaming webgl2 to isWebGl2. What do you think about this?

willeastcott: Maybe. I think we're starting to mix different styles of capitalization for WebGL/GPU. There's `Webgl`, WebGl`, `WebGL`.

PR #5063, opened 2023-02-09
WebXR - Fixed bug where the wrong number of views is being added

yaustar:
Fixes an error in WebXR on some devices that have more than 2 viewports about view being undefined

image

This is because when we add views to an array in the loop, we are also using the length of that array to calculate how many to add which means the total number to add gets smaller each loop.

### DIFF
--- a/src/framework/xr/xr-manager.js
+++ b/src/framework/xr/xr-manager.js
@@ -696,7 +696,8 @@ class XrManager extends EventHandler {
 
         if (lengthNew > this.views.length) {
             // add new views into list
-            for (let i = 0; i <= (lengthNew - this.views.length); i++) {
+            const viewAddCount = (lengthNew - this.views.length);
+            for (let i = 0; i <= viewAddCount; i++) {
                 let view = this.viewsPool.pop();
                 if (!view) {
                     view = {
--- a/src/framework/xr/xr-manager.js
+++ b/src/framework/xr/xr-manager.js
@@ -694,32 +694,28 @@ class XrManager extends EventHandler {
         const lengthOld = this.views.length;
         const lengthNew = pose.views.length;
 
-        if (lengthNew > this.views.length) {
-            // add new views into list
-            while (lengthNew > this.views.length) {
-                let view = this.viewsPool.pop();
-                if (!view) {
-                    view = {
-                        viewport: new Vec4(),
-                        projMat: new Mat4(),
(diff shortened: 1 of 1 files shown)


src/framework/xr/xr-manager.js
```diff
@@ -696,7 +696,8 @@ class XrManager extends EventHandler {
 
         if (lengthNew > this.views.length) {
             // add new views into list
-            for (let i = 0; i <= (lengthNew - this.views.length); i++) {
+            const viewAddCount = (lengthNew - this.views.length);
```
mvaligursky, line 699: I was looking at recently when you were refactoring around there, and though it'd be best to replace the if above as as well as the loop with `while (lengthNew > this.views.length) {`

and similar for the else / for loop in the else branch.

yaustar: Done. I don't have a way of testing the logic in the else branch so can you double check the code please 😅

src/framework/xr/xr-manager.js
```diff
@@ -696,8 +696,7 @@ class XrManager extends EventHandler {
 
         if (lengthNew > this.views.length) {
             // add new views into list
-            const viewAddCount = (lengthNew - this.views.length);
-            for (let i = 0; i <= viewAddCount; i++) {
+            while (lengthNew > this.views.length) {
```
mvaligursky, line 699: now get rid of that if, and also else?
the code should just be
```
while (..) {
}

whilte (..) {
}
```
no need for ifs as they just do the same as the first iteration of the loop.
```

yaustar: Done!

mvaligursky (approved):
Approving, looks good.
Personally I'd remove the `viewsPool` completely as well, I'm not sure we win anything apart from complexity by pooling those objects, I don't expect the count of views changes often / ever.

Maksims:
> Approving, looks good. Personally I'd remove the `viewsPool` completely as well, I'm not sure we win anything apart from complexity by pooling those objects, I don't expect the count of views changes often / ever.

As I recall, when session starts views might be not available straight away. Also, they might get removed due to platform specific behaviours (like minimizing or going into passthrough, and can be other edge cases).

Here is the source of specs: https://www.w3.org/TR/webxr/#xrview-interface

> No guarantee is made about the number of views any XR device uses or their order, nor is the number of views required to be constant for the duration of an XRSession.

PR #5069, opened 2023-02-10
Warn and block Entity transform when Dynamic Rigid Body present

### LINKED ISSUE #2215: Add warning when trying to setPosition on dynamic physics object

### LINKED ISSUE #5042: Entity rendered different from Dynamic RigidBody

MushAsterion:
Fixes #2215 
Fixes #5042 

Warn user when changing Dynamic Rigid Body transform and make sure to sync back from Rigid Body even if disabled from simulation.


willeastcott (approved):
This is great, @MushAsterion. I have to say, I much prefer this approach over the previous PR. So thanks for your feedback @LeXXik. I do still feel a bit uneasy about accessing a private, internal GraphNode property from the rigidbody component. But I can live with it for now. Maybe we can refactor that a bit later. But for now, let's merge! 🚀

PR #5071, opened 2023-02-11
BitPacking functionality

mvaligursky:
A BitPacking namespace with functionality useful for packing bits into a number.

I've done some performance testing as well. The winner were free functions, followed by functions in an object (name space) - this was 17% slower, but still great. This followed by a class with static functions, which was massively slower, and so I've decided to not use that.

Screenshot 2023-02-10 at 11 20 09

### DIFF
--- a/src/core/math/bit-packing.js
+++ b/src/core/math/bit-packing.js
@@ -1,5 +1,5 @@
 /**
-* BitPacking API - functionality for operating on values stored as bits in a number.
+ * BitPacking API - functionality for operating on values stored as bits in a number.
  *
  * @namespace
  * @ignore
--- a/src/core/math/bit-packing.js
+++ b/src/core/math/bit-packing.js
@@ -54,7 +54,7 @@ const BitPacking = {
      * @param {number} storage - Number to test.
      * @param {number} shift - Number of bits to shift the mask.
      * @param {number} [mask] - Mask to limit the number of storage bits. Defaults to 1.
-     * @returns {number} Retuns true if any bits in the mask are set in the storage.
+     * @returns {boolean} Returns true if any bits in the mask are set in the storage.
      */
     any: function (storage, shift, mask = 1) {
         return (storage & (mask << shift)) !== 0;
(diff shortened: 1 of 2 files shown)


src/core/math/bit-packing.js
```diff
@@ -0,0 +1,64 @@
+/**
+ * BitPacking API - functionality allowing to operate on values stored as bits in a number.
```
willeastcott, line 2:
```suggestion
* BitPacking API - functionality for operating on values stored as bits in a number.
```

src/core/math/bit-packing.js
```diff
@@ -0,0 +1,64 @@
...
+        return (storage & shifted) === shifted;
+    },
+
+    /**
+     * Tests if any specified bits are set.
+     *
+     * @param {number} storage - Number to test.
+     * @param {number} shift - Number of bits to shift the mask.
+     * @param {number} [mask] - Mask to limit the number of storage bits. Defaults to 1.
+     * @returns {number} Retuns true if any bits in the mask are set in the storage.
```
willeastcott, line 57:
```suggestion
 * @returns {boolean} Returns true if any bits in the mask are set in the storage.
```

src/core/math/bit-packing.js
```diff
@@ -0,0 +1,64 @@
...
+        return (storage >> shift) & mask;
+    },
+
+    /**
+     * Tests if all specified bits are set.
+     *
+     * @param {number} storage - Number to test.
+     * @param {number} shift - Number of bits to shift the mask.
+     * @param {number} [mask] - Mask to limit the number of storage bits. Defaults to 1.
+     * @returns {number} Retuns true if all bits in the mask are set in the storage.
```
willeastcott, line 44:
```suggestion
     * @returns {boolean} Returns true if all bits in the mask are set in the storage.
```

willeastcott (approved):
Approving with some minor suggestions. I notice this is being introduced without actually being used anywhere. What are the planned use cases? I assume all 4 functions are required for what you have planned?

mvaligursky:
Yes, it's part of how I store render states - this will be a separate PR, just need to finish up. But this functionality looked a good one to be put into a separate PR.

willeastcott:
I'm all for separate PRs. Nice job on the performance investigation too. We should always keep perf in mind.

PR #973, opened 2017-07-20
Changed pc.createCylinder opts.baseRadius to opts.radius

CarlBateman:
Fixed spaces

PR #5073, opened 2023-02-12
More nullish coalescing refactoring

willeastcott:
Make further use of nullish coalescing (`??`) in the codebase. The `Texture` constructor in particular has been heavily simplified (by around 50%).

Mesh shape creation functions have also been simplified.

Minor breaking change:
* `baseRadius` option is now removed from `createCylinder` function. This was [deprecated in July 2017](https://github.com/playcanvas/engine/pull/973). This is highly unlikely to affect anybody.

### DIFF
--- a/src/framework/components/collision/system.js
+++ b/src/framework/components/collision/system.js
@@ -233,9 +233,9 @@ class CollisionSphereSystemImpl extends CollisionSystemImpl {
 // Capsule Collision System
 class CollisionCapsuleSystemImpl extends CollisionSystemImpl {
     createPhysicalShape(entity, data) {
-        const axis = (data.axis !== undefined) ? data.axis : 1;
-        const radius = data.radius || 0.5;
-        const height = Math.max((data.height || 2) - 2 * radius, 0);
+        const axis = data.axis ?? 1;
+        const radius = data.radius ?? 0.5;
+        const height = Math.max((data.height ?? 2) - 2 * radius, 0);
 
         let shape = null;
 
@@ -260,9 +260,9 @@ class CollisionCapsuleSystemImpl extends CollisionSystemImpl {
 // Cylinder Collision System
 class CollisionCylinderSystemImpl extends CollisionSystemImpl {
     createPhysicalShape(entity, data) {
-        const axis = (data.axis !== undefined) ? data.axis : 1;
-        const radius = (data.radius !== undefined) ? data.radius : 0.5;
-        const height = (data.height !== undefined) ? data.height : 1;
+        const axis = data.axis ?? 1;
+        const radius = data.radius ?? 0.5;
+        const height = data.height ?? 1;
 
         let halfExtents = null;
         let shape = null;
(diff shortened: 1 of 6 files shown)


kungfooman:
Just for brain storming, ES6 also allows this kind of style:

```js
function RenderTarget({
    depth = true,
    stencil = false
} = {}) {
    this._depth = depth;
    this._stencil = stencil;
    console.log({depth, stencil})
}
new RenderTarget();
new RenderTarget({depth: false})
```

image

in contrast to:

```js
function RenderTarget(options = {}) {
    this._depth = options.depth ?? true;
    this._stencil = options.stencil ?? false;
    console.log({depth: this._depth, stencil: this._stencil})
}
new RenderTarget();
new RenderTarget({depth: false})
```

image

It moves the "default values" a bit closer together, but on the other hand it kinda introduces ascii soup like `{a=1} = {}`.

src/framework/components/collision/system.js
```diff
@@ -233,9 +233,9 @@ class CollisionSphereSystemImpl extends CollisionSystemImpl {
 // Capsule Collision System
 class CollisionCapsuleSystemImpl extends CollisionSystemImpl {
     createPhysicalShape(entity, data) {
-        const axis = (data.axis !== undefined) ? data.axis : 1;
-        const radius = data.radius || 0.5;
-        const height = Math.max((data.height || 2) - 2 * radius, 0);
+        const axis = data.axis ?? 1;
+        const radius = data.radius ?? 0.5;
```
kungfooman, line 237: Doesn't this change functionality? I guess a zero radius doesn't make sense and previously it would turn it into 0.5:

image

But then again, who would request a zero radius in the first place :thinking:

src/platform/graphics/render-target.js
```diff
@@ -135,7 +130,7 @@ class RenderTarget {
         }
 
         // render image flipped in Y
-        this.flipY = !!options.flipY;
+        this.flipY = options.flipY ?? false;
```
slimbuck, line 133: not sure this one is better, I'd probably keep the old version which forces boolean.

src/framework/components/collision/system.js
willeastcott: I would argue that this was a bug before. While it's weird to request a 0 radius, I think the original intention was to catch `undefined`.

willeastcott:
@kungfooman Interesting. I'll experiment with that style and see how it goes...

src/platform/graphics/render-target.js
willeastcott: I was really going for consistency here since this was out of step with, say, the `depth` option. which doesn't coerce non-booleans into a boolean.

willeastcott: For TypeScript developers, we can guarantee they're passing a boolean. So the code as is is totally correct in that case. However, for JS developers, you're basically advocating doing some level of work to sanitize function parameters. This isn't something the engine does much. I can find some instances. For example, check out this `Slot` code:

https://github.com/playcanvas/engine/blob/bd6256e83eadb065d4a5810959555200a9db4c9a/src/framework/components/sound/slot.js#L83-L89

Values are being explicitly converted to a `number` type where needed. What I would want is a set of explicit rules for the engine to follow to process passed in parameter values. And once we have those rules, we can apply them _everywhere_ and document them. Right now, the codebase isn't consistent.

kungfooman: You can't really sanitize input, for example the "explicit" pitch code conversion you mention works for `NaN` but then already fails at `Infinity`, which happens rather often (division by zero somewhere deep down in the rabbit hole):

image

Also this is "silent failing", so the code may work, but no one really knows why it happens to work for `NaN` but not for `Infinity` (until you debug for a certain amount of time). TypeScript doesn't really help either, it will happily pass `NaN`/`Infinity` around as numbers.

What actually works is generating greedy and pedantic type checks for literally every function that has JSDoc. But this is a performance concern while adding lots of extra code. So people making games in the PlayCanvas editor can always test everything with the type checking debug builds, but then e.g. only running the code in unchecking/fast release builds for the players.

PR #5077, opened 2023-02-20
Fix the destroy callback in Bloom

querielo:
_destroy is defined in this.effect.

### DIFF
--- a/scripts/posteffects/posteffect-bloom.js
+++ b/scripts/posteffects/posteffect-bloom.js
@@ -296,5 +296,5 @@ Bloom.prototype.initialize = function () {
     this.on('destroy', function () {
         queue.removeEffect(this.effect);
         this.effect._destroy();
-    }, this);
+    });
 };
--- a/scripts/posteffects/posteffect-bloom.js
+++ b/scripts/posteffects/posteffect-bloom.js
@@ -295,6 +295,6 @@ Bloom.prototype.initialize = function () {
 
     this.on('destroy', function () {
         queue.removeEffect(this.effect);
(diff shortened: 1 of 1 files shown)


scripts/posteffects/posteffect-bloom.js
```diff
@@ -295,6 +295,6 @@ Bloom.prototype.initialize = function () {
 
     this.on('destroy', function () {
         queue.removeEffect(this.effect);
-        this._destroy();
-    });
+        this.effect._destroy();
+    }, this);
```
willeastcott, line 299: I didn't think the final `this` parameter to `on` is needed here. Can you confirm?

willeastcott:
Thanks for the fix. Is now a good time to do an audit to check if *all* post effects are correctly freeing any allocated resources?

scripts/posteffects/posteffect-bloom.js
querielo: You are right. I just added it because had seen it in other places. Example: https://github.com/playcanvas/engine/blob/3fead3ac2a2a7bf3d09b292a07abed2b4a798cb6/scripts/posteffects/posteffect-bloom.js#L286

querielo:
> Thanks for the fix. Is now a good time to do an audit to check if _all_ post effects are correctly freeing any allocated resources?

It looks like everything is fine with other post effects

willeastcott (approved):
Approving, but I'll leave it to @mvaligursky to merge. 😄 Thanks!

mvaligursky (approved):

PR #5078, opened 2023-02-20
Parse models asynchronously

slimbuck:
Due to the nature of asset loading system the model and animation assets always used a synchronous method for loading assets.

This has recently become an issue with draco loading. The draco decompression step is asynchronous and so model loading happens via a different execution path (it succeeds, but results are not returned correctly).

This PR changes `GlbParser.parse` to be asynchronous and moves model parsing from `Handler.open` (which is synchronous) to `Handler.load` (which is asynchronous).

**Testing Notes:**
This is a big change to the asset loading flow for glb models, json models and animations. With that in mind:
- [x] Test in the editor that all types of models work correctly:
    - [x] json
    - [x] glb
    - [x] animations
    - [x] glb-hierarchy


src/framework/handlers/animation.js
```diff
@@ -30,7 +30,7 @@ class AnimationHandler {
         this.maxRetries = 0;
     }
 
-    load(url, callback) {
+    load(url, callback, asset) {
```
kungfooman, line 33: To quote @yaustar from another PR:

> and traditionally, the callback is the last param

Maksims:
async/await is great, although it has some execution flow differences, in in many cases it should be avoided with callback/promise used instead. A quick pseudo example:

For each asset:
    Load asset
Start update loop

kungfooman:
> async/await is great, although it has some execution flow differences, in many cases it should be avoided with callback/promise used instead.

Thank you for sharing your experience! Async/await is an incredibly powerful tool that can make asynchronous programming easier to understand, maintain, and debug. It is not necessarily the case that callback/promise should be used instead of async/await. In fact, async/await can often be used in conjunction with callback/promise to provide a cleaner, more manageable to handle asynchronous asset loading.

If you don't want to wait for the promises you can use `.then()` instead of `await`. This is e.g. how I load some textures:

```js
function material_freepbr_grass1() {
    const material     = new pc.StandardMaterial();
    material.name      = "freepbr_grass1"
    material.bumpiness = 3;
    material.update();
    const { app } = appSimplified;
    fetch_pc_texture(app, Globals.url + "pbr/" + material.name + "/diffuse.png").then((texture) => { material.diffuseMap = texture; material.update(); });
    fetch_pc_texture(app, Globals.url + "pbr/" + material.name + "/normal.png" ).then((texture) => { material.normalMap  = texture; material.update(); });
    fetch_pc_texture(app, Globals.url + "pbr/" + material.name + "/height.png" ).then((texture) => { material.heightMap  = texture; material.update(); });
    return material;
  }
```

(the material is instantiated synchrounously and updated whenever new textures arrive)

I have to maintain `fetch_pc_texture` and other promisified asset loading functions for myself, but I would like a general `Promise`-based `pc` API.

My main concern would be mostly backwards compatibility.

Backwards compatibility for private methods can be disregarded, so we can always opt-in for some specific/private/unexposed functions that look like callback hell:

```js
// parse the gltf or glb data synchronously. external resources (buffers and images) are ignored.
    static parse(filename, data, device, options, callback) {
        options = options || { };

        // parse the data
        parseChunk(filename, data, function (err, chunks) {
            if (err) {
                callback(err);
            } else {
                // parse gltf
                parseGltf(chunks.gltfChunk, function (err, gltf) {
                    if (err) {
                        callback(err);
                    } else {
                        // parse buffer views
                        parseBufferViewsAsync(gltf, [chunks.binaryChunk], options, function (err, bufferViews) {
                            if (err) {
                                callback(err);
                            } else {
                                // create resources
                                createResources(device, gltf, bufferViews, [], options, function (err, result) {
                                    if (err) {
                                        callback(err);
                                    } else {
                                        callback(null, result);
                                    }
                                });
                            }
                        });
                    }
                });
            }
        });
    }
```

```js
// Simplified Version / treat as pseudo-code / untested
async static parse(filename, data, device, options, callback) {
  options = options || {};
  try {
    // parse the data
    const chunks = await parseChunk(filename, data);
    // parse gltf
    const gltf = await parseGltf(chunks.gltfChunk);
    // parse buffer views
    const bufferViews = await parseBufferViewsAsync(gltf, [chunks.binaryChunk], options);
    // create resources
    const result = await createResources(device, gltf, bufferViews, [], options);
    callback(null, result)
  } catch (err) {
    callback(err);
  }
}
```

Just look at the line difference. You can basically keep your callbacks aswell. I agree that Promises can be quite hard to learn/understand aswell, everything takes time... do you think sooner or later Promise's should be introduced? Or are you 100% against it, or any other ideas?

slimbuck:
Hi @kungfooman, @Maksims,

Yep we're actually in the process of rewriting glb-parser to make heavy use of promises in order to speed up loading times.

This PR simply prepares the model asset handler to continue working once we've made that change.

(We're also moving draco decompression to worker threads, it's great).

Thanks!

kungfooman:
Thank you for the information @slimbuck, I look forward to see the contrast of lines/complexity in the PR :+1:

src/framework/handlers/animation.js
```diff
@@ -52,31 +52,35 @@ class AnimationHandler {
...
             } else {
-                callback(null, response);
+                // parse the result immediately (this used to happen during open)
+                if (path.getExtension(url.original).toLowerCase() === '.glb') {
+                    GlbParser.parse('filename.glb', response, null, null, (err, parseResult) => {
+                        if (err) {
+                            callback(err);
+                        } else {
+                            const animations = parseResult.animations;
+                            if (asset?.data?.events) {
```
mvaligursky, line 66: unrelated to this PR, but this feels strange @ellthompson - we have a single asset.data.events, but create multiple instances of AnimEvents and attach it to all animations. Is this correct?

mvaligursky (approved):
Approved with a minor comment.

src/framework/handlers/model.js
```diff
@@ -75,28 +74,36 @@ class ModelHandler {
...
 
     open(url, data) {
-        for (let i = 0; i < this._parsers.length; i++) {
-            const p = this._parsers[i];
-
-            if (p.decider(url, data)) {
-                return p.parser.parse(data);
-            }
-        }
-        Debug.warn('pc.ModelHandler#open: No model parser found for: ' + url);
```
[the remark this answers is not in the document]
slimbuck: asset failure falls through to loader's failure handler which prints this out https://github.com/playcanvas/engine/blob/main/src/framework/handlers/loader.js#L198

PR #5080, opened 2023-02-20
Instantiate DracoDecoderModule if it is in extensionsUsed

querielo:
According to https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#specifying-extensions
"All values in extensionsRequired MUST also exist in extensionsUsed." So, I think we have to use Draco even if it is not required but used.


slimbuck (approved):
Yep makes sense. Thanks!

mvaligursky (approved):

Issue #2254, opened 2020-07-07
Optimize setup of WebGl state variables


mvaligursky:
device.js contains multiple functions of this format:

![Screen Shot 2020-07-07 at 4 42 56 PM](https://user-images.githubusercontent.com/59932779/86806606-edfef280-c070-11ea-9274-003149c36c44.png)

there are multiple conditions followed by settings state if any of these fail.

I would suggest to follow newer graphics APIs and create objects wrapping groups of these various settings. As an example, we can create RasterizerState object (pseudo code with random properties - needs design)

```
class RasterizerState {
    blendEquation;
    blendAlphaEquation;
    separateAlphaEquation;
    cullMode;
    depthTest;
    depthWrite;
} 
```

when these settings are set up on material, inside material.Update() we will look up dictionary of these RasterizerObjects, and assign a single of them to Material.

While rendering, we'd only need one test to see if RasterizerState changes, instead of many tests.
We could also store transition data for when device is transitioning between these RasterizerState objects, allowing us to pre-evaluate what needs to change on gl device (for example going from RasterizerState A to B, only cullMode needs to change)

so the equivalent function would look like this:
```
setRasteriserState: function (state) {
    if (this.rastState !== state) {   // compare by pointer
        this.rastState = state;

        // apply changes on GL
       gl.setAlphaState(...);
       gl.setDepthState(...);
    }
}
```

Ideally, these state objects would correspond to how WebGPU manages these, to allow for easier transition to that API.

mvaligursky:
I'll close this as both the DepthState and BlendState are done, and we have no need for additional states for now.

PR #5083, opened 2023-02-21
[BREAKING] Implementation of BlendState

mvaligursky:
Related to https://github.com/playcanvas/engine/issues/2254

## Overview

A new BlendState class is implemented, which encapsulates a blending state of rendering pipeline. It handles alpha blending and enabling of per-channel writes. Internally, a bit packing is used to store blend state in a single number (per render target, to allow multiple render targets in the future). This makes blend state setup more costly, but per frame rendering is faster as just a single number needs to be compared to see if blend state needs to be changed. Additionally, it makes it fast to generate a key which is used for render pipeline look up on WebGPU.

### Breaking changes
- although all public functionality has been deprecated and preserved, if some private blending related functionality has been used, it's possible the code needs to be updated to use public functionality.

- To align WebGL and WebGPU:
```
- BLENDMODE_CONSTANT_COLOR and BLENDMODE_CONSTANT_ALPHA has been merged
   to a single BLENDMODE_CONSTANT constant
- BLENDMODE_ONE_MINUS_CONSTANT_COLOR and BLENDMODE_ONE_MINUS_CONSTANT_ALPHA
   has been merged to a single BLENDMODE_ONE_MINUS_CONSTANT constant
- Note that the _COLOR constant is internally used for the color blending, and _ALPHA constant is used
   for alpha blending, so no functionality has been lost, only simplified. This is not exposed on the
   material, and was only useable using private APIs on the material, so the impact is expected to be small.
```
- drawQuadWithShader function signature has changed, and last 'useBlend' parameter has been removed. If value is provided, a warning message is printed out. This parameter was confusing, as it was expected to set all other blending states directly on device apart from this one, which was internally set by the drawQuadWithShader. Now all render state is expected to be set directly on device. Code like this:
```
if (blending) {
    device.setBlendFunction(BLENDMODE_ONE, BLENDMODE_ONE);
    device.setBlendEquation(BLENDEQUATION_ADD);
    drawQuadWithShader(..., blending);
}
```
is now
```
    device.setBlendState(<created state with blend mode fully set up>);
    drawQuadWithShader(...);
}
```

- BLEND_SUBTRACTIVE mode is now supported. The implementation was missing before, so this was behaving like BLEND_NORMAL, but now will function as expected.

### New public API (still breaking change, as old APIs are deprecated)

- BlendState.constructor(blend = false, colorOp = BLENDEQUATION_ADD, colorSrcFactor = BLENDMODE_ONE, colorDstFactor = BLENDMODE_ZERO,
        alphaOp, alphaSrcFactor, alphaDstFactor,
        redWrite = true, greenWrite = true, blueWrite = true, alphaWrite = true)
- BlendState.blend - toggle blending
- BlendState.equals - compares two blend states
- BlendState.DEFAULT - a static access to a blend state with no blending and writing to all channels

- GraphicsDevice.setBlendState(blendState)
	this replaces these properties:
```
blending
blendSrc
blendDst
blendSrcAlpha
blendDstAlpha
separateAlphaBlend
blendEquation
blendAlphaEquation
separateAlphaEquation
writeRed
writeGreen
writeBlue
writeAlpha
```
- Material.blendState property
	this replaces these properties:
```
blend
blendSrc
blendDst
blendEquation
separateAlphaBlend
blendSrcAlpha
blendDstAlpha
blendAlphaEquation
redWrite
greenWrite
blueWrite
alphaWrite
```

### Other changes
- internally, engine has been updated to use blend state for all blending operations, and in some cases, a blend state is set that wasn't before, to enforce correct state to be set up, to avoid leftover state issue
- on WebGl platform, the implementation has been changed to always use separate blending for color & alpha channels, to match WebGPU - this is the same number of API calls anyways, so there is no advantage to use both modes.

### Testing
- tested engine examples against WebGl1 & 2
- tested Editor

### JSDocs for the new class
Screenshot 2023-02-21 at 12 21 04
Screenshot 2023-02-21 at 12 21 16


src/platform/graphics/blend-state.js
```diff
@@ -0,0 +1,222 @@
...
+        this.blend = blend;
+    }
+
+    /**
+     * Enables or disables blending.
+     *
+     * @type {boolean}
+     */
+    set blend(value) {
+        this.target0 = BitPacking.set(this.target0, value ? 1 : 0, blendShift);
```
kungfooman, line 107: Isn't this doing more than necessary?

1) The bit operations involve the property lookup in the `BitPacking` object, which could be just exported as normal functions, e.g. named `bitPackingSet`

2) I am not sure about the exact working of `BitPacking.set`, but isn't this line the same as:

```js
this.target0 |= (value ? 1 : 0) << blendShift;
```
?

I did some testing and it seems similiar:

```js
var diffResults = 0;
for (var start = -100000; start < 100000; start++) {
    for (var i=-100; i<100; i++) {
        const a = BitPacking.set(start, 1, i);
        const b = start | 1 << i;
        if (a != b) {
            diffResults++;
        }
    }
}
console.log("diffResults", diffResults); // 0
```

mvaligursky: 1. I did some performance testing, and having those functions inside an object was acceptable: https://github.com/playcanvas/engine/pull/5071
2. it's not the same. BitPacking.set allows values to be set to 0 or 1, but your implementation with `\=' only allows it to be set to 1, but never cleared to 0.

kungfooman: Thank you, I was blind to the 0 case :eyes: 

```js
var diffResults = 0;
var sameResults = 0;
for (var start = -100000; start < 100000; start++) {
    for (var i=-100; i<100; i++) {
        const a = BitPacking.set(start, 0, i);
        const b = start | 0 << i;
        if (a != b) {
            diffResults++;
        } else {
            sameResults++;
        }
    }
}
console.log("diffResults", diffResults, "sameResults", sameResults);
// diffResults 20000000 sameResults 20000000
```

src/platform/graphics/blend-state.js
```diff
@@ -0,0 +1,222 @@
+import { BitPacking } from "../../core/math/bit-packing.js";
+import { BLENDEQUATION_ADD, BLENDMODE_ONE, BLENDMODE_ZERO } from '../../platform/graphics/constants.js';
+
+// masks (to only keep relevant bits)
+const opMask = 0b111;
+const factorMask = 0b1111;
+
+// shits values to where individual parts are stored
```
slimbuck, line 8:
```suggestion
// shifts values to where individual parts are stored
```

src/platform/graphics/blend-state.js
```diff
@@ -0,0 +1,222 @@
...
+     * @param {boolean} redWrite - True to enable writing of the red channel and false otherwise.
+     * Defaults to true.
+     * @param {boolean} greenWrite - True to enable writing of the green channel and false
+     * otherwise. Defaults to true.
+     * @param {boolean} blueWrite - True to enable writing of the blue channel and false otherwise.
+     * Defaults to true.
+     * @param {boolean} alphaWrite - True to enable writing of the alpha channel and false
+     * otherwise. Defaults to true.
+     */
+    constructor(blend = false, colorOp = BLENDEQUATION_ADD, colorSrcFactor = BLENDMODE_ONE, colorDstFactor = BLENDMODE_ZERO,
```
slimbuck, line 92: I think default blend mode should be SRC_ALPHA, ONE_MINUS_SRC_ALPHA even though blend = false.

slimbuck: And probably alpha should default to ADD, ONE, ONE_MINUS_SRC_ALPHA.

src/scene/materials/material.js
```diff
@@ -239,12 +267,41 @@ class Material {
...
+     * render target. This overwrites blending type set using {@link pc.Material#blendType}, and
+     * offers more control over blending.
+     *
+     * @type { BlendState }
+     */
+    set blendState(value) {
+        if (this._blendState.blend !== value.blend) {
+            this._markBlendDirty();
+        }
+        this._blendState.copy(value);
```
slimbuck, line 294: might as well move this into the if block above?

src/deprecated/deprecated.js
```diff
@@ -572,6 +579,63 @@ GraphicsDevice.prototype.removeShaderFromCache = function (shader) {
     getProgramLibrary(this).removeFromCache(shader);
 };
 
+const _tempBlendState = new BlendState();
+
+GraphicsDevice.prototype.setBlendFunction = function (blendSrc, blendDst) {
```
slimbuck, line 584: slightly concerned about performance impact on existing code setting a bunch of states, but not sure there's much we can do about it.

src/platform/graphics/blend-state.js
mvaligursky: haha thanks!

src/platform/graphics/blend-state.js
mvaligursky: I preserved existing blend mode, so that's what we had, and I'm not sure there's a reason to change it, as when a blend mode is set, those all are changed anyways.

src/deprecated/deprecated.js
mvaligursky: yeah internally in the engine I removed all those, and print deprecated message.
The cost would not be that large though I think.

src/platform/graphics/blend-state.js
```diff
@@ -0,0 +1,222 @@
...
+    get alphaSrcFactor() {
+        return BitPacking.get(this.target0, alphaSrcFactorShift, factorMask);
+    }
+
+    get alphaDstFactor() {
+        return BitPacking.get(this.target0, alphaDstFactorShift, factorMask);
+    }
+
+    set redWrite(value) {
+        this.target0 = BitPacking.set(this.target0, value ? 1 : 0, redWriteShift);
```
GSterbrant, line 158: Either what I said above, or these should also set the redWrite, greenWrite, etc.

src/platform/graphics/blend-state.js
```diff
@@ -0,0 +1,222 @@
...
+        this.target0 = BitPacking.set(this.target0, dstFactor, colorDstFactorShift, factorMask);
+    }
+
+    setAlphaBlend(op, srcFactor, dstFactor) {
+        this.target0 = BitPacking.set(this.target0, op, alphaOpShift, opMask);
+        this.target0 = BitPacking.set(this.target0, srcFactor, alphaSrcFactorShift, factorMask);
+        this.target0 = BitPacking.set(this.target0, dstFactor, alphaDstFactorShift, factorMask);
+    }
+
+    setColorWrite(redWrite, greenWrite, blueWrite, alphaWrite) {
```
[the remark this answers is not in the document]
mvaligursky: It does? it calls those functions, that do it.

src/platform/graphics/blend-state.js
mvaligursky: I'm not sure I follow. All data is stored in target0 at the moment, including redWrite and all that. redWrite is just a setter that does it.

src/scene/materials/material.js
mvaligursky: I don't think I can, as that would only copy it blendState.blend is different, but it needs to be copied even if that is equal and other members of it are not equal.

slimbuck (approved):

src/platform/graphics/blend-state.js
```diff
@@ -0,0 +1,222 @@
...
+     * @param {boolean} greenWrite - True to enable writing of the green channel and false
+     * otherwise. Defaults to true.
+     * @param {boolean} blueWrite - True to enable writing of the blue channel and false otherwise.
+     * Defaults to true.
+     * @param {boolean} alphaWrite - True to enable writing of the alpha channel and false
+     * otherwise. Defaults to true.
+     */
+    constructor(blend = false, colorOp = BLENDEQUATION_ADD, colorSrcFactor = BLENDMODE_ONE, colorDstFactor = BLENDMODE_ZERO,
+        alphaOp, alphaSrcFactor, alphaDstFactor,
+        redWrite = true, greenWrite = true, blueWrite = true, alphaWrite = true) {
```
willeastcott, line 94: Should this take an options object instead? Like a `Texture` for example?

src/platform/graphics/blend-state.js
```diff
@@ -0,0 +1,222 @@
...
+    get alphaWrite() {
+        return BitPacking.all(this.target0, alphaWriteShift);
+    }
+
+    get allWrite() {
+        // return a number with all 4 bits, for fast compare
+        return BitPacking.get(this.target0, allWriteShift, allWriteMasks);
+    }
+
+    copy(rhs) {
```
willeastcott, line 194: Would a `clone` be useful too? We seem to be developing a `copy` + `clone` interface across many API classes.

src/platform/graphics/blend-state.js
mvaligursky: Maybe, just not sure it's worth creating a temporary object. In majority of cases, only the first 4 parameters are used, the rest if a lot more rare.

src/platform/graphics/blend-state.js
mvaligursky: It might be useful, so this has been added now.

Issue #5049, opened 2023-02-07
Picker never releases its render target


willeastcott:
The `Picker` constructor creates a class member called `_renderTarget` here:

https://github.com/playcanvas/engine/blob/374b55fd8d4641106beb88bb5392eabdc84e7bb3/src/framework/graphics/picker.js#L66

This is freed in `releaseRenderTarget` here:

https://github.com/playcanvas/engine/blob/374b55fd8d4641106beb88bb5392eabdc84e7bb3/src/framework/graphics/picker.js#L182-L185

However, `allocateRenderTarget` sets the allocated render target on a class member called `renderTarget` (notice the missing underscore):

https://github.com/playcanvas/engine/blob/374b55fd8d4641106beb88bb5392eabdc84e7bb3/src/framework/graphics/picker.js#L171-L174

So it is not currently possible to release the picker's render target.

PR #5084, opened 2023-02-21
[Fix] Fix to Picker class not releasing its render target

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/5049


slimbuck (commented):
Looks like you've just renamed rendertarget... What am I missing?

mvaligursky:
You missed nothing. I renamed it to match the name we use in the rest of the file.

slimbuck (approved):

Issue #3187, opened 2021-05-22
Application.start when called twice will re-initialise systems


marklundin:
`application.start` will call initialize on the component system and start a new animation loop even if it has already been called before. This causes multiple render loops and it's not clear this is the intended behaviour. Ideally the 2nd call to `start()` is a noop and the Application maintains reference to whether this has already been called

yaustar:
Just to confirm, this is only in the case when calling application.start twice?

marklundin:
Yep. Granted, it's trivial for a developer to track whether an Application has already started, but it seems this should also be tracked in the Application itself?

mvaligursky:
To me calling the API correctly seems a responsibility of a developer - engine should not need to handle a case user initialises the engine multiple times.

marklundin:
I've run into issues accidentally calling start multiple times but it wasn't clear this was the case. I think if it is the devs responsibility, there should be more feedback that calling it multiple times can cause side effects.

PR #5097, opened 2023-02-23
Debug assert when AppBase.start is called more than once.

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/3187


slimbuck (approved):
I do wonder whether this test should still return in release build, but it's debatable so approving.

PR #5099, opened 2023-02-23
Added convenience functions in SoundComponent

### LINKED ISSUE #4885: Prevent a sound from starting repeatedly
Currently the engine does not prevent that a sound is started repeatedly. This ensures that the sound cannot be played correctly, if the user starts the sound from for example the update function. The user could make a statement for each sound to prevent this, but this is not user friendly. Also, the API to do this is unnecessarily complicated and not clearly explained in the manual.

Expectation:
```
if (!this.entity.sound.isPlaying('shoot')) {
    this.entity.sound.play('shoot');
}
```
Actually:
```
if (!this.entity.sound.slot('shoot').isPlaying) {
    this.entity.sound.play('shoot');
}
```

MushAsterion:
Added convenience functions in `SoundComponent` to make API easier for users, adding the following methods:
- `isLoaded(name)`
- `isPlaying(name)`
- `isPaused(name)`
- `isStopped(name)`

Fixes #4885 

### DIFF
--- a/src/framework/components/sound/component.js
+++ b/src/framework/components/sound/component.js
@@ -415,23 +415,35 @@ class SoundComponent extends Component {
     }
 
     /**
+     * Return a property from the slot with the specified name.
+     *
+     * @param {string} name - The name of the {@link SoundSlot} to look for.
+     * @param {string} property - The name of the property to look for.
+     * @returns {*} The value from the looked property inside the slot with specified name. May be undefined if slot does not exist.
+     * @ignore
+     */
+    _getSlotProperty(name, property) {
+        if (!this.enabled || !this.entity.enabled) {
+            return undefined;
+        }
+
+        const slot = this._slots[name];
+        if (!slot) {
+            Debug.warn(`Trying to get ${property} from sound slot with name ${name} which does not exist`);
+            return undefined;
+        }
+
+        return slot[property];
+    }
+
+    /**
      * Returns true if the slot with the specified name is currently playing.
      *
      * @param {string} name - The name of the {@link SoundSlot} to look for.
      * @returns {boolean} True if the slot with the specified name exists and is currently playing.
      */
     isPlaying(name) {
-        if (!this.enabled || !this.entity.enabled) {
-            return false;
-        }
-
-        const slot = this._slots[name];
-        if (!slot) {
-            Debug.warn(`Trying to get information from sound slot with name ${name} which does not exist`);
-            return false;
-        }
-
-        return slot.isPlaying;
+        return this._getSlotProperty(name, 'isPlaying') || false;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/framework/components/sound/component.js
```diff
@@ -414,6 +414,86 @@ class SoundComponent extends Component {
...
+            return false;
+        }
+
+        const slot = this._slots[name];
+        if (!slot) {
+            Debug.warn(`Trying to get information from sound slot with name ${name} which does not exist`);
+            return false;
+        }
+
+        return slot.isPlaying;
```
willeastcott, line 434: Quite a lot of code duplication here. Could you have something like:

```
    _getSlotProperty(slotName, property) {
        if (!this.enabled || !this.entity.enabled) {
            return false;
        }

        const slot = this._slots[slotName];
        if (!slot) {
            Debug.warn(`Trying to get information from sound slot with name ${name} which does not exist`);
            return false;
        }

        return slot[property];
    }
```

MushAsterion: It's done!

src/framework/components/sound/component.js
```diff
@@ -414,6 +414,68 @@ class SoundComponent extends Component {
         return this._slots[name];
     }
 
+    /**
+     * Return a property from the slot with the specified name.
+     *
+     * @param {string} name - The name of the {@link SoundSlot} to look for.
+     * @param {string} property - The name of the property to look for.
+     * @returns {*} The value from the looked property inside the slot with specified name. May be undefined if slot does not exist.
+     * @ignore
```
willeastcott, line 423:
```suggestion
     * @private
```

willeastcott (approved):
Thanks for the PR! ⭐

src/framework/components/sound/component.js
```diff
@@ -414,6 +414,68 @@ class SoundComponent extends Component {
         return this._slots[name];
     }
 
+    /**
+     * Return a property from the slot with the specified name.
+     *
+     * @param {string} name - The name of the {@link SoundSlot} to look for.
+     * @param {string} property - The name of the property to look for.
+     * @returns {*} The value from the looked property inside the slot with specified name. May be undefined if slot does not exist.
```
mvaligursky, line 422: the type could be `boolean | undefined` instaed of '*'?

mvaligursky: or even better, the function should return false in that case .. never undefined? That would simplify other functions that call this function.

mvaligursky (approved):
approved with a small suggestion.

src/framework/components/sound/component.js
mvaligursky: ok fair enough, lets keep it generic then.

PR #5103, opened 2023-02-27
Convert local light shadow rendering to use render passes

mvaligursky:
Render passes are used to render local light (spot & omni) shadows:
- for clustered lights, as those are all in a single shadow atlas, all shadows are rendered inside a single render pass
- for non-clustered lights, as those use shadow map each, each shadow map generates a render pass

Note that lightmapper does not use shadow passes, and so renders shadows using original solution not using render passes. Tested both in clustered and non-clustered mode.

This will allow the local shadows to be supported by WebGPU (separate PR).


slimbuck (approved):
so lightmapper still needs updating to work in future?

mvaligursky:
yes, lightmapper needs work for webgpu, it's one of the things I have not touched at all yet.

Issue #4250, opened 2022-05-16
Split standard material into frontend and backend


slimbuck:
#### Background

Up till now the standard material has been responsible for all parts of the rendering pipeline. Standard material defines an "uber shader" interface for artists to specify the surface properties like diffuse, specular, roughness etc. (We call these properties the material _frontend_). The standard material also implements the runtime lighting system, including lights, IBL, dynamic shadows and everything else needed to render the surface. (We call this the material _backend_).

(As an aside, in deferred rendering the components written into the g-buffer can be thought of as the frontend. While the lighting passes which use the g-buffer and generate the final image are the material backend).

The current standard material structure means that users have limited scope for customising the look of materials. Either users must make do with the features provided by the uber shader interface, or they must customise the shader chunk system. Unfortunately the shader chunk system is undocumented, open to change, difficult to debug and results in very brittle code.

For these reasons, we propose to split out the lighting backend from the monolithic standard material. Doing so will mean we can leverage the runtime lighting system with any arbitrary frontend.

The new lighting backend will be used by the standard material as well as the shader-editor (currently in development).

#### Todo:
- [x] extract the backend parts of [standard.js](https://github.com/playcanvas/engine/blob/main/src/graphics/program-lib/programs/standard.js) (done in #4254)
- [x] restructure shader chunks so there is a formal frontend and backend split (done in #4254)
- [x] provide runtime support to notify application developers if chunks have changed (#4343)
- [x] restructure the options builder into frontend and backend (see [options builder](https://github.com/playcanvas/engine/blob/main/src/scene/materials/standard-material-options-builder.js)) (done in PR #)
- [ ] implement an engine example of custom frontend shader using the new material backend (done in PR #)

slimbuck:
Todo:
- [ ] extract the backend parts of [standard.js](https://github.com/playcanvas/engine/blob/main/src/graphics/program-lib/programs/standard.js)
- [ ] restructure shader chunks so there is a formal frontend and backend split
- [ ] restructure the options builder into frontend and backend (see [options builder](https://github.com/playcanvas/engine/blob/main/src/scene/materials/standard-material-options-builder.js))
- [ ] implement an engine example of custom frontend shader using the new material backend

Maksims:
Will this deprecate shader chunks?

slimbuck:
> Will this deprecate shader chunks and the current approach of providing custom chunks?

No not at all.

[...]

PR #5104, opened 2023-02-27
Further separation of frontend and backend shaders

GSterbrant:
This PR further separates the backend and the frontend by providing a common interface for the backend. 

By using a frontend struct, there is a clear distinction in the shader code between frontend and backend. This allows us to run the lit-shader by just filling out a struct and passing it to evaluateBackend, instead of reading those values from global parameters.

The way code is generated in the lit-shader is also refactored to use the ChunkBuilders, and just like we do in the frontend, we separate declarations, function declarations and the primary function body as separate strings. 

This PR will break almost all backend chunks but leaves the frontend chunks intact. 

Relates to #4250.

The main function of the shader changes from this:

```glsl
void main(void) {
    dReflection = vec4(0);
    ccSpecularLight = vec3(0);
    ccReflection = vec3(0);
    dVertexNormalW = normalize(vNormalW);
    getViewDir();
    getTBN();
    getNormal();
    getAlbedo();
    getMetalness();
    getGlossiness();
    getSpecularity();
    getEmission();
    getClearCoat();
    getClearCoatGlossiness();
    getClearCoatNormal();
    getReflDir();
    ccReflDirW = normalize(-reflect(dViewDirW, ccNormalW));
    getMetalnessModulate();
    addAmbient();
    dDiffuseLight = dDiffuseLight * (1.0 - dSpecularity);
    addReflectionCC();
    ccFresnel = getFresnelCC(dot(dViewDirW, ccNormalW));
    ccReflection.rgb *= ccFresnel;
    addReflection();
    dReflection.rgb *= getFresnel(dot(dViewDirW, dNormalW), dSpecularity);
    dLightDirNormW = light0_direction;
    dAtten = 1.0;
    dAtten *= getLightDiffuse();
    dDiffuseLight += dAtten * light0_color;
    dHalfDirW = normalize(-dLightDirNormW + dViewDirW);
    ccSpecularLight += getLightSpecularCC(dHalfDirW) * dAtten * light0_color * getFresnelCC(dot(dViewDirW, dHalfDirW));
    dSpecularLight += getLightSpecular(dHalfDirW) * dAtten * light0_color * getFresnel(dot(dViewDirW, dHalfDirW), dSpecularity);
    addClusteredLights();
    gl_FragColor.rgb = combineColor();
    gl_FragColor.rgb += dEmission;
    gl_FragColor.rgb = addFog(gl_FragColor.rgb);
    gl_FragColor.rgb = toneMap(gl_FragColor.rgb);
    gl_FragColor.rgb = gammaCorrectOutput(gl_FragColor.rgb);
    gl_FragColor.a = 1.0;
}
```

To this:

```glsl
void main(void) {
    dReflection = vec4(0);
    ccSpecularLight = vec3(0);
    ccReflection = vec3(0);
    dVertexNormalW = normalize(vNormalW);
    getViewDir();
    getTBN();
    Frontend frontend = evaluateFrontend();
    evaluateBackend(frontend);
}
```

Where evaluateFrontend extracts the frontend arguments, and evaluateBackend runs the shader that produces the image using those frontend values.

In essence, this would allow someone to make a custom shader for the entire frontend, while still using the same shading model.


mvaligursky:
Can you please attach an example of generated shader to the description.

mvaligursky:
It feels like the struct should be a fixed code in some new chunk, instead of dynamically generated in the standard.js - otherwise if we have other shader generators that need to use the lit-shader, such as new material type we'll need, or use directly passing just a front end, we'd need to handle it again?

GSterbrant:
@mvaligursky Completely agree, let me do that instead actually.

src/scene/shader-lib/chunks/standard/frag/frontend.js
```diff
@@ -0,0 +1,37 @@
+export default /* glsl */`
+struct Frontend {
+    vec2 dUvOffset;
```
mvaligursky, line 3: do we need to keep 'd' prefix now that these are not global?

mvaligursky: these should be easy to understand names as users will providing values to these at some point in custom shader front ends .. so lets see if we can have a nice consistent clear naming.

GSterbrant: We could dump the d prefix and change the clearcoat and sheen to be prefixed differently, like clearcoat_Specularity and sheen_Glossiness.

mvaligursky: and perhaps no underscores. just `clearcoatSpecularity`.
even ao could maybe expanded to ambentOcclusion, or maybe just occlusion?
and normalW to worldNormal, or normalWorld.

perhaps make a proposal here before changing all the chunks.

src/scene/shader-lib/chunks/standard/frag/frontend.js
```diff
@@ -0,0 +1,37 @@
+export default /* glsl */`
+struct Frontend {
```
mvaligursky, line 2: I also wonder if we should have a better name than Frontend, considering this is going to be a public API.
Unity calls it SurfaceOutput for example.
Maybe SurfaceProperties? Or MaterialProperties?

src/scene/shader-lib/chunks/standard/frag/frontend.js
```diff
@@ -0,0 +1,37 @@
...
+    vec2 uvOffset;
```
mvaligursky, line 3: what is uvOffset - this does not feel self-explanatory. I think we should add comments for all members here.

src/scene/shader-lib/chunks/standard/frag/frontend.js
```diff
@@ -0,0 +1,37 @@
...
+
+    float alpha;
```
mvaligursky, line 5: should this be transparency?

src/scene/shader-lib/chunks/standard/frag/frontend.js
```diff
@@ -0,0 +1,37 @@
...
+
+    float transmission;
+    float thickness;
+
+    vec3 iridescenceFresnel;
+    float iridescence;
+    float iridescenceThickness;
+
+    vec3 specularity;
+    float glossiness;
```
mvaligursky, line 19: in standard material this was renamed to `gloss` .. should we use the same name here?

src/scene/shader-lib/chunks/standard/frag/frontend.js
```diff
@@ -0,0 +1,37 @@
...
+
+    float clearcoatSpecularity;
+    float clearcoatGlossiness;
+    vec3 clearcoatWorldNormal;
+
+    float sheenGlossiness;
+    vec3 sheenSpecularity;
+
+    vec3 lightmap;
+    vec3 lightmapDir;
```
mvaligursky, line 35: I wonder if we should only have `lightmap' here, as lightmapDir is only related to our directional lightmap implementation. Perhaps by exposing ligthmapDir we lock frontend to just this specific basis? Maybe the front end needs to evaluate ligthmap with dir, and give backend just the result?

src/scene/shader-lib/chunks/standard/frag/frontend.js
GSterbrant: Yep, done.

GSterbrant:
The documentation can be found here: https://github.com/playcanvas/developer.playcanvas.com/pull/504

src/scene/shader-lib/chunks/lit/frag/clusteredLight.js
```diff
@@ -543,45 +543,45 @@ void evaluateLight(ClusterLightData light) {
                 
                 // specular
                 #ifdef LIT_SPECULAR_FRESNEL
-                    dSpecularLight += getLightSpecular(halfDir) * dAtten * light.color * dAtten3 * getFresnel(dot(dViewDirW, halfDir), dSpecularity);
+                    dSpecularLight += getLightSpecular(halfDir, dReflDirW, worldNormal, gloss) * dAtten * light.color * dAtten3 * getFresnel(dot(dViewDirW, halfDir), gloss, specularity, iridescence);
```
mvaligursky, line 546: should we pass dReflDirW and ccReflDirW (and similar if any) as function arguments instead of using global, to limit changes to chunk in the future? Or are we ok to use globals for temp variables.

src/scene/shader-lib/chunks/lit/frag/lightSheen.js
```diff
@@ -14,9 +14,9 @@ float sheenV(vec3 normal, vec3 view, vec3 light) {
     return 1.0 / (4.0 * (NoL + NoV - NoL * NoV));
 }
 
-float getLightSpecularSheen(vec3 h) {
-    float D = sheenD(dNormalW, h, sGlossiness);
-    float V = sheenV(dNormalW, dViewDirW, -dLightDirNormW);
+float getLightSpecularSheen(vec3 h, vec3 worldNormal, float sheenGloss) {
+    float D = sheenD(worldNormal, h, sheenGloss);
+    float V = sheenV(worldNormal, dViewDirW, -dLightDirNormW);
```
mvaligursky, line 19: same here, do we want these to be passed in?
dViewDirW, dLightDirNormW

src/scene/shader-lib/chunks/standard/frag/litShaderArgs.js
```diff
@@ -0,0 +1,49 @@
+export default /* glsl */`
+
+struct IridescenceArgs
```
mvaligursky, line 3: add comments to this file please, for each member.

GSterbrant: Done

src/scene/shader-lib/chunks/standard/frag/litShaderArgs.js
```diff
@@ -0,0 +1,85 @@
...
+{
+    // The color of the fresnel effect, set by the backend
```
mvaligursky, line 5: this is 'set by the backend' - should it be specified in the structure that represents the input to the backend?

src/scene/shader-lib/chunks/standard/frag/litShaderArgs.js
```diff
@@ -0,0 +1,85 @@
...
+    // Glossiness of clearcoat layer, range [0..1]
+    float gloss;
+
+    // The normal used for the clearcoat layer
+    vec3 worldNormal;
+};
+
+struct SheenArgs
+{
+    /// Glossiness of the sheen layer, range [0..1]
```
mvaligursky, line 29:
```suggestion
    // Glossiness of the sheen layer, range [0..1]
```

src/scene/shader-lib/chunks/standard/frag/litShaderArgs.js
```diff
@@ -0,0 +1,85 @@
...
+    float gloss;
+
+    // Surface metalness factor, range [0..1]
+    float metalness;
+
+    // Specularity intensity factor, range [0..1]
+    float specularityFactor;
+
+    // Ambient occlusion amount, range [0..1]
+    float ao;
```
mvaligursky, line 65: It seems this term is used in the chunks to occlude both diffuse and specular .. if I understand this correctly, is it correct to call it `ao` and not simply occlusion?

src/scene/shader-lib/chunks/lit/frag/clusteredLight.js
```diff
@@ -539,49 +556,93 @@ void evaluateLight(ClusterLightData light) {
...
+    ClearcoatArgs clearcoat, 
+    SheenArgs sheen, 
+    IridescenceArgs iridescence
+) {
 
     // skip lights if no lights at all
     if (clusterSkip > 0.5)
         return;
-
+
```
mvaligursky, line 645: remove these added spaces

src/scene/shader-lib/programs/lit-shader.js
```diff
@@ -50,13 +51,6 @@ const builtinVaryings = {
 };
 
 class LitShader {
-    /**
```
mvaligursky, line 53: why removing JSDocs from here? I think I added them so that we get member validation inside the function.

src/scene/shader-lib/chunks/standard/frag/litShaderArgs.js
GSterbrant: What do you mean? Pointing this out means that setting this in the frontend will have no effect as the value gets replaced sometime later.

mvaligursky: so why expose this value in a structure the user is expected to fill with data, if the data is ignored / overwritten? It should probably not be part of this struct which represents a public API as an output of front end.

src/scene/shader-lib/programs/lit-shader.js
GSterbrant: This was not intended, probably just a bad merge.

GSterbrant: Fixed.

src/scene/shader-lib/chunks/standard/frag/litShaderArgs.js
GSterbrant: Fixed.

src/scene/shader-lib/chunks/standard/frag/litShaderArgs.js
GSterbrant: It's debatable. We don't provide a specular occlusion term on the material, but we do however use the AO maps to occlude specularity as well.

mvaligursky (approved):
I'm happy with this. Any thoughts on your side @slimbuck or @willeastcott ?

src/scene/shader-lib/chunks/lit/frag/TBNfast.js
```diff
@@ -1,5 +1,5 @@
 export default /* glsl */`
-void getTBN() {
-    dTBN = mat3(dTangentW, dBinormalW, dVertexNormalW);
+void getTBN(vec3 tangent, vec3 binormal, vec3 normal) {
+    dTBN = mat3(tangent, binormal, normal);
```
willeastcott, line 3: Is this chunk actually used anywhere? Still worthwhile keeping?

src/scene/shader-lib/chunks/lit/frag/lightSpecularAnisoGGX.js
```diff
@@ -1,6 +1,6 @@
 export default /* glsl */`
 // Anisotropic GGX
-float calcLightSpecular(float tGlossiness, vec3 tNormalW, vec3 h) {
+float calcLightSpecular(float tGlossiness, vec3 tNormalW, vec3 viewDir, vec3 h, vec3 lightDirNorm, mat3 tbn) {
```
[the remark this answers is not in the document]
willeastcott: And maybe `tGlossiness` to `gloss`?

willeastcott (approved):
Approving with a few comments/questions.

GSterbrant:
Ran it on PC with a 7900 XTX, my M1 MacBook Pro, an iPhone 14 Pro and a Pixel 2 XL with and without this PR on main branch. Did not see any performance difference whatsoever.

src/scene/shader-lib/chunks/lit/frag/lightSpecularAnisoGGX.js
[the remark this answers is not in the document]
GSterbrant: Done!

src/scene/shader-lib/chunks/lit/frag/TBNfast.js
GSterbrant: It's switched on with a material parameter. We could definitely delete this one, but would prefer to do so in a separate PR since it changes the material API.

PR #5106, opened 2023-02-27
Fix memory leak on mesh collider destroy

### LINKED ISSUE #4267: Memory leak when mesh collider is used
When an entity is destroyed, e.g. with `entity.destroy()`, the component system removes the component:
https://github.com/playcanvas/engine/blob/a9abcbef983edadc7346fda851d39996722c7dbf/src/framework/components/system.js#L74-L86

`beforeremove` event triggers `beforeRemove` method of the `CollisionSystemImpl`, which destroys the shape:
https://github.com/playcanvas/engine/blob/a9abcbef983edadc7346fda851d39996722c7dbf/src/framework/components/collision/system.js#L158-L172

`remove` event then triggers the `remove` method, which in turn calls `destroyShape` of the `CollisionMeshSystemImpl`, that implements the mesh collision:
https://github.com/playcanvas/engine/blob/a9abcbef983edadc7346fda851d39996722c7dbf/src/framework/components/collision/system.js#L507-L525

`destroyShape` however, exits early, since the shape was already destroyed in the previous step. Since the mesh collider is created using a `btCompoundShape`, as a result, we only destroy the parent mesh, skipping the children, hence the memory leak.

Consider adding a conditional to  `beforeRemove` method of the `CollisionSystemImpl` when destroying the shape, letting the shape destruction to be handled by the collision implementation:
```js
// Let implementation destroy it properly
if (component.type !== 'mesh') {
    Ammo.destroy(component.data.shape);
    component.data.shape = null;
}
```

MushAsterion:
Fixes #4267 

I know it doesn't match the suggestion from the issue. However I think having a `CollisionSystemImpl#destroyShape` function sound more beneficial than making the parent class aware of an extension. It also removes override of the `CollisionMeshSystemImpl#remove` function and make sure there is an actual shape to destroy.


willeastcott:
Hey @LeXXik - you submitted the original issue. Got any thoughts here? 😄

LeXXik:
@willeastcott so, are we keeping the data object on the collision component then? I thought you might have been cooking an update for it, hence the last one didn't go through. 

@MushAsterion, I think it would be good if we add a test that covers this change. If you are short on time, I can add it later this week. The test would need to cover at least clearing of the collision shapes (primitive, compounds, mesh and triggers). We don't need to test Ammo, so its methods can be stubbed. We just need to make sure we are calling the destroy methods on our side whenever needed.

MushAsterion:
> I think it would be good if we add a test that covers this change. If you are short on time, I can add it later this week. The test would need to cover at least clearing of the collision shapes (primitive, compounds, mesh and triggers). We don't need to test Ammo, so its methods can be stubbed. We just need to make sure we are calling the destroy methods on our side whenever needed.

That makes sense and time is not really a problem to me. However I have to admit that I have absolutely no knowledge or skills on this matter and when I look at the existing tests none already handle Ammo which sound quite complex to mirror especially on complex shapes... I would love to learn so if you have any example I can use to do so I'll be more than happy to give it a try!

LeXXik:
@MushAsterion you can take a look at these tests that handle Ammo, as an example:

https://github.com/playcanvas/engine/pull/4144/commits/2a07dacae527eb3abcabc0efafd884bd969398cc

MushAsterion:
Thank you @LeXXik! Took a bit of time but I managed to mirror Ammo based on the IDL from playcanvas/ammo.js repo and added the tests for destruction. Did my best to keep it simple but it might be still a bit messy...

I wasn't sure about creating a `.mjs` file only for Ammo but this way it keeps it separate since it is quite long. I know not all functions are used by the engine at the moment but as I was on it I just added everything based off the IDL so it won't cause issue if more features are used.

test/ammo.mjs
```diff
@@ -0,0 +1,1635 @@
...
+ */
+function assertVec3(vec, x, y, z) {
+    assertNum(vec.x, x);
+    assertNum(vec.y, y);
+    assertNum(vec.z, z);
+}
+
+// Stub Ammo
+
+// Note: do not use arrow functions in the Ammo root object, since the engine will
```
LeXXik, line 73: This comment is not applicable here and can be removed.

LeXXik (approved):
This is amazing. I'd say you went a bit overkill with the Ammo stub module :) However, as you mentioned it can definitely be used to cover all current and potential future features we add to the engine.

test/ammo.mjs
MushAsterion: It's removed!

src/framework/components/collision/system.js
```diff
@@ -155,6 +154,15 @@ class CollisionSystemImpl {
...
     }
 
+    destroyShape(data) {
+        if (!data.shape) {
+            return;
+        }
+
+        Ammo.destroy(data.shape);
+        data.shape = null;
+    }
```
mvaligursky, line 164:
```suggestion
    destroyShape(data) {
        if (data.shape) {
            Ammo.destroy(data.shape);
            data.shape = null;
        }
    }
```

mvaligursky (approved):
The PR looks amazing. I added a small suggestion to the code, to make it slightly more compact.

The test looks a lot better than expected, I'm happy here.

Approving, but would like @willeastcott's thoughts here as well.

src/framework/components/collision/system.js
MushAsterion: In fact this was to follow contributing guidelines on [early exiting](https://github.com/playcanvas/engine/blob/main/.github/CONTRIBUTING.md#exit-logic-early), are you sure this would really better?

mvaligursky: I understand where you're coming from, but for a short function like this, the early exit provides no advantages.

mvaligursky:
Perhaps you can then split this PR into two .. one which fixes the memory leak, and the test as a separate. This would allow us to merge the fix while waiting for latest Ammo.

src/framework/components/collision/system.js
MushAsterion: Alright, it's committed then

MushAsterion:
Moved the tests to #5322!

mvaligursky (approved):
Perfect, thanks!

PR #5109, opened 2023-02-27
Remove Material Asset on Material change

### LINKED ISSUE #3560: Material changes are lost when Image Element is disabled
If any material changes are done to a Image element, those will be reverted if you disable the entity.

Example: https://playcanvas.com/editor/scene/1244945

This happens on, and have only been tested on chrome Version 94.0.4606.81.

Repo Steps:
* Create new Image Element.
* Assign a Material to the Element
* Write code that in runtime will clone the material.
* Disable and enable the Element.
* The Image Element will now have the initial material assigned to it, as seen in the screenshot bellow. 

image

MushAsterion:
Fixes #3560 

Set `ImageElement#materialAsset` to `null` on `material` change if new material does not match `materialAsset` to avoid overwrite from `ImageElement#onEnable`. Important to use setter instead of `_materialAsset` to remove listeners.


src/framework/components/element/image-element.js
```diff
@@ -1034,13 +1043,15 @@ class ImageElement {
...
 
-            this._materialAsset = _id;
-            if (this._materialAsset) {
-                const asset = assets.get(this._materialAsset);
+            this._materialAsset = null;
+            if (_id) {
+                const asset = assets.get(_id);
                 if (!asset) {
                     this.material = null;
+                    this._materialAsset = _id;
```
querielo, line 1051: `this._materialAsset = _id;` is in both branches of the `if`.

Can we replace only 1037 by `this._materialAsset = _id || null;`?

querielo (approved):

src/framework/components/element/image-element.js
MushAsterion: The point is that if `this._materialAsset` is not `null` when we set `this.material` to `null`, `this.material` setter will call `this.materialAsset` to `null` which we don't want. So I had to make sure it's `null` before assigning to the right value after processing with `material` reset.

It's also not possible to change `this._material` directly as it won't reset the displayed material.

So yeah a bit upside-down but I couldn't find any better way to do it...

willeastcott: @querielo Any further comments here or is this particular conversation resolved?

MushAsterion: I moved to the other solution
```
this._materialAsset = null;
this.material = null;

this._materialAsset = _id;
```

So it's more clear that the `null` assignment is only for material `null`ing and reduces the number of changed lines, as it's only to reduces call from `material` setter and avoid a loop between the two.

In fact the bug was coming from material setter not changing the `_materialAsset` and the enable callback using the `_materialAsset` if any, overriding any existing `_material` if it was from a different ID/no ID. So the main point is to remove `_materialAsset` inside `material` setter which was then making a loop with `materialAsset` setter if `_materialAsset` was non-`null` or different, why the `null`.

willeastcott (approved):

slimbuck (approved):
Approving, sorry for the long wait @MushAsterion !!

mvaligursky:
It looks good, merging in, thanks all!

src/framework/components/element/image-element.js
```diff
@@ -989,6 +989,15 @@ class ImageElement {
         }
 
         this._material = value;
+
+        // Remove material asset if changed
+        if (this._materialAsset) {
+            const asset = this._system.app.assets.get(this._materialAsset);
+            if (!asset || asset.resource !== value) {
+                this.materialAsset = null;
```
yaustar, line 997: Should this line be `this._materialAsset = null`? as calling the setter for `materialAsset` also sets the `_material` to null as well?

willeastcott:
This PR seems to be causing a problem reported by a forum user:

https://forum.playcanvas.com/t/cannot-set-properties-of-alphatest-for-materials-even-on-playcanvas-examples/31387

They're cloning the material on an element and replacing the element's material with the clone. This internally sets materialAsset = null which in turn sets material to null. I'm wondering if we should back this out. Got any thoughts @MushAsterion?

yaustar:
The fix in https://github.com/playcanvas/engine/pull/5109 has not been merged to a release yet @willeastcott

mvaligursky:
You're right! With so many patch releases, I was sure this was out already.

PR #5112, opened 2023-02-27
[Anim Component] Support empty AnimTracks

### LINKED ISSUE #3267: Whole state graph doesn't play animation is there is one missing animation on the component
It would be good if there was some way we could somehow still play the animation graph even if there is an invalid state.

ellthompson:
The anim component can only play when all states have been assigned an animation. However, it is sometimes necessary to test out the behaviour of a state graph before the animation data is available. This PR adds support for empty AnimTracks, which can be assigned to a state without requiring any animation data to be loaded.

When a set of animation assets are loaded into the state graph, any missing states will be filled with empty AnimTracks, allowing the state graph to still play.

Fixes #3267


willeastcott:
Back out change to `package-lock.json` since it's not really relevant to this PR.

src/framework/anim/evaluator/empty-anim-track.js
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * This AnimTrack can be used as a placeholder track when creating a state graph before having all associated animation data available.
+ */
+class EmptyAnimTrack extends AnimTrack {
+    /**
+     * Create a new empty AnimTrack instance.
+     *
+     * @hideconstructor
+     */
+    constructor() {
```
mvaligursky, line 12: this looks like you just need an instance of AnimTrack with empty data instead of a new class?

src/framework/anim/evaluator/anim-evaluator.js
```diff
@@ -184,7 +184,7 @@ class AnimEvaluator {
      * @param {number} deltaTime - The amount of time that has passed since the last update, in
      * seconds.
      */
-    update(deltaTime) {
+    update(deltaTime, outputAnimation = true) {
```
mvaligursky, line 187: document this parameter

mvaligursky: it could be AnimTrack.EMPTY, similarly to Vec3.ZERO for example.

src/framework/anim/evaluator/empty-anim-track.js
mvaligursky: it could be AnimTrack.EMPTY, similarly to Vec3.ZERO for example.

ellthompson: Yep I think a static property on the AnimTrack class would be better, I'll update it to that.

src/framework/anim/evaluator/anim-track.js
```diff
@@ -110,4 +110,9 @@ class AnimTrack {
     }
 }
 
+/**
+ * This AnimTrack can be used as a placeholder track when creating a state graph before having all associated animation data available.
+ */
+AnimTrack.EMPTY = new AnimTrack('empty', Number.MAX_VALUE, [], [], []);
```
mvaligursky, line 116: could this please use the same format as for example static colors:

    static WHITE = Object.freeze(new Color(1, 1, 1, 1));

src/framework/anim/controller/anim-state.js
```diff
@@ -96,6 +98,11 @@ class AnimState {
             node.animTrack = animTrack;
             this._animationList.push(node);
         }
+        this._updateHasAnimations();
+    }
+
+    _updateHasAnimations() {
+        this._hasAnimations = this._animationList.length > 0 && this._animationList.filter(animation => !animation.animTrack || animation.animTrack === AnimTrack.EMPTY).length === 0;
```
mvaligursky, line 105: I think instead of .filter this should use .some

mvaligursky (approved):
Approving with some suggested improvements.

src/framework/anim/evaluator/anim-track.js
kungfooman: Technically it would require a recursive/deep freeze, since the arrays are still mutable

PR #5118, opened 2023-03-01
WebGPU support to shadow atlas (shadow cascades + clustered shadows)

mvaligursky:
- shadow shaders updated to work with WebGPU
- added a private way to display unfiltered texture for debugging, such as shadow map (WebGPU only) - this involved supporting multiple samplers per texture.
- handling renderTarget.flipY when rendering using a viewport (shadow atlas) by flipping viewport as well on WebGPU
- using WebGPU error scopes to report bind group validation errors with additional relevant data (used descriptors ...) to considerably improve understanding of those errors. To be used to validate other WeGPU APIs in the future. Based on https://toji.dev/webgpu-best-practices/error-handling

Missing for full support of clustered shadows:
- support to clear only part of the render target (viewport) - WebGPU does not have a clear function to do this, and it needs to be implemented using a quad draw. Separate PR. This is to handle a case where not all shadows are updated per frame, and so the whole render target cannot get cleared.

https://user-images.githubusercontent.com/59932779/222208711-cc3eea9a-fbd6-4555-9bee-6a59c3875f01.mov

### DIFF
--- a/src/framework/app-base.js
+++ b/src/framework/app-base.js
@@ -1894,7 +1894,12 @@ class AppBase extends EventHandler {
      * @param {Layer} [layer] - The layer to render the texture into. Defaults to {@link LAYERID_IMMEDIATE}.
      * @ignore
      */
-    drawTexture(x, y, width, height, texture, material, layer = this.scene.defaultDrawLayer) {
+    drawTexture(x, y, width, height, texture, material, layer = this.scene.defaultDrawLayer, filterable = true) {
+
+        // only WebGPU supports filterable parameter to be false, allowing a depth texture / shadow
+        // map to be fetched (without filtering) and rendered
+        if (filterable === false && !this.graphicsDevice.isWebGPU)
+            return;
 
         // TODO: if this is used for anything other than debug texture display, we should optimize this to avoid allocations
         const matrix = new Mat4();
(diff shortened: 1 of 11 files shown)


src/framework/app-base.js
```diff
@@ -1894,7 +1894,12 @@ class AppBase extends EventHandler {
      * @param {Layer} [layer] - The layer to render the texture into. Defaults to {@link LAYERID_IMMEDIATE}.
      * @ignore
      */
-    drawTexture(x, y, width, height, texture, material, layer = this.scene.defaultDrawLayer) {
+    drawTexture(x, y, width, height, texture, material, layer = this.scene.defaultDrawLayer, filterable = true) {
```
willeastcott, line 1897: Shouldn't filterable parameter be documented? It'll show up in Intellisense right?

src/platform/graphics/webgpu/webgpu-texture.js
```diff
@@ -184,15 +189,21 @@ class WebgpuTexture {
     }
 
     // TODO: handle the case where those properties get changed
+    // TODO: share a globl map of samplers. Possibly even use shared samplers for bind group,
```
willeastcott, line 192:
```suggestion
// TODO: share a global map of samplers. Possibly even use shared samplers for bind group,
```

willeastcott (approved):
Approving with minor feedback...

src/framework/app-base.js
mvaligursky: good point, done

src/platform/graphics/webgpu/webgpu-texture.js
mvaligursky: thanks!

PR #5126, opened 2023-03-06
Alpha test fix

### LINKED ISSUE #5115: Shadows don’t work when using Vertex Alpha and Alpha Test
We’ve run into a case where we need to use Alpha Test to “cut” a model. Our model has a Vertex Alpha gradient, 0-1, from left to right. On the material, under opacity, we enable Vertex Color (which uses the Vertex Alpha channel), and then set a Alpha Test value. We noticed that realtime shadows don’t respect this value when using Vertex Alpha. When using textures instead of Vertex Alpha though, shadows do respect the Alpha Test value.

The following image is an example of a running application. The top row uses Alpha Test and a gradient texture. The bottom row uses Alpha Test and a Vertex Alpha gradient:
AlphaTestShadows

In the editor, shadows often freak out when using Vertex Alpha and Alpha test. Shadows flicker on and off depending on if you hover the mouse over the viewport or not, and the Alpha Test value that the light “sees” and uses for casting shadows depends on what object you have selected.

Also, using an Alpha Test value of 1 with Vertex Alpha seems to confuse the shader, creating an interference pattern. I’d expect a value of 1 would hide the entire object.

Here’s a test PlayCanvas project. Please take a look and see if this is something that can be fixed:
https://playcanvas.com/project/1044631/overview/alpha-test-shadows

Related Forum thread:
https://forum.playcanvas.com/t/shadows-dont-work-when-using-vertex-alpha-and-alpha-test/29932

slimbuck:
Fixes #5115 

The material alpha test value wasn't being set without an alpha map specified.

### DIFF
--- a/src/scene/renderer/renderer.js
+++ b/src/scene/renderer/renderer.js
@@ -505,6 +505,8 @@ class Renderer {
         // Alpha test
         if (material.opacityMap) {
             this.opacityMapId.setValue(material.opacityMap);
+        }
+        if (material.opacityMap || material.alphaTest > 0) {
             this.alphaTestId.setValue(material.alphaTest);
         }
     }


src/scene/renderer/renderer.js
```diff
@@ -505,6 +505,8 @@ class Renderer {
         // Alpha test
         if (material.opacityMap) {
             this.opacityMapId.setValue(material.opacityMap);
+        }
+        if (material.opacityMap || material.alphaTest > 0) {
```
mvaligursky, line 509: So if the following material has alphaTest === 0, it won't be set and will keep using the previously set value.

slimbuck: If there is no opacity map and alphaTest == 0, then there will be no alpha test shader code?

PR #5129, opened 2023-03-06
Small fixes in examples to avoid debug errors / warnings

mvaligursky:
- avoiding double calling of start, and setting .blend which is now deprecated


slimbuck (commented):
not deprecated?

mvaligursky:
> not deprecated?

not => now

slimbuck (approved):

PR #5139, opened 2023-03-09
Lazily evaluate disjoint timer extension

mvaligursky:
This extension is not typically needed in builds without GPU profiling, so evaluate the extension lazily to avoid the cost at start up.


slimbuck (approved):

willeastcott:
Shouldn't we have a generic approach to lazily wrangling *all* WebGL extensions? I had considered this in the past but never got around to it. Obviously, some extensions will be required in all circumstances....but it still might be worth doing this across the board for consistency.

PR #5145, opened 2023-03-11
Quat: add the setBetweenVectors method and tests for it

querielo:
I have just read the next tweet: https://twitter.com/dejohn_paul/status/1633470425870004228

When I did a gizmo it was handy to have a function that returns a quaternion between two vectors. It'd help the author of the tweet.

So, MR adds the `setBetweenVectors` method to the quaternion class. It also adds second epsilon parameters to Quat.equals, that is used to compare almost equal quaternions.

NOTE: right now, `Quat.equals` considers `new Quat(0.1, 0.2, 0.3, 0.4)` and  `new Quat(-0.1, -0.2, -0.3, -0.4)` as non-equal that is not correct in semantic. Should we fix it?

### DIFF
--- a/src/core/math/quat.js
+++ b/src/core/math/quat.js
@@ -104,14 +104,22 @@ class Quat {
      * Reports whether two quaternions are equal.
      *
      * @param {Quat} rhs - The quaternion to be compared against.
+     * @param {number} [epsilon] - The maximum difference between each component of the two quaternions.
      * @returns {boolean} True if the quaternions are equal and false otherwise.
      * @example
      * var a = new pc.Quat();
      * var b = new pc.Quat();
      * console.log("The two quaternions are " + (a.equals(b) ? "equal" : "different"));
      */
-    equals(rhs) {
-        return ((this.x === rhs.x) && (this.y === rhs.y) && (this.z === rhs.z) && (this.w === rhs.w));
+    equals(rhs, epsilon) {
+        if (epsilon === undefined) {
+            return ((this.x === rhs.x) && (this.y === rhs.y) && (this.z === rhs.z) && (this.w === rhs.w));
+        } else {
+            return (Math.abs(this.x - rhs.x) < epsilon) &&
+                   (Math.abs(this.y - rhs.y) < epsilon) &&
+                   (Math.abs(this.z - rhs.z) < epsilon) &&
+                   (Math.abs(this.w - rhs.w) < epsilon);
+        }
     }
 
     /**
@@ -532,6 +540,46 @@ class Quat {
                 this.y = (m21 + m12) * rs;
             }
         }
+
+        return this;
+    }
+
+    /**
+     * Set the quaternion that represents the shortest rotation from one direction to another.
+     *
+     * @param {import('./vec3').Vec3} from - The direction to rotate from. It should be normalized.
+     * @param {import('./vec3').Vec3} to - The direction to rotate to. It should be normalized.
+     * @returns {Quat} Self for chaining.
+     *
+     * Prood of correctness: https://www.xarg.org/proof/quaternion-from-two-vectors/
+     */
+    setBetweenVectors(from, to) {
+        let dotProduct = 1 + from.dot(to);
+
+        if (dotProduct < Number.EPSILON) {
+            // the vectors point in opposite directions
+            // so we need to rotate 180 degrees around an arbitrary orthogonal axis
+
+            if (Math.abs(from.x) > Math.abs(from.y)) {
+                this.x = -from.z;
+                this.y = 0;
+                this.z = from.x;
+                this.w = 0;
+            } else {
+                this.x = 0;
+                this.y = -from.z;
+                this.z = from.y;
+                this.w = 0;
+            }
+        } else {
+            // cross product between the two vectors
+            this.x = from.y * to.z - from.z * to.y;
+            this.y = from.z * to.x - from.x * to.z;
+            this.z = from.x * to.y - from.y * to.x;
+            this.w = dotProduct;
+        }
+
+        this.normalize();
 
         return this;
     }
--- a/test/core/math/quat.test.mjs
+++ b/test/core/math/quat.test.mjs
@@ -95,6 +95,19 @@ describe('Quat', function () {
             expect(q1.equals(q2)).to.be.false;
         });
 
+        it('checks for equality of two different quaternions that are close enough', function () {
+            const q1 = new Quat(0.1, 0.2, 0.3, 0.4);
+            const q2 = new Quat(0.10000000000000001, 0.2, 0.3, 0.4);
+            const epsilon = 0.000001;
+            expect(q1.equals(q2, epsilon)).to.be.true;
+
+            const q3 = new Quat(0.1 + epsilon - Number.EPSILON, 0.2, 0.3, 0.4);
+            expect(q1.equals(q3, epsilon)).to.be.true;
+
+            const q4 = new Quat(0.1 + epsilon + Number.EPSILON, 0.2, 0.3, 0.4);
+            expect(q1.equals(q4, epsilon)).to.be.false;
+        });
+
     });
 
     describe('#getAxisAngle()', function () {
@@ -544,6 +557,56 @@ describe('Quat', function () {
             const q = new Quat();
             const m = new Mat4();
             expect(q.setFromMat4(m)).to.equal(q);
+        });
+
+    });
+
+    describe('#setBetweenVectors()', function () {
+
+        it('set the identity quaternion between equal directions', function () {
+            const v1 = new Vec3(1, 0, 0);
+            const v2 = new Vec3(1, 0, 0);
+
+            const q1 = new Quat().setBetweenVectors(v1, v2);
+            expect(q1.equals(Quat.IDENTITY)).to.be.true;
+
+
+            const v3 = new Vec3(0, 0, 0);
+            const v4 = new Vec3(0, 0, 0);
+
+            const q2 = new Quat().setBetweenVectors(v3, v4);
+            expect(q2.equals(Quat.IDENTITY)).to.be.true;
+        });
+
+        it('set the quaternion between different direction', function () {
+            const epsilon = 0.00001;
+
+            const v1 = new Vec3(1, 0, 0);
+            const v2 = new Vec3(0, 1, 0);
+
+            const q1 = new Quat().setBetweenVectors(v1, v2);
+            const q2 = new Quat().setFromEulerAngles(0, 0, 90);
+
+            expect(q1.equals(q2, epsilon)).to.be.true;
+
+            const v3 = new Vec3(1, 0, 0);
+            const v4 = new Vec3(1, 1, 0).normalize();
+
+            const q3 = new Quat().setBetweenVectors(v3, v4);
+            const q4 = new Quat().setFromEulerAngles(0, 0, 45);
+
+            expect(q3.equals(q4, epsilon)).to.be.true;
+
+            const q5 = new Quat().setFromEulerAngles(0, 0, 44);
+            expect(q3.equals(q5, epsilon)).to.be.false;
+
+        });
+
+        it('returns this', function () {
+            const q = new Quat();
+            const v1 = new Vec3();
+            const v2 = new Vec3();
+            expect(q.setBetweenVectors(v1, v2)).to.equal(q);
         });
 
     });


LeXXik:
I also often have a need in a utility function that calculates a quaternion between 2 vectors.  I support this.

I think equality check is correct at the moment. That is to say the current implementation checks the equality of the vector components. If there is a disambiguity, we should fix the docs instead. If a rotation equality is needed, it should have own method.

Maksims:
Does the quat persists the direction of an arc between two vectors?
The naming feels a bit awkward. Also, if we just want to return an angle between two vectors, should it be a method on a Vec3 instead?
Vec3.angleTo(Vec3) something like so?
Also, what about Vec2, often needed too.

LeXXik:
Agree with Max here on class ownership. I've extended Vec3 for this and have a separate for quats under Quat class as well.

querielo:
> Does the quat persists the direction of an arc between two vectors?

Yes. There is the next test: https://github.com/playcanvas/engine/pull/5145/files#diff-172bbc76634fad0abc6f263547aa1930cfa62f9f7e45d444f9b4091eec7ac0beR596

> The naming feels a bit awkward.

What name do you suggest? I have just renamed it: setFromDirections (similar to **setFrom**AxisAngle, **setFrom**EulerAngles, **setFrom**Mat4). Is it fine?

> Also, if we just want to return an angle between two vectors, should it be a method on a Vec3 instead?
> Vec3.angleTo(Vec3) something like so?

Yes. The method that returns an angle between two vectors should be in Vec3. But I think that the method that returns a quaternion that represent the shortest rotation between 2 vectors has to be in Quat instead of Vec3.

1. We already have methods Quat.set*
2. Vec3 doesn't "know" about Quat and I would keep it this way.

kungfooman:
I feel slightly uneasy adding an epsilon to the `equals` method:

1) proximity != equality
2) adds an overhead `if` condition to every "normal" equality test, only used for a unit test
3) API would be more expressive calling it what it is:

```js
    // Other name ideas: nearlyEquals, equalsApprox, equalsWithEpsilon etc.
    equalsNearly(rhs, epsilon) {
        return (Math.abs(this.x - rhs.x) < epsilon) &&
            (Math.abs(this.y - rhs.y) < epsilon) &&
            (Math.abs(this.z - rhs.z) < epsilon) &&
            (Math.abs(this.w - rhs.w) < epsilon);
    }
```

4) No other math class (like `Vec3`) offers the capability either

Just for brain storming, maybe someone has better ideas for a fast/unified API?

querielo:
> I feel slightly uneasy adding an epsilon to the `equals` method:

Partly agree.
1) [Strictly/mathematically speaking it depends](https://en.wikipedia.org/wiki/Equivalence_relation). Especially, when we are speaking about floating-point arithmetic, where equality is defined with proximity by design.
2) I'm not sure that there are real use cases for the strict checking (unit tests are exception), because when we are working with floating-point arithmetic it is better use epsilon (especially, with something that uses trigonometry). So, I am not sure that anybody notice it.
3) Agree.
4) Should I change it in PR?

Note: I changed it [for the tests](https://github.com/playcanvas/engine/pull/5145/files#diff-172bbc76634fad0abc6f263547aa1930cfa62f9f7e45d444f9b4091eec7ac0beR590).

kungfooman:
1) Are you talking about different bits representing the same floating point number? That is already covered by `===`:

```js
const a     = new Float64Array([0]);
const a_int = new Uint32Array(a.buffer);
const b     = new Float64Array([-0]);
const b_int = new Uint32Array(b.buffer);
console.log(a[0] === b[0]); // true
// Even though bit-wise they are different:
console.log("a_int", ...a_int); // 0 0
console.log("b_int", ...b_int); // 0 2147483648
```

2) Two engine examples:

https://github.com/playcanvas/engine/blob/b334160ee41627d8be82b7201c2338d987671e8b/src/scene/materials/standard-material-options-builder.js#L349

https://github.com/playcanvas/engine/blob/b334160ee41627d8be82b7201c2338d987671e8b/src/scene/scene.js#L636-L638

4) Good question, personally I like this little helper methods, but `Vec3#distanceSquared` didn't make it into vector classes aswell (e.g. to get rid of `Math.sqrt` for simple/fast distance checks)

querielo:
> 1. Are you talking about different bits representing the same floating point number? That is already covered by `===`:

No. I mean that semantically equivalent expressions can give different results when we use floating-point arithmetic, especially when we use trigonometrical functions.
 
Screenshot 2023-03-13 at 00 22 51

So, we can't just use `equals` since n-th digits of the numbers are differ (precision error, that tends to accumulate), and

```js
    expect(new Quat().setFromDirections(new Vec3(1, 0, 0), new Vec3(0, 1, 0))
        .equals(new Quat().setFromEulerAngles(0, 0, 90))).to.be.true;
```

would always throws an exception. It is not a good idea to compare float variables with `==, ===, !=, !==` especially when they are got from different sources with trigonometrical function and not only.

2) Both of these cases are executed a few times per scene/material lifetime. They are optimizations and probably can return false negative result if the quaternion is changed/computed somehow.

**OFFTOP**: does `options.litOptions.useCubeMapRotation = usingSceneEnv && scene.skyboxRotation && !scene.skyboxRotation.equals(Quat.IDENTITY);` mean that we have to recompile the shader if the skybox is rotated?

src/core/math/quat.js
```diff
@@ -536,6 +543,45 @@ class Quat {
...
+            }
+        } else {
+            // cross product between the two vectors
+            this.x = from.y * to.z - from.z * to.y;
+            this.y = from.z * to.x - from.x * to.z;
+            this.z = from.x * to.y - from.y * to.x;
+            this.w = dotProduct;
+        }
+
+        this.normalize();
```
mvaligursky, line 580:
```suggestion
        return this.normalize();
```

src/core/math/quat.js
```diff
@@ -536,6 +543,45 @@ class Quat {
...
+
+        return this;
```
mvaligursky, line 582:
```suggestion
        return this.normalize();
```

src/core/math/quat.js
```diff
@@ -104,14 +104,21 @@ class Quat {
...
+     * @param {number} [epsilon] - The maximum difference between each component of the two quaternions.
      * @returns {boolean} True if the quaternions are equal and false otherwise.
      * @example
      * var a = new pc.Quat();
      * var b = new pc.Quat();
      * console.log("The two quaternions are " + (a.equals(b) ? "equal" : "different"));
      */
-    equals(rhs) {
-        return ((this.x === rhs.x) && (this.y === rhs.y) && (this.z === rhs.z) && (this.w === rhs.w));
+    equals(rhs, epsilon) {
```
mvaligursky, line 114: I'd vote for moving the version with epsilon to equalsAprox, and leave original function as is.

willeastcott: What about `closeTo`?

src/core/math/quat.js
```diff
@@ -536,6 +543,43 @@ class Quat {
         return this;
     }
 
+    /**
+     * Set the quaternion that represents the shortest rotation from one direction to another.
+     *
+     * @param {import('./vec3').Vec3} from - The direction to rotate from. It should be normalized.
+     * @param {import('./vec3').Vec3} to - The direction to rotate to. It should be normalized.
```
willeastcott, line 550: You don't need to import `Vec3`. It's already imported at the top of the module.

src/core/math/quat.js
mvaligursky: closeTo is not something you'd search for, so you'd likely just miss it.

willeastcott: I wouldn't search for approx either.

willeastcott: The only good thing is that `equalsApprox` would be next to `equals` in the API ref.

mvaligursky: yeah, and when you type `quad.equ...`, IntelliSense would offer both.

src/core/math/quat.js
querielo: Fixed.

src/core/math/quat.js
querielo: Added the new `equalsApprox` method.

src/core/math/quat.js
```diff
@@ -114,6 +114,24 @@ class Quat {
...
+    /**
+     * Reports whether two quaternions are equal using an absolute error tolerance.
+     *
+     * @param {Quat} rhs - The quaternion to be compared against.
+     * @param {number} [epsilon] - The maximum difference between each component of the two quaternions.
+     * @returns {boolean} True if the quaternions are equal and false otherwise.
+     * @example
+     * var a = new pc.Quat();
+     * var b = new pc.Quat();
+     * console.log("The two quaternions are " + (a.equals(b, 1e-9) ? "equal" : "different"));
```
mvaligursky, line 126:
```suggestion
     * console.log("The two quaternions are approximately " + (a.equalsApprox(b, 1e-9) ? "equal" : "different"));
```

mvaligursky (approved):
Looks great, thank you!

test/core/math/quat.test.mjs
```diff
@@ -548,6 +561,56 @@ describe('Quat', function () {
...
+        it('set a quaternion from different directions', function () {
+            const epsilon = 0.00001;
+
+            const v1 = new Vec3(1, 0, 0);
+            const v2 = new Vec3(0, 1, 0);
+
+            const q1 = new Quat().setFromDirections(v1, v2);
+            const q2 = new Quat().setFromEulerAngles(0, 0, 90);
+
+            expect(q1.equalsApprox(q2, epsilon)).to.be.true;
```
kungfooman, line 590: The point about having a custom method is that you can easily set default values, which you don't need to re-define here any longer, so we have shorter and fewer lines

src/core/math/quat.js
```diff
@@ -536,6 +554,43 @@ class Quat {
...
     }
 
+    /**
+     * Set the quaternion that represents the shortest rotation from one direction to another.
+     *
+     * @param {Vec3} from - The direction to rotate from. It should be normalized.
+     * @param {Vec3} to - The direction to rotate to. It should be normalized.
+     * @returns {Quat} Self for chaining.
+     *
+     * Proof of correctness: https://www.xarg.org/proof/quaternion-from-two-vectors/
```
willeastcott, line 564:
```suggestion
     * {@link https://www.xarg.org/proof/quaternion-from-two-vectors/ Proof of correctness}
```

willeastcott: I haven't checked this works....but just thought it'd be cool to actually hyperlink this.

src/core/math/quat.js
```diff
@@ -114,6 +114,24 @@ class Quat {
         return ((this.x === rhs.x) && (this.y === rhs.y) && (this.z === rhs.z) && (this.w === rhs.w));
     }
 
+    /**
+     * Reports whether two quaternions are equal using an absolute error tolerance.
+     *
+     * @param {Quat} rhs - The quaternion to be compared against.
+     * @param {number} [epsilon] - The maximum difference between each component of the two quaternions.
```
willeastcott, line 121: Please document the default value for epsilon.

willeastcott (approved):

test/core/math/quat.test.mjs
querielo: Done :-)

src/core/math/quat.js
querielo: Done

test/core/math/quat.test.mjs
```diff
@@ -100,6 +100,7 @@ describe('Quat', function () {
             const q2 = new Quat(0.10000000000000001, 0.2, 0.3, 0.4);
             const epsilon = 0.000001;
             expect(q1.equalsApprox(q2, epsilon)).to.be.true;
+            expect(q1.equalsApprox(q2)).to.be.true;
```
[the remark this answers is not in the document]
querielo: Yes. To check that there is some default value.

`expect(q1.equalsApprox(q2, epsilon)).to.be.true;` --- to check if it works
`expect(q1.equalsApprox(q2)).to.be.true;` --- to check if the parameter has a default value
`expect(q1.equalsApprox(q3, epsilon)).to.be.true;` --- to check if it returns true if coords differ less than maximum difference `epsilon - Number.EPSILON`.
`expect(q1.equalsApprox(q2, epsilon)).to.be.true;` --- to check if it returns false if coords differ more than maximum difference `epsilon + Number.EPSILON`.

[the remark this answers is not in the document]
kungfooman: Very nice, good work to test all cases :+1:

PR #5147, opened 2023-03-13
[Examples Browser] Support device type switching

### LINKED ISSUE #5027: Publish WebGPU-enabled build of Examples Browser
We are probably now ready to publish a build of the Examples Browser (to https://playcanvas.github.io) which supports WebGPU. Some thoughts:

* This will presumably require us to add WebGPU support to the release builds of the engine.
* The Examples Browser should probably have some kind of UI to select the utilized graphics API. So, for example, a dropdown control which enumerates what's supported in the current browser: WebGL, WebGL2 and/or WebGPU. Select one of these options should reload the active example using that API.

ellthompson:
Adds support for graphics device type switching in the examples browser. Users can choose to run each example using WebGL 1, WebGL 2 or WebGPU. 
image

The browser will attempt to run the current example using the selected device, however if that device is not usable for a given example a fallback device will be used instead. This will be displayed to the user as follows:
image

If the user selects an unsupported device type, then switches to an example which can use that device type, their selection will be restored. This allows a user to click through both WebGL and WebGPU examples while keeping track of their preferred device type.

Only examples with the following property set will support the WebGPU device type. This will allow us to progressively enable WebGPU in different examples as its support matures.
```javascript
static WEBGPU_ENABLED = true;
```

Fixes #5027 


examples/src/examples/animation/locomotion.tsx
```diff
@@ -18,7 +18,7 @@ class LocomotionExample {
         </>;
     }
 
-    example(canvas: HTMLCanvasElement, data: any): void {
+    example(canvas: HTMLCanvasElement, deviceType: typeof pc.DEVICETYPE_WEBGPU | typeof pc.DEVICETYPE_WEBGL1 | typeof pc.DEVICETYPE_WEBGL2, data: any): void {
```
mvaligursky, line 21: could we have an enum for this or similar, to allow a single type?

src/platform/graphics/constants.js
```diff
@@ -1179,6 +1179,18 @@ export const DEVICETYPE_WEBGL2 = 'webgl2';
...
+ * Enum for the device types.
+ *
+ * @readonly
+ * @enum {string}
+ */
+export const DEVICETYPE = {
+    WEBGL1: DEVICETYPE_WEBGL1,
+    WEBGL2: DEVICETYPE_WEBGL2,
+    WEBGPU: DEVICETYPE_WEBGPU
+};
```
willeastcott, line 1192: I'm a little confused why we have now the API `DEVICETYPE_XXX` and `DEVICETYPE.XXX` (the latter being out of step with the rest of the API constants). @mvaligursky, this request for an enum was from you, right? Can you clarify what you wanted a bit further?

mvaligursky: my hope is to have a single type, or a similar solution, to avoid this in every example:
'''
deviceType: typeof pc.DEVICETYPE_WEBGPU | typeof pc.DEVICETYPE_WEBGL1 | typeof pc.DEVICETYPE_WEBGL2
'''

ellthompson: It seems like we can either list out all the different options when typing the deviceType property, be less specific with the type of that property or use an enum to group the different device types together.

The enum works pretty well in this instance but like @willeastcott said it'd be the first instance of `@enum` in the codebase.

mvaligursky: By enum I assumed a type created in the example browser, not in the engine, as that type is only even used by the example browser.

willeastcott: Yeah, that's what I thought you meant too. In the engine source, I have no idea what `@enum` would generate in terms of JSDoc output (API reference) and TS declarations.

examples/src/examples/animation/locomotion.tsx
mvaligursky: maybe use 'any' or 'string' instead? Those pc.DEVICETYPE_*** constants are strings.

src/platform/graphics/constants.js
mvaligursky: I don't think it should, as current implementation matches the rest of this engine, while this makes it different.
If at some point we decide to adapt enums in some form, it would probably need to be an engine-wide decision.

ellthompson: Ok, i've set the type of the `deviceType` property to string for now.

examples/src/app/example.tsx
```diff
@@ -1,15 +1,29 @@
...
+import { Container, Spinner, SelectInput, Panel } from '@playcanvas/pcui/react';
+import { SelectInput as SelectInputClass } from '@playcanvas/pcui';
 import { File } from './helpers/types';
 import examples from './helpers/example-data.mjs';
 // @ts-ignore: library file import
 import { withRouter } from 'react-router-dom';
 import ControlPanel from './control-panel';
+import { Observer } from '@playcanvas/observer';
+
+const DEVICETYPE_WEBGL1 = 'webgl1';
```
mvaligursky, line 11: do we need these here or can they come from engine constants

src/platform/graphics/graphics-device.js
```diff
@@ -39,6 +39,13 @@ class GraphicsDevice extends EventHandler {
      */
     isWebGPU = false;
 
+    /**
```
mvaligursky, line 42: I'd probably remove this public API. You can detect what device you have by:
device.isWebGPU: webgpu
device.webgl2: webgl2
otherwise: webgl1

mvaligursky: actually, thinking about this, maybe leave it in, it's definitely more convenient. I added it a while back removed in favour is easier to use device.isWebGPU (does not need any import, compared to `device.deviceType === some-constant`) but for some cases this is easier to have.

src/platform/graphics/graphics-device.js
```diff
@@ -39,6 +39,13 @@ class GraphicsDevice extends EventHandler {
...
+     * The type of graphics device that the system could initialize. Can be one of pc.DEVICETYPE_WEBGL1, pc.DEVICETYPE_WEBGL2 or pc.DEVICETYPE_WEBGPU.
+     *
+     * @type {import('./constants.js').DEVICETYPE_WEBGL1 | import('./constants.js').DEVICETYPE_WEBGL2 | import('./constants.js').DEVICETYPE_WEBGPU}
+     */
+    activeDeviceType;
```
mvaligursky, line 47: change it to deviceType please, 'active' suggests it can change, which it cannot. And the description:
`A type of the device. Can be one of ...`

examples/src/app/example.tsx
ellthompson: I've updated these to come from the engine build.

examples/src/app/example.tsx
```diff
@@ -71,14 +96,85 @@ interface ExampleState {
...
 
     componentDidMount() {
         window.localStorage.removeItem(this.path);
         this.props.setFiles(this.defaultFiles);
     }
 
     shouldComponentUpdate(nextProps: Readonly<ExampleProps>): boolean {
-        return this.props.match.params.category !== nextProps.match.params.category || this.props.match.params.example !== nextProps.match.params.example;
+        const updateMobileOnFileChange = () => {
+            return window.top.innerWidth < 601 && this.props.files !== nextProps.files;
```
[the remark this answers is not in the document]
ellthompson: Done 👍

src/platform/graphics/graphics-device.js
```diff
@@ -40,11 +40,11 @@ class GraphicsDevice extends EventHandler {
     isWebGPU = false;
 
     /**
-     * The type of graphics device that the system could initialize. Can be one of pc.DEVICETYPE_WEBGL1, pc.DEVICETYPE_WEBGL2 or pc.DEVICETYPE_WEBGPU.
+     * The type of the device. Can be one of pc.DEVICETYPE_WEBGL1, pc.DEVICETYPE_WEBGL2 or pc.DEVICETYPE_WEBGPU.
```
mvaligursky, line 43: Sorry, I should have realised this earlier. Can we rename this to _deviceType, and make deviceType a getter only (read only access), as currently this feels like a property users can change.

ellthompson: Yep that makes sense.

src/platform/graphics/webgl/webgl-graphics-device.js
```diff
@@ -403,7 +403,7 @@ class WebglGraphicsDevice extends GraphicsDevice {
 
             if (gl) {
                 this.webgl2 = (names[i] === DEVICETYPE_WEBGL2);
-                this.activeDeviceType = (names[i] === DEVICETYPE_WEBGL2) ? DEVICETYPE_WEBGL2 : DEVICETYPE_WEBGL1;
+                this.deviceType = (names[i] === DEVICETYPE_WEBGL2) ? DEVICETYPE_WEBGL2 : DEVICETYPE_WEBGL1;
```
mvaligursky, line 406: use `this.webgl2` insteadof `(names[i] === DEVICETYPE_WEBGL2)`

mvaligursky (approved):
Looks great, cannot wait to test this. 
Left few suggestions before merging.

mvaligursky (approved):

PR #5166, opened 2023-03-18
Deprecate `StandardMaterial#shininess`

### LINKED ISSUE #1081: Rename shininess to gloss

willeastcott:
Fixes #1081


slimbuck (approved):
Thanks for this!

PR #5168, opened 2023-03-19
Support for OffscreenCanvas in Web Workers

### LINKED ISSUE #2481: Investigate implementation of OffscreenCanvas API
Investigate implementation of https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas. This would give us:
- possible additional performance for the application, as playcanvas part of the application including webgl would execute on separate thread
- ability to use Atomics library https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics on both webgl and worker threads, allowing us to write a job system which can execute jobs triggered on webgl thread and get results back inside the same frame, without using messages for synchronization.

michaelfranzl:
Fixes #2481

This probably needs the development of a minimal example (and perhaps automatic tests). Since this is my first contribution, please advise on how to proceed.

I'm happy to discuss and improve the implementation if needed.

### DIFF
--- a/src/core/platform.js
+++ b/src/core/platform.js
@@ -88,6 +88,13 @@ const platform = {
     browser: environment === 'browser',
 
     /**
+     * True if running in a Web Worker.
+     *
+     * @type {boolean}
+     */
+    worker: environment === 'worker',
+
+    /**
      * True if running on a desktop or laptop device.
      *
      * @type {boolean}
--- a/src/core/platform.js
+++ b/src/core/platform.js
@@ -91,6 +91,7 @@ const platform = {
      * True if running in a Web Worker.
      *
      * @type {boolean}
+     * @ignore
      */
     worker: environment === 'worker',
 
--- a/src/framework/application.js
+++ b/src/framework/application.js
@@ -115,7 +115,7 @@ class Application extends AppBase {
      * - sound ({@link SoundComponentSystem})
      * - sprite ({@link SpriteComponentSystem})
      *
-     * @param {HTMLCanvasElement} canvas - The canvas element.
+     * @param {HTMLCanvasElement | OffscreenCanvas} canvas - The canvas element.
      * @param {object} [options] - The options object to configure the Application.
      * @param {ElementInput} [options.elementInput] - Input handler for {@link ElementComponent}s.
      * @param {Keyboard} [options.keyboard] - Keyboard handler for input.
--- a/src/platform/graphics/graphics-device.js
+++ b/src/platform/graphics/graphics-device.js
@@ -360,6 +360,7 @@ class GraphicsDevice extends EventHandler {
      * The current client rect.
      *
      * @type {{ width: number, height: number }}
+     * @ignore
      */
     clientRect = {
         width: 0,
@@ -762,15 +763,14 @@ class GraphicsDevice extends EventHandler {
     }
 
     updateClientRect() {
-        if (platform.browser) {
-            const rect = this.canvas.getBoundingClientRect();
-            this.clientRect.width = rect.width;
-            this.clientRect.height = rect.height;
-
-        } else if (platform.worker) {
+        if (platform.worker) {
             // Web Workers don't do page layout, so getBoundingClientRect is not available
             this.clientRect.width = this.canvas.width;
             this.clientRect.height = this.canvas.height;
+        } else {
+            const rect = this.canvas.getBoundingClientRect();
+            this.clientRect.width = rect.width;
+            this.clientRect.height = rect.height;
         }
     }
 
--- a/src/platform/graphics/graphics-device.js
+++ b/src/platform/graphics/graphics-device.js
@@ -356,6 +356,16 @@ class GraphicsDevice extends EventHandler {
         flags: CLEARFLAG_COLOR | CLEARFLAG_DEPTH
     };
 
+    /**
+     * The current client rect.
+     *
+     * @type {{ width: number, height: number }}
+     */
+    clientRect = {
+        width: 0,
+        height: 0
+    };
+
     static EVENT_RESIZE = 'resizecanvas';
 
     constructor(canvas, options) {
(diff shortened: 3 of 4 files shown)


src/core/platform.js
```diff
@@ -87,6 +87,13 @@ const platform = {
      */
     browser: environment === 'browser',
 
+    /**
+     * Convenience boolean indicating whether we're running in a web worker
```
willeastcott, line 91:
```suggestion
     * Convenience boolean indicating whether we're running in a Web Worker.
```

src/framework/application.js
```diff
@@ -92,7 +92,7 @@ class Application extends AppBase {
     /**
      * Create a new Application instance.
      *
-     * @param {HTMLCanvasElement} canvas - The canvas element.
+     * @param {HTMLCanvasElement | OffscreenCanvas } canvas - The canvas element.
```
willeastcott, line 95:
```suggestion
     * @param {HTMLCanvasElement | OffscreenCanvas} canvas - The canvas element.
```

src/platform/graphics/graphics-device.js
```diff
@@ -444,7 +444,7 @@ class GraphicsDevice extends EventHandler {
     }
 
     updateClientRect() {
-        this.clientRect = this.canvas.getBoundingClientRect();
+        this.clientRect = new DOMRect(0, 0, this.canvas.width, this.canvas.height);
```
willeastcott, line 447: Couple of issues here:

1. `DOMRect` isn't available in Node.js so this causes the unit tests to fail.
2. This function is called every frame so this will be generating garbage. Probably not a significant amount - just pointing it out.

src/core/platform.js
michaelfranzl: fixed

src/framework/application.js
michaelfranzl: fixed

src/core/platform.js
```diff
@@ -87,6 +87,13 @@ const platform = {
      */
     browser: environment === 'browser',
 
+    /**
+     * Convenience boolean indicating whether we're running in a Web Worker
```
willeastcott, line 91:
```suggestion
     * Convenience boolean indicating whether we're running in a Web Worker.
```

src/platform/graphics/graphics-device.js
```diff
@@ -444,7 +454,8 @@ class GraphicsDevice extends EventHandler {
     }
 
     updateClientRect() {
-        this.clientRect = this.canvas.getBoundingClientRect();
+        this.clientRect.width = this.canvas.width;
+        this.clientRect.height = this.canvas.height;
```
willeastcott, line 458: My guess is that the `canvas.width` and `canvas.height` properties are the pixel resolution of the canvas, whereas `getBoundingClientRect` would return the CSS width and height. So I would imagine `CameraComponent#worldToScreen` and `CameraComponent#screenToWorld` would break. Maybe needs `this.canvas.style.width` instead? So `parseInt(this.canvas.style.width, 10)`. But I'm not sure if that would work in all circumstances...

michaelfranzl: Unfortunately it would not work. Because Web Workers don't do the page layout, they are unaware of `el.getBoundingClientRect()`, `el.style`, `getComputedStyle(el)`. Only the main/rendering thread knows about layout dimensions and coordinates of mouse button clicks.

Therefore I can think right now of the following independent options:

1. Change `CameraComponent#screenToWorld` and `CameraComponent#worldToScreen` to accept normalized coordinates for the mouse clicks. i.e. let the user do the division of the mouse click coordinates by the canvas screen dimensions, of which they have full knowledge.
2. Pass the canvas screen dimensions into the `Application` constructor.
3. Leave the implementation as it is and require that `OffscreenCanvas` not be scaled by CSS when `CameraComponent#screenToWorld` and `CameraComponent#worldToScreen` are to be used.

I can't judge what would be suitable for the project, but option 3 would be least amount of work while still being correct for the general use case.

issacclee: option 3 seems to be the best fit for now. also, may I enquiry if there's anything else prevent this from being merged to main?

michaelfranzl: @willeastcott Is there still interest in merging this PR? I do need this feature and I think it would be generally useful.

I rebased the PR it on _main_ and addressed option 3 by doing a conditional calculation of the _clientRect_ to not break other existing features.

issacclee: definitely would like to this get merged

src/core/platform.js
```diff
@@ -17,8 +17,10 @@ const detectPassiveEvents = () => {
 };
 
 const ua = (typeof navigator !== 'undefined') ? navigator.userAgent : '';
-const environment = typeof window !== 'undefined' ? 'browser' :
-    typeof global !== 'undefined' ? 'node' : 'worker';
+const environment =
+    typeof window !== 'undefined' ? 'browser' :
+        typeof WorkerGlobalScope !== 'undefined' ? 'webworker' :
+            typeof global !== 'undefined' ? 'node' : 'worker';
```
willeastcott, line 23: So.....what's the difference between 'worker' and 'webworker'?

michaelfranzl: `worker` was defined in #6108 by @marklundin:

> When a user script is parsed in the editor, it's executed in a Worker which doesn't have window or global.

When reading this, it was not clear to me that a Web Worker was meant by "Worker"; it's also coded as a fallback, and not coded explicitly like with `typeof WorkerGlobalScope`. So to be explicit, I introduced `webworker`. But, reading the above definition again, yes, it sounds like a the two are the same and can be unified. If there are no objections, I'll unify them.

In any case, #6108 only introduces `worker` but does not do anything with that value. It looks like #6108 would not have been created if this PR here was already merged.

michaelfranzl: @willeastcott Please re-review.

willeastcott (approved):

willeastcott:
Thanks for the contribution! I removed `platform.worker` from the public API though because it's uncomfortably close to `platform.workers`.

willeastcott:
I've also cherry-picked this to `main_v1` so it will be included in the next minor release of 1.x.

PR #5172, opened 2023-03-20
[Anim Component] Unloaded animation assignment fix

ellthompson:
Fix an issue when assigning unloaded animation assets into the anim component. It will now load any unloaded assets before assigning them to the appropriate state. Any AnimClips from states that have been assigned should also be updated with the new AnimTrack.

### DIFF
--- a/src/framework/components/anim/system.js
+++ b/src/framework/components/anim/system.js
@@ -53,6 +53,7 @@ class AnimComponentSystem extends ComponentSystem {
                     layer._controller._states[stateKey]._animationList.forEach((node) => {
                         if (!node.animTrack || node.animTrack === AnimTrack.EMPTY) {
                             const animationAsset = this.app.assets.get(layer._component._animationAssets[layer.name + ':' + node.name].asset);
+                            // If there is an animation asset that hasn't been loaded, assign it once it has loaded. If it is already loaded it will be assigned already.
                             if (animationAsset && !animationAsset.loaded) {
                                 animationAsset.once('load', () => {
                                     component.layers[i].assignAnimation(node.name, animationAsset.resource);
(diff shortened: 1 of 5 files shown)


src/framework/components/anim/system.js
```diff
@@ -50,7 +51,16 @@ class AnimComponentSystem extends ComponentSystem {
             data.layers.forEach((layer, i) => {
                 layer._controller.states.forEach((stateKey) => {
                     layer._controller._states[stateKey]._animationList.forEach((node) => {
-                        component.layers[i].assignAnimation(node.name, node.animTrack);
+                        if (!node.animTrack || node.animTrack === AnimTrack.EMPTY) {
+                            const animationAsset = pc.app.assets.get(layer._component._animationAssets[layer.name + ':' + node.name].asset);
```
yaustar, line 55:
```suggestion
                            const animationAsset = this.app.assets.get(layer._component._animationAssets[layer.name + ':' + node.name].asset);
```

src/framework/components/anim/system.js
```diff
@@ -50,7 +51,16 @@ class AnimComponentSystem extends ComponentSystem {
...
+                            if (animationAsset && !animationAsset.loaded) {
```
yaustar, line 56: What if the animation asset is loaded for some reason? Should we be calling `component.layers[i].assignAnimation(node.name, animationAsset.resource);` or will that be happening anyway on the load?

ellthompson: If the animationAsset has been loaded the node will already contain a valid animTrack so this wont be reached.

yaustar: Might be worth adding a comment on why we don't have the other path in the code

yaustar:
@mvaligursky Can you have final review. It seems fine from me.

mvaligursky (approved):

PR #5179, opened 2023-03-21
Add `RaycastResult` sorting on `RigidBodyComponentSystem#raycastAll`

### LINKED ISSUE #5162: Get raycastAll results in the correct order
Currently the `raycastAll` method returns the results not in the correct order. While not everyone needs this functionality, I think it is a basic functionality that you would expect from a line going from point A to point B. (On the forum I have also had conversations with other users about this lack).

MushAsterion:
Fixes #5162 

Implements `RaycastResult` distance sorting for `RigidBodyComponentSystem#raycastAll`. As API was never documented as having the results sorted by any manner I don't think that would count as breaking. I also exposed `hitFraction` which was used for the sorting and is a very useful information to know where the hit occurred.

Changes to `RaycastResult` public API:
- \+ `hitFraction`: The fraction the hit happened. Can be used to know which one came first based on distance.

### DIFF
--- a/src/framework/components/rigid-body/system.js
+++ b/src/framework/components/rigid-body/system.js
@@ -23,9 +23,10 @@ class RaycastResult {
      * @param {import('../../entity.js').Entity} entity - The entity that was hit.
      * @param {Vec3} point - The point at which the ray hit the entity in world space.
      * @param {Vec3} normal - The normal vector of the surface where the ray hit in world space.
+     * @param {number} hitFraction - The distance percentage at which the collision occurred from starting point.
      * @hideconstructor
      */
-    constructor(entity, point, normal) {
+    constructor(entity, point, normal, hitFraction) {
         /**
          * The entity that was hit.
          *
--- a/src/framework/components/rigid-body/system.js
+++ b/src/framework/components/rigid-body/system.js
@@ -525,7 +525,8 @@ class RigidBodyComponentSystem extends ComponentSystem {
      *
      * @param {Vec3} start - The world space point where the ray starts.
      * @param {Vec3} end - The world space point where the ray ends.
-     * @returns {RaycastResult[]} An array of raycast hit results (0 length if there were no hits). Results are sorted by distance with closest first.
+     * @returns {RaycastResult[]} An array of raycast hit results (0 length if there were no hits).
+     * Results are sorted by distance with closest first.
      */
     raycastAll(start, end) {
         Debug.assert(Ammo.AllHitsRayResultCallback, 'pc.RigidBodyComponentSystem#raycastAll: Your version of ammo.js does not expose Ammo.AllHitsRayResultCallback. Update it to latest.');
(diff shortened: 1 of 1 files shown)


src/framework/components/rigid-body/system.js
```diff
@@ -46,6 +47,13 @@ class RaycastResult {
          * @type {Vec3}
          */
         this.normal = normal;
+
+        /**
+         * The distance percentage at which the collision occurred from starting point.
```
willeastcott, line 52: When you say 'percentage', is this a normalized value (0..1) or is it in a different range like 0.100?

src/framework/components/rigid-body/system.js
```diff
@@ -540,11 +550,15 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
-                        new Vec3(normal.x(), normal.y(), normal.z())
+                        new Vec3(normal.x(), normal.y(), normal.z()),
+                        hitFractions.at(i)
                     );
+
                     results.push(result);
                 }
             }
+
+            results.sort((a, b) => a.hitFraction - b.hitFraction);
```
willeastcott, line 561: Have you verified that `rayCallback.get_m_collisionObjects` is not already sorted?

src/framework/components/rigid-body/system.js
MushAsterion: It is normalized. Between 0 and 1.

src/framework/components/rigid-body/system.js
MushAsterion: Yep, in fact it seems to be reverse-sorted but considering how many possibilities there are in a 3D space I can't 100% sure confirm it is reverse-sorted. All I can confirm is that it's not sorted. If you want to test it out on a basic app with `raycastAll(new pc.Vec3(0, 2, 2), new pc.Vec3(0, -2, -2))` you will hit box then plane while it goes through the box before the plane.

src/framework/components/rigid-body/system.js
willeastcott: Cool. Can you update the JSDoc to mention that please?

MushAsterion: It's updated!

willeastcott (approved):
I've made a little tweak to the docs. Approving. I'll get one other review organized and then we can merge! 😄

mvaligursky (approved):
Looks great to me, but now that the results are sorted, we should document that somewhere so that users can depend on it. Otherwise we're sorting for no reason.

MushAsterion:
> Looks great to me, but now that the results are sorted, we should document that somewhere so that users can depend on it. Otherwise we're sorting for no reason.

Done, thank you!

willeastcott:
Nice! 🙌

LeXXik:
Just noticed this. but do we have an option not to sort it? I mean its an extra step that not always needed.

MushAsterion:
Nope, but is it really a big deal? I was actually working at the same time on #5180 which is adding a new parameter to the function so I wanted to keep it quite short and simple rather than adding an extra parameter for this.

LeXXik:
I would prefer to go for the simplest and most performant as a base, and any additional functionality that adds a cost would be optional. I do not see a justification to have a sorting enabled by default. We pass additional options to all sorts of methods across the engine, it should be fine to do so here as well. Can it not be part of the options object that adds filtering?

MushAsterion:
Makes sense, there you go then: #5181

yaustar:
I'm with LeXXik, sorting isn't always desirable so now there is extra unwanted cost to raycastAll

PR #5180, opened 2023-03-21
Implements results filtering for `raycastAll` and `raycastFirst`

### LINKED ISSUE #5143: Allow filtering during raycasts
This is a feature request.

Ammo allows to add filters during raycasts. At the moment the engine doesn't allow it.

We should add optional attribute to raycast and raycastAll that would specify the filtering. If not provided, the default behavior is taken. 

Preferably, we allow a predefined filter, e.g. static only, or dynamic only, all but kinematic, etc. 

Custom filter must be supported, in case the user is using own groups and masks.

### LINKED ISSUE #5161: Add raycastFirstByTag extension as API
In the past, Steven created a `raycastFirstByTag` extension. I still use this extension in all sorts of projects. I wonder why this isn't an official API yet since using raycasting and tags is widely used.

https://developer.playcanvas.com/en/tutorials/physics-raycasting-by-tag/

I have modified the extension myself to allow for multiple tags. 

I would like to have this extension (with the above personal adjustment) more easily accessible for all users.

MushAsterion:
Fixes #5143 
Fixed #5161 

- [ ] Requires #5179 to be merged first as `raycastFirst` entity & custom filtering requires `raycastAll` results to be sorted.

Implements filtering for `raycastAll` and `raycastFirst` through a new `filters` parameter. It includes filtering from collision group, collision mask, entity tags and custom function.

Changes to `RigidBodyComponentSystem` public API:
- `raycastFirst(start, end, filters)`
- `raycastAll(start, end, filters)`

Filters declare as follow:
```
{
    collisionGroup: 0,
    collisionMask: 0,
    tags: [ ],
    callback: (entity) => !!entity
}
```

PR set as draft due to #5179 needing to be merged first to work. Ready for review.

### DIFF
--- a/src/framework/components/rigid-body/system.js
+++ b/src/framework/components/rigid-body/system.js
@@ -469,14 +469,32 @@ class RigidBodyComponentSystem extends ComponentSystem {
      *
      * @param {Vec3} start - The world space point where the ray starts.
      * @param {Vec3} end - The world space point where the ray ends.
+     * @param {object} [filters] - Filters to find more precise values.
+     * @param {number} [filters.collisionGroup] - Collision group to apply to the raycast.
+     * @param {number} [filters.collisionMask] - Collision mask to apply to the raycast.
+     * @param {any[]} [filters.tags] - Tags filters. Defines the same way as {@link Tags#has} query but within an object.
+     * @param {Function} [filters.callback] - Custom function to use to filter entities. Must return true to proceed with result. Takes one argument: the entity to evaluate.
      * @returns {RaycastResult} The result of the raycasting or null if there was no hit.
      */
     raycastFirst(start, end) {
+        // Tags and custom callback can only be performed by looking at all results.
+        if (filters?.tags || filters?.callback) {
+            return this.raycastAll(start, end, filters)[0] || null;
+        }
+
         let result = null;
 
         ammoRayStart.setValue(start.x, start.y, start.z);
         ammoRayEnd.setValue(end.x, end.y, end.z);
         const rayCallback = new Ammo.ClosestRayResultCallback(ammoRayStart, ammoRayEnd);
+
+        if (typeof filters?.collisionGroup === 'number') {
+            rayCallback.set_m_collisionFilterGroup(filters.collisionGroup);
+        }
+
+        if (typeof filters?.collisionMask === 'number') {
+            rayCallback.set_m_collisionFilterMask(filters.collisionMask);
+        }
 
         this.dynamicsWorld.rayTest(ammoRayStart, ammoRayEnd, rayCallback);
         if (rayCallback.hasHit()) {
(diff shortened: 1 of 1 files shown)


src/framework/components/rigid-body/system.js
```diff
@@ -547,7 +583,12 @@ class RigidBodyComponentSystem extends ComponentSystem {
             const numHits = collisionObjs.size();
             for (let i = 0; i < numHits; i++) {
                 const body = Ammo.castObject(collisionObjs.at(i), Ammo.btRigidBody);
-                if (body) {
+
+                if (body && body.entity) {
+                    if (options.filterTags && !body.entity.has(...options.filterTags) || options.filterCallback && !options.filterCallback(body.entity)) {
+                        continue;
+                    }
```
slimbuck, line 590: Wondering whether you meant?
```suggestion
                if (body) {
                    if (body.entity) {
                        if (options.filterTags && !body.entity.has(...options.filterTags) || options.filterCallback && !options.filterCallback(body.entity)) {
                            continue;
                        }
                    }
```

MushAsterion: No, in fact the current behaviour to return a result even if there are no entity linked does not make sense as the body that was hit is never returned, therefore the user will have information about a hit present but no information about what was actually hit.

src/framework/components/rigid-body/system.js
```diff
@@ -528,6 +550,12 @@ class RigidBodyComponentSystem extends ComponentSystem {
      * @param {object} [options] - The additional options for the raycasting.
      * @param {boolean} [options.sort] - Whether to sort raycast results based on distance with closest
      * first. Defaults to false.
+     * @param {number} [options.filterCollisionGroup] - Collision group to apply to the raycast.
+     * @param {number} [options.filterCollisionMask] - Collision mask to apply to the raycast.
+     * @param {any[]} [options.filterTags] - Tags filters. Defines the same way as {@link Tags#has}
```
willeastcott, line 555:
```suggestion
     * @param {any[]} [options.filterTags] - Tags filters. Defined the same way as a {@link Tags#has}
```

src/framework/components/rigid-body/system.js
```diff
@@ -479,19 +479,41 @@ class RigidBodyComponentSystem extends ComponentSystem {
      *
      * @param {Vec3} start - The world space point where the ray starts.
      * @param {Vec3} end - The world space point where the ray ends.
-     * @returns {RaycastResult} The result of the raycasting or null if there was no hit.
+     * @param {object} [options] - The additional options for the raycasting.
+     * @param {number} [options.filterCollisionGroup] - Collision group to apply to the raycast.
+     * @param {number} [options.filterCollisionMask] - Collision mask to apply to the raycast.
+     * @param {any[]} [options.filterTags] - Tags filters. Defines the same way as {@link Tags#has}
```
willeastcott, line 485:
```suggestion
     * @param {any[]} [options.filterTags] - Tags filters. Defined the same way as a {@link Tags#has}
```

willeastcott:
This is cool. I'm thinking that this PR could benefit from using the `@example` tag to show some different patterns for calling the raycast function and apply the different types of filters. I found it non-obvious as to how you'd write a tags-based query for example.

MushAsterion:
Thanks for your feedback @willeastcott, here's a new commit with the examples for tags filtering as well as callback filtering, including one with both to make sure users understand that it's an AND filtering and not an OR filtering when applying multiple filters.

src/framework/components/rigid-body/system.js
```diff
@@ -528,7 +551,56 @@ class RigidBodyComponentSystem extends ComponentSystem {
      * @param {object} [options] - The additional options for the raycasting.
      * @param {boolean} [options.sort] - Whether to sort raycast results based on distance with closest
      * first. Defaults to false.
+     * @param {number} [options.filterCollisionGroup] - Collision group to apply to the raycast.
+     * @param {number} [options.filterCollisionMask] - Collision mask to apply to the raycast.
+     * @param {any[]} [options.filterTags] - Tags filters. Defined the same way as a {@link Tags#has}
+     * query but within an object.
```
willeastcott, line 557: So should this be 'within an array'?

willeastcott (approved):

willeastcott:
Yay - another great addition! Thanks. 🙏

mvaligursky:
I believe this PR created an issue here: https://github.com/playcanvas/engine/issues/5367

PR #5181, opened 2023-03-21
Change `raycastAll` sorting default to `false`

MushAsterion:
As requested by @LeXXik in #5179, added an `options` parameter to `raycastAll` to allow non-sorting of results.

Changes to `RigidBodyComponentSystem` public API:
- `raycastFirst(start, end, options)`
- `raycastAll(start, end, options)`

Options declares as follow for not sorting:
```
{
    sort: false
}
```

### DIFF
--- a/src/framework/components/rigid-body/system.js
+++ b/src/framework/components/rigid-body/system.js
@@ -563,7 +563,7 @@ class RigidBodyComponentSystem extends ComponentSystem {
                 }
             }
 
-            if (options.sort === true) {
+            if (options.sort) {
                 results.sort((a, b) => a.hitFraction - b.hitFraction);
             }
         }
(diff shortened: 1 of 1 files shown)


LeXXik:
Hey, thank you for this one! However, I would like to reiterate, that my view on this still stands as the default option (when no options specified) should be non-sorted result. And if a developer wants a sorting, it would be enabled via options or do own sorting on the results that would match their game logic.

MushAsterion:
Then it becomes breaking change, but there you go. Placing #5180 back on draft as if this change gets merged the code will need to be updated.

LeXXik:
Thank you! I don't think it is a breaking change, unless it changes the code that was already released. I will leave it up to @willeastcott to decide.

mvaligursky (approved):

MushAsterion:
Oh alright, I was referring at the current folder state but yeah that makes sense if that's from the released state!

willeastcott (approved):

PR #5183, opened 2023-03-22
Clamp shadow distance to camera far plane

slimbuck:
Shadow cascades weren't taking the camera's far clipping plane into consideration when generating cascades.

This PR clamps shadow distance so cascades are more optimally placed.

With optimal near and far camera clipping planes, shadow resolution is improved:

**Before:**
Screenshot 2023-03-22 at 13 10 44

**After:**
Screenshot 2023-03-22 at 13 11 21


mvaligursky (approved):
Awesome, this works around the case where the user would not set up shadowDistance correctly.

willeastcott (approved):

PR #5186, opened 2023-03-23
Simple planar-renderer script to render planar reflections

mvaligursky:
- PlanarRenderer script, which handles mirroring by plane and rendering to texture
- made Plane class public
- renamed couple of examples to REFLECTION*** pattern to keep those grouped.

TODO:
- integrate the result to standard material, allowing normal maps, roughness ... (separate PR)

Screenshot 2023-03-23 at 15 44 37


src/core/shape/plane.js
```diff
@@ -39,6 +37,16 @@ class Plane {
         }
     }
 
+    /**
+     * The d term for a scalar equation of the form ax + by + cz + d = 0, where a, b and c
+     * are the x, y and z components of the plane's normal.
+     *
+     * @type {number}
+     */
+    get d() {
```
slimbuck, line 46: shouldn't we fix plane before making it public api? why on earth it stores a normal and point is beyond me!

mvaligursky: yep, that makes sense and done. Larger PR, but hey.

src/framework/components/element/element-drag-helper.js
```diff
@@ -158,22 +160,13 @@ class ElementDragHelper extends EventHandler {
...
-
-            _entityRotation.copy(this._element.entity.getRotation()).invert().transformVector(position, position);
-
-            position.mul(this._dragScale);
-
-            return position;
+        if (_plane.intersectsRay(_ray, _point)) {
+            _entityRotation.copy(this._element.entity.getRotation()).invert().transformVector(_point, _point);
+            _point.mul(this._dragScale);
+            return _point;
```
mvaligursky, line 169: I tested this change in the engine example: scroll-view

src/core/shape/plane.js
```diff
@@ -109,7 +107,8 @@ class Plane {
      * @returns {this} A duplicate Plane.
      */
     clone() {
-        return new this.constructor(this.point, this.normal);
+        const clone = new this.constructor();
+        return clone.copy(this);
```
willeastcott, line 111:
```suggestion
        /** @type {this} */
        const cstr = this.constructor;
        return new cstr().copy(this);
```

willeastcott: This seems to be the format for `clone` across the engine.

mvaligursky: hah, I copied from DepthState ;)

GSterbrant:
Why not make the example default to using a 100% scaled reflection buffer? Looks less aliased that way 🙂.

GSterbrant (approved):

mvaligursky:
> Why not make the example default to using a 100% scaled reflection buffer? Looks less aliased that way 🙂.

Sure, done.

src/core/math/mat4.js
```diff
@@ -778,6 +778,42 @@ class Mat4 {
...
     }
 
+    /**
+     * Sets the matrix to a reflection matrix, which can be used as a mirror transformation by the
+     * plane.
+     *
+     * @param {import('./plane.js').Plane} plane - The plane to reflect by.
+     * @returns {Mat4} Self for chaining.
+     */
+    setReflection(plane) {
```
willeastcott, line 788: I'm a bit late to the party here, but this is introducing a dependency of the math API on the shape API. I'm not a big fan of that.

src/core/shape/plane.js
```diff
@@ -1,54 +1,48 @@
...
+     * Defaults to {@link pc.Vec3#ZERO}.
      */
-    constructor(point, normal) {
-        if (point) {
-            this.point.copy(point);
-        }
-        if (normal) {
-            this.normal.copy(normal);
-        }
+    constructor(normal = Vec3.UP, point = Vec3.ZERO) {
```
willeastcott, line 32: Should we have changed the constructor to take a normal and d?

src/core/math/mat4.js
mvaligursky: Good point.
I feel like moving Plane to Math. I even assumed it's there.

src/core/shape/plane.js
```diff
@@ -1,54 +1,48 @@
...
+     * Sets the plane to use a specified normal and pass through a specified point.
      *
-     * @param {Vec3} point - The starting point of the plane.
      * @param {Vec3} normal - The normal of the plane.
+     * @param {Vec3} point - The point on a plane.
      * @returns {Plane} Self for chaining.
      */
-    set(point, normal) {
-        this.point.copy(point);
+    set(normal, point) {
```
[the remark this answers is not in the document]
mvaligursky: yep, same conclusion here. On it.

willeastcott:
Wondering whether `Frustum` should be using this new `Plane` class...

src/core/shape/plane.js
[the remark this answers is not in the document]
mvaligursky: done here https://github.com/playcanvas/engine/pull/5196

PR #5191, opened 2023-03-26
Add static math methods to `Mat4`, `Quat`, `Vec2`, `Vec3` and `Vec4`

MushAsterion:
Added static functions for math calculations to make the code simpler. It can be useful to be able to make calculations without modifying an existing value (even if temporary variables remain the more efficient) as well as ease code import from other engines.

Example of current state with a `Vec3`:
```
const a = new pc.Vec3(4, 9, 16);
const b = new pc.Vec3(2, 3, 4);

const r = new pc.Vec3().add2(a, b);
```

New state:
```
const a = new pc.Vec3(4, 9, 16);
const b = new pc.Vec3(2, 3, 4);

const r = pc.Vec3.add(a, b);
```

Changes to `Mat4` public API:
- static `mul(lhs, rhs)`
- static `mulAffine(lhs, rhs)`

Changes to `Quat` public API:
- static `mul(lhs, rhs)`
- static `slerp(lhs, rhs, alpha)`

Changes to `Vec2` / `Vec3` / `Vec4` public API:
- static `add(lhs, rhs)`
- static `addScalar(rhs, scalar)`
- static `div(lhs, rhs)`
- static `divScalar(rhs, scalar)`
- static `lerp(lhs, rhs, alpha)`
- static `mul(lhs, rhs)`
- static `mulScalar(rhs, scalar)`
- static `sub(lhs, rhs)`
- static `subScalar(rhs, scalar)`

### DIFF
--- a/src/core/math/quat.js
+++ b/src/core/math/quat.js
@@ -80,15 +80,6 @@ class Quat {
         return this;
     }
 
-    static conjugate(rhs) {
-        return new Quat(
-            rhs.x * -1,
-            rhs.y * -1,
-            rhs.z * -1,
-            rhs.w
-        );
-    }
-
     /**
      * Copies the contents of a source quaternion to a destination quaternion.
      *
--- a/src/core/math/quat.js
+++ b/src/core/math/quat.js
@@ -80,6 +80,15 @@ class Quat {
         return this;
     }
 
+    static conjugate(rhs) {
+        return new Quat(
+            rhs.x * -1,
+            rhs.y * -1,
+            rhs.z * -1,
+            rhs.w
+        );
+    }
+
     /**
      * Copies the contents of a source quaternion to a destination quaternion.
      *
@@ -281,6 +290,41 @@ class Quat {
         this.w = q1w * q2w - q1x * q2x - q1y * q2y - q1z * q2z;
 
         return this;
+    }
+
+    /**
+     * Returns the result of multiplying the specified quaternions together.
+     *
+     * @param {Quat} lhs - The quaternion used as the first multiplicand of the operation.
+     * @param {Quat} rhs - The quaternion used as the second multiplicand of the operation.
+     * @returns {Quat} The resulting quaternion.
+     * @example
+     * var a = new pc.Quat().setFromEulerAngles(0, 30, 0);
+     * var b = new pc.Quat().setFromEulerAngles(0, 60, 0);
+     *
+     * // r is set to a 90 degree rotation around the Y axis
+     * // In other words, r = a * b
+     * const r = pc.Quat.mul(a, b);
+     *
+     * console.log("The result of the multiplication is: " + r.toString());
+     */
+    static mul(lhs, rhs) {
+        const q1x = lhs.x;
+        const q1y = lhs.y;
+        const q1z = lhs.z;
+        const q1w = lhs.w;
+
+        const q2x = rhs.x;
+        const q2y = rhs.y;
+        const q2z = rhs.z;
+        const q2w = rhs.w;
+
+        return new Quat(
+            q1w * q2x + q1x * q2w + q1y * q2z - q1z * q2y,
+            q1w * q2y + q1y * q2w + q1z * q2x - q1x * q2z,
+            q1w * q2z + q1z * q2w + q1x * q2y - q1y * q2x,
+            q1w * q2w - q1x * q2x - q1y * q2y - q1z * q2z
+        );
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/core/math/quat.js
```diff
@@ -667,6 +702,70 @@ class Quat {
...
+        // we could rotate around any axis normal to qa or qb
+        if (Math.abs(sinHalfTheta) < 0.001) {
+            return new Quat(lx * 0.5 + rx * 0.5, ly * 0.5 + ry * 0.5, lz * 0.5 + rz * 0.5, lw * 0.5 + rw * 0.5);
+        }
+
+        const ratioA = Math.sin((1 - alpha) * halfTheta) / sinHalfTheta;
+        const ratioB = Math.sin(alpha * halfTheta) / sinHalfTheta;
+
+        // Calculate Quaternion.
+        return new Quat(lx * ratioA + rx * ratioB, ly * ratioA + ry * ratioB, lz * ratioA + rz * ratioB, lw * ratioA + rw * ratioB);
```
willeastcott, line 766: Why isn't this just `return new Quat().slerp(lhs, rhs, alpha);`?

src/core/math/quat.js
```diff
@@ -283,6 +283,41 @@ class Quat {
...
+        const q2y = rhs.y;
+        const q2z = rhs.z;
+        const q2w = rhs.w;
+
+        return new Quat(
+            q1w * q2x + q1x * q2w + q1y * q2z - q1z * q2y,
+            q1w * q2y + q1y * q2w + q1z * q2x - q1x * q2z,
+            q1w * q2z + q1z * q2w + q1x * q2y - q1y * q2x,
+            q1w * q2w - q1x * q2x - q1y * q2y - q1z * q2z
+        );
```
willeastcott, line 318: Why not just `return new Quat().mul(lhs, rhs);`?

willeastcott, line 318: Wondering if adding `Quat#mul2` and using that here would make the API more consistent with vectors/matrices.

src/core/math/quat.js
MushAsterion: I wanted to make less functions call as possible, this way it's just `new` instead of `new` + `slerp` but yeah that's true that it makes the function really long for same result, I can change it if you prefer

src/core/math/quat.js
MushAsterion: `Quat#mul2` already exists, same as the previous review I wanted to make it only call `new` instead of `new` + `mul2` but I can change if you prefer

willeastcott:
I want to add some historical context. In the past, vectors and matrices stored their components in a `Float32Array`. Creating these was (and still is, I believe) very slow. So the math APIs were specifically designed to never allocate vectors/matrices internally. The user was forced to proactively allocate vectors/matrices and were encouraged to allocate them up-front rather than every frame.

Today, vectors just used regular numbers for components. But matrices still use `Float32Array`. So vectors are now much faster to allocate - but matrices should still be quite slow. So maybe its a bit less of an issue for vectors these days.

But still, we do want to avoid PlayCanvas apps generating per-frame garbage that could trigger stalls when the garbage collector runs.

So pros for this PR:

* Makes app code nicer in some circumstances.

Cons:

* Arguably bloats the API.
* Increases the amount of garbage generated by apps.
* Potential performance implications of allocating math objects 'behind the scenes'.

To be honest, I think we should update our contribution guidelines. I think the process should be:

1. Decide to add new feature to the engine.
2. Open an issue and outline the feature you're thinking of adding.
3. Have conversation with the community to get agreement/alignment and understand history/context.
4. If there's consensus/agreement, start implementation and submit PR.
5. Review + merge as usual.

willeastcott:
A side thought: I'm wondering whether it makes sense for `Mat3`/`Mat4` to use a vanilla array of numbers these days. We'd get better precision, faster allocation....and maybe performance in modern browsers is acceptable (or maybe even better).

MushAsterion:
Very honestly when I was working on this PR I knew it might not be well welcomed or might just get refused and that's fine as it's basically a not-so-good practice however I still wanted to make it just in case as personally when I started playing around with Playcanvas I was not optimizing anything and the `add2`/`div2`/`mul2`/`sub2` were still kind of a unclear scenario.

But even with that I wanted to show off with a different vision not based on optimization (as not everyone is optimizing their projects, which is a bit sad for low-end devices) but on readability. On most of the code I've read from examples and some open projects I have not seen a lot with an allocated vector being used, it's rather directly the initial vector or a clone and I wanted to make that practice a bit easier.

When it comes to create an issue you're completely right, I should have done that as I guess more people read there than in PR...

kungfooman:
I somewhat know where you are coming from, since I did the same when I had to port a bunch of C# code once... basically my motivation was to come closer towards operator overloading, as it exists in C#, so it would be easier to create a first approximation towards C# to "make it work" and the later you can concentrate on the optimisation part.

I ended up using a style like this:

```js
/**
 * @example
 * mul(new pc.Quat().setFromEulerAngles(10, 20, 30), new pc.Quat().setFromEulerAngles(1, 2, 3))
 * mul(new pc.Vec3(10, 11, 12), new pc.Vec3(1, 2, 3))
 * @param {T} a 
 * @param {T} b 
 * @template {{clone(): T, mul(T): T}} T
 * @returns {T}
 */
export function mul(a, b) {
  return a.clone().mul(b);
}

const x = mul(
  new pc.Quat().setFromEulerAngles(10, 20, 30),
  new pc.Quat().setFromEulerAngles(1, 2, 3)
);

const y = mul(
  new pc.Vec3(10, 11, 12),
  new pc.Vec3(1, 2, 3)
);
```

Which, arguably, is also the shorest, since the types imply themselves and the types just follow aswell:

image

So basically I share the same sentiment as @willeastcott here, this feels like API bloat, even though I understand the motivation. I wish JS would finally support operator overloading for these kinds of things. QuickJS already supports it.

PR #5196, opened 2023-03-27
Updates to Plane’s class constructor, to take a distance instead of a point

mvaligursky:
based on a feedback.

### DIFF
--- a/src/core/shape/plane.js
+++ b/src/core/shape/plane.js
@@ -37,7 +37,7 @@ class Plane {
     /**
      * Sets the plane based on a specified normal and a point on the plane.
      *
-     * @param {Vec3} point - The point on a plane.
+     * @param {Vec3} point - The point on the plane.
      * @param {Vec3} normal - The normal of the plane.
      * @returns {Plane} Self for chaining.
      */
--- a/src/core/shape/plane.js
+++ b/src/core/shape/plane.js
@@ -14,7 +14,7 @@ class Plane {
     normal = new Vec3();
 
     /**
-     * The distance from the plan to the origin, along its normal.
+     * The distance from the plane to the origin, along its normal.
      *
      * @readonly
      * @type {number}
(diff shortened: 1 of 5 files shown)


src/core/shape/plane.js
```diff
@@ -26,21 +26,22 @@ class Plane {
...
-     * @param {Vec3} [point] - Point position on the plane. The constructor copies this parameter.
-     * Defaults to {@link pc.Vec3#ZERO}.
+     * @param {Vec3} [distance] - The distance from the plane to the origin, along its normal.
+     * Defaults to 0.
      */
-    constructor(normal = Vec3.UP, point = Vec3.ZERO) {
-        this.set(normal, point);
+    constructor(normal = Vec3.UP, distance = 0) {
+        this.normal.copy(normal);
+        this.distance = distance;
```
willeastcott, line 34: Maybe declare this above the constructor too?

src/core/shape/plane.js
```diff
@@ -26,21 +26,22 @@ class Plane {
...
     }
 
     /**
-     * Sets the plane to use a specified normal and pass through a specified point.
+     * Sets the plane based on  a specified normal and a point on a plane.
```
willeastcott, line 38:
```suggestion
     * Sets the plane based on a specified normal and a point on the plane.
```

src/core/shape/plane.js
mvaligursky: the member is there, but not initialized, to avoid initializing it twice - is that what you meant?

src/core/shape/plane.js
```diff
@@ -26,21 +26,22 @@ class Plane {
      *
      * @param {Vec3} [normal] - Normal of the plane. The constructor copies this parameter. Defaults
      * to {@link pc.Vec3#UP}.
```
willeastcott, line 28: You don't need `pc.` here.

src/core/shape/plane.js
```diff
@@ -26,21 +26,22 @@ class Plane {
...
+        this.normal.copy(normal);
+        this.distance = distance;
     }
 
     /**
-     * Sets the plane to use a specified normal and pass through a specified point.
+     * Sets the plane based on  a specified normal and a point on a plane.
      *
-     * @param {Vec3} normal - The normal of the plane.
      * @param {Vec3} point - The point on a plane.
```
willeastcott, line 40: a -> the

willeastcott (approved):

lucaheft:
What is the intention behind changing this and why is this not marked as breaking change?
Creating a plane by normal and point wont work after this. Which will break old code.
Also I dont get why this needs a change. A plane is defined by a normal and a point in space. Why define it by normal and distance from Origin?

mvaligursky:
Fair enough, I marked this PR as breaking, adding the workaround in the description, and added it to a Breaking section of the release notes https://github.com/playcanvas/engine/releases/tag/v1.62.0

LeXXik:
Yeah, I use planes everywhere as well to test intersections :) Didn't voice it, since it was a private API change, though.

PR #5205, opened 2023-03-28
Fix lightmapping with 1.62

GSterbrant:
- Writes to dAtten, dAttenD and dAtten3 from clusteredLighting.js for any shaders that might need those values later (such as the lightmapper).
- Remove redundant passing of the half vector to the phong specular function.
- Fixed calls to addLightMap to correctly include the iridescence fresnel.

### DIFF
--- a/src/scene/shader-lib/chunks/lit/frag/clusteredLight.js
+++ b/src/scene/shader-lib/chunks/lit/frag/clusteredLight.js
@@ -593,6 +593,8 @@ void evaluateLight(
             #endif
         }
     }
+
+    // Write to global attenuation values (for lightmapper)
     dAtten = falloffAttenuation;
     dAttenD = diffuseAttenuation;
     dAtten3 = cookieAttenuation;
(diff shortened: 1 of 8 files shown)


willeastcott (approved):

mvaligursky:
Please make sure lightmappeer works with non-clustered lighting as well.

src/scene/shader-lib/chunks/lit/frag/clusteredLight.js
```diff
@@ -593,6 +593,9 @@ void evaluateLight(
             #endif
         }
     }
+    dAtten = falloffAttenuation;
```
mvaligursky, line 596: add a comment for this pls

GSterbrant: Done.

mvaligursky (approved):

GSterbrant:
Tested light mapper with the baked AO lights w/o clustered lighting. Works.

PR #5208, opened 2023-03-29
Async draco decompression

slimbuck:
This PR moves gltf draco decompression from the main thread onto a configurable set of workers.

A new function called `dracoInitialize` is added, which is used as follows:
```javascript
dracoInitialize({
    jsUrl: 'draco.wasm.js',
    wasmUrl: 'draco.wasm.wasm',
    numWorkers: 2,
    lazyInit: true
});
```

It's also possible to set configuration via the `WasmModule` class.

A followup PR will also load buffers and textures asynchronously.

### Performance

Below are total load times for draco compressed models. The load time includes creating textures and animations and the timings were averaged over 7 runs.

**6.6MB glb vehicle model containing 2MB draco compressed meshes:**

| Platform |  Main | 1 Worker | 2 Workers | 4 Workers |
|-|-|-|-|-|
| Chrome MacOS | 610 | 640 | 520 | 460 |
| Chrome Pixel 3a | 4430 | 2950 | 3040 | 2820 |
| Safari MacOS | 2010 | 2010 | 1940 | 1980 |
| Safari iPhone 8 | 5300 | 5400 |  5270 | 5770 |

**2.3MB backpack model containing 1 draco compressed mesh totalling 38k**

| Platform |  Main | 1 Worker | 2 Workers | 4 Workers |
|-|-|-|-|-|
| Chrome MacOS | 220 | 230 | - | - |
| Chrome Pixel 3a | 740 |  720 | - | - |
| Safari MacOS | 470 |  480 | - | - |
| Safari iPhone 8 |  1270 |  1190 | - | - |


mvaligursky:
Please mention some performance data if possible, it'd be great to see.

src/core/wasm-module.js
```diff
@@ -135,6 +135,8 @@ class WasmModule {
      * @param {string} [config.wasmUrl] - URL of the wasm script.
      * @param {string} [config.fallbackUrl] - URL of the fallback script to use when wasm modules
      * aren't supported.
+     * @param {number} [config.numWorkers] - For modules running on worker threads, the number of
+     * threads to use.
```
mvaligursky, line 139: Is there a default value?

src/framework/parsers/draco-decoder.js
```diff
@@ -0,0 +1,168 @@
+import { WasmModule } from "../../core/wasm-module";
+import { ABSOLUTE_URL } from "../asset/constants";
+import { DracoWorker } from "./draco-worker";
+
+class JobQueue {
```
mvaligursky, line 5: It'd be great to have this documented.
Also, it this a generic class, or is it draco specific?

src/framework/parsers/draco-decoder.js
```diff
@@ -0,0 +1,168 @@
...
+    constructor(workers) {
+        this.workers = [workers, [], []];
```
mvaligursky, line 7: what are we storing here? Could those be separately stored arrays for clarity, and avoiding things like `this.workers[2]`?

mvaligursky: or maybe some of those should be Sets, considering you do indexOf on them

src/framework/parsers/draco-worker.js
```diff
@@ -0,0 +1,259 @@
...
+            case draco.DT_FLOAT32: return 4;
+        }
+        return 1;
+    };
+
+    const attributeSizeInBytes = (attribute) => {
+        return attribute.num_components() * componentSizeInBytes(attribute.data_type());
+    };
+
+    const attributeOrder = {
```
mvaligursky, line 37: maybe we should send this table to the worker as part of the job description, to avoid hardcoded constants?

src/framework/parsers/draco-worker.js
```diff
@@ -0,0 +1,259 @@
...
+            const src = new Uint8Array(draco.HEAPU8.buffer, ptr, ptrSize);
+
+            // pack
+            for (let j = 0; j < mesh.num_points(); ++j) {
+                for (let c = 0; c < sizeInBytes; ++c) {
+                    dst[j * totalVertexSize + offsets[i] + c] = src[j * sizeInBytes + c];
+                }
+            }
+
+            if (!hasNormals && attribute.attribute_type() === 0) {
```
mvaligursky, line 200: I'd vote for not using '0' but have that wrapped by a constant in this file somewhere.

src/framework/parsers/draco-worker.js
```diff
@@ -0,0 +1,259 @@
...
+        for (let i = 0; i < numTriangles; ++i) {
+            const v0 = indices[i * 3 + 0] * 3;
+            const v1 = indices[i * 3 + 1] * 3;
+            const v2 = indices[i * 3 + 2] * 3;
+
+            copy(a, vertices, v0);
+            copy(b, vertices, v1);
+            copy(c, vertices, v2);
+
+            subtract(t1, b, a);
```
mvaligursky, line 96: are positions here always in a format we can use math operations on?

mvaligursky:
Can we depend on workers being available on all browsers we support? (including older versions)

LeXXik:
Thank you for the performance notes! It is interesting that adding a worker would make it actually slower on some devices.

slimbuck:
> Thank you for the performance notes! It is interesting that adding a worker would make it actually slower on some devices.

I wouldn't read too much into the timings because they are quite noisy and the "slowdown" is tiny.

The big win with this change is that decoding is now happening off the main thread and once glb-parser stops stalling for buffers and images to complete loading, we should end up doing a lot more work concurrently than we currently are, meaning load times should be even better. (On devices with multiple CPU cores anyway).

src/core/wasm-module.js
slimbuck: Not really, it's up to the module implementation to decide.

src/core/wasm-module.js
```diff
@@ -147,6 +149,15 @@ class WasmModule {
         }
     }
 
+    /**
+     * Get a wasm module's configuration.
+     * 
+     * @returns {object | undefined} The previously set configuration.
+     */
```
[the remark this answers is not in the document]
slimbuck: Hmm I think it'll be useful...

src/framework/parsers/draco-decoder.js
slimbuck: This job queue is currently only used for draco, but I think i'd like to use it for basis if possible. Will probably make sense to document then.

src/framework/parsers/draco-decoder.js
slimbuck: `workers` stores 3 arrays of workers depending on how many jobs they're actively working on.

`workers[0]` contains workers with no jobs.
`workers[1]` contains workers with 1 job queued.
`workers[2]` contains workers with 2 jobs queued.

We want to keep worker order in each array so we push to the back and splice from the front of the array.

src/framework/parsers/draco-worker.js
slimbuck: these are draco constants, not engine constants unfortunately.

src/core/wasm-module.js
mvaligursky: Maybe mention that.

src/framework/parsers/draco-worker.js
slimbuck: yeah nice, I'll do that and see what else could do with labelling.

src/framework/parsers/draco-decoder.js
mvaligursky: ah I didn't guess that, I assumed [0] is free, [1] is busy, [2] is done or something like that.
We need documentation.

src/framework/parsers/draco-worker.js
slimbuck: yep, this logic hasn't changed.

src/framework/parsers/draco-decoder.js
```diff
@@ -0,0 +1,174 @@
...
+    // worker urls must be absolute
+    const absoluteUrl = (url) => {
+        return ABSOLUTE_URL.test(url) ? url : new URL(window.location.href).origin + url;
+    };
+
+    const jsUrl = absoluteUrl(config.jsUrl);
+    const wasmUrl = absoluteUrl(config.wasmUrl);
+
+    // create workers
+    const numWorkers = Math.max(1, Math.min(16, config.numWorkers || defaultNumWorkers));
```
willeastcott, line 125: So if I specified 0 workers, I'd get 1? I _guess_ that's intentional? Just wondering if it should be `??` instead of `||`.

slimbuck: yeah right, 0 should become 1

mvaligursky (approved):

PR #5219, opened 2023-03-31
Update glb-parser

slimbuck:
This PR modernises glb-parser.js by converting functions to arrow functions and replacing instances of `options && options.foo && options.foo.bar` with `options?.foo?.bar`.


mvaligursky (approved):

src/framework/parsers/glb-parser.js
```diff
@@ -1804,17 +1804,17 @@ const createScenes = function (gltf, nodes) {
...
 
     let cameras = null;
 
     if (gltf.hasOwnProperty('nodes') && gltf.hasOwnProperty('cameras') && gltf.cameras.length > 0) {
 
-        const preprocess = options && options.camera && options.camera.preprocess;
-        const process = options && options.camera && options.camera.process || createCamera;
-        const postprocess = options && options.camera && options.camera.postprocess;
+        const preprocess = options?.camera?.preprocess;
+        const process = options?.camera?.process || createCamera;
```
willeastcott, line 1814: Do we want to standardize on `??` instead of `||` for these kinds of use cases? Which means 'if the left side is null or undefined, use the right side' (rather than a logical op).

src/framework/parsers/glb-parser.js
```diff
@@ -1804,17 +1804,17 @@ const createScenes = function (gltf, nodes) {
...
+const createCameras = (gltf, nodes, options) => {
 
     let cameras = null;
 
     if (gltf.hasOwnProperty('nodes') && gltf.hasOwnProperty('cameras') && gltf.cameras.length > 0) {
 
-        const preprocess = options && options.camera && options.camera.preprocess;
-        const process = options && options.camera && options.camera.process || createCamera;
-        const postprocess = options && options.camera && options.camera.postprocess;
+        const preprocess = options?.camera?.preprocess;
```
willeastcott, line 1813: Wondering if we should prescribe a set coding style for options objects. In my code, I've been using a default parameter (empty object) for options objects. But, of course, that will allocate on every call potentially. Whereas using `?.` everywhere is a bit more verbose and might transpile to more code via Babel?

willeastcott (approved):

src/framework/parsers/glb-parser.js
slimbuck: Not sure how that would work here. Could you give a code snippet?

willeastcott: Well, you'd still need the second `?.` regardless. TBH, it's not a big deal.

src/framework/parsers/glb-parser.js
slimbuck: Yeah updating to ?? would make sense. I have a followup PR already which is based on this one and will update to ?? there (since some have been removed).

PR #5224, opened 2023-04-01
Eliminate `var` from the code base

willeastcott:
It's time. This PR eliminates almost all remaining instances of `var` in the engine codebase. These are almost all in the API ref docs. However, since even IE11 support both `let` and `const`, we should embrace these superior keywords everywhere.


slimbuck (approved):
🥳🥳🥳

PR #5232, opened 2023-04-11
Use RGBA8 for png and jpg

slimbuck:
The image loader was specifying RGB8 pixel format for JPG, but this incurs a format conversion cost when using ImageBitmap.

This PR standardises on RGBA8 instead, speeding up JPG loading in Chrome.


willeastcott (approved):
Are there memory implications we're concerned about for switching the format from RGB8 to RGBA8?

slimbuck:
Nope. GPUs store textures in 32bits anyway, never 24.

PR #5233, opened 2023-04-11
WebGL device support passing in gl context

slimbuck:
This PR adds the option of passing in the rendering context to the graphics device.

Creating the context ahead of time can speed up startup times in some applications.


willeastcott:
> Creating the context ahead of time can speed up startup times in some applications.

Can you elaborate?

slimbuck:
In most applications the main application script (or engine) is first downloaded, then the pc application instance is created and then the rendering context is initialised.

Since creating the rendering context can take some time, I have created a small "bootstrap" module for my application which creates the 2d canvas and rendering context while the main application script is being downloaded. This saves approximately 50ms at startup.

willeastcott (approved):

PR #5235, opened 2023-04-11
Disable ImageBitmap in Safari

slimbuck:
ImageBitmap only loads images faster than HTMLImageElement on Chrome.

Instead of performing an async canvas test for ImageBitmap at startup, we instead simply limit it to Chrome. The async startup test was returning too slowly in some applications and so wasn't reliable anyway.

For future testing, the alpha test functionality has been kept, though moved out of the graphics device. The test can be performed at startup in debug build by enabling the trace `IMG_ALPHA_TEST`.

### DIFF
--- a/src/framework/parsers/texture/img-alpha-test.js
+++ b/src/framework/parsers/texture/img-alpha-test.js
@@ -83,6 +83,4 @@ class ImgAlphaTest {
     }
 }
 
-export {
-    ImgAlphaTest
-};
+export { ImgAlphaTest };
--- a/src/framework/parsers/texture/img-alpha-test.js
+++ b/src/framework/parsers/texture/img-alpha-test.js
@@ -1,6 +1,6 @@
-import { Texture } from '../../../platform/graphics/texture.js';
 import { PIXELFORMAT_RGBA8 } from '../../../platform/graphics/constants.js';
 import { RenderTarget } from '../../../platform/graphics/render-target.js';
+import { Texture } from '../../../platform/graphics/texture.js';
 
 //
 // Current state of ImageBitmap (April 2023):
--- a/src/framework/parsers/texture/img-alpha-test.js
+++ b/src/framework/parsers/texture/img-alpha-test.js
@@ -1,6 +1,6 @@
-import { Texture } from '../texture.js';
-import { PIXELFORMAT_RGBA8 } from '../constants.js';
-import { RenderTarget } from '../render-target.js';
+import { Texture } from '../../../platform/graphics/texture.js';
+import { PIXELFORMAT_RGBA8 } from '../../../platform/graphics/constants.js';
+import { RenderTarget } from '../../../platform/graphics/render-target.js';
 
 //
 // Current state of ImageBitmap (April 2023):
@@ -44,7 +44,7 @@ const testAlpha = (device, image) => {
     texture.destroy();
 
     return data[0] === 1 && data[1] === 2 && data[2] === 3 && data[3] === 63;
-}
+};
 
 // Test whether ImageBitmap correctly loads PNG alpha data.
 const testImageBitmapAlpha = (device) => {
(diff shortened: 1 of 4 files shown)


src/framework/parsers/texture/img-alpha-test.js
```diff
@@ -0,0 +1,88 @@
...
+
+        testImgElementAlpha(device).then((result) => {
+            console.log(`imgElementIsCorrect=${result}`);
+        });
+    }
+}
+
+export {
+    ImgAlphaTest
+};
```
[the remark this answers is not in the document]
slimbuck: done

slimbuck:
So we don't currently use ImageBitmap on Safari at all, because the startup alpha test returns negative. However on Chrome, the startup test for ImageBitmap sometimes returns after textures have already started loading, meaning that in some cases Chrome -isn't- using ImageBitmap when it should.

Regarding your points above:
- If startup time is a priority, then faster loading times wins over main thread locking. I haven't tested thread locking performance on Safari because the profiler doesn't show this level of info, but it also doesn't matter because Safari premultiplies alpha anyway (rendering ImageBitmap mostly useless).
- We can update the test, but it currently just checks for Chrome.
- You're right, but I don't see a way to fix this. In an ideal world we'd have a test that tells us whether ImageBitmap works correctly and how it performs compared to HTMLImageElement. The test wouldn't take any time.
- Here are average timings on MacOS loading a single gltf with 6 textures in jpg & png format (first time is using HTMLImageBitmap, second is using ImageBitmap):
Safari - 550ms -> 650ms.
Firefox - 230ms -> 270ms.
Chrome - 310ms -> 280ms.
These timings include creating the device rendering the first frame.
- I thought there already was, but don't see an issue for the RGB premul issue. I can create a simple repro and report it.

src/framework/parsers/texture/img-alpha-test.js
```diff
@@ -0,0 +1,86 @@
...
+import { Texture } from '../../../platform/graphics/texture.js';
+
+//
+// Current state of ImageBitmap (April 2023):
+//
+// Chrome MacOS and Android (Pixel 3a and Pixel 7 Pro):
+//   Correctly loads PNG alpha and runs faster (significantly so on mobile) than HTMLImageElement.
+//
+// Firefox MacOS:
+//   Correctly loads PNG alpha, but runs significantly slower than HTMLImageElement.
```
willeastcott, line 12: You mentioned a re-test of this showed the performance difference was perhaps not as significant as you thought. Is this still 'accurate'?

slimbuck: Firefox timings are a little noisy, but I'm seeing 300ms vs 200ms load time on a test scene on MacOS.

src/framework/parsers/texture/img.js
```diff
@@ -20,6 +23,13 @@ class ImgParser {
         this.crossOrigin = registry.prefix ? 'anonymous' : null;
         this.maxRetries = 0;
         this.device = device;
+
+        // run image alpha test
+        Debug.call(() => {
```
mvaligursky, line 28: why do we run this in debug mode only?

slimbuck: try reading the PR description ;)

mvaligursky: Ah I see.
Perhaps add `// #if _DEBUG` around `import { ImgAlphaTest }` to exclude the code from non-debug builds.

slimbuck: The code is already excluded from non-debug builds because it's not referenced anywhere (and not exported). Might be worth adding the #if anyway to make that more obvious to readers. Thanks!

marklundin:
Just posting some new info in case it's relevant. It seems the [`ImageDecoder`](https://developer.mozilla.org/en-US/docs/Web/API/ImageDecoder) can also be used here, and it performs comparatively to `createImageBitmap` on Chrome. See [these tests](https://stackblitz.com/edit/vitejs-vite-pakpzuq5?file=src%2Fmain.js,index.html&terminal=dev)

The difference is that it's **not supported by Safari** but can easily be tested with `typeof ImageDecoder === 'undefined'`

PR #5238, opened 2023-04-13
Replaced debug logging of the Element component by Tracing

mvaligursky:
- small refactor


willeastcott (approved):
I'm personally unsure of the value of all this logging. I'm wondering whether @daredevildave added it originally to debug something and simply never got around to removing it. But I'll approve and leave it for you to decide.

mvaligursky:
It could be useful for somebody not very familiar with the system to track some sorting issues, so lets leave it in.

PR #5246, opened 2023-04-14
Enabled additional type checking in the source code

mvaligursky:


slimbuck (approved):
Good job Will! I mean Martin.

willeastcott:
Any reason to keep the `jsconfig.json`? I think it used to do what you've enabled here. I believe the `tsconfig.json` was introduced in 2021 and it's now used for generating the TS declarations and the Typedocs, but probably overrode the `jsconfig.json`. So maybe it's safe to remove now.

kungfooman:
The irony is that `noImplicitAny: true` is required to find the original `length` error, but this is a good improvement anyway 👍 

IMO best would be a fully strict code base without errors, which would also be best for the developer experience.

E.g. this is an error right now with `noImplicitAny: true`:

image

But it would be simple enough to just add it to the type system, as in other classes:

*platform/graphics/graphics-device.js*, simply a non-implemented "abstract" method:
```js
    /**
     * Sets the active shader to be used during subsequent draw calls.
     * 
     * @param {import('./shader.js').Shader} shader - The shader to set to assign to the device.
     * @returns {boolean} True if the shader was successfully set, false otherwise.
     */
    setShader(shader) {
        throw new Error('not implemented');
    }
```

And suddenly we have proper Intellisense and no errors:

image

Would it be feasible to do this rather than disregarding all these type errors? @willeastcott, you are good at doing these large changes, would that be a possibility or is there an argument against doing this?

mvaligursky:
> Would it be feasible to do this (fully strict mode) rather than disregarding all these type errors?

Longer term absolutely. But right now this creates so many "errors" due to missing types, that the real errors would go unnoticed.

willeastcott (approved):

PR #5256, opened 2023-04-20
Refactor of private ShaderPass class to be dynamic instead of hardcoded

mvaligursky:
In the past, a shader pass ids were hardcoded, for example forward, depth, pick, and a whole list of them for combinations of light type and shadow type. The Editor also uses hardcoded number, which was limiting what can be done in the engine (separate PR to address). The shader pass dictates a shader needed for the pass - each pass generates a new shader for its specific requirements.

In order to make the engine more flexible, this has been now changed to a dynamic solution, where renderers can request shader creation for any number of passes. This will allow is to easily add debug rendering ('render albedo, render normal' ..).

For each shader pass, a new define is added to a shader, allowing per pass customization, i.e:
- SHADOW_PASS (existing), but also
- CUSTOM_PASS (based on 'custom' shader pass name)

Each shader pass created can have additional options to be attached, to be used by the generator.

There will be a separate follow up PR taking advantage of this to implement some debug rendering.

Note that few shader passes related to public constants commonly used are pre-created to guarantee their order matches the constants:

```
        add('depth', SHADER_DEPTH);
        add('pick', SHADER_PICK);
        add('shadow', SHADER_SHADOW);
```

Example logging of shader pass creation logging when rendering shadows for few different light types.

Screenshot 2023-04-18 at 12 26 18

### DIFF
--- a/src/scene/shader-pass.js
+++ b/src/scene/shader-pass.js
@@ -66,7 +66,6 @@ class ShaderPass {
      * Allocated shader passes, map of a shader pass name to info.
      *
      * @type {Map<string, ShaderPassInfo>}
-     * @ignore
      */
     passesNamed = new Map();
 
@@ -74,7 +73,6 @@ class ShaderPass {
      * Allocated shader passes, indexed by their index.
      *
      * @type {Array<ShaderPassInfo>}
-     * @ignore
      */
     passesIndexed = [];
 
--- a/src/scene/shader-pass.js
+++ b/src/scene/shader-pass.js
@@ -1,110 +1,148 @@
 import { Debug } from '../core/debug.js';
 import {
-    SHADER_FORWARD, SHADER_FORWARDHDR, SHADER_DEPTH, SHADER_PICK,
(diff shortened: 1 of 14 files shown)


src/scene/shader-pass.js
```diff
@@ -1,110 +1,148 @@
...
+ */
+class ShaderPass {
     /**
-     * Returns true if the shader pass is a forward shader pass.
+     * Allocated shader passes, map of a shader pass name to info.
      *
-     * @param {number} pass - The shader pass.
-     * @returns {boolean} - True if the pass is a forward shader pass.
+     * @type {Map<string, ShaderPassInfo>}
+     * @ignore
```
willeastcott, line 69: AFAIK, if there is an `@ignore` on the class description, you don't need `@ignore` on anything in the class itself.

willeastcott (approved):

src/scene/shader-pass.js
mvaligursky: yep fixed both, thanks

Issue #5259, opened 2023-04-21
Null exception error in SpriteAnimationClip if entity is destroyed before texture atlas loads


yaustar:
Repro steps:

1. Mark texture atlas to not preload
2. Use a sprite from the atlas in the scene
3. This will start an asset load for the texture atlas with callback 
4. Destroy the sprite before the texture atlas loads
5. Exception:

```
[[playcanvas-1.62.0.dbg.js:89590]](https://code.playcanvas.com/playcanvas-1.62.0.dbg.js): Cannot read properties of undefined (reading 'resource')

TypeError: Cannot read properties of undefined (reading 'resource')
    at SpriteAnimationClip._onSpriteAssetLoad (https://code.playcanvas.com/playcanvas-1.62.0.dbg.js:89590:17)
    at SpriteAnimationClip._onTextureAtlasLoad (https://code.playcanvas.com/playcanvas-1.62.0.dbg.js:89611:13)
    at AssetRegistry.fire (https://code.playcanvas.com/playcanvas-1.62.0.dbg.js:947:21)
    at _opened (https://code.playcanvas.com/playcanvas-1.62.0.dbg.js:57406:15)
    at Array._loaded (https://code.playcanvas.com/playcanvas-1.62.0.dbg.js:57428:10)
    at ResourceLoader._onSuccess (https://code.playcanvas.com/playcanvas-1.62.0.dbg.js:58957:30)
    at https://code.playcanvas.com/playcanvas-1.62.0.dbg.js:58911:19
    at handler (https://code.playcanvas.com/playcanvas-1.62.0.dbg.js:102185:8)
    at https://code.playcanvas.com/playcanvas-1.62.0.dbg.js:102260:19
```

Example project: https://playcanvas.com/project/1065057/overview/destroy-sprite-before-texture

PR #5260, opened 2023-04-21
Clean up event listeners on anim clip destroy

yaustar:
Fixes https://github.com/playcanvas/engine/issues/5259

### DIFF
--- a/src/framework/components/sprite/sprite-animation-clip.js
+++ b/src/framework/components/sprite/sprite-animation-clip.js
@@ -289,7 +289,7 @@ class SpriteAnimationClip extends EventHandler {
         asset.off('remove', this._onSpriteAssetRemove, this);
 
         // unbind atlas
-        if (asset.resource && asset.resource.atlas) {
+        if (asset.resource && !asset.resource.atlas) {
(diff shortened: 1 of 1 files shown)


src/framework/components/sprite/sprite-animation-clip.js
```diff
@@ -289,7 +289,7 @@ class SpriteAnimationClip extends EventHandler {
         asset.off('remove', this._onSpriteAssetRemove, this);
 
         // unbind atlas
-        if (asset.resource && asset.resource.atlas) {
+        if (asset.resource && !asset.resource.atlas) {
```
yaustar, line 292: We only set the event listener if `.atlas` is null (see https://github.com/playcanvas/engine/blob/main/src/framework/components/sprite/sprite-animation-clip.js#L303)

So it makes sense to only remove it if it is still null (ie the texture atlas hasn't been loaded yet)

willeastcott (approved):

willeastcott:
Would be good to fully verify this fix before merging.... (nice fix tho!)

yaustar:
I've asked the client to send over details where we can test this fix in a larger project with a large number of sprites, scene changing etc.

yaustar:
Checked against the clients larger project and no visual issues can be seen in areas where there's heavy sprite usuage

PR #5261, opened 2023-04-21
Debug shader output functionality

mvaligursky:
Functionality, allowing a name of the shader pass to be set on camera, to be used for its rendering. There are multiple shader debug passes predefined, allowing debug rendering functionality.

Custom shader pass names can also be used. Possible use cases is where for example user wants to use cheaper shaders when rendering to reflection texture. They can use the shader define to skip some processing in some shader chunks. Alternatively, they can modify StandardMaterialOptions in a callback to disable shader features.

There are two new shader chunks:
- debug-output.js - inserted at the end of the generated shader, outputs data that are written to color render target based on the shader pass define
- debug-process-frontend - inserted before the front end data are passed to backend, allowing those to be modified. For example to render 'Lighting', the albedo is set to mid gray color here.

### New API:
Screenshot 2023-04-21 at 16 45 06
Screenshot 2023-04-21 at 16 45 22

### Example:
Multi-View example updated with dropdown (not all options are visible and scrolling is needing, which is not obvious)

Screenshot 2023-04-21 at 14 42 59

https://user-images.githubusercontent.com/59932779/233651429-4411cb7f-de5c-4b36-ac8d-bd02c67668de.mov

### DIFF
--- a/src/framework/components/camera/component.js
+++ b/src/framework/components/camera/component.js
@@ -135,7 +135,7 @@ class CameraComponent extends Component {
      *
      * Additionally, a new name can be specified, which creates a new shader pass with the given
      * name. The name provided can only use alphanumeric characters and underscores. When a shader
-     * is compiled for the new pass, a define is added to the shader. For example if the name is
+     * is compiled for the new pass, a define is added to the shader. For example, if the name is
      * 'custom_rendering', the define 'CUSTOM_RENDERING_PASS' is added to the shader, allowing the
      * shader code to conditionally execute code only when that shader pass is active.
      *
--- a/src/framework/components/camera/component.js
+++ b/src/framework/components/camera/component.js
@@ -145,7 +145,7 @@ class CameraComponent extends Component {
      * callback can modify the shader generation options specifically for this shader pass.
      *
      * ```javascript
-     * var shaderPassId = camera.setShaderPassName('custom_rendering');
+     * const shaderPassId = camera.setShaderPassName('custom_rendering');
      *
      * material.onUpdateShader = function (options) {
      *    if (options.pass === shaderPassId) {
--- a/src/framework/components/camera/component.js
+++ b/src/framework/components/camera/component.js
@@ -1,10 +1,12 @@
+import { Debug } from '../../../core/debug.js';
+
 import { ASPECT_AUTO, LAYERID_UI, LAYERID_DEPTH } from '../../../scene/constants.js';
 import { Camera } from '../../../scene/camera.js';
+import { ShaderPass } from '../../../scene/shader-pass.js';
 
 import { Component } from '../component.js';
 
 import { PostEffectQueue } from './post-effect-queue.js';
-import { Debug } from '../../../core/debug.js';
 
 /**
  * Callback used by {@link CameraComponent#calculateTransform} and {@link CameraComponent#calculateProjection}.
@@ -93,6 +95,9 @@ class CameraComponent extends Component {
      */
     _disablePostEffectsLayer = LAYERID_UI;
 
+    /** @private */
+    _camera = new Camera();
+
     /**
      * Create a new CameraComponent instance.
      *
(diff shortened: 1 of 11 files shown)


willeastcott:
So cool. I was wondering. Maybe this feature warrants a new example. Maybe with 9 camera (3x3 grid) rendering an object with each viewport labelled with the pass being used.

src/scene/constants.js
```diff
@@ -681,6 +681,76 @@ export const SHADERTYPE_PICK = 'pick';
  */
 export const SHADERTYPE_SHADOW = 'shadow';
 
+/**
+ * Shader used for debug rendering of albedo.
+ *
+ * @type {string}
+ */
+export const SHADERTYPE_DEBUG_ALBEDO = 'debug_albedo';
```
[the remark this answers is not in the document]
willeastcott: Alternatively, should `setShaderPassName` be `setShaderType`?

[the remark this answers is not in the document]
mvaligursky: we already have SHADERTYPE_FORWARD, SHADERTYPE_DEPTH and few more .. which can be passed here as well, so I thought to keep it as a single enum.

src/framework/components/camera/component.js
```diff
@@ -104,13 +109,54 @@ class CameraComponent extends Component {
...
         // postprocessing management
         this._postEffects = new PostEffectQueue(system.app, this);
     }
 
+    /**
+     * Sets the name of the shader pass the camera will use when rendering.
+     *
+     * @param {string} name - The name of the shader pass. Defaults to undefined, which is
+     * equivalent to {@link SHADERTYPE_FORWARD}.
+     * Can be:
```
willeastcott, line 123:
```suggestion
     * equivalent to {@link SHADERTYPE_FORWARD}. Can be:
```

src/framework/components/camera/component.js
```diff
@@ -104,13 +109,54 @@ class CameraComponent extends Component {
...
+     * - {@link SHADERTYPE_DEBUG_GLOSS}
+     * - {@link SHADERTYPE_DEBUG_METALNESS}
+     * - {@link SHADERTYPE_DEBUG_AO}
+     * - {@link SHADERTYPE_DEBUG_EMISSION}
+     * - {@link SHADERTYPE_DEBUG_LIGHTING}
+     * - {@link SHADERTYPE_DEBUG_UV0}
+     *
+     * Additionally, a new name can be specified, which creates a new shader pass with the given
+     * name. The name provided can only use alphanumeric characters and underscores. When a shader
+     * is compiled for the new pass, a define is added to the shader. For example if the name is
```
willeastcott, line 139:
```suggestion
     * is compiled for the new pass, a define is added to the shader. For example, if the name is
```

src/framework/components/camera/component.js
mvaligursky: done, ta

src/framework/components/camera/component.js
```diff
@@ -104,13 +109,74 @@ class CameraComponent extends Component {
...
+     * var shaderPassId = camera.setShaderPassName('custom_rendering');
+     *
+     * material.onUpdateShader = function (options) {
+     *    if (options.pass === shaderPassId) {
+     *        options.litOptions.normalMapEnabled = false;
+     *        options.litOptions.useSpecular = false;
+     *    }
+     *    return options;
+     * };
+     * ```
```
willeastcott, line 157: This should be in an `@example` block.

willeastcott: And don't use `var`. 😉

src/scene/constants.js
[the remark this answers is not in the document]
willeastcott: The API just doesn't quite seem right at the moment. Don't forget, we can always alias constant names. But I'd rather have an API that is consistent/guessable/etc.

What about `RENDERMODE_`? And I think the `_DEBUG_` is kinda redundant, IMHO.

src/framework/components/camera/component.js
mvaligursky: When I use example block, the example is moved to the start which I don't like. Mesh class does the same to keep it ordered. Is there some other way?
Screenshot 2023-04-21 at 15 24 06

slimbuck:
I'm struggling to understand how the new shaderPassName relates to the existing FORWARD, DEPTH, ETC passes.

Is SHADERTYPE_DEBUG_AO assumed to be a FORWARD pass or...?

mvaligursky:
> I'm struggling to understand how the new shaderPassName relates to the existing FORWARD, DEPTH, ETC passes.
> 
> Is SHADERTYPE_DEBUG_AO assumed to be a FORWARD pass or...?

Shader pass name is converted to a number internally, and this is used as a pass that is used all over the place, for example in options.pass, or forward-renderer using pass to access shader from an array on material / mesh instance. FORWARD is 0, SHADOW is 1 .. and new passes that are created on camera are added after those. If the user wants to render shadow depth map, they could use existing DEPTH pass .. they are all the same. That is way I use the same SHADERTYPE_** constants. But I like @willeastcott 's suggestion of renaming them to SHADERPASS_**

willeastcott:
If you _do_ alias to `SHADERPASS_`, I would suggest changing `CameraComponent#setShaderPassName` to simply `CameraComponent#setShaderPass`.

src/scene/constants.js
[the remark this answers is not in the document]
mvaligursky: renamed to SHADERPASS_

src/scene/shader-lib/chunks/chunks.js
```diff
@@ -27,6 +27,8 @@ import cookiePS from './lit/frag/cookie.js';
 import cubeMapProjectBoxPS from './lit/frag/cubeMapProjectBox.js';
 import cubeMapProjectNonePS from './lit/frag/cubeMapProjectNone.js';
 import cubeMapRotatePS from './lit/frag/cubeMapRotate.js';
+import debugOutputPS from './lit/frag/debug-output.js';
+import debugProcessFrontendPS from './lit/frag/debug-process-frontend.js';
```
willeastcott, line 31: Actually, can we consider this naming too please (since once chunk names are chosen, they can be used in app code). I'm not sure about `debugOutput` - the `_DEBUG_` naming has gone from the constants. I can think of non-debug uses for this. And also `debugProcessFrontend` is a bit verbose. Not sure what could be better though. Ideas, @GSterbrant and @slimbuck?

mvaligursky: debugOutputPS could be: **postEndPS**
debugProcessFrontendPS could be: **postFrontEndPS** or processLitArgsPS

I'd probably vote for 1st on these lists.

willeastcott: Yeah, I think they're better. Would still love for @GSterbrant and @slimbuck to chip in too...

slimbuck (approved):

src/scene/shader-lib/chunks/chunks.js
slimbuck: We chatted about this offline and decided this naming isn't great and we'd like to fix it, but not in the PR.

Issue #4807, opened 2022-11-01
AO detail maps


animech-sijo:
PlayCanvas currently supports Diffuse and Normal detail maps (added in #1968). We often need more detail maps though, especially AO, and more rarely Glossiness. Can we add these detail maps?

Previous issue describing why we need detail maps in the first place: #1962.

PR #1968, opened 2020-04-04
Detail Maps

aidinabedi:
Fixes #1962

PR includes adds the following properties to `pc.StandardMaterial` with full documentation:
- `diffuseDetailMap` (texture)
- `diffuseDetailMapChannel` ("rgb", "r", "g", "b", "a")
- `diffuseDetailMapUv` (number)
- `diffuseDetailMapTiling` (vec2)
- `diffuseDetailMapOffset` (vec2)
- `diffuseDetailBlend` (`pc.DETAILBLEND_COLORS_*`)
- `normalDetailMap` (texture)
- `normalDetailMapTiling` (vec2)
- `normalDetailMapOffset` (vec2)
- `normalDetailMapUv` (number)
- `normalDetailMapBumpiness` (number)
- `normalDetailBlend` (`pc.DETAILBLEND_NORMALS_*`)

----

All changes have been tested using a [public PlayCanvas project](https://playcanvas.com/project/671513/overview/detail-maps-test) (simply run it with and without the PR):

### With PR:
image

### Without PR:
image

I confirm I have signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #5270, opened 2023-04-25
AO Detail Map

Mistale:
Fixes [#4807](https://github.com/playcanvas/engine/issues/4807)
Uses the concepts introduced by [#1968](https://github.com/playcanvas/engine/pull/1968).

PR adds the following properties to `pc.StandardMaterial` with documentation and tests:

* `aoDetailMap` (texture)
* `aoDetailMapChannel` ("r", "g", "b", "a", default is "g")
* `aoDetailMapUv` (number)
* `aoDetailMapTiling` (vec2)
* `aoDetailMapOffset` (vec2)
* `aoDetailMode` (`pc.DETAILMODE_*`)

When combined with a baked `aoMap`, this PR allows for high- and low-frequency AO to be mixed without requiring excessively large baked textures. Typical use-cases may be a fence, a car grille or similar where one might want to use a baked AO texture for the entire object, but use a tiling baked AO texture for the intricate details.

### Example:
image (3)
> From left to right: baked AO, baked AO + detail Normal, baked AO + detail Normal + detail AO 


slimbuck:
Hi @Mistale,

Sorry for our slow response on your PR, it's awesome so we'll review/test/merge real soon.

Thanks!!

src/scene/materials/standard-material.js
```diff
@@ -447,18 +447,43 @@ let _params = new Set();
...
+ * (primary) ao map is non-null.
+ * @property {number} aoDetailMapUv Detail (secondary) AO map UV channel.
+ * @property {Vec2} aoDetailMapTiling Controls the 2D tiling of the detail (secondary) AO
+ * map.
+ * @property {Vec2} aoDetailMapOffset Controls the 2D offset of the detail (secondary) AO
+ * map. Each component is between 0 and 1.
+ * @property {number} aoDetailMapRotation Controls the 2D rotation (in degrees) of the detail
+ * (secondary) AO map.
+ * @property {string} aoDetailMapChannel Color channels of the detail (secondary) AO map
+ * to use. Can be "r", "g", "b" or "a".
```
mvaligursky, line 473: This should mention the default value is 'g'

mvaligursky (approved):
I'm approving this, with a small JSDocs suggestion, but would like @GSterbrant to have a look as well.

My question is regarding the aoDetailMode parameter - did you find this useful when testing, versus just using one fixed way to blend those ao maps?

src/scene/materials/standard-material.js
Maksims: Is it common to use default value 'g' instead or 'r' in other similar properties?

GSterbrant: We don't seem to have that mentioned elsewhere in the docs, but we do set 'g' as the default channel in the material. It should be down to how textures are packed, surely, unless there is some bit depth favor for green channels (which would make sense perceptually) for certain file formats?

GSterbrant (approved):
Looks good! Just remember to update the chunk API version to 1.63 before merge. 

It does spawn a conversation I think we've had before on the team which is that perhaps we should add details maps as an intrinsic engine feature instead of having to add a bunch of extra parameters manually for each texture you'd want to detail map.

src/scene/shader-lib/chunks/standard/frag/ao.js
```diff
@@ -4,7 +4,8 @@ void getAO() {
     dAo = 1.0;
 
     #ifdef MAPTEXTURE
-    dAo *= texture2DBias($SAMPLER, $UV, textureBias).$CH;
+    float aoBase = texture2DBias($SAMPLER, $UV, textureBias).$CH;
```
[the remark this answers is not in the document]
mvaligursky: yep, love the idea of automatic optional detail map for any texture. Please create an issue for this @GSterbrant

src/scene/materials/standard-material.js
Mistale: I've added a clarification about the default value. I can't figure out why 'g' is the default for ao either, but I figured I'd just follow the same standard as  `aoMapChannel`. 
When packing ORM-textures for glTF, "r" is used for occlusion. But it may be related to inherent precision issues as @GSterbrant mentions.

Mistale:
> I'm approving this, with a small JSDocs suggestion, but would like @GSterbrant to have a look as well.
> 
> My question is regarding the aoDetailMode parameter - did you find this useful when testing, versus just using one fixed way to blend those ao maps?

I think that using `DETAILMODE_MUL` as default covers a lot of normal use-cases, with one specific exception:
- When multiplying primary and secondary AO together using mul, the results will always vary (slightly) by the pattern of the secondary map, even in areas where the primary occlusion should be dark enough to completely remove details.
In this case a choice between `DETAILMODE_MIN` and `DETAILMODE_MAX` can make the results "flatter" while at the same time allowing the artist to introduce a bias towards the lighter or darker spectrum.

The other options are mainly there for artistic freedom.

src/scene/shader-lib/chunks/chunk-validation.js
```diff
@@ -5,6 +5,7 @@ import { shaderChunks } from './chunks.js';
 const chunkVersions = {
     // frontend
     aoPS: CHUNKAPI_1_57,
+    aoDetailMapPS: CHUNKAPI_1_62,
```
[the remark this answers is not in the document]
Mistale: Just wanted to check if this was intended for me or internally (for the team)? It's my first PR, so I'm not quite sure about how you usually handle some of the details 😃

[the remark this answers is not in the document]
slimbuck: Also just to add this is a great first PE @Mistale - thanks so much!

willeastcott:
This now has two approvals - any objections to merging @mvaligursky and @slimbuck? Last call for comments! ⌛

mvaligursky:
Lets merge it in.

PR #5277, opened 2023-04-26
Limit allowed samples on WebGPU when using multisampled render target

mvaligursky:
WebGPU only supports 1 or 4 samples, enforce this limitation


mvaligursky:
> If device.maxSamples is valid (1 or 4) it might be cleaner to use that?

It's not valid, it's something I set up. This is WebGPU docs:

Screenshot 2023-04-26 at 19 10 51

slimbuck:
So all wgpu devices support 1 or 4 samples?

slimbuck:
...and then maxSamples should always be 4?

slimbuck (approved):

PR #5283, opened 2023-04-27
Temporarily disable ParticleSystems on WebGPU to avoid errors

mvaligursky:
- particle systems are currently not supported on WebGPU - disable them to avoid errors if those are in the scene
- enabled RenderToTexture example for WebGPU, as now it runs (apart from particles)

### DIFF
--- a/examples/src/examples/graphics/render-to-texture.tsx
+++ b/examples/src/examples/graphics/render-to-texture.tsx
@@ -3,6 +3,7 @@ import * as pc from '../../../../';
 class RenderToTextureExample {
     static CATEGORY = 'Graphics';
     static NAME = 'Render to Texture';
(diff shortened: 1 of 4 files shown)


slimbuck (approved):

mvaligursky:
Thanks for the question and nice comments!

Yes, it can definitely be done, and it's on a TODO list. But till it's done, to avoid crashes if people try to use it, it's disabled. Notice the name of this PR: "Temporarily".

seth-ellison:
Yeah! I saw it was temporary, I was just concerned about the description citing the whole, "particle systems are currently not supported on WebGPU" so I wanted to share what I had!

Again, thank you so much for the quick answer! Hope you have a great day!

PR #5285, opened 2023-04-27
Revert rename of the uniform used by the Picker shaders

mvaligursky:
revert of https://github.com/playcanvas/engine/pull/5266 as Editor depends on the Basic shader for some picker meshes which uses generic uColor

### DIFF
--- a/src/framework/graphics/picker.js
+++ b/src/framework/graphics/picker.js
@@ -191,7 +191,7 @@ class Picker {
 
         const device = this.device;
         const self = this;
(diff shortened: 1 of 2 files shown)


willeastcott:
Why not just update the Editor? :)

mvaligursky:
I changed Editor where it uses the constant, but it uses Basic Material in the engine for other things to pick, which does not have a "picker pass" generator, which would need to be added. Considering the original PR was just for clarity, this is not worth doing.

willeastcott (approved):

Issue #5286, opened 2023-04-27
1.62.2 VSM shadows from spotlights have no blur


yaustar:
1.62.2 https://playcanvas.com/editor/scene/1730271?use_local_engine=https://code.playcanvas.com/playcanvas-1.62.2.js
<img width="792" alt="image" src="https://user-images.githubusercontent.com/16639049/234918860-331b8285-e9e2-492e-b6f2-4293466646ba.png">

1.61.3 https://playcanvas.com/editor/scene/1730271?use_local_engine=https://code.playcanvas.com/playcanvas-1.61.3.js
<img width="788" alt="image" src="https://user-images.githubusercontent.com/16639049/234918979-47a21e90-ba3e-4e73-822d-e0d5764f7e12.png">

PR #5288, opened 2023-04-28
[Fix] Apply VMS blur to spot lights when clustered lighting is off

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/5286

- this got removed during a recent refactoring of local lights to use render passes

### DIFF
--- a/src/scene/renderer/shadow-renderer-local.js
+++ b/src/scene/renderer/shadow-renderer-local.js
@@ -156,7 +156,7 @@ class ShadowRendererLocal {
         if (applyVsm) {
             renderPass.after = () => {
                 // after the pass is done, apply VSM blur if needed
-                this.shadowRenderer.renderVms(light, shadowCamera);
+                this.shadowRenderer.renderVsm(light, shadowCamera);
             };
         }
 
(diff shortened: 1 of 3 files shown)


src/scene/renderer/shadow-renderer-local.js
```diff
@@ -152,6 +152,14 @@ class ShadowRendererLocal {
         this.shadowRenderer.setupRenderPass(renderPass, shadowCamera, true);
         DebugHelper.setName(renderPass, `SpotShadow-${light._node.name}`);
 
+        // apply vsm
+        if (applyVsm) {
+            renderPass.after = () => {
+                // after the pass is done, apply VSM blur if needed
+                this.shadowRenderer.renderVms(light, shadowCamera);
```
willeastcott, line 159: Vms? Guess this should be Vsm?

willeastcott (approved):

src/scene/renderer/shadow-renderer-local.js
mvaligursky: done, thanks!

PR #5291, opened 2023-04-28
Glb async loading

slimbuck:
This PR is a followup to #5208 and further improves asynchronous loading of gltf files:
- buffers and textures are downloaded asynchronously
- draco decompression and texture download (typically the two slowest operations) can now happen simultaneously
- extracted the asset parser interface from glb-parser

The code is also made simpler and more readable using Promises.

Speed up will depend on the precise asset contents, number of textures and whether draco compression is used. From my early testing I see 10-20% faster loading on models with textures and draco.

**Testing Notes:**
- [ ] ensure that models (both glb and json), animations and container resources all load correctly (import assets in editor for each type)
- [ ] test all variants of gltf load correctly - glb, gltf with external resources, gltf with embedded resources
- [ ] test that the loading callback functions (preprocess, process, postprocess) all still work correctly
- [ ] load glb models with draco compression and without


src/framework/parsers/glb-container-parser.js
```diff
@@ -0,0 +1,53 @@
...
+        }, asset, this.maxRetries);
+    }
+
+    open(url, data, asset) {
+        return data;
+    }
+
+    patch(asset, assets) {
+
+    }
```
willeastcott, line 48: Kind of annoying you have to specify an empty function here. Can't the asset system handle parsers that don't specify a patch func?

src/framework/parsers/glb-parser.js
```diff
@@ -46,27 +46,14 @@ import { AnimCurve } from '../anim/evaluator/anim-curve.js';
...
-        this.materials = null;
-        this.variants = null;
-        this.meshVariants = null;
-        this.meshDefaultMaterials = null;
-        this.renders = null;
-        this.skins = null;
-        this.lights = null;
-        this.cameras = null;
+    constructor(members) {
+        Object.assign(this, members);
```
willeastcott, line 56: Maybe use spread here - let's use modern JS. 😄

src/framework/parsers/glb-parser.js
```diff
@@ -2022,139 +2014,165 @@ const loadImageAsync = (gltfImage, index, bufferViews, urlBase, registry, option
...
-    const assets = [];        // one per image
-    const textures = [];      // list per image
+    const seenImages = new Set();
 
-    let remaining = gltf.textures.length;
-    const onLoad = (textureIndex, imageIndex) => {
-        if (!textures[imageIndex]) {
-            textures[imageIndex] = [];
+    for (let i = 0; i < gltf.textures.length; ++i) {
+        const gltfTexture = gltf.textures[i];
```
willeastcott, line 2122: Don't you prefer:

```
for (const texture of gltf.texture) {
```

willeastcott (approved):
What an awesome PR!! 🔥  Do you have any stats to add to the PR description to describe the expected real-world impact of this optimization?

src/framework/parsers/glb-parser.js
slimbuck: I've made them `array.map`s instead, thanks!

src/framework/parsers/glb-container-parser.js
mvaligursky: they could be called as `parser?.open(...)` for example

src/framework/parsers/glb-parser.js
```diff
@@ -46,27 +46,14 @@ import { AnimCurve } from '../anim/evaluator/anim-curve.js';
...
-        this.textures = null;
-        this.materials = null;
-        this.variants = null;
-        this.meshVariants = null;
-        this.meshDefaultMaterials = null;
-        this.renders = null;
-        this.skins = null;
-        this.lights = null;
-        this.cameras = null;
+    constructor(members) {
```
mvaligursky, line 55: It'd be good to have jsdocs for this constructor, as 'members' does not mean much to me.
But should be try to go strong typed way anyways, where we try to limit dynamic members in the class? Ideally this class would look like TS class with members specified.

slimbuck: OK I've create member declarations, though without types for now.

PR #5294, opened 2023-05-02
Draco worker initialisation update

slimbuck:
Followup to #5208.

This PR changes draco initialisation on workers. Instead of each worker downloading the glue script and compiling the module, this is done once on the main thread and passed to workers.

This is done primarily to retain compatibility with Facebook games (see #5016).

From what I can see this change does not appear to impact startup time.

### DIFF
--- a/src/framework/parsers/draco-worker.js
+++ b/src/framework/parsers/draco-worker.js
@@ -246,10 +246,10 @@ function DracoWorker(jsUrl, wasmUrl) {
                         return {};
                     }
                 })
-                .then(instance => {
-                    draco = instance;
-                    workQueue.forEach(data => decode(data));
-                });
+                    .then((instance) => {
+                        draco = instance;
+                        workQueue.forEach(data => decode(data));
+                    });
                 break;
             case 'decodeMesh':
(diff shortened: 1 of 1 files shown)


mvaligursky (approved):

src/framework/parsers/draco-decoder.js
```diff
@@ -78,6 +101,44 @@ class JobQueue {
...
+        };
+
+        http.get(url, options, (err, response) => {
+            if (err) {
+                reject(err);
+            } else {
+                resolve(response);
+            }
+        });
+    });
```
willeastcott, line 120: Just throwing it out there, but any reason not to start using `fetch` in the engine codebase?

willeastcott: https://caniuse.com/?search=fetch

willeastcott (approved):

src/framework/parsers/draco-decoder.js
slimbuck: This was left to leverage http retry logic.

PR #5295, opened 2023-05-02
Glb container parse event

pavle-goloskokovic:
This PR introduces change where `parse` event is fired on a container asset to indicate that data was downloaded and parsing is about to begin.

This functionality adds more control over loading process, where e.g. we can start loading low priority assets immediately after container asset has downloaded, without needing to wait for its parsing as well, in effect shortening total loading time.

### DIFF
--- a/src/framework/parsers/glb-container-parser.js
+++ b/src/framework/parsers/glb-container-parser.js
@@ -20,6 +20,7 @@ class GlbContainerParser {
             if (err) {
                 callback(err);
             } else {
+                asset.fire('parse', asset);
                 GlbParser.parse(
                     this._getUrlWithoutParams(url.original),
                     path.extractPath(url.load),


mvaligursky:
any thoughts on this one @slimbuck ?

querielo:
Isn't `preprocess` of `global` is suitable for this?

[Documentation](https://developer.playcanvas.com/en/api/pc.ContainerHandler.html)

```js
const containerAsset = new pc.Asset(filename, 'container', { url: url, filename: filename }, null, {
    global: {
        preprocess() { console.log("parsing is about to begin"); }
    },
});
```

pavle-goloskokovic:
> Isn't `preprocess` is suitable for this case?
> 
> [Documentation](https://developer.playcanvas.com/en/api/pc.ContainerHandler.html)
> 
> ```js
> const containerAsset = new pc.Asset(filename, 'container', { url: url, filename: filename }, null, {
>     global: {
>         preprocess() { console.log("parsing is about to begin"); }
>     },
> });
> ```

Unfortunately not since `preprocess` callback is called within `createResources` method after parsing is already done:
https://github.com/playcanvas/engine/blob/cee326b46c4025b16a3c301fd285c6319e5793a7/src/framework/parsers/glb-parser.js#L2470

willeastcott:
Thanks for the PR, and apologies that it sat unreviewed for so long. We've decided to close it rather than add a new asset event, for the following reasons.

**The existing `options.global.preprocess` hook already covers this.** The earlier comment that it runs "after parsing is already done" is only partly right. It's called at the top of `createResources`, before any mesh, material or texture resource is created:

https://github.com/playcanvas/engine/blob/main/src/framework/parsers/glb-parser.js#L1240-L1245

The only work between the download completing and `preprocess` being called is splitting the GLB chunks, one `JSON.parse` of the glTF JSON, and setting up buffer views and image assets. All of that is light. The expensive synchronous work (building meshes, materials, skins, animations) happens after `preprocess`, so kicking off low-priority loads from that callback overlaps them with the parse just as well as the proposed event would:

```js
const asset = new pc.Asset('scene.glb', 'container', { url }, null, {
    global: {
        preprocess: (gltf) => {
            // download complete, resource creation about to start
            app.assets.load(lowPriorityAsset);
        }
    }
});
```

**A `parse` event on the generic `Asset` class would only ever fire for container assets.** We'd need an `Asset.EVENT_PARSE` constant, JSDoc, TypeScript coverage and a test, and we'd have to document that no other asset type emits it. The `progress` event already has that shape and it's a wart we'd rather not repeat.

For completeness, container assets also emit `progress` (`Asset.fetchArrayBuffer` passes the asset as the progress target), which can serve as a download-complete signal when the server sends a content-length header.

If `global.preprocess` turns out not to work for your loading manager in practice, please open an issue describing the setup and we'll revisit.

PR #5299, opened 2023-05-03
Faster wireframe gen

slimbuck:
Use `Set` to track line hashes and use plain math instead of bit math. Also correctly support meshes with more than 65535 vertices.

This change results in roughly double the speed generation of indices.


LeXXik:
I am curious to read about the info that led you to swap out bitwise operations.

mvaligursky (approved):

slimbuck:
@LeXXik, using set makes a bigger difference than bit operations for performance. I kept the vanilla math over bit operations so it'd work correctly for meshes with more than 65536 vertices.

willeastcott (approved):

PR #5303, opened 2023-05-03
Change to static members of BlendState

mvaligursky:
- deprecated `BlendState.DEFAULT` due to non-obvious naming
- replaced it by `BlendState.NOBLEND`
- added `BlendState.ALPHABLEND` which is commonly used


GSterbrant (approved):

willeastcott (approved):
Really good improvement. Thanks.

PR #5312, opened 2023-05-10
Downgrade draco assert to warning

slimbuck:
The test is not fatal, downgrade to warning.

### DIFF
--- a/src/framework/parsers/glb-parser.js
+++ b/src/framework/parsers/glb-parser.js
@@ -721,7 +721,7 @@ const createDracoMesh = (device, primitive, accessors, bufferViews, meshVariants
             } else {
                 // create vertex buffer
                 const numVertices = decompressedData.vertices.byteLength / vertexFormat.size;
-                Debug.assert(numVertices === accessors[primitive.attributes.POSITION].count, 'mesh has invalid draco sizes');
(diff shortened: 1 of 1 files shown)


src/framework/parsers/glb-parser.js
```diff
@@ -721,7 +721,9 @@ const createDracoMesh = (device, primitive, accessors, bufferViews, meshVariants
             } else {
                 // create vertex buffer
                 const numVertices = decompressedData.vertices.byteLength / vertexFormat.size;
-                Debug.assert(numVertices === accessors[primitive.attributes.POSITION].count, 'mesh has invalid draco sizes');
+                if (numVertices !== accessors[primitive.attributes.POSITION].count) {
+                    Debug.warn('mesh has invalid draco sizes');
+                }
```
mvaligursky, line 726: add the whole thing inside `Debug.call(() => { }` so that even the condition gets stripped out
and maybe add that glb name if possible
and warnOnce could be good as well

yaustar: +1 on some identifier would be great 
image

mvaligursky (approved):

src/framework/parsers/glb-parser.js
slimbuck: Projects don't need reimport TBH. We'll patch the engine with this change now so we can get it out quickly, then add more debugging information later.

PR #5316, opened 2023-05-10
Calculate draco indices

slimbuck:
This PR fixes a bug introduced in #5208.

An old version of the playcanvas glb converter was writing incorrect count values for number of vertices and indices in some cases.

This PR ignores the values written in the file json and uses the counts returned from draco instead.

### DIFF
--- a/src/framework/parsers/glb-parser.js
+++ b/src/framework/parsers/glb-parser.js
@@ -721,16 +721,19 @@ const createDracoMesh = (device, primitive, accessors, bufferViews, meshVariants
             } else {
                 // create vertex buffer
                 const numVertices = decompressedData.vertices.byteLength / vertexFormat.size;
+                const indexFormat = numVertices <= 65535 ? INDEXFORMAT_UINT16 : INDEXFORMAT_UINT32;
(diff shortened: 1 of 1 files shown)


src/framework/parsers/glb-parser.js
```diff
@@ -721,16 +721,19 @@ const createDracoMesh = (device, primitive, accessors, bufferViews, meshVariants
...
+                const indexFormat = numVertices <= 65535 ? INDEXFORMAT_UINT16 : INDEXFORMAT_UINT32;
+                const numIndices = decompressedData.indices.byteLength / (numVertices <= 65535 ? 2 : 4);
+
                 Debug.call(() => {
                     if (numVertices !== accessors[primitive.attributes.POSITION].count) {
-                        Debug.warn('mesh has invalid draco sizes');
+                        Debug.warn('mesh has invalid vertex count');
+                    }
+                    if (numIndices !== accessors[primitive.indices].count) {
+                        Debug.warn('mesh has invalid index count');
```
mvaligursky, line 732: still no way to print the glb name? Without it i's not super useful message.

slimbuck: nope. user doesn't need to do anything anyway.

willeastcott (approved):

Issue #5262, opened 2023-04-22
Mesh decals example issue - incorrect colors on WebGPU


LeXXik:
Mesh decals example is not working correctly under WebGPU. The colors of the decals are not correct. Works fine under WebGL.

https://playcanvas.github.io/#/graphics/mesh-decals

mvaligursky:
Yes, I'm aware and will try to sort it out one day, thanks.

PR #5319, opened 2023-05-11
Update to vertex format handling and morphing on WebGPU

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/5262

Changes:
- WebGPU implementation now handles normalized vertex element formats
- modification to texture based morphing on WebGPU to use uint32 vertex Ids and texture fetch to sample morph target textures, to avoid problems with float conversion / half pixel offset

MeshMorphMany example has been changes to use a morph-stress-test example instead of procedural geometry, and it reveals issues in a cleaner way. The example has been disabled for WebGPU temporarily, due to the issues with vertex-id buffer passing to GPU and will be fixed separately.

Screenshot 2023-05-11 at 11 53 16


slimbuck (approved):

src/scene/shader-lib/chunks/lit/vert/normal.js
```diff
@@ -30,9 +30,16 @@ vec3 getNormal() {
...
 
     #ifdef MORPHING_TEXTURE_BASED_NORMAL
+
+    #ifdef WEBGPU
+            ivec2 morphUV = getTextureMorphCoords();
+            vec3 morphNormal = texelFetch(morphNormalTex, ivec2(morphUV), 0).xyz;
+        #else
+            vec2 morphUV = getTextureMorphCoords();
+            vec3 morphNormal = texture2D(morphNormalTex, morphUV).xyz;
+        #endif
```
willeastcott, line 40: Indentation here looks inconsistent.

mvaligursky: good point: Fixed incorrect indentation in normal.js chunk
thanks.

Issue #5326, opened 2023-05-12
PR 5109 breaks accessing `element.material` when a material asset is set in the Editor


yaustar:
Tutorial project broken by https://github.com/playcanvas/engine/pull/5109

Project: https://playcanvas.com/project/705273

`this.entity.element.material` is null

CC @MushAsterion

mvaligursky:
It seems there is another issue that is related and needs investigation:
https://forum.playcanvas.com/t/cannot-set-properties-of-alphatest-for-materials-even-on-playcanvas-examples/31387

yaustar:
This is fixed in 1.63.5: https://github.com/playcanvas/engine/releases/tag/v1.63.5

PR #5327, opened 2023-05-12
Only removed asset events in setMaterial for image element

yaustar:
Fixes https://github.com/playcanvas/engine/issues/5326

Also tested against the original bug https://github.com/playcanvas/engine/issues/3560

### DIFF
--- a/src/framework/components/element/image-element.js
+++ b/src/framework/components/element/image-element.js
@@ -976,6 +976,19 @@ class ImageElement {
         return this._rect;
     }
 
+    _removeMaterialAssetEvents() {
+        if (this._materialAsset) {
+            const assets = this._system.app.assets;
+            assets.off('add:' + this._materialAsset, this._onMaterialAdded, this);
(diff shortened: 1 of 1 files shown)


src/framework/components/element/image-element.js
```diff
@@ -994,7 +1007,8 @@ class ImageElement {
         if (this._materialAsset) {
             const asset = this._system.app.assets.get(this._materialAsset);
             if (!asset || asset.resource !== value) {
-                this.materialAsset = null;
```
yaustar, line 997: This created a circular call back to `set material` where it gets set to null

mvaligursky (approved):

src/framework/components/element/image-element.js
MushAsterion: It was to remove the events, but wasn't thinking of the circular callback, nice spot!

MushAsterion (approved):
Uh sorry for that, thanks for the correction!

PR #5330, opened 2023-05-15
Add support for PCF1 to directional / spot lights

### LINKED ISSUE #3194: Implement cheaper shadow sampling
In current implementation, PCF3 is the cheapest available shadow sampling. For mobile, especially older devices, even cheaper filtering could be implemented. PCF2, and even possibly single tap sampling (which is getting used in mobile games) to keep the shadow cost down.

ellthompson:
Updates the light component with support for the PCF1 shadow type when either the directional or spot light type has been selected.

image

Also updates the shadow cascades example with this option.

This is the final change necessary to close #3194

### DIFF
--- a/src/scene/light.js
+++ b/src/scene/light.js
@@ -334,7 +334,7 @@ class Light {
             value = SHADOW_VSM8;
 
         this._isVsm = value >= SHADOW_VSM8 && value <= SHADOW_VSM32;
-        this._isPcf = value === SHADOW_PCF5 || value === SHADOW_PCF3 || value === SHADOW_PCF1;
+        this._isPcf = value === SHADOW_PCF1 || value === SHADOW_PCF3 || value === SHADOW_PCF5;
 
         this._shadowType = value;
         this._destroyShadowMap();
--- a/src/scene/light.js
+++ b/src/scene/light.js
@@ -9,7 +9,7 @@ import {
     BLUR_GAUSSIAN,
     LIGHTTYPE_DIRECTIONAL, LIGHTTYPE_OMNI, LIGHTTYPE_SPOT,
     MASK_BAKE, MASK_AFFECT_DYNAMIC,
-    SHADOW_PCF3, SHADOW_PCF5, SHADOW_VSM8, SHADOW_VSM16, SHADOW_VSM32,
+    SHADOW_PCF1, SHADOW_PCF3, SHADOW_PCF5, SHADOW_VSM8, SHADOW_VSM16, SHADOW_VSM32,
     SHADOWUPDATE_NONE, SHADOWUPDATE_REALTIME, SHADOWUPDATE_THISFRAME,
     LIGHTSHAPE_PUNCTUAL, LIGHTFALLOFF_LINEAR
 } from './constants.js';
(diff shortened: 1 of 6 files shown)


src/scene/light.js
```diff
@@ -334,7 +334,7 @@ class Light {
             value = SHADOW_VSM8;
 
         this._isVsm = value >= SHADOW_VSM8 && value <= SHADOW_VSM32;
-        this._isPcf = value === SHADOW_PCF5 || value === SHADOW_PCF3;
+        this._isPcf = value === SHADOW_PCF5 || value === SHADOW_PCF3 || value === SHADOW_PCF1;
```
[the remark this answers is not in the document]
willeastcott: Sorry, the order was bugging me! 😆

slimbuck:
How about point lights - also supported?

ellthompson:
> How about point lights - also supported?

This is lower priority as omni light shadows aren't cheap to render anyway.

src/scene/shader-lib/programs/lit-shader.js
```diff
@@ -1239,6 +1239,8 @@ class LitShader {
                         }
                     } else if (light._shadowType === SHADOW_PCF5) {
                         shadowReadMode = "PCF5x5";
+                    } else if (light._shadowType === SHADOW_PCF1) {
+                        shadowReadMode = "PCF1x1";
```
willeastcott, line 1243: Kinda wish this was a switch statement TBH. But that's not your fault of course! 😄

willeastcott (approved):
LGTM. Get at least 1 more approval before merging though.

mvaligursky:
I wonder why the shadow looks black and white (binary). When looking at clustered-spot-shadows example (disabled normal maps) with PCF1, there is still filtering taking place, using a single bilinear sample. This should take place in your case as well.
Screenshot 2023-05-15 at 15 58 04

src/scene/shader-lib/chunks/lit/frag/shadowStandard.js
```diff
@@ -62,6 +62,26 @@ float getShadowSpotPCF3x3(SHADOWMAP_ACCEPT(shadowMap), vec3 shadowCoord, vec4 sh
     return _getShadowPCF3x3(SHADOWMAP_PASS(shadowMap), shadowCoord, shadowParams.xyz);
 }
 
+float _getShadowPCF1x1(SHADOWMAP_ACCEPT(shadowMap), vec3 shadowCoord, vec3 shadowParams) {
+    float z = shadowCoord.z;
+    vec2 uv = shadowCoord.xy * shadowParams.x; // 1 unit - 1 texel
+    float shadowMapSizeInv = 1.0 / shadowParams.x;
+    vec2 base_uv = floor(uv + 0.5);
```
mvaligursky, line 69: it seems you're forcing pixel snapping / no filtering here, which I don't think you should? This function should look similarly to what PCF1 shadow does for clustered lights, which is a lot cheaper:

```
    float getShadowSpotClusteredPCF1(SHADOWMAP_ACCEPT(shadowMap), vec3 shadowCoord, vec4 shadowParams) {
        return textureShadow(shadowMap, shadowCoord);
    }

```

ellthompson: I've updated this to be in line with clustered light shadows.

src/scene/shader-lib/programs/lit-shader.js
ellthompson: I've switched this over.

mvaligursky (approved):
Looking great!

PR #5332, opened 2023-05-15
Add async download of texture data from graphics memory

slimbuck:
This PR:
- organise pixel format metadata declarations slightly
- updates `Texture.lock()` to work correctly when specifying level and cubemap face
- add support for asynchronously reading texture data from graphics memory
- add a helper function for forcing lost context to `WebglGraphicsDevice` called `debugLoseContext()`.

### DIFF
--- a/src/platform/graphics/constants.js
+++ b/src/platform/graphics/constants.js
@@ -658,7 +658,7 @@ pixelFormatBlockSizes[PIXELFORMAT_ATC_RGBA] = 16;
 export const isCompressedPixelFormat = (format) => {
     return (format >= PIXELFORMAT_DXT1 && format <= PIXELFORMAT_DXT5) ||
            (format >= PIXELFORMAT_ETC1 && format <= PIXELFORMAT_ATC_RGBA);
-}
+};
 
 // get the pixel format array type
 export const getPixelFormatArrayType = (format) => {
@@ -674,7 +674,7 @@ export const getPixelFormatArrayType = (format) => {
             return Uint16Array;
         default:
             return Uint8Array;
-    };
+    }
 };
 
 /**
--- a/src/platform/graphics/constants.js
+++ b/src/platform/graphics/constants.js
@@ -618,11 +618,64 @@ export const PIXELFORMAT_ATC_RGBA = 30;
  */
 export const PIXELFORMAT_BGRA8 = 31;
 
+// map of engine PIXELFORMAT_*** enums to the pixel byte size
+export const pixelFormatByteSizes = [];
+pixelFormatByteSizes[PIXELFORMAT_A8] = 1;
+pixelFormatByteSizes[PIXELFORMAT_L8] = 1;
+pixelFormatByteSizes[PIXELFORMAT_LA8] = 2;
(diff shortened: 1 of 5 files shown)


src/platform/graphics/constants.js
```diff
@@ -618,11 +618,64 @@ export const PIXELFORMAT_ATC_RGBA = 30;
  */
 export const PIXELFORMAT_BGRA8 = 31;
 
+// map of engine PIXELFORMAT_*** enums to the pixel byte size
+export const pixelFormatByteSizes = [];
```
willeastcott, line 622: Kinda wish these were true Maps (you can call the Map constructor with an array for multiple entries). Not a blocker for merging this though, of course.

mvaligursky: arrays have slightly faster access though, so why map?

src/platform/graphics/webgl/webgl-graphics-device.js
```diff
@@ -2062,6 +2062,63 @@ class WebglGraphicsDevice extends GraphicsDevice {
...
+     * @param {number} h - The height of the rectangle, in pixels.
+     * @param {ArrayBufferView} pixels - The ArrayBufferView object that holds the returned pixel
+     * data.
+     * @ignore
+     */
+    async readPixelsAsync(x, y, w, h, pixels) {
+        const gl = this.gl;
+
+        const clientWaitAsync = (flags, interval_ms) => {
+            const sync = gl.fenceSync(gl.SYNC_GPU_COMMANDS_COMPLETE, 0);
```
mvaligursky, line 2081: this is WebGL2 only, do we need some GL1 fallback?

src/platform/graphics/texture.js
```diff
@@ -926,6 +883,40 @@ class Texture {
...
 
+    /**
+     * Download texture's top level data from graphics memory to local memory.
+     *
+     * @ignore
+     */
+    async downloadAsync() {
+        const promises = [];
+        for (let i = 0; i < (this.cubemap ? 6 : 1); i++) {
+            const renderTarget = new RenderTarget({
```
mvaligursky, line 894: I wonder if we can move more of this code to the device, to avoid having to import render target module into the texture, which sounds like a future circular reference.

Also, WebGPU will likely not need to create it.

src/platform/graphics/webgl/webgl-graphics-device.js
slimbuck: yeah good point. we should probably just return synchronously

src/platform/graphics/constants.js
slimbuck: Map works much better, thanks!

willeastcott: @mvaligursky Maps are a bit more deliberate and safer type-wise, in my experience.

src/platform/graphics/texture.js
slimbuck: This is probably ok for now since it's private API?

WebglGraphicsDevice already supported `readPixel` which WebGpu doesn't. So this function and those will need updating when you add WebGpu support.

So yeah, I'll be happy to leave this till you do the deeper refactor.

PR #5333, opened 2023-05-16
Refactor the shadow sampling coordinate code

GSterbrant:
This PR changes the way we calculate the dShadowCoord parameter used for shadow mapping for the different lights. Previously, we would provide a swath of different shadow sampling methods, such as with normal bias, without, for perspective, for orthogonal etc, and pick from them using very complicated conditions.

With this PR, all of the different modifications to the shadow sample are done in a single chunk, which is controlled by a set of preprocessor defines. Each lights have their own shadow coordinate evaluation function generated using this method.


src/scene/shader-lib/programs/lit-shader.js
```diff
@@ -1244,22 +1210,63 @@ class LitShader {
...
+                        // Create shadow coord sampler function for this light
+                        const coordCode = chunks.shadowSampleCoordPS;
+                        func.append(coordCode.replace("$LIGHT", i));
+
+                        func.append("#undef SHADOW_SAMPLE_NORMAL_OFFSET");
+                        func.append("#undef SHADOW_SAMPLE_ORTHO");
+                        func.append("#undef SHADOW_SAMPLE_SOURCE_ZBUFFER");
+                        func.append("#undef SHADOW_SAMPLE_POINT");
+
+                        // Make sure to undefine the shadow sampler defines
```
mvaligursky, line 1236: should this line be 4 lines higher, before those undefs?

mvaligursky (approved):
looks great, much simpler.

src/scene/shader-lib/programs/lit-shader.js
GSterbrant: The code is basically, set defines, generate chunk with defines set, unset the defines - for each light.

mvaligursky: sorry, I just meant to move the comment above those lines that do the under.

willeastcott (approved):

PR #5334, opened 2023-05-17
[Examples] Build update

ellthompson:
This PR removes the need to build each individual example iframe document using a separate build process from the main example application. Developing the examples browser now only requires the use of a single build script `npm run watch` which builds the entire application using rollup.

The rollup configuration now contains three builds:
- app: Builds the examples browser frontend application. This depends on a prebuilt examples manifest file to retrieve the list of categories / examples to display, as well as the TypeScript code used to run each example for the online code editor.
- iframe: Bundles all of the dependencies necessary to run the example iframe. This needs to be able to run independently from the main frontend application. This bundle includes the PlayCanvas dependency from the local engine build directory so any edits to your local engine build will trigger a build of this iframe.
- examples: Bundles all of the example classes. This is included on the iframe page and the required example is selected from this bundle using url query parameters.


mvaligursky:
I had a go. Overall a great improvement! It's so much easier to use with just a single command line and debug engine! Love it.

Small improvement suggestions:
- could 'npm run watch' start he serving as well, to avoid having to start it separately?
- when I visit 'http://localhost:5000/iframe/' I get 'Uncaught No example specified, showing a list of all examples instead!' - this should be handled / no error displayed?
- when I modify script (for example planar-renderer.js while using reflection-planar example), the changes are reflected in the example, but nothing is printed in console window to show me something gets updated. This is a small possible improvement to log something, so I don't need to wonder if anything rebuilt. When I change example code, or engine builds, I can see dist/examples or dist/iframe is updated .. this is great, and I'd like this for scripts as well please. If all it says is 'asset copied' or whatever, that'd be great. If can even print the names of those, that'd be even better.

GSterbrant:
Absolutely love it!

GSterbrant:
@mvaligursky You can just run `develop` right?

@ellthompson If it is at all possible to get a time stamp of when the build was done, that would help with @mvaligursky problem of understanding if changes have been recognized by the watch 🙂

ellthompson:
I've included a timestamp for all builds:
image

Also all static directories / files are now watched and copied on change.

ellthompson:
The iframe page no longer throws errors when displaying the list of all examples.

mvaligursky (approved):

Issue #4230, opened 2022-05-05
Implement support for multi render target (MRT)


mvaligursky:
Currently, only a single color and a single depth buffer is supported. This should be extended to multiple colors buffers, which is supported by WebGl 2 and WebGPU.

Additionally, there is a WebGL1 extension: https://registry.khronos.org/webgl/extensions/WEBGL_draw_buffers/

PR #5336, opened 2023-05-17
Initial MRT implementation for WebGPU

mvaligursky:
partially implements: https://github.com/playcanvas/engine/issues/4230

Initial multi render target rendering using WebGPU. This allows a single draw call to be rendered into more than one color buffer.

## New API:

- **GraphicsDevice. supportsMrt** - true if MRT is supported, which is WebGPU only for now, soon WebGL2 and WebGL1 with an extension.

- options passed to RenderTarget constructor can contain: **options.colorBuffers** - The textures that this render target will treat as a rendering surfaces. If this option is set, the colorBuffer option is ignored.

- **RenderTarget.getColorBuffer(index)** - returns color buffer with index. This is in addition to existing **RenderTarget.colorBuffer**, which returns buffer with index 0.

- new shader chunk **OutputPS**, empty by default. Override allows output to additional color buffers.

## Basic usage:

1. set up chunk that uses a shader pass to output data into additional color outputs:
```
material.chunks.outputPS = `
  #ifdef MYMRT_PASS
      // optionally override value in pcFragColor0 if needed, otherwise outputs the standard shader output
      pcFragColor1 = vec4(litShaderArgs.worldNormal * 0.5 + 0.5, 1.0);
  #endif
`;
```

2. create render target with multiple buffers
```
  const renderTarget = new pc.RenderTarget({
      colorBuffers: [texture0, texture1],
      depth: true
  });
```

3. enable shader pass that matches the shader chunk on a camera using the render target:
```
    textureCamera.camera.setShaderPass('MyMRT');
```

## Render pass tracing

Screenshot 2023-05-17 at 14 13 15

## Example: MRT

https://github.com/playcanvas/engine/assets/59932779/87254539-08a8-4039-88a3-72701829daec

### DIFF
--- a/examples/src/examples/graphics/mrt.tsx
+++ b/examples/src/examples/graphics/mrt.tsx
@@ -1,0 +1,201 @@
+import * as pc from '../../../../';
+
+class MrtExample {
+    static CATEGORY = 'Graphics';
+    static NAME = 'Multi Render Targets';
+    static WEBGPU_ENABLED = true;
+
+    static FILES = {
+
+        // shader chunk which outputs to multiple render targets
+        // Note: gl_FragColor is not modified, and so the forward pass output is used for target 0
+        'output.frag': /* glsl */`
+            #ifdef MYMRT_PASS
+                // output world normal to target 1
+                gl_FragColor1 = vec4(litShaderArgs.worldNormal * 0.5 + 0.5, 1.0);
+
+                // output gloss to target 2
+                gl_FragColor2 = vec4(vec3(litShaderArgs.gloss) , 1.0);
+            #endif
+        `
+    };
+
+    example(canvas: HTMLCanvasElement, deviceType: string, files: { 'output.frag': string, }): void {
+
+        // set up and load draco module, as the glb we load is draco compressed
+        pc.WasmModule.setConfig('DracoDecoderModule', {
+            glueUrl: '/static/lib/draco/draco.wasm.js',
+            wasmUrl: '/static/lib/draco/draco.wasm.wasm',
+            fallbackUrl: '/static/lib/draco/draco.js'
+        });
+
+        const assets = {
+            'board': new pc.Asset('statue', 'container', { url: '/static/assets/models/chess-board.glb' }),
+            helipad: new pc.Asset('helipad-env-atlas', 'texture', { url: '/static/assets/cubemaps/helipad-env-atlas.png' }, { type: pc.TEXTURETYPE_RGBP, mipmaps: false })
+        };
+
+        const gfxOptions = {
+            deviceTypes: [deviceType],
+            glslangUrl: '/static/lib/glslang/glslang.js',
+            twgslUrl: '/static/lib/twgsl/twgsl.js'
+        };
+
+        pc.createGraphicsDevice(canvas, gfxOptions).then((device: pc.GraphicsDevice) => {
+
+            const createOptions = new pc.AppOptions();
+            createOptions.graphicsDevice = device;
+            createOptions.keyboard = new pc.Keyboard(document.body);
+
+            createOptions.componentSystems = [
+                // @ts-ignore
+                pc.RenderComponentSystem,
+                // @ts-ignore
+                pc.CameraComponentSystem,
+                // @ts-ignore
+                pc.LightComponentSystem,
+                // @ts-ignore
+                pc.ScriptComponentSystem,
+                // @ts-ignore
+                pc.ScreenComponentSystem,
+                // @ts-ignore
+                pc.ElementComponentSystem
+            ];
+            createOptions.resourceHandlers = [
+                // @ts-ignore
+                pc.ScriptHandler,
+                // @ts-ignore
+                pc.TextureHandler,
+                // @ts-ignore
+                pc.ContainerHandler,
+                // @ts-ignore
+                pc.FontHandler
+            ];
+
+            const app = new pc.AppBase(canvas);
+            app.init(createOptions);
+
+            // Set the canvas to fill the window and automatically change resolution to be the same as the canvas size
+            app.setCanvasFillMode(pc.FILLMODE_FILL_WINDOW);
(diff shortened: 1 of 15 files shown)


src/platform/graphics/shader-chunks/frag/webgpu.js
```diff
@@ -4,7 +4,22 @@ export default /* glsl */`
...
+layout(location = 1) out highp vec4 pc_fragColor1;
+layout(location = 2) out highp vec4 pc_fragColor2;
+layout(location = 3) out highp vec4 pc_fragColor3;
+layout(location = 4) out highp vec4 pc_fragColor4;
+layout(location = 5) out highp vec4 pc_fragColor5;
+layout(location = 6) out highp vec4 pc_fragColor6;
+layout(location = 7) out highp vec4 pc_fragColor7;
+
 #define gl_FragColor pc_fragColor
+#define gl_FragColor1 pc_fragColor1
```
GSterbrant, line 16: I would suggest calling this render target instead of fragColor. Makes it more consistent with what we call the API on the CPU side.

GSterbrant: So, pc_renderTarget, pc_renderTarget1, etc.

examples/src/examples/graphics/mrt.tsx
```diff
@@ -0,0 +1,201 @@
...
+    static WEBGPU_ENABLED = true;
+
+    static FILES = {
+
+        // shader chunk which outputs to multiple render targets
+        // Note: gl_FragColor is not modified, and so the forward pass output is used for target 0
+        'output.frag': /* glsl */`
+            #ifdef MYMRT_PASS
+                // output world normal to target 1
+                gl_FragColor1 = vec4(litShaderArgs.worldNormal * 0.5 + 0.5, 1.0);
```
GSterbrant, line 15: Should we use the new pc alias for frag color here maybe to illustrate how it should be done?

src/platform/graphics/shader-chunks/frag/webgpu.js
mvaligursky: I would not want to call:
- output 0: **gl_FragColor** 
- output 1: pc_renderTarget1

especially since all of those are the same render target. Just different color buffers.

I prefer consistent:
```
gl_FragColor
gl_FragColor1
...
```

examples/src/examples/graphics/mrt.tsx
mvaligursky: it uses new `gl_FragColor1` here?

src/platform/graphics/shader-chunks/frag/webgpu.js
GSterbrant: Wouldn't it make more sense to get away from using gl_ prefixes and use our own, should make it easier to translate between glsl and wgsl for example.

GSterbrant: And this PR adds support for multiple render targets. How come it makes sense to have multiple render targets yet not refer to them as render targets?

Just not sure about the naming.

GSterbrant (approved):

src/platform/graphics/shader-chunks/frag/webgpu.js
mvaligursky: We could move away, but we don't really want to rename existing gl_FragColor, as lots of shaders use it.
And naming 0th differently than following is not a great option either.
We could perhaps keep the old name and define a set of new names?

mvaligursky: Lets discuss in the afternoon.

mvaligursky: renamed to pcFragColor0 .. 7

willeastcott:
Do we need public API that specifies the max number of draw buffers supported?

mvaligursky:
> Do we need public API that specifies the max number of draw buffers supported? Presumably this can vary.

Yes, definitely, I added it to my list and will add it in the WebGL implementation of MRT soon.

src/platform/graphics/webgpu/webgpu-render-target.js
```diff
@@ -1,6 +1,30 @@
...
 
+/**
+ * Private class storing info about color buffer.
+ *
+ * @ignore
+ */
+class ColorAttachment {
+    /**
+     * @type {GPUTextureFormat}
+     * @private
```
willeastcott, line 12: I don't think you meant to mark this a private.

src/platform/graphics/webgpu/webgpu-render-target.js
```diff
@@ -1,6 +1,30 @@
...
+     */
+    format;
+
+    /**
+     * @type {GPUTexture}
+     * @private
```
willeastcott, line 18: I don't think you meant to mark this as private.

src/platform/graphics/webgpu/webgpu-render-target.js
mvaligursky: That is the only way to avoid type errors, as WebGPU types cannot be exported. So I added those as workarounds really.

willeastcott (approved):

src/platform/graphics/webgpu/webgpu-render-target.js
willeastcott: Ooooh, OK.

src/platform/graphics/webgpu/webgpu-render-target.js
willeastcott: Ooooh, OK.

querielo:
@mvaligursky  how do you visualize "Render pass tracing"?

mvaligursky:
> @mvaligursky how do you visualize "Render pass tracing"?

Check the Tracking docs, I think this is what you're after?
https://developer.playcanvas.com/en/api/pc.Tracing.html

PR #5338, opened 2023-05-19
PCSS shadows

### LINKED ISSUE #2815: Soft Area light shadows
Implement soft area light shadows. Possible implementation is PCSS:
https://www.gamedev.net/blogs/entry/2265228-effect-area-light-shadows-pt-1-pcss/

This could be used at least for the lightmap baking, and during runtime on higher end platforms.

image

GSterbrant:
Fixes #2815 

This PR adds support for PCSS shadows as well as a new example highlighting contact hardening shadows. 

It also adds support for using R32F shadow maps for omni lights with clustered lighting disabled and when using PCSS.

Support for PCSS with clustered lighting is not handled by this PR, the main reason being that we need to pass several new arguments using the clustering textures which I would like to reduce before being completely done with this feature. 

image


src/framework/components/light/component.js
```diff
@@ -127,6 +129,7 @@ const _lightPropsDefault = [];
  * OES_texture_float extension. Falls back to {@link SHADOW_VSM16}, if not supported.
  * - {@link SHADOW_PCF5}: Render depth buffer only, can be used for hardware-accelerated PCF 5x5
  * sampling. Requires WebGL2. Falls back to {@link SHADOW_PCF3} on WebGL 1.0.
+ * - {@link SHADOW_PCSS}: Render depth as color, and use the software sampled PCSS method for shadows.
```
willeastcott, line 132:
```suggestion
 * - {@link SHADOW_PCSS}: Render depth as color and use the software sampled PCSS method for shadows.
```

src/scene/light.js
```diff
@@ -97,7 +97,7 @@ class LightRenderData {
                 return rt.colorBuffer;
             }
 
-            return light._isPcf && light.device.supportsDepthShadow ? rt.depthBuffer : rt.colorBuffer;
+            return (light._isPcf) && light.device.supportsDepthShadow ? rt.depthBuffer : rt.colorBuffer;
```
willeastcott, line 100: Why the brackets?

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
```diff
@@ -0,0 +1,185 @@
...
+    float weight = sampleIndex / count;
+    float radius = sqrt(1.0 - weight * weight);
+    const float GoldenAngle = 2.4;
+
+    float theta = GoldenAngle * sampleIndex + phi;
+
+    return vec3(cos(theta) * radius, weight, sin(theta) * radius);
+}
+
+float GradientNoise(vec2 screenPos) {
```
willeastcott, line 30: Do we have GLSL coding standard/conventions? Should function names be capitalized or camel-case? cc @mvaligursky

src/scene/shader-lib/programs/lit-shader.js
```diff
@@ -507,15 +508,26 @@ class LitShader {
...
         const applySlopeScaleBias = !device.webgl2 && device.extStandardDerivatives && !device.isWebGPU;
+        const customDepth = (isVsm && lightType !== LIGHTTYPE_DIRECTIONAL) || lightType === LIGHTTYPE_OMNI;
 
-        if (lightType === LIGHTTYPE_OMNI || (isVsm && lightType !== LIGHTTYPE_DIRECTIONAL)) {
+        if (customDepth) {
             code += "    float depth = min(distance(view_position, vPositionW) / light_radius, 0.99999);\n";
+        } else if (shadowType === SHADOW_PCSS) {
+            code += "    float depth = linearizeDepth(gl_FragCoord.z, camera_params);\n";
+            if (usePackedDepth) {
+                // After linearising, make sure to pack the range between [0..1]
```
willeastcott, line 528:
```suggestion
                // After linearizing, make sure to pack the range between [0..1]
```

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
mvaligursky: I believe it's the same as Js code, so `gradientNoise`.

src/scene/light.js
GSterbrant: Fixed

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
GSterbrant: Fixed

src/scene/shader-lib/programs/lit-shader.js
GSterbrant: Fixed

examples/src/examples/graphics/contact-hardening-shadows.tsx
```diff
@@ -0,0 +1,367 @@
...
+                app.start();
+
+                app.scene.toneMapping = pc.TONEMAP_ACES;
+                app.scene.skyboxMip = 1;
+                app.scene.ambientLight.set(0, 0, 0);
+                app.scene.ambientLuminance = 0;
+                app.scene.setSkybox(assets.helipad.resources);
+
+                // enable area lights which are disabled by default for clustered lighting
+                app.scene.lighting.areaLightsEnabled = true;
```
mvaligursky, line 115: probably not needed line

mvaligursky: and few app.scene.lighting below as well, those are only used by clustered

examples/src/examples/graphics/contact-hardening-shadows.tsx
```diff
@@ -0,0 +1,367 @@
...
+                app.scene.lighting.shadowsEnabled = true;
+                app.scene.lighting.shadowAtlasResolution = 2048;
+                app.scene.skyboxIntensity = 0.1;
+
+                // set the loaded area light LUT data
+                const luts = assets.luts.resource;
+                app.setAreaLightLuts(luts.LTC_MAT_1, luts.LTC_MAT_2);
+
+                const planeMaterial = new pc.StandardMaterial();
+                //planeMaterial.diffuseMap = assets.color.resource;
```
mvaligursky, line 127: remove comments

examples/src/examples/graphics/contact-hardening-shadows.tsx
```diff
@@ -0,0 +1,367 @@
...
+                    time += dt;
+                    timeDiff += dt;
+
+                    if (data.get('script.cycle')) {
+                        if ((timeDiff / 5) > 1) {
+                            index = (index + 1) % 3;
+                            timeDiff = 0;
+                        }
+    
+                        areaLight.enabled = false;
```
mvaligursky, line 325: we do not want disable lights every frame, just to enable one later, as that is pretty expensive under the hood. Ideally this is something like:
```
areaLight.enabled = index === 0;
directionalLight.enabled = index === 1;
lightOmni.enabled = index === 2;
```

I know it's just an example, but we should show how to do little things properly as well.

src/scene/constants.js
```diff
@@ -287,8 +287,15 @@ export const SHADOW_PCF5 = 4;
...
+/**
+ * Render depth as color for PCSS software filtering.
+ *
+ * @type {number}
+ */
+export const SHADOW_PCSS = 6;
+
 // non-public: number of supported depth shadow modes
-export const SHADOW_COUNT = 6;
+export const SHADOW_COUNT = 7;
```
mvaligursky, line 298: just looking at the codebase, this is no longer needed, I missed its removal when I refactored it, could you please remove SHADOW_COUNT completely.

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
```diff
@@ -0,0 +1,185 @@
+export default /* glsl */`
```
mvaligursky, line 1: please add references to what this is based on as comments

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
```diff
@@ -0,0 +1,185 @@
...
+    const float GoldenAngle = 2.4;
+    float r = sqrt(sampleIndex + 0.5) / sqrt(count);
+
+    float theta = sampleIndex * GoldenAngle + phi;
+
+    float sine, cosine;
+    sine = sin(theta);
+    cosine = cos(theta);
+
+    return vec2(r * cosine, r * sine);
```
mvaligursky, line 17: I'd compact this to a single line

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
```diff
@@ -0,0 +1,185 @@
+export default /* glsl */`
+
+
+#define PCSS_SAMPLE_COUNT 16
```
mvaligursky, line 4: considering the number of samples is a hardcoded define, should we specify vogel disk and sphere offsets as inline constants to avoid their evaluation?

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
```diff
@@ -0,0 +1,185 @@
...
+float PCSSBlockerDistance(TEXTURE_ACCEPT(shadowMap), vec2 sampleCoords[PCSS_SAMPLE_COUNT], vec2 shadowCoords, float searchSize, float z) {
+
+    float blockers = 0.0;
+    float averageBlocker = 0.0;
+    for (int i = 0; i < PCSS_SAMPLE_COUNT; i++) {
+        vec2 offset = sampleCoords[i] * searchSize;
+        vec2 sampleUV = shadowCoords + offset;
+
+    #ifdef GL2
+        float blocker = texture2D(shadowMap, sampleUV).r;
```
mvaligursky, line 60: see if you can use textureShadow instead of texture2D

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
mvaligursky: I see now .. it depends on screen pixel noise and so it can only be partially precomputed. Perhaps even that is worth it?

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
```diff
@@ -0,0 +1,185 @@
...
+#endif
+
+    vec2 samplePoints[PCSS_SAMPLE_COUNT];
+    float noise = GradientNoise( gl_FragCoord.xy ) * 2.0 * PI;
+    for (int i = 0; i < PCSS_SAMPLE_COUNT; i++) {
+        samplePoints[i] = VogelDisk(float(i), float(PCSS_SAMPLE_COUNT), noise);
+    }
+
+    // Calculate the ratio of FOV between 45.0 degrees (tan(45) == 1) and the FOV of the camera    
+    float fovRatioAtDepth = receiverDepth / (cameraParams.x * receiverDepth);
```
mvaligursky, line 88: don't `receiverDepth`'s cancel each other here? So the result is `1/cameraParams.x`

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
mvaligursky: is it fixed? It's still GradientNoise and not gradientNoise.

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
```diff
@@ -0,0 +1,185 @@
...
+    } else {
+
+        float filterRadius = ((receiverDepth - averageBlocker) / averageBlocker) * lightSize * oneOverShadowMapSize * fovRatioAtDepth;
+
+        float shadow = 0.0;
+        float noise = GradientNoise(gl_FragCoord.xy);
+
+        for (int i = 0; i < PCSS_SAMPLE_COUNT; i ++)
+        {
+            vec2 sampleUV = samplePoints[i] * (filterRadius);
```
mvaligursky, line 101: why the brackets around (filterRadius)?

src/scene/shader-lib/chunks/chunks.js
```diff
@@ -363,6 +366,7 @@ const shaderChunks = {
     shadowEVSMPS,
     shadowEVSMnPS,
     shadowSampleCoordPS,
+    shadowPCSSPS,
```
mvaligursky, line 369: alphabetic sorting

src/scene/shader-lib/chunks/common/frag/linearizeDepth.js
```diff
@@ -0,0 +1,32 @@
+export default /* glsl */`
```
mvaligursky, line 1: a lot of this is a duplication from the screenDepth.js. Should we find a solution without the duplication?

src/framework/components/light/component.js
```diff
@@ -87,6 +87,8 @@ const _lightPropsDefault = [];
  * angle is specified in degrees. Affects spot lights only. Defaults to 40.
  * @property {number} outerConeAngle The angle at which the spotlight cone has faded to nothing.
  * The angle is specified in degrees. Affects spot lights only. Defaults to 45.
+ * @property {number} lightSize Controls the size of the light source. Used only for PCSS shadows to
+ * estimate penumra size. Defaults to 0.001.
```
mvaligursky, line 91: 1. in the light code I see `this._lightSize = 10.0;` ... where does the 0.001 come from? What units is this in? Is it an angle, or a penumbra size at some unit distance or something?
2. I wonder if this should be called something else than the lightSize, considering this is used to estimate penumbra size only. Maybe even 'penumbraSize'? This comes back to the unit this is specified in.

src/scene/light.js
```diff
@@ -334,7 +335,7 @@ class Light {
             value = SHADOW_VSM8;
 
         this._isVsm = value >= SHADOW_VSM8 && value <= SHADOW_VSM32;
-        this._isPcf = value === SHADOW_PCF1 || value === SHADOW_PCF3 || value === SHADOW_PCF5;
+        this._isPcf = value === SHADOW_PCF5 || value === SHADOW_PCF3;
```
mvaligursky, line 338: is the PCF1 handling removed here correctly?

src/scene/renderer/forward-renderer.js
```diff
@@ -229,6 +233,14 @@ class ForwardRenderer extends Renderer {
                 this.shadowCascadeDistancesId[cnt].setValue(directional._shadowCascadeDistances);
                 this.shadowCascadeCountId[cnt].setValue(directional.numCascades);
                 this.lightShadowIntensity[cnt].setValue(directional.shadowIntensity);
+                this.lightSizeId[cnt].setValue(directional.lightSize);
+
+                const cameraParams = new Float32Array(4);
```
mvaligursky, line 238: I'd like to avoid the allocation in all 3 cases, and store the array / float32array on the light itself, similarly to how _shadowRenderParams is handled.

mvaligursky:
Could this be optimized?
15.1 MB examples/assets/models/old_tree.glb

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
GSterbrant: Sure but then we lose the ability to change the number of samples.

mvaligursky: in that case, we should have it specified in JS code, which would generate a define for shader + appropriate look up tables. But avoid doing this per pixel.

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
GSterbrant: I need the sampler to be a sampler2D and not samplerShadow, since I need the actual depth and not the binary result from a comparison sample.

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
GSterbrant: True, simplified it.

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
GSterbrant: Deleted.

src/scene/shader-lib/chunks/common/frag/linearizeDepth.js
GSterbrant: I can't change screenDepth without breaking other projects that may rely on it. I thought about moving all the linearization code to this chunk, and then have screenDepth only be specifically for getting depth at a pixel coordinate, but then I realized a lot of people might be using screenDepthPS in their own chunks to get this functionality and it would break things for them.

src/scene/renderer/forward-renderer.js
GSterbrant: Done

src/framework/components/light/component.js
GSterbrant: This is a really good question. The size is supposedly area size, which means it should ideally be (and can be made to be) a 2d vector denoting both height and width. It doesn't really describe the penumbra size therefore. This parameter could also be inferred from area lights in the scenarios where we use them.

src/scene/constants.js
GSterbrant: Fixed

examples/src/examples/graphics/contact-hardening-shadows.tsx
GSterbrant: Done.

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
mvaligursky: yeah that's what I thought, but I'll leave the decision to you if it's worth the gain here.

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
GSterbrant: Done.

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
GSterbrant: Yeah, did a partial precalculation which saves us a few sqrt.

examples/src/examples/graphics/contact-hardening-shadows.tsx
GSterbrant: Done.

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
GSterbrant: Done.

src/framework/components/light/component.js
GSterbrant: So the question is, do we change this to be a 2D vector for the width and height or do we keep it as a uniform value? I think the documentation explains what this field is supposed to do.

examples/src/examples/graphics/contact-hardening-shadows.tsx
```diff
@@ -0,0 +1,358 @@
...
+        pc.WasmModule.getInstance('DracoDecoderModule', demo);
+
+
+        function demo() {
+
+            const assets = {
+                orbitCamera: new pc.Asset('script', 'script', { url: '/static/scripts/camera/orbit-camera.js' }),
+                helipad: new pc.Asset('helipad-env-atlas', 'texture', { url: '/static/assets/cubemaps/helipad-env-atlas.png' }, { type: pc.TEXTURETYPE_RGBP }),
+                cube: new pc.Asset('cube', 'container', { url: '/static/assets/models/playcanvas-cube.glb' }),
+                color: new pc.Asset('color', 'texture', { url: '/static/assets/textures/seaside-rocks01-color.jpg' }),
```
mvaligursky, line 65: can you just double check all these textures are used

GSterbrant: Not needed, deleted.

src/scene/shader-lib/chunks/chunks.js
mvaligursky: move one line up please

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
```diff
@@ -0,0 +1,190 @@
...
+/**
+ * PCSS is a shadow sampling method that provides contact hardening soft shadows. 
+ * Based on: 
+ * - https://www.gamedev.net/tutorials/programming/graphics/effect-area-light-shadows-part-1-pcss-r4971/
+ * - https://github.com/pboechat/PCSS 
+ */
+
+
+#define PCSS_SAMPLE_COUNT 16
+uniform float pcssDiskSamples[PCSS_SAMPLE_COUNT];
```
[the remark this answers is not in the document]
GSterbrant: Okay, I can fix it. But PCSS won't work with WebGPU yet anyways as only clustered is supported. The follow-up PR that does the clustered support would also do the WebGPU support I suppose.

src/framework/components/light/component.js
```diff
@@ -87,6 +87,9 @@ const _lightPropsDefault = [];
  * angle is specified in degrees. Affects spot lights only. Defaults to 40.
  * @property {number} outerConeAngle The angle at which the spotlight cone has faded to nothing.
  * The angle is specified in degrees. Affects spot lights only. Defaults to 45.
+ * @property {number} lightSize Size of the light used for contact hardening shadows. For area lights
+ * acts as a size multiplier with the area light dimensions. For punctual and directional lights
+ * acts as the actual size of the light. Defaults to 10.0.
```
mvaligursky, line 92: Should this perhaps default to 1? Multiplier of 10 for area lights seems strange.

GSterbrant: Good point.

src/framework/components/light/component.js
```diff
@@ -549,6 +561,9 @@ function _defineProps() {
             }
         }
     });
+
+    _lightProps.push("lightSize");
+    _lightPropsDefault.push(10);
```
mvaligursky, line 566: There is a default value here.

src/scene/light.js
```diff
@@ -185,6 +186,7 @@ class Light {
         this._normalOffsetBias = 0.0;
         this.shadowUpdateMode = SHADOWUPDATE_REALTIME;
         this.shadowUpdateOverrides = null;
+        this._lightSize = 10.0;
```
mvaligursky, line 189: There is also a default value here.

mvaligursky:
@GSterbrant - you only changed the default value in the comment, but not the actual values in the code - that needs changing?

PR #5341, opened 2023-05-20
Added ScriptType listen and unlisten API for easier Event subscription handling

yaustar:
Fixes https://github.com/playcanvas/engine/issues/4910

Test project with the debug version of the engine already included. It has buttons to change scene (destroying entities) and to disable/enable entities with scripts that are listening for mouse and touch events.

https://playcanvas.com/project/1078046/overview/listen-api-test-project

PR includes automated tests as well

### DIFF
--- a/src/framework/script/script-type.js
+++ b/src/framework/script/script-type.js
@@ -3,6 +3,7 @@ import { EventHandler } from '../../core/event-handler.js';
 
 import { SCRIPT_INITIALIZE, SCRIPT_POST_INITIALIZE } from './constants.js';
 import { ScriptAttributes } from './script-attributes.js';
+import { c } from "sinon/lib/sinon/spy-formatters.js";
 
 const funcNameRegex = new RegExp('^\\s*function(?:\\s|\\s*\\/\\*.*\\*\\/\\s*)+([^\\(\\s\\/]*)\\s*');
 
@@ -58,6 +59,12 @@ class ScriptType extends EventHandler {
 
     /** @private */
     _postInitialized;
+
+    /**
+     * @type {any[]}
+     * @private
+     */
+    _listeners;
 
     /** @private */
     __destroyed;
@@ -207,6 +214,12 @@ class ScriptType extends EventHandler {
         this.fire(this.enabled ? 'enable' : 'disable');
         this.fire('state', this.enabled);
 
+        // Subscribe or unsubscribe events based on enabled
+        const offOrOn = this.enabled ? 'on' : 'off';
+        this._listeners.forEach((l) => {
+            l.eventHandler[offOrOn](l.name, l.callback, l.scope);
+        });
+
         // initialize script if not initialized yet and script is enabled
         if (!this._initialized && this.enabled) {
             this._initialized = true;
@@ -235,6 +248,41 @@ class ScriptType extends EventHandler {
     }
 
     /**
+     * Attach an event handler to an event.
+     *
+     * @param {EventHandler} eventHandler - Name of the event to bind the callback to.
+     * @param {string} name - Name of the event to bind the callback to.
+     * @param {HandleEventCallback} callback - Function that is called when event is fired. Note
+     * the callback is limited to 8 arguments.
+     * @param {object} [scope] - Object to use as 'this' when the event is fired, defaults to
+     * current this.
+     * @returns {ScriptType} Self for chaining.
+     * @example
+     * this.listen(eventHandler, 'test', function (a, b) {
+     *     console.log(a + b);
+     * }, this);
+     * eventHandler.fire('test', 1, 2); // prints 3 to the console
+     */
+    listen(eventHandler, name, callback, scope) {
+        this._listeners.push({
+            eventHandler,
+            name,
+            callback,
+            scope
+        });
+
+        if (this.enabled) {
+            eventHandler.on(name, callback, scope);
+        }
+
+        return this;
+    }
+
+    unlisten(eventHandler, name, callback, scope) {
+        return this;
+    }
+
+    /**
      * @param {{entity: import('../entity.js').Entity, app: import('../app-base.js').AppBase}} args -
      * The entity and app.
      * @private
@@ -248,6 +296,8 @@ class ScriptType extends EventHandler {
 
         this._enabled = typeof args.enabled === 'boolean' ? args.enabled : true;
         this._enabledOld = this.enabled;
+
+        this._listeners = [];
 
         this.__destroyed = false;
         this.__attributes = { };
--- a/src/framework/script/script-type.js
+++ b/src/framework/script/script-type.js
@@ -3,13 +3,13 @@ import { EventHandler } from '../../core/event-handler.js';
 
 import { SCRIPT_INITIALIZE, SCRIPT_POST_INITIALIZE } from './constants.js';
 import { ScriptAttributes } from './script-attributes.js';
-import { c } from "sinon/lib/sinon/spy-formatters.js";
 
 const funcNameRegex = new RegExp('^\\s*function(?:\\s|\\s*\\/\\*.*\\*\\/\\s*)+([^\\(\\s\\/]*)\\s*');
 
 /**
  * EventListener object used for storing what events on EventHandlers the ScriptType is listening for.
- * @typedef {Object} EventListener
+ *
+ * @typedef {object} EventListener
  * @property {EventHandler} eventHandler - EventHandler to listen for events on.
  * @property {string} name - Name of the event to bind the callback to.
  * @property {HandleEventCallback} callback - Function that is called when event is fired. Note
--- a/src/framework/script/script-type.js
+++ b/src/framework/script/script-type.js
@@ -248,9 +248,10 @@ class ScriptType extends EventHandler {
     }
 
     /**
-     * Attach an event handler to an event.
+     * Listen for an event on an EventHandler. The ScriptType will add and remove the event listeners on the
+     * EventHandler automatically when the ScriptType is enabled, disabled and destroyed.
      *
-     * @param {EventHandler} eventHandler - Name of the event to bind the callback to.
+     * @param {EventHandler} eventHandler - EventHandler to listen for events on.
      * @param {string} name - Name of the event to bind the callback to.
      * @param {HandleEventCallback} callback - Function that is called when event is fired. Note
      * the callback is limited to 8 arguments.
@@ -262,6 +263,7 @@ class ScriptType extends EventHandler {
      *     console.log(a + b);
      * }, this);
      * eventHandler.fire('test', 1, 2); // prints 3 to the console
+     * @see {@link unlisten} to remove listeners to an event on an EventHandler.
      */
     listen(eventHandler, name, callback, scope) {
         this._listeners.push({
@@ -278,7 +280,35 @@ class ScriptType extends EventHandler {
         return this;
     }
 
+    /**
+     * Removes a listener for an event on an EventHandler that was added by {@link listen}.
+     *
+     * @param {EventHandler} eventHandler - EventHandler that was originally listened to.
+     * @param {string} name - Name of the event that was originally listened to.
+     * @param {HandleEventCallback} callback - Function that was used as the callback when the event was originally
+     * listened to.
+     * @param {object} [scope] - Object that was used as the scope when the event was originally listened to
+     * @returns {ScriptType} Self for chaining.
+     * @example
+     * this.unlisten(eventHandler, 'test', callbackFunction, this);
+     * @see {@link listen} to listen to events on an EventHandler.
+     */
     unlisten(eventHandler, name, callback, scope) {
+        let found = false;
+        for (let i = 0; i < this._listeners.length; ++i) {
+            const l = this._listeners[i];
+            if (l.eventHandler === eventHandler && l.name === name && l.callback === callback && l.scope === scope) {
+                this._listeners.splice(i, 1);
+                eventHandler.off(name, callback, scope);
+                found = true;
+                break;
+            }
+        }
+
+        if (!found) {
+            Debug.warn(`${this.__name}: Cannot find listener to remove for event ${name}. Please check where you originally listened for the event.`);
+        }
+
         return this;
     }
 
--- a/src/framework/script/script-type.js
+++ b/src/framework/script/script-type.js
@@ -8,6 +8,16 @@ import { c } from "sinon/lib/sinon/spy-formatters.js";
 const funcNameRegex = new RegExp('^\\s*function(?:\\s|\\s*\\/\\*.*\\*\\/\\s*)+([^\\(\\s\\/]*)\\s*');
 
 /**
+ * EventListener object used for storing what events on EventHandlers the ScriptType is listening for.
+ * @typedef {Object} EventListener
+ * @property {EventHandler} eventHandler - EventHandler to listen for events on.
+ * @property {string} name - Name of the event to bind the callback to.
+ * @property {HandleEventCallback} callback - Function that is called when event is fired. Note
+ * the callback is limited to 8 arguments.
+ * @property {object} [scope] - Object to use as 'this' when the event is fired.
+ */
+
+/**
  * Represents the type of a script. It is returned by {@link createScript}. Also referred to as
  * Script Type.
  *
@@ -61,7 +71,7 @@ class ScriptType extends EventHandler {
     _postInitialized;
 
     /**
-     * @type {any[]}
+     * @type {EventListener[]}
      * @private
      */
     _listeners;
@@ -255,15 +265,11 @@ class ScriptType extends EventHandler {
      * @param {string} name - Name of the event to bind the callback to.
      * @param {HandleEventCallback} callback - Function that is called when event is fired. Note
      * the callback is limited to 8 arguments.
-     * @param {object} [scope] - Object to use as 'this' when the event is fired, defaults to
-     * current this.
+     * @param {object} [scope] - Object to use as 'this' when the event is fired.
      * @returns {ScriptType} Self for chaining.
      * @example
-     * this.listen(eventHandler, 'test', function (a, b) {
-     *     console.log(a + b);
-     * }, this);
-     * eventHandler.fire('test', 1, 2); // prints 3 to the console
-     * @see {@link unlisten} to remove listeners to an event on an EventHandler.
+     * this.listen(this.app.mouse, 'mousemove', this.onMouseMove, this);
+     * @see {@link ScriptType#unlisten} to remove listeners to an event on an EventHandler.
      */
     listen(eventHandler, name, callback, scope) {
         this._listeners.push({
@@ -281,7 +287,7 @@ class ScriptType extends EventHandler {
     }
 
     /**
-     * Removes a listener for an event on an EventHandler that was added by {@link listen}.
+     * Removes a listener for an event on an EventHandler that was added by {@link ScriptType#listen}.
      *
      * @param {EventHandler} eventHandler - EventHandler that was originally listened to.
      * @param {string} name - Name of the event that was originally listened to.
@@ -290,8 +296,8 @@ class ScriptType extends EventHandler {
      * @param {object} [scope] - Object that was used as the scope when the event was originally listened to
      * @returns {ScriptType} Self for chaining.
      * @example
-     * this.unlisten(eventHandler, 'test', callbackFunction, this);
-     * @see {@link listen} to listen to events on an EventHandler.
+     * this.unlisten(this.app.mouse, 'mousemove', this.onMouseMove, this);
+     * @see {@link ScriptType#listen} to listen to events on an EventHandler.
      */
     unlisten(eventHandler, name, callback, scope) {
         let found = false;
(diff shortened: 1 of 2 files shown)


src/framework/script/script-type.js
```diff
@@ -234,6 +257,67 @@ class ScriptType extends EventHandler {
...
+     * @param {string} name - Name of the event that was originally listened to.
+     * @param {HandleEventCallback} callback - Function that was used as the callback when the event was originally
+     * listened to.
+     * @param {object} [scope] - Object that was used as the scope when the event was originally listened to
+     * @returns {ScriptType} Self for chaining.
+     * @example
+     * this.unlisten(this.app.mouse, 'mousemove', this.onMouseMove, this);
+     * @see {@link ScriptType#listen} to listen to events on an EventHandler.
+     */
+    unlisten(eventHandler, name, callback, scope) {
```
yaustar, line 302: This a slow function but I don't expect this to be called very often/at all and the number of subscribed events is generally expected to be quite low generally.

src/framework/script/script-type.js
```diff
@@ -6,6 +6,17 @@ import { ScriptAttributes } from './script-attributes.js';
 
 const funcNameRegex = new RegExp('^\\s*function(?:\\s|\\s*\\/\\*.*\\*\\/\\s*)+([^\\(\\s\\/]*)\\s*');
 
+/**
+ * EventListener object used for storing what events on EventHandlers the ScriptType is listening for.
+ *
+ * @typedef {object} EventListener
```
kungfooman, line 12: `@typedef` is too good to be allowed to use, you need to remove it and mess around with `any` instead. :wink:

yaustar: I did wonder if I should use a typedef or a new class definition 🤔

kungfooman: Proper way by example: https://github.com/playcanvas/engine/blob/6d4766e7dd7cdc2e4ecfac439e4cb4b85b2e0404/src/framework/anim/state-graph/anim-state-graph.js#L17

(while removing neat IntelliSense)

src/framework/script/script-type.js
```diff
@@ -234,6 +260,61 @@ class ScriptType extends EventHandler {
...
+                eventHandler.off(name, callback, scope);
+                found = true;
+                break;
+            }
+        }
+
+        if (!found) {
+            Debug.warn(`${this.__name}: Cannot find listener to remove for event ${name}. Please check where you originally listened for the event.`);
+        }
+    }
```
kungfooman, line 316: I also don't consider using this too often and performance is a one-time-thing and doesn't even look too bad... it could be a bit shorter tho?

```suggestion
    unlisten(eventHandler, name, callback, scope) {
        for (let i = 0; i < this._listeners.length; ++i) {
            const l = this._listeners[i];
            if (l.eventHandler === eventHandler && l.name === name && l.callback === callback && l.scope === scope) {
                this._listeners.splice(i, 1);
                eventHandler.off(name, callback, scope);
                return;
            }
        }
        Debug.warn(`${this.__name}: Cannot find listener to remove for event ${name}. Please check where you originally listened for the event.`);
    }
```

kungfooman:
Would be nice to see this merged, saves a ton of time/nerves handling events :+1: 

Also required to rather easily fix Seemore scene change issue as described in https://github.com/playcanvas/engine/issues/5409#issuecomment-1597271712

Good work as always :100:

src/framework/script/script-type.js
yaustar: Hmm, not sure why I used a flag for found 🤔

yaustar: Now I remember, I used to allow chaining so there was a `return this;` at the end of the function. I removed it as I rarely see chaining used in practise where it already exists and didn't want to limit future expansion of the API unnecessarily

yaustar: Good catch, I've taken on your suggestion

yaustar:
Anyone got time to review this one for 1.65? :)

Maksims:
Just a comment related to this issue: I've made a small override of `on` and `once` and another method of event handler, to return object with `off` method, and in a large projects with long legacy we have not encountered any issue.

And it made it a breeze to manage events that way.

kungfooman:
> Just a comment related to this issue: I've made a small override of `on` and `once` and another method of event handler, to return object with `off` method

Interesting, can you share full example? Is it only an empty object otherwise? If so, what do you think about just returning a function which - when called - removes the event?

Just a little bit like React `useEffect` with a different life-cycle/context:

```js
  useEffect(() => {
    function handleStatusChange(status) {
      setIsOnline(status.isOnline);
    }
    ChatAPI.subscribeToFriendStatus(props.friend.id, handleStatusChange);
    // Specify how to clean up after this effect:
    return function cleanup() {
      ChatAPI.unsubscribeFromFriendStatus(props.friend.id, handleStatusChange);
    };
  });
```

Maksims:
Example of usage:

```javascript
let evt = entity.on('trigger', () => { });

// sometime after
evt.off();
```

willeastcott:
@Maksims That is what the Editor does in various places. As an example:

```
const events = [];

events.push(editor.on('someevent', ...));
events.push(editor.on('otherevent', ...));

events.push(panel.on('clear', () => {
    for (const event of events) {
        event.unbind();
    }
    events.length = 0;
});
```

Maksims:
> @Maksims That is what the Editor does in various places. As an example:
> 
> ```
> const events = [];
> 
> events.push(editor.on('someevent', ...));
> events.push(editor.on('otherevent', ...));
> 
> events.push(panel.on('clear', () => {
>     for (const event of events) {
>         event.unbind();
>     }
>     events.length = 0;
> });
> ```

Indeed.
As I've designed Editor's events API based on exactly this pain of managing events. So within Editor's context - it proved to be a great and flexible approach.

yaustar:
> Just a comment related to this issue: I've made a small override of on and once and another method of event handler, to return object with off method, and in a large projects with long legacy we have not encountered any issue.

With that in mind, do we want to go ahead with this PR or explore other possibilities? 

If so, what is being purposed here? Going with @Maksims comment on returning an object with an `off` method requires the user to still hand this manually in an destroy/disable/enable scenario which I wanted to avoid as that's what people tend to forget/not know how to handle.

kungfooman:
> With that in mind, do we want to go ahead with this PR or explore other possibilities?

Personally I want this merged and not explore 100 possibilities which eventually just get stuck in a rut. I already use it and then I have to patch every PR using this if I want to load for example Seemore to test bug fixes against non-trivial scenes.

You already made sure that this is extensible in future, so nothing is lost - and if someone feels like extending upon this, that can be better done in individual PR's.

Maksims:
Managing events - is somewhat advance topic. It is required in more complex scenarios, as ~80% of events don't need to be managed.

PlayCanvas does not break existing public APIs, so adding an API - should be very considered.
Also, it is important to avoid "multi-path" in API design - to prevent branching during learning. If it is possible to achieve same thing different ways - it is harder to learn such bloated APIs.

Unfortunately this particular issue - has no easy solution.
Proposed API stands out to me personally as a utility library, rather than a first-class API of PlayCanvas way of doing things.

It also does not handle a case where event is already unbound (using `off`), but will be still in a list unless it is unbound only using `unlisten` method. This leads to GC being unable to free things that are reachable from the event function scope in some cases.

kungfooman:
> It also does not handle a case where event is already unbound (using `off`), but will be still in a list unless it is unbound only using `unlisten` method. This leads to GC being unable to free things that are reachable from the event function scope in some cases.

Practically the `unlisten` method isn't necessary because the point is to call `ScriptType#listen` once and then being able to forget about it. Because pretty much everyone forgets (or has forgotten) to `off` their `on`'s.

When the entity (or only the script component) is deleted, the GC gets rid of every `ScriptType` instance including of course every `ScriptType#_listeners`.

Maksims:
> > It also does not handle a case where event is already unbound (using `off`), but will be still in a list unless it is unbound only using `unlisten` method. This leads to GC being unable to free things that are reachable from the event function scope in some cases.
> 
> Practically the `unlisten` method isn't necessary because the point is to call `ScriptType#listen` once and then being able to forget about it. Because pretty much everyone forgets (or has forgotten) to `off` their `on`'s.
> 
> When the entity (or only the script component) is deleted, the GC gets rid of every `ScriptType` instance including of course `ScriptType#_listeners`.

This is only one of the cases for cleaning events.
There are many cases where re-binding events is necessary during lifetime of script component. E.g. bind on enable and unbind on disable. If it would be created using `listen` method and then unbound using `off`, it will lead to stalled references that will not be GC'ed, and developer will not know about it without knowing the details of the implementation of API.

So either it should handle remove of events that were created using `listen` or the alternative API design would be preferred.

yaustar:
This API also unbinds the underlying event during disable and enable.

You are right in its a utility function where the developer can write one line to listen for events that automatically unbinds on disable and destroy and rebind on enable.

Throughout the years I've done support and moderated the forums, the amount of boilerplate code needed to manage events and number of topics created around destroying entities and changing scenes because of dangling events is high.

It's been complained about enough by developers from studios and beginners that it warrants a solution like this where its one call, managed and part of the core API. Not an extra script that is added.

Maksims:
> This API also unbinds the underlying event during disable and enable.

What if it is not desired (some events should be called during disabled state)? But unbind on destroy is desired?
This covers very specific use case. And does not cover few different scenarios, this will lead to still a need to develop own solutions when this does not cover.

Sorry for being critical here, I just believe there is a simpler building block API can be designed that is more flexible and less narrow, while providing a tools to cover different behaviours and needs of a developer.

yaustar:
Then they can use the 'raw' events that currently exist. If there is a solution that doesn't require the developer to have manually listen for the destroy/enable/disable events, then I be happy to hear it but haven't seen any suggestions that cover that use case.

While this is a specific use case, it's a very common one.

Maksims:
So, an combine API of returning events handle with `listen` utility, with handling case of removing event handler from listen list - probably would be a better solution here.
Also, provide an optional mechanic to handle enable/disable when using listen.
Such way developers would use common way to create events, as well as have a utility that does not have hidden behaviours that can bite.

Something like:
```javascript
this.linkEvent(app.on('event', () => { }));
```

We are linking an event to this script instance.
Also, we can link at any point, and unlink using the same event handle.
So it covers the main purpose of this API. Without introducing new events API (alternative way of subscribing).

yaustar:
If we go this route, we would have to be okay breaking API for the on/off functions on the EventHandler as they currently set up for chaining

kungfooman:
Lets say I have a script like this, two `on`'s and two `listen`'s:

```js
HoloVideoPlayer.prototype.initialize = function () {
    this.on("attr:url", this.reload, this);
    this.on('destroy', this.cleanUp, this);
    this.listen(this.app.mouse, "mousemove", this.mousemove, this);
    this.listen(this.app.mouse, "mousedown", this.mousedown, this);
    // allocate all kinds of things down the line...
};
```

I work to ensure that all of the operations done by `initialize` are undone by `cleanUp`. Pretty much allocating resources, textures, mesh instances, meshes, entities, events on self and others:

```js
HoloVideoPlayer.prototype.reload = function () {
    this.cleanUp();
    this.initialize();
}
```

Now something becomes obvious:

```js
HoloVideoPlayer.prototype.cleanUp = function () {
    this.off();
    this.unlisten(this.app.mouse, "mousemove", this.mousemove, this);
    this.unlisten(this.app.mouse, "mousedown", this.mousedown, this);
    // delete other stuff...
}
```

`EventHandler#off` is way easier to work with than `ScriptType#unlisten`. If we had 20 of each, we would still have only one `this.off()`, but 20 `unlisten`'s.

Considering that we want to make it a quick/easy interface, maybe we need to improve `unlisten`?

Maksims:
Here is an example of how I believe such API could work:

```javascript
Script.prototype.initialize = function () {
    this.bindEvents();
};

// handling swap becomes easier
Script.prototype.swap = function (old) {
    // old script events on script instance are removed, so no need to remove them
    // old script linked events are unsubscribed automatically too
    this.bindEvents();
};

Script.prototype.bindEvents = function() {
    this.on('attr:property', this.onProperty);

    // should trigger only when this script is enabled, and be ignored when this script is disabled
    this.linkEvent(this.app.mouse.on('mousedown', this.onMouseDown, this), true);

    // should trigger while this script is either enabled or disabled
    // also, if `otherEntity` gets destroyed, it will lead to `off` on that event,
    // which should automatically be removed from this script linked events
    this.linkEvent(otherEntity.on('someEvent', () => { }, this));
};
```

The API would be:

```javascript
// these are changes to existing API
EventHandler.on(); // returns event handle
EventHandler.once(); // returns event handle

// new APIs
class EventHandle // handle of the event
EventHandle.off(); // unbind an event
EventHandle.silenced // if true, event will not be triggered

ScriptType.linkEvent(eventHandle, [silenceWhenDisable]); // method to link event to script instance life. Optionally can be silenced when script instance is disabled. This method returns eventHandle.
ScriptType.unlinkEvents(); // unbind (off) all linked events, this method is ran automatically when script instance is destroyed

EventHandler.off(); // should ensure linked event handles to script instances to be removed
```

Such API and implementation, will eliminate hanging references of linked event handler that were off'ed elsewhere, also will handle cases where source of the event removes such event (e.g. entity that we subscribe to is destroyed). Additionally it allows developers to have any complexity for event handle's life. And it does not introduces alternative way of making events, which means integrating this approach to existing code, is as easy as adding `ScriptType.linkEvent(` in front of existing lines of code.

The only problem with such approach, is that it requires a breaking change - EventHandler to return EventHandle instance instead of scope of which it was called. Which I have not ever seen to be used by other developers. So to handle such transition, we can create a global project setting, that will define the behaviour, and by default for all existing projects or if not defined will default to old way. For new projects will be enabled by default. For existing projects it will be possible to switch in settings.

Potential impact of such breaking change is extremely tiny. As nor the engine uses "chaining" (which is not chaining) in its code base, nor I personally seen any other project used it.

yaustar:
Do we have an example elsewhere in the engine where creation of an object is 'linked' in a similar way that we need to be considerate of pattern wise?

Maksims:
> Do we have an example elsewhere in the engine where creation of an object is 'linked' in a similar way that we need to be considerate of pattern wise?

I believe this pattern would be new to the engine API design.

src/framework/script/script-type.js
```diff
@@ -234,6 +260,57 @@ class ScriptType extends EventHandler {
...
+     * @param {string} name - Name of the event to bind the callback to.
+     * @param {HandleEventCallback} callback - Function that is called when event is fired. Note
+     * the callback is limited to 8 arguments.
+     * @param {object} [scope] - Object to use as 'this' when the event is fired.
+     * @example
+     * this.listen(this.app.mouse, 'mousemove', this.onMouseMove, this);
+     * @see {@link ScriptType#unlisten} to remove listeners to an event on an EventHandler.
+     */
+    listen(eventHandler, name, callback, scope) {
+        this._listeners.push({
```
kungfooman, line 277:
```suggestion
        eventHandler.on("destroy", () => {
            this.unlisten(eventHandler, name, callback, scope);
        });
        this._listeners.push({
```

kungfooman (changes requested):
Lets say we really need to get rid of the mouse (could be anything really, any entity):

```js
function destroyMouse() {
    const { mouse } = pc.app;
    if (mouse) {
        // For lack of Mouse#destroy:
        mouse.fire("destroy");
        mouse.disableContextMenu();
        mouse.detach();
        mouse.off();
    }
    delete pc.app.mouse;
}
destroyMouse();
```

Right now we would be stuck with it in `ScriptType#_listeners`:

image

src/framework/script/script-type.js
yaustar: That would have to assume that all objects have a destroy event which is not guaranteed

kungfooman: > That would have to assume that all objects have a destroy event which is not guaranteed

Good point! Should we try to remove object references or accept that these objects will remain active in memory then?

As with everything, we have multiple possibilities:

1) Make sure to actually add destroy methods on obvious objects (good style anyway?)
2) Save some kind of GUID reference to objects, which isn't the actual object (would require GUID lookups etc.)
3) Simply accept we reference destroyed objects and trust the garbage collector to eventually free up the memory (as it is right now)
4...) ...any more ideas?

I guess option (3) would work fine in most cases, but this should be a conscious choice and not an oversight decision.

Maksims: Check out WeakMap and WeakSet.

kungfooman: @Maksims Interesting suggestion, maybe I miss something, but we cannot iterate WeakMap and WeakSet by design, so how would we iterate over all event handlers to enable/disable them when needed?

Maksims: Thinking on it more, without at least periodical iteration through listened event handles, or an event on event handler - would be hard to implement self removals.

kungfooman:
I inspected the idea of adding a `EventHandler#destroy` method a bit, which would fire a `this.fire('destroy')` and three classes which extend `EventHandler` already have it:

https://github.com/playcanvas/engine/blob/c55ab2065c5b127e95cdc8898925ec3edea7a36b/src/platform/sound/manager.js#L117-L125

https://github.com/playcanvas/engine/blob/c55ab2065c5b127e95cdc8898925ec3edea7a36b/src/platform/graphics/graphics-device.js#L359-L369

This one feels odd, kinda semantically wrong - remove implies out of hierarchy, but still "working"? Destroying implies you shouldn't touch any properties any longer, because the object is... well... destroyed:

https://github.com/playcanvas/engine/blob/c55ab2065c5b127e95cdc8898925ec3edea7a36b/src/framework/xr/xr-plane.js#L94-L96

At the same time, `EventHandler.destroy` could always call `this.off()`? A quick look over the code base looks like many classes do that already, so we could even reduce the amount of code, pretty much removing duplicate code here and there :thinking:

Maksims:
XrPlane is "remove", because it is not destroyed, as its data is still available to the developer. But it is removed from tracking by the underlying system.

yaustar:
Closing PR as it's not clear of the direction that this should go in. Can alway reopen later down the line.

kungfooman:
> Closing PR as it's not clear of the direction that this should go in. Can alway reopen later down the line.

I think we can just forget about the references in `_listeners`, it's not even a bug and rather simple to debug using Chrome's Heap Snapshot tool... and to put it bluntly, no one else cares about that, the engine is full of examples.

I saw for example exactly one case trying to get rid of the `this.app` reference:

https://github.com/playcanvas/engine/blob/6f82ce824643fa0fae5ec31009c3a50b3594524f/src/framework/script/script-registry.js#L24-L27

No other class would ever consider that.

But forgetting to actually remove event listeners *is* causing bugs.

When I implemented this myself, I implemented some other stuff and came across this event `off`'ing bug aswell:

https://github.com/playcanvas/engine/blob/6f82ce824643fa0fae5ec31009c3a50b3594524f/src/framework/components/camera/system.js#L33-L47

https://github.com/playcanvas/engine/blob/6f82ce824643fa0fae5ec31009c3a50b3594524f/src/framework/components/camera/system.js#L170-L174

Two little code snippets of the same file, anyone wants to guess the bug here?

And this example brings me to another point: instead of in `ScriptType`, we should add this in `EventHandler`, because "events are easy to forget for everyone".

`EventHandler#unlisten`:

```js
    /**
     * Removes a listener for an event on an EventHandler that was added by {@link ScriptType#listen}.
     *
     * @param {EventHandler} [eventHandler] - EventHandler that was originally listened to.
     * @param {string} [name] - Name of the event that was originally listened to.
     * @param {HandleEventCallback} [callback] - Function that was used as the callback when the event was originally
     * listened to.
     * @param {object} [scope] - Object that was used as the scope when the event was originally listened to
     * @example
     * this.unlisten(this.app.mouse, 'mousemove', this.onMouseMove, this);
     * @see {@link ScriptType#listen} to listen to events on an EventHandler.
     */
    unlisten(eventHandler, name, callback, scope) {
        for (let i = 0; i < this._listeners.length; ++i) {
          const l = this._listeners[i];
          if (
            (!eventHandler || l.eventHandler === eventHandler) &&
            (!name         || l.name         === name        ) &&
            (!callback     || l.callback     === callback    ) &&
            (!scope        || l.scope        === scope       )
          ) {
            this._listeners.splice(i, 1);
            l.eventHandler.off(l.name, l.callback, l.scope);
          }
        }
    }
```

With this change, we can simply call `this.unlisten()` to remove all remote events, which is the equivalent for the common `this.off()` for local events.

Anyway, that's my current direction I'm thinking in, would be happy to see more discussion around this... IMO this is rather important to have and I couldn't find another idea as appealing as `listen/unlisten` for non-this event handling.

kungfooman:
I think I found a legit example for `listenEvenWhenDisabled`:

https://github.com/playcanvas/engine/blob/e5c6903d1645d1875d5e047843c3caef3a643d1c/src/scene/shader-lib/program-library.js#L33-L56

Not directly applicable here (since it's not a script), but when it comes to resource handling, we woud want to destroy shaders or other resources.

In the script I also don't see a `off` call, might be an event leak?

And as usual, we don't remove `device` from the instance, "leaking" just as much as `listen/unlisten` would do - so I don't see a strong argument against it.

Maksims:
Made a PR to push "chaining" away, that will lead to better APIs and events management in the future.

PR #5345, opened 2023-05-22
Fix for multiple non clustered lights

ellthompson:
image

The lit shader currently fails to compile when using multiple non clustered lights. This PR removes the redefinition of the `dShadowCoord` variable on line 990.

### DIFF
--- a/src/scene/shader-lib/programs/lit-shader.js
+++ b/src/scene/shader-lib/programs/lit-shader.js
@@ -1250,7 +1250,7 @@ class LitShader {
                             hasCascades = true;
                         }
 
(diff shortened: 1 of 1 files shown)


willeastcott:
Is this a regression? I'm surprised this hasn't been reported before. If so, when did it regress?

mvaligursky (approved):

mvaligursky:
This likely got broken last week when the shadow coordinate generation was refactored and is not part of the released engine. If that is the case, we don't need to patch release this.

willeastcott:
@mvaligursky Oh that's great news. 😄

ellthompson:
> This likely got broken last week when the shadow coordinate generation was refactored and is not part of the released engine. If that is the case, we don't need to patch release this.

That's right, this is only an issue in 1.64.0-dev.

PR #5347, opened 2023-05-23
Update NPM dependencies

willeastcott:
Also include lint fixes as a result of the update.


src/framework/components/camera/component.js
```diff
@@ -148,11 +148,11 @@ class CameraComponent extends Component {
...
      * material.onUpdateShader = function (options) {
-     *    if (options.pass === shaderPassId) {
-     *        options.litOptions.normalMapEnabled = false;
-     *        options.litOptions.useSpecular = false;
-     *    }
-     *    return options;
+     * if (options.pass === shaderPassId) {
+     * options.litOptions.normalMapEnabled = false;
+     * options.litOptions.useSpecular = false;
+     * }
```
mvaligursky, line 154: is this required? It does not look as good as before.

src/scene/graph-node.js
```diff
@@ -912,7 +912,7 @@ class GraphNode extends EventHandler {
     /**
      * Returns cached value of negative scale of the world transform.
      *
-     * @returns {number} -1 if world transform has negative scale, 1 otherwise.
+     * @returns {number} - 1 if world transform has negative scale, 1 otherwise.
```
mvaligursky, line 915: hmm it should be '-1' here, not '- 1'.

src/framework/components/camera/component.js
willeastcott: I think this is a bug in eslint_plugin_jsdoc. I'll investigate.

src/scene/graph-node.js
willeastcott: Fixed.

src/framework/components/camera/component.js
mvaligursky: this has been fixed / avoided in https://github.com/playcanvas/engine/pull/5472

Issue #5367, opened 2023-06-01
raycastFirst error when passing options


mvaligursky:
see details here: https://forum.playcanvas.com/t/raycastfirst-error-when-passing-options/31564

I think the deprecated handling of the callback needs to be refactored / removed.

PR #5368, opened 2023-06-01
Remove no longer compatible handling of deprecated parameter by RigidBody.raycastFirst

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/5367

### DIFF
--- a/src/framework/components/rigid-body/system.js
+++ b/src/framework/components/rigid-body/system.js
@@ -525,14 +525,6 @@ class RigidBodyComponentSystem extends ComponentSystem {
                     new Vec3(normal.x(), normal.y(), normal.z()),
                     rayCallback.get_m_closestHitFraction()
                 );
-
-                // keeping for backwards compatibility
-                if (arguments.length > 2) {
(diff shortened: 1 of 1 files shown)


MushAsterion:
Oh my bad, didn't noticed such stuff. However instead of removing backward compatibility wouldn't it be better to rather check that `arguments[2]` is also a function?

```JavaScript
// keeping for backwards compatibility
if (arguments.length > 2 && typeof arguments[2] === 'function') {
    Debug.deprecated('pc.RigidBodyComponentSystem#rayCastFirst no longer requires a callback. The result of the raycast is returned by the function instead.');

    const callback = arguments[2];
    callback(result);
}
```

mvaligursky:
I considered that too .. but I think it's good to occasionally actually remove the deprecated code instead of keeping it forever.

willeastcott:
Yeah, I'm inclined to agree with @mvaligursky here. That function has been asserting about deprecation of the callback function for many years now. I think folks have had enough time to update their code. 😉

willeastcott (approved):

PR #5383, opened 2023-06-09
Fix safari detection on ipad

slimbuck:
Fixes https://forum.playcanvas.com/t/having-wavy-patterns-again-on-ipad/31380

WebglGraphicsDevice was using the presence of `window.safari` to detect safari browsers, but this only works on desktop, not ios.

To address this we add very simple browser detection logic and expose it via `platform.browserName`. This value is flagged internal for now and not exposed as public API.

This PR also simplifies the existing detection logic without changing functionality.

### DIFF
--- a/src/core/platform.js
+++ b/src/core/platform.js
@@ -38,7 +38,7 @@ const browserName =
 
 const xbox = /xbox/i.test(ua);
 const touch = (environment === 'browser') && ('ontouchstart' in window || ('maxTouchPoints' in navigator && navigator.maxTouchPoints > 0));
-const gamepads = (typeof navigator !== 'undefined') && (!!navigator.getGamepads || !!navigator.webkitGetGamepads);
+const gamepads = (environment === 'browser') && (!!navigator.getGamepads || !!navigator.webkitGetGamepads);
 const workers = (typeof Worker !== 'undefined');
 const passiveEvents = detectPassiveEvents();
 


mvaligursky (approved):

src/core/platform.js
```diff
@@ -88,76 +70,84 @@ const platform = {
...
-     * If it is Android.
+     * True if running on an Android device.
      *
      * @type {boolean}
      */
     android: android,
 
     /**
-     * If it is Windows.
+     * True if running on a Windows mobile device.
```
willeastcott, line 101: Windows Phone was discontinued 7 years ago:

https://en.wikipedia.org/wiki/Windows_Phone

I would remove this property.

slimbuck: done!

src/core/platform.js
```diff
@@ -88,76 +75,77 @@ const platform = {
...
-    passiveEvents: passiveEvents
+    passiveEvents: passiveEvents,
+
+    /**
+     * Get the browser name.
+     *
+     * @type {'chrome' | 'safari' | 'firefox' | 'other' | null}
+     * @ignore
+     */
+    browserName: browserName
```
willeastcott, line 148: I kinda prefer just 'browser' for naming.

willeastcott (approved):

PR #5391, opened 2023-06-14
Fix light size default value

GSterbrant:
Missed changing the default light size.


mvaligursky:
There's another copy of this value in light.js

GSterbrant:
Thanks, fixed that one as well.

mvaligursky (approved):

PR #5401, opened 2023-06-15
AnimBlendTree fix

### LINKED ISSUE #5398: Multiple animation engine examples fail to run
Multiple animation engine examples error on the infinite recursion, possibly related to https://github.com/playcanvas/engine/pull/5392

image

ellthompson:
The findParameter and comsumeTrigger functions should always be called in the AnimController's scope so that they have access to its private variables. Setting these functions as anonymous ensures they are always called with this scope, even when they are passed to and called by the various AnimBlendTree classes.

Fixes #5398


mvaligursky (commented):
Please add the explanation to those functions (similar as the PR's description), as it's not obvious.

PR #5403, opened 2023-06-16
Fix PCSS with WebGL1

### LINKED ISSUE #5397: Contact Hardening Shadow example fails to run on WebGL1
A fallback should be handled, and PCF3 or similar should be used instead by the engine.

Screenshot 2023-06-15 at 12 45 20

GSterbrant:
PCSS wouldn't compile with WebGL1 due to fetching a uniform value through a function. Moving the uniform value fetch to directly in the loop fixed the issue. 

Fixes #5397

### DIFF
--- a/src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
+++ b/src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
@@ -12,10 +12,8 @@ export default /* glsl */`
 uniform float pcssDiskSamples[PCSS_SAMPLE_COUNT];
 uniform float pcssSphereSamples[PCSS_SAMPLE_COUNT];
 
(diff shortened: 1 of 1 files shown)


slimbuck (approved):
Any idea what the rule is here being broken in WebGL1?

mvaligursky (approved):

GSterbrant:
Using an argument to sample from a uniform makes the compiler complain that the value must be constant. What they mean is probably `known at compile time`. 

It is also known at compile time obviously but the compiler probably sees the function outside of the scope of the callee and assumes it can be called anywhere, making it assume the uniform fetch can't be guaranteed to be invariant, which is what provokes the error. 

But the error is simply `Index expression must be constant`.

PR #5413, opened 2023-06-19
Handle anim events during reverse playback

### LINKED ISSUE #3808: Make anim events fire for reversed anims.
### Description

Make anim events work when an animation is running in reverse due to using a negative speed setting.

### Steps to Reproduce

In the Animation Events example, if you swap the sign on the speed, the anim events don't fire:

`113: modelEntity.anim.assignAnimation("Walk", walkTrack, undefined, -0.62);`

### LINKED ISSUE #4426: Anim event set at time 0 is fired at the start and end of a non looping animation
Thread: https://forum.playcanvas.com/t/anim-system-issues/26559/

Project: https://playcanvas.com/project/956782

Anim Event is set on both non looping animations at time 0

The event is fired at the start of the animation and at the end

ellthompson:
Setting an anim component state's speed to anything below zero would stop anim events firing during playback. The `activeEventsForFrame` function has been made agnostic to the direction of playback.

Fixes #3808
Fixes #4426

### DIFF
--- a/src/framework/anim/evaluator/anim-clip.js
+++ b/src/framework/anim/evaluator/anim-clip.js
@@ -153,22 +153,23 @@ class AnimClip {
     }
 
     clipFrameTime(frameEndTime) {
-        AnimClip.eventFrame.start = 0;
-        AnimClip.eventFrame.end = frameEndTime;
-        AnimClip.eventFrame.residual = 0;
+        const eventFrame = AnimClip.eventFrame;
+        eventFrame.start = 0;
+        eventFrame.end = frameEndTime;
+        eventFrame.residual = 0;
 
         // if this frame overlaps with the end of the track, we should clip off the end of the frame time then check that clipped time later
         if (this.isReverse) {
             if (frameEndTime < 0) {
-                AnimClip.eventFrame.start = this.track.duration;
-                AnimClip.eventFrame.end = 0;
-                AnimClip.eventFrame.residual = frameEndTime + this.track.duration;
+                eventFrame.start = this.track.duration;
+                eventFrame.end = 0;
+                eventFrame.residual = frameEndTime + this.track.duration;
             }
         } else {
             if (frameEndTime > this.track.duration) {
-                AnimClip.eventFrame.start = 0;
-                AnimClip.eventFrame.end = this.track.duration;
-                AnimClip.eventFrame.residual = frameEndTime - this.track.duration;
+                eventFrame.start = 0;
+                eventFrame.end = this.track.duration;
+                eventFrame.residual = frameEndTime - this.track.duration;
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/framework/anim/evaluator/anim-clip.js
```diff
@@ -104,35 +105,98 @@ class AnimClip {
...
+        return this.isReverse ? 0 : this._track.events.length - 1;
+    }
+
+    get nextEvent() {
+        return this._track.events[this._eventCursor];
+    }
+
+    get isReverse() {
+        return this._speed < 0;
+    }
```
yaustar, line 118: What if speed was 0. Could that cause an issue with any logic?

ellthompson: Clips won't update events when the speed is 0 so there's no change in behaviour here.

src/framework/anim/evaluator/anim-clip.js
```diff
@@ -104,35 +105,101 @@ class AnimClip {
...
+        if (this._eventCursor >= this.track.events.length) {
+            this._eventCursor = 0;
+        } else if (this._eventCursor < 0) {
+            this._eventCursor = this.track.events.length - 1;
         }
-        let clippedFrameDuration;
+    }
+
+    clipFrameTime(frameEndTime) {
+        const result = {
```
mvaligursky, line 147: can we find a solution that does not allocate the object here?

src/framework/anim/evaluator/anim-clip.js
```diff
@@ -66,6 +66,7 @@ class AnimClip {
 
     set speed(speed) {
         this._speed = speed;
+        this.alignCursorToCurrentTime();
```
mvaligursky, line 69: should this be called only when the sign changes?

mvaligursky: but even then, it seems strange to change the start index for events when the speed changes, as at this point the time has not changed?

src/framework/anim/evaluator/anim-clip.js
ellthompson: Moved this to a static class property.

src/framework/anim/evaluator/anim-clip.js
```diff
@@ -104,35 +114,98 @@ class AnimClip {
...
+        if (this._eventCursor >= this.track.events.length) {
+            this._eventCursor = 0;
+        } else if (this._eventCursor < 0) {
+            this._eventCursor = this.track.events.length - 1;
         }
-        let clippedFrameDuration;
+    }
+
+    clipFrameTime(frameEndTime) {
+        AnimClip.eventFrame.start = 0;
```
mvaligursky, line 156: just do `const eventFrame = AnimClip.eventFrame;` here

src/framework/anim/evaluator/anim-clip.js
ellthompson: The clip events are evaluated using an incoming delta time and the current speed. The clip's time will not necessarily have changed when this evaluation happens. That'll only occur if the user has actively set the time themselves.

mvaligursky (approved):

PR #5414, opened 2023-06-19
Fix `Entity#destroy`

### LINKED ISSUE #4649: Seemore double renders when destroying all the entities and loading the same scene hierarchy again
Related posts: https://github.com/playcanvas/engine/pull/4626#issuecomment-1249425411

image

Launch this project: https://playcanvas.com/project/983477/overview/f-seemore
Press 1

kungfooman:
Fixes #4649

The main issue is simply that `GraphNode` has no `destroy` method and if a `Entity` is inside a `GraphNode`... it will never be deleted.

I can imagine many solutions, but simply to get the conversations going I propose this one...

### DIFF
--- a/src/framework/entity.js
+++ b/src/framework/entity.js
@@ -514,38 +514,12 @@ class Entity extends GraphNode {
             this.c[name].enabled = false;
         }
 
+        super.destroy();
+
         // Remove all components
         for (const name in this.c) {
             this.c[name].system.removeComponent(this);
         }
-
-        // Detach from parent
-        this._parent?.removeChild(this);
-
-        // If an Entity is in a GraphNode, recursive destroy'ing will never get to them.
-        // Therefore we have to find these special cases and destroy iteratively.
-        const isEntityChildOfGraphNode = _ => _ instanceof Entity && _.parent?.constructor === GraphNode;
-        const entitiesInGraphNodes = this.find(isEntityChildOfGraphNode);
-        entitiesInGraphNodes.forEach(e => e.destroy());
-
-        // recursively destroy all children
-        const children = this._children;
-        while (children.length) {
-
-            // remove a child from the array and disconnect it from the parent
-            const child = children.pop();
-            child._parent = null;
-
-            if (child instanceof Entity) {
-                child.destroy();
-            }
-        }
-
-        // fire destroy event
-        this.fire('destroy', this);
-
-        // clear all events
-        this.off();
 
         // remove from entity index
         if (this._guid) {
--- a/src/framework/entity.js
+++ b/src/framework/entity.js
@@ -520,8 +520,13 @@ class Entity extends GraphNode {
         }
 
         // Detach from parent
-        if (this._parent)
-            this._parent.removeChild(this);
+        this._parent?.removeChild(this);
+
+        // If an Entity is in a GraphNode, recursive destroy'ing will never get to them.
+        // Therefore we have to find these special cases and destroy iteratively.
+        const isEntityChildOfGraphNode = _ => _ instanceof Entity && _.parent?.constructor === GraphNode;
+        const entitiesInGraphNodes = this.find(isEntityChildOfGraphNode);
+        entitiesInGraphNodes.forEach(e => e.destroy());
 
         // recursively destroy all children
         const children = this._children;
--- a/src/framework/entity.js
+++ b/src/framework/entity.js
@@ -514,12 +514,12 @@ class Entity extends GraphNode {
             this.c[name].enabled = false;
         }
 
-        super.destroy();
-
         // Remove all components
         for (const name in this.c) {
             this.c[name].system.removeComponent(this);
         }
+
+        super.destroy();
 
         // remove from entity index
         if (this._guid) {
--- a/src/framework/scene-registry.js
+++ b/src/framework/scene-registry.js
@@ -381,11 +381,6 @@ class SceneRegistry {
             while (children.length) {
                 children[0].destroy();
             }
-            Debug.assert(
-                Object.keys(app._entityIndex).length === 0,
-                "Destroyed all entities but these entities survived:",
-                app._entityIndex
-            );
             app.applySceneSettings(sceneItem.data.settings);
         };
 
--- a/src/framework/scene-registry.js
+++ b/src/framework/scene-registry.js
@@ -376,14 +376,16 @@ class SceneRegistry {
         const app = this._app;
 
         const onBeforeAddHierarchy = (sceneItem) => {
-            // Destroy/Remove all nodes on the app.root
-            const rootChildren = app.root.children;
-            while (rootChildren.length > 0) {
-                const child = rootChildren[0];
-                child.reparent(null);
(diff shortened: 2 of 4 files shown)


yaustar:
Just to confirm, that scene basically has:

```
entity
-> graph node
- -> entity
```

kungfooman:
@yaustar yep exactly, testable like:

```js
const isEntityChildOfGraphNode = _ => _ instanceof pc.Entity && _.parent?.constructor === pc.GraphNode;
const entitiesInGraphNodes = pc.app.root.find(isEntityChildOfGraphNode);
entitiesInGraphNodes.map(ent => {
    const {name, path, parent} = ent;
    return {name, path, ent, parent};
});
```

Output in Seemore:

image


More ideas for (possible?) solutions:

1) As in PR, find these special cases before recursively destroy entities "normally" (as it always was)
2) Implement `GraphNode#destroy`?
3) Simply find every entity using `GraphNode#find` and delete them all iteratively?

Probably everything has pros/cons, but so far I tend to (1) or (2)

mvaligursky:
personally I would find the `Implement GraphNode#destroy` option a lot cleaner, any reasons not to go that route?

kungfooman:
I just thought about performance, lets say having 20 skeletons with ~70 `GraphNode`'s and `GraphNode` is a bit "lower level" and calling `destroy` on them even tho they may have no entities might be useless. But I guess in real-life people put entities in `GraphNode`'s all the time to track animated positions or attach particle effects, so being "pure" with GraphNode's is not really a thing.

> personally I would find the `Implement GraphNode#destroy` option a lot cleaner, any reasons not to go that route?

So I pretty much agree, I would also like to see `GraphNode#detachFromParent`, because I see it all over the place handled via nested if's:

image

Or other tricks are `GraphNode#reparent(null)`... which feels "okayish" once you know it, but still not 100% intuitive. I guess therefore it's only used once:

image

Would it be okay to implement `GraphNode#detachFromParent` and `GraphNode#destroy`?

mvaligursky:
> Would it be okay to implement `GraphNode#detachFromParent` and `GraphNode#destroy`?

Absolutely, I'd like that. If possible, create two separate PRs please.

kungfooman:
Debugging and debugging :nerd_face: 

image

Found another reason why it fails...

Called from:

image

Simply said, the "component removal code" is snatching away our entities (which we want to delete recursively) before our recursion ever has a chance to get to them (they just "disappear" 🤯👻💥).

This is easily fixed though, simply first disabled all component, then do the recursion, and then to component deletion.

Or 2nd possible solution: mark the entire hierarchy we want to delete (something like `_destroying` flag) beforehand and then make `GraphNode#removeChild` do nothing, because we know better (namely that this is not a removal but a destruction).

But I like the 1st solution, simply destroying the components after the recursion, so I will research that route a bit more... I realize that historically this stuff was a mine field, a change here breaks a project there etc... please just link me all scenes I shall test this with :mag: (after all, we want to fix bugs, not introduce new ones).

kungfooman:
Ready for review... pretty happy about the result.

I'm testing with Seemore and besides the event bug I also found a room-loading bug in `demo-loader.js`:

https://github.com/kungfooman/seemore/commit/25e565c87a68721f2327e6e9632acc69d3748197

src/scene/graph-node.js
```diff
@@ -480,6 +480,37 @@ class GraphNode extends EventHandler {
         return this;
     }
 
+
+    /**
+     * Detach it from the hierarchy. Then recursively destroy all ancestors.
```
mvaligursky, line 485:
```suggestion
     * Detach a GraphNode from the hierarchy. Then recursively destroy all ancestors.
```

src/framework/entity.js
```diff
@@ -514,34 +514,13 @@ class Entity extends GraphNode {
             this.c[name].enabled = false;
         }
 
+        super.destroy();
```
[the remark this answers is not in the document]
kungfooman: That doesn't work, this would be the result:

image

Reason I described here: https://github.com/playcanvas/engine/pull/5414#issuecomment-1599076644

The components think they can remove entities from the hierarchy without deleting them, so the get orphaned basically and the recursion doesn't find them any longer.

src/framework/scene-registry.js
```diff
@@ -376,14 +376,16 @@ class SceneRegistry {
...
-                const child = rootChildren[0];
-                child.reparent(null);
-                child.destroy?.();
+            // Destroy all nodes on the app.root
+            const { children } = app.root;
+            while (children.length) {
+                children[0].destroy();
             }
-
+            Debug.assert(
```
mvaligursky, line 384: I'm not very familia with the _entityIndex, but could this incorrectly fire if we have entities with GUID that are not added to the root?

mvaligursky: I can imagine entities that are not under the root .. for example in some pool class, or just stored to be displayed occasionally.

kungfooman: Yep, personally I just keep those entities parented and enable/disable when needed, but yes, I can see how that's a possibility.

kungfooman:
Context for latest commit:

image

Output of this code in `src/framework/components/model/component.js`:

```js
            this.removeModelFromLayers();
            const ents = this._model.getGraph().find(_ => _ instanceof Entity);
            if (ents.length) {
                console.log(ents);
            }
            this._model.getGraph().destroy();
```

mvaligursky (approved):
Approving, looks great to me, I don't see any problems now. Thanks!

src/framework/scene-registry.js
```diff
@@ -376,14 +376,11 @@ class SceneRegistry {
...
 
         const onBeforeAddHierarchy = (sceneItem) => {
-            // Destroy/Remove all nodes on the app.root
-            const rootChildren = app.root.children;
-            while (rootChildren.length > 0) {
-                const child = rootChildren[0];
-                child.reparent(null);
-                child.destroy?.();
+            // Destroy all nodes on the app.root
+            const { children } = app.root;
```
willeastcott, line 380: Why not just:

```
const children = app.root.children;
```

src/scene/graph-node.js
```diff
@@ -480,6 +480,37 @@ class GraphNode extends EventHandler {
         return this;
     }
 
+
+    /**
+     * Detach a GraphNode from the hierarchy. Then recursively destroy all ancestors.
```
willeastcott, line 485: Shouldn't this say descendants instead of ancestors?

kungfooman: Wow, totally, thank you for seeing that! It's the original comment from here: https://github.com/playcanvas/engine/blob/039da2e23ca4a6dcc71630e5f3e79afe21553001/src/framework/entity.js#L502-L503

Do you want to make a code suggestion? "recursively destroy all children"?

willeastcott, line 485:
```suggestion
     * Detach a GraphNode from the hierarchy and recursively destroy all children.
```

willeastcott (approved):

PR #5421, opened 2023-06-21
Change behaviour of lightSize

GSterbrant:
This PR changes the way lightSize modifies the softness of PCSS shadows. Instead of using lightSize directly, it's now a factor of how many pixels per meter to fetch, making the parameter more intuitive to use.

Also threw in a bunch of small optimisations.

### DIFF
--- a/src/framework/components/light/component.js
+++ b/src/framework/components/light/component.js
@@ -89,9 +89,9 @@ const _lightPropsDefault = [];
  * angle is specified in degrees. Affects spot lights only. Defaults to 40.
  * @property {number} outerConeAngle The angle at which the spotlight cone has faded to nothing.
  * The angle is specified in degrees. Affects spot lights only. Defaults to 45.
- * @property {number} lightSize Size of the light used for contact hardening shadows. For area lights
- * acts as a size multiplier with the area light dimensions. For punctual and directional lights
- * acts as the actual size of the light. Defaults to 1.0.
+ * @property {number} lightSize Size of penumbra for contact hardening shadows. For area lights
+ * acts as a multiplier with the dimensions of the area light. For punctual and directional lights
+ * it's the area size of the light. Defaults to 1.0.
  * @property {number} falloffMode Controls the rate at which a light attenuates from its position.
  * Can be:
  *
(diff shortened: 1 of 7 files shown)


src/framework/components/light/component.js
```diff
@@ -89,9 +89,9 @@ const _lightPropsDefault = [];
  * angle is specified in degrees. Affects spot lights only. Defaults to 40.
  * @property {number} outerConeAngle The angle at which the spotlight cone has faded to nothing.
  * The angle is specified in degrees. Affects spot lights only. Defaults to 45.
- * @property {number} lightSize Size of the light used for contact hardening shadows. For area lights
- * acts as a size multiplier with the area light dimensions. For punctual and directional lights
- * acts as the actual size of the light. Defaults to 1.0.
+ * @property {number} lightSize Size of penumbra for contact hardening shadows. For area lights
```
mvaligursky, line 92: I wonder if this should be called penumbraSize instead. @willeastcott - thoughts on this? As all this parameter affects is the penumbra size, but it the name suggests it could change the area light size too.

src/scene/shader-lib/programs/lit-shader.js
```diff
@@ -1342,47 +1340,45 @@ class LitShader {
                 }
 
                 // specular / clear coat
-                if (light.affectSpecularity) {
```
mvaligursky, line 1345: just double checking this was line removed on purpose, as @ellthompson recently added it if I remember correctly

src/scene/renderer/forward-renderer.js
```diff
@@ -257,11 +257,13 @@ class ForwardRenderer extends Renderer {
                 this.shadowCascadeDistancesId[cnt].setValue(directional._shadowCascadeDistances);
                 this.shadowCascadeCountId[cnt].setValue(directional.numCascades);
                 this.lightShadowIntensity[cnt].setValue(directional.shadowIntensity);
-                this.lightSizeId[cnt].setValue(directional.lightSize);
+
+                const pixelsPerMeter = 1.0 / (lightRenderData.shadowCamera.renderTarget.width / directional.lightSize);
+                this.lightSizeId[cnt].setValue(pixelsPerMeter);
```
mvaligursky, line 262: I think to make it more understandable, lightSize uniform should be renamed to match its new meaning? `shadowPixelDensity` or something better?

GSterbrant: Fixed, named it shadowSearchArea.

src/framework/components/light/component.js
willeastcott: If the description is `Size of penumbra`, then yes, I'd says `penumbraSize` is better. 😆

PR #5426, opened 2023-06-21
[Anim Component] Fix exit time conditions for non looping animations

### LINKED ISSUE #4427: Anim transition with exit time of 1 is not triggered when conditions match after non looping state has finished
Thread: https://forum.playcanvas.com/t/anim-system-issues/26559

Project https://playcanvas.com/project/956796/overview/cant-exit-non-looping-state

Repro steps: 

Hold P to transition to the punch animation
Wait for the punch animation to finish
Release P

Expected result is that the graph should transition to the Idle state as the punch state has reached the end of the animation and is at exit time 1

Related to #4426 ?

ellthompson:
The anim controller shouldn't update the state time when a non looping animation has completed its playback. Additionally  states that have finished their playback should have their exit time conditions checked as if they were playing.

Fixes #4427

### DIFF
--- a/src/framework/anim/controller/anim-controller.js
+++ b/src/framework/anim/controller/anim-controller.js
@@ -306,8 +306,13 @@ class AnimController {
                     progressBefore -= Math.floor(progressBefore);
                     progress -= Math.floor(progress);
                 }
-                // return false if exit time isn't within the frames delta time
-                if (!(transition.exitTime > progressBefore && transition.exitTime <= progress)) {
+                // if the delta time is 0 and the progress matches the exit time, the exitTime condition has been met
+                if (progress === progressBefore) {
+                    if (progress !== transition.exitTime) {
+                        return null;
+                    }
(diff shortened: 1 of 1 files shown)


src/framework/anim/controller/anim-controller.js
```diff
@@ -516,8 +521,17 @@ class AnimController {
         let state;
         let animation;
         let clip;
-        this._timeInStateBefore = this._timeInState;
-        this._timeInState += dt * this.activeState.speed;
+        // update time when looping or when the active state is not at the end of its duration
+        if (this.activeState.loop || this._timeInState < this.activeStateDuration) {
```
mvaligursky, line 525: `this.activeStateDuration` is really expensive - it searches for animations by name, and it gets called here 4x ?

ellthompson: I've added a flag to the calculation of the current states duration. It'll now be calculated at most once per state transition or assignment of a new animation file.

mvaligursky (approved):

PR #5427, opened 2023-06-22
Removed no longer needed temporarily WebGPU test code

mvaligursky:
- now the real examples can be used instead


willeastcott (approved):
Deleting stuff FTW! 🎉

PR #5432, opened 2023-06-23
Fix point light shadows missing float packing

GSterbrant:
We can skip this if we decide to support using R32F textures for cube shadows.

### DIFF
--- a/src/scene/shader-lib/programs/lit-shader.js
+++ b/src/scene/shader-lib/programs/lit-shader.js
@@ -474,7 +474,8 @@ class LitShader {
         code += this.frontendDecl;
         code += this.frontendCode;
 
-        const usePackedDepth = (!device.supportsDepthShadow) && (shadowType === SHADOW_PCF1 || shadowType === SHADOW_PCF3 || shadowType === SHADOW_PCSS);
+        const usePackedDepth = (!device.supportsDepthShadow) && (shadowType === SHADOW_PCF1 || shadowType === SHADOW_PCF3 || shadowType === SHADOW_PCSS) ||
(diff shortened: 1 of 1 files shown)


mvaligursky (approved):
That looks correct to me.

src/scene/shader-lib/programs/lit-shader.js
```diff
@@ -474,7 +474,8 @@ class LitShader {
         code += this.frontendDecl;
         code += this.frontendCode;
 
-        const usePackedDepth = (!device.supportsDepthShadow) && (shadowType === SHADOW_PCF1 || shadowType === SHADOW_PCF3 || shadowType === SHADOW_PCSS);
+        const usePackedDepth = (!device.supportsDepthShadow) && (shadowType === SHADOW_PCF1 || shadowType === SHADOW_PCF3 || shadowType === SHADOW_PCSS) ||
+            (lightType === LIGHTTYPE_OMNI && !options.clusteredLightingEnabled);
```
slimbuck, line 478:
```suggestion
        const usePackedDepth =
            (!device.supportsDepthShadow) &&
            (shadowType === SHADOW_PCF1 || shadowType === SHADOW_PCF3 || shadowType === SHADOW_PCSS) ||
            (lightType === LIGHTTYPE_OMNI && !options.clusteredLightingEnabled);

```

GSterbrant: Made it more clear that the supportsDepthShadow test and the PCF1/3 and PCSS methods are dependent on it by adding an extra set of parenthesis.

slimbuck (approved):

PR #5438, opened 2023-06-26
Bump allocation for Uniform Buffers on WebGPU

mvaligursky:
Before this PR, each UniformBuffer would allocate its internal GPUBuffer storage, and per frame copy the CPU storage content to it using writeBuffer. This required many writeBuffer calls, which is expensive on both CPU and GPU time.

This PR implement more performant implementation. Under the hood, a one or more large (1MB) gpu buffers are allocated, and a pool of staging buffers of the same size. Individual uniform buffers allocate storage using bump allocator from the staging buffers. Then, just before the command buffers are submitted, a command buffer is added to execute first, which copies the used staging buffers to the gpu buffers. 
Here's an example of used buffers for many example. Note that the number of staging buffers gets larger each time a command buffers are submitted, as they can no longer use already existing staging buffer.

buffer-allocation

This PR also cleans up some temporary solutions introduced in #5423 to limit the number of expensive submit commands per frame. Before, command buffer of each render pass would be submitted separately, while now those are batched to a very small number.

As an example, the shadow cascades example is using a single submit, first copying the staging buffers to gpu buffers, following by a single command buffer render all shadow cascade render passes, followed the the forward pass of the scene:

shadow-cascades

Multi view example similarly renders the whole scene using a single submit for all command buffers:

multi-view

If there are texture uploads done in a frame (typically a very small number of places), for example in this case the bone texture used by the skinning, and clustered lights updated on CPU, we end up with two submits:

texture-uploads

All rendering submitted from the update functions of the script are submitted separately for now (could be a single submit as well), for example reflection-cubemap example which renders the scene using a single submit, and does multiple texture reprojections using draQuadWithShader within the scripts:

reflectio-cubemap

## Performance

CPU frame time for the hierarchy example with 5000 or so meshes:
- WebGPU before: 57ms
- WebGPU now: 48ms (15% improvement)
- for comparison, WebGL time: 14ms

GPU times (these are based on the GPU duration reported by Chrome Profiler only, not sure about their reliability / what else they capture). I do not think this is reliable at all.
- WebGPU before: 12.8ms
- WebGPU now: 11.8ms
- WebGL time: hard to estimate in browser, too many displayed bars.

### DIFF
--- a/src/platform/graphics/dynamic-buffers.js
+++ b/src/platform/graphics/dynamic-buffers.js
@@ -118,8 +118,8 @@ class DynamicBuffers {
      * Create the system of dynamic buffers.
      *
      * @param {import('./graphics-device.js').GraphicsDevice} device - The graphics device.
-     * @param {*} bufferSize - The size of the underlying large buffers.
-     * @param {*} bufferAlignment - Alignment of each allocation.
+     * @param {number} bufferSize - The size of the underlying large buffers.
+     * @param {number} bufferAlignment - Alignment of each allocation.
      */
     constructor(device, bufferSize, bufferAlignment) {
         this.device = device;
--- a/src/platform/graphics/dynamic-buffers.js
+++ b/src/platform/graphics/dynamic-buffers.js
@@ -1,0 +1,218 @@
+import { Debug } from '../../core/debug.js';
+import { math } from '../../core/math/math.js';
+
+/**
+ * A base class representing a single per platform buffer.
+ *
+ * @ignore
+ */
+class DynamicBuffer {
+    /** @type {import('./graphics-device.js').GraphicsDevice} */
+    device;
+
+    constructor(device) {
+        this.device = device;
+    }
+}
+
+/**
+ * A container for storing the used areas of a pair of staging and gpu buffers.
+ *
+ * @ignore
+ */
+class UsedBuffer {
+    /** @type {DynamicBuffer} */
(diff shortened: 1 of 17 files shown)


src/platform/graphics/dynamic-buffers.js
```diff
@@ -0,0 +1,218 @@
+import { Debug } from '../../core/debug.js';
```
willeastcott, line 1: I think this module should be called `dynamic-buffer.js`. (Drop the 's')

src/platform/graphics/dynamic-buffers.js
```diff
@@ -0,0 +1,218 @@
...
+     * @type {UsedBuffer}
+     */
+    activeBuffer = null;
+
+    /**
+     * Create the system of dynamic buffers.
+     *
+     * @param {import('./graphics-device.js').GraphicsDevice} device - The graphics device.
+     * @param {*} bufferSize - The size of the underlying large buffers.
+     * @param {*} bufferAlignment - Alignment of each allocation.
```
willeastcott, line 122: Aren't the types here `number`?

willeastcott (approved):

src/platform/graphics/dynamic-buffers.js
mvaligursky: yep, I missed filling those in, done now, thanks

PR #5441, opened 2023-06-27
Compensate PCSS kernel with depth range for directional lights

GSterbrant:
PCSS reacts differently based on scene size with directional lights. This PR fixes that by compensating the depth differences caused by a change in the depth range.


src/scene/light.js
```diff
@@ -65,6 +65,9 @@ class LightRenderData {
         // scissor rectangle for the shadow rendering to the texture (x, y, width, height)
         this.shadowScissor = new Vec4(0, 0, 1, 1);
 
+        // depth range for cascaded shadows
+        this.depthRange = 0;
```
mvaligursky, line 69: this should be this.depthRange = new Vec2(); instead

src/scene/renderer/shadow-renderer-directional.js
```diff
@@ -173,6 +173,7 @@ class ShadowRendererDirectional {
             // calculate depth range of the caster's AABB from the point of view of the shadow camera
             shadowCamView.copy(shadowCamNode.getWorldTransform()).invert();
             const depthRange = getDepthRange(shadowCamView, visibleSceneAabb.getMin(), visibleSceneAabb.getMax());
+            lightRenderData.depthRange = depthRange;
```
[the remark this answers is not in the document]
GSterbrant: Done.

src/scene/light.js
GSterbrant: Fixed.

src/scene/renderer/shadow-renderer-directional.js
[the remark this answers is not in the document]
mvaligursky: or even better, we should just store `max-min` value here as that is all we need.

src/scene/light.js
```diff
@@ -65,6 +65,9 @@ class LightRenderData {
         // scissor rectangle for the shadow rendering to the texture (x, y, width, height)
         this.shadowScissor = new Vec4(0, 0, 1, 1);
 
+        // depth range for cascaded shadows
```
mvaligursky, line 68: note: this is not only for cascaded shadows, but for all directional lights

src/scene/renderer/shadow-renderer-directional.js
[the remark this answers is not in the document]
GSterbrant: Yeah true, changed it so we calculate the value here directly and just pass it to the forward-renderer.

src/scene/light.js
GSterbrant: Fixed this comment.

mvaligursky (approved):

PR #5442, opened 2023-06-28
Morph: fix options handling and targets validation

kungfooman:
This PR is fixing two issues:

1) `Morph#preferHighPrecision` being `undefined`, because:

https://github.com/playcanvas/engine/blob/1855a582409e35480fe48bc5d3a473597ae149dd/src/framework/parsers/glb-parser.js#L888-L890

Because `assetOptions.morphPreferHighPrecision` is `undefined`.

The intention of the code is clear, but the objects aren't merged as probably intended here. Once an object is given, it will be used as-is no matter what, without any kind of internal merging. So we end up with `undefined` instead of `boolean` (which will make all `=== true/false` fail all the time).

2) Not enough AST shaking to remove the `Debug.assert` from `build/playcanvas.js`:

image


willeastcott (approved):
Nice fix! Thank you! Approved...but I'll leave it to @mvaligursky to review+merge.

PR #5445, opened 2023-06-28
Custom material type

GSterbrant:
The custom material types allows for the user to provide a frontend shader chunk as code, as well as supplying the backend options without having to go through the StandardMaterial type to do so. 

This should finish the last step of #4250. 

An example use of this material can look as such:
```js
const material = new pc.CustomMaterial();
material.setParameter("texture_envAtlas", assets.helipad.resource);
material.setParameter("material_reflectivity", 1.0);
const options = new pc.LitOptions();
options.shadingModel = pc.SPECULAR_BLINN;
options.useSpecular = true;
options.useMetalness = true;
options.occludeSpecular = 1;
options.reflectionSource = 'envAtlas';
options.reflectionEncoding = assets.helipad.resource.encoding;
options.ambientSource = 'envAtlas';
options.ambientEncoding = assets.helipad.resource.encoding;
options.clusteredLightingEnabled = app.scene.clusteredLightingEnabled;
material.litOptions = options;

const shaderChunk = `
void evaluateFrontend() {
    litArgs_emission = vec3(0, 0, 0);
    litArgs_albedo = dVertexNormalW;
    litArgs_metalness = 0.5;
    litArgs_specularity = vec3(1,1,1);
    litArgs_specularityFactor = 1.0;
    litArgs_ior = 0.6667;
    litArgs_gloss = 0.5;
    litArgs_worldNormal = dVertexNormalW;
    litArgs_ao = 0.0;
    litArgs_opacity = 1.0;
}`;
material.shaderChunk = shaderChunk;
material.update();

// create primitive
const primitive = new pc.Entity();
primitive.addComponent('render', {
    type: "sphere",
    material: material
});
```

Note how the options are only for the lighting part of the shader, and the way arguments are passed to it is through a custom bit of code that evaluates the lit shader arguments.

### DIFF
--- a/examples/src/examples/graphics/custom-material.tsx
+++ b/examples/src/examples/graphics/custom-material.tsx
@@ -91,7 +91,7 @@ class CustomMaterialExample {
                 const light = new pc.Entity();
                 light.addComponent("light", {
                     type: "omni",
-                    color: new pc.Color(1, 1, 1),
+                    color: pc.Color.WHITE,
                     intensity: 1,
                     range: 100
                 });
--- a/examples/src/examples/graphics/custom-material.tsx
+++ b/examples/src/examples/graphics/custom-material.tsx
@@ -1,0 +1,147 @@
+import React from 'react';
+
+import * as pc from '../../../../';
+import { Panel, Button } from '@playcanvas/pcui/react';
+import { Observer } from '@playcanvas/observer';
+
+class CustomMaterialExample {
+    static CATEGORY = 'Graphics';
+    static NAME = 'Custom Material';
+
+    controls(data: Observer) {
+        return <></>;
+    }
+
+    example(canvas: HTMLCanvasElement, deviceType: string, data: any): void {
+
+        const assets = {
+            orbitCamera: new pc.Asset('script', 'script', { url: '/static/scripts/camera/orbit-camera.js' }),
+            helipad: new pc.Asset('helipad-env-atlas', 'texture', { url: '/static/assets/cubemaps/helipad-env-atlas.png' }, { type: pc.TEXTURETYPE_RGBP }),
+            font: new pc.Asset('font', 'font', { url: '/static/assets/fonts/arial.json' }),
+            checkerboard: new pc.Asset('checkerboard', 'texture', { url: '/static/assets/textures/checkboard.png' })
+        };
+
+        pc.createGraphicsDevice(canvas).then((device: pc.GraphicsDevice) => {
+
+            const createOptions = new pc.AppOptions();
+            createOptions.graphicsDevice = device;
+            createOptions.mouse = new pc.Mouse(document.body);
+            createOptions.touch = new pc.TouchDevice(document.body);
+            createOptions.keyboard = new pc.Keyboard(document.body);
+
+            createOptions.componentSystems = [
+                // @ts-ignore
+                pc.RenderComponentSystem,
+                // @ts-ignore
+                pc.CameraComponentSystem,
+                // @ts-ignore
+                pc.LightComponentSystem,
+                // @ts-ignore
+                pc.ScriptComponentSystem,
+                // @ts-ignore
+                pc.ElementComponentSystem
+            ];
+            createOptions.resourceHandlers = [
+                // @ts-ignore
+                pc.TextureHandler,
+                // @ts-ignore
+                pc.ContainerHandler,
+                // @ts-ignore
+                pc.ScriptHandler,
+                // @ts-ignore
+                pc.JsonHandler,
+                // @ts-ignore
+                pc.FontHandler
+            ];
+
+            const app = new pc.AppBase(canvas);
+            app.init(createOptions);
+
+            // Set the canvas to fill the window and automatically change resolution to be the same as the canvas size
+            app.setCanvasFillMode(pc.FILLMODE_FILL_WINDOW);
+            app.setCanvasResolution(pc.RESOLUTION_AUTO);
+
+            const assetListLoader = new pc.AssetListLoader(Object.values(assets), app.assets);
+            assetListLoader.load(() => {
+                app.start();
+
+                app.scene.envAtlas = assets.helipad.resource;
+                app.scene.clusteredLightingEnabled = false;
+
+                // Create an Entity with a camera component
+                const camera = new pc.Entity();
+                camera.addComponent("camera", {
+                    clearColor: new pc.Color(0.4, 0.45, 0.5)
+                });
+                camera.addComponent("script");
+                camera.script.create("orbitCamera", {
+                    attributes: {
+                        inertiaFactor: 0.2,
+                        distanceMin: 8,
+                        distanceMax: 50
+                    }
+                });
+                camera.script.create("orbitCameraInputMouse");
+                camera.script.create("orbitCameraInputTouch");
+                camera.translate(0, 1, 4);
+                camera.lookAt(0, 0, 0);
+                app.root.addChild(camera);
+
+                // Create an Entity with a omni light component and a sphere model component.
+                const light = new pc.Entity();
+                light.addComponent("light", {
+                    type: "omni",
+                    color: new pc.Color(1, 1, 1),
+                    intensity: 1,
+                    range: 100
+                });
+                light.translate(0, 3, 0);
+                app.root.addChild(light);
+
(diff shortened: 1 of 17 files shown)


src/scene/materials/custom-material.js
```diff
@@ -0,0 +1,79 @@
...
+import { Material } from './material.js';
+import { custom } from '../shader-lib/programs/custom.js';
+
+/**
+ * A custom material provides a custom front-end part of a lit material
+ *
+ * @augments Material
+ */
+
+
```
willeastcott, line 15: Remove these two blank lines (currently the JSDoc block is detached from the class).

src/scene/materials/custom-material.js
```diff
@@ -0,0 +1,79 @@
...
+class CustomMaterial extends Material {
+    /**
+     * Create a new custom material instance.
+     */
+    constructor() {
+        super();
+
+        this._litOptions = new LitOptions();
+        this._argumentsChunk = "";
```
willeastcott, line 24: You can make these class fields and drop the constructor.

examples/src/examples/graphics/custom-material.tsx
```diff
@@ -0,0 +1,147 @@
...
+                camera.script.create("orbitCameraInputTouch");
+                camera.translate(0, 1, 4);
+                camera.lookAt(0, 0, 0);
+                app.root.addChild(camera);
+
+                // Create an Entity with a omni light component and a sphere model component.
+                const light = new pc.Entity();
+                light.addComponent("light", {
+                    type: "omni",
+                    color: new pc.Color(1, 1, 1),
```
mvaligursky, line 94:
```suggestion
                    color: pc.Color.WHITE,
```

examples/src/examples/graphics/custom-material.tsx
```diff
@@ -0,0 +1,147 @@
...
+            // Set the canvas to fill the window and automatically change resolution to be the same as the canvas size
+            app.setCanvasFillMode(pc.FILLMODE_FILL_WINDOW);
+            app.setCanvasResolution(pc.RESOLUTION_AUTO);
+
+            const assetListLoader = new pc.AssetListLoader(Object.values(assets), app.assets);
+            assetListLoader.load(() => {
+                app.start();
+
+                app.scene.envAtlas = assets.helipad.resource;
+                app.scene.clusteredLightingEnabled = false;
```
mvaligursky, line 69: I assume it works with clustered as well? Worth double checking.

src/scene/shader-lib/programs/custom.js
```diff
@@ -0,0 +1,83 @@
...
+                        const light = options.litOptions.lights[i];
+                        if (!isClustered || light._type === LIGHTTYPE_DIRECTIONAL) {
+                            key += light.key;
+                        }
+                    }
+                } else {
+                    key += m + options.litOptions[m];
+                }
+            }
+        }
```
mvaligursky, line 46: Perhaps this block should be moved to some shared file, as we already have it - it could be a function that returns they key for the lighting, and we do `key += thisfunction();`

GSterbrant: Sure, this is basically key generation for materials with lights.

Maksims:
Will it be possible to see a new API classes and methods and example usage in first post of this PR?

examples/src/examples/graphics/custom-material.tsx
GSterbrant: Yeah, tested it and it works just fine. Enabled WebGPU on this example as well to make sure that works too.

src/scene/materials/custom-material.js
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @augments Material
+ */
+
+class CustomMaterial extends Material {
+    _litOptions = new LitOptions();
+
+    _argumentsChunk = "";
+
+    set litOptions(value) {
```
mvaligursky, line 20: this should have JSDocs as it's a public API

src/scene/materials/custom-material.js
```diff
@@ -0,0 +1,72 @@
...
+        this._litOptions = value;
+    }
+
+    get litOptions() {
+        return this._litOptions;
+    }
+
+    set argumentsChunk(value) {
```
mvaligursky, line 28: this should have JSDocs as it's a public API

mvaligursky: I'd be keen to find some better name here. Perhaps `customChunk`, to match the CustomMaterial. @willeastcott any suggestions? Could also be `materialChunk`, I think I like this even better, or maybe even just `chunk`.

src/scene/materials/custom-material.js
```diff
@@ -0,0 +1,72 @@
...
+        collectLights(LIGHTTYPE_SPOT, sortedLights[LIGHTTYPE_SPOT], lightsFiltered, mask, staticLightList);
+
+        this._litOptions.lights = lightsFiltered;
+        options.customLitArguments = this._argumentsChunk;
+
+        const processingOptions = new ShaderProcessorOptions(viewUniformFormat, viewBindGroupFormat, vertexFormat);
+
+        const library = getProgramLibrary(device);
+        library.register('custom', custom);
+        const shader = library.getProgram('custom', options, processingOptions);
```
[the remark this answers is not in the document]
mvaligursky: ah we do, I see options.customLitArguments being assigned, all good

src/scene/shader-lib/programs/custom.js
```diff
@@ -0,0 +1,62 @@
...
+
+        // global texture bias for standard textures
+        decl.append(`uniform float textureBias;`);
+
+        decl.append(litShader.chunks.litShaderArgsPS);
+        code.append(options.customLitArguments);
+        func.code = `evaluateFrontend();`;
+
+        func.code = `\n${func.code.split('\n').map(l => `    ${l}`).join('\n')}\n\n`;
+        const useUv = [];
```
mvaligursky, line 52: should there be a way to add something to this array? Can the user use multiple UVs in his front end chunk?

mvaligursky:
This looks pretty good! I would not mind to add additional more complex custom material into the example, something that uses the textures as well, to show / test how this works. Perhaps using at least 2 UVs, textures with different encoding and similar.

src/scene/shader-lib/programs/custom.js
GSterbrant: Yeah, added support for generating code for multiple UVs.

src/scene/materials/custom-material.js
```diff
@@ -0,0 +1,105 @@
...
+import { LitOptions } from "./lit-options";
+import { collectLights } from "./lit-material-common.js";
+import { Material } from './material.js';
+import { custom } from '../shader-lib/programs/custom.js';
+
+/**
+ * A custom material provides a custom front-end part of a lit material
+ *
+ * @ignore
+ * @augments Material
```
mvaligursky, line 13: we don't need this line anymore since we use classes. @willeastcott - should we strip all these from the codebase or do they serve some purpose in ES6 world?

src/scene/materials/custom-material.js
```diff
@@ -0,0 +1,105 @@
+import { ShaderProcessorOptions } from '../../platform/graphics/shader-processor-options.js';
+import { LIGHTTYPE_DIRECTIONAL, LIGHTTYPE_OMNI, LIGHTTYPE_SPOT, MASK_AFFECT_DYNAMIC, SHADERDEF_INSTANCING, SHADERDEF_MORPH_NORMAL, SHADERDEF_MORPH_POSITION, SHADERDEF_MORPH_TEXTURE_BASED, SHADERDEF_SCREENSPACE, SHADERDEF_SKIN } from "../constants";
+import { getProgramLibrary } from "../shader-lib/get-program-library";
+import { LitOptions } from "./lit-options";
+import { collectLights } from "./lit-material-common.js";
+import { Material } from './material.js';
+import { custom } from '../shader-lib/programs/custom.js';
+
+/**
+ * A custom material provides a custom front-end part of a lit material
```
mvaligursky, line 10: adjust the description here to be more in line with `shaderChunk` property - so no `front-end part` for example.

mvaligursky (approved):
Approving, but would like @slimbuck 's opinion here as well.

slimbuck (approved):

Issue #5444, opened 2023-06-28
PCSS shadow error


slimbuck:
Setting PCSS shadow on a point light in clustered lighting mode will result in the shadow breaking. (The light's own shadow type should be completely ignored in clustered lighting mode).

PR #5446, opened 2023-06-28
Fix omni light shadows with PCSS and clustered

GSterbrant:
Fixes https://github.com/playcanvas/engine/issues/5444

Omni light shadows when using PCSS would try to output to PCSS style R32F but reading D32F. This PR fixes that bug and also refactors the way we select whether or not we output to depth component or not more clearly.

### DIFF
--- a/src/scene/shader-lib/programs/lit-shader.js
+++ b/src/scene/shader-lib/programs/lit-shader.js
@@ -450,7 +450,7 @@ class LitShader {
 
         // If not a directional light and using clustered, fall back to using PCF3x3 if shadow type isn't supported
         if (lightType !== LIGHTTYPE_DIRECTIONAL && options.clusteredLightingEnabled) {
-            if (shadowType !== SHADOW_PCF1 || shadowType !== SHADOW_PCF3 || shadowType !== SHADOW_PCF5) {
+            if (shadowType === SHADOW_VSM8 || shadowType === SHADOW_VSM16 || shadowType === SHADOW_VSM32 || shadowType === SHADOW_PCSS) {
                 shadowType = SHADOW_PCF3;
             }
         }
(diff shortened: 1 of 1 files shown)


mvaligursky:
It is not related, I suspect that project has a custom shadow catcher chunk override. @GSterbrant - could you please fix it in the project.

src/scene/shader-lib/programs/lit-shader.js
```diff
@@ -520,9 +520,16 @@ class LitShader {
             code += "    gl_FragColor = packFloat(depth);\n";
         } else if (!isVsm) {
             code += "    gl_FragColor = vec4(1.0);\n"; // just the simplest code, color is not written anyway
+            const exportDepth = shadowType === SHADOW_PCF1 || shadowType === SHADOW_PCF3 || shadowType === SHADOW_PCF5 ||
```
mvaligursky, line 523: do we need to handle this differently for WebGl1?
Also, just realised, we'd need to test this (including PCSS) on other iOS devices like iPhone 6s I think that use WebGL2 but I'm not sure they can render to Float textures, but can render to half float, of the other way around, not sure now.

GSterbrant: I think for GL1 we would enter the packed depth flow.

GSterbrant:
@kungfooman The signature of getShadowPCF3x3 has changed in a previous PR to accept a vec4 for the shadowParams. That's probably what breaks your code 🙂

GSterbrant:
@mvaligursky @slimbuck Fixed our project, but @kungfooman has a fork of it, so he needs to move the fix over as well.

src/scene/shader-lib/programs/lit-shader.js
mvaligursky: Can you please confirm this and test all shadow options you can on gl1 and 2

mvaligursky (approved):
All seems fine, logical to what is expected.

PR #5451, opened 2023-06-30
Don't write to FragDepth unless depth value is modified

GSterbrant:
Fixed some blinking issues seen with WebGPU.

Not sure what the cause of this is yet, but this PR makes sure the shadow output shaders export the same values as they did before the shadow export refactor in #5446


mvaligursky (approved):
great, we don't want to write out depth unless we have to, as that comes at a cost.

PR #5456, opened 2023-07-03
Improves PCSS with directional light

GSterbrant:
PCSS with directional lights doesn't take into account the fact that the scene dimensions might affect the orthographic projection, which in turn has an impact on the spread of the penumbra soft shadows. 

This PR accounts for that in two ways, the first is by anchoring the orthographic projection radius by 10, making smaller projections get bigger sample kernels and vice versa.

The other factor is depth range, which is applied to the distance in depth between receiver and shadow map, and determines the width of the filter kernel used to take the final shadow samples.


mvaligursky:
We need to add attribution to both added glb files - see .txt files already existing in the folder.

PR #5463, opened 2023-07-06
Fixes Adreno 610 and 618 bugs

GSterbrant:
Moves the lit shader arguments out of the struct and in to globals. The retain the same naming strategy, with litShaderArgs being the first level, and the clearcoat/sheen/iridescence extensions being another layer, so for example litShaderArgs_sheen_gloss gets the sheen gloss value.

Also splits the metalness modulate into two functions, one for albedo and one for specular.

For whatever reason, the Adreno 610 and 618 devices we have tested has some bug which is provoked by having the lit shader arguments in structs instead of globals. Why this happens is not clear, but it seems to be caused by a bug in the shader compiler on the target, rather than the shader we generate. The issue in question is demonstrated here: https://forum.playcanvas.com/t/strange-rendering-in-android-13/31622/20.

This PR brings with it some chunk changes, but they should be rather minor since we rarely pass the entire LitShaderArgs struct around.

### DIFF
--- a/src/scene/shader-lib/chunks/standard/frag/litShaderArgs.js
+++ b/src/scene/shader-lib/chunks/standard/frag/litShaderArgs.js
@@ -1,24 +1,63 @@
 export default /* glsl */`
 
+// Normal direction in world space
 vec3 litShaderArgs_worldNormal;
+
+// Transparency
 float litShaderArgs_opacity;
+
+// Surface albedo absorbance
 vec3 litShaderArgs_albedo;
+
+// Transmission factor (refraction), range [0..1]
 float litShaderArgs_transmission;
+
+// The f0 specularity factor
 vec3 litShaderArgs_specularity;
+
+// Uniform thickness of medium, used by transmission, range [0..inf]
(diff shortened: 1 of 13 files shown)


src/scene/shader-lib/chunks/standard/frag/litShaderArgs.js
```diff
@@ -1,82 +1,24 @@
 export default /* glsl */`
 
-struct IridescenceArgs
-{
-    // Iridescence effect intensity, range [0..1]
-    float intensity;
+vec3 litShaderArgs_worldNormal;
```
mvaligursky, line 3: please keep the comments from the struct

mvaligursky (approved):
Looks good to me. It's not an ideal solution, but we cannot win against the buggy shader compilers unfortunately. @slimbuck - thoughts here?

GSterbrant:
Did the rename

kungfooman:
I believe this PR broke ar-ground once more in the holometric video example:

image

Issue #5458, opened 2023-07-04
Setting `Scene#skyBoxRotation` invalidate all shader variants


willeastcott:
For the 3D Tiles project, this causes about 80% of a frame to be spent on regenerating shaders:

![image](https://github.com/playcanvas/engine/assets/697563/31c06ada-855e-4525-9904-f3b297af7e1f)

I can work around this by now calling `Scene#skyBoxRotation` every frame, but this engine behavior is currently way to aggressive.

mvaligursky:
Note to myself (or whoever will fix this): `set skyboxRotation` function calls `_resetSky` which sets `updateShaders = true;`. This is used to regenerate shaders that depend on skybox rotation. I think if the matrix is identity, the shader chunk is skipped. This should be matched here and resetSky should perhaps be only called when the matrix switches between identity and non-dentity.

PR #5467, opened 2023-07-07
Only rebuild shaders on sky rotation change if changing away from identity

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/5458

Don't rebuild shaders if rotation changes, only if it changes and it was identity before, which requires shaders to be re-generated to include support for rotation, which is excluded when identity.


src/scene/scene.js
```diff
@@ -626,14 +626,23 @@ class Scene extends EventHandler {
...
                 this._skyboxRotationMat3.setIdentity();
             } else {
                 this._skyboxRotationMat4.setTRS(Vec3.ZERO, value, Vec3.ONE);
                 this._skyboxRotationMat4.invertTo3x3(this._skyboxRotationMat3);
             }
-            this._resetSky();
+
+            // only reset sky / rebuild scene shaders if rotation changed from identity, to include
+            // support for skybox rotation in shaders
+            if (prevIdentity && !newIdentity) {
```
slimbuck, line 643: This should either do `prevIdentity !== newIdentity` or scene should store a flag `skyboxHasRotation` which tracks first time rotation is set (and then doesn't reset again).

mvaligursky: I don't want to rebuild shaders if we go from rotation to no rotation.

slimbuck: Right, but now you're rebuilding every time you go no->rotation. I think you want a flag that says 'we've compiled rotation' and only do that once.

mvaligursky: ok this is done, thanks

PR #5469, opened 2023-07-08
Implement/use `GraphNode#remove`

kungfooman:
This is a follow-up of this PR, to make it a bit cleaner/shorter:

https://github.com/playcanvas/engine/pull/5414#issuecomment-1598473727

`GraphNode#remove` works just like `HTMLElement#remove`:

```js
const div = document.createElement("div");
div.remove();
div.remove();
document.body.append(div);
div.remove(); // only call which does something
div.remove();
```

Equivalent API:

```js
const ent = new pc.Entity();
ent.remove();
ent.remove();
pc.app.root.addChild(ent);
ent.remove(); // only call which does something
ent.remove();
```

### DIFF
--- a/src/framework/components/sprite/component.js
+++ b/src/framework/components/sprite/component.js
@@ -596,11 +596,8 @@ class SpriteComponent extends Component {
         this._hideModel();
         this._model = null;
 
-        if (this._node) {
-            if (this._node.parent)
-                this._node.parent.removeChild(this._node);
-            this._node = null;
-        }
+        this._node?.remove();
+        this._node = null;
 
         if (this._meshInstance) {
             // make sure we decrease the ref counts materials and meshes
--- a/src/scene/graph-node.js
+++ b/src/scene/graph-node.js
@@ -490,7 +490,7 @@ class GraphNode extends EventHandler {
      */
     destroy() {
         // Detach from parent
-        this._parent?.removeChild(this);
+        this.remove();
 
         // Recursively destroy all children
         const children = this._children;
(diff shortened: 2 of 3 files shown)


LeXXik:
Shouldn't this be a private API? I mean, we already have `entity.destroy()`, which presumably does the same?

kungfooman:
> Shouldn't this be a private API? I mean, we already have `entity.destroy()`, which presumably does the same?

No, this reflects how HTML works:

```js
const div = document.createElement("div");
div.innerHTML = "innerhtml of div";
div.remove();
div.remove();
document.body.append(div);
div.remove();
div.remove();
div.remove();
div.remove();
document.body.append(div);
div.remove();
document.body.append(div);
```

You can parent/unparent a `HTMLElement`/`GraphNode` as much as you want, it's not destroyed. Just because you remove a cookie out of the box, it's not destroyed/eaten. You can just put it back or put it in another box.

Just like you can have this code in HTML, you can do the same in PlayCanvas... `destroy` of course also unparents your entity, but completely destroys it recursively aswell (hence the name implying destruction). `GraphNode#remove` is simply removing it from the scene graph.

slimbuck (approved):
Nice!

LeXXik:
Ah, that makes sense! Thank you for the clarification, @kungfooman !

mvaligursky (approved):
Looks great to me, but it'd be nice to add few tests to graph-node.test.mjs

willeastcott (approved):

willeastcott:
Ah, yeah, good point about the unit tests @mvaligursky. Other than that, looks great. Thanks! 🙏

kungfooman:
Thank you all for the quick review! I just added a unit test (and I hope the texting is okay).

PR #5480, opened 2023-07-14
Optimize AssetRegistry.find and findAll

Maksims:
Improves `AssetRegistry.find` and `findAll` performance, by building and managing internal index of assets by names. And avoids casting large asset lists to arrays during these methods to avoid GC stalls.

In a project with around 2.5k of assets, calling `find` was taking around 0.04ms per request. With this patch it takes around 0.00006ms. This is around ~666 times improvement.
`findAll` used to take 0.04ms per request also, now it takes 0.0006ms. ~66 times improvement.

**In specific project, during custom scene loaders that reference assets by names, it used to take around 200ms just finding assets, now it takes 0.25ms.**

_Timings are just a relative measurement and based on hardware._
_It only highlights the performance change between versions._
_Tested on high-end computer, so on mid/low/mobile it is even more apparent improvement._

### DIFF
--- a/src/framework/asset/asset-registry.js
+++ b/src/framework/asset/asset-registry.js
@@ -52,6 +52,12 @@ class AssetRegistry extends EventHandler {
     _urlToAsset = new Map();
 
     /**
+     * @type {Map<string, Set<Asset>>}
+     * @private
+     */
+    _nameToAsset = new Map();
+
+    /**
      * Index for looking up by tags.
      *
      * @private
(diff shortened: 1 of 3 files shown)


willeastcott (approved):

src/framework/asset/asset-registry.js
```diff
@@ -649,7 +698,11 @@ class AssetRegistry extends EventHandler {
      * console.log(`Found ${assets.length} texture assets named 'brick'`);
      */
     findAll(name, type) {
-        return Array.from(this._assets).filter(asset => asset.name === name && (!type || asset.type === type));
+        const items = this._nameToAsset.get(name);
+        if (!items) return [];
+        let results = Array.from(items);
```
willeastcott, line 703:
```suggestion
        const results = Array.from(items);
```

willeastcott:
Nicely done - thanks!! 🙏

PR #5483, opened 2023-07-14
Address PCSS softness

GSterbrant:
Fixes https://github.com/playcanvas/engine/issues/7196

After reviewing the PCSS algorithm some more, I came to the conclusion that the reason why directional lights don't get a nice hardening effect is due to the normalization factor, usually the average of the blocker distances, is too large. This makes sense for directional lights seeing as the average blocker distance essentially becomes the distance between the light origin, which for directional lights is indeed very big, and the blocker, resulting in the normalization of receiver/blocker becoming very small indeed. 

To fix this, we first need to output depth in linear space, which this PR reimplements for PCSS. Then, we need a normalization factor for directional lights which is right now hard coded to be 3 meters. This factor should be provided as a parameter to the light somehow. 

This is the result:
image


src/scene/renderer/forward-renderer.js
```diff
@@ -258,13 +258,11 @@ class ForwardRenderer extends Renderer {
...
 
-                const projectionCompensation = (50.0 / lightRenderData.projectionCompensation);
-                const pixelsPerMeter = 1.0 / (lightRenderData.shadowCamera.renderTarget.width / directional.penumbraSize);
-                this.lightShadowSearchAreaId[cnt].setValue(pixelsPerMeter * projectionCompensation);
+                this.lightShadowSearchAreaId[cnt].setValue(directional.penumbraSize / lightRenderData.shadowCamera.renderTarget.width * lightRenderData.projectionCompensation);
 
                 const cameraParams = directional._shadowCameraParams;
                 cameraParams.length = 4;
-                cameraParams[0] = lightRenderData.depthRangeCompensation;
+                cameraParams[0] = 3.0; // unused
```
willeastcott, line 265: @mvaligursky Need to figure out what to do here, presumably. 😄

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
```diff
@@ -49,10 +49,11 @@ float viewSpaceDepth(float depth, mat4 invProjection) {
     return viewSpace.z;
 }
 
-float PCSSBlockerDistance(TEXTURE_ACCEPT(shadowMap), vec2 sampleCoords[PCSS_SAMPLE_COUNT], vec2 shadowCoords, vec2 searchSize, float z) {
+float PCSSBlockerDistance(TEXTURE_ACCEPT(shadowMap), vec2 sampleCoords[PCSS_SAMPLE_COUNT], vec2 shadowCoords, vec2 searchSize, float z, vec4 cameraParams) {
 
     float blockers = 0.0;
     float averageBlocker = 0.0;
+    //float radius = searchSize * ( z - )
```
willeastcott, line 56: Delete commented out code.

src/scene/shader-lib/chunks/lit/frag/shadowPCSS.js
```diff
@@ -64,20 +65,16 @@ float PCSSBlockerDistance(TEXTURE_ACCEPT(shadowMap), vec2 sampleCoords[PCSS_SAMP
     #endif        
         float isBlocking = step(blocker, z);
         blockers += isBlocking;
-        averageBlocker += blocker * isBlocking;
+        averageBlocker += blocker * isBlocking;
```
willeastcott, line 68:
```suggestion
        averageBlocker += blocker * isBlocking;
```

mvaligursky:
Updated this with main and slightly modified to be up to date. Still a question about the hardcoded constant / working for different sized worlds, but this is an improvement and we should merge it and resolve those separately.

current main:
Screenshot 2024-12-23 at 15 30 07

changes in this PR:
Screenshot 2024-12-23 at 15 26 48

and PCF for comparison:
Screenshot 2024-12-23 at 15 26 55

willeastcott:
Yeah definitely an improvement over what's currently in main!!

willeastcott (approved):

PR #5484, opened 2023-07-17
Refactor EventHandler class

willeastcott:
This PR refactors the `EventHandler` class. It gets rid of the `_callbacksActive` property and generally simplifies the code but maintains the same behavior and (hopefully) the same performance.

It also lifts the limit on passing 8 arguments to an event handler.

### DIFF
--- a/src/core/event-handler.js
+++ b/src/core/event-handler.js
@@ -1,15 +1,8 @@
 /**
- * Callback used by {@link EventHandler} functions. Note the callback is limited to 8 arguments.
+ * Callback used by {@link EventHandler} functions.
  *
  * @callback HandleEventCallback
- * @param {*} [arg1] - First argument that is passed from caller.
- * @param {*} [arg2] - Second argument that is passed from caller.
- * @param {*} [arg3] - Third argument that is passed from caller.
- * @param {*} [arg4] - Fourth argument that is passed from caller.
- * @param {*} [arg5] - Fifth argument that is passed from caller.
- * @param {*} [arg6] - Sixth argument that is passed from caller.
- * @param {*} [arg7] - Seventh argument that is passed from caller.
- * @param {*} [arg8] - Eighth argument that is passed from caller.
+ * @param {...*} args - Arguments that are passed to the event handler.
  */
 
 /**
@@ -29,25 +22,18 @@
  */
 class EventHandler {
     /**
-     * @type {object}
+     * @type {Map<string, {callback: HandleEventCallback, scope: object, once: boolean}[]>}
      * @private
      */
-    _callbacks = {};
-
-    /**
-     * @type {object}
-     * @private
-     */
-    _callbackActive = {};
+    _callbacks = new Map();
 
     /**
      * Reinitialize the event handler.
      *
-     * @private
+     * @ignore
      */
     initEventHandler() {
-        this._callbacks = {};
-        this._callbackActive = {};
+        this._callbacks = new Map();
     }
 
     /**
@@ -56,26 +42,20 @@ class EventHandler {
      * @param {string} name - Name of the event to bind the callback to.
      * @param {HandleEventCallback} callback - Function that is called when event is fired. Note
      * the callback is limited to 8 arguments.
-     * @param {object} [scope] - Object to use as 'this' when the event is fired, defaults to
-     * current this.
-     * @param {boolean} [once=false] - If true, the callback will be unbound after being fired once.
+     * @param {object} scope - Object to use as 'this' when the event is fired, defaults to current
+     * this.
+     * @param {boolean} once - Set to true to have the handler remove itself after being called for
+     * the first time.
      * @private
      */
-    _addCallback(name, callback, scope, once = false) {
-        if (!name || typeof name !== 'string' || !callback)
+    _addCallback(name, callback, scope, once) {
+        if (!name || typeof name !== 'string' || typeof callback !== 'function')
             return;
-
-        if (!this._callbacks[name])
-            this._callbacks[name] = [];
-
-        if (this._callbackActive[name] && this._callbackActive[name] === this._callbacks[name])
-            this._callbackActive[name] = this._callbackActive[name].slice();
-
-        this._callbacks[name].push({
-            callback: callback,
-            scope: scope || this,
-            once: once
-        });
+    
+        if (!this._callbacks.has(name))
+            this._callbacks.set(name, []);
+    
+        this._callbacks.get(name).push({ callback, scope, once });
     }
 
     /**
@@ -93,9 +73,29 @@ class EventHandler {
      * });
      * obj.fire('test', 1, 2); // prints 3 to the console
      */
-    on(name, callback, scope) {
+    on(name, callback, scope = this) {
         this._addCallback(name, callback, scope, false);
+        return this;
+    }
 
+    /**
+     * Attach an event handler to an event. This handler will be removed after being fired once.
+     *
+     * @param {string} name - Name of the event to bind the callback to.
+     * @param {HandleEventCallback} callback - Function that is called when event is fired. Note
+     * the callback is limited to 8 arguments.
+     * @param {object} [scope] - Object to use as 'this' when the event is fired, defaults to
+     * current this.
+     * @returns {EventHandler} Self for chaining.
+     * @example
+     * obj.once('test', function (a, b) {
+     *     console.log(a + b);
+     * });
+     * obj.fire('test', 1, 2); // prints 3 to the console
+     * obj.fire('test', 1, 2); // not going to get handled
+     */
+    once(name, callback, scope = this) {
+        this._addCallback(name, callback, scope, true);
         return this;
     }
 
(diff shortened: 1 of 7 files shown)


willeastcott:
@Maksims Do you have any thoughts on this change? I think you were the last person to make significant changes here. 😄

kungfooman:
I'm *very* happy about this... nothing about the code made sense to me and I felt sure it's just technical debt... 4x senseless shallow array cloning etc.

Please make sure to fix this too:

https://github.com/playcanvas/engine/blob/3e550768b1901e35070e865c44f1e7bd618d9070/src/core/events.js#L14-L25

src/core/event-handler.js
```diff
@@ -165,80 +148,31 @@ class EventHandler {
...
-
-        if (!this._callbackActive[name]) {
-            this._callbackActive[name] = this._callbacks[name];
-        } else {
-            if (this._callbackActive[name] === this._callbacks[name])
-                this._callbackActive[name] = this._callbackActive[name].slice();
-
-            callbacks = this._callbacks[name].slice();
-        }
+    fire(name, ...args) {
```
Maksims, line 156: When this transpired into ES5, does it use `arguments`? If it does, then, unfortunately, this will lead to a lack of V8 optimization on this method, unless V8 team made a change here. If it still the case, then calling `fire` will lead to always unoptimized JS under V8, making it significantly slower in some cases.

Maksims:
The current implementation focuses on a few major things: avoid V8 deoptimization due to the use of `arguments` keyword, and ensure there are no unnecessary allocations. And modifying callbacks during the execution of a callback is safe and has no magical behaviors.

This PR will introduce a lot of breaking changes due to the way the callback array is managed and can lead to infinite callback execution when within callback another event is added of the same name.
Also removing an event during event execution will lead to a change of handlers, which will not work as intended too.
A brutforce solution to this is to clone callbacks on each `fire`, which was the code before. And after memory and performance profiling we identified it to be a significant issue, so I've previously optimized this code to clone array **only** if the callbacks array gets modified during the execution of the callback.

**This PR leads to breaking changes as it changes logic of callbacks execution.**

If you still insist on these changes, please create a set of tests for the current implementation, and measure the memory and performance on large projects.

src/core/event-handler.js
```diff
@@ -165,80 +148,31 @@ class EventHandler {
...
+        if (!name || !this._callbacks.has(name)) return this;
 
-        // TODO: What does callbacks do here?
-        // In particular this condition check looks wrong: (i < (callbacks || this._callbackActive[name]).length)
-        // Because callbacks is not an integer
-        // eslint-disable-next-line no-unmodified-loop-condition
-        for (let i = 0; (callbacks || this._callbackActive[name]) && (i < (callbacks || this._callbackActive[name]).length); i++) {
-            const evt = (callbacks || this._callbackActive[name])[i];
-            evt.callback.call(evt.scope, arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8);
+        const handlers = this._callbacks.get(name);
```
Maksims, line 159: This approach uses the callbacks array, which can be modified during the execution, leading to unintended behaviours.

src/core/event-handler.js
kungfooman: Yea, ES5 code looks like this:

```js
		_proto.fire = function fire(name) {
			if (!name || !this._callbacks.has(name)) return this;
			var handlers = this._callbacks.get(name);
			if (!handlers) return this;
			for (var _len = arguments.length, args = new Array(_len > 1 ? _len - 1 : 0), _key = 1; _key < _len; _key++) {
				args[_key - 1] = arguments[_key];
			}
			for (var i = 0; i < handlers.length; i++) {
				var handler = handlers[i];
				handler.callback.apply(handler.scope, args);
				if (handler.once) {
					handlers.splice(i, 1);
					i--;
				}
			}
			if (handlers.length === 0) {
				this._callbacks.delete(name);
			}
			return this;
		};
```

I didn't profile it, but I trust your intuition here, the extra `for`-loop/`Array` creation doesn't look so appealing either.

kungfooman:
> This PR will introduce a lot of breaking changes due to the way the callback array is managed and can lead to infinite callback execution when within callback another event is added of the same name.

I tried to test that, but I didn't see that happening:

```js
animal = new GraphNode("cat");
animal.on("birth", () => {
  console.log("awww");
    animal.on("birth", () => {
      console.log("awww 2");
  });
});
animal.fire("birth");
```

Do you mean like that?

image

The old implementation is acting differently (not calling the newly added birth event:

image

Personally I've never come across this pattern of adding an event on itself with the same name on itself... is anyone doing this and what is the reasoning?

Not tested, but if we require this, instead of creating temporary arrays in multiple places which causes GC, we could save a version/id in the handler. If we call `on`, the id increases and the newly added callback has "id mismatch" - therefore not called, acting like the old version without GC issues.

Maksims:
Events in complex projects and in the engine, their flow of execution and nesting - can be deep and complex.

Often when we see an exception of an event, notice how many events deep it can be. There is no guarantee that developer will know this special rules, and especially facing an issue of infinite events, or trigger of event while somewhere down the line event was added - this is all "magical" and hidden behaviors from developer. And will be almost impossible to debug or understand without knowing implementation specifics.

Current implementation mitigates that, and have straight forward rules of execution of events, without magical behaviors that developer could stumble on.

Maksims:
> The old implementation is acting differently (not calling the newly added birth event):

That is correct behavior. Only events that were present during `fire` should be triggered.

> Not tested, but if we require this, instead of creating temporary arrays in multiple places which causes GC, we could save a version/id in the handler. If we call `on`, the id increases and the newly added callback has "id mismatch" - therefore not called, acting like the old version without GC issues.

This will not work when removing events within callbacks that were present during `fire`, they should happen still.

kungfooman:
Yep, fully agree, I still have a feeling we dont need to cause GC for array cloning here and there.

> > Not tested, but if we require this, instead of creating temporary arrays in multiple places which causes GC, we could save a version/id in the handler. If we call `on`, the id increases and the newly added callback has "id mismatch" - therefore not called, acting like the old version without GC issues.
> 
> This will not work when removing events within callbacks that were present during `fire`, they should be triggered.

Right, this is also fixed when fixating `n` in the fire/for loop.

> This will not work when removing events within callbacks that were present during `fire`, they should be triggered.

Yep, I didn't test that yet, but also agree that we require unit tests for this case.

Maksims:
Here is a quick description of "activeCallback" mechanics:

If we call `fire`, we do not clone an array of callbacks and go through its execution. We note by `name` that we are currently executing list of callbacks.
If at any point while there is active execution of callbacks we try to modify list of callbacks of that name (by calling `on`, `once`, `off`), then we ensure that currently executed list of callbacks elsewhere is cloned and continues on pre-modified list. Then list of callbacks is modified.

**Benefits of this approach:**
1. Eliminates cloning of callbacks list - that is a source of GC stalls.
2. Eliminates unintended magical behaviors that could have occurred if we would modify callbacks list during its execution (which is the case with this PR).
3. It performs better than brute-cloning of callbacks list when accounted for GC stalls performance hit.

This approach worked great for a very long time, and had no issues with GC, performance or bugs. Why change it?

kungfooman:
> This approach worked great for a very long time, and had no issues with GC, performance or bugs. Why change it?

It's kinda simple to proof, just replace every occurrence of:

```js
this._callbackActive[name] = this._callbackActive[name].slice();
```

with:

```js
this.causeGC(name);
```

and implement it:

```js
    causeGC(name) {
        console.log("cause GC for", name);
        this._callbackActive[name] = this._callbackActive[name].slice();
    }
```

Test some code:

image

Looks like GC on every `EventHandler#fire`, right?

willeastcott:
> This approach worked great for a very long time, and had no issues with GC, performance or bugs. Why change it?

I attempted a refactor because:

1. The code is really hard to read and understand. [This comment](https://github.com/playcanvas/engine/blob/main/src/core/event-handler.js#L195-L197) is concerning.
2. The use of an object essentially in place of a Map is bad practice - just use a Map. I won't go into why - lots has been  written about this.
3. I was curious to see if we could boost performance of this class (since it's called so frequently) but it's hard to iterate with optimization experiments when the code is so brittle and hard to follow.
4. It's a shame we impose a limit on arg count to event handlers.
5. We are proposing changes to event management with `EventHandle` so I thought it would make sense we revisited whether we had a solid foundation in the `EventHandler` class. And while I feel it *works* well, I don't think it's perfectly written.

Maksims:
> Looks like GC on every `EventHandler#fire`, right? Of course we change the callbacks via `on` in this case, and usually it doesn't happen? But then again, we never really know, because:

Please read carefully the previous post:

> If at any point while there is active execution of callbacks we try to modify list of callbacks of that name (by calling on, once, off), **then we ensure that currently executed list of callbacks elsewhere is cloned** and continues on pre-modified list. Then list of callbacks is modified.

The Solution before this optimization was cloning on every `fire`. Not cloning the callbacks array - will lead to infinite callbacks and unintended or missed callbacks.

> 1. The code is really hard to read and understand. [This comment](https://github.com/playcanvas/engine/blob/main/src/core/event-handler.js#L195-L197) is concerning.

The comment is:
```js
// TODO: What does callbacks do here?
// In particular this condition check looks wrong: (i < (callbacks || this._callbackActive[name]).length)
// Because callbacks is not an integer
```
The condition is not wrong, the commenter missread the code. There are brackets around two objects. It will either do `callbacks.lenth` or `this._callbackActive[name].length` if `callbacks` is falsy.

> 2\. The use of an object essentially in place of a Map is bad practice - just use a Map. I won't go into why - [lots has been  written](https://www.zhenghao.io/posts/object-vs-map) about this.

Sure, we can replace objects with Maps if it is beneficial.

> 3\. I was curious to see if we could boost performance of this class (since it's called so frequently) but it's hard to iterate with optimization experiments when the code is so brittle and hard to follow.

Good intent. The current implementation has boosted the performance and GC footprint a lot when we moved away from "always cloning" of callbacks. If this PR is updated to ensure backward compatibility and the same events execution behavior, then will be great to see a performance comparison.

> 4\. It's a shame we impose a limit on arg count to event handlers.

I don't see a reason not to increase that number. Using `arguments` will decrease performance as it requires casting it into an array, and as mentioned before, still is in lists of V8 deoptimizations.

> 5\. We are proposing changes to event management with `EventHandle` so I thought it would make sense we revisited whether we had a solid foundation in the `EventHandler` class. And while I feel it _works_ well, I don't think it's perfectly written.

It was written around 8 years ago. We had more ES5 strict rules at the time. I can make a PR to use more modern features while preserving functionality and existing optimizations. With performance tests.

mvaligursky:
Closing for now as per @willeastcott

PR #5486, opened 2023-07-18
Instancing fix

slimbuck:
Fix a bug where the instancing flag wasn't being set correctly on the mesh instance and users had to implement `updateShader` instead.


mvaligursky:
I'd be good to update the manual as well:
https://developer.playcanvas.com/en/user-manual/optimization/hardware-instancing/

mvaligursky (approved):
Nice one!

PR #5488, opened 2023-07-19
Lit material

slimbuck:
This PR makes the following change to materials:
- remove `CustomMaterial` in favour of `LitMaterial` and its accompanying options and builder
- `LitMaterial` is internal and still in flux (but is concrete enough to start testing its usage)
- simplify and organise some of the `LitOptions` and `StandardMaterialOptions` members
- fix various bugs with options flags and shader hash generation
- rename `LitOptions` -> `LitShaderOptions` to be more descriptive and clear

The following work is still outstanding:
- `StandardMaterial` should ideally inherit from `LitMaterial` so options builder code is shared, but achieving this is not obvious right now
- `LitMaterial` doesn't deal with shader parameters at all - it's left for user to figure out
- `LitMaterial.shaderChunk` should probably support a chunk for forward and another for non-forward passes


examples/src/examples/graphics/index.mjs
```diff
@@ -8,7 +8,7 @@ import ClusteredLightingExample from "./clustered-lighting";
 import ClusteredOmniShadowsExample from "./clustered-omni-shadows";
 import ClusteredSpotShadowsExample from "./clustered-spot-shadows";
 import ContactHardeningShadowsExample from "./contact-hardening-shadows";
-import CustomMaterialExample from "./custom-material";
+import LitMaterialExample from "./lit-material";
```
mvaligursky, line 11: we probably want to remove this to avoid the example going public till it's all ready

examples/src/examples/graphics/lit-material.tsx
```diff
@@ -1,18 +1,13 @@
 import * as pc from '../../../../';
-class CustomMaterialExample {
+
+class LitMaterialExample {
     static CATEGORY = 'Graphics';
-    static NAME = 'Custom Material';
+    static NAME = 'Lit Material';
     static HIDDEN = true;
-    static WEBGPU_ENABLED = true;
+    static WEBGPU_ENABLED = false;
```
mvaligursky, line 7: is something not working on WebGPU?

src/index.js
```diff
@@ -151,6 +151,7 @@ export { createShader, createShaderFromCode } from './scene/shader-lib/utils.js'
 export { ProgramLibrary } from './scene/shader-lib/program-library.js';
 export { shaderChunks } from './scene/shader-lib/chunks/chunks.js';
 export { shaderChunksLightmapper } from './scene/shader-lib/chunks/chunks-lightmapper.js';
+export { ChunkBuilder } from './scene/shader-lib/chunk-builder.js';
```
mvaligursky, line 154: is this needed by the Shed or elsewhere?

src/scene/materials/lit-material.js
```diff
@@ -0,0 +1,262 @@
...
+    usedUvs;
+
+    // custom shader chunk to be added to the shader
+    shaderChunk;
+
+    // lit options
+    litOptions = new LitOptions();
+}
+
+class LitMaterialOptionsBuilder {
```
mvaligursky, line 20: I'd vote for a separate file for this class

mvaligursky (approved):
Added some minor comments, but this looks great!

src/index.js
slimbuck: Yep shed. Probably ok to export?

mvaligursky: yeah I don't see why not. Perhaps add a comment to that file that its used by Shed, so that we don't just modify it considering its private.

src/scene/materials/lit-material.js
slimbuck: Yes agreed, for options and builder. I was going to do that in the followup PR.

examples/src/examples/graphics/lit-material.tsx
slimbuck: Yes the reflections are incorrect, not sure why:

lit-material-example

slimbuck: Perhaps wgpu is generating mipmaps when it shouldn't?

src/scene/materials/lit-material.js
slimbuck: Done

examples/src/examples/graphics/lit-material.tsx
slimbuck: Setting `mipmaps: false` solves this issue so I've re-enabled wgpu on the example.

It does bring up an interesting issue with the difference between webgl and webgpu initialisation. There will likely be more bugs like this as we transition between GL->GPU.

mvaligursky: where did you need to change the mipmaps?

slimbuck: On the asset load options object. All the `helipad-env-atlas` assets were flagged with `mipmaps: false` except for two examples (including this one). I've updated both to specify the flag now.

BTW what I'm referring to with bugs is that the scene sets texture.mipmap=false when used as env-atlas. This works fine with WebGL because the texture is created later on, but under WebGPU this doesn't work.

mvaligursky: Yep ok, all good, glad it's just this, I log out a debug warning when these settings that are not supported to be changed later (currently anyways) are changed.

examples/src/examples/graphics/lit-material.tsx
```diff
@@ -98,27 +99,20 @@ class CustomMaterialExample {
...
-                options.useSpecularColor = true;
-                options.useSpecularityFactor = true;
-                material.litOptions = options;
+                material.shadingModel = pc.SPECULAR_BLINN;
+                material.useSkybox = true;
+                material.hasSpecular = true;
+                material.hasSpecularityFactor = true;
+                material.hasNormals = true;
+                material.hasMetalness = true;
+                material.occludeSpecular = 1;
```
willeastcott, line 115: Should you maybe use the constant here? i.e. `SPECOCC_AO`

slimbuck: Yes! Done! Thanks!

src/scene/shader-lib/programs/lit.js
```diff
@@ -1,58 +1,55 @@
...
 
-    buildPropertiesList(options) {
-        const props = [];
-        for (const prop in options) {
-            if (options.hasOwnProperty(prop) && prop !== "chunks" && prop !== "lights")
-                props.push(prop);
-        }
-        return props.sort();
-    },
+    /** @type { Function } */
```
willeastcott, line 13:
```suggestion
    /** @type {Function} */
```

PR #5490, opened 2023-07-19
Optimization of the render pipeline cache access on WebGPU

mvaligursky:
The initial render pipeline cache was a string concatenation based and its performance was not ideal due to longer strings getting constructed / compared for each draw call.

This PR switches the unique ID from string to an array of numbers, which gets hashed to a single number for the cache lookup. To handle the hash collisions (which are pretty much non existing when testing with less than 100k unique arrays), the array compare is used as well.

Performance improvement in Hierarchy engine examples with over 5k draw calls:

tweet

### DIFF
--- a/src/core/hash.js
+++ b/src/core/hash.js
@@ -15,4 +15,22 @@ function hashCode(str) {
     return hash;
 }
 
-export { hashCode };
+/**
+ * Calculates simple hash value of an array of numbers. Designed for performance, but provides good
+ * distribution with small number of collisions
+ *.
+ * @param {Uint32Array} typedArray - Array of numbers to hash.
+ * @returns {number} Hash value.
+ */
+function fnv1aHashUint32Array(typedArray) {
+    const prime = 16777619;
+    let hash = 2166136261;
+
+    for (let i = 0; i < typedArray.length; i++) {
+        hash ^= typedArray[i];
+        hash *= prime;
+    }
+    return hash >>> 0; // Ensure non-negative integer
+}
+
+export { hashCode, fnv1aHashUint32Array };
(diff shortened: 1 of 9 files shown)


src/platform/graphics/webgpu/webgpu-render-pipeline.js
```diff
@@ -72,10 +74,30 @@ const _stencilOps = [
 // temp array to avoid allocation
 const _bindGroupLayouts = [];
 
+/** @ignore */
+class CacheEntry {
+    /**
+     * Render pipeline
+     *
+     * @type {GPURenderPipeline}
+     * @private
```
willeastcott, line 83: This doesn't look private to me. It's referenced from `WebgpuRenderPipeline`.

src/core/hash.js
```diff
@@ -15,4 +15,22 @@ function hashCode(str) {
     return hash;
 }
 
-export { hashCode };
+/**
+ * Calculates simple hash value of an array of numbers. Designed for performance, but provides good
+ * distribution with small number of collisions
```
willeastcott, line 20:
```suggestion
 * distribution with small number of collisions.
```

willeastcott (approved):

willeastcott:
This is the kind of performance boost I love. ❤️

src/platform/graphics/webgpu/webgpu-render-pipeline.js
mvaligursky: it's private because it exposes GPURenderPipeline type and test:types fails.
I can either remove the type info, or make it private.

willeastcott: Ah, makes sense... :)

slimbuck (approved):

PR #5491, opened 2023-07-20
Optimized the key generation for StencilParameters

mvaligursky:
The only render state which didn't cache the key generation, causing this to be evaluated each frame (for the DEFAULT).

### DIFF
--- a/src/platform/graphics/stencil-parameters.js
+++ b/src/platform/graphics/stencil-parameters.js
@@ -169,8 +169,8 @@ class StencilParameters {
         this._zfail = options.zfail ?? STENCILOP_KEEP;
         this._zpass = options.zpass ?? STENCILOP_KEEP;
 
-        // evaluate key here. This evaluates the key for the DEFAULT instance, which is important,
-        // as during rendering it gets copied and they key would get evaluated each time.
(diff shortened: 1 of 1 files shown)


src/platform/graphics/stencil-parameters.js
```diff
@@ -47,61 +88,104 @@ class StencilParameters {
...
+        this._ref = options.ref ?? 0;
+        this._readMask = options.readMask ?? 0xFF;
+        this._writeMask = options.writeMask ?? 0xFF;
+
+        this._fail = options.fail ?? STENCILOP_KEEP; // keep == 0
+        this._zfail = options.zfail ?? STENCILOP_KEEP;
+        this._zpass = options.zpass ?? STENCILOP_KEEP;
+
+        // evaluate key here. This evaluates the key for the DEFAULT instance, which is important,
+        // as during rendering it gets copied and they key would get evaluated each time.
```
willeastcott, line 173:
```suggestion
        // Evaluate key here. This evaluates the key for the DEFAULT instance, which is important,
        // as during rendering it gets copied and the key would get evaluated each time.
```

willeastcott (approved):

PR #5493, opened 2023-07-21
Remove string.toBool function

mvaligursky:
it’s not used by the engine nor the editor


willeastcott (approved):
Yeah, what the hell is this in the engine for anyway. 😆

PR #5494, opened 2023-07-21
Provide `litArgs` backwards compatibility

kungfooman:
Recently I've been encountering some shader errors in different projects and I wish to have reassurance that previous projects will continue to function properly (I prefer not to rely on luck while performing a `git pull`).

My idea is a simple function which fixes the shader source, but maybe there is a better solution? Any thoughts? 

Example for F12/DevTools:

```js
const tests = [
    'x + litShaderArgs.albedo * y',
    'x + litShaderArgs.sheen.specularity * y',
];
const toLitArgs = src => src.replace(/litShaderArgs([\.a-zA-Z]+)/g, (a, b) => {
    return 'litArgs' + b.replace(/\./g, '_');
});
tests.map(toLitArgs);
```


src/scene/shader-lib/chunk-builder.js
```diff
@@ -1,24 +1,46 @@
...
-    code = '';
+    _code = '';
+
+    set code(newCode) {
+        this._code = newCode;
+    }
+
+    get code() {
+        if (this._code.includes('litShaderArgs')) {
+            this._code = toLitArgs(this._code);
```
mvaligursky, line 23: Personally I'm keen to add it. We discussed how to auto upgrade shaders in the past, but in most cases, the solution would not be simple.
But in this case, I don't see why not.

I'd add `Debug.deprecated` there to inform the user to update the chunks, instead of just silently fixing this forever.

any thoughts @slimbuck ?

kungfooman:
```js
/*
 * @see https://github.com/playcanvas/engine/pull/5463
 */
const toLitArgs = src => src.replace(/litShaderArgs([\.a-zA-Z]+)+/g, (a, b) => {
    const newSource = 'litArgs' + b.replace(/\./g, '_');
    console.warn("Nested struct property access is deprecated, because it's crashing some devices. Please update your custom chunks manually. In particular", a, "should be", newSource);
    return newSource;
});

toLitArgs('litShaderArgs.sheen.specularity'); // Result: 'litArgs_sheen_specularity'
```

Maybe overkill? But I guess as informative as it gets:

image

Other suggestions?

willeastcott: Only thing I would raise is performance. How expensive is it to run these kinds of regexes, particularly on a large fragment shader?

kungfooman:
I added a deprecation warning now, it looks like this in a real project:

image

src/scene/shader-lib/chunk-builder.js
mvaligursky: I was thinking about this, and would suggest a change. I'd move all this backwards compatibility for this class to deprecated.js - perhaps to a function `ChunkBuilder.handleCompatibility`, and call it from here if it exists. That would allow us to have it excluded for module / tree-shaking builds, where you manually update to a new minor version, and so have a control over when you fix the chunks. And we would not have this function as part of the main engine in that case too.

kungfooman: @mvaligursky Thank you for the feedback, I wasn't too happy about using `ChunkBuilder` either, it was just the first solution to be able to continue with other problems and I realized that the `ChunkBuilder#code` was accessed way more often than I thought (causing a bunch of needless `String#includes` checks). Now I just patch `LitShader#+generateFragmentShader` in case `deprecated.js` is included. What do you think about this?

slimbuck:
Just want to say thanks to @kungfooman for this PR and thought I'd add a quick note on my thoughts, because chunks in the engine are so important.

While I understand the desire for this functionality (and very much sympathise with everyone who's projects break due to shader chunk changes), I don't think this is a good approach to maintaining backwards compatibility.

Reasons are because this approach:
- is brittle (especially if we keep adding new fixup code)
- will only ever work for some chunk changes/situations
- means we are either just kicking the problem down the road (until we remove this code from the engine) or this patch and the grep work is done for everyone, forevermore
- if we continue to add to fixup code, over time it will become more difficult to maintain and reason about
- etc, I could go on.

Ideally we would have tools for detecting and updating shader chunks, but in the editor. Then we have no runtime performance cost and are free to implement any snazzy functionality we want. Then we could go crazy and store chunk metadata with engine versions etc etc.

Of course we don't have this functionality in the editor right now. So until then, we probably have to accept this approach.

I just hope we can keep this a one-off and actually remove it as soon as possible.

kungfooman:
Thank you for adding your thoughts @slimbuck!

> I just hope we can keep this a one-off and actually remove it as soon as possible.

Personally I merge `main` rather often, I just need a certain time frame where it is safe to test this and that PR without breaks that make PR testing harder than it really is (manual fixing checkout's, which is removing time/motivation from dealing with the actual issues).

Which time frame would you be happy with? A year would be enough for me and I enjoy to have the deprecation warnings in the meanwhile, which support the upgrading process to make it easy to switch to new engine versions.

kungfooman:
Another related issue which took like 90 minutes to figure out (probably longer if @yaustar didn't help) because it simply fails silently without any deprecation warning:

We have to change:
```js
vec3 getEmission() {
  return vec3(...);
}
```

to:

```js
void getEmission() {
  dEmission = vec3(...);
}
```

Wouldn't it be nice to print a deprecation warning when people still submit shader chunks containing `vec3 getEmission`?

People trip over these things on Discord and they have no clue what is going on:

> I saw a tutorial on how to do it from 2017 and it didn't work for me. Maybe there are newer examples or the engine has changed too much since then?

It doesn't feel good when old projects just slowly decay into non-working states even tho warnings would be easy/obvious to integrate at runtime. Of course these deprecation messages require "extra client work", but what about just putting them in:

```js
        // #if _DEBUG
        ...
        // #endif
```

?

As a developer I use of course the bigger sized debug build, and the "final" project will see and execute absolutely nothing of these backwards compatibility checks and fixes.

slimbuck:
We already have a system in place for this.

The emissive chunk was changed in https://developer.playcanvas.com/en/user-manual/graphics/shader-chunk-migrations/#engine-v157

Your runtime should specify which API version its chunks conform to. We check this at runtime and issue warnings when the API has been updated since then.

For example, if you had specified in your material `mat.chunks.APIVersion='1.62'` (or left APIVersion blank), then you would have received a warning that emissivePS has been updated.

For interest sake, why _didn't_ you see this warning?

src/deprecated/deprecated.js
```diff
@@ -514,6 +515,42 @@ Object.keys(deprecatedChunks).forEach((chunkName) => {
...
+ *
+ * @param {string} src - The fragment shader source which may generate shader errors.
+ * @returns {string} The backwards compatible fragment shader source.
+ */
+function compatibilityForFragmentShader(src) {
+    src = compatibilityForLitArgs(src);
+    return src;
+}
+
+const original_LitShader_generateFragmentShader = LitShader.prototype.generateFragmentShader;
```
mvaligursky, line 548: To be honest, I'm not super keen on the engine monkey patching its internal functions, as its harder to find this is taking place. I'd much preferred a cleaner approach such as this (using just some random naming)

in deprecated.js
```
LitShader.prototype.fixup = function () {
    // do the validation / fixing here
}
```

in lit-shader.js
```
generateFragmentShader() {

    // at some point
    if (this.fixup) {
        code = this.fixup(code);
    }
}
```

src/deprecated/deprecated.js
```diff
@@ -514,6 +515,42 @@ Object.keys(deprecatedChunks).forEach((chunkName) => {
     });
 });
 
+/**
+ * Helper function to ensure a bit of backwards compatibility.
+ *
+ * @example
+ * toLitArgs('litShaderArgs.sheen.specularity'); // Result: 'litArgs_sheen_specularity'
+ * @param {string} src - The shader source which may generate shader errors.
+ * @returns {string} The backwards compatible shader source.
```
mvaligursky, line 524: this needs to have `@ignore` on it as we don't want this to be a a public API

kungfooman:
> For interest sake, why _didn't_ you see this warning?

I *did* see the warning, but it's rather useless. It doesn't tell you anything besides "yea, update your stuff, you know, right".

> Your runtime should specify which API version its chunks conform to. We check this at runtime and issue warnings when the API has been updated since then.

If every "should" in reality would be actualized, we should have less problems and yet in reality we have older projects like @willeastcott's project: https://playcanvas.com/project/1116002/overview/textmesh-sandbox-2

Fork it, launch it, you see this:

image

Show that message to 100 devs, I bet nearly no one knows what's going on or waste too much time on it. After like 90 minutes it dawned on me why it *actually* didn't work by searching through the engine code, but all I really needed would have been *one* message:

`Signature 'vec3 getEmission()' is deprecated, please update your signature to 'void getEmission()' and write the return value into 'dEmission' instead.`

If you show that message to 100 devs on Discord, how many will understand it right away?

Such message would be a simple `String#includes` and has the potenial to make the shader system way easier too use. In terms of runtime overhead and size it's *nothing* inside a `_DEBUG`. It would be stripped entirely out of the production builds.

src/deprecated/deprecated.js
kungfooman: Good idea, I wasn't too happy about the long comment explaining the situation over the method either. First I didn't understand what you meant by `ChunkBuilder.handleCompatibility`, but now I got it. I will rewrite it tomorrow.

mvaligursky: Note that the inside of the `generateFragmentShader` could also be inside Debug.call() to only be included in the debug build.

slimbuck:
Hey @kungfooman I can just imagine your frustration dealing with this. I'm really sorry you've wasted time tracking down the issues.

I agree we could easily improve that message - probably with link to URL as @mvaligursky suggests.

But yeah sorry, I don't agree that adding checks and greps at runtime for every shader chunk change in future is a good idea (if this is what you're advocating for). It is completely unmanageable longer term and doesn't belong in the engine (but does belong in the editor!)

src/deprecated/deprecated.js
mvaligursky: Hi .. if you have some time to do this very soon it'd be awesome, otherwise I'm happy to take over if you're too busy .. it's the last PR we're like to merge before starting the engine release. Please let me know. @kungfooman

kungfooman: Sorry, I just read it, I will fix it right away

mvaligursky (approved):

kungfooman:
> But yeah sorry, I don't agree that adding checks and greps at runtime for every shader chunk change in future is a good idea (if this is what you're advocating for). It is completely unmanageable longer term and doesn't belong in the engine (but does belong in the editor!)

Sure, I also don't want bloated production builds. I see the point of an editor feature and it would be nice if the editor would help aswell in shader programming. In my particular case, I wouldn't have benefited from the editor feature because I found it as a standalone project on GitHub.

Current situation for backwards compatibility of shaders, using importmaps:

```js
/**
 * @param {string} content 
 * @returns {string}
 */
function importFile(content) {
  return "data:text/javascript;base64," + btoa(content);
}
const imports = {
  // Backwards compatibility of shaders
  //"playcanvas": "/playcanvas-engine-pr/src/index.js",
  //"playcanvas": "/playcanvas-engine-pr/build/playcanvas.dbg.mjs/index.js",

  // Lacking backwards compatibility of shaders, use/enable when everything is fixed!
  //"playcanvas": "/playcanvas-engine-pr/build/playcanvas.mjs/index.js",
  "playcanvas": "/playcanvas-engine-pr/build/playcanvas.prf.mjs/index.js",
  //"playcanvas": "/playcanvas-engine-pr/build/playcanvas.min.mjs/index.js",
};
const importmap = document.createElement("script");
importmap.type = "importmap";
importmap.textContent = JSON.stringify({imports});
document.body.appendChild(importmap);

```

src/scene/shader-lib/programs/lit-shader.js
```diff
@@ -125,6 +125,13 @@ class LitShader {
 
         // generated by fshader
         this.fshader = null;
+
+        /**
+         * @type {Function|undefined} - This function may update `this.fshader` in case it's
+         * set in src/deprecated/deprecated.js
+         * @ignore
+         */
+        this.handleCompatibility;
```
mvaligursky, line 134:
```suggestion
        this.handleCompatibility = null;
```

src/scene/shader-lib/programs/lit-shader.js
```diff
@@ -125,6 +125,13 @@ class LitShader {
 
         // generated by fshader
         this.fshader = null;
+
+        /**
+         * @type {Function|undefined} - This function may update `this.fshader` in case it's
```
mvaligursky, line 130:
```suggestion
         * @type {Function|null} - This function may update `this.fshader` in case it's
```

src/scene/shader-lib/programs/lit-shader.js
kungfooman: Sorry, this is wrong, if you assign it, it will overwrite the actual compatibility function from `deprecated.js`. The ESLint rule here is/was deceptive.

mvaligursky: ah, so what's the better solution?

mvaligursky: maybe just to remove this line completely?

PR #5372, opened 2023-06-02
Improvements to Texture implementation of cubemaps for WebGPU

mvaligursky:
- moved few static functions from Texture class to TextureUtils class, allowing those to be called from WebgpuTexture without circular refs.
- implemented support for loading data in TypeArrays into cubemap faces for WebGPU
- implemented support for mipmap generation of cubemaps for WebGPU

PR #5497, opened 2023-07-23
Fix WebGL depth texture support

liamdon:
This PR fixes depth texture support on WebGL.

I am using `texture.lock()` to generate custom 3D textures, and noticed that this stopped working because the texture size passed to `texImage3D` is calculated incorrectly. I believe this may have changed in #5372.
 
I have tested this change on a WebGL-only project that uses a variety of textures including 2D, 3D, and cubemap, with and without mipmaps.

Depth textures are not currently not supported in WebGPU, so we can just pass `depth=1` for all WebGPU textures.

---

### DIFF
--- a/src/platform/graphics/texture-utils.js
+++ b/src/platform/graphics/texture-utils.js
@@ -26,30 +26,32 @@ class TextureUtils {
      *
      * @param {number} width - Texture's width.
      * @param {number} height - Texture's height.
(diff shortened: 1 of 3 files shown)


willeastcott (approved):
LGTM, but I'll leave it to @mvaligursky to have the final word! Thanks for this, @liamdon!!

PR #5501, opened 2023-07-26
EventHandler: refactor, comment and optimize

Maksims:
To follow up on: https://github.com/playcanvas/engine/pull/5484
Slightly improved EventHandler, by moving from `Object` to `Map` for callbacks.
Also added comments to "Active Callback" optimizations to explain it better.
And even found a small optimization which reduced occurrences of unoptimized path when cloning callbacks list. Previously it cloned callbacks list if it there was any active callbacks in place. But now it only does it if previously active callbacks list were modified during its execution.
Behavior is preserved.

### DIFF
--- a/src/core/event-handler.js
+++ b/src/core/event-handler.js
@@ -42,8 +42,7 @@ class EventHandler {
 
     /**
      * Reinitialize the event handler.
-     *
-     * @private
+     * @ignore
      */
     initEventHandler() {
         this._callbacks = new Map();
@@ -59,7 +58,7 @@ class EventHandler {
      * @param {object} scope - Object to use as 'this' when the event is fired, defaults to
      * current this.
      * @param {boolean} once - If true, the callback will be unbound after being fired once.
-     * @private
+     * @ignore
      */
     _addCallback(name, callback, scope, once) {
         if (!name || typeof name !== 'string' || !callback)
@@ -73,7 +72,7 @@ class EventHandler {
         if (this._callbackActive.has(name)) {
             const callbackActive = this._callbackActive.get(name);
             if (callbackActive && callbackActive === this._callbacks.get(name)) {
-                this._callbackActive.set(name, callbackActive.slice());
+                this._callbackActive.set(name, Array.from(callbackActive));
             }
         }
 
--- a/src/core/event-handler.js
+++ b/src/core/event-handler.js
@@ -56,9 +56,9 @@ class EventHandler {
      * @param {string} name - Name of the event to bind the callback to.
      * @param {HandleEventCallback} callback - Function that is called when event is fired. Note
      * the callback is limited to 8 arguments.
-     * @param {object} [scope] - Object to use as 'this' when the event is fired, defaults to
+     * @param {object} scope - Object to use as 'this' when the event is fired, defaults to
      * current this.
-     * @param {boolean} [once] - If true, the callback will be unbound after being fired once.
+     * @param {boolean} once - If true, the callback will be unbound after being fired once.
      * @private
      */
     _addCallback(name, callback, scope, once) {
@@ -276,7 +276,7 @@ class EventHandler {
      * obj.hasEvent('hello'); // returns false
      */
     hasEvent(name) {
-        return (this._callbacks.has(name) && this._callbacks.get(name).length !== 0) || false;
+        return !!this._callbacks.get(name)?.length;
     }
 }
 
--- a/src/core/event-handler.js
+++ b/src/core/event-handler.js
@@ -29,16 +29,16 @@
  */
 class EventHandler {
     /**
-     * @type {object}
+     * @type {Map<string,Array<object>>}
      * @private
      */
-    _callbacks = {};
+    _callbacks = new Map();
 
     /**
-     * @type {object}
+     * @type {Map<string,Array<object>>}
      * @private
      */
-    _callbackActive = {};
+    _callbackActive = new Map();
 
     /**
      * Reinitialize the event handler.
@@ -46,8 +46,8 @@ class EventHandler {
      * @private
      */
     initEventHandler() {
-        this._callbacks = {};
-        this._callbackActive = {};
+        this._callbacks = new Map();
+        this._callbackActive = new Map();
     }
 
     /**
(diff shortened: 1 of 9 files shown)


src/core/event-handler.js
```diff
@@ -178,67 +218,55 @@ class EventHandler {
...
         }
 
-        // TODO: What does callbacks do here?
-        // In particular this condition check looks wrong: (i < (callbacks || this._callbackActive[name]).length)
-        // Because callbacks is not an integer
+        // @ts-ignore: Array is possibly 'undefined'
         // eslint-disable-next-line no-unmodified-loop-condition
-        for (let i = 0; (callbacks || this._callbackActive[name]) && (i < (callbacks || this._callbackActive[name]).length); i++) {
-            const evt = (callbacks || this._callbackActive[name])[i];
+        for (let i = 0; (callbacks || this._callbackActive.get(name)) && (i < (callbacks || this._callbackActive.get(name)).length); i++) {
```
willeastcott, line 242: `this._callbackActive.get(name)` seems to be called in various places. Can that be cached?

src/core/event-handler.js
```diff
@@ -178,67 +218,55 @@ class EventHandler {
...
+            // if we are trying to execute a callback while there is an active execution right now
+            // and the active list has been already modified,
+            // then we go to an unoptimized path and clone callbacks list to ensure execution consistency
+            callbacks = Array.from(callbacksInitial);
         }
 
-        // TODO: What does callbacks do here?
-        // In particular this condition check looks wrong: (i < (callbacks || this._callbackActive[name]).length)
-        // Because callbacks is not an integer
+        // @ts-ignore: Array is possibly 'undefined'
```
willeastcott, line 240: We don't use `@ts-ignore` anywhere across the codebase. I know it's nice to suppress the errors VS Code shows, but for now, we haven't taken a decision to actively use `@ts-ignore`. If we did, there'd be probably many hundreds of them all over the source.

src/core/event-handler.js
Maksims: Value by that key can be modified during this loop execution (when adding/removing callbacks during callback execution).
So it has to get actual value every time.

src/core/event-handler.js
```diff
@@ -58,22 +58,28 @@ class EventHandler {
      * the callback is limited to 8 arguments.
      * @param {object} [scope] - Object to use as 'this' when the event is fired, defaults to
      * current this.
-     * @param {boolean} [once=false] - If true, the callback will be unbound after being fired once.
+     * @param {boolean} [once] - If true, the callback will be unbound after being fired once.
```
willeastcott, line 61: `scope` and `once` should not be marked as optional (square brackets) here because both `on` and `once` do supply these parameters in all circumstances.

src/core/event-handler.js
```diff
@@ -253,7 +276,7 @@ class EventHandler {
      * obj.hasEvent('hello'); // returns false
      */
     hasEvent(name) {
-        return (this._callbacks[name] && this._callbacks[name].length !== 0) || false;
+        return (this._callbacks.has(name) && this._callbacks.get(name).length !== 0) || false;
```
kungfooman, line 279:
```suggestion
        return !!_callbacks.get(nameArg)?.length;
```

Shorter, a method call less, no type error, easier to read?

[playground](https://www.typescriptlang.org/play?strictNullChecks=true&noUnusedLocals=true&exactOptionalPropertyTypes=true&ts=4.6.2&filetype=js#code/PQKhAIAEBcE8AcCm4DeBZAhvAPAZ2gE4CWAdgOYA04GJsA2gLoB8AvuCMAFADGA9ifnAB9bhgA2YgEYZuAa1zgAvOBKIA7uEzwAFAEoA3JxHipM+QDpciaNoBEsRLltU6ARioAmKgGYGBo6IS0nK4ltZ2JLzO4Iz+fALQKhgAtogAggRkSuD2jraGwMDxgiQp6ZnZtpH5nACEtcZBZqFk4aWpGWS6APzmYojk0AAWhgEmwRatNu3lXX0DZMOjQA)

willeastcott: `nameArg`?

kungfooman: > @willeastcott Sorry, that stemmed from my playground test, because for some reason a variable called "name" triggers errors there.

kungfooman:
Why do we use a mixture of `Array.from` and `.slice()` now? Can't we just go for the shorter version, aka keeping `.slice()`?

kungfooman:
Since we are already refactoring...

```
    /**
     * Reinitialize the event handler.
     *
     * @private
     */
    initEventHandler() {
        this._callbacks = new Map();
        this._callbackActive = new Map();
    }
```

This method is *not* private, it has to be called throughout the rest of the engine. This will also fix additional type errors:

image

You also didn't fix this:

https://github.com/playcanvas/engine/blob/6cc5ebce1984a6237b0e556c58b720cb1ff63820/src/core/events.js#L21-L24

It has to be:

```js
        // ...
        target.hasEvent = ev.hasEvent;
        EventHandler.prototype.initEventHandler.apply(target);
        return target;
```

`@private` means that you can refactor the code without a 2nd thought, but this is the 2nd PR with the same mistake, so we shouldn't "over-privatize" what should be public and of course refrain from using private stuff in the rest of the engine, because it's causing refactor bugs that may be hard to find later on.

I guess we can excuse the unit tests for being tests that check internal states, but not normal use-cases from within the engine itself.

willeastcott:
Hmm, presumably `initEventHandler` should be `@ignore` instead.

Maksims:
Updated PR based on recent comments. Used @ignore (Thanks @willeastcott), and used consistently `Array.from` - as it is a modern way to clone an array.

kungfooman:
> and used consistently `Array.from` - as it is a modern way to clone an array.

I'm not really convinced, if it's only about modern syntax, you could have picked the shorter `[...arr]` version aswell, which seems to be slightly faster aswell.

I guess the main issue is performance, both `[...arr]` and `Array.from(...)` are using the iterator protocol, which makes them rather slow:

https://jsben.ch/IjJQG

image

Because they have to deal with cases like:

```js
class Vec3 {
    x = 1;
    y = 2;
    z = 3;
    *[ Symbol.iterator ]() {
		yield this.x;
		yield this.y;
		yield this.z;
	}
}
vec3 = new Vec3;
Array.from(vec3); // Result: [1, 2, 3]
```

Whereas `Array.prototype.slice` knows exactly what to do - because it's the very own designated method for doing this job.

src/framework/components/script/component.js
```diff
@@ -391,7 +391,7 @@ class ScriptComponent extends Component {
             // disable script if it fails to call method
             script.enabled = false;
 
-            if (!script._callbacks || !script._callbacks.error) {
+            if (!script?.hasEvent('error')) {
```
willeastcott, line 394: Do you need the `?` here? This line above seems to assume `script` is valid.

kungfooman: Good catch, I agree, it seems like an old edge case where it was possible that `EventHandler` deleted `_callbacks` :thinking: Or maybe it was always too defensive.

willeastcott (approved):
Awesome. LGTM! @kungfooman, are you happy for me to merge now?

kungfooman:
@willeastcott Yep, PR is looking very good, great job @Maksims :1st_place_medal:

Issue #3334, opened 2021-07-14
Static lights not working after the scene reload


mvaligursky:
issue and a repro project: https://forum.playcanvas.com/t/issue-with-static-lights/19233/25

mvaligursky:
This likely won't get fixed as we're aiming at deprecating static lights, in favour of clustered lighting.

Issue #3229, opened 2021-06-10
Static models with an assigned batch group disappear when reloading the scene


davidpox:
### Description

Static models in a scene that have a Batch Group assigned disappear when the same scene is destroyed and reloaded, regardless of whether or not the batcher has ran. 

[Project Link](https://playcanvas.com/project/803952/overview/batch-minimal)
[Forum Post Link](https://forum.playcanvas.com/t/batched-models-disappear-after-reloading-scene/20637)

Upon first load, you'll see a couple of plane entities laid out. The outer circle of planes have no batch group assigned, whereas anything inside the circle have a single batch group assigned with its layer set to World. Once you reload the scene by pressing 'L', you'll notice that all the planes that had a batch group assigned disappear from the scene, and the ones without a batch group remain in place as expected. 

Tested on Windows 10 (2004 - 19041.985) using Google Chrome v91.0.4472.77, and Firefox v89.0, PlayCanvas v1.42.0

### Steps to Reproduce

1. Create a few entities around the scene and set them to be Static
2. Create a new Batch Group and give it the World layer
3. Assign the new batch group to the entities
4. Load the game, see that all entities load fine
5. Use `SceneRegistry.loadSceneHierarchy` to reload the scene and destroy the old hierarchy
6. Observe that the entities with a batch group have disappeared.

### Screenshots

Screenshot 1 - Before reloading the scene
![Iy78cxYn](https://user-images.githubusercontent.com/7888472/121571020-6a851880-ca1a-11eb-820c-dea962fdfe4a.png)

Screenshot 2 - After reloading the scene
![EIodJzAK](https://user-images.githubusercontent.com/7888472/121571051-71ac2680-ca1a-11eb-8397-7f468a32e13f.png)

mvaligursky:
I tried looking into this, but have not found an issue .. it would need more time and deeper investigation, perhaps another time.

As a workaround (not the same behaviour, I understand) would be to remove the previous scene before new one gets loaded.

```
SceneSwitchButton.prototype.loadScene = function() {
    var oldHierarchy = this.app.root.findByName('MyScene');
    var scene = this.app.scenes.find("GameScene");
    oldHierarchy.destroy();
   
    this.app.scenes.loadSceneHierarchy(scene.url, function (err, parent) {
        if (!err) {
        } else {
            console.error(err);
        }
    });  
};

```

yaustar:
[...]

Issue #3052, opened 2021-03-31
Static enabled spot lights do not light up the scene as expected


yaustar:
Scene: https://playcanvas.com/editor/scene/1122796
Forum thread: https://forum.playcanvas.com/t/issue-with-static-lights/19233/

Spotlight with static disabled
![image](https://user-images.githubusercontent.com/16639049/113150602-28767400-922c-11eb-8cd6-df465eb83ae0.png)

Spotlight with static enabled
![image](https://user-images.githubusercontent.com/16639049/113150685-3d530780-922c-11eb-84f0-ed42b5bdfb82.png)

mvaligursky:
maybe related to https://github.com/playcanvas/engine/issues/2079

Issue #2079, opened 2020-05-14
Static lights are not rendered


thisredone:
This was quite a head-scratcher for me recently.
Marking lights as Static makes them not have an effect on the scene at all.

slimbuck:
Hi @thisredone, could you please create a publicly accessible scene so we can investigate?

Thanks!

thisredone:
There you go: https://playcanvas.com/editor/scene/921825

mvaligursky:
I'm investigating this .. it seems static light only affects static geometry. Should not be the case.

mvaligursky:
The further investigation confirms this. Static lights are only used for static geometry, and ignored for non-static geometry. Changing this behavior will be considered in the future due to complexity of the change.

For further reference:
- we skip static lights here when rendering any geometry, and those are added from staticLightList which is only used for static geometry.
https://github.com/playcanvas/engine/blob/ca4d739737c500bf079e62c05436726f3cbf5be4/src/scene/forward-renderer.js#L993

- also static lights are skipped here when building material / shader
https://github.com/playcanvas/engine/blob/ca4d739737c500bf079e62c05436726f3cbf5be4/src/scene/materials/standard-material-options-builder.js#L263-L264

Maksims:
> it seems static light only affects static geometry.

As I remember, that was the actual design of **static** feature.
When used static, it will affect only static geometry enabling ability for such geometry to be split into separate mesh instances, based on static light bounding boxes. So there is no need to have one material with all lights on it, but will have multiple materials(shaders), with only affecting lights in shader code. Which optimises compilation time, but increases draw calls and number of material switches.

This demo used it: https://playcanv.as/p/yipplmVO/

yaustar:
Another bump from the forum: https://forum.playcanvas.com/t/issue-with-static-lights/19233/21?u=yaustar

PR #5502, opened 2023-07-26
[Breaking] Removed support for static lights

mvaligursky:
This PR removes support for static lights, where meshes would have assigned static lights that affect them, and meshes affected by too many lights would be split into smaller meshes.

The static light setting is now ignored, and all lights are treated as dynamic. A clustered lighting handles this without problem. This is marked as Breaking, as the users might need to enable clustered lighting (which is enabled by default)

leftover bits: 
- MeshInstance.getShaderVariant still accepts static lights list, but does not internally use it. This is to preserve compatibility with the Editor that calls the function directly in few places.
- Light/Model/Render component keep this flag for compatibility reasons, but it does not get used, and will be deprecated in the future, unless some use for those is found.

These tickets are no longer relevant:
- Fixes https://github.com/playcanvas/engine/issues/2079
- Fixes https://github.com/playcanvas/engine/issues/3052
- Fixes https://github.com/playcanvas/engine/issues/3229
- Fixes https://github.com/playcanvas/engine/issues/3334


willeastcott (approved):
LGTM! 👏

PR #5506, opened 2023-07-26
Layer.cullingMask and Camera.cullingMask has been removed as not needed

mvaligursky:
Obsolete functionality no longer used.

### DIFF
--- a/src/scene/camera.js
+++ b/src/scene/camera.js
@@ -40,7 +40,6 @@ class Camera {
         this._clearDepthBuffer = true;
         this._clearStencil = 0;
         this._clearStencilBuffer = true;
-        this._cullingMask = 0xFFFFFFFF;
         this._cullFaces = true;
         this._farClip = 1000;
         this._flipFaces = false;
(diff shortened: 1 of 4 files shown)


willeastcott (approved):
Seems safe enough to me. Even if somebody uses this API, they app should still run, as far as I can tell.

LeXXik:
What would be the right way to migrate an old project that used that functionality? For example:
https://playcanvas.com/project/668042/overview/my-toon-water

mvaligursky:
The Layers need to be used instead. Culling mask is basically like a list of Layers - each bit represents a Layer.

MAG-AdrianMeredith:
This was really handy, definitely not obsolete.  

It allowed us to remove certain meshes from a layer just for a draw call.  True we could have created a bunch of different layers but this seems a more elegant way to do it. 
Is this documentation no longer correct? 
"Mask controlling which LightComponents light this mesh instance, which **CameraComponent** sees it and in which Layer it is rendered."

mvaligursky:
It was a left over from the system before the layers were implemented. I agree the Layers are little bit heavy, and I wish under the hood they used bitflags, but they don't.

I think we've yet to release the docs for the engine, which should take place shortly. Sorry about that.

PR #5508, opened 2023-07-26
Light hash generation does not use strings

mvaligursky:
- optimization to avoid allocations
- added basic tests

### DIFF
--- a/src/core/hash.js
+++ b/src/core/hash.js
@@ -16,14 +16,15 @@ function hashCode(str) {
 }
 
 /**
- * Calculates simple hash value of an array of numbers. Designed for performance, but provides good
- * distribution with small number of collisions.
+ * Calculates simple 32bit hash value of an array of 32bit integer numbers. Designed for
+ * performance, but provides good distribution with small number of collisions. Based on
+ * FNV-1a non-cryptographic hash function.
  *.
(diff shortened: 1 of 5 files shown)


src/core/hash.js
```diff
@@ -19,15 +19,16 @@ function hashCode(str) {
  * Calculates simple hash value of an array of numbers. Designed for performance, but provides good
  * distribution with small number of collisions.
  *.
- * @param {Uint32Array} typedArray - Array of numbers to hash.
+ * @param {number[]|Uint32Array} typedArray - Array of numbers to hash.
```
willeastcott, line 22: The parameter name here no longer matches the function signature.

src/core/hash.js
```diff
@@ -19,15 +19,16 @@ function hashCode(str) {
...
  * @returns {number} Hash value.
+ * @ignore
  */
-function fnv1aHashUint32Array(typedArray) {
+function fnv1aHashUint32Array(array) {
```
willeastcott, line 26: If the function can now also take a number array, is it now misnamed?

mvaligursky: I knew you would not miss it .. renaming it.

willeastcott (approved):
LGTM! 😄

PR #5514, opened 2023-07-28
Optimize LayerComposition layer lookup by name and id using Maps

mvaligursky:
- This being O(1) will be used in follow up PRs for fast access.
- The maps get update each time a list changes, instead of being lazy updated, as this is a very low frequency operations and avoid testing the dirty flag on each access later

### DIFF
--- a/src/scene/composition/layer-composition.js
+++ b/src/scene/composition/layer-composition.js
@@ -29,6 +29,63 @@ class LayerComposition extends EventHandler {
     // Composition can hold only 2 sublayers of each layer
 
     /**
+     * A read-only array of {@link Layer} sorted in the order they will be rendered.
+     *
+     * @type {import('../layer.js').Layer[]}
+     */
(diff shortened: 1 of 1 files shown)


src/scene/composition/layer-composition.js
```diff
@@ -984,6 +1014,21 @@ class LayerComposition extends EventHandler {
         return this._getSublayerIndex(layer, true);
     }
 
+    /**
+     * Update maps of layer IDs and names to match the layer list.
+     *
+     * @ignore
```
willeastcott, line 1020: Wouldn't this be `@private`?

willeastcott (approved):
I ❤️ Maps!

src/scene/composition/layer-composition.js
mvaligursky: yep, thanks!

PR #5518, opened 2023-07-28
Update tsconfig.json

epreston:
Add default `outDir`

Removes 700 error messages from vscode warning that files will be overwritten.  The warnings make type / error checking unusable.

This is the same value used in `npm run build:dts`.  All other commands that use `tsc` override `outDir` at the command line. 
Practical type generation is unchanged, future commands will have a safe / documented location to default to.

From the tsc docs:

> Compiler options specified on the command line override those specified in the tsconfig.json file. -- https://www.typescriptlang.org/docs/handbook/tsconfig-json.html

Fixes contributors using vscode.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,7 @@
 {
     "compilerOptions": {
         "baseUrl": ".",
+        "outDir": "./types",
         "allowJs": true,
         "checkJs": true,
         "module": "es6",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,7 @@
 {
     "compilerOptions": {
         "baseUrl": ".",
-        "outDir": "./types",
+        "outDir": "types",
         "allowJs": true,
         "checkJs": true,
         "module": "es6",


willeastcott:
Hmmm interesting. So does this mean we can remove `--outDir types` from the `build:dts` script then?

willeastcott:
I just tried it - it seems not. When `tsc` runs on executing `build:dts`, it seems to ignore the `tsconfig.json` for some reason.

epreston:
Oh yeahj, when you supply source files on the command line, `tsc` ignores the `tsconfig.json`

need to not specify files or use the `-p tsconfig.json`

It also means you haven't been haven't been compiling with the project settings.  Only rollup config in examples uses the `tsconfig.json` ?

willeastcott:
OK. Well, for now, this seems to be an improvement as it is. Should I just go ahead and merge?

epreston:
Worth double checking with the new typedoc bits first.

Related question: Is it expected that the settings / restrictions in `tsconfig.json` be applied when building types ?

willeastcott:
I believe I added the `tsconfig.json` originally to enable Intellisense in VS Code. So no, I don't think it was added to configure type declarations generation. I think this is safe to merge. After all, it's super-easy to back out if we need to! 😄

willeastcott (approved):

Issue #5521, opened 2023-07-30
Issues Building Examples Browser


epreston:
NPM is unable to resolve some dependencies, warns of two deprecated modules, completes with 12 vulnerabilities (7 moderate, 5 high).

### Description

A dependency is no longer publicly available, another is deprecated, minor cleanup required.

`mini-create-react-context` is no longer maintained or available from NPM.  react-router@5.3.4 was released, where dependency warning was removed.  From what I understand, this is not useful in react 18+ and is a common issue.

### Steps to Reproduce

1. in the `examples` directory, run `npm install`

### Proposed Solution

Update dependencies and address issues in a PR.

epreston:
Some build commands use `node --es-module-specifier-resolution=node`.  

` --es-module-specifier-resolution` has been removed in node 19+, adding to remediation since its related.

epreston:
Example LitMaterial not found.  This blocks thumbnail generation using `npm run build:thumbnails`.

relative URL: `?category=Graphics&example=LitMaterial

kungfooman:
I couldn't build it either some weeks ago (I don't remember the actual errors) and I ended up rewriting it in pure MJS - completely circumventing the build script.

Simply `<script type="module"...` with an importmap.

epreston:
first time dependency installation produces the following messages.

```
npm WARN deprecated sourcemap-codec@1.4.8: Please use @jridgewell/sourcemap-codec instead
npm WARN deprecated rollup-plugin-terser@7.0.2: This package has been deprecated and is no longer maintained. Please use @rollup/plugin-terser
npm WARN deprecated source-map-resolve@0.6.0: See https://github.com/lydell/source-map-resolve#deprecated
npm WARN deprecated puppeteer@14.4.1: < 19.4.0 is no longer supported
```

epreston:
@rollup/plugin-terser is a drop in replacement for rollup-plugin-terser@7.0.2

This has already been done for the engine build but not examples build.  `rollup` is using too much memory and spawning a large number of processes with the deprecated plugin.

willeastcott:
Interesting. Yeah, that definitely needs to be switched over. 👍

epreston:
The last two bits ultimately require an update of `rollup`. The base engine package is already updated.

They are:

```
npm WARN deprecated sourcemap-codec@1.4.8: Please use @jridgewell/sourcemap-codec instead
[...]

PR #5522, opened 2023-07-30
fix: example thumbnail build

epreston:
The file and content exists in the generated site content but the class for this example is commented out.  This bocks thumbnails from being generated for the whole project.

Related: #5521

### DIFF
--- a/examples/src/examples/graphics/index.mjs
+++ b/examples/src/examples/graphics/index.mjs
@@ -8,7 +8,7 @@ import ClusteredLightingExample from "./clustered-lighting";
 import ClusteredOmniShadowsExample from "./clustered-omni-shadows";
 import ClusteredSpotShadowsExample from "./clustered-spot-shadows";
 import ContactHardeningShadowsExample from "./contact-hardening-shadows";
-import LitMaterialExample from "./lit-material";
+// import LitMaterialExample from "./lit-material";
 import GrabPassExample from "./grab-pass";
 import GroundFogExample from "./ground-fog";
 import HardwareInstancingExample from "./hardware-instancing";
@@ -69,7 +69,7 @@ export {
     ClusteredOmniShadowsExample,
     ClusteredSpotShadowsExample,
     ContactHardeningShadowsExample,
-    LitMaterialExample,
+    // LitMaterialExample,
     GrabPassExample,
     GroundFogExample,
     HardwareInstancingExample,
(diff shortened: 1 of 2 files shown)


willeastcott:
Hey @mvaligursky and @slimbuck, I know this example was hidden because it's not ready for public comsumption. But if it's breaking the build, we should either merge this or figure out a way to hide it without breaking stuff. I'll leave it to you to two to decide.

epreston:
It's not being hidden when you build examples.  It is there in the list of examples, just no thumbnail. The thumbnail builds break halfway through leaving only half thumbs.

mvaligursky:
For now we need to remove the example (instead of commenting it out). It works, but the API is not finalised / public yet.
We used to have HIDDEN option in the example being supported before, to skip adding those examples to the list before, but that seems gone now, so we just need to remove its import for now.

mvaligursky:
Yep, we should just remove those two lines with LitMaterialExample in them from the index.mjs

epreston:
The thumbnails are built by enumerating files in a directory.

mvaligursky:
Ah I see. Hmm, this feels pretty hidden way to do this, I can see somebody spending time trying to figure out why this example does not have a thumbnail in the future. Perhaps add a log when this name is filtered out, so that we're aware, thanks.

examples/src/examples/graphics/index.mjs
```diff
@@ -69,7 +69,7 @@ export {
     ClusteredOmniShadowsExample,
     ClusteredSpotShadowsExample,
     ContactHardeningShadowsExample,
-    // LitMaterialExample,
+    LitMaterialExample,
```
mvaligursky, line 72: and here please remove this line, and also the relevant import in the same file - the example is not ready to be shipped yet.

epreston: No worries.

epreston:
I'm knee deep in the examples build stuff.  I'd like to add that HIDDEN option back in, would a PR for that be welcome ?

examples/src/examples/graphics/index.mjs
epreston: Remove or comment ?  So you can quickly undo this ?

mvaligursky:
> I'd like to add that HIDDEN option back in, would a PR for that be welcome

It would be very much appreciated to have it back, thanks for looking at this!

mvaligursky (approved):
Perfect, thanks, commenting out is great.

PR #5523, opened 2023-07-30
chore: update examples build

epreston:
Related: #5521

Addresses tool and dependency warnings
 - from: 12 vulnerabilities (7 moderate, 5 high)
 - to: 0

This resolves:

> mini-create-react-context is no longer maintained or available from NPM. react-router@5.3.4 was released, where dependency warning was removed.

>  `--es-module-specifier-resolution` has been removed in node 19+, adding to remediation since its related.


examples/package-lock.json
```diff
@@ -1,84 +1,92 @@
 {
   "name": "examples-browser",
   "version": "0.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
```
willeastcott, line 4: I think this is the first time I've seen `lockfileVersion` set to 3. Does this mean we're bumping the minimum required node version with this update?

willeastcott: Ah, I just saw the root `package.json` has the same version. So I guess it's fine.

epreston: No. Lockfile 3 is backward compatible to version 7.

epreston: Out of curiosity, what are you targeting these days ?  18 ?

CI looks like it's at 16 but other things look like 18.

willeastcott: I don't think we really have an official policy on a min version for Node. 16 sounds reasonable I guess?

willeastcott:
The GitHub Action that builds the Examples Browser seems to fail with this PR for some reason (JS Heap OOM).

epreston:
CI is failing with:

`FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`

I'll run through it and update.

examples/package-lock.json
epreston: Only a few weeks left of support on 16.  18 is on current Long Term Support.  20 is current.  

Highly recommend 18.

Diagram here. 
https://nodejs.dev/en/about/releases/

willeastcott: OK, I'm fine with that.

epreston:
Happy to split this one up if needed.

willeastcott:
I'm easy either way, @epreston. Generally, I like PRs to be narrow in focus (makes backing out straightforward). But at the same time, this PR isn't too enormous.

epreston:
Thank you. I'll leave this for review then.

willeastcott (approved):
LGTM. Thanks! 🙏

PR #5524, opened 2023-07-31
Fire ‘shader:generate’ when a shader code gets generated

mvaligursky:
The event `shader:generate` is fired on the GraphicsDevice when the shader definition is generated, just before the shader is compiled, to allow the user to inspect / modify the generated shader. Note that there is a further processing happening after this step, where defines are process (unused code is removed), a shader is converted to uniform format on some platforms (currently on WebGPU) and similar.

Typical use case:

- give materials a userId to recognise when its shader is getting compiled:
```
myMaterial.userId = 'redMaterial';
```

- listen to shader:generated events:
```
app.graphicsDevice.on('shader:generate', (info) => {
    if (info.userMaterialId)
        console.log('shader:generate', info);
});
```

This is the expected output, in this case the shader is compiled for a shadow pass and a forward pass. You can find relevant info in the `info.shaderPassInfo'. Note that there could be multiple shader compilations based on material settings / vertex format it's used on and similar.

Screenshot 2023-07-31 at 12 26 29

### DIFF
--- a/src/scene/materials/material.js
+++ b/src/scene/materials/material.js
@@ -69,7 +69,9 @@ class Material {
     name = 'Untitled';
 
     /**
-     * A unique id the user can assign to the material.
+     * A unique id the user can assign to the material. The engine internally does not use this for
+     * anything, and the user can assign a value to this id for any purpose they like. Defaults to
+     * an empty string.
      *
      * @type {string}
      */
--- a/src/scene/materials/material.js
+++ b/src/scene/materials/material.js
@@ -68,6 +68,13 @@ class Material {
      */
     name = 'Untitled';
 
+    /**
+     * A unique id the user can assign to the material.
+     *
+     * @type {string}
+     */
+    userId = '';
+
     id = id++;
 
     variants = {};
--- a/src/scene/shader-lib/program-library.js
+++ b/src/scene/shader-lib/program-library.js
@@ -110,7 +110,7 @@ class ProgramLibrary {
         this.processedCache.set(key, shader);
     }
 
-    getProgram(name, options, processingOptions) {
+    getProgram(name, options, processingOptions, userMaterialId) {
         const generator = this._generators[name];
         if (!generator) {
             Debug.warn(`ProgramLibrary#getProgram: No program library functions registered for: ${name}`);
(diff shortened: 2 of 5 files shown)


src/scene/materials/material.js
```diff
@@ -68,6 +68,13 @@ class Material {
      */
     name = 'Untitled';
 
+    /**
+     * A unique id the user can assign to the material.
```
willeastcott, line 72: Maybe add some words on what this is for and why a user might want to set it?

willeastcott (approved):

src/scene/materials/material.js
mvaligursky: updated

src/scene/materials/material.js
```diff
@@ -68,6 +68,15 @@ class Material {
...
     name = 'Untitled';
 
+    /**
+     * A unique id the user can assign to the material. The engine internally does not use this for
+     * anything, and the user can assign a value to this id for any purpose they like. Defaults to
+     * an empty string.
+     *
+     * @type {string}
+     */
+    userId = '';
```
slimbuck, line 78: why not just use `material.name` for this?

mvaligursky: this is assigned to / managed by the engine. Ideally this can be used by a code and work even if the user renames the material for example.

slimbuck: 👍🏻 makes sense

slimbuck (approved):

src/scene/shader-lib/program-library.js
```diff
@@ -133,11 +133,20 @@ class ProgramLibrary {
...
                 passName = `-${shaderPassInfo.name}`;
             }
 
+            // fire an event to allow the shader to be modified by the user. Note that any modifications are applied
+            // to all materials using the same generated shader, as the cache key is not modified.
+            this._device.fire('shader:generate', {
+                userMaterialId,
+                shaderPassInfo,
+                definition: generatedShaderDef
+            });
```
AndrewRayCode, line 148: nit: it would be great if this had a type so consumers could at least typecast the event `as pc.ShaderGenerateEvent` or similar

PR #5536, opened 2023-08-02
Refactor LayerComposition to avoid its update when mesh transparency changes / mesh is added

mvaligursky:


src/scene/layer.js
```diff
@@ -43,51 +42,22 @@ let layerCounter = 0;
...
-        if (!this.visibleTransparent[index]) {
-            this.visibleTransparent[index] = new VisibleInstanceList();
-        }
+const _tempMaterials = new Set();
 
-        // mark them as not processed yet
-        this.visibleOpaque[index].done = false;
-        this.visibleTransparent[index].done = false;
-    }
+class CulledInstances {
```
willeastcott, line 47: Does this class need an `@ignore` tag?

src/scene/layer.js
```diff
@@ -98,6 +68,30 @@ class InstanceList {
...
+     */
+    shadowCasters = [];
+
+    /**
+     * Visible (culled) mesh instances assigned to this layer. Looked up by the Camera.
+     *
+     * @type {WeakMap<import('./camera.js').Camera, CulledInstances>}
+     * @private
+     */
+    _visibleInstances = new WeakMap();
```
willeastcott, line 93: Ooo. Is this the first use of `WeakMap` in the codebase? Exciting. 😄

src/scene/layer.js
mvaligursky: It should not as it does not have a JSDoc block for it.

src/scene/renderer/renderer.js
```diff
@@ -523,15 +534,12 @@ class Renderer {
...
-        const count = drawCalls.length;
-        for (let i = 0; i < count; i++) {
-            const drawCall = drawCalls[i];
-            if (drawCall.visibleThisFrame) {
-                const skin = drawCall.skinInstance;
-                if (skin && skin._dirty) {
-                    skin.updateMatrixPalette(drawCall.node, _skinUpdateIndex);
-                    skin._dirty = false;
-                }
+        for (const drawCall of drawCalls) {
```
willeastcott, line 537: How does `Set` iteration transpile in Babel? Also, it's not obvious that `drawCalls` is a `Set`. I had to jump around the code a bit to find out. Might be useful to add a JSDoc block here.

src/scene/renderer/renderer.js
```diff
@@ -1055,25 +1044,57 @@ class Renderer {
...
+            const count = meshInstances.length;
+            totalMeshInstances += count;
+
+            for (let j = 0; j < count; j++) {
+                const meshInst = meshInstances[j];
+
+                // clear visibility
+                meshInst.visibleThisFrame = false;
+
+                // collect all meshe instances if we need to update their shaders. Note that there could
```
willeastcott, line 1067:
```suggestion
                // collect all mesh instances if we need to update their shaders. Note that there could
```

willeastcott (approved):
Very cool improvement. And a net reduction of 182 lines of code! 🚀

src/scene/renderer/renderer.js
mvaligursky: yep adding JSDocs for these. This is the generated code:

```
for (var _iterator = _createForOfIteratorHelperLoose(drawCalls), _step; !(_step = _iterator()).done;) {
	      var drawCall = _step.value;
               ...
}
```

src/scene/renderer/renderer.js
```diff
@@ -545,21 +553,21 @@ class Renderer {
...
         }
+
         // #if _PROFILER
         this._morphTime += now() - morphTime;
         // #endif
     }
 
     gpuUpdate(drawCalls) {
-        // skip everything with visibleThisFrame === false
+        // Note that drawCalls can be either a Set or an Array and contains mesh instances
```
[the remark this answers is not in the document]
mvaligursky: and done, thanks

Issue #5399, opened 2023-06-15
Spineboy engine example fails to work when using ES6 build


mvaligursky:
![Screenshot 2023-06-15 at 13 02 13](https://github.com/playcanvas/engine/assets/59932779/8eea1243-f88a-492d-85cb-4d8c2759422e)

mvaligursky:
I don't see the codebase calling https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/preventExtensions and so I'm not sure what is causing this yet.

mvaligursky:
if I build the examples this way, it works:
engine: `target=debug_es6 npm run watch`
examples: `ENGINE_PATH=../build/playcanvas.dbg.mjs/index.js npm run develop`

but when I use `npm run build` for both, it does not

@ellthompson - any suggestions here?

mvaligursky:
Similar issue takes place when using module build on examples that use post-effect, for example area-picker. See the selected line causing the issue, again related to access to `pc.` namespace

![image](https://github.com/playcanvas/engine/assets/59932779/45ea4b27-6c03-4430-8eff-6884df8d9904)

kungfooman:
> See the selected line causing the issue, again related to access to `pc.` namespace

It's because you cannot mix ES5 and ES6... i mean you can, but then you run into issues like that.

So the choice is all-in or all-out:

Works:

```js
function Animal() {}
function Bird(fName, lName, eId) {
  Animal.call(this);
}
bird = new Bird()
```

Doesn't work (because mixing ES5 and ES6):

```js
class Animal {}
function Bird(fName, lName, eId) {
  Animal.call(this);
}
bird = new Bird()
```

![image](https://github.com/playcanvas/engine/assets/5236548/356ee2db-52eb-4302-8e9f-5818ef5cc300)

Personally I don't waste my time with these kinds of issues any longer and I'm simply all-in for ES6.

Neither do I want or need and bundler during the development process, so I rewrote the example system into ES6 for myself to remove all the `*.tsx` sugar into pure `*.mjs`.

Why should we still cling to ES5? We already assume everyone uses WebGPU but ES5 is backwards-compatibility? I don't know how that works together.

Obviously we still want to ship the engine in ES5, but transpiling into ES5 is exactly what bundlers excel at.

mvaligursky:
[...]

Issue #5384, opened 2023-06-09
Engine examples builds twice on engine change when using ES6 build


mvaligursky:
Command to rebuild engine on change:
```
target=debug_es6 npm run watch
```

Command to rebuild examples using module engine build:
```
ENGINE_PATH=../build/playcanvas.dbg.mjs/index.js npm run develop
```

When a file in the engine folder changes, engine rebuilds, and it seems the examples rebuilds twice:
```
[0] [0] bundles src/iframe/index.mjs → dist/iframe...
[0] [0]  Finished at: 09:34:35
[0] [0] created dist/iframe in 5.4s
[0] [0] bundles src/iframe/index.mjs → dist/iframe...
[0] [0]  Finished at: 09:34:40
[0] [0] created dist/iframe in 5.2s
```

Is my understanding correct that it does the same build twice? If so, can this be avoided.

slimbuck:
The same happens for model-viewer and our other engine apps. Would also love to see it solved too (I've not had a chance to look into the cause myself).

mvaligursky:
Not that this does not seem to happen with non-modules build of the engine. Is it the problem with the module build creating / updating multiple files in the build folder, each of those triggering the build of examples?

mvaligursky:
This has been fixed in #5555

PR #5539, opened 2023-08-03
chore(examples): rollup type completion - lint warnings

epreston:
add type completion for rollup
jsdoc for inline plugin
address lint warnings
explicit __dirname so this can be used as umd or cjs

Fixes lint warnings.

### DIFF
--- a/examples/rollup.config.js
+++ b/examples/rollup.config.js
@@ -1,14 +1,24 @@
-import resolve from "@rollup/plugin-node-resolve";
+import date from 'date-and-time';
+import fs from 'node:fs';
+import fse from 'fs-extra';
+import { fileURLToPath } from 'node:url';
+import path from 'node:path';
+
+// 1st party Rollup plugins
+import alias from '@rollup/plugin-alias';
 import commonjs from "@rollup/plugin-commonjs";
 import replace from '@rollup/plugin-replace';
-import typescript from 'rollup-plugin-typescript2';
+import resolve from "@rollup/plugin-node-resolve";
 import terser from '@rollup/plugin-terser';
-import alias from '@rollup/plugin-alias';
+
+// 3rd party Rollup plugins
 import sourcemaps from 'rollup-plugin-sourcemaps';
-import date from 'date-and-time';
-import path from 'path';
-import fs from 'fs';
-import fse from 'fs-extra';
+import typescript from 'rollup-plugin-typescript2';
+
+/** @typedef {import('rollup').RollupOptions} RollupOptions */
+/** @typedef {import('rollup').Plugin} RollupPlugin */
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
 const ENGINE_PATH = process.env.ENGINE_PATH ? path.resolve(process.env.ENGINE_PATH) : path.resolve(`../build/playcanvas${process.env.NODE_ENV === 'development' ? '.dbg' : ''}.js`);
 const PCUI_PATH = process.env.PCUI_PATH || 'node_modules/@playcanvas/pcui';
@@ -33,6 +43,12 @@ function timestamp() {
     };
 }
 
+/**
+ * This plugin copies static files from source to destination.
+ *
+ * @param {staticFiles} targets - Array of source and destination objects.
+ * @returns {RollupPlugin} The plugin.
+ */
 function copyStaticFiles(targets) {
     function watch(src) {
         const srcStats = fs.statSync(src);
@@ -57,21 +73,21 @@ function copyStaticFiles(targets) {
     }
     return {
         name: 'copy-and-watch',
-        async load() {
+        load() {
             return 'console.log("This temp file is created when copying static files, it should be removed during the build process.");';
         },
-        async buildStart() {
+        buildStart() {
             if (process.env.NODE_ENV !== 'development') return;
             targets.forEach((target) => {
                 watch.bind(this)(target.src, target.dest);
             });
         },
-        async generateBundle() {
+        generateBundle() {
             targets.forEach((target) => {
                 fse.copySync(target.src, target.dest, { overwrite: true });
             });
         },
-        async writeBundle() {
+        writeBundle() {
             fs.unlinkSync('dist/copy.tmp');
         }
     };
@@ -91,6 +107,7 @@ const tsCompilerOptions = {
     }
 };
 
+/** @type {RollupOptions[]} */
 const builds = [
     {
         input: 'src/app/index.tsx',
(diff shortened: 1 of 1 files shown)


examples/rollup.config.js
```diff
@@ -1,14 +1,24 @@
...
 import terser from '@rollup/plugin-terser';
-import alias from '@rollup/plugin-alias';
+
+// 3rd party Rollup plugins
 import sourcemaps from 'rollup-plugin-sourcemaps';
-import date from 'date-and-time';
-import path from 'path';
-import fs from 'fs';
-import fse from 'fs-extra';
+import typescript from 'rollup-plugin-typescript2';
```
willeastcott, line 16: We should switch to the official typescript plug-in at some point

willeastcott (approved):

examples/rollup.config.js
epreston: Yes. I'm preparing a PR but it still breaks sourcemaps or produces very large ones.  The issue is the configuration is not consistent.

Quick question that can resolve it.  Do the examples javascript bundles need to all be umd ? Can they all be ESM ?

This builds 3 javascript bundles, 2 in ESM, one in UMD.  In this configuration typescript is doing the work to make them es5 instead of rollup.  If I had a direction on the outcome, I could make this build system consistent.

mvaligursky: These are the primary goals for the examples, but I'm not entirely sure how to get there (not my area of expertise):
- the result needs to be ES5 so that we can run them on old devices / browsers
- in debug build, we need source maps so that we can see the engine source code in the original form

At the moment the examples are built using ES5 engine build (unless you specify the module build on command line). The reason is that we could not get all examples to work due to some internal reasons (for now) using ES6 engine. See these two issues for the reasons we're aware of:
https://github.com/playcanvas/engine/issues/5399
https://github.com/playcanvas/engine/issues/5384

Thanks for doing an awesome job on this! Much appreciated.

epreston: That's what I need, thank you.  

Are we married to this build system ?  I could replace this with `vite` and 8 lines of config for a great dev experience.

In a way where you'll still have all the same utility (library swapping) and equivalent build products. It uses all the same tools under the hood, provides dev web server, hmr, consistency. I understand going rollup for the smallest, most treeshakable code for the engine library builds.  For everything else, cruise control.

If for any other reason, just being able to build without stopping the web server or having to mystery processes running after termination.

mvaligursky: Vite sounds interesting, and we might consider it, but we have many repos (internal / external) that use rollup, and for simplicity it's useful for us to keep the builds consistent - otherwise we find it hard to maintain.

But I'd be curious to see what @willeastcott has to say on this.

kungfooman: Cruise control would be 100% build step free - nothing to worry about at all during the dev cycle :wink: And a plain ES5/ES6 output once the dust has settled.

epreston: Vite is like "managed rollup", definitely keep fine grained rollup for specific build products.  For most of the development time, its build step free.

epreston: To be clear, I'm just talking about "examples" project. Only to polish the front end heavy or "new to library" developer experience.  Get them in and going with training wheels and a soft chair.

epreston: Please excuse all the messages: 

one last thing, the examples are built using React 18.  That does not support IE 11. (see: https://react.dev/blog/2022/03/08/react-18-upgrade-guide#dropping-support-for-internet-explorer)

In short, since march 31, examples browser works on the lowest supported target of React 18.  If you haven't had complaints, nobody is requiring es5 for examples, certainly not for old browser support.

willeastcott: Some thoughts/comments from me:

The example browser has never supported IE11 since PCUI doesn't support IE11 (I believe because of it's reliance on flexbox and maybe some other CSS features). There's a mode that you can run examples from the examples browser outside of the application and that's how we used to test in IE11, IIRC. But we've pretty much given up supporting IE11, even in the engine codebase, now.

On Vite, as @mvaligursky says, sounds interesting. I just had a quick look at their site. But my immediate concern is just introducing another technology which folks have to learn. Right now, once you know Rollup, can understand straight away how our repos are built. I'm not averse to switching to other build systems, but my preference would be that we keep things consistent. Also, I want to make sure we examine all the options. One thing I've been thinking about recently (and it seems @kungfooman has too) is whether ESM examples + import maps would be a better way to go. Dump the TypeScript mode. Remove the build system for the examples browser. Remove the React stuff and use vanilla PCUI. And just make it more simple. This is how the PCUI examples work and I _much_ prefer developing/debugging against those:

https://github.com/playcanvas/pcui/tree/main/examples

epreston: Yes, you are very right. This can be set up so examples can be run in browser or bundler without modification.  @mvaligursky invited me to comment, I got excited and wrote too much.

See: https://github.com/playcanvas/engine/issues/5541

Adjusting the examples that way is a great mark of quality. So is keeping typescript and doing the same for it. ESM + TS presented in a way that works the same bundle to browser is like having a mini playground / validation tool. It would be impressive.

If you feel adventurous, invite me to demonstrate it on something like `spine` that doesn't have a build system.  You'll see its as familiar as rollup (cause it is, check rollups front page) with a managed config.

kungfooman: I'm pretty much averse to anything that I feel is unnecessary extra steps (like `npm build`, `npm serve`, TS sugar, JSX sugar, ...)

If every path is so hardcoded that you need `npm serve` or otherwise a user can't even open a simple `index.html` via lighttpd on `/var/www` then I think something is wrong and not resilient enough, so I ended up changing multiple things I required at once.

I could make a PR with what I have, but it requires more work (there are many examples left that need to be reworked from TSX to ESM), even though I invested quite some time already to make the process work in the first place.

If we could put this on a branch and more people are willing to contribute to that, we could work it out over time, without asking too much from one individual.

I'm not advocating for completely removing the examples build system (it's fine to create minimized bundles and we need thumbnails generation), but to only use what modern browser technology is giving us to enable a fast and fun dev experience.

epreston: All those goals can be met. PR's incoming.

kungfooman: @epreston Nice, what's your plan? Can you break it down and make an issue for the first PR? First step should be imo `*.tsx -> *.jsx` and `*.ts -> *.mjs` (too keep the amount of changes reviewable and everything intact)

epreston: - Fix all the things external to engine blocking build improvements.
- Remove the workarounds in place to consume legacy code
- Predictable and easy to rationalise development
- Learn and share

PR #5540, opened 2023-08-03
Layer composition does not maintain list of shadow casters per light

mvaligursky:
follow up to #5536
contains changes already approved here: https://github.com/playcanvas/engine/pull/5549 (sorry about it, read the details there)

- shadow casters for culling are fetched from layers directly, to avoid expensive composition update when a caster mesh instance is added to a layer.


src/scene/renderer/shadow-renderer.js
```diff
@@ -155,24 +156,63 @@ class ShadowRenderer {
...
-                    count++;
+                    visible.push(meshInstance);
                 }
             }
         }
+    }
+
+    /**
+     * Culls the list of shadow casters used by the light by the camera, storing visible mesh
+     * instances in the specified array
```
willeastcott, line 176:
```suggestion
     * Culls the list of shadow casters used by the light by the camera, storing visible mesh
     * instances in the specified array.

```

willeastcott (approved):

PR #5547, opened 2023-08-03
fix(examples): undefined imageEntity

epreston:
Remove warning: Entity has no component with name element

no image entity with the name of `button` exists when addcomponent is called.  It is a good placeholder to show were background? images are added.

### DIFF
--- a/examples/src/examples/user-interface/button-basic.tsx
+++ b/examples/src/examples/user-interface/button-basic.tsx
@@ -79,7 +79,7 @@ class ButtonBasicExample {
 
                 // Button
                 const button = new pc.Entity();
-                button.addComponent("button", {});
+                button.addComponent("button");
                 button.addComponent("element", {
                     anchor: [0.5, 0.5, 0.5, 0.5],
                     height: 40,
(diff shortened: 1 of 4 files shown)


willeastcott:
I'm not keen on having examples with commented out code. I'd rather just add a `button` component with no options (if that works OK - I'm assuming that's legal from an API perspective).

epreston:
I'll check, test, and resubmit.

examples/src/examples/user-interface/button-basic.tsx
```diff
@@ -79,9 +79,7 @@ class ButtonBasicExample {
 
                 // Button
                 const button = new pc.Entity();
-                button.addComponent("button", {
-                    imageEntity: button
-                });
+                button.addComponent("button", {});
```
willeastcott, line 82:
```suggestion
                button.addComponent("button");
```

willeastcott: And is the `button` entity referenced anywhere else?

epreston: In the onclick events.

epreston:
I've taken my first step writing code that uses the engine.  It feels familiar, like scripting the dom.  Nice :-)

mvaligursky (approved):
looks good, fixes the warning too.

Issue #2626, opened 2020-12-17
Optimize adding elements into arrays in Layer


mvaligursky:
adding elements into arrays, especially meshInstances and shadowCasters, is using .indexOf to test if element is already in the array. This is very expensive for large arrays and should be accelerated by perhaps using a Map.

Maksims:
Set has very fast check if it is in the Set.

mvaligursky:
Done in https://github.com/playcanvas/engine/pull/5549, closing

PR #5548, opened 2023-08-04
Optimize add/remove of meshes instances to a Layer using sets

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/2626


mvaligursky:
closing, wrong parent branch

PR #5549, opened 2023-08-04
Optimize add/remove of meshes instances to a Layer using sets

### LINKED ISSUE #2626: Optimize adding elements into arrays in Layer

mvaligursky:
Fixes #2626

### DIFF
--- a/src/scene/layer.js
+++ b/src/scene/layer.js
@@ -77,12 +77,28 @@ class Layer {
     meshInstances = [];
 
     /**
+     * Mesh instances assigned to this layer, stored in a set.
+     *
+     * @type {Set<import('./mesh-instance.js').MeshInstance>}
+     * @ignore
+     */
+    meshInstancesSet = new Set();
+
+    /**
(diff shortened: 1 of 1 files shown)


willeastcott (approved):
Did you detect a performance improvement?

mvaligursky:
> Did you detect a performance improvement?

I did not test, but based on my notes, adding 5000  mesh instances in the Hierarchy example was 70ms on my previous Mac, and I'm confident that is now gone, as we don't do indexOf on the array anymore, which was the reason for the high cost.

mvaligursky:
Hmm I used @slimbuck 's suggestion here and created a branch that depends on another branch. This PR got approved, I merged it .. but not to main but to that other branch. And so the same code was added to this PR :( https://github.com/playcanvas/engine/pull/5540
(just review it again there pls)

Issue #5542, opened 2023-08-03
LitShader recompile


LeXXik:
There is a similar closed ticket here:
https://github.com/playcanvas/engine/issues/5512

However the original issue of it was fixed in 1.65.0. This recompilation seems to be caused by another reason, hence opening a new ticket.

Repro project:
https://playcanvas.com/editor/scene/1815575

Just disable Light 1 entity and enable it back. When you enable it, compilation takes place again, even though the shader should already be in the cache under id 1. This doesn't happen with Light 2. The difference between both lights is Light 1 has real time shadow update.

https://github.com/playcanvas/engine/assets/5677782/2560ea64-10f8-4761-a3d3-8d28ada206e1

_Originally posted in https://github.com/playcanvas/engine/issues/5512#issuecomment-1660745031_

PR #5550, opened 2023-08-04
[Fix] Avoid additional shaders creation caused by out of order lights

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/5542

When lights stored by the Layer were in different order, a different key would be generated causing new shaders to be created.

To avoid this, the lights the Layer stores are sorted by their key.

Additionally, small refactor related to shader key generation has been done, making those multiline and easier to read, and storing those on the Shader instance in debug mode, to allow us to easily compare why a shader was generated (see the generation and processing keys)

Screenshot 2023-08-04 at 14 25 05

### DIFF
--- a/src/platform/graphics/shader-processor-options.js
+++ b/src/platform/graphics/shader-processor-options.js
@@ -75,7 +75,7 @@ class ShaderProcessorOptions {
     }
 
     /**
-     * Generate unique key represending the processing options.
+     * Generate unique key representing the processing options.
      *
      * @returns {string} - Returns the key.
      */
--- a/src/platform/graphics/shader-processor-options.js
+++ b/src/platform/graphics/shader-processor-options.js
@@ -77,14 +77,20 @@ class ShaderProcessorOptions {
     /**
      * Generate unique key representing the processing options.
      *
+     * @param {import('./graphics-device.js').GraphicsDevice} device - The device.
      * @returns {string} - Returns the key.
      */
-    generateKey() {
+    generateKey(device) {
         // TODO: Optimize. Uniform and BindGroup formats should have their keys evaluated in their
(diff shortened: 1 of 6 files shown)


src/platform/graphics/shader-processor-options.js
```diff
@@ -75,16 +75,22 @@ class ShaderProcessorOptions {
...
-    generateKey() {
+    generateKey(device) {
         // TODO: Optimize. Uniform and BindGroup formats should have their keys evaluated in their
         // constructors, and here we should simply concatenate those.
-        return JSON.stringify(this.uniformFormats) +
-        JSON.stringify(this.bindGroupFormats) +
-        this.vertexFormat?.renderingHashString;
+        let key = JSON.stringify(this.uniformFormats) + JSON.stringify(this.bindGroupFormats);
+
+        // WebGPU shaders are processes per vertex format
```
willeastcott, line 88: processed

src/scene/shader-lib/program-library.js
```diff
@@ -6,6 +6,7 @@ import { Shader } from '../../platform/graphics/shader.js';
 import { SHADER_FORWARD, SHADER_DEPTH, SHADER_PICK, SHADER_SHADOW } from '../constants.js';
 import { ShaderPass } from '../shader-pass.js';
 import { StandardMaterialOptions } from '../materials/standard-material-options.js';
+import { hashCode } from '../../core/hash.js';
```
willeastcott, line 9: Ideally this import would be at the top because it's from the core

willeastcott (approved):

src/platform/graphics/shader-processor-options.js
mvaligursky: cheers!

PR #5551, opened 2023-08-04
Simplify some shadow penumbra math / avoid divide by zero

mvaligursky:
- small improvement

### DIFF
--- a/src/scene/renderer/forward-renderer.js
+++ b/src/scene/renderer/forward-renderer.js
@@ -259,7 +259,7 @@ class ForwardRenderer extends Renderer {
                 this.lightShadowIntensity[cnt].setValue(directional.shadowIntensity);
 
                 const projectionCompensation = (50.0 / lightRenderData.projectionCompensation);
-                const pixelsPerMeter = 1.0 / (lightRenderData.shadowCamera.renderTarget.width / directional.penumbraSize);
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

kungfooman:
I'm a bit confused by this, I get that `1 / (a / b)` is the same as `b / a`, but how does it prevent div by zero?

```js
for (let a = -10; a <= 10; a++) {
  for (let b = -10; b <= 10; b++) {
    const x = 1 / (a / b);
    const y = b / a;
    if (x !== y) {
      console.log("!=", x, y);
    }
  }
}
```

Output:

image

The numbers just seem to become more "numerically stable".

mvaligursky:
in our case of `1 / (a / b)` - `a` was a render target width, so never zero, but `b` was a softness, so a user can set that to 0 and generate NaNs.
`b / a` is the same result, but we never divide by `b`, only by `a`.

And it's a simpler code too .. no need for two divides.

kungfooman:
Thank you for explaining, funny thing is I cannot generate a NaN with any combination, the Infinity turns into 0 again:

```js
width = 512;
penumbraSize = 0;
1.0 / (width / penumbraSize); // is 0
```

image

But anyway, code is way cleaner/shorter/better now without "internal" infinities.

Only possibility to generate a `NaN` is when both `width` and `penumbraSize` is zero.

PR #5552, opened 2023-08-04
Shader generators are converted to modules

mvaligursky:
- new base class for shader generators, called `ShaderGenerator`, implements functionality from now deleted `common.js` as static functions (so that they can be called from outside of generators too)
- all shader generator (basic, standard, particle, sky, pass-through) extend this class, and generate and export a single instance of it


src/scene/shader-lib/programs/particle.js
```diff
@@ -1,28 +1,27 @@
...
 import { BLEND_ADDITIVE, BLEND_MULTIPLICATIVE, BLEND_NORMAL } from '../../constants.js';
 import { shaderChunks } from '../chunks/chunks.js';
+import { ShaderGenerator } from './shader-generator.js';
 
-import { gammaCode, tonemapCode } from './common.js';
-
-const particle = {
-    generateKey: function (options) {
+class ShaderGeneratorParticle extends ShaderGenerator {
+    generateKey(options) {
```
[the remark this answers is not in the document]
mvaligursky: standard.js contains some members as well .. I'm not sure I'd want to change those and all functions in it to static, instances seems more flexible for the future too.

src/scene/shader-lib/programs/basic.js
```diff
@@ -9,10 +9,10 @@ import {
 } from '../../constants.js';
 import { ShaderPass } from '../../shader-pass.js';
 
-import { begin, end, fogCode, skinCode } from './common.js';
+import { ShaderGenerator } from './shader-generator.js';
 
-const basic = {
-    generateKey: function (options) {
+class ShaderGeneratorBasic extends ShaderGenerator {
```
slimbuck, line 14: What's the purpose of this extends?

mvaligursky: To give all those classes base class / shared type information.

slimbuck: I don't see that used anywhere?

The structure of this whole class hierarchy is very non-standard (and for me confusing).

ShaderGenerator isn't a real base class because it defines no base functionality. Rather it is a set of static helper functions.

Then ShaderGeneratorBasic and the rest implement what looks like a consistent set of functions (which presumably program-libraray expects, but never defines).

Ultimately the structure/interface of ShaderGenerator isn't defined anywhere and is left as implicit.

I don't think there needs to be any inheritance here and probably program-library should define what structure a 'generator' takes. Each implementation of generator just implements that interface.

mvaligursky: At some point I'm going to add type information to the program-library, and this is the type of the generator. Ideally this would be an interface, but unfortunately JS does not have a concept of this. But this is the only way to assign a single (base) type to multiple classes.

slimbuck: If program-library just uses duck typing on generators, why do generators need a class hierarchy?

mvaligursky: I added some JSDocs to the program-libray that use the base class type - type safety is where we're going to with the engine changes.

slimbuck (approved):
Approving under the assumption this mess is cleaned up in future :)

Issue #5437, opened 2023-06-25
WebGPU `requestAdapter` failure isn't handled


kungfooman:
![image](https://github.com/playcanvas/engine/assets/5236548/4c6eef7c-ce9d-4a4e-9cf1-9c461ef502a8)

This messes up the examples for me too, because they default to WebGPU:

![image](https://github.com/playcanvas/engine/assets/5236548/9c9b5461-3fe5-40e7-b6a3-fd6e787d6039)

I'm on Linux/Chrome, which still doesn't support WebGPU :hourglass: :eyes: 

But it's exporting `navigator.gpu` anyway:

![image](https://github.com/playcanvas/engine/assets/5236548/5a693999-c81e-4494-a324-c3103fe733c5)

kungfooman:
@ellthompson Can you set default renderer to WebGL2?

You can do so via `value: DEVICETYPE_WEBGL2` in `examples/src/app` `Example` class:

```js
                  <SelectInput
                    id='deviceTypeSelectInput'
                    options={[
                        { t: deviceTypeNames[DEVICETYPE_WEBGL1], v: DEVICETYPE_WEBGL1 },
                        { t: deviceTypeNames[DEVICETYPE_WEBGL2], v: DEVICETYPE_WEBGL2 },
                        { t: deviceTypeNames[DEVICETYPE_WEBGPU], v: DEVICETYPE_WEBGPU }
                    ]}
                    value={DEVICETYPE_WEBGL2}
                    onSelect={this.onSetPreferredGraphicsDevice}
                    prefix='Active Device: '
                    // @ts-ignore this is setting a legacy ref
                    ref={this.deviceTypeSelectInputRef}
                />
```

mvaligursky:
Ideally we'd want to handle requestAdapter returning null, and fallback to WebGL2 device, but that's not straightforward.

mvaligursky:
This should be fixed by #5721

PR #5555, opened 2023-08-07
Examples es6

kungfooman:
Paraphrasing some thoughts of @willeastcott into a list with small changes:

- Rewrite /examples/ as ESM modules
- Using import maps for every library (react, react-dom, pcui etc.)
- Dump the TypeScript mode
- Simplifying the build system of the examples browser
- And just make it more simple
- This is how the PCUI examples work and I much prefer developing/debugging against those: https://github.com/playcanvas/pcui/tree/main/examples

This is heavily work-in-progress and I need to make more changes until it's testable locally, I just wanna flash out the PR already for discussion.

However, I already figured some technical improvements aswell, e.g. right now:

image

In my version with dumped TS mode, no problem to change the shader:

image

I also reworked the code that we can use a `async example` method, so we are not forced to use `.then(...)`.


Maksims:
This is a huge amount of work.
Does this PR goes well with [engine guidelines](https://github.com/playcanvas/engine/blob/main/.github/CONTRIBUTING.md)? So the engine promotes recommended and well-supported syntax and patterns?

kungfooman:
> This is a huge amount of work. Does this PR goes well with [engine guidelines](https://github.com/playcanvas/engine/blob/main/.github/CONTRIBUTING.md)? So the engine promotes recommended and well-supported syntax and patterns?

I try to use clean patterns as much as possible:

 - I rewrote many `window` property accesses, when it doesn't had to be on `window`
 - Instead of polling `pc.app` every 50ms, I just return `app` from the `example` function
 - that example function can be `async example` aswell, removing needless `.then(...)` indentations
 - simplify code, whenever something is not immediately obvious what the code is doing (I add a JSDoc including `@example`)
 - Separation of Concerns: I moved JS code squeezed insde `*.html` into `*.mjs`
 
Please feel free to point out specific issues you find :+1:

willeastcott:
So this is super-interesting. Thanks for submitting this, @kungfooman! So a few comments:

1. You have a whole bunch of conflicts to resolve in this PR. Might be a good idea to resolve those sooner rather than later.
2. Should the examples default to run on the unbuilt engine modules?
3. Can I run any given engine example outside of the examples browser? I can imagine that'd be super useful for iterating on new examples or fixing bugs.
4. Are there any issues will deploying/running the examples on GitHub Pages?
5. I love the idea of moving to async/await to simplify the example code.
6. How should we progress this PR? I believe this could be polished to a state it can be merged. But I don't want to get caught up in small details right now because this is a big change. It'd be good to resolve the bigger issues at the start. So what do you see as the 'big issues' that need to be discussed now?

kungfooman:
Thank you for the input @willeastcott 

> 1. You have a whole bunch of conflicts to resolve in this PR. Might be a good idea to resolve those sooner rather than later.

Merged `main` :+1: 

> 2. Should the examples default to run on the unbuilt engine modules?

The sheer amount of files is the biggest problem so far, locally everything works fine, but CDN's like unpkg block after too many requests. Solution is to ship bundled *.mjs files again (we had that in the past, but we changed `dir`/`file` setting in `rollup` to help tree-shaking). Now we have tree-shaking, but CDN's block us after too many requests... :sweat_smile:

This is maybe the first change we can do in an extra PR to support modules better.

> 3. Can I run any given engine example outside of the examples browser? I can imagine that'd be super useful for iterating on new examples or fixing bugs.

Right now only via `/examples/src/iframe/?category=Animation&example=BlendTrees2DDirectional` (as it was).

I like the idea of adding a `*.html` for every example containing only what's absolutely required. And then instead of the iframe logic stuff, we simply set the source of the iframe to the "standalone" example `*.html`... makes everything even a little bit easier again and less error prone.

> 4. Are there any issues will deploying/running the examples on GitHub Pages?

huge todo right now

> 5. I love the idea of moving to async/await to simplify the example code.

Yep, every removed level of intendation makes it feel a little bit easier :100: 

> 6. How should we progress this PR? I believe this could be polished to a state it can be merged. But I don't want to get caught up in small details right now because this is a big change. It'd be good to resolve the bigger issues at the start. So what do you see as the 'big issues' that need to be discussed now?

1) We need an official bundled-into-one-file `playcanvas.whole.mjs` (`playcanvas.entirety.mjs`, `playcanvas.singlefile.mjs`?) which we can reference in import maps

2) Once the dust has settled (2-3 days), I would like some support converting every example. It's a rather simple process I can make a video about, but it still takes quite some time to change/test everything.

willeastcott:
> The sheer amount of files is the biggest problem so far, locally everything works fine, but CDN's like unpkg block after too many requests. Solution is to ship bundled *.mjs files again (we had that in the past, but we changed dir/file setting in rollup to help tree-shaking). Now we have tree-shaking, but CDN's block us after too many requests... 😅

It feels like the importmap should point to `src\index.js` for development and `build\playcanvas.mjs` when running from production.

examples/src/examples/graphics/render-asset.mjs
```diff
@@ -0,0 +1,118 @@
...
+        });
+    });
+    return app;
+}
+
+export class RenderAssetExample {
+    static CATEGORY = 'Graphics';
+    static NAME = 'Render Asset';
+    static WEBGPU_ENABLED = true;
+    static example = example;
```
mvaligursky, line 117: could we embed the example function here?

kungfooman: Yes, that's still possible. I thought about that these classes could be removed entirely to make it simpler to use examples (for standalone .html files later). They don't contain "methods" (aka no access to `this`) and it's just adding a level of intendation. Having it not in there also helps with removal of Babel/Prettier and also the handmade "parsers" (just to remove level of intendation for code reload in Monaco). Another reason is to have a simple toString() which resembles an entire function:

image

So I pretty much just saw advantages to rewrite the code like that to make the code shorter/smaller/easier to read.

kungfooman:
> Nice stuff. Got any instructions on how to get it to run locally? I'm keen to give it a spin.

Hey @mvaligursky, thank you for the interest! Do you have lighttpd or Apache running? On Linux, you can just put the engine in `/var/www` or e.g. `C:/xampp/htdocs`

1) Fetch PR: `git fetch origin pull/5555/head:examples_es6` + git check it out 
2) `cd examples`
3) `npm install` (the importmap directs the browser into `examples/node_modules/*`)
4) `npm run build:example:data` (this is required because otherwise we would have to increase bundle size - importing all examples in Examples UI just to display a list of all examples)
5) Go to http://127.0.0.1/playcanvas-engine/examples/app.html#/misc/hello-world

I'm still a bit annoyed that we need (4), but it can be quite easily refactored (I already didn't require it, but removed it for the sake of reducing tersered bundle size)... but my main goal is to finish this PR and not break anything right now, so that the old `npm run build` and `npm run serve` is just like it was.

Happy to hear your feedback :+1:

Maksims:
Should PlayCanvas Spine be included as a dependency in package.json instead of a file in this repo?

kungfooman:
> Should PlayCanvas Spine be included as a dependency in package.json instead of a file in this repo?

I was about to suggest that since I'm coming closer to completion here and want to clean up everything that is not "really" required. I think it was a mistake to add `scripts/spine/playcanvas-spine.3.8.js` in the first place... I don't wanna repeat that with the MJS version of it (it's ~ 420 kB).

Maksims:
I believe to ensure github sees the file as renamed, for safety it should be renamed and committed firest, before changing its content and committing again.

kungfooman:
> I believe to ensure github sees the file as renamed, for safety it should be renamed and committed firest, before changing its content and committing again.

Yep, good point, maybe it helps to just rename the files back and forth with same content in two commits... I will try it once I'm done with current feature I'm working on (trying to gravite towards near-instant-reload times instead of waiting for 2s).

mvaligursky:
> it should be renamed and committed firest

I typically use this to handle file renames: `git add -A .`

mvaligursky:
> quick quick https://www.youtube.com/watch?v=fPH13kNNDss 😅

Video unavailable
This video is private

kungfooman:
@mvaligursky Sorry, now it should work (just rushed to put it on YouTube before leaving PC)

mvaligursky:
Editing controls looks amazing!

kungfooman:
This is pretty much ready for review, a live demo for testing is here: http://playcanvas.killtube.org/examples_es6/examples/dist/

You can reevaluate every example by pressing Shift+Enter (works with code + shaders). React controls are updated immediately right now.

Hot reload updating is quite fast now, previously is took around 2 seconds for a hot reload, now it's nearly instant (depends on how many assets are loaded).

mvaligursky:
I see this when loading examples from your link .. is this getting fixed?

Screenshot 2023-08-22 at 18 29 17

I'll play more with it all tomorrow and let you know if I find any problems, thanks so much!

Maksims:
Tested on mobile, unfortunately code does not get loaded.
Also, is it possible to remove playcanvas-spine library from the repository, and add it as a package.json dev dependency?

kungfooman:
> I see this when loading examples from your link .. is this getting fixed?

That's kinda funny, I fixed it (the issue is a different path for `npm run serve` vs "served by normal httpd" version), so I added a `` fallback construct via `<object data="../arkit.png" ...>...</object>`... it works to catch the right image, but it still shows this error, so I have to change that, because it spams the console... will work it out right away

@Maksims Good points, I will just remove the playcanvas-spine.mjs for now, because I'm running examples in ES5 mode right now to keep things simple. Then we can have dedicated issues which don't feel too confusing (or simply wasting time because it's not obvious what has to be done) hopefully.

I didn't test it on mobile one time so far, so I have to see what's breaking that. Thank you both for testing!

kungfooman:
Hm, first time I tried it on my old Android from 2013 (with new OS/Chrome tho, Android device is connected via USB so I can remote debug via Chrome protocol), worked right away:

Locally, via `npm run serve`:

image

Link I shared here:

image

What kind of device are you testing with?

Maksims:
> What kind of device are you testing with?

Try in portrait.
Screenshot_2023-08-23-10-53-23-95_40deb401b9ffe8e1df2f1cc5ba480b12

kungfooman:
Oh right, I think I need an orientationchange event

kungfooman:
Fixed every issue mentioned here.


**Live demo**

http://playcanvas.killtube.org/examples_es6/examples/dist/

Before PR for direct comparison: 

https://playcanvas.github.io/#/animation/blend-trees-1d

In the case of resizing the browser window towards portrait mode, my version already works better:

Old one after a bit of resizing:

image

I converted the functional components towards class-based components aswell, because functional state change handling is lunacy...

mvaligursky:
> AR examples failed to load, while they work on current version.

those probably need https and not just http. Same as WebGPU - running that requires https.

kungfooman:
Currently I have this error on Chrome/iPhone: `TypeError: undefined is not an object (evaluating 'screen.orientation.addEventListener'`).. gotta figure out best solution here, maybe this deprecated API has better overall support: https://developer.mozilla.org/en-US/docs/Web/API/Window/orientationchange_event

kungfooman:
Fixed orientation change issue and spawned a HTTPS proxy for `cd examples && npm run serve`:

https://examples.killtube.org/

However, I don't have any VR/AR testing device (technically I have one, but Zuckerberg made sure to break it via planned obsolescence).

Maybe someone can look into this AR issue deeper, if it still fails to work? I didn't work with AR yet.

Maksims:
Unfortunately latest version is not working for me:
image

kungfooman:
I just had a small bug to fix and was running `npm run build` with `npm run serve` in a background `screen` session, not sure if this is related, because I've never seen that bug before :thinking: Can you try Ctrl+R with DevTools open?

kungfooman:
Stupid me, it's an issue I reported some time back with original examples: https://github.com/playcanvas/engine/issues/5437

kungfooman:
AR works for me, tested every example, but not sure about real devices:

image

image

(first it renders black, but it works after mouse-clicking once - same behaviour in old editor).

So more testing/feedback would be great

mvaligursky:
so what's the latest on running this locally? I got the branch, did `npm install`, and then just open the app.html on local host? That does not seem to work for me.

kungfooman:
> so what's the latest on running this locally? I got the branch, did `npm install`, and then just open the app.html on local host? That does not seem to work for me.

The easiest/shortest is to just `npm run build` after `npm install`, it does the two required sub-steps which make `app.html` work:

`npm run build:example:data && npm run build:standalone`

It is now rather easy to aim for "100% pure and build-free" MJS, but I don't want to squeeze that any longer in this PR (it already feels too big).

If you use `app.html`, it is already build-free for the UI part (you don't need to `npm run build` after changing files in `examples/src/app/*`), but changing `examples/src/examples/*` themselves requires a build step to generate the standalone files, which the UI will load.

mvaligursky:
@kungfooman - I get this when trying to build it - any idea what the problem is?

```
npm run build

> examples-browser@0.0.0 build
> npm run build:example:data && npm run build:standalone && npm run build:directory && cross-env NODE_ENV=production rollup -c


> examples-browser@0.0.0 build:example:data
> node ./scripts/example-data.mjs

node:internal/errors:477
    ErrorCaptureStackTrace(err);
    ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package '/Users/mvaligursky/code/EXTERNAL/kungfoo-examples/examples/node_modules/playcanvas/' imported from /Users/mvaligursky/code/EXTERNAL/kungfoo-examples/examples/src/examples/input/gamepad.mjs
    at new NodeError (node:internal/errors:388:5)
    at legacyMainResolve (node:internal/modules/esm/resolve:317:9)
    at packageResolve (node:internal/modules/esm/resolve:897:14)
    at moduleResolve (node:internal/modules/esm/resolve:959:20)
    at defaultResolve (node:internal/modules/esm/resolve:1174:11)
    at ESMLoader.resolve (node:internal/modules/esm/loader:605:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:318:18)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:80:40)
    at link (node:internal/modules/esm/module_job:78:36) {
  code: 'ERR_MODULE_NOT_FOUND'
}

Node.js v18.4.0
```

kungfooman:
@mvaligursky Is that on Mac? It seems like it doesn't know about `playcanvas` because of a missing `cd examples && npm link ..`

Did you do `npm install` in `/examples/` dir?

If it's on Mac, I have an old one on which I can try it aswell.

Lets make sure everything is installed and build:

```sh
cd /Users/mvaligursky/code/EXTERNAL/kungfoo-examples/
npm install
npm run build
cd examples
npm install
npm run build
```

mvaligursky:
Hey .. I forgot to build the engine in that folder, my bad.

So now I get this on the last of your steps (`npm run build`)

```
npm run build

> examples-browser@0.0.0 build
> npm run build:example:data && npm run build:standalone && npm run build:directory && cross-env NODE_ENV=production rollup -c


> examples-browser@0.0.0 build:example:data
> node ./scripts/example-data.mjs

file:///Users/mvaligursky/code/EXTERNAL/kungfoo-examples/examples/node_modules/@playcanvas/pcui/react/dist/module/src/components/SliderInput/index.mjs:10
const IS_CHROME = /Chrome\//.test(navigator.userAgent);
                                  ^

ReferenceError: navigator is not defined
    at file:///Users/mvaligursky/code/EXTERNAL/kungfoo-examples/examples/node_modules/@playcanvas/pcui/react/dist/module/src/components/SliderInput/index.mjs:10:35
    at ModuleJob.run (node:internal/modules/esm/module_job:198:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:409:24)
    at async loadESM (node:internal/process/esm_loader:85:5)
    at async handleMainPromise (node:internal/modules/run_main:61:12)

Node.js v18.4.0

```

And yes, this is Mac.

kungfooman:
Nice, we are very close then, that's an issue I only had on an older node version and this is the only occurance why PCUI can't be imported in node. We have to fix that in 

https://github.com/playcanvas/pcui/blob/1a087dae207fc0276d49289c9075c126d0ec3434/src/components/SliderInput/index.ts#L11

For my server with an older node version I just had it fixed quickly directly in `node_modules`:

```js
const IS_CHROME = /Chrome\//.test(globalThis.navigator?.userAgent);
```

kungfooman:
I made sure that @mvaligursky's PR is reflected again: https://examples.killtube.org/#/graphics/multi-view

kungfooman:
@mvaligursky We have it fixed now in PCUI, so we just need a new release. Then I can update towards that release in `package.json` and then it works after `npm install` without manual changes in `examples/node_modules`

kungfooman:
I realized the Null renderer and MiniStats don't play nice together right now, gotta see how to fix that...

image

mvaligursky:
> Null renderer and MiniStats don't play nice together

we have some code in examples that disable MiniStats for WebGPU .. I think we need to do the same for the Null renderer.

willeastcott:
> Could you publish PCUI?

Done: https://github.com/playcanvas/pcui/releases/tag/v4.1.2

kungfooman:
If a user clicks on MiniStats and happens to use Null renderer, it will simply display a message now:

image

@willeastcott Thank you, now the build process has no issues any longer :100:

mvaligursky:
Any thoughts on this when building the examples?

Screenshot 2023-09-04 at 09 28 53

kungfooman:
Hey @mvaligursky, thank you for the review!

I think I worked it all out now: https://examples.killtube.org/#/misc/mini-stats

1) Fixed circular dependency
2) Remove MiniStats "not supported" alert/message
3) MiniStats is running correct engine again
4) If you start the examples like this: `ENGINE_PATH=../src/index.js npm run develop`

You have access to all files again:

image

Changes in the engine are immediately watched/mirrored, you just need a `Ctrl+r`.

> we no longer need to run npm run build, which is great, but it's needed by npm run build:thumbnails. Can this command just serve internally or something along the lines?

Right now we still need `npm run build` in any case, to generate the standalone files e.g. aswell after they changed - I'm not sure what you mean with serving internally, can you elaborate?

-----

Suddenly Monaco and React is acting strange and the error messages are not exactly super helpful, just throwing errors in areas that always used to work (IntelliSense in Monaco)... need to investigate... I have a feeling it's maybe some CDN updates that changed, but unsure:

image

kungfooman:
> So basically, can `build:thumbnails` internally set up serving, so that I don't have to?

That's a good idea, that will certainly make it easier to use and less "magic knowledge" will be required... I will see what I can do :mag: 

I will start with the React errors as they break some controls UI and/or spam the console, it feels completely random that they suddenly appear... maybe a sign to ditch `pcuiReact` with vanilla `pcui` and refactor/cleanup a bit more.

kungfooman:
> So basically, can `build:thumbnails` internally set up serving, so that I don't have to?

Implemented this here: https://github.com/playcanvas/engine/pull/5555/commits/6f92a8ef26099752f278d0e555a2a79c4ce40ae7

I picked a new port (12321), so it doesn't conflict with the default 5000. Works nicely, biggest issue I had was with the `serve` package - very peculiar set of opinions (simply removing `.html?miniStats=false` from URL's...).

I think this is ready to review again - I don't know how often I thought it's ready by now :see_no_evil:

mvaligursky:
> I think this is ready to review again - I don't know how often I thought it's ready by now 🙈

Awesome, I'll have a look again, thanks!

mvaligursky:
@kungfooman - have you had chance to update readme yet? Things we need to document, if there is any change since current readme:
- how to run in dev mode, before `npm run develop`, which was using es5 debug build of the engine - does that still work, or do we have a new way to work in debug mode of the engine / examples with live update? Does this use ES6 or ES5 build of the engine? Can we use either?
- how to build using release mode, when we need to publish. Any changes to publishing? Especially in relation to PCUI - before we'd copy over the build folder but leave PCUI folder in place, is that still needed?

kungfooman:
@mvaligursky Sorry, I was a bit sidetracked by inspecting the broken examples but you just merged both fixes (template string + Preprocessor bug).

Actions I have to take:

1) Go over the `README.md` and make sure that everything is up to date - I could also clarify the requirement for the PCUI folder
2) I also realized I need to trigger `npm run build:standalone` when the examples change in dev/watch more

> * how to run in dev mode, before `npm run develop`, which was using es5 debug build of the engine - does that still work, or do we have a new way to work in debug mode of the engine / examples with live update? Does this use ES6 or ES5 build of the engine? Can we use either?

You can run both ES5 and ES6 by setting ENGINE_PATH just like in `README.md` right now:

```sh
ENGINE_PATH=../build/playcanvas.mjs/index.js npm run develop
```

> * how to build using release mode, when we need to publish. Any changes to publishing? Especially in relation to PCUI - before we'd copy over the build folder but leave PCUI folder in place, is that still needed?

Good question, the PCUI statement in `README.md`:

> 4) Copy the contents of the `./dist` directory to the root of the [playcanvas.github.io](https://github.com/playcanvas/playcanvas.github.io) repository. Be sure not to wipe the contents of the `pcui` subdirectory in that repository.

Has no relation to the examples browser. It's simply the link for pcui documentation: https://playcanvas.github.io/pcui/

For example googling `playcanvas pcui`:

image

mvaligursky:
Thanks for the update! I'm testing it and will add some comments in random order:

- when I use `npm run develop` on the examples, it seems to use a release engine. Can we default to a debug engine? Either es5 (my preference, as that's the same engine we use by the Editor currently, so just a single watch command) or es6.

- Actually, `ENGINE_PATH=../build/playcanvas.dbg.js npm run develop` does not work. Not required if the module version works, which I'll test next. In that case, `npm run develop` should default to a module engine.

- In the engine folder I did `npm run build:extras` and `target=debug_es6 npm run watch`, and in the editor folder I did `ENGINE_PATH=../build/playcanvas.dbg.mjs/index.js npm run develop` and I get:

```
mvaligursky@Q25P7P6TVX examples % ENGINE_PATH=../build/playcanvas.dbg.mjs/index.js npm run develop

> examples-browser@0.0.0 develop
> cross-env concurrently --kill-others "npm run watch" "npm run serve"

[0] 
[0] > examples-browser@0.0.0 watch
[0] > npm run build:example:data && npm run build:standalone && npm run build:directory && cross-env concurrently --kill-others "npm run rollup:watch"
[0] 
[1] 
[1] > examples-browser@0.0.0 serve
[1] > serve dist -l 5000 --no-request-logging --config ../serve.json
[1] 
[1]  UPDATE  The latest version of `serve` is 14.2.1
[1]  INFO  Accepting connections at http://localhost:56446
[0] 
[0] > examples-browser@0.0.0 build:example:data
[0] > node ./scripts/example-data.mjs
[0] 
[0] node:internal/errors:477
[0]     ErrorCaptureStackTrace(err);
[0]     ^
[0] 
[0] Error [ERR_MODULE_NOT_FOUND]: Cannot find package '/Users/mvaligursky/Snapchat/Dev/EXTERNAL/kungfoo-examples/examples/node_modules/playcanvas/' imported from /Users/mvaligursky/Snapchat/Dev/EXTERNAL/kungfoo-examples/examples/src/examples/graphics/area-lights.mjs
[0]     at new NodeError (node:internal/errors:388:5)
[0]     at legacyMainResolve (node:internal/modules/esm/resolve:317:9)
[0]     at packageResolve (node:internal/modules/esm/resolve:897:14)
[0]     at moduleResolve (node:internal/modules/esm/resolve:959:20)
[0]     at defaultResolve (node:internal/modules/esm/resolve:1174:11)
[0]     at ESMLoader.resolve (node:internal/modules/esm/loader:605:30)
[0]     at ESMLoader.getModuleJob (node:internal/modules/esm/loader:318:18)
[0]     at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:80:40)
[0]     at link (node:internal/modules/esm/module_job:78:36) {
[0]   code: 'ERR_MODULE_NOT_FOUND'
[0] }
[0] 
[0] Node.js v18.4.0
[0] npm run watch exited with code 1
```

What am I missing?

mvaligursky:
Options 1 sounds good to me to be honest. A separate PR would be good I think?

kungfooman:
I hope I tested every command mentioned here, I can't find any edge case right now that doesn't work.

I like this command: `ENGINE_PATH=../src/index.mjs npm run develop` - doesn't require a 2nd shell to build the engine and serves/rebuilds everything.

The last issue I still have to fix is to rebuild the standalone Example HTML files when the Example MJS code is changed.

There are historically quite some ways to declare a preferred engine:

 - `Example#ENGINE` (lowest priority) - either `PERFORMANCE`, `DEBUG` or `undefined`
 - `NODE_ENV` (2nd lowest priority) - either `production` or `development`
 - `ENGINE_PATH` (highest priority)

If none of these sources are given, we simply pick build/playcanvas.js (ES5)

It's not super complicated, but also not entirely obvious any longer... I had the idea to add another button to the Controls like:

```html
Engine: <select>
<option>playcanvas.js
<option>playcanvas.min.js
<option>playcanvas.dbg.js
<option>playcanvas.prf.js
<option>ENGINE_PATH (if given)
</select>
```

It would default to the preferred engine that was given, but then it would be just two mouse clicks to use another engine anyway for quick testing. Not sure if it would overcomplicate the UI or be helpful :sweat_smile:

mvaligursky:
> but then it would be just two mouse clicks to use another engine anyway for quick testing. Not sure if it would overcomplicate the UI or be helpful

I would be helpful for occasionally testing something by us, but not for majority of the users - so I would probably vote for not adding it, what you have is convenient enough for the rare case we need to test other engine.

mvaligursky:
Nice progress! 
Again, random comments as I test it:

I did `npm install` and `npm build` on the engine, and `npm install` and `npm run build:thumbnails` in the examples folder and get this:

```
> examples-browser@0.0.0 build:thumbnails
> node ./scripts/thumbnails.mjs

node:fs:1394
  handleErrorFromBinding(ctx);
  ^

Error: ENOENT: no such file or directory, mkdir '/Users/mvaligursky/Snapchat/Dev/EXTERNAL/kungfoo-examples/examples/scripts/..//dist/thumbnails'
    at Object.mkdirSync (node:fs:1394:3)
    at file:///Users/mvaligursky/Snapchat/Dev/EXTERNAL/kungfoo-examples/examples/scripts/thumbnails.mjs:28:8
    at ModuleJob.run (node:internal/modules/esm/module_job:198:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:409:24)
    at async loadESM (node:internal/process/esm_loader:85:5)
    at async handleMainPromise (node:internal/modules/run_main:61:12) {
  errno: -2,
  syscall: 'mkdir',
  code: 'ENOENT',
  path: '/Users/mvaligursky/Snapchat/Dev/EXTERNAL/kungfoo-examples/examples/scripts/..//dist/thumbnails'
}
```

Then I run `ENGINE_PATH=../src/index.mjs npm run develop` and go to my local host to find examples running. Issues:
- in the CODE section, I can scroll with the mouse wheel, but I cannot click on the scroll bars and drag them around. Clicking around works, just not dragging.
- Not a new issue, but I would love to have the `Active Device` selection sticky. If I set it to WebGL and reload the page, ideally it would stick to WebGL and not WebGPU.
- It's so awesome I can change the engine code, reload and have the changes reflected!! I'm not sure what the default engine is used when `npm run develop`, but I think it should default to this, as it is the most convenient way to develop.
- if I change code in mini-stats.js (probably any extras), it does not get reflected - can this be done? This works well if I change any scripts - reload loads those changes.
- if I go to MiniStats example, which does not work on WebGPU, it still shows in the controls that it's using WebGPU, even though it uses WebGL. Can this be up to date please?

I wanted to try this: `ENGINE_PATH=../build/playcanvas.mjs/index.js npm run develop` from readme, but each example gives me this:
```
VM29605 index.js:1 Uncaught SyntaxError: Cannot use import statement outside a module (at VM29605 index.js:1:1)
VM29602 Graphics_Layers.html:131 Uncaught (in promise) ReferenceError: pc is not defined
    at main (VM29602 Graphics_Layers.html:131:29)
```
Screenshot 2023-09-14 at 10 39 33

LeXXik:
This is becoming a great PR 👍

mvaligursky:
Please ping me when this is ready for the review, and I'll keep testing, as I'm keen to get this in a final state.

kungfooman:
> Please ping me when this is ready for the review, and I'll keep testing, as I'm keen to get this in a final state.

Yep, gonna do that @mvaligursky, right now I'm trying to figure out the scrolling bug, seems to be a common issue:

https://github.com/microsoft/monaco-editor/issues/1853

Even though I only get that in "npm run serve" mode... which is strange. So I need to investigate a bit more :mag:

kungfooman:
`@todo` right now, if you don't find more issues:

> * if I go to MiniStats example, which does not work on WebGPU, it still shows in the controls that it's using WebGPU, even though it uses WebGL. Can this be up to date please?
> * Not a new issue, but I would love to have the Active Device selection sticky. If I set it to WebGL and reload the page, ideally it would stick to WebGL and not WebGPU.

Epreston aIso started the separation process of the main engine `rollup.config.js` (shaderchunks function in separate file), it would come in handy here aswell if we continue with that, because I thought about triggering a fresh build of `build/playcanvas.js` in the examples `rollup.config.js`, because it's so easy to forget... happened to me all the time and you also wrote twice about it. It would also simplify the instructions in the `README` a bit :thinking:

kungfooman:
I fixed every mentioned and unmentioned issue, but I have no time right now to go once more over every example and WebGPU still crashes on Linux so far in the engine: https://github.com/playcanvas/engine/issues/5437 (which makes my Linux testing a bit useless for this aspect).

So I would love to hear feedback from different platforms.

@mvaligursky Last issue with WebGPU in DeviceSelector can be tested here: https://examples.killtube.org/#/misc/mini-stats

image

mvaligursky:
Awesome as always, thanks, I'll test it out tomorrow!

mvaligursky:
Hi @kungfooman - another go at trying this as much as I can, and some unsorted comments as I go:

1. Could we please pick a fixed port, makes it easier to just reload the same page each time: `http://localhost:55040`
2. thumbnail generation worked without any issues.
3. switch to shader-burn example in WebGPU mode. In code editor switch to shader.frag, and change the second line to `color = vec4(0.0, 0.2, 0.0, 1.0);` - just changed the first number to zero. Click on the reload .. and the change works. Select WebGL1 and observe this. It seems changing any shader and switching platform does this.
Screenshot 2023-09-26 at 10 25 21
4. remove the debug log please. Logs it when I minimize the controls, but likely other cases.
Screenshot 2023-09-26 at 10 30 59
5. the text fields here are slightly smaller than before - numbers are cropped:
Screenshot 2023-09-26 at 10 49 54
6. `loaders-gl` example needs to be disabled for WebGPU - it's not supported and crashes. Same for `text-emojis`
7. Running it on Android, I don't get the CODE in the editor, it has `// loading` in it. Same on the iPhone.

Apart from this, all looks really good. It's so close now.

kungfooman:
Thank you for another review @mvaligursky, very glad that you pointed these issues out.

image

Changing `flex` value from 1 to 1.5 does the trick for me here (it shrinks the slider bar a little), if no one has another design preference (I didn't change any CSS, I think it came through a PCUI update).

> Could we please pick a fixed port, makes it easier to just reload the same page each time: http://localhost:55040

Hm, it defaults to 5000, picking a new port when 5000 is blocked (maybe an old zombie process running?). Shall we kill the old `npm run serve` already on 5000 when we try to start a new instance?

(I will add more commits for your other points)

mvaligursky:
> Hm, it defaults to 5000, picking a new port when 5000 is blocked

Right I see .. ok no change needed, i just had my examples serving from the main branch creating the conflict.

kungfooman:
Fixed everything once more and went over every example on Desktop to make sure it works... also tested bunch of examples on iPhone to make sure the CodeEditor issue you mentioned is fixed.

Thank you very much for your constant reviewing @mvaligursky (I know how much time it can take :sweat_smile:)

I updated the build here again to latest commit, if someone wants to test:

https://examples.killtube.org/#/graphics/shadow-cascades

examples/scripts/thumbnails.mjs
```diff
@@ -63,5 +68,26 @@ async function takeScreenshots() {
...
+    });
+}
+async function main() {
+    console.log('Spawn server on', port);
+    // We need this kind of command:
+    // npx serve dist --config ../serve.json
+    // Reason: https://github.com/vercel/serve/issues/732
+    // (who *ever* thought that stripping .html was a good idea in the first place...)
+    const server = spawn('serve', ['dist', '-l', port, '--no-request-logging', '--config', '../serve.json']);
+    await sleep(1000); // give a second to spawn server
```
mvaligursky, line 83: is there a better way to do this?

examples/src/app/helpers/example-data.mjs
```diff
@@ -1,38 +1,48 @@
...
+            realClass = realExamples[category][name + "Example"];
+        } catch (e) {
+            // slow process of porting everything to proper MJS
+            debugger;
+            return;
+        }
+        if (!realClass) {
+            return;
+        }
+        */
```
mvaligursky, line 26: remove this unless useful

examples/scripts/thumbnails.mjs
kungfooman: Ditching `serve` package for having strange/breaking peculiar default behaviour in favor of `http-server` would make the code cleaner, but I'm not sure who and for what reason picked `serve`, maybe it does something which we would miss with `http-server` again... so I just added a `serve` config to fix that behaviour so far - but I also don't like it. We could probably just use clean/normal JS API to spawn a server aswell: `import { createServer } from 'http'` (or express)

examples/src/app/helpers/example-data.mjs
kungfooman: Gonna remove it, only comes in handy a bit later for 100% build step free ES6 version

examples/src/app/helpers/strings.mjs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/**
+ * @example
+ * toKebabCase('BlendTrees1D'); // Outputs: 'blend-trees-1d'
+ * toKebabCase('LightsBakedAO'); // Outputs 'lights-baked-a-o'
+ * @param {string} str - The string.
+ * @returns String in kebab-case-format
+ */
+function toKebabCase(str) {
```
mvaligursky, line 18: I'd keep it simple, perhaps a solution 2 from here: https://www.geeksforgeeks.org/how-to-convert-a-string-into-kebab-case-using-javascript/
we don't need AO to be -a-o, it's fine to be -ao
and also 1d does not need to be -1d but can be -1-d

same goes for other functions in this file .. if there's a simpler way, I'm up for it.

examples/src/app/jsx.mjs
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+export const jsx = React.createElement;
+export const fragment = (...args) => jsx(React.Fragment, null, ...args);
+import { createRoot } from "react-dom/client";
+
+/**
+ * @param {Function} fn - todo
```
mvaligursky, line 7: this could do with a small update ;)

Also some two line description of what the file is for please.

examples/src/app/sidebar.mjs
```diff
@@ -0,0 +1,257 @@
...
+        this.mergeState({
+            collapsed: !collapsed,
+        });
+        // console.log("SideBar#toggleCollapse> was ", collapsed);
+    }
+
+    onLayoutChange() {
+        if (!this.state.filteredCategories && document.body.offsetWidth < MIN_DESKTOP_WIDTH) {
+            // @ts-ignore
+            //const sideBar = document.getElementById('sideBar');
```
mvaligursky, line 109: useful to keep?

examples/src/app/sidebar.mjs
```diff
@@ -0,0 +1,257 @@
...
+    }
+    render() {
+        //console.log("SideBar#render> state", JSON.stringify(this.state, null, 2));
+        const { observer, collapsed, orientation } = this.state;
+        const panelOptions = {
+            headerText: "EXAMPLES",
+            collapsible: true,
+            collapsed: false,
+            id: 'sideBar',
+            //class: ,
```
mvaligursky, line 226: remove?

examples/src/app/helpers/strings.mjs
mvaligursky: yep sure, just be aware that if there's a simpler way to do all this, even if the end strings are different than in current examples, feel free to use it.

examples/src/app/jsx.mjs
kungfooman: Thank you! Fixed all `- todo` and removed these two functions (I just worked things out in the beginning with them).

mvaligursky:
I reviewed the code, and apart from few minor comments it's looking great.
I'm happy to merge as soon as you fix those first two issues from my last review, the rest can be done separately, thanks!

kungfooman:
> 2\. Painter example, in WebGPU mode. Spam the reload button in the code section. It gets to an error:

Thank you for testing all these cases :+1: I could trigger WebGL2 issues too by reloading engine while still reloading. Now I drop reload-while-reloading events and I can spam reload as much as I possibly can (Shift+Enter enter enter enter...), but no more errors whatsoever any longer:

image

mvaligursky:
I loaded it from the link from your last message, in WebGPU mode. When I fast click on the reload button, I get few of these errors. When I after that tried to resize the window, I got the other errors.

Screenshot 2023-09-28 at 14 47 52

kungfooman:
Thank you @mvaligursky, this happens because the examples don't remove their dangling events via e.g. `app.on('destroy', ...)`:

```js
        window.addEventListener("resize", function () {
            app.resizeCanvas(canvas.width, canvas.height);
        });
```

Good thing is that every example already has a resize event registered, so we don't require these three lines in the examples any longer. So we have two options:

1) simply remove these three lines from every example
2) add `app.on('destroy', ...)` into examples (may have a role model value for event handling)

Kinda tending to (1) right now, you have have a opinion on this one?

mvaligursky:
I like option 1 too, but I don't like the fact we're hiding an important part of an example from the users, and so if 2 is not that complex, I'd go for it instead - perhaps if possible keep the `addEventListener` and `app.on('destroy'..` close together to make it obvious those are related.

kungfooman:
I would like to mark this ready for review once we have solved this issue: https://github.com/playcanvas/engine/issues/5729 (right now it causes errors in WebGPU mode)

(and then I can go over every example once more on Linux/Windows-tablet and see if I can find any other issues)

mvaligursky:
@5729 is solved!

kungfooman:
I went over every example on a Surface Pro 9, testing WebGPU (hot-reloading every example to make sure it works):

 - Reflection Box has no video (known WebGPU issue)
 - Particles Spark doesn't spark anything
 - Particles Snow first works, but not snowing any longer after hot-reload
 - Mesh Morph Many doesn't morph anything
 - Lights Baked AO is just gray
 - Lights Baked is just black
 - AreaPicker throws an error: `device.updateBegin() is not a function`

So I just have to mark these examples as WebGPU-disabled, otherwise I think this is nearly ready for review :see_no_evil:

mvaligursky:
Particles / morphing / light baking and area picker are not supported on WebGPU, so we need to mark those as such.
The reflection box has no video, but lets leave this as supported on WebGPU, as the main part of the example (box reflection works).

kungfooman:
Hey @mvaligursky thank you for the review again! I'm running out of time for today, but I will have a fix for that tomorrow (I imagine the solution to be quite simple, just like you say)

kungfooman:
It turned out that `<div id="app">` interfered here, resulting in two possible `app`'s:

image

image

I feel like renaming the `<div id="app">` into `<div id="appOuter">`:

```html
        <div id="app">
            <div id="appInner">
                <!--A link without href, which makes it invisible. Setting href in an example would trigger a download when clicked.-->
                <div style="width:100%; position:absolute; top:10px">
                    <div style="text-align: center;">
                        <a id="ar-link" rel="ar" download="asset.usdz">
                            <img src="./arkit.png" id="button" width="200"/>
                        </a>    
                    </div>
                </div>
                <canvas id='application-canvas'></canvas>
            </div>
        </div>
```

At least I'm more used to that `app` resolves to the pc application instead of a `div` :thinking:

kungfooman:
I can't find your post any longer, but you asked for hiding the minimap aswell:

image

Now you can press F1, enter "mini" and press "enter" and the state will be saved in `localStorage`... I mostly have F12/DevTools open too and everything looked squeezed, so this helps a bit :sweat_smile: 

Test with latest build: https://examples.killtube.org/#/graphics/particles-snow

mvaligursky:
Looks great! I got one error one time - I was reasonably fast switching examples, but cannot reproduce it. Not a blocker, this can be fixed separately.
Screenshot 2023-10-12 at 08 36 01

I'm potentially working around it here: https://github.com/playcanvas/engine/pull/5745

mvaligursky (approved):
Looks great to me, approving. Any thoughts @willeastcott @slimbuck ?

kungfooman:
> Looks great! I got one error one time - I was reasonably fast switching examples, but cannot reproduce it. Not a blocker, this can be fixed separately.

Yep, I was aware of this issue, it happens if you pick/click the same example twice: the first time the app destroys normally (I use that to stop my devices from rendering/overheating/battery-draining on an old 2013 Android device) and the second time you trigger the issue that you cannot destroy an app twice:

image

So I would be happy about a PR like you did already, just so that we can call `app.destroy()` multiple times without errors in console. Functionally this doesn't cause any error in the Examples browser, it just doesn't look nice to see errors in console :sweat_smile: 

(usually we can all `destroy` methods as often as we want, so `AppBase` behaves a bit odd I guess)

extras/package.json
```diff
@@ -0,0 +1,14 @@
+{
```
mvaligursky, line 1: question about this file - why do we need it?

kungfooman: It allows us to do `cd examples && npm link ../extras`

kungfooman:
I just realized that `iframeEval('window.ready === true')` doesn't crash but also returns the wrong value (strangely), while:

`iframeEval('ready === true')` works, but crashes when it's not defined yet (in the very beginning of loading the example)... I have to rewrite that a bit to get the right information from the iframe again without possibility of a undefined-access.

PR #5561, opened 2023-08-14
Add/use pc.call for scripts/posteffects/*.js

kungfooman:
Partly fixes issues described in https://github.com/playcanvas/engine/issues/5399

This allows us to keep using ES5 scripts, which extend ES6 classes (required for some examples in the Examples browser)

(I don't wanna add engine changes in https://github.com/playcanvas/engine/pull/5555 and it depends on this)

### DIFF
--- a/scripts/posteffects/posteffect-blend.js
+++ b/scripts/posteffects/posteffect-blend.js
@@ -10,7 +10,7 @@
  * @property {number} mixRatio The amount of blending between the input and the blendMap. Ranges from 0 to 1.
  */
 function BlendEffect(graphicsDevice) {
-    pc.PostEffect.call(this, graphicsDevice);
+    pc.call(pc.PostEffect, this, graphicsDevice);
 
     var fshader = [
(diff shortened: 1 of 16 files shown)


Maksims:
Based on this PR `pc.call` - is a public API. It is "very global" and non-descriptive out of context. If this to be used internally, please mark it `@internal`. If it is to be recommended way of doing things by developers, then API should be designed with longevity in mind.

kungfooman:
> Based on this PR `pc.call` - is a public API.

Yes, anyone who wants to continue writing ES5 code requires this function to extend PC classes "the old ES5 way".

Another possibility would be to polyfill `Function.prototype.call` to bring ES5 features back into ES6.

Or adding ES6 versions of every script, but I also don't like the code duplication...

Maybe the polyfill is the best option - we wouldn't even need to change one script :thinking:

kungfooman:
(closing because we don't really need any changes if we accept bring-back-ES5-features-into-ES6 polyfill)

PR #5563, opened 2023-08-15
NullGraphicsDevice implementation allowing to run the application without rendering

mvaligursky:
- a NullGraphicsDevice implementation that allows to run PlayCanvas engine without graphics output (server environment for example). 
- engine examples framework is updated to add NULL option as a graphics device - examples all work but output nothing.

Screenshot 2023-08-15 at 12 49 21

### DIFF
--- a/examples/src/app/example.tsx
+++ b/examples/src/app/example.tsx
@@ -3,7 +3,7 @@ import { withRouter } from 'react-router-dom';
 import { Container, Spinner, SelectInput, Panel } from '@playcanvas/pcui/react';
 import { SelectInput as SelectInputClass } from '@playcanvas/pcui';
 import { Observer } from '@playcanvas/observer';
-import { DEVICETYPE_WEBGL1, DEVICETYPE_WEBGL2, DEVICETYPE_WEBGPU } from '../../../build/playcanvas.js';
+import { DEVICETYPE_WEBGL1, DEVICETYPE_WEBGL2, DEVICETYPE_WEBGPU, DEVICETYPE_NULL } from '../../../build/playcanvas.js';
 import { File } from './helpers/types';
 import examples from './helpers/example-data.mjs';
 import ControlPanel from './control-panel';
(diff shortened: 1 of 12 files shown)


Maksims:
This is great. Will be more efficient when running it server-side using node.js.
We will be updating https://github.com/meta-space-org/playnetwork to use this feature.

examples/src/app/example.tsx
```diff
@@ -12,7 +12,8 @@ import { MIN_DESKTOP_WIDTH } from './constants';
 const deviceTypeNames = {
     [DEVICETYPE_WEBGL1]: 'WebGL 1',
     [DEVICETYPE_WEBGL2]: 'WebGL 2',
-    [DEVICETYPE_WEBGPU]: 'WebGPU'
+    [DEVICETYPE_WEBGPU]: 'WebGPU',
+    [DEVICETYPE_NULL]: 'NULL'
```
willeastcott, line 16:
```suggestion
    [DEVICETYPE_NULL]: 'Null'
```

willeastcott: Personal preference here but I think it's best not to capitalize here. When I see `NULL`, I think C++!

src/scene/graphics/scene-grab.js
```diff
@@ -54,6 +54,18 @@ class SceneGrab {
         // create depth layer
         this.layer = null;
 
+        // null device does not support scene grab
+        if (this.device.isNull) {
```
willeastcott, line 58: Instead of needing a property for `isNull`, could you just do:

```javascript
if (this.device instanceof NullGraphicsDevice) {
```

willeastcott (approved):

src/scene/graphics/scene-grab.js
mvaligursky: but that requires an import, adding a dependency - and we would not be able to tree-shake the null device.

willeastcott: OK, fair enough.

PR #5567, opened 2023-08-16
Mat3 and Mat4 updates

slimbuck:
This PR:
- updates Mat3 and Mat4 `inverse` and `transpose` functions to take an optional source parameter.
- moves `Mat4.inverseTo3x3` function to `Mat3.invertMat4` bringing it inline with the others. (This function was internal. I added docs but kept @ignore. Might be ok to expose this now @willeastcott?).
- adds tests for Mat3 and a few additional tests to Mat4.

### DIFF
--- a/src/core/math/mat3.js
+++ b/src/core/math/mat3.js
@@ -177,7 +177,7 @@ class Mat3 {
     /**
      * Generates the transpose of the specified 3x3 matrix.
      *
-     * @param {Mat3} [src] - The matrix to transpose. If not set, the current matrix is used.
+     * @param {Mat3} [src] - The matrix to transpose. If not set, the matrix is transposed in-place.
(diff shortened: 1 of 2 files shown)


src/core/math/mat3.js
```diff
@@ -177,20 +177,34 @@ class Mat3 {
     /**
      * Generates the transpose of the specified 3x3 matrix.
      *
+     * @param {Mat3} [src] - The matrix to transpose. If not set, the current matrix is used.
```
willeastcott, line 180: Do you think something like this might read better:

```
The matrix to transpose. If not set, the matrix is transposed in-place.
```

slimbuck: Yep that sounds better. I've updated all instances thanks!

PR #5569, opened 2023-08-16
WorldClusters allocation is now dynamic, removed from composition update

mvaligursky:
Refactoring to limit the amount of LayerComposition updates.

Before: When a light was added / removed from the layer, the expensive composition update would execute. Part of the update was to assign world clusters (textures storing clustered lights) to individual layers based on their lights.

Now: The world clusters assignment takes place at the start of the frame. The layers lazily update their light hash values and world clusters are shared between layers based on this hash.

The is still a cost of computing a hash for a layer on frames lights are added / removed on it - but this is reasonably cheap.
Note that this is needed specifically for WebGL1 path, on other platforms we could use bitflag to encode layers and reject lights from a single clusters structure at runtime. One day.

Note that the composition still updates when lights are added / removed, and further refactoring is needed, this PR only addresses the clustered part of it.

### DIFF
--- a/src/scene/composition/layer-composition.js
+++ b/src/scene/composition/layer-composition.js
@@ -2,7 +2,6 @@ import { TRACEID_RENDER_ACTION } from '../../core/constants.js';
 import { Debug } from '../../core/debug.js';
 import { Tracing } from '../../core/tracing.js';
 import { EventHandler } from '../../core/event-handler.js';
-import { set } from '../../core/set-utils.js';
 import { sortPriority } from '../../core/sort.js';
 
 import {
@@ -12,10 +11,8 @@ import {
 } from '../constants.js';
 
 import { RenderAction } from './render-action.js';
-import { WorldClusters } from '../lighting/world-clusters.js';
 
 const tempSet = new Set();
-const tempClusterArray = [];
 
 /**
  * Layer Composition is a collection of {@link Layer} that is fed to {@link Scene#layers} to define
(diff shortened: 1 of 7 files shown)


src/scene/composition/layer-composition.js
```diff
@@ -579,8 +445,8 @@ class LayerComposition extends EventHandler {
                     ' Meshes: ', ('?' + layer.meshInstances.length).padStart(5) +
                     (' RT: ' + (ra.renderTarget ? ra.renderTarget.name : '-')).padEnd(30, ' ') +
                     ' Clear: ' + clear +
-                    ' Lights: (' + layer._clusteredLightsSet.size + '/' + layer._lightsSet.size + ')' +
-                    ' ' + (ra.lightClusters !== this._emptyWorldClusters ? (ra.lightClusters.name) : '').padEnd(10, ' ') +
+                    // ' Lights: (' + layer._clusteredLightsSet.size + '/' + layer._lightsSet.size + ')' +
+                    // ' ' + (ra.lightClusters !== this._emptyWorldClusters ? (ra.lightClusters.name) : '').padEnd(10, ' ') +
```
willeastcott, line 449: Maybe just delete?

src/scene/layer.js
```diff
@@ -665,6 +679,12 @@ class Layer {
         }
     }
 
+    markLightDirty() {
```
willeastcott, line 682: Should this be `markLightsDirty`?

src/scene/composition/layer-composition.js
mvaligursky: yep, done, cheers

src/scene/layer.js
mvaligursky: yep, done

willeastcott (approved):

Issue #5554, opened 2023-08-07
Feature Request - Displacement Map


defektu:
Displacement map vertex shader would be a cool feature to implement.

Managed to prepare a project with displacement effect by replacing transformVS. 

Here is the [Project Link](https://playcanvas.com/project/1123367) (Working properly with lights and shadows)

### Screenshots
Displacement Effect in use:
![image](https://github.com/playcanvas/engine/assets/33038942/fa511fdb-f5e8-4ff5-8d1d-2f4cdf1a69a7)

# My try on trying to create it in engine: (can provide a PR but it is not working as intended)
- Tried to implement the effect with `if(options.useDisplacement) code+= #define DISPLACEMENT` in shader definition. 
- Thus, in vert/transform.js `#ifdef DISPLACEMENT` seems not picking while rendering shadows. 

### Screenshots
![image](https://github.com/playcanvas/engine/assets/33038942/2add5ed3-bf93-43c6-bd57-b86edbbc23b2)

## Issues while engine implementation :
- Displacement state needs to be stated in transformVS. (`matrix_normal * sample_texture` will throw error otherwise because of undefined texture)
- This leads to some issues like lights not getting definitions from lit options. Which is causing lighting artifacts.
- Another approach might be `code += displacementVS`.

# Recommendations:
- Soft limiting in UI displacementMapFactor within -2, 2 range would be good.
- Offsetting displacementMapFactorOffset would be good implementation for manually defining boundaries

mvaligursky:
Hey ... that's pretty exciting. Feel free to create a PR and we can discuss it there perhaps?

defektu:
Created a draft pull request #5573

mvaligursky:
Superseded by #9246, which carries the current design for this: vertex displacement as a separate property composable with the 2.22 parallax stack, sharing the height map and its zero level. Much of the discussion here predates the 2.22 parallax rework (#9207/#9213/#9225/#9245), so filing fresh keeps the proposal readable. Thanks @defektu for the original prototype - the shadow-pass problem it uncovered is called out there as the main piece of engine work.

PR #5573, opened 2023-08-17
Displacement Effect

defektu:
Fixes https://github.com/playcanvas/engine/issues/5554

@mvaligursky requested PR from #5554 to look into it.

Known issues with PR is in issues thread. 

It might be a poor implementation in terms of engine's workflow.

`LitShader-forward-proc` & `LitShader-ShadowPass_1_0-proc` does not get `#define DISPLACEMENT` argument possibly because of my poor implementation.

Project file in #5554 works perfectly (probably due to shader chunk rebuilds when replacing a chunk)

Did not implemented checks for implementation cause of draft PR


mvaligursky:
Would you have any way to use this? Engine example or Editor project?

defektu:
Scene for changing [transformVS ](https://playcanvas.com/editor/scene/1830188)
and [engine example ](https://playcanvas.com/editor/scene/1836257)

mvaligursky:
I had a quick look, and I think:

1. you should use heightMap for this instead of introducing new texture channel. The `useDisplacement` would then use this map as displacement instead of for parallax.
2. I think you need something like this to have normals attribute added to the vertex shader:
Screenshot 2023-08-18 at 11 34 13

Give these a go please and let me know if there are further issues - there might be with supplying UVs to the vertex shader too, but maybe not with the heightMap.

defektu:
Today I noticed standart.js has `all other passes require only opacity` part. So `ShadowPass` is not getting any info about anything except opacity.

mvaligursky:
Yep, we'll need to adjust this to pass on the height if used as well.

defektu:
Also `vertex_normals` needed. But it does not pass into shadow pass as semantics as well `Vertex shader attribute "vertex_normal" is not mapped to a semantic in shader definition, shader [Shader Id 2 LitShader-ShadowPass_1_0-proc] `

defektu:
Shaderpass does not get any litOptions (turns out to be undefined when set) 
`getShaderVariant` passes `this.shaderOptBuilder.updateMinRef` if it is not `FORWARD_PASS`. So do you have any suggestions about this topic. Besides shadowPass it is working as intended (displacing the surface).
image

mvaligursky:
I suspect the StandardMaterialOptionsBuilder will need to be updated to generate required options in the shadow pass - probably in _updateSharedOptions, which I think is shared between forward and shadow pass, but not sure now based on you seeing litOptions being undefined. 

Perhaps update the PR with what you have, using the height texture, and I'll see how I can fix the shadows using your example.

defektu:
Rolled back previous changes and added `mesh-displacement` example draft.

defektu:
Somehow I got it working with shared options. I will cleanup everything and send PR.

defektu:
Pushed working version with `minimalOptions = false`. It is working as intended like this.
image

defektu:
Effect fully works for now but noticed env goes blank (weird frustrum culling behaviour causes that)
Can't state if it is caused by PR or something else.

mvaligursky:
I fixed the culling that few days ago .. you're just probably not up to date.
Is it all ready for the review you reckon? Ping me when it is, I'll check it out, thanks!

defektu:
Yep it is fully working PR right now.

defektu:
Yes it is up and running. Uv transformations are not implemented. `displacementOffset` is nice to have and can be implemented by introducing new material attribute.

defektu:
Pushed new commit for `displacementOffset` too.

defektu:
Hi @mvaligursky, It's all ready to review.

Rebased branch with latest updates I accidentally force repushed everything again sorry about that. 
Culling is fixed with updates from origin. Also noticed that `heightMap` still affects when displacement is used, this causes some kind of weird artifacts with big displacements. 

I am suggesting seperating `heightMapFactor` and adding `displacementFactor`.
This will gives possibility to remove `useDisplacement` boolean. We can disable effect if `displacementFactor == 0`.
That might be better for UI cleanness with single slider to do everyting at once.

mvaligursky:
> I am suggesting seperating `heightMapFactor` and adding `displacementScale`.

Great idea! Perhaps we should call it just `displacement`, default it to 0?

examples/src/examples/graphics/mesh-displacement.tsx
```diff
@@ -0,0 +1,301 @@
...
+                    <BooleanInput id='displacment-effect' binding={new BindingTwoWay()} link={{ observer: data, path: 'script.useDisplacement.enabled' }} />
+                </LabelGroup>
+                <LabelGroup text='Height Map Factor'>
+                    <SliderInput binding={new BindingTwoWay()} link={{ observer: data, path: 'script.useDisplacement.heightMapFactor' }} min={-2.0} max={2.0}/>
+                </LabelGroup>
+                <LabelGroup text='Displacement Offset'>
+                    <SliderInput binding={new BindingTwoWay()} link={{ observer: data, path: 'script.useDisplacement.displacementOffset' }} min={-2.0} max={2.0}/>
+                </LabelGroup>
+            </Panel>
+            <Panel headerText='Point light'>
```
mvaligursky, line 24: I have not tried running it (yet), but it seems the point light control is not related to the displacement - ideally please strip out unrelated things to keep the example well focused.

defektu:
There is `specularityFactor`. Since effect uses `heightMap`, `displacementFactor` and `displacementOffset` is good?

src/scene/materials/standard-material.js
```diff
@@ -800,6 +800,11 @@ class StandardMaterial extends Material {
             this._setParameter('material_heightMapFactor', getUniform('heightMapFactor'));
         }
 
+        if (this.useDisplacement) {
+            this._setParameter('material_displacementOffset', this.displacementOffset);
+            console.log(this.displacementOffset)
```
mvaligursky: remove log when done pls

src/scene/shader-lib/chunks/common/vert/transform.js
```diff
@@ -59,13 +59,14 @@ mat4 getModelMatrix() {
...
 uniform float material_heightMapFactor;
+uniform float material_displacementOffset;
 #endif
 vec4 getPosition() {
     dModelMatrix = getModelMatrix();
 
     #ifdef DISPLACEMENT
     vec4 dMapSample = texture2D(texture_heightMap, vertex_texCoord0.xy);
-    float displacer = dMapSample.r * material_heightMapFactor * 10.;
+    float displacer = dMapSample.r * material_heightMapFactor * 10. + material_displacementOffset;
```
mvaligursky, line 69: I'm not sure about the `*10 ` part here?

defektu: That can be removed when `material_heightMapFactor` is replaced with new attribute.

mvaligursky:
> There is `specularityFactor`. Since effect uses `heightMap`, `displacementFactor` and `displacementOffset` is good?

yeah actually, I missed you added an offset too. Sounds good in that case to me.

src/scene/shader-lib/chunks/common/vert/transform.js
slimbuck: Wouldn't it be more useful for vertex displacement to be implemented as an overridable shader chunk instead?

defektu:
Hey @mvaligursky, Updated PR & ready for review. 

Decided to keep `useDisplacement` attribute for code readability. It sets based on `displacementFactor`. Cleaned up mesh-displacement example.

I guess it is good to keep `useDisplacement` attribute in material parameters for debugging?

Not a code dev mainly so ignore commit chains <3

examples/src/examples/graphics/mesh-displacement.tsx
```diff
@@ -0,0 +1,225 @@
+import React from 'react';
+import * as pc from '../../../../';
+
+import { BindingTwoWay, LabelGroup, Panel, SliderInput, SelectInput, BooleanInput } from '@playcanvas/pcui/react';
+import { Observer } from '@playcanvas/observer';
+class MeshDisplacementExample {
+    static CATEGORY = 'Graphics';
+    static NAME = 'Mesh Displacement';
+    static WEBGPU_ENABLED = false;
```
mvaligursky, line 9: just curious - does it not work on WebGPU, or have you just not tried?

examples/src/examples/graphics/mesh-displacement.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+    example(canvas: HTMLCanvasElement, deviceType: string, data: any): void {
+		
+        const assets = {
+            orbitCamera: new pc.Asset('script', 'script', { url: '/static/scripts/camera/orbit-camera.js' }),
+            helipad: new pc.Asset('helipad-env-atlas', 'texture', { url: '/static/assets/cubemaps/helipad-env-atlas.png' }, { type: pc.TEXTURETYPE_RGBP, mipmaps: false }),
+            normal: new pc.Asset("normal", "texture", { url: "/static/assets/textures/seaside-rocks01-normal.jpg" }),
+            diffuse: new pc.Asset("diffuse", "texture", { url: "/static/assets/textures/seaside-rocks01-color.jpg" }),
+            other: new pc.Asset("other", "texture", { url: "/static/assets/textures/seaside-rocks01-gloss.jpg" }),
+            height: new pc.Asset("other", "texture", { url: "/static/assets/textures/seaside-rocks01-gloss.jpg" })
```
mvaligursky, line 32: no need to load the same texture twice

examples/src/examples/graphics/mesh-displacement.tsx
```diff
@@ -0,0 +1,225 @@
...
+                app.start();
+
+                app.scene.toneMapping = pc.TONEMAP_ACES;
+                app.scene.skyboxMip = 1;
+                app.scene.ambientLight.set(0, 0, 0);
+                app.scene.ambientLuminance = 0;
+                app.scene.setSkybox(assets.helipad.resources);
+
+                // enable area lights which are disabled by default for clustered lighting
+                app.scene.clusteredLightingEnabled = false;
```
mvaligursky, line 89: comment says enable, code says disable

examples/src/examples/graphics/mesh-displacement.tsx
```diff
@@ -0,0 +1,225 @@
...
+                material.bumpiness = 0.7;
+                material.useMetalness = true;
+                material.heightMap = assets.height.resource;
+                material.displacementFactor = data.get('script.useDisplacement.displacementFactor');
+                material.displacementOffset = data.get('script.useDisplacement.displacementOffset');
+                material.update();
+
+                const sphere = createSphere(0, 2.5, 0, material, true);
+
+                app.scene.envAtlas = assets.helipad.resource;
```
mvaligursky, line 156: move this line up to where other app.scene.* are set up

examples/src/examples/graphics/mesh-displacement.tsx
mvaligursky: ah I guess not due to the use of SHADOW_PCSS

src/scene/materials/standard-material-options-builder.js
```diff
@@ -287,6 +288,10 @@ class StandardMaterialOptionsBuilder {
         options.litOptions.useIridescence = stdMat.useIridescence && stdMat.iridescence !== 0.0;
         options.litOptions.useMetalness = stdMat.useMetalness;
         options.litOptions.useDynamicRefraction = stdMat.useDynamicRefraction;
+
+        options.litOptions.useDisplacement = stdMat.useDisplacement;
```
mvaligursky, line 292: I suspect only useDisplacement needs to be in the options, as likely this is the only flag used by the generator. Not the values - those are just uniforms, and do not change the generated shader.

src/scene/materials/standard-material.js
```diff
@@ -816,6 +816,13 @@ class StandardMaterial extends Material {
             this._setParameter('material_heightMapFactor', getUniform('heightMapFactor'));
         }
 
+        this.useDisplacement = this.displacementFactor !== 0;
```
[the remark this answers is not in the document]
mvaligursky: Actually. I think we should leave the flag on the material, but let user set it.
The reason for that is that if the value is set to 0, we do not want to have to compile a new shader because the flag has changed - it should use the same shader with zero displacement. Shader should only be regenerated when the user actually changes the flag on the material.

src/scene/materials/standard-material.js
```diff
@@ -865,7 +872,8 @@ class StandardMaterial extends Material {
 
         // Minimal options for Depth and Shadow passes
         const shaderPassInfo = ShaderPass.get(device).getByIndex(pass);
-        const minimalOptions = pass === SHADER_DEPTH || pass === SHADER_PICK || shaderPassInfo.isShadow;
+        const minimalOptions = this._useDisplacement ? false : pass === SHADER_DEPTH || pass === SHADER_PICK || shaderPassInfo.isShadow;
```
mvaligursky, line 875: this seems to be the only use of `_useDisplacement` - should the `_` be there? If it works without it, do we need this change at all? (maybe I'm just missing something)

src/scene/shader-lib/chunks/common/vert/transform.js
```diff
@@ -56,10 +56,21 @@ mat4 getModelMatrix() {
...
 }
-
+#ifdef DISPLACEMENT
+uniform sampler2D texture_heightMap;
+uniform float material_displacementFactor;
+uniform float material_displacementOffset;
+#endif
 vec4 getPosition() {
     dModelMatrix = getModelMatrix();
+
```
mvaligursky, line 66: I'd leave this original line on top:
```
```

and then add the displacement this way
```
#ifdef DISPLACEMENT
...
localPos == vertex_normal * displacer;
#endif
```

examples/src/examples/graphics/mesh-displacement.tsx
defektu: > just curious - does it not work on WebGPU, or have you just not tried?

Tried it without SHADOW_PCSS too but to many errors came up. So I thought it might be something else.

src/scene/materials/standard-material.js
defektu: Was thinking about changing it to private at first, then drifted. `_` gonna be removed.

src/scene/materials/standard-material-options-builder.js
defektu: `Lit-shader.js` uses this. Won't generate if does not set

`this.needsNormal ||= options.useDisplacement;` and       
 ```
if (options.useDisplacement) {
           code += "#define DISPLACEMENT\n";
}
```

File: src/scene/materials/standard-material.js (hunk 9 above)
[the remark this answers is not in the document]
Comment (defektu) [reply src/scene/materials/standard-material.js:819]: Not much to change here then, just removing the line.

File: examples/src/examples/graphics/mesh-displacement.tsx (hunk 4 above)
Comment (mvaligursky) [reply examples/src/examples/graphics/mesh-displacement.tsx:9]: yep fair enough, I'll check it later

File: src/scene/materials/standard-material-options-builder.js (hunk 8 above)
Comment (mvaligursky) [reply src/scene/materials/standard-material-options-builder.js:292]: yes, that's what I said too. useDisplacement needs to be there, but not the other two values (following two lines)

File: src/scene/materials/standard-material-options-builder.js [hunk 12]
```diff
@@ -287,6 +288,10 @@ class StandardMaterialOptionsBuilder {
...
+        options.litOptions.displacementFactor = stdMat.displacementFactor;
+        options.litOptions.displacementOffset = stdMat.displacementOffset;
```
mvaligursky, line 294: these two lines should not be here, we don't need a new shader each time a value changes, which this would cause as options are used to generate hash

defektu:
Added draft docs for attributes.

Things to consider:
- This effect uses UV0 and R channel of height map.
- You can't offset or transform map yet.
- If there is no heightMap, `useDisplacement` will break vertex chunk.

src/scene/shader-lib/chunks/common/vert/transform.js
defektu: I am not sure, you are the guys to call the shot.

willeastcott:
Hey @mvaligursky and @slimbuck - can we sync on this PR? I'm keen to make a decision on what to do with it.

Maksims:
Is there a way to change heightMap channel for displacement?
Also will it apply tiling and offsets?

Another question, will it be viable/common when devs use displacement in combination with parallax? In such case the heightMap textures might be different or the same?

defektu:
Hi @Maksims, I think this vertex shader needs some special attributes for offsetting and tiling. Concurrent options are bound to fragment shader. Don't think that back propagation is not possible unless changing where uniforms defined. 

Not really experienced with engine, so this needs crosscheck.

defektu:
Back from service. Ready to do changes if needed.

mvaligursky:
This should be a lot easier (very easy) to do using a ShaderMaterial for custom shaders - which would support skinned and instanced meshes too, see more info here: https://github.com/playcanvas/engine/issues/6835

Path to add it to the StandardMaterial is similar as before, but little simplified. But to add it to it I think we should ideally wait for https://github.com/KhronosGroup/glTF/issues/948 to shake out (could be a while!), to add exactly what is needed.

If you're keen to try and write an engine example using ShaderMaterial, that'd be a welcome addition. As that would likely be a separate PR, I'll close this one, but very much appreciate your time and efforts here @defektu .

PR #5575, opened 2023-08-18
Vec and Quat source args

slimbuck:
Follow up PR to #5567 to allow specifying source on Vec2, Vec3, Vec4 and Quat operations.

### DIFF
--- a/src/core/math/quat.js
+++ b/src/core/math/quat.js
@@ -74,10 +74,11 @@ class Quat {
         return new cstr(this.x, this.y, this.z, this.w);
     }
 
-    conjugate() {
-        this.x *= -1;
-        this.y *= -1;
-        this.z *= -1;
+    conjugate(src = this) {
+        this.x = src.x * -1;
+        this.y = src.y * -1;
+        this.z = src.z * -1;
+        this.w = src.w;
 
         return this;
     }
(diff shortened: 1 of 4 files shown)


willeastcott:
Anything to do for `Color`?

kungfooman:
Is the main goal to reduce amount of code we have to write or make things faster (or both)?

Like usually we would do: `mat.copy(otherMat).invert()`
And now we just do: `mat.invert(otherMat)`?

Is this making calculations faster, because we can get rid of .copy (benchmark would be interesting) or what is the philosophy behind these changes?

slimbuck:
These changes bring a little consistency with existing functions like lerp, add, cross, etc which already fill `this`. In some cases as you say, it'll negate the need for an additional copy and also make calling code a little more succinct. I see this change as adding some expressive power to the math API without breaking backwards compatibility.

And for me personally, I like the mental model that `this` is the target storage for math operations.

slimbuck:
> Anything to do for `Color`?

Don't see anything.

willeastcott (approved):

LeXXik:
Those look like static methods to me. I feel like if I'd want to change `this` storage I should be doing something like `pc.Mat4.invert(other)`, rather than using one random instance to change another.

PR #5582, opened 2023-08-23
Refactor shader caching to handle mesh instance on layers with different lights

mvaligursky:
Cache of shaders on the mesh instance only supported a single set of lights, by storing _lightHash value. This does not work when the same mesh instance is used on multiple layers, with different sets of lights, causing shader cache miss each time.

This has been refactored now to be supported. Further refactoring has been done to storing bind groups that match the shaders.

Some deprecated functionality, in the past used by the Editor, has been removed now as well, as it's being removed from the Editor.

### DIFF
--- a/src/deprecated/deprecated.js
+++ b/src/deprecated/deprecated.js
@@ -868,11 +868,11 @@ ForwardRenderer.prototype.renderComposition = function (comp) {
     getApplication().renderComposition(comp);
 };
 
-ForwardRenderer.prototype.updateShader = function (meshInstance, objDefs, unused, pass, sortedLights) {
-    Debug.deprecated('pc.ForwardRenderer#updateShader is deprecated, use pc.MeshInstance#updatePassShader.');
-    const scene = meshInstance.material._scene || getApplication().scene;
-    return meshInstance.updatePassShader(scene, pass, sortedLights);
-};
+// ForwardRenderer.prototype.updateShader = function (meshInstance, objDefs, unused, pass, sortedLights) {
+//     Debug.deprecated('pc.ForwardRenderer#updateShader is deprecated, use pc.MeshInstance#updatePassShader.');
+//     const scene = meshInstance.material._scene || getApplication().scene;
+//     return meshInstance.updatePassShader(scene, pass, sortedLights);
+// };
 
 MeshInstance.prototype.syncAabb = function () {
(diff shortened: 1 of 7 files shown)


src/scene/materials/material.js
```diff
@@ -79,7 +79,14 @@ class Material {
 
     id = id++;
 
-    variants = {};
+    /**
+     * The cache of shader variants generated for this material. They key represents the unique
```
willeastcott, line 83:
```suggestion
     * The cache of shader variants generated for this material. The key represents the unique
```

src/deprecated/deprecated.js
```diff
@@ -868,11 +868,11 @@ ForwardRenderer.prototype.renderComposition = function (comp) {
...
-ForwardRenderer.prototype.updateShader = function (meshInstance, objDefs, unused, pass, sortedLights) {
-    Debug.deprecated('pc.ForwardRenderer#updateShader is deprecated, use pc.MeshInstance#updatePassShader.');
-    const scene = meshInstance.material._scene || getApplication().scene;
-    return meshInstance.updatePassShader(scene, pass, sortedLights);
-};
+// ForwardRenderer.prototype.updateShader = function (meshInstance, objDefs, unused, pass, sortedLights) {
+//     Debug.deprecated('pc.ForwardRenderer#updateShader is deprecated, use pc.MeshInstance#updatePassShader.');
+//     const scene = meshInstance.material._scene || getApplication().scene;
+//     return meshInstance.updatePassShader(scene, pass, sortedLights);
+// };
```
willeastcott, line 875: Delete?

willeastcott (approved):

src/deprecated/deprecated.js
mvaligursky: removed, thanks

src/scene/mesh-instance.js
```diff
@@ -59,6 +59,85 @@ class Command {
     }
 }
 
+/**
+ * Internal helper class for storing the shader and related mesh bind group in the shader cache.
+ *
+ * @ignore
+ */
+class ShaderInstance {
```
willeastcott, line 67: Should this go in its own module?

mvaligursky: I decided against it for now, as it's just internal to mesh instance, not even exported.

Issue #5541, opened 2023-08-03
Improve the building of /extras


mvaligursky:
Currently, we add engine non-core functionality to `/extras` folder, and the build system produces `playcanvas-extras.js`. This was a good initial implementation, but it is not flexible / convenient enough, as we're hoping to add more non-core functionality to extras. The possible solution here would be this, but we're open to discussion of better options as well:

- move `/extras` folder to the engine source instead of being outside, so that the new folder would be `/src/extras`
- currently the engine build is based around `/src/index.js' - both ES5 and module builds use this
- we could add additional file, perhaps called`src/index.mjs' that would be import the index.js but additionally it would import the modules from `src/extras`.
- the module build of the engine would use this `index.mjs`, and the module build of the engine would contain extras modules
- the ES5 build of the engine would use existing `index.js` and would not contain extras
- for ES5 use (in the Editor primarily), we'd build extras into individual functional modules (per subfolder in extras), so instead of playcanvas-extras.js, we'd have playcanvas-mini-stats.js and playcanvas-exporters.js as an example, allowing the user to only include what they need.

Would import maps be useful here?

epreston:
Initial ideas:

In a world of treeshaking, having a separate "extras" only makes sense for commonjs formatted modules.  Past that, every folder, file, class, declaration is an "extra" that will be discarded / ignored / not downloaded if not referenced.

The "what doe we do with extras" is only an issue for "commonjs" and below.

mvaligursky:
> The "what do we do with extras" is only an issue for "commonjs" and below.

Yep, fully agreed, that's the design is trying to cover.

epreston:
If I was setting a goal for this in todays world, I would structure it for consistency / readability / ease of use in the ESM world.

Then I would use whatever clever tricks are necessary to extract commonjs build products.  Yep, import maps where its graceful in html examples that run off bundles from a CDN if needed.

things that I can see will help:

Add an exports section to the existing `package.json`  (example only)
```json
  "exports": {
    ".": {
      "import": "./build/engine.module.js",
      "require": "./build/engine.cjs"
    },
    "./examples/*": "./examples/*",
    "./src/*": "./src/*",
  },
```

A browserlist definition instead of importing out of date polyfils in `package.json`
Have rollup select and introduce the polyfils based on language and browser support targets.

```json
  "browserslist": [
    "> 1%, not dead, not ie 11, not op_mini all"
  ],
```

mvaligursky:
[...]

PR #5589, opened 2023-08-24
chore: move extras source into src

epreston:
This moves the "extras" source into "src" while ensuring all build tools produce the same outputs without warning.  It also updates configuration files to ensure that autocomplete, type compilation, and module resolution function correctly with small improvements to the developer experience.

Supports: #5541

Details:

move extras into source, step one of the plan in #5541 
create a single autocomplete namespace for tools

Build:
- update build commands
- update build config paths

Jsdoc: 
- add autocomplete and validation to jsdoc config
- exclude extras from source config to ensure jsdocs are not changed

tsconfig:
- add paths so autocomplete / type compiler can resolve imports
- add second path alias so imports don't need to know where "extras" is

things can import from 'playcanvas' as if they were using the npm bundle but it is aliased to the current source during development.  saves a rebuild, reduces the risk of getting de-synchronised.
inversely, things can reference 'playcanvas-extras' for imports but get aliased to the extras source during dev.

avoid issues with cjs mjs js modes with lint config


epreston:
This is only supposed to be the file move, nothing else.  Sorted now.

PR #5591, opened 2023-08-24
chore: update the module builds to contain all the modules

epreston:
This updates the ESM builds to contain all the modules.  

> The final bundler in a project will keep code based on usage and "treeshake" the rest.  Having developers include parts of the library separately creates friction for no benefit.

Status:

- [x] module builds additionally import the modules from src/extras`.
- [x] module build of the engine would contain extras modules
- [x] ES5 build of the engine would not contain extras
- [x] external `fflate` for module builds <-- confirm this

Supports: #5541 

This builds on the previous PR, #5589, to change the build products as described above.


epreston:
Issue module output.

kungfooman:
Having multiple roots in `/src/` feels more confusing to me than not and that some PC builds seem to come with bundled `fflate` and some others not just makes it even more confusing to me.

And I'm not sure why in the first place... if `/extras/` shall be a module, just `cd examples && npm init` and then you can `npm link` that in e.g. `/examples/` via `cd examples && npm link ../extras` and use it via `import { Whatever } from 'playcanvas-extras'`.

If "monorepo" means that we have to overcomplicate the normal engine rollup script with every future "extras" issue (like `fflate` and everything else we are going to add in "extras"), I'm not thrilled about this - effectively turning the engine build script into a nanny that tries to build everything but easily breaks if you don't follow every edge case logic for every single special need there is.

I think it would be better to aim for a simpler core engine rollup script and one for `/extras/`. That way, each developer could work within their own package and fine-tune their own build script. This would allow for simpler maintenance, less merge conflicts and fewer complications overall - and if there is an complication, you can track them down quicker.

If monorepo means to have more nice/useful scripts in this repo (e.g. examples  / extras with separate build scripts to reduce complexity), I look forward to changes.

In short, I would aim for nuking every occurrence of "extras" in `rollup.config.js` and move it over into a `extras/rollup.config.js`, which then represents a distinct and easy-to-judge-about package with a dedicated README.md to inform the devs.

epreston:
@kungfooman yep, that's what its looking like. This might just be an exploration to tease out issues and good ideas.

epreston:
Because I built this PR on another PR that hasn't been merged, its made it look very complicated. 

This PR should only be 10 lines in 3 files.  Short and clear.  

Instead this has all the changes...  The CI was failing because it depends on the other PR.  I rebased and made a mess.

epreston:
I'm going to redo this.  

It was a valuable exercise but this isn't workable.

PR #5593, opened 2023-08-25
Refactored shadow casting directional lights collection

mvaligursky:
Those lights used to be collected during composition update, but now are obtained from layers at runtime, to avoid composition update when any lights / light properties change

### DIFF
--- a/src/scene/layer.js
+++ b/src/scene/layer.js
@@ -2,7 +2,7 @@ import { Debug } from '../core/debug.js';
 import { hash32Fnv1a } from '../core/hash.js';
 
 import {
-    LIGHTTYPE_DIRECTIONAL, LIGHTTYPE_OMNI, LIGHTTYPE_SPOT,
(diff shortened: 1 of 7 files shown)


src/scene/layer.js
```diff
@@ -734,6 +747,32 @@ class Layer {
...
+            this._splitLightsDirty = false;
+
+            const splitLights = this._splitLights;
+            for (let i = 0; i < splitLights.length; i++)
+                splitLights[i].length = 0;
+
+            const lights = this._lights;
+            for (let i = 0; i < lights.length; i++) {
+                const light = lights[i];
+                if (light.enabled) {
```
slimbuck, line 762: is splitLights updated when lights are enabled and disabled?

mvaligursky: yes, as light is removed from the layer at that point which sets markLightsDirty to true

slimbuck (approved):

PR #5604, opened 2023-08-31
Layer composition no longer handles lights

mvaligursky:


src/scene/renderer/renderer.js
```diff
@@ -872,6 +888,63 @@ class Renderer {
...
+    collectLights(comp) {
+
+        // build a list and of all unique lights from all layers
+        this.lights.length = 0;
+        this.localLights.length = 0;
+
+        // stats
+        const stats = this.scene._stats;
+        stats.dynamicLights = 0;
+        stats.bakedLights = 0;
```
slimbuck, line 900: Does scene._stats exist in non-profile build? Might be worth putting this in #if _PROFILER.

mvaligursky: It exists, I checked.

slimbuck: And is there a reason not to remove this profile code from non-profile builds?

mvaligursky: done, but it's not really worth it I think, adding ifdefs for two trivial lines.

slimbuck: you don't think it's useful to the people reading this source code and people reading the compiled release build?

for example the line further down ```stats.lights = this.lights.length;``` - is this also profile build detail or no?

mvaligursky: Number of lights stat we have available, so I output that in all builds. This is similar to many others stats we output in all builds if trivially available. If there's a cost to compute those, then profile build only.

slimbuck: This is crazy approach imo. How will anyone other than you know what's going on? There are stats declared, but they're not updated, or maybe they are. How can we trust them? The stats we're talking about here - are they profile build only or not? How do we use the values in a release build? How is someone maintaining the codebase meant to keep track of any of this if declaration and updating is done ad-hoc?

It would be much better to explicitly make every stat either profile-only or not. If they're general stats then they should available somewhere, otherwise what's the point?

mvaligursky: It's an existing solution, where stats exist in every engine, but most of them are not updated unless in the profile engine. Something I was eying on refactoring for sure, as it's a pretty messy system at the moment.

Issue #3485, opened 2021-09-15
Custom attributes on material's shader definitions.


Maksims:
When constructing a `Mesh` it is possible to add custom attributes using `setVertexStream` and `SEMANTIC_ATTR*` constant.
But when the shader is created for StandardMaterial, unfortunately it is not adding relevant varying semantic to `definition.attributes` object, so no attributes will be passed.

The workaround is to patch `standard` generator `createShaderDefinition`, and force add a relevant attribute to the definition, but it is required to check if vertex shader code contains a particular substring in order to identify the right shader for the custom attribute.

It would be great to have a generic way for Material to define relevant semantic to attribute mapping functionality.
Something like:

```js
material.defineAttribute('vertex_custom_attribute', pc.SEMANTIC_ATTR15);
```

So shader for it would include relevant attributes in definitions.

mvaligursky:
It should likely also allow to set semantic used by the instancing, which is currently hardcoded. See the change here which modified it .. this should be settable by the user instead of hardcoded in the engine https://github.com/playcanvas/engine/commit/dc126317ea0d67d9542a00c9bb44481c3a64fd10#diff-9ca32de26f8325e11d4a0936116c4a9a8c8bbcce24630887f90f72070dfb64fc

LeXXik:
Bumping up. Would be great to have.

PR #5607, opened 2023-09-03
Add setAttribute to standard material

LeXXik:
Fixes https://github.com/playcanvas/engine/issues/3485

Adds a `setAttribute` to StandardMaterial.

Example:
```js
const transformVS = `
attribute vec3 offset;

mat4 getModelMatrix() {
    return matrix_model;
}

vec4 getPosition() {
    localPos.y += offset.y;
    vec4 posW = dModelMatrix * vec4(localPos, 1.0);
    dPositionW = posW.xyz;
    return matrix_viewProjection * posW;
}

vec3 getWorldPosition() {
    return dPositionW;
}
`;

const positions = [
    0, 0, 0,
    1, 0, 0,
    1, 0, 1,
    0, 0, 1
];
const indices = [ 0, 2, 1, 0, 3, 2 ];

const offset = [];
for (let i = 0; i < 12; i++) {
    offset.push(2);
}

const mesh = new pc.Mesh(graphicsDevice);
mesh.setPositions(positions);
mesh.setIndices(indices);
mesh.setNormals(pc.calculateNormals(positions, indices));
mesh.setVertexStream(pc.SEMANTIC_ATTR15, offset, 3);
mesh.update();

const material = new pc.StandardMaterial();
  
material.setAttribute('offset', pc.SEMANTIC_ATTR15);
material.chunks.transformVS = transformVS;
material.chunks.APIVersion = pc.CHUNKAPI_1_65;

const meshInstance = new pc.MeshInstance(mesh, material);

entity.addComponent('render', {
    meshInstances: [ meshInstance ],
});
```

### DIFF
--- a/src/platform/graphics/shader-processor-options.js
+++ b/src/platform/graphics/shader-processor-options.js
@@ -15,9 +15,6 @@ class ShaderProcessorOptions {
     /** @type {import('./vertex-format.js').VertexFormat[]} */
     vertexFormat;
 
-    /** @type {Map} */
-    userAttributes;
-
     /**
      * Constructs shader processing options, used to process the shader for uniform buffer support.
      *
@@ -27,17 +24,14 @@ class ShaderProcessorOptions {
      * the bind group.
      * @param {import('./vertex-format.js').VertexFormat} [vertexFormat] - Format of the vertex
      * buffer.
-     * @param {Map} [userAttributes] - Semantics map of the user set vertex attributes.
      */
-    constructor(viewUniformFormat, viewBindGroupFormat, vertexFormat, userAttributes) {
+    constructor(viewUniformFormat, viewBindGroupFormat, vertexFormat) {
 
         // construct a sparse array
         this.uniformFormats[BINDGROUP_VIEW] = viewUniformFormat;
         this.bindGroupFormats[BINDGROUP_VIEW] = viewBindGroupFormat;
 
         this.vertexFormat = vertexFormat;
-
-        this.userAttributes = userAttributes;
     }
 
     /**
--- a/src/platform/graphics/shader-processor-options.js
+++ b/src/platform/graphics/shader-processor-options.js
@@ -15,6 +15,9 @@ class ShaderProcessorOptions {
     /** @type {import('./vertex-format.js').VertexFormat[]} */
     vertexFormat;
 
+    /** @type {Map} */
+    userAttributes;
+
     /**
      * Constructs shader processing options, used to process the shader for uniform buffer support.
      *
@@ -24,6 +27,7 @@ class ShaderProcessorOptions {
      * the bind group.
      * @param {import('./vertex-format.js').VertexFormat} [vertexFormat] - Format of the vertex
      * buffer.
+     * @param {Map} [userAttributes] - Semantics map of the user set vertex attributes.
      */
     constructor(viewUniformFormat, viewBindGroupFormat, vertexFormat) {
 
@@ -32,6 +36,8 @@ class ShaderProcessorOptions {
         this.bindGroupFormats[BINDGROUP_VIEW] = viewBindGroupFormat;
 
         this.vertexFormat = vertexFormat;
+
+        this.userAttributes = userAttributes;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/platform/graphics/shader-processor-options.js
```diff
@@ -15,6 +15,9 @@ class ShaderProcessorOptions {
     /** @type {import('./vertex-format.js').VertexFormat[]} */
     vertexFormat;
 
+    /** @type {Map} */
+    userAttributes;
```
mvaligursky, line 19: is it used in this class?

src/scene/shader-lib/program-library.js
```diff
@@ -151,6 +151,12 @@ class ProgramLibrary {
             const generatedShaderDef = this.generateShaderDefinition(generator, name, generationKey, options);
             Debug.assert(generatedShaderDef);
 
+            if (processingOptions.userAttributes && processingOptions.userAttributes.size > 0) {
```
mvaligursky, line 154: if standard material has the attributes, the standard shader generator should handle this, and not a program-library.

src/scene/materials/standard-material.js
```diff
@@ -643,6 +645,22 @@ class StandardMaterial extends Material {
...
+     *
+     * @param {string} name - The name of the parameter to set.
+     * @param {string} semantic - Semantic to map the vertex data. Must match with the semantic set on vertex stream
+     * of the mesh.
+     * @example
+     * mesh.setVertexStream(pc.SEMANTIC_ATTR0, offset, 3);
+     * material.setAttribute('offset', pc.SEMANTIC_ATTR0);
+     */
+    setAttribute(name, semantic) {
+        if (!this.userAttributes.has(semantic)) {
```
mvaligursky, line 659: I'd say remove the if, just set it .. that'd allow to override existing ones too.

LeXXik:
Hmm, it looks like the CI job is failing. Similar to https://github.com/playcanvas/developer.playcanvas.com/pull/520. Not sure what to do.

src/scene/materials/standard-material.js
```diff
@@ -868,6 +884,12 @@ class StandardMaterial extends Material {
         library.register('standard', standard);
         const shader = library.getProgram('standard', options, processingOptions, this.userId);
 
+        if (this.userAttributes) {
```
mvaligursky, line 887: I don't think we want to do this here this way, as the getProgram internally does shader caching, and the cache key needs to know about those additional attributes, otherwise it will provide the same shader regardless of the need of the additional attributes.

I think here we need to convert userAttributes to options before calling the getProgram. That internally calls standard.js generator, where the options would be to generate the key and also generate the matching shader.

LeXXik: I moved it to options builder. Would that be fine? Or do you want it to be added to options before that?

src/scene/materials/standard-material-options-builder.js
```diff
@@ -63,6 +63,12 @@ class StandardMaterialOptionsBuilder {
 
     _updateSharedOptions(options, scene, stdMat, objDefs, pass) {
         options.forceUv1 = stdMat.forceUv1;
+
+        // USER ATTRIBUTES
+        if (stdMat.userAttributes) {
+            options.userAttributes = Object.fromEntries(stdMat.userAttributes.entries());
```
mvaligursky, line 69: We should add it to `options.litOptions` here.

src/scene/materials/standard-material.js
```diff
@@ -868,6 +884,12 @@ class StandardMaterial extends Material {
         library.register('standard', standard);
         const shader = library.getProgram('standard', options, processingOptions, this.userId);
 
+        if (options.userAttributes) {
```
mvaligursky, line 887: This block should move to LitShader's constructor

src/scene/materials/standard-material-options-builder.js
```diff
@@ -63,6 +63,12 @@ class StandardMaterialOptionsBuilder {
 
     _updateSharedOptions(options, scene, stdMat, objDefs, pass) {
         options.forceUv1 = stdMat.forceUv1;
+
+        // USER ATTRIBUTES
+        if (stdMat.userAttributes) {
+            options.litOptions.userAttributes = Object.fromEntries(stdMat.userAttributes.entries());
```
mvaligursky, line 69: All is good now, just one little thing - could you please add 'userAttributes' member to the lit-shader-options.js as well.

mvaligursky (approved):
Nice and thanks!

PR #5613, opened 2023-09-06
rollup-shader-chunks.mjs: simplify + fix template string bug

kungfooman:
Fixes bug indirectly introduced via https://github.com/playcanvas/engine/pull/5602 (it wasn't a bug before because the filter was broken :sweat_smile: ):

image

The reasoning is quite simple, we used `JSON.stringify` to add `"` around the string, but that renders every occurance of `${...}` useless afterwards (hence they landed in the build folder like `" ... ${decode} ..."`.

Affects every build target:

image

Fixes examples: Paint Mesh + Reflection Box

@epreston If you have time, can you review?

### DIFF
--- a/utils/rollup-shader-chunks.mjs
+++ b/utils/rollup-shader-chunks.mjs
@@ -30,17 +30,15 @@ export function shaderChunks({
         transform(source, shader) {
             if (!enabled || !filter(shader)) return;
 
-            source = source.replace(/\/\* *glsl *\*\/\s*`(.*?)`/gs, function (match, glsl) {
-                return JSON.stringify(
-                    glsl
-                        .trim() // trim whitespace
-                        .replace(/\r/g, '') // Remove carriage returns
-                        .replace(/ {4}/g, '\t') // 4 spaces to tabs
-                        .replace(/[ \t]*\/\/.*\n/g, '') // remove single line comments
-                        .replace(/[ \t]*\/\*[\s\S]*?\*\//g, '') // remove multi line comments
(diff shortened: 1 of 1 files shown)


epreston:
JSON.stringify has always been used by this plugin, even before the change.  I'll take a look.

kungfooman:
> JSON.stringify has always been used by this plugin, even before the change. I'll take a look.

Yep, the trick is to include the backtick in the regex replace and simply keep it as it is dropping `JSON.stringify`, to not change the string kind (turning template string into normal string).

epreston:
Only one fragment is using string template literal replacement.

`src\scene\shader-lib\chunks\common\frag\reproject.js` at line 58.

```
${decode}
${encode}
```

`JSON.Stringify` ensures that the contend assigned to var or const is faithfully stored / reproduced.  I'll need another hour to review.

mvaligursky (approved):
I take it we're all in favour of merging this in.

epreston:
I think so.  I'm just doing a quick check on why other libraries go to pains to avoid template literals in the final build project.

epreston:
Once this is merged, I want to add to this.

With the correction in this PR, this will produce the following for ES6:

```js
var shader_fragment_begin = `
  // shader code
`;
```

Other tools I use produce the following:

```js
const shader_fragment_begin = (
  /* glsl */
  `
  // shader code
  `
);
```

The parens ensure it gets evaluated, evaluated once, const ensures it is optimal and dependable. It still has the benefits of the template literal and supports the use case of that file.

PR #5616, opened 2023-09-07
Refactor rendering in MiniStats

mvaligursky:
- removed support for multi draw calls, which is no longer used
- rendering using a mesh instead of setting up everything on the device manually


mvaligursky:
Is that something that worked with the previous version of MiniStats? I have not touched anything related to this.

LeXXik:
Nope, I don't think it ever worked. Maybe some pre-AppBase did. Not sure.

slimbuck (approved):

PR #5617, opened 2023-09-08
Don't drop newline of single line comments

### LINKED ISSUE #5614: Shader Preprocessor suddenly broken?
Demo: [Graphics_GroundFog.html.zip](https://github.com/playcanvas/engine/files/12538237/Graphics_GroundFog.html.zip)

(This is *not* a .zip, I added .zip extension because GitHub doesn't allow .html files)

You can add the demo in engine root and then you see something like this:

image

Full error:

```
playcanvas.js:7943 Uncaught TypeError: Cannot read properties of undefined (reading 'keep')
    at Function._preprocess (playcanvas.js:7943:34)
    at Function.run (playcanvas.js:7838:18)
    at new Shader (playcanvas.js:8013:38)
    at Object.createShaderFromCode (playcanvas.js:18321:13)
    at Graphics_GroundFog.html:143:27
    at AssetListLoader._loadingComplete (playcanvas.js:69378:21)
    at playcanvas.js:69391:13
```

kungfooman:
Fixes #5614

Introduced by https://github.com/playcanvas/engine/pull/5602

Sorry @epreston for bringing this PR up again, but it caused another bug because you rewrote the RegExp.

You can't just remove newlines because the internal C-style `Preprocessor` depends on them, e.g.

For example code like this:

```c
#ifndef LINEARIZE_DEPTH
#ifndef CAMERAPLANES
#define CAMERAPLANES
uniform vec4 camera_params; // x: 1 / camera_far,      y: camera_far,     z: camera_near,        w: is_ortho
#endif
```

Turns into:

```c
#ifndef LINEARIZE_DEPTH
#ifndef CAMERAPLANES
#define CAMERAPLANES
uniform vec4 camera_params;#endif
```

Or code like this:

```c
#else // GL2
#ifndef UNPACKFLOAT
```

Turns into:

```c
#else#ifndef UNPACKFLOAT
```

Examples from:

https://github.com/playcanvas/engine/blob/1fcdbd804e23efb5387f9cfe53ba339ae8bb295a/src/scene/shader-lib/chunks/common/frag/screenDepth.js#L16-L37

### DIFF
--- a/utils/rollup-shader-chunks.mjs
+++ b/utils/rollup-shader-chunks.mjs
@@ -35,7 +35,7 @@ export function shaderChunks({
                     .trim() // trim whitespace
                     .replace(/\r/g, '') // Remove carriage returns
                     .replace(/ {4}/g, '\t') // 4 spaces to tabs
-                    .replace(/[ \t]*\/\/.*\n/g, '') // remove single line comments
+                    .replace(/[ \t]*\/\/.*/g, '') // remove single line comments
                     .replace(/[ \t]*\/\*[\s\S]*?\*\//g, '') // remove multi line comments
                     .concat('\n') // ensure final new line
                     .replace(/\n{2,}/g, '\n'); // condense 2 or more empty lines to 1


epreston:
Good edge case, happy for any improvements and to share.

Reviewing now.

epreston:
Best to match the "\n" and replace with "\n" ?  Something like:

```
.replace(/[ \t]*\/\/.*\n/g, '\n')
```

Is that over cooking it ?  Played with it for a few minutes.  I can't find an edge case where this is required but, thought I should ask anyway.

Maksims:
You can use `m` option in regexp for enabling multiline queries. Then the regexp will be: `/\/\/.*$/gm`.

kungfooman:
You don't need a multiline regexp for a single line comment... existing regexp also takes care of removing whitespaces before the single line comment.

Maksims:
> You don't need a multiline regexp for a single line comment... existing regexp also takes care of removing whitespaces before the single line comment.

It works, but query can be less efficient, as it has to scan whole text. While using `$` can optimise the regexp execution.

kungfooman:
@Maksims I'm not trusting unbased performance claims without an actual performance test, but if you want to spend the time, I would be happy to change the regexp. Right now I just want to make sure we have no broken shaders in the engine.

epreston:
It's always possible to improve. The complete chain of 6 regex processes the the full source code in 16 ms or less.

`rollup --peft -c` to confirm.

I'd like to ensure we fix this edge case which is a little more urgent.  I have this in prod on my other projects.  I feel this is an important change.

mvaligursky (approved):
Thanks guys for the fix!

PR #5621, opened 2023-09-11
Cubemap with the new prefiltered data is ignored when it's used by a …

figo2264:
…material.

Fixes #

The old style prefiltered cubemap which uses dds for the faces is causing problem on 'restoring webgl context.

https://forum.playcanvas.com/t/what-cause-webgl-context-lost/32886/2?u=sooyong_kim

The suggestion is to switch prefiltered data with the new type that uses 'png'

It works fine with skybox, but it's not working when the cubemap is used by a material.

image

The new type of prefiltered data only uses a spherical texture. Not like old prefiltered cubemap which passes 6 textures for each x, y, and z, it only passes one texture, the rests are marked as null.

### DIFF
--- a/src/scene/materials/standard-material.js
+++ b/src/scene/materials/standard-material.js
@@ -818,6 +818,14 @@ class StandardMaterial extends Material {
 
         const isPhong = this.shadingModel === SPECULAR_PHONG;
 
+        if (this._prefilteredCubemaps && this._prefilteredCubemaps[0]) {
+            const isOtherTexturesNull = this._prefilteredCubemaps.slice(1).every(cubemap => !cubemap);
+
+            if (isOtherTexturesNull) {
+                this.envAtlas = this._prefilteredCubemaps[0];
+            }
+        }
+
         // set overridden environment textures
         if (this.envAtlas && this.cubeMap && !isPhong) {
             this._setParameter('texture_envAtlas', this.envAtlas);


mvaligursky (approved):
Thanks, looks good, but I'll leave this to @slimbuck for a final approval / merge.

figo2264:
I think the issue is not completely solved.
The mipmap level seems to be not working properly.

Need to check more.

mvaligursky:
The single png image (envatlas) uses a single mipmap (only level 0) which encodes mipmaps inside of it, so one level is all is needed. But please do check and let us know.

If it helps, in the engine we load those envAtlases this way:

```
new pc.Asset('helipad-env-atlas', 'texture', { url: '/static/assets/cubemaps/helipad-env-atlas.png' }, { type: pc.TEXTURETYPE_RGBP, mipmaps: false }),
```

slimbuck:
I've just tried testing this locally. I created a scene with a global cubemap and a material with a cubemap override.

I invoke `pc.app.graphicsDevice.debugLoseContext()` (to test a lost context state) it seems to recover fine for me.

What problems are you having with the existing code @figo2264 ?

mvaligursky:
Did the recovered context recover the global cubemap as well? I don't think we have a code that handles that, as internally we only store the generated envAtlas which is not recovered?

figo2264:
> I've just tried testing this locally. I created a scene with a global cubemap and a material with a cubemap override.
> 
> 
> 
> I invoke `pc.app.graphicsDevice.debugLoseContext()` (to test a lost context state) it seems to recover fine for me.
> 
> 
> 
> What problems are you having with the existing code @figo2264 ?

if uses the old dds for the cubemap faces, it doesn't recover properly on iOS or iPadOS.

slimbuck:
Oh I see - so you're using the old style cubemaps still. Why not switch to envAtlas @figo2264?

figo2264:
@slimbuck we were using the old style cubemap with dds and faced "webgl context lost" issue on an apple device randomly. Especially when it comes back from the background to the foreground. 
So i was suggested to use the new envatlas. 
it works almost except that the material can overwrite the environment. 
so i dig the engine code and found that the new envatlas only uses a single texture.

Another problem i found is that the all the mip
ap level is reflected to the target texture so the reflection isn't correct.

slimbuck:
Can you share the project with me so I can take a look?

As I say, I tried this locally and it all worked for me!

slimbuck:
> Did the recovered context recover the global cubemap as well? I don't think we have a code that handles that, as internally we only store the generated envAtlas which is not recovered?

This was envAtlas, should work!

figo2264:
@slimbuck i can create a sample project and share it with u.

slimbuck:
Closing in favour of #5627.

PR #5622, opened 2023-09-11
Vec2.rotate(degrees), Vec2.angle, Vec2.angleTo methods

Maksims:
Adds two methods to Vec2 to help with converting to radians and rotating it.

### New APIs:
```javascript
// pc.Vec2
vec.angle() // returns Euler angle of a vector
vec.angleTo(vector) // returns shortest Euler angle from one vector to another
vec.rotate(degrees) // rotate a vector by degrees (by Euler angle)
```

### DIFF
--- a/src/core/math/vec2.js
+++ b/src/core/math/vec2.js
@@ -410,7 +410,7 @@ class Vec2 {
     }
 
     /**
-     * Rotate a vector by Euler angle.
+     * Rotate a vector by an angle in degrees.
      *
      * @param {number} degrees - The number to degrees to rotate the vector by.
      * @returns {Vec2} Self for chaining.
--- a/src/core/math/vec2.js
+++ b/src/core/math/vec2.js
@@ -1,3 +1,5 @@
+import { math } from './math.js';
+
 /**
  * A 2-dimensional vector.
  *
@@ -408,20 +410,20 @@ class Vec2 {
     }
 
     /**
-     * Rotate a vector by radians.
+     * Rotate a vector by Euler angle.
      *
-     * @param {number} radians - The number to rotate by in radians.
+     * @param {number} degrees - The number to rotate by Euler angle.
      * @returns {Vec2} Self for chaining.
      * @example
-     * const vec = new pc.Vec2(0, 10);
+     * const v = new pc.Vec2(0, 10);
      *
-     * vec.rotate(Math.PI * 0.25); // rotates by 45 degrees
+     * v.rotate(45); // rotates by 45 degrees
      *
      * // Outputs [7.071068.., 7.071068..]
-     * console.log("Vector after rotation is: " + vec.toString());
+     * console.log("Vector after rotation is: " + v.toString());
      */
-    rotate(radians) {
-        const angle = Math.atan2(this.x, this.y) + radians;
+    rotate(degrees) {
+        const angle = Math.atan2(this.x, this.y) + (degrees / math.RAD_TO_DEG);
         const len = Math.sqrt(this.x * this.x + this.y * this.y);
         this.x = Math.sin(angle) * len;
         this.y = Math.cos(angle) * len;
@@ -429,17 +431,33 @@ class Vec2 {
     }
 
     /**
-     * Returns the angle in radians of the specified 2-dimensional vector.
+     * Returns the Euler angle of the specified 2-dimensional vector.
      *
-     * @returns {number} The angle in radians of the specified 2-dimensional vector.
+     * @returns {number} The Euler angle of the specified 2-dimensional vector.
      * @example
-     * const vec = new pc.Vec2(6, 0);
-     * const rad = vec.radians();
-     * // Outputs 1.5707963..
-     * console.log("The angle of the vector is: " + rad);
+     * const v = new pc.Vec2(6, 0);
+     * const angle = v.angle();
+     * // Outputs 90..
+     * console.log("The angle of the vector is: " + angle);
      */
-    radians() {
-        return Math.atan2(this.x, this.y);
+    angle() {
+        return Math.atan2(this.x, this.y) * math.RAD_TO_DEG;
+    }
+
+    /**
+     * Returns the shortest Euler angle between two 2-dimensional vectors.
+     *
+     * @param {Vec2} rhs - The 2-dimensional vector to calculate angle to.
+     * @returns {number} The Euler angle of the shortest angle between two 2-dimensional vectors.
+     * @example
+     * const a = new pc.Vec2(0, 10); // up
+     * const b = new pc.Vec2(1, -1); // down-right
+     * const angle = a.angleTo(b);
+     * // Outputs 135..
+     * console.log("The angle between vecotrs a and b: " + angle);
+     */
+    angleTo(rhs) {
+        return Math.atan2(this.x * rhs.y + this.y * rhs.x, this.x * rhs.x + this.y * rhs.y) * math.RAD_TO_DEG;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


mvaligursky (commented):
I think rotate should take an angle in degrees, like all ublic facing functionality in math.
And the `radians` should be `get angle` that returns the angle in degrees as well.

Maksims:
> I think rotate should take an angle in degrees, like all public facing functionality in math. And the `radians` should be `get angle` that returns the angle in degrees as well.

Good point. I've updated the PR and added additional method: `angleTo`.

src/core/math/vec2.js
```diff
@@ -407,6 +409,57 @@ class Vec2 {
...
+     * @example
+     * const v = new pc.Vec2(0, 10);
+     *
+     * v.rotate(45); // rotates by 45 degrees
+     *
+     * // Outputs [7.071068.., 7.071068..]
+     * console.log("Vector after rotation is: " + v.toString());
+     */
+    rotate(degrees) {
+        const angle = Math.atan2(this.x, this.y) + (degrees / math.RAD_TO_DEG);
```
mvaligursky, line 426: how about
```
const angle = (this.angle() + degrees) / math.RAD_TO_DEG;
```

Even though I understand you're trying to avoid a function call.

mvaligursky: or even better
```
const angle = (this.angle() + degrees) * math.DEG_TO_RAD;
```

mvaligursky: actually, I think I like your code better, avoid the function and multiply. Just replace the / by *math.DEG_TO_RAD I think.

mvaligursky:
Also, please update the PR's title.

slimbuck:
Could you add a `src = this` param to `rotate()`? I did this for some vec2, vec3 and quat functions recently, eg https://github.com/playcanvas/engine/blob/main/src/core/math/vec2.js#L416. It makes the interface a little more consistent, a little more flexible and in some cases will result in one fewer vector copy.

Maksims:
@slimbuck, not sure where would `src` argument would go in `rotate` method. As this method is within semantics of `mulScalar` or `addScalar`, where only modifier argument is provided.

slimbuck:
It would go like this:

```javascript
    rotate(degrees, src = this) {
        const angle = Math.atan2(src.x, src.y) + (degrees * math.DEG_TO_RAD);
        const len = Math.sqrt(src.x * src.x + src.y * src.y);
        this.x = Math.sin(angle) * len;
        this.y = Math.cos(angle) * len;
        return this;
    }
```

You're right rotate could be considered closer to `mulScalar` than `normalize` since rotate takes takes a parameter like mulScalar, but I'd say we could easily add src parameter to `mulScalar` one day too, why not?

Maksims:
> but I'd say we could easily add src parameter to `mulScalar` one day too, why not?

Sure, shall then this be left as is, to ensure consistency. And if `src` to be added to similar methods, that would come from a future PR?

slimbuck:
Sure :)

src/core/math/vec2.js
```diff
@@ -407,6 +409,57 @@ class Vec2 {
...
+     * Returns the shortest Euler angle between two 2-dimensional vectors.
+     *
+     * @param {Vec2} rhs - The 2-dimensional vector to calculate angle to.
+     * @returns {number} The Euler angle of the shortest angle between two 2-dimensional vectors.
+     * @example
+     * const a = new pc.Vec2(0, 10); // up
+     * const b = new pc.Vec2(1, -1); // down-right
+     * const angle = a.angleTo(b);
+     * // Outputs 135..
+     * console.log("The angle between vecotrs a and b: " + angle);
```
willeastcott, line 457:
```suggestion
     * console.log("The angle between vectors a and b: " + angle);
```

src/core/math/vec2.js
```diff
@@ -407,6 +409,57 @@ class Vec2 {
...
+     * const angle = v.angle();
+     * // Outputs 90..
+     * console.log("The angle of the vector is: " + angle);
+     */
+    angle() {
+        return Math.atan2(this.x, this.y) * math.RAD_TO_DEG;
+    }
+
+    /**
+     * Returns the shortest Euler angle between two 2-dimensional vectors.
```
willeastcott, line 448:
```suggestion
     * Returns the shortest Euler angle in degrees between two 2-dimensional vectors.
```

src/core/math/vec2.js
```diff
@@ -407,6 +409,57 @@ class Vec2 {
...
+     *
+     * @param {Vec2} rhs - The 2-dimensional vector to calculate angle to.
+     * @returns {number} The Euler angle of the shortest angle between two 2-dimensional vectors.
```
willeastcott, line 451:
```suggestion
     * @returns {number} The shortest angle in degrees between two 2-dimensional vectors.
```

src/core/math/vec2.js
```diff
@@ -407,6 +409,57 @@ class Vec2 {
...
+    rotate(degrees) {
+        const angle = Math.atan2(this.x, this.y) + (degrees * math.DEG_TO_RAD);
+        const len = Math.sqrt(this.x * this.x + this.y * this.y);
+        this.x = Math.sin(angle) * len;
+        this.y = Math.cos(angle) * len;
+        return this;
+    }
+
+    /**
+     * Returns the Euler angle of the specified 2-dimensional vector.
```
willeastcott, line 434:
```suggestion
     * Returns the angle of the specified 2-dimensional vector.
```

src/core/math/vec2.js
```diff
@@ -407,6 +409,57 @@ class Vec2 {
...
+     *
+     * @returns {number} The Euler angle of the specified 2-dimensional vector.
```
willeastcott, line 436:
```suggestion
     * @returns {number} The angle in degrees of the specified 2-dimensional vector.
```

src/core/math/vec2.js
```diff
@@ -407,6 +409,57 @@ class Vec2 {
         return this;
     }
 
+    /**
+     * Rotate a vector by Euler angle.
```
willeastcott, line 413:
```suggestion
     * Rotate a vector by an angle in degrees.
```

src/core/math/vec2.js
```diff
@@ -407,6 +409,57 @@ class Vec2 {
...
+     *
+     * @param {number} degrees - The number to rotate by Euler angle.
```
willeastcott, line 415:
```suggestion
     * @param {number} degrees - The number to degrees to rotate the vector by.
```

src/core/math/vec2.js
```diff
@@ -407,6 +409,57 @@ class Vec2 {
...
+     * Returns the Euler angle of the specified 2-dimensional vector.
+     *
+     * @returns {number} The Euler angle of the specified 2-dimensional vector.
+     * @example
+     * const v = new pc.Vec2(6, 0);
+     * const angle = v.angle();
+     * // Outputs 90..
+     * console.log("The angle of the vector is: " + angle);
+     */
+    angle() {
```
willeastcott, line 443: Question: should this be a function or a getter?

Maksims: I believe it should be a function, to stay consistent with similar methods, such as: length()

willeastcott (approved):

src/core/math/vec2.js
willeastcott: That's a good and fair point. But I'm wondering whether length, for example, should also be a getter these days. If we *do* decide to leverage getters in the math API, that's for another PR... 😄

mvaligursky (approved):

PR #5623, opened 2023-09-11
Easy ES5 ES6 Distinguishability via file extension only

kungfooman:
This fixes an issue we encounter in https://github.com/playcanvas/engine/pull/5555 and makes sure that we can use a simple conditional which always works:

```js
ENGINE_PATH.includes('.mjs')
```

The only workaround otherwise would be to peek into the file via some kind of `/...import.../magic` regexp, which just increases technical debt while not doing anything more than... figuring out if a file is ES5 or ES6 - and that's exactly the job of the `*.mjs` file extension :rainbow: 


mvaligursky:
If you did `git add -A .` it would be marked as rename, instead of this large diff.

mvaligursky:
@slimbuck @willeastcott do you foresee any issue with this change?

kungfooman:
> If you did `git add -A .` it would be marked as rename, instead of this large diff.

Unfortunately not, I just tried it:

image

I could finalize that PR, but it would be the same. Only solution is to break it down in 2 PR's:

1) Rename `src/index.js` -> `src/index.mjs`
2) Add deprecated `src/index.js`

kungfooman:
> have you tried new branch, and first commit doing that?

Yep, did a new branch directly from `main`, only one commit:

https://github.com/kungfooman/playcanvas-engine/commits/easy-es5-es6-distinguishability-git-A

> but not a big deal anyways

:+1:

mvaligursky (approved):

willeastcott (commented):
I'm fine with this. Also wondering whether it makes sense to migrate engine source to .mjs wholesale.

willeastcott (approved):

epreston:
Sorry to be late to the conversation.

There's no need to do any of this.  Need to add `"type": "module",` to the `package.json`

This covers node, browsers, tools, etc.

Basic idea is when the source (not the build products) are es6 modules, you mark it in the package.json and all tools work on that assumption.

kungfooman:
> There's no need to do any of this. Need to add `"type": "module",` to the `package.json`

What you are describing is the ESM algorithm: https://nodejs.org/api/esm.html

image

For whatever reason you dislike part (2) of it - it's there for a reason, if you like it or not.

epreston:
Nothing to dislike. 

Leave some room buy reserving file extensions to get a further, meaningful distinction.  Allow for some files to be interpreted in multiple ways instead locking it down (so you can force cjs require on plain javascript).  Tools are smart enough to look this up once. Making node tools happy in a browser project is not my primary concern.

PR #5631, opened 2023-09-13
MiniStats render on UI layer

mvaligursky:
- updated mini-stats to render on the UI instead of IMMEDIATE layer where it was added during refactoring recently, to avoid mini-stats getting post-processed in a typical case
- updated few examples to include UI layer in their cameras to render mini-stats


willeastcott (approved):

kungfooman:
Very nice update, I just have one issue still in the Reflection Planar example:

image

Not sure if I merged wrong or do you have it aswell?

mvaligursky:
> Very nice update, I just have one issue still in the Reflection Planar example:

Fix here, thanks! https://github.com/playcanvas/engine/pull/5642

PR #5632, opened 2023-09-14
Renamed engine/src/*.js to *.mjs

mvaligursky:


epreston:
@mvaligursky you don't want to do this, its just hiding the issue that the package is not marked as es6.

mvaligursky:
I think we can do both, mark the package as well as rename the files to make it clear what is what, as we have a combination of js and mjs files in the project.

epreston:
I'll have it all cleaned up in the #5635 PR, promise. 

The "mjs" thing is confusing, and in another change, will not even be in build products.

epreston:
Sorted.  The only place in the whole code base that requires a node "cjs" extension is the "karma" config.  All these "mjs" renames are redundant.  Tools start with the correct assumptions.

willeastcott:
We still have that `index.mjs` created by @kungfooman in the root of `src`. That was one of the motivations for adopting .mjs everywhere (for consistency). TBH, I kinda OK either way.

mvaligursky:
I was keen to use mjs to distinguish module vs common js files (for example those in `/scripts` folder. When we'll add module scripting support to the Editor, we'll likely have a separate set of those scripts in module format, so it would be good to make it obvious to developed / our users.

epreston:
I'll help out with that.  I think I understand what you mean.  At the moment, just want to avoid the equivalent of calling a `vec4` a `color`; might have the same shape but create some subtle issues in interpretation down the line.

epreston:
> I was keen to use mjs to distinguish module vs common js files (for example those in `/scripts` folder. When we'll add module scripting support to the Editor, we'll likely have a separate set of those scripts in module format, so it would be good to make it obvious to developed / our users.

I am hoping this implementation will be that everything is a module, wrapped in an es6 proxy.  Something similar to:

https://2ality.com/2016/11/trace-globals-proxy.html

Feeds global variables / translates the scripts dependencies.  You get interop, version independence, and isolate user code from crashing editor.

epreston:
I found a better way to explain it.  Only 2 files are strictly commmonjs "node only" files in the scripts directory.

`earcut.js` in the `/scripts` is a commonjs file that can not run in the browser.  We want to ensure it does not get bundled and used in that way.  At the moment, someone could make that mistake.

if we marked it as `earcut.cjs`, tools could would warn when bundling for browser targets.

here's the other question, how many es6 "node only" files are in that directory ?  Everything is just 'js'.  Marking everything as "node only" module code with `mjs` leaves you in the same position, you have to guess or work it out.

mvaligursky:
`\scripts` folder does not contain any scripts that we use with node or bundling, it's all just code we dynamically load and execute in the Editor as part of a scripting system, without any bundling and running on ES5 as well. 
In the future we'll replace those by proper modules, when we add some form of bundling into Editor projects, but it's not something we need to solve at the moment, that's larger long term projects.

epreston:
Sorry for the confusion. 

Are we talking about the "scripts" in the root of the engine repository ?  Here's what I see:

- cjs vs mjs aren't much of an issue in an es5 / es6 world with these files.   When you say, commonjs, you mean IIFE or UMD modules ?  Only 2 files have hints of commonjs.

- opentype and earcut in the textmesh folder use "require" wont work in a browser without "Browserify".  Maybe earcut will be ok, looks like its checking if "require is defined".

- All the others look like IIFE's that will work everywhere.  (we fixed spine)

I can make sure that every file in that directory is es5 es6 ready.  Might not be much to do.

mvaligursky:
> I can make sure that every file in that directory is es5 es6 ready. Might not be much to do.

I'm not sure we need to at the moment, a lot of this will be moved t modules when we come to that, no need to touch it till then.

mvaligursky:
> I'm 5 minutes from finishing. Looks like you can use them as they are in es6.
go ahead, it won't hurt for sure, nice!

> Have time to talk design for running user scripts ?
Not yet. It's a really large topic. The engine part of that is reasonably simple, but tooling and support in the Editor are the large and complicated part.

epreston:
Ok, keep me in mind. I've done some work in the "run arbitrary user code safely" space.

kungfooman:
> Reserve "mjs" for "node.js only" code that needs to be interpreted as es6 modules please.
> 
> ```
> An MJS file is a source code file containing an ES Module (ECMAScript Module) for use with a Node.js application.
> ```

I don't know where you got that quote from, but a google search leads to an Indian dev on StackOverflow... not exactly what I consider a based and definitive statement.

On the other side, a team whose skills and developments which power a huge chunk of the entire JS ecosystem - Google's V8 - states: https://v8.dev/features/modules#mjs

image

> This is a browser library and taking that approach creates other issues.

When I write importmaps I always like to see `*.mjs` extensions, exactly for the "crystal clear" reasoning. You find libraries like:

https://www.npmjs.com/package/@monaco-editor/react?activeTab=code

image

If - like you propose - we would ONLY do `package.json` magic spells, then we would need to fetch a `package.json` in the Examples browser and decide if we load the code as CJS or ESM - but that's overkill and just produces faulty technical debt solutions, if you can just set the right extension and shortcut the module format detection.

epreston:
Very valid.  Its been 5 years since that article, typescript is part of the code base, I invite you to look at other popular libraries.

A distinction between web / and node javascript is far more valuable for file extensions.  ESM code runs universally on both.

Look to the future:

I was able to find 1 file in code base that truly needed to be "cjs".  Everything else is ESM, plain JS, or a variant of an IIFE. There are relatively few files that would pass as es5.  Why tattoo the code base weeks from living in a future world.

epreston:
Keep in mind, this does not affect the build products.  I understand you like clarity, extension or a word are both great to know what you are importing.  It's rare to read a well built project that would disagree with you on that.

importmaps are a luxury for people without bandwidth, target device support, or performance constraints.  Most of the world bundles and the benefits are amazing for developers, users, and consumers of the library.

There's no magic, the package.json directs tools consuming the library to what they need and the format to find it.  See the "exports" section.  In a way, it's an importmap you don't have to think about.

epreston:
> If - like you propose - we would ONLY do package.json magic spells, then we would need to fetch a package.json in the Examples browser and decide if we load the code as CJS or ESM - but that's overkill and just produces faulty technical debt solutions, if you can just set the right extension and shortcut the module format detection.

That would never be the case and difficult design to justify.  You can't load cjs to begin with in the browser.  Commonly people on this project confuse "iife" and "cjs".  You can consume "iife" directly from ES modules.  There is no additional thought or effort.

In fact, this where I would like to discuss using "importmaps" for a very high quality developer experience with much less complexity than what is there now.

mvaligursky:
Closing this as at this stage we're not going to do the rename.

PR #5635, opened 2023-09-14
fix: define es6 package as es6 package

epreston:
The source (not the build products) for the project are es6 modules.  Inform all tools that this is an es6 project.

At the moment, the package.json does not specify this correctly so tools assume "cjs".  Some entry points are marked with "mjs" to correct this, but, there will always be a new edge case where incorrect assumption is uncovered.

In other words, this sets the "source" assumption to be "modules", so "js" is assumed to be "mjs", "cjs" can still be used to mark tooling code that needs to be interpreted as node only "cjs". "mjs" is reserved for node only code that needs to be evaluated as a module.

Renaming all the files to `.mjs` is pointless once the project is marked correctly.

References: #5623
References: #5555

Fixes the odd amount of effort expended juggling file extensions for js files.


mvaligursky (approved):

mvaligursky:
Looks good to me.
Any thoughts @kungfooman ?

epreston:
To remain consistent with standards:

"cjs" - node only code written in commmonjs
"mjs" - node only code using es6 modules
"js" - browser and tooling code written to the package standard

The package.json directs a few things for consumers of the library. There's a few of odd naming conventions and configuration settings in place to "make things work" because the package type was not set.  I'd like to talk about build products and simplify this in another PR.

kungfooman:
Looks good to me too, even though I don't have the time right now to check every consequence... but that will come by itself.

Wouldn't we need a `scripts/package.json` to "unset" them from being ESM now?

About renaming every engine file to .mjs... I basically just like the idea because in some contexts you simply cannot quickly judge the content otherwise without peeking into the file (and even then it could be a wrong assumption) + e.g. file filtering in VSCode would be more precise.

epreston:
Does not look like a "package.json" is required for them.

Only question mark is opentype and earcut in the scripts/textmesh folder, everything is either an IIFE (which works everywhere) or just simple javascript in a declarative es5 style (which works everywhere).

If they are "node only commonjs", they should be named cjs to avoid bundling them for browsers by mistake.

Using the correct application of "cjs", "mjs" and "js" keeps things very clear.  Marking browser code as node code, in order to know what is es5 and es6 is odd.

epreston:
Update: finished checking `/scripts`.  text-mesh is already an es6 class using assuming a global of "earcut" is defined.  I guess that is not used by the es5 editor.   The rest don't show any issues.  I don't see anything stopping this code from being executed in an es5 or es6 environment.

kungfooman:
Can you explain how you test the changes here? You talk about "tools", but which tool breaks and how do you notice that the breakage is fixed with the changes here - to make it obvious what we are dealing with here

Since you define type=module in `/package.json`, I guess we can delete this file too?

https://github.com/playcanvas/engine/blob/3032623a5742f5a7e0635cbe45a6518f9773cc17/src/package.json#L1-L3

epreston:
Happy to help, its a big topic there's a few perspectives so help me narrow this down.

Yes, that is not required and can be removed as well.

epreston:
My approach to tool testing:

I maintain public template projects on github for all the major bundlers under this account.  (vite, webpack, parcel, variant configurations of rollup)  I quickly spin them up to ensure a dependency can be consumed gracefully.

I maintain a public template project specifically for playcanvas under this account to test how the library is consumed by 3rd partiess.

epreston:
The `package.json` is the traffic director for tools, a source of meta for a project, we can also cram other tool configurations in there.

Through the `type`, `types`, `exports`, `main`, `module`, `sideEffects` properties we communicate the expectations of what 3rd parties can consume.  These third parties are bundlers, people, editors, and common tools.  You kind of build a support grid from here.  (I want to write out the grid, happy to do it)

In that grid, this setting communicates to:

- node import
- bundlers consuming source
- people using the library
- editors resolving symbols

Through the `type` property specifically, we communicate the expectations of the source code used to create the build products.  Which is ES modules.  

This will be essential in divesting "extras" from core code and having a good developer experience.

When its set correctly on each project dependency in your app, you don't think about what is cjs, es modules, iife, you just import and it works.

Setting it correctly leaves room for overrides by file extension.  The overrides become redundant on a modern project, "mjs" and "cjs" can be used as a distinction between "browser" client code and "node" code.  Since only node needs the "cjs" extension, it gets earmarked with extensions.

This project is less than an handful of of files away from living in a ES2022 world. Literally a weekend and 3 PRs.  

I agree, this has more meaning in projects which have more of a mixture of frontend, backend, and CLI.  Seeing as some PC projects sometimes assume they are all cloned into the same parent folder as Engine, it's a valid concern.

kungfooman:
@epreston Thank you for your feedback, currently `examples/rollup.config.js` is using a file extension check:

https://github.com/playcanvas/engine/blob/3032623a5742f5a7e0635cbe45a6518f9773cc17/examples/rollup.config.js#L147-L151

That implies you cannot do `ENGINE_PATH=../src/index.js` right now.

Now every problem has a bunch of different solutions... and testing for `.mjs` is:

1) easy to implement, harder to introduce bugs or increase complexity where it's not needed
2) quicker for both human and machine - skipping extra parsing of JSON files

But yes, I also see your point... in a way I like to keep everything `*.js` aswell... after all we can already do:

```js
ret = await import("http://127.0.0.1/playcanvas-engine/build/playcanvas.js")
```

Important difference is:

image

And this difference can then be used to check if it was ESM or not... in case of ESM, we have to do `window.pc = ret2;` to support both UMD and ESM.

You talked about this article yesterday: https://2ality.com/2016/11/trace-globals-proxy.html

And I'm not sure how that relates to importing or how that helps, because it is eval-ing a string, so it doesn't "hook" into the `import` system (to get accessed globals for assigning them to `window`?).

> That would never be the case and difficult design to justify. You can't load cjs to begin with in the browser. Commonly people on this project confuse "iife" and "cjs". You can consume "iife" directly from ES modules. There is no additional thought or effort.

Right, everything is confusing :sweat_smile: And iife is unified in umd, what the `build/playcanvas.js` is, therefore we can consume it via `await import(url)`. I also saw you talking about ES6 flavor in regards to ESM yesterday, which is also not the same: we had ES6 const/let/classes way before ESM import/from/export (can't find your comment any longer to check again).

Anyway, I think this discussion was already fruitful and I will remove some dependence on MJS file extension checking in #5555 while not introducing more code... not sure if I can do so 100% yet.

epreston:
That's a check on build products.  It's not related to this.

There is no need to support UMD for any platform or use case.  It's the full technical equivalent of IIFE with less compatibility and no defined standard.  Delivering an IIFE supports the oldest legacy and ESM in one shot.

About the ES6 proxies, examples, editor: I'm writing up a discussion combining your ideas with what I see.

kungfooman:
I retract my LGTM after https://github.com/playcanvas/engine/pull/5635/commits/738d19dff5d88f48a20c84568b4ded82115da45d

I know your stance and all, but it doesn't make sense to me. I've attempted to explain the reasoning behind the JS/MJS change multiple times, but it appears you haven't quite understood yet or you simply don't want to see the issue.

epreston:
You want to use the most granular definition when it could be reserved, leaving space for it to be ignored.  It's an odd measure aligned to managing importmaps by hand.

willeastcott:
My feeling is that we should consider parking the JS->MJS rename for now and allow @epreston to run with this for a bit. We can reassess the rename later - I mean, it was just a casual suggestion from me in the first place. Unless there's something blocking the merging of the rewritten Examples Browser?

kungfooman:
> Unless there's something blocking the merging of the rewritten Examples Browser?

I liked it for making one decision as simple as it gets, but I can add some code to figure out the difference at build time if `ENGINE_PATH` is UMD or MJS, so the Examples browser can consume both again (build outputs and various ESM `index.js` from different engine PR's and quick testing etc.).

epreston:
Need some help ?

playcanvas.js should be the IIFE.
playcanvas.mjs should be the es module build

The current examples browser does not consume source, it consumes build products, it also contains an example of making the distinction.

kungfooman:
> The current examples browser does not consume source, it consumes build products, it also contains an example of making the distinction.

It's funny that you want to unify UMD and ESM so hard, but then you have such a hard opinion about `build products` vs. `source`.

It's both... `consumable source code`. One main difference is that you need to invest **work** to create a `build product`, whereas `source` is **free**.

So the *price* of a `build product` is:

1) You need to handle multiple open shells for `npm run build/watch`
2) Many extra resources / file operations on your disc
3) The `rollup` processing in itself
4) The cognitive overhead to simply get started

That's all wasting time and energy for nothing. When I'm testing a PR or do bug fixing, I do none of that. I use `source` for the quickest possible workflow. This allows you to work as fast as possible, time is money etc. :money_mouth_face: 

> Need some help ?

Thank you for asking, I pretty much just need a node/fs function right now that reliably tells me if a file is UMD or EJS if you want to work it out already. I don't know how much time you have at hand, but I have other things on my TODO list aswell :sweat_smile:

epreston:
I don't want to unify anything.  UMD is an abandoned pattern that hasn't become a standard.  It has drawbacks, incompatibilities with the last 8 years of standards, and you can ship 2 other formats covering all use cases, never having an issue.  Mix and match them, etc.

The source is at a ES11 language level, trying to consume it directly as ES5 or make any distinctions  like that won't go very far.  The reason for the build system is to translate that down on at least one build product to the lowest commmon support target which is ES5-ish.  This is why build products are consumed by examples to create an ES5 build.

I understand the confusion and complexity.  Things can be simplified; external requirements are forcing some extra machinery.

epreston:
To answer your question:

You'll already know from the build file name what you are importing.  Importing either incorrectly won't build.  Still..

Off the top of my head, if an extra check at runtime is required for some reason, the node function you are looking for is the humble "import" statement; the same why you would when you want "side effects" from a polyfill.

```js
import './file.js';
```

It works on both modules and IIFEs.  If the global variable "pc" is defined (the side effect), you imported the IIFE.  If not, you have the module.

kungfooman:
> Need some help ?
>
> playcanvas.js should be the IIFE.
> playcanvas.mjs should be the es module build

@epreston I don't have a sense we speak the same language here, but thank you for trying to help, I will implement my requirements myself.

I'm not sure why you bring up `IIFE` all the time now, maybe you are confusing that with `UMD`. `UMD` literally means `Universal Module Definition` and 100 % *encompasses* `IIFE` - that's why it's called *universal* :bulb: 

And our ES5 `build/playcanvas.js` build is `UMD` and *not* `IIFE`.

epreston:
No worries.

The code you have linked and reviewed is for selecting build products from the build directory. Should be safe to use the existing code "as is" for selecting between an ES5 or modular build.  It has nothing to do with source.

Read up on UMD. Having universal in the name hasn't helped it.  A UMD is an IIFE with less compatibility.

There are a few issues and omissions with the build products.  That's one of them.

kungfooman:
> Read up on UMD. Having universal in the name hasn't helped it. A UMD is an IIFE with less compatibility.

Wrong. A `IIFE` literally just wraps your code in a function to spare you from leaking global module variables into `globalThis`. It doesn't attempt to make sure your code name (`pc`) is added *into* `globalThis`. UMD *does*, while doing everything that `IIFE` does (wrapping your code so it doesn't leak anything).

I wrote:

> And our ES5 `build/playcanvas.js` build is `UMD` and _not_ `IIFE`.

You wrote:

> playcanvas.js should be the IIFE.

Response: You can't admit that you were wrong once.

epreston:
Dig into it when you have time.

kungfooman:
> Dig into it when you have time.

I don't know whom you are trying to fool here, but anyway:

https://github.com/playcanvas/engine/blob/3032623a5742f5a7e0635cbe45a6518f9773cc17/rollup.config.mjs#L288-L291

I'm out of this PR.

epreston:
A UMD is a pattern of writing and IIFE with a few extra `if statements`

Here it is in it's full form.

```
// myModuleName.js
(function (root, factory) {
    if (typeof define === 'function' && define.amd) {
        // AMD. Register as an anonymous module.
        define(['exports', 'b'], factory);
    } else if (typeof exports === 'object' && typeof exports.nodeName !== 'string') {
        // CommonJS
        factory(exports, require('b'));
    } else {
        // Browser globals
        factory((root.myModuleName = {}), root.b);
    }
}(typeof self !== 'undefined' ? self : this, function (exports, b) {
    // Use b in some fashion.

    // attach properties to the exports object to define
    // the exported module properties.
    exports.action = function () {};
}));
```

AMD and commonjs are useless in the browser space.  The library does not run in node, AMD needs a runtime to work in a browser.  These extra `if statements` make it break if used in ESM.

You are pointing to a mistake in the build configuration.

willeastcott:
I just looked at three and they seemed to have marked their UMD build as deprecated: https://github.com/mrdoob/three.js/blob/dev/utils/build/rollup.config.js#L154

epreston:
Yep.  It was useful in 2016, today, not so much with a supported standard built into browsers.  It's a liability if someone uses the "import" statement with the wrong file.  

Some people like the global namespace with everything in it, one file, instead of a collection of them.  Same as this project does with some files in the scripts folder.

You do the chart and you get the same support matrix.

willeastcott:
Soooo.....we mergin'? 😄

PR #5639, opened 2023-09-17
tools: add check typedocs command

epreston:
Edit: Correct references to jsdoc

Generates detailed and summary report of tsdoc and type issues related to tsdocs in Javascript files.  Output is to the console.

Running the following from the root of the project will produce a report.  
(run with `npx` at beginning if typescript is not installed globally)

```bash
tsc src/index.js --allowJs --checkJs --noEmit --target ES6
```

This PR wraps this into a command in the `package.json`.  `npx` is not required here, it will run against the version of typescript installed by this project.

```json
{
  "scripts": {
    "check:tsdocs": "tsc src/index.js --allowJs --checkJs --noEmit --target ES6",
  },
}
```

The same command can be added as a step in the github checks in the future.
(see `./github/workflows/ci.yml`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,7 +144,7 @@
     "build:types:extras": "tsc extras/index.js --allowJs --declaration --emitDeclarationOnly --outDir build/playcanvas-extras.mjs",
     "build:sourcemaps": "npm run build -- -m",
     "build:publish": "npm run build && npm run build:types",
-    "check:jsdocs": "tsc src/index.js --allowJs --checkJs --noEmit --target ES6",
+    "check:tsdocs": "tsc src/index.js --allowJs --checkJs --noEmit --target ES6",
     "watch": "npm run build -- -w",
     "watch:release": "npm run build:release -- -w",
     "watch:debug": "npm run build:debug -- -w",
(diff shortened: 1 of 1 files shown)


willeastcott:
So, a few things:

1. This isn't so much a JSDoc thing but a mechanism to dump all problems reported by the TypeScript compiler to the command line.
2. We'll be dropping JSDoc in the (hopefully) not too distant future in preference for [Typedoc output](https://developer.playcanvas.com/en/typedocs/).
3. I'm not sure how we'd be able to hook this up to GitHub Actions since there's just so many errors at the moment.

epreston:
There might some confusion there.  Title is supposed to be TypeDoc.

TSC combines the TSDoc/ JSDoc comments above functions and variables, with any typescript issues.  There's no typescript and this is only checking JS files so, all the issues are TSDoc / JSDoc issues.

The TSDoc reporter and vscode language server (autocomplete engine) are just running this in the background.

epreston:
Corrected the name of the command. Old habits.

willeastcott:
So, actually, running the `typedocs` script already output warnings/errors that affect the docs building process. Have you tried that out?

epreston:
That's only showing 47 warnings.  It's focused on the final docs.  Missing issues in tsdoc headers.

Are those the priority ?  Want those addressed ?

willeastcott:
Yeah, those are the ones that affect the actual doc output. For example, we need to figure out how to get callbacks represented correctly in the docs. There's a few other issues too. An issue is that fixing some of these things for Typedoc might break them for JSDoc.

epreston:
Sounds good. I'll start there. Need the practice.

PR #5641, opened 2023-09-17
docs: update drawTexture typedoc

epreston:
This is a small test PR to ensure the process / method is correct.

Addresses warning:

```
[warning] Failed to resolve link to "GraphicsDevice#getCopyShader" in comment for drawTexture.shader.
```

Jsdocs location: pc.drawTexture
Typedocs location: drawTexture

The reference does not exist on GraphicsDevice.  On WebglGraphicsDevice, it's marked as ignored so links do not work in jsdocs or typedocs.  

This assumes the link should be kept.  Removing the "ignore" documents "getCopyShader" in both tools.
Should the "ignore" be honoured and link removed ?

It is documenting the default value.

Notes:

- getCopyShader only exists on WebglGraphicsDevice
- drawTexture only supports WebGL device
- The link reference only works if it is not ignored.

Fixes jsdoc link, typedoc link, removes typedoc warning.

### DIFF
--- a/src/scene/graphics/quad-render-utils.js
+++ b/src/scene/graphics/quad-render-utils.js
@@ -81,7 +81,7 @@ function drawQuadWithShader(device, target, shader, rect, scissorRect) {
 /**
  * Draws a texture in screen-space. Mostly used by post-effects.
  *
- * @param {import('../../platform/graphics/webgl/webgl-graphics-device.js').WebglGraphicsDevice} device - The graphics device used to draw
+ * @param {import('../../platform/graphics/graphics-device.js').GraphicsDevice} device - The graphics device used to draw
  * the texture.
  * @param {import('../../platform/graphics/texture.js').Texture} texture - The source texture to be drawn. Accessible as
  * `uniform sampler2D * source` in shader.
--- a/src/scene/graphics/quad-render-utils.js
+++ b/src/scene/graphics/quad-render-utils.js
@@ -81,14 +81,14 @@ function drawQuadWithShader(device, target, shader, rect, scissorRect) {
 /**
  * Draws a texture in screen-space. Mostly used by post-effects.
  *
- * @param {import('../../platform/graphics/graphics-device.js').GraphicsDevice} device - The graphics device used to draw
+ * @param {import('../../platform/graphics/webgl/webgl-graphics-device.js').WebglGraphicsDevice} device - The graphics device used to draw
  * the texture.
  * @param {import('../../platform/graphics/texture.js').Texture} texture - The source texture to be drawn. Accessible as
  * `uniform sampler2D * source` in shader.
  * @param {import('../../platform/graphics/render-target.js').RenderTarget} [target] - The destination render target.
  * Defaults to the frame buffer.
  * @param {import('../../platform/graphics/shader.js').Shader} [shader] - The shader used for rendering the texture.
- * Defaults to {@link GraphicsDevice#getCopyShader}.
+ * Defaults to {@link WebglGraphicsDevice#getCopyShader}.
  * @param {import('../../core/math/vec4.js').Vec4} [rect] - The viewport rectangle to use for the
  * texture, in pixels. Defaults to fullscreen (`0, 0, target.width, target.height`).
  * @param {import('../../core/math/vec4.js').Vec4} [scissorRect] - The scissor rectangle to use for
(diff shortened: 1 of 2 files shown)


src/scene/graphics/quad-render-utils.js
```diff
@@ -81,14 +81,14 @@ function drawQuadWithShader(device, target, shader, rect, scissorRect) {
...
- * @param {import('../../platform/graphics/graphics-device.js').GraphicsDevice} device - The graphics device used to draw
+ * @param {import('../../platform/graphics/webgl/webgl-graphics-device.js').WebglGraphicsDevice} device - The graphics device used to draw
  * the texture.
  * @param {import('../../platform/graphics/texture.js').Texture} texture - The source texture to be drawn. Accessible as
  * `uniform sampler2D * source` in shader.
  * @param {import('../../platform/graphics/render-target.js').RenderTarget} [target] - The destination render target.
  * Defaults to the frame buffer.
  * @param {import('../../platform/graphics/shader.js').Shader} [shader] - The shader used for rendering the texture.
- * Defaults to {@link GraphicsDevice#getCopyShader}.
+ * Defaults to {@link WebglGraphicsDevice#getCopyShader}.
```
willeastcott, line 91: An alternative here is to simply remove the `Defaults to...` sentence and keep `getCopyShader` out of the docs. @mvaligursky?

src/scene/graphics/quad-render-utils.js
```diff
@@ -81,14 +81,14 @@ function drawQuadWithShader(device, target, shader, rect, scissorRect) {
 /**
  * Draws a texture in screen-space. Mostly used by post-effects.
  *
- * @param {import('../../platform/graphics/graphics-device.js').GraphicsDevice} device - The graphics device used to draw
+ * @param {import('../../platform/graphics/webgl/webgl-graphics-device.js').WebglGraphicsDevice} device - The graphics device used to draw
```
willeastcott, line 84: AFAIK, `drawTexture` should/will work with WebGPU. So this should be a generic `GraphicsDevice`. @mvaligursky to confirm.

mvaligursky: yes this needs to stay as GraphicsDevice as it is multi-platform

src/scene/graphics/quad-render-utils.js
```diff
@@ -100,7 +100,7 @@ function drawTexture(device, texture, target, shader, rect, scissorRect) {
     const useBlend = arguments[6];
     Debug.call(() => {
         if (useBlend !== undefined) {
-            Debug.warnOnce('pc.drawTexture no longer accepts useBlend parameter, and blending state needs to be set up using GraphicsDevice.setBlendState.');
+            Debug.warnOnce('pc.drawTexture no longer accepts useBlend parameter, and blending state needs to be set up using WebglGraphicsDevice.setBlendState.');
```
mvaligursky, line 103: and this needs to stay as GraphicsDevice too

src/scene/graphics/quad-render-utils.js
epreston: Theres a debug gaurd that stops that.  Can we define an abstract method for this and the "draw" method ?

mvaligursky: yes we should do that, similarly to how `setStencilState` does it. The problem is that JS does not have interfaces, and so if we change the function in Webgl class we have to remember to change it in the base class, there is not error reporting for this, which is bit limited. But we add those functions specifically for the documentation purpose, so we should do it here too.

epreston: I don't see getCopyShader on WebGPU (called by drawTexture).  The debug warning might be correct, this looks like it will have issues on WebGPU.

src/scene/graphics/quad-render-utils.js
epreston: Reference removed in 44a127d
`getCopyShader` set back to `@ignore` in ef35b41

src/scene/graphics/quad-render-utils.js
mvaligursky: yep, at some point I'll clean this up, for now it does not work on WebGPU if you use the default shader (which is not a case when post-effects are used, as they supply a shader)

src/scene/graphics/quad-render-utils.js
willeastcott: Fixed.

src/scene/graphics/quad-render-utils.js
willeastcott: I backed out this change.

willeastcott (approved):

PR #5642, opened 2023-09-18
Fix the order of layers in ReflectionPlanar example to render MiniStats last

mvaligursky:
fixes the issue mentioned here: https://github.com/playcanvas/engine/pull/5631#issuecomment-1722456861


kungfooman:
Works like a charm:

image

willeastcott (approved):

PR #5644, opened 2023-09-18
When the context is lost, silently ignore failed shader compilation

mvaligursky:
When the context is lost, all gl functionality should silently work, but `gl.createShader` returns null (which documentation of the function does not mentioned). This generates an error in the application.

The solution in this PR is to silently ignore shader compilation / linking errors when the device is lost, as those are compiled again when the device is restored, avoiding logging errors without real error message.

Related to https://forum.playcanvas.com/t/what-cause-webgl-context-lost/32886

Tested using CompileShader example, modified to compile a shader each frame - the errors are gone and all works correctly when the context is restored.

### DIFF
--- a/src/platform/graphics/webgl/webgl-shader.js
+++ b/src/platform/graphics/webgl/webgl-shader.js
@@ -149,13 +149,18 @@ class WebglShader {
         if (this.glProgram)
             return;
 
+        // if the device is lost, silently ignore
+        const gl = device.gl;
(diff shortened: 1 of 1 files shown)


slimbuck (approved):

figo2264:
Can I simply cheery pick webgl-shader.js file and apply it to 1.65.3 version?

mvaligursky:
Yep that should work without a problem @figo2264

PR #5650, opened 2023-09-20
Fix reallocation of a depth render target in the SceneGrab class

querielo:
There's a scenario where a camera has a custom render target without a depth buffer texture, and the SceneGrab class incorrectly assumes that the grabbed depth texture has to be the size of the device. This can cause issues when the color buffer of the render target and a canvas have different sizes.

To fix this issue, I suggest that the code should check the size of the color buffer in the absence of a depth buffer, since they should have the same size. This will ensure that the reallocation of SceneGrab.depthRenderTarget works as expected.

I hope this helps! Let me know if you have any questions.


mvaligursky:
If the render target does not have depth, we should probably just log error and early out / fail as what is the reason to execute depth grab if there is no depth buffer?

querielo:
Hm. I like your suggestion.

It just looks like there are two independent ways to grab depth buffers. "Depth grabpass" simplify usage "getLinearScreenDepth" or something like this. But a user can instantiate RenderTarget the next way:
```
const renderTarget = new pc.RenderTarget({
    colorBuffer: texture
    depth: true
});
```

It looks like "_depthBuffer === undefined"

src/scene/graphics/scene-grab.js
```diff
@@ -261,7 +261,12 @@ class SceneGrab {
                     }
 
                     // reallocate RT if needed
-                    if (self.shouldReallocate(this.depthRenderTarget, camera.renderTarget?.depthBuffer)) {
+                    if (
+                        self.shouldReallocate(
```
mvaligursky, line 265: I think a simpler fix might be to pass a sourceRenderTarget to shouldReallocate, instead of a source texture, and that way it will get the size in all cases. So in this case, just `camera.renderTarget` instead of `camera.renderTarget?.depthBuffer ?? camera.renderTarget?.colorBuffer`. All 4 calls to this function in the file would need to be adjusted the same simple way.

mvaligursky:
> In this case, `renderTarget._depthBuffer === undefined`

Yep, that's a valid point!

src/scene/graphics/scene-grab.js
querielo: Done

src/scene/graphics/scene-grab.js
```diff
@@ -211,7 +211,7 @@ class SceneGrab {
                 if (camera.renderSceneColorMap) {
 
                     // allocate / resize existing RT as needed
-                    if (self.shouldReallocate(this.colorRenderTarget, camera.renderTarget?.colorBuffer, true)) {
+                    if (self.shouldReallocate(this.colorRenderTarget, camera.renderTarget, true)) {
```
mvaligursky, line 214: Also rename sourceTexture to sourceRT in the shouldReallocate function please.

src/scene/graphics/scene-grab.js
```diff
@@ -346,7 +346,10 @@ class SceneGrab {
                 if (camera.renderSceneDepthMap) {
 
                     // reallocate RT if needed
-                    if (!this.depthRenderTarget?.colorBuffer || self.shouldReallocate(this.depthRenderTarget, camera.renderTarget?.depthBuffer)) {
+                    if (
```
mvaligursky, line 349: single line here pls

src/scene/graphics/scene-grab.js
querielo: No problem. Done

src/scene/graphics/scene-grab.js
querielo: Done

src/scene/graphics/scene-grab.js
```diff
@@ -123,19 +123,19 @@ class SceneGrab {
...
     }
 
-    shouldReallocate(targetRT, sourceTexture, testFormat) {
+    shouldReallocate(targetRT, sourceRT, testFormat) {
 
         // need to reallocate if format does not match
         if (testFormat) {
             const targetFormat = targetRT?.colorBuffer.format;
-            const sourceFormat = this.getSourceColorFormat(sourceTexture);
+            const sourceFormat = this.getSourceColorFormat(sourceRT);
```
mvaligursky, line 131: I completely missed that this parameter is used to get the format from the texture .. so RT here won't work :(

mvaligursky: I guess so. Just make sure we're not using the format in the case you're changing, as we don't want to compare against color format.
split the line is ok .. just no lines with just a `(` on it like you had.

mvaligursky (approved):
Happy with this, thanks!

querielo:
Thanks!

I have met another case. A camera with posteffects + Depth grabpass + Render target

https://github.com/playcanvas/engine/blob/d753e00475dc4033f1fe8d4540de9b3eff39a364/src/framework/components/camera/post-effect-queue.js#L114

Issue #3090, opened 2021-04-13
Post-effects applied on camera with render target issue


sinlyu:
related to #2941 #3069 

forum post:
[here](https://forum.playcanvas.com/t/apply-bloom-only-to-specific-materials/18455/40?u=sinlyu)

description and explanation is given in the title and the forum post.

mvaligursky:
- the problem seems to take place when the post processing is set up on camera using the Editor, but render target is set up later using the script.
- also some viewport scaling ratio issue seems to be there

mvaligursky:
also it seems the sizing is based on backbuffer instead of render target, needs fixing.
https://github.com/playcanvas/engine/blob/master/src/framework/components/camera/post-effect-queue.js#L60
https://github.com/playcanvas/engine/blob/master/src/framework/components/camera/post-effect-queue.js#L272

mvaligursky:
This should now be fixed, but as the repro no longer works due to API changes since it was created, I cannot confirm.
I'll close it. Please reopen if this does not work, with an up to date repro, sorry.

PR #5651, opened 2023-09-20
Fix resizing of render targets in the PostEffectQueue class

querielo:
Issue: the resizing of render targets is not properly handled for cameras with renderTarget by PostEffectQueue.

Suggest: add check the size of the camera render target before resizing render targets of effects.


mvaligursky:
possibly related to https://github.com/playcanvas/engine/issues/3090

figo2264:
Is it possible render different size on a specific camera?
For example, the main orbit camera has FHD render size.
If a certain action occurs, the orbit camera is disabled but a hidden Camera A is activated with a HD resolution, then capture the scene. 
Is it possible?

Previously, I tried but somehow the capture image from the camera A is corrupted.

LeXXik:
@figo2264 this repository is for bug reports and feature requests. Please, visit forum for help and support.

PR #5653, opened 2023-09-20
Extracted generic render pass for a cookie renderer and picker to a new modules

mvaligursky:
- just a refactor on a way to standalone render passes

### DIFF
--- a/src/framework/graphics/picker.js
+++ b/src/framework/graphics/picker.js
@@ -4,19 +4,16 @@ import { ADDRESS_CLAMP_TO_EDGE, FILTER_NEAREST, PIXELFORMAT_RGBA8 } from '../../
 import { GraphicsDevice } from '../../platform/graphics/graphics-device.js';
 import { RenderTarget } from '../../platform/graphics/render-target.js';
 import { Texture } from '../../platform/graphics/texture.js';
-import { DebugGraphics } from '../../platform/graphics/debug-graphics.js';
 
-import { SHADER_PICK } from '../../scene/constants.js';
 import { Camera } from '../../scene/camera.js';
 import { Layer } from '../../scene/layer.js';
 
 import { getApplication } from '../globals.js';
-import { Debug, DebugHelper } from '../../core/debug.js';
-import { BlendState } from '../../platform/graphics/blend-state.js';
-import { RenderPass } from '../../platform/graphics/render-pass.js';
+import { Debug } from '../../core/debug.js';
+import { RenderPassPicker } from './render-pass-picker.js';
 
 const tempSet = new Set();
-const tempMeshInstances = [];
+// const tempMeshInstances = [];
 
 /**
  * Picker object used to select mesh instances from screen coordinates.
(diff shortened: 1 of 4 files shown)


src/framework/graphics/picker.js
```diff
@@ -4,19 +4,16 @@ import { ADDRESS_CLAMP_TO_EDGE, FILTER_NEAREST, PIXELFORMAT_RGBA8 } from '../../
...
 import { getApplication } from '../globals.js';
-import { Debug, DebugHelper } from '../../core/debug.js';
-import { BlendState } from '../../platform/graphics/blend-state.js';
-import { RenderPass } from '../../platform/graphics/render-pass.js';
+import { Debug } from '../../core/debug.js';
+import { RenderPassPicker } from './render-pass-picker.js';
 
 const tempSet = new Set();
-const tempMeshInstances = [];
+// const tempMeshInstances = [];
```
willeastcott, line 16: Delete

src/framework/graphics/picker.js
```diff
@@ -194,73 +190,12 @@ class Picker {
...
-                }
-            }
 
-            DebugGraphics.popGpuMarker(device);
-        });
 
-        DebugHelper.setName(renderPass, `RenderPass-Picker`);
+
+
+        /////////// !!!!!!!! todo  - this should be done once and RT resized when needed
```
willeastcott, line 197: Uhm....maybe turn down the volume! 🤣

src/scene/renderer/render-pass-cookie-renderer.js
```diff
@@ -0,0 +1,161 @@
...
+// shader runs for each face, with inViewProj matrix representing a face camera
+const textureCubeBlitFragmentShader = `
+    varying vec2 uv0;
+    uniform samplerCube blitTexture;
+    uniform mat4 invViewProj;
+    void main(void) {
+        vec4 projPos = vec4(uv0 * 2.0 - 1.0, 0.5, 1.0);
+        vec4 worldPos = invViewProj * projPos;
+        gl_FragColor = textureCube(blitTexture, worldPos.xyz);
+    }`;
```
willeastcott, line 40: Personally I wouldn't indent here...

willeastcott (approved):

src/scene/renderer/render-pass-cookie-renderer.js
mvaligursky: I find it a lot easier to scan the code and see start / end

src/framework/graphics/picker.js
mvaligursky: haha yeah sorry I missed it even when it was so obvious! :)

src/scene/renderer/render-pass-cookie-renderer.js
willeastcott: Is that so?

https://github.com/playcanvas/engine/blob/main/src/platform/graphics/webgl/webgl-graphics-device.js#L46-L84

😉

mvaligursky: I'll add the tag, thanks. I'll also indent those shaders you mentioned :)
See MorphInstance for a good example.

PR #5655, opened 2023-09-20
Extract the color grab functionality to a render pass

mvaligursky:
- standalone `RenderPassColorGrab` which implements color grab on all platforms
- simplified integration - does not depend on layer callbacks to do things, and is scheduled at the right time as a render pass in the render graph
- camera creates / owns the render pass, which is allocated when the color grab is enabled on the camera. This solves the issue with the layer solution, where if we needed more than a single color grab per scene (two or more cameras), render target would be re-allocate (to match the size) each frame per camera. Now each camera owns one.

### DIFF
--- a/src/platform/graphics/render-pass.js
+++ b/src/platform/graphics/render-pass.js
@@ -176,6 +176,9 @@ class RenderPass {
         this._execute = execute;
     }
 
+    destroy() {
+    }
+
     /**
      * @param {import('../graphics/render-target.js').RenderTarget} renderTarget - The render
      * target to render into (output). This function should be called only for render passes which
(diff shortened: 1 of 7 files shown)


src/platform/graphics/render-pass.js
```diff
@@ -176,6 +176,9 @@ class RenderPass {
         this._execute = execute;
     }
 
+    destroy() {
```
willeastcott, line 179: Are you expecting to start deallocating stuff in the base class here? If not, I'm not sure I see the value of an empty `destroy` function.

willeastcott (approved):

src/platform/graphics/render-pass.js
mvaligursky: Yes I am looking at managing render target allocations and resizing in the base class.

PR #5657, opened 2023-09-21
Handle ios 17 basis wasm regression

slimbuck:
iOS 17 introduced a wasm regression which results in some basis files failing to transcode (see https://bugs.webkit.org/show_bug.cgi?id=260252 for more context).

It appears the transcode only fails on images containing mipmaps (which have no alpha channel) and it's always the very last mipmap that fails to transcode.

This PR modifies the basis transcode logic to handle this specific failure case by copying the texture data from the second last mipmap level instead of completely failing the image transcode.

The resulting final mipmap will be very slightly incorrect, but the difference should be unnoticeable.


mvaligursky (approved):

willeastcott (approved):
Very smart workaround. Love it.

PR #5658, opened 2023-09-21
Removed CookieRenderer by making it fully integrated into cookie render pass

mvaligursky:
- refactor on the way to make render passes more standalone, easier to schedule.


src/scene/renderer/render-pass-cookie-renderer.js
```diff
@@ -1,8 +1,9 @@
+import { Debug } from '../../core/debug.js';
```
willeastcott, line 1: What about a folder `renderer/render-passes`?

src/scene/renderer/render-pass-cookie-renderer.js
```diff
@@ -74,6 +77,52 @@ class RenderPassCookieRenderer extends RenderPass {
...
+
+        // pick lights we need to update the cookies for
+        const filteredLights = this._filteredLights;
+        this.filter(lights, filteredLights);
+
+        // enabled / disable the pass
+        this.executeEnabled = filteredLights.length > 0;
+    }
+
+    filter(lights, filteredLights) {
```
willeastcott, line 103: I'm guessing this is much faster than `Array#filter`?

willeastcott (approved):

src/scene/renderer/render-pass-cookie-renderer.js
mvaligursky: maybe if we have more of those, yeah

src/scene/renderer/render-pass-cookie-renderer.js
mvaligursky: this avoids temporary array allocation, so why not

PR #5659, opened 2023-09-21
Compressed texture formats working on webgpu

slimbuck:
Few bits to get compressed textures working in WebGPU.


mvaligursky (approved):
Amazing!

PR #5662, opened 2023-09-21
Add equalsApprox to vec2, vec3, vec4 (and tests)

querielo:
The PR adds equalsApprox to vec2, vec3, vec4 and fixes the 4 point of [this review comment](https://github.com/playcanvas/engine/pull/5145#issuecomment-1465164912)

> No other math class (like Vec3) offers the capability either

### DIFF
--- a/src/core/math/vec2.js
+++ b/src/core/math/vec2.js
@@ -268,6 +268,23 @@ class Vec2 {
     }
 
     /**
+     * Reports whether two vectors are equal using an absolute error tolerance.
+     *
+     * @param {Vec2} rhs - The vector to be compared against.
+     * @param {number} [epsilon] - The maximum difference between each component of the two
+     * vectors. Defaults to 1e-6.
+     * @returns {boolean} True if the vectors are equal and false otherwise.
+     * @example
+     * const a = new pc.Vec2();
+     * const b = new pc.Vec2();
+     * console.log("The two vectors are approximately " + (a.equalsApprox(b, 1e-9) ? "equal" : "different"));
+     */
+    equalsApprox(rhs, epsilon = 1e-6) {
+        return (Math.abs(this.x - rhs.x) < epsilon) &&
+            (Math.abs(this.y - rhs.y) < epsilon);
+    }
+
+    /**
      * Returns the magnitude of the specified 2-dimensional vector.
      *
      * @returns {number} The magnitude of the specified 2-dimensional vector.
--- a/src/core/math/vec3.js
+++ b/src/core/math/vec3.js
@@ -295,6 +295,24 @@ class Vec3 {
     }
 
     /**
+     * Reports whether two vectors are equal using an absolute error tolerance.
+     *
+     * @param {Vec3} rhs - The vector to be compared against.
+     * @param {number} [epsilon] - The maximum difference between each component of the two
+     * vectors. Defaults to 1e-6.
+     * @returns {boolean} True if the vectors are equal and false otherwise.
+     * @example
+     * const a = new pc.Vec3();
+     * const b = new pc.Vec3();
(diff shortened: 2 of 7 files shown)


Maksims:
"Float math is great".
Is it worth checking first if values are actually equal so to early return true, and if not, then checking if their abs difference are within epsilon?

querielo:
@Maksims, it depends on real-world vectors. Probably, if the vectors are equal and not computed, we will get a performance boost. However, if you use the method, that means you computed two vectors and equals may not be suitable for your situation.

Here is a small performance test: 
https://jsbench.me/jvlmu8cc2d/1

querielo:
It seems that we will only achieve a performance boost if there are too many identical vectors.

On my computer, the times are equal for the following vectors
```
let unique = [
	... new Array(30).fill(new Vec4(1,2,3,4)),
	new Vec4(1,2,3,4 + 0.0000001),
	new Vec4(1,2,3,4 - 0.0000001),
	new Vec4(1,2,3,4 + 0.00001),
	new Vec4(1,2,3,4 - 0.00001),
	new Vec4(1 - 0.0000001,2,3,4),
	new Vec4(1 + 0.0000001,2,3,4),
	new Vec4(1 - 0.00001,2,3,4),
	new Vec4(1 + 0.00001,2,3,4),
]
```

Link: https://jsbench.me/jvlmu8cc2d/2

mvaligursky (approved):
Nice work, thanks!

willeastcott:
What about `Color`?

mvaligursky:
You don't typically do math on Colors much .. that's why our Color class is so lite with almost no math functionality.

willeastcott:
OK, fair enough.

Also, as folks probably know, I'm always hesitant about bloating the API unnecessarily. Any reason why this is not sufficient:

```
if (a.distance(b) < epsilon) {
    console.log('approximately equal');
}
```

kungfooman:
> Also, as folks probably know, I'm always hesitant about bloating the API unnecessarily. Any reason why this is not sufficient:

I have the same objection, that's why I usually try to find cases in the engine that could be shortened/simplified through a PR in the engine itself to prove the "real world need".

The problem with `Vec3#distance` is usually that people say it's slooow because of the `Math.sqrt` and then people use `Vec3#lengthSq` instead while multiplying the epsilon with itself to make up for the missing square root:

https://github.com/playcanvas/engine/blob/2bfd57bac9a7f31c1306ecb13541784ce1075027/src/framework/input/element-input.js#L101-L106

(maybe that's at least one case which could be simplified with this PR and maybe there are others)

LeXXik:
If I rarely need an approximate comparison, I go with Will's example. Not sure I'd expect it to be part of API.

mvaligursky:
We already have this function for Quat, so it's nice to be consistent. In the past I used those function for vectors in different engines.

querielo:
Added 2 tests: https://jsbench.me/jvlmu8cc2d/3

The 4th test is the @willeastcott's [suggestion](https://github.com/playcanvas/engine/pull/5662#issuecomment-1731120899)
The 5th test is the @kungfooman's [suggestion](https://github.com/playcanvas/engine/pull/5662#issuecomment-1731153629)

Please, check if I implemented them correctly. 

Note: there is no distance function in vec4.


My result:
Screenshot 2023-09-22 at 12 32 37

querielo:
Looks like in edge cases the original equalApprox provides the stablest result.

* Many different vectors in the first coordinate
Screenshot 2023-09-22 at 12 43 53
* Many different vectors in the last coordinate
Screenshot 2023-09-22 at 12 47 49
* Many equal vectors
Screenshot 2023-09-22 at 12 51 26

PR #5663, opened 2023-09-22
Update mesh.js

chrissterling:
Fix bug in getIndices which incorrectly modifies an array to have an array in the first element instead of setting the array to the numbers.

Fixes #

### DIFF
--- a/src/scene/mesh.js
+++ b/src/scene/mesh.js
@@ -723,7 +723,7 @@ class Mesh extends RefCountedObject {
             } else {
                 // destination data is array
                 indices.length = 0;
-                indices.push(streamIndices);
+                indices.push(...streamIndices);
             }
         } else {
             // get data from IndexBuffer


src/scene/mesh.js
```diff
@@ -723,7 +723,7 @@ class Mesh extends RefCountedObject {
             } else {
                 // destination data is array
                 indices.length = 0;
-                indices.push(streamIndices);
+                indices.push(...streamIndices);
```
mvaligursky, line 726: Nice, but lets avoid using spread operator here on these potentially larger arrays due to https://github.com/nodejs/node/issues/16870
(We've had an experience in other code with this)

kungfooman: Looks like a good-old for-loop would handles both cases equally here (untested assumption)

mvaligursky: yep

mvaligursky: @chrissterling - would you have time to do this change please?

mvaligursky:
Thanks for finding the issue and the fix, that was really helpful, we've merged slightly modified fix now, so I'm closing this one.

Issue #4844, opened 2022-11-14
Screen space ambient occlusion should only affect indirect light


animech-sijo:
The post process AO added in #3266 is much needed. However, it naively multiplies the AO with the final image (`gl_FragColor.rgb = inCol.rgb * ssao;`)

This is a problem when using realtime lights, since lit and shadowed areas are darkened equally by the AO:
![SSAO](https://user-images.githubusercontent.com/56446466/201687516-27b80e35-12ed-49ef-aaf6-83c21726a441.jpg)

Compare this to using baked AO textures in the ambient slot (in which lighting is calculated correctly):
![BakedAO](https://user-images.githubusercontent.com/56446466/201687521-a84ae6ed-413b-4a3a-beed-ba062ab463f1.jpg)

SSAO should behave the same way AO textures do. The current behavior produces results that look "dirty".

mvaligursky:
Yes, we discussed both approaches, but at the end decided to implement it as a post-process (with the mentioned limitation) as a more performant solution. In this case, the scene is rendered, and its depth buffer is used to generates ssao, which is then blended into the scene.

The better solution, which we might do at some point in the future, renders the scene into depth buffer first. Then using this depth texture it generates ssao texture (this is the same step as the current implementation does). Then the whole scene is rendered again, while the ssao texture is used during the lighting computation. The extra cost here is having to re-render the scene one more time to have access to its depth before the scene with lighting is rendered.

PR #5671, opened 2023-09-25
Fallback Webgl1 solution for depth grab is replaced by a RenderPassDepth

mvaligursky:
For WebGL1, we no longer use a custom Depth layer with render target and some manual handling of visible mesh instances under the hood, but have a render pass that is scheduled to run before the camera's render pass to render the depth.

- The main camera's render pass is no longer split into two, saving performance
- This removes last warning of this type inside the engine: `DEPRECATED: pc.Layer#renderTarget`
- new RenderPassDepth will be extended in the future to provide depth rendering before the camera for other effects, such as https://github.com/playcanvas/engine/issues/4844

### New public API:
```
GraphicsDevice.isWebGL1
GraphicsDevice.isWebGL2
```

to match existing
```
GraphicsDevice.isWebGPU
```

### DIFF
--- a/src/platform/graphics/webgl/webgl-graphics-device.js
+++ b/src/platform/graphics/webgl/webgl-graphics-device.js
@@ -385,6 +385,8 @@ class WebglGraphicsDevice extends GraphicsDevice {
 
         this.gl = gl;
         this.webgl2 = typeof WebGL2RenderingContext !== 'undefined' && gl instanceof WebGL2RenderingContext;
+        this.isWebGl2 = this.webgl2;
+        this.isWebGl1 = !this.webgl2;
         this._deviceType = this.webgl2 ? DEVICETYPE_WEBGL2 : DEVICETYPE_WEBGL1;
 
         // pixel format of the framebuffer
--- a/src/scene/graphics/render-pass-depth.js
+++ b/src/scene/graphics/render-pass-depth.js
@@ -14,7 +14,7 @@ const lights = [[], [], []];
 const _depthUniformNames = ['uSceneDepthMap', 'uDepthMap'];
 
 /**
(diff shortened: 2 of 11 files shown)


src/platform/graphics/webgl/webgl-graphics-device.js
```diff
@@ -385,6 +385,8 @@ class WebglGraphicsDevice extends GraphicsDevice {
 
         this.gl = gl;
         this.webgl2 = typeof WebGL2RenderingContext !== 'undefined' && gl instanceof WebGL2RenderingContext;
+        this.isWebGl2 = this.webgl2;
+        this.isWebGl1 = !this.webgl2;
```
willeastcott, line 389: Should `webgl2` be moved to `deprecated.js`?

src/scene/graphics/render-pass-depth.js
```diff
@@ -0,0 +1,164 @@
...
+
+const webgl1DepthClearColor = new Color(254.0 / 255, 254.0 / 255, 254.0 / 255, 254.0 / 255);
+const tempMeshInstances = [];
+const lights = [[], [], []];
+
+// uniform names (first is current name, second one is deprecated name for compatibility)
+const _depthUniformNames = ['uSceneDepthMap', 'uDepthMap'];
+
+/**
+ * A render pass implementing rendering a depth. In current implementation, the depth is encoded in
```
willeastcott, line 17:
```suggestion
 * A render pass implementing rendering of depth. In current implementation, the depth is encoded in
```

src/platform/graphics/webgl/webgl-graphics-device.js
mvaligursky: yes, I will do in a separate PR as it will have many changes.

willeastcott (approved):

PR #5675, opened 2023-09-25
Bundles Refactor

### LINKED ISSUE #5626: Asset Bundles - improvements and making it public
Asset Bundles - are a great way to reduce number of network requests by bundling assets into single .tar file.
Currently engine support for it usable, but it is not public and needs improvements. It also relies on Editor bundling functionality and would be great if described well, so that engine-only developers can create .tar files themself.

## Current "issue" with bundles:

If asset is in a bundle which is not loaded, and we try to load that asset, it will fire an error. This has couple of issues:

1. Developers who implements lazy-loading, now should check not only if asset is loaded, but also if asset is referenced in a bundle before trying to load it.
2. When we reference an asset by components, engine will load an asset for you when its needed for rendering - this is very powerful, and simplifies lazy-loading big way. But if asset is referenced by a bundle then this fails and will fire errors unless bundle is loaded in advance, which for non-preloading cases makes it hard to manage.
3. Because asset can be in multiple bundles, it makes it somewhat non-obvious what to load here.
4. Bundles do throw `blob` requests in a network tab, this preferably should be avoided, to improve network debugging.
5. .tar potentially can be streamed, instead of loading whole file first, this will spread assets parsing CPU impact when lazy-loading bundles, as well as allow parsing while network is still downloading, so improve speed at which bundled assets will load in overall.

## The potential solution to this problem:

Make `AssetRegistry.load` to take bundles into the account, and load bundles when trying to load asset which is referenced by a bundle. Instead of throwing an error (current behavior).

Add optional options argument to `AssetRegistry.load` method, that can assist in picking the right bundle with two properties `bundlesIgnore` and `bundlesFilter`:
```js
app.assets.load(asset, {
    bundlesFilter: (bundles) => {
        return bundles[0];
    }
});
```
When asset registry tries to load an asset, that is referenced in a bundle, developer has an option to ignore bundles by providing `bundlesIgnore` flag, which by default is `false` so assets will load from bundles.
And then within list of referenced bundles, if any of the bundles are already loading/loaded, we ensure asset is marked as `loading` and then don't do anything else, as it will trigger `load` event of an asset once bundle is loaded. If none of the referenced bundles are loading/loaded, by default the smallest bundle will be called for loading.
If developer provides `bundlesFilter` callback, then instead of smallest bundle by default, the developer can pick a specific bundle from a list of referenced bundles. If nothing is returned, default behavior prevails.

**This solution will not require any change in engine lazy-loading behaviors across components. It will load non-preloaded assets which are in bundles as expected without throwing an error (current behavior). And provide control to the developer if such needed.**

### Use Cases:

Currently to take bundles into account:
```js
const asset = app.assets.get(42);

asset.ready(() => {
    // asset is loaded
});

if (!asset.loading && !asset.loaded) {
    const bundles = app.bundles.listBundlesForAsset(asset);
    if (bundles) {
        asset.loading = true;
        let bundle;
        
        for(let i = 0; i < bundles.length; i++) {
            // if any of bundles is already loading
            // then do not try to load any other bundle
            if (bundles[i].loading) {
                bundle = null;
                break;
            }
            bundle = bundles[i];
        }
        
        if (bundle)
            app.assets.load(bundle); // start loading bundle
    } else {
        app.assets.load(asset);
    }
}
```

With new behavior, basic lazy-loading will work regardless of in-bundle or not:
```js
const asset = app.assets.get(42);

asset.ready(() => {
    // asset is loaded
});

app.assets.load(asset);
```

If a developer need to force load asset not from bundle:
```js
app.assets.load(asset, { bundlesIgnore: true });
```

If a developer need to pick which bundle to load asset from:
```js
app.assets.load(asset, {
    bundlesFilter: (bundles) => {
        const ind = Math.floor(bundles.length * Math.random()); // random bundle just for example
        return bundles[ind];
    }
});
```

Maksims:
Fixes #5626

Refactor and improve behavior of Bundles:

1. Loading asset is now consistent and behaves the same regardless if it is in a bundle or not. As it will trigger bundle loading.
2. Loading bundles now uses `fetch` instead of `XMLHttpRequest`. This provides a source for a `ReadableStream`.
3. Bundle tar is now stream parsed, so its assets trigger `load` as they individually are loaded before whole bundle is downloaded. This spreads parsing time and improves overall loading speed.
4. Various text assets (e.g. `text`, `json`, `template`, ..) now do not create Blob, which does not pollute network tab with Blob url requests and loads such assets at sync speed when it gets available from bundle. Similar approach can be applied to other binary formats (separate PR).
5. Loading a bundle now will trigger `load` for all of its assets once they are downloaded.
6. As loading and untar'ing a bundle takes little time, this PR removed WebWorker path.
7. When loading an asset, by default if there are suitable bundles, then loaded bundles are prioritized first, then not-loaded bundles sorted by size (smallest first). This logic can be controlled by a developer using `bundlesFilter`.

This makes working with bundles much easier, and does not require any changes to existing projects that use bundles. It actually provides a simplification so developers can load async bundles and its assets much easier without custom code.
As well as slightly speeds up the loading as it spreads parsing as it downloads.

### New APIs:
```javascript
// pc.AssetRegistry
assets.load(asset, { // optional options
    bundlesIgnore: true, // force asset loading not from a bundle
    bundlesFilter: (bundles) => { // if there are suitable bundles, then this will be called provided all suitable bundles
        return bundles[0]; // you can control which bundle to load asset from
    }
})
```

### In case of no cache, loading is 18-35% faster depending on latency.
Worse latency = more benefits from bundles.
Larger bundles (more assets in them) = more benefit from bad latency.
Preloaded vs post-loaded assets improvements might vary depending on bundles contents.
When loading from cache, there is no effect (marginal error).


willeastcott:
Is is possible to have some perf numbers in the PR description please? I think the concern before was the bundles didn't reliably give better startup times than unbundled assets. So if we can now demonstrate that, I think I'd be much happier to promote bundle usage to users.

marklundin:
Given the example:

```

With It's not immediately clear what the expected behaviour would be, which I assume is that ``

**Comment by marklundin** [comment]
Given the following example, it's not totally clear to me whether `bundlesIgnore` takes precedence or if it loads from the first bundle.  

```javascript
// pc.AssetRegistry
assets.load(asset, {
    bundlesIgnore: true,
    bundlesFilter: (bundles) => { // by default if multiple assets contain a bundle, it chooses either already loaded bundle, or if none available the smallest file size bundle
        return bundles[0]; // you can provide custom method to control this logic
    }
})
```

This could be potentially be simplified with something along the following lines

```javascript
// pc.AssetRegistry
assets.load(asset, { // optional options object
    useBundle: (bundles) => { // by default (undefined) or if true and multiple assets contain a bundle, it chooses either already loaded bundle, or if none available the smallest file size bundle. If false load asset directly
      return bundles[0] // you can provide custom method to control this logic, if falsy defaults to logic as if useBundle: false 
   }
})
```

Examples would be:

```javascript 
assets.load(asset) // Asset loads from bundle if it exists in one
assets.load(asset, { useBundle : true }) // Same as above
assets.load(asset, { useBundle : false }) // Asset does not loads from bundle and loads directly
assets.load(asset, { useBundle : bundle => bundles[0] }) // Asset loads from specific bundle if it exists
```

Maksims:
Below are benchmarks of a specific project that I believe is a good case for bundles.  
It has some preload assets to have menu screen with some interaction (renders at that moment). And post loads more assets so the application is fully playable. Assets (/files/ directory) are hard cached (will load from local cache).
Timings start from when DOM is loaded.
Averages from 5 runs on each case.

## Conclusions:

### In case of no cache, loading is 18-35% faster depending on latency.  
Worse latency = more benefits from bundles.
Larger bundles (more assets in them) = more benefit from bad latency.
When loading from cache, there is no effect (marginal error).

### Test Types:

no bundles, no cache: 312 requests, 12.3Mb download
no bundles, cache: 309 requests, 4.1Kb download
bundles, no cache: 65 requests (163 blobs), 12.4Mb download
bundles, cache: 62 requests (163 blobs), 4.1Kb download

### Timing Names:
**First Frame** - when first `update` is called.
**Renderable** - when assets are preloaded and shaders compiled (second frame usually).
**Playable** - when preload and post load assets are loaded and app is fully playable.

## Timings (good latency, Latvia - Frankfurt):

| Platform | Bundles | Cache | First Frame | Renderable | Playable | Advantage (higher is better) |
| --- | --- | --- | --- | --- | --- | --- |
| PC (fibre) | | |3.46s | 4.63s | 5.71s | - |
| PC (fibre) | :white_check_mark: | | 2.32s | 3.60s | 4.63s | +19% |
| PC (fibre) | | :white_check_mark: | 1.83s | 3.14s | 3.81s | - |
| PC (fibre) | :white_check_mark: | :white_check_mark: | 1.83s | 3.10s | 3.96s | -4% |
| Mobile (5G) | | | 5.45s | 7.69s | 9.42s | - |
| Mobile (5G) | :white_check_mark: | | 4.19s | 6.52s | 7.69s | +18%  |
| Mobile (5G) | | :white_check_mark: | 3.46s | 5.82s | 6.88s | - |
| Mobile (5G) | :white_check_mark: | :white_check_mark: | 3.88s | 6.21s | 7.32s | -6% |
| Mobile (4G) | | | 6.62s | 8.93s | 11.45s | - |
| Mobile (4G) | :white_check_mark: | | 4.49s | 7.3s | 9.24s | +19% |
| Mobile (3G) | | | 7.84s | 10.11s | 14.46s | - |
| Mobile (3G) | :white_check_mark: | | 6.42s | 8.7s | 13.88s | +4% |

## Timings (bad latency 150ms+, Latvia - San Francisco):

| Platform | Bundles | Cache | First Frame | Renderable | Playable | Advantage (higher is better) |
| --- | --- | --- | --- | --- | --- | --- |
| PC (fibre) | | | 7.84s | 9.02s | 12.06s | - |
| PC (fibre) | :white_check_mark: | | 4.09s | 5.45s | 7.81s | +35% |
| PC (fibre) | | :white_check_mark: | 1.87s | 3.11s | 3.96s | - |
| PC (fibre) | :white_check_mark: | :white_check_mark: | 1.98s | 3.3s | 3.89s | +2% |
| Mobile (4G) | | | 13.31s | 15.6s | 20.67s | - |
| Mobile (4G) | :white_check_mark: | | 7.89s | 10.17s | 13.39s | +35% |

Maksims:
> ```js
> bundlesIgnore
> ```

When `bundlesIgnore` is provided, then `bundlesFilter` is not called as there are no bundles to filter. Same happens if asset is not in a bundle, and `bundlesFilter` is provided - it will not be called, as there are no bundles to filter.
`bundlesFilter` is called only when there are suitable bundles (that contain asset) are provided, and custom behavior is required (I still haven't came up with the case tbh).
Within the engine, we try to avoid multi-typed arguments (`boolean` | `function`), as it makes it harder to learn and IDE tooltips are harder to read.

Here is current PRs ways:

```javascript
// default to be loaded from bundle if any is suitable
assets.load(asset);

// force asset to be loaded not from a bundle
assets.load(asset, {
    bundlesIgnore: true
});

// choose a random bundle (if any suitable) to load from
assets.load(asset, {
    bundlesFilter: (bundles) => {
        const ind = Math.ceil(Math.random() * bundles.length);
        return bundles[ind];
    }
});
```

src/framework/bundle/bundle.js
```diff
@@ -1,23 +1,59 @@
...
-            if (files[i].url) {
-                this._blobUrls[files[i].name] = files[i].url;
-            }
-        }
+    /**
+     * Add file to a Bundle.
+     *
+     * @param {string} url - A url of a file.
+     * @param {DataView} data - A DataView of a file.
+     * @internal
```
willeastcott, line 50: AFAIK, there is not `@internal` tag. I guess you meant `@ignore`?

src/framework/bundle/bundle-registry.js
```diff
@@ -4,89 +4,104 @@
  * @ignore
  */
 class BundleRegistry {
+    /**
+     * index of bundle assets
```
willeastcott, line 8:
```suggestion
     * Index of bundle assets.
```

src/framework/bundle/bundle-registry.js
```diff
@@ -4,89 +4,104 @@
...
+     * @type {Map<number, import('../asset/asset.js').Asset>}
+     * @private
+     */
+    _idToBundle = new Map();
+
+    /**
+     * index of asset id to set of bundle assets
```
willeastcott, line 15:
```suggestion
     * Index of asset id to set of bundle assets.
```

src/framework/bundle/bundle-registry.js
```diff
@@ -4,89 +4,104 @@
...
+     * @type {Map<number, Set<import('../asset/asset.js').Asset>>}
+     * @private
+     */
+    _assetToBundles = new Map();
+
+    /**
+     * index of file urls to set of bundle assets
```
willeastcott, line 22:
```suggestion
     * Index of file url to set of bundle assets.
```

src/framework/bundle/bundle-registry.js
```diff
@@ -4,89 +4,104 @@
...
+     * @type {Map<string, Set<import('../asset/asset.js').Asset>>}
+     * @private
+     */
+    _urlsToBundles = new Map();
+
+    /**
+     * index of file requests to load callbacks
```
willeastcott, line 29:
```suggestion
     * Index of file request to load callbacks.
```

src/framework/bundle/bundle.js
```diff
@@ -1,23 +1,59 @@
...
 /**
- * Represents the resource of a Bundle Asset, which contains an index that maps URLs to blob URLs.
+ * Represents the resource of a Bundle Asset, which contains an index that maps URLs to DataViews.
  *
  * @ignore
  */
-class Bundle {
+class Bundle extends EventHandler {
+    /**
+     * index of file urls to to DataView
```
willeastcott, line 10:
```suggestion
     * Index of file url to to DataView
```

src/framework/bundle/bundle.js
```diff
@@ -1,23 +1,59 @@
...
+     * @type {Map<string, DataView>}
+     * @private
+     */
+    _index = new Map();
+
+    /**
+     * if Bundle has all files loaded
```
willeastcott, line 17:
```suggestion
     * If Bundle has all files loaded.
```

src/framework/bundle/bundle.js
```diff
@@ -27,29 +63,42 @@ class Bundle {
...
     destroy() {
-        for (const key in this._blobUrls) {
-            URL.revokeObjectURL(this._blobUrls[key]);
-        }
-        this._blobUrls = null;
+        this._index.clear();
+    }
+
+    /**
+     * True if all files of a Bundle are loaded
```
willeastcott, line 89:
```suggestion
     * True if all files of a Bundle are loaded.
```

src/framework/handlers/untar.js
```diff
@@ -1,369 +1,202 @@
...
-                file[fieldName] = fieldValue;
-            }
-        }
-    };
+import { EventHandler } from '../../core/event-handler.js';
 
+/**
+ * An utility class for un-tar'ing archives, from a fetch requests.
+ * It processes files in tar in streamed manned, so assets parsing
+ * can happen in parallel instead of all at once at the end.
```
willeastcott, line 6:
```suggestion
 * A utility class for untaring archives, from a fetch request. It processes files from a tar files
 * in a streamed manner, so asset parsing can happen in parallel instead of all at once at the end.
```

src/framework/handlers/untar.js
```diff
@@ -1,369 +1,202 @@
...
+            this.pump(res.done, res.value);
+        }).catch((err) => {
+            this.fire('error', err);
+        });
     }
 
     /**
-     * @param {MessageEvent} e - The message event from the worker.
-     * @private
+     * This method is called multiple times when stream provides a data.
```
willeastcott, line 110:
```suggestion
     * This method is called multiple times when the stream provides data.
```

src/framework/handlers/untar.js
```diff
@@ -1,369 +1,202 @@
...
     /**
-     * Untars the specified array buffer using a Web Worker and returns the result in the callback.
+     * Attempt to read file from an available data buffer
      *
-     * @param {ArrayBuffer} arrayBuffer - The array buffer that holds the tar archive.
-     * @param {Function} callback - The callback function called when the worker is finished or if
-     * there is an error. The callback has the following arguments: {error, files}, where error is
-     * a string if any, and files is an array of file descriptors.
+     * @returns {boolean} True if file were sucessfully read and potentially more
+     * data is available for processing.
```
willeastcott, line 143:
```suggestion
     * @returns {boolean} True if file was successfully read and more data is potentially available for
     * processing.
```

src/framework/handlers/untar.js
```diff
@@ -1,369 +1,202 @@
...
+        if (this.headerRead) {
+            // buffer might be not long enough
+            if (this.bytesReceived < (this.bytesRead + this.fileSize)) {
+                return false;
+            }
+
+            // normal file
+            if (this.fileType === '' || this.fileType === '0') {
+                const dataView = new DataView(this.data.buffer, this.bytesRead, this.fileSize);
+                // const blob = URL.createObjectURL(new Blob([dataView]));
```
willeastcott, line 176: Delete commented out code?

willeastcott:
Just wanted to ask another question about bundles. Obviously, these work fine in the Editor. But are they also useful for engine developers? So I believe I originally recommended the TAR format because anyone can open a terminal window and tar up some files. But is that all a bundle is? A vanilla TAR file? Or is there extra metadata that would make it hard for engine-only users to create?

Maksims:
> Just wanted to ask another question about bundles. Obviously, these work fine in the Editor. But are they also useful for engine-only developers? So I believe I originally recommended the TAR format because anyone can open a terminal window and tar up some files. But is that all a bundle is? A vanilla TAR file? Or is there extra metadata that would make it hard for engine-only users to create?

It is just a tar file. Tests run over a bundle that I've produced using command line `tar`. The only rule that `asset.file.url` of assets in bundle should match folder structure in tar. Also bundle asset should have a list of ids in `asset.data.assets`. So they are perfectly usable by engine-only users, and easy to create/manage.

mvaligursky:
It'd be great to have an engine example demonstrating its use.

Maksims:
Added engine-only example.
Described in that example a process of creating a bundle, as simple as:
```bash
cd engine/examples/
tar cvf bundle.tar assets/models/geometry-camera-light.glb assets/models/torus.png
```
Then make sure `asset.file.url` matches that tar's folder structure.

src/framework/handlers/bundle.js
```diff
@@ -41,59 +37,38 @@ class BundleHandler {
...
-                }
-            } else {
-                callback('Error loading bundle resource ' + url.original + ': ' + err);
-            }
-        });
-    }
+        fetch(url.load, {
+            mode: 'cors',
+            credentials: 'include'
+        }).then((res) => {
```
kungfooman, line 43: Does `fetch` still handle `maxRetries` (I can't find it in diff)?

slimbuck:
This PR is looking great @Maksims!

One thing I'd like to address is the new asset `openBinary` function. We already have similar functionality using `fetchArrayBuffer` which is used for loading embedded images from glb files.

`fetchArrayBuffer` will either return `asset.file.contents` if it is provided or it will download and return the file data using url in the normal way. The helper means that all callers don't to have to worry either way. This fast path results in much less packing and unpacking of data when it's already available, much like `openBinary`. One advantage with `asset.file.contents` though is that users can also provide the asset file contents.

So yeah, ideally the asset system would have just one method for this fast path that works everywhere.

Maksims:
[slimbuck](https://github.com/slimbuck) I've looked into `fetchArrayBuffer`, and could not figure out how to combine the Bundle and fetchArrayBuffer ways of opening data. First of all, opening data from DataView is different per asset type handlers, so either way we need a custom "pre-processing" per handler type, e.g. for JSON it is: `JSON.parse(this.decoder.decode(data));` where we have to decode it first, and then parse it.
Also `fetchArrayBuffer` if failed to use `asset.file.contents` will make `arraybuffer` request which is not desirable.

LeXXik:
What effect will it have on the size of the application config file? We have some apps with thousands of assets. This creates a config JSON of several megabytes. The browser trying to read such a large file is taking too long on itself, and that is without PlayCanvas application parsing it afterwards. It would be great, if bundles would be a solution to it. If each asset is still referenced in the config file, then it won't help us.

Less http requests are still good, though. We are currently mitigating those with creating archives, then unpacking and generating assets at runtime. Its only suitable for some assets, though, e.g. we can't do it for materials, etc. I suppose bundles would help us in this regard?

I guess what I am trying to ask is whether an asset of any type can be invisible to config file, live only in a bundle, and be loaded only by referencing to the correct bundle at load time?

Maksims:
Hi @LeXXik. Bundles only pack actual files in a tar. Meta data of an asset (json data in config.json) is not affected. Bundle meta json (data in config.json) only stores an asset IDs as an array, so impact on config.json size is minimal.

I had a similar challenge as you describe, and I solved it by a pre-processing config.json after build, where I used tags to mark some assets in a way that allowed me to split one config.json into multiple asset-meta.json files. Which then I loaded based ln application logic.
In another case with similar challenge, I actually used database with API to query for a lists of assets based on application logic.

Maksims:
We've tested this PR deployed with the engine in production project, and it performed all as expected.
@willeastcott let me know please what is holding a review.

src/framework/handlers/bundle.js
kpal81xd: Could probably write something like this to wrap fetch to handle retries and timeout:

```js
const fetchExtra = async (url, { retry = false, maxRetries = 5, timeout = 5000, ...options } = {}) => {
    const timeoutPromise = new Promise((_, reject) => setTimeout(() => reject(new Error('Timed out.')), timeout));
    const fetchPromise = new Promise(async (resolve, reject) => {
        if (!retry) {
            try {
                resolve(await fetch(url, options));
            } catch (e) {
                reject(e);
            }
        }

        let attempts = 0;
        while (attempts < maxRetries) {
            try {
                resolve(await fetch(url, options));
                break;
            } catch (e) {
                attempts++;
                if (attempts === maxRetries) {
                    reject(e);
                    break;
                }
            }
        }
    });

    return await Promise.race([ fetchPromise, timeoutPromise ]);
};

```

devcem:
Is there any update on this? This could be potentially increase our C2P values on our web games.

willeastcott:
Hey, @slimbuck and @mvaligursky - I want us to set aside some time this week to properly review this PR and finally get it merged. 🙏

mvaligursky:
Hi @Maksims - we're keen to get to review this / finalize this. When you get a chance, could you please merge main to it and resolve the conflicts.

src/framework/handlers/css.js
```diff
@@ -8,6 +8,14 @@ class CssHandler {
      */
     handlerType = "css";
 
+    /**
+     * TextDecoder for decoding binary data.
+     *
+     * @type {TextDecoder}
+     * @private
+     */
+    decoder = new TextDecoder('utf-8');
```
marklundin, line 17: Is there a solution here for older devices without TextDecoder?

Maksims: Which devices would that be in particular?
It has great support (97.8%): https://caniuse.com/textencoder
Only devices with software that has not been updated since 2016/18 potentially will not support it, but that is like 6-8 year outdated.
IE11 is not supported anymore already by the engine.

marklundin: The UMD build is still [targeting some fairly old browsers](https://github.com/playcanvas/engine/blob/53d2ce4824453c0eee3801dd417ba3bceb7791b1/utils/rollup-es5-options.mjs#L22-L24). It's reasonable to require a recent browser to use bundles, but the engine probably shouldn't throw an error if not. Could we lazily initialize so it only happens when a bundle is loaded?

src/framework/asset/asset-registry.js
```diff
@@ -362,16 +384,24 @@ class AssetRegistry extends EventHandler {
      *     app.assets.load(assetToLoad);
      * });
      */
-    load(asset) {
+    load(asset, options) {
         // do nothing if asset is already loaded
         // note: lots of code calls assets.load() assuming this check is present
         // don't remove it without updating calls to assets.load() with checks for the asset.loaded state
-        if (asset.loading || asset.loaded) {
+        if ((asset.loading || asset.loaded) && !options?.force) {
```
kpal81xd, line 391: Add the `[options.force]` to options JSDoc

examples/src/examples/loaders/bundle.mjs
```diff
@@ -0,0 +1,133 @@
...
+        });
+    });
+}
+
+export class BundleExample {
+    static CATEGORY = 'Loaders';
+    static NAME = 'Bundle';
+    static WEBGPU_ENABLED = true;
+    static example = example;
+}
```
kpal81xd, line 133: This example will need to be updated to match new format.

src/framework/bundle/bundle.js
```diff
@@ -1,23 +1,59 @@
...
+     * Fired when all files of a Bundle has been loaded.
      *
-     * @param {object[]} files - An array of objects that have a name field and contain a
-     * getBlobUrl() function.
+     * @event Bundle#load
+     * @example
+     * bundle.on("load", function () {
+     *     console.log("All Bundle files has been loaded");
+     * });
      */
```
kpal81xd, line 43: define each event name as a static property on the class like the rest of the engine using the form `EVENT_<name>`

src/framework/handlers/css.js
kpal81xd: Pretty sure you can easily polyfill TextDecoder such as https://stackoverflow.com/questions/40662142/polyfill-for-textdecoder

src/framework/handlers/loader.js
```diff
@@ -155,10 +184,25 @@ class ResourceLoader {
...
+                    if (options && options.bundlesFilter) {
+                        bundle = options.bundlesFilter(bundles);
+                    }
+
+                    if (!bundle) {
+                        // prioritize smallest bundle
+                        bundles?.sort((a, b) => {
+                            return a.file.size - b.file.size;
+                        });
+                        bundle = bundles[0];
```
kpal81xd, line 202: If bundles is `undefined` or `null` the sort will be fine from conditional chaining but indexing it at `0` will throw an error.

kpal81xd (changes requested):

src/framework/asset/asset-registry.js
```diff
@@ -362,16 +384,24 @@ class AssetRegistry extends EventHandler {
      *     app.assets.load(assetToLoad);
      * });
      */
-    load(asset) {
+    load(asset, options) {
```
slimbuck, line 387: asset already has an options field. why add another one here?

src/framework/asset/asset-registry.js
slimbuck, line 391: `options.force` isn't documented above

src/framework/handlers/shader.js
```diff
@@ -32,6 +40,10 @@ class ShaderHandler {
         });
     }
 
+    openBinary(data) {
```
slimbuck, line 43: Please add documentation to `handler.js` on what this function is, when it is used, what the data parameter is etc.

Maksims: Added individual functions to parsers, as not every handler can/should support openBinary method.

src/framework/asset/asset-registry.js
Maksims: Not sure I understand the question. These properties are required to pass situational flags based on where `load` is called from.

Maksims:
Updated this PR:
1. Example
2. TextDecoder is now lazily initialized same as in the other places in the engine.
3. Added docs where needed. openBinary should be added only on specific handlers where it is necessary, so not adding it to `ResourceHandler`.

Tested on existing project - works as before.

src/framework/handlers/bundle.js
kpal81xd: @Maksims the fetch method still needs to handle max retries

kpal81xd (approved):

PR #5678, opened 2023-09-26
Deprecate device.webgl2 in favor of device.isWebGL2

mvaligursky:
The consistent API on Graphics device is now:
```
isWebGPU
isWebGL2
isWebGL1
```

`.webgl2` is still function, but logs a one time warning message

### DIFF
--- a/src/scene/shader-lib/programs/lit-shader.js
+++ b/src/scene/shader-lib/programs/lit-shader.js
@@ -467,7 +467,7 @@ class LitShader {
 
         let code = this._fsGetBeginCode();
 
-        if (device.extStandardDerivatives && !device.webgl2 && !device.isWebGPU) {
+        if (device.extStandardDerivatives && device.isWebGl1) {
             code += 'uniform vec2 polygonOffset;\n';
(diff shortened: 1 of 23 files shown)


src/scene/shader-lib/programs/lit-shader.js
```diff
@@ -870,7 +870,7 @@ class LitShader {
                 func.append(chunks.shadowPCSSPS);
             }
 
-            if (!(device.webgl2 || device.extStandardDerivatives || device.isWebGPU)) {
+            if (!(device.isWebGl2 || device.isWebGPU || device.extStandardDerivatives)) {
```
willeastcott, line 873: Seems that `isWebGl2` doesn't capitalize the `L` but `isWebGPU` capitalizes the `PU`. Kinda inconsistent.

mvaligursky: true, thanks, changing to all caps, as in the PR description

willeastcott (approved):
Nice. 😄

PR #5693, opened 2023-09-28
Fire destroy event in AppBase#destroy

kungfooman:
Provides a solution for https://github.com/playcanvas/engine/pull/5555#issuecomment-1739243685

I was a bit surprised that we didn't have this yet. :thinking:

With this PR we have a simple/intuitive way to clean up events:

```js
function resize() {
    app.resizeCanvas(canvas.width, canvas.height);
}
window.addEventListener("resize", resize);
app.on('destroy', () => {
    window.removeEventListener("resize", resize);
});
```

List of stuff we always need to clean: window events, setInterval id's

### DIFF
--- a/src/framework/app-base.js
+++ b/src/framework/app-base.js
@@ -2037,6 +2037,7 @@ class AppBase extends EventHandler {
 
         this.tick = null;
 
+        this.fire('destroy', this); // fire destroy event
         this.off(); // remove all events
 
         this._soundManager?.destroy();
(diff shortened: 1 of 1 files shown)


src/framework/app-base.js
```diff
@@ -2037,6 +2037,7 @@ class AppBase extends EventHandler {
 
         this.tick = null;
 
+        this.fire('destroy', this); // fire destroy event
```
mvaligursky, line 2040: I wonder if we should fire this at the start of the `destroy` function instead of here, giving the user an option to use the AppBase instance while handling this event (for example to save some data), instead of at a point where it's mostly not valid anymore.

willeastcott: 100% agree here.

kungfooman: Yep, good point, just moved it up

mvaligursky (approved):

PR #5694, opened 2023-09-28
fix array cloning

LeXXik:
Fixes #5663

(in case @chrissterling doesn't have a chance to fix it)

### DIFF
--- a/src/scene/mesh.js
+++ b/src/scene/mesh.js
@@ -723,8 +723,8 @@ class Mesh extends RefCountedObject {
             } else {
                 // destination data is array
                 indices.length = 0;
-                for (const index of streamIndices) {
-                    indices.push(index);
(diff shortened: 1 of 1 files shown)


mvaligursky:
in most tests I see, `for` loop is consistently faster than `for of` loop, and so I'd suggest we use that instead.

https://blog.bitsrc.io/measuring-performance-of-different-javascript-loop-types-c0e9b1d193ed

or
Screenshot 2023-10-04 at 10 48 26
https://levelup.gitconnected.com/which-is-faster-for-for-of-foreach-loops-in-javascript-18dbd9ffbca9

willeastcott:
Agreed. While I like `for...of` loops in certain circumstances, they don't transpile particularly nicely with Babel. PCUI uses them a lot, for example, but we tend to avoid them for performance critical engine stuff.

PR #5702, opened 2023-10-01
New API: add a scaled vector

### LINKED ISSUE #5114: Add vector + scaled vector math
A feature request here.

I know, we don't want to bloat the vector math with convenience methods. However, this one I found I am using too often. Almost every time I do any vector math - when I am trying to add a scaled vector (without changing the original) to another vector.

Normally, one would do something like:
```js
// a + b*2;
const temp = new pc.Vec3();
temp.copy(b).mulScalar(2).add(a);
```

It requires you to allocate a temp vector, if you don't want to change the original. The only way to avoid allocation, you'd need to do a component wise writeout, which is verbose and would better be as part of the vector math methods:

```js
a.addScaled(b, scalar);

// where
a.x += b.x * scalar;
a.y += b.y * scalar;
a.z += b.z * scalar;
```

LeXXik:
Fixes #5114

Adds a new API to vectors: add a scaled vector. Adds a scaled vector to the current instance and returns itself as result. Does not modify the vector being added.

```js
const vec = new pc.Vec3();
vec.addScaled(pc.Vec3.UP, 2); // [0, 2, 0]
```


src/core/math/vec2.js
```diff
@@ -103,6 +103,27 @@ class Vec2 {
...
     }
 
+    /**
+     * Adds a 2-dimensional vector scaled by scalar value. Does not modify the vector being added.
+     *
+     * @param {Vec2} rhs - The vector to add to the specified vector.
+     * @param {*} scalar - The number to multiply the added vector with.
+     * @returns {Vec2} Self for chaining.
+     * @example
+     * var vec = new pc.Vec2(1, 2);
```
willeastcott, line 113:
```suggestion
     * const vec = new pc.Vec2(1, 2);
```

src/core/math/vec3.js
```diff
@@ -114,6 +114,28 @@ class Vec3 {
...
     }
 
+    /**
+     * Adds a 3-dimensional vector scaled by scalar value. Does not modify the vector being added.
+     *
+     * @param {Vec3} rhs - The vector to add to the specified vector.
+     * @param {*} scalar - The number to multiply the added vector with.
+     * @returns {Vec3} Self for chaining.
+     * @example
+     * var vec = new pc.Vec3(1, 2, 3);
```
willeastcott, line 124:
```suggestion
     * const vec = new pc.Vec3(1, 2, 3);
```

src/core/math/vec4.js
```diff
@@ -127,6 +127,29 @@ class Vec4 {
...
     }
 
+    /**
+     * Adds a 4-dimensional vector scaled by scalar value. Does not modify the vector being added.
+     *
+     * @param {Vec4} rhs - The vector to add to the specified vector.
+     * @param {*} scalar - The number to multiply the added vector with.
+     * @returns {Vec4} Self for chaining.
+     * @example
+     * var vec = new pc.Vec4(1, 2, 3, 4);
```
willeastcott, line 137:
```suggestion
     * const vec = new pc.Vec4(1, 2, 3, 4);
```

willeastcott:
I guess my only slight hesitation is the naming of `addScalar` and `addScaled`....just wondering if that is the best option....

LeXXik:
I've also considered `addScaledVector`, similar to threejs, but its too long for my taste, so I went with a shorter version. Suggestions welcome.

marklundin:
I've seen `addScaledVec` and `addScaledVector` in other api's which is more descriptive and easier to distinguish against `addScalar`

LeXXik:
Added a poll for voting https://github.com/playcanvas/engine/discussions/5707

willeastcott:
@marklundin It does make it clearer. But at the same time, the `VecX` APIs are so terse/succinct. Looks down the API page and it's really very concise (and guessable, I'd argue). So I'd probably err on a short option over a longer one in this case. It's just the minor clash with `addScalar` that make me pause. I'm very, very torn.

marklundin:
Yep agree, it's just that slight collision with `addScalar` that gives hesitation. However there's no real ambiguity what `addScalar` does, so maybe it's a non issue?

willeastcott:
Yes, I guess so. @mvaligursky, thoughts?

src/core/math/vec2.js
```diff
@@ -103,6 +103,27 @@ class Vec2 {
         return this;
     }
 
+    /**
+     * Adds a 2-dimensional vector scaled by scalar value. Does not modify the vector being added.
+     *
+     * @param {Vec2} rhs - The vector to add to the specified vector.
+     * @param {*} scalar - The number to multiply the added vector with.
```
mvaligursky, line 110: maybe `number` instead of `*`?

mvaligursky:
I'm happy with `addScaled` I think. even though my mind calls this `mad` from old shader assembly instruction (`multiply and add`), even though that instruction handles vector being multiplied by both scalar or another vector.

willeastcott:
OK, agreed. Once that JSDoc is updated to `number`, we can merge.

src/core/math/vec3.js
```diff
@@ -114,6 +114,28 @@ class Vec3 {
         return this;
     }
 
+    /**
+     * Adds a 3-dimensional vector scaled by scalar value. Does not modify the vector being added.
+     *
+     * @param {Vec3} rhs - The vector to add to the specified vector.
+     * @param {*} scalar - The number to multiply the added vector with.
```
mvaligursky, line 121: and this

src/core/math/vec4.js
```diff
@@ -127,6 +127,29 @@ class Vec4 {
         return this;
     }
 
+    /**
+     * Adds a 4-dimensional vector scaled by scalar value. Does not modify the vector being added.
+     *
+     * @param {Vec4} rhs - The vector to add to the specified vector.
+     * @param {*} scalar - The number to multiply the added vector with.
```
mvaligursky, line 134: and even this please :)

mvaligursky (approved):
Thanks!

PR #5704, opened 2023-10-02
Add RenderTarget|null typing + remove dead code

kungfooman:
This fixes a few of type issues, since `RenderTarget` is `null` in many situations, and then it goes on and on to e.g. `drawQuadWithShader` which doesn't need a `target`, e.g. example with bloom:

image

### DIFF
--- a/src/platform/graphics/render-pass.js
+++ b/src/platform/graphics/render-pass.js
@@ -192,15 +192,15 @@ class RenderPass {
     }
 
     /**
-     * @param {import('../graphics/render-target.js').RenderTarget|null} renderTarget - The render
+     * @param {import('../graphics/render-target.js').RenderTarget|null} [renderTarget] - The render
      * target to render into (output). This function should be called only for render passes which
      * use render target, or passes which render directly into the default framebuffer, in which
      * case a null or undefined render target is expected.
      */
-    init(renderTarget) {
+    init(renderTarget = null) {
 
         // null represents the default framebuffer
-        this.renderTarget = renderTarget || null;
+        this.renderTarget = renderTarget;
 
         // defaults depend on multisampling
         this.samples = Math.max(this.renderTarget ? this.renderTarget.samples : this.device.samples, 1);
--- a/src/platform/graphics/render-pass.js
+++ b/src/platform/graphics/render-pass.js
@@ -93,7 +93,13 @@ class RenderPass {
     /** @type {string} */
     name;
 
-    /** @type {import('../graphics/render-target.js').RenderTarget|null|undefined} */
+    /**
+     * The render target for this render pass:
+     *  - `undefined`: render pass does not render to any render target
+     *  - `null`: render pass renders to the backbuffer
+     *  - Otherwise, renders to the provided RT.
+     * @type {import('../graphics/render-target.js').RenderTarget|null|undefined}
+     */
     renderTarget;
 
     /**
--- a/src/platform/graphics/render-pass.js
+++ b/src/platform/graphics/render-pass.js
@@ -93,7 +93,7 @@ class RenderPass {
     /** @type {string} */
     name;
 
-    /** @type {import('../graphics/render-target.js').RenderTarget} */
+    /** @type {import('../graphics/render-target.js').RenderTarget|null} */
     renderTarget;
 
     /**
@@ -186,7 +186,7 @@ class RenderPass {
     }
 
     /**
-     * @param {import('../graphics/render-target.js').RenderTarget} renderTarget - The render
+     * @param {import('../graphics/render-target.js').RenderTarget|null} renderTarget - The render
      * target to render into (output). This function should be called only for render passes which
      * use render target, or passes which render directly into the default framebuffer, in which
      * case a null or undefined render target is expected.
--- a/src/platform/graphics/render-pass.js
+++ b/src/platform/graphics/render-pass.js
@@ -93,7 +93,7 @@ class RenderPass {
     /** @type {string} */
     name;
 
-    /** @type {import('../graphics/render-target.js').RenderTarget|null} */
+    /** @type {import('../graphics/render-target.js').RenderTarget|null|undefined} */
     renderTarget;
 
     /**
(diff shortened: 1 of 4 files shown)


src/scene/graphics/quad-render-utils.js
```diff
@@ -68,8 +68,7 @@ function drawQuadWithShader(device, target, shader, rect, scissorRect) {
     // the same framebuffer. The workaround here is to store multi-sampled color buffer, instead of only resolving it, which is wasted
     // memory bandwidth. Without this we end up with a black result (or just UI), as multi-sampled color buffer is never written to.
     if (device.isWebGPU && target === null) {
-        const samples = target?.samples ?? device.samples;
-        if (samples > 1)
+        if (device.samples > 1)
```
kungfooman, line 71: This is/was dead code, we are testing `target?.samples` even though the `if` makes sure that `target === null`

slimbuck: nice catch, now might as well collapse the nested ifs?

kungfooman: > nice catch, now might as well collapse the nested ifs?

nice observation, gonna do that

kungfooman:
We have in some functions strict "no-undefined-allowed" assertions:

```js
function drawQuadWithShader(device, target, shader, rect, scissorRect) {
    // a valid target or a null target (framebuffer) are supported
    Debug.assert(target !== undefined);
```

But then we still have places where `undefined` creeps in again:

image

Should we try to keep it `RenderTarget|null` everywhere? From a clean type perspective, it makes most sense to me here :thinking:

slimbuck (approved):

slimbuck:
> Should we try to keep it RenderTarget|null everywhere? From a clean type perspective, it makes most sense to me here 🤔

Probably a question for @mvaligursky

src/platform/graphics/render-pass.js
```diff
@@ -93,7 +93,7 @@ class RenderPass {
     /** @type {string} */
     name;
 
-    /** @type {import('../graphics/render-target.js').RenderTarget} */
+    /** @type {import('../graphics/render-target.js').RenderTarget|null} */
```
mvaligursky, line 96: this can be `undefined` as well, which means 'not rendering to any render target'. `null` means 'rendering the backbuffer'.

mvaligursky:
> Should we try to keep it RenderTarget|null everywhere? From a clean type perspective, it makes most sense to me here

`undefined` is valid in some cases, so we cannot force it to null in those cases.

kungfooman:
> `undefined` is valid in some cases, so we cannot force it to null in those cases.

Interesting, so I tried to find the these methods called with `undefined` as `renderTarget` in a running app (UI ScrollView example):

I found these two:

https://github.com/playcanvas/engine/blob/bbb79618ccdd7dcdf8db1230c27d638e07c16e51/src/platform/graphics/render-pass.js#L188-L221

https://github.com/playcanvas/engine/blob/bbb79618ccdd7dcdf8db1230c27d638e07c16e51/src/framework/components/camera/component.js#L905-L917

If these are called correctly, I can just mark them as `[renderTarget]` (meaning `|undefined`).

I feel uncertain here though, in first code snippet, line 197:

```js
this.renderTarget = renderTarget || null; 
```

So it could be called with `undefined` but will be immediately updated to `null` in case of `undefined`? Would be really nice to get the types right here, because they cause type invalidations in RuntimeTypeInspector.js every frame:

image

src/platform/graphics/render-pass.js
kungfooman: Thank you! I added `undefined` aswell in next commit

mvaligursky:
So the `RendePass.renderTarget` is undefined if nothing is set to it, so that is the default uninitialized state. 
if `RenderPass.init` is called, it can only update that member to a valid rt, or NULL (backbuffer), but never to uninitialized.

so only renderTarget needs to have all 3 types allowed, but init function only needs two.

src/platform/graphics/render-pass.js
```diff
@@ -93,7 +93,7 @@ class RenderPass {
     /** @type {string} */
     name;
 
-    /** @type {import('../graphics/render-target.js').RenderTarget} */
+    /** @type {import('../graphics/render-target.js').RenderTarget|null|undefined} */
```
mvaligursky, line 96: feel free to document these types here for clarity, as it's not obvious:
- `undefined`: render pass does not render to any render target
- `null`: render pass renders to the backbuffer
- `otherwise`: renders to the provided RT

src/platform/graphics/render-pass.js
```diff
@@ -186,7 +186,7 @@ class RenderPass {
     }
 
     /**
-     * @param {import('../graphics/render-target.js').RenderTarget} renderTarget - The render
+     * @param {import('../graphics/render-target.js').RenderTarget|null} renderTarget - The render
```
mvaligursky, line 189: so yeah, here only the optional RenderTarget type should be provided as you suggested.

src/platform/graphics/render-pass.js
kungfooman: Nice, added some markdown to it:

image

kungfooman:
> if RenderPass.init is called, it can only update that member to a valid rt, or NULL (backbuffer), but never to uninitialized.
>
> so only renderTarget needs to have all 3 types allowed, but init function only needs two.

Hm, we still call `RenderPass#init` with `undefined` aswell, shouldn't we then just do `init(renderTarget = null)` so the type system picks up that calling `init` with `undefined` is allowed?

image

(this happens in `ForwardRenderer#addMainRenderPass`)

mvaligursky:
> init(renderTarget = null)

perfect! Go for it.

src/platform/graphics/render-pass.js
kungfooman: Nice, changed this as discussed below here via `[renderTarget]` and `init(renderTarget = null)` in https://github.com/playcanvas/engine/pull/5704/commits/12272f20eb1954077bbbd9c5fe054fddfec9687c

mvaligursky (approved):
Nice one, thanks!

PR #5706, opened 2023-10-03
Minor fix: there is no need to bind _addLayer

querielo:
Because it is called with **this** as its context.


willeastcott (approved):
Well spotted, @querielo! Thank you. 🙏

PR #5710, opened 2023-10-04
Don't create WebglShaderInput if there is no attribute

AndrewRayCode:
`webgl-shader.js` has this check:

```js
if (definition.attributes[info.name] === undefined) {
    console.error(`Vertex shader attribute "${info.name}" is not mapped to a semantic in shader definition, shader [${shader.label}]`, shader);
    shader.failed = true;
}
```

Right after that if statement, regardless of if the attribute is defiend or not, it calls:

```js
const shaderInput = new WebglShaderInput(device, definition.attributes[info.name], device.pcUniformType[info.type], location);
```

We've aleady established `definition.attributes[info.name]` is undefined, so this line probably shouldn't be called. Inside `WebglShaderInput`, `definition.attributes[info.name]` becomes the `name` parameter, and the function tries to call
`name.substring(name.length - 3)`, which errors.

This whole code path can fire if you define a custom vertex shader with the code

```glsl
in vec2 vertex_texCoord0;
```

And no diffuse map set, so PlayCanvas doesn't think there's an attribute in the shader definition, and tries to error looping over the attributes returned from webgl.

Fixes #

This PR prevents the runtime Javascript error. Playcanvas already logs an error in this case, but then the existing code path crashes.

### DIFF
--- a/src/platform/graphics/webgl/webgl-shader.js
+++ b/src/platform/graphics/webgl/webgl-shader.js
@@ -346,11 +346,10 @@ class WebglShader {
             if (definition.attributes[info.name] === undefined) {
                 console.error(`Vertex shader attribute "${info.name}" is not mapped to a semantic in shader definition, shader [${shader.label}]`, shader);
                 shader.failed = true;
+            } else {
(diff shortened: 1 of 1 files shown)


mvaligursky (approved):
Great fix, thanks!

PR #5715, opened 2023-10-04
WebGPU texture anisotropy validation fix

slimbuck:
According to the WebGPU spec anisotropy can only be enabled when filtering is linear for both min and mag.

This PR adds changes the WebGPU sampler builder to only set valid anisotropy value.

### DIFF
--- a/src/platform/graphics/webgpu/webgpu-texture.js
+++ b/src/platform/graphics/webgpu/webgpu-texture.js
@@ -284,9 +284,12 @@ class WebgpuTexture {
                 }
             }
 
-            // ensure anisotropy is only used only when filtering is correctly
+            // ensure anisotropic filtering is only set when filtering is correctly
             // set up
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

src/platform/graphics/webgpu/webgpu-texture.js
```diff
@@ -285,6 +284,12 @@ class WebgpuTexture {
                 }
             }
 
+            // ensure anisotropy is only used only when filtering is correctly
+            // set up
+            descr.maxAnisotropy = (descr.minFilter === 'linear' && descr.magFilter === 'linear') ?
```
mvaligursky, line 289: based on the spec, mip filter needs to be linear too

image

slimbuck: indeed - thanks!

mvaligursky (approved):

PR #5738, opened 2023-10-10
Refactoring to LayerComposition internal implementation

mvaligursky:
- LayerCompositions has additional maps to look up index of sublayers, plus few related optimisations
- RenderAction no longer stores sublayer index
- LayerComposition is updated when a layer.enable state changes - so we no longer need to handle this during rendering
- Mesh Instance culling using cameras / layers directly instead of composition / RenderActions, to be easy to use with custom RenderPassRenderAction instances that are not part of the composition.

### DIFF
--- a/src/scene/composition/render-action.js
+++ b/src/scene/composition/render-action.js
@@ -8,7 +8,7 @@ class RenderAction {
     constructor() {
 
         // the layer
-        /** @type {import('../layer.js').Layer} */
+        /** @type {import('../layer.js').Layer|null} */
         this.layer = null;
 
         // true if this uses transparent sublayer, opaque otherwise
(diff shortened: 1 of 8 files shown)


src/scene/composition/render-action.js
```diff
@@ -7,12 +7,13 @@
 class RenderAction {
     constructor() {
 
-        // index into a layer stored in LayerComposition.layerList
-        this.layerIndex = 0;
-
         // the layer
+        /** @type {import('../layer.js').Layer} */
```
willeastcott, line 11:
```suggestion
        /** @type {import('../layer.js').Layer|null} */
```

willeastcott (approved):

PR #5741, opened 2023-10-11
Custom render passes on a camera

mvaligursky:
- an API (private for now) on the Camera to allow render passes for its rendering to be added to it.
- those passes are then inserted into the frame graph by forward renderer, allowing rendering the camera using custom render passes.

Suggestion: Hide whitespaces when reviewing.

### DIFF
--- a/src/scene/renderer/render-pass-shadow-local-clustered.js
+++ b/src/scene/renderer/render-pass-shadow-local-clustered.js
@@ -7,7 +7,7 @@ import { RenderPass } from "../../platform/graphics/render-pass.js";
  * @ignore
  */
 class RenderPassShadowLocalClustered extends RenderPass {
-    constructor(device, shadowRenderer, shadowRendererLocal, localLights) {
+    constructor(device, shadowRenderer, shadowRendererLocal) {
         super(device);
 
         this.requiresCubemaps = false;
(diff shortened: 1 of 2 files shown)


src/scene/composition/render-action.js
```diff
@@ -59,6 +62,12 @@ class RenderAction {
         this.viewBindGroups.length = 0;
     }
 
+    setupClears(camera, layer) {
+        this.clearColor = camera?.clearColorBuffer || layer.clearColorBuffer;
```
slimbuck, line 66: I assume `||` is correct here instead of `??`?

src/scene/frame-graph.js
```diff
@@ -24,7 +24,28 @@ class FrameGraph {
      */
     addRenderPass(renderPass) {
         Debug.assert(renderPass);
-        this.renderPasses.push(renderPass);
+
+        const passes = this.renderPasses;
+
+        const beforePasses = renderPass.beforePasses;
```
slimbuck, line 30: Why not?
`passes = renderPass.beforePasses.filter(p => p.enabled)`

slimbuck (approved):

src/scene/composition/render-action.js
mvaligursky: both work here equally.

src/scene/frame-graph.js
mvaligursky: this runs every frame, so performance is the reason
https://plainenglish.io/blog/are-for-loops-better-than-arrays-filter-or-foreach-methods

PR #5745, opened 2023-10-12
Allow the AppBase to be cleanly destroyed even when not initialized

mvaligursky:
avoiding undefined access

### DIFF
--- a/src/framework/app-base.js
+++ b/src/framework/app-base.js
@@ -264,7 +264,10 @@ class AppBase extends EventHandler {
     init(appOptions) {
         const device = appOptions.graphicsDevice;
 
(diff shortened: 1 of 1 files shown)


mvaligursky:
Closing as no go. I was hoping to avoid construction / destruction without a device, but it's more complicated than expected and not worth it.

kungfooman:
Can't we just fix this error?

image

Basically only what you added already:

```js
        const canvasId = this.graphicsDevice?.canvas?.id;
        if (canvasId !== undefined) {
            AppBase._applications[canvasId] = null;
        }
```

mvaligursky:
fair enough, I'll do that that.
new PR.

PR #5746, opened 2023-10-12
Allow AppBase to be destroyed if GraphicsDevice does not have a canvas

mvaligursky:
discussed here https://github.com/playcanvas/engine/pull/5745#issuecomment-1759171148

### DIFF
--- a/src/framework/app-base.js
+++ b/src/framework/app-base.js
@@ -1927,7 +1927,7 @@ class AppBase extends EventHandler {
             return;
         }
 
-        const canvasId = this.graphicsDevice.canvas.id;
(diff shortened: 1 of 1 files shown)


kungfooman:
Hm, we still can't call `app.destroy()` twice:

image

The first PR felt already like the solution, just without the change in `AppBase#init` maybe?

image

mvaligursky:
maybe just don't call destroy if destroyed already

kungfooman:
> maybe just don't call destroy if destroyed already

I will just go for this then:

```js
if (app.graphicsDevice) {
    app.destroy();
}
```

slimbuck (approved):

PR #5747, opened 2023-10-12
Set model matrix when instancing

slimbuck:
Don't skip setting model matrix when instancing is used.


willeastcott (approved):

mvaligursky (approved):
Default implementation of instancing does not use the matrix, so that should cause no compatibility issues.

PR #5757, opened 2023-10-16
Initial framework for compute shaders on WebGPU

mvaligursky:
Initial framework, allowing creation and dispatch of WebGPU computer shaders. Note that support for passing data in / out of the compute shader is not yet supported.

All API is currently private, but expected public API:
- `Shader` options object passed to constructor accepts compute shader source code
- `Compute` - new class wrapping an instance of the compute shader along with its parameters (buffers, textures)

Example of its use:
```
const shader = new Shader(device, {
    name: 'ComputeShader',
    shaderLanguage: SHADERLANGUAGE_WGSL,
    cshader: `
        @compute @workgroup_size(8, 8)
        fn main(@builtin(global_invocation_id) global_id : vec3u) {
        }
    `
});

const compute = new Compute(device, shader);
compute.dispatch(4);
```

### DIFF
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -105,6 +105,13 @@ export const TRACEID_BINDGROUPFORMAT_ALLOC = 'BindGroupFormatAlloc';
 export const TRACEID_RENDERPIPELINE_ALLOC = 'RenderPipelineAlloc';
 
 /**
+ * Logs the creation of compute pipelines. WebBPU only.
+ *
+ * @type {string}
+ */
+export const TRACEID_COMPUTEPIPELINE_ALLOC = 'ComputePipelineAlloc';
+
+/**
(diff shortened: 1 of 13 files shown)


src/core/constants.js
```diff
@@ -104,6 +104,13 @@ export const TRACEID_BINDGROUPFORMAT_ALLOC = 'BindGroupFormatAlloc';
  */
 export const TRACEID_RENDERPIPELINE_ALLOC = 'RenderPipelineAlloc';
 
+/**
+ * Logs the creation of compute pipelines. WebBPU only.
```
slimbuck, line 108:
```suggestion
 * Logs the creation of compute pipelines. WebGPU only.
```

slimbuck (approved):

PR #5758, opened 2023-10-17
Buffer update speedup

slimbuck:
When updating an existing GL buffer it is faster to call `bufferSubData` instead of `bufferData` (which creates a new gl-side buffer each time).

This change roughly halves the time taken to update splat buffers.


willeastcott (approved):

willeastcott:
Wow, these are the kinds of speedups I ❤️

mvaligursky (approved):
Great change!

PR #5759, opened 2023-10-17
Use subimage versions during texture update

slimbuck:
This PR updates the texture upload logic similar to #5758.

Updating the data of an existing gl-side texture is relatively slow using `gl.texImage2D` because a new gl-side texture is created every time. It's faster to use `gl.texSubImage2D` instead, which will update the existing texture data.

This PR requires thorough testing, especially with editor projects, to ensure edge cases are correctly covered.

Testing Notes:
- each creation and update path needs testing: 2d and cubemap, pixel data and browser interface sources

### DIFF
--- a/src/platform/graphics/webgl/webgl-texture.js
+++ b/src/platform/graphics/webgl/webgl-texture.js
@@ -59,6 +59,8 @@ class WebglTexture {
 
     _glPixelType;
 
+    _glCreated;
+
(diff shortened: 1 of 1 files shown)


marklundin:
This is great @slimbuck, also I did not know this, so I've learned something new 👏

mvaligursky (approved):

Issue #5001, opened 2023-01-25
Implement GLTF exporter improvements


mvaligursky:
- full support for the material properties export (currently only few common material properties are exported).
- export of material properties that map into glTF extensions
- export skins
- export animations

LeXXik:
We are missing this feature. Bumping up!

leonidaspir:
Indeed full glTF exporting is quite useful and we need in a number of cases. 🙏

willeastcott:
I'm not sure when we will be able to schedule more work on the glTF exporter...but I would love to improve it when time allows.

@LeXXik - what are the specific features you would like added first?

If any of our open-source contributors want to tackle any of these missing features, we'd appreciate it!

slimbuck:
Hey @LeXXik, @leonidaspir,

Can you explain where you're using glTF export atm?

Thanks!

LeXXik:
Hi, guys. We are not currently using it, because it is lacking the features we need. We want to export a user-customized character, together with its materials and animations, so that we can import it back later.

willeastcott:
Just to update this issue, `GltfExporter` now supports all material extensions! 🎉

tanaydimri:
Hey all,
Really excited to see this initiative has progressed quite a bit. This and USDZ exporter has been the primary reason that has kept us from using Playcanvas in our primary stack. 
I gave the new API a try the other day and all else seemed to have exported just fine, except for the fact that seconds UV channels were being ignored. We usually author all our assets with dedicated seconds UV channel 1 for occlusion. 
Wondering if it would be okay for me to submit a PR for this?

Great to be back on the playcanvas forums as always!!

mvaligursky:
> Wondering if it would be okay for me to submit a PR for this?

absolutely, please do!
 
> Great to be back on the playcanvas forums as always!!

and welcome. also our discord is very popular, join there too if you're interested.

PR #5762, opened 2023-10-18
(WIP) glTF export: improve material export

querielo:
This pull request improves the material export in the glTF exporter.

* added support for exporting normal maps, metalness maps, and emissive maps.
* added support for exporting textures with texture transforms (tiling, offset, and rotation). 
* added support for exporting alpha test and alpha cutoff values for materials that use the "none" blend mode.

Please review. Thank you!

Screenshot 2023-11-10 at 14 18 22


extras/exporters/gltf-exporter.js
```diff
@@ -350,13 +391,25 @@ class GltfExporter extends CoreExporter {
...
-                    material.alphaMode = "BLEND";
+                    material.alphaMode = 'BLEND';
+                } else if (blendType === BLEND_NONE) {
+                    if (alphaTest !== 0) {
+                        material.alphaMode = 'MASK';
+                        material.alphaCutoff = alphaTest;
+                    } else {
+                        // It is a default value, so we don't need to write it
+                        // material.alphaMode = 'OPAQUE';
+                    }
```
willeastcott, line 402: Personally, I would just delete the else clause here.

willeastcott (approved):
This is a great contribution! It's great to see the glTF exporter get some love. Hey, @marklundin, got any idea how we can better make use of the glTF exporter in the Editor? Maybe a context menu option when right clicking on entities in the Hierarchy view?

mvaligursky:
related to https://github.com/playcanvas/engine/issues/5001

extras/exporters/gltf-exporter.js
querielo: Done

mvaligursky:
The code assumes the textures have their channels matching the glTF specification here. We should test for that (materials in PlayCanvas specify which channel is used) and only export the textures if the channels match.

Ideal solution in the future would be that if the channels do not match, we build the required textures from source textures.

Also, the normal map can have this format, which we should probably warn about and not export either, till we have code to convert it:

https://github.com/playcanvas/engine/blob/47d1f4c24b13ce15877b5cd79aaa204fa3abab92/src/platform/graphics/constants.js#L1049-L1056

The goal here being that we can only export valid glTF and print a warning otherwise.

marklundin:
Yep great point @willeastcott, would love to surface this in a more intuitive way for users. It's a great piece of functionality and adds a lot of value. I'll have a think about this.

Great PR too @querielo 🔥

querielo:
> Ideal solution in the future would be that if the channels do not match, we build the required textures from source textures.

I'll fix it later when I have free time.

mvaligursky:
Please try to avoid make this PR larger and keep it focused on the material export improvements this PR is about, as this is getting hard to review now. Follow up work is awesome, but a separate PR please.

mvaligursky (approved):
Approving, tested and works well, well done and thanks!

PR #5766, opened 2023-10-24
Specify type definitions as dependencies

willeastcott:
We ship type definitions for the engine that contain references to WebXR types. This means that TS projects now have to specify `@types/webxr` as a dependency:

https://github.com/playcanvas/playcanvas-editor-ts-template/pull/12/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R34

This would be unnecessary if we just made types `dependencies` instead of `devDependencies`.

I believe the WebGPU types don't need to be `dependencies` but, for consistency, I'm moving those to `dependencies` too. It will avoid any potential future issues around WebGPU type dependencies.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,10 @@
       "name": "playcanvas",
       "version": "1.67.0-dev",
       "license": "MIT",
+      "dependencies": {
+        "@types/webxr": "^0.5.5",
+        "@webgpu/types": "^0.1.35"
+      },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


slimbuck (approved):

mvaligursky:
This is great, it probably means I no longer need to make module functions private if they use webgpu types in JSDocs.

willeastcott:
I believe that's correct, yes.

PR #5768, opened 2023-10-26
Make `npm run build` also trigger `npm run build:types`

kungfooman:
Since we rimraf `/build/` we don't rebuild the types, which makes `cd examples && npm run build` fail (because Monaco editor depends on `build/playcanvas.d.ts`).

Solution is quite easy (simply trigger `npm run build:types`), but side-effects for example could be increased RAM usage, so the build could fail on cheap/overused VPS servers (which struggle with RAM).


willeastcott (approved):

willeastcott:
Thanks for this! 🙏

PR #5769, opened 2023-10-26
Adding rollup replace

marklundin:
This PR covers an edge-case when using 'use_local_engine' during development with Script ES Modules. 

In this scenario the engine is running on 'localhost:51000' meaning that any modules imported via the engine are relative to localhost and not the document. This means that user ESM Scripts are incorrectly imported from localhost.

Rather than introducing any logic into the engine itself (this is only applies for local dev) or pointing to window.location. This PR simply allows us to replace symbols and inject the host url during development via 
```
ASSET_BASE_URL=https://launch.local-playcanvas.com/ && npm run watch
```
At final build, this will have no change on the build output.


isumygin-sc (approved):

isumygin-sc:
tests are broken for some reason

kungfooman:
Tests should work after committing `package-lock.json`

slimbuck:
We're already using jscc for text replacement in the rollup build. Could you use that instead?

kungfooman:
I have a feeling that `jscc` would even replace assignments, e.g.

```js
const ASSET_BASE_URL = '';`
```

would turn into:

```js
const https://launch.local-playcanvas.com/ = '';
```

(kinda low-level C preprocessor)

marklundin:
OK closing this 👍 jscc covers this usecase

PR #5774, opened 2023-11-01
WebXR manual room capture

Maksims:
With Quest 3 and similar devices, environment is captured at the special time instead of real-time like for Android devices.
So the extension to the [WebXR Plane Detection Specs](https://immersive-web.github.io/real-world-geometry/plane-detection.html#dom-xrsession-initiateroomcapture) provides a new method for XRSession to initiate manual room capture.

This can be useful when developer want to use AR with Plane Detection or Real-World Meshes, and needs an up to date information, or user might have not pre-captured their environment at all. So developer can provide an option to trigger manual capture right from the AR session.
Our current API integration design takes asynchronous nature of real-world objects like planes into the account, so changes in underlying data will be reflected immediately without a need to refresh a page.

### New APIs:
```javascript
// pc.XrManager
xr.initiateRoomCapture((err) => { })
// Start manual room capture.
// If such feature is not supported, error with according message will be provided in callback.
// If capture was a successful, then callback will fire without an error.
```

### DIFF
--- a/src/framework/xr/xr-manager.js
+++ b/src/framework/xr/xr-manager.js
@@ -26,6 +26,13 @@ import { XrAnchors } from './xr-anchors.js';
  */
 
 /**
+ * Callback used by manual room capturing.
+ *
+ * @callback XrRoomCapture
+ * @param {Error|null} err - The Error object or null if manual room capture was successful.
+ */
+
+/**
  * Manage and update XR session and its states.
  *
  * @augments EventHandler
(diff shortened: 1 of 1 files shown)


src/framework/xr/xr-manager.js
```diff
@@ -25,6 +25,13 @@ import { XrAnchors } from './xr-anchors.js';
  * @param {Error|null} err - The Error object or null if operation was successful.
  */
 
+/**
+ * Callback used by manual room capturing.
+ *
+ * @callback XrRoomCapture
```
kungfooman, line 31: I think we usually add `Callback` like `XrRoomCaptureCallback`, but I started to wonder, do we still want to add more callbacks everywhere when Promises are becoming the new standard to prevent callback hell?

Maksims: Well spotted.
Yes, we want a **consistent API**, so we use callbacks as it is a common practice around the engine.
Promises do not prevent callback hell, async/await do, but they are not always desirable.

Refactoring / adding promise alternatives to callback-like methods should be done as a separate PR if users will demand a need for and report that there is a callback hell (so far API is clean and devs have not highlighted it as an issue).

marklundin (approved):

slimbuck (approved):

PR #5778, opened 2023-11-01
WebXR fix Hit Test providing multiple results

Maksims:
Based on WebXR specs, it is possible to query hit test results for specific input sources, but it is done using `profile` array string. As left / right controllers have the same profile string, then hit test source can report information from various input sources. This PR fixes that, and ensures that HitTestSource fires `result` event only for the input source it is related to.

Additionally underlying systems can provide multiple results per hit test source as ray can pass through multiple planes. Previously only devices with a single result were present, so this PR updates that, and ensures that single (closest) result is returned. If there will be a need for returning multiple results, then additional PR with `resultsAll` callback probably can be made.

Example with AR Hit Test is updated to include additional input sources (screen taps, controllers and hands).

### DIFF
--- a/src/framework/xr/xr-hit-test-source.js
+++ b/src/framework/xr/xr-hit-test-source.js
@@ -132,26 +132,44 @@ class XrHitTestSource extends EventHandler {
      * @private
      */
     updateHitResults(results, inputSource) {
-        if (inputSource && inputSource.hitTestSourcesSet.has(this))
+        if (inputSource && !inputSource.hitTestSourcesSet.has(this))
             return;
+
+        let origin = poolVec3.pop();
+        if (!origin) origin = new Vec3();
+
+        if (inputSource) {
+            origin.copy(inputSource.getOrigin());
+        } else {
+            origin.copy(this.manager.camera.getPosition());
+        }
+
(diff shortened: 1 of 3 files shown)


src/framework/xr/xr-hit-test-source.js
```diff
@@ -132,23 +132,44 @@ class XrHitTestSource extends EventHandler {
      * @private
      */
     updateHitResults(results, inputSource) {
+        if (inputSource && !inputSource.hitTestSourcesSet.has(this))
+            return;
+
+        let origin = poolVec3.pop();
+        if (!origin) origin = new Vec3();
```
mvaligursky, line 139:
```suggestion
        const origin = poolVec3.pop() ?? new Vec3();
```

mvaligursky: and the same can be done for position / rotation below

mvaligursky (approved):
Approved with a minor comments

src/framework/xr/xr-hit-test-source.js
Maksims: Good tip!

PR #5782, opened 2023-11-02
Examples: Use actual component to hide spinner

### LINKED ISSUE #5781: Examples loading should hide after loading is complete.
Recently engine examples were reworked which is great, but it affected few examples.
One of the effect is that loading spinner does not hide after loading is complete. This leads to examples with transparent canvas to appear as they infinitely loading.

### Solution:

After example is finished loading it should hide the spinner.

kungfooman:
Fixes #5781

(I would still like some feedback from @Maksims about which example/OS was affected to test this on my side, in the meantime I will update the live browser so we can easily test it)

Edit:

This PR doesn't just `display:none` but removing it from the DOM. Testable via `document.getElementsByClassName("spin")`:

image

Updated Examples browser for this PR: https://examples.killtube.org/#/misc/hello-world

### DIFF
--- a/examples/src/app/code-editor.mjs
+++ b/examples/src/app/code-editor.mjs
@@ -78,14 +78,12 @@ class CodeEditor extends TypedComponent {
         /** @type {Record<string, string>} */
         const files = event.files;
         this.mergeState({ files, selectedFile: 'example.mjs' });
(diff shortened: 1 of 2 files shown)


mvaligursky (approved):
Approving, looks good, thank you.

Issue #5775, opened 2023-11-01
Engine 1.66+ fails to create XR context with default GraphicsDevice settings.


Maksims:
### Description

Regression with engine 1.66+.

When starting WebXR VR session, it creates a XRWebGLLayer that creates a framebuffer based on provided settings. We create it with `alpha:true`. But by default GraphicsDevice is created with `alpha:false`. This leads to an error when engine tries to blit buffers as their texture format mismatches.

When framebuffers are re-created (on WebXR Session start), engine should create underlying buffers with forced `alpha:true`.

### Steps to Reproduce

1. Using a VR device (emulator wont replicate an issue), navigate to: https://playcanvas.github.io/#/xr/vr-basic
2. Click on viewport to start VR session.

### Issue:

No rendering, and console spams with warnings of mismatched framebuffers.

### Temporary workaround:

Add extra option to Application constructor:
```js
const app = new pc.Application(canvas, {
    graphicsDeviceOptions: {
        alpha: true // this
    }
});
```

PR #5784, opened 2023-11-03
[Fix] When the default framebuffer is updated, refresh info about its format

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/5775

When the default framebuffer changes (for example entering or exiting XR), detect its pixel format, as that is required to create a matching multi-sampled framebuffer.

Without this, the multi-sampled framebuffer could have different format (alpha channel) and resolve blit would fail.

### DIFF
--- a/src/platform/graphics/webgl/webgl-graphics-device.js
+++ b/src/platform/graphics/webgl/webgl-graphics-device.js
@@ -383,8 +383,7 @@ class WebglGraphicsDevice extends GraphicsDevice {
         this._deviceType = this.isWebGL2 ? DEVICETYPE_WEBGL2 : DEVICETYPE_WEBGL1;
 
         // pixel format of the framebuffer
-        const alphaBits = gl.getParameter(gl.ALPHA_BITS);
-        this.backBufferFormat = alphaBits ? PIXELFORMAT_RGBA8 : PIXELFORMAT_RGB8;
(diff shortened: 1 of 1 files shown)


src/platform/graphics/webgl/webgl-graphics-device.js
```diff
@@ -787,10 +786,24 @@ class WebglGraphicsDevice extends GraphicsDevice {
         this.backBuffer.impl.suppliedColorFramebuffer = frameBuffer;
     }
 
+    // Update framebuffer format based on the current framebuffer, as this is use to create matching multi-sampled framebuffer
+    updateBackbufferFormat(framebuffer) {
+        const gl = this.gl;
+        gl.bindFramebuffer(gl.FRAMEBUFFER, framebuffer);
+        const alphaBits = this.gl.getParameter(this.gl.ALPHA_BITS);
```
slimbuck, line 793: any performance worries here? i thought getParameter was real slow.

mvaligursky: this happens when entering / exiting XR .. so no worries? We do this already when the engine starts up. It's not that slow.

slimbuck: oic, thought this was per frame 👍🏻

slimbuck (approved):

PR #5785, opened 2023-11-03
Initial implementation of Gaussian Splat support in Extras

mvaligursky:
- initial version based on the code in https://github.com/playcanvas/model-viewer/
- engine example
Screenshot 2023-11-03 at 11 36 25

### DIFF
--- a/extras/splat/splat-container-resource.js
+++ b/extras/splat/splat-container-resource.js
@@ -29,7 +29,7 @@ class SplatContainerResource extends ContainerResource {
         this.splatData = splatData;
 
         // calculate splat focal point
-        this.splatData.calcFocalPoint(this.focalPoint);
+        this.updateFocalPoint();
 
         // calculate custom aabb
         this.splatData.calcAabb(this.customAabb);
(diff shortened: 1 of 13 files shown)


kungfooman:
Funny, GS is the new hot topic and I also mistyped it as "Gassian" to a friend when I mentioned it this morning :sweat_smile:

extras/splat/splat-container-resource.js
```diff
@@ -0,0 +1,127 @@
...
+class SplatContainerResource extends ContainerResource {
+    device;
+
+    splatData;
+
+    splat;
+
+    focalPoint = new Vec3();
+
+    // TODO: this can be set on render component -> mesh instance, remove this one?
```
slimbuck, line 22: Yeah this should be removed and probably also `focalPoint` (and the functions below dealing with them).

The aabb should probably be stored on `Splat` (passed into constructor) then `SplatInstance` can set its `MeshInstance` aabb in constructor.

The editor will need to calcualte custom aabb when splats edited, but I will probably use SplatData directly for that.

slimbuck (approved):

Issue #5780, opened 2023-11-02
WebXR Raw Camera Access


Maksims:
~~Since Quest 3 got released, it gets access to camera's raw textures.~~
We would want to have an API that gives an easy and integrated way to access and update such textures.
Specs: https://immersive-web.github.io/raw-camera-access/

EDIT:
Android has access to raw camera texture.

Maksims:
Below is an example of an potential API. It implements access to views that session has.
Due to nature of WebXR API, when session starts views are not available straight away, and will become available on the first pose data availability. So events should be used to detect when views are added.
It is designed to include depth information in this API in the future, and deprecate XrDepthSensing as it only assumed a single view information (mobile phone), but MR HMD devices now can provide multiple depth textures per view (two eyes).

```javascript
// constants
// it comes from XRView eye property
XREYE_NONE = 'none' // usually it is a mobile screen
XREYE_LEFT = 'left'
XREYE_RIGHT = 'right'

// pc.XrManager
xr.views // XrViews object that provides access to views
xr.start(camera, pc.XRTYPE_AR, pc.XRSPACE_LOCALFLOOR, {
    cameraColor: true // new option to enable access to camera color information
});

// pc.XrViews
// this class represents number of views, their capabilities and access to individual views
views.supportedColor // true if color information is supported
views.availableColor // true if color information is available.
// It can be supported but not available due to lack of session eatures request or e.g.
// Quest 3 - implements necessary features, but does not expose `camera-access` to users.
views.get(eye) // returns XrView or null based on XREYE_*
views.list // an array of XrView instances
views.size // number of views
views.on('add', (XrView) => { }); // event when new XrView is added
views.on('remove', (XrView) => { }); // event when XrView is removed

// pc.XrView
// this class represents a specific view with its data
view.eye // returns a string (constants XREYE_*). 'none' - means it is not an eye, and
// most likely is a mobile screen
view.textureColor // returns a texture with color information. If color information is not
// available, texture will be blank 4x4
```

Example of how to access view color texture:

```javascript
app.xr.start(camera, pc.XRTYPE_AR, pc.XRSPACE_LOCALFLOOR, {
    cameraColor: true
});

const views = app.xr.views;

if (views.supportedColor) {
    // camera color might not be supported
    views.on('add', (view) => {
        if (!views.availableColor) {
            // camera color might be unavailable
            return;
        }
        
        // get access to color texture
        const texture = view.textureColor;
    });
}


// ...
// or later when views has been created:
[...]

PR #5786, opened 2023-11-04
WebXR Raw Camera Access (color)

Maksims:
This PR integrates Raw Camera Access API, that provides access to color image of a XR views. Along the way it introduces XrView concept which aligns with WebXR specs. It gives access to each view (single on mobile phones, and two on HMD devices), and will be extended with Depth Sensing API in the future. As current Depth Sensing API assumes a single view, so it will not work well with Quest 3's ability of providing depth maps for each view.

Test link: https://examples.playcanvas.meta.space/#/xr/ar-camera-color

Discussions: https://github.com/playcanvas/engine/issues/5780

```javascript
// constants
// it comes from XRView eye property
XREYE_NONE = 'none' // monoscopic screens, e.g. mobile phones
XREYE_LEFT = 'left'
XREYE_RIGHT = 'right'

// pc.XrManager
xr.views // XrViews interface that provides an access to views and their capabilities
xr.start(camera, pc.XRTYPE_AR, pc.XRSPACE_LOCALFLOOR, {
    cameraColor: true // new option to request camera color information
});

// pc.XrViews
// this class provides an access to views, and their capabilities
views.supportedColor // true if color information is supported
views.availableColor // true if color information is available.
// This information is only available after XR session has started.
// It can be supported but not available due to lack of session features request, user permissions
// or underlying system capabilities
// Quest 3 - implements necessary features, but does not expose `camera-access` to users.
views.get(eye) // returns XrView or null based on XREYE_*
views.list // an array of XrView instances
views.on('add', (view) => { }); // event when new XrView is added
views.on('remove', (view) => { }); // event when XrView is removed

// pc.XrView
// this class represents a specific view with its data
view.eye // returns a string (constants XREYE_*). 'none' - means it is monoscopic display such as mobile phone screen
view.textureColor // returns a texture with color information. If color information is not available it will be null
view.viewport // Vec4 that represents a viewport are that this view is occupying of the screen
```


Maksims:
There was a comment from @mvaligursky:

> To be honest, I'd strongly prefer we use some a multi-platform solution here instead of accessing WebGL directly. That means render target and drawQuadWithShader functionality. Platform stuff is inside `/platform` folder only as much as possible. In the future WebGPU will add support for XR and we'd want it to just work.

I've tried multiple ways:
1. Use RenderTargets - but it does not provide a way to create one with raw WebGLTexture, and `copy` method does pretty much what we do in the current implementation.
2. Direct WebGLTexture use - this works, and is most efficient way, but it does not provide flipY, and has a rule for ["opaque texture"](https://immersive-web.github.io/layers/#opaque-texture) that WebGLTexture object should not be used outside of its RAF.
3. Direct WebGL functions use with blit function, that provides a way to flipY the buffer, and avoids quad a render.

Currently WebXR integration is only WebGL compatible, there are WebXR-WebGPU bindings Specs coming along, but we do not have devices that do support both APIs yet, so this is for the future PRs.

src/framework/xr/xr-view.js
```diff
@@ -0,0 +1,338 @@
...
+        device.setFramebuffer(this._frameBuffer);
+        gl.framebufferTexture2D(
+            gl.FRAMEBUFFER,
+            attachmentBaseConstant,
+            gl.TEXTURE_2D,
+            this._textureColor.impl._glTexture,
+            0
+        );
+
+        // bind buffers
```
mvaligursky, line 285: I'd recommend:
- call checkFramebufferStatus in debug mode only (the rest of the engine does `Debug.call(() => this._checkFbo(device, target));` - but only when the framebuffer is created, not each frame, as this is expensive call.

src/framework/xr/xr-view.js
```diff
@@ -0,0 +1,338 @@
...
+        const device = this._manager.app.graphicsDevice;
+        const gl = device.gl;
+        const attachmentBaseConstant = device.isWebGL2 ? gl.COLOR_ATTACHMENT0 : (device.extDrawBuffers?.COLOR_ATTACHMENT0_WEBGL ?? gl.COLOR_ATTACHMENT0);
+
+        const width = this._xrCamera.width;
+        const height = this._xrCamera.height;
+
+        if (!this._textureColor) {
+            // color texture
+            this._textureColor = new Texture(this._manager.app.graphicsDevice, {
```
mvaligursky, line 243: use `device` instead of `this._manager.app.graphicsDevice`

src/framework/xr/xr-views.js
```diff
@@ -0,0 +1,196 @@
...
+        return this._list;
+    }
+
+    /**
+     * How many views are available. Views can be added/removed mid-session by underlying
+     * WebXR system.
+     *
+     * @type {number}
+     */
+    get size() {
```
mvaligursky, line 101: is this needed? the user can do `.list.length`

src/framework/xr/xr-views.js
```diff
@@ -0,0 +1,196 @@
...
+        // remove views
+        for (const [eye, view] of this._index) {
+            if (this._indexTmp.has(eye))
+                continue;
+
+            view.destroy();
+            this._index.delete(eye);
+            const ind = this._list.indexOf(view);
+            if (ind !== -1) this._list.splice(ind, 1);
+            this.fire('remove', view);
```
mvaligursky, line 160: should you fire the event before destroying the view? Otherwise the user gets destroyed view.

mvaligursky:
Fantastic PR @Maksims , thanks for this! I added few small comments, but overall this is great.

src/framework/xr/xr-view.js
Maksims: Good tip. I've noticed that frame buffer were not ready to be used straight away, how to check/skip the `blit` without checks?

mvaligursky: we don't do any checks like this in the engine .. the buffer should be ready straight away, so that's something strange going on there. Do you get any webgl errors?

src/framework/xr/xr-views.js
mvaligursky: ok, keep consistent then.

src/framework/xr/xr-view.js
Maksims: Yes, I get this error:
```
[.WebGL-0x200cb2200] GL_INVALID_FRAMEBUFFER_OPERATION: Framebuffer is incomplete: No attachments and default size is zero.
```

mvaligursky: so that feels like maybe the texture from `texture = binding.getCameraImage` is zero size initially or something .. maybe there's a way to test for this instead and early out, same as when it's null

Maksims: From my observation, Android Chrome likely has multiple textures behind the scenes, where they provide swizzled access to. As they likely render into one texture while providing access to other. That is partly why they are "opaque", so we don't try to read texture which they might be writing to.

I found a workaround. Basically do not try to copy data on the first frame of session, but do it on the second - that worked well. I do create a texture on initialize to have more stable behavour for the user.

src/framework/xr/xr-view.js
```diff
@@ -101,10 +101,24 @@ class XrView {
...
 
-        if (this._manager.views.supportedColor)
+        if (this._manager.views.supportedColor) {
             this._xrCamera = this._xrView.camera;
 
-        this._updateTextureColor();
+            // color texture
+            if (this._manager.views.availableColor && this._xrCamera) {
+                this._textureColor = new Texture(this._manager.app.graphicsDevice, {
+                    format: PIXELFORMAT_RGB8,
```
mvaligursky, line 110: Just a quick check - is this always RGB8? Or could there be an alpha channel?

Maksims: Currently Specs [do not provide such information](https://immersive-web.github.io/raw-camera-access/#issue-63423cac) and from my testing it was the format provided.

src/framework/xr/xr-view.js
```diff
@@ -0,0 +1,333 @@
...
+        const texture = binding.getCameraImage(this._xrCamera);
+        if (!texture)
+            return;
+
+        const device = this._manager.app.graphicsDevice;
+        const gl = device.gl;
+
+        if (!this._frameBufferSource) {
+            // create frame buffer to read from
+            this._frameBufferSource = gl.createFramebuffer();
```
mvaligursky, line 253: Last suggestion. I assume the device can be lost while this is active, and those framebuffers would no longer be valid / work, and so this should be handled.

the device fires this when the device is lost:
```           
 this.fire('devicelost');
```

So you should listen to this, and simply set gpu resources to null, so that they re-created the next frame, for example:

```
this._frameBufferSource = null;
```

It'd be good to test this by forcing device lost if possible on the device you're testing this on, to make sure nothing is missed here.

Maksims: I've improved device lost handling by camera, but apparently XrManager needs to do a better job for handling device lost also, and re-create XRWebGLLayer. I will make another PR for that.

Maksims: Implemented handling of WebGL context loss by WebXR here: https://github.com/playcanvas/engine/pull/5849

mvaligursky (approved):
Great work, thanks so much.

Issue #5699, opened 2023-09-29
WebXR Mesh Detection


Maksims:
New Mesh Detection API is in the works, and soon will be available on Quest 3 (after October 10).

This API provides more detailed information about the real world in form of meshes, their polygons and most importantly semantic labels.
This provides a few new possibilities:
1. Generate a mesh that represents real-world, that can be used for e.g. physics.
2. Get semantic labels of meshes like: wall, window, screen, art on the wall, etc. And use it to drive your application logic.
3. Faster RayCasting - which can be done in sync instead of using Hit Detection API.

This issue is for discussion of potential API to integrate into PlayCanvas Engine.

Specs: https://immersive-web.github.io/real-world-meshing/
Labels: https://github.com/immersive-web/semantic-labels

PR #5791, opened 2023-11-05
WebXR Mesh Detection

Maksims:
Integration of [WebXR Mesh Detection API](https://immersive-web.github.io/real-world-meshing/).

Tested on Quest 3. This data on Quest 3 comes from pre-scanned information and added furniture. You can trigger a request for [manual room capture](https://github.com/playcanvas/engine/pull/5774), so users can provide refined information.
Includes an example of building geometry and labeling them.
This API design based on [planeDetection API](https://github.com/playcanvas/engine/pull/3068).

Test link: https://examples.playcanvas.meta.space/#/xr/ar-mesh-detection

This API can enable many cool AR applications:
1. Efficient depth based occlusion of virtual geometry.
2. Physics interactions of virtual objects with mesh that approximates real-world.
3. Using semantic labels it is possible to identify where screens are, table surfaces, and more.
4. And much more.

New APIs:
```javascript
// pc.XrManager
xr.meshDetection // XrMeshDetection interface that provides an access to XrMeshe's
xr.start(camera, pc.XRTYPE_AR, pc.XRSPACE_LOCALFLOOR, {
    meshDetection: true // new option to request mesh detection features
});

// pc.XrMeshDetection
// this class provides an access to meshes.
// XrMesh instances can be added/removed mid-session by underlying system, due to real-time scanning, tracking state changes, manual scanning, etc.
meshDetection.supported // true if mesh detection is supported
meshDetection.available // true if mesh detection is available. This information is only available when session has started.
meshDetection.meshes = [ ]; // an array of XrMesh instances
meshDetection.on('available', () => { }); // fired when mesh detection becomes available
meshDetection.on('unavailable', () => { }); // fired when mesh detection becomes unavailable
meshDetection.on('add', (xrMesh) => { }); // fired when a new XrMesh have been added
meshDetection.on('remove', (xrMesh) => { }); // fired when XrMesh have been removed

// pc.XrMesh
// this class represents a specific XrMesh, its vertices data and attributes
xrMesh.label // string - a semantic label of this mesh
// can be any of: https://github.com/immersive-web/semantic-labels/blob/master/labels.json
// but not limited to this list. e.g. Quest 3 also uses "global mesh".
xrMesh.vertices // Float32Array - local positions of vertex data. This data can change during lifetime of XrMesh, when "change" event fired.
xrMesh.indices // Uint32Array - indices for vertices to construct triangles. This data can change during lifetime of XrMesh, when "change" event fired.
xrMesh.getPosition() // global position of XrMesh (Vec3), this data can change without firing "change" event.
xrMesh.getRotation() // global rotation of XrMesh (Quat), this data can change without firing "change" event.
xrMesh.on('change', () => { }); // fired when any of its attributes: vertices, indices, label have been changed.
xrMesh.on('remove', () => { }); // fired when XrMesh have been removed
```

Fixes https://github.com/playcanvas/engine/issues/5699

https://github.com/playcanvas/engine/assets/636263/51f81a2c-87df-4dd7-83e6-1eb91eb3cbdd


examples/src/examples/xr/ar-mesh-detection.mjs
```diff
@@ -0,0 +1,272 @@
...
+            app.keyboard.on('keydown', function (evt) {
+                if (evt.key === pc.KEY_ESCAPE && app.xr.active) {
+                    app.xr.end();
+                }
+            });
+
+            app.xr.on('start', function () {
+                message("Immersive AR session has started");
+
+                // TODO
```
mvaligursky, line 117: what's left todo here?

examples/src/examples/xr/ar-mesh-detection.mjs
```diff
@@ -0,0 +1,272 @@
...
+            message("WebXR is not supported");
+        }
+    });
+
+    return app;
+}
+
+class ArMeshDetectionExample {
+    static CATEGORY = 'XR';
+    static NAME = 'AR Mesh Detection';
```
mvaligursky, line 268: remove the NAME property, it's no longer used.

src/framework/xr/xr-mesh.js
```diff
@@ -0,0 +1,179 @@
...
+     * @type {number}
+     * @private
+     */
+    _id;
+
+    /**
+     * @type {import('./xr-mesh-detection.js').XrMeshDetection}
+     * @private
+     */
+    _meshes;
```
mvaligursky, line 25: this could be perhaps better called `meshDetection` .. meshes implies it's an array of some meshes

src/framework/xr/xr-mesh.js
```diff
@@ -0,0 +1,179 @@
...
+     *     // mesh attributes have been changed
+     * });
+     */
+
+    /**
+     * Unique identifier of a mesh.
+     *
+     * @type {number}
+     */
+    get id() {
```
mvaligursky, line 94: any reason this is exposed?
It implies that perhaps if the same mesh gets added later after it was removed, it would have the same Id.

Maksims: It is for convenience, and consistency with XrPlanes. As XrMesh can be added/removed during the session, and developers might use ID's for their indexes and identification. Perhaps with common use of Map and being able to use object as a key, instead of old-school object<id, XrMesh> type index, this is not needed anymore.

examples/src/examples/xr/ar-mesh-detection.mjs
Maksims: Triggered manual room capture here.

Maksims:
Made a quick update based on comments, and removed ID as it is nowadays redundant, as Map is widely used for indexes.

Issue #3974, opened 2022-02-02
Collision using unloaded asset bug


ma-vejsada:
### Description

Collision component throws an error when using unloaded render asset. Black screen will appear and project could not be launched.

Shortened error:
```
TypeError: meshes is null
    createPhysicalShape https://launch.playcanvas.com/editor/scene/js/engine/playcanvas.dbg.js?version=1.51.6:51339
    doRecreatePhysicalShape https://launch.playcanvas.com/editor/scene/js/engine/playcanvas.dbg.js?version=1.51.6:51398
    loadAsset https://launch.playcanvas.com/editor/scene/js/engine/playcanvas.dbg.js?version=1.51.6:51377
    ...
```

### Steps to Reproduce

1. import custom model (as render asset, with hierarchy)
2. uncheck preloading (both template and render)
3. create collision component on empty gameobject
4. assign imported render asset to collision component
5. launch project
6. this error appears

### Possible solution

1. Implement asset resource check.
2. Implement asset awaiter, to wait until asset is loaded, then use the mesh.

ma-vejsada:
example project: https://playcanvas.com/project/877512/overview/collision-error

LeXXik:
Possibly duplicate:
https://github.com/playcanvas/engine/issues/3638

willeastcott:
Should have been closed by https://github.com/playcanvas/engine/pull/5799

PR #5799, opened 2023-11-06
Fix collision initialization for container assets.

jpauloruschel:
Fixes https://github.com/playcanvas/engine/issues/3638 and https://github.com/playcanvas/engine/issues/3974.

This PR implements an additional check when using Assets that have a Container Asset on Collision components. The new code will check if the asset's `containerAsset` is also loaded, and if not, load it. As noted in the linked Issues, this is required so that the `meshInstances` are actually loaded on the asset.

For use-cases where the Asset does not have a containerAsset or if it's already loaded, there's no change.

Sample Project with bug: https://playcanvas.com/editor/scene/1894310


mvaligursky (approved):
Looks great, thanks!

slimbuck (approved):

Issue #5729, opened 2023-10-06
WebGPU `AppBase#destroy` triggers reading-push-of-null error


kungfooman:
@LeXXik found this originally and this is a standalone repro:

Live test: https://killtube.org/~playcanvas/examples_es6/many-apps.html

Code:

```html
<!DOCTYPE html>
<html>
<head>
    <meta charset="utf-8">
    <title>PlayCanvas Many Apps</title>
    <meta name='viewport' content='width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no' />
    <style>
        body {
            margin: 0;
            overflow: hidden;
        }
    </style>
</head>
<body>
    <div>
        <button onclick="createAnotherApp('webgl2')">createAnotherApp('webgl2')</button>
        <button onclick="createAnotherApp('webgpu')">createAnotherApp('webgpu')</button>
        <button onclick="destroyLast()">destroyLast()</button>
    </div>
    <script type="module">
        import * as pc from './src/index.js';
        window.pc = pc;
        const apps = [];
        function destroyLast() {
            const lastApp = apps.pop();
            if (!lastApp) {
                return;
            }
            const { canvas } = lastApp.graphicsDevice;
            lastApp.destroy();
            canvas.remove();
        }
        async function createAnotherApp(deviceType) {
            const canvas = document.createElement('canvas');
            document.body.appendChild(canvas);
            const app = await example({
                canvas,
                deviceType,
                // relative from playcanvas.js due import(), only absolute works
                glslangPath: '/playcanvas-engine/examples/src/lib/glslang/',
                twgslPath: '/playcanvas-engine/examples/src/lib/twgsl/'
             });
            apps.push(app);
        }
        async function example({ canvas, deviceType, glslangPath, twgslPath }) {
            const gfxOptions = {
                deviceTypes: [deviceType],
                glslangUrl: glslangPath + 'glslang.js',
                twgslUrl: twgslPath + 'twgsl.js'
            };
            const device = await pc.createGraphicsDevice(canvas, gfxOptions);
            const createOptions = new pc.AppOptions();
            createOptions.graphicsDevice = device;
            createOptions.componentSystems = [
                pc.RenderComponentSystem,
                pc.CameraComponentSystem,
                pc.LightComponentSystem
            ];
            createOptions.resourceHandlers = [
                pc.TextureHandler,
                pc.ContainerHandler
            ];
            const app = new pc.AppBase(canvas);
            app.init(createOptions);
[...]

PR #5802, opened 2023-11-08
Build webgpu module full from relative path

slimbuck:
Fixes #https://github.com/playcanvas/editor/issues/1070

Build the full url module path from relative path because when engine is loaded with CORS-cross-origin the base url is about:blank.

### DIFF
--- a/src/platform/graphics/webgpu/webgpu-graphics-device.js
+++ b/src/platform/graphics/webgpu/webgpu-graphics-device.js
@@ -174,9 +174,17 @@ class WebgpuGraphicsDevice extends GraphicsDevice {
         // temporary message to confirm Webgpu is being used
         Debug.log("WebgpuGraphicsDevice initialization ..");
 
+        // build a full URL from a relative path
+        const buildUrl = (relativePath) => {
+            const url = new URL(window.location.href);
+            url.pathname = relativePath;
+            url.search = '';
+            return url.toString();
+        };
+
         const results = await Promise.all([
-            import(`${twgslUrl}`).then(module => twgsl(twgslUrl.replace('.js', '.wasm'))),
-            import(`${glslangUrl}`).then(module => module.default())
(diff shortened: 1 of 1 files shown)


mvaligursky (approved):
Nice!

kungfooman:
I was running into the same issue for my issue-example here: https://github.com/playcanvas/engine/issues/5729

I tested this with that example, but changed it to use a relative path:

image

Error:

image

slimbuck:
Hi @kungfooman,

Please would you submit your additional change as followup PR? Then we can all discuss.

Thanks!

src/platform/graphics/webgpu/webgpu-graphics-device.js
```diff
@@ -174,9 +174,17 @@ class WebgpuGraphicsDevice extends GraphicsDevice {
         // temporary message to confirm Webgpu is being used
         Debug.log("WebgpuGraphicsDevice initialization ..");
 
+        // build a full URL from a relative path
+        const buildUrl = (relativePath) => {
+            const url = new URL(window.location.href);
```
marklundin, line 179: I'm not sure we should have a `window` reference which would break in node. Can we just update the `createGraphicsOption` to include the full path, and assume any path passed in is correct?

kungfooman: I think it's fine as long it's not e.g. in a variable declaration in global scope, or do you run into this code and trigger a reference error in node?

(before it was a simple `<script src=...=>` aswell and if you wanted to use that you would have already required the jsdom shim lib in node to have access to `window` etc. via node)

marklundin: Yeah I just wonder if we just require the path to be correct when it's passed in, rather than making the engine make figure it out. The createGraphics options just has the full url. Not sure, might be missing something obvious

mvaligursky: > correct when it's passed in

I agree with this. At the end, the application itself passes this path in, they should be able to build a correct path easier than the engine.

Issue #5801, opened 2023-11-07
Serious memory leak: Ammo btRigidBodies are (almost) never destroyed


ThatStevenGuy:
### Description
Hey everyone,

We have very large worlds in one of our games with roughly 3-5k rigid bodies per scene (only a handful are active at a time). We noticed after switching scenes a couple of times, Ammo would eventually crash with an `abort(OOM)`-error indicating it exhausted its 64 Mb heap. After some digging, it turns out almost none of Ammo's underlying `btRigidBody`'s in our scenes are actually destroyed.

### Cause
The issue stems from `RigidBodyComponentSystem`. This class listens for the removal of rigid body components so it can destroy the underlying `btRigidBody`. It does this by binding its `onRemove`-method to the `remove`-event of `ComponentSystem` during initialization. This `remove`-event however requires a signature of `(entity: Entity, componentData: any)`, but the `onRemove`-method has a signature of `(entity: Entity, component: RigidBodyComponent)` (note the type of the second parameter). 

As a result, whenever a rigid body component is destroyed and the `remove`-event is fired, `RigidBodyComponentSystem.onRemove` will receive a non-component argument for its `component`-parameter. This causes the destruction of the underlying body to always fail, since the method can't find the `body`-field on what it thinks is a rigid body component, but is actually an argument of type `RigidBodyComponentData`.

Small side note: the only time `RigidBodyComponentSystem.onRemove` works as expected is through `RigidBodyComponent.createBody()`, which invokes `onRemove` to destroy any previous `btRigidBody` instance.

### Steps to Reproduce
1. In the Editor, create a new scene;
2. Under the `Root` entity, find the default `Box` entity and attach a collision component and a rigid body component to it (default values are fine);
3. Press play. Make sure the URL includes `debug=true`;
4. Open your browser's developer tools and find the file `rigid-body/system.js` which contains `RigidBodyComponentSystem`. Place a breakpoint on the first line of the `onRemove`-method;
5. Destroy the box by invoking `pc.app.root.children[0].children[2].destroy();` in the JavaScript console;
6. The breakpoint you've set will be hit. Note that the `component`-argument is in fact of the type `RigidBodyComponentData` which only contains an `enabled`-field, not a `body`-field. The check for the existence of `component.body` will always fail, and the body will remain in Ammo's memory;

### Quick-Fix
A quick way to resolve the issue is to make a new script named `fix.js` and insert the following:

```
let Fix = pc.createScript('fix');

[...]

PR #5803, opened 2023-11-08
Fixes RigidBody remove component

### LINKED ISSUE #4145: Potential memory leak in Rigidbody Component
When a Rigidbody Component is removed from the entity, the Component System calls its `onRemove` method and passes the entity and a component data object. Since rigidbody component was migrated to the new class, it dropped the reliance on the component data object. Since then `onRemove` method is incorrectly checking for the private `body` property.

```js
entity.removeComponent('rigidbody');
```

`pc.ComponentSystem#removeComponent()`: Notice it passes a component data as a last argument.
https://github.com/playcanvas/engine/blob/1967983c1cb975c2dfd2cd2232a6d4ebf50e644f/src/framework/components/system.js#L85

`pc.RigidBodyComponentSystem#onRemove` however thinks it is a component and tests its `body` property, which will always fail, since its component data is not used.
https://github.com/playcanvas/engine/blob/1967983c1cb975c2dfd2cd2232a6d4ebf50e644f/src/framework/components/rigid-body/system.js#L410-L418

Consider moving it to the `onBeforeRemove`, which receives actual component:
image

marklundin:
This PR fixes #4145 and #5801 where the RigidBodyComponentSystem would never remove a rigid body from Ammo when `removeComponent` was called

### DIFF
--- a/src/framework/components/rigid-body/component.js
+++ b/src/framework/components/rigid-body/component.js
@@ -503,8 +503,12 @@ class RigidBodyComponent extends Component {
         }
 
         if (shape) {
-            if (this._body)
-                this.system.onRemove(entity, this);
+            if (this._body) {
+                this.system.removeBody(this._body);
+                this.system.destroyBody(this._body);
+
+                this._body = null;
+            }
 
             const mass = this._type === BODYTYPE_DYNAMIC ? this._mass : 0;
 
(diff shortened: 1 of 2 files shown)


slimbuck (approved):
Should we expect any other changes in behaviour with this change (other than the intended memory free)?

Issue #4878, opened 2022-11-23
Hardware instancing improvements


mvaligursky:
- We have a basic gpu instancing system, which allows world matrices of meshes to be specified in a vertex buffer, and a mesh at those location is instance-rendered on the GPU. 

We could do with some improvements:
- currently, only Matrix4 can be stored in the VB. Ideally we allow custom VB format to be used, with additional data (colors or anything else), and allow those to be accessible in the vertex shader code.
- currently the Matrix4 (4 x Vec4) is mapped to those attributes: SEMANTIC_ATTR12 .. 15, and ideally this is flexible, as well as location of additional attributes in the instancing VB.
- currently, when instancing is set on a MeshInstance, the frustum culling is turned off for it, and it gets submitted to the GPU at all times. We could:
  - allow user to per frame specify a single AABB for all instances, and use culling
  - we could even automatically calculate AABB in some cases (add up AABBs of meshes at their locations).

Related issue: https://github.com/playcanvas/engine/issues/3485

mvaligursky:
Regarding the culling, this might work, but ideally we have a nicer solution:
https://forum.playcanvas.com/t/how-to-check-which-meshes-are-getting-rendered/28519/10

Maksims:
A few other features that would be great for instancing solutions based on some use cases:
1. **Dynamic quantity vs Static quantity** - sometimes a number of instances are static, but sometimes it is desirable to have a dynamic number of instances which can increase/decrease per frame.
2. **Instance Object** - would be great to have a very slim instance object that would store associated data based on instance vb attribute mapping, in engine-friendly format. For example it could just have one vec3 for position and one vec3 for color. So the instance object would have two accessors for these properties in form of Vec3 and Color.
3. **Dynamic pool** - when using instancing with dynamically created/deleted instances, we had to create a pool of IDS, sorted incrementally, where we would always try to re-use smallest ID of an instance from a pool. Then sometimes we would check how many active instances there are and how many we are drawing, and re-sort them periodically if threshold is reached. This dynamic re-sorting allowed us to have a large buffer, with only small portion of it rendered when needed, so it reduces number of triangles dramatically. This also can be combined with culling techniques.
4. **More abstractions** - currently creating a custom instanced mesh requires definition of: TypedArray, VertexFormat, VertexBuffer, Shader, Material and MeshInstance. It would be great to have some abstractions to reduce number of low-level preparations, e.g. InstanceObject would simplify data setting.

[...]

PR #5804, opened 2023-11-08
Fix to splats frustum culling

mvaligursky:
- Instancing used to turn off frustum culling of the mesh instance. The API now has additional parameter to control this. If the case a user sets up a customAabb on the render / model component, they can enable culling for the mesh instancing. Note that there is no culling per instance, but a single custom aabb is used to cull all instances.
- this is used by gaussian splat rendering to skip rendering when splat is off-screen

### Changed API:

```
// note the new cull parameter, which defaults to false which was the original behaviour
MeshInstance.setInstancing(vertexBuffer, cull = false)
```

### DIFF
--- a/extras/splat/splat-container-resource.js
+++ b/extras/splat/splat-container-resource.js
@@ -87,13 +87,12 @@ class SplatContainerResource extends ContainerResource {
         });
 
         // set custom aabb
-        entity.render.customAabb = this.customAabb;
+        entity.render.customAabb = splat.aabb.clone();
 
         // when the render component gets deleted, destroy the splat instance
         entity.render.system.on('beforeremove', (entity, component) => {
             splatInstance.destroy();
         }, this);
-
 
         return entity;
     }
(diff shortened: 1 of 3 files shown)


LeXXik:
Ah, that was a thing. I had a screen space shader for a polyline that used instanced geometry. And when the culling was on, the skybox went black among other issues. I disabled culling as a temp solution, but should revisit it after this is released. Curious if adding a frustum culling per instance would be hard?

slimbuck (approved):

mvaligursky:
You can do this in the meantime:

```
meshInstance.setInstancing(vertexBuffer);
meshInstance.cull = true;
```

Per instance culling is doable, but:
- I'd need to assume instancing VB contains matrices. Currently (officially) it does, but I want to extend it to other things. And people already using it to store other things in those matrices. https://github.com/playcanvas/engine/issues/4878
- I'd need to have an internal VB to copy data to
- I'd need to cull per camera, so added cost, even though the user can often do a single cull with larger bounds for example

So I suspect we might not add per instancing culling, and leave users to do this if needed, and update their index buffer dynamically with the results.

PR #5807, opened 2023-11-08
Guards against zero canvas size

### LINKED ISSUE #5806: Error when canvas display is none
When the a canvas is hidden using css 

```javascript 
document.body.querySelector('#application-canvas').style.display = "none";
``` 
The graphics device throws

`Framebuffer creation failed with error code FRAMEBUFFER_INCOMPLETE_ATTACHMENT, render target: WebglFramebuffer MSAA`

See this repro https://launch.playcanvas.com/1406528?debug=true and https://forum.playcanvas.com/t/engine-1-66-assert-failed-framebuffer-creation-failed/33734/3

marklundin:
This adds a guard to not update the back buffer when the canvas size is 0, which happens when the css display is set to 'hidden'

Fixes #5806

### DIFF
--- a/src/platform/graphics/webgl/webgl-graphics-device.js
+++ b/src/platform/graphics/webgl/webgl-graphics-device.js
@@ -797,7 +797,8 @@ class WebglGraphicsDevice extends GraphicsDevice {
     updateBackbuffer() {
 
         const resolutionChanged = this.canvas.width !== this.backBufferSize.x || this.canvas.height !== this.backBufferSize.y;
-        if (this._defaultFramebufferChanged || resolutionChanged) {
+        const isCanvasNonZero = this.canvas.width > 0 && this.canvas.height > 0;
+        if (isCanvasNonZero && (this._defaultFramebufferChanged || resolutionChanged)) {
 
             // if the default framebuffer changes (entering or exiting XR for example)
             if (this._defaultFramebufferChanged) {


src/platform/graphics/webgl/webgl-graphics-device.js
```diff
@@ -797,7 +797,8 @@ class WebglGraphicsDevice extends GraphicsDevice {
     updateBackbuffer() {
 
         const resolutionChanged = this.canvas.width !== this.backBufferSize.x || this.canvas.height !== this.backBufferSize.y;
-        if (this._defaultFramebufferChanged || resolutionChanged) {
+        const isCanvasNonZero = this.canvas.width > 0 && this.canvas.height > 0;
+        if (isCanvasNonZero && (this._defaultFramebufferChanged || resolutionChanged)) {
```
mvaligursky, line 801: Have you tested this? I would expect this to fail as well, as the canvas is zero, our multi-sampled buffer is non-zero and we blit from one to the other.

marklundin: Yep, it doesn't throw the error. I think maybe because the canvas size can be zero, but the buffer can't

mvaligursky: I'm not super keen on this to be honest .. I think just reporting a warnOnce would be a better option. Otherwise the user might create a zero sized buffer and we'll keep rendering to some larger buffer. We should not hide this, just surface it.

marklundin: Right, yeah thats a valid point

PR #5808, opened 2023-11-08
Splat updates

slimbuck:
This is a followup PR to #5785 and adds a few bits required in SuperSplat:
- export a few more interfaces
- fix for camera sorting with modelView matrix
- early out sort when camera hasn't moved in webworker
- remove tiny splats, helps with very large scenes like garden.ply
- add `updated` event to sorter so editors can re-draw the scene even when camera is static


extras/splat/splat-sorter.js
```diff
@@ -163,7 +188,10 @@ class SplatSorter {
                 data: oldData
             }, [oldData]);
 
-            this.vertexBuffer.setData(newData);
+            setTimeout(() => {
```
willeastcott, line 191: Just wondering why the `setTimeout` is required...

willeastcott (approved):

extras/splat/splat-sorter.js
slimbuck: Without the setTimeout, the postMessage above will only be enqueued after setData has finished executing (setData can be slooow). With the setTimeout, postMessage is sent immediately and the worker can start sorting for the next round while setData completes.

I've confirmed this using Chrome profiler, but it's possible we can do better.

willeastcott: Worthy of a comment in the code?

willeastcott (approved):

mvaligursky (approved):

PR #5809, opened 2023-11-09
Expose custom shader modifications of the splat shader

mvaligursky:
- in vertex shader, both object and / or world space position can be modified
- in fragment shader, the color can be modified

https://github.com/playcanvas/engine/assets/59932779/e378dd88-56a7-43de-8782-b69773977188

### DIFF
--- a/extras/splat/splat-instance.js
+++ b/extras/splat/splat-instance.js
@@ -32,11 +32,12 @@ class SplatInstance {
 
     lastCameraDirection = new Vec3();
 
-    constructor(splat, cameraEntity, debugRender = false) {
+    constructor(splat, options) {
         this.splat = splat;
 
         // material
-        this.material = splat.createMaterial(debugRender);
+        const debugRender = options.debugRender ?? globalDebugRender;
+        this.material = splat.createMaterial(debugRender, options);
 
         // mesh
         const device = splat.device;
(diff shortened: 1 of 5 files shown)


extras/splat/splat-instance.js
```diff
@@ -32,11 +32,12 @@ class SplatInstance {
 
     lastCameraDirection = new Vec3();
 
-    constructor(splat, cameraEntity, debugRender = false) {
+    constructor(splat, options) {
```
slimbuck, line 35: Might be worth documenting these options.

extras/splat/splat-instance.js
```diff
@@ -32,11 +32,12 @@ class SplatInstance {
...
         this.splat = splat;
 
         // material
-        this.material = splat.createMaterial(debugRender);
+        const debugRender = options.debugRender;
+        this.material = splat.createMaterial(options);
```
slimbuck, line 40: It would be good if we also supported user passing in material.

extras/splat/splat.js
```diff
@@ -63,8 +63,8 @@ class Splat {
         this.centerTexture.destroy();
     }
 
-    createMaterial(debugRender = false) {
-        const material = createSplatMaterial(this.device, debugRender);
+    createMaterial(options) {
```
slimbuck, line 66: Think we should document options.

slimbuck (commented):
AWESOME

extras/splat/splat-instance.js
mvaligursky: Yep. At the moment none of the splat code has jsdocs, so that's all for a separate PR.

extras/splat/splat-instance.js
mvaligursky: separate PR. But not sure how that would work with a compute shader implementation on WebGPU, so I would not do that yet.

slimbuck (approved):

PR #5811, opened 2023-11-09
[Examples] Fix shader code indentation

kungfooman:
I was just testing the latest Gaussian Splat example and figured the shader code received a wrong indentation:

image

Basically it happened through the last line here:

image

It happens to fit twice and then the spaces-replacement would catch a line twice. I'm sure we could also use a multiline-replacement-regex searching for "start of line", but I don't know how well supported that is across browser etc., so I just kept it somewhat simple.


mvaligursky (approved):
perfect, thanks! I way prefer to indent the shaders to make it more readable when working on examples, than formatting it so that it looks fine in browser. Now we have both.

Issue #3759, opened 2021-12-01
rotate() behaves differently to rotateLocal() at 120fps


yaustar:
Thread: https://forum.playcanvas.com/t/rotate-having-issues-that-rotatelocal-does-not/23153

Test project that forces dt to be 1/120 to replicate the issue: https://launch.playcanvas.com/1285029?debug=true

https://user-images.githubusercontent.com/16639049/144244505-2205d545-5e09-4ec4-8e71-708c95af7e27.mp4

I can't see anything in the Engine code that is obvious to cause the rotation to go backwards at a particular point and only at 120fps.

At 60fps, they rotate fine

kungfooman:
Using RTI.js it is almost trivial to fix bugs like this now:

1) Go to http://pc.runtimetypeinspector.org/#/misc/hello-world

2) Replace app-on-update code with:

```js
box.setEulerAngles(70, 11, 84);
app.on('update', (/** @type {number} */ dt) => {
    dt = 1 / 120;
    box.rotate(5 * dt, 6 * dt, 7 * dt);
});
```
3) Figure out how RTI works

(I just did some website refactoing, so I hope the DNS is working for everyone, gotta add HTTPS again too)

willeastcott:
Ooooh, interesting! So does the `rotation` quat here need to be normalized after the `mul2` op?

https://github.com/playcanvas/engine/blob/5a6ff8668e0c0f3deb56ce89d4897b523b020abd/src/scene/graph-node.js#L1629

kungfooman:
> Ooooh, interesting! So does the `rotation` quat here need to be normalized after the `mul2` op?

Just did a little test:

```js
let diff = 0;
for (let i=0; i<100; i+=0.2)
for (let j=0; j<100; j+=0.2)
for (let k=0; k<100; k+=0.2)
{
    box.setEulerAngles(i, j, k);
    const tmp = Math.abs(1 - box.getRotation().length());
    if (tmp > diff) {
        console.log("new diff", tmp.toFixed(10), "for", {i, j, k});
        diff = tmp;
    }
}
```

Last outputs:

```
VM455:9 new diff 0.0000000125 for {i: 0, j: 52.0000000000002, k: 98.80000000000086}
VM455:9 new diff 0.0000000128 for {i: 0, j: 54.80000000000024, k: 97.60000000000085}
VM455:9 new diff 0.0000000129 for {i: 0.2, j: 56.80000000000027, k: 97.20000000000084}
[...]

PR #5817, opened 2023-11-10
RuntimeTypeInspector.js: Add RTI build target + Examples integration

### LINKED ISSUE #2529: Investigate function parameter validation for public Engine API
To help engine customers catch issues early instead of getting unhandled exceptions / undefined behavior, we should consider validating parameters passed into public API functions / possibly some internal functions.

Validation:
- should run in the debug mode only to not impact performance
- could be specified outside of the function and code generated in debug mode to do validation
- ideally without having to add #ifdef
- unified approach with unified error reporting
- ideally should check parameter types, unexpected null / undefined parameters and similar

And likely some other validations - please add comments with ideas and suggestions on how this could be implemented.

kungfooman:
Fixes #2529

I'm excited to introduce [RuntimeTypeInspector.js](https://runtimetypeinspector.org/), which I believe will benefit PlayCanvas as a whole and our debugging workflows in specific. RTI is designed to type-check the arguments of every function and method. It detects type errors while the engine is running in a way that goes beyond the capabilities of TypeScript, as demonstrated in [this video](https://www.youtube.com/watch?v=o5ipQe2rVKQ) (there is another one on the website debugging the Microsoft Holometric Video PC project).

During the development of RTI it already helped me to open issues or make PR's for various issues (e.g. NaN bugs).

I believe that RTI will help improve productivity and efficiency for both engine developers and users, by catching errors early on and reporting them in detail.

Runtime Type Inspector is implemented using Babel AST traversal with custom JSDoc parsing (powered by TypeScript itself for highest compatibility), customized to PlayCanvas requirements. That means for example supporting polymorphing `this` types.

TLDR: It adds type assertions for runtime type checking, for example:

image

Being able to set breakpoints exactly where the errors are happening for the first time helps to track down problems that are otherwise difficult to pinpoint (and saves a lot of time :sweat_smile:).

Try for yourself: http://pc.runtimetypeinspector.org/

Just make sure to pick the right JS context - the exampleIframe:

image

I'm working on this for quite some time by now, a mix of fun and frustration - rewriting code on the AST level first seemed tricky, but given enough time, it was a rather straightforward way to automate type validations. I'm excited to read feedback and suggestions. :pray: 


Maksims:
This is pretty cool!
A couple of questions:
1. Is there a way to suppress reporting more than N errors if they are originating from the same place? Spam of errors on every application update - is not much of use.
2. Is there a way to improve the stack within console log reporting so it points to the right place of the error? This will allow easier debugging with Dev Tools, but also provide a deep linking from Launcher to IDE.

kungfooman:
Hey @Maksims, thank you for the feedback!

> 1. Is there a way to suppress reporting more than N errors if they are originating from the same place? Spam of errors on every application update - is not much of use.

Yep, for this there is this button:

image

Currently it supports a "warn once" or "spam all the time" mode, but do you mean even more fine-grained control with N + source?

(I just uploaded to latest HEAD + fixed a issue with the spam-button-state, now it saves/restores properly from `localStorage`)

> 2\. Is there a way to improve the stack within console log reporting so it points to the right place of the error? This will allow easier debugging with Dev Tools, but also provide a deep linking from Launcher to IDE.

It's a `console.error` and if you click on it, the stack shows up:

image

Then you can click on the file location under `assertType` and you are at the "error source" and can just put a breakpoint:

image

So currently it takes two clicks to get to the place of the type error, but I'm not sure what you mean with deep linking from Launcher to IDE, could you elaborate a bit on that?

Thank you again!

Maksims:
> Yep, for this there is this button:

It did not work, unfortunately. I tried it, made no difference. Still, had errors spammed:
image

> It's a `console.error` and if you click on it, the stack shows up:

Ah, I did open dev tools *after* the errors were reported, it does not contain stack in such place, and all errors "originated" from the same place.
Is there a way to preserve their original stack? E.g. if the exceptions happen in the user script, but it reports first `playcanvas.rti.js:504`, then it is less clear where the error happened. This is actually an issue already with the current engine, where some errors like providing wrong arguments, are reported from within the engine (that's where they happen), but it is not straight away apparent to the user that it was due to passing wrong arguments. Especially when these arguments are not directly applied, but this is slightly unrelated.
It just highlights an issue: if an error is reported not from the user script, then they first think it is not their doing. So if an error happened somewhere within the user script or engine, the stack should specify errors from there, not from one single place.

Editor can try their best to filter through stack lines, and figure out if it can provide a deep link to a specific script in IDE, as it is really useful.

kungfooman:
> It did not work, unfortunately. I tried it, made no difference. Still, had errors spammed:

Hm, maybe the intention of the button isn't clear, when "Spam type reports" is checked it means that every type error is calling `console.error` (and if it is unchecked, the type errors are only logged once). If that doesn't work, it is probably because I updated the file and then it needs a Ctrl+R with DevTools open to clear the cache :thinking:

Thank you for pointing out the stack trace issue when DevTools isn't open from the beginning - I wasn't aware of that, so I will save the stack traces aswell for better inspection (would be nice to see in UI too).

> Editor can try their best to filter through stack lines, and figure out if it can provide a deep link to a specific script in IDE, as it is really useful.

Right, I made a little experiment what kind of information we can get out of the stack trace for Editor integration:

```js
const info = () => {
  try {
    throw new Error(); // Dummy exception to retrieve stack trace
  } catch(e) {
    const {stack} = e;
    if (stack.toLowerCase().includes('script')) {
      const lines = stack.split('\n');
      const i = lines.findIndex(_ => _.includes('ScriptComponent'));
      let scriptLine = lines[i - 1];
      scriptLine = scriptLine.replace(/at/, '').trim().split(' ')[0];
      console.warn('Type error in script: ', scriptLine);
    }
  }
}
info(), false; // false so it doesn't breakpoint here
```

This is "breakpoint code" we can enter here:

image

(clicking on the function source)

image

Then you have integrated the "breakpoint code" into the code flow and now you can play around with stack traces:

image

Thinking about the editor, we could just add "quick links" to the actual scripts now. Thank you for the suggestion!

kpal81xd:
Im unsure about directly including run time inspector specifics in the engine at best it can support loading a custom engine build but no custom code

kungfooman:
Hey @kpal81xd, can you elaborate more on what you mean by adding RTI specifics into the engine? Only a RTI build contains PC specific extra validations (e.g. checking that a `Mat4` contains no `NaN` values)

Thank you for checking this out already!

kungfooman:
I'm finally coming to a resolution here, it just feels so spammy for many examples still, even tho they run fine (simply based on wrong types in the engine). My main goal is "0 errors" - so developers can start editing examples and only see type errors they introduced themselves during their coding session.

If anyone has time, I would love to have these issues fixed:

 - http://pc.runtimetypeinspector.org/#/graphics/paint-mesh - scopeId seems to be a left over and has to be removed?
 - http://pc.runtimetypeinspector.org/#/user-interface/button-sprite - we are setting `length` of number arrays and move them around and calculating all kinds of stuff with them (leading to vectors filled with `undefined`) - useless calculations, so there is potential for a speedup aswell via small refactor
 - http://pc.runtimetypeinspector.org/#/graphics/lights-baked-a-o - filling of matrices with 'NaN' values
 - significantly reduce error counts, especially the ones that occur every frame

Any volunteers? :sweat_smile:

LeXXik:
A good cause 👍 I might tackle some of those next week.

kungfooman:
Now the vercel preview comes with RTI :see_no_evil:

willeastcott:
> Now the vercel preview comes with RTI 🙈

This is actually really cool - I can imagine going down a rabbit hole fixing a bunch of these issues one weekend! 🚀

willeastcott:
Hi @kungfooman, thanks for the enormous effort here, and sorry it has sat so long without a decision. Doing a pass over long-open PRs, so here is a clear one.

We are not going to merge an RTI build target into the engine repo. That is the position @kpal81xd stated in 2024 and it still holds: the engine ships `rel`/`dbg`/`prf`/`min` and we want to keep the build matrix and devDependency set to what we support directly. Since then the build has also moved from Rollup to esbuild (`build.mjs` + `utils/esbuild-build-target.mjs`) and the examples app to Vite, so the build and Examples parts of this branch (`rollup.config.mjs`, `Bottombar.mjs`, `rollup-build-examples.mjs`) no longer have anything to attach to. That is a big part of why it is now 92 commits and unmergeable.

What I would love to keep:

1. **The JSDoc/type fixes.** About fifteen of the `src/` changes in this branch are still valid against `main` (`Debug.assert` signature, `ViewCube` SVG element types, `AssetReference` null handling, `MeshInstance` nullable `mesh`/`material`, `AnimTarget` `'number'` type, `AnimCurveTarget` typedef, the `KeyboardEvent` optional params, `ShaderGenerator.definesHash`, `DeviceCache.get`, `ElementDragHelper._onMove`, `RenderComponent.asset`, `Texture` levels, `VertexIteratorAccessor` stale `scopeId` doc, `material.js` view formats, `AnimBinder.encode` example). Those are exactly the kind of PR we merge quickly. If you would rather not split them out yourself, say so and we will open a PR crediting this branch.

2. **RTI as an external tool.** The most useful form for the engine is one that does not require a merged build target: a plugin that consumes engine source or `playcanvas.mjs` plus `playcanvas.d.ts` from a fresh clone and produces an instrumented bundle on your side. Every type error that surfaces is a legitimate engine JSDoc bug we want reported, and I am very happy to keep merging those fixes.

3. **#2529 stays open.** This branch is the best evidence so far that runtime parameter validation catches real bugs. I will note there what we learned.

Closing this one so it does not keep drifting. Thanks again, and apologies for the wait.

kungfooman:
Hi Will, thank you for your quick decision after three years! I'm noticing some "inconsistencies", shall we re-evaluate?

> Those are exactly the kind of PR we merge quickly.

Check out https://github.com/playcanvas/engine/pull/7349 - you quickly resolved it in your own PR after only more than 19 months.

> That is a big part of why it is now 92 commits and unmergeable.

It's so unmergeable that I merged it in half an hour this morning. It runs happily on: https://pc.runtimetypeinspector.org/#/misc/hello-world

It collected 92 commits because:
 1) I merged back `main` 25 times
 2) I reacted to constructive feedback from @Maksims (very slowly, on the same day)
 3) I extended RTI all the time to find more bugs - the kind of deep bugs that were open issues for years because no one else could: https://github.com/playcanvas/engine/issues/3759

The third point is also exactly the reason why it doesn't make sense to have RTI living outside of this repository - I only found the bug through encoding PlayCanvas-specific mathematical invariants.

And to be clear, this PR is relatively **simple**. The fact it contains 33 file changes is because of your own behaviour. If you actually had done what you are proclaiming and promised, the 33 file changes would drastically decrease (it's all just spammy JSDoc type error fixes that you finally merged after years).

If you had *actually* merged my JSDoc fixes, I would have made more PRs and wouldn't be forced to put them in here as well (increasing commits and file count).

willeastcott:
By unmergeable, I mean by me.

My preference is still that RTI is an external tool.

kungfooman:
So because you can't merge it no one is allowed to benefit from structurally improved code quality?

The problem with "external tool" is what I explained. It's domain-specific knowledge that needs to be encoded to make it able to catch difficult-to-find bugs.

willeastcott:
> So because you can't merge it no one is allowed to benefit from structurally improved code quality?

It's just a datapoint that played into closing the PR. I'm trying to clean up the repo be decisive about issues and PRs that have been open and untouched for over a year.

> The problem with "external tool" is what I explained. It's domain-specific knowledge that needs to be encoded to make it able to catch difficult-to-find bugs.

I'm confident that this could be workable as an external tool. For example, an external tool can use the engine as a submodule and perform any tweaks necessary on the upstream fileset for the tool to work.

PR #5821, opened 2023-11-13
Update to Splat engine example + small fixes

mvaligursky:
- clone `centers` for each instance, as its ownership gets transferred to a worker
- generate unique shader id based on vs and fs
- updated visuals, to test integration with 3D scene and work of multiple splats together

Splat example:
Screenshot 2023-11-14 at 11 28 17

Splat-many example:
Screenshot 2023-11-14 at 11 29 44


LeXXik:
Looks great! The lighting on both Wills breaks the immersion, though :)

extras/splat/splat-container-resource.js
```diff
@@ -85,9 +85,19 @@ class SplatContainerResource extends ContainerResource {
         // set custom aabb
         entity.render.customAabb = splat.aabb.clone();
 
+        // HACK: store splat instance on the render component, to allow it to be destroye in the following code
```
willeastcott, line 88:
```suggestion
        // HACK: store splat instance on the render component, to allow it to be destroyed in the following code
```

willeastcott (approved):

PR #5826, opened 2023-11-15
[Fix] Fix texture / render target resize on WebGL2

mvaligursky:
- recent optimization in #5759 introduced the error by storing the data on an incorrect object (Texture vs WebglTexture)

### DIFF
--- a/src/platform/graphics/webgl/webgl-texture.js
+++ b/src/platform/graphics/webgl/webgl-texture.js
@@ -345,7 +345,7 @@ class WebglTexture {
                         device.setUnpackFlipY(false);
                         device.setUnpackPremultiplyAlpha(texture._premultiplyAlpha);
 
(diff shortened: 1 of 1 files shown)


src/platform/graphics/webgl/webgl-texture.js
```diff
@@ -574,7 +574,7 @@ class WebglTexture {
         texture._gpuSize = texture.gpuSize;
         texture.adjustVramSizeTracking(device._vram, texture._gpuSize);
 
-        texture._glCreated = true;
+        texture.this = true;
```
Maksims, line 577: Probably a typo here.

mvaligursky: yeah, thanks!

slimbuck (approved):

PR #5830, opened 2023-11-17
New GraphicsDevice.getRenderableHdrFormat function

mvaligursky:
## New public API:

Screenshot 2023-11-17 at 14 02 58

New public constant: `BlendState.ADDBLEND` - predefined constant for additive blending

- few small fixes
- removed no longer needed `WebglGraphicsDevice. getHdrFormat`

### DIFF
--- a/src/platform/graphics/graphics-device.js
+++ b/src/platform/graphics/graphics-device.js
@@ -10,7 +10,7 @@ import {
     BUFFER_STATIC,
     CULLFACE_BACK,
     CLEARFLAG_COLOR, CLEARFLAG_DEPTH,
-    PRIMITIVE_POINTS, PRIMITIVE_TRIFAN, SEMANTIC_POSITION, TYPE_FLOAT32
+    PRIMITIVE_POINTS, PRIMITIVE_TRIFAN, SEMANTIC_POSITION, TYPE_FLOAT32, PIXELFORMAT_111110F, PIXELFORMAT_RGBA16F, PIXELFORMAT_RGBA32F
 } from './constants.js';
 import { BlendState } from './blend-state.js';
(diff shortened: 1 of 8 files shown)


src/platform/graphics/graphics-device.js
```diff
@@ -809,6 +817,47 @@ class GraphicsDevice extends EventHandler {
...
+     * - {@link PIXELFORMAT_111110F}
+     * - {@link PIXELFORMAT_RGBA16F}
+     * - {@link PIXELFORMAT_RGBA32F}
+     *
+     * @param {boolean} [filterable] - If true, the format aso needs to be filterable. Defaults to
+     * true.
+     * @returns {number|undefined} The first supported renderable HDR format or undefined if none is
+     * supported.
+     */
+    getRenderableHdrFormat(formats = [PIXELFORMAT_111110F, PIXELFORMAT_RGBA16F, PIXELFORMAT_RGBA32F], filterable = true) {
```
willeastcott, line 835: Just wondering whether for API simplicity whether it's better to just have:

```
const supported = device.isRenderable(PIXELFORMAT_111110F);
```

mvaligursky: These are optional parameters, and typically the user would not pass any, to get any format from smallest size to largest.

It's similar to the device creation where we specify GL1, GL2 and GPU array in any other. Simple to call without having to implement logic on the user side.

mvaligursky: I already call this function twice in this code, and one more time in follow up PR .. and it's convenient to not have to implement format logic at each place.

slimbuck (approved):

willeastcott (approved):

Issue #5187, opened 2023-03-24
Create modern looking bloom post-effect


mvaligursky:
Currently, the engine only has a simple bloom effect, and a modern implementation is needed.

references:
https://learnopengl.com/Guest-Articles/2022/Phys.-Based-Bloom
https://www.froyok.fr/blog/2021-09-ue4-custom-lens-flare/
https://twitter.com/iced_coffee_dev/status/1617580408576565251?t=OYAlQMbMAx4PYCKlx7HplQ&s=03

SashaRX:
I experimented with it on this tutorial   https://playcanv.as/b/491d8a85

SashaRX:
> That looks cool, @SashaRX. Did you experience any problems or limitations using the technique from that article?

In general, everything went really fast, despite my poor programming skills). I still haven't worked out how to use blendState and pc.drawQuadWithShader. Here it uses render to hdr render target, I use linear space, it's possible that this is not correct everywhere, since I didn't specify that albedo should be srgb for example, but it seems to be correct.

MAG-AdrianMeredith:
Looks good I'll be interested to see how that compares to our version.  I made a custom enhancement to playcanvas hdr mode so that it renders to 11bits not 16bit.  This was a lot faster to render on my m1 macbook (3-4ms).  The one issue still remaining is alpha blended objects look different like they are being double tonemapped or something....

mvaligursky:
I'm definitely keen to have this working on 11bits as well, to have a wider device support as well.
Both results look fantastic.

GSterbrant:
I am also adding this link to the UE current bloom method for visibility: https://www.froyok.fr/blog/2021-12-ue4-custom-bloom/

mvaligursky:
Another awesome work on this:
https://forum.playcanvas.com/t/physically-based-hdr-bloom/31759

mvaligursky:
@GSterbrant - you did some work on this as well, could you please create a PR for this (just attach the file in current state), link the PR here and close the PR (for now), just to keep reference to that file.

mvaligursky:
https://twitter.com/turanszkij/status/1680836739399667713

MAG-AdrianMeredith:
I like that custom ue4 one.  I believe I can share the source to my implementation if you're interested though its still not 100% imo

mvaligursky:
Ideally we'd implement in using this system when ready: https://github.com/playcanvas/engine/issues/5687

Issue #5687, opened 2023-09-28
Implement a system to describe the custom frame rendering using render passes


mvaligursky:
Current situation:
- Render of the frame is handled by the forward renderer, which pretty much forces few render passes to render if more of less fixed order, and handles shadows, rendering to camera, grab passes, post effects, main forward pass rendering and not many others. 

Goal:
- Make clearly defined interface to describe a frame / part of the frame using modular render passes. This will be a unified interface, which can be used to describe a post-effect, or a way the camera renders it's content, or even a whole frame. Initially the interface will be internal, to allow the engine ti implement many new features, but when it stabilises, the goal is to expose it as an API to give users controls of their rendering.

High level design (subject to change)
- each render pass is a class which extends the RenderPass, and implements its callbacks as needed (before, execute, after). It also stores dependent render passes, that execute after (and possibly even before if needed) the render pass itself.
- this will allow us to write many isolated render passes, that implement some basic functionality, and combine them together to create more complex render passes / post-effects or even whole frames.

Example of few render passes and their combination to a frame:
`RenderPassForward` - renders opaque / transparent meshes to render target as needed
`RenderPassExtractBright` - a render pass which takes source texture, and extracts the bright parts of it to the destination texture (render target).
'RenderPassBlur' - a render pass that blurs source texture into a render target texture.
'RenderPassBloom' - a combined render pass, which first uses RenderPassExtractBright, and after that runs the blur pass(s) to create a bloom effect, stored in a texture (without writing it back to the framebuffer).
'RenderPassDepth' - this is already created, allows the camera to be rendered out to generate depth buffer.
'RenderPassSSAO' - takes source depth texture, generates occlusion buffer, and after that runs multiple RenderPassBlur passes to blur it. The result is ssao buffer that can be used at a later stage.
'RenderPassTestFrame' - an example of putting those things together to create a frame. It would start with RenderPassDepth and RenderPassSSAO to generate occlusion texture, followed by RenderPassForward to render the scene, then RenderPassBloom to generate the bloom buffer, and finally it will run some RenderPassCompose to blend the bloom on top of the output of the RenderPassForward, into the backbuffer.

Many details are not described, but the flexibility of the system should be visible here.

Related to https://github.com/playcanvas/engine/issues/5686 and https://github.com/playcanvas/engine/issues/4271

MAG-AdrianMeredith:
[...]

PR #5833, opened 2023-11-20
HDR physically based bloom + render pass examples

mvaligursky:
Fixes: https://github.com/playcanvas/engine/issues/5187
Related Issue: https://github.com/playcanvas/engine/issues/5687

## New Engine API (private for now, while in preview)
- **CameraComponent.renderPasses** - an array of render passes the camera will use to render, instead of its default rendering.
- **RenderPassShaderQuad** - new render pass, a base class for render passes that render quad using a shader.

## Extras
Multiple render passes in `/extras`, used to implement HDR Bloom:
- RenderPassUpscale/Downscale - quad based passes, used to upscale and downscale the image
- RenderPassBloom - render pass using upscaling and downscaling to generate bloom texture
- RenderPassComposite - a render pass compositing the scene texture with bloom texture. This will be expanded by multiple post-effects as part of the same shader / render pass: color-corrections, distortion, screen-texture and many others.

## Examples
Two examples:
- RenderPassExample: simple example showing a custom render pass (tint) and how to use it to tint the rendered scene
Screenshot 2023-11-20 at 16 05 25

- PostprocessingExample: example using new render passes from `/extras` to set up a scene with HDR bloom. Control over render passes allows user to render 3d scene to lower resolution HDR texture, apply bloom and then render UI on top of it in full resolution.
Screenshot 2023-11-20 at 16 04 52


slimbuck (approved):

MAG-AdrianMeredith:
Very nice, render pass implementation over post effects looks nice.  Look forward to comparing it to ours.

mvaligursky:
Yep, the render passes are a really nice way to put this together.
If you at some point write another bloom implementation using passes, we can easily add that to extras too, no problem with that.

MAG-AdrianMeredith:
@mvaligursky Have you tried this with a directional light present?  I'm getting errors here 
image
the shadowCamera.renderTarget is undefined.   it works otherwise and could just be me but its worth checking

MAG-AdrianMeredith:
image

PR #5836, opened 2023-11-21
Fix for video textures

slimbuck:
Fix due to #5759 for video textures.

### DIFF
--- a/src/platform/graphics/webgl/webgl-texture.js
+++ b/src/platform/graphics/webgl/webgl-texture.js
@@ -537,8 +537,10 @@ class WebglTexture {
                             mipObject
                         );
 
-                        texture._width = w;
-                        texture._height = h;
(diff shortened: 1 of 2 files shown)


src/platform/graphics/webgl/webgl-texture.js
```diff
@@ -512,10 +512,13 @@ class WebglTexture {
                         }
                     }
 
+                    const w = mipObject.width || mipObject.videoWidth;
```
mvaligursky, line 515: ?? would probably be better

mvaligursky (approved):

src/platform/graphics/webgl/webgl-texture.js
slimbuck: For video textures width and height are 0.

PR #5842, opened 2023-11-22
Ensure screen scale is always valid

### LINKED ISSUE #5841: 2D screen broken when setting canvas display to none
Since 1.66.0, 2D screens are no longer working after hiding the canvas.

### Steps to Reproduce

1. Add a 2D screen to scene with some elements
2. Hide the canvas using `document.getElementById('application-canvas').style.setProperty('display', 'none')`
3. Unhide the canvas using `document.getElementById('application-canvas').style.removeProperty('display')`
4. The 2D screen is now gone forever

3D screens are still working correctly though.

NewboO:
Fixes #5841 

When hiding canvas, resolution is set to 0x0, thus screen scale computation is giving NaN and it will never recover from it. This PR ensures the scale can't be NaN but default to 1 instead.

However I'm not sure about coding standards here. Is this syntax acceptable or does it need some explicit early exit condition by checking resolution components (and maybe referenceResolution components too)?

### DIFF
--- a/src/framework/components/screen/component.js
+++ b/src/framework/components/screen/component.js
@@ -124,9 +124,9 @@ class ScreenComponent extends Component {
         // Using log of scale values
         // This produces a nicer outcome where if you have a xscale = 2 and yscale = 0.5
         // the combined scale is 1 for an even blend
-        const lx = Math.log2(resolution.x / referenceResolution.x);
-        const ly = Math.log2(resolution.y / referenceResolution.y);
-        return Math.pow(2, (lx * (1 - this._scaleBlend) + ly * this._scaleBlend)) || 1;
+        const lx = Math.log2((resolution.x || 1) / referenceResolution.x);
+        const ly = Math.log2((resolution.y || 1) / referenceResolution.y);
+        return Math.pow(2, (lx * (1 - this._scaleBlend) + ly * this._scaleBlend));
     }
 
     _onResize(width, height) {
--- a/src/framework/components/screen/component.js
+++ b/src/framework/components/screen/component.js
@@ -126,7 +126,7 @@ class ScreenComponent extends Component {
         // the combined scale is 1 for an even blend
         const lx = Math.log2(resolution.x / referenceResolution.x);
         const ly = Math.log2(resolution.y / referenceResolution.y);
-        return Math.pow(2, (lx * (1 - this._scaleBlend) + ly * this._scaleBlend));
+        return Math.pow(2, (lx * (1 - this._scaleBlend) + ly * this._scaleBlend)) || 1;
     }
 
     _onResize(width, height) {


src/framework/components/screen/component.js
```diff
@@ -126,7 +126,7 @@ class ScreenComponent extends Component {
         // the combined scale is 1 for an even blend
         const lx = Math.log2(resolution.x / referenceResolution.x);
         const ly = Math.log2(resolution.y / referenceResolution.y);
-        return Math.pow(2, (lx * (1 - this._scaleBlend) + ly * this._scaleBlend));
+        return Math.pow(2, (lx * (1 - this._scaleBlend) + ly * this._scaleBlend)) || 1;
```
mvaligursky, line 129: I would probably prefer to change the two lines above to avoid the NaN completely, could you test that please.
```
const lx = Math.log2(resolution.x / (referenceResolution.x || 1));
```

mvaligursky:
Nice find!

src/framework/components/screen/component.js
NewboO: The original issue is caused by `resolution` but I did a few more tests with `referenceResolution` which is user defined and a negative value can also break the 2D screen.

For completeness, here is what happens at moment:
```
                          |    resolution.x < 0    |    resolution.x = 0   |   resolution.x > 0
--------------------------------------------------------------------------------------------------
referenceResolution.x < 0 | OK (but can't happen?) | NaN if scaleBlend = 1 |         NaN
                          |                        |      0 otherwise      |
--------------------------------------------------------------------------------------------------
referenceResolution.x = 0 |           NaN          |          NaN          | NaN if scaleBlend = 1
                          |                        |                       |  Infinity otherwise
--------------------------------------------------------------------------------------------------
referenceResolution.x > 0 |           NaN          | NaN if scaleBlend = 1 |     OK (obviously)
                          |                        |      0 otherwise      |
```
And same for `y` components too except for `scaleBlend` condition (0 instead of 1).

It should be noted that when it returns 0, it also breaks the 2D screen while Infinity works (though it doesn't show anything until it's back to a more usual value).

I'm not sure how negative and 0 values in `referenceResolution` should be handled (I don't see how `resolution` could be negative at all but maybe it should be considered too), but at the moment nothing prevents the user to do that (not even a warning in the editor) so at least it should be possible to recover from it.

mvaligursky: I would not worry about the case the user sets a referenceResolution to a negative value. You can pass incorrect values to many things in the engine and break it completely. We only need to handle the resolution being 0 really.

mvaligursky (approved):

src/framework/components/screen/component.js
NewboO: Ok, resolution || 1 then.

PR #5847, opened 2023-11-25
[Examples] Skip hidden examples from being listed

### LINKED ISSUE #5812: [Examples] Make HIDDEN flag to hide examples from build
We used to have a system where HIDDEN flag would allow the example for local development, but exclude it from the published build, which is used to submit work-in-progress examples.

`lit-material.mjs` currently has this 
`    static HIDDEN = true;`
but it is not used, and the example is excluded instead manually this way:
```
// export * from "./lit-material";   // note that this is also excluded in read-dir.mjs
```

Perhaps we could:
- ignore examples with HIDDEN when using `npm run build`
- include them when doing `npm run develop`

kungfooman:
Fixes #5812 

### DIFF
--- a/examples/scripts/example-data.mjs
+++ b/examples/scripts/example-data.mjs
@@ -18,7 +18,8 @@ for (const category_ in realExamples) {
     exampleData[category] = {};
     const examples = realExamples[category_];
     for (const exampleName_ in examples) {
-        if (examples[exampleName_].HIDDEN) {
+        const release = process.env.NODE_ENV !== 'development';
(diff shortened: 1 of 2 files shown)


mvaligursky (approved):

kungfooman:
> Nice one @kungfooman . Does it show the examples when `npm run develop` and only exclude them when `npm run build`?

Hey @mvaligursky, thank you for pointing that out! I will add that asap

mvaligursky:
Amazing, thanks!

PR #5848, opened 2023-11-25
WebXR Depth Camera

Maksims:
~~This PR is dependent on https://github.com/playcanvas/engine/pull/5786 (Camera Access).~~

With release of Quest 3, Depth Sensing API has been added to its browser, which also affected the [Specs](https://www.w3.org/TR/webxr-depth-sensing-1/). Previously only one type of devices were capable of providing depth information: Android Phones, which assumed a single monoscope view. So XrDepthSensing was designed with assumption of a single view. Now Quest 3 exposes depth data per view, which is multiple depth textures. This does require a redesign of the access to depth data. So part of this API we are deprecating XrDepthSensing in favor of XrView access to depth information.

This API is pretty low-level. It also provides a more precise real-time depth based hit testing using getDepth method, which is more precise than XrHitTest and reacts to moving objects in real world.

In the future PR I will add material flag for "Depth Occlusion", that will automatically occlude virtual geometry with real world geometry, with underlying implementation that makes use of Depth Sensing for users matter of just a checkbox.

### New APIs:

```javascript
// pc.XrViews
views.supportedDepth // true if depth information is supported
views.availableDepth // true if depth information is available.
// This information is only available after XR session has started.
// It can be supported but not available due to lack of session features request, user permissions
// or underlying system capabilities
views.depthPixelFormat // PIXELFORMAT_LA8 or PIXELFORMAT_F32 based on underlying AR capabilities

// pc.XrView
view.textureDepth // returns a texture with depth information. If depth information is not available it will be null
view.depthUvMatrix // Mat4 that should be used to transform depth texture for a correct projection
view.depthValueToMeters // a coefficient for depth pixel information to be multiplied by, which will convert raw depth value to meters
view.getDepth(u, v) // returns a float in meters of based on depth in a view or null if depth information is not available at the present frame. U and V arguments are 0..1 floats, where 0,0 is top-left corner.
```

### Examples:

1. **Camera Depth** - renders a plane in front of a camera with custom shader that supports CPU and GPU paths to render a depth texture.
2. **Depth Sensing Placer** - placing an object based on depth sensing information.

### TODO:

- [x] GPU path (waiting for Quest 3 to either deploy OS v60 that makes Depth Sensing public, or get access to OS beta for development).
- [x] Example with depth map
- [x] Example with a depth hit test (object placer)


Maksims:
This PR is ready for a review.

Added an example for object placement using depth information. Bear in mind that it is only available when depth sensing uses CPU path. GPU path does not provide such functionality, but potentially can be emulated by reading pixel from a depth texture in a separate PR.

This is useful technique for more precise object placement when hit-test is not providing reliable information. I've also used both of those techniques, and pick the closest point which provided better results for reliable objects placement.

src/scene/renderer/forward-renderer.js
```diff
@@ -628,6 +628,7 @@ class ForwardRenderer extends Renderer {
                     this.viewId3.setValue(view.viewMat3.data);
                     this.viewProjId.setValue(view.projViewOffMat.data);
                     this.viewPosId.setValue(view.positionData);
+                    this.viewIndex.setValue(v);
```
mvaligursky, line 631:
```suggestion
                    this.viewIndexId.setValue(v);
```

src/scene/renderer/renderer.js
```diff
@@ -220,6 +220,7 @@ class Renderer {
         this.farClipId = scope.resolve('camera_far');
         this.cameraParams = new Float32Array(4);
         this.cameraParamsId = scope.resolve('camera_params');
+        this.viewIndex = scope.resolve('view_index');
```
mvaligursky, line 223:
```suggestion
        this.viewIndexId = scope.resolve('view_index');
```

src/framework/xr/xr-view.js
```diff
@@ -314,17 +518,48 @@ class XrView {
...
     destroy() {
+        this._depthInfo = null;
+
         if (this._textureColor) {
             this._textureColor.destroy();
             this._textureColor = null;
         }
 
+        if (this._textureDepth) {
+            this._textureDepth.destroy();
```
mvaligursky, line 559: So we override `this._textureDepth.impl._glTexture` and then destroy it.
1. Was _glTexture already allocated before? I suspect not, so we're not leaking it.
2. can we destroy _glTexture provided by XR?

mvaligursky (commented):
Very nice PR @Maksims . I added few minor comments / questions, but this is pretty much ready for merge. Thanks!

src/framework/xr/xr-view.js
```diff
@@ -117,12 +149,45 @@ class XrView {
...
 
-                this._manager.app.graphicsDevice?.on('devicelost', this._onDeviceLost, this);
+        if (this._manager.views.supportedDepth && this._manager.views.availableDepth) {
+            this._textureDepth = new Texture(device, {
+                format: this._manager.views.depthPixelFormat,
+                arrayLength: (viewsCount === 1) ? 0 : viewsCount,
+                mipmaps: false,
+                addressU: ADDRESS_CLAMP_TO_EDGE,
+                addressV: ADDRESS_CLAMP_TO_EDGE,
+                minFilter: FILTER_LINEAR,
```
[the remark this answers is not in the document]
Maksims: The depth texture always comes with lower resolution than framebuffer resolution of each eye. So interpolation helps to smooth out the edges. But now you mentioned it, if we don't upload actual texture, then we are not setting the `texParameteri` of it, right? So these flags pretty much do nothing?

src/framework/xr/xr-view.js
Maksims: We don't allocate it indeed, but for the safety, when the view is destroyed (end of the session), I do destroy that texture also. Subsequent sessions work well, so it seems to be good.

mvaligursky (approved):

PR #5850, opened 2023-11-25
Simplify and optimize some GraphNode find functions

NewboO:
Follow-up of #4603

This PR simplify `GraphNode.find` and `GraphNode.findOne`.
It also avoids creating and merging lots of arrays in `GraphNode.find` which means less garbage.
In addition, `GraphNode.findByName` now simply calls `GraphNode.findOne` so the traversing logic isn't duplicated anymore.

### DIFF
--- a/src/scene/graph-node.js
+++ b/src/scene/graph-node.js
@@ -24,6 +24,27 @@ const target = new Vec3();
 const up = new Vec3();
 
 /**
+ * Helper function that handles signature overloading to receive a test function.
+ *
+ * @param {FindNodeCallback|string} attr - Attribute or lambda.
+ * @param {*} [value] - Optional value in case of `attr` being a `string`
+ * @returns {FindNodeCallback} Test function that receives a GraphNode and returns a boolean.
+ * @ignore
+ */
+function createTest(attr, value) {
+    if (attr instanceof Function) {
+        return attr;
+    }
+    return (node) => {
+        let x = node[attr];
+        if (x instanceof Function) {
+            x = x();
+        }
+        return x === value;
+    };
+}
+
+/**
  * Callback used by {@link GraphNode#find} and {@link GraphNode#findOne} to search through a graph
  * node and all of its descendants.
  *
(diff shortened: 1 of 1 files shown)


Maksims:
Do you have a benchmarks with various hierarchy cases?

src/scene/graph-node.js
```diff
@@ -534,40 +555,13 @@ class GraphNode extends EventHandler {
...
 
-            for (let i = 0; i < len; ++i) {
-                const descendants = this._children[i].find(attr, value);
-                if (descendants.length)
-                    results = results.concat(descendants);
-            }
-        }
+        this.forEach((node) => {
+            if (test(node))
+                results.push(node);
```
slimbuck, line 563: shouldn't this recurse into descendants?

src/scene/graph-node.js
```diff
@@ -597,39 +591,16 @@ class GraphNode extends EventHandler {
...
-        let result = null;
 
-        if (attr instanceof Function) {
-            const fn = attr;
+        if (test(this))
+            return this;
 
-            result = fn(this);
+        for (let i = 0; i < len; ++i) {
+            const result = this._children[i].findOne(test);
```
slimbuck, line 601: why not recurse on a function that directly accepts the test? this way we don't have to keep checking if it's a function at each and every invocation?

src/scene/graph-node.js
mvaligursky: check `this.forEach` implementation

NewboO:
> Do you have a benchmarks with various hierarchy cases?

I ran somes on one of my projects which has a hierarchy of ~800 nodes.
Benchmarked from root and from a deeper node (only 80 descendants), finding by name both existing (100 occurrences to return from root, 5 from the deeper node) and non-existing:

```
GraphNode.find, existing name

                       |  old  |  new
--------------------------------------
800+ nodes (100k runs) |  5.3s |  2.5s
--------------------------------------
80 nodes (1M runs)     | 0.98s | 0.75s


GraphNode.find, non-existing name

                       |  old  |  new
--------------------------------------
800+ nodes (100k runs) |  4.6s |  2.5s
--------------------------------------
80 nodes (1M runs)     | 0.63s | 0.70s


GraphNode.findOne, existing name relatively deep

                       |   old  |   new
----------------------------------------
800+ nodes (100k runs) |  0.29s |  0.17s
----------------------------------------
80 nodes (1M runs)     | 0.027s | 0.036s


GraphNode.findOne, non-existing name

                       |  old  |  new
--------------------------------------
800+ nodes (100k runs) |  4.5s |  2.7s
--------------------------------------
80 nodes (1M runs)     | 0.56s | 0.69s
```

So it may be sometimes a bit slower when only a few nodes need to be checked, but it gets way faster on bigger hierarchies.

src/scene/graph-node.js
```diff
@@ -23,6 +23,50 @@ const matrix = new Mat4();
...
+ * node is found.
+ * @ignore
+ */
+function findNode(node, test) {
+    if (test(node))
+        return node;
+
+    const len = node._children.length;
+    for (let i = 0; i < len; ++i) {
+        const result = findNode(node._children[i], test);
```
slimbuck, line 62: Since in this code we're going for _super speed_ it probably makes sense to cache `node._children` in a local variable above.

slimbuck:
This PR seems to me a wonderful example of how simplifying code can make it more readable, maintainable and run faster all at the same time. Great job @NewboO !!

src/scene/graph-node.js
NewboO: Couldn't say how much sense it makes nowadays as it may depends on the JS engine but it probably doesn't hurt (on my end it's sometimes faster sometimes not, always by a very small margin). I also noticed `forEach` weren't caching length and using `i++` instead of `++i` (which on my end seems to make no difference on chrome but sometimes a big one on firefox).

mvaligursky (approved):
Great work! I'm happy to merge this. @slimbuck ?

PR #5851, opened 2023-11-25
Add Entity.findScript[s] functions

NewboO:
Follow-up of #4603

This PR adds `Entity.findScript` and `Entity.findScripts` based on what already exists for components.

### DIFF
--- a/src/framework/entity.js
+++ b/src/framework/entity.js
@@ -383,15 +383,15 @@ class Entity extends GraphNode {
      * Search the entity and all of its descendants for the first script instance of specified type.
      *
      * @param {string|Class<import('./script/script-type.js').ScriptType>} nameOrType - The name or type of {@link ScriptType}.
-     * @returns {import('./script/script-type.js').ScriptType} A script instance of specified type, if the entity or any of its descendants
+     * @returns {import('./script/script-type.js').ScriptType|undefined} A script instance of specified type, if the entity or any of its descendants
      * has one. Returns undefined otherwise.
      * @example
      * // Get the first found "playerController" instance in the hierarchy tree that starts with this entity
      * var controller = entity.findScript("playerController");
      */
     findScript(nameOrType) {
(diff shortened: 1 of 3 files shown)


src/framework/entity.js
```diff
@@ -379,6 +379,36 @@ class Entity extends GraphNode {
...
+     * @param {string|Class<import('./script/script-type.js').ScriptType>} nameOrType - The name or type of {@link ScriptType}.
+     * @returns {import('./script/script-type.js').ScriptType} A script instance of specified type, if the entity or any of its descendants
+     * has one. Returns undefined otherwise.
+     * @example
+     * // Get the first found "playerController" instance in the hierarchy tree that starts with this entity
+     * var controller = entity.findScript("playerController");
+     */
+    findScript(nameOrType) {
+        const entity = this.findOne(node => node.c && node.c.script && node.c.script.has(nameOrType));
+        return entity && entity.c.script.get(nameOrType);
```
mvaligursky, line 394: could probably be shortened to
```suggestion
        const entity = this.findOne(node => node.c?.script?.has(nameOrType));
        return entity?.c.script.get(nameOrType);
```

src/framework/entity.js
```diff
@@ -379,6 +379,36 @@ class Entity extends GraphNode {
         });
     }
 
+    /**
+     * Search the entity and all of its descendants for the first script instance of specified type.
+     *
+     * @param {string|Class<import('./script/script-type.js').ScriptType>} nameOrType - The name or type of {@link ScriptType}.
+     * @returns {import('./script/script-type.js').ScriptType} A script instance of specified type, if the entity or any of its descendants
```
mvaligursky, line 386: add `|undefined` to the return type please. (we should fix the findComponent too)

src/framework/entity.js
mvaligursky: entity can be undefined, so `entity?.` should be used

mvaligursky (approved):

mvaligursky:
Thanks!

Issue #5846, opened 2023-11-25
Examples Error reporting


Maksims:
Currently if there is an exception in example code, it does not report a sensible stack trace and it is harder to debug due to that.

It would be great if error reporting would show a good stack trace, presented in UI within examples page with an ability to click and navigate to line of code even inside of code editor. Just like Launcher does with IDE (Editor).

kungfooman:
Maybe just show it directly as editor annotation without any extra clicks? I implemented something similiar for another live editor:

https://github.com/enkimute/ganja.js/pull/108

Maksims:
Yes, that definitely will be useful also.
When remote debugging with mobile devices (phones, VR headsets), pre process is usually involves Chrome/Safari remote debugger with access to console logs. So a stack trace with correct error message and line of code is very useful.

kungfooman:
I worked something out:

```js
const iframe = document.getElementById("exampleIframe");
iframe.contentWindow.onerror = function (message, file, line, column, errorObj) {
  // Chrome only atm, but we would add a try/catch in PR
  if (errorObj) {
    console.log('Error stack: ', errorObj.stack);
  }
  console.log({message, file, line, column, errorObj});
  line -= 2; // map to correct line in editor
  const lineText = editor.getModel().getLineContent(line);
  const oldDecorators = [];
  const newDecorator = {
    range: new monaco.Range(line, 0, line, lineText.length),
    options: {
      // isWholeLine: true,
      className: 'squiggly-error',
      hoverMessage: {value: message},
    }
  };
  editor.deltaDecorations(oldDecorators, [newDecorator]);
};
```

If you insert some test code that fails in the Monaco editor:

```js
    setTimeout(() => {
        1n / 2;
    }, 1000);
```

Then Shift+Enter to hot-reload the example:

![image](https://github.com/playcanvas/engine/assets/5236548/d9ea8ba2-cb91-4b5f-aa8c-21c1cb6012fe)

Console looks like:

![image](https://github.com/playcanvas/engine/assets/5236548/21b7cea0-9fbb-4981-8296-8db8fdfd1e3a)

So this works at least if the error happens in the code that happens to be in the Monaco editor. But we can also trigger errors from within:

```js
    app.mouse = {}; // trying to call update somewhere, so it triggers an error
```

So I pretty much see two cases:

1) Error happens in Monaco editor: display as code decoration and UI
2) Error happens somewher else: only UI

[...]

PR #5854, opened 2023-11-26
[Examples] Fix error handling and display errors in Monaco editor

kungfooman:
Partly implements #5846

This PR is implementing what's needed to deal with errors in the UI, so if someone else wants to add nicer error displaying, they can do so quite easily, just listen to: `window.addEventListener('exampleError', event => console.log("TEST", event.detail));`

So far errors are only added in the UI:

image

The `allowRestart` variable also wasn't reset in case of an error, so once an error happened, the hot-reload wouldn't do anything afterwards.

### DIFF
--- a/examples/scripts/generate-standalone-files.mjs
+++ b/examples/scripts/generate-standalone-files.mjs
@@ -294,23 +294,33 @@ ${exampleClass.example.toString()}
                 // console.log("Dispatch exampleLoading!");
                 const event = new CustomEvent("exampleLoading"); // just notify to clean UI, but not during hot-reload
                 window.top.dispatchEvent(event);
+            } else {
+                window.top.dispatchEvent(new CustomEvent("exampleHotReload"));
             }
-            const example = resolveFunction(files['example.mjs']);
             files['example.mjs'] = files['example.mjs'].toString();
-            const app = await example({
-                canvas,
-                deviceType,
-                data,
-                assetPath,
-                scriptsPath,
-                ammoPath,
-                basisPath,
-                dracoPath,
-                glslangPath,
-                twgslPath,
-                pcx,
-                files,
-            });
+            let app;
+            try {
+                const example = resolveFunction(files['example.mjs']);
+                app = await example({
+                    canvas,
+                    deviceType,
+                    data,
+                    assetPath,
+                    scriptsPath,
+                    ammoPath,
+                    basisPath,
+                    dracoPath,
+                    glslangPath,
+                    twgslPath,
+                    pcx,
+                    files,
+                });
+            } catch (e) {
+                console.error(e, "Stack:", e.stack);
+                window.top.dispatchEvent(new CustomEvent('exampleError', {detail: e}));
+                allowRestart = true;
+                return;
+            }
             ready = true;
             class ExampleLoadEvent extends CustomEvent {
                 constructor(deviceType) {
--- a/examples/scripts/generate-standalone-files.mjs
+++ b/examples/scripts/generate-standalone-files.mjs
@@ -255,11 +255,32 @@ ${exampleClass.example.toString()}
             data = new observer.Observer({});
             main(files);
         }
+        /**
+         * Native error handling triggered by an error. We send them to UI so we can
+         * display them in Monaco editor or elsewhere.
+         */
+        function nativeError(event) {
+            const {lineno, colno, filename, error, message} = event;
+            const stack = error.stack || 'missing stack';
+            console.error(
+                'Error:\\n', error, '\\n',
+                'Stack:\\n', stack, '\\n',
+                {lineno, colno, filename, error, message}
+            );
+            if (event.error.name === 'SyntaxError') {
+                // Monaco displays syntax errors already
+                return;
+            }
+            const detail = {lineno, colno, filename, error, message};
+            const responseEvent = new CustomEvent("exampleError", {detail});
+            window.top.dispatchEvent(responseEvent);
+        }
         window.addEventListener('requestFiles', requestFiles);
         window.addEventListener('showStats'   , showStats   );
         window.addEventListener('hideStats'   , hideStats   );
         window.addEventListener('destroy'     , destroy     );
         window.addEventListener('hotReload'   , hotReload   );
+        window.addEventListener('error'       , nativeError );
         function updateControls() {
             const event = new CustomEvent("updateFiles", {
                 detail: {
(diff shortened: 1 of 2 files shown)


mvaligursky (approved):
Very cool.

examples/src/app/code-editor.mjs
```diff
@@ -59,27 +65,76 @@ class CodeEditor extends TypedComponent {
...
+        const error = event.detail;
+        if (!error.message) {
+            return;
+        }
+        const message = `**Error**\n${error.message}\n\n**Stack**\n ${error.stack}`;
+        const lines = error.stack.split('\n');
+        const firstLineWithAt = lines.find(line => / *at/.test(line));
+        // -2 to map to correct line in editor
+        const line = firstLineWithAt.split(':').at(-2) - 2;
+        const lineText = editor.getModel().getLineContent(line);
```
marklundin, line 118: Nice work @kungfooman 🙏 I may be wrong, but is it better to use `setModelMarkers` here?  https://microsoft.github.io/monaco-editor/playground.html?source=v0.44.0#example-extending-language-services-model-markers-example. Not 100% clear on the difference but seems it's more generally used for errors

kungfooman: Interesting, thank you @marklundin... that API even looks a bit simpler, just `monaco.editor.removeAllMarkers()` instead of passing around "delta arrays".

kungfooman:
I'm testing on Chrome Mobile and I can't find a way to read the actual message for the red squiggles:

image

This is code to test suggestion by @marklundin in any Examples instance via F12/DevTools:

```js
editor.getModel().setValue(`12345
abcd
234.56
12345
abcd
234.5612345
abcd
234.56
12345
abcd
234.56`);
function validate(model) {
	const markers = [];
	// lines start at 1
	for (let i = 1; i < model.getLineCount() + 1; i++) {
		const range = {
			startLineNumber: i,
			startColumn: 1,
			endLineNumber: i,
			endColumn: model.getLineLength(i) + 1,
		};
		const content = model.getValueInRange(range).trim();
		const number = Number(content);
		if (Number.isNaN(number)) {
			markers.push({
				message: "not a number",
				severity: monaco.MarkerSeverity.Error,
				startLineNumber: range.startLineNumber,
				startColumn: range.startColumn,
				endLineNumber: range.endLineNumber,
				endColumn: range.endColumn,
			});
		} else if (!Number.isInteger(number)) {
			markers.push({
				message: "not an integer",
				severity: monaco.MarkerSeverity.Warning,
				startLineNumber: range.startLineNumber,
				startColumn: range.startColumn,
				endLineNumber: range.endLineNumber,
				endColumn: range.endColumn,
			});
		}
	}
	monaco.editor.setModelMarkers(model, "owner", markers);
}
validate(editor.getModel());
```

Same problem, I can't find a way to read the actual errors/warnings:

image

By chance I found this mobile friendly version: https://github.com/Splizard/vscode-mobile ... @Splizard does this happen to work in your version?

Splizard:
@kungfooman
I've got a stylus (S pen) on my mobile, which will trigger these sorts of popup on hover, documentation, errors etc. No solution for exclusively touch yet. 

(We've only just started the project with a couple of PRs, you are welcome to join the discussions area of the project). 

The primary point of contention for better touch interaction is scrolling, versus selection, versus viewing this sort of contextual information. 

Personally, I think a medium-long press could be suitable to show this information?

kpal81xd (commented):
Hi just for the new changes make sure that single values in brackets and curly braces need to have spaces too e.g. `{details}` and `[newDecorators]`. The rest looks sound tho!

kungfooman:
Thank you for the review @kpal81xd! I will fix the linting and I also have to change what @marklundin suggested, I just don't have the time right now :zany_face:

examples/scripts/generate-standalone-files.mjs
```diff
@@ -255,11 +255,32 @@ ${exampleClass.example.toString()}
             data = new observer.Observer({});
             main(files);
         }
+        /**
+         * Native error handling triggered by an error. We send them to UI so we can
+         * display them in Monaco editor or elsewhere.
+         */
+        function nativeError(event) {
+            const {lineno, colno, filename, error, message} = event;
```
kpal81xd, line 263: Curly braces should have spaces after the leading `{` and before the trailing `}` if containing more than one element.

kungfooman:
> Hi just for the new changes make sure that single values in brackets and curly braces need to have spaces too e.g. `{details}` and `[newDecorators]`. The rest looks sound tho!

BTW only curly braces are the only exceptions here, we don't waste horizontal space in other locations:

https://github.com/playcanvas/engine/blob/2345be2217680280b0f450399add206a35b199cd/src/framework/parsers/glb-parser.js#L917-L918

kpal81xd:
> Hey @Splizard, thank you for sharing this information! I had the chance to try it out with a Surface Pro 9 pen and it also works in Monaco (although I must say using the scroll bar is not ideal 😅).
> 
> > Personally, I think a medium-long press could be suitable to show this information?
> 
> Yep, I wonder how hard that would even be, it may just require some touch events that simulate a mouse hover event via `window.dispatchEvent(...)`.
> 
> I also wonder about a good UI location to place errors that can't be located in the Examples editor, maybe like an "error tab" which shows error message, filename, stack trace etc. 🤔

I think this error tab is a good idea and seems much more integrated than just solely having errors on highlight esp for mobile in particular

kungfooman:
> I think this error tab is a good idea and seems much more integrated than just solely having errors on highlight esp for mobile in particular

Thank you for the feedback @kpal81xd!

Right, it was kind of disappointing that Monaco doesn't support touch for reading such error messages, sometimes it could be the only clue (I love Android debugging via Chrome Remote Debugging Protocol, but I always had a bad experience with e.g. Apple devices).

Maybe it could look like VSCode is displaying errors?

image

If it's possible, could you take over this PR?

examples/scripts/generate-standalone-files.mjs
```diff
@@ -294,23 +331,44 @@ ${exampleClass.example.toString()}
...
+                    data,
+                    assetPath,
+                    scriptsPath,
+                    ammoPath,
+                    basisPath,
+                    dracoPath,
+                    glslangPath,
+                    twgslPath,
+                    pcx,
+                    files,
```
willeastcott, line 353:
```suggestion
                    files
```

examples/src/app/code-editor.mjs
```diff
@@ -59,27 +63,74 @@ class CodeEditor extends TypedComponent {
...
 
+    /**
+     * @param {CustomEvent} event - Custom event containing error thrown in iframe.
+     */
+    handleExampleError(event) {
+        const { editor } = this;
+        if (!editor) {
+            return;
+        }
+        // error can be an actual error or e.g. only a string in case of throw "oh ohhh"
```
willeastcott, line 106: "oh ohhh"? 🤔

willeastcott (approved):

kpal81xd:
@willeastcott this PR was going to be closed in place of this GH-6260 😅

kpal81xd:
This PR is closed in favour of GH-6260

PR #5574, opened 2023-08-17
Change ResourceHandler from interface to base class

willeastcott:
This PR is a reasonably simple refactor of the `ResourceHandler`s. Specifically, it switches the effectively unused 'interface' `ResourceHandler` (despite JavaScript not supporting interfaces) into a base class.

I confirm I have read the [contributing guidelines](https://github.com/playcanvas/engine/blob/master/.github/CONTRIBUTING.md) and signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #5868, opened 2023-11-29
Add JSDoc for extras/splat ply-reader/splat-material/splat

kungfooman:
Also adding `jsconfig.js` and fixing a small issue with `find` for `UInt8Array`'s (it would skip a good/last result) because it used `<` and not `<=`.

Motivation is to have a better IDE experience (static type-checking and Intellisense) and also to have more types in playcanvas-extras that we can validate via RTI (dynamic type-checking for NaN's for example).

### DIFF
--- a/extras/splat/ply-reader.js
+++ b/extras/splat/ply-reader.js
@@ -13,27 +13,30 @@ const dataTypeMap = new Map([
 ]);
 
 /**
+ * @typedef {Int8Array|Uint8Array|Int16Array|Uint16Array|Int32Array|Uint32Array|Float32Array|Float64Array} DataType
+ */
+
+/**
  * @typedef {object} PlyProperty
- * @property {string} type
- * @property {string} name
- * @property {Float32Array} storage
- * @property {number} byteSize
+ * @property {string} type - E.g. 'float'.
+ * @property {string} name - E.g. 'x', 'y', 'z', 'f_dc_0' etc.
+ * @property {DataType} storage - Data type, e.g. instance of Float32Array.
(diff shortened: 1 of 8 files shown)


kungfooman:
I think I found another bug:

image

The property `app.loader.maxRetries` doesn't exist, but they do exist on each `containerHandler`, can I just fix that here, like `containerHandler.maxRetries`?

mvaligursky:
This might be related
https://github.com/playcanvas/engine/pull/5574

extras/splat/ply-reader.js
```diff
@@ -12,25 +12,32 @@ const dataTypeMap = new Map([
...
-    properties;
-}
-
-// asynchronously read a ply file data
+/**
+ * @typedef {Int8Array|Uint8Array|Int16Array|Uint16Array|Int32Array|Uint32Array|Float32Array|Float64Array} DataType
+ */
+
+/**
+ * @typedef {object} PlyProperty
```
mvaligursky, line 20: do we even use these types anywhere?

kungfooman: yep, `PlyElement` is used once in `ply-reader.js` and three times in `splat-data.js`:

```js
class SplatData {
    /** @type {import('./ply-reader').PlyElement[]} */
    elements;

    /** @type {import('./ply-reader').PlyElement} */
    vertexElement;

    /**
     * @param {import('./ply-reader').PlyElement[]} elements - The elements.
     * @param {boolean} [performZScale] - Whether to perform z scaling.
     */
    constructor(elements, performZScale = true) {
```

slimbuck (approved):
Yep, looks great thanks!

PR #5873, opened 2023-12-01
[Examples] Fix build time getting progressively slower

### LINKED ISSUE #5872: [Engine Examples] - local build time gets progressively slower
I locally build examples using `npm run develop`, and then modify the source code of both the examples and the engine. I've noticed that the build time gets longer each build, forcing me to restart the command every 20 or so builds.

See how this progressed from 0.8s to 2.2s over about 15 builds:

Screenshot 2023-12-01 at 12 00 48

Note that I also run `target=debug_es5 npm run watch` in the engine to have up to date engine build, which might but might not be related.

pinging @kungfooman in case he has some ideas here.

kungfooman:
Fixes #5872

Before we had watchers in `examples/rollup.config.js` that would simply copy build outputs over into `/dist/`:

```js
    { src: '../build/playcanvas.d.ts', dest: 'dist/playcanvas.d.ts' },
    { src: '../build/playcanvas.js', dest: 'dist/iframe/playcanvas.js' },
    { src: '../build/playcanvas.dbg.js', dest: 'dist/iframe/playcanvas.dbg.js' },
    { src: '../build/playcanvas.prf.js', dest: 'dist/iframe/playcanvas.prf.js' },
    { src: '../build/playcanvas-extras.js', dest: 'dist/iframe/playcanvas-extras.js' },
```

This worked nice with running `npm run develop` and if you spawn another shell with `target=debug_es5 npm run watch` it keeps working for a while aswell, but over time it does more and more, the two rollup processes get in some kind of "create / watch - do stuff / create do stuff again" loop or something.

This can be fixed when we stop putting watchers on build outputs and instead use the rollup target configuration functions (e.g. `function buildTarget(buildType, moduleFormat)`) directly in `examples/rollup.config.js` for building the mentioned build outputs.

This is not complicated by itself, but what makes the PR a bit bigger is the separation of a few functions that we need to call from the Examples rollup script.


mvaligursky:
Nice one @kungfooman , much appreciated. The separation into modules is great!
I'll merge this now and give it a good test over the next few days as I work on some new engine features.

mvaligursky:
So far this has worked great, no issues .. and maybe even fixed another issue, where sometimes changes from extras were ignored and I have to manually build extras to get those in again.

PR #5877, opened 2023-12-04
[Fix] Fix compatibility with deprecated Layer.renderTarget

mvaligursky:
Fix the camera where a Layer specifies renderTarget and the camera matrices were not updated for it (aspect ratio). Issue introduced in refactoring in #5738

### DIFF
--- a/src/scene/renderer/renderer.js
+++ b/src/scene/renderer/renderer.js
@@ -1077,9 +1077,8 @@ class Renderer {
         for (let i = 0; i < numCameras; i++) {
             const camera = comp.cameras[i];
 
-            // update camera and frustum
-            camera.frameUpdate(camera.renderTarget);
-            this.updateCameraFrustum(camera.camera);
+            let currentRenderTarget;
+            let cameraChanged = true;
             this._camerasRendered++;
 
             // for all of its enabled layers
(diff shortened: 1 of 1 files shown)


src/scene/renderer/renderer.js
```diff
@@ -1088,6 +1087,17 @@ class Renderer {
                 const layer = comp.getLayerById(layerIds[j]);
                 if (layer && layer.enabled) {
 
+                    // update camera and frustum when the render target changes
+                    // TODO: This is done here to handle the backwards compatibility with the deprecated Layer.renderTarget,
+                    // when this is no longer needed, this code can be moved up to execute once per camera.
+                    const renderTarget = camera.renderTarget ?? layer.renderTarget;
```
slimbuck, line 1093: could renderTarget be null at this point (and currentRenderTarget undefined)?

mvaligursky: Nope, on the camera or Layer you either specify the render target, or it's undefined. There is no distinction between undefined and null here.

This matches handling here:
https://github.com/playcanvas/engine/blob/e0360352900618c608f4f11de2e4f7b6cb5919b3/src/scene/composition/layer-composition.js#L281-L287

slimbuck: Should you be checking for renderTarget !== undefined here since on first loop cameraChange is true?

mvaligursky: the camera needs to update at least once, even for undefined RT.
Before it updated before the loop, now it's done inside the loop.

PR #5880, opened 2023-12-05
Fix docs 404 for XrAnchorCreateCallback

kungfooman:
Little oversight of @Maksims, currently we have a 404 in the docs:

https://developer.playcanvas.com/en/api/pc.XrAnchors.html#create

What makes it a bit worse ... the HTTP redirection goes into an endless loop:

https://developer.playcanvas.com/en/en/en/en/en/en/en/en/en/en/en/en/en/en/en/en/en/en/en/en/api/pc.XrAnchorCreate.html

The reason is pretty much that internally a callback is defined as a typedef, but instead of having properties, callbacks have params. So in order to differentiate between the two, the `name.endsWith('Callback')` logic was introduced. We could change that logic of course and check for "does the typedef has params?", but I personally like the Callback ending (since the ending makes it immediately clear what it is).

### DIFF
--- a/src/framework/xr/xr-anchors.js
+++ b/src/framework/xr/xr-anchors.js
@@ -5,7 +5,7 @@ import { XrAnchor } from './xr-anchor.js';
 /**
  * Callback used by {@link XrAnchors#create}.
  *
- * @callback XrAnchorCreate
+ * @callback XrAnchorCreateCallback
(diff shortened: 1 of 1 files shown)


mvaligursky (approved):
Nice one. All good @Maksims before I merge this?

Maksims:
LGTM!

mvaligursky:
Thanks @kungfooman !

PR #5881, opened 2023-12-05
Introduction of RenderPassCamera - pass describing typical render passes for the camera

mvaligursky:
- new class RenderPassCamera, which internally uses other passes passes to render a typical rendering by the camera, including new post-processing. The only unsupported passes for now are depth grab passes (separate PR).

This allows new post-processing to be used by simply creating this pass and specifying it on a camera:
```
// Use a render pass camera, which is a render pass that implements typical rendering of a camera.
// Internally this sets up additional passes it needs, based on the options passed to it.
const renderPassCamera = new pcx.RenderPassCamera(app, {
    camera: cameraEntity.camera,    // camera used to render those passes
    samples: 2,                     // number of samples for multi-sampling
    sceneColorMap: true             // true if the scene color should be captured
});

// and set up these rendering passes to be used by the camera, instead of its default rendering
cameraEntity.camera.renderPasses = [renderPassCamera];
```
it exposes properties / internal passes to allow their control, for example:
```
renderPassCamera.bloomEnabled = true;

renderPassCamera.composePass.gradingEnabled = true;
renderPassCamera.composePass.gradingSaturation = 2;
```

### DIFF
--- a/extras/render-passes/render-pass-camera.js
+++ b/extras/render-passes/render-pass-camera.js
@@ -1,0 +1,181 @@
+import {
+    LAYERID_SKYBOX,
+    LAYERID_IMMEDIATE,
+    PIXELFORMAT_RGBA8,
+    ADDRESS_CLAMP_TO_EDGE,
+    FILTER_LINEAR,
+    RenderPass,
+    RenderPassColorGrab,
+    RenderPassRenderActions,
+    RenderTarget,
+    Texture
+} from "playcanvas";
+import { RenderPassBloom } from "./render-pass-bloom.js";
+import { RenderPassCompose } from "./render-pass-compose.js";
+
+class RenderPassCamera extends RenderPass {
+    app;
+
+    scenePass;
+
+    composePass;
+
+    bloomPass;
+
+    _bloomEnabled = true;
+
+    _renderTargetScale = 1;
+
+    constructor(app, options = {}) {
+        super(app.graphicsDevice);
+        this.app = app;
+        this.options = this.sanitizeOptions(options);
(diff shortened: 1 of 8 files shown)


extras/render-passes/render-pass-camera.js
```diff
@@ -0,0 +1,181 @@
...
+    RenderPass,
+    RenderPassColorGrab,
+    RenderPassRenderActions,
+    RenderTarget,
+    Texture
+} from "playcanvas";
+import { RenderPassBloom } from "./render-pass-bloom.js";
+import { RenderPassCompose } from "./render-pass-compose.js";
+
+class RenderPassCamera extends RenderPass {
```
slimbuck, line 16: This name makes it sound like a type of camera!

If this this class implements a render pass with default post processing, shouldn't the name describe that instead?

mvaligursky: maybe, suggest a name?
I called it RenderPassCamera, as it represents a default render pass setup for a camera

slimbuck: The correct name (if there is such a thing) would probably be `DefaultPostProcessRenderPass`. Quite a mouth (eye) full.

mvaligursky: but it does not just do post-processing, it renders the scene, grabs color, applies post effects .. in the future I'll add depth pre-pass, SSAO, SSR and similar. Maybe I'll call it `RenderPassFrame`

LeXXik:
Worth changing the title?

PR #5882, opened 2023-12-06
Add JSDoc typeof support via JSDoc plugin

kungfooman:
We had many issues around `typeof` and at some point simply settled on... the wrong type.

This caused a trail of `ts-ignore` comments, but I never realized that JSDoc has a plugin for exactly this issue: jsdoc-typeof-plugin

So this PR simply integrates that plugin and since now we have a correct type, we can let go of maaany `ts-ignore` comments.

### DIFF
--- a/conf-api.json
+++ b/conf-api.json
@@ -1,7 +1,8 @@
 {
     "plugins": [
         "plugins/markdown",
-        "./node_modules/jsdoc-tsimport-plugin/index.js"
+        "./node_modules/jsdoc-tsimport-plugin/index.js",
+        "./node_modules/jsdoc-typeof-plugin"
     ],
     "recurseDepth": 10,
     "source": {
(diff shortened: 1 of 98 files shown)


mvaligursky:
Yes, fantastic find!
I think there are few more as well in those app create options .. for example lightmapper and batcher should be typeof, as we pass in type, and not the instance.

kungfooman:
> I think there are few more as well in those app create options .. for example lightmapper and batcher should be typeof, as we pass in type, and not the instance.

Yep, absolutely, I thought we could chip away at it one piece at a time (to keep PR size non-overwhelming/reviewable). 

If we make some fire under https://github.com/playcanvas/engine/pull/5574 we can also remove lots of `ts-ignore`'s again (currently it's not possible, because it's considered `@interface` while `typeof` requires a "real" class).

PR #5887, opened 2023-12-08
WebXR HitTestSource fix

### LINKED ISSUE #5884: WebXR hit test bugs
After #5778 hit tests are being raised even if there aren't any results.

The previous code raised an event for every result in the array, while the new code doesn't check if any results were received. The first hit test is reported as (0, 0, 0).

I fixed this locally by adding the following test, but there may be a better way.
```javascript 
    updateHitResults(results, inputSource) {
        if (results.length === 0)
            return;
```

Also, if you initiate hit testing with `xr.hitTest.start` (and specify profile 'generic-touchscreen', if that matters), then `hitTestSourcesSet` is not populated with the hit test source and no hit tests are reported.

These two issues mean model-viewer XR mode no longer works.

cc @Maksims

Maksims:
PR #5778 introduced a bug where the absence of hit test results would still provide a callback with the null result at the position of the controller.

Fixes #5884


willeastcott (approved):

src/framework/xr/xr-hit-test-source.js
```diff
@@ -123,7 +127,11 @@ class XrHitTestSource extends EventHandler {
                 this.updateHitResults(transientResult.results, inputSource);
             }
         } else {
-            this.updateHitResults(frame.getHitTestResults(this._xrHitTestSource));
+            const results = frame.getHitTestResults(this._xrHitTestSource);
+            if (results.length)
+                return;
+
+            this.updateHitResults(results);
```
slimbuck, line 134:
```suggestion
            if (results.length) {
                this.updateHitResults(results);
            }

```

slimbuck (commented):
Thanks for this fix Maks!

Model-viewer is still not receiving hit results for touchscreen because of [this line](https://github.com/playcanvas/engine/blob/main/src/framework/xr/xr-hit-test-source.js#L136). Looks like the inputSource isn't being added to `hitTestSourcesSet` when using `xr.hitTest.start()`.

Maksims:
Thank you @slimbuck, I did a typo there. Should work well not: gaze based hit test works well, screen input sources too, and controller hit tests. Tested on Android phone and Quest 3.

slimbuck:
Thanks @Maksims , but how about?:

> Model-viewer is still not receiving hit results for touchscreen because of [this line](https://github.com/playcanvas/engine/blob/main/src/framework/xr/xr-hit-test-source.js#L136). Looks like the inputSource isn't being added to hitTestSourcesSet when using xr.hitTest.start().

Maksims:
> Thanks @Maksims , but how about?:
> 
> > Model-viewer is still not receiving hit results for touchscreen because of [this line](https://github.com/playcanvas/engine/blob/main/src/framework/xr/xr-hit-test-source.js#L136). Looks like the inputSource isn't being added to hitTestSourcesSet when using xr.hitTest.start().

I have tested Android and Quest 3 from examples, they work as intended. Will re-check again then.

slimbuck:
Thanks Maks. I just tested with your latest changes and the following callback is still never invoked:

```javascript
// register for touchscreen hit test
xr.hitTest.start({
    profile: 'generic-touchscreen',
    entityTypes: [XRTRACKABLE_POINT, XRTRACKABLE_PLANE, XRTRACKABLE_MESH],
    callback: (err: Error | null, hitTestSource: XrHitTestSource) => {
        ...
    }
});
```

Maksims:
Ok, now it should work well with profiles provided.
But bear in mind: when using profile that will cover multiple input sources (multiple controllers of the same profile (left/right), multi-touch, etc), it will trigger `result` for each of those input sources.
Best would be to listen for added/removed input sources, and while there is added input source of required profile, start hit test for it, and ignore any other input sources until that active one is removed. That way in multi-touch scenario any concurrent taps will not lead to overtaking hit test results.

slimbuck:
Hi @Maksims,

I can confirm this works with model-viewer - thanks!

I am just wondering about your last comment. If I start hitTest using profile `generic-touchscreen`, won't I only get callbacks from touchscreen devices? (And devices presumably only have a single touchscreen so this isn't an issue?).

Maksims:
> I am just wondering about your last comment. If I start hitTest using profile `generic-touchscreen`, won't I only get callbacks from touchscreen devices? (And devices presumably only have a single touchscreen so this isn't an issue?).

Indeed, it will only cover touchscreen devices.
Just for a model-viewer, the problem is that:
1. Single tap screen
2. Get a callback
3. Without releasing start moving tap around
4. Then add another tap
5. You now have multi-touch scenario, and result callbacks will be fired for both.

So if you don't use `XrInputSource.id` to know which tap started, then moved and finished. Then you will start using hit test result from another tap.
That is why it is recommended to use explicit hit test on the input source, that way you know when to start listening for results, for which particular input source, and so on.

This is not related to this PR.

slimbuck:
BTW is it possible to perform a hitTest with a screen-space pixel offset? Currently we perform hitTest directly under the finger always, but we'd like to perform a hitTest at, say, touch-position + vec2(40, 20) pixels. Is this currently possible?

Maksims:
> BTW is it possible to perform a hitTest with a screen-space pixel offset? Currently we perform hitTest directly under the finger always, but we'd like to perform a hitTest at, say, touch-position + vec2(40, 20) pixels. Is this currently possible?

There is a way to provide XRRay for ray offset, but it has not been exposed through integration. It would be a separate PR.

slimbuck:
Yep of course, created https://github.com/playcanvas/engine/issues/5927 to track.

willeastcott:
@slimbuck Is this mergeable now?

slimbuck (approved):

slimbuck:
Yes! Approved

PR #5893, opened 2023-12-12
fix opacity conversion when 0.0, remove extraneous clamp

pjbaron:
The opacity value ranges from NEGATIVE_INFINITY to POSITIVE_INFINITY and at opacity === 0.0 the sigmoid will return 0.5.
The original code would convert this to an alpha of 255 instead of 128.
It is possible that the condition is intended to catch an undefined opacity, so I have made that explicit. If this is not the case, then the ternary condition can be removed entirely. However if there is a possibility of opacity === null, that should be handled as well.
The sigmoid function cannot return < 0 or > 1 so the clamp is not needed.

Fixes #
I don't think there's bug report, but I've been having problems with opacity and went into the engine to find out why.

### DIFF
--- a/extras/splat/splat.js
+++ b/extras/splat/splat.js
@@ -198,8 +198,9 @@ class Splat {
                 data[i * 4 + 2] = math.clamp((0.5 + SH_C0 * c2[i]) * 255, 0, 255);
             }
 
-            // opacity
-            data[i * 4 + 3] = opacity ? math.clamp(sigmoid(opacity[i]) * 255, 0, 255) : 255;
+            
+            // opacity (-inf..+inf, at 0.0 it returns 0.5)
+            data[i * 4 + 3] = (opacity !== undefined) ? sigmoid(opacity[i]) * 255 : 255;
         }
 
         texture.unlock();


pjbaron:
sorry about the lint fail, it appears my editor inserted spaces on an empty line

extras/splat/splat.js
```diff
@@ -198,8 +198,9 @@ class Splat {
                 data[i * 4 + 2] = math.clamp((0.5 + SH_C0 * c2[i]) * 255, 0, 255);
             }
 
-            // opacity
-            data[i * 4 + 3] = opacity ? math.clamp(sigmoid(opacity[i]) * 255, 0, 255) : 255;
+            
+            // opacity (-inf..+inf, at 0.0 it returns 0.5)
+            data[i * 4 + 3] = (opacity !== undefined) ? sigmoid(opacity[i]) * 255 : 255;
```
willeastcott, line 203: But this previous code was just checking that the `Float32Array` `opacity` array was defined. Not a _value_ from that array. If anything is wrong in this function, it's the JSDocs. The `opacity` param should be marked as optional presumably?

slimbuck:
Hi @pjbaron,

What problems have you been having with opacity?

Can you provide us a repo we can test with?

Thanks!

extras/splat/splat.js
kungfooman: I agree, based on `if (c0 && c1 && c2) {` they should all be optional.

Code is called from here:

https://github.com/playcanvas/engine/blob/f9bac39b27f9df00d1ca6285a3c1c36b5af77fd4/extras/splat/splat-container-resource.js#L50

And I found a PLY that has other prop names aswell (using the prop names 'red', 'green' and 'blue')

pjbaron: @willeastcott oh damn, my apologies! The hazards of making a public pull request after a long day... Yes you're correct of course, the opacity condition is an array existence check.
The clamp is definitely not needed though.

pjbaron:
@slimbuck it was the separate clones of 'center' which were making my morph targets appear transparent, not opacity as I had previously thought. I commented with more details about that on a separate issue about optimisation/redesign of the splatting system.

extras/splat/splat.js
willeastcott: @pjbaron No worries! 😄 Do you wanna update the code to reflect this?

mvaligursky:
Closing this as the solution does not seem needed. Please update the description / change the PR based on the feedback if this is not correct.

PR #5899, opened 2023-12-14
Disable MSAA for splat examples

mvaligursky:
- it almost doubles the GPU cost to have it enabled, for no visibie benefit as splats are soft-edged already


willeastcott (approved):

willeastcott:
But the 'splat many' example has regular geometry too, right?

mvaligursky:
Sure does. But the perf gain is significant enough to ignore some uglier edges.

PR #5901, opened 2023-12-18
Exposing sky for Editor 

kpal81xd:
Implemenation of the engine side sky properties outlined in [playcanvas/editor#1080](https://github.com/playcanvas/editor/issues/1080)

### DIFF
--- a/src/scene/scene.js
+++ b/src/scene/scene.js
@@ -219,12 +219,12 @@ class Scene extends EventHandler {
         this._skyboxLuminance = 0;
         this._skyboxMip = 0;
 
-        this._skyboxScale = new Vec3(1, 1, 1);
         this._skyboxPosition = new Vec3();
         this._skyboxRotationShaderInclude = false;
         this._skyboxRotation = new Quat();
         this._skyboxRotationMat3 = new Mat3();
         this._skyboxRotationMat4 = new Mat4();
+        this._skyboxScale = new Vec3(1, 1, 1);
 
         this._skyboxTripod = new Vec3();
 
@@ -618,21 +618,6 @@ class Scene extends EventHandler {
     }
 
     /**
-     * The scale of the skybox to be displayed. Defaults to {@link Vec3.ONE}.
-     *
-     * @type {Vec3}
-     */
-    set skyboxScale(value) {
-        if (!this._skyboxScale.equals(value)) {
-            this._skyboxScale.copy(value);
-        }
-    }
-
-    get skyboxScale() {
-        return this._skyboxScale;
-    }
-
-    /**
      * The position of the skybox to be displayed. Defaults to {@link Vec3.ZERO}.
      *
      * @type {Vec3}
@@ -675,6 +660,21 @@ class Scene extends EventHandler {
 
     get skyboxRotation() {
         return this._skyboxRotation;
+    }
+
+    /**
+     * The scale of the skybox to be displayed. Defaults to {@link Vec3.ONE}.
+     *
+     * @type {Vec3}
+     */
+    set skyboxScale(value) {
+        if (!this._skyboxScale.equals(value)) {
+            this._skyboxScale.copy(value);
+        }
+    }
+
+    get skyboxScale() {
+        return this._skyboxScale;
     }
 
     /**
--- a/src/scene/scene.js
+++ b/src/scene/scene.js
@@ -10,7 +10,7 @@ import { Mat4 } from '../core/math/mat4.js';
 import { GraphicsDeviceAccess } from '../platform/graphics/graphics-device-access.js';
 import { PIXELFORMAT_RGBA8, ADDRESS_CLAMP_TO_EDGE, FILTER_LINEAR } from '../platform/graphics/constants.js';
 
-import { BAKE_COLORDIR, FOG_NONE, GAMMA_SRGB, LAYERID_IMMEDIATE } from './constants.js';
+import { BAKE_COLORDIR, FOG_NONE, GAMMA_SRGB, LAYERID_IMMEDIATE, SKYTYPE_INFINITE } from './constants.js';
 import { LightingParams } from './lighting/lighting-params.js';
 import { Sky } from './skybox/sky.js';
 import { Immediate } from './immediate/immediate.js';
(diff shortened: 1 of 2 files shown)


src/scene/scene.js
```diff
@@ -551,6 +555,19 @@ class Scene extends EventHandler {
         return this._skyboxCubeMap;
     }
 
+    /**
+     * Type of sky. Defaults to {@link SKYTYPE_INFINITE}.
+     *
+     * @type {number}
+     */
+    set skyType(value) {
```
mvaligursky, line 563: we do not need to expose these new properties, those are already exposed as scene.sky.***
for example `skyType` is `scene.sky.type`

mvaligursky: and so the code that needs to access them just need to access them on `.sky`

src/scene/scene.js
```diff
@@ -694,14 +756,27 @@ class Scene extends EventHandler {
         this.lightmapMaxResolution = render.lightmapMaxResolution;
         this.lightmapMode = render.lightmapMode;
         this.exposure = render.exposure;
+        this.skyType = render.skyType ?? SKYTYPE_INFINITE;
         this._skyboxIntensity = render.skyboxIntensity ?? 1;
         this._skyboxLuminance = render.skyboxLuminance ?? 20000;
         this._skyboxMip = render.skyboxMip ?? 0;
 
+        if (render.skyboxScale) {
```
mvaligursky, line 764: here the sky properties should be handled similarly to lighting properties.
So based on 
`this.lighting.applySettings(render);` it should be
`this.sky..applySettings(render);`

src/scene/scene.js
kpal81xd: Ahh okay I wasnt sure if I should have forwarded the get call to this scene class but I will remove that

src/scene/scene.js
kpal81xd: Ahh thanks yea I see. Should i move the existing properties over too like rotation and intensity?

mvaligursky: Not in this PR. It would change the existing public API, and so needs different handling.

kpal81xd: Alright thanks

src/scene/skybox/sky.js
```diff
@@ -59,6 +75,19 @@ class Sky {
         this.projectedSkydomeCenterId = this.device.scope.resolve('projectedSkydomeCenter');
     }
 
+
+    applySettings(render) {
+        if (render.skyPosition) {
```
mvaligursky, line 80: scale and position should be set on the this.node instead of new properties, the same way the example does it:
```
this.node.setLocalScale(new pc.Vec3(data.get('data.skybox.scale') ?? [1, 1, 1]));
this.node.setLocalPosition(new pc.Vec3(data.get('data.skybox.position') ?? [0, 0, 0]));
```

and the added `position` and `scale` properties are not needed.

src/scene/skybox/sky.js
```diff
@@ -59,6 +75,19 @@ class Sky {
...
+            this.position = new Vec3(render.skyPosition[0], render.skyPosition[1], render.skyPosition[2]);
+        }
+        if (render.skyScale) {
+            this.scale = new Vec3(render.skyScale[0], render.skyScale[1], render.skyScale[2]);
+        }
+        if (render.skyTripod) {
```
mvaligursky, line 86: I think lets use name `skyCenter` instead of `skyTripod` to match the naming inside the engine

src/scene/skybox/sky.js
willeastcott: Is it too late to reconsider exposing the `node` property? Maybe `scale`, `position` and `rotation` properties would be kinda nice. Then we could just alias `Scene#skyboxRotation` to `Sky#rotation`.

kpal81xd: Okay yea I was wondering if I was going to set them or leave that bit to you

mvaligursky: we need both rotations though. The node positions / rotates / scales the mesh itself.
but scene.skyboxRotation rotates the texture.

willeastcott: I don't understand why the texture can't also be rotated with a quaternion.

src/scene/skybox/sky.js
```diff
@@ -59,6 +59,15 @@ class Sky {
         this.projectedSkydomeCenterId = this.device.scope.resolve('projectedSkydomeCenter');
     }
 
+
+    applySettings(render) {
+        this.node.setLocalPosition(new Vec3(render.skyPosition ?? [0, 0, 0]));
+        this.node.setLocalScale(new Vec3(render.skyScale ?? [1, 1, 1]));
```
mvaligursky, line 66: and for completeness, expose 'rotation' as well please - this is handy for box projection if the room geometry is not aligned with the axes.
maybe those 3 that we set on the node should be `skyMeshPosition` `skyMeshRotation` and `skyMeshScale`, to not confuse them with the existing `skyboxRotation`, that rotates the texture

src/scene/skybox/sky.js
kpal81xd: Maybe alias `Scene#skyboxRotation` to `Scene#skyTextureRotation` for more clarity then?

mvaligursky: maybe in a separate PR. I'd avoid doing it in this PR, this one should focus on the task that is needed.

src/scene/skybox/sky.js
kpal81xd: I suggested in the other comment to do `skyTextureRotation` for the existing rotation prop but since you said it would change the existing API this seems more logical

mvaligursky (approved):
I'm happy with this. @willeastcott ?

willeastcott (approved):
It's fine. However, the two things that do still come to mind are:
1. Why can't rotation be shared between infinite and non-infinite skies?
2. Why does the engine have these `applySettings` functions? I feel like this should be Editor code. This is a general observation about loading scene settings though and not to be addressed in this PR.

mvaligursky:
> * Why does the engine have these `applySettings` functions? I feel like this should be Editor code. This is a general observation about loading scene settings though and not to be addressed in this PR.

I think Editor could set those directly, but the other use case is when the published project is loaded. Those settings are stored in data, and are applied to the scene. Similar to how components accept loaded data to initialize. So I'm not sure we can avoid this.

Issue #5159, opened 2023-03-16
Implement dither transparency as one solution for order independent transparency


mvaligursky:
Not a general solution, but cheap and useful.
https://danielilett.com/2020-04-19-tut5-5-urp-dither-transparency/

![image](https://user-images.githubusercontent.com/59932779/225690784-0cf3ea3b-b62e-496e-ac88-3cb2f0fbac47.png)

querielo:
I'm not sure that it has to be a part of the engine since it can be done without it. Also, it is not clear what kind of pattern should be used: Bayer, white noise, blue noise and so on. 

### Example.

**Source code**: https://github.com/querielo/playcanvas-dither-transparency

**Project**: https://playcanvas.com/project/1051789/overview/dither-transparency

LeXXik:
Is the project private?

querielo:
Ouch. I forgot about it :-)

Fixed.

https://playcanv.as/p/qdYtm4jw/
https://playcanvas.com/project/1051789/overview/dither-transparency

yaustar:
> I'm not sure that it has to be a part of the engine since it can be done without it. 

IIRC, the goal was to have a script that was officially supported by PlayCanvas rather than adding it to the engine directly. So it would live here: https://github.com/playcanvas/engine/tree/main/scripts

mvaligursky:
Script vs in engine is not something that was decided yet, it can go either way.

PR #5903, opened 2023-12-19
Opacity dithering support

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/5159

- Order independent transparency using dithering.
- uses procedural Bayer pattern with 64 dither levels (8x8 grid)
- dithering can be independently enabled for the shadows as well - this works the best with soft shadows such as VSM, otherwise a pattern is more obvious.

new API / example on how to enable:
```
StandardMaterial.opacityDither = true;
StandardMaterial.opacityShadowDither = true;

material.blendType would ideally be set to pc.BLEND_NONE
```

https://github.com/playcanvas/engine/assets/59932779/1c1da4bc-7c60-4d0c-90f4-8515d350d944

Limitations:
- if two objects are behind each other, with the same opacity, their pattern is the same and that causes the object behind to not be seen. Possible improvement here would be to have some per drew call random ID (meshInstance.id would work well) and the pattern would be offset along x and y based on this number. This is not done in this PR.
- Bayer pattern is used. Other great option without obvious pattern include a blue noise. This was considered, but would require a small blue noise texture to be available, and so this can be done in a separate PR if needed / alternatively the dither chunk can be overriden and implement this.


kpal81xd (approved):

src/scene/materials/standard-material.js
```diff
@@ -375,6 +375,10 @@ let _params = new Set();
  * @property {boolean} opacityFadesSpecular Used to specify whether specular and reflections are
  * faded out using {@link StandardMaterial#opacity}. Default is true. When set to false use
  * {@link Material#alphaFade} to fade out materials.
+ * @property {boolean} opacityDither Used to specify whether opacity is dithered, which allows
+ * transparency without alpha blending. Defaults is false.
+ * @property {boolean} opacityShadowDither Used to specify whether shadow opacity is dithered, which
+ * allows shadow transparency without alpha blending. Defaults is false.
```
willeastcott, line 381:
```suggestion
 * @property {boolean} opacityDither Used to specify whether opacity is dithered, which allows
 * transparency without alpha blending. Defaults to false.
 * @property {boolean} opacityShadowDither Used to specify whether shadow opacity is dithered, which
 * allows shadow transparency without alpha blending. Defaults to false.
```

willeastcott (approved):

Issue #3361, opened 2021-07-23
Implement progressive multi-frame rendering


mvaligursky:
The algorithm is described in GPU Pro 7, chapter 3.1
<img width="599" alt="Screenshot 2021-07-23 at 13 52 47" src="https://user-images.githubusercontent.com/59932779/126784207-d25b29e8-4946-4b3e-a36d-942e33472e39.png">

It allows for image quality to improve when the camera is static (and the scene is not animated) by accumulation of slightly shifted frames. This generates high quality antialiasing (TAA), and can also be used to generate soft shadows. All this without requiring (many) additional resources.

Additionally, we could possibly add support for progressive scaling here .. rendering the scene in lower resolution when the camera moves to keep the performance higher, and use full resolution when the camera stops.

mvaligursky:
additional paper
https://hpi.de/doellner/publications.html?tx_extbibsonomycsl_publicationlist%5Bcontroller%5D=Document&tx_extbibsonomycsl_publicationlist%5BfileName%5D=08-0022-limberger.pdf&tx_extbibsonomycsl_publicationlist%5BintraHash%5D=1f75864af25ed4bb218cc636619ffe45&tx_extbibsonomycsl_publicationlist%5BuserName%5D=jan.klimke&cHash=7046e2af99b3af0768dd42f80831fdb7

figo2264:
Just wonder when this feature is going to be added.

mvaligursky:
There are no immediate plans to work on it, unless somebody wants to help.

yaustar:
No short term plans from our team unfortunately. I've added the tag 'good first PR' so if anyone would like to help and contribute, they should be able to find this ticket more easily :)

figo2264:
I believe this is quite important issue for most of users who is using Playcanvas as industrial purpose.
Hope it will be added soon. 🙂


[cid:fbf25413-1d9c-4227-be80-153dd2211b40]
Technical Director

Sooyong Kim | Luis Kim


CP: 82-10-6763-5205
Office: 82-70-8622-5160
Fax: 82-70-8622-5161
E-mail: ***@***.***

Creative Design Production Studio | TEMEPST
<http://www.tmpst.co.kr/>http://www.tmpst.co.kr<http://www.tmpst.co.kr/>


________________________________

yaustar:
It would be great to have the work that was done for the Model Viewer rendering as a drop in Editor script

[...]

PR #5909, opened 2023-12-22
Initial experimental support for static TAA

mvaligursky:
this mostly implements: https://github.com/playcanvas/engine/issues/3361

### New API
- **CameraComponent.jitter** - a number representing the amount of jitter, defaults to0 (disabled)
- **GraphicsDevice.setBlendColor** - made the existing function public, as it is required by few blend modes.

### Other changes:
- support for TAA, but only for the static scene / static camera. Full TAA in later PRs.
- support for `setBlendColor` on WebGPU platform

https://github.com/playcanvas/engine/assets/59932779/dfd75dee-395f-412a-9e24-3abdc7782528

### DIFF
--- a/extras/index.js
+++ b/extras/index.js
@@ -16,3 +16,4 @@ export { RenderPassCompose } from './render-passes/render-pass-compose.js';
 export { RenderPassDownSample } from './render-passes/render-pass-downsample.js';
 export { RenderPassUpSample } from './render-passes/render-pass-upsample.js';
 export { RenderPassBloom } from './render-passes/render-pass-bloom.js';
+export { RenderPassTAA } from './render-passes/render-pass-taa.js';
--- a/src/scene/renderer/renderer.js
+++ b/src/scene/renderer/renderer.js
@@ -1,5 +1,6 @@
 import { Debug, DebugHelper } from '../../core/debug.js';
 import { now } from '../../core/time.js';
+import { Vec2 } from '../../core/math/vec2.js';
 import { Vec3 } from '../../core/math/vec3.js';
 import { Mat3 } from '../../core/math/mat3.js';
 import { Mat4 } from '../../core/math/mat4.js';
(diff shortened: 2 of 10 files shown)


extras/index.js
```diff
@@ -16,3 +16,4 @@ export { RenderPassCompose } from './render-passes/render-pass-compose.js';
 export { RenderPassDownSample } from './render-passes/render-pass-downsample.js';
 export { RenderPassUpSample } from './render-passes/render-pass-upsample.js';
```
willeastcott, line 17: Unrelated to this PR but I just noticed that `UpSample` should really be `Upsample`. See: https://en.wikipedia.org/wiki/Upsampling

src/scene/renderer/renderer.js
```diff
@@ -56,10 +57,32 @@ const _fixProjRangeMat = new Mat4().set([
...
+    new Vec2(0.0625, 0.888889),
+    new Vec2(0.5625, 0.037037),
+    new Vec2(0.3125, 0.370370),
+    new Vec2(0.8125, 0.703704),
+    new Vec2(0.1875, 0.148148),
+    new Vec2(0.6875, 0.481481),
+    new Vec2(0.4375, 0.814815),
+    new Vec2(0.9375, 0.259259),
+    new Vec2(0.03125, 0.592593)
+];
```
willeastcott, line 78: Why bother creating `Vec2` instances? Won't an array of arrays be sufficient here?

willeastcott (approved):

src/scene/renderer/renderer.js
mvaligursky: I don't know, I guess performance difference is minimal, it's a vector so I created a vector, but it could have been an array of just object too.

extras/index.js
mvaligursky: Noted, thanks, will rename those next year.

PR #5911, opened 2023-12-22
Fix to fringe offset to be independent of the render target resolution

mvaligursky:


extras/render-passes/render-pass-compose.js
```diff
@@ -264,7 +264,8 @@ class RenderPassCompose extends RenderPassShaderQuad {
         }
 
         if (this._fringingEnabled) {
-            this.fringingIntensityId.setValue(this.fringingIntensity / this.sceneTexture.height);
+            // relative fo a fixed texture resolution to preserve size regardless of the resolution
```
kpal81xd, line 267:
```suggestion
            // relative to a fixed texture resolution to preserve size regardless of the resolution
```

kpal81xd (approved):

PR #5912, opened 2023-12-26
On WebGPU flip the image for drawDepthTexture

erikdubbelboer:
Without this the depth buffer is rendered upside down. See WebGPU for https://playcanvas.github.io/#/graphics/ground-fog

This was the best way I found to fix it, making changes to drawTexture to fix it for depth textures makes other textures wrong instead that are already correct.


mvaligursky:
A good fix, but I'd prefer this instead, which matches the way the texture is sampled by the user in that example as well.

https://github.com/playcanvas/engine/pull/5930

PR #1374, opened 2018-09-24
Enable parallel shader compilation.

willeastcott:
Enable the use of the [KHR_parallel_shader_compile](https://www.khronos.org/registry/webgl/extensions/KHR_parallel_shader_compile/) extension to speed up shader compilation and hence app load times.

Note: at time of writing, KHR_parallel_shader_compile is only enabled in Chrome Canary AFAIK (and even then, you must enable WebGL draft extensions via chrome://flags).

I confirm I have signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #5914, opened 2023-12-26
Link WebGL shaders in parallel

erikdubbelboer:
In https://github.com/playcanvas/engine/pull/1374/files `KHR_parallel_shader_compile` was added but was never used. This extension is not needed for parallel compiling and linking.

We can link the shader in parallel right away without blocking. Only `getProgramParameter(prog, gl.LINK_STATUS)` will block until both compiling and linking are done.

Before this change every shader would be linked in serial as we would block right after linking until it is done before continuing with the next shader.

### DIFF
--- a/src/platform/graphics/webgl/webgl-graphics-device.js
+++ b/src/platform/graphics/webgl/webgl-graphics-device.js
@@ -1097,7 +1097,6 @@ class WebglGraphicsDevice extends GraphicsDevice {
         this.extCompressedTextureS3TC = this.getExtension('WEBGL_compressed_texture_s3tc', 'WEBKIT_WEBGL_compressed_texture_s3tc');
         this.extCompressedTextureATC = this.getExtension('WEBGL_compressed_texture_atc');
         this.extCompressedTextureASTC = this.getExtension('WEBGL_compressed_texture_astc');
-        this.extParallelShaderCompile = this.getExtension('KHR_parallel_shader_compile');
 
         // iOS exposes this for half precision render targets on both Webgl1 and 2 from iOS v 14.5beta
         this.extColorBufferHalfFloat = this.getExtension("EXT_color_buffer_half_float");
--- a/src/platform/graphics/webgl/webgl-shader.js
+++ b/src/platform/graphics/webgl/webgl-shader.js
@@ -58,9 +58,11 @@ class WebglShader {
     constructor(shader) {
         this.init();
 
-        // kick off vertex and fragment shader compilation, but not linking here, as that would
-        // make it blocking.
+        // kick off vertex and fragment shader compilation.
         this.compile(shader.device, shader);
+
+        // link the shader right away, this is not blocking
+        this.link(shader.device, shader);
 
         // add the shader to recently created list
         WebglShader.getBatchShaders(shader.device).push(shader);
@@ -305,6 +307,7 @@ class WebglShader {
             linkStartTime = now();
         });
 
+        // check the link status of a shader is a blocking operation
         const linkStatus = gl.getProgramParameter(glProgram, gl.LINK_STATUS);
         if (!linkStatus) {
 


kungfooman:
Can you test the change in the shader compilation speed test example?

Doing so:

1) `cd examples`
2) `npm install` if not done already
3) `npm run develop`
4) Go to http://127.0.0.1:5000/index.html#/graphics/shader-compile

I just made a crude testing counting seconds on my clock and it feels ~1-2s slower

DevTools log for each:

image

image

(there is some fluctuation, but that's a usual result I get)

slimbuck:
Noice!

@mvaligursky has actually done a lot of research in this area and ended up creating a ticket on chrome bugs because it doesn't parallelize correctly and it's very difficult to hide the stall on main thread (I don't have the ticket to hand).

Did you find this change made a measurable difference to compile speed and/or stall? Did you perhaps try on Safari and Chrome on Windows and Macos?

LeXXik:
Martin's research with links:
https://github.com/mvaligursky/webgl-parallel_shader_compile

erikdubbelboer:
@kungfooman which OS and which browser are you testing this on?

Here are my results for the ShaderCompiler example with my changes. I'm running on an Apple M2 Max.
```
Safari 17.2.1
before: 4.0s
after:  2.8s

Chrome 120.0.6099.109
before: 3.5s
after:  3.5s

Firefox 122.0b3
before: 2.8s
after:  2.8s
```
It seems indeed that Safari is the only browser to really properly parallelize the compiling and linking.

@LeXXik interesting research, but since we aren't using `KHR_parallel_shader_compile` I'm not sure it's relevant.

Even if the browser doesn't link the shaders in parallel it shouldn't make it slower I would assume. Calling `gl.linkProgram` all at once or one by one while waiting should then just produce the same result. This is what the results on my laptop seem to show as well.

kungfooman:
@erikdubbelboer Thank you for profiling! I'm on Linux Mint + Chrome 120.0.6099.71

Same system, but testing with Firefox 122.0a1 (2023-12-07) (64-bit):

Before: 5769.0ms
After: 5468.0ms (I tested multiple times and PR speedup is ~200-300ms)

LeXXik:
Adding Win11 Chrome 121.0.6167.16:

Before:
1679.6ms
1600.1ms
1571.6ms
1451.8ms
1369.7ms

After:
1210.7ms
1245.1ms
1203.0ms
1206.1ms
1187.0ms

erikdubbelboer:
https://developers.poki.com/player-device-report/ shows that only ~60% of the Poki users have `KHR_parallel_shader_compile`. Less than I expected.

Maksims:
> https://developers.poki.com/player-device-report/ shows that only ~60% of the Poki users have `KHR_parallel_shader_compile`. Less than I expected.

Poki's statistics comes from a very specific audience, generally it covers a lot of old Android's. Take it with a grain of salt.
Caniuse shows 92%: https://caniuse.com/mdn-api_khr_parallel_shader_compile

mvaligursky:
> Before this change every shader would be linked in serial as we would block right after linking until it is done before continuing with the next shader.

We actually don't link shaders in serial like you mention, but most of them get link called right we call compile on all of them. See here: https://github.com/playcanvas/engine/blob/a460bb662bc39e453f9afef592df9ed6d3f58626/src/platform/graphics/webgl/webgl-shader.js#L106

Note that this is used only for forward shaders, but not shadow shaders - that is still to be done.

The reason it is done this way is this article: https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#compile_shaders_and_link_programs_in_parallel

The thinking here is that if Link step is blocking (which was the case on iOS when I tested this), that at least a compilation steps run in parallel. Calling link in between compile steps would make all compilations serial as well.

It seems Safari now does not make Link blocking like it used to, and that's a good news (for newer version of the browsers). 

I think we'd need to test the behaviour here more on multiple versions / platforms (mobile) to conclude this improves the performance, as I'm not sure I see a reason.

mvaligursky:
Closing in favour of https://github.com/playcanvas/engine/pull/6076. 

Thanks @erikdubbelboer !

Issue #5888, opened 2023-12-09
Running examples


LeXXik:
I can't seem to start the examples. After using `npm install` and `npm run develop`, I get the following error (Win11, VSCode, bash):

```
[1] 
[1] > examples-browser@0.0.0 serve
[1] > serve dist -l 5000 --no-request-logging --config ../serve.json
[1] 
[0] 
[0] > examples-browser@0.0.0 watch
[0] > npm run prebuild && cross-env concurrently --kill-others "npm run rollup:watch"
[0] 
[1]  UPDATE  The latest version of `serve` is 14.2.1
[1]  INFO  Accepting connections at http://localhost:5000
[0] 
[0] > examples-browser@0.0.0 prebuild
[0] > npm run build:example:data && npm run build:standalone && npm run build:directory
[0] 
[0] 
[0] > examples-browser@0.0.0 build:example:data
[0] > node ./scripts/example-data.mjs
[0] 
[0] 
[0] > examples-browser@0.0.0 build:standalone
[0] > node ./scripts/generate-standalone-files.mjs
[0] 
[0] 
[0] > examples-browser@0.0.0 build:directory
[0] > node ./scripts/example-directory.mjs
[0] 
[0] [0] 
[0] [0] > examples-browser@0.0.0 rollup:watch
[0] [0] > cross-env NODE_ENV=development rollup -c -w
[0] [0] 
[0] [0] [!] Error: EISDIR: illegal operation on a directory, read
[0] [0] Error: EISDIR: illegal operation on a directory, read
[0] [0]     at Object.readSync (node:fs:757:3)
[0] [0]     at tryReadSync (node:fs:438:20)
[0] [0]     at Object.readFileSync (node:fs:484:19)
[0] [0]     at Object.<anonymous> (E:\dev\playcanvas\esm\engine\examples\rollup.config.js:549:35)
[0] [0]     at Module._compile (node:internal/modules/cjs/loader:1267:14)
[0] [0]     at Object.require.extensions.<computed> [as .js] (E:\dev\playcanvas\esm\engine\examples\node_modules\rollup\dist\shared\loadConfigFile.js:621:20)
[0] [0]     at Module.load (node:internal/modules/cjs/loader:1125:32)
[0] [0]     at Function.Module._load (node:internal/modules/cjs/loader:965:12)
[0] [0]     at Module.require (node:internal/modules/cjs/loader:1149:19)
[0] [0]     at require (node:internal/modules/helpers:121:18)
[0] [0]

[...]

PR #5919, opened 2023-12-26
Fix npm run build:thumbnails

erikdubbelboer:
Building thumbnails fails for `ClusteredLightingExample` and `MiniStatsExample` because they have `ENGINE = 'PERFORMANCE'` which doesn't work with thumbnail generation and fails with

```
Failed to fetch dynamically imported module: http://localhost:12321/iframe/playcanvas.prf.js
```

### DIFF
--- a/examples/rollup.config.js
+++ b/examples/rollup.config.js
@@ -218,22 +218,25 @@ const targets = [
 // We skip building PlayCanvas ourselves when ENGINE_PATH is given.
 // In that case we have a watcher which copies all necessary files.
 if (ENGINE_PATH === '') {
-    /** @type {RollupOptions|undefined} */
-    let target;
+    /** @type {buildTarget} */
+    const pushTarget = (...args) => {
+        targets.push(buildTarget(...args));
+    };
     if (NODE_ENV === 'production') {
         // Outputs: dist/iframe/playcanvas.js
-        target = buildTarget('release', 'es5', '../src/index.js', 'dist/iframe');
+        pushTarget('release', 'es5', '../src/index.js', 'dist/iframe');
+        // Outputs: dist/iframe/playcanvas.dbg.js
+        pushTarget('debug', 'es5', '../src/index.js', 'dist/iframe');
+        // Outputs: dist/iframe/playcanvas.prf.js
+        pushTarget('profiler', 'es5', '../src/index.js', 'dist/iframe');
     } else if (NODE_ENV === 'development') {
         // Outputs: dist/iframe/playcanvas.dbg.js
-        target = buildTarget('debug', 'es5', '../src/index.js', 'dist/iframe');
+        pushTarget('debug', 'es5', '../src/index.js', 'dist/iframe');
     } else if (NODE_ENV === 'profiler') {
         // Outputs: dist/iframe/playcanvas.prf.js
-        target = buildTarget('profiler', 'es5', '../src/index.js', 'dist/iframe');
+        pushTarget('profiler', 'es5', '../src/index.js', 'dist/iframe');
     } else {
         console.warn("NODE_ENV is neither production, development nor profiler.");
-    }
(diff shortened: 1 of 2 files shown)


kungfooman:
Thank you very much for pointing this out!

Now this points to a deeper issue, namely that `npm run build` is supposed to build all three targets:
 - dist/iframe/playcanvas.js
 - dist/iframe/playcanvas.dbg.js
 - dist/iframe/playcanvas.prf.js

Because every example is able to choose the one they want:

```js
export class ClusteredLightingExample {
    static CATEGORY = 'Graphics';
    static ENGINE = 'PERFORMANCE';
    static WEBGPU_ENABLED = true;
    static example = example;
}
```

So what I suggest is rewriting `rollup.config.js` a bit at the end:

```js
// We skip building PlayCanvas ourselves when ENGINE_PATH is given.
// In that case we have a watcher which copies all necessary files.
if (ENGINE_PATH === '') {
    /** @type {buildTarget} */
    const pushTarget = (...args) => {
        targets.push(buildTarget(...args));
    };
    if (NODE_ENV === 'production') {
        // Outputs: dist/iframe/playcanvas.js
        pushTarget('release', 'es5', '../src/index.js', 'dist/iframe');
        // Outputs: dist/iframe/playcanvas.dbg.js
        pushTarget('debug', 'es5', '../src/index.js', 'dist/iframe');
        // Outputs: dist/iframe/playcanvas.prf.js
        pushTarget('profiler', 'es5', '../src/index.js', 'dist/iframe');
    } else if (NODE_ENV === 'development') {
        // Outputs: dist/iframe/playcanvas.dbg.js
        pushTarget('debug', 'es5', '../src/index.js', 'dist/iframe');
    } else if (NODE_ENV === 'profiler') {
        // Outputs: dist/iframe/playcanvas.prf.js
        pushTarget('profiler', 'es5', '../src/index.js', 'dist/iframe');
    } else {
        console.warn("NODE_ENV is neither production, development nor profiler.");
    }
}
```

If we don't care to build all three targets (takes some time), we can also just do:

```
ENGINE_PATH=../src/index.js npm run build
```

(nothing is built, only the ESM files are copied into `examples/dist` to reduce time to create thumbnails)

In any case, we need all three targets for production (for `npm run build` + `npm run serve` aswell). Would you mind to edit `rollup.config.js`?

erikdubbelboer:
@kungfooman that also works, I have applied your changes instead 👍

examples/rollup.config.js
```diff
@@ -218,23 +218,25 @@ const targets = [
 // We skip building PlayCanvas ourselves when ENGINE_PATH is given.
 // In that case we have a watcher which copies all necessary files.
 if (ENGINE_PATH === '') {
-    /** @type {RollupOptions|undefined} */
-    let target;
+    const pushTarget = (...args) => {
+        targets.push(buildTarget(...args));
+    };
```
kungfooman, line 223:
```suggestion
    /** @type {buildTarget} */
    const pushTarget = (...args) => {
        targets.push(buildTarget(...args));
    };
```

(lets keep the type for full typing / IntelliSense)

willeastcott:
@kungfooman I'm trying to test this PR on my Win10 laptop but `npm run build:thumbnails` doesn't work there:

image

Looks like this is the line to blame:

```js
    const server = spawn('serve', ['dist', '-l', port, '--no-request-logging', '--config', '../serve.json']);
```

This seems to assume `serve` is globally installed maybe? Anyway, the `serve` command isn't found and the process errors out.

kungfooman:
Hey @willeastcott, we got a solution for that here: https://github.com/playcanvas/engine/issues/5888#issuecomment-1848424290

(issue is still open and requires a PR)

Maksims:
Same issue for me. And it used to work, but recently stopped.
Building on Windows also.

PR #5920, opened 2023-12-28
Add example with multiple applications at the same time

### LINKED ISSUE #5732: Engine example demonstrating multiple applications at the same time
I'd be great to have an engine example similar to the repro here: https://github.com/playcanvas/engine/issues/5729, where multiple applications can be created at the same time and destroyed.
Perhaps at the start, all 4 types would be created (gl1, gl2, gpu, null) with buttons to add more of each time, plus remove any of those (or last of each type) or similar. That'd be a great way to regularly test app destruction too.

erikdubbelboer:
Fixes #5732

Screenshot 2023-12-28 at 12 36 49

### DIFF
--- a/examples/scripts/generate-standalone-files.mjs
+++ b/examples/scripts/generate-standalone-files.mjs
@@ -76,7 +76,7 @@ function generateExampleFile(category, example, exampleClass) {
                         </a>    
                     </div>
                 </div>
-                <canvas id='application-canvas'></canvas>
+                ${exampleClass.NO_CANVAS ? '' : '<canvas id="application-canvas"></canvas>'}
             </div>
         </div>
         <script src='./playcanvas-observer.js'></script>
@@ -171,6 +171,7 @@ ${exampleClass.example.toString()}
             }
             return specifiedEngine;
         }
+        let app;
         let ready = false; // Used in indicate if UI can render Controls
         let started = false;
         let miniStats;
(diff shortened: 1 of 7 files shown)


src/scene/lighting/lights-buffer.js
```diff
@@ -249,7 +250,7 @@ class LightsBuffer {
         // textures
         this._lightsTexture8Id.setValue(this.lightsTexture8);
 
-        if (LightsBuffer.lightTextureFormat === LightsBuffer.FORMAT_FLOAT) {
+        if (LightsBuffer.lightTextureFormat === LightsBuffer.FORMAT_FLOAT && this._lightsTextureFloatId) {
```
erikdubbelboer, line 253: These are fixes for the `NullGraphicsDevice` crashing when you create another Device after it, because of their shared `LightsBuffer` global state. Creating a `NullGraphicsDevice` should not affect this state as it doesn't use the `LightsBuffer`.

examples/scripts/generate-standalone-files.mjs
```diff
@@ -171,6 +171,7 @@ ${exampleClass.example.toString()}
             }
             return specifiedEngine;
         }
+        let app;
```
kungfooman, line 174: Can you check if the tween example still works? I just remember that it also declared app in global scope in a script and two app's are crashing on each other (pitfalls of UMD)

erikdubbelboer: You are right, the tween example broken. I don't see why tween.js should create a global app variable, so I have changed that. I checked all other examples and they all work now.

kungfooman: Thank you, good call :+1:

mvaligursky (approved):
Thanks, this looks perfect!

src/scene/lighting/lights-buffer.js
mvaligursky: related: https://github.com/playcanvas/engine/pull/5935

Issue #4769, opened 2022-10-20
Add GPGPU


querielo:
Problem: I experimented [with water surface simulation](https://playcanv.as/b/1a320545), and found that there is no library for GPGPU in Playcanvas. So, a user has to create an environment for that themselves.

**Suggestion**: add GPUCompute renderer that can be used as GPGPU.
**Primitives**: 
1) variables are textures
2) the computation is based on a fragment shader 
  a) it uses textures as variables
  b) it happens before frame rendering
  c) a result of the computation is texture (another variable)
  d) a variable can have itself as a dependency (!)

kungfooman:
So you want PlayCanvas to implement basically something like https://github.com/gpujs/gpu.js?

Is your demo open source, so we can see the code of what you are currently doing to achieve the water splash effect? I only see the link to your running example.

Since GPU's become more and more powerful, I think this opens all kinds of doors, but I'm still confused what changes would be required. Right now gpu.js is a 2nd class citizen and you ask it to be 1st class in PlayCanvas?

querielo:
> So you want PlayCanvas to implement basically something like https://github.com/gpujs/gpu.js?

Yes, or maybe something simplier. At least I don't understand how to pass result of computation from gpu.js to Playcanvas Texture without CPU<->GPU round trip.


> Is your demo open source, so we can see the code of what you are currently doing to achieve the water splash effect? I only see the link to your running example.

I need to polish the code and improve naming. Maybe I do it later. But you can find idea here:
<img width="734" alt="Screen Shot 2022-10-31 at 22 28 45" src="https://user-images.githubusercontent.com/104348270/199082999-986368ff-067f-4999-84dc-98e53f838734.png">

LeXXik:
@querielo I've made one exactly like you want like a year ago or so. Each variable is a texture, and supports itself as a dependency. The only reason I haven't published it yet, is I wasn't able to get my hands on some nice example that would make use if it. I was thinking to use boids, and move a swarm around. Made a simple butterfly animated in vertex shader for it. Wanted to make a bunch of them fly around a flower or something. Still in my todo list since then :D If you have a good and simple shader example, we could make a public project together.

querielo:
@LeXXik  I thought about particles that visualize disintegrated objects that has to be hidden (like dead players or inactive game objects). So instead of hiding objects in one frame the objects are dissolved with random walking particles.
Or we could do magic stuff like fireballs, lightning, firework :-)

[...]

PR #5921, opened 2023-12-28
Support for integer/data textures in WebGL2 and WebGPU

liamdon:
Add support for many of the [signed/unsigned integer formats](https://webgl2fundamentals.org/webgl/lessons/webgl-data-textures.html#:~:text=Sized%0AFormat,Texture%0Afilterable) supported by WebGL 2.

Integer textures are very useful for emulating compute shaders in WebGL using data textures, voxel rendering and more.

With that said, new formats do require a few lines of repetitive definition code - 573 net lines here. I won't be offended if you decide that the potential use-cases are too niche to justify the added weight.

Happy Holidays!

Example:

https://github.com/playcanvas/engine/assets/453578/12953eca-0dcd-4825-a5a3-8b1b2f335d32


src/platform/graphics/constants.js
```diff
@@ -655,12 +782,36 @@ export const pixelFormatInfo = new Map([
...
+    [PIXELFORMAT_RG16I,       { name: 'RG16I', size: 4, channelSize: 2 }],
+    [PIXELFORMAT_RG16UI,       { name: 'RG16UI', size: 4, channelSize: 2 }],
+    [PIXELFORMAT_RG32I,       { name: 'RG32I', size: 8, channelSize: 4 }],
+    [PIXELFORMAT_RG32UI,       { name: 'RG32UI', size: 8, channelSize: 4 }],
+    [PIXELFORMAT_RGBA8I,       { name: 'RGBA8I', size: 4, channelSize: 1 }],
+    [PIXELFORMAT_RGBA8UI,       { name: 'RGBA8UI', size: 4, channelSize: 1 }],
+    [PIXELFORMAT_RGBA16I,       { name: 'RGBA16I', size: 8, channelSize: 2 }],
+    [PIXELFORMAT_RGBA16UI,       { name: 'RGBA16UI', size: 8, channelSize: 2 }],
+    [PIXELFORMAT_RGBA32I,       { name: 'RGBA32I', size: 16, channelSize: 4 }],
+    [PIXELFORMAT_RGBA32UI,       { name: 'RGBA32UI', size: 16, channelSize: 4 }]
```
willeastcott, line 805: Seems some superfluous spaces have been inserted here. Can you remove them, please?

willeastcott (approved):
Impressive PR - thanks for all the effort on this! 🙏  LGTM, but deferring to @mvaligursky for final approval before we can merge.

src/platform/graphics/constants.js
```diff
@@ -618,6 +618,133 @@ export const PIXELFORMAT_ATC_RGBA = 30;
  */
 export const PIXELFORMAT_BGRA8 = 31;
 
+/**
+ * 8-bit signed integer single-channel (R) format (WebGL2 only).
```
mvaligursky, line 622: I suggest changing `WebGL2 only` to `Not supported on WebGL1` on all these lines. As WebGPU supports those.

src/platform/graphics/constants.js
```diff
@@ -655,12 +782,36 @@ export const pixelFormatInfo = new Map([
     [PIXELFORMAT_PVRTC_4BPP_RGBA_1, { name: 'PVRTC_4BPP_RGBA_1', blockSize: 8 }],
     [PIXELFORMAT_ASTC_4x4, { name: 'ASTC_4x4', blockSize: 16 }],
     [PIXELFORMAT_ATC_RGB, { name: 'ATC_RGB', blockSize: 8 }],
-    [PIXELFORMAT_ATC_RGBA, { name: 'ATC_RGBA', blockSize: 16 }]
+    [PIXELFORMAT_ATC_RGBA, { name: 'ATC_RGBA', blockSize: 16 }],
+
+    // uncompressed integer formats (WebGL2 Only)
```
mvaligursky, line 787:
```suggestion
    // uncompressed integer formats (Not supported on WebGL1)
```

src/platform/graphics/constants.js
```diff
@@ -655,12 +782,36 @@ export const pixelFormatInfo = new Map([
...
+    [PIXELFORMAT_R8I,       { name: 'R8I', size: 1, channelSize: 1 }],
```
mvaligursky, line 788: `channelSize` does not seem to be used anywhere else apart from `isIntegerPixelFormat` function. Should we just add 'int` boolean flag instead?

src/platform/graphics/constants.js
```diff
@@ -1169,58 +1339,150 @@ export const TYPE_FLOAT32 = 6;
  */
 export const TYPE_FLOAT16 = 7;
 
-export const UNIFORMTYPE_BOOL = 0;
+/**
+ * Uniform and sampler types
+ */
```
mvaligursky, line 1344: not sure we need this block?

src/platform/graphics/constants.js
```diff
@@ -1169,58 +1339,150 @@ export const TYPE_FLOAT32 = 6;
...
+// Uniforms
+export const UNIFORMTYPE_FLOAT = 0;
```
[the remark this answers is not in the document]
mvaligursky: Even though these are not public members, and so should not be used directly, I'd think it'd be safer to avoid modifications.

src/platform/graphics/texture.js
```diff
@@ -201,6 +202,12 @@ class Texture {
 
         this._format = options.format ?? PIXELFORMAT_RGBA8;
         this._compressed = isCompressedPixelFormat(this._format);
+        this._integerFormat = isIntegerPixelFormat(this._format);
+        if (this._integerFormat) {
+            options.mipmaps = false;
+            options.minFilter = FILTER_NEAREST;
```
mvaligursky, line 208: We should handle setters of these function to ignore unsupported values as well for integer formats

src/platform/graphics/webgpu/constants.js
```diff
@@ -42,3 +45,21 @@ gpuTextureFormats[PIXELFORMAT_ASTC_4x4] = 'astc-4x4-unorm';
 gpuTextureFormats[PIXELFORMAT_ATC_RGB] = '';
 gpuTextureFormats[PIXELFORMAT_ATC_RGBA] = '';
 gpuTextureFormats[PIXELFORMAT_BGRA8] = 'bgra8unorm';
+gpuTextureFormats[PIXELFORMAT_R8I] = 'r8sint';
+gpuTextureFormats[PIXELFORMAT_R8UI] = 'r8uint';
+gpuTextureFormats[PIXELFORMAT_R16I] = 'r16sint';
+gpuTextureFormats[PIXELFORMAT_R16UI] = 'r16uint';
+gpuTextureFormats[PIXELFORMAT_R32I] = 'r32sint';
+gpuTextureFormats[PIXELFORMAT_R32UI] = 'r32uint';
+gpuTextureFormats[PIXELFORMAT_RG8I] = 'r8sint';
```
mvaligursky, line 54:
```suggestion
gpuTextureFormats[PIXELFORMAT_RG8I] = 'rg8sint';
```

src/platform/graphics/webgpu/constants.js
```diff
@@ -42,3 +45,21 @@ gpuTextureFormats[PIXELFORMAT_ASTC_4x4] = 'astc-4x4-unorm';
...
+gpuTextureFormats[PIXELFORMAT_RG8UI] = 'rg8uint';
+gpuTextureFormats[PIXELFORMAT_RG16I] = 'rg16sint';
+gpuTextureFormats[PIXELFORMAT_RG16UI] = 'rg16uint';
+gpuTextureFormats[PIXELFORMAT_RG32I] = 'rg32sint';
+gpuTextureFormats[PIXELFORMAT_RG32UI] = 'rg32uint';
+gpuTextureFormats[PIXELFORMAT_RGBA8I] = 'r8sint';
```
mvaligursky, line 60:
```suggestion
gpuTextureFormats[PIXELFORMAT_RGBA8I] = 'rgba8sint';
```

src/platform/graphics/uniform-buffer-format.js
```diff
@@ -96,8 +142,7 @@ class UniformFormat {
                 this.invalid = true;
         });
 
-        let elementSize = uniformTypeToNumElements[type];
-        Debug.assert(elementSize, `Unhandled uniform format ${type} used for ${name}`);
+        let elementSize = this.numElements;
```
mvaligursky, line 145: what is `this.numElements'? That is never set.

mvaligursky: needs to be `this.numComponents`

src/platform/graphics/uniform-buffer.js
```diff
@@ -1,69 +1,33 @@
...
+            dst[offset + i * numComponents + j] = value[i * numComponents + j];
+        }
+    }
+}
+
 // Uniform buffer set functions - only implemented for types for which the default
 // array to buffer copy does not work, or could be slower.
 const _updateFunctions = [];
-
-_updateFunctions[UNIFORMTYPE_FLOAT] = function (uniformBuffer, value, offset) {
```
mvaligursky, line 15: Why are these removed? Have you tested a generic approach to perform better?

src/platform/graphics/uniform-buffer.js
```diff
@@ -1,69 +1,33 @@
...
+    for (let i = 0; i < numComponents; i++) {
+        dst[offset + i] = value[i];
+    }
+}
+
+function updateArray(uniformBuffer, value, offset, numComponents, count) {
+    const dst = uniformBuffer.getStorageForType(value.type);
+    for (let i = 0; i < count; i++) {
+        for (let j = 0; j < numComponents; j++) {
+            dst[offset + i * numComponents + j] = value[i * numComponents + j];
```
mvaligursky, line 23: I have not checked the specs, but I thought there was an alignment (padding) to follow as well here? Similarly to how UNIFORMTYPE_MAT3 format is handled ... vec3 is padded to vec4 and similar rules.

src/platform/graphics/uniform-buffer.js
```diff
@@ -284,6 +233,46 @@ class UniformBuffer {
         } else {
             this.storageFloat32 = null;
             this.storageInt32 = null;
+            this.storageUint32 = null;
+        }
+    }
+
+    getStorageForType(type) {
```
mvaligursky, line 240: I suspect this is compiled to an array lookup by most browsers, but I'd prefer we don't rely on this and avoid executing this for each uniform we set - that could be costly.

src/platform/graphics/webgl/webgl-graphics-device.js
```diff
@@ -613,44 +667,46 @@ class WebglGraphicsDevice extends GraphicsDevice {
                 uniformValue[2] = scopeZ;
             }
         };
-        this.commitFunction[UNIFORMTYPE_VEC4]  = function (uniform, value) {
+        this.commitFunction[UNIFORMTYPE_IVEC3]  = function (uniform, value) {
```
mvaligursky, line 670: personally I'd avoid changing / reordering things in the same PR as adding new functionality, as that makes it tricky to make sure all is correct.

mvaligursky (commented):
Great PR, but I left a bunch of comments / fixes / suggestions before this can be merged.

mvaligursky:
Just to add here - I'd love to have an engine example here showing the use of some of these. Any thoughts on this @liamdon ?

liamdon:
Thank you for all the comments @mvaligursky! I'll address the issues above.

Regarding the engine example: yes, I thought about doing this but wanted to make sure you were OK with the idea before investing time in an example. I'll add an example to this PR, but likely over the weekend.

examples/scripts/generate-standalone-files.mjs
```diff
@@ -136,6 +136,10 @@ ${exampleClass.example.toString()}
                     console.warn('Picked WebGPU but example is not supported on WebGPU, defaulting to WebGL2');
                     return 'webgl2';
                 }
+                if (last === 'webgl1' && ${exampleClass.WEBGL1_DISABLED === true}) {
```
mvaligursky, line 139: I'd prefer to avoid 'WEBGL1_DISABLED' flag, as there is no fallback for that device - falling back to webgl2 won't work on device that only support webgl 1.

Other examples that do not support webgl1 simply do not execute code that does not work on webgl 1 .. and then render nothing.

for example

```
if (!device.isWebGL1) {
    // create / use integer textures
}

File: examples/src/examples/graphics/integer-textures.mjs [hunk 16]
```diff
@@ -0,0 +1,600 @@
...
+    );
+
+    // Write the initial simulation state to the integer texture
+    const resetData = () => {
+        const sourceTextureData = sourceTexture.lock();
+        for (let x = 0; x < sourceTexture.width; x++) {
+            for (let y = 0; y < sourceTexture.height; y++) {
+                const i = (y * sourceTexture.width + x);
+                if (x > sourceTexture.width * 0.3 && x < sourceTexture.width * 0.7 && y > sourceTexture.height * 0.7 && y < sourceTexture.height * 0.8) {
+                    sourceTextureData[i] = 4;
```
mvaligursky, line 187: add some comments here to make it easier to follow by people

examples/src/examples/graphics/integer-textures.mjs
```diff
@@ -0,0 +1,600 @@
...
+        cameraEntity.addComponent("camera", {
+            clearColor: new pc.Color(0.4, 0.45, 0.5),
+            farClip: 500
+        });
+
+        // add camera to the world
+        cameraEntity.setPosition(0, 5, 15);
+        cameraEntity.lookAt(0, 5, 0);
+        app.root.addChild(cameraEntity);
+        if (cameraEntity.camera === undefined) throw new Error('Camera component expected');
```
mvaligursky, line 224: no need for this line, camera component is added above.

examples/src/examples/graphics/integer-textures.mjs
```diff
@@ -0,0 +1,600 @@
...
+        // simulation visualization texture
+        const gameScreen = new pc.Entity();
+        gameScreen.addComponent('render', {
+            castShadows: true,
+            receiveShadows: false,
+            castShadowsLightmap: false,
+            lightmapped: false,
+            type: "plane"
+        });
+        if (!gameScreen.render) throw new Error('Render component expected');
```
mvaligursky, line 274: not needed

examples/src/examples/graphics/integer-textures.mjs
```diff
@@ -0,0 +1,600 @@
...
+        screen.addComponent("screen", {
+            referenceResolution: new pc.Vec2(1280, 720),
+            scaleBlend: 0.5,
+            scaleMode: pc.SCALEMODE_BLEND,
+            screenSpace: true
+        });
+        app.root.addChild(screen);
+
+        // Help text
+        const textMarkup = new pc.Entity();
```
mvaligursky, line 373: maybe use this https://github.com/playcanvas/engine/pull/5922

examples/src/examples/graphics/integer-textures.mjs
```diff
@@ -0,0 +1,600 @@
...
+    return app;
+}
+
+export class IntegerTextureExample {
+    static CATEGORY = 'Graphics';
+    static WEBGPU_ENABLED = true;
+    static WEBGL1_DISABLED = true;
+    static example = example;
+    static controls = controls;
+    static sharedShaderChunks = {
```
mvaligursky, line 432: could this also be part of FILES so that the built in code editor displays this file?

mvaligursky: ah actually that is not needed, you inlined this into the shaders directly.

src/platform/graphics/constants.js
```diff
@@ -618,6 +618,133 @@ export const PIXELFORMAT_ATC_RGBA = 30;
...
+ * @type {number}
+ */
+export const PIXELFORMAT_R8I = 32;
+
+/**
+ * 8-bit unsigned integer single-channel (R) format (WebGL2 only).
+ *
+ * @type {number}
+ */
+export const PIXELFORMAT_R8UI = 33;
```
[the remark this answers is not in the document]
mvaligursky: any thoughts on this?

src/platform/graphics/constants.js
```diff
@@ -1220,9 +1424,90 @@ export const uniformTypeToName = [
...
+    'isamplerCube',
+    'usamplerCube',
+    'isampler3D',
+    'usampler3D',
+    'isampler2DArray',
+    'usampler2DArray'
 ];
 
+// Map used in uniform-buffer.js to convert uniform type to storage type
+// used in the uniform buffer.
```
mvaligursky, line 1457: Cleanup this comment a bit please, it says twice the same.

src/scene/shader-lib/utils.js
```diff
@@ -40,9 +42,10 @@ function createShader(device, vsName, fsName, useTransformFeedback = false) {
  * attribute names to semantics SEMANTIC_*. This enables the engine to match vertex buffer data as
  * inputs to the shader. Defaults to undefined, which generates the default attributes.
  * @param {boolean} [useTransformFeedback] - Whether to use transform feedback. Defaults to false.
+ * @param {string} [fragmentOutputType] - Output fragment shader type. Defaults to vec4.
  * @returns {Shader} The newly created shader.
  */
-function createShaderFromCode(device, vsCode, fsCode, uniqueName, attributes, useTransformFeedback = false) {
+function createShaderFromCode(device, vsCode, fsCode, uniqueName, attributes, useTransformFeedback = false, fragmentOutputType = 'vec4') {
```
mvaligursky, line 48: - I think its time we add `options` parameter here instead of just another parameter now that we need to add a new one. 
- I think instead of a single output type, we need an array of those, allowing us to specify different output per render target in MRT scenario
- other parts of the code need to be adjusted to handle this, but that looks straightforward - I like the way you handle it.

src/platform/graphics/shader-chunks/frag/gles3.js
```diff
@@ -1,45 +1,50 @@
 export default /* glsl */`
-layout(location = 0) out highp vec4 pc_fragColor;
+
+#ifndef outType
```
mvaligursky, line 3: So we'd need a numbered outTypes here, one per pc_fragColor*

src/platform/graphics/shader-utils.js
```diff
@@ -51,19 +51,27 @@ class ShaderUtils {
...
+        const getDefines = (gpu, gl2, gl1, isVertex, options) => {
 
             const deviceIntro = device.isWebGPU ? gpu :
                 (device.isWebGL2 ? gl2 : ShaderUtils.gl1Extensions(device, options) + gl1);
 
             // a define per supported color attachment, which strips out unsupported output definitions in the deviceIntro
             let attachmentsDefine = '';
+
+            // Define the fragment shader output type, vec4 by default
+            if (!isVertex) {
```
mvaligursky, line 70: perhaps merge this with a loop bellow (and handle an array of out types)

liamdon:
Thanks for the additional comments @mvaligursky. It was a good idea to implement the example as I found some issues with transpiling shaders to WGSL, which are now fixed.

For these compute-ish use-cases, it's very useful to be able to specify the output type! So you saw the change I made there - I will update it to better support the MRT case.

A future PR I've been thinking about is a higher-level compute API that operates on any TypedArray, with shaders that have a GLSL-ish syntax similar to other PlayCanvas shaders. On WebGPU, this could be transpiled to a compute shader, while on WebGL it will use a fragment shader / renderTarget approach similar to my example. The selection of a texture format would be transparent to the user, they will just pass in a TypedArray and we will select an approach and create a texture based on the type and stride options.

liamdon:
@mvaligursky updated to use an options object in shader creation, and allow specifying an output type per color attachment.

See relevant changes:
[shader-utils.js](https://github.com/playcanvas/engine/blob/63febb9af1ea7b578b0d78327bfaa714ea6285b7/src/platform/graphics/shader-utils.js#L68-L82)
[gles3 chunk](https://github.com/playcanvas/engine/pull/5921/files#diff-b90b6c93d71096ad38b8ae1822fb79bd7c3abcbef0583cc187caeabfc4239e4f)
[webgpu chunk](https://github.com/playcanvas/engine/pull/5921/files#diff-fde9f197297e15f152a96361bfe38b9127be4f0b401143e182774555843eb990) 

Note that in the webgpu header we're defining a default outType for all targets _inside the preprocessor_ because unlike in gles3, these targets are not wrapped with a directive. In theory they should always be defined by `ShaderUtils.createDefinition` but I don't want shader compilation to fail if they're created via some other path.

liamdon:
@mvaligursky I think the only thing left to resolve is the question about the `UI` naming for `PIXELFORMAT_`.

I shared my perspective [here](https://github.com/playcanvas/engine/pull/5921/files/a5aff3e6ca16d196c40bc9dd96a380c0e7d2fd6d#diff-38f2220b1ae9045c0793b9fefa07ead038eebd7b5934cdcf78ab47352b833a07R633), which is that we should keep `UI` for now as it maps to the OpenGL format names. However, as the engine maintainer, you have the final call; if you prefer the single character naming I am happy to update it.

src/platform/graphics/shader-utils.js
```diff
@@ -50,22 +50,35 @@ class ShaderUtils {
      * @param {string} [options.fragmentDefines] - The fragment shader defines.
      * @param {string} [options.fragmentExtensions] - The fragment shader extensions code.
      * @param {string} [options.fragmentPreamble] - The preamble string for the fragment shader.
-     * @param {boolean} [options.useTransformFeedback] - Whether to use transform feedback. Defaults
-     * to false.
+     * @param {boolean} [options.useTransformFeedback] - Whether to use transform feedback. Defaults to false
```
mvaligursky, line 53:
```suggestion
     * @param {boolean} [options.useTransformFeedback] - Whether to use transform feedback. Defaults to false.
```

src/platform/graphics/shader-utils.js
```diff
@@ -50,22 +50,35 @@ class ShaderUtils {
...
             let attachmentsDefine = '';
-            for (let i = 0; i < device.maxColorAttachments; i++) {
-                attachmentsDefine += `#define COLOR_ATTACHMENT_${i}\n`;
+
+            // Normalize fragmentOutputTypes to an array
+            let fragmentOutputTypes = options.fragmentOutputTypes ?? ['vec4'];
+            if (!Array.isArray(fragmentOutputTypes)) {
+                fragmentOutputTypes = [fragmentOutputTypes ?? 'vec4'];
+            }
+
```
mvaligursky, line 73: slightly simplified.
```
            let fragmentOutputTypes = options.fragmentOutputTypes ?? 'vec4';
            if (!Array.isArray(fragmentOutputTypes)) {
                fragmentOutputTypes = [fragmentOutputTypes];
            }
```
But also, this can be inside the `if (!isVertex)` condition, it does not need to run otherwise.

src/platform/graphics/shader-utils.js
```diff
@@ -50,22 +50,35 @@ class ShaderUtils {
...
+            // Define the fragment shader output type, vec4 by default
+            if (!isVertex) {
+                let outType = '';
```
mvaligursky, line 76: remove this line and just add a `const` to where its used below

src/platform/graphics/webgpu/webgpu-bind-group-format.js
```diff
@@ -9,11 +9,15 @@ const samplerTypes = [];
 samplerTypes[SAMPLETYPE_FLOAT] = 'filtering';
 samplerTypes[SAMPLETYPE_UNFILTERABLE_FLOAT] = 'non-filtering';
 samplerTypes[SAMPLETYPE_DEPTH] = 'comparison';
+samplerTypes[SAMPLETYPE_INT] = 'comparison';
+samplerTypes[SAMPLETYPE_UINT] = 'comparison';
```
mvaligursky, line 13: should these be `non-filtering` instead of `comparison`?

src/scene/shader-lib/utils.js
```diff
@@ -12,15 +12,18 @@ import { ShaderGenerator } from './programs/shader-generator.js';
...
  * @param {string} fsName - The fragment shader chunk name.
- * @param {boolean} [useTransformFeedback] - Whether to use transform feedback. Defaults to false.
+ * @param {object} [shaderDefinitionOptions] - Additional options that will be added to the shader definition.
+ * @param {boolean} [shaderDefinitionOptions.useTransformFeedback] - Whether to use transform feedback. Defaults to false.
+ * @param {string | string[]} [shaderDefinitionOptions.fragmentOutputTypes] - Fragment shader output types, which default to vec4. Passing a string will set the output type for all color attachments. Passing an array will set the output type for each color attachment.
+ * @see ShaderUtils.createDefinition
  * @returns {Shader} The newly created shader.
  */
-function createShader(device, vsName, fsName, useTransformFeedback = false) {
+function createShader(device, vsName, fsName, shaderDefinitionOptions) {
```
mvaligursky, line 21: both functions in this file are part of the public API and we cannot change their signature. Meaning we can add the options at the end, but after the `useTransformFeedback`, which should not be removed.

Alternatively (probably a better option), as `useTransformFeedback` is pretty uncommon, would be to do what you do, but handle a case where shaderDefinitionOptions is a boolean and not an object, allowing backwards compatibility.

src/platform/graphics/constants.js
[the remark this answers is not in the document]
mvaligursky: I'd like to follow this existing convention of our API here instead of WebGL convention:
`UNIFORMTYPE_IVEC3, UNIFORMTYPE_UVEC3, UNIFORMTYPE_BVEC3`

src/scene/graphics/render-pass-shader-quad.js
```diff
@@ -89,10 +89,10 @@ class RenderPassShaderQuad extends RenderPass {
         return this._shader;
     }
 
-    createQuadShader(name, fs) {
+    createQuadShader(name, fs, fragmentOutputType = 'vec4') {
         return createShaderFromCode(this.device, RenderPassShaderQuad.quadVertexShader, fs, name, {
             aPosition: SEMANTIC_POSITION
-        });
+        }, false, fragmentOutputType);
```
mvaligursky, line 95: and this accept options now

mvaligursky:
Nice updates, few more small changes and we're done.

> A future PR I've been thinking about is a higher-level compute API that operates on any TypedArray

I'd say this is something that would be outside of the core engine. Possibly a separate repo, or at least part of the `/extras`. But sounds very exciting! Existing issue on this: https://github.com/playcanvas/engine/issues/4769

It would be good to start discussion on this before writing code to see where this would belong I think.

liamdon:
@mvaligursky addressed all your comments above - I really appreciate the rounds of review on this, the PR is much better now!

Want to make sure you saw my note on `comparison` [here](https://github.com/playcanvas/engine/pull/5921/files/d4bd68990098593f87bbb4d6334102e5415a4b77#diff-5b598c9b7e99921dcc61f368853c19a5e798363010ee5e7d744ef842ec957a4fR13). I think we should leave this as-is and potentially change it if we discover that we're setting up our samplers incorrectly elsewhere. For now, we are definitely covering the most common use-case of using `texelFetch` with an int or uint texture.

src/platform/graphics/constants.js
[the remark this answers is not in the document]
mvaligursky: Please address this one as well.

mvaligursky:
> Want to make sure you saw my note on `comparison` [here](https://github.com/playcanvas/engine/pull/5921/files/d4bd68990098593f87bbb4d6334102e5415a4b77#diff-5b598c9b7e99921dcc61f368853c19a5e798363010ee5e7d744ef842ec957a4fR13).

I'm not sure you added the note. Perhaps it'd be the best to add a comment in the code to explain.

src/platform/graphics/shader-utils.js
```diff
@@ -50,22 +50,34 @@ class ShaderUtils {
      * @param {string} [options.fragmentDefines] - The fragment shader defines.
      * @param {string} [options.fragmentExtensions] - The fragment shader extensions code.
      * @param {string} [options.fragmentPreamble] - The preamble string for the fragment shader.
-     * @param {boolean} [options.useTransformFeedback] - Whether to use transform feedback. Defaults
-     * to false.
+     * @param {boolean} [options.useTransformFeedback] - Whether to use transform feedback. Defaults to false.
+     * @param {string | string[]} [options.fragmentOutputTypes] - Fragment shader output types, which default to vec4. Passing a string will set the output type for all color attachments. Passing an array will set the output type for each color attachment.
```
willeastcott, line 54:
```suggestion
     * @param {string | string[]} [options.fragmentOutputTypes] - Fragment shader output types,
     * which default to vec4. Passing a string will set the output type for all color attachments.
     * Passing an array will set the output type for each color attachment.
```

src/platform/graphics/shader.js
```diff
@@ -54,6 +54,7 @@ class Shader {
      * WebGPU platform.
      * @param {boolean} [definition.useTransformFeedback] - Specifies that this shader outputs
      * post-VS data to a buffer.
+     * @param {string | string[]} [definition.fragmentOutputTypes] - Fragment shader output types, which default to vec4. Passing a string will set the output type for all color attachments. Passing an array will set the output type for each color attachment.
```
willeastcott, line 57:
```suggestion
     * @param {string | string[]} [definition.fragmentOutputTypes] - Fragment shader output types,
     * which default to vec4. Passing a string will set the output type for all color attachments.
     * Passing an array will set the output type for each color attachment.
```

src/platform/graphics/webgpu/webgpu-bind-group-format.js
```diff
@@ -10,10 +10,21 @@ samplerTypes[SAMPLETYPE_FLOAT] = 'filtering';
 samplerTypes[SAMPLETYPE_UNFILTERABLE_FLOAT] = 'non-filtering';
 samplerTypes[SAMPLETYPE_DEPTH] = 'comparison';
 
+// Using 'comparison' instead of 'non-filtering' may seem unusual,
+// but currently we will get a validation error if we use 'non-filtering'
+// along with texelFetch/textureLoad. 'comparison' works very well
+//  for the most common use-case of integer textures, texelFetch.
+// We may be able to change how we initialize the sampler elsewhere
+// to support 'non-filtering' in the future.
```
willeastcott, line 18:
```suggestion
// Using 'comparison' instead of 'non-filtering' may seem unusual, but currently we will get a
// validation error if we use 'non-filtering' along with texelFetch/textureLoad. 'comparison' works
// very well for the most common use-case of integer textures, texelFetch. We may be able to change
// how we initialize the sampler elsewhere to support 'non-filtering' in the future.
```

src/scene/graphics/render-pass-shader-quad.js
```diff
@@ -89,10 +89,25 @@ class RenderPassShaderQuad extends RenderPass {
...
-            aPosition: SEMANTIC_POSITION
-        });
+    /**
+     * Creates a quad shader from the supplied fragment shader code.
+     *
+     * @param {string} name - A name of the shader.
+     * @param {string} fs - Fragment shader source code.
+     * @param {object} [shaderDefinitionOptions] - Additional options that will be added to the shader definition.
+     * @param {boolean} [shaderDefinitionOptions.useTransformFeedback] - Whether to use transform feedback. Defaults to false.
+     * @param {string | string[]} [shaderDefinitionOptions.fragmentOutputTypes] - Fragment shader output types, which default to vec4. Passing a string will set the output type for all color attachments. Passing an array will set the output type for each color attachment.
```
willeastcott, line 99:
```suggestion
     * @param {string | string[]} [shaderDefinitionOptions.fragmentOutputTypes] - Fragment shader
     * output types, which default to vec4. Passing a string will set the output type for all color
     * attachments. Passing an array will set the output type for each color attachment.
```

src/scene/shader-lib/utils.js
```diff
@@ -39,23 +54,37 @@ function createShader(device, vsName, fsName, useTransformFeedback = false) {
  * @param {Object<string, string>} [attributes] - Object detailing the mapping of vertex shader
  * attribute names to semantics SEMANTIC_*. This enables the engine to match vertex buffer data as
  * inputs to the shader. Defaults to undefined, which generates the default attributes.
- * @param {boolean} [useTransformFeedback] - Whether to use transform feedback. Defaults to false.
+ * @param {boolean | Record<string, boolean | string | string[]>} [useTransformFeedback] - Whether to use transform feedback. Defaults to false.
+ * @param {object} [shaderDefinitionOptions] - Additional options that will be added to the shader definition.
+ * @param {boolean} [shaderDefinitionOptions.useTransformFeedback] - Whether to use transform feedback. Defaults to false.
+ * @param {string | string[]} [shaderDefinitionOptions.fragmentOutputTypes] - Fragment shader output types, which default to vec4. Passing a string will set the output type for all color attachments. Passing an array will set the output type for each color attachment.
```
willeastcott, line 60:
```suggestion
 * @param {boolean | Record<string, boolean | string | string[]>} [useTransformFeedback] - Whether
 * to use transform feedback. Defaults to false.
 * @param {object} [shaderDefinitionOptions] - Additional options that will be added to the shader
 * definition.
 * @param {boolean} [shaderDefinitionOptions.useTransformFeedback] - Whether to use transform
 * feedback. Defaults to false.
 * @param {string | string[]} [shaderDefinitionOptions.fragmentOutputTypes] - Fragment shader
 * output types, which default to vec4. Passing a string will set the output type for all color
 * attachments. Passing an array will set the output type for each color attachment.
```

src/scene/graphics/render-pass-shader-quad.js
willeastcott, line 99:
```suggestion
     * @param {object} [shaderDefinitionOptions] - Additional options that will be added to the
     * shader definition.
     * @param {boolean} [shaderDefinitionOptions.useTransformFeedback] - Whether to use transform
     * feedback. Defaults to false.
     * @param {string | string[]} [shaderDefinitionOptions.fragmentOutputTypes] - Fragment shader
     * output types, which default to vec4. Passing a string will set the output type for all color
     * attachments. Passing an array will set the output type for each color attachment.
```

liamdon:
@willeastcott No problem! Do you use a formatter or IDE extension that flags those for you, or just eyeballing?

willeastcott:
Just eyeballed it. 😄

liamdon:
Awesome, just enabled that. There's always a new VSCode feature to discover!

I realized that my earlier replies/comments to you were not submitted @mvaligursky, I did not toggle on the Github review mode 🤦‍♂️ 
[Here](https://github.com/playcanvas/engine/pull/5921/files#r1453716072) is that last comment I was talking about, but I also added similar comment in the code to clarify the use of `'comparison'`. After we've merged this PR, I will experiment with using/supporting other types of samplers with integer textures. But, I think the PR as-is will cover the most common way they are used.

Let me know if I missed anything else, otherwise I think we are good to go!

src/platform/graphics/webgpu/webgpu-bind-group-format.js
mvaligursky: Perfect, thanks for the details.

mvaligursky:
The only remaining issue for me is to change the pixel format `'UI` to `U`, to keep the type a single letter.
We already have `F` for float (PIXELFORMAT_RGB16F), we're adding `I` for int types, and I'd prefer `U` for uint types instead of `UI`.
You mention we can later rename those .. but lets avoid that, as deprecating them and having new names means the engine needs to support both, and it creates a possibly more work to fix examples, Editor projects and similar.

Single letter type is something I'm already using for uniform types.

Naming compatibility with WebGL naming is not very relevant in our API.

liamdon:
Done - was not a strong opinion against that from me, but I wanted to present the 'WebGL' reasoning in case it was important. We are now using a single character for all the type suffixes.

PR #5922, opened 2023-12-29
Show configurable example descriptions

### LINKED ISSUE #5178: Add descriptions to engine-only examples
We have some good engine-only examples. Some examples are self-explanatory, but some are hard to understand what they are showing, without reading through the code. It would be good to add some UI way to show a description of what the example is showing and how to use it.

For example, Component Properties loads an example of a box and some lights moving. It is not immediately obvious what the example is demonstrating without reading the code.

image

erikdubbelboer:
Add new `static DESCRIPTION = '...';` to example classes. This description is a HTML string and can contain tags such as `<a>`.

Fixes #5178

### DIFF
--- a/examples/src/app/example.mjs
+++ b/examples/src/app/example.mjs
@@ -36,6 +36,7 @@ import * as ReactPCUI from '@playcanvas/pcui/react';
  * @property {Function} controls - Controls function from example.
  * @property {'code' | 'parameters'} show - Used in case of mobile view.
  * @property {Record<string, string>} files - Files of example (controls, shaders, example itself)
+ * @property {string} description - Description of example.
  */
 
 /** @type {typeof Component<Props, State>} */
@@ -50,7 +51,8 @@ class Example extends TypedComponent {
         //controls: () => jsx('pre', null, 'Status: initial'),
         controls: () => undefined,
         show: 'code',
-        files: {'example.mjs': '// loading'}
+        files: {'example.mjs': '// loading'},
+        description: '',
     };
 
     /**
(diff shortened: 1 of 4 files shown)


examples/src/app/example.mjs
```diff
@@ -218,67 +242,76 @@ class Example extends TypedComponent {
     }
 
     renderPortrait() {
```
erikdubbelboer, line 244: Best to watch this part with "Hide whitespace" enabled as it's just more indentation for the addition of the `fragment()` and is really only 2 lines of actual change.

examples/src/examples/animation/component-properties.mjs
```diff
@@ -296,6 +296,7 @@ async function example({ canvas, deviceType, data, assetPath, glslangPath, twgsl
 class ComponentPropertiesExample {
     static CATEGORY = 'Animation';
     static WEBGPU_ENABLED = true;
+    static DESCRIPTION = 'This example demonstrates how to use the Anim Component to animate the properties of other Components.';
```
erikdubbelboer, line 299: For now I have only added a description to this example. Let me know if the code is correct so I can add text to more examples that aren't clear from their name.

kungfooman:
You really got some holidays :sweat_smile: Very nice to see some action on this. Can you post a small screenshot how it looks?

erikdubbelboer:
It looks like this on desktop:
Screenshot 2023-12-29 at 15 59 10

And like this on mobile:
Screenshot 2023-12-29 at 15 59 21

Maksims:
This is useful. Can this be controlled from code, so we can change it to communicate some messages through it?
In XR examples I had to implement an element for that, but generic one would be great.

erikdubbelboer:
The original task suggested "Ideally it would be something rendered using html, to not pollute the example code with element components and similar", so it's currently a static config and not controllable from the code. In theory it would be possible to add this of course.

mvaligursky (approved):
Fantastic, exactly what we needed!

Issue #5735, opened 2023-10-09
WebGPU has incorrect tangents


slimbuck:
Which you can confirm by loading

https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/NormalTangentTest

in model viewer.

mvaligursky:
related: https://github.com/playcanvas/engine/pull/5924

willeastcott:
@mvaligursky The related PR you link to is now merged. I just checked to see if this still looks wrong in WebGPU and it does. Do you know what the problem is?

erikdubbelboer:
The related PR only fixed it on webgl. I wasn't able to figure out why it's broken on WebGPU.

PR #5924, opened 2023-12-29
Add example to test normals and tangents (and fix WebGL)

erikdubbelboer:
This fixes normals and tangents in the WebGL2 shader. WebGPU is still broken, see: https://github.com/playcanvas/engine/issues/5735

Screenshot 2024-03-31 at 11 53 32

See also https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/NormalTangentTest


examples/src/examples/graphics/normals-and-tangents.mjs
```diff
@@ -0,0 +1,121 @@
...
+    app.root.addChild(directionalLight);
+
+    app.start();
+    return app;
+}
+
+class NormalsAndTangentsExample {
+    static CATEGORY = 'Graphics';
+    static WEBGPU_ENABLED = true;
+    static DESCRIPTION = '<a href="https://github.com/KhronosGroup/glTF-Sample-Models/tree/main/2.0/NormalTangentTest">Model and textures by Ed Mackey</a>, copyright 2017-2018 Analytical Graphics Inc. CC-BY 4.0 <a href="https://creativecommons.org/licenses/by/4.0/">https://creativecommons.org/licenses/by/4.0/</a>.';
```
erikdubbelboer, line 117: Depends on https://github.com/playcanvas/engine/pull/5922 to show the copyright in the description.

erikdubbelboer:
I tried to find a way to fix these issues but I'm not sure where to start. Should this be fixed in the GLB parser or in somewhere else? If anyone can point me in the right direction I can try and fix the issues in this pull request as well.

slimbuck:
Oh that's interesting. I never realised our lighting basis is wrong on the back faces. (We were aware of WebGPU issues though). Thanks for pointing this out!

erikdubbelboer:
@slimbuck if you point me in the right direction, I can try and fix it. Are the normals and tangents calculated at a specific place? I couldn't easily find it as I'm not familiar enough with the source yet.

slimbuck:
I would have expected backfaces to be handled by this bit of code here: https://github.com/playcanvas/engine/blob/main/src/scene/shader-lib/programs/lit-shader.js#L989

The actual tangents and binormals are calculated in this shader chunk:
https://github.com/playcanvas/engine/blob/main/src/scene/shader-lib/chunks/lit/frag/TBNderivative.js

erikdubbelboer:
I have fixed the calculations for both WebGL and WebGPU. I have also fixed the WebGPU version looking so weird, this was caused by setting `mipmaps: true` on the `envAtlas`.

src/scene/shader-lib/chunks/lit/frag/TBNderivative.js
```diff
@@ -20,6 +20,19 @@ void getTBN(vec3 tangent, vec3 binormal, vec3 normal) {
...
+        #else
+            dTBN = mat3(T * invmax, -B * invmax, gl_FrontFacing ? normal : -normal);
+        #endif
+    #else
+        #ifdef WEBGPU
+            dTBN = mat3(T * invmax, -B * invmax, -normal);
+        #else
+            dTBN = mat3(T * invmax, -B * invmax, normal);
+        #endif
+    #endif
```
slimbuck, line 36: Webgpu lighting space should not be different to webgl, so this normal negation is terribly confusing.

Are textures inverted (upside down) on webgpu vs webgl @mvaligursky? (If so then tangents and binormals as calculated would be inverted).

I think we need to get to the bottom of why normals needs flipping here and fix it at the source. My suspicion is we have an inverted transform somewhere else in the graphics pipeline under webgpu.

examples/src/examples/graphics/normals-and-tangents.mjs
mvaligursky: We don't need to display the attribution, we just need to have it stored in the file along with the glb. See this as an example of what we do: https://github.com/playcanvas/engine/blob/main/examples/assets/models/IridescentDishWithOlives.txt

src/scene/shader-lib/chunks/lit/frag/TBNderivative.js
slimbuck: Actually I just noticed, looking in lit-shader.js, it seems we are already attempting to handle the normal for double-sided lighting case at https://github.com/playcanvas/engine/blob/main/src/scene/shader-lib/programs/lit-shader.js#L989. This calc must have a 🐛...

mvaligursky:
I'd be definitely great to have this example as a test for normals / tangents.

src/scene/shader-lib/chunks/lit/frag/TBNderivative.js
erikdubbelboer: You are right, it seems the issue is that it was handled in the wrong place. I have added a negatively scaled mesh to the example and I have gone though all variations of the shader to maker sure they all work correctly now.

src/scene/shader-lib/chunks/lit/frag/TBNfast.js
```diff
@@ -1,5 +1,13 @@
 export default /* glsl */`
 void getTBN(vec3 tangent, vec3 binormal, vec3 normal) {
-    dTBN = mat3(tangent, binormal, normal);
+    #ifdef TWO_SIDED_LIGHTING
+        dTBN = mat3(
+            (gl_FrontFacing ? tangent  : -tangent ) * twoSidedLightingNegScaleFactor,
+            (gl_FrontFacing ? binormal : -binormal) * twoSidedLightingNegScaleFactor,
+            (gl_FrontFacing ? normal   : -normal  ) * twoSidedLightingNegScaleFactor
+        );
```
slimbuck, line 8: Could simplify to something like
```suggestion
    dTBN = mat3(tangent, binormal, normal) * (-float(gl_FrontFacing) * twoSidedLightingNegScaleFactor);
```

willeastcott: I'd vote for:

```
    dTBN = mat3(tangent, binormal, normal);

    #ifdef TWO_SIDED_LIGHTING
    dTBN *= ((gl_FrontFacing ? 1.0 : -1.0) * twoSidedLightingNegScaleFactor);
    #endif
```

erikdubbelboer: Fixed.

erikdubbelboer:
I have rebased the example on main with the new examples refactor. What needs to happen to get this merged now?

willeastcott:
Thanks for doing that, @erikdubbelboer - we'll take another look ASAP! 😄

examples/src/examples/graphics/normals-and-tangents/config.mjs
```diff
@@ -0,0 +1,6 @@
+/**
+ * @type {import('../../../../types.mjs').ExampleConfig}
+ */
+export default {
+    WEBGPU_ENABLED: true,
```
mvaligursky, line 5:
```suggestion
    HIDDEN: true,
    WEBGPU_ENABLED: true
```

mvaligursky: I'm suggesting to hide it from public .. this is more of an internal example for us to validate how things work, and not an example to show the public how to use the engine.

mvaligursky:
It seems the description only shows 'before' state, but not 'after' state - could you please update it.

slimbuck:
Hi @erikdubbelboer,

Looking at the changes, I think it makes sense to handle twoSidedLighting in just one place - after dTBN has been calculated.

This is probably best done in a new shader chunk. So I went ahead and implemented this approach in https://github.com/playcanvas/engine/compare/main...slimbuck:erik-normals-and-tangents.

Could you take a look and let me know what you think? This approach also means we probably fix the case where `TBNObjectSpace.js` is used.

If you're happy with this I can push directly to your branch?

Thanks!

erikdubbelboer:
@slimbuck great, I have merged your changes.

@mvaligursky I think this pull is ready now.

slimbuck (approved):

mvaligursky (approved):
This is much simpler now, thanks!

PR #5926, opened 2024-01-02
Fix some ShaderPass JSDoc

kungfooman:
I looked a bit into the shader pass examples and figured a type issue via RuntimeTypeInspector.js, which was also a nice opportunity to simply extend the documentation a bit.

I renamed `ShaderPassInfo#initShaderDefines` to `ShaderPassInfo#buildShaderDefines` because all types need to be set in the constructor to fix this kind of type error:

image


mvaligursky (approved):
Nice cleanup, thanks!

PR #5928, opened 2024-01-05
Gizmo System

### LINKED ISSUE #2904: Add a gizmo system
The Editor has a great translate/rotate/scale gizmo.

gizmo

It would be great if the engine supported a generic gizmo system that could be shared between various applications. Here are some feature I would like to support in this system:

* Shouldn't be dependent on framebuffer picker (because this relies on the GPU and is inefficient). Just use math to perform ray-intersection tests.
* Support translate/rotate/scale but potentially any type of gizmo that could depend on hover and pick+drag type input.
* Configurable graphical representation (size on screen, thickness/size of graphical elements, etc)
* Shading controls (flat shaded vs lit)

kpal81xd:
Fixes #2904 

## Overview

- A reworked port of the gizmo system from the editor to engine extras. Allows for gizmo customisation and modular for easy extension with new gizmos. 
- Added an example to showcase translation gizmos.

### Translation
image

### Rotation
image

### Scaling
image

### Draft 1
- [x] Full test coverage 
- [x] JSDocs
- [x] Add shading
- [x] Replace torus for rotate gizmo with custom mesh
- [x] Uniform scaling
- [x] Fix camera facing rotation
- [x] Fixed rotation stuttering at small glancing angles of axes (local rotation)
- [x] Separate intersection and rendering meshes

### Draft 2
- [x] Plane to show movement axis lines rather than plane axis
- [x] Change fly to orbit camera (remove distance option)
- [x] Add intersection tolerance as an option 
- [x] Sort intersection mesh by depth
- [x] Fixed center box transparency when hovering
- [x] Add starting line for rotation
- [x] Added picker selection in example
- [x] Change rotation to use mouse displacement rather than 3D world position (see https://threejs.org/examples/misc_controls_transform.html)

## Possible features
- Add text indicator to show linear or angular distance moved for transformations - useful for fine control


extras/gizmo/gizmo.js
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    attach(nodes) {
+        this.nodes = nodes;
+        this.updateGizmoPosition();
+        this.updateGizmoRotation();
+
+        window.addEventListener('pointermove', this._onPointerMove);
+        window.addEventListener('pointerdown', this._onPointerDown);
+        window.addEventListener('pointerup', this._onPointerUp);
```
willeastcott, line 103: I love that this is based on Pointer Events. Touch input for free! ❤️

Maksims:
That's cool! Gizmos are hard, especially for multi-selection and weird cases of non-uniform parent scales, etc.
An easy API to bring generic gizmos, be able to show/hide them, and get events to apply on selection - is important to get them usable. And I think can be the most challenging part.

extras/gizmo/gizmo-transform.js
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    // _createLight(angles) {
+    //     const light = new Entity('light');
+    //     light.addComponent('light', {
+    //         layers: [this.layerGizmo.id]
+    //     });
+    //     light.setEulerAngles(angles);
+    //     return light;
+    // }
```
willeastcott, line 321: Seems as though you originally planned to light the gizmo geometry. For this first version, what is the plan? To just go with flat shading for now, switch to lit geometry or have an option for lighting?

extras/gizmo/gizmo-transform.js
```diff
@@ -0,0 +1,364 @@
...
+        // lighting
+        // const light = this._createLight(new Vec3(45, 0, -45));
+        // this.gizmo.addChild(light);
+
+        // center
+        this._center = new Entity('center');
+        this.gizmo.addChild(this._center);
+    }
+
+    set axisXColor(value) {
```
willeastcott, line 333: This seems to take an array of numbers....but shouldn't it take a `Color`?

extras/gizmo/gizmo.js
```diff
@@ -0,0 +1,275 @@
...
+        }
+        return selection;
+    }
+
+    _rayIntersectsTriangle(origin, dir, v0, v1, v2, out) {
+        const e1 = new Vec3();
+        const e2 = new Vec3();
+        const h = new Vec3();
+        const s = new Vec3();
+        const q = new Vec3();
```
willeastcott, line 242: Just a note to say that it might be worth using preallocated vectors here to generate less garbage.

willeastcott:
Something else for the TODO list: JSDocs, at least for any API you consider public?

willeastcott:
I have a question: are the intersection shapes for selection/picking independent of the graphical configuration of the gizmo? From the quick scan of the code, they look the same. I would imagine that if a gizmo axis was set to render as a very thin line, this would make it much harder to pick. Feels like they should be completely separate.

extras/gizmo/gizmo-transform.js
kpal81xd: Yes I am planning to remove it because of local rotation lighting issues. I could add it back but I feel like that that is extra and something the user add themselves when customising

extras/gizmo/gizmo-transform.js
kpal81xd: The controls PCUI element for the color picker returns an array rather than `Color` hence why it is configured in this way.

kpal81xd:
> Something else for the TODO list: JSDocs, at least for any API you consider public?

Yes of course I am in the process of adding those.

kpal81xd:
> I have a question: are the intersection shapes for selection/picking independent of the graphical configuration of the gizmo? From a quick scan of the code, they look the same. I would imagine that if a gizmo axis was set to render as a very thin line, this would make it much harder to pick. Feels like they should be completely separate.

Yes right now they are the same but ideally you would want to have a separate entity representing the bounds for selection.

extras/gizmo/axis-shapes.js
```diff
@@ -0,0 +1,407 @@
...
+            material: this._defaultColor,
+            castShadows: false
+        });
+        this._updateArrow();
+        this.entity.addChild(this._arrow);
+        this.meshInstances.push(...this._arrow.render.meshInstances);
+    }
+
+    _updateLine() {
+        this._line.setLocalPosition(new Vec3(0, this._gap + this._lineLength * 0.5, 0));
```
marklundin, line 145: Just incase this get's called frequently, is it worth avoiding the `new Vec3()`'s if possible

extras/gizmo/axis-shapes.js
```diff
@@ -0,0 +1,407 @@
...
+    constructor(options = {}) {
+        super(options);
+
+        this._device = options.device;
+        this._ringRadius = options.ringRadius ?? this._ringRadius;
+        this._createDisk(options.layers ?? []);
+    }
+
+    _createDisk(layers) {
+        const mesh = createTorus(this._device, {
```
marklundin, line 298: Can we control the number of segments here either directly, or infer based on the size? It's more noticeable when the scale is large.

image

extras/gizmo/axis-shapes.js
kpal81xd: Ahh yes I must have missed that one thanks!

extras/gizmo/axis-shapes.js
kpal81xd: I will check the createTorus method but I should be able to

extras/gizmo/gizmo.js
```diff
@@ -0,0 +1,351 @@
...
+        }
+        scale = Math.max(scale * this._size, MIN_GIZMO_SCALE);
+        tmpV1.set(scale, scale, scale);
+        this.gizmo.setLocalScale(tmpV1);
+
+        this.fire('scale:set', tmpV1);
+
+    }
+
+    _getSelection(x, y) {
```
mvaligursky, line 233: This feels rather expensive. Maybe the gizmo should not use entities / render components at all, and only create meshInstances and insert those to the layer directly.
We should also not call getPositions / getIndices each time, as those get extracted from storage in the vertex buffer. We should keen our collision data stored directly in a format that is convenient for ray to triangle intersection. I'd add a Triangle class to the engine (constructed from 3 points). And it would have intersect function with the ray. Then this code would store the array of triangles directly.

extras/gizmo/gizmo.js
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    _updateRotation() {
+        if (this._coordSpace === 'local') {
+            tmpV1.set(0, 0, 0);
+            for (let i = 0; i < this.nodes.length; i++) {
+                const node = this.nodes[i];
+                tmpV1.add(node.getEulerAngles());
+            }
+            tmpV1.scale(1.0 / this.nodes.length);
```
mvaligursky, line 208: possible divide by zero here too

extras/gizmo/gizmo.js
willeastcott: I agree that the engine would benefit from a `Triangle` class that can handle a `Ray` instersection. But I kinda like the idea that this initial implementation is self-contained. Could this be a follow up PR?

mvaligursky:
I'd be great to add an orbit camera script to the example, as I always find myself look around to see what changes in the scene. It would also validate that gizmo can work together with other system accepting input.

mvaligursky:
I also currently get these, I assume it's WIP:
Screenshot 2024-01-08 at 16 21 43

mvaligursky:
When I change the camera FOV, the gizmo size changes. Should it be consistent size on the screen?

kpal81xd:
> When I change the camera FOV, the gizmo size changes. Should it be consistent size on the screen?

That was because of the broken `updateScale` error. Im fixing that now

kpal81xd:
> I also currently get these, I assume it's WIP: Screenshot 2024-01-08 at 16 21 43

This has also been fixed now.

extras/gizmo/gizmo-transform.js
```diff
@@ -0,0 +1,476 @@
...
+     *
+     * @type {number}
+     */
+    snapIncrement = 1;
+
+    /**
+     * Creates a new GizmoTransform object.
+     *
+     * @param {import('playcanvas').AppBase} app - The application instance.
+     * @param {Entity} camera - The camera entity.
```
mvaligursky, line 182: camera passed in should be of type CameraComponent and not Entity.

extras/gizmo/tri.js
```diff
@@ -0,0 +1,74 @@
...
+
+class Tri {
+    v0;
+
+    v1;
+
+    v2;
+
+    constructor(v0 = new Vec3(), v1 = new Vec3(), v2 = new Vec3()) {
+        if (v0.length === 3) {
```
LeXXik, line 23: Vec3 has a `length` method, so this will always be true.

extras/gizmo/tri.js
```diff
@@ -0,0 +1,74 @@
...
+            this.v0 = v0[0];
+            this.v1 = v0[1];
+            this.v2 = v0[2];
+        } else {
+            this.v0 = v0;
```
mvaligursky, line 28: we should not keep references to user allocated Vec3's but need to copy data to our own Vec3's. For example see how Plane class does it in the engine.

extras/gizmo/gizmo.js
```diff
@@ -0,0 +1,330 @@
...
+     * @example
+     * const gizmo = new pcx.Gizmo();
+     * gizmo.attach([boxA, boxB]);
+     */
+    attach(nodes = []) {
+        this.nodes = nodes;
+        this._updatePosition();
+        this._updateRotation();
+
+        window.addEventListener('pointerdown', this._onPointerDown);
```
marklundin, line 296: It's worth adding a [check to see if](https://github.com/playcanvas/engine/blob/4ce20f98907ac0395a1fbeaeda21acdfc7f7875f/src/core/platform.js#L84) `window` exists, just in case this somehow gets pulled into a non browser environment. Edge case I know.

mvaligursky:
Looking at this image:
image

It makes me wonder if the gizmo should be rendered on top (or at least have an option to do so, if not already). I can imagine a situation where with some arrangement of objects (or even just a single large cube?) the gizmo would be completely invisible, as it would be inside of the object.

Our editor renders gizmos on top, like this:
Screenshot 2024-01-15 at 15 06 50

willeastcott:
@mvaligursky It _is_ on top. 😄 The screenshot is misleading. The rotate gizmo now no longer renders the back part, as if there was a fully transparent sphere writing to the depth buffer now.

willeastcott (approved):
What an achievement. Nice job, @kpal81xd! 🚀

extras/gizmo/tri.js
```diff
@@ -0,0 +1,73 @@
...
+const EPSILON = 1e-6;
+
+class Tri {
+    v0 = new Vec3();
+
+    v1 = new Vec3();
+
+    v2 = new Vec3();
+
+    constructor(v0 = new Vec3(), v1 = new Vec3(), v2 = new Vec3()) {
```
mvaligursky, line 21: I'd keep the constructor simple:
- do not create it from an array (you're not using it)
- do not create default `new Vec3` - those are unused when created from the array
- just expect 3 Vec3's being specified. Creating a triangle with three zero vertices is not very useful.

extras/gizmo/tri.js
```diff
@@ -0,0 +1,73 @@
...
+
+    set(v0, v1, v2) {
+        this.v0.copy(v0);
+        this.v1.copy(v1);
+        this.v2.copy(v2);
+
+        return this;
+    }
+
+    intersectRay(origin, dir, out = new Vec3(), epsilon = EPSILON) {
```
mvaligursky, line 33: I would not default `out` to `new Vec3` - the user cannot get the result in that case anyways. It can just be left undefined, and test for it when writing to it.

Alternatively, leave the signature as is, but return the Vec3 (with data) in case of intersection, and null if no intersection.

extras/gizmo/mesh-tri-data.js
```diff
@@ -0,0 +1,74 @@
...
+const tmpV1 = new Vec3();
+const tmpV2 = new Vec3();
+const tmpV3 = new Vec3();
+
+/**
+ * The class for holding mesh triangle data.
+ */
+class MeshTriData {
+    /**
+     * The local to parent transform matrix.
```
mvaligursky, line 20: what parent? Is this a hierarchy?
It's just a world transform for the triangles.

kpal81xd: So that matrix is for the transformation of the mesh relative to its parent specifically for an `AxisShape` consisting of more than one primitive mesh.

extras/gizmo/gizmo.js
```diff
@@ -0,0 +1,325 @@
...
+    _getProjFrustumWidth() {
+        const gizmoPos = this.gizmo.getPosition();
+        const cameraPos = this.camera.entity.getPosition();
+        const dist = tmpV1.copy(gizmoPos).sub(cameraPos).dot(this.camera.entity.forward);
+        return dist * Math.tan(this.camera.fov * math.DEG_TO_RAD / 2);
+    }
+
+    _createLayer() {
+        const layerMap = this.app.scene.layers.layerIdMap;
+        if (layerMap.has(GIZMO_LAYER_ID)) {
```
mvaligursky, line 186: Don't hardcode GIZMO_LAYER_ID. Just test by the name if the layer exists, and leave whatever default ID a new layer gets.

But additionally, I'd remove the layer creation from the gizmo class completely .. a layer should be passed in in the constructor, giving user the flexibility to add it to any layer they already have.

mvaligursky: it should also not modify layers on the camera .. the user should do this.

extras/gizmo/gizmo.js
```diff
@@ -0,0 +1,325 @@
...
+            }
+            this.fire('key:up');
+        };
+
+        app.on('update', () => this._updateScale());
+
+        app.on('destroy', () => this.detach());
+    }
+
+    set coordSpace(value) {
```
mvaligursky, line 159: in public API, we use exported constants instead of strings, otherwise I'm not sure what even the options here are apart from 'world'.

extras/gizmo/gizmo.js
```diff
@@ -0,0 +1,325 @@
...
+
+    /**
+     * Creates a new Gizmo object.
+     *
+     * @param {import('playcanvas').AppBase} app - The application instance.
+     * @param {import('playcanvas').CameraComponent} camera - The camera component.
+     * @example
+     * const gizmo = new pcx.Gizmo(app, camera);
+     */
+    constructor(app, camera) {
```
mvaligursky, line 104: do we need a `destroy` function too? That would remove the event handles from the window, remove the gizmo entity from the hierarchy root and similar.

mvaligursky: disconnect from app.on('update' ...

mvaligursky: we actually need to make sure all mesh instances get destroyed too, otherwise we leak GPU memory that is not garbage collected

extras/gizmo/axis-shapes.js
```diff
@@ -0,0 +1,650 @@
...
+    const sides = opts.sides ?? 20;
+    const sectorAngle = opts.sectorAngle ?? 2 * Math.PI;
+
+    // Variable declarations
+    const positions = [];
+    const normals = [];
+    const uvs = [];
+    const indices = [];
+
+    for (let i = 0; i <= sides; i++) {
```
mvaligursky, line 49: this seems like a copy of createTorus in the engine .. do we need a copy of it, or can we just call what we have? I have not compared, but ideally we don't duplicate code like this.

extras/gizmo/axis-shapes.js
```diff
@@ -0,0 +1,650 @@
...
+        normals: [],
+        uvs: [],
+        indices: [],
+        colors: []
+    };
+
+    mesh.getPositions(options.positions);
+    mesh.getNormals(options.normals);
+    mesh.getIndices(options.indices);
+    mesh.getUvs(0, options.uvs);
```
mvaligursky, line 106: the gizmo mesh is not textured, does it need UVs? How about normals .. does it need those?

extras/gizmo/axis-shapes.js
```diff
@@ -0,0 +1,650 @@
...
+        this._lineThickness = value ?? 1;
+        this._updateHead();
+        this._updateLine();
+    }
+
+    get lineThickness() {
+        return this._lineThickness;
+    }
+
+    set lineLength(value) {
```
mvaligursky, line 221: default value to 1, instead of doing `?? 1` on the next line for all these setters.

extras/gizmo/axis-shapes.js
```diff
@@ -0,0 +1,650 @@
...
+        this.entity.setLocalPosition(this._position);
+        this.entity.setLocalEulerAngles(this._rotation);
+        this.entity.setLocalScale(this._scale);
+
+        // head
+        this._head = new Entity('head:' + this.axis);
+        this.entity.addChild(this._head);
+        this._updateHead();
+        let mesh = createShadowMesh(this.device, this._head, 'cone');
+        let meshInstance = new MeshInstance(mesh, this._defaultColor);
```
mvaligursky, line 260: Reminder: please make sure these meshInstances are destroyed at some point (or entity owning those using render component is destroyed)

extras/gizmo/axis-shapes.js
```diff
@@ -0,0 +1,650 @@
...
+
+    get arrowLength() {
+        return this._arrowLength;
+    }
+
+    _createArrow() {
+        this.entity = new Entity('arrow:' + this.axis);
+        this.entity.setLocalPosition(this._position);
+        this.entity.setLocalEulerAngles(this._rotation);
+        this.entity.setLocalScale(this._scale);
```
mvaligursky, line 253: Could this be extracted to a function on a parent class? It's called multiple times.

extras/gizmo/axis-shapes.js
```diff
@@ -0,0 +1,650 @@
...
+
+        // box
+        this._box = new Entity('box:' + this.axis);
+        this.entity.addChild(this._box);
+        this._updateBox();
+        let mesh = createShadowMesh(this.device, this._box, 'box');
+        let meshInstance = new MeshInstance(mesh, this._defaultColor);
+        this._box.addComponent('render', {
```
mvaligursky, line 422: even this could be turned into a reuseable function to avoid repetition

extras/gizmo/axis-shapes.js
kpal81xd: It is almost the same but this version allows for the sector size to be set. I was hoping this would replace the version inside the engine.

extras/gizmo/axis-shapes.js
willeastcott: I don't think setters support default parameters.

extras/gizmo/axis-shapes.js
kpal81xd: The UVs are not needed but the normals are required for shadow calculation

extras/gizmo/axis-shapes.js
mvaligursky: feel free to just add the extra parameters you need to the one in the engine

extras/gizmo/axis-shapes.js
mvaligursky: ah ok

examples/src/examples/misc/gizmos.mjs
```diff
@@ -57,7 +57,7 @@ function controls({ observer, ReactPCUI, React, jsx, fragment }) {
                     jsx(BooleanInput, {
                         type: 'toggle',
                         binding: new BindingTwoWay(),
-                        link: { observer, path: 'gizmo.legacyRotation' }
+                        link: { observer, path: 'gizmo.useLegacyRotation' }
```
mvaligursky, line 60: I think it'd be better to have a more descriptive name then legacy - this does not help people decide which one they want to use.

kpal81xd: Yea Im just trying to think of a concise way to describe the difference.

extras/gizmo/axis-shapes.js
kpal81xd: I was planning of leaving that for a follow up PR since this PR has no core engine changes

mvaligursky (approved):
Approving, great effort, this works really well!

PR #5941, opened 2024-01-15
Updated dithered transparency to support blue noise

mvaligursky:
Update to transparency dithering introduced here: #5903, which is not released yet and so API changes are fine.

New API (booleans were changed to enums to allow different options)
```
StandardMaterial.opacityDither = DITHER_BAYER8 / DITHER_BLUENOISE / DITHER_NONE;
StandardMaterial.opacityShadowDither = DITHER_BAYER8 / DITHER_BLUENOISE / DITHER_NONE;
```

- added internal data storing 32x32 4 channel blue noise, which will be used for multiple effects together with TAA
- the data gets also uploaded as a texture (initially used for blue noise dithering only)
- when the camera jitter is enabled (TAA), both bayer and blue noise pattern are randomly per frame offset by a blue noise based offset

Existing bayer dithering:
Screenshot 2024-01-15 at 13 50 21

Blue noise based (better with TAA):
Screenshot 2024-01-15 at 13 50 37

Related: https://github.com/playcanvas/engine/issues/5943


src/core/math/blue-noise.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    value() {
+        this.next();
+        return data[this.seed] / 255;
+    }
+
+    vec4(dest = new Vec4()) {
+        this.next();
+        return dest.set(data[this.seed], data[this.seed + 1], data[this.seed + 2], data[this.seed + 3]).mulScalar(1 / 255);
```
slimbuck, line 46: This will error if `this.seed` is currently `data.length - 1 or 2 or 3`. Either next must modulo `data.length - 4` or these accesses require %.

src/core/math/blue-noise.js
```diff
@@ -0,0 +1,50 @@
...
+    next() {
+        this.seed = (this.seed + 4) % data.length;
+    }
+
+    value() {
+        this.next();
+        return data[this.seed] / 255;
+    }
+
+    vec4(dest = new Vec4()) {
```
slimbuck, line 44: This api seems weird. Why expose `next` if `value` and `vec4` both call it? Shouldn't vec4 be called `vec4Value` or similar?

src/core/math/blue-noise.js
```diff
@@ -0,0 +1,50 @@
...
+class BlueNoise {
+    seed = 0;
+
+    constructor(seed = 0) {
+        this.seed = seed * 4;
+        initData();
+    }
+
+    next() {
+        this.seed = (this.seed + 4) % data.length;
```
slimbuck, line 36: Strange that next assumes size 4. Perhaps 4 should be a function parameter.

mvaligursky: The data is RGBA texture format, so basically a 4 grayscale blue noise arrays, stored in the interleaved format.
To get a single stream of blue noise .. we need to sample each 4th value from the array, so the 4 is not something that should be exposed as a parameter - this is just a function of the data available.

src/core/math/blue-noise.js
mvaligursky: next is meant as a private function - it moves to a next pixel (so 4 indices). Only value and vec4 are public (we can add vec2 and vec3 later too)

src/core/math/blue-noise.js
mvaligursky: this.seed is always pointing to a start value of the 4 (RGBA), so there are always 3 following values available

src/core/math/blue-noise.js
slimbuck: OIC you're stepping through the 32x32 image and returning values in sequence. Is that still blue noise?

slimbuck: This is the best quick answer I could find:

Screenshot 2024-01-15 at 15 18 48

examples/src/examples/graphics/dithered-transparency.mjs
```diff
@@ -41,7 +49,7 @@ function controls({ observer, ReactPCUI, React, jsx, fragment }) {
  * @param {import('../../options.mjs').ExampleOptions} options - The example options.
  * @returns {Promise<pc.AppBase>} The example application.
  */
-async function example({ canvas, deviceType, assetPath, glslangPath, twgslPath, scriptsPath, data }) {
+async function example({ canvas, deviceType, assetPath, glslangPath, twgslPath, scriptsPath, data, pcx }) {
```
kpal81xd, line 52: `pcx` is unused.

src/core/math/blue-noise.js
mvaligursky: It's likely there are better 1D blue noise sequences we could store as some larger arrays, but this is close enough to give us a better random generator compared to just random values - the important attributes are the lack of clustering and even distribution, without storing additional data.

examples/src/examples/graphics/dithered-transparency.mjs
mvaligursky: removing, thanks

slimbuck (approved):

PR #5942, opened 2024-01-15
Remove type global function

slimbuck:
I noticed the following unused core function being included in an optimised, tree-shaken build of an engine app:

Screenshot 2024-01-15 at 13 36 39

This PR changes the implementation so the function will be removed correctly.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -27,26 +27,25 @@ const data = { }; // Storage for exported entity data
  * @returns {string} The type string: "null", "undefined", "number", "string", "boolean", "array", "object", "function", "date", "regexp" or "float32array".
  * @ignore
  */
-const type = (() => {
-    const _typeLookup = { };
-    ['Array', 'Object', 'Function', 'Date', 'RegExp', 'Float32Array'].forEach((name) => {
-        _typeLookup[`[object ${name}]`] = name.toLowerCase();
-    });
+function type (obj) {
+    if (obj === null) {
+        return 'null';
+    }
 
-    return (obj) => {
-        if (obj === null) {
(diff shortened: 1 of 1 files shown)


src/core/core.js
```diff
@@ -48,13 +32,19 @@ function type(obj) {
         return 'null';
     }
 
-    const type = typeof obj;
-
-    if (type === 'undefined' || type === 'number' || type === 'string' || type === 'boolean') {
-        return type;
+    const typeString = typeof obj;
+    if (['undefined', 'number', 'string', 'boolean'].includes(typeString)) {
```
mvaligursky, line 36: this array should not be constructed each time the function runs but be global

src/core/core.js
```diff
@@ -48,13 +32,19 @@ function type(obj) {
...
+        return typeString;
     }
 
-    return _typeLookup[Object.prototype.toString.call(obj)];
+    return {
```
mvaligursky, line 40: and the same goes for this object, no need to construct it each time

src/core/core.js
slimbuck: I tested your theory because I don't believe static arrays or objects are constructed each time the function is called. (Just like a number 52.3 or "string" appearing in a function isn't "constructed" each time). Only a stupid JS implementation would do that...

In Chrome this is true: no extra cpu or memory overhead using global or locally declared static array or object.

However in safari the locally declared version runs 25% slower.

So I'll move it.

mvaligursky: Interesting, good to know!

mvaligursky (approved):

PR #5944, opened 2024-01-16
[Experimental] Dithered transparency using Spatiotemporal Blue Noise

liamdon:
_This is an experimental draft, not yet a merge proposal_

I was excited to see the recent blue noise PR (#5941), great stuff.

That PR uses noise published on [this blog](https://momentsingraphics.de/BlueNoise.html), and on that blog there is also a discussion about [3D blue noise](https://momentsingraphics.de/3DBlueNoise.html) and why it doesn't work well for stable temporal transparency dithering.

However, there have been some recent advances in this area - NVIDIA published a paper on "[Spatiotemporal Blue Noise Masks](https://research.nvidia.com/publication/2022-07_spatiotemporal-blue-noise-masks)" and included some [C++ code](https://github.com/NVIDIAGameWorks/SpatiotemporalBlueNoiseSDK) to generate them. This is a type of 3D blue noise that has the nice property of being visually stable if you step through the 3rd dimension over time.

Using this special type of noise, we can create dithered transparency without requiring TAA or camera jittering - simply step through the Z axis of the noise volume/array every frame and then sample from that slice in the normal way.

At a high enough framerate, the results are really good for transparency. But they are impossible to capture in either a screenshot (not temporal) or in a screen recording (compression artifacts ruin it). I suppose this could also be a downside of the technique - your game won't look good while streaming 😉 

Anyway, thought you might want to take a look at [my branch](https://github.com/liamdon/engine/tree/spatioTemporalBlueNoise-arr) to see it in action on your machine - it could be a direction that's worth pursuing if we want TAA-independent dithered transparency. 

If helpful, I can also share my hacky fork of the [SpatiotemporalBlueNoiseSDK](https://github.com/NVIDIAGameWorks/SpatiotemporalBlueNoiseSDK) repo - it was rather painful to compile on macOS, and I added some additional python scripts to wrangle the output into an array of base64 slices.

------------------------


mvaligursky:
It looks great, but the amount of data this adds to the engine makes me not very keen on this.
I wonder if we could try using the 4 channels we have a loop over those each frame. Not as good as 32, but could be an improvement.
Alternatively, we could allow the users set the data from outside, but that's not very convenient to use.

liamdon:
Feel free to close this PR anytime you want, I just wanted to put this on your radar as a potential option as you explore transparency/TAA.

mvaligursky:
> I'll try out the "4 slices" version - it's best to generate a new 32x32x4 rather than simply chopping our current array, as the algorithm will ensure it "loops" smoothly. This would still be around 5.3kb of base64 characters, though.

That's the size we currently have, I'm ok with this. At the moment its 32x32 RGBA. So feel free to just update data to new 4 layers.

mvaligursky:
Closing but great experiment!

Issue #5692, opened 2023-09-28
Add WebGPU support to lightmapper


mvaligursky:
Most likely the lightmapper should be converted to render passes.

mvaligursky:
Note that this was partially done in #5953 but is partially disabled by #6016

PR #5953, opened 2024-01-19
Partial support for Lightmapper on WebGPU

mvaligursky:
partial solution to https://github.com/playcanvas/engine/issues/5692

- basic baking works, including smoothness filter
- all light types are supported, including ambient cubemap
- Limitation: local light shadows are not supported, so those are skipped (with a warning)
- directional light shadows work

Screenshot 2024-01-19 at 10 52 21


src/framework/lightmapper/lightmapper.js
```diff
@@ -854,12 +850,24 @@ class Lightmapper {
...
+                    Debug.warnOnce('Lightmapper on WebGPU does not yet support spot and omni shadows.');
+                    return true;
+                }
+
                 this.renderer._shadowRendererLocal.cull(light, comp, casters);
-            }
 
-            const insideRenderPass = false;
-            this.renderer.shadowRenderer.render(light, this.camera, insideRenderPass);
+                // TODO: this needs to used render passes to work on WebGPU
```
willeastcott, line 867:
```suggestion
                // TODO: this needs to use render passes to work on WebGPU
```

willeastcott (approved):

src/scene/shader-lib/chunks/common/vert/transform.js
```diff
@@ -115,6 +115,9 @@ vec4 getPosition() {
     vec4 screenPos;
     #ifdef UV1LAYOUT
     screenPos = vec4(vertex_texCoord1.xy * 2.0 - 1.0, 0.5, 1);
+        #ifdef UV1LAYOUT_FLIP
+        screenPos.y *= -1.0;
+        #endif
```
erikdubbelboer, line 120: I'm curious, why this way while in other places we just use [`#ifdef WEBGPU`](https://github.com/search?q=repo%3Aplaycanvas%2Fengine+%22%23ifdef+WEBGPU%22&type=code) instead?

mvaligursky: good point, I'll change it to that too, thanks

PR #5954, opened 2024-01-19
JSDocs for extras/exporters

mvaligursky:
- only constructors and build functions are exposed.

### DIFF
--- a/extras/exporters/core-exporter.js
+++ b/extras/exporters/core-exporter.js
@@ -20,11 +20,17 @@ const textureBlitFragmentShader = `
 
 class CoreExporter {
     /**
+     * Create a new instance of the exporter.
+     */
+    // eslint-disable-next-line no-useless-constructor
+    constructor() { }
+
+    /**
      * Converts a texture to a canvas.
      *
      * @param {Texture} texture - The source texture to be converted.
      * @param {object} options - Object for passing optional arguments.
-     * @param {Color} [options.color] - The tint color to modify the texture with.
+     * @param {import('playcanvas').Color} [options.color] - The tint color to modify the texture with.
      * @param {number} [options.maxTextureSize] - Maximum texture size. Texture is resized if over the size.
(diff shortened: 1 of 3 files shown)


extras/exporters/gltf-exporter.js
```diff
@@ -120,7 +120,19 @@ const textureSemantics = [
...
 
+/**
+ * Implementation of the GLTF 2.0 format exporter.
+ */
 class GltfExporter extends CoreExporter {
+    /**
+     * Creates an instance of the GltfExporter.
+     * @function
+     */
+    // No actual constructor code here, just JSDoc.
```
willeastcott, line 131: TBH, I think this is unnecessary (I'm not even convinced this is doing to the right thing anyway). If the superclass' constructor has something like `Create a new instance of the exporter.`, I think that's sufficient to be shown in the subclass too.

extras/exporters/usdz-exporter.js
```diff
@@ -103,9 +112,16 @@ class UsdzExporter extends CoreExporter {
...
      * @type {object}
+     * @ignore
      */
     files;
 
+    /**
+     * Creates an instance of the UsdzExporter.
+     * @function
+     */
+    // No actual constructor code here, just JSDoc.
```
willeastcott, line 123: Same, I wouldn't include this.

willeastcott (approved):

extras/exporters/core-exporter.js
```diff
@@ -19,12 +19,18 @@ const textureBlitFragmentShader = `
     }`;
 
 class CoreExporter {
+    /**
+     * Create a new instance of the exporter.
+     */
+    // eslint-disable-next-line no-useless-constructor
+    constructor() { }
```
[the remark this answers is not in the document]
willeastcott: Should really use `eslint-disable-line`.

PR #5957, opened 2024-01-19
Remove VertexArrayObject polyfill for IE11

mvaligursky:
We no longer support IE11, and so this extension can go, as the VAO functionality is otherwise well covered by OES_vertex_array_object extension on WebGL1:

Screenshot 2024-01-19 at 15 39 21

on https://developers.poki.com/player-device-report/ support for this extension is 5%, which matches 5% of devices support only WebGL1:

Screenshot 2024-01-19 at 15 40 12


marklundin:
The ES5 engine build does seem to [target IE11](https://github.com/playcanvas/engine/blob/main/utils/rollup-es5-options.mjs#L21). Is that an issue?

mvaligursky:
related: https://github.com/playcanvas/developer.playcanvas.com/pull/554

mvaligursky:
We'll need to fix the build to remove IE11 support as well.

willeastcott (approved):

PR #5963, opened 2024-01-22
ESM ScriptType loading

marklundin:
This PR adds support for es modules using the existing ScriptType system. It conditionally loads scripts as a module depending on the file name (ending in`.mjs`). This can be used in the editor [as-is](https://playcanvas.com/editor/scene/1927088) although there is work necessary in the editor to parse Scripts as modules.

#### Issues

1. Scripts cannot define attributes as getter/setters. https://github.com/playcanvas/engine/pull/5910#issuecomment-1903888527
2. Script class names can be mangled during a build, which would `registerScript` https://github.com/playcanvas/engine/pull/5910#discussion_r1453353618


src/framework/handlers/script.js
```diff
@@ -97,28 +100,65 @@ class ScriptHandler {
...
-                done = true; // prevent double event firing
-                callback(null, url, element);
-            }
-        };
-        // set the src attribute after the onload callback is set, to avoid an instant loading failing to fire the callback
-        element.src = url;
 
-        head.appendChild(element);
+        const path = url.split('?')[0];
+        const isEsmScript = path.endsWith('.esm.js');
```
slimbuck, line 105: wondering why not `.mjs`?

marklundin: Open to suggestion, however one thing the V8 team and Mozilla pointed out is that '.js' will get served with the correct Content-Type headers that would already exist on the server. 'mjs', 'jsm' etc may not have those configured and would need to be manually setup . https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#aside_%E2%80%94_.mjs_versus_.js

Maksims: Although build systems, today do support mjs pretty well, e.g. Babel.
We do have page that does require setting up server already: https://developer.playcanvas.com/en/user-manual/publishing/web/self-hosting-for-beginners/#content-types
It worth adding there configs too.

src/framework/handlers/script.js
```diff
@@ -97,28 +100,65 @@ class ScriptHandler {
...
+
+        if (isEsmScript) {
+
+            const baseUrl = platform.browser ? window.location.origin : '';
+            const importUrl = new URL(url, baseUrl);
+            importUrl.searchParams.append('cacheBust', new Date().valueOf().toString());
```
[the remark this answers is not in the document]
marklundin: Ahh the `file.hash` changes on asset save?

src/framework/handlers/script.js
marklundin: That's true. I'm not particularly wedded to either. Currently the editor doesn't let you change the file suffix, so that would need to be changed, and I'm not sure what it reall buys you. The nice thing about keeping it '.js' is that the user can just change directly as-is. Also one other thing, there's gonna be a distinction between ScriptType modules, and just plain vanilla modules (config/utils/etc). Arguably these shouldn't __only__ work if they're '.mjs'. Also potentially, if we enforced the ESM ScriptType at the file name level then we can guard against it being added as a script in the editor, rather than having to load it and check through exports.

mvaligursky:
Personally, I'd like to see an example of this, perhaps with limitations in comments do demonstrate what works and what does not. It does not need to be an engine example (but could be), but at least one in the PR description, so that we know what this enables / what needs to be avoided, for this to work in engine only / editor environment.

marklundin:
@mvaligursky There isn't any changes in the engine use, it's only where files get loaded in the Script Handler. Here's a Editor example, which will work if you run the launcher or edit code. Although don't parse the script in the editor as this needs an editor update 

https://playcanvas.com/editor/scene/1927088

mvaligursky:
Great example. So how do we let people know they can use it?

mvaligursky:
> The example doesn't work for me or do I need to run it against this PR?

Yes you would need that.

src/framework/handlers/script.js
```diff
@@ -120,6 +145,38 @@ class ScriptHandler {
...
     }
+
+    _loadModule(url, callback) {
+
+        // if we're in the browser, we need to use the full URL
+        const baseUrl = platform.browser ? window.location.origin : import.meta.url;
+        const importUrl = new URL(url, baseUrl);
+
+        // @ts-ignore
+        import(importUrl.toString()).then((module) => {
```
mvaligursky, line 156: So this uses import, which based on https://caniuse.com/?search=import is supported by Chrome from 2017, Safari 2018, Firefox 2019 - I wonder how we could handle the compatibility issues here.

src/framework/handlers/script.js
```diff
@@ -120,6 +145,38 @@ class ScriptHandler {
 
         head.appendChild(element);
     }
+
+    _loadModule(url, callback) {
+
+        // if we're in the browser, we need to use the full URL
+        const baseUrl = platform.browser ? window.location.origin : import.meta.url;
```
slimbuck, line 152: Can an es5 version of the engine load module scripts? If so then `import.meta.url` would fail, right?

slimbuck: (Even though that'll only ever be used in node).

marklundin: The target is [marginally higher](https://github.com/playcanvas/engine/blob/020e5f3b762e48cf6759d8430e85dabf36cf9d8b/utils/rollup-es5-options.mjs#L22-L24) than ES5 now, but babel should transpile this according to the preset.env. See [here](https://babeljs.io/docs/babel-plugin-syntax-import-meta)

src/framework/handlers/script.js
marklundin: There has to be an assumption that if a user chooses to use ES modules, then there is a neccessary baseline level of support required. [ES6 is around 97.2%](https://caniuse.com/es6-module) which is only [0.7% less than our legacy build](https://browsersl.ist/#q=%3E+0.1%25%2C+partially+supports+webgl%2C+not+dead).

If someone uses the old script mechanism, the ESM path won't get triggered so it will just work as is.

src/framework/handlers/script.js
mvaligursky: That sounds good, just double check it does and we're not missing some settings please.

marklundin:
> Will the script that was extended from ScriptType be parsed by Editor as well?

How do you mean @LeXXik? Do you mean subclassing something that extends ScriptType?

```
class CustomBase extends ScriptType { static attributeDefinitions: { ... } }
class MyClass extends CustomBase {}
```

This should work and pick up the correct attributes if you attach MyClass to a Script Component, ie, it will 'inherit' the CustomBase attributes. Although, for the moment you can't import a relative module from the asset registry. Not until we address the pathing issues.

Maksims:
> Will the script that was extended from ScriptType be parsed by Editor as well?

I think you just mean if current Editor workflow will work pretty much the same with a ESM script? It will.
```javascript
class Test extends pc.ScriptType { }
pc.registerScript(Test, 'test');
```
This already works in Editor.

With ESM script no need to call `pc.registerScript` (you still should be able to though), instead you would export, like so:
```javascript
export Test;
```

And asset loader will automatically add it to the script registry.

marklundin:
Yep, all exports from a module will be tested to see if they are a valid ScriptType, so the following are both valid Scripts

```javascript
class Test extends ScriptType {};
class OtherTest extends ScriptType {};

export Test;
export default OtherTest;
```

LeXXik:
Right, I was referring to the child class, as in
```js
class A extends pc.ScriptType {}
class B extends A {} // can't parse B in Editor at the moment
```

Maksims:
> Right, I was referring to the child class, as in
> 
> ```js
> class A extends pc.ScriptType {}
> class B extends A {} // can't parse B in Editor at the moment
> ```

That would be the Editor. Technically it could already be improved to load other scripts. With ESM Modules and import maps - this becomes easier. You would need to do this:

/scripts/scriptA.js
```javascript
class A extends pc.ScriptType {}
export A;
```

/scripts/scriptB.js
```javascript
import A from '/scripts/scriptA.js';
class B extends A {}
export B;
```

Maksims (approved):
LGTM!

slimbuck (approved):
🎉

marklundin:
Quietly pretty excited about this one!

PR #5968, opened 2024-01-23
Move splat functionality from extras to core engine

mvaligursky:


src/scene/gsplat/gsplat-data.js
```diff
@@ -72,17 +70,17 @@ const calcSplatMat = (result, data) => {
     ]);
 };
 
-class SplatData {
-    /** @type {import('./ply-reader').PlyElement[]} */
+class GSplatData {
+    // /** @type {import('./ply-reader').PlyElement[]} */
```
slimbuck, line 74: remove // on this line and below

slimbuck (approved):

src/scene/gsplat/gsplat-data.js
mvaligursky: will do in a follow up PR, thanks

PR #5971, opened 2024-01-24
Add missing XR exports

### LINKED ISSUE #5966: Missing TypeDoc classes
A number of engine classes including XR ones such as [XRHand](https://github.com/playcanvas/engine/blob/main/src/framework/xr/xr-hand.js) and [XRFinger](https://github.com/playcanvas/engine/blob/main/src/framework/xr/xr-finger.js) are not being picked up by typedoc and therefore no documentation is available for them. This seems to be because they're only used as [`@private`](https://github.com/playcanvas/engine/blob/2c9437f6fd2a07e36457db07fb89990660ad7fd8/src/framework/xr/xr-joint.js#L39) members of other classes and presumably typedoc skips them as non public api, and they're not re-exported by index.js.

The current developer api documentation however [does include](https://developer.playcanvas.com/en/api/pc.XrHand.html) [these classes](https://developer.playcanvas.com/en/api/pc.XrFinger.html), and there are inbound links in the developer site.

Options are we could force typedoc to include these, either by exporting them, or simply just adding them to the list of typedoc entry points. Or we can keep as-is and just remove any references to them. Any thoughts here @mvaligursky @Maksims ?

marklundin:
Re-exports XRFinger and XrHand in the index.js output. This also includes them in Typedoc output

Fixes #5966 


src/index.js
```diff
@@ -364,6 +364,8 @@ export { XrAnchors } from './framework/xr/xr-anchors.js';
 export { XrAnchor } from './framework/xr/xr-anchor.js';
 export { XrPlaneDetection } from './framework/xr/xr-plane-detection.js';
 export { XrPlane } from './framework/xr/xr-plane.js';
+export { XrHand } from './framework/xr/xr-hand.js';
```
mvaligursky, line 367: please sort those alphabetically.

mvaligursky (approved):

src/index.js
marklundin: Sorry, missed this. See https://github.com/playcanvas/engine/pull/5972

PR #5980, opened 2024-01-25
Adding GSplatComponentSystem to the Application by default

mvaligursky:
automatically support gsplats


src/framework/application.js
```diff
@@ -181,7 +182,8 @@ class Application extends AppBase {
             SpriteComponentSystem,
             LayoutGroupComponentSystem,
             LayoutChildComponentSystem,
-            ZoneComponentSystem
+            ZoneComponentSystem,
+            GSplatComponentSystem
```
willeastcott, line 186: This is one of those lists where it looks random but there are probably some weird reasons for the ordering.

willeastcott (approved):

src/framework/application.js
mvaligursky: yeah it used to be sorted this way and it's possible there is some internal dependency between some of them, so I just preserved it for now.

PR #5986, opened 2024-01-25
Expose AppOptions and correct its types

mvaligursky:
- based on https://github.com/playcanvas/engine/pull/5882
- exposes AppOptions class as an public API
- fixed up some types in it (now that we can), which allows us to remove many `ts-ignore` tags from the examples


mvaligursky:
What am I missing here @kungfooman ?

`Error: build/playcanvas.d.ts(22879,30): error TS2693: 'ResourceHandler' only refers to a type, but is being used as a value here.`

kungfooman:
@mvaligursky You can't do `typeof` on an interface, that's why I tried to push it here a bit: https://github.com/playcanvas/engine/pull/5882#issuecomment-1845080088

In short, we have to come to a solution here: https://github.com/playcanvas/engine/pull/5574 (looks kinda minor, but no response since November last year)

mvaligursky:
I suspected this :(
Well, we're switching from jsdocs to tsdocs just now .. so that issue might go away I hope.

mvaligursky:
Actually, that won't help. We need the fix / refactor you mentioned.

willeastcott (approved):

PR #5997, opened 2024-01-26
Simple polish to the IntegerTextures example

mvaligursky:
- removed ground / light / shadow
- added skydome instead
- darkened the background to make the sand more visible

Before:
Screenshot 2024-01-26 at 14 15 04

Now:
Screenshot 2024-01-26 at 14 30 32


willeastcott (approved):

liamdon:
Looks much better, and also more on-brand too with the dark background 🙂

mvaligursky:
This was the easy bit after @liamdon did all the hard work here :)

PR #5999, opened 2024-01-26
Gizmo uniform scaling update

kpal81xd:
Implemented new calculation logic for uniform scaling.

### DIFF
--- a/extras/gizmo/transform-gizmo.js
+++ b/extras/gizmo/transform-gizmo.js
@@ -282,6 +282,14 @@ class TransformGizmo extends Gizmo {
     _isRotation = false;
 
     /**
+     * Internal state if transform should use uniform scaling.
+     *
+     * @type {boolean}
+     * @protected
+     */
+    _useUniformScaling = false;
+
+    /**
      * Internal state for if the gizmo is being dragged.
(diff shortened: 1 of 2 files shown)


slimbuck (approved):

extras/gizmo/transform-gizmo.js
```diff
@@ -537,7 +546,27 @@ class TransformGizmo extends Gizmo {
...
         }
 
-        if (isAllAxes) {
+        if (isUniform) {
+            // calculate point distance from gizmo
+            tmpV1.copy(point).sub(gizmoPos).normalize();
+
+            switch (axis) {
+                case 'x':
+                    tmpV2.copy(this.root.up).add(this.root.forward.mulScalar(-1)).normalize();
```
willeastcott, line 555: This is not a massive deal, but here, you're writing into the internal `Vec3` stored on the `GraphNode` and negating it. This could potentially cause some subtle bugs in app code where somebody might query the `forward` vector and then use it slightly later after you've written to it here. Now, in practice, you shouldn't just save references to the forward/up/right vectors but query them when you need them. But I think it's safer not to write to it.

PR #6001, opened 2024-01-27
Implement changes to WebGPU timestamp-query

### LINKED ISSUE #5989: Implement changes to WebGPU `timestamp-query`
The spec has changed and the current functionality is currently no longer available. This needs to be looked at and addressed as some point. This was introduced in Chrome 121.

Comment from Chrome implementators: ` The spec was recently changed because we found that writeTimestamp wouldn't work (or wouldn't work consistently) across backends, but pass timing could. It may be added back later.`

Related issue: #5973 
WebGPU spec change: https://github.com/gpuweb/gpuweb/pull/2190

erikdubbelboer:
Remove call to `GPURenderPassEncoder.writeTimestamp` as it is not part of the spec anymore, and was removed by Chrome. It is also not needed anymore.

The new interface, `GPURenderPassDescriptor.timestampWrites` was already implemented so nothing is needed there.

Fixes #5989


mvaligursky (approved):
Great!
Hopefully this will come back one day.

PR #6003, opened 2024-01-27
Don't upload texture if locked using TEXTURELOCK_READ

liamdon:
The API docs suggest that you can call `texture.lock()` with a `mode` of either `TEXTURELOCK_READ` or `TEXTURELOCK_WRITE`, but in this option doesn't do anything.

This is a small change so that a lock that's opened with `TEXTURELOCK_READ` will not call `texture.upload()` when unlocked. It also uses the mode to check the state, instead of overloading `texture._lockedLevel` for this purpose.

I also have some related questions about how we should handle mip updates via API - please see #6004

### DIFF
--- a/src/platform/graphics/texture.js
+++ b/src/platform/graphics/texture.js
@@ -790,10 +790,17 @@ class Texture {
         options.face ??= 0;
         options.mode ??= TEXTURELOCK_WRITE;
 
-        if (options.mode === TEXTURELOCK_NONE) {
-            Debug.log("pc.Texture#lock: To unlock a texture, call texture.unlock(). Defaulting to 'TEXTURELOCK_READ'.", this);
-            options.mode = TEXTURELOCK_READ;
-        }
+        Debug.assert(
+            this._lockedMode === TEXTURELOCK_NONE,
+            'The texture is already locked. Call `texture.unlock()` before attempting to lock again.',
+            this
+        );
+
+        Debug.assert(
+            options.mode === TEXTURELOCK_READ || options.mode === TEXTURELOCK_WRITE,
+            'Cannot lock a texture with TEXTURELOCK_NONE. To unlock a texture, call `texture.unlock()`.',
+            this
+        );
 
         this._lockedMode = options.mode;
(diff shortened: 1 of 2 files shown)


erikdubbelboer (changes requested):
You can completely remove `_lockedLevel` as it's unused after your changes.

slimbuck:
TBH the team had internally mumbled wanting to completely rewrite the texture API, but this is often a pipe-dream. Breaking backwards compatibility is very painful for us and our users.

I love this PR and the issue you created with explanation. It is well thought out and a clear improvement to the API - thanks so much!

src/platform/graphics/texture.js
```diff
@@ -773,6 +790,12 @@ class Texture {
         options.face ??= 0;
         options.mode ??= TEXTURELOCK_WRITE;
 
+        if (options.mode === TEXTURELOCK_NONE) {
+            Debug.log("pc.Texture#lock: To unlock a texture, call texture.unlock(). Defaulting to 'TEXTURELOCK_READ'.", this);
+            options.mode = TEXTURELOCK_READ;
+        }
```
mvaligursky, line 796: I'd just have
```
Debug.assert(options.mode === TEXTURELOCK_READ || options.mode === TEXTURELOCK_WRITE);
```
as nothing else is allowed.

liamdon:
> related to #4237

Ha, I actually forgot about this! I guess this API has been on my mind for a while 😄

willeastcott (approved):
Nice PR - approving (although agree with the feedback from @mvaligursky)

src/platform/graphics/texture.js
liamdon: Done!

mvaligursky (approved):
Nice, thanks!

liamdon:
Made some changes to assertions, now you can only have one lock in place at a time. If those changes look good, feel free to merge. 

I'll probably follow up (soon-ish) with a PR that implements my proposal in #6004; it may be easier to consider/discuss when you see a concrete implementation.

Issue #5925, opened 2024-01-01
Support WGSL shaders


erikdubbelboer:
PlayCanvas is currently using [glslang](https://github.com/KhronosGroup/glslang) and [twgsl](https://github.com/BabylonJS/twgsl) to transpile GLSL into WGSL. This doesn't support all new features in WGSL which are not usable now.

Things like
- `@group(0) @binding(0) var externalTexture : texture_external;` for fast [video texture support](https://github.com/playcanvas/engine/issues/5691).
- `@builtin(instance_index) instanceIdx : u32` for [hardware instancing](https://webgpu.github.io/webgpu-samples/samples/instancedCube#../../shaders/instanced.vert.wgsl).

The shader generator will need to be changed and all current shaders will need to be rewritten. It will need to be possible for developers to provide shader chunks for both GLSL and WGSL at the same time for backwards compatibility.

mvaligursky:
related: https://github.com/playcanvas/engine/issues/5691

mvaligursky:
Closing this as done, will be released with the next engine release.
Internally, all shaders use WGSL chunks, enabled in https://github.com/playcanvas/engine/pull/7665

PR #6006, opened 2024-01-28
Add WGSL example

erikdubbelboer:
Add first basic support for creating WGSL shaders in applications. Since WGSL shaders aren't parsed yet the bindgroup formats need to be specified manually.

Working towards #5925

Screenshot 2024-01-28 at 05 09 33

### DIFF
--- a/examples/src/examples/graphics/wgsl-shader.mjs
+++ b/examples/src/examples/graphics/wgsl-shader.mjs
@@ -101,7 +101,8 @@ async function example({ canvas, deviceType, files, glslangPath, twgslPath }) {
 
 class WgslShaderExample {
     static CATEGORY = 'Graphics';
-    static WEBGPU_ENABLED = 'force';
+    static WEBGPU_REQUIRED = true;
+    static HIDDEN = true;
     static FILES = {
         'shader.wgsl': `
             struct ub_mesh {
--- a/examples/src/examples/graphics/wgsl-shader.mjs
+++ b/examples/src/examples/graphics/wgsl-shader.mjs
@@ -1,0 +1,143 @@
+import * as pc from 'playcanvas';
+
+/**
+ * @param {import('../../options.mjs').ExampleOptions} options - The example options.
+ * @returns {Promise<pc.AppBase|null>} The example application.
+ */
+async function example({ canvas, deviceType, files, glslangPath, twgslPath }) {
+    const gfxOptions = {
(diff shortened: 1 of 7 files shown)


examples/scripts/generate-standalone-files.mjs
```diff
@@ -214,6 +217,9 @@ ${exampleClass.example.toString()}
                 return;
             }
             const deviceType = app?.graphicsDevice?.deviceType;
+            if (!deviceType) {
+                return;
+            }
```
erikdubbelboer, line 222: Don't try to create a `MiniStats` instance if we have no device.

examples/src/examples/graphics/wgsl-shader.mjs
```diff
@@ -0,0 +1,143 @@
...
+        // animate the amount as a sine wave varying from 0 to 1
+        material.setParameter('amount', (Math.sin(time * 4) + 1) * 0.5);
+    });
+    return app;
+}
+
+class WgslShaderExample {
+    static CATEGORY = 'Graphics';
+    static WEBGPU_ENABLED = 'force';
+    static FILES = {
```
mvaligursky, line 105: lets add
```
    static HIDDEN = true;
```
for now, as till the support is nicer, I would not people to start officially using it.

This will give us the example when `npm run develop` but not when `npm run build` which is used for publishing.

examples/src/examples/graphics/wgsl-shader.mjs
```diff
@@ -0,0 +1,143 @@
...
+        time += dt;
+        // animate the amount as a sine wave varying from 0 to 1
+        material.setParameter('amount', (Math.sin(time * 4) + 1) * 0.5);
+    });
+    return app;
+}
+
+class WgslShaderExample {
+    static CATEGORY = 'Graphics';
+    static WEBGPU_ENABLED = 'force';
```
mvaligursky, line 104: lets change `force` to `required` please

examples/src/examples/graphics/wgsl-shader.mjs
erikdubbelboer: Done. Added `WEBGPU_REQUIRED` and `HIDDEN`.

mvaligursky (approved):
Looks good to me. I'll have a go at making a small change, but happy to get this in first.

Issue #2974, opened 2021-03-17
Implement async read back of textures for WebGl2


mvaligursky:
This would avoid main thread being blocked waiting for GPU to finish rendering and transfer back the texture.
https://forum.babylonjs.com/t/speeding-up-readpixels/12739

Supported browsers:
https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/fenceSync

mvaligursky:
adding @slimbuck here as well as he was interested to work on this as well.

mvaligursky:
forum thread: https://forum.playcanvas.com/t/cant-read-from-a-texture-locked-with-texturelock-read-it-doesnt-seem-to-do-anything/26402/2

mvaligursky:
WebGL implementation: https://github.com/playcanvas/engine/pull/5332

mvaligursky:
webgpu example implementation: https://www.sobyte.net/post/2022-02/webgpu-buffer-mapping/#22-gpu-to-cpu

erikdubbelboer:
This is sort of implemented in https://github.com/playcanvas/engine/pull/6009, [here](https://github.com/playcanvas/engine/blob/0709cfad6d3f4100cd7746745a522cfdb2fef3ff/src/platform/graphics/webgpu/webgpu-compute.js#L60-L95) and [here](https://github.com/playcanvas/engine/blob/0709cfad6d3f4100cd7746745a522cfdb2fef3ff/src/platform/graphics/webgpu/webgpu-buffer.js#L95-L100).

mvaligursky:
mostly done here, but not yet public till more solid support is added
https://github.com/playcanvas/engine/pull/6370

willeastcott:
What's the status on this one @mvaligursky?

mvaligursky:
Still the same as my comment above. Not full API is functional.

mvaligursky:
This was now marked as public, closing, will be released with the next engine release as `Texure.read`

PR #6009, opened 2024-01-28
Compute shader example

erikdubbelboer:
Initial work on further implementing WebGPU Compute shaders and adding an example for it. This slightly changes the API to allow for more work to submitted in a simple pass and to allow for reading the storage texture for the result.

@mvaligursky could you have a look and let me know what you think?

Building on the work of https://github.com/playcanvas/engine/pull/5760

Kinda implements https://github.com/playcanvas/engine/issues/2974 for WebGPU.


src/platform/graphics/webgpu/webgpu-compute.js
```diff
@@ -39,7 +46,52 @@ class WebgpuCompute {
...
+    /**
+     * Get a buffer that contains the data of the specified texture.
+     * This needs to be called before dispatch! But can be called before device.startComputePass().
+     *
+     * @param {import('../texture.js').Texture} texture - The texture to get the buffer for.
+     * @returns {import('../buffer.js').Buffer} The buffer.
+     */
+    getBuffer(texture) {
+        // Calculate bytes per pixel, assuming RGBA8 format (4 bytes per pixel)
+        const bytesPerPixel = 4;
```
erikdubbelboer, line 62: This needs to be dynamic based on the `texture`.

erikdubbelboer:
I have rewritten the example to show how it can do some actual work with in and output. I have also added support for buffers as in and output, instead of just textures, as buffers allow for more useful work to be done using atomics.

src/platform/graphics/webgpu/webgpu-compute.js
```diff
@@ -39,7 +54,69 @@ class WebgpuCompute {
...
+     *
+     * @param {import('../texture.js').Texture} texture - The texture to get the buffer for.
+     * @returns {import('../buffer.js').Buffer} The buffer.
+     */
+    getTextureBuffer(texture) {
+        const formatInfo = pixelFormatInfo.get(texture.format);
+        const bytesPerPixel = formatInfo.size;
+
+        // Calculate bytes per row, ensuring it's a multiple of 256
+        const bytesPerRow = Math.ceil((texture.width * bytesPerPixel) / 256) * 256;
```
mvaligursky, line 76: use `math.roundUp`

mvaligursky:
And another update / building block: https://github.com/playcanvas/engine/pull/6186

mvaligursky:
Another related change: #6187

mvaligursky:
One more change, with different API to what you proposed here, but I believe it's simpler and more obvious. I'd appreciate your feedback @erikdubbelboer https://github.com/playcanvas/engine/pull/6201

I'll leave this PR opened as it still supports few additional buts I have not integrated yet. Your effort here is much appreciated.

erikdubbelboer:
@mvaligursky great that you have been working on this. All your changes look good and are in line with what I wrote as well.

One thing I'm wondering is, if compute shaders should be submitted with the normal render pass or if we should submit them right away. Like this: https://github.com/erikdubbelboer/engine/commit/19963f1f2643578bd9906fc5d21eadff6ddba531
I feel like this shouldn't work with the texture-gen example as that uses uniform buffers, which uses dynamic buffers, which are only submitted on `WebgpuGraphicsDevice.submit()`. Which then doesn't happen before we submit the compute shader. But I tested it and it does work, even when I only submit compute work before any call to `WebgpuGraphicsDevice.submit()`. Do you understand what is happening here?

mvaligursky:
UniformBuffers support both using dynamic buffers, but also it's own persistent buffer. I opted to use persistent buffers in this case, and so it works.

I thought about not submitting computes at the same time, but the most common use case we'd probably need is to integrate compute with render passes. For example the scene renders using render passes, then compute uses those textures to generates some buffers for postprocessing, and then render pass consumes it. In those cases it'd need to be submitted in order. So that's what I focused for now. Same for indirect drawing.

At some point we can relax this if there's no render pass dependency, but I have not solve this yet - what the simple API would be here.

At least that's my thinking currently, with not a large level of experience with all this.

erikdubbelboer:
But if you submit them right away they also get executed before the render commands are submitted. I tried various variations of calling `computeDispatch` in the texture-gen example and it all works. It will just finish earlier and make the render pass wait less.
The only way it doesn't work is if you call `computeDispatch` after the render pass, but that's always going to go wrong.

mvaligursky:
closing as all is now in the engine, thanks again @erikdubbelboer 

Final PR: https://github.com/playcanvas/engine/pull/6233

PR #6023, opened 2024-01-30
Add support for debug rendering obbs

slimbuck:
Add optional matrix so debug AABB rendering can be oriented.

### DIFF
--- a/src/framework/app-base.js
+++ b/src/framework/app-base.js
@@ -1802,7 +1802,7 @@ class AppBase extends EventHandler {
      * app.drawWireAlignedBox(min, max, pc.Color.RED);
      * @ignore
      */
-    drawWireAlignedBox(minPoint, maxPoint, color = Color.WHITE, depthTest = true, layer = this.scene.defaultDrawLayer, mat = undefined) {
+    drawWireAlignedBox(minPoint, maxPoint, color = Color.WHITE, depthTest = true, layer = this.scene.defaultDrawLayer, mat) {
         this.scene.immediate.drawWireAlignedBox(minPoint, maxPoint, color, depthTest, layer, mat);
     }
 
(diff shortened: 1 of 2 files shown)


mvaligursky:
We discussed this here: https://github.com/playcanvas/engine/discussions/3073#discussioncomment-590090
I was keen to add it back then, but it was agreed to keep those outside of the code engine. Perhaps this can be still outside?

mvaligursky:
Lets add it, and create a ticket to move all these non-pubic rendering helpers to extras for now?

src/framework/app-base.js
```diff
@@ -1794,15 +1794,16 @@ class AppBase extends EventHandler {
...
      * @example
      * // Render a red wire aligned box
      * const min = new pc.Vec3(-1, -1, -1);
      * const max = new pc.Vec3(1, 1, 1);
      * app.drawWireAlignedBox(min, max, pc.Color.RED);
      * @ignore
      */
-    drawWireAlignedBox(minPoint, maxPoint, color = Color.WHITE, depthTest = true, layer = this.scene.defaultDrawLayer) {
-        this.scene.immediate.drawWireAlignedBox(minPoint, maxPoint, color, depthTest, layer);
+    drawWireAlignedBox(minPoint, maxPoint, color = Color.WHITE, depthTest = true, layer = this.scene.defaultDrawLayer, mat = undefined) {
```
willeastcott, line 1805: Why do you need the `= undefined` here?

mvaligursky:
Created an issue: https://github.com/playcanvas/engine/issues/6024

src/framework/app-base.js
slimbuck: Because I'm still used to c++ parameter rules I guess?! Removed, thanks.

willeastcott (approved):

PR #6027, opened 2024-01-30
Gizmo Ortho Rotation fixes

kpal81xd:
- Fixes orthographic rotation speed
- Fixed orthographic facing logic for rotation axes and facing rotation (mimics rotation of camera instead of lookAt)

### DIFF
--- a/extras/gizmo/transform-gizmo.js
+++ b/extras/gizmo/transform-gizmo.js
@@ -3,6 +3,7 @@ import {
     CULLFACE_NONE,
     CULLFACE_BACK,
     PROJECTION_PERSPECTIVE,
+    PROJECTION_ORTHOGRAPHIC,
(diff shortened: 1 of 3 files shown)


extras/gizmo/transform-gizmo.js
```diff
@@ -628,6 +629,9 @@ class TransformGizmo extends Gizmo {
                 }
             } else {
                 angle = mouseWPos.dot(tmpV2.normalize()) * ROTATE_SCALE;
+                if (this._camera.projection === PROJECTION_ORTHOGRAPHIC) {
+                    angle /= this._camera.orthoHeight;
```
mvaligursky, line 633: it should not be zero, but handle it just in case.

mvaligursky (approved):

PR #6036, opened 2024-02-02
Gizmo grid and mesh shaders

kpal81xd:
- Replaces mesh rendering to use similar shader to grid lines to override depth setting for camera clipping.

Gizmo fully rendered without being clipped (cube is still clipped):
image


examples/src/examples/misc/gizmos.mjs
```diff
@@ -695,6 +778,36 @@ class GizmosExample {
     static WEBGPU_ENABLED = false;
     static controls = controls;
     static example = example;
+    static FILES = {
```
mvaligursky, line 781: Not sure the example needs to have non-clipping version of the grid, the example is about gizmos. This makes it more complex outside of its scope.

kpal81xd: Alright ill revert those changes 👍

slimbuck (approved):

PR #6040, opened 2024-02-05
RenderPassPrepass implementation

mvaligursky:
- prepass render pass, currently priming the depth buffer only, before the main pass
- this pass will be updated to render additional render targets such as velocity buffer in follow up PRs
- disabled for now till some pass uses it (TAA will)


slimbuck (approved):

Maksims:
Just a quick question - could it be called `RenderPrepass`?

mvaligursky:
Perhaps. For now all render pass names have format `RenderPassXXX` .. but I'm not sure that was the best decision ever ;)

PR #6041, opened 2024-02-05
Examples cleanup (2nd pass)

kpal81xd:
- Removed unused files
- Organises common files into folders
- Moves files from src into app where possible
- Removed unnecessary building of static data
- Renamed scripts and added comments to be more informative
- Extracted standalone html template into file for better readability
- Checked thumbnails in (prevent regenerating when cloning)


willeastcott (approved):
LGTM!

willeastcott:
I recommend updating this PR to include the thumbs in WebP format instead of PNG.

kungfooman:
I use `app.html` and the importmap because it allows you to circumvent `npm run ...` commands - CTRL+R with DevTools open and your changes are reflected instantly (without any scripts running anywhere).

I like to have the screenshots in the repo and luckily the dist/thumbnails folder is only 8 MB right now.

We could probably also add the generated HTML files and make them work as-is (so devs can simply clone the repo into any web root and are able to hack away with only a `npm install` - and then the importmap does the dependency resolution into `node_modules`).

kpal81xd:
> I use `app.html` and the importmap because it allows you to circumvent `npm run ...` commands - CTRL+R with DevTools open and your changes are reflected instantly (without any scripts running anywhere).
> 
> I like to have the screenshots in the repo and luckily the dist/thumbnails folder is only 8 MB right now.
> 
> We could probably also add the generated HTML files and make them work as-is (so devs can simply clone the repo into any web root and are able to hack away with only a `npm install` - and then the importmap does the dependency resolution into `node_modules`).

Ahh okay that makes sense. Yea im gonna redo the import map and get the examples generated properly instead of the current template system to allow for actual `import` statements

PR #6047, opened 2024-02-08
Update post effect queue type mismatch

battall:
currently, first argument of addEffect() function is typed as PostEffect which is wrong because it refers to PostEffect class that is declared on the same file, it should have been pointing to pc.PostEffect. but pc. prefix removed from docs so renamed it to PostEffectEntry, as it is assigned to variable named "newEntry".

Fixes #
renamed PostEffect class declared in post-effect-queue.js to PostEffectEntry. it is not exported so there is no need for renaming in other files.


mvaligursky:
Thanks for pointing this out @battall - I made a further changes to get it to pass the tests, and so will close this. Much appreciated!

PR #6048, opened 2024-02-09
TAA reprojection

mvaligursky:
Improvement to the initial TAA implementation (#5909):
- Support for moving camera

Other changes:
- partial support for velocity buffer rendering, will be finished in a follow up PR
- new TAA example 

https://github.com/playcanvas/engine/assets/59932779/4b00761b-9f0a-4473-a607-f205aef30e98

Limitations:
- No support for moving objects - velocity buffer implementation needs to be finalized for this.
- The scene gets slightly soft when TAA is enabled. This is typically improved by a sharpening pass, which will be added in a separate PR.


Maksims:
Velocity buffer can be very useful for WebXR Space Warp technique. https://developer.oculus.com/documentation/web/webxr-space-warp/

I guess the format of vector data and the way it is stored is important here. If it matches between implementations - that would be just amazing!

mvaligursky:
> Velocity buffer can be very useful for WebXR Space Warp technique. 

created an issue: https://github.com/playcanvas/engine/issues/6116

examples/src/examples/graphics/taa/example.mjs
```diff
@@ -0,0 +1,203 @@
...
+    app.start();
+
+    // setup skydome with low intensity
+    app.scene.envAtlas = assets.envatlas.resource;
+    app.scene.skyboxMip = 0;
+    app.scene.exposure = 1.0;
+
+    // render in HDR mode
+    app.scene.toneMapping = pc.TONEMAP_LINEAR;
+    app.scene.gammaCorrection = pc.GAMMA_NONE;
```
willeastcott, line 74: Linear with gamma disabled is HDR mode?

extras/render-passes/render-pass-taa.js
```diff
@@ -1,50 +1,151 @@
...
+        return vec4(clamped, historyColor.a);
+    }
+
+    void main()
+    {
+        vec2 uv = uv0;
+
+        #ifdef WEBGPU
+            // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+            // This hack is needed on webgpu, which makes TAA to work but the resulting image is upside-down.
```
willeastcott, line 69:
```suggestion
            // This hack is needed on webgpu, which makes TAA work but the resulting image is upside-down.
```

examples/src/examples/graphics/taa/example.mjs
mvaligursky: well that, and additionally those render passes render into HDR texture format.
And compose pass at the end applies tone mapping and converts it to gamma space.

But let me adjust the comment a bit.

extras/render-passes/render-pass-taa.js
```diff
@@ -1,50 +1,151 @@
...
+     * The index of the history texture to render to.
      *
      * @type {number}
      */
-    accumulationIndex = 0;
+    historyIndex = 0;
 
-    accumulationTexture = null;
+    /**
+     * @type {Texture}
```
willeastcott, line 116:
```suggestion
     * @type {Texture|null}
```

willeastcott (approved):

Issue #6030, opened 2024-01-31
Examples Rework


kpal81xd:
This issue is for the discussion of reworking the examples to make the development and experimentation experience easier. Stage 1 of the refactor and cleanup has already been completed and merged in #6005.

### Stage 1
- [x] Fix linting issues
- [x] Speed up thumbnail generation
- [x] Maintain state for collapsing on SideBar

### Stage 2
- [ ] Remove wrapping `example` and `controls` functions
- [ ] Support for `imports` 
- [ ] Abstract app configuration into separate file (boilerplate code)
- [ ] Replace `jsx` wrapper for controls to use jsx components (Will probably require PCUI rebuild/state optimisation)

mvaligursky:
> * Abstract app configuration into separate file (boilerplate code)

This was considered during last refactoring, but at the end keeping all / as much code required to run the example in the same file was decided is a good way to show people what needs to be done to get an app running, as examples are in a way tutorials too. Kind of self contained examples, similar to a completely self-contained html files examples used to be a long time ago.

mvaligursky:
Love the idea of `imports`, and would appreciate some more details on what you're hoping to do this, what the end goal is.

Few other points could do with more information as well to make it more clear what the goal here is.

marklundin:
It would be great if we can remove the direct dependancy on the `node_modules` of the engine, and just symlink it directly using [npm workspace](https://docs.npmjs.com/cli/v7/using-npm/workspaces/) which doesn't require building the parent engine first

mvaligursky:
The examples no longer need to build the engine first. Just `npm run develop` in the examples, that builds the engine it needs.

marklundin:
But `npm run build` in example doesn't work unless you build the engine first.

kungfooman:
IIRC the only missing thing is the engine `playcanvas.d.ts` generation for Monaco examples editor to make `/examples/` fully standalone, which would be nice to add here too

Maksims:
Currently the example code, shows the core of the example stuff, but at the same time, it is not easily replicable from the scratch.
Maybe it is not related to examples, but amount of build processes, the wrapping, lack of debugging and proper error handling - all this moves away from users being able to take an example, and expand from it.

In contrast, three.js example - is a html file, with everything in it, you can just take it and experiment with it, without need of complex build processes, understanding of project structure, etc.

willeastcott:
The PCUI examples browser is like that:

[...]

PR #6060, opened 2024-02-16
Examples Overhaul

kpal81xd:
Refers #6030

This PR aims to align the examples experience to of using the engine in a standalone app.

## Changes
- Removed `example` function and arguments and put body into example script global scope (app is now accessed by using `export { app };` at the bottom of the script)
- Moved arguments from `example` into modules to load in for each script
- Acquiring the canvas element is handled by the example rather than being passed in as an argument
- Converted multiple paths (`assetPath`, `glslPath` etc) to a `rootPath` which serves all static files
- Added support for imports of ES modules
- Added syntax highlighting for shaders

## Demo
This PR has a lot of changes it so I have provided a demo serving these changes.
https://kpal81xd.github.io/playcanvas-examples

## Preview
image

## ToDo
- [ ] Convert `observer` to use module
- [x] Move config for each example to its own separate `config.mjs` from `examples.config.mjs`
- [x] Fix broken thumbnail generation


examples/src/app/components/Example.mjs
```diff
@@ -232,7 +256,7 @@ class Example extends TypedComponent {
             null,
             jsx(controls, {
                 // @ts-ignore
-                observer: window.observerData,
+                observer,
```
willeastcott, line 259: Is the `@ts-ignore` above still needed?

examples/src/examples/graphics/reflection-planar/controls.mjs
```diff
@@ -0,0 +1,7 @@
+/**
+ * @param {import('../../../app/components/Example.mjs').ControlOptions} options - The options.
+ * @returns {JSX.Element} The returned JSX Element.
+ */
+export function controls({ fragment }) {
+    return fragment();
```
willeastcott, line 6: For examples that don't specify controls, would it be better to allow for `controls.mjs` to be left unspecified?

examples/iframe/observer.mjs
```diff
@@ -0,0 +1,6 @@
+let data;
+function refresh() {
+    data = new observer.Observer({});
```
willeastcott, line 3: Can't you use the module version of `Observer`?

examples/iframe/files.mjs
```diff
@@ -0,0 +1,7 @@
+import config from '@examples/config';
+const files = {
+    'example.mjs': '',
+    'controls.mjs': ''
+};
+Object.assign(files, config.FILES ?? {});
+export default files;
```
willeastcott, line 7:
```suggestion
import config from '@examples/config';

const files = {
    'example.mjs': '',
    'controls.mjs': '',
    ...config.FILES
};

export default files;
```

willeastcott (approved):
This PR is marvelous! This is a superb refinement of the Examples Browser. Congrats! 👏

kungfooman:
Any reflection on the trade-offs on `examples/examples.config.mjs`? Now all the files for every example are in one huge file and not per-example any longer - at least I feel uneasy about that. Having one file for each example allows for easier organization and modification of individual examples and one-big-file-for-all-examples feels like making it more:
 - difficult to maintain
 - switching file contexts for doing same-example work
 - git conflicts even tho people work on completely different examples
 - extra step that requires more mental overhead than just attaching files into the old static `Example` class

Turning the iframe functions into OOP feels a bit over-engineered, many changes like:

```diff
- setTimeout(iframeResize, 50);
+ setTimeout(() => iframe.fire('resize'), 50);
```

examples/src/examples/graphics/reflection-planar/controls.mjs
kpal81xd: It just included it just in the case where someone would want to experiment with the controls since the code editor window doesn't show if there are no controls specified

examples/src/app/components/Example.mjs
kpal81xd: No I will remove it :)

examples/iframe/observer.mjs
kpal81xd: Yes that is true I will add this too

kpal81xd:
> Any reflection on the trade-offs on `examples/examples.config.mjs`? Now all the files for every example are in one huge file and

- So actually I was thinking of converting that to put a `config.mjs` file in each of the example folders now along with `controls.mjs` and `example.mjs` so it would not all have to be in a single file

- As for the iframe it was just adding scope to forward any events from the window over to the iframe in a more abstract context to make it easier to add new ones if needed

examples/src/examples/graphics/reflection-planar/controls.mjs
kpal81xd: I have removed the empty files in favor of a template now 👍

kungfooman:
Thank you for the quick changes, you work so fast :100:

What's up with all the file name changes? We have four conventions by now:

 - loader.js for `ExampleLoader` - filename only somewhat related to class name
 - ministats.js for `MiniStats` - lowercase of class name
 - We use kebab-case.js in normal engine that you converted to PascalCase.js in examples
 - PascalCase.js that you converted to in some places

(I don't care too much, just wondering)

I build it locally and build process worked like a charm so far, just need some time to test around here and there to get a picture

examples/src/app/events.js
```diff
@@ -1,37 +1,19 @@
 /**
- * @typedef {object} Loading
+ * @typedef {CustomEvent} LoadingEvent
```
kungfooman, line 2: `LoadingEvent` doesn't seem to be typed any longer, but we can make use of the template:

```ts
interface CustomEvent<T = any> extends Event {
    /** Returns any custom data event was created with. Typically used for synthetic events. */
    readonly detail: T;
    /** @deprecated */
    initCustomEvent(type: string, bubbles?: boolean, cancelable?: boolean, detail?: T): void;
}
```
For example:

```js
/**
 * @typedef {object} LoadingEventDetail
 * @property {boolean} showDeviceSelector - Show device selector.
 */
/**
 * @typedef {CustomEvent<LoadingEventDetail>} LoadingEvent
 */
/** @type {LoadingEvent} */
const test = {
    detail: {
        showDeviceSelector: true
    }
}
```

examples/src/app/index.mjs
```diff
@@ -13,7 +16,4 @@ function main() {
     root.render(jsx(MainLayout, null));
 }
 
-window.onload = () => {
-    // Just a little timeout to give browser some time to "breathe"
-    setTimeout(main, 50);
-};
+main();
```
kungfooman, line 19: It sounds silly to give it time to breathe, but Chrome would randomly freeze on me during reloads... not sure if this is/was a Linux issue tho lol

kungfooman:
I tested on my OnePlus 11 (Chrome/Android) and portrait mode has some problems:

Screenshot from PC (just scale devtools until examples window is going into portrait mode):

image

And on Android I have the opposite effect (being a bit too wide):

image

Not a big deal, just looks a bit strange: 

image

Probably easier to fix in an extra PR, some examples also have so many options that I can't scroll to them...

otherwise - wow, I like the GLSL formatting, it seems to have some token issues here and there (mixed blue/white tokens):

image

Nothing that is a blocker here that I came across, good work :+1:

kungfooman (approved):
:100:

kpal81xd:
> Thank you for the quick changes, you work so fast 💯
> 
> What's up with all the file name changes? We have four conventions by now:
> 
> * loader.js for `ExampleLoader` - filename only somewhat related to class name
> * ministats.js for `MiniStats` - lowercase of class name
> * We use kebab-case.js in normal engine that you converted to PascalCase.js in examples
> * PascalCase.js that you converted to in some places
> 
> (I don't care too much, just wondering)
> 
> I build it locally and build process worked like a charm so far, just need some time to test around here and there to get a picture

So all of it should be kebab case with the exception of React Components which I wanted to convert to PascalCase

examples/src/app/events.js
kpal81xd: Oh okay thanks I replace these!

examples/src/app/index.mjs
kpal81xd: I have never seen this issue before or could replicate it but let me know if it still persists and Ill look into it 👍

examples/package.json
```diff
@@ -10,7 +10,7 @@
     "build:metadata": "node ./scripts/metadata.mjs",
     "build:sharing": "node ./scripts/sharing-html.mjs",
     "build:standalone": "node ./scripts/standalone-html.mjs",
-    "build:thumbnails": "npm run build && node ./scripts/thumbnails.mjs",
+    "build:thumbnails": "node ./scripts/thumbnails.mjs",
```
kungfooman, line 13: We discussed this a bit here: https://github.com/playcanvas/engine/pull/5555#issuecomment-1719108268

Basically it allows for a cold start - but I also agree that it takes quite some time before building thumbnails. @mvaligursky Do you have a opinion on this?

kpal81xd: Well currently this would only be required when modifying or creating new examples now since thumbnails are included in the repo itself now. 

Yea having the thumbnails resolved using the imports I don't like since it splits how the example names are generated from folder and file names to the names specified in the import `index.js` files. One workaround would be setting file path in the `config` to read after the import is resolved but that seems messy.

Additionally you don't need to add example to some index file when generating a new one you can simply add the file within the specified folder.

marklundin:
What is the build process for this in relation to the engine? Does it still rely on building the engine first, or can we just import directly?

```
npm install && npm build
cd ./examples
npm install && npm build
```

kungfooman:
> What is the build process for this in relation to the engine? Does it still rely on building the engine first, or can we just import directly?
> 
> ```
> npm install && npm build
> cd ./examples
> npm install && npm build
> ```

Yea, it still relies on `playcanvas.d.ts` from engine, so you still need the double build

mvaligursky:
I've noticed that if I reload the example (even without making a change to it), I get this:

Screenshot 2024-02-20 at 14 28 32

PR #6063, opened 2024-02-19
ResourceHandler type fix (constructor parameters)

kpal81xd:
- Fixes mismatched constructor types of child and parent classes for `ResourceHandler` which caused linting errors.

### DIFF
--- a/src/framework/handlers/anim-clip.js
+++ b/src/framework/handlers/anim-clip.js
@@ -12,8 +12,9 @@ import { ResourceHandler } from './handler.js';
  * @ignore
  */
 class AnimClipHandler extends ResourceHandler {
-    constructor() {
-        super('animclip');
+    constructor(app) {
+        super(app);
+        this.handlerType = 'animclip';
     }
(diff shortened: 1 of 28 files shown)


willeastcott (approved):

src/framework/handlers/anim-clip.js
```diff
@@ -12,8 +12,9 @@ import { ResourceHandler } from './handler.js';
  * @ignore
  */
 class AnimClipHandler extends ResourceHandler {
-    constructor() {
-        super('animclip');
+    constructor(app) {
+        super(app);
+        this.handlerType = 'animclip';
```
mvaligursky, line 17: why not make the super constructor accept two parameters: `app`, and the `handlerType`?
so that this would be `super(app, 'animclip');`

kpal81xd: I thought that might have caused a constructor type clash but it apparently does not so Ive modified it with this change 👍

PR #6064, opened 2024-02-19
Small cleanup of SkinInstance

mvaligursky:
- make sure the initial bone texture lock has unlock, to avoid recently introduced asserts for non-matching locking
- skin-instance stores uniform array with texture sizes, to avoid filling this in for each drawcall


willeastcott (approved):

willeastcott:
Oh, weird how this was mergeable after I dismissed my review. I was gonna say that adding the essentially public API (despite it not being in the docs) `SkinInstance#boneTextureSize` isn't super-ideal. But hey.

PR #6073, opened 2024-02-21
Added ES6 bundle output

marklundin:
This PR adds an additional ES6 bundle `playcanvas.esm.js` as a build output. It uses the same ES6 build configuration as the unbundled `playcanvas.mjs` but instead simply outputs to a single output. 

We plan on upgrading the launcher and exported projects to use this ES6 build to support ESM Scripts.

### DIFF
--- a/utils/rollup-build-target.mjs
+++ b/utils/rollup-build-target.mjs
@@ -105,7 +105,7 @@ function buildTarget(buildType, moduleFormat, input = 'src/index.js', buildDir =
 
     const outputExtension = {
         es5: '.js',
-        es6: shouldBundle ? '.esm.js' : '.mjs'
+        es6: shouldBundle ? '.mjs' : ''
     };
 
     /** @type {Record<string, ModuleFormat>} */
--- a/utils/rollup-build-target.mjs
+++ b/utils/rollup-build-target.mjs
@@ -169,8 +169,6 @@ function buildTarget(buildType, moduleFormat, input = 'src/index.js', buildDir =
         es6: moduleOptions(buildType)
     };
 
-    console.log(babelOptions[moduleFormat])
-
     return {
         input,
         output: outputOptions,
--- a/utils/rollup-build-target.mjs
+++ b/utils/rollup-build-target.mjs
@@ -56,9 +56,10 @@ const stripFunctions = [
  * @param {'es5'|'es6'} moduleFormat - The module format.
  * @param {string} input - Only used for Examples to change it to `../src/index.js`.
  * @param {string} [buildDir] - Only used for examples to change the output location.
+ * @param {Boolean} [shouldBundle] - Whether the target should be bundled.
  * @returns {RollupOptions} One rollup target.
  */
-function buildTarget(buildType, moduleFormat, input = 'src/index.js', buildDir = 'build') {
+function buildTarget(buildType, moduleFormat, input = 'src/index.js', buildDir = 'build', shouldBundle = false) {
     const banner = {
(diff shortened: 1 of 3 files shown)


kungfooman:
Very nice, people needed this for quite some time - unbundled ESM builds were causing:
 - no problem for local development, but long request times for every individual file over internet
 - services like UNPKG have request limits and simply drop further ones

marklundin:
> Very nice, people needed this for quite some time - unbundled ESM builds were causing:
> 
> * no problem for local development, but long request times for every individual file over internet
> * services like UNPKG have request limits and simply drop further ones

Unfortunately even the ES Modules build includes [these polyfils](https://github.com/playcanvas/engine/blob/main/src/index.js#L12-L22) which are likely redundant

utils/rollup-build-target.mjs
```diff
@@ -104,7 +105,7 @@ function buildTarget(buildType, moduleFormat, input = 'src/index.js', buildDir =
 
     const outputExtension = {
         es5: '.js',
-        es6: '.mjs'
+        es6: shouldBundle ? '.esm.js' : '.mjs'
```
kungfooman, line 108: In the current file extension scheme `.js` means UMD, shouldn't we keep `.mjs`? Maybe something like `playcanvas.bundled.mjs`

mvaligursky:
Should we have other flavours of this bundle? At least debug, performance, and release (normal).

marklundin:
> Should we have other flavours of this bundle? At least debug, performance, and release (normal).

Yep there's the same outputs, dbg, prf and min

slimbuck:
Had a little discussion among the team and we'd vote for:
`playcanvas.js` - umd build, as before
`playcanvas.mjs` - the ESM bundled version
`playcanvas` - the RSM unbundled version

And of course .dgb, .min, .profile version of each.

kungfooman:
We also need to change in that case:

https://github.com/playcanvas/engine/blob/631493694dd3fb23c9c79fa77c6f4a6380c3aa1c/package.json#L21

marklundin:
> Had a little discussion among the team and we'd vote for: `playcanvas.js` - the UMD build, as before `playcanvas.mjs` - the ESM bundled version `playcanvas` - the ESM unbundled version
> 
> And of course .dgb, .min, .profile version of each.

Love this. Done! 

Assuming the unbundled 'playcanvas' dir doesn't need to have 'playcanvas.prf', 'playcanvas.min', 'playcanvas.dbg' variants

marklundin:
> We also need to change in that case:
> 
> https://github.com/playcanvas/engine/blob/631493694dd3fb23c9c79fa77c6f4a6380c3aa1c/package.json#L21

Yep agreed. Have updated

marklundin:
What would be the naming convention on these folders? Having the dot suffix
doesn't feel right. kebab case 'playcanvas-dbg/index.js' might work if we do

mvaligursky:
If we expect people to use the unbundled versions, we need to provide the variants, as they cannot easily replicate our processes to strip out debug info, and they need debug version at the very least.

slimbuck:
For consistency naming the folders `playcanvas`, `playcanvas.dbg`, `playcanvas.min`, `playcanvas.prf` seems fine to me.

slimbuck:
Though in reality the `.min` unbundled build has no use whatsoever (the rest do).

marklundin:
Have added unbundled es6 variants excluding 'min' @slimbuck @mvaligursky

slimbuck (approved):
<3

PR #1855, opened 2020-02-05
Async shader compilation (KHR_parallel_shader_compile)

Maksims:
Fixes #1474

Restored code of: https://github.com/playcanvas/engine/pull/1685
After so many years of using git, I still mess up :D

I confirm I have signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #6076, opened 2024-02-21
Link WebGL shaders immediately after compilation

mvaligursky:
Based on https://github.com/playcanvas/engine/pull/5914, thanks @erikdubbelboer for the work on this.

I tested and confirmed the behaviour on Chrome (and also Safari) has changed compared to my findings in https://github.com/mvaligursky/webgl-parallel_shader_compile

Chrome can now compile 50 large shaders without locking the main thead (as long as the result is not needed), unlike before. Likely some change in ANGLE, as that's where the expected fix was supposed to land.

It seems link right after compile step is beneficial. Before, we would trigger compilation step on all shaders in a pass, and then trigger linking on all of them, to get the best speed. Now it seems no work is done during compile step (on Chrome), and so unless we call the link step, browser does not start the compilation at all. 

By this change we're gaining little time it takes to trigger the compilation for all shaders - this time now seems to be used to link the first few shaders.

I also stripped out some code managing the delayed linking, as that is no longer needed.

### DIFF
--- a/src/platform/graphics/webgl/webgl-graphics-device.js
+++ b/src/platform/graphics/webgl/webgl-graphics-device.js
@@ -1353,15 +1353,6 @@ class WebglGraphicsDevice extends GraphicsDevice {
     }
 
     /**
-     * Called after a batch of shaders was created, to guide in their optimal preparation for rendering.
-     *
-     * @ignore
-     */
-    endShaderBatch() {
-        WebglShader.endShaderBatch(this);
-    }
-
-    /**
      * Set the active rectangle for rendering on the specified device.
      *
      * @param {number} x - The pixel space x-coordinate of the bottom left corner of the viewport.
--- a/src/platform/graphics/webgl/webgl-shader.js
+++ b/src/platform/graphics/webgl/webgl-shader.js
@@ -34,18 +34,8 @@ class CompiledShaderCache {
     }
 }
 
-// class used to hold a list of recently created shaders forming a batch, to allow their more optimized compilation
-class ShaderBatchCache {
-    shaders = [];
-
-    loseContext(device) {
-        this.shaders = [];
-    }
-}
-
 const _vertexShaderCache = new DeviceCache();
 const _fragmentShaderCache = new DeviceCache();
-const _shaderBatchCache = new DeviceCache();
 
 /**
  * A WebGL implementation of the Shader.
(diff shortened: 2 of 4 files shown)


willeastcott (approved):

Maksims:
Perhaps with changes in ANGLE, async shader compilation potentially works now?

mvaligursky:
yep, based on my test project https://github.com/mvaligursky/webgl-parallel_shader_compile running on up to date Chrome (121) async compilation is finally useful, main thread framerate is now solid even though we compile lots of large shaders.

And so I'm thinking how to use it.

My current plan (still thinking about it) would be to add some flag to the material (public API), to indicate that any meshes using this material can be skipped from being rendered if the shader is not compiledyet. People would need to set it, by default likely false to not force any changes to existing projects.

I'd check for this flag inside forward-renderer render loop, but for the forward pass only. So ignored for shadow, picker and similar. Also would need to disable it for lightmapper. The thinking here is that shadow / picker shaders are fast to compile, and we handle 'render once' and similar there, so sometimes we do not want to skip rendering (don't want the mesh missing in render-once shadow map). On the other hand, forward shaders are slow to compile, so we specifically want to skip those.

I'll try something like this and see what benefits we get.

Maksims:
My [previous attempt](https://github.com/playcanvas/engine/pull/1855) to implement it was pretty much as you've described: it would force compile for certain types of variants (shadowmaps, pickers, etc). And provide a way to async compile for materials.

Also, I thought if you would call `material.update`, you could provide a callback, so you can know when it was compiled.
Something like:

```javascript
material.async = true;
material.chunk.endPS = shaderCode;
material.update((err) => {
    // shader has been recompiled
});
```

This gives control to the developer and also provides a way to implement some loading bars.

Also a global counter of a number of shaders that are currently compiling - would be good.

mvaligursky:
Not sure a callback is super useful, as we compile multiple variants based on lights / passes / mesh vertex format and similar, not a single shader.

The same goes for compiling shaders during loading - at that point we do not now how the material will be used - what layers it needs to collect the lights from, what are the scene settings going to be at the time we need to render it. This is hard to do. I was looking at restoring the functionality of the shader cache we have .. dump those, and those we can precompile. But there are several issues there the way shader generation works for different platforms, that could cause shaders not needed for the platform to be compiled. Again, non trivial solution here would be needed.

MAG-AdrianMeredith:
Thanks for all your hard work everyone.  From a UX perspective I'd like to be able to have the ability for a "compile shaders" stage during loading so that we can load up the assets of the scene then pre-eptively compile as much as possible.  We previously attempted this by disabling culling on the camera to force everything to draw (but its currently disable as our lod system kept triggering it)

PR #6077, opened 2024-02-21
WebXR FrameRate

Maksims:
The developer will be able to control the target frame rate of an XR session based on a provided list of supported frame rates.

XR device can change framerate based on its underlying mechanics, but the user has control over it also. For example, Quest 3 provides these supported frame rates: [ 72, 80, 90, 120 ], and by default it will use 90 Hz. But you can ask it to render with 120 Hz if your app can manage such performance.

New APIs:
```javascript
// pc.XrManager
xr.frameRate // (number) a target frame rate of the current session, or null if not supported or no XR session is available
xr.supportedFrameRates // (Float32Array) list of supported frame rates, or null if not supported or no XR session is available
xr.updateTargetFrameRate(frameRate: number, (err) => { }); // async method to update frame rate based on list of supported values
xr.on('frameratechange', (frameRate: number) => { }); // event when framerate has been changed 
```

### DIFF
--- a/src/framework/xr/xr-manager.js
+++ b/src/framework/xr/xr-manager.js
@@ -623,6 +623,25 @@ class XrManager extends EventHandler {
     }
 
     /**
+     * @param {number} frameRate - Target frame rate. It should be any value from the list
+     * of supportedFrameRates.
+     * @param {Function} [callback] - Callback that will be called when frameRate has been
+     * updated or failed to update with error provided.
+     */
+    updateTargetFrameRate(frameRate, callback) {
+        if (!this._session?.updateTargetFrameRate)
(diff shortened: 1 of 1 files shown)


mvaligursky (approved):

willeastcott (approved):

src/framework/xr/xr-manager.js
```diff
@@ -633,6 +639,30 @@ class XrManager extends EventHandler {
         });
     }
 
+    /**
+     * Update target frame rate of an XR session to one of supported value provided by
+     * supportedFrameRates list.
+     *
```
mvaligursky, line 645:
```suggestion
     *
```

PR #6087, opened 2024-02-25
Mesh collision optimization

LeXXik:
This is a follow up on #5818 

After we have created a mesh collider, the engine proceeds with setting the scale of the mesh. This is an expensive operation in Ammo, as it triggers another loop over all verts and recalculating AABB, which was already calculated once when we created `btBvhTriangleMeshShape`.
Since we are already looping over verts on JS side, we can apply scale in place instead.

This. further improves collider generation (24k verts) from `~181 ms` down to `~115 ms`.

Note:
This optimization is only for Render Asset types. There is no change to legacy Model Assets, which will continue using Ammo scaling as they do now.

### DIFF
--- a/src/framework/components/collision/system.js
+++ b/src/framework/components/collision/system.js
@@ -445,7 +445,7 @@ class CollisionMeshSystemImpl extends CollisionSystemImpl {
                     this.createAmmoMesh(meshes[i], tempGraphNode, shape, scale, data.checkVertexDuplicates);
                 }
             }
-            
+
             return shape;
         }
 
--- a/src/framework/components/collision/system.js
+++ b/src/framework/components/collision/system.js
@@ -332,7 +332,7 @@ class CollisionMeshSystemImpl extends CollisionSystemImpl {
     // special handling
     beforeInitialize(component, data) {}
 
-    createAmmoMesh(mesh, node, shape, checkDupes = true) {
+    createAmmoMesh(mesh, node, shape, scale, checkDupes = true) {
         const system = this.system;
         let triMesh;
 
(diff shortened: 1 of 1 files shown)


LeXXik:
We could probably change that for Model Assets as well, but I am not too familiar with the legacy format. Can the nodes on model mesh instances be of any other scale, than 1?

mvaligursky:
Thanks for the PR @LeXXik .
Yes, the model component works the same way, mesh instances have GraphNodes and those would have scales matching those from render component.

src/framework/components/collision/system.js
```diff
@@ -407,9 +411,11 @@ class CollisionMeshSystemImpl extends CollisionSystemImpl {
 
         const triMeshShape = new Ammo.btBvhTriangleMeshShape(triMesh, true /* useQuantizedAabbCompression */);
 
-        const scaling = system._getNodeScaling(node);
-        triMeshShape.setLocalScaling(scaling);
-        Ammo.destroy(scaling);
+        if (!scale) {
```
mvaligursky, line 414: I wonder if instead of passing the scale parameter to the function, and then either using it, or extracting it from the node, we should not pass it in, and always extract it from the node? And apply directly to vertices that you introduced here. I'm just not sure what the advantage is in having to pass it in, considering we can extract it from the node?

LeXXik: Hmm, in case of Render Asset that node is a temp node, which will always has a scale 1, so not needed. In case of Mesh Asset it will be a node from the mesh instance. I wasn't sure if nodes on model mesh instances could have a different scale, so left it there as is.

mvaligursky (approved):
Looks good, thanks!

PR #6090, opened 2024-02-26
Add rollup import transform

marklundin:
This PR addresses a long standing issue with the UMD build not running on devices without dynamic `import()` support. See #6011.

This issue stemmed from the use of dynamic `import()` statement, where `import` is a reserved keyword and would throw a SyntaxError during parsing on older browsers, even if the code path was never executed.

This PR adds a rollup transform to the UMD build only, which wraps `import()` statements in a dynamic function. This is only evaluated at runtime, thereby avoiding the parsing on load error. Note that this does not polyfill dynamic imports, it will simply prevent parsing errors on browser without support.

### Caveats
This does put a limitation on CSP for UMD builds, however this only applies if *all of the following conditions are met*

1. The UMD build is used (not applicable to the ES6 build) **AND**
2. Hosting blocks `unsafe-eval` in their CSP **AND**
3. When using either WebGPU or importing an ESM ScriptType

### DIFF
--- a/utils/rollup-dynamic-import-transform.mjs
+++ b/utils/rollup-dynamic-import-transform.mjs
@@ -17,7 +17,7 @@ export function dynamicImportTransform() {
              * @returns {object} - The transformed code and map.
              */
             return {
-                code: code.replace(/import\(/g, 'new Function("modulePath", "return import(modulePath)")('),
+                code: code.replace(/([^\w])import\(/g, '$1new Function("modulePath", "return import(modulePath)")('),
                 map: null
             };
         }
--- a/utils/rollup-dynamic-import-transform.mjs
+++ b/utils/rollup-dynamic-import-transform.mjs
@@ -3,7 +3,7 @@
  * in a `new Function('import("modulePath")')` statement, in order to avoid parsing errors in older browsers
  * without support for dynamic imports.
  *
- * Note that whilst this will prevent parsing errors, it will
+ * Note that whilst this will prevent parsing errors, it can trigger CSP errors.
  */
 
 export function dynamicImportTransform() {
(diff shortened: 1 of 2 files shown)


utils/rollup-dynamic-import-transform.mjs
```diff
@@ -0,0 +1,25 @@
+/**
+ * This rollup plugin transform code with dynamic import statements and wraps them
+ * in a `new Function('import("modulePath")')` statement, in order to avoid parsing errors in older browsers
+ * without support for dynamic imports.
+ *
+ * Note that whilst this will prevent parsing errors, it will
```
mvaligursky, line 6: hmm looks half finished sentence? :)

mvaligursky (approved):
Looks great, but please give some time for additional opinions before merging.

@kungfooman perhaps?

kungfooman (approved):
Looks very good 👍

utils/rollup-dynamic-import-transform.mjs
```diff
@@ -0,0 +1,25 @@
...
+        name: 'dynamic-import-transform',
+        transform(code, id) {
+            /**
+             * Transforms the code by replacing `import(` with `new Function("return import")(`.
+             * @param {string} code - The code to transform.
+             * @param {string} id - The id of the code.
+             * @returns {object} - The transformed code and map.
+             */
+            return {
+                code: code.replace(/import\(/g, 'new Function("modulePath", "return import(modulePath)")('),
```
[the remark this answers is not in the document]
kpal81xd: @marklundin  Just change the regex to be `/[^\w]import/g` which unmatch any character in front of `import`

kpal81xd, line 20:
```suggestion
                code: code.replace(/([\w^])import\(/g, '$1new Function("modulePath", "return import(modulePath)")('),
```

kpal81xd (approved):

slimbuck (approved):

PR #6094, opened 2024-02-27
Ditherered transparency example - enable dorth write when dithering is on

mvaligursky:
this makes sure the skybox does not overwrite it


LeXXik:
dorth = depth

slimbuck (approved):

PR #6108, opened 2024-03-05
Added worker environment

marklundin:
When a user script is parsed in the editor, it's executed in a Worker which doesn't have `window` or `global`.  

This means that when the following ESM Script is parsed in the editor, it will import the engine and immediately throw because neither `window` nor `global` exists.

```javascript
import { ScriptType } from 'playcanvas';
class MyRotator extends ScriptType {}
```

This PR adds `environment: 'worker'` and `global: self` as additional values which prevents this error.

### DIFF
--- a/src/core/platform.js
+++ b/src/core/platform.js
@@ -17,7 +17,8 @@ const detectPassiveEvents = () => {
 };
 
 const ua = (typeof navigator !== 'undefined') ? navigator.userAgent : '';
-const environment = (typeof window !== 'undefined') ? 'browser' : 'node';
+const environment = typeof window !== 'undefined' ? 'browser' :
+    typeof global !== 'undefined' ? 'node' : 'worker';
 
 // detect platform
 const platformName =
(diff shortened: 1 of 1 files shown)


mvaligursky (approved):
Looks fine. What's the `globalThis`?

marklundin:
globalThis is  global which works across everything, but it's not available pre Chrome-71 era, so falls back to using `environment`. I considered a polyfill, but it's so small

src/core/platform.js
```diff
@@ -62,19 +63,22 @@ const platform = {
...
 
     /**
      * The global object. This will be the window object when running in a browser and the global
-     * object when running in nodejs.
+     * object when running in nodejs and self when running in a worker.
      *
      * @type {object}
      */
-    global: (environment === 'browser') ? window : global,
+    global: globalThis ??
```
slimbuck, line 78: Won't this throw on browsers that didn't support globalThis? You might need to do `typeof globalThis !== undefined` before attempting to reference it.

marklundin: Yes you're 100% right. Good spot. Will patch this

PR #6111, opened 2024-03-05
Update banner to mention MIT license

willeastcott:
Update built engine banner to explicitly call out that the build is under an MIT license.

### DIFF
--- a/utils/rollup-get-banner.mjs
+++ b/utils/rollup-get-banner.mjs
@@ -7,13 +7,14 @@ import { version, revision } from './rollup-version-revision.mjs';
  * @returns {string} - The banner.
  */
 function getBanner(config) {
-    return [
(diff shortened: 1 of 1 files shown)


mvaligursky (approved):

lucaheft:
I would prefer to actually insert the full license, because if the engine is used in the context of a website. There is no source tree.
I know this will add a few bytes to the file size, but is the proper way if I understand correctly.

willeastcott:
@lucaheft If it's good enough for React, isn't it sufficient here?

https://unpkg.com/react@16.7.0/umd/react.production.min.js

lucaheft:
Welp i guess so. I'm no legal expert, but if they are doing it like that it's propably alright. Thanks for adding!

PR #6112, opened 2024-03-06
Update README files to show ESM usage

willeastcott:
Tweak the repo README:

* Switch from UMD example usage to ESM.
* Link to the full engine 'getting started' guide on the Developer Site.
* Remove reference to `code.playcanvas.com` (we don't want to encourage devs to use this)
* Simplify the build commands table.
* Translate to kr, zh and ja.


Maksims:
The beauty of the current code is that you can just copy it into clean html, and launch it.

willeastcott:
@Maksims Yeah, I totally agree with you. But we are deprecating both the UMD builds and code.playcanvas.com sometime in the not too distant future. We need new devs to dive right in with ESM....event if the results aren't quite so immediate.

mvaligursky (approved):

PR #6114, opened 2024-03-06
LightComponent properties port to get/set

kpal81xd:
Port of `LightComponent` from using `Object.defineProperty` to using `get/set`

- [x] TS linter
- [x] API reference


package.json
```diff
@@ -146,8 +146,8 @@
     "build": "rollup -c",
     "build:release": "rollup -c --environment target:release",
     "build:debug": "rollup -c --environment target:debug",
-    "build:dts": "tsc src/index.js --outDir types --allowJs --declaration --emitDeclarationOnly",
-    "build:dts:extras": "tsc extras/index.js --allowJs --declaration --emitDeclarationOnly --outDir build/playcanvas-extras",
+    "build:dts": "tsc --project tsconfig.engine.json",
+    "build:dts:extras": "tsc --project tsconfig.extras.json",
```
willeastcott, line 150: Try to keep this PR focused on the `LightComponent`. If you wanna revise the `package.json`, consider doing that in another PR.

utils/types-fixup.mjs
```diff
@@ -165,56 +165,6 @@ import { Font } from '../../../framework/font/font.js';
...
-path = './types/framework/components/light/component.d.ts';
-dts = fs.readFileSync(path, 'utf8');
-dts = dts.replace(regexConstructor, '$&\n' + getDeclarations(lightComponentProps));
-// We need to import types that are newly introduced in the property list above
-dts += `
-import { Color } from '../../../core/math/color.js';
-import { Vec2 } from '../../../core/math/vec2.js';
-`;
-fs.writeFileSync(path, dts);
-
```
willeastcott, line 217: Good riddance to this stuff! 👋 So satisfying.

src/framework/components/light/component.js
```diff
@@ -178,6 +54,841 @@ class LightComponent extends Component {
         this._cookieMatrix = null;
     }
 
+    /**
+     * @type {import('./data.js').LightComponentData}
+     */
+    get data() {
+        const record = this.system.store[this.entity.getGuid()];
+        return record ? record.data : null;
+    }
```
willeastcott, line 63: Isn't this defined in `Component` superclass? Why do we need this anyway? The new component architecture doesn't use the data block. All component data is now stored on the component itself.

kpal81xd: This was purely to define the type of data and there is no way to to override a `getter`

src/framework/components/light/component.js
```diff
@@ -178,6 +54,841 @@ class LightComponent extends Component {
...
+    set penumbraSize(value) {
+        this.light.penumbraSize = value;
+    }
+
+    get penumbraSize() {
+        return this.light.penumbraSize;
+    }
+
+    /** @ignore */
+    _setValue(name, value, setFunc, skipEqualsCheck) {
```
willeastcott, line 884: Am I right that 4 properties enable `skipEqualsCheck`? Why would you ever want to skip the equals check?

willeastcott: Also, doesn't the `Light` class do an equals check? Is there some inefficiency there?

kpal81xd: I am not sure this is all old code I think i will do a second pass to clean it up - This PR was just mainly to remove the type fixing for LightComponent but I can extend it to a LightComponent refactor

willeastcott: Yeah, probably best not to expand the remit of this PR beyond the bare minimum otherwise it could introduce bugs.

kpal81xd:
@willeastcott I just added a comment for when I update the component to remove that data method

src/framework/components/light/component.js
```diff
@@ -178,6 +54,842 @@ class LightComponent extends Component {
...
+    }
+
+    get enabled() {
+        return this.data.enabled;
+    }
+
+    /**
+     * The light.
+     *
+     * @type {import('../../../scene/light.js').Light}
```
willeastcott, line 82: Should this have an @ignore?

src/framework/components/light/component.js
```diff
@@ -178,6 +54,842 @@ class LightComponent extends Component {
...
+            this.light.bakeArea = math.clamp(newValue, 0, 180);
+        });
+    }
+
+    get bakeArea() {
+        return this.data.bakeArea;
+    }
+
+    /**
+     * The distribution of subdivision of the camera frustum for
```
willeastcott, line 331: Use full 100 characters line length.

src/framework/components/light/component.js
```diff
@@ -178,6 +54,842 @@ class LightComponent extends Component {
...
+        });
+    }
+
+    get cascadeDistribution() {
+        return this.data.cascadeDistribution;
+    }
+
+    /**
+     * Normal offset depth bias. Valid range is 0 to 1. Defaults to
+     * 0.
```
willeastcott, line 351:
```suggestion
     * Normal offset depth bias. Valid range is 0 to 1. Defaults to 0.
```

src/framework/components/light/component.js
```diff
@@ -178,6 +54,842 @@ class LightComponent extends Component {
...
+        });
+    }
+
+    get normalOffsetBias() {
+        return this.data.normalOffsetBias;
+    }
+
+    /**
+     * The range of the light. Affects omni and spot lights only. Defaults to
+     * 10.
```
willeastcott, line 367:
```suggestion
     * The range of the light. Affects omni and spot lights only. Defaults to 10.
```

src/framework/components/light/component.js
```diff
@@ -178,6 +54,842 @@ class LightComponent extends Component {
...
+        });
+    }
+
+    get outerConeAngle() {
+        return this.data.outerConeAngle;
+    }
+
+    /**
+     * Controls the rate at which a light attenuates from its position.
+     * Can be:
```
willeastcott, line 415:
```suggestion
     * Controls the rate at which a light attenuates from its position. Can be:
```

willeastcott (approved):
Approving with minor feedback.

src/framework/components/light/component.js
```diff
@@ -178,6 +54,839 @@ class LightComponent extends Component {
...
+
+    get light() {
+        return this.data.light;
+    }
+
+    /**
+     * The type of light. Can be:
+     *
+     * - "directional": A light that is infinitely far away and lights the entire scene from one
+     * direction.
```
mvaligursky, line 95: ideally instead of string types, we use the exported enums here

kpal81xd: I will prob go back and change fixup the types when upgrading the class again to move the majority of properties to the component instead of the data class

PR #6119, opened 2024-03-07
Examples build fix

kpal81xd:
- Fixed examples building issue with importing from subpath


kungfooman:
The comment above can then also be removed:

```js
// Don't include all of 'playcanvas' for these defines, it just
// causes bigger bundles and prolongs the build time by ~3s.
```

(not sure how it currently pans out)

PR #6123, opened 2024-03-07
ScriptType handlers type fix

kpal81xd:
- Embeds empty `ScriptType` handlers into class to remove post type gen fixup

### DIFF
--- a/src/framework/script/script-type.js
+++ b/src/framework/script/script-type.js
@@ -356,39 +356,37 @@ class ScriptType extends EventHandler {
     }
 
     /**
-     * @function
-     * @name ScriptType#[initialize]
-     * @description Called when script is about to run for the first time.
+     * Called when script is about to run for the first time.
      */
+    initialize() {}
 
     /**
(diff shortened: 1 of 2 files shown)


mvaligursky (approved):
Nice one!

willeastcott:
For the record, I'm worried about this. Imagine a scene with 10,000 entities. Every frame, all entity functions will now be called. That's a lot of calls to empty functions (where unlike C++, they'll still enter/exit presumably). It'll be cheap to call them, but can you say categorically this change doesn't reduce performance?

I'm also concerned that types of these functions are changing here. This may affect our TS devs.

Maksims:
This also will lead to an issue with `swap` method. Hot-reloading only should happen when `swap` method is defined, as it is up to the developer to inherit the state.
Unwanted reloads that will lead to exceptions will be very unwanted, especially in the context of collaborative development, where during development artist might get an error in Launcher at some point because the developer just saved a file.

PR #6136, opened 2024-03-11
Adding `typedocs` to gitignore

mvaligursky:
and removing `docs`, as a follow up on https://github.com/playcanvas/engine/pull/6132

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 .vscode/
 build
 coverage
-docs
+typedocs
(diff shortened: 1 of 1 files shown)


willeastcott:
This is not needed after https://github.com/playcanvas/engine/pull/6132

willeastcott:
The build doesn't generate a `typedocs` folder any more. It just generates a `docs` folder (the default for TypeDoc).

PR #6140, opened 2024-03-11
Removed class replacement typefix

kpal81xd:
Removed Class wrapper and uses `import` with `typeof` directly instead of finding and replacing in types

- [x] TS linter
- [x] API reference


willeastcott (approved):
This looks great...and you're certain the types and docs don't suffer any problems/changes because of this?

PR #6147, opened 2024-03-12
ES6 createScript 

kpal81xd:
- Modifies `scriptType` extension of `ScriptType` to use ES6 syntax

### DIFF
--- a/src/framework/script/script-create.js
+++ b/src/framework/script/script-create.js
@@ -65,19 +65,12 @@ function createScript(name, app) {
     if (reservedScriptNames.has(name))
         throw new Error(`Script name '${name}' is reserved, please rename the script`);
 
-    const scriptType = function (args) {
-        EventHandler.prototype.initEventHandler.call(this);
-        Script.prototype.initScriptType.call(this, args);
-    };
+    class ScriptWithAttributes extends Script {
+        attributes = new ScriptAttributes(ScriptWithAttributes);
+    }
 
-    scriptType.prototype = Object.create(Script.prototype);
-    scriptType.prototype.constructor = scriptType;
-
-    scriptType.extend = Script.extend;
-    scriptType.attributes = new ScriptAttributes(scriptType);
-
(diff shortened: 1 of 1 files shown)


src/framework/script/script.js
```diff
@@ -65,16 +64,13 @@ function createScript(name, app) {
     if (reservedScriptNames.has(name))
         throw new Error(`Script name '${name}' is reserved, please rename the script`);
 
-    const scriptType = function (args) {
-        EventHandler.prototype.initEventHandler.call(this);
-        ScriptType.prototype.initScriptType.call(this, args);
-    };
+    class scriptType extends ScriptType {}
```
willeastcott, line 67: What about something like:

```
    class ScriptTypeWithAttributes extends ScriptType {
        attributes = new ScriptAttributes(this);
    }

    registerScript(ScriptTypeWithAttributes, name, app);
    return ScriptTypeWithAttributes;
    ```

Comment (kpal81xd) [reply src/framework/script/script.js:67]: I did try that but I got a linting error for using `this` as a parameter. Seems to work fine tho if I just ignore the linting error

Comment (kpal81xd) [reply src/framework/script/script.js]: Actually so does the existing method - The linting just didnt show before 👍

**Review by marklundin — APPROVED** [review]

File: src/framework/script/script.js [hunk 2]
```diff
@@ -65,19 +64,13 @@ function createScript(name, app) {
...
-        ScriptType.prototype.initScriptType.call(this, args);
-    };
-
-    scriptType.prototype = Object.create(ScriptType.prototype);
-    scriptType.prototype.constructor = scriptType;
-
-    scriptType.extend = ScriptType.extend;
-    scriptType.attributes = new ScriptAttributes(scriptType);
+    class ScriptTypeWithAttributes extends ScriptType {
+        // @ts-ignore
```
willeastcott, line 68: Why is the `@ts-ignore` needed?

willeastcott (approved):

PR #6148, opened 2024-03-12
ScriptType handler embed

kpal81xd:
Embeds handlers for `ScriptType` into base class and checks for extended class methods using `Object.getPrototypeOf(instance).hasOwnProperty(SCRIPT_<method>)` 

- [x] TS linter
- [x] API reference

#### N.B
- handler types are now not optional as there is a base class implementation (will not be called if no override due to  `hasOwnProperty` check on extended class)


Maksims:
How does it handle inheritance?

````js
class Foo extends SctipType {
    initialize() {
        // ...
    }
    update(dt) {
        // ...
    }
}

class Bar extends Foo {
    initialize() {
        super();
        // ...
    }
}
````

If Bar is added as a script, will it call an update for its instances?

kpal81xd:
> How does it handle inheritance?
> 
> ```js
> class Foo extends SctipType {
>     initialize() {
>         // ...
>     }
>     update(dt) {
>         // ...
>     }
> }
> 
> class Bar extends Foo {
>     initialize() {
>         super();
>         // ...
>     }
> }
> ```
> 
> If Bar is added as a script, will it call an update for its instances?

It wont unfortunately so will most likely need to close this PR

Issue #6117, opened 2024-03-07
Implement KHR_materials_dispersion glTF extension support


mvaligursky:
https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_dispersion
https://github.com/KhronosGroup/glTF/pull/2340

<img width="764" alt="Screenshot 2024-03-07 at 10 52 17" src="https://github.com/playcanvas/engine/assets/59932779/46604319-0544-4c23-a49b-ab8e301bc81e">

![image](https://github.com/playcanvas/engine/assets/59932779/345554df-d13a-4f16-8a15-931050cdcf0b)

test model: https://github.com/cx20/gltf-test/tree/master/tutorialModels/DispersionTest

PR #6151, opened 2024-03-12
Support for KHR_materials_dispersion glTF extension

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/6117

Works on all platforms. Added new engine example.

Screenshot 2024-03-12 at 16 11 19


src/scene/materials/standard-material.js
```diff
@@ -279,6 +279,8 @@ let _params = new Set();
  * indices of refraction, the one around the object and the one of its own surface. In most
  * situations outer medium is air, so outerIor will be approximately 1. Then you only need to do
  * (1.0 / surfaceIor).
+ * @property {number} dispersion The strength of the angular separation of colors (chromatic
+ * aberration)transmitting through a volume. Defaults to 0, which is equivalent to no dispersion.
```
willeastcott, line 283:
```suggestion
 * aberration) transmitting through a volume. Defaults to 0, which is equivalent to no dispersion.
```

willeastcott (approved):
Looks great!

PR #6166, opened 2024-03-15
Support for dithering using interleaved gradient noise

mvaligursky:
A new type of noise, implemented without a texture lookup by directly generating the noise in the shader.

Comparison of noise patterns:
Screenshot 2024-03-15 at 11 43 49

It's a noise that's supposed to work well with TAA, as it's based on a 3x3 grid, which is what TAA typically uses for rejection.  Details: https://blog.demofox.org/2022/01/01/interleaved-gradient-noise-a-different-kind-of-low-discrepancy-sequence/

Adjusted the Dithered Transparency example to support this noise and also TAA for testing.


slimbuck (approved):

slimbuck:
nice!

slimbuck:
do you have any comparison screenshots old vs new?

mvaligursky:
> do you have any comparison screenshots old vs new?

Not really, it looks about the same :(

mvaligursky:
Or to be more specific:
- for static camera, both blue noise and IGN do great job, and TAA resolves both to almost pattern-free transparency (depending on the transparency level)
- but for moving camera, both noise types smudge the transparency about the same .. no major improvement from IGN at the moment - this is where I hoped to get an improvement, as that's where the 3x3 pixel rejection takes place.

PR #6169, opened 2024-03-15
ElementComponent properties port to get/set 

kpal81xd:
Port of `ElementComponent` from using `Object.defineProperty` to using `get/set`

- [x] TS linter
- [x] API reference

### DIFF
--- a/src/framework/asset/asset-localized.js
+++ b/src/framework/asset/asset-localized.js
@@ -13,10 +13,8 @@ class LocalizedAsset extends EventHandler {
         this._disableLocalization = false;
 
         /** @type {number} */
-        // @ts-ignore
         this._defaultAsset = null;
         /** @type {number} */
-        // @ts-ignore
         this._localizedAsset = null;
     }
 
--- a/src/framework/asset/asset-localized.js
+++ b/src/framework/asset/asset-localized.js
@@ -12,10 +12,17 @@ class LocalizedAsset extends EventHandler {
         this._autoLoad = false;
         this._disableLocalization = false;
 
+        /** @type {number} */
+        // @ts-ignore
         this._defaultAsset = null;
+        /** @type {number} */
+        // @ts-ignore
         this._localizedAsset = null;
     }
 
+    /**
+     * @param {Asset | number} value - The asset or id.
+     */
     set defaultAsset(value) {
         const id = value instanceof Asset ? value.id : value;
 
@@ -39,6 +46,9 @@ class LocalizedAsset extends EventHandler {
         return this._defaultAsset;
     }
 
+    /**
+     * @param {Asset | number} value - The asset or id.
+     */
     set localizedAsset(value) {
         const id = value instanceof Asset ? value.id : value;
         if (this._localizedAsset === id) {
@@ -48,7 +58,6 @@ class LocalizedAsset extends EventHandler {
         if (this._localizedAsset) {
             this._app.assets.off('add:' + this._localizedAsset, this._onLocalizedAssetAdd, this);
             this._unbindLocalizedAsset();
-            this._localizedAsset = null;
         }
 
         this._localizedAsset = id;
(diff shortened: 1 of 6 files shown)


src/framework/asset/asset-localized.js
```diff
@@ -12,10 +12,17 @@ class LocalizedAsset extends EventHandler {
...
+        /** @type {number} */
+        // @ts-ignore
         this._defaultAsset = null;
+        /** @type {number} */
+        // @ts-ignore
         this._localizedAsset = null;
     }
 
+    /**
+     * @param {Asset | number} value - The asset or id.
```
willeastcott, line 24: I think this is not public API so consider adding `@ignore`.

src/framework/asset/asset-localized.js
```diff
@@ -39,6 +46,9 @@ class LocalizedAsset extends EventHandler {
         return this._defaultAsset;
     }
 
+    /**
+     * @param {Asset | number} value - The asset or id.
```
willeastcott, line 50: I think this is not public API so consider adding `@ignore`.

src/framework/components/element/component.js
```diff
@@ -1044,26 +1002,753 @@ class ElementComponent extends Component {
...
+     *      -  `[color="#0000ff"]blue text[/color] `
+     * 2.  `outline ` - override the element's  `outlineColor ` and  `outlineThickness ` properties.
+     * Example:
+     *      -  `[outline color="#ffffff" thickness="0.5"]text[/outline] `
+     * 3.  `shadow ` - override the element's  `shadowColor ` and  `shadowOffset ` properties.
+     * Examples:
+     *      -  `[shadow color="#ffffff" offset="0.5"]text[/shadow] `
+     *      -  `[shadow color="#000000" offsetX="0.1" offsetY="0.2"]text[/shadow]
+     *
+     * ` Note that markup tags are only processed if the text element's  `enableMarkup ` property is
```
willeastcott, line 1383:
```suggestion
     *      -  `[shadow color="#000000" offsetX="0.1" offsetY="0.2"]text[/shadow]`
     *
     * Note that markup tags are only processed if the text element's  `enableMarkup ` property is
```

src/framework/components/element/component.js
```diff
@@ -1044,26 +1002,753 @@ class ElementComponent extends Component {
...
+     * set to true.
```
willeastcott, line 1384: This did not translate over from the original `@property` formatting correctly. Compare against the original...

src/framework/asset/asset-localized.js
```diff
@@ -12,10 +12,17 @@ class LocalizedAsset extends EventHandler {
         this._autoLoad = false;
         this._disableLocalization = false;
 
+        /** @type {number} */
+        // @ts-ignore
```
willeastcott, line 16: This PR does introduce quite a lot of `@ts-ignore`s into the codebase. What's the plan here?

kpal81xd: A lot of these classes will have to be changed to properly represent their types. I dont like having them either but its whole separate job it itself and I dont want to bundle it together in this PR.

src/framework/components/element/component.js
kpal81xd: Fixed

src/framework/components/element/component.js
```diff
@@ -806,7 +755,11 @@ class ElementComponent extends Component {
         this._setWidth(wr - wl);
 
         // update position
-        p.x = (this._localAnchor.z - this._localAnchor.x) - value - (this._calculatedWidth * (1 - this._pivot.x));
+        p.x =
+            this._localAnchor.z -
+            this._localAnchor.x -
+            value -
+            this._calculatedWidth * (1 - this._pivot.x);
```
willeastcott, line 762: I'm a bit unsure about all these Prettier edits. I mean, is it _really_ prettier? 🤔  That said, I don't feel super-strongly, so I'll just hit approve...

willeastcott (approved):

src/framework/components/element/component.js
kpal81xd: Yea a lot of that im reverting its just for the new code ive added a lot of this i need to go back over and revert

Issue #6045, opened 2024-02-07
WebXR issue on PICO Neo 3


Sycmasters:
I'm working on a WebXR experience, and everything was functioning well until the update to version 1.68.1 from 1.68. Since the update, all my WebXR projects have stopped working on the PICO Neo 3 headset. However, they continue to work on Meta Quest headsets. Even PC tutorials aren't working.

Here's a link to the tutorial I'm referencing. If this works, there shouldn't be any issues:
https://developer.playcanvas.com/tutorials/webxr-vr-lab/

![image](https://github.com/playcanvas/engine/assets/6709040/72a37466-391b-441d-b591-08bc62ab695e)

The problem occurs specifically on the PICO Browser.

### Steps to Reproduce
1. Open the PICO browser.
2. Navigate to [WebXR VR Lab Tutorial](https://developer.playcanvas.com/tutorials/webxr-vr-lab/).
3. Attempt to enter WebXR mode.

willeastcott:
Hey @Maksims, do you know what's new in 1.68.1 that could have caused this?

Maksims:
There were no changes between 1.68.0 and 1.68.1 that affected anything in the engine related to WebXR or feature detection.
I unfortunately do not have PICO device to test.

To ensure it is only a version related, I've made 2 builds of the VR Starter Kit:
1.68.0 https://playcanv.as/b/548eb056
1.68.1 https://playcanv.as/b/86ea0a35

@Sycmasters please let me know if you do see a problem between these two versions also. If not, then the issue must be due to something else.

Sycmasters:
Hi @Maksims

There's no difference in functionality between version 1.68.0 and 1.68.1; neither works for my project. I've tested it, and the last version that works correctly is 1.67.x.

![image](https://github.com/playcanvas/engine/assets/6709040/b5c5d976-c09c-40af-9038-cffab07f58b3)

So, the issue started with version 1.68.0, while everything operates as expected in version 1.67.x.

Maksims:
> Hi @Maksims
> 
> There's no difference in functionality between version 1.68.0 and 1.68.1; neither works for my project. I've tested it, and the last version that works correctly is 1.67.x.
> 
> So, the issue started with version 1.68.0, while everything operates as expected in version 1.67.x.

So the issue is with the 67 to 68 update. Do you have console logs?

Sycmasters:
[...]

PR #6181, opened 2024-03-18
WebXR XRSession.enabledFeatures missing workaround

Maksims:
WebXR Specs are updated periodically and it is up to vendors to keep up with updates.
Despite that, XRSession.enabledFeatures was introduced in the 24 August 2022 spec update, but not all platforms have updated their implementations since, leading to that property missing on some platforms.

This is an unfortunate reality of the web, but we can use alternative paths where it is not available.

This PR does exactly that. Instead of using the `enabledFetaures` list, when it is not available we have to detect if a feature is available in different ways, depending on a specific feature's capabilities. Docs already mention that the availability flag on a feature - does not necessarily available on the session start.

Fixes https://github.com/playcanvas/engine/issues/6045


src/framework/xr/xr-anchors.js
```diff
@@ -380,8 +388,24 @@ class XrAnchors extends EventHandler {
      * @ignore
      */
     update(frame) {
-        if (!this._available)
+        if (!this._available) {
+            // enabledFeatures - is not available, requires alternative way to check feature availability
+            if (!this.manager.session.enabledFeatures && !this._checkingAvailability) {
+                this._checkingAvailability = true;
+
+                frame.createAnchor(new XRRigidTransform(new Vec3(), new Quat()), this.manager._referenceSpace) // eslint-disable-line no-undef
```
willeastcott, line 396: I'm guessing Intellisense in VS Code is going to complain that you're passing in a `Vec3` and `Quat` here. But as it happens, the `XRRigidTransform` constructor parameters are optional and if you leave them unspecified, the equivalent effect will be made. So consider just deleting them (and the imports).

mvaligursky (approved):
This looks great, thanks!

src/framework/xr/xr-anchors.js
willeastcott: Also, instead of specifying `eslint-disable-line no-undef`, it's probably better to add `XRRigidTransform` to the ESLint ignore list in `package.json`.

PR #6183, opened 2024-03-18
A simple compute example, support for compute parameters

mvaligursky:
This is similar to how material instances keep their parameters. Those get pushed to the global scope during rendering, and the bind groups can be updated based on them.

This is now applied to instances of compute as well, allowing per compute parameters. This will allow us in separate PRs to batch multiple compute dispatches in a single compute pass, while allowing per compute parameters.


examples/src/examples/compute/texture-gen/config.mjs
```diff
@@ -0,0 +1,32 @@
+/**
+ * @type {import('../../../../types.mjs').ExampleConfig}
+ */
+export default {
+    HIDDEN: true,
+    WEBGPU_REQUIRED: true,
+    FILES: {
+        'compute-shader.wgsl': `
```
kpal81xd, line 8:
```suggestion
        'compute-shader.wgsl': /* wgsl */`
```

kpal81xd (approved):

examples/src/examples/compute/texture-gen/config.mjs
mvaligursky: any extension that supports it? I cannot find one.

kpal81xd: image

Issue #518, opened 2016-02-16
skinning doesn't work with low bone limit and no bone texture support


daredevildave:
If there is no bone texture support and skin splitting is required the skinning doesn't work properly (deformed).

To try this add these lines to Doom 3 Gangnam style or other project with skinned characters.

```
app.graphicsDevice.supportsBoneTextures = false;
app.graphicsDevice.boneLimit = 20;
```

willeastcott:
I just verified that the skin splitting does work. I suspect maybe you executed those lines after the model JSON has been parsed. To see the skin partitioner work, just have the texture_float extension return null and set the max vertex uniforms to 128. Then you'll get a bone limit of about 21. And it does actually work. The problem is that the engine has changed how it handles material mappings in the last few years. The skin partitioner does handle the mappings when they are passed in as a parameter. However, these days, the call has been changed to pass null. So materials are incorrectly assigned. And the extra mesh instances generated from the split simply get the default material:

![image](https://user-images.githubusercontent.com/697563/36958761-f2e814be-1ff2-11e8-9194-1b78a2b545a2.png)

The current skin splitter is really bad for another reason too. It manipulates the source JSON model format. It doesn't split the runtime format. This mean that if you load a glTF skinned model, say, the skin partitioner cannot split it.

I'm wondering if we should consider ripping it out and falling back to CPU skinning where necessary. I mean, we already have a CPU morphing path.

mvaligursky:
Another option here would be to fallback on the RGBA8/16 textures to store matrices. It would be lots more texture samples to get 4x3 matrix of floats, but should still be faster than CPU skinning.

These are the stats on vertex texture sampling support from about a year ago when webglstats still worked.
vertex texture reads: 4: 97.7%, 8: 79.9%, 16: 67.0%

willeastcott:
Wow, that's a cool idea, @mvaligursky. Deleting all that skin partitioning code from the engine would be great.

mvaligursky:
are we ok to drop support for up to 2% of devices (based on stats from a year ago) and do this? Or would we still need to do some backup CPU skinning or similar.

willeastcott:
[...]

PR #6188, opened 2024-03-19
Remove json model mesh splitting to work around uniform limits

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/518

- meshes are no longer split into larger number of meshes using smaller number of bones
- these days we have pretty much guaranteed 256 vectors, which equates to around 60 bones - which is a reasonable number for low level devices. This path is also only used on WebGL1.
- this partitioning code was used only for json models, and not glb

Screenshot 2024-03-19 at 15 29 38


willeastcott (approved):
I wrote this and I'm glad to see it go! 🚀

PR #6195, opened 2024-03-21
StandardMaterial types fixup

kpal81xd:
Adds types directly as attributes to class in `StandardMaterial`. Removes the need to inject types after in fixup file.

- [x] TS linter
- [x] API reference


src/scene/materials/standard-material.js
```diff
@@ -970,9 +2083,9 @@ const defineAggProp = (prop) => {
 
     const setterFunc = function (value) {
         const oldValue = this[internalName];
-        if (!oldValue.equals(value)) {
+        if (!oldValue?.equals(value)) {
```
willeastcott, line 2086: What in this PR requires this function to be updated? Just curious.

kpal81xd: That probably wont be needed anymore actually I will double check

src/scene/materials/standard-material.js
```diff
@@ -563,6 +60,1618 @@ class StandardMaterial extends Material {
...
     userAttributes = new Map();
 
+    /**
+     * Used to fade out materials when {@link StandardMaterial#opacityFadesSpecular} is set to false.
+     *
+     * @type {boolean}
+     */
+    alphaFade;
+
+
```
willeastcott, line 70: Can you please go through and make sure only 1 line separates class field declarations please (not 2)?

src/scene/materials/standard-material.js
kpal81xd: So I found out this was because when defining the properties in the class with no value that is actually defining them with `undefined` and triggers the set handler before the default internal properties are set with `reset`.

willeastcott: Oh...maybe consider adding a comment about that in the code then.

kpal81xd:
This PR will not be required in this form once shader compilation has been reworked

PR #6201, opened 2024-03-25
Implementation of StorageBuffer on WebGPU

mvaligursky:
related to https://github.com/playcanvas/engine/pull/6009

- implementation of the StorageBuffer class, to be used with WebGPU's Compute for read/write data.

intended public API
```
StorageBuffer.constructor(device, byteSize, bufferUsage);
StorageBuffer.destroy();

// clear part of the buffer to 0
StorageBuffer.clear(offset, size);

// async read of the GPU buffer
StorageBuffer.read(offset, size, allocatedTypeArray) : Promise<ArrayBuffer>;

// async upload (not implemented yet, follow up PR)
StorageBuffer.write(data, offset);
```

- Compute now handles StorageBuffer as input / output. This will be slightly expanded in the follow up PRs (for example read-only buffer)
- new example which demonstrate how to create a compute shader that generates a histogram, from the rendered image, downloads the results to CPU and renders it using lines. This is not the most performant implementation, as it requires async data download to the CPU, but this example demonstrates exactly that case, where a result of compute is needed on CPU. We'll have another example where compute generated storage buffer gets consumed by the GPU directly vertex of fragment shader).

Screenshot 2024-03-25 at 16 21 10

### DIFF
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -96,6 +96,14 @@ export const TRACEID_VRAM_VB = 'VRAM.Vb';
 export const TRACEID_VRAM_IB = 'VRAM.Ib';
 
 /**
+ * Logs the vram use by the storage buffers.
+ *
+ * @type {string}
+ * @category Debug
+ */
+export const TRACEID_VRAM_SB = 'VRAM.Sb';
+
+/**
  * Logs the creation of bind groups.
  *
  * @type {string}
--- a/src/platform/graphics/compute.js
+++ b/src/platform/graphics/compute.js
@@ -60,7 +60,7 @@ class Compute {
      * @param {import('./shader.js').Shader} shader - The compute shader.
      * @param {string} [name] - The name of the compute instance, used for debugging only.
      */
-    constructor(graphicsDevice, shader, name = 'Noname') {
+    constructor(graphicsDevice, shader, name = 'Unnamed') {
         this.device = graphicsDevice;
         this.shader = shader;
         this.name = name;
--- a/src/platform/graphics/compute.js
+++ b/src/platform/graphics/compute.js
@@ -24,6 +24,9 @@ class Compute {
      */
     shader = null;
 
+    /** @type {string} */
+    name;
+
     /**
      * @type {Map<string, ComputeParameter>}
      * @ignore
@@ -55,10 +58,12 @@ class Compute {
      * @param {import('./graphics-device.js').GraphicsDevice} graphicsDevice -
      * The graphics device.
      * @param {import('./shader.js').Shader} shader - The compute shader.
+     * @param {string} [name] - The name of the compute instance, used for debugging only.
      */
-    constructor(graphicsDevice, shader) {
+    constructor(graphicsDevice, shader, name = 'Noname') {
         this.device = graphicsDevice;
         this.shader = shader;
+        this.name = name;
 
         if (graphicsDevice.supportsCompute) {
             this.impl = graphicsDevice.createComputeImpl(this);
@@ -69,8 +74,8 @@ class Compute {
      * Sets a shader parameter on a compute instance.
      *
      * @param {string} name - The name of the parameter to set.
-     * @param {number|number[]|Float32Array|import('./texture.js').Texture} value - The value for
-     * the specified parameter.
+     * @param {number|number[]|Float32Array|import('./texture.js').Texture|import('./storage-buffer.js').StorageBuffer} value
+     * - The value for the specified parameter.
      */
     setParameter(name, value) {
         let param = this.parameters.get(name);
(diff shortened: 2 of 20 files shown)


src/core/constants.js
```diff
@@ -95,6 +95,14 @@ export const TRACEID_VRAM_VB = 'VRAM.Vb';
  */
 export const TRACEID_VRAM_IB = 'VRAM.Ib';
 
+/**
+ * Logs the vram use by the storage buffers.
```
willeastcott, line 99: Does this need a `(WebGPU only)`?

src/platform/graphics/compute.js
```diff
@@ -55,10 +58,12 @@ class Compute {
      * @param {import('./graphics-device.js').GraphicsDevice} graphicsDevice -
      * The graphics device.
      * @param {import('./shader.js').Shader} shader - The compute shader.
+     * @param {string} [name] - The name of the compute instance, used for debugging only.
      */
-    constructor(graphicsDevice, shader) {
+    constructor(graphicsDevice, shader, name = 'Noname') {
```
willeastcott, line 63: Maybe `Unnamed` would be better. We already use that for GraphNodes.

src/platform/graphics/storage-buffer.js
```diff
@@ -0,0 +1,69 @@
+import { Debug } from '../../core/debug.js';
+import { TRACEID_VRAM_SB } from '../../core/constants.js';
+import { BUFFERUSAGE_STORAGE } from './constants.js';
+
+let id = 0;
+
+/**
+ * A storage buffer represents a memory which both the CPU and the GPU can access. Typically it is
+ * used to provide data for compute shader, and to store the result of the computation.
+ * Note that this class is only supported on WebGPU platform.
```
willeastcott, line 10:
```suggestion
 * Note that this class is only supported on the WebGPU platform.
```

src/core/constants.js
mvaligursky: I would report 0 on other platforms, so it works.

src/platform/graphics/webgpu/webgpu-buffer.js
```diff
@@ -42,18 +54,18 @@ class WebgpuBuffer {
 
         if (!this.buffer) {
             // size needs to be a multiple of 4
+            // note: based on specs, descriptor.size must be a multiple of 4 If descriptor.mappedAtCreation is true
```
willeastcott, line 57:
```suggestion
            // note: based on specs, descriptor.size must be a multiple of 4 if descriptor.mappedAtCreation is true
```

src/platform/graphics/webgpu/webgpu-graphics-device.js
```diff
@@ -817,6 +823,104 @@ class WebgpuGraphicsDevice extends GraphicsDevice {
...
+
+    /**
+     * Read a content of a storage buffer.
+     *
+     * @param {import('./webgpu-buffer.js').WebgpuBuffer} storageBuffer - The storage buffer.
+     * @param {number} [offset] - The offset of data to read. Defaults to 0.
+     * @param {number} [size] - The size of data to read. Defaults to the full size of the buffer.
+     * @param {ArrayBufferView} [data] - Typed array to populate with the data read from the storage
+     * buffer. When typed array is supplied, enough space needs to be reserved, otherwise only
+     * partial data is copied. If not specified, a data is returned in an Uint8Array. Defaults to
```
willeastcott, line 858:
```suggestion
     * partial data is copied. If not specified, the data is returned in an Uint8Array. Defaults to
```

src/platform/graphics/compute.js
mvaligursky: GraphNode uses `Untitled`, and that looks strange in logging where I log names. Happy with `Unnamed`.

willeastcott (approved):
Approving with minor comments. Nice! 👏

src/platform/graphics/webgpu/webgpu-graphics-device.js
```diff
@@ -817,6 +823,104 @@ class WebgpuGraphicsDevice extends GraphicsDevice {
...
+        const commandEncoder = this.commandEncoder ?? this.wgpu.createCommandEncoder();
+
+        commandEncoder.clearBuffer(storageBuffer.buffer, offset, size);
+
+        // if we created the encoder
+        if (!this.commandEncoder) {
+            DebugHelper.setLabel(commandEncoder, 'ReadStorageBuffer-Encoder');
+            const cb = commandEncoder.finish();
+            DebugHelper.setLabel(cb, 'ReadStorageBuffer-CommandBuffer');
+            this.addCommandBuffer(cb);
```
erikdubbelboer, line 846: Why not just `this.wgpu.queue.submit([cb]);`. This doesn't need to happen in sync with the normal frame rate?

mvaligursky: Well that depends, but there are situation where this needs to happen in sync, so this is the more generic behaviour. As mentioned elsewhere, we can relax it later in some cases.

PR #6206, opened 2024-03-27
Removed undollar script

kpal81xd:
Extracted out texture parser into separate class and extended each parser from abstract class

- [x] TS linter
- [x] API reference


src/index.js
```diff
@@ -337,14 +337,15 @@ export { ShaderHandler } from './framework/handlers/shader.js';
...
-export { TextureHandler, TextureParser } from './framework/handlers/texture.js';
+export { TextureHandler } from './framework/handlers/texture.js';
 export { TextureAtlasHandler } from './framework/handlers/texture-atlas.js';
 
 // FRAMEWORK / INPUT
 export { ElementInput, ElementInputEvent, ElementMouseEvent, ElementSelectEvent, ElementTouchEvent } from './framework/input/element-input.js';
 
 // FRAMEWORK / PARSERS
 export { JsonStandardMaterialParser } from './framework/parsers/material/json-standard-material.js';
+export { TextureParser } from './framework/parsers/texture/texture.js';
```
mvaligursky, line 348: Not sure we need to expose it, it seems the class and its descendants are all `@ignore`.

src/framework/parsers/texture/texture.js
```diff
@@ -0,0 +1,45 @@
+/**
+ * @interface
+ * @name TextureParser
+ * @description Interface to a texture parser. Implementations of this interface handle the loading
+ * and opening of texture assets.
+ *
+ * @category Graphics
+ */
+class TextureParser {
```
mvaligursky, line 9: Should this be `@ignore`? Or is it used anywhere in the public API.

src/index.js
kpal81xd: Yea i wasnt sure since it was exposed before.

kpal81xd: Might be used in the editor thats the ony thing I can think of?

mvaligursky (approved):
Great change!

src/framework/parsers/texture/texture.js
```diff
@@ -0,0 +1,43 @@
...
+     * @description Load the texture from the remote URL. When loaded (or failed),
+     * use the callback to return an the raw resource data (or error).
+     * @param {object} url - The URL of the resource to load.
+     * @param {string} url.load - The URL to use for loading the resource.
+     * @param {string} url.original - The original URL useful for identifying the resource type.
+     * @param {import('../../../framework/handlers/handler.js').ResourceHandlerCallback} callback - The callback used when
+     * the resource is loaded or an error occurs.
+     * @param {import('../../asset/asset.js').Asset} [asset] - Optional asset that is passed by
+     * ResourceLoader.
+     */
```
willeastcott, line 21: This can be converted to a 'modern' JSDoc block now (e.g. drop `@function`, `@description` and `@name` tags).

src/framework/parsers/texture/texture.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * Interface to a texture parser. Implementations of this interface handle the loading
```
willeastcott, line 2: I'd change the description to 'Base class' instead of interface now.

PR #6208, opened 2024-03-27
Clear vertex buffer on empty instance buffer inside drawInstance

heretique:
This PR handles a corner case that we found when doing instanced draw calls.
Inside the `Renderer.drawInstance` method if the `instancingData.count` is zero there is no `device.draw` called which leaves the `GraphicsDevice.vertexBuffers` list uncleared (that list is only cleared on a successful draw call) thus making next draw call to bind more vertex buffers then needed (plus create extra temporary VAOs) and in some cases if the material changes from a draw call to the next and it has different attribute requirements it breaks the draw call (sometimes with visible artifacts) and WebGL logs warnings:

```
FIREFOX:  WebGL warning: drawElementsInstanced: Indexed vertex fetch requires 462 vertices, but attribs only supply 376.
EDGE: [.WebGL-0x11c0055cb00] GL_INVALID_OPERATION: Vertex buffer is not big enough for the draw call.
```

It took a long time to figure this one out so we're hoping this PR would benefit others who also rely a lot on instancing.

image


src/platform/graphics/graphics-device.js
```diff
@@ -661,6 +661,14 @@ class GraphicsDevice extends EventHandler {
         }
     }
 
+    /**
+     * Clears the vertex buffer set on the graphics device. This is called automatically by the
+     * renderer.
+     */
```
mvaligursky, line 667: it needs `@ignore` tag to not make it a public API

heretique: Oh, missed that, I'll take care of it.

kpal81xd (approved):

mvaligursky (commented):
Looks great, thanks for this!

PR #6219, opened 2024-04-03
Engine build rework

kpal81xd:
### Overview

This PR aims to clean up and optimize the building process of the engine. This includes: 

- [ ] Moved types build entirely to rollup (custom plugins to handle generation and fixup)
- [ ] Moving engine types output to be contained within release build (technically don't need to specify types file directly can use unbundled types)
- [ ]  Combined babel options into a single function
- [ ] Bundled ES6 builds now use unbundled ES6 builds as input (can skip rerunning plugins)
- [ ] Minifed build type uses release build as input (will only apply output plugins e.g terser)

### Results (on M1 Macbook Pro)
- ~1.5s for each bundled build (from >5s)
- 5s for minified build (from >10s)


willeastcott:
Random comment: in the build system, we have had `es5` and `es6` and names for the build targets....but I don't think that's accurate. I think `umd` and `esm` are more accurate, no?

mvaligursky:
I'd be great to compare build file sizes before and after to make sure we get the same(ish) results.

kpal81xd:
> Random comment: in the build system, we have had `es5` and `es6` and names for the build targets....but I don't think that's accurate. I think `umd` and `esm` are more accurate, no?

Yep I agree

kpal81xd:
> I'd be great to compare build file sizes before and after to make sure we get the same(ish) results.

Yea im gonna do some file comparisons just ensure nothing drastic was changed in the build process

package.json
```diff
@@ -18,17 +18,13 @@
   ],
   "license": "MIT",
   "main": "build/playcanvas.js",
-  "module": "build/playcanvas.mjs",
-  "types": "build/playcanvas.d.ts",
+  "module": "build/playcanvas/index.js",
```
willeastcott, line 21: So I see the `types` property has been dropped from the `package..json`. How are types to be found by clients of the package now?

package.json
```diff
@@ -83,25 +79,9 @@
...
-    "build/playcanvas-extras/index.js",
+    "build/playcanvas*",
     "scripts",
-    "LICENSE",
-    "package.json",
-    "README.md",
-    "README-ja.md",
-    "README-kr.md",
-    "README-zh.md"
+    "README*.md"
```
willeastcott, line 84: So presumably the `LICENSE` and `package.json` are implicitly included and don't need to be explicitly listed here?

utils/rollup-build-target.mjs
```diff
@@ -50,144 +49,191 @@ const stripFunctions = [
...
-        name: 'pc',
-        preserveModules: !shouldBundle
-    };
+    return plugins;
+}
 
-    const loc = `${buildDir}/${outputFile[buildType]}${outputExtension[moduleFormat]}`;
-    outputOptions[shouldBundle ? 'file' : 'dir'] = loc;
+/**
+ * Build a target that rollup is supposed to build (bundled and unbundled).
```
willeastcott, line 141:
```suggestion
 * Build a target that Rollup is supposed to build (bundled and unbundled).
```

utils/rollup-build-target.mjs
```diff
@@ -50,144 +49,191 @@ const stripFunctions = [
...
+ *
+ * @param {object} options - The build target options.
+ * @param {'umd'|'esm'} options.moduleFormat - The module format.
+ * @param {'debug'|'release'|'profiler'|'min'} options.buildType - The build type.
+ * @param {string} [options.input] - Only used for Examples to change it to `../src/index.js`.
```
willeastcott, line 146:
```suggestion
 * @param {string} [options.input] - Only used for examples to change it to `../src/index.js`.
```

utils/rollup-script-target.mjs
```diff
@@ -3,38 +3,71 @@ import { babel } from '@rollup/plugin-babel';
...
-import { es5Options } from './rollup-es5-options.mjs';
-import { spacesToTabs } from './rollup-spaces-to-tabs.mjs';
+import { babelOptions } from './rollup-babel-options.mjs';
+import { spacesToTabs } from './plugins/rollup-spaces-to-tabs.mjs';
 
 /** @typedef {import('rollup').RollupOptions} RollupOptions */
 
 /**
- * Build an ES5 target that rollup is supposed to build.
+ * Build a target that rollup is supposed to build (bundled and unbundled).
```
willeastcott, line 12:
```suggestion
 * Build a target that Rollup is supposed to build (bundled and unbundled).
```

willeastcott (approved):
This is a wonderful refinement of the build system. 👏

package.json
kpal81xd: So the unbundled types are now in the same directory as the unbundled so they are recognised when importing from there. If you were to use the UMD build the types are infered directly from playcanvas.d.ts since its name matches the package name. I might still include `types` just in case some issues arise using older versions of node or npm

PR #6226, opened 2024-04-04
Compute shaders can read / write index and vertex buffers

mvaligursky:
adjusted APIs to allow vertex / index buffer to be marked as accessible by compute shaders:

- **IndexBuffer** and **VertexBuffer** constructor now take options parameter, and can specify:
    options.storage = true;

- **Mesh** constructor now takes options parameter, which can be:
    options.storageVertex = true;
    options.storageIndex = true;

- **procedural API** - createMesh, createSphere and similar have additional valid options in the options class:
    options.storageVertex = true;
    options.storageIndex = true;

- vertex or index buffer can be attached to a compute shader:
```
    compute.setParameter('vb', mesh.vertexBuffer);
```

**Example modifying the vertex buffer:**

https://github.com/playcanvas/engine/assets/59932779/1914d4b3-1a14-4e7f-8b0c-71812fa396b8


erikdubbelboer (approved):

src/platform/graphics/vertex-buffer.js
```diff
@@ -21,8 +21,11 @@ class VertexBuffer {
      * @param {number} numVertices - The number of vertices that this vertex buffer will hold.
      * @param {number} [usage] - The usage type of the vertex buffer (see BUFFER_*). Defaults to BUFFER_STATIC.
      * @param {ArrayBuffer} [initialData] - Initial data.
+     * @param {object} [options] - Object for passing optional arguments.
+     * @param {boolean} [options.storage] - Defines if the vertex buffer can be used as a storage
+     * buffer by a compute shader. Defaults to false. Only supported on WebGPU.
      */
-    constructor(graphicsDevice, format, numVertices, usage = BUFFER_STATIC, initialData) {
+    constructor(graphicsDevice, format, numVertices, usage = BUFFER_STATIC, initialData, options) {
```
willeastcott, line 28: Hmm. This constructor signature is getting a bit messy. If you pass options, you need to set a value of `initialData`. which a lot of devs never pass. Feels like if this was written from scratch, `usage` and `initialData` would be in `options`. IIRC, `initialData` was added originally as a bit of a hack...

src/scene/procedural.js
```diff
@@ -232,6 +232,10 @@ function calculateTangents(positions, normals, uvs, indices) {
  * @param {number[]} [opts.blendWeights] - An array of 4-dimensional bone weights where each
  * component is in the range 0 to 1 and the sum of the weights should equal 1.
  * @param {number[]} [opts.indices] - An array of triangle indices.
+ * @param {boolean} [opts.storageVertex] - Defines if the vertex buffer of the mesh can be used as
+ * a storage buffer by a compute shader. Defaults to false. Only supported on WebGPU.
+ * @param {boolean} [opts.storageIndex] - Defines if the index buffer of the mesh can be used as
+ * a storage buffer by a compute shader. Defaults to false. Only supported on WebGPU.
```
willeastcott, line 238: I'm not sure why I feel uneasy about adding new WebGPU-only options to this API. I almost feel like shape creation should be very simple and pure. Like maybe it should return just bare position, normal, uv and index arrays. And you create the mesh yourself in whatever form you like (e.g. whether you want it to be used by a compute shader etc).

src/platform/graphics/vertex-buffer.js
mvaligursky: I agree, but the devs can pass `uninitialised` for those to get the default values.
But maybe it's time we move last 3 or so parameters to options?

mvaligursky: Looking at this, I think a separate PR would be a good idea here too.

willeastcott (approved):

willeastcott:
Approving despite my reservations - but sounds like things can be addressed in follow up PRs. 😄

PR #6228, opened 2024-04-08
Extras merged into engine

kpal81xd:
- Adds extras to engine source (paths all relative now)
- Remaps pcx to pc in UMD build (legacy support)
- Removes adding extras to examples (examples includes extras types

#### N.B
- fflate now included in build

#### File sizes
- `playcanvas.min.js` 1.6MB -> 1.7MB 
- `playcanvas.min.mjs` 1.5MB -> 1.6MB
- `playcanvas.dbg.js` 17.1MB -> 18.1MB
- `playcanvas.dbg.mjs` 15.1MB -> 15.8MB
- `playcanvas.prf.js` 2.9MB -> 3MB
- `playcanvas.prf.mjs` 2.5MB -> 2.6MB
- `playcanvas.js` 2.9MB -> 3MB
- `playcanvas.mjs` 2.5MB -> 2.6MB
- `gzip` 12.7MB -> 13.3MB (N.B. includes unbundled ESM builds)

#### TODO
- [ ] handle backwards compatibility in editor


mvaligursky:
What's the size impact for the UMD build, especially minified, and also gzipped version of it.

mvaligursky:
Also, should we add `@category` tags for the API docs to be well organized?

kpal81xd:
> Also, should we add `@category` tags for the API docs to be well organized?

Like extras category on top of the existing categories?

mvaligursky:
> > Also, should we add `@category` tags for the API docs to be well organized?
> 
> Like extras category on top of the existing categories?

Not sure what would be the best. Maybe even `Exporters`, `Gizmo`, `RenderPasses` and similar? What do you think @willeastcott ?

willeastcott:
I think RenderPasses should be under Graphics. But the rest is fine.

mvaligursky:
Would you have this in kB? I'd like to see if this is 10k or more like a 100k difference, which is hard to see with the rounding:
`playcanvas.min.js 1.6MB -> 1.7MB`

Also, this specific target only gzip comparison as well please.

kpal81xd:
> Would you have this in kB? I'd like to see if this is 10k or more like a 100k difference, which is hard to see with the rounding: `playcanvas.min.js 1.6MB -> 1.7MB`
> 
> Also, this specific target only gzip comparison as well please.

`playcanvas.min.js` 1616650B -> 1709861B
`playcanvas.min.js (gzip)` 406757B -> 433084B

marklundin:
Are extras only added to the UMD build or ESM too?

kpal81xd:
> Are extras only added to the UMD build or ESM too?

Both

slimbuck:
I can't remember now, what are the reasons behind this change? We use fflate only when exporting usdz models. Adding all this to the core engine seems... wrong.

kpal81xd:
> I can't remember now, what are the reasons behind this change? We use fflate only when exporting usdz models. Adding all this to the core engine seems... wrong.

Simplifies out entire build process, also allows extras to work properly with the standalone engine and will include the extras types too

mvaligursky:
The end goal is to use ESM build everywhere and tree-shaking, so the size increase won't be a problem longer term.
The question is if we're ok with this till then. Gzipped minified build is 26k larger, which is an increase of 6.4%. To me this is acceptable, 26k is a small cost to pay temporarily.

kpal81xd:
> The end goal is to use ESM build everywhere and tree-shaking, so the size increase won't be a problem longer term. The question is if we're ok with this till then. Gzipped minified build is 26k larger, which is an increase of 6.4%. To me this is acceptable, 26k is a small cost to pay temporarily.

So is this good to merge or does it need more discussion?

mvaligursky:
> I know this is largely semantics, but should the extras not live inside src if we're folding it into the engine?

yep, notice the renames, for example:
`extras/mini-stats/mini-stats.js → src/extras/mini-stats/mini-stats.js`

utils/plugins/rollup-import-validation.mjs
```diff
@@ -14,7 +14,8 @@ export function engineLayerImportValidation(rootFile) {
         'core': 0,
         'platform': 1,
         'scene': 2,
-        'framework': 3
+        'framework': 3,
+        'extras': 4
```
kpal81xd, line 18: @mvaligursky This is what you meant by adding the layer for the extras?

mvaligursky: yep perfect.

mvaligursky:
Note that before the engine with this change is released, we need to release the Editor to handle the lack of extras, in a backwards compatible way.

mvaligursky (approved):
Looks great!

willeastcott (approved):

PR #6229, opened 2024-04-09
ESM attributes rename

marklundin:
Currently ESM Scripts attributes are defined as follows

```javascript
class MyScript extends ScriptType {
    static attributesDefinition = { ... }
}
```

The static `attributesDefinition` is both verbose and confusingly names as we already use `attributes`. This PR uses the existing static `attributes` property which makes for a cleaner, more intuitive interface, and helps users migrate

```javascript
class MyScript extends ScriptType {
    static attributes = { ...  }
}
```

When the script is registered, this is used to populate the ScriptAttributes class and so becomes equivalent to using MyScript.attributes.add(...)`. Ini addition the `attributes` property [can already be](https://github.com/playcanvas/engine/blob/4ab0aceb0a4440eebb854a6e24e07dcf02805d66/src/framework/script/script-type.js#L253) a [plain object](https://github.com/playcanvas/engine/blob/4ab0aceb0a4440eebb854a6e24e07dcf02805d66/src/framework/script/script-type.js#L305), so this doesn't change things significantly, but it improves the overall usability of the class.

### DIFF
--- a/src/framework/handlers/script.js
+++ b/src/framework/handlers/script.js
@@ -155,13 +155,13 @@ class ScriptHandler extends ResourceHandler {
                 const extendsScriptType = scriptClass.prototype instanceof ScriptType;
 
                 if (extendsScriptType) {
-                    const attributes = new ScriptAttributes(scriptClass);
                     if (scriptClass.attributes) {
+                        const attributes = new ScriptAttributes(scriptClass);
                         for (const key in script.attributes) {
                             attributes.add(key, scriptClass.attributes[key]);
                         }
+                        scriptClass.attributes = attributes;
                     }
-                    scriptClass.attributes = attributes;
                     registerScript(scriptClass, scriptClass.name.toLowerCase());
                 }
             }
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

marklundin:
Any thoughts on this @Maksims too?

Maksims:
How about this:
```javascript
class MyScript extends ScriptType { }
MyScript.attributes.add('attr', { type: 'text' });
```

This is a current way of defining attributes and would be the easiest way to migrate from non-ESM scripts.
I do agree that having a static property on a class would make it nice too, but thought of it before and we have this unfortunate collision of names here.

marklundin:
I'm following the [decorators proposal](https://github.com/tc39/proposal-decorators) currently at stage 3. I'm hoping it goes all the way but we'll have to wait and see. It looks like a great candidate for script attributes if it does land 💪 

The `MyScript.attributes.add('attr', { type: 'text' });` syntax should still work, but this PR is as just an alternative to the current `attributesDefinition` which I think works for now. The ordering point is valid, but IIRC it's only numeric keys that get sorted, the rest should retain insertion order.

Maksims:
Then it all sounds great!

PR #6233, opened 2024-04-10
Particles example using compute shader simulation

mvaligursky:
- feature to upload data to compute buffer
- example that simulates 1M of particles using compute shader, renders them using WGSL vertex / fragment shader by reading data from the simulated storage buffer

screenshot:
Screenshot 2024-04-10 at 11 24 41

very heavily compressed video, destroying the visuals completely:

https://github.com/playcanvas/engine/assets/59932779/d73428f4-04ca-4f17-93f7-028c611b5cf6

### DIFF
--- a/examples/src/examples/compute/particles/config.mjs
+++ b/examples/src/examples/compute/particles/config.mjs
@@ -1,0 +1,145 @@
+/**
+ * @type {import('../../../../types.mjs').ExampleConfig}
+ */
+export default {
+    HIDDEN: true,
+    WEBGPU_REQUIRED: true,
+    FILES: {
+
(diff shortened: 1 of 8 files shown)


Maksims:
The codec required to play that video is not available by default on Windows.

mvaligursky:
> The codec required to play that video is not available by default on Windows.

Swapped to h264 .. but this looks even worse ;)

LeXXik:
What a blurry mess :D Thank you for the PR, though!

mvaligursky:
I tried few options .. but particles are so small that their are invisible when the browser is scaled down (likely on mobile too). Even just making them round. I tried making them larger, but then it's just a solid color, hard to tell particles apart.

PR #6237, opened 2024-04-10
Add Script alias

marklundin:
Having some discussions with @willeastcott, with the new ESM Script, we want to keep the default script boilerplate as clean as possible. This PR exports an alias for `ScriptType` named `Script`

```javascript
import { Script } from 'playcanvas';

class Rotator extends Script {
  initialize() {
    ...
  }
}
```

UPDATE:
ScriptType has been added to the deprecation.js file
Improved dicunmentation for Script to include ESM example


kpal81xd:
What was the reason for this?

mvaligursky (dismissed):
Looks good.
Should we rename `ScriptType` to `Script` in that module, and export it as `ScriptType` for compatibility?

willeastcott:
> Should we rename ScriptType to Script in that module, and export it as ScriptType for compatibility?

Yeah, because otherwise, the API ref will continue to show `ScriptType`. The aliasing should be done in the backwards compatibility module.

marklundin:
@willeastcott are we happy to simply rename the class and keep the `script-type.js` filename?

mvaligursky:
Rename filename as well, we keep those in sync.

willeastcott:
Maybe do:

* `script.js` -> `script-create.js` (or `script-register.js`)
* `script-type.js` -> `script.js`

willeastcott:
BTW, we should consider deprecating and hiding `registerScript`.

mvaligursky:
We also need to consider the impact in the Editor. Will the Editor be able to still use at least few previous engine versions? Do we need to add some type checking to it or similar? Maybe it all just works, but we need to check.

marklundin:
> * `script.js` -> `script-create.js` (or `script-register.js`)
> * `script-type.js` -> `script.js`

Makes sense, although strictly speaking this would be a breaking change right?

> We also need to consider the impact in the Editor. Will the Editor be able to still use at least few previous engine versions? Do we need to add some type checking to it or similar? Maybe it all just works, but we need to check.

Yeah it should be fine, editor only uses ScriptType when parsing and `pc.Script === pc.ScriptType`

marklundin:
Actually maybe not a breaking change since we haven't yet [published the ESM build ](https://www.npmjs.com/package/playcanvas?activeTab=code). Mainly thinking about `import { Script } from 'playcanvas/framework/script/script-type.js'

marklundin:
This has not been updated

- Renamed script.js -> script-create.js
- Renamed script-type.js -> script.js
- Updated jsdoc references of {ScriptType} => {Script}

Verified with type docs...
image

*Note that now ScriptType does not exist, it does not generate a corresponding url. It would be great to retain the original, to indicated to users that it has been renamed

src/framework/script/script.js
```diff
@@ -1,156 +1,393 @@
...
+     * @description Called when a Script that already exists in the registry
+     * gets redefined. If the new Script has a `swap` method in its prototype,
+     * then it will be executed to perform hot-reload at runtime.
+     * @param {Script} old - Old instance of the script to copy data to the new instance.
+     */
 }
-/* eslint-enable jsdoc/check-examples */
 
-export { createScript, registerScript, getReservedScriptNames };
+export { Script, Script as ScriptType };
```
willeastcott, line 393: I would handle the aliasing in `deprecated.js`.

marklundin: You mean adding `export { Script as ScriptType }` to deprecated.js?

willeastcott: I was thinking something like: https://github.com/playcanvas/engine/blob/c5177982f9c9064df1d5a881a513a302054d60ae/src/deprecated/deprecated.js#L725

src/framework/script/script.js
```diff
@@ -7,11 +7,20 @@ import { ScriptAttributes } from './script-attributes.js';
...
+ * from the engine.
  *
- * The type is to be extended using its JavaScript prototype. There is a list of methods that will
- * be executed by the engine on instances of this type, such as:
+ * You can create a Script using either {@link createScript} or by extending the class directly.
+ *
+ * ```javascript
+ * class Rotator extends Script {
+ *    update() {
+ *      this.entity.rotate(0, 0.1, 0);
```
Maksims, line 20: Should be 4 spaces instead of 2.

src/framework/script/script-create.js
```diff
@@ -0,0 +1,156 @@
+import { Debug } from '../../core/debug.js';
+import { EventHandler } from '../../core/event-handler.js';
+
+import { script } from '../script.js';
+import { AppBase } from '../app-base.js';
+
+import { ScriptAttributes } from './script-attributes.js';
+import { ScriptType } from './script.js';
```
willeastcott, line 8: `ScriptType` -> `Script`

src/framework/script/script-create.js
```diff
@@ -0,0 +1,156 @@
...
+        Debug.error('This project is using the legacy script system. You cannot call pc.createScript().');
+        return null;
+    }
+
+    if (reservedScriptNames.has(name))
+        throw new Error(`Script name '${name}' is reserved, please rename the script`);
+
+    const scriptType = function (args) {
+        EventHandler.prototype.initEventHandler.call(this);
+        ScriptType.prototype.initScriptType.call(this, args);
```
willeastcott, line 70: `ScriptType` -> `Script`

marklundin:
jsdocs update

image

marklundin:
Any more updates on this @willeastcott? Keen to get this merged

PR #6240, opened 2024-04-10
Refactored VertexBuffer constructor to hide more params into options

mvaligursky:


slimbuck (approved):
This looks great, though I'd probably remove all the `usage: BUFFER_STATIC` since that's the default.

LeXXik:
Is this a breaking change?

mvaligursky:
It's not, old API is supported but prints a deprecated warning.

PR #6244, opened 2024-04-11
WebXR fix overlaping rAF callbacks

Maksims:
Previously in Quest 3 and Emulator as well as the other places when launching the WebXR session, the browser requestAnimationFrame would not trigger, and you would need to queue up rAF from the XRSession instead of window.
But with recent changes, they've changed it, so you can have multiple request animation frames, from the XRSession and from the window, which leads to double application updates per frame. This is an issue of course.

This PR fixes that and avoids overlapping rAF callbacks per frame by cancelling any existing rAF callbacks when calling AppBase.tick.

For rendering in an overlay browser window while in a WebXR session, we would need to figure out how to handle that separately, and ensure we do avoid multiple application update calls per actual frame.

### DIFF
--- a/src/framework/app-base.js
+++ b/src/framework/app-base.js
@@ -2118,9 +2118,9 @@ const makeTick = function (_app) {
         if (application.frameRequestId) {
             application.xr?.session?.cancelAnimationFrame(application.frameRequestId);
             cancelAnimationFrame(application.frameRequestId);
+            application.frameRequestId = null;
         }
 
-        application.frameRequestId = null;
         application._inFrameUpdate = true;
 
         setApplication(application);
--- a/src/framework/app-base.js
+++ b/src/framework/app-base.js
@@ -2114,6 +2114,13 @@ const makeTick = function (_app) {
         if (!application.graphicsDevice)
             return;
 
+        // cancel any hanging rAF to avoid multiple rAF callbacks per frame
+        if (application.frameRequestId) {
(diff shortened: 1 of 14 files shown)


src/framework/app-base.js
```diff
@@ -2114,6 +2114,13 @@ const makeTick = function (_app) {
         if (!application.graphicsDevice)
             return;
 
+        // cancel any hanging rAF to avoid multiple rAF callbacks per frame
+        if (application.frameRequestId) {
+            if (application.xr?.session)
+                application.xr.session.cancelAnimationFrame(application.frameRequestId);
```
kpal81xd, line 2120: Can you not use optional chaining for the entire line?

willeastcott (approved):

src/framework/app-base.js
```diff
@@ -2114,6 +2114,12 @@ const makeTick = function (_app) {
         if (!application.graphicsDevice)
             return;
 
+        // cancel any hanging rAF to avoid multiple rAF callbacks per frame
+        if (application.frameRequestId) {
+            application.xr?.session?.cancelAnimationFrame(application.frameRequestId);
+            cancelAnimationFrame(application.frameRequestId);
+        }
+
```
willeastcott, line 2122: Should the `application.frameRequestId = null;` line below move inside the `if`?

willeastcott:
Nice job, @Maksims! Thanks.

PR #6250, opened 2024-04-12
Basic shader generation refactoring

mvaligursky:
This PR uses recently added support for `#include` in the Preprocessor (https://github.com/playcanvas/engine/pull/6242) to refactor the basic shader generation. Instead of a JS code that glues strings and chunks together, it creates a list of defines based on the generation options, and the vertex and fragment code is simply processed with those. This makes it a lot easier to understand the shader code, which is now specified as a single string.

This is a prototype of where we want to go with the standard shader in the future.

**New public API:**
- Shader.constructor not accepts list of includes and defines for both vertex and fragment shaders, which are used to drive the shader code processing.

### DIFF
--- a/src/core/preprocessor.js
+++ b/src/core/preprocessor.js
@@ -44,12 +44,17 @@ class Preprocessor {
      * Run c-like preprocessor on the source code, and resolves the code based on the defines and ifdefs
      *
      * @param {string} source - The source code to work on.
-     * @param {Map<string, string>} [includes] - An object containing key-value pairs of include names and their
-     * content.
+     * @param {Map<string, string>} [defines] - A map containing key-value pairs of define names
+     * and their values. These are used for resolving #ifdef style of directives in the source.
+     * @param {Map<string, string>} [includes] - A map containing key-value pairs of include names
+     * and their content. These are used for resolving #include directives in the source.
      * @param {boolean} [stripUnusedColorAttachments] - If true, strips unused color attachments.
      * @returns {string|null} Returns preprocessed source code, or null in case of error.
      */
-    static run(source, includes = new Map(), stripUnusedColorAttachments = false) {
+    static run(source, defines, includes = new Map(), stripUnusedColorAttachments = false) {
+
+        // shallow clone as we will modify the map
+        defines = defines ? new Map(defines) : new Map();
 
         // strips comments, handles // and many cases of /*
         source = source.replace(/\/\*[\s\S]*?\*\/|([^\\:]|^)\/\/.*$/gm, '$1');
@@ -60,7 +65,6 @@ class Preprocessor {
             .join('\n');
 
         // generate defines to remove unused color attachments
-        const defines = new Map();
         if (stripUnusedColorAttachments) {
 
             // find out how many times pcFragColorX is used (see gles3.js)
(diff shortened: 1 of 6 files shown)


src/core/preprocessor.js
```diff
@@ -44,12 +44,17 @@ class Preprocessor {
...
-     * content.
+     * @param {Map<string, string>} [defines] - A map containing key-value pairs of define names
+     * and their values. These are used for resolving #ifdef style of directives in the source.
+     * @param {Map<string, string>} [includes] - A map containing key-value pairs of include names
+     * and their content. These are used for resolving #include directives in the source.
      * @param {boolean} [stripUnusedColorAttachments] - If true, strips unused color attachments.
      * @returns {string|null} Returns preprocessed source code, or null in case of error.
      */
-    static run(source, includes = new Map(), stripUnusedColorAttachments = false) {
+    static run(source, defines, includes = new Map(), stripUnusedColorAttachments = false) {
```
slimbuck, line 54: maybe also make `defines = new Map()` so you don't need the test below

mvaligursky: but then it nothing is passed in, I'd create a map and then clone it. I'd prefer to avoid it.

slimbuck (approved):
as the weird clown says, I'm lovin it

src/scene/shader-lib/programs/basic.js
```diff
@@ -46,124 +151,66 @@ class ShaderGeneratorBasic extends ShaderGenerator {
...
-            fshader += shaderChunks.packDepthPS;
-        }
+        includes.set('shaderPassDefines', shaderPassInfo.shaderDefines);
+        includes.set('fogCode', ShaderGenerator.fogCode(options.fog));
+        includes.set('alphaTestPS', shaderChunks.alphaTestPS);
+        includes.set('packDepthPS', shaderChunks.packDepthPS);
 
-        // FRAGMENT SHADER BODY
-        fshader += ShaderGenerator.begin();
+        this.addDefine(defines, options.vertexColors, 'VERTEX_COLORS');
```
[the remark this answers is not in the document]
mvaligursky: yep works out shorter as well, done.

src/scene/shader-lib/programs/basic.js
```diff
@@ -3,14 +3,118 @@ import {
...
 import { shaderChunks } from '../chunks/chunks.js';
-
-import {
-    SHADER_DEPTH, SHADER_PICK
-} from '../../constants.js';
 import { ShaderPass } from '../../shader-pass.js';
-
 import { ShaderGenerator } from './shader-generator.js';
 
+const vShader = `
```
kpal81xd, line 9:
```suggestion
const vShader = /* glsl */`
```

src/scene/shader-lib/programs/basic.js
```diff
@@ -3,14 +3,118 @@ import {
...
+            vColor = vertex_color;
+        #endif
+
+        #ifdef DIFFUSE_MAP
+            vUv0 = vertex_texCoord0;
+        #endif
+    }
+`;
+
+const fShader = `
```
kpal81xd, line 61:
```suggestion
const fShader = /* glsl */`
```

PR #6252, opened 2024-04-15
Getter for viewInvMatrix property (for model viewer)

kpal81xd:
- Adds getter for `_viewInvMatrix` in XRView for model viewer (hidden from docs)

### DIFF
--- a/src/framework/xr/xr-view.js
+++ b/src/framework/xr/xr-view.js
@@ -85,13 +85,13 @@ class XrView extends EventHandler {
      * @type {Mat4}
      * @private
      */
-    _viewInvMat = new Mat4();
+    _invViewMat = new Mat4();
 
     /**
      * @type {Mat4}
      * @private
      */
-    _viewInvOffMat = new Mat4();
+    _invViewOffMat = new Mat4();
 
     /**
      * @type {XRCamera}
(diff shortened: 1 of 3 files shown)


src/framework/xr/xr-view.js
```diff
@@ -324,6 +324,14 @@ class XrView extends EventHandler {
         return this._viewInvOffMat;
     }
 
+    /**
+     * @type {Mat4}
+     * @ignore
+     */
+    get viewInvMat() {
```
willeastcott, line 331: I know you've set this as `@ignore` but I don't like the name if this is intended as API. `invViewMat` is better. I've head the phrase 'inverse view matrix' before. I've never heard anyone say 'view inverse matrix'.

mvaligursky, line 331: Camera component exposes viewMatrix only, I'd be keen to not expose this at all.

willeastcott: Also, the naming isn't consistent with the `CameraComponent`'s `viewMatrix` and `projectionMatrix` properties.

mvaligursky: Why is per-eye view matrix needed?

kpal81xd: Okay sounds like this is a bigger refactor ill close this PR and create at a later date

kpal81xd: It technically isnt but the model viewer gets the inverse matrix using `const mat = xr.views.list[0]._viewInvMat;`
but ive replaced it with `const mat = xr.camera.camera.viewMatrix.clone().invert();`

PR #6254, opened 2024-04-15
Do not create a sky shader variant based on skyIntensity being equal to 1

mvaligursky:
- before, a sky shader variant would be created based on skyIntensity being equal to 1 or not. 
- this PR always includes intensity multiplication in the shader, and so this is not needed
- also slight refactor of skydome code generation - handling of mipmap selection using define instead of string replace


willeastcott (approved):
Nice!

PR #6255, opened 2024-04-15
Require all supported WebGPU limits when creating a device

mvaligursky:
- we were getting just a default limits, but can get higher limits if requesting them.


slimbuck (approved):
any idea why this isn't default behaviour?

mvaligursky:
> any idea why this isn't default behaviour?

`The only reason to not require better limits is when testing to ensure it works on all WebGPU supported devices.`

PR #6256, opened 2024-04-15
Mark WebGPU compute examples public

mvaligursky:
- I’m happy with the current compute support and the API and so these can now be public


erikdubbelboer (approved):
Great work @mvaligursky!

kpal81xd (approved):

PR #6261, opened 2024-04-18
Procedural geometry refactoring, createMesh and similar deprecation

mvaligursky:
Refactoring goals:
- separate geometry functionality, from Mesh functionality. Current function would simply create a mesh, and if CPU processing / access to data was needed, those would have to be extracted from the mesh VBs.
- simplify a way to add new geometry shapes while allowing three-shaking if not used

## New API

- **Geometry** - a base class to store geometry in CPU memory arrays
- **SphereGeometry**,  **PlaneGeometry** and similar - extended classes to fill arrays of the base class with appropriate vertex and index data
- **Mesh.fromGeometry** - a static function to create a Mesh instance from the Geometry

## Deprecated API
- **createMesh** - replaced by Mesh.fromGeometry
- **createSphere**, **createPlane** and similar - replaced by SphereGeometry and similar classes

All deprecate functions are still fully functional, and print a deprecated warning.

## Upgrading guide

Before:
```
const cylinderMesh = pc.createCylinder(app.graphicsDevice, { capSegments: 200 });
```
Now:
```
const cylinderMesh = pc.Mesh.fromGeometry(app.graphicsDevice, new pc.CylinderGeometry({ capSegments: 200 }));
```


LeXXik:
Can this be used as input for the particle emitter? I remember there was an issue previously.

kpal81xd:
Nice look way cleaner now

slimbuck (approved):

mvaligursky:
> @mvaligursky Whats the intension of keeping geometry utils separate? why not add those to the geometry base class?

Those are already public functions, and sometimes also called with arrays of positions / indices without having to ues Geometry class. So the base class just calls them as one of the customers of them.

kpal81xd (approved):

PR #6267, opened 2024-04-19
Add convex hull support to CollisionComponent

### LINKED ISSUE #123: Convex Hulls

willeastcott:
Hints to the physics engine that a collision mesh should be treated as a convex hull. This enables dynamic body meshes to collide with each other.

Here it is in action with `app.timeScale` set to `0.1` so you can observe the collisions more easily:

convexhull

Adds new API:

`CollisionComponent#convexHull: boolean`

Needs to be exposed in the Editor. For now, you can experiment using the following script:

```js
var ConvexHull = pc.createScript('convexHull');

ConvexHull.attributes.add('convexHull', { type: 'boolean', default: false });

ConvexHull.prototype.initialize = function() {
    this.entity.collision.convexHull = this.convexHull;

    this.on('attr', (name, value, prev) => {
        this.entity.collision.convexHull = value;
    });
};
```

Fixes #123

### DIFF
--- a/src/framework/components/collision/system.js
+++ b/src/framework/components/collision/system.js
@@ -30,6 +30,7 @@ const _schema = [
     'radius',
     'axis',
     'height',
+    'convexHull',
     'asset',
     'renderAsset',
     'shape',
@@ -202,6 +203,7 @@ class CollisionSystemImpl {
             radius: src.data.radius,
             axis: src.data.axis,
             height: src.data.height,
+            convexHull: src.data.convexHull,
             asset: src.data.asset,
             renderAsset: src.data.renderAsset,
             model: src.data.model,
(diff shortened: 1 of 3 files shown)


src/framework/components/collision/system.js
```diff
@@ -331,6 +333,29 @@ class CollisionMeshSystemImpl extends CollisionSystemImpl {
...
+            // No need to calculate the aabb here. We'll do it after all points are added.
+            hull.addPoint(point, false);
+        }
+
+        Ammo.destroy(point);
+
+        hull.recalcLocalAabb();
+        hull.setMargin(0.01);   // Note: default margin is 0.04
+
+        shape.addChildShape(this.system._getNodeTransform(node), hull);
```
mvaligursky, line 356: I assume the hull will be destroyed when the shape is destroyed, no need to do it manually.

mvaligursky (approved):

src/framework/components/collision/system.js
willeastcott: Correct.

LeXXik (approved):

src/framework/components/collision/system.js
```diff
@@ -331,6 +333,29 @@ class CollisionMeshSystemImpl extends CollisionSystemImpl {
...
+
+        const point = new Ammo.btVector3();
+
+        const positions = [];
+        mesh.getPositions(positions);
+
+        for (let i = 0; i < positions.length; i += 3) {
+            point.setValue(positions[i] * scale.x, positions[i + 1] * scale.y, positions[i + 2] * scale.z);
+
+            // No need to calculate the aabb here. We'll do it after all points are added.
```
LeXXik, line 347: Not sure this comment is valid. We don't calculate aabb inside the loop either way.

willeastcott: This comment is to explain why `false` is passed as the second parameter of `addPoint`. See: https://pybullet.org/Bullet/BulletFull/btConvexHullShape_8cpp_source.html

LeXXik: Ah, you are right, of course.

PR #6268, opened 2024-04-19
Suport for reading from StorageTexture by a shader (WebGPU)

mvaligursky:
By default, a StorageTexture can be only used with `write-only` mode. When WebGPU exposes storage texture `read` support for WGSL language, the StorageTexture can also be used in read mode.

This extension has been added in Chrome 124, and this PR adds support for it: https://developer.chrome.com/blog/new-in-webgpu-124

Also, few bind group parameters were marked as optional.

### DIFF
--- a/src/platform/graphics/graphics-device.js
+++ b/src/platform/graphics/graphics-device.js
@@ -208,6 +208,20 @@ class GraphicsDevice extends EventHandler {
     supportsCompute = false;
 
     /**
+     * True if the device can read from StorageTexture in the compute shader. By default, the
+     * storage texture can be only used with the write operation.
+     * When a shader uses this feature, it's recommended to use a requires-directive to signal the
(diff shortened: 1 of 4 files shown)


src/platform/graphics/graphics-device.js
```diff
@@ -207,6 +207,20 @@ class GraphicsDevice extends EventHandler {
      */
     supportsCompute = false;
 
+    /**
+     * True if the device can read from StorageTexture in the compute shader. By default, the
+     * storage texture can be only used with the write operation.
+     * When a shader uses this feature, it's recommended to use a requires-directive to signal the
```
willeastcott, line 213:
```suggestion
     * When a shader uses this feature, it's recommended to use a `requires` directive to signal the
````

willeastcott (approved):

PR #6284, opened 2024-04-24
Be clear that Node 18+ is required

### LINKED ISSUE #6236: Engine requires Node 18.0 to build
Due to some dependencies updating the engine now only builds with Node 18.

We should document this version, add an [engines field](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines) and optionally [add an `.nvmrc`](https://github.com/nvm-sh/nvm?tab=readme-ov-file#nvmrc) or equivalent to the root to set a version if using nvm

image

willeastcott:
* Explicitly declare in README that Node 18+ is required.
* Add a .`nvmrc`.
* Add an `engines` field to the `package.json`.

Fixes #6236 


marklundin:
LGTM. Nice one!

marklundin (approved):

mvaligursky:
Could we also add this to `.nvmrc` to avoid to have to edit it locally?
`registry=https://registry.npmjs.org/`

Also, we should do the same for `./examples` I think.

willeastcott:
@mvaligursky That goes in `.npmrc`, not `.nvmrc`.

kungfooman:
I used to build the engine on my old Ubuntu VM/server and it doesn't support node v18:

https://github.com/rollup/rollup/issues/5497

There are hundreds of other people with the same issue: https://stackoverflow.com/questions/72921215/getting-glibc-2-28-not-found

The "solution" is "install v16"... software disintegrates quicker than you can fix it these days :disappointed:

PR #6288, opened 2024-04-25
Examples build speedup

### LINKED ISSUE #6230: Examples - build step takes progressively longer each time
Running `npm run develop` locally. Each time an engine / example change is saved, the build takes longer and longer.

Screenshot 2024-04-09 at 15 23 03

kpal81xd:
Fixes #6230 

- Disables build output for standalone files (not used)
- Disables copying over of assets that are unlikely to change for each rebuild
- Converted generate standalone and file copy to be done on `buildEnd` hook
- Outputs into `cache` directory to remove deletion call on build writing
- Changes execSync call to use node directly instead of npm (caused slowdown) 

Build takes ~750ms on M1 Macbook


examples/utils/plugins/rollup-copy-static.mjs
```diff
@@ -24,13 +30,17 @@ export function copyStatic(nodeEnv, targets) {
...
             }
         },
-        generateBundle() {
-            targets.forEach((target) => {
+        buildEnd() {
+            for (let i = 0; i < targets.length; i++) {
+                const target = targets[i];
+                if (target.once && copied.has(target.src)) {
+                    if (log) {
+                        console.log(`${YELLOW_OUT}skipped copying ${BOLD_OUT}${target.src}${REGULAR_OUT}`);
```
mvaligursky, line 38: I assume this don't log out on each rebuild?

mvaligursky (approved):
Sounds fantastic!

PR #6291, opened 2024-04-26
Added mesh particles example

mvaligursky:
- new example to use as a test when refactoring particle shaders
- switched rendering of mesh particles to use 32bit indices to avoid limits

Screenshot 2024-04-26 at 12 35 51

### DIFF
--- a/src/scene/particle-system/particle-emitter.js
+++ b/src/scene/particle-system/particle-emitter.js
@@ -13,7 +13,7 @@ import {
     BUFFER_DYNAMIC,
     CULLFACE_NONE,
     FILTER_LINEAR, FILTER_NEAREST,
-    INDEXFORMAT_UINT16,
+    INDEXFORMAT_UINT32,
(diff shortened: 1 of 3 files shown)


src/scene/particle-system/particle-emitter.js
```diff
@@ -13,7 +13,7 @@ import {
     BUFFER_DYNAMIC,
     CULLFACE_NONE,
     FILTER_LINEAR, FILTER_NEAREST,
-    INDEXFORMAT_UINT16,
+    INDEXFORMAT_UINT32,
```
willeastcott, line 16: Why did Uint16 used to be the limit?

mvaligursky: Because for WebGL1, uint32 index was an extension. For WebGL2, this is part of the API and we use it all the time.

willeastcott (approved):

Issue #5892, opened 2023-12-12
Remove WebGL1 support


mvaligursky:
As WebGL2 support is almost on all devices (that can run PlayCanvas content): https://web3dsurvey.com/webgl2, it's time to consider removal of WebGL1 support.

There are two stages to this:
- add a deprecated warning
- remove the actual support in some future release

willeastcott:
Because why not:
https://twitter.com/willeastcott/status/1734574645871771931

LeXXik:
The time has come!
_cheerful noises in the background_

Maksims:
In recent year, the only cases of WebGL 1 I've seen in production is a combination of old hardware and outdated software.
The performance of basic content on these devices is really bad, and would require application developer to reduce the quality of the app drastically.

There are various platforms with their stats, biased based on the audience (gaming websites, caniuse, etc).

I believe it is time to go for WebGL 1.

What would a major benefits would be for the engine by removing WebGL1? e.g. Transform Feedback (does WebGPU have alternative, Compute Shaders?), download size reduction, stability, some GLSL features, what else?

mvaligursky:
> What would a major benefits would be for the engine by removing WebGL1
We could use Uniform buffers on WebGL2 (to match WebGPU). We could add it now, but we'd have two systems.

mvaligursky:
related: https://github.com/playcanvas/engine/pull/5957

erikdubbelboer:
As mentioned as a response to the above tweet as well, Poki is still seeing around 5% of it's users using WebGL1: https://developers.poki.com/player-device-report/

liamdon:
Looks like Three.js is dropping WebGL1 support in their next release: https://github.com/mrdoob/three.js/pull/27836

mvaligursky:
Yep, the timing is ticking here as well .. I suspect we won't have it for too long either.

Maksims:
> Looks like Three.js is dropping WebGL1 support in their next release: [mrdoob/three.js#27836](https://github.com/mrdoob/three.js/pull/27836)

Bear in mind that updating three.js version in your project is optional, while with PlayCanvas Editor it will be enforced. This makes significant difference.

mvaligursky:
The plan to **phase out WebGL1** support from the PlayCanvas engine is following:
- **Removal from Engine**: We'll eliminate WebGL1 support from the engine. Existing engine-only users can choose to use any engine version that still supports WebGL1.
[...]

PR #6293, opened 2024-04-26
Internal class to expose engine features

mvaligursky:
related to: https://github.com/playcanvas/engine/issues/5892, but could be useful many times we need to make the Editor backwards compatible.

### DIFF
--- a/src/framework/features.js
+++ b/src/framework/features.js
@@ -1,0 +1,11 @@
+/**
+ * Pure static class exposing a set of flags of engine features, allowing the external tools like
+ * the Editor to work with multiple engine versions easily.
+ *
+ * @ignore
+ */
+class Features {
+    static removedWebGL1 = true;
+}
+
+export { Features };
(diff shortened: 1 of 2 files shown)


LeXXik:
I think you wanted to link another issue.

kpal81xd (approved):

mvaligursky:
any thoughts on this one @willeastcott ?

willeastcott:
I guess I'm not convinced of why this is useful yet...

kpal81xd:
@mvaligursky was this a more general solution for showing UI components in the editor based on engine version?

mvaligursky:
Yes for UI components. But it might also more important soon as we've decided to keep the Editor backwards compatible with 1.70 - this might play role here.
But happy to wait and see if the Editor actually needs it, which we should see when you start stripping out the WebGL1 support from launcher / publisher @kpal81xd

willeastcott:
Yeah, can we wait until we need this please? 😄

mvaligursky:
closing for now

PR #6297, opened 2024-04-26
WebGL device ignores WebGL1 request and always creates WebGL2 device

mvaligursky:
- WebGL device ignores WebGL1 request and always creates WebGL2 device


kpal81xd:
Prob worth still add a description?

kpal81xd (approved):

src/platform/graphics/webgl/webgl-graphics-device.js
```diff
@@ -380,8 +377,7 @@ class WebglGraphicsDevice extends GraphicsDevice {
         if (options.gl) {
             gl = options.gl;
         } else {
-            const preferWebGl2 = (options.preferWebGl2 !== undefined) ? options.preferWebGl2 : true;
-            const names = preferWebGl2 ? ["webgl2", "webgl", "experimental-webgl"] : ["webgl", "experimental-webgl"];
+            const names = ["webgl2", "webgl", "experimental-webgl"];
```
kpal81xd, line 380: So there will still be an internal fallback to webgl1 if webgl2 fails?

examples/src/examples/misc/multi-app/example.mjs
```diff
@@ -21,9 +21,7 @@ const createApp = async function (deviceType) {
         device = new pc.WebgpuGraphicsDevice(canvas, {});
         await device.initWebGpu(rootPath + '/static/lib/glslang/glslang.js', rootPath + '/static/lib/twgsl/twgsl.js');
     } else if (deviceType === 'webgl1' || deviceType === 'webgl2') {
-        device = new pc.WebglGraphicsDevice(canvas, {
```
yaustar, line 24: Do we want to add a warning if `webgl1` is requested saying deprecated or something?

mvaligursky: No really, we'll string it out from examples completely so it cannot be requested by the examples

PR #6312, opened 2024-04-29
Cleanup of WebGLGraphicsDevice to strip out WebGL1 code

mvaligursky:
general cleanup

### DIFF
--- a/src/platform/graphics/webgl/webgl-graphics-device.js
+++ b/src/platform/graphics/webgl/webgl-graphics-device.js
@@ -931,10 +931,8 @@ class WebglGraphicsDevice extends GraphicsDevice {
     get extDisjointTimerQuery() {
         // lazy evaluation as this is not typically used
         if (!this._extDisjointTimerQuery) {
-            if (this.isWebGL2) {
-                // Note that Firefox exposes EXT_disjoint_timer_query under WebGL2 rather than EXT_disjoint_timer_query_webgl2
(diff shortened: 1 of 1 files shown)


src/platform/graphics/webgl/webgl-graphics-device.js
```diff
@@ -192,10 +192,7 @@ function testTextureFloatHighPrecision(device) {
  */
 class WebglGraphicsDevice extends GraphicsDevice {
     /**
-     * The WebGL context managed by the graphics device. The type could also technically be
-     * `WebGLRenderingContext` if WebGL 2.0 is not available. But in order for IntelliSense to be
-     * able to function for all WebGL calls in the codebase, we specify `WebGL2RenderingContext`
-     * here instead.
+     * The WebGL2 context managed by the graphics device.
```
willeastcott, line 195: This should really say `Create a new WebglGraphicsDevice instance.` And the class description should be update to cover what the `WebglGraphicsDevice` is about (including talking about the fact that it's a WebGL 2.0-based device with no support for WebGL 1.0.

willeastcott (approved):

src/platform/graphics/webgl/webgl-graphics-device.js
mvaligursky: This is not a doc for the constructor, but just a `gl` member. The rest is as you described already.

willeastcott: Oh, yeah, not sure why I misread that. Anyway, this class should have its own class description instead of cloning the `GraphicsDevice` one. That comment still stands.

PR #6323, opened 2024-04-30
Gizmo cleanup and bug fixes

kpal81xd:
- Split `calcPoint` into separate overloads for each subclass (and renamed to `screenToPoint`
- Replaced custom ray plane intersection with built in shapes
- Cleaner code for rotation calculation (facing and non-facing)
- Fixed facing guide angle rotation bug
- Fixed orthographic local space rotation bug
- Updated render materials when modifying `colorAlpha`
- Updated lighting for example
- Added instructions to example
- Persistent coordSpace on changing gizmo types

### Preview 

image


src/extras/gizmo/rotate-gizmo.js
```diff
@@ -360,6 +373,44 @@ class RotateGizmo extends TransformGizmo {
...
+        plane.intersectsRay(ray, point);
+
+        // calculate angle
+        const facingDir = tmpV1.sub2(ray.origin, gizmoPos).normalize();
+        const facingDot = plane.normal.dot(facingDir);
+        if (axis === 'face' || Math.abs(facingDot) > FACING_THRESHOLD) {
+            // plane facing camera so based on mouse position around gizmo
+            tmpQ1.copy(this._camera.entity.getRotation()).invert();
+
+            // transform point so its facing the camera
```
willeastcott, line 398:
```suggestion
            // transform point so it's facing the camera
```

examples/src/examples/misc/gizmos/example.mjs
```diff
@@ -226,6 +231,32 @@ function createColorMaterial(color) {
...
+instr.addComponent('element', {
+    pivot: new pc.Vec2(0.5, 0.5),
+    anchor: new pc.Vec4(0, 1, 1, 0.8),
+    margin: new pc.Vec4(0, 0, 0, 0),
+    fontAsset: assets.font.id,
+    fontSize: 18,
+    text: 'Translate (1), Rotate (2), Scale (3)\nWorld/Local (X)\nPerspective (P), Orthographic (O)',
+    type: pc.ELEMENTTYPE_TEXT
+});
+screen.addChild(instr);
```
willeastcott, line 258: Is this nice and readable on mobile?

src/extras/gizmo/tri-data.js
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * The class for holding triangle data.
+ *
+ * @ignore
+ */
+class TriData {
+    /**
+     * The priority of the triangle data (Used for intersection ordering).
+     * priority = 0 - no priority
+     * priority > 0 - higher value represents a higher priority
```
willeastcott, line 21: I know this isn't API, but if this was, this wouldn't render right. You probably want bullets here. Otherwise, it'd all be on the same line.

examples/src/examples/misc/gizmos/example.mjs
kpal81xd: The text seems very pixelated on mobile as do a lot of the other text examples

src/scene/geometry/geometry.js
```diff
@@ -91,6 +98,15 @@ class Geometry {
         Debug.assert(this.indices, 'Geometry must have indices set');
         this.tangents = calculateTangents(this.positions, this.normals, this.uvs, this.indices);
     }
+
+    /**
+     * Generates triangle information from the positions and triangle indices.
+     */
+    calculateTris() {
```
willeastcott, line 105: I would suggest, for now, removing this function and have `TriData` call `calculateTris`. And also for now, just keep `calculateTris` private/not exported. We can expose it later should we wish, but let's be conservative for now.

examples/src/examples/misc/gizmos/example.mjs
willeastcott: I think you just need to do:

```
app.graphicsDevice.maxPixelRatio = window.devicePixelRatio;
```

src/scene/geometry/geometry-utils.js
```diff
@@ -1,5 +1,10 @@
 import { Vec2 } from '../../core/math/vec2.js';
 import { Vec3 } from '../../core/math/vec3.js';
+import { Tri } from '../../core/shape/tri.js';
+
+const tmpV1 = new Vec3();
+const tmpV2 = new Vec3();
+const tmpV3 = new Vec3();
```
willeastcott, line 7: Can this go?

PR #6324, opened 2024-04-30
Remove uniform buffer based skinning and batching

mvaligursky:
WebGPU and WebGL2 support texture based skinning, and so the limited uniform buffer skinning and batching used for some WebGL1 devices is removed.


src/scene/skin-instance.js
```diff
@@ -134,10 +128,8 @@ class SkinInstance {
     uploadBones(device) {
 
         // TODO: this is a bit strange looking. Change the Texture API to do a reupload
-        if (device.supportsBoneTextures) {
-            this.boneTexture.lock();
-            this.boneTexture.unlock();
-        }
+        this.boneTexture.lock();
```
slimbuck, line 131: call `texture.upload()` instead

mvaligursky: yep

willeastcott (approved):
Great to see this go!

PR #6327, opened 2024-04-30
Remove parts of shader chunks only used on WebGL1

mvaligursky:
- removed code in the else branches of `#ifdef GL2`

### DIFF
--- a/src/scene/shader-lib/chunks/common/frag/msdf.js
+++ b/src/scene/shader-lib/chunks/common/frag/msdf.js
@@ -1,13 +1,7 @@
 export default /* glsl */`
 uniform sampler2D texture_msdfMap;
 
-#ifdef GL_OES_standard_derivatives
 #define USE_FWIDTH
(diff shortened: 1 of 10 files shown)


src/scene/shader-lib/chunks/common/frag/msdf.js
```diff
@@ -1,13 +1,7 @@
 export default /* glsl */`
 uniform sampler2D texture_msdfMap;
 
-#ifdef GL_OES_standard_derivatives
 #define USE_FWIDTH
```
willeastcott, line 4: Can you not just remove `USE_FWIDTH` then, since it's always defined?

willeastcott (approved):

src/scene/shader-lib/chunks/common/frag/msdf.js
mvaligursky: yep, did that, thanks

PR #6330, opened 2024-05-01
Removed WebGL1 clustered lighting fallback for storing floats in 8bit texture

mvaligursky:
- as now we have guaranteed float textures


slimbuck (approved):
A thing of beaty!

PR #6332, opened 2024-05-01
Disable treeshaking polyfills

### LINKED ISSUE #6328: Engine 1.70.0 is missing polyfills
Forum thread: https://forum.playcanvas.com/t/engine-release-v1-70-0/35666/9

polyfill that is missing (applies to few other I checked): https://github.com/playcanvas/engine/blob/main/src/polyfill/object-entries.js

If I search the string from polyfill `Object.fromEntries() requires a single iterable argument`, I can find it in 1.69.0 but not 1.70.0

https://code.playcanvas.com/playcanvas-1.69.0.js
https://code.playcanvas.com/playcanvas-1.70.0.js

kpal81xd:
Fixes #6328 

Adds a custom plugin to disable treeshaking on certain module paths (polyfills in this case)

### DIFF
--- a/utils/plugins/rollup-treeshake-ignore.mjs
+++ b/utils/plugins/rollup-treeshake-ignore.mjs
@@ -1,0 +1,25 @@
+/**
+ * A rollup plugin that filters out modules from treeshaking
+ *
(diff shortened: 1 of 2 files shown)


mvaligursky (approved):
Nice!

PR #6335, opened 2024-05-02
Examples mobile fixes

### LINKED ISSUE #3991: Examples control panel on mobile takes up the whole screen
See the screenshot.
Ideally it would only take as much space as needed by the controls. Perhaps even it should never take more than 50% of the screen and have a scroll bar instead. When displayed, the example could then be resized to the other half, so all of it is visible.

image

### LINKED ISSUE #5215: [Examples] When switching examples on Android, empty UI covers the screen
When switching examples on Android, this UI that covers half the screen shows up. To hide it, I need to maximise first, and then minimize - when makes it hard to capture some promo videos.

android UI

kpal81xd:
Fixes #3991
Fixes #5215

- Upgraded mobile code editor for mobile with proper highlighting and theme (read only)
- Fixed controls and code editor styling (clipped under Examples header)
- Made long controls scrollable
- Increased device pixel ratio to of the window (capped at 2)
- Gizmo distance and size invariant for different aspect ratios

### Preview
**Code editor**
image

**Scrollable & Smaller controls**
image

### DIFF
--- a/examples/src/app/components/CodeEditor.mjs
+++ b/examples/src/app/components/CodeEditor.mjs
@@ -43,7 +43,7 @@ let monacoEditor;
 
 // eslint-disable-next-line jsdoc/require-property
 /**
- * @typedef {object} Props
+ * @typedef {Record<string, any>} Props
  */
 
 /**
@@ -314,7 +314,7 @@ class CodeEditor extends TypedComponent {
         const options = {
             value,
             language,
-            theme: 'vs-dark',
+            theme: 'playcanvas',
             loading: null,
             beforeMount: this.beforeMount.bind(this),
             onMount: this.editorDidMount.bind(this),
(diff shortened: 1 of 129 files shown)


mvaligursky:
> Increased device pixel ratio to of the window (capped at 2)

That would possibly have a negative performance impact on fragment limited examples.

Is there a good way to make the Controls panel less wide so that you can see the example you're adjusted on the side?

mvaligursky (approved):
Looks great!

mvaligursky:
Great solution! Merge away.

willeastcott (approved):
Love the polish here - awesome!

Maksims:
How does UI look on high DPI screens, like retina Macs?
Best would be to solve it using CSS instead of whole screen DPI.

PR #6336, opened 2024-05-03
Remove final WebGL1 extensions and simplify code that used them

mvaligursky:
- few remaining WebGL1 only extensions stripped out
- now that half16 texture format is guaranteed, removed all fallback formats to store area lights LUTs


src/scene/gsplat/gsplat.js
```diff
@@ -149,7 +149,7 @@ class GSplat {
         if (preferHighPrecision) {
             halfFormat = false;
         } else {
-            halfFormat = device.extTextureHalfFloat;
+            halfFormat = true;
```
willeastcott, line 152: Can't you just do `halfFormat = !preferHighPrecision;`?

willeastcott: Or just `return !preferHighPrecision;` 😄

willeastcott (approved):

src/scene/gsplat/gsplat.js
mvaligursky: yep. I was trying to avoid changing the code much as Donovan is in the middle of it .. but have just removed this function now, thanks.

PR #6339, opened 2024-05-07
Examples file loader rework

kpal81xd:
- Moves external file loading from config to actual file loading (use `localImport` function)
- Moves flags from config to comment flags in example
- Flattened examples folders into files with preprended exampleName (e.g. `hello-world.example.mjs`)

### Preview

image


mvaligursky:
Could we also rename the example files so that they're not all called `example.mjs`? Happy if this is done in a separate PR too.

examples/src/examples/user-interface/text-typewriter/example.mjs
```diff
@@ -1,3 +1,4 @@
+// @flag WEBGPU_ENABLED
```
mvaligursky, line 1: are you using some existing `@flag` functionality? If not, could we give it some better name, perhaps `@example` or something?

kpal81xd:
> Could we also rename the example files so that they're not all called `example.mjs`? Happy if this is done in a separate PR too.

Since there is still a controls file for now I was going to rename the files to `<exampleName>.example.mjs` and `<exampleName.controls.mjs`?

examples/src/examples/user-interface/text-typewriter/example.mjs
kpal81xd: No this is custom and the word flag is just arbitrary just to signify its a config flag option so can be called anything

examples/src/examples/graphics/ground-fog/shader.frag
```diff
@@ -0,0 +1,37 @@
+
+            uniform sampler2D uTexture;
```
willeastcott, line 2: Trim all the unnecessary whitespace in this file (and all other shader files that have the same problem).

kpal81xd: Yep working on it

examples/README.md
```diff
@@ -61,6 +57,20 @@ export { app };
 
 This is the only file that's required to run an example. The code defined in this function is executed each time the example play button is pressed. It takes the example's canvas element from the DOM and usually begins by creating a new PlayCanvas `Application` or `AppBase` using that canvas.
 
+Examples can also contains comments which allow you to define the default configuration for your examples as well as overrides to particular settings such as `deviceType`. Check the possible values to set in `ExampleConfig` in `scripts/utils.mjs` file for the full list.
```
willeastcott, line 60:
```suggestion
Examples can also contain comments which allow you to define the default configuration for your examples as well as overrides to particular settings such as `deviceType`. Check the possible values to set in `ExampleConfig` in `scripts/utils.mjs` file for the full list.
```

examples/scripts/utils.mjs
```diff
@@ -0,0 +1,85 @@
+import fs from 'fs';
+
+/**
+ * @param {object} obj - The object.
+ * @returns {string} - The stringifiied object
```
willeastcott, line 5:
```suggestion
 * @returns {string} - The stringified object
```

Issue #6315, opened 2024-04-29
WebGPU: BindGroups aren't being reused


erikdubbelboer:
When using https://github.com/brendan-duncan/webgpu_inspector to look at what is happening, it's very obvious that BindGroups are constantly being created, up to a 20000 sometimes, and then being garbage collected. This then causes a lag spike.

<img width="1254" alt="Screenshot 2024-04-29 at 22 27 44" src="https://github.com/playcanvas/engine/assets/522870/26bc58b2-d7f8-46d6-bd69-9af50b1892ec">

It would be more optimal to reuse BindGroups.

mvaligursky:
BindGroup is actually re-used as much as possible. Only when it's content changes a new one is created. The reason a bind group is re-created:
- when a different texture needs to be attached - this cannot be avoided, but is low frequency.
- when a different buffer storing the uniforms is assigned to it. Most UniformBuffers are allocated from a large pools (1MB blocks currently), and a dynamic offset to those is used. Space in the buffers is allocated in order of rendering, so if the camera moves around, and the amount of meshes causes more than one 1MB buffers is used, some meshes switch between those buffers and a new BindGroup must be reallocated.

Plan to improve this:
- Material Uniform buffers: https://github.com/playcanvas/engine/issues/6157 - without this, a lot of uniforms are each frame being added to dynamic buffers, making it more likely to have to recreate the BindGroup. When material uniforms will be moved to per material persistent UBs, this allocation size will be a lot smaller.
- I also tried to play with the size of these pools (1MB). If I make them larger, it's more likely an allocation would be in the same buffer. But with WebGPU rendering few frames ahead, and buffer needs to be uploaded to GPU each time we submit CBs (which is often only 1-2 times a frame), the total amount of used memory gets pretty high.

I wish I had better options at the moment, but as they say here https://github.com/gpuweb/gpuweb/issues/915, `It's pretty clear that bindgroup creation will be one of the biggest hotspots of the API.`

Thanks for the issue. I'll close it for now, as I do not see a way to improve this, but will always keen an eye on it.

mvaligursky:
Actually, let me re-open this. I see a solution, I'll try it soon:

- as mentioned, the problem is those bind groups that use uniform buffers allocated from those large dynamic buffers. When they switch to a different buffer, bind groups need to be re-created.
- those are used for bing groups / uniform buffers of meshes. Scene / materials use persistent UBs / bind groups.
[...]

PR #6341, opened 2024-05-07
Initial refactor to how bind groups can be used with non-persistent UBs

mvaligursky:
- related to the suggested design here: https://github.com/playcanvas/engine/issues/6315#issuecomment-2093250305
- this is a first try at doing this, and applied to WebgpuClearRenderer only, but it works.
- basically each dynamic buffer from which we bump allocate non-persistent uniform buffers has a cache of bind groups based on uniform buffer size (as this size is part of the bind group). And this means, all buffers using the same size can share a single bind group, and avoid their allocation each time the bump allocator switches to a different buffer
- also did a small refactor to the way uniforms are set to the UB, allowing users to avoid using the scope in the middle and set values directly to the buffer.

The next PR will apply this to the mesh UB, which is a main reason for the allocations.

### DIFF
--- a/src/platform/graphics/dynamic-buffer.js
+++ b/src/platform/graphics/dynamic-buffer.js
@@ -1,0 +1,50 @@
+import { DebugHelper } from '../../core/debug.js';
+import { BindGroupFormat, BindUniformBufferFormat } from './bind-group-format.js';
+import { BindGroup } from './bind-group.js';
+import { SHADERSTAGE_FRAGMENT, SHADERSTAGE_VERTEX, UNIFORM_BUFFER_DEFAULT_SLOT_NAME } from './constants.js';
+
+/**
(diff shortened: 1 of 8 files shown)


src/platform/graphics/dynamic-buffer.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+class DynamicBuffer {
+    /** @type {import('./graphics-device.js').GraphicsDevice} */
+    device;
+
+    /**
+     * A cache of bind groups for each uniform buffer size, which is used to avoid creating a new
+     * bind group for each uniform buffer.
+     *
+     * @type {Map<number, BindGroup>}
```
slimbuck, line 19: adding these to the code is so incredibly useful <3

slimbuck (approved):

PR #6343, opened 2024-05-08
Examples removed linux chrome fallback

kpal81xd:
Removes fallback to webgl2 if using linux chrome (Enabled Android WebGPU)


mvaligursky (approved):

mvaligursky:
The reason for the removal is that is also stopped Android WebGPU to work

Issue #6344, opened 2024-05-08
Improve Gaussian Splats performance


cognitiveplus:
Recently Adobe researchers published a paper called [GS-LRM: Large Reconstruction Model for 3D Gaussian Splatting](https://sai-bi.github.io/project/gs-lrm/).

The page contains real-time demos featuring this viewer: [antimatter15](https://github.com/antimatter15) / [splat](https://github.com/antimatter15/splat).

What's interesting about it is that it produces solid 60fps on M1 Air even with advanced scenes, while the same PLY imported into PlayCanvas results in less FPS (with Device Pixel Ratio unchecked).

For example:
https://sai-bi.github.io/project/gs-lrm/viewer/index.html?is_object=false&url=https://huggingface.co/datasets/saibi/gs-lrm/resolve/main/scene_assets/002000_gaussians.ply
https://playcanv.as/index/7d0ab8f5 (https://playcanvas.com/project/1214533/)

From viewer description it looks like it's doing splat sorting on CPU in a separate worker thread on async basis:
> the final approach that i settled on is to run the sorting process on the CPU in a webworker, which happens a bit more slowly (at roughly 4fps whereas the main render is at 60fps), but that's fine because most of the time when you are moving around the z order doesn't actually change very fast (this results in momentary artifacts when jumping directly between different camera orientations on opposite sides).

I guess PC might introduce similar configuration for 3DGS, with selector in inspector allowing to choose a sorting algo.

mvaligursky:
We're doing splat sorting on a separate thread, and that bit is pretty fast. - faster than 4fps for sure I think. 1M splats in about 4ms on M1 Max.
But we have other bottlenecks we're investigating currently.

MAG-AdrianMeredith:
The sorting is definitely the slowest part in my experience, it causes a lot of hitching as the gpu upload locks rendering so the fps meter might be reporting a good fps but its dropping frames.  
Its still awesome though, especially for a first release

mvaligursky:
related: https://github.com/playcanvas/engine/pull/6352

willeastcott:
Closing this since it was addressed by https://github.com/playcanvas/engine/pull/6357

Released as part of v1.71.0:

https://github.com/playcanvas/engine/releases/tag/v1.71.0

PR #6352, opened 2024-05-10
Update GS rendering

slimbuck:
This PR updates our runtime GS renderer. GPU render performance is improved as well as sorting performance and _jank_ during camera movement.

Before:
Screenshot 2024-05-10 at 16 07 53

After:
Screenshot 2024-05-10 at 16 08 24

The code changes are summarised as follows:
- at load time we reorder splat data to be gpu cache friendly (using morton order)
    - this is skipped when loading the compressed PLY format because data is already ordered
- the cpu sorter writes results to a texture and calculates number of splats in front of the camera
- render with a two phase approach:
    - phase1: (preprocess) transform splat data into camera space - store the resulting data in render order (this is only required when camera moves)
    - phase2: (render) consume preprocess data and render quads

Other changes:
- splats are rendered in batches of 128 quad mesh instances to improve occupancy

Issues:
- the load time reordering is time and memory intensive. we may want to perform this on a worker.
- this technique uses more gpu memory than before (to store the phase1 output). this might be an issue for mobile devices.
- phase1 output stores splat XYZW to a RGBA16F buffer. This appears to be enough precision, but more testing is needed to confirm.
- a followup PR will restructure shaders to accommodate the rendering customisations we need in, for example [SuperSplat](https://github.com/playcanvas/supersplat)
- VR mode may require work. This will be a followup PR.
- The texture upload API doesn't support uploading partial data, so we can't upload only the required sorting data.

TODO:
- implement rendering directly from compressed format (we currently decompress the data at load time). this should result in good memory and performance speedup and should fit nicely into the existing rendering pipeline.
- investigate fragment bandwidth optimisation by:
    - switch to rendering the splats front to back (instead of back to front as we do now)
    - render half the splats
    - stencil out/mark pixels that are fully opaque
    - render second half of the scene skipping already filled pixels


mvaligursky:
What's the impact to WebXR rendering using two viewports?

Maksims:
Is it possible to configure size of batches? As in case of 1m splats, it will lead to ~7.8k drawcalls?

src/scene/gsplat/gsplat-data.js
```diff
@@ -408,6 +412,77 @@ class GSplatData {
...
+        const y = this.getProp('y');
+        const z = this.getProp('z');
+
+        const minX = x.reduce((a, b) => Math.min(a, b), x[0]);
+        const minY = y.reduce((a, b) => Math.min(a, b), y[0]);
+        const minZ = z.reduce((a, b) => Math.min(a, b), z[0]);
+
+        const maxX = x.reduce((a, b) => Math.max(a, b), x[0]);
+        const maxY = y.reduce((a, b) => Math.max(a, b), y[0]);
+        const maxZ = z.reduce((a, b) => Math.max(a, b), z[0]);
```
willeastcott, line 441: This looks lovely and tidy, but I do wonder whether something like this would be significantly faster:

```
function getMinMax(arr) {
    let min = arr[0];
    let max = arr[0];
    for (let i = 1; i < arr.length; i++) {
        if (arr[i] < min) min = arr[i];
        if (arr[i] > max) max = arr[i];
    }
    return { min, max };
}

const { min: minX, max: maxX } = getMinMax(x);
const { min: minY, max: maxY } = getMinMax(y);
const { min: minZ, max: maxZ } = getMinMax(z);
```

src/scene/gsplat/gsplat-instance.js
```diff
@@ -26,11 +143,22 @@ class GSplatInstance {
...
+    /** @type {import('../../platform/graphics/texture.js').Texture} */
+    v1v2Texture;
+
+    centerTexture;
+
+    colorTexture;
+
+    v1v2RenderTarget;
+
+    v1v2RenderPass;
```
willeastcott, line 160: Add types?

src/scene/gsplat/gsplat-instance.js
```diff
@@ -120,6 +307,10 @@ class GSplatInstance {
 
     createMaterial(options) {
         this.material = createGSplatMaterial(options);
+        // this.material.setParameter('splatOrder', this.orderTexture);
```
willeastcott, line 310: Delete?

willeastcott (approved):
A true work of art. 🖼️

willeastcott:
@slimbuck Would it be reasonable for this PR to close #6344?

slimbuck:
> Is it possible to configure size of batches? As in case of 1m splats, it will lead to ~7.8k drawcalls?

@Maksims, the 128-quad mesh is instanced as many times as necessary with single draw call.

slimbuck:
> What's the impact to WebXR rendering using two viewports?

Yeah thanks I forgot to mention in the issues section. Plan is to fix/address VR in a followup PR.

slimbuck:
> @slimbuck Would it be reasonable for this PR to close #6344?

Hopefully! :D

src/scene/gsplat/gsplat-instance.js
```diff
@@ -167,11 +358,48 @@ class GSplatInstance {
...
+        const device = this.splat.device;
+        const scope = device.scope;
+
+        scope.resolve('splatColor').setValue(this.splat.colorTexture);
+        scope.resolve('splatOrder').setValue(this.orderTexture);
+        scope.resolve('transformA').setValue(this.splat.transformATexture);
+        scope.resolve('transformB').setValue(this.splat.transformBTexture);
+        scope.resolve('transformC').setValue(this.splat.transformCTexture);
+
+        const cameraMatrix = new Mat4();
```
mvaligursky, line 378: these seem like per frame allocations?

mvaligursky (approved):

src/scene/gsplat/gsplat-data.js
slimbuck: Yep thanks, I've made the change.

slimbuck:
Closing in favour of #6357.

Issue #6345, opened 2024-05-09
App crashes when drawing lines to disabled layer


LeXXik:
If the layer that you draw lines to is disabled, the batch arrays never get cleared, overflowing and the app crashes.

Repro: https://playcanvas.com/project/1212210/overview/draw-line-issue

![image](https://github.com/playcanvas/engine/assets/5677782/1bad8c78-55fc-44d8-bf60-8fffd7e02a61)

PR #6353, opened 2024-05-10
Fix: Clean up lines for rendering even when the Layer is disabled

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/6345

- cleanup all the time, regardless of lines getting rendered

### DIFF
--- a/src/scene/immediate/immediate-batch.js
+++ b/src/scene/immediate/immediate-batch.js
@@ -91,13 +91,15 @@ class ImmediateBatch {
                 this.meshInstance = new MeshInstance(this.mesh, this.material, identityGraphNode);
             }
 
-            // clear lines when after they were rendered as their lifetime is one frame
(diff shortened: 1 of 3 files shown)


willeastcott (approved):

src/scene/immediate/immediate-batch.js
```diff
@@ -91,14 +91,16 @@ class ImmediateBatch {
...
-            this.positions.length = 0;
-            this.colors.length = 0;
-
             // inject mesh instance into visible list to be rendered
             visibleList.push(this.meshInstance);
         }
     }
+
+    clear() {
+        // clear lines when after they were rendered as their lifetime is one frame
```
willeastcott, line 100:
```suggestion
        // clear lines after they are rendered as their lifetime is one frame
```

PR #6354, opened 2024-05-10
Fix buildUrl relative path support

erikdubbelboer:
`buildUrl` never worked with relative paths as it would always replace the path in the url, instead of adding to it.
Using the other mode of the URL constructor we can use it to always give us a correct url:
```
  `new URL('glslang/glslang.js', 'https://example.com/foo/bar.html')` is `https://example.com/foo/glslang/glslang.js`
  `new URL('/glslang/glslang.js', 'https://example.com/foo/bar.html')` is `https://example.com/glslang/glslang.js`
  `new URL('glslang/glslang.js', 'https://example.com/')` is `https://example.com/glslang/glslang.js`
  `new URL('/glslang/glslang.js', 'https://example.com/')` is `https://example.com/glslang/glslang.js`
```
This bug currently prevents Playcanvas Editor exports from being run on a non root path with WebGPU. The editor export will pass just `glslang.js` as path.
So on a url like https://94176748-9ef8-42c9-a44e-a95b70ec5680.poki-gdn.com/7a8f1fc6-7162-473a-b1dc-4479ed5d60cf/index.html 
the old code would turn it into https://94176748-9ef8-42c9-a44e-a95b70ec5680.poki-gdn.com/glslang.js 
while the actual paths should be https://94176748-9ef8-42c9-a44e-a95b70ec5680.poki-gdn.com/7a8f1fc6-7162-473a-b1dc-4479ed5d60cf/glslang.js

### DIFF
--- a/src/platform/graphics/webgpu/webgpu-graphics-device.js
+++ b/src/platform/graphics/webgpu/webgpu-graphics-device.js
@@ -1,6 +1,5 @@
 import { TRACEID_RENDER_QUEUE } from '../../../core/constants.js';
 import { Debug, DebugHelper } from '../../../core/debug.js';
-import { path } from '../../../core/path.js';
 
 import {
     PIXELFORMAT_RGBA8, PIXELFORMAT_BGRA8, DEVICETYPE_WEBGPU,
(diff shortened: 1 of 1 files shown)


slimbuck (approved):
yep this is great, thanks!

PR #6356, opened 2024-05-13
Remove handling of WebGL1 seamless cubemap sampling

mvaligursky:
WebGL2/WebGPU support this by default


src/scene/graphics/reproject-texture.js
```diff
@@ -500,8 +500,8 @@ function reprojectTexture(source, target, options = {}) {
     const params = [
         0,
         specularPower,
-        source.fixCubemapSeams ? 1.0 / source.width : 0.0,          // source seam scale
-        target.fixCubemapSeams ? 1.0 / target.width : 0.0           // target seam scale
+        0.0,          // source seam scale
+        0.0           // target seam scale
```
willeastcott, line 504: Is there even any point to passing these values to the shader now?

willeastcott (approved):

src/scene/graphics/reproject-texture.js
mvaligursky: yep done now

PR #6357, opened 2024-05-13
Update GS rendering v2.0

slimbuck:
Having done more profiling, it seems that the order of splat data is by far the most important element when rendering large GS scenes on GPU.

Therefore this PR supersedes #6352: it takes the best parts, but without the additional memory requirements of the preprocess step.

Before:
Screenshot 2024-05-10 at 16 07 53

After:
Screenshot 2024-05-10 at 16 08 24

So, this PR:
- Uploads splat order data to texture instead of uploading duplicate vertex data.
- Renders splats by instancing a 128-quad mesh.
- Refrains from rendering splats behind the camera.
- Reorder uncompressed data at startup to match compressed format.
- Default to half float for storing uncompressed data.

Issues:
- reordering millions of splats at load time is not fast. For fast load time, use the compressed format. (We will investigate improving reorder performance in future)
- we might want to expose a load-time option for users to disable splat reorder on load

Next steps:
- Implement rendering directly from compressed format. This will cut runtime memory usage 4x.


src/scene/gsplat/gsplat-instance.js
```diff
@@ -58,53 +60,81 @@ class GSplatInstance {
...
 
         // create sorter
         if (!options.dither || options.dither === DITHER_NONE) {
             this.sorter = new GSplatSorter();
-            this.sorter.init(mesh.vertexBuffer, this.centers, !this.splat.device.isWebGL1);
+            this.sorter.init(this.orderTexture, this.centers);
+            this.sorter.on('updated', (count) => {
+                // limit splat render count to exclude those behind the camera.
+                // NOTE: the last instance rendered may include non-existant splat
+                // data. this should be ok though. as the data is filled with 0's.
```
willeastcott, line 135:
```suggestion
                // data. this should be ok though as the data is filled with 0's.
```

willeastcott (approved):

src/scene/gsplat/gsplat-data.js
```diff
@@ -88,6 +88,10 @@ class GSplatData {
         if (!this.isCompressed && performZScale) {
             mat4.setScale(-1, -1, 1);
             this.transform(mat4);
+
+            // reorder uncompressed splats in morton order for better memory access
+            // efficiency during rendering
+            this.reorderData();
```
mvaligursky, line 94: should this only happen when `performZScale`?

src/scene/gsplat/gsplat-data.js
```diff
@@ -408,6 +412,83 @@ class GSplatData {
...
+            morton[i] = encodeMorton3(ix, iy, iz);
+        }
+
+        // generate indices
+        const indices = new Uint32Array(this.numSplats);
+        for (let i = 0; i < this.numSplats; i++) {
+            indices[i] = i;
+        }
+        // order splats by morton code
+        indices.sort((a, b) => morton[a] - morton[b]);
```
mvaligursky, line 467: If this is the expensive part, a bucket sort could be used instead.

src/scene/gsplat/gsplat-instance.js
```diff
@@ -58,53 +60,81 @@ class GSplatInstance {
...
-            indices[triBase + 0] = base;
-            indices[triBase + 1] = base + 1;
-            indices[triBase + 2] = base + 2;
-            indices[triBase + 3] = base;
-            indices[triBase + 4] = base + 2;
-            indices[triBase + 5] = base + 3;
+        // number of quads to combine into a single instance. this is to increase occupancy
+        // in the vertex shader.
+        const splatInstanceSize = 128;
+        const numSplats = Math.ceil(splat.numSplats / splatInstanceSize) * splatInstanceSize;
```
mvaligursky, line 77: could use math.roundUp

mvaligursky (approved):

src/scene/gsplat/gsplat-data.js
slimbuck: yeah right. i didn't want to but am forced to add a reorder option. thanks!

src/scene/gsplat/gsplat-data.js
slimbuck: this isn't a fast bit, but generating the morton order is far slower. i will possibly investigate reimplementing this whole function in wasm and then also try bucketing.

PR #6360, opened 2024-05-14
Build fixes for 1.70

kpal81xd:
Includes #6340 changes for 1.70


kpal81xd:
@mvaligursky I made sure to only include those same changes in case you are merging in other build changes to 1.70

mvaligursky (approved):
Looks good. I assume this was well tested against 1.70

kpal81xd:
> Looks good. I assume this was well tested against 1.70

Yep checked the build and the importing too with ts and vite

PR #6365, opened 2024-05-14
Removed Trigger double Body deletion

### LINKED ISSUE #6364: Ammo.js "table index is out of bounds" following Trigger destruction
Hey there!

I ran into some troubles with Ammo.js and a trigger body. For some reasons I couldn't reproduce it on certain triggers but for some standalone triggers Ammo.js throws an error "Runtime Error: table index is out of bounds" on Node.

image

Affected triggers seems to be only when trigger is not a child of any rigid body.

MushAsterion:
Fixes #6364 

Trigger#destroy was calling `destroyBody` while already being called earlier from `this.disable` within the same function leading to Ammo.js trying to destroy an empty reference

### DIFF
--- a/src/framework/components/collision/trigger.js
+++ b/src/framework/components/collision/trigger.js
@@ -79,12 +79,7 @@ class Trigger {
     }
 
     destroy() {
-        const body = this.body;
-        if (!body) return;
-
         this.disable();
-
-        this.app.systems.rigidbody.destroyBody(body);
     }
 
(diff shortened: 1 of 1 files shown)


mvaligursky:
Closed?

MushAsterion:
Yes sorry, it seems the change is not fixing the original issue and I'm definitely blind as the first only remove the body and don't destroy it so this one is fine. Sorry for the early PR

PR #6369, opened 2024-05-15
Small JSDoc normalization/fixes

kungfooman:
Problem with TS style types is:

1) Can't add inline-documentation to the properties.
2) Quickly escalate line length (was > 100).

So I normalized that into JSDoc style and added some description what it is about. `targets.once` is also optional, which the TS type didn't point out, leading to type mismatch.


willeastcott (approved):
Thanks! 🙏

mvaligursky (approved):

PR #6371, opened 2024-05-15
Render compressed GS data

slimbuck:
This is a followup PR to #6357.

Up until now when loading compressed gaussian splat PLY files, we have simply decompressed the data at load time and rendered the scene as usual.

This PR adds a render path to load and render the compressed data directly, cutting memory consumption by 4x for the raw splat data and speeding up rendering by a modest 4%.

For the following bicycle scene:
| | Old | New |
| --- | --- | --- |
| Render time | 13.4ms | 13.0ms |
| Total VRAM | 208MB | 123MB | 
| Load Time | 6.0s | 1.8s |

NOTE: load time above includes the examples browser startup.

Screenshot 2024-05-15 at 14 00 55

This PR makes the following changes:
- adds a `createIter` function to `GSplatData`, which constructs an iterator for accessing gaussian splat data. the iterator works the same whether the underlying data is compressed or not.
- makes use of the new iterator in the rest of `GSplatData`, making most code smaller and simpler.
- adds compressed versions of `gsplat` and `gsplat-material` for rendering compressed data directly.
- adds `asset.data` options which control whether to decompress the data at load time (useful for tools like super-splat) and also whether to reorder data for better GPU render performance.
- speed up gs bounding box calculation by using splat bounding sphere (i.e. overestimation) and rename the exact bound calculation `calcAabbExact`.

Todo:
- it may be possible to speed up rendering by early-out based on chunk bound (to be investigated)
- we might prefer to optimise the compressed format:
   - store v1v2 instead of rotation/scale
   - store rotation/scale/v1v2 in halfs instead of floats


src/scene/gsplat/gsplat-data.js
```diff
@@ -12,70 +12,173 @@ const quat2 = new Quat();
...
+        const max_scale_x = gsplatData.getProp('max_scale_x', 'chunk');
+        const max_scale_y = gsplatData.getProp('max_scale_y', 'chunk');
+        const max_scale_z = gsplatData.getProp('max_scale_z', 'chunk');
+
+        const position = gsplatData.getProp('packed_position');
+        const rotation = gsplatData.getProp('packed_rotation');
+        const scale = gsplatData.getProp('packed_scale');
+        const color = gsplatData.getProp('packed_color');
+
+        this.read = (i) => {
```
Maksims, line 75: These functions will be allocated for each new instance, and V8 might be unable to optimize it as it depends on higher scope data. Which can lead to worse performance of that method due to deopt, especially taking into the account that it is called a lot of times. It is best to define it as a class member.

Maksims: Same goes to each method defined in constructor. It is best to have it as members of a class, instead of allocating new method for each instance. Also if it is possible to avoid calling a function for very high frequency methods (e.g. `lerp`), it is best to inline its code.

slimbuck: I tried your suggestions, but don't see any difference in performance loading/decompressing/uploading a 6M splat scene. This is on chrome macos. (I would have been surprised if there was a difference tbh).

Maksims: With using inline lerp instead of a function, I get 3 times the performance in benchmarks, same goes for unpacking methods.
As we are dealing with millions of GS's, even small save can lead to significant loading time saves, so inlining these methods instead of using functions - does speed up it significantly.

slimbuck: Hey @Maksims,

Yeah you're right. Speed really matters, and especially when loading compressed splats (the runtime case).

This code is really for the general case of accessing data. It's used when we decompress the data for use in editors like supersplat. It would be almost impossible to "micro-optimise" it to run fast because of the suboptimal structure.

So I've added a fast path for the functions we use at load time instead - the aabb calc and the splat center extraction. Load time processing has gone from 640ms -> 60ms.

I wouldn't have done this extra work if it wasn't for your nudging - so thank you so much!! <3<3

src/scene/gsplat/gsplat-compressed-material.js
```diff
@@ -0,0 +1,377 @@
...
+
+import { hashCode } from "../../core/hash.js";
+import { ShaderUtils } from "../../platform/graphics/shader-utils.js";
+import { shaderChunks } from "../shader-lib/chunks/chunks.js";
+import { ShaderGenerator } from "../shader-lib/programs/shader-generator.js";
+import { ShaderPass } from "../shader-pass.js";
+
+const splatCoreVS = /* glsl */ `
+
+uniform mat4 matrix_model;
```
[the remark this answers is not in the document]
slimbuck: Yes there is some duplication, but I'd like to keep the two materials separate until the optimisations are final.

mvaligursky (approved):

src/scene/gsplat/gsplat-compressed-material.js
```diff
@@ -0,0 +1,361 @@
...
+
+const splatMainFS = `
+    void main(void)
+    {
+        gl_FragColor = evalSplat();
+    }
+`;
+
+/**
+ * @typedef {object} SplatMaterialOptions - The options.
```
willeastcott, line 317: Did you verify that this symbol doesn't make it to the API reference?

src/scene/gsplat/gsplat-compressed-material.js
```diff
@@ -0,0 +1,361 @@
...
+ * @property {string} [vertex] - Custom vertex shader, see SPLAT MANY example.
+ * @property {string} [fragment] - Custom fragment shader, see SPLAT MANY example.
+ * @property {string} [dither] - Opacity dithering enum.
+ */
+
+/**
+ * @param {SplatMaterialOptions} [options] - The options.
+ * @returns {Material} The GS material.
```
willeastcott, line 325: And this too...

src/scene/gsplat/gsplat-compressed.js
```diff
@@ -0,0 +1,155 @@
...
+        return result;
+    }
+
+    /**
+     * Evaluates the texture size needed to store a given number of elements.
+     * The function calculates a width and height that is close to a square
+     * that can contain 'count' elements.
+     *
+     * @param {number} count - The number of elements to store in the texture.
+     * @returns {Vec2} An instance of Vec2 representing the width and height of the texture.
```
willeastcott, line 123:
```suggestion
     * @returns {Vec2} The width and height of the texture.
```

willeastcott (approved):
🔥🔥🔥

src/scene/gsplat/gsplat-compressed-material.js
slimbuck: yep, it's not exported

src/scene/gsplat/gsplat-compressed-material.js
slimbuck: Yes, this function doesn't appear.

PR #6393, opened 2024-05-17
Support for Picker on WebGPU

mvaligursky:
- New public API for both **WebGL and WebGPU**:
```
// async version of the existing API returning a Promise resolving to an array of MeshInstances
Picker.getSelectionAsync(x, y, width, height)
```

- existing API (**WebGL** only, blocking):
```
Picker.getSelection(x, y, width, height)
```
example of use:
```
picker.getSelectionAsync(10, 20, 5, 5).then((meshInstances) => {
    console.log(meshInstances);
});
```
- updated the picker and gizmos example to use the async version on both platforms.

### DIFF
--- a/src/platform/graphics/texture.js
+++ b/src/platform/graphics/texture.js
@@ -2,7 +2,6 @@ import { Debug } from '../../core/debug.js';
 import { TRACEID_TEXTURE_ALLOC, TRACEID_VRAM_TEXTURE } from '../../core/constants.js';
 import { math } from '../../core/math/math.js';
 
-import { RenderTarget } from './render-target.js';
 import { TextureUtils } from './texture-utils.js';
 import {
     isCompressedPixelFormat,


src/platform/graphics/texture.js
```diff
@@ -1000,8 +1003,34 @@ class Texture {
...
+     * write the pixel data to. If not provided, a new buffer will be created. The type of the buffer
+     * must match the texture's format.
+     * @param {boolean} [options.immediate] - If true, the read operation will be executed as soon as
+     * possible. This has a performance impact, so it should be used only when necessary. Defaults
+     * to false.
+     * @returns {Promise<Uint8Array|Uint16Array|Uint32Array|Float32Array>} A promise that resolves
+     * with the pixel data of the texture.
+     * @ignore
+     */
+    read(x, y, width, height, options = {}) {
```
slimbuck, line 1032: Is `read` meant to supersede `downloadAsync`? Seems there is some duplication here. Couldn't there just be a single function that does it all?

Also, I prefer the name `downloadAsync` (to match `upload`) more than `read`.

src/framework/graphics/picker.js
```diff
@@ -76,50 +91,110 @@ class Picker {
      */
     getSelection(x, y, width = 1, height = 1) {
         const device = this.device;
+        if (device.isWebGPU) {
```
slimbuck, line 94: This seems real messy. I would prefer one way of getting texture data on both (even if one implementation must emulate the other).

src/platform/graphics/texture.js
mvaligursky: yes, downloadAsync is "deprecated" and will be removed. I left it there as there likely is some user of this.

`read` matches `StorageBuffer` `read` and `write`, `download` seems more internet term which seems confusing to use on the Texture.

src/framework/graphics/picker.js
mvaligursky: This is backwards compatibility, which a lot of users depends on, and so I left it there. Deprecating this would create a breaking change for no reason for now, people can move over to the newer API without us forcing it at this stage, as that could be a larger impact change. Including the Editor.

src/platform/graphics/texture.js
mvaligursky: I removed `downloadAsync` - only the model viewer seems to be using it, and we can easily swap it to use `read` when we update it to the new engine

src/framework/graphics/picker.js
mvaligursky: There is one way to get the data on both, called `Texture.read` - this is async and works on both WebGL2 and WebGPU.
Public API `Picker.getSelectionAsync` uses it, and works on both platforms.
Existing `Picker.getSelection` is left there an a synchronous WebGL2 only method, to avoid a breaking change, and prints an error on WebGPU platform, forcing people to upgrade at that point.

src/platform/graphics/texture.js
slimbuck: So you'll rename `upload` to `write` as well? You think the functions `read`/`write` describes what is happening?

mvaligursky: `write` function would take user data, set it on the object and upload internally.
`upload` just uploads already provided data
`read` rives user a copy of data for reading

I think `read` is a pretty descriptive and simple name for what the user wants to do - read the content of the texture. Note that for now this is a non-public API, and we can revise this in a follow up PRs.

slimbuck (approved):

PR #6407, opened 2024-05-21
[BREAKING] Remove deprecated AudioSourceComponent component

mvaligursky:
- this component has been deprecated and removed, and in the next engine version a SoundComponent has to be used instead.


slimbuck (approved):

willeastcott:
I think it's important to verify that loading a scene with an audiosource component just gracefully fails to add the component but carries on regardless (rather than crashing).

mvaligursky:
> I think it's important to verify that loading a scene with an audiosource component just gracefully fails to add the component but carries on regardless (rather than crashing).

Yep all seems good, we get this warning in the Editor but no problems / crashes:

Screenshot 2024-06-06 at 10 42 14

PR #6584, opened 2024-05-22
[BREAKING] Remove support for legacy scripts

mvaligursky:
- removed support for long time deprecated and disabled scripting system


kungfooman:
I guess we can remove `pc.script.attribute` aswell?

I'm not 100% certain, but I never used it and quick code search only seems to be related to legacy system:

https://github.com/search?q=%22pc.script.attribute%28%22&type=code

mvaligursky:
> I guess we can remove `pc.script.attribute` aswell?

yep, gone now, thanks

mvaligursky:
Tested against a project with the legacy scripts, and we get no crashes from it, the behaviour is the same as before removing the component.

PR #6588, opened 2024-05-22
Examples app bundle optimizations

### LINKED ISSUE #6368: [Examples] `npm run develop` slower/bigger than it has to be
https://github.com/playcanvas/engine/blob/4075e5b7c292d66ff0a872f4cb604468430298cf/examples/src/app/components/DeviceSelector.mjs#L1-L7

Previously we had:

```js
// Don't include all of 'playcanvas' for these defines, it just
// causes bigger bundles and prolongs the build time by ~3s.
import {
    DEVICETYPE_WEBGL1, DEVICETYPE_WEBGL2, DEVICETYPE_WEBGPU, DEVICETYPE_NULL
} from 'playcanvas/src/platform/graphics/constants.js';
```

(hence the comment)

For some reason it was refactored and the comment was neither read nor removed.

The Examples UI doesn't really require PlayCanvas, it's like a fancy UI to pick content for an `<iframe>`, as the only function is to render the UI. It depends on exactly three constants for the `DeviceSelector`. These three constants add like 2-3 seconds build time and append an unused extra of ~1.7MB to generate `examples/dist/index.js` on every `Ctrl+S` now.

So possible actions are:

1) Keep it slow and oversized
2) Reimplement how it was
3) Just write them out without messing around with the imports, for example:
```js
const DEVICETYPE_NULL = 'null';
const DEVICETYPE_WEBGL2 = 'webgl2';
const DEVICETYPE_WEBGPU = 'webgpu';
```
4) Any better ideas?

kpal81xd:
Fixes #6368

Minimised app bundle sizes:
- None: `3998454B`
- Treeshake smallest: `2240229B`
- No reference to playcanvas : `2349782B`


willeastcott (approved):

kungfooman:
Does this still parse the entirety of PlayCanvas and add 3s for each `Ctrl+S` during `npm run develop`?

mvaligursky:
> Does this still parse the entirety of PlayCanvas and add 3s for each `Ctrl+S` during `npm run develop`?

For last few weeks, the rebuild is under 1s on M1.

kungfooman:
> For last few weeks, the rebuild is under 1s on M1.

I see random times from 4-8s during development, so it definitely still affects people. It's nice that it's fast on M1 and the Apple file system, but I'm using e.g. ZFS for specific file system abilities.

kpal81xd:
> > For last few weeks, the rebuild is under 1s on M1.
> 
> I see random times from 4-8s during development, so it definitely still affects people. It's nice that it's fast on M1 and the Apple file system, but I'm using e.g. ZFS for specific file system abilities.

What system can you repro is this on?

kungfooman:
> What system can you repro is this on?

On Linux + ZFS, did a clean checkout: `gh pr checkout 6588`

image

The <2s is from manually updating code to:

```js
const DEVICETYPE_NULL = 'null';
const DEVICETYPE_WEBGL2 = 'webgl2';
const DEVICETYPE_WEBGPU = 'webgpu';
```

(back and forth to test timing (made a mistake hence the error), so we are still wasting all the CPU/IO for nothing basically)

kpal81xd:
> > What system can you repro is this on?
> 
> On Linux + ZFS, did a clean checkout: `gh pr checkout 6588`
> 
> image
> 
> The <2s is from manually updating code to:
> 
> ```js
> const DEVICETYPE_NULL = 'null';
> const DEVICETYPE_WEBGL2 = 'webgl2';
> const DEVICETYPE_WEBGPU = 'webgpu';
> ```
> 
> (back and forth to test timing (made a mistake hence the error), so we are still wasting all the CPU/IO for nothing basically)

Hmm okay replacing with just the constants might acc be better since on first build you dont need to wait to build the engine anymore just the types

kungfooman:
Nice work, just tested and it's way faster now :rocket:

PR #6589, opened 2024-05-22
Keep correct paths for build output even when build is invoked outside of repo root

liamdon:
This fixes a (possibly niche) problem that occurs when PlayCanvas is included in a monorepo with a central build system like [turborepo](https://turbo.build/) or [pnpm workspaces](https://pnpm.io/workspaces).

The problem: build output paths in the `build` directory are relative to the cwd of the shell executing the npm script. 

#### Example 
If you are using a monorepo-wide runner and have a structure like this:
`<root>/packages/vendor/playcanvas`

Then the expected output of `build/playcanvas/src/index.js` instead becomes `build/playcanvas/packages/vendor/playcanvas/src/index.js`. This causes subsequent build steps to fail. 

#### Solution
To fix the issue above, we find the actual repo root of `playcanvas` and set _this_ as the base path using the rollup option `preserveModulesRoot`. Now the output paths are correct regardless of where you run `npm run build` from.

-------

### DIFF
--- a/utils/rollup-build-target.mjs
+++ b/utils/rollup-build-target.mjs
@@ -18,6 +18,15 @@ import { version, revision } from './rollup-version-revision.mjs';
 import { getBanner } from './rollup-get-banner.mjs';
 import { babelOptions } from './rollup-babel-options.mjs';
 
+import { dirname, resolve as pathResolve } from 'path';
+import { fileURLToPath } from 'url';
+
+// Find path to the repo root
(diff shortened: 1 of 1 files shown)


kpal81xd:
@liamdon how are you building the engine? Because you can use `npm run <cmd> --prefix <dir>` to run a script referencing a particular path

kpal81xd (approved):

PR #6590, opened 2024-05-23
Gsplat rendering fix

slimbuck:
Fixes #

Changes in this PR:
* refrain from rendering the out-of-bounds splats.
* fix opacity decompress to handle 0 and 1.

### DIFF
--- a/src/scene/gsplat/gsplat-data.js
+++ b/src/scene/gsplat/gsplat-data.js
@@ -563,7 +563,7 @@ class GSplatData {
             data.f_dc_0[i] = (c.x - 0.5) / SH_C0;
             data.f_dc_1[i] = (c.y - 0.5) / SH_C0;
             data.f_dc_2[i] = (c.z - 0.5) / SH_C0;
-            data.opacity[i] = -Math.log(1 / c.w - 1);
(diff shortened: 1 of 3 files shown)


src/scene/gsplat/gsplat-data.js
```diff
@@ -563,7 +563,7 @@ class GSplatData {
             data.f_dc_0[i] = (c.x - 0.5) / SH_C0;
             data.f_dc_1[i] = (c.y - 0.5) / SH_C0;
             data.f_dc_2[i] = (c.z - 0.5) / SH_C0;
-            data.opacity[i] = -Math.log(1 / c.w - 1);
+            data.opacity[i] = (c.w <= 0) ? -40 : (c.w >= 1) ? 40 : -Math.log(1 / c.w - 1);
```
mvaligursky, line 566: it'd be good to comment on what -40 and 40 represents

mvaligursky (approved):

src/scene/gsplat/gsplat-data.js
slimbuck: Sure, good point thanks.

slimbuck: TBH I think it was a mistake to store scale & opacity values as exponents/sigmoid values the way they're stored in the source PLY data. We should be converting these to linear/direct values at load time, once, and storing these in the compressed format.

PR #6602, opened 2024-05-23
New texture formats: R8 and RG8

mvaligursky:
- exposed two new texture formats: `PIXELFORMAT_R8` and `PIXELFORMAT_RG8`
- both of these are supported on both WebGL2 and WebGPU, and are renderable
- these are a replacement for `PIXELFORMAT_L8` and `PIXELFORMAT_LA8` that are WebGL only and not renderable
- the old formats are kept for compatibility reason, as their behaviour is slightly different vs new formats (R is not automatically replicated into RGB when sampling and similar)


src/platform/graphics/constants.js
```diff
@@ -528,23 +528,24 @@ export const INDEXFORMAT_UINT16 = 1;
 export const INDEXFORMAT_UINT32 = 2;
 
 /**
- * 8-bit alpha.
+ * 8-bit alpha. Only supported on the WebGL. Use {@link PIXELFORMAT_R8} instead.
```
willeastcott, line 531:
```suggestion
 * 8-bit alpha. Only supported on WebGL. Use {@link PIXELFORMAT_R8} instead.
```

src/platform/graphics/constants.js
```diff
@@ -528,23 +528,24 @@ export const INDEXFORMAT_UINT16 = 1;
...
  *
  * @type {number}
  * @category Graphics
  */
 export const PIXELFORMAT_A8 = 0;
 
 /**
- * 8-bit luminance.
+ * 8-bit luminance. Only supported on the WebGL. Use {@link PIXELFORMAT_R8} instead.
```
willeastcott, line 539:
```suggestion
 * 8-bit luminance. Only supported on WebGL. Use {@link PIXELFORMAT_R8} instead.
```

src/platform/graphics/constants.js
```diff
@@ -528,23 +528,24 @@ export const INDEXFORMAT_UINT16 = 1;
...
  *
  * @type {number}
  * @category Graphics
  */
 export const PIXELFORMAT_L8 = 1;
 
 /**
- * 8-bit luminance with 8-bit alpha.
+ * 8-bit luminance with 8-bit alpha. Only supported on the WebGL. Use {@link PIXELFORMAT_RG8}
```
willeastcott, line 547:
```suggestion
 * 8-bit luminance with 8-bit alpha. Only supported on WebGL. Use {@link PIXELFORMAT_RG8}
```

willeastcott:
So should we deprecate L8 and LA8? Or alias them to R and RG?

mvaligursky:
> So should we deprecate L8 and LA8? Or alias them to R and RG?

We could deprecate them (hide the docs). I won't move that to deprecated as that needs a cleanup first.
Aliasing is not ideal (I started with that), as their behaviour is different.

Maksims:
> So should we deprecate L8 and LA8? Or alias them to R and RG?

LA8 format can be used by WebXR Depth Sensing as it can provide depth information either in R32F or LA8 formats.

mvaligursky:
> > So should we deprecate L8 and LA8? Or alias them to R and RG?
> 
> LA8 format can be used by WebXR Depth Sensing as it can provide depth information either in R32F or LA8 formats.

Yep, exactly the reason I didn't just remove it, as I've seen it used there. Does it support RB8 format by the way?

mvaligursky:
It seems from the WebXR API we just get a typed array .. so I guess internally at some point we can switch this to use RG texture format, to be eventually compatible with WebGPU when XR is there. But for now this should still just work.

willeastcott (approved):

Maksims:
> It seems from the WebXR API we just get a typed array .. so I guess internally at some point we can switch this to use RG texture format, to be eventually compatible with WebGPU when XR is there. But for now this should still just work.

There are two paths: one is CPU that provides data as a buffer, that we use to construct a texture.
Or second path is GPU accelerated (e.g. Quest 3) that currently provides texture as R32F, but specs state that it can be LA8.

Maksims:
Unfortunately, removal of LA8 did break depth sensing for mobile.

mvaligursky:
> Unfortunately, removal of LA8 did break depth sensing for mobile.

Ah thank's for the update. I changed my mind during the PR lifetime and only deprecated those formats, but I see I failed to change back two lines, making these deprecated formats unusable. Following PR will fix it.

mvaligursky:
done here now https://github.com/playcanvas/engine/pull/6649 @Maksims

PR #6623, opened 2024-05-24
Exports and types fixes

kpal81xd:
Refers to type exports https://forum.playcanvas.com/t/models-no-longer-cast-shadows-after-upgrading-playcanvas-engine/35857/11

- Adds separate exports for debug - `playcanvas/debug` and profiler - `playcanvas/profiler`
- Adds scripts to exports
- Cleans up ts and js config


willeastcott (approved):
Do you recommend we patch 1.71 because of this?

kpal81xd:
> Do you recommend we patch 1.71 because of this?

Yes I'm creating another PR for `main_v1`

PR #6630, opened 2024-05-25
Fix mesh collider creation

### LINKED ISSUE #6405: Cannot create collision mesh using accessors
The engine fails to create the collision mesh, when component's property accessors are used. This works:

```js
this.entity.addComponent('collision', {
    type: 'mesh',
    renderAsset: this.app.root.findByName('cube').render.asset
});
```

This doesn't:
```js
this.entity.addComponent('collision')
this.entity.collision.type = 'mesh';
this.entity.collision.renderAsset = this.app.root.findByName('cube').render.asset;
```

Repro:
https://playcanvas.com/project/1218921/overview/collision-mesh-issue

The issue is because base collision system implementation creates an empty model before initialization:
https://github.com/playcanvas/engine/blob/e0277bdc0d30f6c833a0dad125bcb7f45de8cc4c/src/framework/components/collision/system.js#L49-L54

Then when we try to assign a render asset, it first looks into model, which exists, but has no mesh instances and as a result fails to create a collision mesh:

https://github.com/playcanvas/engine/blob/e0277bdc0d30f6c833a0dad125bcb7f45de8cc4c/src/framework/components/collision/system.js#L450-L466

LeXXik:
Fixes #6405

The fix is to simply change the order of checks - first check render, then model.

### DIFF
--- a/src/framework/components/collision/system.js
+++ b/src/framework/components/collision/system.js
@@ -456,15 +456,7 @@ class CollisionMeshSystemImpl extends CollisionSystemImpl {
             const entityTransform = entity.getWorldTransform();
             const scale = entityTransform.getScale();
 
-            if (data.model) {
-                const meshInstances = data.model.meshInstances;
-                for (let i = 0; i < meshInstances.length; i++) {
-                    this.createAmmoMesh(meshInstances[i].mesh, meshInstances[i].node, shape, null, data.checkVertexDuplicates);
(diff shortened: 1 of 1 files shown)


kpal81xd:
Nice

kpal81xd (approved):

willeastcott (approved):

PR #6640, opened 2024-05-28
First person script update

kpal81xd:
- First person script updated to match editor template
- Updated example to use same assets as template

image


examples/assets/models/lowpoly__fps__tdm__game__map.glb
willeastcott, line 1: Can we rename this to `fps-map.glb` or something? 😄

willeastcott (approved):
So nice to see this example get some love!

PR #6641, opened 2024-05-29
Fixed trigger body not set to null once destroyed

MushAsterion:
When a trigger body is destroyed its reference is currently kept within the Trigger object until a new body is created.

Only changing this to remove the reference to the destroyed `Ammo.btRigidBody`.


willeastcott (approved):
Thanks!!

PR #6643, opened 2024-05-30
Fix transform not destroyed on Ammo Hull creation

MushAsterion:
When calling `CollisionComponentSystem#_getNodeTransform`, it creates an `Ammo.btTransform`.

This transform is already correctly destroyed on other use of the function but on the Hull implementation, fixing through this PR.


willeastcott (approved):
Well spotted. Thanks @MushAsterion! 🙏

PR #6644, opened 2024-05-30
Examples JSDoc tokenizer fixes

kpal81xd:
- JSDoc tokenizer fixes
- Inline type casting for canvas decl

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -45,10 +45,8 @@ Each example consists of two modules to define its behavior:
 ```js
 import * as pc from 'playcanvas';
 
-const canvas = document.getElementById('application-canvas');
-if (!(canvas instanceof HTMLCanvasElement)) {
-    throw new Error('No canvas found');
-}
+const canvas = /** @type {HTMLCanvasElement} */ (document.getElementById('application-canvas'));
+window.focus();
 
 const app = new pc.Application(canvas, {});
 
--- a/examples/src/examples/animation/blend-trees-1d.example.mjs
+++ b/examples/src/examples/animation/blend-trees-1d.example.mjs
@@ -2,10 +2,8 @@ import * as pc from 'playcanvas';
 import { data } from 'examples/observer';
 import { deviceType, rootPath } from 'examples/utils';
 
-const canvas = document.getElementById('application-canvas');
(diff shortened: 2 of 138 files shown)


examples/README.md
```diff
@@ -45,10 +45,8 @@ Each example consists of two modules to define its behavior:
 ```js
 
-const canvas = document.getElementById('application-canvas');
-if (!(canvas instanceof HTMLCanvasElement)) {
-    throw new Error('No canvas found');
-}
+const canvas = /** @type {HTMLCanvasElement} */ (document.getElementById('application-canvas'));
```
willeastcott, line 48: Why the outer brackets?

kpal81xd: Mantadory for jsdoc type casting

kpal81xd: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#casts

willeastcott: Can you put the JSDoc on the line above?

examples/src/examples/animation/blend-trees-1d.example.mjs
```diff
@@ -2,10 +2,8 @@ import * as pc from 'playcanvas';
 import { data } from 'examples/observer';
 import { deviceType, rootPath } from 'examples/utils';
 
-const canvas = document.getElementById('application-canvas');
-if (!(canvas instanceof HTMLCanvasElement)) {
-    throw new Error('No canvas found');
-}
+const canvas = /** @type {HTMLCanvasElement} */ (document.getElementById('application-canvas'));
```
kungfooman, line 5: Casting is not knowing... fake type safety that breaks at any refactor

examples/README.md
willeastcott: OK, I take it back. You can't. 👍

kpal81xd: No casting can only be inline since the jsdoc type is refering to the rhs of the expression not the left

willeastcott (approved):

examples/src/examples/animation/blend-trees-1d.example.mjs
kpal81xd: application-canvas is hard coded into the html file so its guaranteed.

PR #6645, opened 2024-05-30
Multi app fix

### LINKED ISSUE #6309: MultiApp Example hot reload WebGL2 not rendering
After hot reload WebGL2 app not rendering

image

### Notes
- WebGL2 and WebGPU work independently just not together
- WebGL2 appears to render inside the same context? (two cubes showing on WebGPU app)
- Adding delay of 100ms between device creation resolves the issue

### LINKED ISSUE #6388: Question on multi-app
There was a question on Discord, so decided to put it here, as I am not sure if it is an error. If you change the multi-app example to start with two WebGL2 instances, only the last one starts. Is this an expected behavior? If set to WebGPU - both apps start normally.

image

image

kpal81xd:
Fixes #6388
Fixes #6309

- Wraps add app in async function so can await on app creation
- Passed in application when creating entity (See below for explanation)

The application passed into the entity is used to register new component systems of which one is the `CameraComponent`. The camera component's layers are then used to construct the render passes. If there is an existing application the current application reference will be set to the first application every update frame. Therefore if you construct a new application without starting it and create a new entity it will use the first application reference and in turn render using its graphics device in turn.

### DIFF
--- a/examples/iframe/utils.mjs
+++ b/examples/iframe/utils.mjs
@@ -105,6 +105,11 @@ export function updateDeviceType(config) {
         return;
     }
 
+    if (config.WEBGL_DISABLED && config.WEBGPU_DISABLED) {
+        console.warn('Both WebGL and WebGPU are disabled. Using NullGraphicsDevice instead.');
+        deviceType = 'null';
+        return;
+    }
     if (config.WEBGPU_DISABLED) {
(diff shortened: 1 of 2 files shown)


willeastcott (approved):
Great to have this fixed! 👏

PR #6648, opened 2024-06-03
Allow components ordering

### LINKED ISSUE #6647: Allow entity components enabling order on hierarchy change
When working with custom components (e.g. added via code), we need some components to be enabled before others. At the moment the components are called in random order:

https://github.com/playcanvas/engine/blob/55a3ff8845f0cf30829aeec515f24ac98b6baf4d/src/framework/entity.js#L503-L520

For example, a body component and a constraint component for physics. The constraint component needs a body to be attached to. If the body component is enabled later, it will be added to the physics after the constraint, which will fail creating the constraint as it will fail to locate the body.

There is no elegant way of doing it today. We have to introduce a component dependency into another component that actually doesn't need to be aware if other components and their state. We also have to add some sort of systems to queue up the creation commands to wait for the other component to be enabled, which is pretty brittle.

LeXXik:
Fixes #6647 

The PR allows components to specify in which order they get enabled when multiple components are present on an Entity.


src/framework/components/system.js
```diff
@@ -50,6 +51,9 @@ class ComponentSystem extends EventHandler {
         entity[this.id] = component;
         entity.c[this.id] = component;
 
+        entity.oc.push(component);
```
mvaligursky, line 54: it seems you're only adding to the `oc` array, but never removing from it?

LeXXik: Completely forgot about it :D

LeXXik:
There is a potential (future?) conflict with a script component, which I noticed have the following line:

https://github.com/playcanvas/engine/blob/55a3ff8845f0cf30829aeec515f24ac98b6baf4d/src/framework/components/script/component.js#L228-L232

However, I am not sure what that is.

mvaligursky:
I'm just wondering if this is too much. This allows sorting per component instance, with the order being stored for each instance of a component. I suspect that in reality, we need sorting based on the component type, and so we can avoid storing this per component instance, and store it as a static member of the component type.

Additionally, I think currently the array gets sorted each time a component is added. It's likely not a problem, as typically we only have a small number of components per entity. And each entity stores the sorted array as well, which is not ideal.

We could have a getter to get sorted components, and sort them at that point, as this is called at low frequency, and would avoid storing the array.

Thoughts? Just trying to make this as lightweight as possible.

LeXXik:
Agreed, good points.

LeXXik:
Having an order as a static is a better option. Thanks!

I tried to remove the sorted array store and move it to a getter, but not sure it is better than before. Now it gets sorted each time when `onHierarchyStateChanged` gets called. Instead of only once, when a component was added.

LeXXik:
How about this variant?

src/core/sort.js
```diff
@@ -6,10 +6,27 @@
  */
 const cmpPriority = (a, b) => a.priority - b.priority;
 
+
+/**
+ * @param {{order: number}} a - First object with `order` property.
```
mvaligursky, line 11: change `property` to `static property` to make it clear please

src/core/sort.js
```diff
@@ -6,10 +6,27 @@
...
+ * @param {{order: number}} b - Second object with `order` property.
+ * @returns {number} A number indicating the relative position.
+ * @ignore
+ */
+const cmpOrder = (a, b) => a.constructor.order - b.constructor.order;
```
mvaligursky, line 16: and maybe even rename it to `cmpStaticOrder` for clarity

src/framework/entity.js
```diff
@@ -601,6 +604,24 @@ class Entity extends GraphNode {
         return clone;
     }
 
+    _getSortedComponents() {
+        _sortedArray.length = 0;
+
+        let needSort = false;
+        for (const component of Object.values(this.c)) {
```
mvaligursky, line 611: use this loop as before, to avoid the array allocation inside `Object.values`:
```
        for (const type in components) {
            if (components.hasOwnProperty(type)) {
```

src/framework/entity.js
```diff
@@ -601,6 +604,24 @@ class Entity extends GraphNode {
...
+            if (component.constructor.order !== 0) {
+                needSort = true;
+            }
```
mvaligursky, line 614: avoid `if`:
```
needSort |= component.constructor.order !== 0;
```

src/framework/entity.js
```diff
@@ -601,6 +604,24 @@ class Entity extends GraphNode {
...
+            _sortedArray.push(component);
+        }
+
+        if (needSort) {
```
mvaligursky, line 618: Could be a tiny speed boost

```suggestion
        if (needSort && _sortedArray.length > 1) {
```

mvaligursky:
This feels right to me. I added some small comments.
I also would not mind few units tests to be added to test this if possible.

Maksims:
Will it be more efficient to insert a component into its correct position at the moment of adding it, and avoid sorting altogether?

mvaligursky:
> Will it be more efficient to insert a component into its correct position at the moment of adding it, and avoid sorting altogether?

The component are not stored in an array, but an object. And we're trying to avoid storing additional array per entity.

src/core/sort.js
```diff
@@ -6,10 +6,27 @@
...
  */
 export const sortPriority = arr => arr.sort(cmpPriority);
+
+/**
+ * @param {Array<{order: number}>} arr - Array to be sorted in place where each element contains
+ * an object with a static `order` property.
+ * @returns {Array<{order: number}>} In place sorted array.
+ * @ignore
+ */
+export const sortOrder = arr => arr.sort(cmpStaticOrder);
```
mvaligursky, line 32: also rename to `sortStaticOrder`

mvaligursky (approved):
One last tiny rename suggestion, but otherwise awesome, thanks!

mvaligursky (approved):

Issue #6667, opened 2024-06-05
Visual artifacts on WebGPU


LeXXik:
There is a white line visible on mesh, when using WebGPU:

Repro: https://playcanvas.com/project/1225964/overview/webgpu-shader-issue

![image](https://github.com/playcanvas/engine/assets/5677782/66796250-02c5-4d13-bea2-b68381c69287)

![image](https://github.com/playcanvas/engine/assets/5677782/b58123f9-8486-456f-9626-50d325bc8fc7)

mvaligursky:
This is related to us changing some texture settings after the texture has been created, which WebGPU does not handle (yet). Relevant warning:

<img width="870" alt="Screenshot 2024-06-06 at 09 49 04" src="https://github.com/playcanvas/engine/assets/59932779/bc7cc406-24d5-4f17-b170-0ea8fa093e95">

PR #6669, opened 2024-06-06
Cubemap fix for webgpu

slimbuck:
Fixes https://github.com/playcanvas/engine/issues/6667

This PR changes how we specify envAtlas texture rendering state when loading cubemap assets.

Instead of overriding states after the fact, we supply the correct values at load time.


src/framework/handlers/cubemap.js
```diff
@@ -317,7 +313,15 @@ class CubemapHandler extends ResourceHandler {
                     url: assetId,
                     filename: assetId
                 } : assetId;
-                texAsset = new Asset(cubemapAsset.name + '_part_' + i, 'texture', file);
+
+                const data = file.url.search('.dds') === -1 ? {
```
mvaligursky, line 317: add a comment pls

mvaligursky (approved):

src/framework/handlers/cubemap.js
slimbuck: done

PR #6657, opened 2024-06-04
RenderPass based SSAO

mvaligursky:
- implementation of SSAO using render passes
- few small improvements compared to the implementation using the post-effects (increasing number of samples does not modify the brightness, exposed additional controls ..)

no SSAO
Screenshot 2024-06-04 at 16 28 04

un-blured SSAO
Screenshot 2024-06-04 at 16 28 14

blurred SSAO
Screenshot 2024-06-04 at 16 28 23

Issue #6658, opened 2024-06-04
Improve render-pass based SSAO


mvaligursky:
initial implementation: https://github.com/playcanvas/engine/pull/6657

possible improvements:
- currently the SSAO depends on the depth-prepass. This could be optional and depth-drab pass can be used instead to obtain the depth.
- when the SSAO uses the depth-pass, it runs before the main forward passes of the scene. This allows us to implement https://github.com/playcanvas/engine/issues/4844 properly. 
- when SSAO is used together with TAA, randomize sampling using TAA's random values, to allow smaller number of SSAO samples to be used, and depending on TAA instaed of resolve them over few frames
- improve the shaders to use half precision to gain some performance.
- consider downscaling the depth buffer before the SSAO runs, or at least generate mipmaps from it. The SSAO shader we use allows sampling using mipmaps, to allow better quality larger size filter.

PR #6684, opened 2024-06-10
RenderPassSsao: improve SSAO blur performance

querielo:
This PR suggests splitting one large blur pass from "[RenderPass based SSAO](https://github.com/playcanvas/engine/pull/6657)" into four smaller blur passes without compromising quality.

| Main | PR |
| -----|----|
| Screenshot 2024-06-10 at 22 55 37 | Screenshot 2024-06-10 at 22 54 22 |

Tested on MacBook 14" Pro (2021), used `device.maxPixelRatio = window.devicePixelRatio;` to increase resolution of framebuffers.


mvaligursky:
Hi @querielo - that's definitely a good way to speed it up. But I wonder why doing 4 passes instead of just typical two separable passes?

querielo:
Hi. @mvaligursky 

The main idea is that the first two passes weaken the high-frequency signal. The next two passes are used to eliminate low-frequency signal (there are strided/interleaved blurs with a large step, applied diagonally).

Experimentally, it seems that increasing the filter kernel size is necessary to remove low-frequency SSAO patterns. For example, using a 17x17 kernel on my computer results in a noticeable slowdown, but it appears to achieve the effect of four passes.

The suggested four-pass approach is just one way to implement blurring. An advanced developer could add `RenderPassDepthAwareBlur` to `afterPasses` themselves, depending on the specific SSAO pattern generated.

| 4 passes | 2 passes, 11x11 | 2 passes, 17x17 |
| -----|----|----|
| Screenshot 2024-06-11 at 21 12 36 | Screenshot 2024-06-11 at 21 39 57 | Screenshot 2024-06-11 at 21 11 34 |

mvaligursky:
I had a bit of a play with your branch. My findings:
- two pass BOX blur with 9 samples seems to be equivalent to the current 9x9 filter, as expected. I assume this is faster.
- two pass GAUSSIAN blur with 9 samples is a lot lower quality compared to that - see especially around the yellow torches
- four pass you had set up seems to match in quality with the two pass BOX with 9 samples - so I'm not sure we need to use 4 passes.

I think the main reason you need 4 passes is the use of the GAUSSIAN weights instead of BOX.

mvaligursky:
And so my recommendation is: switch it to 2 pass BOX filtering. We can expose the number of taps to the user as that controls the quality vs the cost.

querielo:
@mvaligursky I switched the blur to 2 pass BOX filtering.

src/core/math/math.js
```diff
@@ -277,6 +277,42 @@ const math = {
...
         return inclusive ? num >= min && num <= max : num > min && num < max;
+    },
+
+    /**
+     * Computes the Gaussian function value for a given input.
+     *
+     * @param {number} x - The input value for which the Gaussian function is to be calculated.
+     * @param {number} sigma - The standard deviation of the distribution.
+     * @returns {number} - The Gaussian function value.
+     */
```
mvaligursky, line 288: add `@ignore`, no need for this to be public

src/core/math/math.js
```diff
@@ -277,6 +277,42 @@ const math = {
...
+    gauss(x, sigma) {
+        return Math.exp(-(x * x) / (2.0 * sigma * sigma));
+    },
+
+    /**
+     * Generates Gaussian weights for a given kernel size.
+     *
+     * @param {number} kernelSize - The size of the kernel for which the Gaussian weights are to be computed.
+     * @returns {number[]} - An array containing the normalized Gaussian weights.
+     */
```
mvaligursky, line 298: add `@ignore`, no need for this to be public

querielo:
Offtop: It looks like 6f29e1b005f2ab2a38f75fb45739387561d1d50a breaks the background.
Screenshot 2024-06-12 at 13 25 18

src/extras/render-passes/render-pass-ssao.js
```diff
@@ -308,6 +351,8 @@ class RenderPassSsao extends RenderPassShaderQuad {
     }
 
     createRenderTarget(name) {
+        // TODO: consider using a pool of 2 texture buffers
```
mvaligursky, line 354: considering we're down to 2 blurs, remove the comment

src/extras/render-passes/render-pass-depth-aware-blur.js
```diff
@@ -8,18 +11,87 @@ import { shaderChunks } from '../../scene/shader-lib/chunks/chunks.js';
...
+     * @param {import('../../platform/graphics/texture.js').Texture} [options.resizeSource] - The source texture to be blurred.
+     * @param {number} [options.scaleX] - The scale factor for the render target width. Defaults to 1.
+     * @param {number} [options.scaleY] - The scale factor for the render target height. Defaults to 1.
+     * @param {number} [options.kernelSize] - The size of the blur kernel. Defaults to 7.
+     * @param {boolean} [options.depthAware] - Whether the blur should be depth-aware. Defaults to false.
+     * @param {BLUR_GAUSSIAN|import('../../scene/constants.js').BLUR_BOX} [options.type] - The type of blur to apply. Defaults to BLUR_GAUSSIAN.
+     * @param {Vec2} [options.direction] - The direction of the blur. Defaults to (1, 0).
+     * @param {string} [options.channels] - The color channels to apply the blur to ('r'|'g'|'b'|'a'|'rg'|..|'ba'|'rgb'|'gba'|'rgba'). Defaults to 'rgba'.
+     */
+    init(renderTarget = null, options = {}) {
```
mvaligursky, line 29: I would not override the `init` function, but instead add a `setup` function which does all this.

src/extras/render-passes/render-pass-depth-aware-blur.js
```diff
@@ -8,18 +11,87 @@ import { shaderChunks } from '../../scene/shader-lib/chunks/chunks.js';
...
+
+        const kernelWidth = kernelSize / 2 | 0;
+        const middle = kernelWidth + (1 - kernelSize % 2);
+
+        loop += `
+                sum += texture2D(sourceTexture, uv0).${channels} * ${weightCoefs[middle].toFixed(4)};
+                totalWeight += ${weightCoefs[middle].toFixed(4)};`;
+
+        // TODO: move calculating UV coordinates to the vertex shader and pass them as varying
+        for (let i = 1; i <= kernelWidth; i++) {
```
mvaligursky, line 66: we're trying to move away from generating shader from javascript as much as possible, as it's a lot harder to understand and modify. We try to create a single shader string as much as possible, and then in code generate a list of defines to pre-pend. See `RenderPassCompose` as an example. It'd be great to modify this in a similar fashion.

mvaligursky: even though the shader here is trying to be pretty generic, which makes it harder (with the types and channels)

mvaligursky:
> Offtop: It looks like [6f29e1b](https://github.com/playcanvas/engine/commit/6f29e1b005f2ab2a38f75fb45739387561d1d50a) breaks the background. 

It just got brighter as a result of this https://github.com/playcanvas/engine/pull/6687 I suspect?

mvaligursky:
Looking much better, thanks!

I'd suggest to remove the option / API to change the blur type. Box looks better than Gaussian, and so the SSAO should just use Box. Only expose values that the user would benefit from adjusting.

src/extras/render-passes/render-pass-depth-aware-blur.js
```diff
@@ -1,93 +1,197 @@
...
-            }
+        const { width, height } = this.sourceTexture;
+        sourceInvResolutionValueTmp[0] = 1.0 / width;
+        sourceInvResolutionValueTmp[1] = 1.0 / height;
 
-            float bilateralWeight(in float depth, in float sampleDepth) {
-                float diff = (sampleDepth - depth);
-                return max(0.0, 1.0 - diff * diff);
-            }
+        this.sourceInvResolutionId?.setValue(sourceInvResolutionValueTmp);
```
mvaligursky, line 110: no need for those `?` there as those are transpiled to `if` .. we always set those up, so they're never undefined.

src/extras/render-passes/render-pass-depth-aware-blur.js
```diff
@@ -1,93 +1,197 @@
...
-        this.sourceInvResolutionValue[0] = 1.0 / width;
-        this.sourceInvResolutionValue[1] = 1.0 / height;
-        this.sourceInvResolutionId.setValue(this.sourceInvResolutionValue);
+    updateShader() {
+        const shaderName = `BlurShader-kernelsize-${this.kernelSize}`;
 
-        super.execute();
+        const defines = `#define KERNEL_SIZE ${this.kernelSize}\n`;
+
+        // CHECK: should we destroy the shader?
```
mvaligursky, line 178: shaders are expensive to compile, so we don't destroy them, to make it faster when the shader is needed again

src/extras/render-passes/render-pass-depth-aware-blur.js
```diff
@@ -1,93 +1,197 @@
...
+            return;
+        }
 
-                float ao = sum / totalWeight;
+        this._kernelSize = Math.max(1, Math.floor(value));
 
-                // simple dithering helps a lot (assumes 8 bits target)
-                // this is most useful with high quality/large blurs
-                // ao += ((random(gl_FragCoord.xy) - 0.5) / 255.0);
+        this.updateShader();
```
mvaligursky, line 142: ideally don't call updateShader directly, just set a _shaderDirty flag - this makes it easy to add more properties that modify the shader, and also better handle the case where the property is changed multi times per frame. (not that this is typical, but happens).

querielo:
> I'd suggest to remove the option / API to change the blur type. Box looks better than Gaussian, and so the SSAO should just use Box. Only expose values that the user would benefit from adjusting.

Are you certain?
Stage: [https://engine-m742xowmh-playcanvas.vercel.app/#/graphics/ambient-occlusion](https://engine-m742xowmh-playcanvas.vercel.app/#/graphics/ambient-occlusion)
To me, it seems that enlarging the kernel size of Box displaces dark areas from their correct position. However, Gaussian necessitates a larger kernel to conceal the SSAO pattern.

https://github.com/playcanvas/engine/assets/104348270/fdf04009-01de-4ffb-8765-94e13c8a9a20

src/extras/render-passes/render-pass-depth-aware-blur.js
querielo: VS Code TS checker highlighted this error, suggesting that sourceInvResolutionId could be undefined as it is not defined during creation. I'll check if Playcanvas linter highlights it.

src/extras/render-passes/render-pass-depth-aware-blur.js
querielo: Then I will remove the comment.

mvaligursky:
> @mvaligursky Are you certain? Stage: https://engine-m742xowmh-playcanvas.vercel.app/#/graphics/ambient-occlusion To me, it seems that enlarging the kernel size of Box displaces dark areas from their correct position more quickly than hiding SSAO artifacts. However, Gaussian necessitates a larger kernel to conceal the SSAO pattern.

yeah interesting. It almost feel like the the bilateral weight function should be improved

```
float bilateralWeight(in float depth, in float sampleDepth) {
    float diff = (sampleDepth - depth);
    return max(0.0, 1.0 - diff * diff);
}
```

to detect the depth discontinuity better, as currently it blurs over those small edges. Maybe we can try something like this to have a control over it by the sigma value:

```
float bilateralWeight(in float depth, in float sampleDepth, in float sigma) {
    float diff = (sampleDepth - depth);
    return exp(-diff * diff / (2.0 * sigma * sigma));
}
```
but agreed, the Gaussian blur in general should be slightly better, at a higher cost, so maybe leave it in.

mvaligursky:
Hi @querielo - have you had some time to look at these?

mvaligursky:
@querielo - please let me know if you have time to finished this PR. I'm keen to continue on some improvements too (https://github.com/playcanvas/engine/issues/6658) but would prefer to avoid larger conflicts with the changes.

If you don't have time, I can take over this PR.

MAG-AdrianMeredith:
> @querielo - please let me know if you have time to finished this PR. I'm keen to continue on some improvements too (#6658) but would prefer to avoid larger conflicts with the changes.
> 
> If you don't have time, I can take over this PR.

guess thats a no...

mvaligursky:
yep, it's on my list to get to in a week or so. Definitely not forgotten.

querielo:
Yes. Sorry, I'm too busy right now and therefore don't have time to fix it.

mvaligursky:
closing this due to https://github.com/playcanvas/engine/pull/6870
Thanks @querielo

PR #6685, opened 2024-06-10
Add JSDoc for getters

### LINKED ISSUE #6652: HIde undocumented setters and getters from public API
It seems those are currently visible as public API. Example in the CameraComponent:

https://api.playcanvas.com/classes/Engine.CameraComponent.html#renderSceneColorMap

check the code it links .. there is no JSDocs block, but it's still exposed as public on the docs.

Screenshot 2024-06-03 at 15 39 57

willeastcott:
Fixes #6652

Before:
Any undocumented symbols would be ignored in the API reference by TypeDoc _except_ getters, because we only used to document the setter. This meant that some deliberately undocumented getters were making it into the API.

Now:
TypeDoc is now configured to ignore _all_ undocumented symbols. Even getters. To handle this, this PR documents all getters considered to be part of the API.

This PR also improves on and fixes many formatting issues in the getter/setter API. And we no longer have to hack things to get TypeDoc to generate a valid API reference.


src/framework/components/screen/component.js
```diff
@@ -231,13 +242,19 @@ class ScreenComponent extends Component {
         this._elements.forEach(element => element._onScreenSpaceChange());
     }
 
+    /**
+     * Gets whether the ScreenComponent will render its child {@link ElementComponent}s in screen
+     * space instead of world space.
```
kungfooman, line 247: I saw you added dashes to "object space" everywhere, should screen space and world space have the same treatment?

willeastcott: I've just done a Google search and it seems more common to not use a hyphen. So I've backed all that out.

kungfooman: Thank you, there still is `object-space` now?

willeastcott: Oh yeah! Forgot about that one. Thanks. Fixed.

src/core/shape/oriented-box.js
LeXXik, line 1: Hmm, so if we take `worldTransform` as a property to look at. Do both `_worldTransform` and accessors need to specify the type?

willeastcott: If you don't specify the type on the `get`, ESLint complains about the JSDoc block having no `@returns` tag. The TypeScript compiler `tsc` can probably infer types...but what we have here works.

LeXXik: How about removing it from the `_property` then? It seems the whole block is no needed there.

willeastcott: I think you're right about that. But this PR is already enormous and at least it works. So I would rather get this merged and then any refinement can happen in a follow up PR.

src/extras/gizmo/transform-gizmo.js
```diff
@@ -381,59 +381,79 @@ class TransformGizmo extends Gizmo {
...
+     * Sets whether snapping is enabled. Defaults to false.
      *
      * @type {boolean}
      */
     set snap(value) {
         this._snap = this.root.enabled && value;
     }
 
+    /**
+     * Gets whether snapping is enabled. Defaults to false.
```
LeXXik, line 393: ... whether or not .... ?

src/core/shape/oriented-box.js
kungfooman: As long there is still a constructor assignment to infer the types from the type comment block seems superfluous indeed, but if there isn't any constructor assignment we could decrease Intellisense type covering for developers (not sure how many cases that would be overall, probably a few).

src/extras/gizmo/transform-gizmo.js
willeastcott: I think that's implicit.

src/framework/asset/asset.js
```diff
@@ -302,14 +312,19 @@ class Asset extends EventHandler {
         }
     }
 
+    /**
+     * Gets the file details or null if no file.
+     *
+     * @type {object}
```
LeXXik, line 318: {object | null}

src/framework/asset/asset.js
```diff
@@ -357,12 +382,18 @@ class Asset extends EventHandler {
...
+     *
+     * @type {object[]}
+     */
     get resources() {
         return this._resources;
     }
 
     /**
-     * If true the asset will be loaded during the preload phase of application set up.
+     * Sets whether to preload an asset. If true the asset will be loaded during the preload phase
```
LeXXik, line 395: If true, the ...

willeastcott: Fixed.

src/framework/components/collision/component.js
```diff
@@ -201,12 +224,17 @@ class CollisionComponent extends Component {
...
+     *
+     * @type {Quat}
+     */
     get angularOffset() {
         return this.data.angularOffset;
     }
 
     /**
-     * The radius of the sphere, capsule, cylinder or cone-shaped collision  volumes.
+     * Sets the radius of the sphere, capsule, cylinder or cone-shaped collision  volumes.
```
LeXXik, line 237: double space before "volumes"

src/framework/components/collision/component.js
```diff
@@ -215,12 +243,17 @@ class CollisionComponent extends Component {
         this._setValue('radius', arg);
     }
 
+    /**
+     * Gets the radius of the sphere, capsule, cylinder or cone-shaped collision  volumes.
```
LeXXik, line 247: Also here

src/framework/asset/asset.js
willeastcott: There are literally hundreds of these kinds of corrections needed across the codebase. I'd rather leave fixing those out of this PR to simplify things for now (although you might see I have make one or two fixes as I went through documenting the getters). I'd say we just leave them to a follow up PR.

src/framework/components/collision/component.js
```diff
@@ -229,55 +262,76 @@ class CollisionComponent extends Component {
...
+     *
+     * @type {Asset|number|null}
+     */
     get asset() {
         return this.data.asset;
     }
 
     /**
-     * The render asset of the mesh collision volume - can also be  an asset id. Defaults to null.
+     * Sets the render asset or asset id of the mesh collision volume. Defaults to null.
```
LeXXik, line 314: "Defaults to null" doesn't seem right on a setter. Should be on a getter probably.

src/framework/components/collision/component.js
willeastcott: Well spotted! Thanks. Fixed.

src/framework/components/collision/component.js
willeastcott: Fixed.

src/framework/components/collision/component.js
```diff
@@ -636,7 +697,8 @@ class CollisionComponent extends Component {
     }
 
     /**
-     * @description Returns the world position for the collision shape taking into account of any offsets.
+     * Returns the world position for the collision shape taking into account of any offsets.
```
LeXXik, line 701: ...shape, taking...

src/framework/components/collision/component.js
```diff
@@ -654,7 +716,8 @@ class CollisionComponent extends Component {
     }
 
     /**
-     * @description Returns the world rotation for the collision shape taking into account of any offsets.
+     * Returns the world rotation for the collision shape taking into account of any offsets.
```
LeXXik, line 720: shape, taking

willeastcott: Fixed!

src/framework/components/collision/component.js
willeastcott: Fixed!

src/framework/components/element/component.js
```diff
@@ -796,7 +866,8 @@ class ElementComponent extends Component {
     }
 
     /**
-     * The width of the text rendered by the component. Only works for {@link ELEMENTTYPE_TEXT} types.
+     * Gets the width of the text rendered by the component. Only works for
+     * {@link ELEMENTTYPE_TEXT} types.
```
LeXXik, line 870: You can safely break links to new line for tighter line fill
```
... for {@link
ELEMENTTYPE_TEXT} types
```

src/framework/components/element/component.js
```diff
@@ -1027,14 +1132,21 @@ class ElementComponent extends Component {
     }
 
     /**
-     * The minimum size that the font can scale to when autoFitWidth or autoFitHeight are true.
+     * Sets the minimum size that the font can scale to when `autoFitWidth` or `autoFitHeight` are
```
LeXXik, line 1135: good place to use a link {@link autoFitWidth} or {@link autoFitWidth | Auto Fit Width}

src/framework/components/gsplat/component.js
```diff
@@ -69,10 +69,9 @@ class GSplatComponent extends Component {
     }
 
     /**
-     * If set, the object space bounding box is used as a bounding box for visibility culling of
-     * attached gsplat. This allows a custom bounding box to be specified.
+     * Sets a custom object space bounding box for visibility culling of the attached gsplat.
      *
-     * @type {import('../../../core/shape/bounding-box.js').BoundingBox}
+     * @type {import('../../../core/shape/bounding-box.js').BoundingBox|null}
```
LeXXik, line 74: space before and after |

src/framework/components/gsplat/component.js
```diff
@@ -81,13 +80,19 @@ class GSplatComponent extends Component {
...
+     */
     get customAabb() {
         return this._customAabb;
     }
 
     /**
-     * A {@link GSplatInstance} contained in the component. If not set or loaded, it returns null.
+     * Sets a {@link GSplatInstance} on the component. If not set or loaded, it returns null.
      *
+     * @type {import('../../../scene/gsplat/gsplat-instance.js').GSplatInstance|null}
```
LeXXik, line 95: here as well

src/framework/components/gsplat/component.js
```diff
@@ -118,6 +123,12 @@ class GSplatComponent extends Component {
         }
     }
 
+    /**
+     * Gets the {@link GSplatInstance} on the component.
+     *
+     * @type {import('../../../scene/gsplat/gsplat-instance.js').GSplatInstance|null}
```
LeXXik, line 129: and here

LeXXik: Hmm, multiple instances. Perhaps a different PR for those.

mvaligursky: we have lots of instances of non-spaces around `|` in the engine, I'm not sure we have a strict rule around that at the moment.

src/framework/components/element/component.js
willeastcott: Done! I have added lots of other links to this module too.

src/framework/components/gsplat/component.js
willeastcott: Yeah, we can standardize something in another PR.

src/scene/mesh-instance.js
```diff
@@ -370,12 +370,17 @@ class MeshInstance {
...
+     *
+     * @type {number}
+     */
     get renderStyle() {
         return this._renderStyle;
     }
 
     /**
-     * The graphics mesh being instanced.
+     * Sets the graphics mesh being instanced.
```
mvaligursky, line 383: This makes it sounds like mesh instancing is taking place here, I wonder if we could use clearer wording.

mvaligursky (approved):
Huge effort, even scrolling through this took a while!

src/scene/mesh-instance.js
willeastcott: It's the same wording as before, so in the interest in getting this merged, I'll save that for a follow up PR. 😉

PR #6686, opened 2024-06-11
[Postprocessing example] Scale down TAA jitter based on renderPassCamera.renderTargetScale

liamdon:
At a lower resolution scale, the jitter becomes very obvious. I noticed that when we scale it down by the resolution scale, and it looks a lot better. At first I tried a linear scale (`jitter * resolution`), but using a squared resolutionScale looks even better imo. I'm not sure why, though!

My question for @mvaligursky is: should `RenderPassCameraFrame` scale down the jitter for you based on the render scale? Or is this something we'd leave up to application developers?

### DIFF
--- a/examples/src/examples/graphics/post-processing.controls.mjs
+++ b/examples/src/examples/graphics/post-processing.controls.mjs
@@ -17,7 +17,7 @@ export function controls({ observer, ReactPCUI, React, jsx, fragment }) {
                     binding: new BindingTwoWay(),
                     link: { observer, path: 'data.scene.scale' },
                     min: 0.2,
-                    max: 1,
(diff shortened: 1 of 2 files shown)


mvaligursky:
Based on my testing, I'd prefer we leave this up to the users to set up, instead of forcing internally - as it lets them balance the sharpness vs aliasing.

But as you've done, I think the example is fine to pre-set this based on the resolution, but only when the user changes the resolution. The user should still be able to adjust this slider separately. I've done this for the bias in the AmbientOcclusion example as well this way, if the user changes scaling, I change the bias to avoid issues, but still allow user to change the bias.

PR #1921, opened 2020-03-15
Optimize BoundingBox and fix inaccuracy

aidinabedi:
Fixes #1903

PR includes the following:
- Fix documentation for all public `pc.BoundingBox` methods (`copy`, `clone`, `compute`, `setMinMax`).
- Optimize `pc.BoundingBox` for both `min`/`max` and `center`/`halfExtents` by always keeping in sync.
- Support arrays (of length 3) as paramters (in addition to `pc.Vec3`) in `pc.BoundingBox#setMinMax`.
- Use new optimized `pc.BoundingBox#setMinMax` in `glb-parser.js` and `json-model.js` parser.
- Add new properties `min` and `max` to `pc.BoundingBox` (without re-calcuating on every get).
- Deprecate old "getters" (`getMin` and `getMax`) in `pc.BoundingBox`, while keeping backward-compatibility.

**PLEASE NOTE:** The behavior of `pc.BoundingBox` is still exactly the same. Only the internals have been improved. This PR is fully backward-compatible.

----
I confirm I have signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

Issue #1903, opened 2020-03-04
BoundingBox inaccuracy


RacoAnimech:
### Description

```pc.BoundingBox``` stores two vectors: ```center``` and ```halfExtents```. The minimum and maximum values can be calculated from these values using ```getMin``` and ```getMax```. The problem with this is rounding errors. The currently only way to get the actual minimum and maximum is to loop over the vertexbuffer and calculate them manually.

To fix this problem ```pc.BoundingBox``` has to be remade to store the minimum and maximum values and calculate center and halfextents from them.

### Steps to Reproduce

Run this code in any project.
```js
let min = new pc.Vec3(0.3,1,1);
let max = new pc.Vec3(2.7,1,1);
let aabb = new pc.BoundingBox();
aabb.setMinMax(min, max);
console.log(aabb.getMin().x);
// output: 0.2999999999999998
console.log(aabb.getMin().x === min.x);
// output: false
```

Maksims:
This is related to general issue with JavaScript Numbers precision. Basically what happens is this:
```javascript
0.2 + 0.4 = 0.6000000000000001
```
You can read more on this: https://stackoverflow.com/questions/11695618/dealing-with-float-precision-in-javascript

RacoAnimech:
> This is related to general issue with JavaScript Numbers precision.  

Yes it is, but it's also preventable. Mathematically speaking storing a bounding box as center and halfextents or minimum and maximum should not matter since we can calculate them from each other. But when implemented they become distinct because of these rounding errors.

Right now ```BoundingBox``` stores center and halfextent and calculates minimum and maximum with rounding errors. An example of when these errors matter is when the ```BoundingBox``` is calculated from a vertex buffer. The minimum and maximum values might actually not be the minimum and maximum of the vertexbuffer. I suggest that ```BoundingBox``` should instead store minimum and maximum and calculate center and halfextents from these. This would solve the problem.

slimbuck:
Yep @RacoAnimech. Presumably center & halfExtent was chosen so that _fastIntersectRay() could be fast.

I would be very surprised if the rounding on 64bits of floating point precision isn't enough though. What problems are you actually seeing with the inaccuracy?

RacoAnimech:
I'm writing an exporter and min max values are required. It feels wrong to export incorrect values and  looping over the vertexbuffer is not an optimal solution, especially when the values are already calculated.

slimbuck:
Yes I understand what you mean.

[...]

PR #6691, opened 2024-06-12
[BREAKING] BoundingBox and OrientedBox copy parameters in the constructor

mvaligursky:
related to https://github.com/playcanvas/engine/pull/2088

BoundingBox and OrientedBox no longer reference arguments passed into their constructors, but clone those, to avoid hard to find issues.

Search of the engine and editor code base reveals no internal issues with this, the problematic cases were refactored out  already.

Marked as `BREAKING` as this might affect user code if not used correctly.

### DIFF
--- a/src/core/shape/bounding-box.js
+++ b/src/core/shape/bounding-box.js
@@ -1,4 +1,3 @@
-import { Debug } from '../debug.js';
 import { Vec3 } from '../math/vec3.js';
 
 const tmpVecA = new Vec3();
@@ -18,14 +17,14 @@ class BoundingBox {
      *
      * @type {Vec3}
      */
-    center;
+    center = new Vec3();
 
     /**
      * Half the distance across the box in each axis.
      *
      * @type {Vec3}
      */
-    halfExtents;
+    halfExtents = new Vec3(0.5, 0.5, 0.5);
 
     /**
      * @type {Vec3}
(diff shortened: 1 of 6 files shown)


kungfooman:
Very nice, you could add `@readonly` too so that developers receive an Intellisense warning when they are on the brink of introducing a hard to find issue.

mvaligursky:
> Very nice, you could add `@readonly` too so that developers receive an Intellisense warning when they are on the brink of introducing a hard to find issue.

Yeah you're right, people need to set / copy to them instead of assigning.

slimbuck:
Can we consider fixing the implementation of `BoundingBox` to store min/max instead of center/extents?

I can't find the issue/pr on the engine repo, but this has come up before and generally speaking storing min/max is preferable, since it handles edge cases better and is more accurate.

For something so core to engine it sure would be nice to fix.

mvaligursky:
> Can we consider fixing the implementation of `BoundingBox` to store min/max instead of center/extents?

This one: https://github.com/playcanvas/engine/issues/1903

slimbuck:
> > Can we consider fixing the implementation of `BoundingBox` to store min/max instead of center/extents?
> 
> This one: #1903

Yes thanks! There was also this PR which wasn't takes due to possible breaking API https://github.com/playcanvas/engine/pull/1921

mvaligursky:
That's for a separate PR to figure out.

slimbuck (approved):

Issue #4290, opened 2022-05-31
Texture API issues


slimbuck:
Some observations of the current texture API:
- `setSource` only works with 'browser interface' objects (not arrays of data) and the function correctly tracks which mipmap levels are dirty using `_levelsUpdated`.
- `getSource` will return browser interface objects and also arrays of data which it strictly probably shouldn't (since it is matched with `setSource`).
- `lock` and `unlock` work with arrays data and the entire texture is always re-uploaded if any mipmap level is locked, whether for read or write.

Ideally we'd have an API that:
- has a single way of updating both types of texture data
- tracks which face/mipmap levels have been updated and only upload those to GPU
- makes it simple to get, set and dirtify individual faces/levels

Also, the current API assumes textures have CPU-side data which is uploaded to GPU. However we often have GPU-side texture data which we want to copy back to CPU.

@mvaligursky are there any additional restrictions or requirements imposed by webgpu on texture handling?

mvaligursky:
Not at the moment, but here are some best practices for WebGPU: https://github.com/toji/webgpu-best-practices/blob/main/img-textures.md

yaustar:
From forum: https://forum.playcanvas.com/t/pre-uploading-of-pc-texture/3212/

Users are requesting the ability to manually upload textures to the GPU so they can control the potential framerate stutter when loading in sections of an open world. Eg, they could upload textures over several frames.

ArztSamuel:
>  Eg, they could upload textures over several frames.

To add to that: currently the textures are loaded "on demand" (i.e. when they are first drawn). We would like to be able to upload large textures during our initial loading screen instead, in order to prevent frame stutters later in the game. We would be fine with all textures being uploaded in 'one frame' in that case. It's just the overall timing of that frame we are concerned with.

LeXXik:
Regarding the timing - we work around it by drawing a bunch of cubes with all materials that will be used in game during the first frame (which is hidden by a loading screen, so the player doesn't see them), then remove them and show the game.

heretique:
Hi everyone, I recently started looking into adding support to update particular textures/slices from a Texture2DArray. In it's current iteration texture 2d array support is not flexible enough and not that useful. While investigating this I was reading all the discussions from #6004, this ticket and the references mentioned here.
I was also considering additional use cases:

[...]

PR #6693, opened 2024-06-13
Texture array and cubemaps single slice updates

heretique:
Hello PC team,
this PR adds single slice updates for texture 2d arrays and cubemaps. 

Please note this is a draft, work in progress PR but I wanted to get it out so that I can get feedback early on.
Please also note that I have [another branch where I was a little more ambitious](https://github.com/playcanvas/engine/compare/main...heretique:playcanvas-engine:texture-updates-improvements) with larger internal changes for the textures but that proved to generate more chained changes because there are a lot of internal places where the engine uses the internals ("private" fields) of the textures bypassing the provided API. So after discussions on #4290 I went with the least possible changes, that I'll list bellow:

 - added a `_dimension` private field to use for identifying the "type" of the texture. `_type` is already used for something else
 - added a single `_slices` field to indicate the number of faces for a cubemap, slices/layers for a texture 2d array, depth for a 3d texture
 - removed `arrayLength`, `_volume`, `_cubemap` internal fields but based the API getters and setters on the newly introduced `_dimension` and `_slices` fields
 - added an `immediate` option at texture creation, `setSource` and `unlock` to force an immediate upload of texture data to GPU
 - this may be debatable but I removed the possibility of setting any mipmap level from `setSource`. Given the name and the fact that it is supposed to set the texture source from a "browser interface" it doesn't make much sense, plus in all the codebase only level 0 it is used. This should force users to use `lock` API which is more low level.
 - `lock` now allows you to update mip level 0 for single slices of a texture 2d array or cubemap face. Why this limitation? because of the way the `_levelsUpdated` were used before. In the first place I don't believe updating a single face using `lock` worked before because `_levelsUpdated` was not being set at all and there is a check in the `webgl` implementation that specifically checks for `_levelsUpdated[0][face]`. I decided to do a similar thing for texture 2 arrays as well and allow setting mip 0 and fixed it as well for cubemap. I believe it makes more sense as if any mip level is missing they still need to be generated using `gl.generateMipmaps` on WebGL.

This last point brings me to the big issue that I encountered, `gl.generateMipmap` does a full mipmap chain generation on all types of textures, this may make sense for 2D and 3D textures but not so much for cubemaps and even less for texure 2d arrays. It does make sense at texture creation time, but this last type can contain hundreds of textures and re-generating the full mipmap chain after a single slice upload is wasteful and can block the main loop.
The solution, implement a mipmap renderer for WebGL like the one we have for WebGPU. I already started with that, although not pushed yet because there's a big issue with that. For WebGL most of the texture data is lazily uploaded during main render loop right inside the `draw` method. By the time we reach `setTexture` a part of the GL state is already set, shaders bound, etc. If I do my own `WebglMipmapRenderere.generate` call it will probably mess up the state pretty bad as I need to bind framebuffers, shaders, etc. for rendering the mipmaps. We will probably need to upload dirty textures before entering the `draw` method and here's where I'm kindly asking for guidance/suggestions on how to do this without breaking anything 😁. I'll update the PR with the mipmaprenderer in the following days when I have some time to polish it. 

### DIFF
--- a/src/platform/graphics/texture.js
+++ b/src/platform/graphics/texture.js
@@ -102,7 +102,7 @@ class Texture {
      * @param {string} [options.name] - The name of the texture. Defaults to null.
      * @param {number} [options.width] - The width of the texture in pixels. Defaults to 4.
      * @param {number} [options.height] - The height of the texture in pixels. Defaults to 4.
-     * @param {number} [options.slices] - The number of depth slices in a 3D texture, the number of textures
+     * @param {number} [options.layers] - The number of depth layers in a 3D texture, the number of textures
      * in a texture array or the number of faces for a cubemap.
      * @param {string} [options.dimension] - The texture dimension type. Can be:
      * - {@link TEXTUREDIMENSION_2D}
@@ -232,7 +232,7 @@ class Texture {
         Debug.assert(this.device, "Texture constructor requires a graphicsDevice to be valid");
         Debug.assert(!options.width || Number.isInteger(options.width), "Texture width must be an integer number, got", options);
         Debug.assert(!options.height || Number.isInteger(options.height), "Texture height must be an integer number, got", options);
-        Debug.assert(!options.slices || Number.isInteger(options.slices), "Texture slices must be an integer number, got", options);
+        Debug.assert(!options.layers || Number.isInteger(options.layers), "Texture layers must be an integer number, got", options);
 
         this.name = options.name ?? '';
 
@@ -244,9 +244,9 @@ class Texture {
         this._width = Math.floor(options.width ?? 4);
         this._height = Math.floor(options.height ?? 4);
 
-        this._slices = Math.floor(options.slices ?? (this._dimension === TEXTUREDIMENSION_CUBE ? 6 : 1));
+        this._layers = Math.floor(options.layers ?? (this._dimension === TEXTUREDIMENSION_CUBE ? 6 : 1));
 
-        Debug.assert((this._dimension === TEXTUREDIMENSION_CUBE ? this._slices === 6 : true), "Texture cube map must have 6 slices");
+        Debug.assert((this._dimension === TEXTUREDIMENSION_CUBE ? this._layers === 6 : true), "Texture cube map must have 6 layers");
 
         this._format = options.format ?? PIXELFORMAT_RGBA8;
         this._compressed = isCompressedPixelFormat(this._format);
@@ -300,7 +300,7 @@ class Texture {
         if (this._levels) {
             this.upload(options.immediate ?? false);
         } else {
-            this._levels = (this.cubemap || this.array) ? [Array(this._slices).fill(null)] : [null];
+            this._levels = (this.cubemap || this.array) ? [Array(this._layers).fill(null)] : [null];
         }
 
         // track the texture
@@ -348,10 +348,10 @@ class Texture {
      *
      * @param {number} width - The new width of the texture.
      * @param {number} height - The new height of the texture.
-     * @param {number} [slices] - The new number of slices for the texture. Defaults to 1.
+     * @param {number} [layers] - The new number of layers for the texture. Defaults to 1.
      * @ignore
      */
-    resize(width, height, slices = 1) {
+    resize(width, height, layers = 1) {
 
         // destroy texture impl
         const device = this.device;
@@ -360,7 +360,7 @@ class Texture {
 
         this._width = Math.floor(width);
         this._height = Math.floor(height);
-        this._slices = Math.floor(slices);
+        this._layers = Math.floor(layers);
         this._updateNumLevel();
 
         // re-create the implementation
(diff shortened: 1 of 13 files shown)


src/platform/graphics/texture.js
```diff
@@ -91,7 +102,15 @@ class Texture {
      * @param {string} [options.name] - The name of the texture. Defaults to null.
      * @param {number} [options.width] - The width of the texture in pixels. Defaults to 4.
      * @param {number} [options.height] - The height of the texture in pixels. Defaults to 4.
-     * @param {number} [options.depth] - The number of depth slices in a 3D texture.
+     * @param {number} [options.slices] - The number of depth slices in a 3D texture, the number of textures
```
mvaligursky, line 105: we need to design the API to support cubemap arrays for when we get to adding those

mvaligursky: I'd probably look at WebGPU API for some inspirations and even naming

https://www.w3.org/TR/webgpu/#dictdef-gputextureviewdescriptor

mvaligursky: and https://www.w3.org/TR/webgpu/#gputexturedescriptor

src/platform/graphics/texture.js
```diff
@@ -15,11 +15,25 @@ import {
...
+    isIntegerPixelFormat, FILTER_NEAREST, TEXTURELOCK_NONE, TEXTURELOCK_READ,
+    TEXTUREDIMENSION_2D,
+    TEXTUREDIMENSION_3D,
+    TEXTUREDIMENSION_2D_ARRAY,
+    TEXTUREDIMENSION_CUBE
 } from './constants.js';
 
 let id = 0;
 
+const PROPERTY_MIN_FILTER = 1;
```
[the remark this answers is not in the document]
heretique: I think I can cherry pick that commit in a separate PR

src/platform/graphics/texture.js
heretique: I'll have a look at the WebGPU API you mentioned  but I'd leave the cubemap arrays support for a separate PR. I'm afraid of too many cascading changes if I change too much of the internals. See my other branch I mentioned in the description

mvaligursky: Yep not suggesting to add a support, just to make sure the changes to the API will support it (`slices` specifically)

heretique:
Yeah, with regards to mipmaps I played with texture destruction on the alternate branch.
There I introduced the concept of texture operation with a new field `_operation` and these values where available:
const TEXTURE_OPERATION_NONE = 0;
const TEXTURE_OPERATION_UPLOAD = 1;
const TEXTURE_OPERATION_UPLOAD_PARTIAL = 2;
const TEXTURE_OPERATION_CLEAR_MIPS = 4;

This allowed me to

src/platform/graphics/texture.js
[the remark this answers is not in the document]
heretique: Done, I've create #6697 but also #6698 for optional immediate data upload to GPU.

src/platform/graphics/texture.js
heretique: I renamed the "slices" to "layers" to make it more consistent with WebGPU naming but also WebGL (framebufferTextureLayer)

heretique:
A small update:
I moved the texture data upload outside of `WebglGraphicsDevice.setTexture` to `WebglGraphicsDevice._uploadDirtyTextures` and it is now called inside `WebglGraphicsDevice.startRenderPass` taking a hint from how WebGPU does this. This will allow me to properly use the custom mipmap renderer that is going to be used when updating a single texture layer without the fear of affecting current WebGL state or the need to implement a state tracker for rolling back state.
The move does seem to affect only two examples (I've tested almost all of them): "Light Baked" and "Lights Baked AO". The colored lights in both examples are missing. I'm currently investigating the issue.

heretique:
@mvaligursky I'm having a hard time trying to figure out why lightmapping doesn't properly work after I moved the texture GPU uploads to `WebglGraphicsDevice._uploadDirtyTextures` inside `WebglGraphicsDevice.startRenderPass`, the lightmapper stops working properly. When texture upload is in `WebglGraphicsDevice.setTexture` it works fine. So must be some order of initialization thing but I spent a couple of day on it and I couldn't figure it out.
Do you have any suggestion where I should look? I ran the lightmapper step by step but probably I'm overlooking something.
Here's how the example looks with my changes:

image

and how it should look properly:

image

Doing some Spector.js captures I'm also noticing differences between textures bound to `texture_lightMap` and `texture_dirLightMap` uniforms:

image

Any help is appreciated, thank you!

mvaligursky:
No clue what the problem could be. From the screenshot, it seems baking of directional light works, but omni does not work. But not sure what the problem could be.
Maybe try reversing the array of the lights to see if omni works and directional does not.

heretique:
I finally managed to figure out the light baking issue. It was fixed with commit [43b99d0](https://github.com/playcanvas/engine/pull/6693/commits/43b99d05c8d2dc6032ce813add098b149edb7371).
Now the thing that remains is to implement a mipmap renderer like the one WebGPU uses, when updating a single or only a few of the textures (layers) of the 2D texture array.

mvaligursky:
Cool! I think perhaps the part where textures are uploaded before rendering, instead of during rendering, is very standalong, similar to what is done on WebGPU platform. It should perhaps be extracted to a separate PR?

heretique:
> Cool! I think perhaps the part where textures are uploaded before rendering, instead of during rendering, is very standalong, similar to what is done on WebGPU platform. It should perhaps be extracted to a separate PR?

Sure, I could try and make it a separate PR. There's a similar standalone one that I extracted from this #6698

PR #6702, opened 2024-06-14
[BREAKING] Changes to the HDR rendering

mvaligursky:
This PR is marked as breaking, as it will affect projects that use the removed (and replaced) functionality. Note that the current implementation (removed functionality) of HDR was limited / not much documented and the it's not very likely it was being used in many cases.

## New API
```
// class storing few rendering parameters, allowing us to set those on the `Scene`,
// but also override them on a `CameraComponent`
class RenderingParams
    .gammaCorrection
    .toneMapping

Scene.rendering // exposes an instance of RenderingParams
CameraComponent.rendering // exposes an override on the camera

// in the future, additional properties will be added - for example fog settings and similar.
```

The main goals here:
- allow easy way to override camera parameters, that are also propagated to the shader generation, and each camera uses matching shaders using recently improved shader caching on MeshInstances
- this allows us to switch some cameras to HDR (linear, no tone mapping) while leaving others in Gamma space.

In the past this was done using multiple deprecated (see below) constants with not-clear meaning and implementation, which was only partially functional.

## Removed API
```
// constants:
    SHADER_FORWARDHDR
    SHADERPASS_FORWARD_HDR
    GAMMA_SRGBHDR
    GAMMA_SRGBFAST

// functions:
Layer.shaderPass 
```

## Deprecated API
```
Scene.toneMapping // now forwards to Scene.rendering.toneMapping
Scene.gammaCorrection // now forwards to Scene.rendering.gammaCorrection
```


MAG-AdrianMeredith:
This all sounds great (though it will break a ton of things for me).  Our use case for example is a user may have turned off HDR bloom rendering but then want to create a reflection probe capture rendered in HDR.  Right now theres a bunch of trickery with the shaderPass flag but my understanding now is with this change its a camera param so no trickery required?

mvaligursky:
Yep exactly, now it will just be an option on the camera. Easy to use, no magic behind the scene.

slimbuck (approved):

mvaligursky:
@MAG-AdrianMeredith - see the linked follow up PR, please test against that branch if possible.

MAG-AdrianMeredith:
yeah I saw thanks

PR #6708, opened 2024-06-15
Improve typings and docs for XrPlane

willeastcott:
Also enhances the API docs.

### DIFF
--- a/src/framework/xr/xr-plane.js
+++ b/src/framework/xr/xr-plane.js
@@ -162,7 +162,7 @@ class XrPlane extends EventHandler {
      * Array of DOMPointReadOnly objects. DOMPointReadOnly is an object with `x y z` properties
      * that defines a local point of a plane's polygon.
      *
-     * @type {object[]}
+     * @type {DOMPointReadOnly[]}
      * @example
      * // prepare reusable objects
      * const vecA = new pc.Vec3();
--- a/src/framework/xr/xr-plane.js
+++ b/src/framework/xr/xr-plane.js
@@ -152,9 +152,19 @@ class XrPlane extends EventHandler {
     }
 
     /**
-     * Plane's specific orientation (horizontal or vertical) or null if orientation is anything else.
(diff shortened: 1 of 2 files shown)


src/framework/xr/xr-plane.js
```diff
@@ -159,10 +161,12 @@ class XrPlane extends EventHandler {
     }
 
     /**
-     * Array of DOMPointReadOnly objects. DOMPointReadOnly is an object with `x y z` properties
-     * that defines a local point of a plane's polygon.
+     * Gets the array of points that define the polygon of the plane in its local coordinate space.
```
LeXXik, line 164: "an array". I believe the very first occurrence of a thing uses "a", and "the" afterwards.

LeXXik (approved):

src/framework/xr/xr-plane.js
willeastcott: The points are not created in response to calling the getter. The array lives on the WebXR XRPlane and the getter simply returns them.

LeXXik (approved):

src/framework/xr/xr-plane.js
```diff
@@ -150,24 +152,36 @@ class XrPlane extends EventHandler {
     }
 
     /**
-     * Plane's specific orientation (horizontal or vertical) or null if orientation is anything else.
+     * Gets the plane's specific orientation. This can be "horizontal" for planes that are parallel
+     * to the ground, "vertical" for planes that are perpendicular to the ground, or `null` if the
+     * orientation is different or unknown.
      *
      * @type {string|null}
+     * @example
```
mvaligursky, line 160: do we want to list the possible strings in the `@type` tag instead of `string`?

mvaligursky (approved):

src/framework/xr/xr-plane.js
willeastcott: Done.

Issue #6699, opened 2024-06-14
Compressed gsplat loads very slowly when the asset is created by scripting


ludovit-as:
With the new version, loading gaussian splats with a script, as in the linked project has become much slower and sometimes even crashes the browser (more likely to crash at higher resolutions, and also when more of the splat is visible in the first frame).

If the splat is barely visible on the first frame, the problem goes away. This can be demonstrated by enabling the `offsetFirst` attribute on the linked project.

Project to reproduce: https://playcanvas.com/project/1228429/
Related forum post: https://forum.playcanvas.com/t/gsplat-loads-very-slowly-and-sometimes-crashes-after-udpate/36049

slimbuck:
I think the first frame is rendering all particles at the same place large on screen and this takes too long, causing the browser to kill the tab. Will continue investigating.

slimbuck:
A fix for this was released yesterday in [Engine v1.71.6](https://github.com/playcanvas/engine/releases/tag/v1.71.6).

You can test it in the editor if you enable "Use Release Candidate" in Launch settings.

ludovit-as:
It works as expected now. Thank you very much for the quick and effective fix!

PR #6710, opened 2024-06-17
Gsplat fixes

slimbuck:
Fixes: #6699 

This PR fixes two issues:
* only render the gsplat instance after we've received the sorted order data (otherwise there is a visible snap on the first frame).
* refrain from rendering out-of-bounds splats when they're compressed (similar to #6590 for uncompressed)


willeastcott (approved):

mvaligursky:
Please merge this to the `main` branch too

PR #6712, opened 2024-06-17
Clear references of temporary array when no longer needed

LeXXik:
Follow up on #6647

I think it would be better to clear the temp array of old references immediately after consuming, so we don't keep them forever.

### DIFF
--- a/src/framework/entity.js
+++ b/src/framework/entity.js
@@ -521,6 +521,8 @@ class Entity extends GraphNode {
                 }
             }
         }
+
+        components.length = 0;
     }
 
     /** @private */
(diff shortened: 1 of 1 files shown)


src/framework/entity.js
```diff
@@ -605,8 +609,6 @@ class Entity extends GraphNode {
     }
 
     _getSortedComponents() {
-        _sortedArray.length = 0;
```
mvaligursky, line 608: I'd add `Debug.assert(_sortedArray.length === 0);` to validate, as it's a bit of an easy to break concept.

mvaligursky (approved):

mvaligursky (approved):

mvaligursky:
It causes this issue: https://github.com/playcanvas/engine/issues/6717

Issue #6709, opened 2024-06-17
Store light / material / ambient / fog colors always in linear space


mvaligursky:
Based on the `Scene.gammaCorrection`, we store those colors either in gamma (this is the default), or linear space (when gamma is disabled). This causes two problems:
- when gammaCorrection changes, the scene looks visually different, as the shader math is the same, but its input colors uniforms are different. So when we switch to HDR, the scene is too bright.
- we cannot correctly handle multiple cameras, one using sRGB and one linear, as colors are adjusted based on a global gamma setting.

Suggestion:
- always store these colors in linear only (as is the default behaviour currently)
- Scene.gammaCorrection should only affect the way pixels are written to render target, not encoding of inputs

I would expect no impact to majority or maybe even all users, as our HDR rendering path was hardly functional till now.

some of the relevant code:

https://github.com/playcanvas/engine/blob/5a6ff8668e0c0f3deb56ce89d4897b523b020abd/src/scene/materials/standard-material.js#L1101-L1109

https://github.com/playcanvas/engine/blob/5a6ff8668e0c0f3deb56ce89d4897b523b020abd/src/scene/renderer/forward-renderer.js#L150-L154

https://github.com/playcanvas/engine/blob/5a6ff8668e0c0f3deb56ce89d4897b523b020abd/src/scene/renderer/forward-renderer.js#L231

https://github.com/playcanvas/engine/blob/5a6ff8668e0c0f3deb56ce89d4897b523b020abd/src/scene/renderer/forward-renderer.js#L780-L784

MAG-AdrianMeredith:
This sounds like it would explain a lot, as someone who mostly runs playcanvas in hdr mode we've struggled a while with consistency over the two modes (e.g. gsplats are very different right now).  I'll be sure to give the PR a test if and when there is one

mvaligursky:
> This sounds like it would explain a lot, as someone who mostly runs playcanvas in hdr mode we've struggled a while with consistency over the two modes (e.g. gsplats are very different right now). I'll be sure to give the PR a test if and when there is one

yep, there will be a PR today / tomorrow for this.

Issue #3715, opened 2021-11-18
Use gamma textures format (WebGl2)


mvaligursky:
Currently, when the engine samples a texture stored in gamma space, it uses gammaCorrectInput function to bring the values to linear space for lighting, costing us a pow instruction per sample, and also bilinear interpolation when sampling takes place in sRGB space, which is slightly incorrect.

On Webgl2 devices, and also on Webgl1 devices with EXT_sRGB (https://developer.mozilla.org/en-US/docs/Web/API/EXT_sRGB), we could use hardware support for sRGB textures.

See here: (http://www.realtimerendering.com/blog/webgl-2-new-features/)
`The sRGB texture will be automatically converted to linear space when being fetched in the shader. ` 

The disadvantage here is that the texture needs to be marked as sRGB when we create a WebGl texture, meaning likely that we would need to expose a texture import setting in the Editor to allow textures to be marked as sRGB (diffuse and similar texture need sRGB, normal map, roughness and others don't as they are in linear space).

slimbuck:
Our gamma space textures also currently have their mipmaps generated in gamma space, which is wrong. Marking texture sRGB will result in linear-space generated mipmaps.

We also need to address compressed textures in the editor. Textures are assumed gamma space (and converted to linear for mipmaps) - so linear-space textures currently have the inverse mipmap problem. (Though I believe normal maps are correctly handled).

mvaligursky:
Note that ideally we'd use sRGB framebuffers as well, which automatically encode linear -> sRGB on write.
https://hackmd.io/@jgilbert/sRGB-WebGL

mvaligursky:
Closing this as this is now supported by the engine v2.1.0. Support to the Editor is coming soon too.

PR #6714, opened 2024-06-18
[BREAKING] Colors of lights, fog and materials are always supplied in sRGB space, for both LDR and HDR

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/6709

Marking this as BREAKING, as due to fixes for HDR rendering, this might look different. Also see the API change listed below.

Follow up to https://github.com/playcanvas/engine/pull/6702

Unified the final output between LDR (default) and HDR rendering. Before this PR, the HDR rendering would be typically brighter. The main reason is that the HDR mode expected colors to be set in linear space, unlike in SDR where those were expected in sRGB. Now all inputs are unified, always in sRGB, so that the scene set up / GLBs looks the same in LDR and HDR.

- added constant TONEMAP_NONE. Before, internally, we'd just use hardcoded -1, but this needs to be available as public API for HDR
- light color as uniform is always in linear space. Before, when HDR was used, the user was expected to assign linear colors. Now we always expect sRGB colors and convert them to linear. This applies to ambient light, but also directional, omni and spot, including clustered lights.
- material color properties were expected be in the gamma space or linear space depending on scene.gammaCorrection. Now they're always expected to be in gamma space, and we always convert them to linear uniforms
- similarly to materials, the fog color is now always expected in sRGB, including in HDR mode, and we convert it to linear uniforms
- UI now works in the linear space, which means the source (emissive) texture is converted from gamma to linear, and a linear to gamma is applied at the end of the shader. Note that this will be in most cases optimized out when this is done: https://github.com/playcanvas/engine/issues/3715

## New API
- `Color.linear(source)` - converts sRGB to linear
- `Color.gamma(source)` - converts linear to sRGB
- `Color.mulScalar(scalar)` - multiplies with scalar, applied to RGB only
- `TONEMAP_NONE` - new constants, used in HDR rendering

## Changed API [BREAKING]
`StandardMaterial.useGammaTonemap` has been deprecated, and `StandardMaterial.useTonemap` has been added instead. We no longer expose the gamma control on the standard material.


src/core/math/math.js
```diff
@@ -277,6 +277,16 @@ const math = {
...
         return inclusive ? num >= min && num <= max : num > min && num < max;
+    },
+
+    /**
+     * Convert a value in sRGB color space to linear color space.
+     *
+     * @param {number} value - the sRGB value to convert to linear.
+     * @returns {number} The value in linear color space.
+     */
+    gammaToLinear(value) {
```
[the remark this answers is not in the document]
mvaligursky: how about
`Color.linear(src)` - converts gamma to linear
`Color.gamma(src)` - converts from linear to gamma
`src` defaults to `this` as is used in Vec3

or just `Color.linear` and `Color.gamma`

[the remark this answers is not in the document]
willeastcott: Yep, sounds good.

mvaligursky:
There is more work to do on the UI side, as that needs gamma changes now to fit as well.

MAG-AdrianMeredith:
How do you access RenderingParams?  its not initialised on the camera by default and doesn't seem to be exported

mvaligursky:
> How do you access RenderingParams? its not initialised on the camera by default and doesn't seem to be exported

Nice, adding the export shortly, sorry.

src/core/math/math.js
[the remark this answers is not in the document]
mvaligursky: done

src/scene/constants.js
```diff
@@ -626,6 +626,14 @@ export const TONEMAP_ACES2 = 4;
  */
 export const TONEMAP_NEUTRAL = 5;
 
+/**
+ * No tonemapping nor the exposure is applied. Used for HDR rendering.
```
willeastcott, line 630:
```suggestion
 * No tonemapping or exposure is applied. Used for HDR rendering.
```

src/scene/constants.js
```diff
@@ -626,6 +626,14 @@ export const TONEMAP_ACES2 = 4;
...
+ *
+ * @type {number}
+ * @category Graphics
+ */
+export const TONEMAP_NONE = 6;
```
willeastcott, line 635: How dangerous would it be to renumber these constants so that `TONEMAP_NONE` was 0? I mean, no one is setting via numbers in code. But presumably numbers are stored in the database...

src/scene/materials/standard-material.js
```diff
@@ -1096,17 +1098,12 @@ function _defineColor(name, defaultValue) {
...
-            uniform[0] = Math.pow(color.r, 2.2);
-            uniform[1] = Math.pow(color.g, 2.2);
-            uniform[2] = Math.pow(color.b, 2.2);
-        } else {
-            uniform[0] = color.r;
-            uniform[1] = color.g;
-            uniform[2] = color.b;
-        }
+        // uniforms are always in linear space
+        const uniformColor = _tempColor.linear(color);
```
willeastcott, line 1103: Why do you need a temp var here? Just set `uniform` from `_tempColor` components.

willeastcott (approved):
Very cool.

src/scene/constants.js
mvaligursky: yeah the Editor projects give us numbers .. and so I would prefer to not break that.

src/scene/materials/standard-material.js
mvaligursky: yep done, thanks

PR #6715, opened 2024-06-18
GSplat speedup

slimbuck:
In this PR we:
* speed up the load-time reordering of uncompressed splat data by roughly 1/3rd for large scenes
    * time to reorder the 6 million splat bicycle scene goes from 3.2s to 2.3s on Chrome Macbook M1
* simplify code slightly by moving some of the load-time logic from `GSplatData` to `PLYParser`
* small render speedup by removing the `id` varying in uncompressed shader when it's not needed

### DIFF
--- a/src/scene/gsplat/gsplat-data.js
+++ b/src/scene/gsplat/gsplat-data.js
@@ -182,28 +182,10 @@ class GSplatData {
 
     // /**
     //  * @param {import('./ply-reader').PlyElement[]} elements - The elements.
-    //  * @param {boolean} [performZScale] - Whether to perform z scaling.
-    //  * @param {object} [options] - The options.
-    //  * @param {boolean} [options.performZScale] - Whether to perform z scaling.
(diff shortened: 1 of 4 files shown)


src/scene/gsplat/gsplat-data.js
```diff
@@ -182,28 +182,10 @@ class GSplatData {
 
     // /**
     //  * @param {import('./ply-reader').PlyElement[]} elements - The elements.
-    //  * @param {boolean} [performZScale] - Whether to perform z scaling.
-    //  * @param {object} [options] - The options.
-    //  * @param {boolean} [options.performZScale] - Whether to perform z scaling.
-    //  * @param {boolean} [options.reorder] - Whether to reorder the data.
     //  */
-    constructor(elements, options = {}) {
+    constructor(elements) {
```
willeastcott, line 186: Can we uncomment the JSDoc above?

slimbuck: These are being naughty - they reference types in a higher level module

PR #6716, opened 2024-06-18
GSplat speedup v1

slimbuck:
#6715 for engine v1.

### DIFF
--- a/src/scene/gsplat/shader-generator-gsplat.js
+++ b/src/scene/gsplat/shader-generator-gsplat.js
@@ -16,7 +16,10 @@ const splatCoreVS = /* glsl */ `
 
     varying vec2 texCoord;
     varying vec4 color;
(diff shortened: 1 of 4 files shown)


src/scene/gsplat/shader-generator-gsplat.js
```diff
@@ -16,7 +16,10 @@ const splatCoreVS = /* glsl */ `
 
     varying vec2 texCoord;
     varying vec4 color;
-    varying float id;
+
+    #ifndef DITHER_NONE
```
mvaligursky, line 20: I have not checked the code, but don't we use `id` for the dithering?

slimbuck: yep! :)

mvaligursky (approved):

PR #6718, opened 2024-06-19
Add array pool for sorted components

### LINKED ISSUE #6717: An assert in the engine example related to sorted components
in the example folder run `npm run develop` and load the ReflectionBox example. This is what you get:

Screenshot 2024-06-19 at 09 39 28

LeXXik:
Fixes #6717 

Adds an array pool for sorted components.

### DIFF
--- a/src/core/sort.js
+++ b/src/core/sort.js
@@ -1,4 +1,8 @@
 /**
+ * @typedef {import('../framework/components/component').Component} Component
+ */
+
+/**
  * @param {{priority: number}} a - First object with priority property.
  * @param {{priority: number}} b - Second object with priority property.
  * @returns {number} A number indicating the relative position.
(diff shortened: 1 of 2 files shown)


src/core/sort.js
```diff
@@ -1,3 +1,7 @@
+/**
+ * @typedef {import('../framework/components/component').Component} Component
```
mvaligursky, line 2: we should not be importing anything from framework (which is the top level layer in the engine) into core (which is the bottom level).

I'd just move those two functions to where we need them, as at this point they're no longer general.

src/framework/entity.js
```diff
@@ -13,10 +13,22 @@ import { getApplication } from './globals.js';
...
 /**
- * @type {Array<import('./components/component.js').Component>}
+ * @type {Array<Array<import('./components/component.js').Component>>}
  * @ignore
  */
-const _sortedArray = [];
+const tmpPool = [];
+
+const getTempArray = () => {
+    return tmpPool.pop() || [];
```
mvaligursky, line 22: ?? instead of || is more correct

src/framework/entity.js
```diff
@@ -523,6 +535,7 @@ class Entity extends GraphNode {
         }
 
         components.length = 0;
+        releaseTempArray(components);
```
mvaligursky, line 538: move `components.length = 0;` inside the release function

mvaligursky (approved):
looks good, thanks

PR #6720, opened 2024-06-19
Update setFromMat4 and add tests

slimbuck:
Fixes: #3759

This PR:
* Updates the matrix to quaternion implementation to be more accurate (quaternions were drifting out of normalisation before).
* Add `Quat.mulScalar`
* Also add a suit of tests

### DIFF
--- a/src/core/math/quat.js
+++ b/src/core/math/quat.js
@@ -289,6 +289,21 @@ class Quat {
     }
 
     /**
+     * Multiplies each element of a quaternion by a number.
+     *
+     * @param {number} scalar - The number to multiply by.
+     * @param {Quat} [src] - The quaternion to scale. If not set, the operation is done in place.
+     * @returns {Quat} Self for chaining.
+     */
+    mulScalar(scalar, src = this) {
+        this.x = src.x * scalar;
+        this.y = src.y * scalar;
+        this.z = src.z * scalar;
+        this.w = src.w * scalar;
+        return this;
+    }
+
+    /**
      * Returns the result of multiplying the specified quaternions together.
      *
      * @param {Quat} lhs - The quaternion used as the first multiplicand of the operation.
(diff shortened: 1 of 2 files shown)


src/core/math/quat.js
```diff
@@ -461,103 +476,59 @@ class Quat {
...
+        } else {
+            if (m00 < -m11) {
+                this.set(m20 + m02, m12 + m21, 1 - m00 - m11 + m22, m01 - m10);
+            } else {
+                this.set(m12 - m21, m20 - m02, m01 - m10, 1 + m00 + m11 + m22);
             }
         }
 
-        return this;
+        return this.mulScalar(1.0 / this.length());
```
willeastcott, line 531: Maybe put a comment about why your need to renormalize?

willeastcott:
Any performance difference?

src/core/math/quat.js
slimbuck: done

slimbuck:
> Any performance difference?

Performance difference should be negligible.

src/core/math/quat.js
```diff
@@ -461,103 +476,63 @@ class Quat {
...
+                this.set(m12 - m21, m20 - m02, m01 - m10, 1 + m00 + m11 + m22);
             }
         }
 
-        return this;
+        // Instead of scaling by 0.5 / Math.sqrt(t) (to match the original implementation),
+        // instead we normalize the result. It costs 3 more adds and muls, but we get
+        // a stable result and in some cases normalization is required anyway, see
+        // https://github.com/blender/blender/blob/v4.1.1/source/blender/blenlib/intern/math_rotation.c#L368
+        return this.mulScalar(1.0 / this.length());
```
kungfooman, line 535: Isn't this just what `Quat#normalize` is, e.g. `return this.normalize()`? Then we also don't need to extend the API by adding a new method.

slimbuck: Yep, similar, except we don't need to test for length 0. I wasn't sure the effort here was necessary, but we decided adding `mulScalar` to match the other types is probably a win.

PR #6735, opened 2024-06-20
ESM example

kpal81xd:
- Adds `fileImport` util method for importing files outside blob context
- Adds `Rotator` script


mvaligursky (approved):
Cool. I assume this Rotator script when dropped to the Editor will work without any changes needed?

examples/assets/scripts/misc/rotator.mjs
```diff
@@ -0,0 +1,15 @@
+import { Script } from 'playcanvas';
+
+class Rotator extends Script {
+    /**
+     * @attribute
```
kungfooman, line 5: That will be a lot of linting fun when people start using `@assetType` instead of `@type {Asset<SomeType>}` too :sweat_smile:

kpal81xd: Well I think we decided on `@resource` over `@assetType` in the end

kpal81xd:
> Cool. I assume this Rotator script when dropped to the Editor will work without any changes needed?

Yes it should do.

examples/assets/scripts/misc/rotator.mjs
kungfooman: Yea I know, we gotta whitelist those new tags now?

marklundin: We have Monaco support for this, we need to see how this works in vscode

marklundin:
the `fileImport` feels a little awkward. Is this just to resolve relative to the utils.mjs?

kpal81xd:
> the `fileImport` feels a little awkward. Is this just to resolve relative to the utils.mjs?

Yea its to do with the path resolution for import inside blobs being different

PR #6736, opened 2024-06-20
Standard material handles textures in sRGB format

mvaligursky:
1. Deprecated old and replaced with new texture formats, to match `RGB8` and `RGBA8` formats, as they're the same but in addition handles sRGB conversion:
```
PIXELFORMAT_SRGB -> PIXELFORMAT_SRGB8
PIXELFORMAT_SRGBA -> PIXELFORMAT_SRGBA8;
```

2. Added support for SRGBA8 on WebGPU

3. When standard material uses a texture in sRGB space, we skip in-shader conversion of sRGB->linear, as sampling it gives us a linear value already.


src/platform/graphics/constants.js
```diff
@@ -970,24 +971,24 @@ export const pixelFormatInfo = new Map([
...
+    [PIXELFORMAT_ETC1,              { name: 'ETC1', blockSize: 8, srgb: true }],
+    [PIXELFORMAT_ETC2_RGB,          { name: 'ETC2_RGB', blockSize: 8, srgb: true }],
+    [PIXELFORMAT_ETC2_RGBA,         { name: 'ETC2_RGBA', blockSize: 16, srgb: true }],
+    [PIXELFORMAT_PVRTC_2BPP_RGB_1,  { name: 'PVRTC_2BPP_RGB_1', blockSize: 8, srgb: true }],
+    [PIXELFORMAT_PVRTC_2BPP_RGBA_1, { name: 'PVRTC_2BPP_RGBA_1', blockSize: 8, srgb: true }],
+    [PIXELFORMAT_PVRTC_4BPP_RGB_1,  { name: 'PVRTC_4BPP_RGB_1', blockSize: 8, srgb: true }],
+    [PIXELFORMAT_PVRTC_4BPP_RGBA_1, { name: 'PVRTC_4BPP_RGBA_1', blockSize: 8, srgb: true }],
+    [PIXELFORMAT_ASTC_4x4,          { name: 'ASTC_4x4', blockSize: 16, srgb: true }],
+    [PIXELFORMAT_ATC_RGB,           { name: 'ATC_RGB', blockSize: 8, srgb: true }],
+    [PIXELFORMAT_ATC_RGBA,          { name: 'ATC_RGBA', blockSize: 16, srgb: true }],
```
willeastcott, line 991: Wait, they're all SRGB all the time? Can't I have a linear normal map, say, using these compressed formats?

willeastcott (approved):

src/platform/graphics/constants.js
mvaligursky: these we convert to linear manually.
Soon I'll add compressed texture formats stored in sRGB as well, separate PR.

PR #6743, opened 2024-06-21
Support for compressed sRGB formats

mvaligursky:
- Both WebGL and WebGPU supports sRGB version of compressed texture formats
- compressed texture formats are used when the asset has `{ srgb = true; }` option, similarly to non-compressed formats
- glb parser requests srgb texture formats for base and emissive color textures, as those are stored in sRGB space
- updated couple of fragment shaders in the examples: as glb supplied diffuse texture is not stored in sRGB format, which is automatically sampled as a linear value, the custom shader now needs to convert the result back to sRGB. The shader can no longer work completely in sRGB space.

partially implements https://github.com/playcanvas/engine/issues/3715


LeXXik:
By the way, is there a decode function for 4 component vectors? I remember we needed a texture with alpha channel, but seems $DECODE variants are only for vec3.

mvaligursky:
We don't have one as transparency is in linear space. But I guess if you store other data in the alpha channel (say a greyscale color), you might need to use a custom decode function.

LeXXik:
Right, custom data. Alright, thanks.

src/framework/parsers/glb-parser.js
```diff
@@ -2070,7 +2094,11 @@ const createImages = (gltf, bufferViews, urlBase, registry, options) => {
                 }
 
                 // create and load the asset
-                const asset = new Asset(name, 'texture', file, null, options);
+                const data = {
+                    srgb: srgb
+                };
```
willeastcott, line 2099:
```suggestion
                const data = { srgb };
```

src/framework/handlers/basis-worker.js
```diff
@@ -171,7 +167,7 @@ function BasisWorker() {
             // pvrtc
             case BASIS_FORMAT.cTFPVRTC1_4_RGB:
             case BASIS_FORMAT.cTFPVRTC1_4_RGBA:
-                return isPOT(width, height) && ((width === height) || webgl2);
+                return true;
```
slimbuck, line 170: shouldn't this still be `return isPOT(width, height);` ?

src/platform/graphics/constants.js
```diff
@@ -946,8 +946,95 @@ export const PIXELFORMAT_R8 = 52;
...
+ * @type {number}
+ * @category Graphics
+ */
+export const PIXELFORMAT_ASTC_4x4_SRGB = 63;
+
+/**
+ * Information about pixel formats.
+ *
+ * srgb: true if sampling the format returns value in sRGB space, otherwise in linear space
+ * linearFormat: the corresponding linear format for sRGB formats (sampled in linear space)
```
slimbuck, line 1033: Erm, it's the other way around. RGBA8 is a linear format, stored and sampled in linear space.

The pixel formats you're mapping to here are the sRGB versions. This should probably be called `srgbFormat` or similar.

Also wondering why the linear formats have been flagged as 'srgb: true' (in a previous PR)? `PIXELFORMAT_DXT1` is definitely not srgb and neither are the others.

src/platform/graphics/constants.js
```diff
@@ -946,8 +946,95 @@ export const PIXELFORMAT_R8 = 52;
...
 
-// map of engine PIXELFORMAT_*** enums to information about the format
-// .srgb: true if sampling the format returns value in sRGB space, otherwise in linear space
+/**
+ * Format equivalent to {@link PIXELFORMAT_DXT1} but sampled in linear color space.
+ *
+ * @type {number}
+ * @category Graphics
+ */
+export const PIXELFORMAT_DXT1_SRGB = 54;
```
slimbuck, line 955: Probably want to add support for SRGB8_ALPHA8  for the very common case of RGBA8 textures?

src/platform/graphics/constants.js
```diff
@@ -1020,9 +1119,21 @@ export const isIntegerPixelFormat = (format) => {
...
 
+/**
+ * Returns the linear format for the given sRGB format. If it does not exist, the input format is
+ * returned. For example for {@link PIXELFORMAT_RGBA8} the return value is {@link PIXELFORMAT_RGBA8}.
+ *
+ * @param {number} format - The texture format.
+ * @returns {number} The linear format for the given sRGB format.
+ * @ignore
+ */
+export const getPixelFormatLinearMatch = (format) => {
```
slimbuck, line 1130: Since this function maps linear formats to srgb, it could probably have a better name, maybe something like `mapFormatLinearToSrgb`?

src/framework/handlers/basis-worker.js
mvaligursky: well spotted, thanks

src/platform/graphics/constants.js
mvaligursky: we have PIXELFORMAT_SRGBA8, is this what you mean?

src/platform/graphics/constants.js
mvaligursky: I updated the comments to better explain the meaning, please review this

src/platform/graphics/constants.js
```diff
@@ -976,20 +1065,32 @@ export const pixelFormatInfo = new Map([
     [PIXELFORMAT_BGRA8,         { name: 'BGRA8', size: 4 }],
 
     // compressed formats
-    [PIXELFORMAT_DXT1,              { name: 'DXT1', blockSize: 8, srgb: true }],
-    [PIXELFORMAT_DXT3,              { name: 'DXT3', blockSize: 16, srgb: true }],
-    [PIXELFORMAT_DXT5,              { name: 'DXT5', blockSize: 16, srgb: true }],
+    [PIXELFORMAT_DXT1,              { name: 'DXT1', blockSize: 8, srgb: true, linearFormat: PIXELFORMAT_DXT1_SRGB }],
```
slimbuck, line 1068: I'm still wondering why DXT1 format is flagged as `srgb: true` and why the mapping is called 'linearFormat'? This is the exact opposite of what it should be.

DXT1 format is 100% linear - it is stored and sampled in linear space (no conversion anywhere) and has nothing to do with srgb.

DXT1_SRGB is data stored in gamma space. The gpu will convert the texture data to linear space for sampling and interpolation and then return the result in linear space for us. This format should be flagged as `srgb: true`.

So this line of code should be something like:
```javascript
[PIXELFORMAT_DXT1, { name: 'DXT1', blockSize: 8, srgbFormat: PIXELFORMAT_DXT1_SRGB }],
```

Also not sure what `srgb` flag is currently used for, but you could probably just use the presence of `srgbFormat` and remove the flag entirely.

src/platform/graphics/constants.js
```diff
@@ -1020,9 +1121,22 @@ export const isIntegerPixelFormat = (format) => {
...
+/**
+ * Returns for the format which can be used to sample the sRGB stored data in linear space, with
+ * automatic sRGB to linear conversion. If it does not exist, the input format is returned. For
+ * example for {@link PIXELFORMAT_RGBA8} the return value is {@link PIXELFORMAT_RGBA8}.
+ *
+ * @param {number} format - The texture format.
+ * @returns {number} The format allowing linear sampling of the texture.
+ * @ignore
+ */
+export const getPixelFormatLinearSampling = (format) => {
```
slimbuck, line 1133: This function takes a linear format and returns its srgb equivalent.

I suggest before this function should be called `mapLinearToGamma` or similar because `getPixelFormatLinearSampling` actually expresses the opposite.

src/framework/parsers/texture/basis.js
```diff
@@ -46,6 +46,7 @@ class BasisParser extends TextureParser {
 
     // our async transcode call provides the neat structure we need to create the texture instance
     open(url, data, device, textureOptions = {}) {
+        const format = textureOptions.srgb ? pixelFormatLinearToGamma(data.format) : data.format;
```
slimbuck, line 49: It probably makes sense to do this in `Texture` constructor rather than every texture parser (looks like you missed `Img`, for example).

src/platform/graphics/constants.js
```diff
@@ -1021,12 +1121,29 @@ export const isIntegerPixelFormat = (format) => {
...
+ * sample is in sRGB space and needs to be decoded to linear space.
+ *
  * @param {number} format - The texture format.
- * @returns {boolean} - Whether sampling the texture with this format returns a linear value.
+ * @returns {boolean} Whether sampling the texture with this format returns a sRGB value.
  * @ignore
  */
-export const isLinearFormat = (format) => {
-    return !pixelFormatInfo.get(format)?.srgb;
+export const isColorSampleSrgb = (format) => {
```
slimbuck, line 1144: I would prefer something like `requiresManualGamma` or `requiresLegacyGamma` and then have docs refer to this feature as a legacy workaround for not having srgb pixel formats.

src/platform/graphics/constants.js
```diff
@@ -946,21 +946,109 @@ export const PIXELFORMAT_R8 = 52;
...
-    [PIXELFORMAT_R8,            { name: 'R8', size: 1, srgb: true }],
-    [PIXELFORMAT_L8,            { name: 'L8', size: 1 }],
-    [PIXELFORMAT_LA8,           { name: 'LA8', size: 2 }],
-    [PIXELFORMAT_RG8,           { name: 'RG8', size: 2, srgb: true }],
-    [PIXELFORMAT_RGB565,        { name: 'RGB565', size: 2, srgb: true }],
-    [PIXELFORMAT_RGBA5551,      { name: 'RGBA5551', size: 2, srgb: true }],
-    [PIXELFORMAT_RGBA4,         { name: 'RGBA4', size: 2, srgb: true }],
-    [PIXELFORMAT_RGB8,          { name: 'RGB8', size: 4, srgb: true }],
-    [PIXELFORMAT_RGBA8,         { name: 'RGBA8', size: 4, srgb: true }],
+    [PIXELFORMAT_A8,            { name: 'A8', size: 1, ldr: true }],
```
slimbuck, line 1042: I still think it would be neater to flag the few hdr formats as `hdr` rather than the relatively many formats as `ldr`. (Integer formats aren't color formats and so shouldn't get either flag).

slimbuck (approved):
Approved with some comments

src/framework/parsers/texture/basis.js
mvaligursky: `Img` handles it too, but as it only supported a single format, it's handled differently.

I considered handling it in Texture constructor, but did't like that the user passes in `format`, and would also pass in `srgb` flag, and under the hood we'd swap the format to a different format. I think code like this belongs to where it is now.

src/platform/graphics/constants.js
mvaligursky: `requiresManualGamma` it is, I like it.

Issue #5943, opened 2024-01-15
Issues and possible improvements to transparency rendering


mvaligursky:
The default framebuffer WebGL uses for rendering is RGB(A)8 buffer. The values written to it are expected to be in the display color space, which typically means sRGB. As this framebuffer is created by the canvas, we have no control over it, and it is not created with sRGB (EXT_sRGB) flag. Meaning the shader writing to it needs to manually encode values from linear to sRGB space.

### LDR rendering
- when rendering directly to the default framebuffer, the engine correctly encodes values to sRGB space at the end of shaders. This causes alpha blending to not be correct, as the blending takes place in sRGB space, instead of correctly in linear space, with the result getting encoded to sRGB. This is how PlayCanvas currently operates if rendering directly to the default framebuffer.
- proposal to support linear workflow: https://www.w3.org/Graphics/Color/Workshop/slides/Russell.pdf (it does not seem to be progressing, see https://groups.google.com/a/chromium.org/g/blink-dev/c/tA7Av4fLQFY/m/oUw_k949AAAJ). 
- a workaround is rendering to a created framebuffer using sRGB, which would alpha blend correctly, and blit the framebuffer to the default framebuffer at the end of the frame. This is expensive, and so not typically used.
- we also implement dithered transparency. As this does not depend on alpha blending, it's not affected by blending in gamma space and works correctly. The only problem is that when enabled, this looks visually different to alpha blending, as that is incorrect.

### HDR rendering
- When we render in HDR mode, we use float format of the internal render target in linear space. This makes alpha blending to work in linear space, which is correct.
- We have problem with transparency dithering here if tone mapping is used - it does not match the correct alpha blending. The problem is that with dithering, the discarded and visible pixels are tone mapped, but as tone-mapping is not linear, it does not match the alpha blending,  as non-blended values are tonemapped.
- Transparency dithering works correctly if tone mapping is used, as long as TAA is used as well - as TAA resolves transparency by blending with history buffer, and this blending takes place before the tone mapping.

**References:**
https://github.com/KhronosGroup/WebGL/issues/2474
https://github.com/mrdoob/three.js/issues/23019
https://stackoverflow.com/questions/51032480/how-do-you-implement-a-gamma-correct-workflow-in-webgl2
https://groups.google.com/g/angleproject/c/5ITMg4_m8Ug
https://github.com/playcanvas/engine/issues/5287

MAG-AdrianMeredith:
[...]

PR #6752, opened 2024-06-26
Render targets can use sRGB color buffer

mvaligursky:
partially implements https://github.com/playcanvas/engine/issues/3715

- when a color buffer with sRGB format (for example SRGBA8) is used by RenderTarget to render to, the generated shaders skip in-shader linear to gamma conversion, as hardware does it.
- this is not implemented for the main backbuffer yet. Some explanation for this is here: https://github.com/playcanvas/engine/issues/5943 but a separate PR will try to address it for WebGPU.
- shaders of few examples are updated to work correctly in linear space, as sampling from render targets in those examples are linear values


src/platform/graphics/render-target.js
```diff
@@ -482,6 +482,16 @@ class RenderTarget {
     get height() {
         return this._colorBuffer?.height || this._depthBuffer?.height || this._device.height;
     }
+
+    /**
+     * True if the first color buffer format is sRGB.
```
willeastcott, line 487:
```suggestion
     * Gets whether the first color buffer format is sRGB.
```

willeastcott (approved):

PR #6755, opened 2024-06-27
[BREAKING] Remove long time deprecated PHONG specular model

mvaligursky:
- removed from engine v2, and the Blinn-phone or ggx are now the only options


willeastcott (approved):

Maksims:
Would be great to have very simple material shader for high performance on low-end mobile. Poki games still use a lot of simple shaders to ensure highest performance possible.

mvaligursky:
I'd say currently the BasicMaterial is the simplest shaders we have. I does not handle lighting though, not even a single direction light, so maybe that could be added. But then, it's not exposed in the Editor at all, so not very convenient.

willeastcott:
If we refactored `StandardMaterial` to just be a container for bare properties, it'd be simple to have other material types such as `PhongMaterial`.

Maksims:
> I'd say currently the BasicMaterial is the simplest shaders we have. I does not handle lighting though, not even a single direction light, so maybe that could be added. But then, it's not exposed in the Editor at all, so not very convenient.

Yep, Editor UI for it would be amazing.

PR #6759, opened 2024-06-27
[BREAKING] Remove StandardMaterial.diffuseTint and simply always apply it

mvaligursky:
- before: When the diffuseTint is true, the diffuse color is used. But only when diffuse texture nor vertex colors were used. This is too confusing, and generates shader variants to save on a uniform and a single multiply.
- now: tint is always on, and applied even when diffuse texture or vertex colors are used

A separate PR will apply this to additional tints of the material.


slimbuck (approved):

Maksims:
This will definitely break a lot of materials for people, as they might have tweaked diffuse color, especially when it is set to weird greys from imported models most of the time.

mvaligursky:
Hi @Maksims - yes, we'll be adjusting the materials in the Editor when the project loads with engine 2, and handle it. If the tint switch is false, we'll set the color to white.

PR #6762, opened 2024-06-28
[BREAKING] Remove StandardMaterial.emissiveTint and simply always apply it 

mvaligursky:
This is similar to #6759

- StandardMaterial.emissiveTint is always assumed true, and the tint is always applied.
- the emissive color defaults to BLACK as before, which makes materials correctly non-emissive by default
- but when an emissiveMap is assigned to a material, the emissive colors needs to be set WHITE for the map to work (user gets a warning if that is not the case)


slimbuck:
For emissive I actually think this is too simple. Behavior will go against expectation when user sets an emissive map and sees nothing, because emissiveColor is black (the default). Editor would also require magic code to flip colors when map is set etc.

So in this case it would probably be better to have an emissiveColor (r,g,b) and a separate emissiveMapFactor (float probably).

Emissive is: color + map * factor.

Default emissiveColor is (0,0,0), default factor is 1.

Maksims:
> So in this case it would probably be better to have an emissiveColor (r,g,b) and a separate emissiveMapFactor (float probably).
> 
> Emissive is: color + map * factor.
> 
> Default emissiveColor is (0,0,0), default factor is 1.

That would change it's behaviour, which would be worse. As atm it is very useful to "colorize" a single channel emissive. Because map is multiplied by color. Not having that multiply would loose that feature that we use very often.

slimbuck:
> > So in this case it would probably be better to have an emissiveColor (r,g,b) and a separate emissiveMapFactor (float probably).
> > Emissive is: color + map * factor.
> > Default emissiveColor is (0,0,0), default factor is 1.
> 
> That would change it's behaviour, which would be worse. As atm it is very useful to "colorize" a single channel emissive. Because map is multiplied by color. Not having that multiply would loose that feature that we use very often.

In which case factor can be an RGB. :)

Maksims:
> > > So in this case it would probably be better to have an emissiveColor (r,g,b) and a separate emissiveMapFactor (float probably).
> > > Emissive is: color + map * factor.
> > > Default emissiveColor is (0,0,0), default factor is 1.
> > 
> > 
> > That would change it's behaviour, which would be worse. As atm it is very useful to "colorize" a single channel emissive. Because map is multiplied by color. Not having that multiply would loose that feature that we use very often.
> 
> In which case factor can be an RGB. :)

That would eliminate one of the points of removing the tint, is to free up the uniform slot.
I believe there is no holy-grail solution here, and users will have to deal with a new mechanics. Unfortunately tint flag does improve some UX for code API and Editor.
Important thing is not to introduce magical behaviors that are even harder to explain and communicate.

slimbuck:
One of the reasons to remove tint is so we don't generate shader permutations based on this flag. I guess we could keep the tint flag and just not generate a shader permutation from it. 

But then emissive would work differently from the other material channels (diffuse, specular etc) and the api wouldn't be uniform and predictable.

I don't think adding an extra uniform is too big a cost in this case, but really @mvaligursky is probably best placed to weigh up the pros and cons here.

mvaligursky:
I don't mind the extra cost of the uniform (that will be mostly invisible cost once we switch to uniform buffers), but I do not like that most (all?) other channels will have a single tint-like value while emissive will have two.

I think I'd prefer to have Debug.warnOnce when the material with emissiveMap is used while the emissive is black, to point the possible issue to the user.

Maksims:
Also, the mapping to GLB specs assumes only two fields: emissiveTexture and emissiveFactor (which is emissive color as it is in the engine now). So introducing another color might confuse and complicate what should be used or adjusted.

mvaligursky:
Added warnOnce

Screenshot 2024-07-02 at 11 05 32

slimbuck (approved):
Approved!, but please wait for @willeastcott to also check before merging.

src/scene/materials/standard-material.js
```diff
@@ -755,12 +756,14 @@ class StandardMaterial extends Material {
...
-            this._setParameter('material_emissive', getUniform('emissive'));
-        }
-        if (this.emissiveIntensity !== 1) {
-            this._setParameter('material_emissiveIntensity', this.emissiveIntensity);
-        }
+        Debug.call(() => {
+            if (this.emissiveMap && this.emissive.r === 0 && this.emissive.g === 0 && this.emissive.b === 0) {
+                Debug.warnOnce(`Emissive map is set but emissive color is black, making the map invisible. Set emissive color to white to make the map visible. Rendering [${DebugGraphics.toString()}]`, this);
+            }
+        });
```
willeastcott, line 763: I can imagine situations where this warning will trigger for intentional situations - like if I'm lerping emissive up and down over time. That might be annoying in debug.

mvaligursky: I thought about that as well. Ideally you'd lerp `emissiveIntensity` in that case and not the color

src/scene/materials/standard-material.js
```diff
@@ -298,10 +299,10 @@ const _tempColor = new Color();
...
  * @property {Color} emissive The emissive color of the material. This color value is 3-component
  * (RGB), where each component is between 0 and 1.
- * @property {boolean} emissiveTint Multiply emissive map and/or emissive vertex color by the
- * constant emissive value.
- * @property {Texture|null} emissiveMap The emissive map of the material (default is null). Can be
- * HDR.
+ * @property {Texture|null} emissiveMap The emissive
+ * map of the material (default is null). Can be HDR. When the emissive map is applied, the emissive
+ * color is multiplied by the texel color in the map. Since the emissive color is black by default,
+ * the emissive map won't be visible unless the emissive color is changed.
```
willeastcott, line 305:
```suggestion
* @property {Texture|null} emissiveMap The emissive map of the material (default is null). Can be
* HDR. When the emissive map is applied, the emissive color is multiplied by the texel color in the
* map. Since the emissive color is black by default, the emissive map won't be visible unless the
* emissive color is changed.
```

willeastcott:
Is `emissiveIntensity` useful any more? Can't you just set `emissive` to [2, 2, 2], for example?

glTF just specifies:

* `emissiveTexture`
* `emissiveFactor` (defaults to `[0, 0, 0]`.

`emissiveTexture` is always multiplied against `emissiveFactor`. If `emissiveTexture` is not specified, `[1, 1, 1]` is used as the emissive color (but by default, the result is still black because `emissiveFactor` is `[0, 0, 0]`).

mvaligursky:
In theory yes, but it's not convenient for the uses.
We specify color in sRGB space and that is only valid for colors 0..1 range.
And then we need a linear intensity multiplier to have bright emissive.

At runtime, we convert color from sRGB to linear, and multiply intensity .. and use this as the uniform value.
Unity engine does the same for example.

If we only stored linear HDR emissive color, for the Editor we'd need to convert it to intensity and sRGB color for the inspector. Or we could have an HDR color picker that handles it, which we don't.

The same is for engine only users .. for them it's more convenient to specify colors in sRGB space, as that's what they normally work with.

willeastcott (approved):
OK, approving, but somewhat nervous about people encountering broken projects....and everyone using Spline being forced to get the latest library.

mvaligursky:
For the Editor projects, the plan is to upgrade the materials so this should be a no problem for them, if material uses emissive map, we'll set the emissive color to white if it was black before.

Maksims:
> For the Editor projects, the plan is to upgrade the materials so this should be a no problem for them, if material uses emissive map, we'll set the emissive color to white if it was black before.

Only in cases where `emissiveTint` was `false` also.

PR #6763, opened 2024-06-28
Gizmo fixes

### LINKED ISSUE #6671: Gizmo sizing bug
Gizmo appears different sizes under different circumstances.

Seems to be related to the `Gizmo._deviceStartSize` value which should just removed completely. (See [here](https://github.com/playcanvas/engine/blob/main/src/extras/gizmo/gizmo.js#L257)).

### LINKED ISSUE #6695: Gizmo is not interactable when close
### Description
Gizmo is not interactable when close

### Steps to Reproduce

1. go to examples https://playcanvas.github.io/#/misc/gizmos
2. get really close to the gizmo
3. it is not possible to interact with the gizmo

https://github.com/playcanvas/engine/assets/100127317/348d1824-5ceb-4d2c-83f8-f6751b47bdf4

kpal81xd:
Fixes #6671 
Fixes #6695 

- Gizmo scale now just dependent on FOV and distance from camera
- Selection ray fired from near plane instead of a distance of 1 from the camera
- Added JSDoc to all methods in gizmo classes 
- Used event names instead of raw strings for callbacks and fire methods
- Used axis constants instead of raw string values


src/extras/gizmo/axis-shapes.js
```diff
@@ -51,7 +51,7 @@ const SHADER = {
             // store z/w for later use in fragment shader
             vZW = gl_Position.zw;
             // disable depth clipping
-            gl_Position.z = 0.0;
+            // gl_Position.z = 0.0;
```
willeastcott, line 54: Delete?

examples/src/examples/misc/gizmos.example.mjs
```diff
@@ -122,7 +122,8 @@ data.set('camera', {
 });
 const camera = new pc.Entity('camera');
 camera.addComponent('camera', {
-    clearColor: new pc.Color(0.1, 0.1, 0.1)
+    clearColor: new pc.Color(0.1, 0.1, 0.1),
+    farClip: 1000
```
willeastcott, line 126: Why won't the example work with the default far clip?

kpal81xd: It does it you can just see the the axis lines clipping when zooming out

src/extras/gizmo/axis-shapes.js
kpal81xd: It looks cleaner when clipping with this so would prefer to have it in but the selection area and rendered area do not align

src/extras/gizmo/rotate-gizmo.js
```diff
@@ -264,28 +265,40 @@ class RotateGizmo extends TransformGizmo {
         return this._shapes.x.tolerance;
     }
 
+    /**
+     * @param {string} prop - The property.
+     * @param {any} value - The value.
```
willeastcott, line 270: Isn't the type here `number`?

src/extras/gizmo/scale-gizmo.js
```diff
@@ -315,40 +315,67 @@ class ScaleGizmo extends TransformGizmo {
         return this._shapes.xyz.tolerance;
     }
 
+    /**
+     * @param {string} prop - The property name.
+     * @param {any} value - The property value.
```
willeastcott, line 320: `number`?

src/extras/gizmo/scale-gizmo.js
```diff
@@ -315,40 +315,67 @@ class ScaleGizmo extends TransformGizmo {
...
+     */
     _setArrowProp(prop, value) {
         this._shapes.x[prop] = value;
         this._shapes.y[prop] = value;
         this._shapes.z[prop] = value;
     }
 
+    /**
+     * @param {string} prop - The property name.
+     * @param {any} value - The property value.
```
willeastcott, line 331: `number`?

src/extras/gizmo/transform-gizmo.js
```diff
@@ -480,12 +485,22 @@ class TransformGizmo extends Gizmo {
...
+     */
     _colorSemi(color) {
         const clone = color.clone();
         clone.a = this._colorAlpha;
         return clone;
     }
 
+    /**
+     * @param {string} axis - The axis to update.
+     * @param {any} value - The value to set.
```
willeastcott, line 501: `Color`?

src/extras/gizmo/translate-gizmo.js
```diff
@@ -305,18 +311,31 @@ class TranslateGizmo extends TransformGizmo {
         return this._shapes.face.tolerance;
     }
 
+    /**
+     * @param {string} prop - The property to set.
+     * @param {any} value - The value to set.
```
willeastcott, line 316: OK, I'm gonna stop pointing these out now....maybe there's a good reason it's `any`! 😄

willeastcott (approved):

src/extras/gizmo/translate-gizmo.js
kpal81xd: All of these are any because its a generic property update so the property could technically be any type

PR #6771, opened 2024-07-01
Fallback to int-texture based morphing instead of attribute when float texture rendering is not supported

mvaligursky:
- before, we'd use float / half-float render target to which morph targets get blended. If rendering to those format was not supported, we'd use vertex attribute based morphing, which was limited to 4 positions + 4 normals
- now that we don't need to handle WebGL1, but only WebGL2 - which on about 1% of devices still does not support float / half-float render targets, we fall back to 16bit per component integer textures, and encode the position / normals to maximize the precision.
- this allows us to remove support for vertex attribute based morphing


src/scene/shader-lib/programs/basic.js
```diff
@@ -139,7 +139,7 @@ class ShaderGeneratorBasic extends ShaderGenerator {
         if (options.useInstancing)          key += '_inst';
         if (options.useMorphPosition)       key += '_morphp';
         if (options.useMorphNormal)         key += '_morphn';
-        if (options.useMorphTextureBased)   key += '_morpht';
+        if (options.useMorphTextureBasedInt)key += '_morphi';
```
willeastcott, line 142: Can you insert an extra space here.

willeastcott (approved):
Great improvement.

src/scene/shader-lib/programs/basic.js
mvaligursky: done thanks

PR #6774, opened 2024-07-02
WebXR - fix depth sensing on quest 3

Maksims:
A recent update of OculusBrowser depth sensing implementation broke access to it.
There was an update to the WebXR Depth Sensing specs and implementation of it on Meta Quest: https://github.com/immersive-web/depth-sensing/pull/49
This PR ensures that we take that edge case into account.

### DIFF
--- a/src/framework/xr/xr-manager.js
+++ b/src/framework/xr/xr-manager.js
@@ -477,19 +477,24 @@ class XrManager extends EventHandler {
             if (options && options.depthSensing && this.depthSensing.supported) {
                 opts.optionalFeatures.push('depth-sensing');
 
-                const usagePreference = [XRDEPTHSENSINGUSAGE_CPU];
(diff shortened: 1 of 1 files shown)


willeastcott:
Can a fix be achieved without sniffing the user agent? I'm not a big fan of doing that...
Is Quest not adhering to the spec? Is it a bug? Would you expect the edge case to be removed at some point?

Maksims:
> Can a fix be achieved without sniffing the user agent? I'm not a big fan of doing that...

Unfortunately, I'm not aware of other ways of detecting Quest browser.

> Is Quest not adhering to the spec? Is it a bug? Would you expect the edge case to be removed at some point?

It is somewhat weird behaviour currently, I will definitely raise an issue with their implementation, but currently it fixes a problem. Maybe they will improve it in the future.

willeastcott:
OK. Well, I'll approve + merge for now, but we should return to this at a later date...

willeastcott (approved):

PR #6777, opened 2024-07-02
Use object destructuring with vectors

willeastcott:
Less code === better! 🚀 

### DIFF
--- a/src/core/math/mat3.js
+++ b/src/core/math/mat3.js
@@ -376,9 +376,7 @@ class Mat3 {
     transformVector(vec, res = new Vec3()) {
         const m = this.data;
 
-        const x = vec.x;
-        const y = vec.y;
-        const z = vec.z;
+        const { x, y, z } = vec;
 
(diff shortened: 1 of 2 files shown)


LeXXik:
I prefer the old way, without object allocation.

mvaligursky (approved):

mvaligursky:
It does not allocate an object, even though the syntax seems to suggest it is.

willeastcott:
Yeah. It transpiles to exactly what it was before via Babel @LeXXik. It's just more concise code.

willeastcott:
Yeah, I know I could have destructured in the parameter itself, but somehow, I thought that was a little bit ugly. 😱

kungfooman:
> Yeah, I know I could have destructured in the parameter itself, but somehow, I thought that was a little bit ugly. 😱

Yep, I agree on that, things can be made so short that it just looks like ASCII soup afterwards :sweat_smile:

PR #6780, opened 2024-07-03
Minor refactor of Curve class

willeastcott:
The main thing I wanted to do was remove the dependency on the `extend` function.

### DIFF
--- a/src/core/math/curve.js
+++ b/src/core/math/curve.js
@@ -82,7 +82,7 @@ class Curve {
      *
      * @param {number} time - Time to add new key.
      * @param {number} value - Value of new key.
-     * @returns {number[]} [time, value] pair.
+     * @returns {number[]} The newly created `[time, value]` pair.
      */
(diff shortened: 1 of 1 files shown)


src/core/math/curve.js
```diff
@@ -154,7 +156,7 @@ class Curve {
     clone() {
         /** @type {this} */
         const result = new this.constructor();
-        result.keys = extend(result.keys, this.keys);
+        result.keys = this.keys.map(key => [...key]);
```
mvaligursky, line 159: the extend function seemed to do a deep copy, but the new solution does a shallow copy only, which is probably not intended?

willeastcott: The `keys` are an array of arrays of two numbers. So I can't see how this would go wrong.

mvaligursky: yep that should be fine , great!

mvaligursky (approved):

PR #6784, opened 2024-07-04
Refactor `Frustum` to use `Plane`

willeastcott:

### DIFF
--- a/src/core/shape/frustum.js
+++ b/src/core/shape/frustum.js
@@ -43,22 +43,20 @@ class Frustum {
      * frustum.setFromMat4(projMat);
      */
     setFromMat4(matrix) {
-        const vpm = matrix.data;
+        const [
+            m00, m01, m02, m03,
+            m10, m11, m12, m13,
+            m20, m21, m22, m23,
+            m30, m31, m32, m33
+        ] = matrix.data;
         const planes = this.planes;
 
-        const setPlane = (plane, a, b, c, d) => {
-            plane.normal.set(vpm[3] + a * vpm[0], vpm[7] + b * vpm[4], vpm[11] + c * vpm[8]);
-            plane.distance = vpm[15] + d * vpm[12];
-            plane.normalize();
-        };
-
-        // Extract the numbers for the planes
-        setPlane(planes[0], -1, -1, -1, -1); // RIGHT
-        setPlane(planes[1], 1, 1, 1, 1);     // LEFT
-        setPlane(planes[2], 1, 1, 1, 1);     // BOTTOM
-        setPlane(planes[3], -1, -1, -1, -1); // TOP
-        setPlane(planes[4], -1, -1, -1, -1); // FAR
-        setPlane(planes[5], 1, 1, 1, 1);     // NEAR
+        planes[0].set(m03 - m00, m13 - m10, m23 - m20, m33 - m30).normalize(); // RIGHT
+        planes[1].set(m03 + m00, m13 + m10, m23 + m20, m33 + m30).normalize(); // LEFT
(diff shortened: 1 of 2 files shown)


src/core/shape/frustum.js
```diff
@@ -135,7 +133,7 @@ class Frustum {
         let c = 0;
         for (let p = 0; p < 6; p++) {
             const plane = planes[p];
-            const d = plane[0] * scx + plane[1] * scy + plane[2] * scz + plane[3];
+            const d = plane.normal.x * scx + plane.normal.y * scy + plane.normal.z * scz + plane.distance;
```
mvaligursky, line 125: I'd probably do `const { normal, distance } = plane[p];` in a hope it might run faster. Likely no difference though.

mvaligursky (approved):

src/core/shape/frustum.js
```diff
@@ -34,66 +43,22 @@ class Frustum {
...
 
-        // Extract the NEAR plane
-        plane = planes[5];
-        plane[0] = vpm[3] + vpm[2];
-        plane[1] = vpm[7] + vpm[6];
-        plane[2] = vpm[11] + vpm[10];
-        plane[3] = vpm[15] + vpm[14];
-        normalize(plane);
+        // Extract the numbers for the planes
+        setPlane(planes[0], -1, -1, -1, -1); // RIGHT
```
mvaligursky, line 56: hmm not sure about this last commit?
for example in the original code:
```
plane[0].distance = vpm[15] - vpm[12];
plane[4].distance = vpm[15] - vpm[14];
```

but new code is always
```
plane.distance = vpm[15] + d * vpm[12];
```

but with some optional sign

src/core/shape/frustum.js
willeastcott: Done. Thanks!

src/core/shape/frustum.js
willeastcott: Yep, I messed up. I've reimplemented this. So much nicer! 🚀

src/core/shape/plane.js
```diff
@@ -11,7 +11,6 @@ class Plane {
      * The normal of the plane.
      *
      * @type {Vec3}
-     * @readonly
```
kungfooman, line 14: Instead of removing it in `Plane` class I would rather add `@readonly` to `Frustum#planes` aswell. It doesn't mean that you shouldn't update the content, but only that you guarantee the instance of the array will be the same all the time. E.g. instead of "GC causing cleaning" like many people would do via `frustum.planes = []` it  be invalid and instead `frustum.planes.length = 0` should be used.

willeastcott: It is perfectly valid to do whatever you like to the plane's normal Vec3. Replace it, edit it, whatever - nothing will break. So it's right to remove it. As for the `planes` property, again, if you _really_ want to replace the array with something else, is that really a problem?

kungfooman: For example https://github.com/playcanvas/engine/pull/2088, issue is the ease of introducing problems in which, for instance, `Vec3.ZERO` has been overwritten. So some developers basically tried to settle on prefering ownership of variables in classes and not mixing around internal variables with other internal variables in the hope it creates code that is easier to quickly argue about, easier to maintain and allowing more idiomatic code (e.g. ability to use Vec3 constants in constructors).

If that is not the case in your observation, you could of course also remove all the other `@readonly`'s aswell that were added just some time ago. But idk, clearly different people pull in different directions which creates some friction, so it might be time to finally reach a consensus on this point.

mvaligursky: agreed we need to keep the `@readonly` here, to avoid possible issues we've just removed in other shapes.

PR #6787, opened 2024-07-05
Add support for assigning properties to script instances

### LINKED ISSUE #6785: Create a ESM Script with attributes
Scripts components have a `create()` method for instantiating scripts with attributes.

```javascript
scripts.create('scriptName', { attributes: { myNum: 1 })`
```

This depends on an attribute schema existing. ESM Scripts use [JSDoc tags](https://github.com/playcanvas/editor/issues/1148) for their schema which is unavailable in engine only environments, **meaning attributes cannot be set on a script**

For ESM scripts in engine only, properties must be set after a script is instantiated:

```javascript
const script = scripts.create('scriptName')`
script.myNum = 1
```

Ideally properties can also be set in-line on a script instance, at the same time as it's created.

### 1. Re-use `attribute` property
In the editor, a schema will exist, so it will be used to create the attributes, when it doesn't, the values are simply assigned as properties onto the script instance
```javascript
scripts.create('scriptName', { attributes: { myNum: 1 }})
```

- **Pros**
Simple. Re-uses existing api. 

 - **Cons**
**Inconsistent behaviour between editor/engine.** - In editor, this will validate and deep copy values. In engine, it will not.
**semantics** Attributes don't inherently exist in engine only, so this is overloading the term.

### 2. Separate method for engine only
Create a new method eg `scripts.createWithProperties('scriptName', { myNum: 1 })` (TBD) that assigns values when instantiated
```javascript
scripts.createWithProperties('scriptName', { myNum: 1 })
```

- **Pros**
Clear and separate behaviour. Does not overload the `attributes` property

 - **Cons**
Creates an additional method which could lead to confusion. `create()` still exists and would work for instantiating a script with no props/attributes

### 3. Create an additional property
Attributes are applied first if they and schema exist, and `properties` are then assigned after.
```javascript
scripts.create('scriptName', { properties: { myNum: 1 }})
```

 - **Pros** 
No additional method. Clear distinction between behaviour of `attributes` and `properties` (attributes require schema, props do not)

 - **Cons**
Adds additional  property. ??

Any thoughts @mvaligursky @willeastcott @Maksims @slimbuck

marklundin:
This PR adds a `properties` field to the `scripts.create(name, { properties })` which is used to assign values onto a script instance _before_ `initialize()` is called.

Fixes #6785


kpal81xd (approved):

mvaligursky (approved):
It'd be great to add an example to Script class module on how to use this.

PR #6804, opened 2024-07-11
Refactor the way shaders get created for particle systems

mvaligursky:
- remove manual handling of shader creation, where shaders would need to be recreated when the camera was supplied on the second frame
- instead, use the same system as other material / shader types and create shader when needed by the renderer, using it's supplied parameters - new internal `ParticleMatrial` class handles it
- slight clean up to the way Editor triggers a custom mode for 2D particles

- exposed few parameters to the engine example to allow for testing
Screenshot 2024-07-11 at 10 51 58

### DIFF
--- a/src/scene/particle-system/particle-material.js
+++ b/src/scene/particle-system/particle-material.js
@@ -1,0 +1,74 @@
+import { Debug } from '../../core/debug.js';
+import { ShaderProcessorOptions } from '../../platform/graphics/shader-processor-options.js';
+import {
+    GAMMA_NONE,
+    PARTICLEORIENTATION_SCREEN,
+    SHADER_FORWARD,
+    TONEMAP_LINEAR
+} from '../constants.js';
+import { getProgramLibrary } from '../shader-lib/get-program-library.js';
+import { Material } from '../materials/material.js'
+import { particle } from '../shader-lib/programs/particle.js';
+
+/**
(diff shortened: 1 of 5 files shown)


slimbuck (approved):

src/scene/particle-system/particle-material.js
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+/**
+ * A material for rendering particle geometry by the particle emitter.
+ *
+ * @category Graphics
+ */
+class ParticleMaterial extends Material {
+    /**
+     * The color of the particles.
```
willeastcott, line 24: This description was wrong BTW. 😄

src/scene/particle-system/particle-material.js
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * @import { ParticleEmitter } from './particle-emitter.js'
+ */
+
+/**
+ * A material for rendering particle geometry by the particle emitter.
+ *
+ * @category Graphics
+ */
```
willeastcott, line 21: Do we really want this to be public API?

src/scene/particle-system/particle-material.js
willeastcott: Oh, I just realized `ParticleMaterial` isn't exported from the root `index.js` so I guess it's fine.

PR #6805, opened 2024-07-11
[BREAKING] Remove StandardMaterial.ambientTint

mvaligursky:
Marked as breaking as it has a potential to change the brightness of the scene. The workaround is to set the tint to WHITE.

- remove the ambientTint property, and make it behave as always true
- change the default value from grey to white, to make it a noop when multiplied by the scene ambient

### DIFF
--- a/src/deprecated/deprecated.js
+++ b/src/deprecated/deprecated.js
@@ -690,6 +690,7 @@ function _deprecateTint(name) {
 
 _deprecateTint('diffuseTint');
 _deprecateTint('emissiveTint');
+_deprecateTint('ambientTint');
 
 _defineAlias('specularTint', 'specularMapTint');
 _defineAlias('aoVertexColor', 'aoMapVertexColor');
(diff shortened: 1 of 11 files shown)


Maksims:
This one potentially has the biggest chance to affect scenes (compared to diffuseTint and emissiveTint).

Tested projects with this code:
```js
editor.assets.list().filter((a) => {
    return a.get('type') === 'material'
        && a.get('data.ambientTint') === false
        && (a.get('data.ambient.0') !== 1 || a.get('data.ambient.1') !== 1 || a.get('data.ambient.2') !== 1);
}).length;
```

In our existing projects, it will affect 70-100% of materials.

mvaligursky:
Yep, same as with the other tints, we'll process the project as they switch to V2 of the engine, and if the ambientTint was false, we'll set the ambient color to 1,1,1. So that should solve majority of Editor projects.

slimbuck (approved):

PR #6812, opened 2024-07-12
Remove opacityTint internal flag

mvaligursky:
- not a breaking change as this was internal flag only, used to optimize one uniform
- this was causing a shader recompilation if alpha value change / blending mode changes, and so is removed.

### DIFF
--- a/src/framework/components/element/system.js
+++ b/src/framework/components/element/system.js
@@ -392,8 +392,6 @@ class ElementComponentSystem extends ComponentSystem {
         material.emissiveMap = this._defaultTexture;
         material.opacityMap = this._defaultTexture;
         material.opacityMapChannel = 'a';
-        material.opacityTint = true;
-        material.opacity = 0; // use non-1 opacity to compile shader correctly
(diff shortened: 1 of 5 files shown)


src/framework/components/element/system.js
```diff
@@ -392,8 +392,6 @@ class ElementComponentSystem extends ComponentSystem {
         material.emissiveMap = this._defaultTexture;
         material.opacityMap = this._defaultTexture;
         material.opacityMapChannel = 'a';
-        material.opacityTint = true;
-        material.opacity = 0; // use non-1 opacity to compile shader correctly
```
slimbuck, line 396: did you mean to remove this line? default opacity is 1.

mvaligursky: yeah it was added to force shader compilation .. later this is set to what it actually needs to be

slimbuck: NASTY ALERT, thanks

slimbuck (approved):

PR #6822, opened 2024-07-16
Gizmo size invariant to canvas resizing

kpal81xd:
Uses the bounding css dimensions of the canvas to keep the gizmo size invariant when resizing the canvas.


kpal81xd:
Better to keep scale invariant outside gizmo and handled by user

PR #6826, opened 2024-07-16
[BREAKING] Gizmo constants

### LINKED ISSUE #6791: Gizmo SHAPEAXIS constants
See:

* https://api.playcanvas.com/classes/Engine.TransformGizmo.html#enableShape
* https://api.playcanvas.com/classes/Engine.TransformGizmo.html#isShapeEnabled

The `SHAPEAXIS` constants are not exported in the root `index.js` so are not public API. So either these functions should be internal, or we should make the constants public API.

kpal81xd:
Fixes #6791

- Renames `GIZMO_<placeholder>` to `GIZMOCOORD_<placeholder>` (public API)
- Renames `SHAPEAXIS_<placeholder>` to `GIZMOAXIS_<placeholder>` and exposes to public API

### DIFF
--- a/src/extras/gizmo/gizmo.js
+++ b/src/extras/gizmo/gizmo.js
@@ -6,6 +6,8 @@ import { EventHandler } from '../../core/event-handler.js';
 import { PROJECTION_PERSPECTIVE } from '../../scene/constants.js';
 import { Entity } from '../../framework/entity.js';
 
+import { GIZMOCOORD_LOCAL, GIZMOCOORD_WORLD } from './constants.js';
+
 /**
  * @import { AppBase } from '../../framework/app-base.js'
  * @import { CameraComponent } from '../../framework/components/camera/component.js'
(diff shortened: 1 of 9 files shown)


src/extras/gizmo/gizmo.js
```diff
@@ -33,15 +33,15 @@ const ORTHO_SCALE_RATIO = 0.32;
  * @type {string}
  * @category Gizmo
  */
-export const GIZMO_LOCAL = 'local';
+export const GIZMOCOORD_LOCAL = 'local';
```
mvaligursky, line 36: I would consider moving all constants from few modules here to `constants.js` as we do in the engine to keep consistent. This also allows the import of constants, without importing the rest of the module, which often avoids circular references to be created.

src/extras/gizmo/translate-gizmo.js
```diff
@@ -305,7 +305,7 @@ class TranslateGizmo extends TransformGizmo {
      * @type {number}
      */
     set axisCenterTolerance(value) {
-        this._shapes.face.tolerance = value;
+        this._shapes.f.tolerance = value;
```
mvaligursky, line 308: not a big deal here, but I find .face more readable than .f, same with the constant _F vs _FACE

willeastcott: @kpal81xd I think you missed this thread. 😄

PR #6829, opened 2024-07-18
Updated lint config

mvaligursky:
When I updated dependencies here: https://github.com/playcanvas/engine/commit/3e9547e8b08c68adf22ea0d4c305964e02fb95f9

We got error that the lint no longer supports `--ext` option.

The solution is to move it to the config. The npm config does not support this, and so an external config needs to be used.


mvaligursky:
closing, reverted the upgrade for now

Issue #2962, opened 2021-03-13
[Feature Request] Add shader recompile mechanism


scarletsky:
This is useful for changing macro such as `#define` and other usage.

A simple implementation can be:

```ts
interface ShaderDefinitionDefines {
  [name: string]: string;
}

interface ShaderDefinition {
  vshader: string;
  fshader: string;
  attributes: ShaderDefinitionAttributes;
  defines?: ShaderDefinitionDefines;
}

class Shader {
  // new interface
  recompile() {
    this.destroy();
    this.device.createShader(this);
  }
}

class GraphicsDevice {
  compileAndLinkShader(shader) {
    // update the shader source by definition.defines
    var glVertexShader = this.compileShaderSource(definition.vshader, true);
    var glFragmentShader = this.compileShaderSource(definition.fshader, false);
    // ...
  }
}
```

mvaligursky:
We've discussed something similar for graph based shaders .. and hopefully the implementation will cover both that as well as  chunked based shader.

scarletsky:
@mvaligursky That's great ! When will the graph based shader release ? 😃

PR #6831, opened 2024-07-18
[BREAKING] Implementation of ShaderMaterial - Material to be used with custom shaders

mvaligursky:
This PR changes the way a material with custom shader gets created.

BEFORE
```
// create shader
const shader = pc.createShaderFromCode(app.graphicsDevice, 
        files['shader.vert'], files['shader.frag'], 'myShader', { aPosition: pc.SEMANTIC_POSITION });

// create material
const material = new pc.Material();
material.shader = shader;
```

NOW
```
const shaderDesc = {
        uniqueName: 'DecalShader',
        vertexCode: files['shader.vert'],
        fragmentCode: files['shader.frag'],
        attributes: { aPosition: pc.SEMANTIC_POSITION }
};
const decalMaterial = new pc.ShaderMaterial(shaderDesc);
```

**Advantage:** Instead of creating the shader at this time, only the building blocks for the shader are assigned to the material. This allows the engine to create the shader at a later time, when it has more context. This allows for few features:
- (DONE) The engine can create multiple shader variations, as it does with StandardMaterial. Currently, a define with the shader pass name is added to the shader, and this allows the shader code to generate different code for forward pass, different for picker pass, shadows, or user defined passes. Simply use `#ifdef ..#endif` in the supplied shader to provide different code for different pass.

example fragment shader:
```
// when the shader for a pass needed, we inject a define at the start of the shader, and compile it. Inserted line:
#define FORWARD_PASS      // for the forward pass
#define MYCUSTOMNAME_PASS   // for custom pass enabled on the camera

// inside your part of the shader, you can do
#ifdef MYCUSTOMNAME_PASS
    // do something special
#endif
```

- (FUTURE) - We can expose a list of DEFINES that you can specify on an instance of ShaderMaterial, and under the hood we'll compile a different version of it as needed. For example, you can do something like this (API is not finalized): `materialInstance.setDefine('SNOW', true);`
and handle it inside the shader using `#ifdef SNOW`
It will solve this: https://github.com/playcanvas/engine/issues/2962

- (NEAR FUTURE) - allow your shader to automatically handle different output encoding. Currently most shaders encode to sRGB space, and you have no control over this. This makes the shader incompatible with HDR rendering, where a linear output is expected. We'll have a build-in define for this, allowing you to handle both cases in a single shader, and appropriate variant will be used based on the RenderTarget it is rendered to.

### DIFF
--- a/src/deprecated/deprecated.js
+++ b/src/deprecated/deprecated.js
@@ -613,10 +613,10 @@ GraphNode.prototype.setName = function (name) {
 
 Object.defineProperty(Material.prototype, 'shader', {
     set: function (value) {
-        Debug.deprecated(`pc.Material#sahder is deprecated, use pc.ShaderMaterial instead.`);
+        Debug.deprecated(`pc.Material#shader is deprecated, use pc.ShaderMaterial instead.`);
     },
     get: function () {
-        Debug.deprecated(`pc.Material#sahder is deprecated, use pc.ShaderMaterial instead.`);
+        Debug.deprecated(`pc.Material#shader is deprecated, use pc.ShaderMaterial instead.`);
         return null;
     }
 });
--- a/src/deprecated/deprecated.js
+++ b/src/deprecated/deprecated.js
@@ -611,25 +611,15 @@ GraphNode.prototype.setName = function (name) {
     this.name = name;
 };
 
-Material.prototype.getName = function () {
-    Debug.deprecated('pc.Material#getName is deprecated. Use pc.Material#name instead.');
-    return this.name;
-};
-
-Material.prototype.setName = function (name) {
-    Debug.deprecated('pc.Material#setName is deprecated. Use pc.Material#name instead.');
-    this.name = name;
-};
-
-Material.prototype.getShader = function () {
-    Debug.deprecated('pc.Material#getShader is deprecated. Use pc.Material#shader instead.');
-    return this.shader;
-};
-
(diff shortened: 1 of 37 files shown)


slimbuck:
Can we name it `Desc` instead of `Descr` pretty please? 🙏🏻

src/deprecated/deprecated.js
```diff
@@ -611,25 +611,15 @@ GraphNode.prototype.setName = function (name) {
...
-};
+Object.defineProperty(Material.prototype, 'shader', {
+    set: function (value) {
+        Debug.deprecated(`pc.Material#sahder is deprecated, use pc.ShaderMaterial instead.`);
+    },
+    get: function () {
+        Debug.deprecated(`pc.Material#sahder is deprecated, use pc.ShaderMaterial instead.`);
+        return null;
+    }
+});
```
kpal81xd, line 622: Normally deprecated versions would still support the functionality but with a warning but this only warns with no implementation

kpal81xd (approved):

src/deprecated/deprecated.js
Maksims: There is a typo: pc.Material#sahder

mvaligursky: good point, so far we used deprecated message for both deprecated and removed functionality, but perhaps we could split it up.

slimbuck (approved):

PR #6848, opened 2024-07-24
Debug error removeComponent correct method name

kpal81xd:
Correct Debug error method name for removeComponent


mvaligursky:
We already have an error in the entity constructor that should trigger when entity does not have the app

kpal81xd:
> We already have an error in the entity constructor that should trigger when entity does not have the app

Ahh okay this should throw an error however not just a `Debug.assert`. It gets stripped out of debug build so will silently fail until you attempt to add a component

mvaligursky (approved):

willeastcott (approved):

PR #6850, opened 2024-07-24
Gizmo examples restructure

kpal81xd:
- Creates separate gizmo examples category (allows for future gizmos to be added)
- Created separate examples for translate rotate and scale gizmos
- Moves verbose controls from editor to each individual gizmo example for better clarity
- Refactored editor example gizmo handler to not interface with controls observer

image


mvaligursky (approved):
Perhaps move the Editor example to the same category as well?

kpal81xd:
> Perhaps move the Editor example to the same category as well?

I was going to leave it in misc since itll have gizmos grid and multi-cam etc so will be a combination of categories not just gizmos

Issue #6835, opened 2024-07-19
Remove BasicMaterial and allow better ways to create custom shaders using ShaderMaterial


mvaligursky:
# Current status

BasicMaterial is a simple engine-provided material with the following key aspects:
- it generates low cost shader which supports skinning, morphing, vertex colors and a single texture with fog and picker compatibility
- it does not support lighting and does not provide a way to add a simple lighting, nor it provides a way for users to customize the generated shader
- does not support VSM shadows https://github.com/playcanvas/engine/issues/3298
- internally the engine and the Editor only use it to render lines, which is trivial to replace: https://github.com/playcanvas/engine/pull/6834

# Plan for engine 2.0.0
The plan is to remove BasicMaterial from the engine. There are two replacement paths, one existing and one which will require some further work.

## StandardMaterial
StandardMaterial can be used as a complete replacement, when using emissive texture slot with vertex format. It handles all features of the BasicMaterial (and many more), and generates code with very similar / same performance.
It is exposed in the Editor inspector / Editor workflow.
This should be the main replacement path for users not requiring custom modification.

## ShaderMaterial

This is an under development material / shader system allowing users to use custom shaders, by providing a custom shader source code, which the engine implements or helps user to implement additional features. Main features / unfinished work:
- [x] internally, the shader code processing in the engine now supports `#include` pragma, allowing an existing shader chunks to be included in the custom shaders. Currently this is used to allow user to easily handle gamma / tone-mapping and fog.
- [x] support for handling of render passes was added here https://github.com/playcanvas/engine/pull/6831
- [ ] support for (semi)automatic shader support with skinning and morphing should be added
- [ ] easier support to handle shadow and picker pass needs to be added, using `#include`
- [ ] support by the batching
- [ ] write an engine example demonstrating a most the these features

When this is done, it will give users an option to create a custom look shader, which can be skinned, which will handle tone-mapping and fog and many other features, opening a way for custom looking applications (targeting mostly non-photorealistic games here)

I look forward to a feedback here.

willeastcott:
@mvaligursky Is this complete now?

mvaligursky:
There are still those two remaining improvements I would not mind to get to at some point.

kungfooman:
LLM's still use `pc.BasicMaterial` all the time, how can we polyfill for still having simple unlit geometry and same performance?

Currently I have:

```js
// Polyfill for simple unlit geometry in PlayCanvas v2
[...]

PR #6857, opened 2024-07-29
[BREAKING] BasicMaterial has been deleted

mvaligursky:
If an alternative is needed for the removed BasicMaterial, a StandardMaterial can be used with emissive color / map instead.

related to https://github.com/playcanvas/engine/issues/6835


slimbuck:
It might be worth implementing BasicMaterial as a ShaderMaterial instead of removing it completely? This would function as an engine-internal example of ShaderMaterial usage.

mvaligursky:
If we need an example, we should have it in the examples folder, not internal in the engine.

slimbuck (approved):

PR #6862, opened 2024-07-30
Removed references to no longer used _shader from the Material

mvaligursky:
- all materials use and store shaders in variants cache, and the single _shader is no longer used

### DIFF
--- a/src/scene/batching/batch-manager.js
+++ b/src/scene/batching/batch-manager.js
@@ -877,9 +877,6 @@ class BatchManager {
         }
 
         batch2.meshInstance.castShadow = batch.meshInstance.castShadow;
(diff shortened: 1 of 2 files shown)


willeastcott (approved):

src/scene/batching/batch-manager.js
```diff
@@ -876,9 +876,6 @@ class BatchManager {
             batch2.meshInstance.skinInstance = new SkinBatchInstance(this.device, nodes, this.rootNode);
         }
 
-        batch2.meshInstance.castShadow = batch.meshInstance.castShadow;
```
willeastcott, line 879: Did you mean to delete this line?

Issue #3526, opened 2021-09-29
Support glTF extension: EXT_mesh_gpu_instancing


issacclee:
[EXT_mesh_gpu_instancing](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Vendor/EXT_mesh_gpu_instancing/README.md) is a multi vendor extensions supported by many other WebGL engine that can load gltf.
Would Playcanvas consider support such extension?
Am I right to assume the Batch function has already implemented whatever is needed to support such extension?

mvaligursky:
It's definitely something we're keen to support, but don't have it on the short term road map.

We currently support these technologies, which would likely be used to implement this:
- hardware instancing: http://playcanvas.github.io/#/graphics/hardware-instancing
- batching http://playcanvas.github.io/#/graphics/batching-dynamic

As hardware instancing is not supported on all hardware, we'd probably use batching as a backup.

mvaligursky:
If you have some example GLBs using this feature, feel free to attach them here if possible to share them.

issacclee:
Hi @mvaligursky ,
I will upload some glb with instancing extension tomorrow

issacclee:
[instanceExample1.zip](https://github.com/playcanvas/engine/files/7257676/instanceExample1.zip)
Hi @mvaligursky ,
Please find the attached zip file as example files with EXT_mesh_gpu_instancing extensions.
There are three glb/gltf files in zip:
1. before_instance.glb: original glb with one mesh referenced by 6 nodes.
2.  testinstance.gltf: generated based on before_instance.glb with EXT_mesh_gpu_instancing extensions
3. testinstance.glb: glb version of testinstance.gltf

Thanks

issacclee:
![image](https://user-images.githubusercontent.com/687746/135411579-3e8c704c-22ff-470d-97af-7504d5fdd277.png)
BTW, this is what it looks like when I put testinstance.glb into the babylon playground viewer. Babylon playground supports gltf file with  EXT_mesh_gpu_instancing extension.

mvaligursky:
closing at the support has been added: https://github.com/playcanvas/engine/pull/6869

PR #6869, opened 2024-08-01
Support for EXT_mesh_gpu_instancing extension

mvaligursky:
Fixes #https://github.com/playcanvas/engine/issues/3526

Example:
image


src/scene/mesh-instance.js
```diff
@@ -58,12 +58,26 @@ class InstancingData {
     /** @type {VertexBuffer|null} */
     vertexBuffer = null;
 
+    /**
+     * True if the vertex buffer is destroyed when the mesh instance is destroyed.
+     *
+     * @type {boolean}
```
willeastcott, line 64: Should this be `@private`?

willeastcott: Ah, you reference it in `glb-container-resource.js`. So maybe `@ignore` instead?

willeastcott (approved):

src/scene/mesh-instance.js
mvaligursky: we removed all `@ignore` tags from non-exported classes.

willeastcott: Oh, this is `InstancingData`. I thought it was `MeshInstance`. OK, forget it.

PR #6872, opened 2024-08-02
Remove float rendering precision test for old WebGL devices

mvaligursky:
This is relevant for GPU chips introduced 12 years ago - like PowerVR SGX543 (iPhone 4s), Intel HD Graphics 4000, Adreno 305.
This precision test was only used for empirical exponent constants for VSM32 shadows to slightly improve it's quality.

I think it's time to remove it for engine v2.

Also remove a WebGL1 path in one function.

Suggestion: diff it with white spaces differences hidden


slimbuck (approved):
Nice!

PR #6877, opened 2024-08-06
Updated ESLint config

kpal81xd:
Updates ESlint config and applies all quick fixes


willeastcott (approved):
I've checked maybe half of this PR. Looks good. I'll assume the rest is OK too! 😄

PR #6881, opened 2024-08-14
fix touch event when target - is not an element

Maksims:
We've noticed in many games published on the web through Analytics Error Reporting that there is a case when touchmove event target - is not an element. It could be a `window`.
Which leads to an exception.

This PR fixes that and also re-uses the existing method within element-input.js as they are identical.

### DIFF
--- a/src/platform/input/touch-event.js
+++ b/src/platform/input/touch-event.js
@@ -11,16 +11,20 @@ function getTouchTargetCoords(touch) {
     let totalOffsetX = 0;
     let totalOffsetY = 0;
     let target = touch.target;
-    while (!(target instanceof HTMLElement)) {
+
+    while (!(target instanceof HTMLElement) && target) {
         target = target.parentNode;
     }
-    let currentElement = target;
 
-    do {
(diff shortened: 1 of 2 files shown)


src/platform/input/touch-event.js
```diff
@@ -11,16 +11,20 @@ function getTouchTargetCoords(touch) {
...
-    } while (currentElement);
+    if (target) {
+        let currentElement = target;
+
+        do {
+            totalOffsetX += currentElement.offsetLeft - currentElement.scrollLeft;
+            totalOffsetY += currentElement.offsetTop - currentElement.scrollTop;
+            currentElement = currentElement.offsetParent;
+        } while (currentElement);
+    }
```
slimbuck, line 27:
```suggestion
    while (target) {
        totalOffsetX += target.offsetLeft - target.scrollLeft;
        totalOffsetY += target.offsetTop - target.scrollTop;
        target = target.offsetParent;
    }
```

slimbuck (approved):
Nice, approving with code suggestion.

willeastcott (approved):

willeastcott:
Thank you, @Maksims! 🤗

PR #6882, opened 2024-08-16
Improve ply parsing performance

kemchenj:
This PR improves the PLY parsing speed by about 63% for the guitar scene. **The parsing time goes from 21.02 ms to 7.84 ms** (tested on an M1 MacBook Air). And the benchmark script can be found [here](https://github.com/kemchenj/engine/blob/cd90840bc763195aab98d91b6bb328b8764da4c2/ply-benchmark.js).

This optimization reduces string comparison overhead during ply parsing. Instead of comparing strings for each property of each splat, I try to do the comparison at once and made a temporary `read` function for each property. This approach reduces the number of string comparisons, resulting in faster parsing.

### DIFF
--- a/src/framework/parsers/ply.js
+++ b/src/framework/parsers/ply.js
@@ -418,6 +418,25 @@ const readPly = async (reader, propertyFilter = null) => {
 
         // calculate the size of an input element record
         const inputSize = element.properties.reduce((a, p) => a + p.byteSize, 0);
+        const propertyParsingFunctions = element.properties.map((p) => {
+            /* eslint-disable brace-style */
+            if (p.storage) {
+                switch (p.type) {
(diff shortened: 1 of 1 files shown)


src/framework/parsers/ply.js
```diff
@@ -404,10 +404,29 @@ const readPly = async (reader, propertyFilter = null) => {
     // allocate element storage
     elements.forEach((e) => {
         e.properties.forEach((p) => {
+            p.read = (streamBuf) => {
```
slimbuck, line 407: Please store the read functions in a separate array (which can be referenced by property index) instead of assigning it temporarily to the property instance.

slimbuck (approved):
Now this is a thing of beauty ❤️ - thank you!

willeastcott (approved):

PR #6888, opened 2024-08-21
[BREAKING] Remove support for Layer.renderTarget and some callbacks on the Layer

mvaligursky:
- removed Layer.renderTarget support, which has been deprecated for few years now. renderTarget can be set on the CameraComponent only now.
- this made multiple callbacks on the Layer no longer relevant, and those were removed as well: 
	- onPreRender
	- onPreRenderOpaque
	- onPreRenderTransparent
	- onPostRender
	- onPostRenderOpaque
	- onPostRenderTransparent
- two callbacks were moved from the Layer to a more appropriate place at the CameraComponent:
	- onPreCull
	- onPostCull
- Layer.layerReference has been removed, this have been unused for some time
- Layer.onDrawCall callback has been removed. This was a callback called during rendering when all render state has been setup, allowing the user to apply some overrides. With the move to render pipelines and uniform buffers, only a very limited subset of state could be change here, making it pointless. This will also further limited when we implement parallel render pipeline compilation, as even those few remining bits that can be changed currently will no longer be applicable.

Callbacks available now, all on the CameraComponent:
- onPreCull
- onPostCull
- onPreRender
- onPostRender

There will likely be additional callbacks added for the render pass architecture, but that a separate work in the future.

### DIFF
--- a/src/scene/composition/layer-composition.js
+++ b/src/scene/composition/layer-composition.js
@@ -77,7 +77,7 @@ class LayerComposition extends EventHandler {
     subLayerEnabled = []; // more granular control on top of layer.enabled (ANDed)
 
     /**
-     * An array of {@link CameraComponent}
+     * An array of {@link CameraComponent}s.
      *
      * @type {CameraComponent[]}
      * @ignore
--- a/src/scene/composition/layer-composition.js
+++ b/src/scene/composition/layer-composition.js
@@ -77,21 +77,12 @@ class LayerComposition extends EventHandler {
     subLayerEnabled = []; // more granular control on top of layer.enabled (ANDed)
 
     /**
-     * A read-only array of {@link CameraComponent} that can be used during rendering. e.g.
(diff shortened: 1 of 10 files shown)


src/scene/composition/layer-composition.js
```diff
@@ -77,21 +77,12 @@ class LayerComposition extends EventHandler {
     subLayerEnabled = []; // more granular control on top of layer.enabled (ANDed)
 
     /**
-     * A read-only array of {@link CameraComponent} that can be used during rendering. e.g.
-     * Inside {@link Layer#onPreCull}, {@link Layer#onPostCull}, {@link Layer#onPreRender},
-     * {@link Layer#onPostRender}.
+     * An array of {@link CameraComponent}
```
willeastcott, line 80:
```suggestion
     * An array of {@link CameraComponent}.
```

willeastcott (approved):
This is great - lovely to see this cleaned up. The future is bright!

PR #6892, opened 2024-08-23
Added few keywords to the package

mvaligursky:


package.json
```diff
@@ -8,8 +8,11 @@
     "3D",
     "2D",
     "VR",
+    "AR",
     "WebGL",
     "WebGL2",
+    "WebGPU",
+    "WebXR",
```
willeastcott, line 15: Can we standardize on lower case keywords?

willeastcott: And gltf

willeastcott (approved):

package.json
willeastcott: I've made some updates myself.

PR #6896, opened 2024-08-28
Fix 3D Gaussian Splat rendering in VR

### LINKED ISSUE #6053: GSplats stereoscopic rendering (viewport size issues)
Currently GSplats has to use [viewport dimensions](https://github.com/playcanvas/engine/blob/7f9cb585a83ee6f6986e94cb2e3da4c95f18c27b/src/scene/gsplat/gsplat-instance.js#L138-L144) which affects the way they are rotated and scaled. Aspect ratio of this viewport is the main driving element. In WebXR stereoscopic scenario, viewport is split into two so the the renderer of the splats look wrong.

The quick workaround I've made is to detect if there are multiple XR views, and split viewport width in two:

```javascript
const viewport = [0, 0];

pc.GSplatInstance.prototype.updateViewport = function () {
    const device = this.splat.device;

    viewport[0] = device.width;
    viewport[1] = device.height;

    if (app.xr.active && app.xr.views.list.length === 2)
        viewport[0] /= 2;

    this.material.setParameter('viewport', viewport);
};
```

But it needs a slightly better solution, for getting an actual viewport of a view.

willeastcott:
Fixes #6053 

Without fix:

image

With fix:

image

Thank you to @Maksims for the suggestion for fixing this. 🙏 

### DIFF
--- a/src/scene/gsplat/gsplat-instance.js
+++ b/src/scene/gsplat/gsplat-instance.js
@@ -174,6 +174,16 @@ class GSplatInstance {
         const device = this.splat.device;
         viewport[0] = device.width;
         viewport[1] = device.height;
+
+        // adjust viewport for stereoscopic VR sessions
+        if (this.cameras.length > 0) {
+            const camera = this.cameras[0];
+            const xr = camera.xr;
+            if (xr && xr.active && xr.views.list.length === 2) {
+                viewport[0] /= 2;
+            }
+        }
+
         this.material.setParameter('viewport', viewport);
     }
 


slimbuck (approved):

src/scene/gsplat/gsplat-instance.js
```diff
@@ -174,6 +174,16 @@ class GSplatInstance {
         const device = this.splat.device;
         viewport[0] = device.width;
         viewport[1] = device.height;
+
+        // adjust viewport for stereoscopic VR sessions
+        if (this.cameras.length > 0) {
+            const camera = this.cameras[0];
```
mvaligursky, line 180: we should probably loop over all cameras to see if any of these is xr. I suspect that if the shadows are used, the cameras[0] would be a shadow camera.

Even better, this should be called per camera somehow, instead of globally once per frame - in case the splat is rendered to a texture as well as main framebuffer.

willeastcott: Yeah, it's not great and almost certainly needs to be reworked.

MAG-AdrianMeredith:
Nice, Does this also add support for multiple cameras generally or is it just a VR fix.  Right now it does the culling and ordering for only the last camera, its acutally quite a cool effect but not one i think is intentional ;)

mvaligursky:
It does not handle it, all that still works just for a single camera. That is actually similar to particle system (when CPU sorting is used)

PR #6898, opened 2024-08-29
OutlineRenderer and an engine example

mvaligursky:
- Outline rendering tech extracted from the Editor and cleaned up and modernised, delivered as a module in `/extras`
- Engine example showing its use
- cleaned up `MeshInstance.setParameter` to avoid obsolete and undocumented functionality, and also to handle refactoring to shader passes, where the passes no longer have fixed numerical IDs.

## New public API

A replacement for removed callbacks on the Layer class.

```
CameraComponent.onPreRenderLayer(Layer, transparent)
CameraComponent.onPostRenderLayer(Layer, transparent)
```

Screenshot 2024-08-29 at 13 25 50

### DIFF
--- a/src/scene/materials/standard-material-options.js
+++ b/src/scene/materials/standard-material-options.js
@@ -8,6 +8,13 @@ import { LitShaderOptions } from '../shader-lib/programs/lit-shader-options.js';
  */
 class StandardMaterialOptions {
     /**
+     * The set of defines used to generate the shader.
+     *
+     * @type {Map<string, string>}
+     */
+    defines = new Map();
+
+    /**
      * If UV1 (second set of texture coordinates) is required in the shader. Will be declared as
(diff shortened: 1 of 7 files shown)


src/scene/materials/standard-material-options.js
```diff
@@ -7,6 +7,13 @@ import { LitShaderOptions } from '../shader-lib/programs/lit-shader-options.js';
  * @category Graphics
  */
 class StandardMaterialOptions {
+    /**
+     * The set of defines used to generate the shader.
+     *
+     * @type {Map<string, string>}
+     */
+    defines = new Map();
```
slimbuck, line 15: this meant to be here?

slimbuck (approved):
Approving, but example needs fixing

PR #6899, opened 2024-08-30
Allow per camera fog settings

mvaligursky:
- before, a fog parameters were a scene settings that all cameras were using
- these have now been moved to scene.rendering which can be overwritten on a camera, allowing per camera control
- this is a non-breaking change as existing Scene.fog settings is only deprecated

Old API
```
Scene.fog
Scece.fogColor
Scene.fogStart
Scene.fogEnd
Scene.fogDensity
```

New API
```
Scene.rendering.fog
Scece.rendering.fogColor
Scene.rendering.fogStart
Scene.rendering.fogEnd
Scene.rendering.fogDensity
```

### DIFF
--- a/examples/src/examples/graphics/shader-hatch.example.mjs
+++ b/examples/src/examples/graphics/shader-hatch.example.mjs
@@ -208,7 +208,7 @@ assetListLoader.load(() => {
         }
         if (propertyName === 'fog') {
             // turn on/off fog and set up its properties
-            app.scene.fog = value ? pc.FOG_LINEAR : pc.FOG_NONE;
+            app.scene.rendering.fog = value ? pc.FOG_LINEAR : pc.FOG_NONE;
             app.scene.fogColor = new pc.Color(0.8, 0.8, 0.8);
             app.scene.fogStart = 100;
             app.scene.fogEnd = 300;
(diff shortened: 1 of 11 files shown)


willeastcott:
I have an alternative proposal:

* Create a new class called `Fog`. Similar to `Sky`. This holds all the fog-related properties.
* Add instance of `Fog` as `Scene#fog` (and alias all fog properties on `Scene` to properties on `Fog`).
* Add `CameraComponent#fog`. If `null`, then use `Scene` fog.

Note that I don't think we should deprecate or hide scene-based fog.

mvaligursky:
The rendering params class I added few months ago is a class designed to contain the properties that are used during the shader generation. Is has gamma settings / tonemapping as well, and those settings are passed to the shader generator, and their hash is used at runtime to look up a shader needed. I would prefer to not create another class for fog properties, as I would need to duplicate this functionality.

We could create rendering.fog subclass, but I think keeping those all in RenderingParams is probably ok

mvaligursky:
But what I can do instead is to move all other fog parameters to be part of RenderingParams.

mvaligursky:
All fog params are now part of the RenderingParams

PR #6900, opened 2024-08-30
Propsal to update deprecated and removed functionality

mvaligursky:
This is a proposed change, and if this gets approved, this change will be applied to most of deprecated.js.

- add `Debug.removed` for functionality that has been completely removed, vs just deprecated (which still works as before)
- move getters / setters and other bits from deprecated.js to the appropriate module:
    - as Object.defineProperty was used in the deprecated.js, this was a module with side-effects and its imports cannot be part of tree-shaking, making tree-shaking less efficient (according to @marklundin )
    - the syntax of setter and getter in a class is nicer than Object.defineProperty syntax

### DIFF
--- a/src/core/debug.js
+++ b/src/core/debug.js
@@ -26,6 +26,18 @@ class Debug {
     }
 
     /**
+     * Removed warning message.
+     *
+     * @param {string} message - The message to log.
+     */
+    static removed(message) {
+        if (!Debug._loggedMessages.has(message)) {
+            Debug._loggedMessages.add(message);
+            console.warn(`REMOVED: ${message}`);
+        }
+    }
+
+    /**
      * Assertion deprecated message. If the assertion is false, the deprecated message is written to the log.
      *
      * @param {boolean|object} assertion - The assertion to check.
--- a/src/scene/batching/batch.js
+++ b/src/scene/batching/batch.js
@@ -1,3 +1,4 @@
+import { Debug } from '../../core/debug.js';
 import { BoundingBox } from '../../core/shape/bounding-box.js';
 
 /**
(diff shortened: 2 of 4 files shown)


marklundin:
I think the `@deprecated` flag will be better in terms of api documentation. Do we really need the `removed` though? I understand it may help migration, but I kinda feel it should be fully removed

mvaligursky:
It's there exactly to help with the migration. If we simply remove functions, for example setter, the user code still works and the user has no idea we removed it. This is very hard to find.

src/scene/batching/batch.js
```diff
@@ -92,6 +93,16 @@ class Batch {
         this.meshInstance.aabb = this._aabb;
         this.meshInstance._aabbVer = 0;
     }
+
+    /**
+     * @deprecated
+     * @private
+     * @returns {undefined}
```
willeastcott, line 100: Getters/setters use `@type`, not `@returns`. See other examples.

src/scene/batching/batch.js
```diff
@@ -92,6 +93,16 @@ class Batch {
         this.meshInstance.aabb = this._aabb;
         this.meshInstance._aabbVer = 0;
     }
+
+    /**
+     * @deprecated
+     * @private
```
willeastcott, line 99: This getter isn't private. It public, but presumably ignored from the documentation? So I suspect this should be `@ignore` instead.

mvaligursky: updated, good points

willeastcott (approved):

src/core/debug.js
```diff
@@ -25,6 +25,18 @@ class Debug {
         }
     }
 
+    /**
+     * Removed warning message.
+     *
+     * @param {string} message - The message to log.
```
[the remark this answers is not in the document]
mvaligursky: the Debug class is not exported by the index, all this is private.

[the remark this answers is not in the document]
willeastcott: LOL, my mistake. 👍

PR #6902, opened 2024-09-02
[Examples/Vehicle] Fix Ammo null function error on restart

### LINKED ISSUE #6724: Engine-only example crash
Originally reported in #6723 

Needs an investigation, if the error is caused by something else.

Repro:

- open Vehicle example
- click the code refresh button

image

kungfooman:
Fixes #6724 

These are just my current findings, the engine still seems to leak rigidbody Ammo objects/pointers... but at least this makes an Example refresh work in the first place. This allows for further fixes without just crashing like it currently does here: https://playcanvas.vercel.app/#/physics/vehicle

About https://github.com/playcanvas/engine/commit/7b0c745cbe480e8dde1d4e82e62fc278562cc30c / Removing `Ammo.destroy(origin);`:

The origin is part of `btTransform`:

https://github.com/kripken/ammo.js/blob/1ed8b58c7058a5f697f2642ceef8ee20fdd55e10/bullet/src/LinearMath/btTransform.h#L32-L41

You don't just get that pointer and call `free` on it - at least it doesn't make sense to me. Probably at best nothing happens and worst case is messing up internal allocation structures while causing other subtle bugs. Any other opinions on this?

Feedback about all this highly welcome and some testing.

I can't even logically explain why the `null function error` occurs, since why should it matter if we reuse some old allocated pointers or free them and recreate those anew - it feels like a black box. :see_no_evil:

Vercel quick link: https://engine-r0b3y912i-playcanvas.vercel.app/#/physics/vehicle

### DIFF
--- a/examples/src/examples/physics/vehicle.example.mjs
+++ b/examples/src/examples/physics/vehicle.example.mjs
@@ -89,17 +89,12 @@ assetListLoader.load(() => {
     app.root.addChild(ground);
 
     // Create 4 wheels for our vehicle
-    /**
-     * @todo use .map ...
-     * @type {pc.Entity[]}
-     */
-    const wheels = [];
-    [
+    const wheels = [
         { name: 'Front Left Wheel', pos: new pc.Vec3(0.8, 0.4, 1.2), front: true },
         { name: 'Front Right Wheel', pos: new pc.Vec3(-0.8, 0.4, 1.2), front: true },
         { name: 'Back Left Wheel', pos: new pc.Vec3(0.8, 0.4, -1.2), front: false },
         { name: 'Back Right Wheel', pos: new pc.Vec3(-0.8, 0.4, -1.2), front: false }
-    ].forEach(function (wheelDef) {
+    ].map(function (wheelDef) {
         // Create a wheel
         const wheel = new pc.Entity(wheelDef.name);
         wheel.addComponent('script');
@@ -110,7 +105,7 @@ assetListLoader.load(() => {
             }
         });
         wheel.setLocalPosition(wheelDef.pos);
-        wheels.push(wheel);
+        return wheel;
     });
 
     // Create a physical vehicle
--- a/examples/src/examples/physics/vehicle.example.mjs
+++ b/examples/src/examples/physics/vehicle.example.mjs
@@ -94,7 +94,7 @@ assetListLoader.load(() => {
         { name: 'Front Right Wheel', pos: new pc.Vec3(-0.8, 0.4, 1.2), front: true },
         { name: 'Back Left Wheel', pos: new pc.Vec3(0.8, 0.4, -1.2), front: false },
         { name: 'Back Right Wheel', pos: new pc.Vec3(-0.8, 0.4, -1.2), front: false }
-    ].map(function (wheelDef) {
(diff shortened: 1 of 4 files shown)


kungfooman:
I wanted to keep track of every object and then it turned out that it's not even possible to destroy `btVehicleTuning` because of:

https://github.com/kripken/ammo.js/blob/1ed8b58c7058a5f697f2642ceef8ee20fdd55e10/ammo.idl#L936-L945

So I just deleted the `NoDelete` keyword, rebuild `Ammo.js` via `cmake`/`emcc` and `Ammo.destroy`'ed the vehicle tuning class aswell just like the others - turns out I can refresh the example more than 100x without *one* strange occurance of twisted or detached tires.

examples/src/examples/physics/vehicle.example.mjs
```diff
@@ -89,17 +89,12 @@ assetListLoader.load(() => {
...
-     */
-    const wheels = [];
-    [
+    const wheels = [
         { name: 'Front Left Wheel', pos: new pc.Vec3(0.8, 0.4, 1.2), front: true },
         { name: 'Front Right Wheel', pos: new pc.Vec3(-0.8, 0.4, 1.2), front: true },
         { name: 'Back Left Wheel', pos: new pc.Vec3(0.8, 0.4, -1.2), front: false },
         { name: 'Back Right Wheel', pos: new pc.Vec3(-0.8, 0.4, -1.2), front: false }
-    ].forEach(function (wheelDef) {
+    ].map(function (wheelDef) {
```
willeastcott, line 97:
```suggestion
    ].map((wheelDef) => {
```

willeastcott (approved):

LeXXik:
> So I just deleted the `NoDelete` keyword ... and `Ammo.destroy`'ed the vehicle tuning class ...

Hmm, I might be not correct here, but I don't think Emscripten actually destroys it. WebIDL "nodelete" keyword means that this class cannot be deleted, since it has no public destructor (e.g. it is private). I think ems simply suppresses the error, when you try to destroy it. Similar to how you won't get an error, if you try to destroy an object a second time:

```js
const vec = new Ammo.btVector3(0, 0, 0);
Ammo.destroy(vec);
Ammo.destroy(vec); // no error
```

kungfooman:
@LeXXik Yea, that description about no-destructor I read aswell and it makes absolutely no sense to me. You don't need a destructor to free memory, it just feels like an arbitrary and even harmful limitation. Historically classes were basically "structs with functions" and in order to have a struct, you have to allocate it's memory... and then obviously free that again for good housekeeping - which that keyword makes impossible.

We can discuss this out further in a new issue/PR, in this one I'm just happy to have the actual bug fixed.

willeastcott:
> We can discuss this out further in a new issue/PR, in this one I'm just happy to have the actual bug fixed.

So folks are happy for me to go ahead and merge this now?

LeXXik:
> So folks are happy for me to go ahead and merge this now?

Yes, this is a good PR, regardless of the bug.

LeXXik:
@willeastcott suggesting for v1

willeastcott:
@LeXXik Done. I've just cherry picked it to `main_v1`.

PR #6907, opened 2024-09-03
Option to apply SSAO during lighting instead of during postprocessing

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/4844
Try it here: https://engine-hx2mdhbb7-playcanvas.vercel.app/#/graphics/ambient-occlusion

Until now, the SSAO was applied as a post-processing effect. The problem is that the ambient occlusion then incorrectly darkens even the areas hit by lights.

For the render pass architecture, an option to apply AO during lighting in forward passes has been added.

Before - notice how uniformly the SSAO looks
Screenshot 2024-09-03 at 15 23 34

Now - almost no AO is visible in the lit areas, but ambient areas receive AO as before
Screenshot 2024-09-03 at 15 23 48

Detail before
Screenshot 2024-09-03 at 15 32 37

and now
Screenshot 2024-09-03 at 15 32 53

Detail before
Screenshot 2024-09-03 at 15 34 53

and now
Screenshot 2024-09-03 at 15 35 04

### DIFF
--- a/examples/src/examples/graphics/ambient-occlusion.controls.mjs
+++ b/examples/src/examples/graphics/ambient-occlusion.controls.mjs
@@ -10,11 +10,16 @@ export function controls({ observer, ReactPCUI, React, jsx, fragment }) {
             { headerText: 'Ambient Occlusion' },
             jsx(
                 LabelGroup,
-                { text: 'enabled' },
-                jsx(BooleanInput, {
-                    type: 'toggle',
+                { text: 'Type' },
+                jsx(SelectInput, {
                     binding: new BindingTwoWay(),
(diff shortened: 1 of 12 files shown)


MAG-AdrianMeredith:
Looks fantastic, there was always something "off" about the ssao that irked me and this is a huge improvement.
I've noticed a bug though if you apply scale its no longer aligned with the scene 
image

mvaligursky:
> this is a huge improvement

Totally agree, the impact is small but large at the same time.

> I've noticed a bug though if you apply scale its no longer aligned with the scene

And I know exactly why .. will fix, thanks!

PR #6909, opened 2024-09-04
WebXR: fix depth sensing for Oculus Quest v68+

Maksims:
This PR fixes Depth Sensing for Meta Quest and all WebXR AR sessions in v2 engine.

With a recent update of specs and introduction of a new texture format (R16UI), Quest v68+ started using that format. Also there was issues introduced with two PRs: #6773 (was not a full cleanup) and #6831 (defines should be used instead of code concatenation).

### DIFF
--- a/examples/src/examples/xr/ar-camera-depth.example.mjs
+++ b/examples/src/examples/xr/ar-camera-depth.example.mjs
@@ -116,7 +116,6 @@ const updateShader = (array, float) => {
     shaderDepthFloat = float;
 
     const key = 'textureDepthSensing_' + array + float;
-    let frag = fragShader;
 
     if (shaderDepthArray) materialDepth.setDefine('XRDEPTH_ARRAY', true);
     if (shaderDepthFloat) materialDepth.setDefine('XRDEPTH_FLOAT', true);
@@ -124,7 +123,7 @@ const updateShader = (array, float) => {
     materialDepth.shaderDesc = {
         uniqueName: key,
         vertexCode: vertShader,
-        fragmentCode: frag,
+        fragmentCode: fragShader,
         attributes: {
             aPosition: pc.SEMANTIC_POSITION,
             aUv0: pc.SEMANTIC_TEXCOORD0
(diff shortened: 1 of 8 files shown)


src/framework/xr/constants.js
```diff
@@ -213,7 +213,15 @@ export const XRDEPTHSENSINGUSAGE_GPU = 'gpu-optimized';
 export const XRDEPTHSENSINGFORMAT_L8A8 = 'luminance-alpha';
 
 /**
- * Float 32 - indicates that depth sensing preferred raw data format is Float 32.
+ * Unsigned Short - indicates that depth sensing preferred raw data format as Unsigned Short 16 bit.
```
willeastcott, line 216:
```suggestion
 * Unsigned Short - indicates that depth sensing preferred raw data format is Unsigned Short (16 bit).
```

src/framework/xr/constants.js
```diff
@@ -213,7 +213,15 @@ export const XRDEPTHSENSINGUSAGE_GPU = 'gpu-optimized';
...
+ *
+ * @type {string}
+ * @category XR
+ */
+export const XRDEPTHSENSINGFORMAT_R16U = 'unsigned-short';
+
+/**
+ * Float 32 - indicates that depth sensing preferred raw data format is Float 32 bit.
```
willeastcott, line 224:
```suggestion
 * Float 32 - indicates that depth sensing preferred raw data format is Float (32 bit).
```

mvaligursky:
So this is for engine v2. I wonder if this fix would work for engine v1 and should be merged there too?

willeastcott (approved):

mvaligursky (approved):

examples/src/examples/xr/ar-camera-depth.example.mjs
```diff
@@ -126,9 +118,8 @@ const updateShader = (array, float) => {
     const key = 'textureDepthSensing_' + array + float;
     let frag = fragShader;
```
willeastcott, line 119: Delete this line...

willeastcott: ...and set `fragShader` in `materialDepth.shaderDesc` instead.

Maksims:
Actually, this needs to be cherry-picked to v1 too. As currently it crashes without this patch.

mvaligursky:
Unfortunately it cannot be simply cherry-picked as it uses some v2 API, for example
```
materialDepth.setDefine('XRDEPTH_ARRAY', true);
```

Issue #5714, opened 2023-10-04
WebGPU MSAA render target with depth buffer fails


slimbuck:
It is not possible to create a multisampled render target and specify a target depth texture roughly like so:

```javascript
const colorBuffer = createTexture(widthPixels, heightPixels, PIXELFORMAT_RGBA8);
const depthBuffer = createTexture(widthPixels, heightPixels, PIXELFORMAT_DEPTH);
const renderTarget = new RenderTarget({
    name: 'rt',
    colorBuffer: colorBuffer,
    depthBuffer: depthBuffer,
    flipY: false,
    samples: device.maxSamples,
    autoResolve: false
});
```

The resulting error is:
![Screenshot 2023-10-04 at 17 30 00](https://github.com/playcanvas/engine/assets/11276292/b8d97747-e3cc-4c3d-a554-8648a8c5edf4)

mvaligursky:
Yep, this is not supported at the moment. What needs to happen for this to work:
- in WebgpuRenderTarget, when samples > 1 and a depthBuffer is passed in, we need to allocate multi-sampled depth buffer for rendering. We cannot specify the provided single-sampled buffer as a resolve target, as WebGPU cannot automatically resolve depth. 
- then in the endPass for WebGPU, just before `passEncoder.end` (so using existing pass encoder) use WebgpuResolver to manually resolve the depth, which it already handles.
- we possibly might need to update framegraph for the render pass in question to store the multisampled depth, as by default I think it just gets discarded.

PR #6917, opened 2024-09-06
WebGPU partial support for multisampled RT using supplied depth texture

mvaligursky:
- implements support for WebGPU rendering to a supplied depth texture when MSAA is enabled. This internally allocates additional MSAA depth buffer for rendering.
- refactored to store depth related parts of WegpuRenderTarget in a DepthAttachment private class
- solves first point here: https://github.com/playcanvas/engine/issues/5714#issuecomment-1750424649
- other points to be done separately

### DIFF
--- a/src/platform/graphics/webgpu/webgpu-render-target.js
+++ b/src/platform/graphics/webgpu/webgpu-render-target.js
@@ -12,10 +12,13 @@ const stringIds = new StringIds();
 
 /**
  * Private class storing info about color buffer.
+ *
+ * @private
  */
 class ColorAttachment {
     /**
      * @type {GPUTextureFormat}
+     * @private
      */
     format;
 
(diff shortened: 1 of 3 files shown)


src/platform/graphics/webgpu/webgpu-render-target.js
```diff
@@ -12,6 +12,8 @@ const stringIds = new StringIds();
 
 /**
  * Private class storing info about color buffer.
+ *
+ * @private
```
willeastcott, line 16: I'm guessing this is unnecessary because the root `index.js` does not export this class.

willeastcott (approved):

src/platform/graphics/webgpu/webgpu-render-target.js
mvaligursky: yeah but I had to add it to avoid the type errors coming from WebGPU types

PR #6919, opened 2024-09-06
[Fix] Do not try to resolve color buffer if the RT does not have any, WebGL

mvaligursky:
- small fix to avoid resolve blit call (that seems to be ingored anyways)

### DIFF
--- a/src/platform/graphics/render-pass.js
+++ b/src/platform/graphics/render-pass.js
@@ -276,7 +276,7 @@ class RenderPass {
         }
 
         // color
-        const numColorOps = rt?._colorBuffers?.length ?? 1;
+        const numColorOps = rt ? rt._colorBuffers?.length : 1;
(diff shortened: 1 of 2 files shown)


willeastcott (approved):

src/platform/graphics/render-pass.js
```diff
@@ -276,7 +276,7 @@ class RenderPass {
         }
 
         // color
-        const numColorOps = rt?._colorBuffers?.length ?? 1;
+        const numColorOps = rt ? rt._colorBuffers?.length : 1;
```
willeastcott, line 279: Wait....is this right? If `rt` is defined but `rt._colorBuffers` is not, `numColorOps` will be set to `undefined` right?

mvaligursky: you're right, fixed, thanks

PR #6926, opened 2024-09-08
Fix trigger crash

### LINKED ISSUE #2572: Cloning an trigger and enabling/disabling the clone many times in the same frame causes Ammo crash
Repo: https://playcanvas.com/editor/scene/1037430
Press 1, then 2, then 1 again. It should crash then.

I can't explain this one, there's some combination of disabling on initialisation and enabling/disabling on clone causes some odd behaviour.

Moving 
```
DisableMeOnInit.prototype.initialize = function() {
    this.entity.enabled = false;
};
``` 
To postInitialize fixes the issue:
```
DisableMeOnInit.prototype.postInitialize = function() {
    this.entity.enabled = false;
};
```

LeXXik:
Fixes #2572 

Prevents adding a trigger to the dynamics world, if it is already there.

### DIFF
--- a/src/framework/components/collision/trigger.js
+++ b/src/framework/components/collision/trigger.js
@@ -119,9 +119,13 @@ class Trigger {
         const body = this.body;
         if (!body) return;
 
-        const systems = this.app.systems;
-        systems.rigidbody.addBody(body, BODYGROUP_TRIGGER, BODYMASK_NOT_STATIC ^ BODYGROUP_TRIGGER);
-        systems.rigidbody._triggers.push(this);
+        const system = this.app.systems.rigidbody;
+        const exists = system._triggers.find(trigger => trigger === this);
+
+        if (!exists) {
+            system.addBody(body, BODYGROUP_TRIGGER, BODYMASK_NOT_STATIC ^ BODYGROUP_TRIGGER);
+            system._triggers.push(this);
+        }
 
         // set the body's activation state to active so that it is
         // simulated properly again


src/framework/components/collision/trigger.js
```diff
@@ -119,9 +119,13 @@ class Trigger {
         const body = this.body;
         if (!body) return;
 
-        const systems = this.app.systems;
-        systems.rigidbody.addBody(body, BODYGROUP_TRIGGER, BODYMASK_NOT_STATIC ^ BODYGROUP_TRIGGER);
-        systems.rigidbody._triggers.push(this);
+        const system = this.app.systems.rigidbody;
+        const exists = system._triggers.find(trigger => trigger === this);
```
willeastcott, line 123:
```suggestion
        const exists = system._triggers.indexOf(trigger) !== -1;
```

src/framework/components/collision/trigger.js
```diff
@@ -119,9 +119,13 @@ class Trigger {
         const body = this.body;
         if (!body) return;
 
-        const systems = this.app.systems;
-        systems.rigidbody.addBody(body, BODYGROUP_TRIGGER, BODYMASK_NOT_STATIC ^ BODYGROUP_TRIGGER);
-        systems.rigidbody._triggers.push(this);
+        const system = this.app.systems.rigidbody;
+        const exists = system._triggers.indexOf(trigger) !== -1;
```
willeastcott, line 123:
```suggestion
        const exists = system._triggers.indexOf(this) !== -1;
```

willeastcott (approved):

LeXXik:
For some reason, when I originally was writing it, indexOf was returning me -1, so I went with `find`. Probably was debug console cache issue or something.

willeastcott:
@LeXXik I've tweaked the code to make `enable` and `disable` more symmetrical. Let me know what you think.

LeXXik:
@willeastcott looks good. Even though removing a body from world if it is not there should be "safe", there is no harm anyhow.

PR #6940, opened 2024-09-10
RenderPass based SSAO works with multisampling

mvaligursky:
When render pass based ssao is enabled, and the samples > 1:
- share multisampled depth buffer with main scene, similar to how it was done on WebGPU in https://github.com/playcanvas/engine/pull/6932
- when prepass rendering is done, resolve the depth buffer, as that resolved depth is used by SSAO

Other changes
- skip few load ops on render targets and do clear when we don't need original data
- extracted multisampled texture cache to a separate module, as that is shared between webgl and webgpu

### DIFF
--- a/src/platform/graphics/multi-sampled-texture-cache.js
+++ b/src/platform/graphics/multi-sampled-texture-cache.js
@@ -4,7 +4,7 @@ import { DeviceCache } from './device-cache.js';
 /**
  * Reference counted cache storing multi-sampled versions of depth buffers, which are reference
  * counted and shared between render targets using the same user-specified depth-buffer. This is
- * needed for in cases where the user provided depth buffer is used for depth-pre-pass and then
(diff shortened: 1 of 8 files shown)


src/platform/graphics/multi-sampled-texture-cache.js
```diff
@@ -0,0 +1,25 @@
+import { RefCountedKeyCache } from '../../core/ref-counted-key-cache.js';
+import { DeviceCache } from './device-cache.js';
+
+/**
+ * Reference counted cache storing multi-sampled versions of depth buffers, which are reference
+ * counted and shared between render targets using the same user-specified depth-buffer. This is
+ * needed for in cases where the user provided depth buffer is used for depth-pre-pass and then
```
willeastcott, line 7:
```suggestion
 * needed for the cases where the user provided depth buffer is used for depth-pre-pass and then
```

PR #6941, opened 2024-09-10
Revert "Enable loading of texture assets in unit tests"

marklundin:
Fixes https://github.com/playcanvas/engine/issues/6942
Reverts playcanvas/engine#6928

The 'node-canvas' doesn't [compile on Apple Silicon](https://github.com/Automattic/node-canvas/wiki/Installation%3A-Mac-OS-X), so `npm install` currently fails on OS X - Silicon. The PR reverts back to using the playcanvas/canvas-mock, until we find a better solution


kpal81xd (approved):

willeastcott:
Can't you just brew install those dependencies for now?

marklundin:
We could do, but I guesswe'd need to update the README.md for MacOS to say something like

Ensure you have [Brew](https://brew.sh/) installed 

```sh
brew install pkg-config cairo pango libpng jpeg giflib librsvg
npm install
```

Not sure, seems a bit overkill for just the tests.

 Haven't tested but Skia seems to work on Silicon https://www.npmjs.com/package/skia-canvas

kungfooman:
Looks like this `cairo.h` error exists since 2021: https://github.com/Automattic/node-canvas/issues/1773

Feels strange if we first need to build all kinds of fancy and often unused C++/Python hell deps in order to generate an engine build artifact - You can't use the highway in Australia because some side road in San Francisco is blocked! :sweat_smile: 

I guess we could add a new folder with an extra `package.json` that *does* install `canvas` NPM dependency for some extra tests that the previous `Canvas` mock wasn't capable of and somehow only test with that when it was possible to install `canvas`. Maybe using NPM postinstall hook for installing "error prone" deps or something. :thought_balloon:

marklundin:
Ok I will revert this whilst we consider alternatives, even if we do end up use a post-install

PR #6954, opened 2024-09-13
Gsplat followup orientation and shader update

slimbuck:
This PR continues on from #6946 by:
* Orientating the GS scene in the instantiated gsplat resource (since we no longer reorientate splat data at load time). 
* Reorientate compressed data at load time to match uncompressed data.
* Update shaders so now uncompressed and compressed fragments are identical.

Also includes load time speedups:
* add a fast path for PLY files containing only float32 data (i.e. uncompressed GS data)
* speedup code which does packaging of transform, covariant and SH data for gpu
* implement fast path for calculating centers and focal point

These speedups result in `bicycle.ply` load time in Supersplat going from 23sec -> 9.5sec.

### DIFF
--- a/src/framework/parsers/gsplat-resource.js
+++ b/src/framework/parsers/gsplat-resource.js
@@ -66,6 +66,12 @@ class GSplatResource {
             instance: splatInstance
         });
 
+        // the ply scene data no longer gets automatically rotated on load, so do
+        // it here instead.
+        if (!this.splatData.isCompressed) {
+            entity.setLocalEulerAngles(0, 0, 180);
+        }
+
         // set custom aabb
         component.customAabb = splatInstance.splat.aabb.clone();
 
--- a/src/framework/parsers/ply.js
+++ b/src/framework/parsers/ply.js
@@ -340,7 +340,7 @@ const readFloatPly = async (streamBuf, elements, littleEndian) => {
         streamBuf.head += toRead * inputSize;
     }
 
-    return new GSplatData(elements); 
+    return new GSplatData(elements);
 };
 
 const readGeneralPly = async (streamBuf, elements, littleEndian) => {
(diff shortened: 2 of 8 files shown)


mvaligursky:
> Reorientate compressed data at load time to match uncompressed data.

Is there some cost here?

src/framework/parsers/gsplat-resource.js
```diff
@@ -66,6 +66,12 @@ class GSplatResource {
             instance: splatInstance
         });
 
+        // the ply scene data no longer gets automatically rotated on load, so do
+        // it here instead.
+        if (!this.splatData.isCompressed) {
+            entity.setLocalEulerAngles(0, 0, 180);
```
mvaligursky, line 72: not sure I'm keen on this, entity transform is owned by the user, we should not modify it. Better just have the entity upside down and the user needs to rotate it the way it needs to

slimbuck:
> Is there some cost here?

There is unfortunately. But this is only done when a compressed scene is uncompressed at load time. This should only be necessary for tooling like supersplat.

Ultimately we should probably stop reorienting compressed data on save.

src/framework/parsers/gsplat-resource.js
slimbuck: What's the danger in returning a correctly oriented entity?

It saves every user from having to worry about reorienting only uncompressed PLY (compressed PLY doesn't need this). And as you say, user owns the transform, they can do with it what they want.

It's either this or we create a child node with this transform, which I think is probably worse.

mvaligursky: They might have script that rotates it by specifying angles, and by default it would not work the same as for other entities, but would flip it upside down.

child entity is worse, agreed.
I'd say we do nothing here, let them handle it.
but ok, maybe pre-rotate it, and give trouble to smaller number of users, I guess I can live with it, even though I wont like it. I'd expect every entity the engine creates has identity transform.

mvaligursky (approved):

PR #6967, opened 2024-09-18
Tint color added to the color grading in the composite pass

mvaligursky:
extra color grading option

### DIFF
--- a/src/extras/render-passes/render-pass-compose.js
+++ b/src/extras/render-passes/render-pass-compose.js
@@ -28,10 +28,14 @@ const fragmentShader = /* glsl */ `
 
     #ifdef GRADING
         uniform vec3 brightnessContrastSaturation;
+        uniform vec3 tint;
 
         // for all parameters, 1.0 is the no-change value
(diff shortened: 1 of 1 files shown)


src/extras/render-passes/render-pass-compose.js
```diff
@@ -229,6 +235,7 @@ class RenderPassCompose extends RenderPassShaderQuad {
         this.ssaoTextureId = scope.resolve('ssaoTexture');
         this.bloomIntensityId = scope.resolve('bloomIntensity');
         this.bcsId = scope.resolve('brightnessContrastSaturation');
+        this.tintId = scope.resolve('tint');
```
willeastcott, line 238: Do we need to worry about name collisions? `tint` is quite generic.

willeastcott (approved):

src/extras/render-passes/render-pass-compose.js
mvaligursky: nope, they can be shared, as long as each rendering sets up the required value to it

PR #6969, opened 2024-09-19
RenderPassCameraFrame exposes HDR formats

mvaligursky:
- this allows the users to specify preference of the HDR format used for rendering
- also, when no HDR format is supported, the bloom is automatically disabled as non-supported

### DIFF
--- a/src/extras/render-passes/render-pass-camera-frame.js
+++ b/src/extras/render-passes/render-pass-camera-frame.js
@@ -21,6 +21,8 @@ export const SSAOTYPE_LIGHTING = 'lighting';
 export const SSAOTYPE_COMBINE = 'combine';
 
 class CameraFrameOptions {
+    formats;
+
     samples = 1;
 
     sceneColorMap = true;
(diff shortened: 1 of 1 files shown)


marklundin (approved):

src/extras/render-passes/render-pass-camera-frame.js
```diff
@@ -142,11 +144,19 @@ class RenderPassCameraFrame extends RenderPass {
 
     needsReset(options) {
         const currentOptions = this.options;
+
+        // helper to compare arrays
+        const arraysNotEqual = (arr1, arr2) => arr1 !== arr2 &&
+            (!(Array.isArray(arr1) && Array.isArray(arr2)) ||
+            arr1.length !== arr2.length ||
+            !arr1.every((value, index) => value === arr2[index]));
```
willeastcott, line 152: Did you forget about this:
https://github.com/playcanvas/engine/blob/main/src/core/array-utils.js#L4

mvaligursky: I spent few minutes looking for this and didn't find it .. I though I removed it in 2.0 :)
Will adjust, thanks!

PR #6973, opened 2024-09-19
Fix for -Infinity compressed GS data

slimbuck:
This PR implements a workaround for https://github.com/playcanvas/supersplat/issues/189.

In cases with infinity scales, chunk a and b values are equal and we can just take a.

### DIFF
--- a/src/scene/gsplat/gsplat-compressed-data.js
+++ b/src/scene/gsplat/gsplat-compressed-data.js
@@ -43,6 +43,7 @@ class SplatCompressedIterator {
         };
 
         const lerp = (a, b, t) => {
+            // choose a when a === b so we don't do math on -Infinity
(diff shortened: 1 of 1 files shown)


src/scene/gsplat/gsplat-compressed-data.js
```diff
@@ -42,7 +42,7 @@ class SplatCompressedIterator {
             }
         };
 
-        const lerp = (a, b, t) => a * (1 - t) + b * t;
+        const lerp = (a, b, t) => a === b ? a : a * (1 - t) + b * t;
```
mvaligursky, line 45: it cold do with a comment, as otherwise it's strange thing to do :)

slimbuck: Well only slightly strange. If a === b, then lerping between a and b is always a (or b if you prefer).

Issue #6751, opened 2024-06-25
Add support baseVertex for GPURenderPassEncoder


AlexAPPi:
Hi!

It would be nice to have baseVertex support for the mesh.

https://developer.mozilla.org/en-US/docs/Web/API/GPURenderPassEncoder/drawIndexed

https://github.com/playcanvas/engine/blob/03920ded44f6cee81ab16a904551b8b1ddecaeb5/src/platform/graphics/webgpu/webgpu-graphics-device.js#L521

https://github.com/playcanvas/engine/blob/03920ded44f6cee81ab16a904551b8b1ddecaeb5/src/scene/mesh.js#L256

mvaligursky:
WebGL 2 adds support for this using ** WEBGL_draw_instanced_base_vertex_base_instance** but its support seems to be very low (less than 1%)

AlexAPPi:
> WebGL 2 adds support for this using **WEBGL_draw_instanced_base_vertex_base_instance** but its support seems to be very low (less than 1%) here https://web3dsurvey.com/webgl2

If you focus on chrome-like browsers, they have support; this greatly speeds up the work with indexes, for example for the LOD system.

mvaligursky:
I see this:

<img width="928" alt="Screenshot 2024-06-27 at 12 31 56" src="https://github.com/playcanvas/engine/assets/59932779/dcae9836-5764-4710-8cec-19ad503baf77">

My MacOS on latest Chrome does not expose this extension.

What data do you see?

AlexAPPi:
Sorry, I'm talking about webGPU support.

https://developer.mozilla.org/en-US/docs/Web/API/GPURenderBundleEncoder/drawIndexed#browser_compatibility

AlexAPPi:
Closed https://github.com/playcanvas/engine/pull/6977

mvaligursky:
this needs to stay open till the implementation is merged in.

PR #6978, opened 2024-09-20
Implementation of baseVertex support

AlexAPPi:
Fixes https://github.com/playcanvas/engine/issues/6751

Added support primitive baseVertex for WebGPU and Batching.


src/scene/mesh.js
```diff
@@ -202,10 +202,11 @@ class Mesh extends RefCountedObject {
      *
      * - `base` is the offset of the first index or vertex to dispatch in the draw call.
      * - `count` is the number of indices or vertices to dispatch in the draw call.
+     * - `baseVertex` is the number added to each index value before indexing into the vertex buffers. (available only for WebGPU and Batching)
      * - `indexed` specifies whether to interpret the primitive as indexed, thereby using the
      * currently set index buffer.
      *
-     * @type {{type: number, base: number, count: number, indexed?: boolean}[]}
+     * @type {{type: number, base: number, count: number, baseVertex?: number, indexed?: boolean}[]}
```
mvaligursky, line 209: we should initialize it to 0 here as well, instead of leaving it undefined

AlexAPPi: In most cases, this parameter will be of little use, and there is no need to store it in memory for each primitive.

Maksims:
Is there an alternative of base vertex but for the WebGL to ensure parity between renderers?

mvaligursky:
> > Is there an alternative of base vertex but for the WebGL to ensure parity between renderers?
> 
> There is no cheap alternative for WebGL2. Only if you change the array of indices directly, via Mesh.setIndices

The real solution is for WebGL to support WEBGL_draw_instanced_base_vertex_base_instance on more devices. Till then this is a WebGPU feature (as compute shaders for example)

willeastcott:
LOL. 0.2% support.

https://web3dsurvey.com/webgl2/extensions/WEBGL_draw_instanced_base_vertex_base_instance

mvaligursky:
> LOL. 0.2% support.
> 
> https://web3dsurvey.com/webgl2/extensions/WEBGL_draw_instanced_base_vertex_base_instance

that is slightly less (-0.04%) than a month ago.

I don't get it. WebGPU supports it, meaning my mac book pro supports it. Some browsers support it on WebGL .. I assume that's not Chrome then?

Searching online, Chrome supports it behind a flag .. when they enable it, this will go to 50%

MAG-AdrianMeredith:
"Will come to all browsers shortly". that was from 2021. haha, webgl/webgpu standards progress summed up perfectly

PR #6979, opened 2024-09-20
Fix compound colliders

### LINKED ISSUE #6976: [BUG] Uncaught RangeError: Maximum call stack size exceeded
### Description

Colliding two compound colliders with kinematics giving error  **Uncaught RangeError: Maximum call stack size exceeded**

I tried to simulate the issue
* Simulated project : [https://playcanvas.com/project/1256880/overview/test-call-stack-error](https://playcanvas.com/project/1256880/overview/test-call-stack-error)
* image

LeXXik:
Fixes #6976 

The issue was that when we loop over compounds during physics update, we go over all nodes of the compound entity:

https://github.com/playcanvas/engine/blob/91148f0b35c88c1ad2d92e7ec086ca929b06f213/src/framework/components/collision/component.js#L694

Since we use `entity.forEach`, that includes the parent itself, which adds parent compound shape as a child to itself. When 2 such compounds collide, the bullet physics locks in a loop, causing max stack exception.

While at it, also changed the frequency of the children shapes transform updates. Instead of removing and adding them every frame if parent moved, they are now updated only when their own transforms changed or they are recreated.


willeastcott (approved):

Maksims:
I think it was caused by my assumption that forEach would not include self into the loop. It has been 5 years since.
If tested and works well, good fix!

LeXXik:
@willeastcott this should be good for v1 as well.

willeastcott:
@LeXXik I have just cherry-picked to `main_v1`.

Issue #6975, opened 2024-09-20
Add HDR output support to WebGPU


mvaligursky:
Introduced in Chrome 129: https://developer.chrome.com/blog/new-in-webgpu-129#hdr_support_with_canvas_tone_mapping_mode

<img width="876" alt="Screenshot 2024-09-20 at 10 36 07" src="https://github.com/user-attachments/assets/4b66cc18-441e-43cb-9b43-b0365887f4ab">

explainer: https://github.com/ccameron-chromium/webgpu-hdr/blob/main/EXPLAINER.md

PR #6980, opened 2024-09-20
Add support for Chrome HDR mode for WebGPU

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/6975

This adds support for HDR displays for WebGPU (new in Chrome 129). 

There is a new public constant that can be used to request HDR mode:
`DISPLAYFORMAT_HDR` // this is now public

also, new API to test if HDR is active:
`GraphicsDevice.isHdr`

Notes:
- few examples were updated to test this.
- old bloom was removed from the example as that is LDR implementation
- WebGPU implements single non-specific tonemapping, so it does not exactly match our tonemapping options. They plan to expose additional options in the future.
- there is an issue with some more advanced examples, I'll investigate more and likely open chromium issue, as the browser rendering gets broken in those cases.

### DIFF
--- a/src/platform/graphics/graphics-device.js
+++ b/src/platform/graphics/graphics-device.js
@@ -105,6 +105,14 @@ class GraphicsDevice extends EventHandler {
     isWebGL2 = false;
 
     /**
+     * True if the back-buffer is using HDR format, which means that the browser will display the
+     * rendered images in high dynamic range mode. This is true if the displayFormat is set to
+     * {@link DISPLAYFORMAT_HDR} when creating the graphics device, and HDR is supported by the
+     * device.
+     */
+    isHdr = false;
+
+    /**
      * The scope namespace for shader attributes and variables.
      *
      * @type {ScopeSpace}
--- a/src/platform/graphics/graphics-device.js
+++ b/src/platform/graphics/graphics-device.js
@@ -106,9 +106,9 @@ class GraphicsDevice extends EventHandler {
 
     /**
      * True if the back-buffer is using HDR format, which means that the browser will display the
-     * rendered images in high dynamic range mode. This is true if the displayFormat is set to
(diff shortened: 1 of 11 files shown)


slimbuck (approved):
The naming is becoming a little confusing since we can already render to HDR backbuffer.

mvaligursky:
> The naming is becoming a little confusing since we can already render to HDR backbuffer.

How can we already render to HDR backbuffer?

slimbuck:
SHADER_FORWARDHDR

mvaligursky:
ahh that .. that was removed in v2 .. confusion solved ;)
But that would not rendering to HDR backbuffer, only HDR render target.

slimbuck:
> HDR backbuffer, only HDR render target.

yeah this bit, confusing

slimbuck:
Only mentioning so we can perhaps make the difference explicit in docs. Though it may not be necessary since as you say, it's gone from v2.0.

src/platform/graphics/graphics-device.js
```diff
@@ -104,6 +104,14 @@ class GraphicsDevice extends EventHandler {
      */
     isWebGL2 = false;
 
+    /**
+     * True if the back-buffer is using HDR format, which means that the browser will display the
+     * rendered images in high dynamic range mode. This is true if the displayFormat is set to
```
willeastcott, line 109: Link `displayFormat`?

willeastcott (approved):

src/platform/graphics/graphics-device.js
mvaligursky: we cannot link to options.displayFormat directly I think, but I link a function accepting this options object.

MAG-AdrianMeredith:
kewl, this looks so much nicer than my hack from a while back

PR #6985, opened 2024-09-23
Use fallback if wasm compile fails

slimbuck:
This is to address the issue in https://forum.playcanvas.com/t/unhandled-promise-rejection-compileerror-webassembly-module-problem/37038


mvaligursky (approved):

src/framework/handlers/basis.js
```diff
@@ -87,7 +87,15 @@ const prepareWorkerModules = (config, callback) => {
                 }
             })
             .catch((err) => {
-                callback(err, null);
+                Debug.warn(`compile() failed for ${config.wasmUrl} (${err}), using fallback script.`);
+
+                http.get(config.fallbackUrl, options, (err, response) => {
```
willeastcott, line 92: Is the fallback script mandatory? I thought some WASM modules out there didn't have a fallback. So assuming it's _not_ guaranteed to be present, this should probably check it's set?

willeastcott (approved):

src/framework/handlers/basis.js
willeastcott: Oh, right yeah, this is for Basis specifically. I thought it was a generic code path for loading modules.

slimbuck:
After investigating a bit more I found the following:
- when using the basis fallback script, the engine has a bug where it passes `null` instead of `undefined` to `self.BASIS()` in basis-worker.js. This causes the fallback to error and fail.
- with the previous fix in place, the current build of basis.js fallback script fails to load due to a closure bug.
- recompiling the basis module with latest toolchain and basis code fixes the issue.
- the recompiled basis.wasm file is 50k larger
- we can recover this size increase by removing support for pvr and atc. with this in place we **_save_** 150K on the wasm binary size.
- but this change requires us to update the engine and remove support for the same. this change is great though and results in a bunch of code being removed.
- I have made the above changes in this PR, but have tested none of it.

Since the latest version of basis transcoder supports HDR formats, we should add support for compressed HDR as part of this work. However that requires supporting BC7 texture formats, which @mvaligursky recently (today!) added [to engine v2](https://github.com/playcanvas/engine/pull/6988).

Since the user who originally opened the issue has found a workaround (using android instead of iphone), this PR will now be closed till we have time to complete the work and test it, most likely in engine v2.

PR #6992, opened 2024-09-26
Gizmo updates

### LINKED ISSUE #6981: Simplify Gizmo creation
Currently we need this code to create a gizmo:

```js
const gizmoLayer = new pc.Layer({
    name: 'Gizmo',
    clearDepthBuffer: true,
    opaqueSortMode: pc.SORTMODE_NONE,
    transparentSortMode: pc.SORTMODE_NONE
});
const layers = app.scene.layers;
layers.push(gizmoLayer);
camera.camera.layers = camera.camera.layers.concat(gizmoLayer.id);

// create gizmo
const gizmo = new pc.TranslateGizmo(app, camera.camera, gizmoLayer);
gizmo.attach([box]);
```

The upper/big block of code is just... complicated boilerplate code that no one can remember (at least not me). The last two lines do what a person really wants from a gizmo.

Ideally I would like the required code for adding a gizmo to be short like this:

```js
const gizmo = new pc.TranslateGizmo(app, camera.camera);
gizmo.attach(box); // Accept non-arrays, just like e.g. `HTMLElement#append`
```

It would be nice if the gizmo system ensures creation of the layer if it doesn't exist yet?

Considering we are also running into issues like #6949 it's nice to not make everyone refactor layer code if it happens to need a change. The gizmo system would just do what is needed for certain updates.

### LINKED ISSUE #6990: Gizmo doesn't follow what it is attached to
Testable in: https://playcanvas.vercel.app/#/gizmos/transform-translate

If you move the box afterwards, it suddenly isn't attached any longer:

```js
gizmo.attach([box]);
box.translate(1, 2, 3);
```

Result:

image

The only way around this seems to be to call:

```js
gizmo._updatePosition()
```

But this is an internal function and even if it were public, it would still be strange to "reinforce" an attachment - as attaching something implies a strong link between things.

kpal81xd:
Fixes #6981
Fixes #6990

- Gizmo follows nodes when attached
- Updated api to allow `gizmo.attach` to accept a single node
- Layer is created internally and reused by other gizmos unless specified
- constructor modified to omit `app` parameter
- Refactored `AxisShapes` into separate `Shape` classes
- Added `shading` flag to `TransformGizmo` types and set default to false

### API changes
- `new Gizmo(app: AppBase, camera: CameraComponent, layer: Layer)` -> `new Gizmo(camera: CameraComponent, layer: Layer)`
- **NEW:** `const layer = Gizmo.createLayer(app: AppBase, layerName?: string = 'Gizmo', layerIndex?: number = -1)`

N.B The `createLayer` static function will create a new layer and will add it to the scene. The Gizmo class constructor will append the layer to the camera's set of layers.

### Preview
image


src/extras/gizmo/gizmo.js
```diff
@@ -230,19 +231,35 @@ class Gizmo extends EventHandler {
...
      *
-     * @param {AppBase} app - The application instance.
      * @param {CameraComponent} camera - The camera component.
-     * @param {Layer} layer - The render layer.
+     * @param {Layer} [layer] - The render layer.
      * @example
      * const gizmo = new pc.Gizmo(app, camera, layer);
      */
-    constructor(app, camera, layer) {
+    constructor(camera, layer) {
```
mvaligursky, line 239: ideally when we make an API breaking change like this, we try to assert to warn user of the change. See the MeshInstance constructor as an example.

mvaligursky:
Can you please list API changes in the descriptor

kungfooman:
Very nice work!

I just tested it and I'm running into this error when refreshing the translation example (happens in `nodes:detach` event):

image

src/extras/gizmo/gizmo.js
```diff
@@ -258,9 +275,22 @@ class Gizmo extends EventHandler {
         this._device.canvas.addEventListener('pointermove', this._onPointerMove, true);
         this._device.canvas.addEventListener('pointerup', this._onPointerUp);
 
-        app.on('update', () => this._updateScale());
+        this._app.on('update', () => {
+            this._updatePosition();
+            this._updateRotation();
+            this._updateScale();
```
kungfooman, line 281: Could we only update when position actually changed (simple distance-squared check against epsilon should suffice to not spam with update events)?

kpal81xd:
> Very nice work!
> 
> I just tested it and I'm running into this error when refreshing the translation example (happens in `nodes:detach` event):
> 
> image

I cant seem to repro this what did you change in the example exactly?

kungfooman:
> I cant seem to repro this what did you change in the example exactly?

It's a strange issue, this is how I can reproduce:

1) No code changes
2) Load http://localhost:5555/index.html#/gizmos/transform-translate
3) Drag blue translation manipulator once
4) Refresh example (pressing example refresh button or Shift+Enter)
5) Repeat until you see the error... strangely enough first it "just worked" a couple of times, then the bug happened all the time. I counted max 12 refresh attempts until you should see the error, mostly it happens at third time already

kungfooman:
> Ive tried this multiple times and still no error

Hm, maybe it only happens for WebGL 2? I get the highest error rate when I drop it on the Controls DOM:

image


Should error after ~ 3-5 attempts

kpal81xd:
> > Ive tried this multiple times and still no error
> 
> Hm, maybe it only happens for WebGL 2? I get the highest error rate when I drop it on the Controls DOM:
> 
> image
> 
> Should error after ~ 3-5 attempts

Okay ive refactored a bunch of stuff so that line isnt used anymore

kpal81xd:
@kungfooman okay found the error mesh instances were trying to be referenced after being destroyed oops

src/extras/gizmo/gizmo.js
```diff
@@ -25,9 +26,14 @@ const tmpM2 = new Mat4();
...
 
 // constants
-const MIN_GIZMO_SCALE = 1e-4;
+const LAYER_NAME = 'Gizmo';
+const MIN_SCALE = 1e-4;
 const PERS_SCALE_RATIO = 0.3;
 const ORTHO_SCALE_RATIO = 0.32;
+const UPDATE_EPSILON = 1e-6;
+
+let gizmoLayer = null;
```
willeastcott, line 35: Will this work with multi-applications on the same page?

src/extras/gizmo/tri-data.js
```diff
@@ -55,13 +55,26 @@ class TriData {
         return this._priority;
     }
 
+    /**
+     * Sets the transform of the triangle data.
+     * @param {Vec3} [pos] - The position of the transform.
```
willeastcott, line 60:
```suggestion
     * Sets the transform of the triangle data.
     *
     * @param {Vec3} [pos] - The position of the transform.
```

willeastcott (approved):

src/extras/gizmo/gizmo.js
```diff
@@ -230,19 +236,35 @@ class Gizmo extends EventHandler {
     /**
      * Creates a new Gizmo object.
      *
-     * @param {AppBase} app - The application instance.
      * @param {CameraComponent} camera - The camera component.
-     * @param {Layer} layer - The render layer.
+     * @param {Layer} [layer] - The render layer.
```
mvaligursky, line 240: explain what happens if no layer is passed in here.

src/extras/gizmo/gizmo.js
```diff
@@ -230,19 +236,35 @@ class Gizmo extends EventHandler {
...
-        this._device = app.graphicsDevice;
         this._camera = camera;
-        this._layer = layer;
+        this._app = camera.system.app;
+        this._device = this._app.graphicsDevice;
+
+        if (layer) {
+            this._layer = layer;
+        } else {
+            if (!gizmoLayer) {
```
mvaligursky, line 255: you should also check if the gizmo layer already exists (and the user didn't pass it in)

kpal81xd: I dont think this would be ideal since if the user created their own gizmo layer they could name it whatever they wanted. So either the user manages the gizmo layer entirely or gizmo manages the layers

kpal81xd: Also this would cause the layer that you created to be removed from the camera and scene if passed as an argument to the gizmo

src/extras/gizmo/gizmo.js
kpal81xd: Just fixed 😅

src/extras/gizmo/gizmo.js
LeXXik: Can there be 2 gizmos in a scene?

kpal81xd: Yes the editor example uses all 3 of the transform gizmos

src/extras/gizmo/gizmo.js
```diff
@@ -230,19 +235,38 @@ class Gizmo extends EventHandler {
...
+            if (!layerStore.has(this._app)) {
+                const gizmoLayer = new Layer({
+                    name: LAYER_NAME,
+                    clearDepthBuffer: true,
+                    opaqueSortMode: SORTMODE_NONE,
+                    transparentSortMode: SORTMODE_NONE
+                });
+                this._app.scene.layers.push(gizmoLayer);
+                this._camera.layers = this._camera.layers.concat(gizmoLayer.id);
+                layerStore.set(this._app, { layer: gizmoLayer, refs: 0 });
```
mvaligursky, line 264: the Map() stores reference to the App used as a key .. so unless it's removed from the map, it won't be garbage collected. Not ideal. Use DeviceCache which clears the device data when the device is destroyed.

src/extras/gizmo/gizmo.js
mvaligursky: layer name is unique .. we have a layer names map. So we need to make sure to not create a layer if one with the name exists.

src/extras/gizmo/gizmo.js
```diff
@@ -227,22 +230,45 @@ class Gizmo extends EventHandler {
      */
     intersectData = [];
 
+    /**
+     * Creates a new gizmo layer and adds it to the scene.
+     *
+     * @param {AppBase} app - The app.
+     * @param {string} [layerName] - The layer name. Defaults to 'Gizmo'.
+     * @param {number} [layerIndex] - The layer index. Defaults to -1.
```
mvaligursky, line 238: clarify what happens in the default case (added to the end?)
maybe use undefined instead, does not need a default
then the code can be 
const index = layerIndex ?? app.scene.layers.layerList.length;

kpal81xd: The convention is that index -1 is the end of the list so I would think thats clear from the default but I can make it more explicit

src/extras/gizmo/shape/arc-shape.js
```diff
@@ -0,0 +1,115 @@
...
+        // arc/circle
+        this._createRenderComponent(this.entity, [
+            this._createTorusMesh(this._sectorAngle),
+            this._createTorusMesh(360)
+        ]);
+        this.drag(false);
+    }
+
+    set tubeRadius(value) {
+        this._tubeRadius = value ?? 0.1;
```
mvaligursky, line 64: are you handling a case the user passes in undefined?

kpal81xd: This techincally isnt public API but yes

mvaligursky: ok. strange to call `something.tubeRadius = undefined;` but ok ;)

src/extras/gizmo/gizmo.js
mvaligursky: mention it in the JSDoc please.

mvaligursky (approved):

PR #6998, opened 2024-09-30
Improvements to non-clustered omni light shadows

mvaligursky:
- removed webgl1 based omni shadows that were encoding depth in the RGBA8, and using depth buffer instead
- swapped 9 sample based omni PCF3x3 but more optimized 4 sample based
- bunch fo cleanup of the related shadow code
- added PCF1 support to omni shadows


slimbuck (approved):

willeastcott (approved):
Lovely simplification 🪄

Issue #6949, opened 2024-09-12
Gizmo not rendered when using renderpass


MAG-AdrianMeredith:
### Description

When activating renderpass on a camera, the gizmo is no longer visible.  

![image](https://github.com/user-attachments/assets/e34fea07-d0ee-4f06-b53b-7596b4e29bb7)

### Steps to Reproduce

1. go to https://playcanvas.vercel.app/#/gizmos/transform-translate example
2. add the following code to enable renderpass
`  const currentOptions = {
        camera: camera.camera, // camera used to render those passes
        samples: 0, // number of samples for multi-sampling
        sceneColorMap: true, // true if the scene color should be captured
        bloomEnabled: true,

        // disabled TAA as it currently does not handle dynamic objects
        prepassEnabled: false,
        taaEnabled: false
    };
      // Use a render pass camera frame, which is a render pass that implements typical rendering of a camera.
        // Internally this sets up additional passes it needs, based on the options passed to it.
        const renderPassCamera = new pc.RenderPassCameraFrame(app, currentOptions);

        // and set up these rendering passes to be used by the camera, instead of its default rendering
        camera.camera.renderPasses = [renderPassCamera];`
3. gizmo is no longer visible

mvaligursky:
It works if you add this render pass setup code to the camera after the gizmoLayer has been added to the camera.
I'll try to add some warning when the layers are modified after the render pass has been set up or something.

mvaligursky:
by the way, when you need to see what is happening with render passes (or even without them as they're used under the hood anyways), use

```
    pc.Tracing.set(pc.TRACEID_RENDER_PASS, log);
    pc.Tracing.set(pc.TRACEID_RENDER_PASS_DETAIL, log);
```

typically enable it for one frame, not every frame.

MAG-AdrianMeredith:
Needing to run functions in a certain order is not ideal and error prone.  Is there a layer added event or similar that could allow the render pass to sync up with the changes?

mvaligursky:
Yeah it's on a longer term list for sure. But having a very quick look, it might not be too complicated to update that one pass internally.

MAG-AdrianMeredith:
Yep there is I can see what needs to happen. 

1. hook into the camera component layer change events
2. pass that to the renderpasses (add similar layer change events)
3. each render pass would then be able to optionally handle it if needed (similar to exising onEnable function etc)

 if its still open after the migration I'll take a look at it since I have a workaround for now
 
 Thanks

[...]

PR #7001, opened 2024-10-01
Improvement to layer handling by render passes

mvaligursky:
related to https://github.com/playcanvas/engine/issues/6949

- When a RenderPassCamera creates forward render passes, it includes even disabled layers, and the layers can be later toggled on and off as needed
- When a layer from the composition is deleted, a warning is printed and the layer is ignored if still included in forward passes. Those need to be re-created to avoid this.
- when PostEffects are used on the same camera as render passes, a warning is logged and post-effects are skipped, as this is not compatible

### DIFF
--- a/src/scene/renderer/render-pass-forward.js
+++ b/src/scene/renderer/render-pass-forward.js
@@ -58,6 +58,10 @@ class RenderPassForward extends RenderPass {
         this.renderer = renderer;
     }
 
+    get rendersAnything() {
+        return this.renderActions.length > 0;
+    }
+
     addRenderAction(renderAction) {
         this.renderActions.push(renderAction);
     }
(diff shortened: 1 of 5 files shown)


src/scene/renderer/render-pass-forward.js
```diff
@@ -114,19 +118,18 @@ class RenderPassForward extends RenderPass {
...
+        const { layerList, subLayerList } = composition;
         let clearRenderTarget = firstLayerClears;
 
         let index = startIndex;
         while (index < layerList.length) {
 
             const layer = layerList[index];
             const isTransparent = subLayerList[index];
-            const enabled = layer.enabled && subLayerEnabled[index];
             const renderedbyCamera = cameraComponent.camera.layersSet.has(layer.id);
```
MAG-AdrianMeredith, line 129: typo, renderedByCamera

slimbuck (approved):

PR #7006, opened 2024-10-03
#7004 Ensure outliner only adds enabled render components

### LINKED ISSUE #7004: Overlay Renderer (pc 2.0) causes hidden objects to be visible
### Description
When using the new Overlayer renderer (not the legacy post process).  If an entity contains hidden children they are made unhidden causing unexpected items to appear.  This is because by default it takes all the render components (hidden or not) and adds them into the intermediate layer and so they will now always be drawn.

### Steps to Reproduce

1.  Create an entity with a hidden render component
2. Add an outline (pc 2.0) to it
3. Object becomes visible

MAG-AdrianMeredith:
Fixes #7004 

Ensure only enabled rendercomponents have their meshcomponents added to the layer


mvaligursky (approved):
Thanks!

PR #7009, opened 2024-10-07
Do not allow gizmo rendering to use debug output

mvaligursky:
- always renderi gizmos using standard output, and not debug output

old
Screenshot 2024-10-07 at 10 39 19

vs new
Screenshot 2024-10-07 at 10 38 49


kpal81xd (approved):

mvaligursky:
Cherry picked to v1 as well.

MAG-AdrianMeredith:
@mvaligursky This is broken for me?  chunks doesn't exist on shadermaterial and therefore throws and exception.  What am I missing?

image

PR #6642, opened 2024-05-30
ESM Multi camera

kpal81xd:
- Added base camera ESM script
- Added multi-camera ESM script which extends base camera
- Added example for multi-camera and controls for attributes

### Preview
image

PR #7010, opened 2024-10-07
ESM Orbit Camera 

marklundin:
Swaps the Orbit Camera out for the ESM version. If this get's approved I can generate the same for all other scripts.

There's a bigger discussion on discovery and how we can provide these to the user, but for now, this is a simple swap

- updating attributes declarations
- `var` -> `let`/`const`
- `attr:*` events to getters/setters
-  callbacks to arrow functions where sensible

### DIFF
--- a/scripts/camera/orbit-camera.mjs
+++ b/scripts/camera/orbit-camera.mjs
@@ -1,0 +1,711 @@
+import { 
+    Script, Entity, math, Vec3, BoundingBox, Quat, Vec2, PROJECTION_PERSPECTIVE,
+    MOUSEBUTTON_LEFT, MOUSEBUTTON_RIGHT, MOUSEBUTTON_MIDDLE, EVENT_TOUCHCANCEL,
+    EVENT_TOUCHSTART, EVENT_TOUCHEND, EVENT_TOUCHMOVE, EVENT_MOUSEDOWN, EVENT_MOUSEMOVE, EVENT_MOUSEUP, EVENT_MOUSEWHEEL
+} from 'playcanvas';
+
+export class OrbitCamera extends Script {
+    /**
+     * @attribute
+     * @title Distance Max
+     * @resource 
+     * @range
+     * @type {number}
+     */
+    set distanceMax(v) {
+        this._distanceMax = v;
+        this._distance = this._clampDistance(this._distance);
+    }
+
+    get distanceMax() {
+        return this._distanceMax;
+    }
+
+    /**
+     * @attribute
+     * @title Distance Min
+     * @type {number}
+     */
+    set distanceMin(v) {
+        this._distanceMin = v;
+        this._distance = this._clampDistance(this._distance);
+    }
+
+    get distanceMin() {
+        return this._distanceMin;
+    }
+
+    /**
+     * @attribute
+     * @title Pitch Angle Max (degrees)
+     * @type {number}
+     */
+    set pitchAngleMax(v) {
+        this._pitchAngleMax = v;
+        this._pitch = this._clampPitchAngle(this._pitch);
+    }
+
+    get pitchAngleMax() {
+        return this._pitchAngleMax;
+    }
+
+    /**
+     * @attribute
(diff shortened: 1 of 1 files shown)


mvaligursky:
It'd be great to update one of the examples to use this, maybe cascaded shadow maps? It'd be great to see how it gets used.

marklundin:
> It'd be great to update one of the examples to use this, maybe cascaded shadow maps? It'd be great to see how it gets used.

Yeah it should just be a case of doing

```javascript
import { OrbrtCamera }  from '/scripts/camera/orbit-camera.mjs'
entity.add(OrbitCamera)
```

unfortunately it's not easy to import things in the examples browser 😩 

> Not sure about deleting orbit-camera.js - lots of engine examples use it?

yeah I guess, we'd need to update everything in sync

mvaligursky:
I'm importing the script in the examples this way (from a different folder, but that's a good start for now)
https://github.com/playcanvas/engine/pull/7011

marklundin:
Yeah ideally we only have scripts in one location. We don't really want to have a copy in the examples/misc

mvaligursky:
Agreed, for me that is THE location. But we need to surface those scripts to the Editor and that will change then.

scripts/camera/orbit-camera.mjs
```diff
@@ -0,0 +1,709 @@
...
+    /** @private */
+    _distanceMax = 0;
+
+    /** @private */
+    _frameOnStart = true;
+
+    /** @private */
+    _focusEntity;
+
+    static fromWorldPoint = new Vec3();
```
Maksims, line 213: This is a module script, means it has own namespace, so declaring temporary vectors in that namespace will make it easier to reference within the script, and wont need to write `const fromWorldPoint = OrbitCameraInputMouse.fromWorldPoint;` later.

Maksims:
These new scripts seem to not have `name` on them, will they still work when using scene loaders and referenced scripts by names (not class references)?

mvaligursky:
> Please don't force people to switch by removing currently used scripts.

We do not want to maintain each script in two forms. We're planning to convert all (maybe apart from post-effects) scripts stored in the engine to the new format. People are welcome to use old format if they already have them in their project though. Those are 1:1 replacement though, so it should be easy for people to switch to new format? Or do you see some complications with this?

scripts/camera/orbit-camera.mjs
marklundin: Yep, some of these are scoped to a class  ie `OrbitCamera` and `OrbitCameraInputMouse`

Maksims:
> Or do you see some complications with this?

I don't think Editor has complete feature parity and similar level of UX yet to force such a switch.
Previously any new systems would be provided as a more convenient option until devs decided to switch, without forcing by force-deprecating or removing features/code.

marklundin:
Curious to know what you think is missing?

At the moment the UX is to manually find a script and copy/paste it into the project. This could be much better, but for now it works and is identical workflow for both. An ESM script can be added to a project containing Classic scripts and vice-versa and it won't affect your existing scripts.

LeXXik:
> At the moment the UX is to manually find a script and copy/paste it into the project.

I personally prefer to import the orbit script from the store, so I don't need to search/copy/paste it. Can it be added there?

mvaligursky:
> > At the moment the UX is to manually find a script and copy/paste it into the project.
> 
> I personally prefer to import the orbit script from the store, so I don't need to search/copy/paste it. Can it be added there?

We're working on a way to surface engine provided scripts to the Editor, and the discussion is ongoing here. Stay tuned. Store is one way, but we're hoping for simply importing them from the "playcanvas" or similar.

kpal81xd:
This version of the orbit camera will be deprecated in favor of extending of the base camera defined in GH-6642 with the multi-camera

kpal81xd:
@marklundin This can be closed now since Orbit camera in included in GH-7111

Issue #5381, opened 2023-06-08
Prevent NaN with disk area lights


GSterbrant:
The original LTC (https://github.com/selfshadow/ltc_code/) implementation for disk area lights can produce NaN values on some GPUs with some floating point buffer formats. 

Blender solves this by adding two safeguards, first a normalization on the cross product vector of the disk tangents (V3): 
<img width="419" alt="image" src="https://github.com/playcanvas/engine/assets/107400752/64776fd6-9b74-4835-86bd-f2d0df5e71e7">

And a clamp of the form factor calculation with 0 to not get negative values:

<img width="689" alt="image" src="https://github.com/playcanvas/engine/assets/107400752/81ab6017-9ae4-442f-b83c-4eae65abcfe9">

The blender LTC code can be found here: https://github.com/blender/blender/blob/main/source/blender/draw/engines/eevee/shaders/ltc_lib.glsl

I have verified this issue on a Windows machine with a 7900 XTX and R10G11B11F/R16G16B16A16F formats, and that the Blender implementation fixes this issue.

PR #7015, opened 2024-10-09
[Fix] Fix to area lights (disk, sphere) occasionally generating -Inf pixels

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/5381

using the solution suggested in the issue, based on Blender's eevee renderer solution

thanks @GSterbrant !

This stops the -Inf values getting generates for few pixels of Sphere lights in the ClusteredAreaLights demo, allowing HDR bloom without spreading large black squares.


willeastcott (approved):

mvaligursky:
merged to main_v1 as well

PR #7023, opened 2024-10-10
Zero out setTimeout id

slimbuck:
I have noticed that `cancelAnimationFrame` actually appears on performance captures in Chrome, but cancelling should be almost always unnecessary.

In this PR I nullify the timeout id at the end of the update loop so cancelling is skipped during the next invocation.

I can't think of negative impact or side effect of this change, can you @Maksims ?

### DIFF
--- a/src/framework/app-base.js
+++ b/src/framework/app-base.js
@@ -2090,6 +2090,8 @@ const makeTick = function (_app) {
         if (application._destroyRequested) {
             application.destroy();
         }
+
+        application.frameRequestId = null;
     };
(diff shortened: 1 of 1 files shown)


Maksims:
I believe the case of canceling was necessary for WebXR, either during session creation, and/or during WebXR active session to avoid rendering to non-xr canvas while session is running.
That needs to be tested.

slimbuck:
Actually the entire tick implementation needs a rewrite. Closing for now.

PR #7029, opened 2024-10-11
Check xr policy

### LINKED ISSUE #6692: Permissions policy violation: xr-spatial-tracking
If the document doesn't allow to use WebXR Device API, then a DOM exception is thrown:

image

Details:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Permissions-Policy/xr-spatial-tracking

LeXXik:
Fixes #6692 

Checks if the WebXR API is allowed to be used before using it (`navigator.xr.isSessionSupported`).

### DIFF
--- a/src/framework/xr/xr-manager.js
+++ b/src/framework/xr/xr-manager.js
@@ -669,6 +669,14 @@ class XrManager extends EventHandler {
      * @private
      */
     _sessionSupportCheck(type) {
+        const featurePolicy = document?.featurePolicy;
+        if (featurePolicy) {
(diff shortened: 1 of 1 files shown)


Maksims:
I believe this will break access to WebXR features for users that have not yet granted permission for it per origin. As allowsFeature returns false if it was either denied or were not requested by origin (by default it will false).

While "support" detects if WebXR is supported on the platform, not necessary allowed.

So this PR will break WebXR for pretty much all new users.

PR #7038, opened 2024-10-15
Ammo pointer fix

### LINKED ISSUE #7036: Example 'Compound Collision' no longer works correctly
repro: https://playcanvas.vercel.app/#/physics/compound-collision

Notice that the legs and backrest of the chair no longer colliders with the ground.

Screenshot 2024-10-15 at 11 21 33

LeXXik:
Fixes #7036 

Correct way to get a pointer in Emscripten now is via a `.getPointer(object)` method. Reading `.ptr` property is not, which can be `undefined`, which made the `undefined === undefined` test return the same first child shape.

Also removed the underscore from the class method, as it is used by another class.

### DIFF
--- a/src/framework/components/collision/component.js
+++ b/src/framework/components/collision/component.js
@@ -626,13 +626,13 @@ class CollisionComponent extends Component {
      * @returns {number|null} The shape's index in the child array of the compound shape.
      * @private
      */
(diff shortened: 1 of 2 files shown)


mvaligursky (approved):

mvaligursky:
Thanks for the quick fix here!

PR #7040, opened 2024-10-15
Combine render passes where feasible

mvaligursky:
Optimization: when two (or more) following render passes render to the same render target + few other conditions are met, render them in a single pass. This is an optimization for tiled architecture (mostly mobile), to avoid the first pass having to write out the render target to main memory and then load it back in the following pass, keeping it all in the fast tile memory.

Typically a UI rendering is merged into RenderPassCompose (post-processing). We save the color buffer + depth buffer store and then load, which is costly especially when MSAA is used.
Screenshot 2024-10-15 at 16 06 49

### DIFF
--- a/src/scene/frame-graph.js
+++ b/src/scene/frame-graph.js
@@ -90,6 +90,41 @@ class FrameGraph {
             }
         }
 
+        // merge passes if possible
+        for (let i = 0; i < renderPasses.length - 1; i++) {
+            const firstPass = renderPasses[i];
+            const firstRT = firstPass.renderTarget;
+            const secondPass = renderPasses[i + 1];
+            const secondRT = secondPass.renderTarget;
+
+            // if the render targets are different, we can't merge the passes
+            // also only merge passes that have a render target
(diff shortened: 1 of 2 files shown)


willeastcott (approved):

src/scene/frame-graph.js
```diff
@@ -90,6 +90,41 @@ class FrameGraph {
...
+        // merge passes if possible
+        for (let i = 0; i < renderPasses.length - 1; i++) {
+            const firstPass = renderPasses[i];
+            const firstRT = firstPass.renderTarget;
+            const secondPass = renderPasses[i + 1];
+            const secondRT = secondPass.renderTarget;
+
+            // if the render targets are different, we can't merge the passes
+            // also only merge passes that have a render target
+            if (firstRT !== secondRT || firstRT === undefined) {
```
LeXXik, line 102: `!firstRT` ?

mvaligursky: nope, a NULL is valid case, just not undefined. null means rendering to backbuffer.

PR #7042, opened 2024-10-16
Fix GLB parser index buffer initialization

### LINKED ISSUE #5869: GLB parser incorrect index buffer construction
When GLB parser creates an index buffer, it passes a typed array to the constructor, instead of an array buffer. This results in view being a storage in the index buffer, instead of array buffer.

https://github.com/playcanvas/engine/blob/dc20e8e6c4555bd6a3418ee313a6fc57ca7c1d58/src/framework/parsers/glb-parser.js#L835

LeXXik:
Fixes #5869 

Fixes a bug, where GLB parser passes a typed array instead of an array buffer to the index buffer constructor.

### DIFF
--- a/src/framework/parsers/glb-parser.js
+++ b/src/framework/parsers/glb-parser.js
@@ -767,7 +767,7 @@ const createMesh = (device, gltfMesh, accessors, bufferViews, vertexBufferDict, 
                     indices = new Uint16Array(indices);
                 }
 
-                const indexBuffer = new IndexBuffer(device, indexFormat, indices.length, BUFFER_STATIC, indices);
+                const indexBuffer = new IndexBuffer(device, indexFormat, indices.length, BUFFER_STATIC, indices.buffer);
                 mesh.indexBuffer[0] = indexBuffer;
                 mesh.primitive[0].count = indices.length;
             } else {


LeXXik:
This breaks the GLB examples.

LeXXik:
Hmm, the examples work fine on my localhost, though...

LeXXik:
@mvaligursky hmm, why does vercel have this webgpu index buffer?

image

vs current implementation (and one on my localhost):
https://github.com/playcanvas/engine/blob/c66a67f8a2a432aca63da7170a0cf0d83ea80bf9/src/platform/graphics/webgpu/webgpu-index-buffer.js#L18-L21

LeXXik:
For some reason, the index buffer implementation has a null buffer in vercel preview deployment, but not when running the examples locally:

image

vs

image

LeXXik:
Ok, the issue is that when Index Buffer sets data, it uses byte length with bytes count to check the size. It expects it to be an array buffer, but receives and acts as if it is a typed array:

https://github.com/playcanvas/engine/blob/c66a67f8a2a432aca63da7170a0cf0d83ea80bf9/src/platform/graphics/index-buffer.js#L160-L171

Array Buffer in this case has a byte length of 4288, while the typed array has 72 with an offset:

image

So, the numbers will match only if a typed array is passed, when using 72, instead of 4288. @mvaligursky please advice.

LeXXik:
I suppose one option could be to change the docs, so the index buffer takes a typed array, instead of array buffer, and change the procedural so that it passes a typed array as well. Not sure about the code base though, but it seems a typed array is used everywhere or we would notice this earlier.

mvaligursky:
Thanks for digging into this, and sorry it sat open for so long.

This has since been resolved the other way, along the lines you suggested in your last comment here: #5869 was closed by #8988, which documents the actual contract instead of changing the parser. `IndexBuffer` now documents `initialData`, `lock()` and `setData()` as `{ArrayBuffer|ArrayBufferView}`, stored by reference.

That also avoids the breakage you spotted: `setData` validates `data.byteLength` against `numBytes`, so passing `indices.buffer` reports the whole glTF buffer view's length rather than the typed array's, and the size check fails.

So the GLB parser passing the typed array directly is now the documented, correct behaviour. Closing this one out - thanks again for the investigation, it's what informed the fix.

PR #7045, opened 2024-10-16
Gizmo clipping plane fix

kpal81xd:
Fixes gizmo clipping by the camera

Here you can see the gizmo has not been clipped but the grid lines have:

image

### DIFF
--- a/src/extras/gizmo/shape/shape.js
+++ b/src/extras/gizmo/shape/shape.js
@@ -51,7 +51,7 @@ const shaderDesc = {
         uniform mat4 matrix_viewProjection;
         void main(void) {
             gl_Position = matrix_viewProjection * matrix_model * vec4(vertex_position, 1.0);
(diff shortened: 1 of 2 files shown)


willeastcott (approved):

src/extras/gizmo/shape/shape.js
```diff
@@ -47,21 +47,19 @@ const shaderDesc = {
         attribute vec3 vertex_position;
         attribute vec4 vertex_color;
         varying vec4 vColor;
-        varying vec2 vZW;
         uniform mat4 matrix_model;
         uniform mat4 matrix_viewProjection;
         void main(void) {
             gl_Position = matrix_viewProjection * matrix_model * vec4(vertex_position, 1.0);
+            gl_Position.z = clamp(gl_Position.z, -abs(gl_Position.w), abs(gl_Position.w)) * sign(gl_Position.z);
```
slimbuck, line 54: No need for * sign at the end

PR #7049, opened 2024-10-18
Gizmo material chunks shader fix

kpal81xd:
Fixed issue listed in https://github.com/playcanvas/engine/pull/7009 regarding ensuring debug output is not used for gizmos

### DIFF
--- a/src/extras/gizmo/shape/shape.js
+++ b/src/extras/gizmo/shape/shape.js
@@ -301,7 +301,9 @@ class Shape {
         material.shader = shader;
         material.cull = this._cull;
         material.blendType = BLEND_NORMAL;
-        material.chunks.debugOutputPS = '';   // do not apply debug output shader code to gizmo
+        if (material.chunks) {
+            material.chunks.debugOutputPS = '';   // do not apply debug output shader code to gizmo
+        }
         material.update();
 
         const meshInstances = [];


mvaligursky (approved):

willeastcott (approved):

src/extras/gizmo/shape/shape.js
```diff
@@ -301,7 +301,9 @@ class Shape {
         material.shader = shader;
         material.cull = this._cull;
         material.blendType = BLEND_NORMAL;
-        material.chunks.debugOutputPS = '';   // do not apply debug output shader code to gizmo
+        if (material.chunks) {
```
LeXXik, line 304: `material.chunks?.debugOutputPS = '';` ?

mvaligursky: Screenshot 2024-10-18 at 12 33 24

mvaligursky: this would try to do `undefined = '';`

LeXXik: Does debugOutputPS exist only in debug build?

mvaligursky: there is no difference between builds. But some materials don't override chunks so chunks is undefined.

PR #7051, opened 2024-10-21
Rendering to a single mip level of RenderTarget (WebGPU only)

mvaligursky:
- added support for rendering to a mipLevel of a texture
- due to some complications on WebGL (could not get it to work), this is currently supported on WebGPU only
- changes to the way RT evaluates its width and height to take this into account, and return size of a rendered mip level to handle viewport set up for it

test screenshot - rendering to mipLevel 1 of the render target, the others are untouched (black)
Screenshot 2024-10-21 at 12 25 54

### DIFF
--- a/src/platform/graphics/render-target.js
+++ b/src/platform/graphics/render-target.js
@@ -82,7 +82,7 @@ class RenderTarget {
     _mipLevel;
 
     /**
-     * True if the mipmaps should b automatically generated for the color buffer(s) if it contains
+     * True if the mipmaps should be automatically generated for the color buffer(s) if it contains
      * a mip chain.
      *
      * @type {boolean}
--- a/src/platform/graphics/render-target.js
+++ b/src/platform/graphics/render-target.js
@@ -327,7 +327,7 @@ class RenderTarget {
     resize(width, height) {
 
         if (this.mipLevel > 0) {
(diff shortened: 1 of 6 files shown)


src/platform/graphics/render-target.js
```diff
@@ -74,6 +75,33 @@ class RenderTarget {
...
     _face;
 
+    /**
+     * @type {number}
+     * @private
+     */
+    _mipLevel;
+
+    /**
+     * True if the mipmaps should b automatically generated for the color buffer(s) if it contains
```
willeastcott, line 85:
```suggestion
     * True if the mipmaps should be automatically generated for the color buffer(s) if it contains
```

willeastcott (approved):

willeastcott:
I'm very surprised this doesn't work on WebGL. Do you have much confidence we can solve that?

mvaligursky:
> I'm very surprised this doesn't work on WebGL. Do you have much confidence we can solve that?

It should be doable, based on online PRs / discussions, this was fixed in ANGLE and Chrome in 2020.

mvaligursky:
actually, after some testing I've confirmed it works on WebGL 2 as well.
The difference was that the minFilter was set to pc.FILTER_LINEAR, which:
- on WebGL2, disables using mipmaps and only uses mipmap 0
- on WebGPU, still uses the nearest mipmap, as there is no equivalent to no mipmaps

PR #7061, opened 2024-10-24
Fixed examples device type logging

### LINKED ISSUE #7019: Tiny examples warning that should not be displayed
Repro steps:

- run any example, for example HelloWorld, in WebGPU mode
- switch to Compute -> Histogram example (which is only supported in WebGPU mode)
- it console log has warning:  `WebGL is disabled. Using WebGPU instead.`
- Ideally this should not be logged, as WebGL is not even considered as we're in persistent WebGPU mode from the previous example.

Tiny issue, but logging anyways.

kpal81xd:
Fixes #7019

### DIFF
--- a/examples/iframe/utils.mjs
+++ b/examples/iframe/utils.mjs
@@ -113,7 +113,7 @@ export function updateDeviceType(config) {
     deviceType = DEVICE_TYPES.includes(savedDevice) ? savedDevice : 'webgl2';
 
     if (params.deviceType && DEVICE_TYPES.includes(params.deviceType)) {
-        console.warn("Overwriting default device: ", params.deviceType);
+        console.warn("Overriding default device: ", params.deviceType);
         deviceType = params.deviceType;
         return;
     }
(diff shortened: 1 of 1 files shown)


examples/iframe/utils.mjs
```diff
@@ -109,30 +109,29 @@ export let deviceType = 'webgl2';
  * @param {{ WEBGPU_DISABLED: boolean; WEBGL_DISABLED: boolean; }} config - The configuration object.
  */
 export function updateDeviceType(config) {
+    const savedDevice = localStorage.getItem('preferredGraphicsDevice') ?? 'webgl2';
+    deviceType = DEVICE_TYPES.includes(savedDevice) ? savedDevice : 'webgl2';
+
     if (params.deviceType && DEVICE_TYPES.includes(params.deviceType)) {
-        console.warn("Overwriting default deviceType from URL: ", params.deviceType);
+        console.warn("Overwriting default device: ", params.deviceType);
```
willeastcott, line 116: Should this say 'overriding'?

willeastcott (approved):

PR #7066, opened 2024-10-25
Compatibility of engine v1 with data format of engine v2

mvaligursky:
- In the engine v2 standard material data json file, the following tint parameters are being removed by the editor, and are always assumed to be true. If the engine v1 finds these parameters missing in the json file, they are set to true.

```
ambientTint
emissiveTint
diffuseTint
sheenTint
```

- `StandardMaterial.useGamma` gets forwarded to `StandardMaterial.useGammaTonemap`
- using two files to separate the code from the engine as much as possible while avoiding circular references
- few tint parameters now default to true to match the v2 behaviour and those flags being removed from the material asset by the Editor migration soon


kpal81xd (approved):

kpal81xd (approved):
Make sure unit tests pass

slimbuck (approved):
nice!

mvaligursky:
> Make sure unit tests pass

Of course, that blocks the merge.

PR #7072, opened 2024-10-28
Gizmo rotation fix for orthographic

kpal81xd:
- Fixes and simplifies ray creation for transform gizmo


slimbuck (approved):
Approving, but please address lint issues before merge.

PR #7079, opened 2024-10-30
[BETA] ESM Grid Renderer Script

kpal81xd:
Ports grid into a script from supersplat editor with changes:
- Resolution `resolution: GridRenderer.RESOLUTION_HIGH | GridRenderer.RESOLUTION_MEDIUM | GridRenderer.RESOLUTION_LOW`
- Half extents `halfExtents: Vec2`
- Grid can be recolored `colorX: Color`, `colorZ: Color`

### Preview 
image


examples/src/examples/gizmos/transform-rotate.example.mjs
```diff
@@ -71,6 +73,7 @@ camera.addComponent('camera', {
     farClip: 1000
 });
 camera.addComponent('script');
+camera.script.create(InfiniteGrid);
```
willeastcott, line 76: Wait, why would the grid be tagged on the camera?

scripts/infinite-grid.js
```diff
@@ -0,0 +1,278 @@
...
+            gl_FragColor = vec4(vec3(0.7), levelAlpha);
+            gl_FragDepth = writeDepth(levelAlpha) ? calcDepth(pos) : 1.0;
+            return;
+        }
+
+        discard;
+    }
+`;
+
+class InfiniteGrid extends Script {
```
willeastcott, line 176: Hey @slimbuck - what sort of configuration/customization options could we provide here?

scripts/infinite-grid.js
```diff
@@ -0,0 +1,278 @@
...
+        levelSize = 2.0 / 1000.0;
+        levelAlpha = pristineGrid(levelPos.xz, ddx * 0.1, ddy * 0.1, vec2(levelSize)) * fade;
+        if (levelAlpha > epsilon) {
+            vec3 color;
+            vec2 loc = abs(levelPos.xz);
+            if (loc.x < levelSize) {
+                if (loc.y < levelSize) {
+                    color = vec3(1.0);
+                } else {
+                    color = vec3(0.2, 0.2, 1.0);
```
marklundin, line 140: We might want to expose these colours as attributes

kpal81xd: Oh true will need to recreate the shader then

scripts/infinite-grid.js
kpal81xd: I could add a size option which includes infinite for smaller grids

examples/src/examples/gizmos/transform-rotate.example.mjs
kpal81xd: Oh it needs the camera for the view projection matrix and for switching between perspective and orthographic modes

willeastcott: @mvaligursky or @slimbuck might be able to advise here but I would have expected the implementation to maybe add the grid to a new layer and in a callback before rendering the layer, get the current camera info. I'm all for making script usage easy and guessable...and I would not have guessed that a grid would need to be tagged on an entity that also has a camera component.

kpal81xd: Yea I think the API for it specifically needs to be adjusted for sure.

marklundin:
Just FYI, the grid breaksup over these objects

image

kpal81xd:
> Just FYI, the grid breaksup over these objects
> 
> image

Yea so thats the dithering alpha @slimbuck we should prob add a flag for this what do you think?

slimbuck:
Make sure the grid is the last opaque object rendered. Then only semitrans objects will expose the dithering.

marklundin:
The grid doesn't scale correctly when the screen aspect ratio changes

https://github.com/user-attachments/assets/490eb267-daf9-49e4-b4d6-26b5b59b2aa2

marklundin:
> The grid doesn't scale correctly when the screen aspect ratio changes
> 
>  Screen.Recording.2024-11-01.at.15.11.30.mov

This seems to be related to getFrustumCorners! And I've just seen the corresponding fix https://github.com/playcanvas/engine/pull/7077/files

kpal81xd:
> > The grid doesn't scale correctly when the screen aspect ratio changes
> > Screen.Recording.2024-11-01.at.15.11.30.mov
> 
> This seems to be related to getFrustumCorners! And I've just seen the corresponding fix https://github.com/playcanvas/engine/pull/7077/files

@slimbuck

scripts/grid.js
```diff
@@ -0,0 +1,386 @@
...
+        return this._colorZ;
+    }
+
+    /**
+     * @param {CameraComponent} camera - The camera component.
+     */
+    attach(camera) {
+        this._camera = camera;
+
+        this._prerender = () => {
```
mvaligursky, line 331: any reason this prerender call does not execute inside onPreRenderLayer? It belongs there just before the drawing itself, to guarantee no other code can change it inbetween.

scripts/grid.js
```diff
@@ -0,0 +1,386 @@
...
+        this._device = this.app.graphicsDevice;
+
+        // create shader
+        const shader = createShaderFromCode(this._device, vertexShader, fragmentShader, 'grid', {
+            vertex_position: SEMANTIC_POSITION
+        });
+        this._quadRender = new QuadRender(shader);
+
+        // set initial size
+        this._set('half_extents', this._halfExtents);
```
mvaligursky, line 257: these _set calls need to execute inside the onPreRenderLayer as well

scripts/grid.js
```diff
@@ -0,0 +1,386 @@
+import {
```
willeastcott, line 1: Does this file not have to have a `.mjs` extension? If you load a script asset via the engine's asset system, it make the assumption a `.js` is non-ESM and `.mjs` is ESM. There are separate code paths. cc @marklundin

scripts/grid.js
```diff
@@ -0,0 +1,386 @@
...
+    /**
+     * Set the value of a uniform in the shader.
+     *
+     * @param {string} name - The name of the uniform.
+     * @param {Color|Vec3|number} value - The value to set.
+     * @private
+     */
+    _set(name, value) {
+        if (value instanceof Color) {
+            this._device.scope.resolve(name).setValue([value.r, value.g, value.b]);
```
marklundin, line 273: @mvaligursky there's no performance impact in polling `resolve()`  here right?

mvaligursky: resolve is just a lookup into a Map, so not much cost.

MAG-AdrianMeredith:
> The grid doesn't scale correctly when the screen aspect ratio changes
> 
>  Screen.Recording.2024-11-01.at.15.11.30.mov

Is the grid being affected by postprocessing here?

marklundin:
Yep, it's got some post on it, but unrelated to the scaling issue which was to do with the frustum calculation

MAG-AdrianMeredith:
> Yep, it's got some post on it, but unrelated to the scaling issue which was to do with the frustum calculation

But is it meant to?

mvaligursky:
> But is it meant to?

Sure, whatever renders in the world gets post-processed.

mvaligursky:
@kpal81xd - keen to get this ready for the release?

kpal81xd:
> @kpal81xd - keen to get this ready for the release?

Id say no since I have another PR for a grid that uses a mesh instead of being just a renderer WIP so I want to evaluate the use case of both

scripts/esm/grid-renderer.mjs
```diff
@@ -309,14 +309,11 @@ class GridRenderer extends Script {
 
         const targetLayer = layer ?? GridRenderer.createLayer(this.app, undefined, 1);
 
-        const camera = this.entity.camera;
-        if (!camera) {
-            throw new Error('GridRenderer script must be attached to a camera entity.');
-        }
+        this.app.scene.on('prerender:layer', (camera, layer, transparent) => {
+            if (!camera.layers.includes(targetLayer.id)) {
+                camera.layers = camera.layers.concat(targetLayer.id);
```
mvaligursky, line 314: I really don't like changing camera.layers from inside the rendering callback.
That means at least one frame will be rendered incorrectly, and could even cause some crashes, as for example meshes on that layer were not culled for the camera, but in the middle of the frame, camera new needs to render them.

kpal81xd: Ahh i see what would you suggest I change it to instead?

kpal81xd: So ive currently changed it to store the cameras in the prerender but update the layers in an update loop. Would that solve the issue?

kpal81xd: I guess theres no other way then and the grid has to be attached to the camera

mvaligursky: Or the grid could be assigned a layer, that's what it really needs right? So that the mesh instance is added to that layer.
Inside the preRender callback, a uniform with the camera properties can be changed on the material, I believe there was this need as well.

kpal81xd: So the grid does have a layer set so I could add a hook on there but does that pass the camera as parameter?

Not sure what you mean about the material tho

mvaligursky: this one does:
https://github.com/playcanvas/engine/blob/c26ecf6ac9195b559dfa9f03df8f1ceda7be0e61/src/scene/scene.js#L109-L120

mvaligursky: I thought you needed to set some camera view matrix or similar uniform, not entirely sure. Ignore if not.

kpal81xd: Not set I just need to get the projection and the frustum corners

PR #7081, opened 2024-10-31
View Cube extra

kpal81xd:
- Ported view cube from supersplat to extras
- Refactored and added customization to components

### API
```ts
colorX: Color;
colorY: Color
colorZ: Color;
colorNeg: Color;
radius: number;
textSize: number;
lineThickness: number;
lineLength: number;
```

### Preview
image


src/extras/gizmo/view-cube.js
```diff
@@ -0,0 +1,457 @@
+import { EventHandler } from '../../core/event-handler.js';
```
willeastcott, line 1: Should `view-cube.js` really be in the gizmo folder? It's not technically a gizmo (doesn't subclass from it).

willeastcott (approved):

src/extras/gizmo/view-cube.js
kpal81xd: I put it in there since it is kind of a gizmo since it allows you to move the camera in the scene but doesnt extend the gizmo class itself. Also the folder contains more than just the raw gizmo classes themselves.

PR #7082, opened 2024-10-31
[BETA] ESM Orbit Camera

kpal81xd:
- Renamed old orbit camera to `legacy-orbit-camera.js`
- Added new orbit camera dervived from the multi-camera

### Preview
image


kpal81xd:
Its likely that I will follow up this PR with the Multi-camera extending this class since they are very similar

kpal81xd:
Closing as new orbit camera example part of GH-7111

Issue #5856, opened 2023-11-27
Application cannot start in Firefox


LeXXik:
Forum: https://forum.playcanvas.com/t/back-screen-in-latest-firefox-version-120-0-playcanvas-version-166-3/33964/1

Repro: https://playcanvas.com/project/1167343/overview/roll-a-ball

Launch a game in latest Firefox. A black screen is shown.

Console:
![image](https://github.com/playcanvas/engine/assets/5677782/75d60357-c5c5-4b4c-ae0b-4bfb47c66179)

marklundin:
Is this FF120? I can't replicate

LeXXik:
![image](https://github.com/playcanvas/engine/assets/5677782/3ec20307-4ab8-4bd1-a166-ae5e7cb9111c)

kungfooman:
Interesting, I just suggested PlayCanvas to someone yesterday and he mentioned the same issue, but it worked with WebGL1 for him :thinking:

mvaligursky:
I'm on Mac and it seems to work for me. Is this a Window issue perhaps?

<img width="727" alt="Screenshot 2023-11-27 at 10 43 00" src="https://github.com/playcanvas/engine/assets/59932779/84296ee6-5d34-4dd7-b686-295d6d8ee4cc">

LeXXik:
I've tried to run under Ubuntu (WSL in Windows) and it runs fine.

marklundin:
OK I can replicate this on Windows FF120. Works in engine 165.5, broken in 1.66.2

LeXXik:
Engine examples work fine, except these (black screen as well):
positional
mini stats
obj
video texture
transform feedback
point cloud
point cloud simulation
particles snow / random sprites / anim index
paint mesh
orbit
fly
first person

These take very long time to load:
shader compile
lights physical units

mvaligursky:
I can repro on Windows as well, Firefox 120. I captured this in Spector, and under the hood rendering seems ok to a multi-sampled buffer. At the very end, we resolve this multi-sampled buffer to a single sampled default backbuffer (NULL), maybe that does not work for some reason in FF on Windows. API calls to do it seem correct to me.

![Screenshot 2023-11-27 at 11 01 55](https://github.com/playcanvas/engine/assets/59932779/c927c0e0-cfa2-458b-bd13-6cf287b8dd55)

I suspect that if the multi-sampling was disabled, this would work.

mvaligursky:
[...]

PR #7096, opened 2024-11-06
Disable MSAA on Android Firefox version newer than 123.0

mvaligursky:
related to Disable MSAA on Android Firefox version newer than 123.0

updated the workaround to disable MSAA in this case as well


slimbuck (approved):

willeastcott:
Is a bug logged against FF?

mvaligursky:
yes, details in https://github.com/playcanvas/engine/issues/5856

PR #7100, opened 2024-11-07
Patch to migrate material to default to physical v1

kpal81xd:
- Sets Engine V1 material to default to physical


mvaligursky (approved):
looks good, so a missing shaderModel is the problem as the data.shader was removed on the Editor side right?

kpal81xd:
> looks good, so a missing shaderModel is the problem as the data.shader was removed on the Editor side right?

Yea all materials were assumed to be Phong so this fixes the issue

PR #7109, opened 2024-11-13
Batcher buffers merging optimization

Maksims:
Batcher consists of multiple parts:
1. Collecting information on batch groups.
2. Figuring out individual batches based on mesh differences.
3. Merging meshes buffers with additional transformation.

This PR focuses on the third case by decomposing some math functions that are called frequently, which results in significant performance gains in some of our tests.

In a scene with 749 batches, on test machine currently it takes:
~1658ms to apply position transformations and 612ms on normals transformations.

With this PR these numbers become:
~216ms to apply position transformations and 155ms on normals transformations.

In this specific case we have **x7.7 time speed up** on positions and **x3.9 times speed up** on normals.
We've tested multiple cases, and gains can vary from x1.5 to x10, depending on number of meshes per batch and how large these meshes are.

### DIFF
--- a/src/scene/batching/batch-manager.js
+++ b/src/scene/batching/batch-manager.js
@@ -706,24 +706,33 @@ class BatchManager {
                         // transform position, normal and tangent to world space
                         if (!dynamic && stream.numComponents >= 3) {
                             if (semantic === SEMANTIC_POSITION) {
+                                const m = transform.data;
+                                let x, y, z;
+
                                 for (let j = 0; j < totalComponents; j += stream.numComponents) {
-                                    vec.set(subarray[j], subarray[j + 1], subarray[j + 2]);
-                                    transform.transformPoint(vec, vec);
-                                    subarray[j] = vec.x;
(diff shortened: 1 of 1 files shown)


src/scene/batching/batch-manager.js
```diff
@@ -706,24 +706,33 @@ class BatchManager {
...
-                                    vec.set(subarray[j], subarray[j + 1], subarray[j + 2]);
-                                    transform.transformPoint(vec, vec);
-                                    subarray[j] = vec.x;
-                                    subarray[j + 1] = vec.y;
-                                    subarray[j + 2] = vec.z;
+                                    x = subarray[j];
+                                    y = subarray[j + 1];
+                                    z = subarray[j + 2];
+
+                                    subarray[j] = x * m[0] + y * m[4] + z * m[8] + m[12];
```
mvaligursky, line 717: looks great, just please add comment that this is Mat4.transformPoint, and same for the one bellow.

mvaligursky:
cherry picked to engine v2 as well

PR #7111, opened 2024-11-14
Multicam API updates

kpal81xd:
- Added attributes to enable orbit/fly/pan controls `enableOrbit: boolean; enablePan: boolean; enableFly: boolean`
- Set default focus to use camera position as start and origin as focus point
- Added smoothing parameter to `focus` and `resetZoom`
- Added `refocus` method for simpler focusing API `refocus(point: Vec3, start?: Vec3 | null = null, zoomDist?: number | null = null, smoothing: boolean = true)`
- Added attributes `pitchMin: number` and `pitchMax: number` for constraining the pitch angles
- Added attributes `zoomDistMin: number` and `zoomDistMax: number` for constraining the absolute zoom distance 

### Preview
image

### DIFF
--- a/examples/src/examples/camera/multi.controls.mjs
+++ b/examples/src/examples/camera/multi.controls.mjs
@@ -72,8 +72,9 @@ export const controls = ({ observer, ReactPCUI, React, jsx, fragment }) => {
                     binding: new BindingTwoWay(),
                     link: { observer, path: 'attr.lookDamping' },
                     min: 0,
-                    max: 0.99,
-                    step: 0.01
+                    max: 0.999,
+                    step: 0.001,
+                    precision: 3
                 })
             ),
             jsx(
@@ -83,8 +84,9 @@ export const controls = ({ observer, ReactPCUI, React, jsx, fragment }) => {
                     binding: new BindingTwoWay(),
                     link: { observer, path: 'attr.moveDamping' },
                     min: 0,
-                    max: 0.99,
-                    step: 0.01
+                    max: 0.999,
+                    step: 0.001,
+                    precision: 3
                 })
             ),
             jsx(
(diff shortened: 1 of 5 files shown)


willeastcott:
Feedback:

* pitch should go from 90 looking down y axis to -90 looking up y axis.
* `pitchMin` -> `minPitch` and `pitchMax` -> `maxPitch`
* `focus` attribute doesn't stick if applied after the script create.

willeastcott:
Suggestion: rename from `MultiCamera` to the more generic and obvious `CameraControls`.

scripts/camera-controls.mjs
```diff
@@ -0,0 +1,935 @@
...
+        this._camera = camera;
+        this._camera.entity.setLocalEulerAngles(0, 0, 0);
+
+        window.addEventListener('wheel', this._onWheel, PASSIVE);
+        window.addEventListener('keydown', this._onKeyDown, false);
+        window.addEventListener('keyup', this._onKeyUp, false);
+        window.addEventListener('pointerdown', this._onPointerDown);
+        window.addEventListener('pointermove', this._onPointerMove);
+        window.addEventListener('pointerup', this._onPointerUp);
+        window.addEventListener('contextmenu', this._onContextMenu);
```
willeastcott, line 876: I have a suspicion that you don't want to be adding listeners to the window here, at least for the pointer events. The current examples browser creates a `pc.Mouse` that is attached to `document.body`. For the Web Components examples browser, the page runs each example in an iframe and on iOS, it was not possible to use `CameraControls`. So I submitted this commit to listen on the canvas element instead:
https://github.com/playcanvas/web-components/commit/d3e83be09b7ddc512a9f758a1778bed52cf98e4b
I'm not sure if `document.body` would have worked too, but yeah, there do seem to be problems with just using `window`.

slimbuck: Could be good if user was also able to override the default element.

willeastcott (approved):
Very cool!

PR #7127, opened 2024-11-20
Support compressed.ply spherical harmonics

slimbuck:
This PR:
- adds support for loading and rendering compressed spherical harmonic data in `compressed.ply` files
- the details of the format will be described elsewhere with the release in SuperSplat, but in summary:
    - 3 palettes of spherical harmonics coefficients are stored per file (one for each sh band)
    - each splat stores 9 indices into this data (r,g,b indices for band 1, 2, 3), totalling 128bits
- this results in compressed files being roughly twice the size with 3 bands of spherical harmonics
- small type fix for specifying texture data

Notes:
- we are adding support here for 0 or 3 bands of spherical harmonics, but may add explicit support for 1 and 2 bands in future
- the palette data is stored in halfs for the space savings, but it might be preferable to use floats

### DIFF
--- a/src/platform/graphics/texture.js
+++ b/src/platform/graphics/texture.js
@@ -186,7 +186,7 @@ class Texture {
      * - {@link FUNC_NOTEQUAL}
      *
      * Defaults to {@link FUNC_LESS}.
-     * @param {Uint8Array[]|HTMLCanvasElement[]|HTMLImageElement[]|HTMLVideoElement[]|Uint8Array[][]} [options.levels]
+     * @param {Uint8Array[]|Uint16Array[]|Uint32Array|HTMLCanvasElement[]|HTMLImageElement[]|HTMLVideoElement[]|Uint8Array[][]} [options.levels]
      * - Array of Uint8Array or other supported browser interface; or a two-dimensional array
      * of Uint8Array if options.arrayLength is defined and greater than zero.
      * @param {boolean} [options.storage] - Defines if texture can be used as a storage texture by
--- a/src/platform/graphics/texture.js
+++ b/src/platform/graphics/texture.js
@@ -186,7 +186,7 @@ class Texture {
      * - {@link FUNC_NOTEQUAL}
      *
      * Defaults to {@link FUNC_LESS}.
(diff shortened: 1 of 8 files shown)


willeastcott (approved):
What an achievement. Is it possible to put some examples numbers in the PR....along with info about expected quality loss.

src/platform/graphics/texture.js
```diff
@@ -186,7 +186,7 @@ class Texture {
      * - {@link FUNC_NOTEQUAL}
      *
      * Defaults to {@link FUNC_LESS}.
-     * @param {Uint8Array[]|HTMLCanvasElement[]|HTMLImageElement[]|HTMLVideoElement[]|Uint8Array[][]} [options.levels]
+     * @param {Uint8Array[]|Uint16Array[]|Uint32Array[]|Float32Array[]|HTMLCanvasElement[]|HTMLImageElement[]|HTMLVideoElement[]|Uint8Array[][]} [options.levels]
```
mvaligursky, line 189: We could just use `ArrayBufferView` which represents any typed array. This is what mesh API for vertex data uses:
`     * @param {number[]|ArrayBufferView} data - Vertex data for the specified semantic.`

src/scene/gsplat/gsplat-compressed-data.js
```diff
@@ -9,9 +9,34 @@ import { GSplatData } from './gsplat-data.js';
 
 const SH_C0 = 0.28209479177387814;
 
+// https://fgiesen.wordpress.com/2012/03/28/half-to-float-done-quic/
```
mvaligursky, line 12: maybe should be moved to FloatPacking module, along the `float2Half`

mvaligursky (approved):

src/scene/gsplat/gsplat-compressed-data.js
slimbuck: done

slimbuck:
After more thorough testing it seems the palette approach results in much more complication with arguable advantage. Closing this PR in favour of a simpler approach that results in larger files, but does not suffer from color degradation.

PR #7128, opened 2024-11-21
Make camera controls script play nice with XR

willeastcott:
* Disable script update when XR session is active
* Restore camera pos/rot when exiting XR


slimbuck (approved):

mvaligursky (approved):

mvaligursky:
@willeastcott - good to merge or other plans with this?

willeastcott:
> @willeastcott - good to merge or other plans with this?

Needs testing - I'll do that today.

scripts/esm/camera-controls.mjs
```diff
@@ -315,6 +315,23 @@ class CameraControls extends Script {
...
+        const position = new Vec3();
+        const rotation = new Quat();
+        this.app.xr.on('start', () => {
+            // Store the position and rotation of the camera
+            position.copy(this.entity.getPosition());
+            rotation.copy(this.entity.getRotation());
+
+            // Set the camera's position to the ground
+            const pos = this.entity.getPosition();
+            this.entity.setPosition(pos.x, 0, pos.z);
```
Maksims, line 328: XR will override this position on every update, as now there is no parent node, offsetting the camera from origin is only done by moving in XR. So this line won't do anything.
For offsetting position in XR, a parent node is required.

willeastcott: OK, I have removed that for now. Everything else OK?

Maksims: This also implies that initial setup of a scene, should have camera at the origin, and content somewhere in front of it, to make sure when entering in XR we are not inside of a content.

willeastcott: How should we handle that then? I don't want to reintroduce the magical hidden parent node in the `camera-controls.mjs` script again...

Maksims: Do a proper parent node for the camera entity. And only use it for XR (not camera orbiting). And reset it when exiting XR.

willeastcott: So I should create a new script called xr-navigation.mjs I guess...

PR #7129, opened 2024-11-21
CameraFrame API in the engine

mvaligursky:
Fixes https://github.com/playcanvas/engine/pull/7129
(separate improvements / changes can be tracked separately)

This adds a new CameraFrame class as part of the engine, which is a public API for the currently exposed render pass based rendering.
The script which exposes this to the Editor now uses an instance of this CameraFrame, and simply passes attributes to it, without containing any logic. This allows it to not access any private engine functionality.

This is the new public API (too long to copy here)
https://github.com/playcanvas/engine/blob/mc-engine-camera-frame/src/extras/render-passes/camera-frame.js

- also updated engine examples to use CameraFrame module directly, without using the script.
- updated eslint config to ignore some custom tags we added to MJS scripts

### DIFF
--- a/src/extras/render-passes/camera-frame.js
+++ b/src/extras/render-passes/camera-frame.js
@@ -28,7 +28,10 @@ import { CameraFrameOptions, RenderPassCameraFrame } from './render-pass-camera-
  * - {@link TONEMAP_ACES2}
  * - {@link TONEMAP_NEUTRAL}
  *
- * @property {number} sharpness - The sharpness of the frame buffer, 0-1 range. Defaults to 0.
+ * @property {number} sharpness - The sharpening intensity, 0-1 range. This can be used to increase
+ * the sharpness of the rendered image. Often used to counteract the blurriness of the TAA effect,
+ * but also blurriness caused by rendering to a lower resolution render target by using
+ * rendering.renderTargetScale property. Defaults to 0.
  */
 
 /**
@@ -92,7 +95,10 @@ import { CameraFrameOptions, RenderPassCameraFrame } from './render-pass-camera-
 /**
  * @typedef {Object} Taa
  * @property {boolean} enabled - Whether Taa is enabled. Defaults to false.
- * @property {number} jitter - The intensity of the camera jitter, 0-1 range. Defaults to 1.
+ * @property {number} jitter - The intensity of the camera jitter, 0-1 range. The larger the value,
+ * the more jitter is applied to the camera, making the anti-aliasing effect more pronounced. This
+ * also makes the image more blurry, and rendering.sharpness parameter can be used to counteract.
+ * Defaults to 1.
  */
 
 /**
--- a/src/extras/render-passes/camera-frame.js
+++ b/src/extras/render-passes/camera-frame.js
@@ -69,9 +69,18 @@ import { CameraFrameOptions, RenderPassCameraFrame } from './render-pass-camera-
  * @typedef {Object} Vignette
  * @property {number} intensity - The intensity of the vignette effect, 0-1 range. Defaults to 0,
  * making it disabled.
- * @property {number} inner - The inner radius of the vignette effect, 0-3 range. Defaults to 0.5.
- * @property {number} outer - The outer radius of the vignette effect, 0-3 range. Defaults to 1.
- * @property {number} curvature - The curvature of the vignette effect, 0.01-10 range. Defaults to 0.5.
+ * @property {number} inner - The inner distance of the vignette effect measured from the center of
+ * the screen, 0-3 range. This is where the vignette effect starts. Value larger than 1 represents
+ * the value off screen, which allows more control. Defaults to 0.5, representing half the distance
+ * from center.
+ * @property {number} outer - The outer distance of the vignette effect measured from the center of
+ * the screen, 0-3 range. This is where the vignette reaches full intensity. Value larger than 1
+ * represents the value off screen, which allows more control. Defaults to 1, representing the full
+ * screen.
+ * @property {number} curvature - The curvature of the vignette effect, 0.01-10 range. The vignette
+ * is rendered using a rectangle with rounded corners, and this parameter controls the curvature of
+ * the corners. Value of 1 represents a circle. Smaller values make the corners more square, while
+ * larger values make them more rounded. Defaults to 0.5.
  */
 
 /**
--- a/src/extras/render-passes/camera-frame.js
+++ b/src/extras/render-passes/camera-frame.js
@@ -1,0 +1,331 @@
+import { Debug } from '../../core/debug.js';
+import { Color } from '../../core/math/color.js';
+import { math } from '../../core/math/math.js';
+import { PIXELFORMAT_111110F, PIXELFORMAT_RGBA16F, PIXELFORMAT_RGBA32F } from '../../platform/graphics/constants.js';
+import { SSAOTYPE_NONE } from './constants.js';
+import { CameraFrameOptions, RenderPassCameraFrame } from './render-pass-camera-frame.js';
+
+/**
+ * @import { AppBase } from '../../framework/app-base.js'
+ * @import { CameraComponent } from '../../framework/components/camera/component.js'
+ */
+/**
+ * @typedef {Object} Rendering
+ * @property {number[]} renderFormats - The render formats of the frame buffer, in order of
+ * preference. Defaults to [{@link PIXELFORMAT_111110F}, {@link PIXELFORMAT_RGBA16F}, 
+ * {@link PIXELFORMAT_RGBA32F}].
+ * @property {boolean} stencil - Whether the render buffer has a stencil buffer. Defaults to false.
+ * @property {number} renderTargetScale - The scale of the frame buffer, 0.1-1 range. Defaults to 1.
+ * @property {number} samples - The number of samples of the render buffer, 1-4 range. Defaults to 1.
+ * @property {boolean} sceneColorMap - Whether rendering generates a scene color map. Defaults to false.
+ * @property {boolean} sceneDepthMap - Whether rendering generates a scene depth map. Defaults to false.
+ * @property {number} toneMapping - The tone mapping. Defaults to {@link ToneMapping.LINEAR}. Can be:
+ *
+ * - {@link TONEMAP_LINEAR}
+ * - {@link TONEMAP_FILMIC}
+ * - {@link TONEMAP_HEJL}
+ * - {@link TONEMAP_ACES}
+ * - {@link TONEMAP_ACES2}
+ * - {@link TONEMAP_NEUTRAL}
+ *
+ * @property {number} sharpness - The sharpness of the frame buffer, 0-1 range. Defaults to 0.
+ */
+
+/**
+ * @typedef {Object} Ssao
+ * @property {string} type - The type of the SSAO determines how it is applied in the rendering
+ * process. Defaults to {@link SSAOTYPE_NONE}. Can be:
+ *
+ * - {@link SSAOTYPE_NONE}
+ * - {@link SSAOTYPE_LIGHTING}
+ * - {@link SSAOTYPE_COMBINE}
+ * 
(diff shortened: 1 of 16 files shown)


slimbuck (approved):

src/extras/render-passes/camera-frame.js
```diff
@@ -0,0 +1,331 @@
...
+ * @property {number} jitter - The intensity of the camera jitter, 0-1 range. Defaults to 1.
+ */
+
+/**
+ * Implementation of a simple to use camera rendering pass, which supports SSAO, Bloom and
+ * other rendering effects.
+ *
+ * @category Render Pass
+ */
+class CameraFrame {
```
willeastcott, line 95: I can't help but feel that an `enabled` property per effect would make for a more logical, guessable API.

src/extras/render-passes/camera-frame.js
```diff
@@ -0,0 +1,331 @@
...
+        sharpness: 0.0
+    };
+
+    /**
+     * SSAO settings.
+     *
+     * @type {Ssao}
+     */
+    ssao = {
+        type: SSAOTYPE_NONE,
```
willeastcott, line 118: I personally have trouble understanding/visualizing the 3 SSAO types. I would prefer an `enabled` flag and have it do the most common or nicest looking code path, but provide a further flag along the lines of 'ignoreLighting' or something.

src/extras/render-passes/camera-frame.js
```diff
@@ -0,0 +1,331 @@
...
+    };
+
+    /**
+     * Bloom settings.
+     *
+     * @type {Bloom}
+     */
+    bloom = {
+        intensity: 0,
+        lastMipLevel: 1
```
willeastcott, line 135: Cryptic.

src/extras/render-passes/camera-frame.js
```diff
@@ -0,0 +1,331 @@
...
+ * @property {number} power - The power of the SSAO effect, 0.1-10 range. Defaults to 6.
+ * @property {number} minAngle - The minimum angle of the SSAO effect, 1-90 range. Defaults to 10.
+ * @property {number} scale - The scale of the SSAO effect, 0.5-1 range. Defaults to 1.
+ */
+
+/**
+ * @typedef {Object} Bloom
+ * @property {number} intensity - The intensity of the bloom effect, 0-0.1 range. Defaults to 0,
+ * making it disabled.
+ * @property {number} lastMipLevel - The last mip level of the bloom effect, 0-12 range. Defaults to 1.
```
willeastcott, line 56: I struggle to understand what this will do. Can we expose this in a more obvious/friendly way?

src/extras/render-passes/camera-frame.js
```diff
@@ -0,0 +1,331 @@
...
+
+/**
+ * @typedef {Object} Ssao
+ * @property {string} type - The type of the SSAO determines how it is applied in the rendering
+ * process. Defaults to {@link SSAOTYPE_NONE}. Can be:
+ *
+ * - {@link SSAOTYPE_NONE}
+ * - {@link SSAOTYPE_LIGHTING}
+ * - {@link SSAOTYPE_COMBINE}
+ *
```
willeastcott, line 42: I personally have trouble understanding/visualizing the 3 SSAO types. I would prefer an `enabled` flag and have it do the most common or nicest looking code path, but provide a further flag along the lines of 'ignoreLighting' or something.

src/extras/render-passes/camera-frame.js
```diff
@@ -0,0 +1,331 @@
...
+ * @import { AppBase } from '../../framework/app-base.js'
+ * @import { CameraComponent } from '../../framework/components/camera/component.js'
+ */
+/**
+ * @typedef {Object} Rendering
+ * @property {number[]} renderFormats - The render formats of the frame buffer, in order of
+ * preference. Defaults to [{@link PIXELFORMAT_111110F}, {@link PIXELFORMAT_RGBA16F},
+ * {@link PIXELFORMAT_RGBA32F}].
+ * @property {boolean} stencil - Whether the render buffer has a stencil buffer. Defaults to false.
+ * @property {number} renderTargetScale - The scale of the frame buffer, 0.1-1 range. Defaults to 1.
```
willeastcott, line 18: Not obvious what this does.

src/extras/render-passes/camera-frame.js
```diff
@@ -0,0 +1,331 @@
...
+
+/**
+ * @import { AppBase } from '../../framework/app-base.js'
+ * @import { CameraComponent } from '../../framework/components/camera/component.js'
+ */
+/**
+ * @typedef {Object} Rendering
+ * @property {number[]} renderFormats - The render formats of the frame buffer, in order of
+ * preference. Defaults to [{@link PIXELFORMAT_111110F}, {@link PIXELFORMAT_RGBA16F},
+ * {@link PIXELFORMAT_RGBA32F}].
```
willeastcott, line 16: What influences my choice here? Are these only applicable when the graphics device is created in HDR mode?

src/extras/render-passes/camera-frame.js
```diff
@@ -0,0 +1,331 @@
...
+ * @property {number} contrast - The contrast of the grading effect, 0.5-1.5 range. Defaults to 1.
+ * @property {number} saturation - The saturation of the grading effect, 0-2 range. Defaults to 1.
+ * @property {Color} tint - The tint color of the grading effect. Defaults to white.
+ */
+
+/**
+ * @typedef {Object} Vignette
+ * @property {number} intensity - The intensity of the vignette effect, 0-1 range. Defaults to 0,
+ * making it disabled.
+ * @property {number} inner - The inner radius of the vignette effect, 0-3 range. Defaults to 0.5.
```
willeastcott, line 72: What is a radius of 1? Center of screen to left/right edge? Or to corner?

src/extras/render-passes/camera-frame.js
```diff
@@ -0,0 +1,331 @@
...
+ * @property {number} outer - The outer radius of the vignette effect, 0-3 range. Defaults to 1.
+ * @property {number} curvature - The curvature of the vignette effect, 0.01-10 range. Defaults to 0.5.
```
willeastcott, line 74: I mean, vignettes _are_ round. But I guess that's not what is meant here by curvature...

src/extras/render-passes/camera-frame.js
```diff
@@ -0,0 +1,331 @@
...
+ */
+
+/**
+ * @typedef {Object} Fringing
```
willeastcott, line 78: What does fringing do? What's it for?

src/extras/render-passes/camera-frame.js
```diff
@@ -0,0 +1,331 @@
...
+    /**
+     * @type {RenderPassCameraFrame|null}
+     * @private
+     */
+    renderPassCamera = null;
+
+    /**
+     * Creates a new CameraFrame instance.
+     *
+     * @param {AppBase} app - The application.
```
willeastcott, line 193: Can't you do: `cameraComponent.system.app`?

src/extras/render-passes/camera-frame.js
```diff
@@ -0,0 +1,331 @@
...
+     * @param {CameraComponent} cameraComponent - The camera component.
+     */
+    constructor(app, cameraComponent) {
```
willeastcott, line 196: How come the `CameraFrame` takes a `cameraComponent` instead of doing:

```
entity.camera.frame = new CameraFrame();
```

src/extras/render-passes/camera-frame.js
mvaligursky: updated

src/extras/render-passes/camera-frame.js
mvaligursky: explained. This vignette has control over how round it is.

src/extras/render-passes/camera-frame.js
```diff
@@ -0,0 +1,331 @@
...
+/**
+ * @typedef {Object} Fringing
+ * @property {number} intensity - The intensity of the fringing effect, 0-100 range. Defaults to 0,
+ * making it disabled.
+ */
+
+/**
+ * @typedef {Object} Taa
+ * @property {boolean} enabled - Whether Taa is enabled. Defaults to false.
+ * @property {number} jitter - The intensity of the camera jitter, 0-1 range. Defaults to 1.
```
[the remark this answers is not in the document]
mvaligursky: This controls camera jitter of up to one pixel range. In majority cases this should be set to 1 to get good antialiasing. Allowing smaller version is only if the user prefers sharper image with more aliasing issue.
But to compensate for the loss of sharpness, the sharpness parameter needs to be set by the user, typically to a valud of 1, which represents a single pixel as well, which brings the sharpness (mostly) back.
If the user prefers to have screen size dependency here, they can implement it easily, but this should not be enforced here.
I'll improve the docs to explain.

src/extras/render-passes/camera-frame.js
mvaligursky: CameraFrame is a module used to simplify set up for the user, and manages and assigns render passes to the camera component. Camera component works directly with render passes, not with CameraFrame, as there can be other implementations of a helper class like this, using different frame setup.

src/extras/render-passes/camera-frame.js
mvaligursky: added descriptions to all objects now

src/extras/render-passes/camera-frame.js
mvaligursky: renamed to blurLevel and improved the description, thanks

src/extras/render-passes/camera-frame.js
```diff
@@ -0,0 +1,331 @@
...
+ * @property {number} intensity - The intensity of the SSAO effect, 0-1 range. Defaults to 0.5.
+ * @property {number} radius - The radius of the SSAO effect, 0-100 range. Defaults to 30.
+ * @property {number} samples - The number of samples of the SSAO effect, 1-64 range. Defaults to 12.
+ * @property {number} power - The power of the SSAO effect, 0.1-10 range. Defaults to 6.
+ * @property {number} minAngle - The minimum angle of the SSAO effect, 1-90 range. Defaults to 10.
+ * @property {number} scale - The scale of the SSAO effect, 0.5-1 range. Defaults to 1.
+ */
+
+/**
+ * @typedef {Object} Bloom
```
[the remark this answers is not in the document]
mvaligursky: currently only TAA and Grading has enabled property. Others have some form of intensity parameter, which disables it when set to 0.
It seems adding enabled there would require two parameters change to use it instead of just a single, so I'm not sure I like that.

src/extras/render-passes/camera-frame.js
mvaligursky: I improved the descriptions here:
SSAOTYPE_LIGHTING: 
 * SSAO is applied during the lighting calculation stage, allowing it to blend seamlessly with scene
 * lighting. This results in ambient occlusion being more pronounced in areas where direct light is
 * obstructed, enhancing realism.

SSAOTYPE_COMBINE:
 * SSAO is applied as a standalone effect after the scene is rendered. This method uniformly
 * overlays ambient occlusion across the image, disregarding direct lighting interactions. While
 * this may sacrifice some realism, it can be advantageous for achieving specific artistic styles.

mvaligursky:
I'll merge this for now, further feedback / changes will be address by a follow up PR @willeastcott

src/extras/render-passes/camera-frame.js
MAG-AdrianMeredith: Barging in here, I'd prefer both an enabled property and a type property, defaulting to lighting?  I wouldn't be suprised if theres a different ssao code path in the future so a type field might be nice

mvaligursky: if we implement different types of SSAO, it would be related to the type of generation of ssao texture.
The current type related to the way ssao texture is consumed by other passes.
So we'd add a different type for that I think.

PR #7135, opened 2024-11-23
Beta release of JointComponent

willeastcott:
Releases the `JointComponent` API in beta. For anyone with physics experience, please take a look and offer your thoughts/comments.

Also added a new rope bridge engine example:

https://github.com/user-attachments/assets/aeccddca-2350-4636-89b4-7335dd0a1f78

API reference:

image

### DIFF
--- a/src/framework/components/joint/component.js
+++ b/src/framework/components/joint/component.js
@@ -43,58 +43,99 @@ const properties = [
  * @beta
  */
 class JointComponent extends Component {
+    /** @private */
     _constraint = null;
+    /** @private */
     _entityA = null;
+    /** @private */
     _entityB = null;
+    /** @private */
     _breakForce = 3.4e+38;
+    /** @private */
     _enableCollision = true;
 
     // Linear X degree of freedom
+    /** @private */
     _linearMotionX = MOTION_LOCKED;
+    /** @private */
     _linearLimitsX = new Vec2(0, 0);
+    /** @private */
     _linearSpringX = false;
+    /** @private */
     _linearStiffnessX = 0;
+    /** @private */
     _linearDampingX = 1;
+    /** @private */
     _linearEquilibriumX = 0;
 
     // Linear Y degree of freedom
+    /** @private */
     _linearMotionY = MOTION_LOCKED;
+    /** @private */
     _linearLimitsY = new Vec2(0, 0);
+    /** @private */
     _linearSpringY = false;
+    /** @private */
     _linearStiffnessY = 0;
+    /** @private */
     _linearDampingY = 1;
+    /** @private */
     _linearEquilibriumY = 0;
 
     // Linear Z degree of freedom
+    /** @private */
     _linearMotionZ = MOTION_LOCKED;
+    /** @private */
     _linearLimitsZ = new Vec2(0, 0);
+    /** @private */
     _linearSpringZ = false;
+    /** @private */
     _linearStiffnessZ = 0;
+    /** @private */
     _linearDampingZ = 1;
+    /** @private */
     _linearEquilibriumZ = 0;
 
     // Angular X degree of freedom
+    /** @private */
     _angularMotionX = MOTION_LOCKED;
+    /** @private */
     _angularLimitsX = new Vec2(0, 0);
+    /** @private */
     _angularSpringX = false;
+    /** @private */
     _angularStiffnessX = 0;
+    /** @private */
     _angularDampingX = 1;
+    /** @private */
     _angularEquilibriumX = 0;
 
     // Angular Y degree of freedom
+    /** @private */
     _angularMotionY = MOTION_LOCKED;
+    /** @private */
     _angularLimitsY = new Vec2(0, 0);
+    /** @private */
     _angularSpringY = false;
+    /** @private */
     _angularStiffnessY = 0;
+    /** @private */
     _angularDampingY = 1;
+    /** @private */
     _angularEquilibriumY = 0;
 
     // Angular Z degree of freedom
+    /** @private */
     _angularMotionZ = MOTION_LOCKED;
+    /** @private */
     _angularLimitsZ = new Vec2(0, 0);
+    /** @private */
     _angularSpringZ = false;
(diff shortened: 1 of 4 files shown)


LeXXik:
I always welcome a constraints component, which I think was missing for a while. There are so many cases where it could be used for interactivity. However, I am not sure of the design. Not that it is wrong, but that there is no definite standard for it so far, so it is up to discussion.

For example, if the ball falling on the bridge disables an entity it touches, then I would expect the bridge would behave as if it would have been cut in the middle. However, that is not the case, since a constraint is its own entity and lives its own life. The bridge would probably fall apart with a constraint still alive that is very soft, like a chewing gum.

Also, I remember making a game for one company, where you would draw a wheel for a vehicle to traverse a terrain. Once you drew a wheel, it would replace the old one, which was connected with a hinge constraint. There was an Ammo crash, when PlayCanvas was trying to destroy a rigidbody (due to hierarchy change) before the constraint. There should be a test case to cover that at least.

willeastcott:
@LeXXik The main things I'm still unsure of related to the following:

* Should I collapse the `X`, `Y`, `Z` number properties into `Vec3` properties? e.g. `angularDampingX`, `angularDampingY` and `angularDampingZ` becomes `angularDamping`. This will collapse the API into something much simpler, of course. But I'm wondering whether this might be annoying if you only are concerned with configuring a one or maybe two axes. Maybe there'd be other down sides I'm not considering?
* The current implementation just wraps the `btGeneric6DofSpringConstraint`. Ammo has specific classes for handling other joint types (like hinge, ball-socket, etc). I'm wondering if those specialized joint types are more stable than the generic 6dof joint.

src/framework/components/joint/component.js
```diff
@@ -28,253 +28,1020 @@ const properties = [
...
+        return this._angularStiffnessZ;
+    }
+
+    /**
+     * Sets the breaking force threshold for the constraint.
+     *
+     * @type {number}
+     */
+    set breakForce(force) {
+        if (this._constraint && this._breakForce !== force) {
```
mvaligursky, line 583: maybe this should follow other setters, where the internal value gets updated even if _constraint is null, and only the setBreakingImpulseThreshold does not get called

src/framework/components/joint/component.js
```diff
@@ -28,253 +28,1020 @@ const properties = [
...
 
+    /**
+     * Sets the first entity in the constraint.
+     *
+     * @type {Entity}
+     */
     set entityA(body) {
         this._destroyConstraint();
         this._entityA = body;
         this._createConstraint();
```
mvaligursky, line 626: What's the cost of destroying and creating them? If the component gets initially created, would this execute multiple times? For example in case

```
joint.addComponent('joint', {
            entityA: planks[i],
            entityB: planks[i + 1],
            enableCollision: false
        });
```

mvaligursky (approved):

willeastcott:
After building this PR, I came to a realization. You often want to attach multiple constraints to a single rigid body. For example, imagine the joints attached to the upper torso of a rag doll: neck, left shoulder, right shoulder, mid-spine. Therefore, on refelction, I think it would be better if the `JointComponent` worked similarly to the `SoundCompeont` and supported multiple joints to be defined. So I will try to devise an `addJoint`/`removeJoint` type of API. Stay tuned for a full refactor.

src/framework/components/joint/component.js
willeastcott: This is a good point. When I refactor this stuff, I'll bypass the setters on component creation.

LeXXik:
>  I will try to devise an addJoint/removeJoint type of API

Yeah, I did it in a similar fashion here:
https://github.com/Gamebop/physics/blob/master/src/physics/jolt/front/constraint/component.mjs

Note - in my implementation, I do destroy a constraint connecting both bodies, when one of the bodies is destroyed. Because of that I also need to keep track of which joints connect which bodies. If one entity is connected to another, then both entities would get a constraint component. A constraint component in my case is not a joint in itself, but a host of different joints that can be added to a body.

Maksims:
> After building this PR, I came to a realization. You often want to attach multiple constraints to a single rigid body. For example, imagine the joints attached to the upper torso of a rag doll: neck, left shoulder, right shoulder, mid-spine. Therefore, on reflection, I think it would be better if the `JointComponent` worked similarly to the `SoundComponent` and supported multiple joints to be defined. So I will try to devise an `addJoint`/`removeJoint` type of API. Stay tuned for a full refactor.

But why not use one joint per entity as I've suggested?
So that you create multiple Joint entities and reference needed entityA and entityB. Won't need to do complex UI like Sound Component. And it reduce amount of API needed, and depth complexity of joint component.

willeastcott:
Closing as out of date — superseded by #8891, which landed the full refactor promised in this thread.

That PR resolves everything left open here:

- **Typed joints** — `fixed`/`ball`/`hinge`/`slider`/`6dof` (`JOINTTYPE_*`), each backed by the most appropriate Bullet constraint rather than everything going through `btGeneric6DofSpringConstraint`.
- **Grouped properties** — `Vec3` spring/damping/equilibrium families and `Vec2` limit pairs instead of per-axis `X`/`Y`/`Z` scalars.
- **Component placement** — the standalone joint entity is retained, with its transform defining the joint frame and `entityA`/`entityB` naming the constrained bodies (per Maksims' suggestion above). The `addJoint`/`removeJoint` multi-joint direction was dropped.
- **Destroy-order safety** — constraints now tear down synchronously before any `btRigidBody` is destroyed, via new internal `simulationenabled`/`simulationdisabled` events on `RigidBodyComponent`. This addresses the Ammo crash LeXXik raised, and it's covered by unit tests in `test/framework/components/joint/`.
- Both of mvaligursky's review notes are in the shipped code: the break-impulse setter updates its backing field regardless of whether a constraint exists, and property application during `addComponent` no longer churns the constraint.

The `rope-bridge` example didn't carry over; #8891 added `physics/joints` and `physics/ragdoll` instead.

PR #7137, opened 2024-11-25
EventHandle.off - performance improvements

Maksims:
No logic changes.

Since EventHandle was implemented, developer lives for events management have been simplified. At the same time there is a new opportunity for optimization. This PR does just that, when using EventHandle.off, instead of scanning callbacks and comparing `name`, `callback` and `scope`, now it uses EventHandle itself and checks if it is in array, if so, removes it.

This leads to massive performance improvements when .off is called on EventHandle vs .off on EventHandler.

One of these cases is [onDisable](https://github.com/playcanvas/engine/blob/079caff78cf969825445f0f367b8117f0816c49f/src/framework/components/render/component.js#L827-L830) on RenderComponent (and similar components). On large scenes with a lot of render components `scene._callbacks['set:layers']` - becomes a very large array, the same applies to `app.layers._callbacks`.

In some of our projects, disabling render components leads to a massive lag spike, in one specific project disabling big chunk of a scene takes ~3500ms, with this optimization time is reduces to ~900ms.

To benefit from this optimization, we need to use EventHandle within engine as much as possible, instead of a `off` on EventHandler's.

PR's with EventHandle.off in critical places coming after this one.

### DIFF
--- a/src/core/event-handle.js
+++ b/src/core/event-handle.js
@@ -32,7 +32,7 @@ class EventHandle {
 
     /**
      * @type {string}
-     * @private
+     * @ignore
      */
     name;
 
@@ -81,7 +81,7 @@ class EventHandle {
      */
     off() {
         if (this._removed) return;
-        this.handler.off(this.name, this.callback, this.scope);
+        this.handler.offByHandle(this);
     }
 
     on(name, callback, scope = this) {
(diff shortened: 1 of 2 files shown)


willeastcott (approved):
Happy to merge @mvaligursky?

mvaligursky (approved):
Yep this looks fantastic.

mvaligursky:
also cherry-picked to v2

PR #7138, opened 2024-11-25
Optimization: use EventHandle in Components for scene and layers

Maksims:
Depends on: https://github.com/playcanvas/engine/pull/7137

This PR uses EventHandle.off instead of EventHandler.off to improve performance when disabling renderable components. As a scene and layers callaback arrays can get large in big projects.
In large scenes, disabling renderable components became ~4-8 times faster based on our tests. In one specific scene, disabling half of a scene used to take ~3550ms now it takes ~600ms (~5.9 times improvement).

### DIFF
--- a/src/framework/components/light/component.js
+++ b/src/framework/components/light/component.js
@@ -1247,14 +1247,10 @@ class LightComponent extends Component {
 
         this.light.enabled = true;
 
-        this._evtLayersChanged?.off();
         this._evtLayersChanged = scene.on('set:layers', this.onLayersChanged, this);
 
         if (layers) {
-            this._evtLayerAdded?.off();
             this._evtLayerAdded = layers.on('add', this.onLayerAdded, this);
-
-            this._evtLayerRemoved?.off();
             this._evtLayerRemoved = layers.on('remove', this.onLayerRemoved, this);
         }
 
--- a/src/framework/components/light/component.js
+++ b/src/framework/components/light/component.js
@@ -37,6 +37,24 @@ import { properties } from './data.js';
  */
 class LightComponent extends Component {
     /**
+     * @type {import('../../../core/event-handle.js').EventHandle|null}
+     * @private
+     */
+    _evtLayersChanged = null;
+
+    /**
+     * @type {import('../../../core/event-handle.js').EventHandle|null}
+     * @private
+     */
+    _evtLayerAdded = null;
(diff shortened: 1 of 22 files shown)


willeastcott:
So this is for merging into `main_v`. I guess that's fine - we just cherry-pick afterwards to `main`. But we normally seem to go the other way....

Maksims:
> So this is for merging into `main_v`. I guess that's fine - we just cherry-pick afterwards to `main`. But we normally seem to go the other way....

Editor uses engine v1, and most (except one) of our projects are v1, so we work with it in the first place.

src/framework/components/light/component.js
```diff
@@ -1224,12 +1242,20 @@ class LightComponent extends Component {
...
+        const scene = this.system.app.scene;
+        const layers = scene.layers;
+
         this.light.enabled = true;
 
-        this.system.app.scene.on('set:layers', this.onLayersChanged, this);
-        if (this.system.app.scene.layers) {
-            this.system.app.scene.layers.on('add', this.onLayerAdded, this);
-            this.system.app.scene.layers.on('remove', this.onLayerRemoved, this);
+        this._evtLayersChanged?.off();
```
mvaligursky, line 1250: why do you call `off` on all these here? They should be always null here, unless there is an engine bug right?
In which case, maybe `Debug.assert(this._evtLayersChanged == null);` should be better?

Maksims: This code, does not change behavior from old one.

There might a bug indeed. onEnable can be called for RenderComponent multiple times, there is no check if it is enabled or not.
So this code does not change the behavior, as before that fix, removes would happen also, regardless if there was onEnable called, even multiple times.
Fixing a multi-calls, should be a separate PR.

Maksims: You are right, off was not called here. Do you want me to remove this line then?

mvaligursky: yep please. Keen to do an engine release later today, if you have a chance to remove those soonish it'd be awesome.

mvaligursky: note that this needs to be done in all files I think, off was not called in any of them.

mvaligursky (approved):

mvaligursky:
cherry picked to main as well

PR #7139, opened 2024-11-25
Per frame randomized SSAO sampling & debug mode in composite pass

mvaligursky:


src/extras/render-passes/camera-frame.js
```diff
@@ -236,6 +239,13 @@ class CameraFrame {
         intensity: 0
     };
 
+    /**
+     * Debug rendering.
+     *
+     * @type {''|'scene'|'ssao'|'bloom'|'vignette'}
```
willeastcott, line 245: Would `null` be better than `''`?

mvaligursky: yep, will change this, thanks

willeastcott (approved):

src/extras/render-passes/camera-frame.js
mvaligursky: done now

PR #7146, opened 2024-11-27
[BREAKING] Removed support for VSM8 shadow type

mvaligursky:
- due to low quality / incorrect rendering, this has been removed as this was mostly a fallback for WebGL1 devices
- if project depends on it, it should be switched to VSM16 which is well supported on WebGL2 devices, and fallbacks for PCF3


slimbuck (approved):
are there any data migration consequences?

mvaligursky:
> are there any data migration consequences?

perhaps. Needed work is captured here: https://github.com/playcanvas/editor/issues/1233

PR #7147, opened 2024-11-27
ESM Grid Mesh Script

kpal81xd:
This is an alternative implementation of the Grid script in GH-7079

### Overview
- Uses a material to construct grid instead of directly rendering to quad
- Allows shader to take advantage of entity transform 
- Resolution `resolution: Grid.RESOLUTION_HIGH | Grid.RESOLUTION_MEDIUM | Grid.RESOLUTION_LOW`
- Grid can be recolored `colorX: Color`, `colorZ: Color`

### Preview
image


scripts/esm/grid.mjs
```diff
@@ -0,0 +1,319 @@
...
+     * @type {Vec2}
+     * @private
+     */
+    _halfExtents = new Vec2();
+
+    /**
+     * @type {Color}
+     * @private
+     */
+    _colorX = new Color(1, 0.3, 0.3);
```
mvaligursky, line 163: should some of these properties be exposed as attributes, to be visible in the Editor?

scripts/esm/grid.mjs
```diff
@@ -0,0 +1,319 @@
...
+     */
+    _colorZ = new Color(0.3, 0.3, 1);
+
+    /**
+     * @type {number}
+     * @private
+     */
+    _resolution = Grid.RESOLUTION_HIGH;
+
+    constructor(args) {
```
mvaligursky, line 177: what are the args this constructor accepts? Typically, a script is not constructed as such right?

mvaligursky:
Is there a way for the user to control which layer the grid renders to? Or is the expectation that the user knows there is a render component and uses properties of it?

scripts/esm/grid.mjs
mvaligursky: do those show as attributes in the inspector of the Editor?

scripts/esm/grid.mjs
mvaligursky: @marklundin - this is a bit strange, is this how users are expected to handle this in the scripts?

kpal81xd:
> Is there a way for the user to control which layer the grid renders to? Or is the expectation that the user knows there is a render component and uses properties of it?

The expectation is for the user to set the render component themselves if they want finer control such as which layer the grid renders to otherwise it will create one itself

scripts/esm/grid.mjs
marklundin: There's nothing inherently wrong with this, but `initialize()` is the more canonical use here. Is there a specific reason why this shouldn't happen in the initialize phase @kpal81xd?

scripts/esm/grid.mjs
kpal81xd: Yes @marklundin  added support for that

scripts/esm/grid.mjs
marklundin: It's the `script.create(Grid, { properties })` which adds properties in the `initialize` phase. Linking back to the original discussion here https://github.com/playcanvas/engine/issues/6785. Let's discuss

marklundin (approved):

PR #7151, opened 2024-11-28
Store, use and restore any existing onPreRenderLayer function

### LINKED ISSUE #7150: Outline renderer overwrites any onPreRenderLayer function on a camera
### Outline renderer overwrites any `onPreRenderLayer` function on a camera

If the camera being used to render a scene has an `onPreRenderLayer` function assigned the outliner code will overwrite that function when activated and reset it to null when de-activated. This is happening in the `frameUpdate` method. The existing function, if present, should be preserved and called as well as the outline renderer's own `onPreRenderLayer` function.

MAG-ME:
Fixes #7150

This change fixes the reported issue by
- Storing any existing onPreRenderLayer function
- Calling that function as well as performing the function provided by the outliner
- Restoring the old function when the outliner is deactivated

### DIFF
--- a/src/extras/renderers/outline-renderer.js
+++ b/src/extras/renderers/outline-renderer.js
@@ -332,7 +332,7 @@ class OutlineRenderer {
 
         // store the current function before we overwrite it
         const onPreRenderLayer = sceneCameraEntity.camera.onPreRenderLayer;
-        
+
         // function called before the scene camera renders a layer
         sceneCameraEntity.camera.onPreRenderLayer = (layer, transparent) => {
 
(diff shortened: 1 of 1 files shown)


mvaligursky:
yep I agree with @Maksims here completely.

MAG-ME:
While I cannot disagree about with the comment about only having a single callback being the real issue here that would be a breaking change so how do we solve this issue?

mvaligursky:
Let me add events in addition to existing callbacks over the next few days. This way people that depend on callbacks can keep using them. And we'll also have events, and switch internal code / scripts to use those.

MAG-ME:
Thank you @mvaligursky, that would be much appreciated.

PR #7154, opened 2024-11-29
A utility script to allow WebGL GPU capture using Xcode

mvaligursky:
- adding a manual page on this shortly.

### DIFF
--- a/scripts/utils/mac-gpu-profiling.js
+++ b/scripts/utils/mac-gpu-profiling.js
@@ -30,7 +30,7 @@ MacGPUProfiling.prototype.initialize = function () {
 };
 
 // Async function for WebGPU initialization
-MacGPUProfiling.prototype.initWebGPU = async function() {
+MacGPUProfiling.prototype.initWebGPU = async function () {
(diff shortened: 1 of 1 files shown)


scripts/utils/mac-gpu-profiling.js
```diff
@@ -0,0 +1,96 @@
+/**
+ * This script allows GPU Profiling on Mac using Xcode's GPU Frame Capture. Please read the instructions
+ * in the manual: https://developer.playcanvas.com/user-manual/optimization/gpu-profiling/
+ */
+var MacGPUProfiling = pc.createScript('MacGPUProfiling');
```
willeastcott, line 5: Can we just commit ESM scripts from now on?

willeastcott (approved):

scripts/utils/mac-gpu-profiling.js
mvaligursky: will do at a later date, when we update the examples build process to access ESM scripts from this folder.

PR #7156, opened 2024-11-29
Camera callbacks replaced by the events

mvaligursky:
based on the discussion here: https://github.com/playcanvas/engine/pull/7151

The callbacks on the `CameraComponent`:
```
onPreRender
onPreRenderLayer
onPostRender
onPostRenderLayer
onPreCull
onPostCull
```
where replaced by matching events fired on `Scene` to allow multiple subscribes.

The callbacks were removed, and the user get an error when they try to access them, suggesting a replacement.

Screenshot 2024-11-29 at 15 08 33
Screenshot 2024-11-29 at 15 08 44

### DIFF
--- a/src/scene/constants.js
+++ b/src/scene/constants.js
@@ -1061,7 +1061,7 @@ export const DITHER_IGNNOISE = 'ignnoise';
  * @type {string}
  * @category Graphics
  */
-export const EVENT_PRERENDER = 'preRender';
+export const EVENT_PRERENDER = 'prerender';
 
 /**
  * Name of event fired after the camera renders the scene.
@@ -1069,7 +1069,7 @@ export const EVENT_PRERENDER = 'preRender';
  * @type {string}
  * @category Graphics
  */
-export const EVENT_POSTRENDER = 'postRender';
+export const EVENT_POSTRENDER = 'postrender';
 
 /**
  * Name of event fired before a layer is rendered by a camera.
@@ -1077,7 +1077,7 @@ export const EVENT_POSTRENDER = 'postRender';
  * @type {string}
  * @category Graphics
  */
-export const EVENT_PRERENDER_LAYER = 'preRenderLayer';
+export const EVENT_PRERENDER_LAYER = 'prerender:layer';
 
 /**
  * Name of event fired after a layer is rendered by a camera.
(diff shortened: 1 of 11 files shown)


willeastcott:
Please search for `@event` in the codebase. Generally speaking, the standard format for event name strings in the codebase is all lower case (rather than camel-case). There _are_ instances of `name:subname` which I'm not averse to in some situations.

mvaligursky:
> The string is used directly.

Maybe we should consider changing it, as that's not consistent with all other constants engine exports that users use.
But that's a separate discussion / PR.

willeastcott:
> > The string is used directly.
> 
> Maybe we should consider changing it, as that's not consistent with all other constants engine exports that users use. But that's a separate discussion / PR.

Well, it's consistent with the event documentation at least. 😄  Honestly, take a look.

src/scene/constants.js
```diff
@@ -1054,3 +1054,51 @@ export const DITHER_BLUENOISE = 'bluenoise';
...
  */
 export const DITHER_IGNNOISE = 'ignnoise';
+
+/**
+ * Name of event fired before the camera renders the scene.
+ *
+ * @type {string}
+ * @category Graphics
+ */
+export const EVENT_PRERENDER = 'prerender';
```
willeastcott, line 1064: Do we even need these as constants? What do they give us?

src/scene/scene.js
```diff
@@ -67,6 +67,80 @@ class Scene extends EventHandler {
...
+     *    console.log(`Camera ${camera.entity.name} rendered the scene`);
+     * });
+     */
+    static EVENT_POSTRENDER = 'postrender';
+
+    /**
+     * Fired before the camera renders a layer. The handler is passed the {@link CameraComponent},
+     * the {@link Layer} that will be rendered, and a boolean parameter set to true if the layer is
+     * transparent. This is called during rendering to a render target or a default framebuffer, and
+     * additional rendering can be performed here, for example using ${@link QuadRender#render}.
```
willeastcott, line 98: What's with the `$` here?

src/scene/constants.js
mvaligursky: I use them internally in the engine, as string typing is error-prone.
but yeah, I don't think it's needed in general.

src/scene/scene.js
willeastcott: `npm run eswill`

src/scene/constants.js
willeastcott: Then at least keep them out of the API ref.

mvaligursky: we have others that are public and used by the way .. I just followed this as an example :)

Screenshot 2024-11-29 at 15 58 10

mvaligursky: removed.

willeastcott (approved):
Nice! 👏

PR #7160, opened 2024-12-02
Texture can be constructed with a specified number of mip levels

mvaligursky:
- texture could have been be created with `options.mipmaps`, which internally allocated a full mip map chain
- now `options.mipLevelCount` can be specified, which limits the mipmaps to specified number - useful when you need just one of two additional mip levels.
- we used to not allow mipmaps for integer texture formats, even though it's a valid use case. This is now allowed, but automatic mipmap generation for those texture is disablef. The user can manually render to the levels though.

### DIFF
--- a/src/platform/graphics/texture.js
+++ b/src/platform/graphics/texture.js
@@ -89,6 +89,12 @@ class Texture {
     /** @protected */
     _storage = false;
 
+    /** @protected */
+    _mipLevelCount = 0;
+
+    /** @protected */
+    _mipLevelCountRequested;
+
     /**
      * Create a new Texture instance.
      *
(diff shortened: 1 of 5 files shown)


src/platform/graphics/texture.js
```diff
@@ -644,6 +661,15 @@ class Texture {
         return this._mipmaps;
     }
 
+    /**
+     * Gets the number of mip levels.
+     *
+     * @type {number}
+     */
+    get mipLevelCount() {
```
willeastcott, line 669: `numLevels`?

Precedents:

* `IndexBuffer#numIndices`
* `LightCompnent#numCascades`
* `ParticleSystemComponent#numPartices`
* `VertexBuffer#numVertices`
* ...and quite a lot of internal properties following this naming.

mvaligursky: yeah good point. I so wish JS was a real language, where I can just 'rename' symbol :(
We have many uses of this name and only some need to change.

willeastcott (approved):

PR #7161, opened 2024-12-02
ESM Preload Fix

marklundin:
This PR fixes https://github.com/playcanvas/editor/issues/1236 along with https://github.com/playcanvas/editor-ui/pull/460.

This removes the `?hash=` appended to ESM paths during preloading. This ensures it matches the cache key  used for explicit imports in user code, resulting in the same module. 


mvaligursky (dismissed):
do we need to cherry pick to main_v1 as well?

marklundin:
Yep we do

Maksims:
By removing hash, this will lead to lack of cache invalidations. This will break what developers usually do.
Currently, everything under files/assets folder can be hard cached, because assets pipeline will ensure hash is generated if file changes. So url with hash is hard cached. This is the case for all assets, including scripts.

By changing this, developers will have do more sophisticated rules on caching, and won't be able to use hard caching for scripts.

This is undesirable. Please do not remove change such well working rule.

marklundin:
Could you clarify? The hash param is not included in the classic scripts preload either, so this better aligns the two

Maksims:
> Could you clarify? The hash param is not included in the classic scripts preload either, so this better aligns the two

I've just checked, and you are correct, there is no hash. Has this been changed at any point? As there was a rule: every asset within assets/files used hashes, so it could be hard cached (200 local storage). Everything outside of that folder is soft-cached (204).

marklundin:
This has been verified with the corresponding editor fix

willeastcott (approved):

PR #7168, opened 2024-12-04
Export internal CameraShaderParams class which is currently required by the Editor

mvaligursky:
not exposed in docs


willeastcott:
Why is this needed by the Editor?

mvaligursky:
This class is used by the shader generator, and Editor has a case where it manually generates depth shader. Not a good solution, but that's what it is. I used to use RenderParams, but I removed those and forgot to expose this new way.

willeastcott:
If I approve this, can we consider this temporary with a view to a better solution based on public API in the future?

mvaligursky:
> If I approve this, can we consider this temporary with a view to a better solution based on public API in the future?

Of course.

willeastcott (approved):

PR #7171, opened 2024-12-04
Add basic XR navigation ESM script

willeastcott:
One handed navigation in XR. Works with both hands and tracked controllers.

https://github.com/user-attachments/assets/0fbbe46c-132d-4c07-bd41-1584d6faaf99

### DIFF
--- a/scripts/esm/xr-navigation.mjs
+++ b/scripts/esm/xr-navigation.mjs
@@ -1,3 +1,4 @@
+/* eslint-disable-next-line import/no-unresolved */
 import { Color, Script, Vec3 } from 'playcanvas';
 
 /** @import { XrInputSource } from 'playcanvas' */
--- a/scripts/esm/xr-navigation.mjs
+++ b/scripts/esm/xr-navigation.mjs
@@ -1,0 +1,122 @@
+import { Color, Script, Vec3 } from 'playcanvas';
+
+/** @import { XrInputSource } from 'playcanvas' */
+
+export class XrNavigation extends Script {
+    /** @type {Set<XrInputSource>} */
+    inputSources = new Set();
+
+    /** @type {Map<XrInputSource, boolean>} */
+    activePointers = new Map();
+
+    validColor = new Color(0, 1, 0);    // Green for valid teleport
+
+    invalidColor = new Color(1, 0, 0);   // Red for invalid teleport
+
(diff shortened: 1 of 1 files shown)


scripts/esm/xr-navigation.mjs
```diff
@@ -0,0 +1,123 @@
...
+            0,
+            origin.z + direction.z * t
+        );
+    }
+
+    tryTeleport(inputSource) {
+        const origin = inputSource.getOrigin();
+        const direction = inputSource.getDirection();
+
+        const hitPoint = this.findPlaneIntersection(origin, direction);
```
mvaligursky, line 70: personally I'd just use `Plane.intersectsRay` here, but this works too (and is faster)

scripts/esm/xr-navigation.mjs
```diff
@@ -0,0 +1,123 @@
...
+            if (!this.activePointers.get(inputSource)) continue;
+
+            const start = inputSource.getOrigin();
+            const direction = inputSource.getDirection();
+
+            const hitPoint = this.findPlaneIntersection(start, direction);
+
+            if (hitPoint) {
+                // Draw line to intersection point
+                this.app.drawLine(start, hitPoint, this.validColor);
```
mvaligursky, line 90: you could move all the drawing code to a separate function, allowing users to easily extend the script and implement a custom drawing by overriding the function

mvaligursky (approved):

slimbuck (approved):
<3<3

scripts/esm/xr-navigation.mjs
```diff
@@ -0,0 +1,123 @@
+/* eslint-disable-next-line import/no-unresolved */
+import { Color, Script, Vec3 } from 'playcanvas';
+
+/** @import { XrInputSource } from 'playcanvas' */
+
+export class XrNavigation extends Script {
```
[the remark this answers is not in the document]
willeastcott: I've added a class description.

scripts/esm/xr-navigation.mjs
willeastcott: Yes, customizing visuals is important but I'd like to leave solving that to a follow up PR.

scripts/esm/xr-navigation.mjs
willeastcott: I'll keep that in mind for a follow up PR.

PR #7174, opened 2024-12-05
Updated the way shaders include tonemapping chunks

mvaligursky:
- moved tonemapping chunks to a sub-folder
- created a new tonemappingPS chunk which includes them and uses one based on the define
- updated shader generators to use this chunk with an appropriate define


src/extras/render-passes/render-pass-compose.js
```diff
@@ -13,6 +11,11 @@ import { ShaderGenerator } from '../../scene/shader-lib/programs/shader-generato
 // before the tone-mapping.
 
 const fragmentShader = /* glsl */ `
+
+    #include "tonemappingPS"
```
willeastcott, line 15: Oh man, I wish we didn't have that `PS` suffix. 😬

willeastcott (approved):

src/extras/render-passes/render-pass-compose.js
mvaligursky: agreed :)

PR #7175, opened 2024-12-06
Update to render passes - generate scene half sized texture

mvaligursky:
Before: The scene gets rendered to a render target. The bloom then blurs it by downscaling and upscaling, starting at half size. The first bloom pass would just cheaply (box filter) downscale to half size. That makes the final bloom texture half size of the scene.

Now: The cheap downscale to half res is pulled out of bloom, as that is something other effects (DOF) needs as well, so lets generate it only one time. Bloom then uses it.

Small change is that the bloom texture ends up being a quarter size and not half size, so slightly cheaper, but I cannot see any difference visually, so happy with this.

### DIFF
--- a/src/platform/graphics/render-target.js
+++ b/src/platform/graphics/render-target.js
@@ -315,7 +315,7 @@ class RenderTarget {
         if (this.width !== width || this.height !== height) {
 
             if (this.mipLevel > 0) {
-                Debug.warn('Only render target rendering to mipLevel 0 can be resized, ignoring.', this);
+                Debug.warn('Only a render target rendering to mipLevel 0 can be resized, ignoring.', this);
                 return;
             }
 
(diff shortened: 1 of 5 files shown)


src/platform/graphics/render-target.js
```diff
@@ -312,13 +312,13 @@ class RenderTarget {
...
 
-        if (this.mipLevel > 0) {
-            Debug.warn('Only render target rendering to mipLevel 0 can be resized, ignoring.', this);
-            return;
-        }
-
         if (this.width !== width || this.height !== height) {
 
+            if (this.mipLevel > 0) {
+                Debug.warn('Only render target rendering to mipLevel 0 can be resized, ignoring.', this);
```
willeastcott, line 318:
```suggestion
                Debug.warn('Only a render target rendering to mipLevel 0 can be resized, ignoring.', this);
```

PR #7185, opened 2024-12-09
Further gsplat shader refinements

slimbuck:
This PR is a followup to #7169 and further refines the chunk layout for gsplat material:
- Gaussians now have 3 data structures:
    - `SplatSource` - stores the splat's data uv, id and render order
    - `SplatCenter` - stores the splat's center in camera and clip space
    - `SplatCorner` - stores the splat's corner offset (from center) in clip space and the corner uv
- The functions `initSource`, `initCenter` and `initCorner` populate the respective structures and are placed in their own chunks for easy customisation
- Each chunk now declares and uses its own global variables (uniforms and attributes)

Also:
- Reflection vector for SH evaluation is done in the main function so it can easily be customised
- Slightly faster early-out when gaussian is behind the camera
- Disable clipping of splats by the near and far clipping planes (clamp depth instead)
- Replace `tex_params` uniform with call to glsl `textureSize`


src/scene/shader-lib/chunks/gsplat/vert/gsplatCommon.js
```diff
@@ -1,20 +1,25 @@
...
     ivec2 uv;           // splat uv
     vec2 cornerUV;      // corner coordinates for this vertex of the gaussian (-1, -1)..(1, 1)
 };
 
-struct ProjectedState {
+// stores the camera and clip space position of the gaussian center
+struct SplatCenter {
+    vec3 view;          // center in view space
+    vec4 proj;          // center in clip space
     mat4 modelView;     // model-view matrix
```
mvaligursky, line 15: With clustered lighting, I had precision issue when a mat4 was stored in a struct, so I had to move it out. Not sure it would have the same impact here, likely not as shadow flickering is a specific issue.

mvaligursky (approved):

willeastcott (approved):

PR #7188, opened 2024-12-09
Compressed material key hash update

slimbuck:
Compressed gsplat material wasn't generating key with defines.

### DIFF
--- a/src/scene/gsplat/gsplat-compressed-material.js
+++ b/src/scene/gsplat/gsplat-compressed-material.js
@@ -361,7 +361,8 @@ class GSplatCompressedShaderGenerator {
     generateKey(options) {
         const vsHash = hashCode(options.vertex);
         const fsHash = hashCode(options.fragment);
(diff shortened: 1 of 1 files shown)


kpal81xd (approved):

src/scene/gsplat/gsplat-compressed-material.js
```diff
@@ -361,7 +361,8 @@ class GSplatCompressedShaderGenerator {
     generateKey(options) {
         const vsHash = hashCode(options.vertex);
         const fsHash = hashCode(options.fragment);
-        return `splat-${options.pass}-${options.gamma}-${options.toneMapping}-${vsHash}-${fsHash}-${options.dither}}`;
+        const defines = options.defines?.sort().join('-') ?? '';
```
mvaligursky, line 364: I use this in other generators:
        const definesHash = ShaderGenerator.definesHash(options.defines);

mvaligursky: ah

PR #7194, opened 2024-12-13
Optimize RenderComponent destroy

Maksims:
RenderComponent has a property `rootBone`, which is just a reference to an entity, that is used for skinning. Currently, it uses overcomplex EntityReference which creates a number of event handlers in various systems, which is extremely inefficient. `rootBone` - is a simple property that just is either `null` or some Entity.
When creating/destroying entities, EntityReference has to subscribe and unsubscribe to a number of events, which leads to an unnecessary workload.

So this PR just simplifies `rootBone` property, which does not change any behaviours. While optimising destruction of all RenderComponents (not skinned especially!!).

### Here are test results, of destroying 4706 entities with RenderComponents (not skinned, primitive cubes):
**Before: 640ms**
**After: 84ms**

# This is ~7.5 times speedup, or ~87% improvement.

In our use case, we need to parametrically create a lot of entities, and remove them on user interaction, so any lag spike is extremely noticeable. I've also noticed that changing rootBone when an entity is disabled, will not be taken in account when an entity is enabled, but this is an existing current problem and should be fixed in a separate PR.

Also, a lot of other components use EntityReference, which is best to go. It is unnecessarily complex and inefficient.

### DIFF
--- a/src/framework/components/render/component.js
+++ b/src/framework/components/render/component.js
@@ -172,8 +172,7 @@ class RenderComponent extends Component {
      *
      * @param {import('./system.js').RenderComponentSystem} system - The ComponentSystem that
      * created this Component.
-     * @param {Entity} entity - The Entity that this Component is
-     * attached to.
+     * @param {Entity} entity - The Entity that this Component is attached to.
      */
     constructor(system, entity) {
         super(system, entity);
--- a/src/framework/components/render/component.js
+++ b/src/framework/components/render/component.js
@@ -719,6 +719,7 @@ class RenderComponent extends Component {
                 this._rootBone = value;
             } else if (isString) {
                 this._rootBone = this.system.app.getEntityFromIndex(value) || null;
+                Debug.assert(this._rootBone, 'Failed to find rootBone Entity by GUID');
             } else {
                 this._rootBone = null;
             }
--- a/src/framework/components/render/component.js
+++ b/src/framework/components/render/component.js
@@ -12,8 +12,6 @@ import { AssetReference } from '../../asset/asset-reference.js';
 
 import { Component } from '../component.js';
 
-import { EntityReference } from '../../utils/entity-reference.js';
-
 /**
  * The RenderComponent enables an {@link Entity} to render 3D meshes. The {@link RenderComponent#type}
  * property can be set to one of several predefined shape types (such as `box`, `sphere`, `cone`
@@ -53,9 +51,6 @@ import { EntityReference } from '../../utils/entity-reference.js';
  * - [Spinning Cube](https://playcanvas.github.io/#/misc/hello-world)
  * - [Primitive Shapes](https://playcanvas.github.io/#/graphics/shapes)
  * - [Loading Render Assets](https://playcanvas.github.io/#/graphics/render-asset)
- *
- * @property {import('../../entity.js').Entity} rootBone A reference to the entity to be used as
- * the root bone for any skinned meshes that are rendered by this component.
  *
  * @category Graphics
  */
@@ -136,10 +131,12 @@ class RenderComponent extends Component {
     _material;
 
     /**
-     * @type {EntityReference}
-     * @private
+     * A reference to the entity to be used as the root bone for any skinned meshes that
+     * are rendered by this component.
+     *
+     * @type {import('../../entity.js').Entity|null}
      */
-    _rootBone;
+    _rootBone = null;
 
     /**
      * @type {import('../../../core/event-handle.js').EventHandle|null}
@@ -177,8 +174,6 @@ class RenderComponent extends Component {
         super(system, entity);
 
         // the entity that represents the root bone if this render component has skinned meshes
-        this._rootBone = new EntityReference(this, 'rootBone');
-        this._rootBone.on('set:entity', this._onSetRootBone, this);
 
         // render asset reference
         this._assetReference = new AssetReference(
@@ -275,7 +270,6 @@ class RenderComponent extends Component {
      * @type {string}
      */
     set type(value) {
-
         if (this._type !== value) {
             this._area = null;
             this._type = value;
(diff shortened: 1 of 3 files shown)


src/framework/components/render/component.js
```diff
@@ -934,23 +938,20 @@ class RenderComponent extends Component {
...
                 if (mesh.skin && !meshInstance.skinInstance) {
-                    meshInstance.skinInstance = SkinInstanceCache.createCachedSkinInstance(mesh.skin, this._rootBone.entity, this.entity);
+                    meshInstance.skinInstance = SkinInstanceCache.createCachedSkinInstance(mesh.skin, this._rootBone, this.entity);
                 }
             }
         }
     }
 
     _cloneMeshes(meshes) {
-
```
mvaligursky, line 953: no need to remove empty lines in this PR :)

Maksims: Couldn't resist, as most places in this file do not have these unnecessary empty lines in the same context.
Do you want me to add them back?

willeastcott: Personally, I don't mind removing them. Our coding standards don't state you should add a blank line after the first line of a function. I we want to introduce that rule, ok, we can debate it - but I'm more keen that we enforce consistency across the whole codebase.

src/framework/components/render/component.js
```diff
@@ -136,10 +131,12 @@ class RenderComponent extends Component {
     _material;
 
     /**
-     * @type {EntityReference}
-     * @private
+     * A reference to the entity to be used as the root bone for any skinned meshes that
+     * are rendered by this component.
+     *
+     * @type {import('../../entity.js').Entity|null}
```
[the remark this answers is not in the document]
Maksims: Missed that, added a proper property comments.

src/framework/components/render/component.js
```diff
@@ -703,27 +697,46 @@ class RenderComponent extends Component {
...
+     * Sets the root bone entity (or entity guid) for the render component.
+     *
+     * @type {import('../../entity.js').Entity|string|null}
      */
-    _onSetRootBone(entity) {
-        if (entity) {
-            this._onRootBoneChanged();
+    set rootBone(value) {
+        if (this._rootBone !== value) {
+            const isString = typeof (value) === 'string';
```
willeastcott, line 706:
```suggestion
            const isString = typeof value === 'string';
```

src/framework/components/render/component.js
```diff
@@ -136,10 +131,13 @@ class RenderComponent extends Component {
     _material;
 
     /**
-     * @type {EntityReference}
+     * A reference to the entity to be used as the root bone for any skinned meshes that
+     * are rendered by this component.
+     *
+     * @type {import('../../entity.js').Entity|null}
```
willeastcott, line 137: Do this instead:

https://github.com/playcanvas/engine/blob/9af3f9efa2f0a45d24368fea680dba9408cd7eed/src/extras/exporters/gltf-exporter.js#L25-L27

willeastcott: In other words, once at the top (after regular imports) and then you can use the `Entity` type throughout the module.

src/framework/components/render/component.js
```diff
@@ -703,27 +697,46 @@ class RenderComponent extends Component {
     }
 
     /**
-     * @param {import('../../entity.js').Entity} entity - The entity set as the root bone.
-     * @private
+     * Sets the root bone entity (or entity guid) for the render component.
+     *
+     * @type {import('../../entity.js').Entity|string|null}
```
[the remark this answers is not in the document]
Maksims: Yes. Old definition was wrong, as it was accepting strings and actively does it.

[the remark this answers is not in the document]
Maksims: `asset` accessor does the same with asset id.

src/framework/components/render/component.js
```diff
@@ -170,15 +172,13 @@ class RenderComponent extends Component {
      *
      * @param {import('./system.js').RenderComponentSystem} system - The ComponentSystem that
      * created this Component.
-     * @param {import('../../entity.js').Entity} entity - The Entity that this Component is
+     * @param {Entity} entity - The Entity that this Component is
```
willeastcott, line 175: This should all fit on one line now. 😄

willeastcott (approved):
LGTM. Nice! 👏  Are you gonna do a follow up to get this merged to main?

Maksims:
> LGTM. Nice! 👏 Are you gonna do a follow up to get this merged to main?

I'm happy for the team to do the cherry-picking and manage multiple versions on their own.
Not sure that it is friendly to outside contributors to ask them to manage multiple versions as I'm for example am not familiar with v2, and we don't use it yet as it is not available in Editor, so I cannot test this PR against main, only main_v1.

src/framework/components/render/component.js
```diff
@@ -703,27 +700,46 @@ class RenderComponent extends Component {
...
+            }
+
+            if (this._rootBone) {
+                this._clearSkinInstances();
+            }
+
+            if (value instanceof GraphNode) {
+                this._rootBone = value;
+            } else if (isString) {
+                this._rootBone = this.system.app.getEntityFromIndex(value) || null;
```
mvaligursky, line 721: I would be tempted to add Debug.warn if this._rootBone is null after this line - to inform the user that the guid didn't resolve.

src/framework/components/render/component.js
```diff
@@ -852,6 +868,10 @@ class RenderComponent extends Component {
         this._evtLayersChanged?.off();
         this._evtLayersChanged = null;
 
+        if (this._rootBone) {
+            this._clearSkinInstances();
```
mvaligursky, line 872: Ideally this would not be here, but I see how it makes it easier / more consistent, and the impact is low.

mvaligursky (approved):
Approving, but would not mind the warn call to be added as mentioned. I did testing of this, including the Editor, with various entity orders in hierarchy and duplication and so on, and I cannot see any problem here. Really well done @Maksims .

PR #7198, opened 2024-12-15
[FIX] Handle case where pbrMetallicRoughness is not present

### LINKED ISSUE #6989: Gltf pbrMetallicRoughness not loaded correctly
If the gltf material does not specify pbrMetalicRoughness at [this point](https://github.com/playcanvas/engine/blob/main/src/framework/parsers/glb-parser.js#L1128) in the gltf parser, then the default values should be used instead (as specified by the [gltf spec](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-material)).

The PlayCanvas material should have metalness enabled in this case and roughness set to 1 (i.e. inverseGloss enabled and gloss set to 1 on the material).

This hasn't been noticed before, because the default PC material happens to look roughly correct.

willeastcott:
Fixes #6989

### DIFF
--- a/src/framework/parsers/glb-parser.js
+++ b/src/framework/parsers/glb-parser.js
@@ -1112,17 +1112,22 @@ const extensionIridescence = (data, material, textures) => {
 const createMaterial = (gltfMaterial, textures) => {
     const material = new StandardMaterial();
 
+    if (gltfMaterial.hasOwnProperty('name')) {
+        material.name = gltfMaterial.name;
+    }
+
     // glTF doesn't define how to occlude specular
     material.occludeSpecular = SPECOCC_AO;
 
     material.diffuseVertexColor = true;
-
     material.specularTint = true;
     material.specularVertexColor = true;
 
-    if (gltfMaterial.hasOwnProperty('name')) {
-        material.name = gltfMaterial.name;
-    }
+    // Set glTF spec defaults
+    material.specular.set(1, 1, 1);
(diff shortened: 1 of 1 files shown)


src/framework/parsers/glb-parser.js
```diff
@@ -1112,17 +1112,22 @@ const extensionIridescence = (data, material, textures) => {
...
     material.specularTint = true;
     material.specularVertexColor = true;
 
-    if (gltfMaterial.hasOwnProperty('name')) {
-        material.name = gltfMaterial.name;
-    }
+    // Set glTF spec defaults
+    material.specular.set(1, 1, 1);
+    material.gloss = 1;
+    material.glossInvert = true;
```
mvaligursky, line 1129: I'm not sure about the gloss being 1 here, I think it should be 0.
The [spec](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-material-pbrmetallicroughness): 
```
roughnessFactor | The factor for the roughness of the material. | default: 1
```

And so I think gloss needs to be zero, which makes the roughness 1 when glossInvert is enabled.
Before, in StandardMaterial, it would use this default value (close to 0)
```
    _defineFloat('gloss', 0.25);
```

willeastcott: Agreed. Done.

mvaligursky (approved):

PR #7200, opened 2024-12-15
[FIX] Correctly set texture dimensions from video source

### LINKED ISSUE #4571: fitMode doesn't work on video textures
If you fork [this project](https://playcanvas.com/project/721265/overview/video-in-ui) and attempt to set the fit mode to cover for example, it won't work. The video will not play on the UI according to the fit mode you chose. It works with static images just fine, but it seems that on textures that are updated at runtime, it fails.

willeastcott:
Fixes #4571 

### DIFF
--- a/src/platform/graphics/texture.js
+++ b/src/platform/graphics/texture.js
@@ -955,8 +955,13 @@ class Texture {
                     this._levelsUpdated[mipLevel] = true;
                 }
 
-                width = source.width;
-                height = source.height;
+                if (source instanceof HTMLVideoElement) {
+                    width = source.videoWidth;
+                    height = source.videoHeight;
+                } else {
+                    width = source.width;
+                    height = source.height;
+                }
             }
         }
 


src/platform/graphics/texture.js
```diff
@@ -955,8 +955,13 @@ class Texture {
                     this._levelsUpdated[mipLevel] = true;
                 }
 
-                width = source.width;
-                height = source.height;
+                if (source instanceof HTMLVideoElement) {
```
slimbuck, line 958: not sure how applicable these days, but in the graphics device we test for presence of HTMLVideoElement as well:
`(typeof HTMLVideoElement !== 'undefined' && texture instanceof HTMLVideoElement)`

slimbuck (approved):

src/platform/graphics/texture.js
willeastcott: Oh, that's probably for running in Node? There seem to already be a couple of other `instanceof HTMLVidelElement` tests in the platform _without_ the additional guard so I think we're fine.

mvaligursky (approved):

Issue #7202, opened 2024-12-16
ComonenSystems add/beforeremove events issue.


Maksims:
# The problem:

ComponentSystem has a very large subscriber lists for add/beforeremove events with mostly irrelevant subscribers. Leading to higher cost of adding/removing components.

## Details: 

Many ComponentSystems do provide a way to inform when component has been added or is about to be removed, and it provides a single event name for that: `add` and `beforeremove` on ComponentSystem. As an argument, it provide an Entity and a Component.
Then every single system's Component, will subscribe to such event, only to check then if then entity is the same as this Component's entity.

This leads to a single callbacks list that has a very large number of subscribers, where only a few of hundreds/thousands of subscribers are relevant.

So then on every added/remove Component, this will lead to wasted time going through irrelevant callbacks and calling them only to return early.
This has also an extra overhead due to high cost of removing event handles from massive arrays. With the introduction of a more efficient EventHandle.off, this can be improved, but this does not solve the original problem.

# The solution:

If a subscriber needs to know when it is about to be removed or added, it then should be subscribed and handled on that subscriber (Component) itself. This will avoid large callback lists on a ComponentSystem, and will be way more efficient when adding/removing entities with components.

mvaligursky:
Investigated the current engine at c4e2dba6e. This issue is partly resolved, but LayoutGroupComponent still has the global subscription pattern.

- #7204 migrated buttons to targeted notifications; #7206 did the same for scroll views and scrollbars.
- ComponentSystem now fires `beforeremove` directly on the component before system-level cleanup.
- Each layout group still subscribes to `add` and `beforeremove` on both the element and layout-child systems (four global subscriptions). Every matching system event calls every layout group, and `_isSelfOrChild()` also scans each group's immediate children using `indexOf()`.

An in-memory probe using the current source and NullGraphicsDevice confirmed the unnecessary dispatch, even with disabled groups:

| Layout groups | Unrelated layout-child add/remove cycles | Layout-group callbacks | Relevant callbacks |
| ---: | ---: | ---: | ---: |
| 100 | 100 | 20,000 | 0 |
| 1,000 | 100 | 200,000 | 0 |
| 5,000 | 100 | 1,000,000 | 0 |

Bulk removal also repeatedly scans the four global listener lists. Switching to EventHandle.off() alone would not address dispatch, and its current implementation still uses `indexOf()` and `splice()`.

[...]

Issue #7195, opened 2024-12-13
EntityReference - is convoluted, complex and inefficient.


Maksims:
Various Components use EntityReference to somewhat simplify events management, but its convoluted implementation, has a lot of inefficient and unnecessary mechanics, like subscribing to a large EventHandlers, and removing such subscriptions often.
So when such Components are added or removed, there is a significant amount of time spent in EntityReference.

Having 100 lines of comments describing what it does, and then having a complex implementation that is still hard to understand - makes it even worse.

The events management is inefficient as there was no EventHandle at the time, so events removals from ComponentSystems - is slow. By utilizing EventHandle.off some speedup can be achieved. It also subscribes to a large EventHandlers like ComponentSystem only to check in callback if it relates to our component - this of course leads to a very large callback lists - which just slows everything down on fires.

A simple example is RenderComponent.rootBone - which is just a reference to another entity. And can be a simple property, and does not need to be a complex system behind it. [This PR](https://github.com/playcanvas/engine/pull/7194) showcases this. And leads to ~87% speedup improvement for component removal.

Other Components need to do a similar job by simplifying properties, and implementing their logic directly inside of a component, instead of using convoluted EntityReference.

## This will lead to massive speedups and move logic implementation to its relevant place - the Component itself.

Maksims:
Here is described problem in simplified code example:

```js
class ComponentSystem extends EventHandler {

}

class Component {
    constructor() {
        // in every EntityReference
        this.system.on('add', this.onAdd, this);
        this.system.on('beforeremove', this.onRemove, this);
        //
    }

    onAdd(entity) {
        if (this.entity === entity) {
            // do something
        }
    }

    // { in an EntityReference:
    onRemove(entity) {
        if (this.entity === entity) {
            this.system.off('add', this.onAdd, this);
            this.system.off('beforeremove', this.onRemove, this);
        }
    }
    //
}
```

Each EntityReference would subscribe to whole system, and then check if `this.entity === entity`. So when we create a new component, we subscribe to whole system again, and when we remove component, we unsubscribe, so if there is a lot of these component, or even components subscribing to the same system (like Element, Button, Scrollbar, ScrollView, and possibly others), this leads to an enormous list of callbacks which do `if (this.entity === entity) {` to check if it relates to them.

So the cost of creating and removeing of a new component grows exponentially with size of a scene. And the size does not need to be crazy to start getting lag spikes on every hierarchy modification (adding, removing, reparenting).

[...]

PR #7204, opened 2024-12-16
ButtonComponent remove EntityReference & optimize add/remove speed.

Maksims:
Related to #7195

This PR does not change any APIs.

This PR removes the dependency on EntityReference by ButtonComponent. Instead, it manages its own events by subscribing and unsubscribing to imageEntity's ElementComponent's life and its relevant properties.
Also, it removes reliance on ComponentSystem's global event lists (described here: https://github.com/playcanvas/engine/issues/7202), to ensure it does not get slower exponentially with the number of Button and Element Components in the scene.

Due to improvements for global event lists on ElementComponentSystem, Element and similar UI components also benefit from this PR. Similar PR for them will come later to keep each PR manageable and narrow.

We had a few projects, which are very UI heavy, with inventory buttons, a lot of UI elements, dynamically created icons all over and so on. We've noticed that the time it takes to modify hierarchy in runtime grows significantly as the project scales in size. This PR addresses some of that.

# Here are some tests:

Create a number of entities with button and element components and then remove them.

## Current engine:

**Quantity:** 777
**Add:** ~145ms (avg. per 1k: ~187ms)
**Remove:** ~87ms (avg. per 1k: ~112ms)
**Average time per 1,000**: 

**Quantity:** 2840
**Add:** ~877ms (avg. per 1k: ~309ms)
**Remove:** ~839ms  (avg. per 1k: ~295ms)

## With this PR:

**Quantity:** 777
**Add:** ~114ms (avg. per 1k: ~147ms)
**Remove:** ~25ms (avg. per 1k: ~32ms)

**Quantity:** 2840
**Add:** ~286ms (avg. per 1k: ~101ms)
**Remove:** ~93ms (avg. per 1k: ~33ms)

# Results:

From the tests, you can see that the cost of creating an Entity with Button and Element components grows exponentially (3rd test shows) with a number of similar Entities already in the scene.
And with this PR this is not the case, and the cost is constant.

## The speed benefits are (it goes higher the more similar entities are in the scene):
Creation speed **~21%-67%+ improvement.**
Deletion speed **~71%-89%+ improvement.**

### DIFF
--- a/src/framework/components/button/component.js
+++ b/src/framework/components/button/component.js
@@ -3,12 +3,17 @@ import { now } from '../../../core/time.js';
 import { math } from '../../../core/math/math.js';
 import { Color } from '../../../core/math/color.js';
 
-import { EntityReference } from '../../utils/entity-reference.js';
+import { GraphNode } from '../../../scene/graph-node.js';
 
 import { Component } from '../component.js';
 
 import { BUTTON_TRANSITION_MODE_SPRITE_CHANGE, BUTTON_TRANSITION_MODE_TINT } from './constants.js';
 import { ELEMENTTYPE_GROUP } from '../element/constants.js';
+
+/**
+ * @import { EventHandle } from '../../../core/event-handle.js'
+ * @import { Entity } from '../../entity.js'
+ */
 
 const VisualState = {
     DEFAULT: 'DEFAULT',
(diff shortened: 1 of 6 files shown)


src/framework/components/button/component.js
```diff
@@ -908,21 +1025,24 @@ class ButtonComponent extends Component {
...
     _applySprite(spriteAsset, spriteFrame) {
-        spriteFrame = spriteFrame || 0;
+        const element = this._imageEntity?.element;
+        if (!element) {
+            return;
+        }
 
-        if (this._imageReference.hasComponent('element')) {
-            this._isApplyingSprite = true;
+        spriteFrame = spriteFrame || 0;
```
mvaligursky, line 1033: should this be just a default value for the parameter?
```
_applySprite(spriteAsset, spriteFrame = 0) {
```

mvaligursky:
cherry picked to main_v2

PR #7205, opened 2024-12-17
Camera controls updates

kpal81xd:
### Overview
- Attribute refactor
- Added `zoomDamping` separate from `rotateDamping` and `moveDamping`
- Emits events `clamp:position` and `clamp:look` for more complex restriction of camera controls such as setting custom bounds
- Checks for double attach and detach calls (Used for disabling controls to prioritize other events)

### Examples
- Updated examples and controls for multi/fly/orbit for renamed attributes
- Replaced editor example orbit camera with camera controls
- Fixed touch control issues on mobile


vincentwoo:
I'm very interested in an easy to use option for the basic camera-controls that allow for clamping the camera within a radius of the origin (or defined point) as well as keeping it above the ground plane, in case that's on the menu!

scripts/esm/camera-controls.mjs
```diff
@@ -401,15 +459,43 @@ class CameraControls extends Script {
...
+    }
+
     /**
      * @private
-     * @param {number} value - The value to clamp.
-     * @returns {number} - The clamped value.
+     * @param {Vec2} look - The value to clamp.
      */
-    _clampPitch(value) {
+    _clampLook(look) {
```
mvaligursky, line 475: private, but I'd still suggest to rename this for consistency

mvaligursky (approved):
Looks good to me.
I'm typically used to dumping work the opposite way, a larger value would stop the rotation quicker, but this works as well.

PR #7206, opened 2024-12-17
Scrollview remove entity reference

Maksims:
This PR depends on https://github.com/playcanvas/engine/pull/7194 and https://github.com/playcanvas/engine/pull/7204 (and includes them already), merge them first please so diff is more concise.

This PR does not change API or behavior.

As mentioned in https://github.com/playcanvas/engine/issues/7195 - EntityRefence need to go. This PR completes that, by removing a last dependency on it and also removing reliance on ComponentSystem's global event lists, which leads to exponentially higher costs of creating/removing related components (Element especially).

There is a significant performance improvement when creating/removing such components similar to other two PR's mentioned above. This also benefits ElementComponent's creation/removal performance, as every ScrollView and Scrollbar components were subscribing to ElementComponentSystem's global lists.


mvaligursky:
@Maksims - could you please merge main to it and resolve the conflict

Maksims:
> @Maksims - could you please merge main to it and resolve the conflict

Done.

mvaligursky (approved):
This looks good, but I wonder if we could avoid some code duplication here, which adds some complexity. It could do with some wrapper of some of the code?

Maksims:
> This looks good, but I wonder if we could avoid some code duplication here, which adds some complexity. It could do with some wrapper of some of the code?

I thought of some more unified way to handle entity setting, subscribe/unsubscribe mechanics also, but every case is so slightly different. This PR reduces overall amount of code in the Engine. I guess some refactoring is possible as a separate PR in the future, but that would require to refactor whole component's code.

mvaligursky:
merged to v2

PR #7215, opened 2024-12-19
[Fix] Fix to recent change to RenderComponent.rootBone when instantiating glb container

mvaligursky:


willeastcott:
Wait, so `rootBone` was a valid property for `addComponent` before but not now? @Maksims, what's the deal with that?

mvaligursky:
It is still valid now as well, but all the bones the skinning uses (the sub-hierarchy) needs to be created before the rootBone is assigned, as the code inside rootBone property connects to those bones.
Before, this was on a delay till entity was enabled.
But this should not cause any issues, typically people do not clone hierarchy and create components at the same time.

willeastcott (approved):

PR #7216, opened 2024-12-19
[Fix] Change gloss default value to 1

mvaligursky:
it seems my suggestion here was not correct: https://github.com/playcanvas/engine/pull/7198#discussion_r1889896613

as a result, few engine examples rendered like this:
Screenshot 2024-12-19 at 15 54 03


slimbuck (approved):

willeastcott (approved):

LeXXik:
Looks normal.

image

PR #7220, opened 2024-12-20
DOF - near blur is optional

mvaligursky:
- made the near blur optional, to allow cheaper DOF when only the far blur is needed
- skips blurring the near part of the texture
- CoC texture is single channel instead of two channel

new public API:
```
CameraFrame.dof.nearBlur  // defaults to false
```


slimbuck (approved):
not sure it should be default off, but if doom does it that way...

PR #7223, opened 2024-12-20
Switch to node-canvas for running unit tests

willeastcott:
Switch from `skia-canvas` to `@napi-rs/canvas` for running unit tests in Node (since it supports Apple Silicon). Note that `canvas` should support Apple Silicon soon too when v3.0.0 is released and compatible with `jsdom`. We should switch to that when we are able.

This PR means we can now use vanilla DOM API to set up the unit tests.

We can also drop `xhr2`.


mvaligursky (approved):
Cool!

PR #7239, opened 2024-12-24
Render pass tracing now logs values used to clear color/depth/stencil buffer

mvaligursky:
- updated Color.toString to handle HDR colors as well by printing those as an array instead of #xxxxxx
- use it for render pass clear values logging

### DIFF
--- a/src/core/math/color.js
+++ b/src/core/math/color.js
@@ -243,7 +243,14 @@ class Color {
         }
 
         let s = `#${((1 << 24) + (Math.round(r * 255) << 16) + (Math.round(g * 255) << 8) + Math.round(b * 255)).toString(16).slice(1)}`;
-        s += alpha === true ? a.toString(16).padStart(2, '0') : '';
+        if (alpha === true) {
+            const aa = Math.round(a * 255).toString(16);
+            if (this.a < 16 / 255) {
+                s += `0${aa}`;
+            } else {
+                s += aa;
+            }
+        }
         return s;
     }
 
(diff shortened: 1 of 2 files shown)


src/core/math/color.js
```diff
@@ -226,24 +226,31 @@ class Color {
...
+     * @param {boolean} asArray - If true, the output will be an array of numbers.
      * @returns {string} The color in string form.
      * @example
      * const c = new pc.Color(1, 1, 1);
      * // Outputs #ffffffff
      * console.log(c.toString());
      */
-    toString(alpha) {
-        let s = `#${((1 << 24) + (Math.round(this.r * 255) << 16) + (Math.round(this.g * 255) << 8) + Math.round(this.b * 255)).toString(16).slice(1)}`;
+    toString(alpha, asArray) {
```
willeastcott, line 236: Worth updating this?
https://github.com/playcanvas/engine/blob/main/test/core/math/color.test.mjs#L176

src/core/math/color.js
```diff
@@ -226,24 +226,31 @@ class Color {
      * is the same format as used in HTML/CSS.
      *
      * @param {boolean} alpha - If true, the output string will include the alpha value.
+     * @param {boolean} asArray - If true, the output will be an array of numbers.
```
willeastcott, line 229:
```suggestion
     * @param {boolean} [asArray] - If true, the output will be an array of numbers. Defaults to false.
```

willeastcott (approved):

PR #7242, opened 2024-12-27
Bun build support

kpal81xd:
This PR stems from attempting to build the engine from source as a submodule within a bun-based project. Ideally, the engine should be able to be built in bun successfully as a submodule without conflicting with the parent project node module types.

### Overview
- Fixes type searching to be limited to local node modules (stops looking in folders above if engine is submodule)


willeastcott:
Can you add some context about the motivation for this PR please?

willeastcott:
So what's the Bun lockfile? How and when does it get updated?

kpal81xd:
> So what's the Bun lockfile? How and when does it get updated?

So the bun lockfile works the same as package-lock.json but instead of being a json object its a binary.

willeastcott:
> > So what's the Bun lockfile? How and when does it get updated?
> 
> So the bun lockfile works the same as package-lock.json but instead of being a json object its a binary.

I expressed some concern to ChatGPT for its take and it chimed with what I was thinking:

> Your inclination not to commit a bun.lockb file to the repository is valid, especially given the PlayCanvas Engine's primary reliance on npm as its package manager. Including a lock file for a package manager that the repository doesn't officially support can introduce confusion, increase maintenance overhead, and potentially give the impression that bun is officially supported when it is not.

If you're using a submodule, can't you use a local override or generate a file into your submodule with a pre-build step or something?

kpal81xd:
> > > So what's the Bun lockfile? How and when does it get updated?
> 
> > 
> 
> > So the bun lockfile works the same as package-lock.json but instead of being a json object its a binary.
> 
> 
> 
> I expressed some concern to ChatGPT for its take and it chimed with what I was thinking:
> 
> 
> 
> > Your inclination not to commit a bun.lockb file to the repository is valid, especially given the PlayCanvas Engine's primary reliance on npm as its package manager. Including a lock file for a package manager that the repository doesn't officially support can introduce confusion, increase maintenance overhead, and potentially give the impression that bun is officially supported when it is not.
> 
> 
> 
> If you're using a submodule, can't you use a local override or generate a file into your submodule with a pre-build step or something?

So I've tested building the engine with bun and that works fine too with the same package lock. Yes you do not need to have bun lock file but for installation performance it's advised to have it by Bun themselves.

As for the using the engine as a sub module the types scoping is required regardless as types of the sub module will conflict is you are using a different runtime in the engine to your parent project.

willeastcott:
Can we compromise on the types update then and back out the bun lockfile for now?

kpal81xd:
> Can we compromise on the types update then and back out the bun lockfile for now?

Yep that works for me

willeastcott (approved):

PR #7248, opened 2025-01-06
Support 1 and 2 bands of SH in uncompressed gsplat data

slimbuck:
This PR updates gsplat rendering to support 1 and 2 bands of spherical harmonic data.

Up until now scenes had either 0 or 3 bands. However users can now generate 1 or 2 band scenes.

### DIFF
--- a/src/scene/gsplat/gsplat.js
+++ b/src/scene/gsplat/gsplat.js
@@ -48,7 +48,7 @@ class GSplat {
     /** @type {Texture} */
     transformBTexture;
 
-    /** @type {number} */
+    /** @type {0 | 1 | 2 | 3} */
     shBands;
 
     /** @type {Texture | undefined} */
@@ -318,7 +318,7 @@ class GSplat {
             for (let j = 0; j < numCoeffs; ++j) {
                 c[j * 3] = src[j][i];
                 c[j * 3 + 1] = src[j + numCoeffs][i];
-                c[j * 3 + 2] = src[j + numCoeffs][i];
+                c[j * 3 + 2] = src[j + numCoeffs * 2][i];
             }
 
             // calc maximum value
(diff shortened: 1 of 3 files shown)


src/scene/gsplat/gsplat.js
```diff
@@ -120,12 +130,18 @@ class GSplat {
...
         result.setParameter('numSplats', this.numSplatsVisible);
-        if (this.hasSH) {
-            result.setDefine('SH_BANDS', 3);
+        if (this.shBands > 0) {
+            result.setDefine('SH_BANDS', this.shBands);
             result.setParameter('splatSH_1to3', this.sh1to3Texture);
-            result.setParameter('splatSH_4to7', this.sh4to7Texture);
-            result.setParameter('splatSH_8to11', this.sh8to11Texture);
-            result.setParameter('splatSH_12to15', this.sh12to15Texture);
+            if (this.shBands > 1) {
```
mvaligursky, line 136: it'd be simpler to keep the original code and simply let it assign parameters as null or so

mvaligursky (approved):

src/scene/gsplat/gsplat.js
slimbuck: I would have except 2 bands uses `splatSH_8` and 3 bands uses `splatSH_8to11`...

slimbuck: I mean, maybe that's fine too

willeastcott (approved):

src/scene/gsplat/gsplat.js
```diff
@@ -48,15 +48,18 @@ class GSplat {
     /** @type {Texture} */
     transformBTexture;
 
-    /** @type {Boolean} */
-    hasSH;
+    /** @type {number} */
```
willeastcott, line 51:
```suggestion
    /** @type {0 | 1 | 2 | 3} */
```

slimbuck: done, thanks

PR #7256, opened 2025-01-09
[Fix] Remove PVRT texture formats with sRGB encoding as not supported on any platform

mvaligursky:
PTRTC textures are stored in non sRGB equivalent, and StandardMaterial generated shaders automatically convert sRGB values to linear by shader code.

Note that this means PTRTC cannot be used in shaders that do not handle automatic conversion (for example Particle System textures).


willeastcott (approved):
Deleting stuff is so satisfying!

mvaligursky:
> Deleting stuff is so satisfying!

Less so considering I only added those 2 months ago :)

PR #7288, opened 2025-01-16
Script Types & ESLint

kpal81xd:
- Adds eslint rule for ESM scripts to eslint config
 - Adds root types to re-export playcanvas types for ESM scripts to use correct typing

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -47,7 +47,8 @@ export default [
     {
         files: ['scripts/**/*.mjs'],
         rules: {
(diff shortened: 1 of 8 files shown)


willeastcott (approved):
Looks good. But wondering if we could keep `playcanvas.d.ts` out of the root.

kpal81xd:
> Looks good. But wondering if we could keep `playcanvas.d.ts` out of the root.

Yea unfortunately not since it exploits the fact that the `'playcanvas'` import is actually a relative path with no extension and not a module

PR #7289, opened 2025-01-16
Refactor first-person controller script

marklundin:
Refactors First Person Controller script to use `initialize()` instead of the constructor

### DIFF
--- a/scripts/esm/first-person-controller.mjs
+++ b/scripts/esm/first-person-controller.mjs
@@ -713,7 +713,6 @@ class FirstPersonController extends Script {
      */
     _gamePadInput;
 
-
     /**
      * @type {number}
      * @private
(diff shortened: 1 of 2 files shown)


scripts/esm/first-person-controller.mjs
```diff
@@ -907,11 +873,10 @@ class FirstPersonController extends Script {
      * @range [0, 0.4]
      */
     set mobileDeadZone(value) {
-        if (value === this._mobileDeadZone) {
```
kpal81xd, line 910: Here make sure to account for when value is null by using ??

kpal81xd (approved):

scripts/esm/first-person-controller.mjs
```diff
@@ -942,7 +955,7 @@ class FirstPersonController extends Script {
      * @range [0, 0.4]
      */
     set gamePadDeadZoneLow(value) {
-        this._gamePadDeadZoneLow = value;
+        this._gamePadDeadZoneLow = value ?? this._gamePadDeadZoneLow;
         if (this._gamePadInput) {
             this._gamePadInput.deadZoneLow = value;
```
marklundin, line 960: Since we're checking for nullish, should this not be `this._gamePadInput.deadZoneLow
 = this._gamePadDeadZoneLow`

PR #7296, opened 2025-01-18
Update the AABB during rendering only if it hasn't already been updated in the current frame

querielo:


src/scene/mesh-instance.js
```diff
@@ -942,10 +945,11 @@ class MeshInstance {
      * date, which forward-renderer takes care of. This function should not be called elsewhere.
      *
      * @param {Camera} camera - The camera to test visibility against.
+     * @param {number} aabbUpdateIndex - An index used to detect if AABB has already been updated this frame.
      * @returns {boolean} - True if the mesh instance is visible by the camera, false otherwise.
      * @ignore
      */
-    _isVisible(camera) {
+    _isVisible(camera, aabbUpdateIndex) {
```
querielo, line 952: Should I use `_isVisible(camera, aabbUpdateIndex=NaN) {` here?

PR #7298, opened 2025-01-20
Unified handling of core shader generation defines

mvaligursky:
In a move towards define driven uber-shaders:

- all core shader generation defines are handled in a unified way, mostly outside of the generators themselves, and are supplied as options. (GAMMA, TONEMAP, FOG, RENDERPASS, ..)
- removed gamma1_0 and gamma 2_2 chunks, and created a single gamma chunk driven by a define
- removed fogExp, forExp2, forLinear, fogNone chunks and create a single fog chunk driven by a define

Added a `Test/Global Shader Properties` example:
- hidden from published example, only for local testing (and so it's not very pretty)
- renders instances of all shaders generators (sky, gsplat, lit-shader, standard-material, particles, shader-material)
- enables tonemapping, fog and gamma to be changed to validate its functionality
- currently gsplat does not support fog, all else works

Screenshot 2025-01-20 at 12 52 31


scripts/esm/camera-frame.mjs
```diff
@@ -1,6 +1,6 @@
 // Camera Frame v 1.0
 
-/* eslint-disable-next-line import/no-unresolved */
+
```
willeastcott, line 3:
```suggestion
```

willeastcott (approved):

PR #7306, opened 2025-01-21
Do not log warning when material data contains sheenGlossTint, which was removed

mvaligursky:
remove one of the warnings

image


src/scene/materials/standard-material-parameters.js
```diff
@@ -188,7 +188,8 @@ const standardMaterialRemovedParameters = {
     diffuseTint: 'boolean',
     sheenTint: 'boolean',
     conserveEnergy: 'boolean',
-    useGamma: 'boolean'
+    useGamma: 'boolean',
```
kpal81xd, line 191: Maybe worth adding useGammaTonemap too

kpal81xd (approved):
Approving with 1 comment

PR #7312, opened 2025-01-27
VecN/Color array conversion support

kpal81xd:
### Motivation
When using Vectors/Colors it is very common to want to serialize and deserialize data to either be used for UI (in conjunction with observers) or to synchronise over a network

### Overview
- Adds `fromArray(offset: number = 0)` methods on VecN and Color classes to allow setting from arrays
- Adds `toArray(out: any[], offset: number = 0)` methods to VecN and Color classes to for explicit conversion to array (avoids potential class fields being added from using `Object.values`

### Testing
- [x] Add unit tests


src/core/math/vec2.js
```diff
@@ -572,8 +572,13 @@ class Vec2 {
      * console.log("The result of the vector set is: " + v.toString());
      */
     set(x, y) {
-        this.x = x;
-        this.y = y;
+        if (x.length === 2) {
```
slimbuck, line 575:
```suggestion
        if (x?.length === 2) {
```

src/core/math/color.js
```diff
@@ -112,17 +112,25 @@ class Color {
...
      * @param {number} [a] - The value for the alpha (0-1), defaults to 1.
      * @returns {Color} Self for chaining.
      */
     set(r, g, b, a = 1) {
-        this.r = r;
-        this.g = g;
-        this.b = b;
-        this.a = a;
+        const length = r.length;
+        if (length === 3 || length === 4) {
```
mvaligursky, line 123: I would not want commonly used functions to gain conditions. I would create custom functions. for example `setFromArray`

kpal81xd: I was thinking that as the alternative if there was a performance cost

src/core/math/vec2.js
mvaligursky: x is not optional parameter, no need to `?`?

Maksims:
I wander if it should be on `set` method or `copy` method. As copy - feels a bit more appropriate, as it copies from another thing. While set using array - feels weird, as Color or Vec - is not an array-like structure.

src/core/math/color.js
Maksims: If other object will be provided here with `length` property, it will get confused. So a more naive to JS checking if it is array is probably better: `Array.isArray(r)`

kpal81xd: Doing some testing `typeof` is slightly faster than using `Array.isArray` so I would use that. Probably still better to hoist it into a separate method however

src/core/math/vec2.js
kpal81xd: Tests are failing apparently so its being set to undefined somewhere but thats a separate issue

kpal81xd:
> I wander if it should be on `set` method or `copy` method. As copy - feels a bit more appropriate, as it copies from another thing. While set using array - feels weird, as Color or Vec - is not an array-like structure.

Copy implies that the target you are copying from is of the same type.

src/core/math/color.js
```diff
@@ -220,6 +220,25 @@ class Color {
...
+     * Set the values of the vector from an array.
+     *
+     * @param {number[]} array - The array to set the vector values from.
+     * @returns {Color} Self for chaining.
+     * @example
+     * const c = new pc.Color();
+     * c.fromArray([1, 0, 1, 1]);
+     * // c is set to [1, 0, 1, 1]
+     */
+    fromArray(array) {
```
mvaligursky, line 233: ChatGPT is suggesting this:
```
fromArray([this.r, this.g, this.b, this.a = 1]) {
    return this;
}
```

kungfooman:
I understand the wish for `#toArray`, but not the updat efor `#set`, since that's just `.set(...arr)`

src/core/math/color.js
```diff
@@ -254,6 +273,23 @@ class Color {
...
+     * Converts the color to an array of numbers.
+     *
+     * @param {boolean} [alpha] - If true, the output array will include the alpha value.
+     * @returns {number[]} The color as an array of numbers.
+     * @example
+     * const c = new pc.Color(1, 1, 1);
+     * // Outputs [1, 1, 1, 1]
+     * console.log(c.toArray(true));
+     */
+    toArray(alpha) {
```
mvaligursky, line 286: and this here

```
toArray(alpha) {
    return alpha ? [this.r, this.g, this.b, this.a] : [this.r, this.g, this.b];
}
```

src/core/math/color.js
kpal81xd: Yea that nots valid syntax lol but i can write this instead
```js
    fromArray(array) {
        [this.r, this.g, this.b, this.a = 1] = array;

        return this;
    }
```

kpal81xd:
> I understand the wish for `#toArray`, but not the update for `#set`, since that's just `.set(...arr)`
> 
> Edit: same goes for `#fromArray`, we have always argued against bloating the API when it is not necessary

It was mainly from typechecking standpoint as we can convert from type `number[]` whereas `set(...arr)` requires arr to be a tuple `[number, number, ...]`. Yes you can cast but the majority of the time you will be dealing with `number[]` over tuples. This can also get super annoying converting from an array multiple times and usually ends up just being added anyway as a helper.

kungfooman:
So `#fromArray` accepts `number[]` as an argument, which implies that even `vec3.set([1])` is "valid" now (so we are ending up with `{x: 1, y: undefined, z: undefined}`). You just watered down type-checking.

kpal81xd:
> So `#fromArray` accepts `number[]` as an argument, which implies that even `vec3.set([1])` is "valid" now (so we are ending up with `{x: 1, y: undefined, z: undefined}`). You just watered down type-checking.

That can easily be fixed by just added undefined defaults

LeXXik:
Can we add an optional index to the to/from methods? I have similar methods, but I use it with buffers and need to point the values location.

kpal81xd:
> Can we add an optional index to the to/from methods? I have similar methods, but I use it with large arrays and need to point to the values location.

I can do it for `fromArray` but `toArray` just generates a new array

src/core/math/color.js
```diff
@@ -254,6 +275,30 @@ class Color {
...
+     * a new array is created.
+     * @param {number} [offset] - The zero-based index at which to start copying elements to the
+     * array. Default is 0.
+     * @returns {number[]} The color as an array of numbers.
+     * @example
+     * const c = new pc.Color(1, 1, 1);
+     * // Outputs [1, 1, 1, 1]
+     * console.log(c.toArray(true));
+     */
+    toArray(alpha, out = [], offset = 0) {
```
willeastcott, line 292: I'm not keen on `alpha` being the first parameter here. It would be nice to have symmetry for arguments for `fromArrray` and `toArray` (e.g. first arg is `array` and second arg is `offset`). Then you could add `alpha` as the third arg for `Color`.

willeastcott: Also (super minor thing buuuut...), share the same name for the array arg for both `fromArray` and `toArray`: `array` or `arr` maybe. But not keen on `out`.

willeastcott (approved):

marklundin:
@kpal81xd Is there something like `ArrayLike<number>` we can use? `number[]` isn't happy with TypedArrays.

mvaligursky:
> @kpal81xd Is there something like `ArrayLike<number>` we can use here? `@type number[]` isn't happy with TypedArrays.

I used {number[]|ArrayBufferView} in few places.

src/core/math/vec2.js
```diff
@@ -655,6 +674,26 @@ class Vec2 {
...
+    /**
+     * Converts the vector to an array.
+     *
+     * @param {number[]} [arr] - The array to populate with the color components. If not specified,
+     * a new array is created.
+     * @param {number} [offset] - The zero-based index at which to start copying elements to the
+     * array. Default is 0.
+     * @returns {number[]} The vector as an array.
+     * @example
+     * const v = new pc.Vec3(20, 10);
```
kungfooman, line 686: And this copypaste had to become `Vec2`

mvaligursky:
> If array is provided into `toArray`, should it check if it can fit values in, to avoid array resizing within that function?

I would not expect so, that's not worth the cost overhead. It's up to the user to make sure array is in the right size if they don't want resize.

kungfooman:
> If array is provided into `toArray`, should it check if it can fit values in, to avoid array resizing within that function?

Aren't we just reinventing an opinionated/non-configurable `Array#splice`?

What if we just did:

```js
pc.Vec3.prototype[Symbol.iterator] = function *() {
  yield this.x;
  yield this.y;
  yield this.z;
}
a = new pc.Vec3(1, 2, 3);
b = new pc.Vec3(11, 22, 33);
c = new pc.Vec3(111, 222, 333);
[...a, ...b, ...c];
arr = [10, 20, 30, 40, 50, 60, 70, 80]
console.log(...arr);
// At pos 5, delete three add three from a at same place:
arr.splice(5, 3, ...a);
console.log(...arr);
// At pos 2, delete nothing, splice c into arr
arr.splice(2, 0, ...c);
console.log(...arr);
```

Outputs:

image

PR #7313, opened 2025-01-28
Camera controls focus damping

kpal81xd:
The damping for the rotation and position affect the focusing logic. This separates out the focusing damping from the user induced rotation and position damping for orbiting/flying/panning.


scripts/esm/camera-controls.mjs
```diff
@@ -222,6 +230,14 @@ class CameraControls extends Script {
      */
     enableFly = true;
 
+    /**
+     * @attribute
+     * @title Focus Damping
+     * @description The focus damping. A higher value means more damping. A value of 0 means no damping.
```
willeastcott, line 236: Maybe explicitly mention the `focus` function here?

willeastcott: Don't we just put JSDoc descriptions at the top of the block and not use the `@description` tag?

willeastcott: Sure, but leaving out the `@description` tags is considered the 'old' way of doing it. Modern JSDoc blocks drop the tag (its implicit) - tagless text should be treated as the description for the block.

kpal81xd: Ahh yes i do vaguely remember Mark mentioning. I will submit a follow up PR since there will be multiple scripts I will need to do this for

marklundin: Yep the `@description` tag isn't necessary for the attributes tooltip. It will just take the first sentence of the summary which is generally the first block of text . https://github.com/playcanvas/attribute-parser/blob/aea38f4731736d58068a89f37dbc3e059e032b0f/test/fixtures/misc.valid.js#L4-L9

PR #7315, opened 2025-01-28
Script Destroy

kpal81xd:
- Fixes issue where destroy callback is not fired if a class is provided to the script component create method


willeastcott (approved):

mvaligursky:
If not already, please merge this to V1 as well @kpal81xd

PR #7323, opened 2025-01-31
Improvements to sRGB support by the Texture

mvaligursky:
### New API

- `Texture.srgb(boolean)` - allows the format of the texture to be switched between sRGB <-> linear. Expensive operation, intended mostly for use by the Editor, to give user an instant impact of the change. Warns of the high cost.

### Fixes

- fix to PIXELFORMAT_ETC2_SRGB format, which was not using SRGB format and so the texture was not being automatically decoded to linear format during sampling

### Other changes

- basis decoding used to prefer ETC1 format over ETC2 formats for RGB textures. This works the other way now, as support for these formats on android is 99.99% anyways, and ETC2 format has sRGB equivalent

### DIFF
--- a/src/platform/graphics/texture.js
+++ b/src/platform/graphics/texture.js
@@ -15,9 +15,8 @@ import {
     isIntegerPixelFormat, FILTER_NEAREST, TEXTURELOCK_NONE, TEXTURELOCK_READ,
     TEXPROPERTY_MIN_FILTER, TEXPROPERTY_MAG_FILTER, TEXPROPERTY_ADDRESS_U, TEXPROPERTY_ADDRESS_V,
     TEXPROPERTY_ADDRESS_W, TEXPROPERTY_COMPARE_ON_READ, TEXPROPERTY_COMPARE_FUNC, TEXPROPERTY_ANISOTROPY,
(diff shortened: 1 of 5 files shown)


src/platform/graphics/texture.js
```diff
@@ -781,6 +790,53 @@ class Texture {
...
 
+    /**
+     * Sets the texture's internal format to an sRGB or linear equivalent of its current format.
+     * When set to true, the texture is stored in sRGB format and automatically converted to linear
+     * space when sampled. When set to false, the texture remains in a linear format. Changing this
+     * property recreates the texture on the GPU, which is an expensive operation, so it is
+     * preferable to create the texture with the correct format from the start. If the texture
+     * format has no sRGB variant, this operation is ignored.
+     *
+     * @type {boolean}
```
kpal81xd, line 801: Added `@ignore` so its hidden from docs?

kpal81xd (approved):
Approving with 1 comment

src/platform/graphics/texture.js
mvaligursky: I don't want to hide it.

PR #7325, opened 2025-02-03
Fix to msdf chunk to handle gamma space correctly

mvaligursky:
related to change where UI shader is in linear space: https://github.com/playcanvas/engine/pull/6714
and this forum reported issue: https://forum.playcanvas.com/t/textelement-outline-appears-darker-in-engine-v2/38945

This fixes the applyMsdf to operate in linear space as well

### DIFF
--- a/src/scene/shader-lib/chunks/common/frag/msdf.js
+++ b/src/scene/shader-lib/chunks/common/frag/msdf.js
@@ -26,6 +26,12 @@ varying vec2 shadow_offset;
 #endif
 
 vec4 applyMsdf(vec4 color) {
+
+    // Convert to linear space before processing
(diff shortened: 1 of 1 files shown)


src/scene/shader-lib/chunks/common/frag/msdf.js
```diff
@@ -26,6 +26,12 @@ varying vec2 shadow_offset;
 #endif
 
 vec4 applyMsdf(vec4 color) {
+
+    // Convert to linear space before processing
+    // TODO: ideally this would received the color in linear space, but that would require larger changes
```
willeastcott, line 31:
```suggestion
    // TODO: ideally this would receive the color in linear space, but that would require larger changes
```

willeastcott (approved):

PR #7336, opened 2025-02-07
When Texture.srgb is changed, shaders are reevaluated to be recreated

mvaligursky:
- this is depended on by the Editor’s inspector, when the srgb property of the texture changes. Shaders need to be updated to respect new texture encoding.

### DIFF
--- a/src/platform/graphics/graphics-device.js
+++ b/src/platform/graphics/graphics-device.js
@@ -392,6 +392,14 @@ class GraphicsDevice extends EventHandler {
         height: 0
     };
 
+    /**
+     * A very heavy handed way to force all shaders to be rebuilt. Avoid using as much as possible.
+     *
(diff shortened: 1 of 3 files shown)


src/platform/graphics/graphics-device.js
```diff
@@ -392,6 +392,14 @@ class GraphicsDevice extends EventHandler {
         height: 0
     };
 
+    /**
+     * A very heavy handed way to force all shaders to be rebuilt. Avoid using as much as possible.
+     *
+     * @type {boolean}
+     * @ignore
+     */
+    _shadersDirty = false;
```
slimbuck, line 401: Wondering whether 'forceRebuildShaders` or similar is more descriptive / scary?

slimbuck (approved):

Issue #7297, opened 2025-01-20
Manually created batches can't be destroyed using BatchManager


OlegGedzjuns:
### Description
It is possible to manually prepare and create batches using BatchManager `prepare()` and `create()` methods.
BatchManager also has `destroyBatch()` method, but calling it with a manually created batch causes a crash.

Link to example: https://playcanvas.com/project/1295595/overview/manual-batching-issue

Also according to destroy method implementation, batches from `_batchList` array are never removed.
https://github.com/playcanvas/engine/blob/9e277aa3aa1e119d918b9c87d5b13ece553bb5fe/src/scene/batching/batch-manager.js#L927

mvaligursky:
you should call batch.destroy instead of private batcher.destroyBatch.
I just made it public.

```
        //this.app.batcher.destroyBatch(batch);
        batch.destroy(this.app.scene, [pc.LAYERID_WORLD]);
```

PR #7340, opened 2025-02-10
Make Batch.destroy public as it’s useful for destroying manually created batch groups

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/7297

### DIFF
--- a/src/scene/batching/batch.js
+++ b/src/scene/batching/batch.js
@@ -3,6 +3,7 @@ import { BoundingBox } from '../../core/shape/bounding-box.js';
 
 /**
  * @import { MeshInstance } from '../mesh-instance.js'
+ * @import { Scene } from '../scene.js'
  */
 
 /**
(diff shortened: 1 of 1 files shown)


kpal81xd (approved):

src/scene/batching/batch.js
```diff
@@ -57,7 +58,12 @@ class Batch {
         this.batchGroupId = batchGroupId;
     }
 
-    // Removes the batch meshes from all layers and destroys it
+    /**
+     * Removes the batch meshes from the layers and destroys it.
```
willeastcott, line 62: This should either be:

```
Removes the batch meshes from the layers and destroys them.
```

Or:

```
Removes the batch from the layers and destroys it.
```

willeastcott: Or maybe:

```
Removes the batch's mesh instance from its layers and destroys it.
```

willeastcott (approved):

willeastcott (approved):

PR #7343, opened 2025-02-11
Skybox shaders on WebGPU use native WGSL chunks

mvaligursky:
- upgrade to ShaderMaterial shader generation to better support WGSL shaders
- converted skydome shaders to use native WGSL shader generation and avoid GLSL->WGSL transpilation completely

Note: Fog does not currently functions as `gl_FragCoord.z / gl_FragCoord.w` needs replacement - this are not global variables in WGSL

### DIFF
--- a/src/scene/shader-lib/chunks-wgsl/common/frag/fog.js
+++ b/src/scene/shader-lib/chunks-wgsl/common/frag/fog.js
@@ -1,0 +1,44 @@
+export default /* glsl */`
+
+#if (FOG != NONE)
+    uniform fog_color : vec3f;
+    
+    #if (FOG == LINEAR)
+        uniform fog_start : f32;
+        uniform fog_end : f32;
+    #else
+        uniform fog_density : f32;
+    #endif
+#endif
+
+uniform dBlendModeFogFactor : f32;
+
+fn getFogFactor() -> f32 {
+    //let depth = gl_FragCoord.z / gl_FragCoord.w;
+    let depth = 1.0;
+
+
+
+
+
+    var fogFactor : f32 = 0.0;
+
+    #if (FOG == LINEAR)
+        fogFactor = (uniform.fog_end - depth) / (uniform.fog_end - uniform.fog_start);
+    #elif (FOG == EXP)
+        fogFactor = exp(-depth * uniform.fog_density);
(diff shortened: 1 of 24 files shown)


src/scene/shader-lib/chunks-wgsl/common/frag/fog.js
```diff
@@ -0,0 +1,44 @@
...
+        uniform fog_end : f32;
+    #else
+        uniform fog_density : f32;
+    #endif
+#endif
+
+uniform dBlendModeFogFactor : f32;
+
+fn getFogFactor() -> f32 {
+    //let depth = gl_FragCoord.z / gl_FragCoord.w;
```
willeastcott, line 17: Delete?

src/scene/shader-lib/chunks-wgsl/common/frag/fog.js
```diff
@@ -0,0 +1,44 @@
...
+    let depth = 1.0;
+
+
+
+
```
willeastcott, line 22: Wow, you love new lines! 🤣

mvaligursky: hah, those are to make it obvious the code is unfinished .. a place I need to clean up before creating PR :)

willeastcott:
Rather than having a `chunks-wgsl` folder, would it be better to place chunks side by side? I can imagine wanting to quickly compare a GLSL vs a WGSL chunk and having them in the same folder could be handy. Maybe a naming like `fog.glsl.js` and `fog.wgsl.js`?

mvaligursky:
> Rather than having a `chunks-wgsl` folder, would it be better to place chunks side by side? I can imagine wanting to quickly compare a GLSL vs a WGSL chunk and having them in the same folder could be handy. Maybe a naming like `fog.glsl.js` and `fog.wgsl.js`?

I actually want to rename the current `chunks` folder to `chunks-glsl` in the follow up PR and leave those completely separate, as there will not be 1:1 mapping all the time.

Maksims:
What about to have shader files as their actual extension, so glsl and wgsl files would be in same location.
This also will enable syntax highlight correctly.
And searching through files will be improved.

And resolve imports as a part of build process?

PR #7346, opened 2025-02-12
Processor handles loop of includes

mvaligursky:
Includes support loop, with an index injection into the included text. This will be used for shader generation of looped objects, for example lights.

This syntax can be now used:
```
#include "lightChunk, LIGHTCOUNT"
```

from JS, it can be driven this way:
```
defines.set('LIGHTCOUNT', lightCount);
for (i = 0, i < lightCount; i++) {
	defines.set(`LIGHTTYPE${i}`, 'omni');     // #define LIGHTTYPE0 omni
```

lightChunk example (part that gets included). All instances of `{i}` are replaced by the loop index:
```
#ifdef LIGHTTYPE{i} == omni
	uniform vec3 light{i}Color
#endif

```

### DIFF
--- a/src/core/preprocessor.js
+++ b/src/core/preprocessor.js
@@ -416,7 +416,7 @@ class Preprocessor {
                                     includeSource = result;
 
                                 } else {
-                                    console.error(`include Count identifier "${countIdentifier}" not resolved while preprocessing ${Preprocessor.sourceName} on line: ${source.substring(match.index, match.index + 100)}...`, { source: originalSource });
+                                    console.error(`Include Count identifier "${countIdentifier}" not resolved while preprocessing ${Preprocessor.sourceName} on line: ${source.substring(match.index, match.index + 100)}...`, { source: originalSource });
                                     error = true;
                                 }
                             }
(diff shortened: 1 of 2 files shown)


src/core/preprocessor.js
```diff
@@ -388,15 +391,37 @@ class Preprocessor {
...
+
+                                    // add the include count times
+                                    let result = '';
+                                    for (let i = 0; i < count; i++) {
+                                        result += includeSource.replace(LOOP_INDEX, String(i));
+                                    }
+                                    includeSource = result;
+
+                                } else {
+                                    console.error(`include Count identifier "${countIdentifier}" not resolved while preprocessing ${Preprocessor.sourceName} on line: ${source.substring(match.index, match.index + 100)}...`, { source: originalSource });
```
willeastcott, line 419:
```suggestion
                                    console.error(`Include Count identifier "${countIdentifier}" not resolved while preprocessing ${Preprocessor.sourceName} on line: ${source.substring(match.index, match.index + 100)}...`, { source: originalSource });
```

Maksims:
Would square brackets be more suiting?
```
#ifdef LIGHTTYPE[i] == omni
```

mvaligursky:
- `[i]` would increase the chance of incorrectly replacing `[i]` you're not expecting, as that is often use to address arrays. `{i}` is more unique and unlikely to cause conflicts
- '{i}' also closer matches to '${i}` in JS side.

PR #7347, opened 2025-02-12
Add visibility conditions for SSAO, Bloom, Grading, Vignette, Fringin…

marklundin:
Added conditional attribute tags to the  `camera-frame.mjs` to show and hide them

### DIFF
--- a/scripts/esm/camera-frame.mjs
+++ b/scripts/esm/camera-frame.mjs
@@ -1,5 +1,3 @@
-// Camera Frame v 1.0
-
 import { CameraFrame as EngineCameraFrame, Script, Color } from 'playcanvas';
 
 /** @enum {number} */
@@ -108,10 +106,14 @@ class Ssao {
      */
     type = SsaoType.NONE;
 
+    /**
+     * @visibleif {type !== 'none'}
+     */
     blurEnabled = true;
 
     /**
      * @range [0, 1]
+     * @visibleif {type !== 'none'}
      * @precision 3
      * @step 0.001
      */
@@ -119,6 +121,7 @@ class Ssao {
 
     /**
      * @range [0, 100]
+     * @visibleif {type !== 'none'}
      * @precision 3
      * @step 0.001
      */
@@ -126,6 +129,7 @@ class Ssao {
 
     /**
      * @range [1, 64]
+     * @visibleif {type !== 'none'}
      * @precision 0
      * @step 1
      */
@@ -133,6 +137,7 @@ class Ssao {
 
     /**
      * @range [0.1, 10]
+     * @visibleif {type !== 'none'}
      * @precision 3
      * @step 0.001
      */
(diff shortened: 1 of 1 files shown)


scripts/esm/camera-frame.mjs
```diff
@@ -108,45 +106,54 @@ class Ssao {
      */
     type = SsaoType.NONE;
 
+    /**
+     * @visibleif {type !== 'none'}
```
willeastcott, line 110: @mvaligursky This is another reason why I wanted an `enabled` property on SSAO. It's gonna look quite incongruous against the other effects in the Editor UI (where everything else can be toggle in the effect header bar.

scripts/esm/camera-frame.mjs
```diff
@@ -158,6 +165,7 @@ class Bloom {
     enabled = false;
 
     /**
+     * @visibleif {enabled}
```
willeastcott, line 168: This is quite elegant! 👍

willeastcott (approved):

scripts/esm/camera-frame.mjs
marklundin: could we have a `type=combine|lighting` and a separate enabled field?

mvaligursky: sure, feel free to change this script

Maksims:
Please refer to this issue: https://github.com/playcanvas/editor/issues/1264
We don't hide stuff in Editor almost ever. The only exceptions is where selected field has related fields that only relate to it, for example based on fog type, you have different settings.
Also, we only disable things where changing them is simply not possible, for example in Settings > Layers, you are not able to change default layers name. But if changing field is conditionally possible, for example if render component is enabled its field is still relevant - then when render component is disabled we do not disable it UI, as editing it would require enabling/disabling component, which is unnecessary extra movement for the user.

So please, refer to the issue described above, and lest get UX consistent within Editor.

Maksims:
> We show/hide fields in nearly all component inspectors (Light, Sound, Particle System, Camera, Render, UI). In addition we also enable/disable fields in Light Component.

If the choice of related fields depend on the type of a related field, then yes. In some edge cases like "shadow" also, but in majority cases, especially when disabling whole component - we do not hide fields, as you have it there for a reason. If you need to remove it - you have removal for that. So hiding fields based on enable/disable of a block - is not used practice. Otherwise we would hide position, rotation, scale, etc fields of Entity if it was disabled?

Same applies to all components - their enabled/disabled state, does not hide their content, as you want to adjust it regardless of its current enable/disable state. It is only for things that you are very unlikely to dynamically change, for example changing fog type - is very unlikely.

So for post-effects - this is something you want to adjust, regardless if it is enabled or disabled in Editor. Because user might enable it in settings later. That is why hiding it - is not the way. Minimizing it when selected if it is disabled - is way better approach, like we do with material inspector - we minimize blocks when they don't have related data (same as not enabled), but we still allow you to go in and see fields.

PR #7348, opened 2025-02-13
Make camera controls more sensitive to trackpad pinch

vincentwoo:
On browsers unfortunately trackpad pinch-zoom is sent over as a wheel event. However, the magnitude of these events are decidedly smaller in practice, leading to very unsatisfying pinch zooming. Multiplying deltas by 10 when they are less than 25 in my testing is an effective fix for this.

Fixes #

### DIFF
--- a/scripts/esm/camera-controls.mjs
+++ b/scripts/esm/camera-controls.mjs
@@ -710,7 +710,9 @@ class CameraControls extends Script {
      */
     _onWheel(event) {
         event.preventDefault();
-        this._zoom(event.deltaY);
+        let delta = event.deltaY;
+        delta = Math.abs(delta) > 25 ? delta : delta * 10;
+        this._zoom(delta);
     }
 
     /**


Maksims:
I would be very hesitant to add such magical behaviour as people use different scroll speeds based on their OS's, and magical number of 25 and then 10 - might be unconsistent across various devices.
Especially considering it will change behavior for existing projects.

What would be a better solution, is to detect what kind of input it is from the event, to differentiate between mouse/touch/touchpad, and then do a decision based on that. It requires a various testing across different platforms and devices.
And perhaps for now you can do that decision on application level, instead of engine level.

vincentwoo:
Yeah I think that is perfectly reasonable, but I also think this hack is quite easy to delete when you build the input differentiation system. Up to you!

willeastcott:
Hey @kpal81xd, this PR has been open for a few months now. Can you bring it to a conclusion when you have a moment please?

willeastcott:
Just bumping this again, @kpal81xd - can you make a call on what you want to do with this please?

kpal81xd:
I agree with @Maksims having an arbitrary magic number is not ideal - would be better to detect when its a trackpad pinch and deal with it that way.

Tracked in #8533.

Issue #7351, opened 2025-02-14
Post Effects break when canvas display is changed


marklundin:
If you toggle the canvas visibility using the `display:none;` when using the Camera Frame, the post effects will stop working

Repro
1. Open [this example](https://stackblitz.com/edit/vitejs-vite-qzo4hgkp?file=src%2Fmain.js)
2. Hide the canvas by clicking "Toggle Canvas Visibility" button
3. Show the canvas by pressing the button again
4. Nothing is rendered

Note that if you remove the CameraFrame, toggling the visibility of the canvas will work, and the scene will render correctly. It looks like a render target is not being resized in the render passes as they have a width and height of 0.

mvaligursky:
duplicate of https://github.com/playcanvas/engine/issues/5806
improved the situation here: https://github.com/playcanvas/engine/pull/7354

closing, as the conclusion is the same as in the linked duplicate issue.

PR #7354, opened 2025-02-14
Moved zero render target checked to a better place

mvaligursky:
this is to handle single samplered RTs as well, not only multisampled

Related to https://github.com/playcanvas/engine/issues/7351

repro using engine example
Screenshot 2025-02-14 at 16 03 46

note that this is not related to CameraFrame only, but any render target rendering, even old post-effects.

### DIFF
--- a/src/platform/graphics/webgl/webgl-render-target.js
+++ b/src/platform/graphics/webgl/webgl-render-target.js
@@ -150,6 +150,12 @@ class WebglRenderTarget {
 
         } else {
 
+            Debug.call(() => {
+                if (target.width <= 0 || target.height <= 0) {
+                    Debug.warnOnce(`Invalid render target size: ${target.width} x ${target.height}`, target);
+                }
+            });
+
             // ##### Create main FBO #####
             this._glFrameBuffer = gl.createFramebuffer();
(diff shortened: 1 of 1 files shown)


src/platform/graphics/webgl/webgl-render-target.js
```diff
@@ -150,6 +150,12 @@ class WebglRenderTarget {
 
         } else {
 
+            Debug.call(() => {
```
marklundin, line 153: This will just warn though.  It won't resize the render targets back right?

mvaligursky: true.
this will cause webgl errors as well.
as discussed in the previous duplicate issue, the engine does not support this way of hiding the application, and it should not be used.

marklundin: it does work without the post though. If you do the same thing without the camera frame it will work correctly

mvaligursky: yes because normally we do not render to any render target
with camera frame or old post effects we render to render target, and currently we do not support when those are 0 size

willeastcott (approved):

PR #7367, opened 2025-02-21
ensure postrender callback is cleaned up in outline renderer

### LINKED ISSUE #7366: exception on outline renderer when a second one is created and the previous one is destroyed
When creating an outline renderer after previously destroying it and exception is thrown.  This is caused by the fact that the 'postrender' event is not unregistered when calling destroy

Image

Image

MAG-AdrianMeredith:
Fixes #7366 


mvaligursky (approved):
Nice one, thanks!

PR #7365, opened 2025-02-21
fix exception on hiding/showing grid

MAG-AdrianMeredith:
Fixes #7364 

I confirm I have read the [contributing guidelines](https://github.com/playcanvas/engine/blob/master/.github/CONTRIBUTING.md) and signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #7374, opened 2025-02-24
Fixed grid enabled/disabled

### LINKED ISSUE #7364: Grid Script throws exception when enabling after setting to disabled
It doesn't seem possible to show/hide the grid from scripts folder

Repo steps

1. go to example https://playcanvas.vercel.app/#/misc/editor
2. Modify code to set grid to globalThis
3. reload, select the correct frame in dev tools
4. type grid.enabled = false; and grid disappears as expected
5. type grid.enabled = true; and exception is thrown

Image

kpal81xd:
Fixes #7364
Extends GH-7365

- Sets visible property on meshInstance instead of removing and adding from the render component
- Cleans up render component if added by script

> **NOTE** There meshInstance will still be in the draw call list however since it is a single mesh instance the cost will be marginal

### DIFF
--- a/scripts/esm/grid.mjs
+++ b/scripts/esm/grid.mjs
@@ -12,8 +12,6 @@ import {
     Vec2
 } from 'playcanvas';
 
-/** @import { AppBase, Entity } from 'playcanvas' */
-
 const tmpVa = new Vec2();
 
 const EPISILON = 1e-3;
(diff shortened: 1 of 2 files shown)


scripts/esm/grid.mjs
```diff
@@ -320,6 +320,10 @@ class Grid extends Script {
     get resolution() {
         return this._resolution;
     }
+
+    destroy() {
+        this.entity.render.meshInstances = [];
```
willeastcott, line 325: If the script adds a render component, shouldn't destroy leave the entity in the state it was beforehand? In other words, shouldn't `removeComonent('render')` be done here instead?

willeastcott: Actually, to be precise, the script adds a render component if there wasn't one already. So maybe this should remove it if it was responsible for adding it.

willeastcott (approved):

PR #7382, opened 2025-02-26
[Fix] Create a dummy framebuffer on NullDevice to avoid undefined access

mvaligursky:
it’s required by few functions in the engine, so create a dummy RT for this purpose

### DIFF
--- a/src/platform/graphics/null/null-graphics-device.js
+++ b/src/platform/graphics/null/null-graphics-device.js
@@ -2,6 +2,7 @@ import {
     DEVICETYPE_NULL
 } from '../constants.js';
 import { GraphicsDevice } from '../graphics-device.js';
+import { RenderTarget } from '../render-target.js';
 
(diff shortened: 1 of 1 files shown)


slimbuck (approved):

src/platform/graphics/null/null-graphics-device.js
```diff
@@ -17,6 +18,14 @@ class NullGraphicsDevice extends GraphicsDevice {
         this.isNull = true;
         this._deviceType = DEVICETYPE_NULL;
         this.samples = 1;
+
+        this.backBuffer = new RenderTarget({
+            name: 'Framebuffer',
```
slimbuck, line 23: maybe call it null-framebuffer or similar

mvaligursky: had that but seemed strange ;)
anyways, I don't think this matters on Null renderer, nobody captures / debugs this as it does nothing

PR #7383, opened 2025-02-26
Fix gsplat discoloration

slimbuck:
This PR fixes an issue with some splats with spherical harmonics being discolored.

We were clamping r, g, b even with spherical harmonics.

Note that our compressed format was handling this correctly so no fix is needed there!

Before:

Screenshot 2025-02-26 at 12 26 59

After:

Screenshot 2025-02-26 at 12 27 28


mvaligursky (approved):
a big improvement!

PR #7384, opened 2025-02-26
Grid render component fix

kpal81xd:
- Forces render component to be created by script instead of provided by user

### DIFF
--- a/scripts/esm/grid.mjs
+++ b/scripts/esm/grid.mjs
@@ -180,14 +180,16 @@ class Grid extends Script {
     _resolution = Grid.RESOLUTION_HIGH;
 
     initialize() {
-
-        // ensure the entity has a render component
-        if (!this.entity.render) {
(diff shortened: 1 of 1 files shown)


scripts/esm/grid.mjs
```diff
@@ -180,15 +180,17 @@ class Grid extends Script {
...
-        // ensure the entity has a render component
-        if (!this.entity.render) {
-            this.entity.addComponent('render', {
-                castShadows: false
-            });
-            this._addedRender = true;
+        // check if the entity already has a render component
+        if (this.entity.render) {
+            console.warn('The entity already has a render component.');
+            return;
```
mvaligursky, line 186: should _addedRender be removed completely?

kpal81xd: Yea ill probably do that and i think ill change the warn to the error

mvaligursky (approved):

PR #7386, opened 2025-02-27
Adjusted syntax of variables the preprocessor uses for injection

mvaligursky:
use curly brackets around the define variable name to mark it for injection into shader code (we resolve it before passing it to platform compilers)

Before, the define had to be named
```
define _INJECT_COUNT 4
```
and the code using it would be 
```
for (int i = 0; i < _INJECT_COUNT; i++)
```

now we have
```
define {COUNT} 4
```
and the code using it would be 
```
for (int i = 0; i < {COUNT}; i++)
```


slimbuck (approved):
looks a _lot_ neater <3

LeXXik:
Does it allow spaces between the variable name and curly braces?

mvaligursky:
It does not, and will not.

PR #7395, opened 2025-03-03
Update to the way we strip unused MRT outputs

mvaligursky:
This PR updates the way we strip unused fragment outputs (MRT related) due to an [issue on iOS 15](https://github.com/playcanvas/engine/pull/5459).

Before, the shader code was created as a single string, and we simply prepended some defines to remove unused outputs, and let out preprocessor strip them out.

Now those outputs can be used inside includes, and so we had to change the way unused outputs are defined, to be done at the end of the preprocessing step. This excludes using defines to strip them, as we don't want to run preprocessor twice just for this, and we do custom line stripping at the end instead.

### DIFF
--- a/src/core/preprocessor.js
+++ b/src/core/preprocessor.js
@@ -42,6 +42,9 @@ const INCLUDE = /include[ \t]+"([\w-]+)(?:\s*,\s*([\w-]+))?"\r?(?:\n|$)/g;
 // loop index to replace, in the format {i}
 const LOOP_INDEX = /\{i\}/g;
 
+// matches color attachments, for example: pcFragColor1
+const FRAGCOLOR = /(pcFragColor[1-8])\b/g;
+
 /**
  * Pure static class implementing subset of C-style preprocessor.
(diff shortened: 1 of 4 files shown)


kpal81xd (approved):

src/core/preprocessor.js
```diff
@@ -123,6 +110,43 @@ class Preprocessor {
...
+            // find out how many times pcFragColorX is used (see gles3.js)
+            const counts = new Map();
+            const matches = source.match(FRAGCOLOR);
+            matches?.forEach((match) => {
+                const index = parseInt(match.charAt(match.length - 1), 10);
+                counts.set(index, (counts.get(index) ?? 0) + 1);
+            });
+
+            // if there's any attachment used only one time (only as a declaration, without actual use)
+            const anySingleUse = Array.from(counts.values()).some(count => count === 1);
```
kpal81xd, line 126: Would it be more efficient just to track the anySingleUse while iterating over the matches? It removes the count traversal

PR #7398, opened 2025-03-03
CameraControls rework (Input/Controller overhaul)

kpal81xd:
### Overview
This PR aims to split the camera controls down into two core components:

- **Input Sources** which represent a collection of deltas generated from one or more input devices. These can range from a mouse movement delta to a gamepad joystick delta. 
- **Input Controllers** which represent the constrained transformations that are to be performed e.g. orbiting around a particular focus point or camera flying (unconstrained) .

Together the deltas from an Input can be collected into a "frame" of deltas which are then applied to the specified controller to perform the required action.

### Input Sources
**SingleGestureSource** - Single touch or virtual joystick input 
**DualGestureSource** - Split screen touch or virtual joystick input
**MultiTouchSource** - Multi touch input (includes pinch handling)
**KeyboardMouseSource** - Keyboard and mouse input
**GamepadSource** - Game pad input

### Input Controllers
**FlyController** - Applies free-fly mechanics to a base transform
**OrbitController** - Applies orbiting mechanics to a base transform

### Design Doc
https://docs.google.com/document/d/1r2ywuzxbte6DwlEJzWgyD-fTDNyp-7P0PKLFefrgV9Y/edit?usp=sharing


src/extras/models/model.js
```diff
@@ -0,0 +1,31 @@
+import { Mat4 } from '../../core/math/mat4.js';
+
+class Model {
```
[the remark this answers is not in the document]
kpal81xd: Yea the naming is variable to change right now. The reason why I haven't named those controls is that they don't handle the inputs themselves rather they accept inputs and apply some transformations depending on constraints returning the final state

src/extras/inputs/joystick.js
```diff
@@ -0,0 +1,98 @@
+import { math } from '../../core/math/math.js';
+import { Vec2 } from '../../core/math/vec2.js';
+import { EventHandler } from '../../core/event-handler.js';
+
+const tmpVa = new Vec2();
+
+class Joystick extends EventHandler {
+    /**
+     * @type {string}
+     * @static
```
willeastcott, line 10: You don't need to explicitly use `@static`. `tsc` can see the `static` keyword in the code.

willeastcott: You also don't need the `@type` tag either because the field is set to a string literal. Again, `tsc` is smart enough to notice this.

willeastcott: You _do_ however, need the `@event` tag. Check out other events across the codebase. Normally, there's an `@example` block too (but this doesn't seem like a public class so probably not required here).

src/extras/inputs/joystickdouble-input.js
```diff
@@ -0,0 +1,165 @@
...
+        return this._leftJoystick;
+    }
+
+    get rightJoystick() {
+        return this._rightJoystick;
+    }
+
+    /**
+     * @private
+     * @param {PointerEvent} event - The pointer event.
```
willeastcott, line 52: Nitpick: we generally use a convention of putting visibility tags after other tags. Again, adhere to what you see elsewhere.

src/extras/inputs/gamepad-input.js
```diff
@@ -0,0 +1,50 @@
...
+            return;
+        }
+
+        this._camera = null;
+
+        super.detach();
+    }
+
+    /**
+     * @override
```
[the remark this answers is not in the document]
kpal81xd: Technically yes its just being explicit

src/extras/inputs/joystick.js
kpal81xd: So for type I just like to be explicit with typings for attributes but I can remove the static flag and add the event one

src/extras/inputs/gamepad-input.js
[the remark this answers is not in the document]
kpal81xd: I have done it before in other fields like in gizmos

src/extras/inputs/joystick.js
willeastcott: I get it. But I do think consistency is king. And including redundant tags just bloats the code ultimately.

kpal81xd: Well on the note of bloat having it in Typescript would be the best option ultimately 👀

src/extras/inputs/joystickdouble-input.js
```diff
@@ -0,0 +1,165 @@
...
+    /**
+     * @param {HTMLElement} element - The element.
+     */
+    attach(element) {
+        super.attach(element);
+
+        this._element = element;
+        this._element.addEventListener('pointerdown', this._onPointerDown);
+        this._element.addEventListener('pointermove', this._onPointerMove);
+        this._element.addEventListener('pointerup', this._onPointerUp);
```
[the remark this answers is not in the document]
kpal81xd: I ended up removing it because of some weird behavior iirc but will need to verify again

src/extras/inputs/joystick.js
willeastcott: True that. Let's hold that thought just for the time being though! 😅

willeastcott:
> The CameraControls script it now more difficult to understand from the end user since the majority of the complexity is now handled internally.

I don't think that's a big deal, provided the Editor UI for the script is easy to visually parse. Worth including a screenshot of that?

src/extras/inputs/keyboardmouse-input.js
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    /**
+     * @param {string} code - The code.
+     * @param {number} value - The value.
+     * @private
+     */
+    _setKey(code, value) {
+        switch (code) {
+            case 'KeyW':
```
willeastcott, line 165: Is there an issue with non-QWERTY keyboards (say France's AZERTY) with referencing characters rather than physical key locations for WASD?

src/extras/controllers/orbit-controller.js
```diff
@@ -0,0 +1,394 @@
...
+import { Ray } from '../../core/shape/ray.js';
+import { Plane } from '../../core/shape/plane.js';
+import { Controller } from './controller.js';
+
+/**
+ * @import { CameraComponent } from '../../framework/components/camera/component.js'
+ */
+
+/**
+ * @typedef {object} OrbitInputFrame
```
willeastcott, line 15: I have a feeling this will appear in the API reference. We have an issue right now that because we use `typedoc-plugin-missing-exports` for the engine, all `@typedef`s appear in the manual, regardless of whether they're public API or now. I meant to start an issue on TypeDoc about this.

willeastcott:
It looks like this PR doesn't add any new APIs to the API reference manual. What's the plan for that?

kpal81xd:
> > The CameraControls script it now more difficult to understand from the end user since the majority of the complexity is now handled internally.
> 
> I don't think that's a big deal, provided the Editor UI for the script is easy to visually parse. Worth including a screenshot of that?

So an example would be accumulating the desktop inputs:

image

Theres a lot of reformatting of frame inputs to accommodate the frames for the models

src/extras/controllers/orbit-controller.js
kpal81xd: Yep they were so added a manual exclusion to the typedoc config for now since i want them hidden

src/extras/inputs/keyboardmouse-input.js
kpal81xd: So the 'KeyW' represents the key code just named in the QWERTY format (see https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code)

kpal81xd:
> It looks like this PR doesn't add any new APIs to the API reference manual. What's the plan for that?

So I do not want to add any public API until this is finalized since this will change when I add support for first/third person controllers and integration into the engine platform input system also

scripts/esm/camera-controls.mjs
```diff
@@ -1,1163 +1,892 @@
...
+    JoystickTouchInputDevice,
+    KeyboardMouseInputDevice,
+    MultiTouchInputDevice,
+    FlyTransformer,
+    OrbitTransformer,
+    Script,
+    Mat4,
+    Vec2,
+    Vec3
+} from 'playcanvas';
```
willeastcott, line 14: Order imports alphabetically. TBH, I thought ESLint enforced that...not sure.

scripts/esm/camera-controls.mjs
```diff
@@ -1,1163 +1,892 @@
...
-        position.add(tmpV1);
+    get pitchRange() {
+        return this._pitchRange;
     }
 
     /**
-     * @private
-     * @param {number} value - The value to clamp.
-     * @returns {number} - The clamped value.
+     * The yaw range.
```
willeastcott, line 501: Again, state range. Degress?

src/extras/inputs/gamepad-input-device.js
```diff
@@ -0,0 +1,56 @@
+import { InputDelta, InputDevice } from './input.js';
+
+/**
+ * Game pad input device class
+ *
+ * @category InputDevice
```
willeastcott, line 6: Will the `InputDevice` and `Input` category split be confusing?

src/extras/inputs/input.js
```diff
@@ -0,0 +1,84 @@
...
+            this._instance[i] += offsets[i] || 0;
+        }
+    }
+
+    flush() {
+        this._instance.fill(0);
+    }
+}
+
+class InputDevice {
```
willeastcott, line 32: `GamepadInputDevice` subclasses this and appears in the docs so ideally, this would be there too.

src/extras/inputs/keyboardmouse-input-device.js
```diff
@@ -0,0 +1,259 @@
+import { InputDelta, InputDevice } from './input.js';
+
+/** @type {AddEventListenerOptions & EventListenerOptions} */
+const PASSIVE = { passive: false };
+
+const array9 = new Array(9).fill(0);
```
willeastcott, line 6:
```suggestion
const array9 = new Array(9).fill(0);

```

src/extras/transformers/fly-transformer.js
```diff
@@ -0,0 +1,205 @@
...
+ * @param {number} damping - The damping.
+ * @param {number} dt - The delta time.
+ * @returns {number} - The lerp rate.
+ */
+const lerpRate = (damping, dt) => 1 - Math.pow(damping, dt * 1000);
+
+/**
+ * The fly controller.
+ *
+ * @category Transformers
```
willeastcott, line 31: I'm thinking it might make more sense to group `InputDevice` and `Transformers`  classes together. Maybe under `Camera Controls`?

scripts/esm/camera-controls.mjs
kpal81xd: It does but linting is skipped in scripts because they are external - another reason as to why its probably better to have them as a separate package

src/extras/inputs/gamepad-input-device.js
kpal81xd: These will be unified together at some point but process inputs differently at the movement - one uses events the other uses a frame of delats to represent the input

PR #7399, opened 2025-03-04
Refactoring fo envMultiply and envConst chunks

mvaligursky:
merged chunks and updated their use: 
`	envMultiply and envConst -> envProc`

### DIFF
--- a/src/scene/shader-lib/chunks-wgsl/common/frag/envProc.js
+++ b/src/scene/shader-lib/chunks-wgsl/common/frag/envProc.js
@@ -1,0 +1,13 @@
+export default /* wgsl */`
+#ifdef LIT_SKYBOX_INTENSITY
+    uniform skyboxIntensity : f32;
+#endif
+
+fn processEnvironment(color : vec3f) -> vec3f {
+    #ifdef LIT_SKYBOX_INTENSITY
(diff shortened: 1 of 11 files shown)


slimbuck (approved):

src/scene/shader-lib/chunks-wgsl/common/frag/envProc.js
```diff
@@ -0,0 +1,13 @@
+export default /* wgsl */`
+#ifdef LIT_SKYBOX_INTENSITY
```
willeastcott, line 2: It is worth having a shader variant just to avoid a multiply?

src/scene/shader-lib/chunks-wgsl/common/frag/envProc.js
```diff
@@ -0,0 +1,13 @@
+export default /* wgsl */`
```
willeastcott, line 1: I'm not 100% keen on the `envProc` naming. `skybox`?

mvaligursky: we already have skybox chunk that does the skybox. This is just a small chunk used by both skybox and lit-shader

PR #7401, opened 2025-03-04
Refactor reflection-source related chunk and code

mvaligursky:
- exposed hardcoded string names representing reflection source as constants, and used those internally
```
export const REFLECTIONSRC_NONE = 'none';
export const REFLECTIONSRC_ENVATLAS = 'envAtlas';
export const REFLECTIONSRC_ENVATLASHQ = 'envAtlasHQ';
export const REFLECTIONSRC_CUBEMAP = 'cubeMap';
export const REFLECTIONSRC_SPHEREMAP = 'sphereMap';
```

- updated chunks to use defines to inject the decoding functions
```
$DECODE -> {reflectionDecode} or {reflectionCubemapDecode}
```
in chunks:
```
  reflectionEnvPS
  reflectionEnvHQPS
  reflectionCubePS
  reflectionSpherePS

```

### DIFF
--- a/src/scene/shader-lib/programs/lit-shader.js
+++ b/src/scene/shader-lib/programs/lit-shader.js
@@ -494,7 +494,8 @@ class LitShader {
             this.fDefineSet(options.useIridescence, 'LIT_IRIDESCENCE');
         }
 
-        this.fDefineSet(true, 'LIT_FRESNEL_MODEL', fresnelNames[options.fresnelModel]);
+        this.fDefineSet(this.needsNormal, 'LIT_NEEDS_NORMAL');
+        this.fDefineSet(this.lighting, 'LIT_LIGHTING');
         this.fDefineSet(options.useMetalness, 'LIT_METALNESS');
         this.fDefineSet(options.enableGGXSpecular, 'LIT_GGX_SPECULAR');
         this.fDefineSet(options.useSpecularityFactor, 'LIT_SPECULARITY_FACTOR');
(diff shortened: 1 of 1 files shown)


slimbuck (approved):

src/scene/shader-lib/programs/lit-shader-options.js
```diff
@@ -252,11 +252,11 @@ class LitShaderOptions {
     toneMap = -1;
 
     /**
-     * One of "envAtlasHQ", "envAtlas", "cubeMap", "sphereMap".
+     * One of REFLECTIONSRC_*** constants.
      *
      * @type {string}
      */
-    reflectionSource = null;
+    reflectionSource = REFLECTIONSRC_NONE;
```
willeastcott, line 259: I prefer to avoid introducing new public API constants unless there's a really good reason. Why not just tighten up the type to:

```
* @type {"envAtlasHQ" | "envAtlas" | "cubeMap" | "sphereMap"}
```

willeastcott (approved):

src/scene/shader-lib/programs/lit-shader-options.js
mvaligursky: those constants are already in the public API here: https://api.playcanvas.com/classes/Engine.LitShaderOptions.html#reflectionsource

and for consistency, I'm turning those into enums as all other similar constants, instead of just strings.

Are you suggesting we change this and avoid new constants in preference to strings?

mvaligursky: Definitely worth a discussion. Ideally we keep the codebase consistent, instead of the mix of two styles.

PR #7409, opened 2025-03-05
Preprocessor reports errors in mismatched #ifdef #endif

mvaligursky:
- avoid crashes, report useful error and give up preprocessing

### DIFF
--- a/src/core/preprocessor.js
+++ b/src/core/preprocessor.js
@@ -227,7 +227,7 @@ class Preprocessor {
         let error = false;
 
         let match;
-        while ((match = KEYWORD.exec(source)) !== null) {
+        while ((match = KEYWORD.exec(source)) !== null && !error) {
 
             const keyword = match[1];
             switch (keyword) {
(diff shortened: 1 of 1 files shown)


src/core/preprocessor.js
```diff
@@ -379,6 +379,11 @@ class Preprocessor {
                     const endif = ENDIF.exec(source);
 
                     const blockInfo = stack.pop();
+                    if (!blockInfo) {
+                        console.error(`Shader preprocessing encountered "#${endif[1]}" without proceding #if #ifdef #ifndef while preprocessing ${Preprocessor.sourceName} on line:\n ${source.substring(match.index, match.index + 100)}...`, { source: originalSource });
```
willeastcott, line 383:
```suggestion
                        console.error(`Shader preprocessing encountered "#${endif[1]}" without a preceding #if #ifdef #ifndef while preprocessing ${Preprocessor.sourceName} on line:\n ${source.substring(match.index, match.index + 100)}...`, { source: originalSource });
```

willeastcott (approved):

PR #7411, opened 2025-03-06
WebXR fix frustum culling (right eye)

### LINKED ISSUE #5787: WebXR Frustum Culling - Issue
Current WebXR Frustum Culling in multi-view session constructs Frustrum from the left view, which leads to early culling on the right edge of the right view.

We need to construct combined Frustrum from both view projections.

Maksims:
Fixes #5787

This PR changes the way view projection matrix is calculated when in stereo XR, by constructing average view matrix and offsetting it slightly backwards. This solves an issue where in right view (eye) objects were frustum culled too early (at the edge of right eye's frustum).

There is potentially a better way, by constructing a frustum that is based on left plane of left's frustum and right plane of right's frustum, but I'm not that smart to figure it out 😃 

### DIFF
--- a/src/scene/renderer/renderer.js
+++ b/src/scene/renderer/renderer.js
@@ -3,6 +3,7 @@ import { now } from '../../core/time.js';
 import { Vec2 } from '../../core/math/vec2.js';
 import { Vec3 } from '../../core/math/vec3.js';
 import { Vec4 } from '../../core/math/vec4.js';
+import { Quat } from '../../core/math/quat.js';
 import { Mat3 } from '../../core/math/mat3.js';
 import { Mat4 } from '../../core/math/mat4.js';
 import { BoundingSphere } from '../../core/shape/bounding-sphere.js';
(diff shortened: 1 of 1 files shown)


willeastcott:
I thought this was interesting:

https://computergraphics.stackexchange.com/a/8468

Maksims:
> I thought this was interesting:
> 
> https://computergraphics.stackexchange.com/a/8468

Even the current solution will make this case much better, but with such wide angles, perhaps a different method for culling (not frustrum) should be used then. And that would be a separate PR.

willeastcott:
I think that over the 180 degree threshhold, you need to cull per eye. But below that, I think we simply need to have a `Frustum#add` or `Frustum#merge` function.

Maksims:
> I think that over the 180 degree threshhold, you need to cull per eye. But below that, I think we simply need to have a `Frustum#add` or `Frustum#merge` function.

As mentioned:
> I'm not that smart to figure it out

`Frustum#merge` - would be great here, but bear in mind that this is not that simple, and in many cases, frustums are not mergeable.

For now, this PR improves a current situation, by removing culling issues on right eye on most XR devices I've tested. Future PR's can improve that. But I believe that this issue been there for far too long, and decided to fix it within my mathematical abilities.

PR #7416, opened 2025-03-07
Finalizing the lit-shader refactoring by moving shader code to chunks

mvaligursky:
Final part of the refactoring, where the extracted code has been moved to chunks. When this is done for front end part of StandardMaterial, this will allow users complete control over the shader, including lighting and all.

This is how the fragment shaders get generated now:
- varyings come from vertex shader generation
- front end code comes from standard shader front end generation
- but all the rest comes from chunks, no more string concatenation in JS code

Screenshot 2025-03-07 at 16 24 29


src/scene/shader-lib/programs/lit-shader.js
```diff
@@ -114,37 +126,31 @@ class LitShader {
...
-            default:
-                return '';
+    /**
+     * Helper function to define a value in the fragment shader.
+     *
+     * @param {boolean} condition - The define is added if the condition is true.
+     * @param {string} name - The define name.
+     * @param {string} [value] - The define value.
+     */
+    fDefineSet(condition, name, value = '') {
```
willeastcott, line 144: Did you mean `fsDefineSet`?

willeastcott (approved):

PR #7424, opened 2025-03-11
Pin monaco version to 0.33.0

kpal81xd:
N/A


kungfooman:
Why is this required?

PR #7426, opened 2025-03-12
Shadow catcher shader support + script + example

mvaligursky:
- engine support creating a material with shadowCatcher property. The generated shader accumulates all directional shadows, and outputs those in RGB.
- some existing shader optimization - early out when the light color / intensity is black or the attenuatipon is 0, to skip shadow sampling (only non-direcional lights partially did this)
- added a script to simplify integration of shadow catcher functionality to applications
- example to demonstrate it's use

Screenshot 2025-03-12 at 14 21 11


willeastcott (approved):
LGTM, but I think this needs an approval from @slimbuck since he's done a bunch of work on shadow catcher in the past.

willeastcott:
Oh actually, I do have some feedback about the example. The skybox has the sun but there are two rotating directional shadows. It looks very 'wrong'. Personally, I would have one shadow and have it line up with the sun direction. If you need multiple (and moving) shadows, the consider switching the scene for something else more appropriate. And BTW, the statue isn't really a good choice - a car perhaps, would be more suitable. So yeah, just generally, scene composition is not great right now.

mvaligursky:
Updated example, but keeping statue as it seems to fit now.

slimbuck (approved):

Issue #7415, opened 2025-03-07
OutlineRenderer wrong colors if entity/render is disabled.


Maksims:
OutlineRenderer has a bug: if entity/render is disabled before adding it to the outliner, then the color and other things of outline will be wrong (uses original meshInstance's material state).
To replicate an issue, modify this example: https://playcanvas.vercel.app/#/graphics/outlines-colored

Code below replicates it well.
On line 128, it will disable entity before adding, and then enable during render leading to wrong outline color.
Moving line 128 to after adding it to outliner - works as a workaround.

```javascript
import { rootPath, deviceType } from 'examples/utils';

const canvas = /** @type {HTMLCanvasElement} */ (document.getElementById('application-canvas'));
window.focus();

// set up and load draco module, as the glb we load is draco compressed
pc.WasmModule.setConfig('DracoDecoderModule', {
    glueUrl: `${rootPath}/static/lib/draco/draco.wasm.js`,
    wasmUrl: `${rootPath}/static/lib/draco/draco.wasm.wasm`,
    fallbackUrl: `${rootPath}/static/lib/draco/draco.js`
});

const assets = {
    laboratory: new pc.Asset('statue', 'container', { url: `${rootPath}/static/assets/models/laboratory.glb` }),
    orbit: new pc.Asset('orbit', 'script', { url: `${rootPath}/static/scripts/camera/orbit-camera.js` }),
    helipad: new pc.Asset(
        'helipad-env-atlas',
        'texture',
        { url: `${rootPath}/static/assets/cubemaps/helipad-env-atlas.png` },
        { type: pc.TEXTURETYPE_RGBP, mipmaps: false }
    )
};

const gfxOptions = {
    deviceTypes: [deviceType],
    glslangUrl: `${rootPath}/static/lib/glslang/glslang.js`,
    twgslUrl: `${rootPath}/static/lib/twgsl/twgsl.js`
};

const device = await pc.createGraphicsDevice(canvas, gfxOptions);
const createOptions = new pc.AppOptions();
createOptions.graphicsDevice = device;
createOptions.mouse = new pc.Mouse(document.body);
createOptions.touch = new pc.TouchDevice(document.body);

createOptions.componentSystems = [
    pc.RenderComponentSystem,
    pc.CameraComponentSystem,
    pc.LightComponentSystem,
    pc.ScriptComponentSystem
];
createOptions.resourceHandlers = [
    pc.ScriptHandler,
    pc.TextureHandler,
[...]

PR #7427, opened 2025-03-12
Outline renderer prepares even disabled entities 

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/7415 by handling even disabled entities / components


src/extras/renderers/outline-renderer.js
```diff
@@ -187,6 +187,8 @@ class OutlineRenderer {
     addEntity(entity, color, recursive = true) {
         const meshInstances = this.getMeshInstances(entity, recursive);
 
+        console.log(meshInstances.length);
```
willeastcott, line 190: Delete.

willeastcott (approved):

PR #7431, opened 2025-03-13
Validate shader vertex attributes are available in vertex buffers

mvaligursky:


LeXXik:
Does it need to be in the production build?

mvaligursky:
> Does `validateAttributes` needs to be in the production build?

Gone, thanks

slimbuck (approved):

PR #7434, opened 2025-03-14
Reproject texture functionality uses native WGSL shaders on WebGPU

mvaligursky:
- updated support code slightly to give it the needed functionality
- converted reproject chunks to WGSL - some complexity with this related to luck of polymorphic functions in WGSL


slimbuck (approved):

src/platform/graphics/shader-chunks/frag/shared-wgsl.js
```diff
@@ -0,0 +1,17 @@
+export default /* glsl */`
+
+// convert clip space position into texture coordinates for sampling scene grab textures
```
willeastcott, line 3: "scene grab textures"? What are those?

mvaligursky: color grab and depth grab textures

mvaligursky: this is existing webgl shader chunk, just translated to wgsl

src/scene/shader-lib/chunks-wgsl/common/frag/encode.js
```diff
@@ -0,0 +1,46 @@
+export default /* glsl */`
```
willeastcott, line 1: Is there no wgsl syntax highlighter yet?

willeastcott (approved):

src/scene/shader-lib/chunks-wgsl/common/frag/encode.js
mvaligursky: there is and using it on few chunks, let me update the rest, cheers

PR #7442, opened 2025-03-17
Use less memory loading compressed.ply

slimbuck:
This PR updates the loading code for compressed.ply files that contain spherical harmonics.

The PLY load was reading the raw spherical harmonic data and then `GsplatCompressedData` was packing the data for GPU.

This resulted in almost double the memory requirement for spherical harmonic data at load time.

This PR changes PLY reader to stream in and pack the data in one step, thereby almost halving the memory allocation at load time.

From a quick chrome devtools memory snapshot, loading the 375MB compress.ply bicycle scene has gone from allocating 854MB to 515MB.

## Notes
- we can get further savings by storing fewer copies of unneccessary centers buffers, this will be done separately because the change will require structural changes in the engine


willeastcott (approved):
Brilliant. I _can't wait_ to get this deployed!

PR #7445, opened 2025-03-18
Convert internal morph target blending shader to WGSL on WebGPU

mvaligursky:
## Morph changes

- morph target blending shader was generated using GLSL and string concatenation, then transpiled to WGSL.
- now the shader has been converted to chunks, and the final shader (for the required number of weights) gets generated using defines.
- these newly created chunks were also converted to WGSL to avoid GLSL transpilation on WebGPU

## Other changes

Support for uniform arrays had been added to WGSL parsing / processing to support morphing. There are some WGSL limitation / requirements regarding uniform arrays: the elements in the array need to be 16byte aligned. This is by default not satisfied for f32 and Vec2f types and their integer variants, and so those types are internally wrapped in structs that are aligned.

Example:
```
uniform a: array<vec2f, 10>
var data = a[3];
```

gets internally transformed to
```
struct WrappedVec2F { @size(16) element: vec2f }
uniform a: array<WrappedVec2F, 10>
var data = a[3].element;
```

by the way, this is what our transpilation pipeline does as well / it's the recommended way to transfer GLSL arrays to WGSL, the only thing to be away of is that access is done using `.element`. The transpilation would handle this bit as well.

### DIFF
--- a/src/platform/graphics/webgpu/webgpu-shader-processor-wgsl.js
+++ b/src/platform/graphics/webgpu/webgpu-shader-processor-wgsl.js
@@ -553,7 +553,7 @@ class WebgpuShaderProcessorWGSL {
             // array uniforms
             if (uniform.count > 0) {
 
-                // if the type is one of the onces that are not by default 16byte aligned, which is
+                // if the type is one of the ones that are not by default 16byte aligned, which is
                 // a requirement for uniform buffers, we need to wrap them in a struct
                 // for example: array<f32, 5> becomes  array<WrappedF32, 5>
                 if (wrappedArrayTypes.hasOwnProperty(typeString)) {
--- a/src/platform/graphics/webgpu/webgpu-shader-processor-wgsl.js
+++ b/src/platform/graphics/webgpu/webgpu-shader-processor-wgsl.js
@@ -73,6 +73,15 @@ const textureFormat2SampleType = {
     'u32': SAMPLETYPE_UINT
 };
 
+const wrappedArrayTypes = {
+    'f32': 'WrappedF32',
+    'i32': 'WrappedI32',
+    'u32': 'WrappedU32',
+    'vec2f': 'WrappedVec2F',
+    'vec2i': 'WrappedVec2I',
+    'vec2u': 'WrappedVec2U'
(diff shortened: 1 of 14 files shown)


src/platform/graphics/webgpu/webgpu-shader-processor-wgsl.js
```diff
@@ -517,9 +547,25 @@ class WebgpuShaderProcessorWGSL {
...
         ubFormat.uniforms.forEach((uniform) => {
-            const typeString = uniformTypeToNameWGSL[uniform.type][0];
+            let typeString = uniformTypeToNameWGSL[uniform.type][0];
             Debug.assert(typeString !== undefined, `Uniform type ${uniform.type} is not handled.`);
-            code += `    ${uniform.name}: ${typeString}${uniform.count ? `[${uniform.count}]` : ''},\n`;
+
+            // array uniforms
+            if (uniform.count > 0) {
+
+                // if the type is one of the onces that are not by default 16byte aligned, which is
```
willeastcott, line 556:
```suggestion
                // if the type is one of the ones that are not by default 16byte aligned, which is
```

src/scene/shader-lib/chunks-wgsl/internal/morph/vert/morph.js
```diff
@@ -0,0 +1,14 @@
...
+export default /* wgsl */`
+    attribute vertex_position: vec2f;
+    varying uv0: vec2f;
+
+    @vertex
+    fn vertexMain(input: VertexInput) -> VertexOutput {
+      var output: VertexOutput;
+      output.position = vec4f(input.vertex_position, 0.5, 1.0);
+      output.uv0 = input.vertex_position * 0.5 + vec2f(0.5, 0.5);
+      return output;
```
willeastcott, line 12:
```suggestion
        var output: VertexOutput;
        output.position = vec4f(input.vertex_position, 0.5, 1.0);
        output.uv0 = input.vertex_position * 0.5 + vec2f(0.5, 0.5);
        return output;
```

willeastcott (approved):

PR #7447, opened 2025-03-18
GS sorter speedup

slimbuck:
This PR updates the sorter implementation resulting in a substantial speed improvement by:
- removing center NaNs instead of testing for them during sort
- refrain from subtracting camera position when calculating sort distances since it's uneccessary

The 6M bicycle scene sort on Mac M1 goes from:

Screenshot 2025-03-18 at 12 02 35

To:

Screenshot 2025-03-18 at 12 01 43


willeastcott (approved):

vincentwoo:
nice

PR #7448, opened 2025-03-18
Improved validation of vertex attributes required by the shader

mvaligursky:
Validation introduced in https://github.com/playcanvas/engine/pull/7431 had false positives in case the user requested additional attributes on the Material, but those were not used in all variants (for example an instancing-capable shader used on non-instanced mesh)

This PR fixes it. We validate attributes from VertexBuffers not against all attributes shader has available, but against the list of those extracted from the final shader. For WebGL those are extracted using `getActiveAttrib`, for WebGPU from one of our shader processors (GLSL or WGSL)

### DIFF
--- a/src/platform/graphics/shader.js
+++ b/src/platform/graphics/shader.js
@@ -42,7 +42,7 @@ class Shader {
     meshBindGroupFormat;
 
     /**
-     * The attributes that this shader code uses, the location is the key, the value is the name.
+     * The attributes that this shader code uses. The location is the key, the value is the name.
      * These attributes are queried / extracted from the final shader.
      *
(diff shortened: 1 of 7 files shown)


src/platform/graphics/shader.js
```diff
@@ -41,6 +41,15 @@ class Shader {
      */
     meshBindGroupFormat;
 
+    /**
+     * The attributes that this shader code uses, the location is the key, the value is the name.
```
willeastcott, line 45:
```suggestion
     * The attributes that this shader code uses. The location is the key, the value is the name.
```

willeastcott (approved):

PR #7459, opened 2025-03-20
Clustered lighting optimisations - removed WebGL1 compatibility solutions

mvaligursky:
## Main goals done
- lights were stored in an 8bit and a float texture. 8bit texture is now removed and so one additional texture slot is available
- WebGL1 didn't have bit packing / uint arithmetic, and so many 1bit data were stored in 8bits of 8bit texture. Those are now bit packed, requiring less bandwidth used by the texture sampling
- using hardware half->float decompression to store data, instead of manual decompression
- storing data pre-cached in half float formats in a light class instead of compressing each frame

I expect slight CPU performance win, but this was not tested.

## GPU improvements (4K on M1 Mac):

11% improvement in this example
before: 1.8ms
now: 1.6ms
Screenshot 2025-03-20 at 16 58 57

16% improvement in this example
before: 4.2ms
now: 3.5ms
Screenshot 2025-03-20 at 16 59 52


examples/src/examples/graphics/clustered-spot-shadows.controls.mjs
```diff
@@ -98,6 +98,17 @@ export const controls = ({ observer, ReactPCUI, React, jsx, fragment }) => {
                     value: observer.get('settings.shadowIntensity')
                 })
             ),
+            jsx(
+                LabelGroup,
+                { text: 'Cookies Intensity' },
```
willeastcott, line 103:
```suggestion
                { text: 'Cookie Intensity' },
```

src/scene/light.js
```diff
@@ -1077,6 +1133,50 @@ class Light {
 
         this.key = key;
     }
+
+    /**
+     * Updates 32bit flags used by the clsutered lighting. This only stores constant data.
```
willeastcott, line 1138:
```suggestion
     * Updates 32bit flags used by the clustered lighting. This only stores constant data.
```

willeastcott (approved):

PR #7479, opened 2025-03-25
Instantiate GLB's with correct AppBase instance

### LINKED ISSUE #7471: Error when sharing Assets between Application instances
Consider a single page displaying a gallery models/assets. There are  multiple `AppBase` instances and a single set of `Assets` shared across multiple app instances. 

This [currently throws a rendering error](https://codesandbox.io/p/sandbox/7prvzh?file=%2Fsrc%2Fmain.js%3A10%2C1) presumable because resource cannot be shared. However, it's not clear for the user what is wrong, or how to fix it. If assets can't be shared we should provide better feedback and warnings

This is a real world user issue See https://github.com/playcanvas/react/issues/80

marklundin:
Ensures that the correct `AppBase` instance is used when creating GLB's via `instantiateRenderEntity()`.  This uses the `Asset` reference to the app rather than `getApplication()` which can be incorrect.

Includes a tests. Fixes #7471 

### DIFF
--- a/src/framework/parsers/glb-container-resource.js
+++ b/src/framework/parsers/glb-container-resource.js
@@ -123,7 +123,7 @@ class GlbContainerResource {
         // helper function to recursively clone a hierarchy of GraphNodes to Entities
         const cloneHierarchy = (root, node, glb) => {
 
-            const entity = new Entity();
+            const entity = new Entity('Untitled', this._assets._loader._app);
             node._cloneInternal(entity);
 
             // first entity becomes the root
--- a/src/framework/parsers/glb-container-resource.js
+++ b/src/framework/parsers/glb-container-resource.js
@@ -71,7 +71,7 @@ class GlbContainerResource {
     }
 
     instantiateModelEntity(options) {
-        const entity = new Entity();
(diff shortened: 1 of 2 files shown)


src/framework/parsers/glb-container-resource.js
```diff
@@ -123,7 +123,7 @@ class GlbContainerResource {
         // helper function to recursively clone a hierarchy of GraphNodes to Entities
         const cloneHierarchy = (root, node, glb) => {
 
-            const entity = new Entity();
+            const entity = new Entity('Untitled', this._assets._loader._app);
```
mvaligursky, line 126: just pass `undefined` instead of 'Untitled', let the underlaying code handle the default

mvaligursky: also update instantiateModelEntity the same way pls

marklundin: done

mvaligursky (approved):
Look good. It's only a partial fix, there are still many cases this can load, as there are many entities created by us or the clients without passing the app to the constructor, but I hit this specific case before as well, so good to have this fixed.

PR #7487, opened 2025-03-27
[BREAKING] Refactor standard material chunks dealing with texture channels

mvaligursky:
If you are not overriding any of the shader chunks listed bellow, you can skip this, this has no effect on your project.

More commonly modified chunks:
```
- diffusePS
- emissivePS
- opacityPS
```
Rarely commonly modified chunks:
```
- parallaxPS
- normalDetailMapPS
- normalMapPS
- diffuseDetailMapPS
- transmissionPS
- thicknessPS
- iridescencePS
- iridescenceThicknessPS
- sheenPS
- sheenGlossPS
- metalnessPS
- iorPS
- specularityFactorPS
- specularPS
- glossPS
- aoDetailMapPS
- aoPS
- clearCoatPS
- clearCoatGlossPS
- clearCoatNormalPS
- lightmapDirPS
- lightmapSinglePS
```
All listed front end shader chunks, dealing with StandardMaterial texture sampling / vertex color access were modified. If you are overriding any of these, you need to update them to a new format.

To make the transition easier, engine in debug mode prints exactly what needs to be replaced in each chunk:
Screenshot 2025-03-28 at 11 40 52

The user can simply apply those changes (search and replace), the source:
Screenshot 2025-03-28 at 11 41 59

the modified chunk:
Screenshot 2025-03-28 at 11 42 07

Benefits:
- we now have global defines that drive those chunks, instead of per chunk defines
- we do NOT need to do any string replacement in javascript, but simply apply standard preprocessor to the chunk
- this allows us to simply convert chunks to WGSL and retain their functionality, and then further create an uber shader for the standard material

### DIFF
--- a/src/scene/materials/standard-material-options-builder.js
+++ b/src/scene/materials/standard-material-options-builder.js
@@ -219,18 +219,18 @@ class StandardMaterialOptionsBuilder {
 
         const equalish = (a, b) => Math.abs(a - b) < 1e-4;
 
-        options.specularTint = specularTint ? 2 : 0;
-        options.specularityFactorTint = specularityFactorTint ? 1 : 0;
-        options.metalnessTint = (stdMat.useMetalness && stdMat.metalness < 1) ? 1 : 0;
-        options.glossTint = 1;
+        options.specularTint = specularTint;
+        options.specularityFactorTint = specularityFactorTint;
+        options.metalnessTint = (stdMat.useMetalness && stdMat.metalness < 1);
+        options.glossTint = true;
         options.diffuseEncoding = stdMat.diffuseMap?.encoding;
         options.diffuseDetailEncoding = stdMat.diffuseDetailMap?.encoding;
         options.emissiveEncoding = stdMat.emissiveMap?.encoding;
         options.lightMapEncoding = stdMat.lightMap?.encoding;
         options.packedNormal = isPackedNormalMap;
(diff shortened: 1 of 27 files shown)


slimbuck:
Wouldn't `STD_AO_TEXTURE` imply defined so you can drop the `_ENABLED`?

mvaligursky:
> Wouldn't `STD_AO_TEXTURE` imply defined so you can drop the `_ENABLED`?

STD_AO_TEXTURE is injected define due to curly brackets: {STD_AO_TEXTURE}
and so it is not left as a define in the code
and so #ifdef cannot test against it
we need a define without {}

slimbuck:
> > Wouldn't `STD_AO_TEXTURE` imply defined so you can drop the `_ENABLED`?
> 
> STD_AO_TEXTURE is injected define due to curly brackets: {STD_AO_TEXTURE} and so it is not left as a define in the code and so #ifdef cannot test against it we need a define without {}

Can't the loop variable be something longer, so all the defines users must deal with are then shorter?

mvaligursky:
There is no loop here, no string expansion.

I can shorten the names, but not much else.
so instead of `STD_OPACITY_TEXTURE_UV` I could do `STD_OPACITY_UV`,
or `STD_CLEARCOAT_VERTEX_CHANNEL` could be `STD_CLEARCOAT_CH`

but maybe this makes it less descriptive. Maybe that's ok.

PR #7493, opened 2025-03-29
[FIX] [WebGL2] Use correct internal format for WebGL 565/5551/RGBA4 formats

liamdon:
I ran into some issues creating textures with these formats and found I had to change the internal format. WebGL2 accepts the sized internal formats while WebGL1 doesn't.

I'm not sure how we're handling WebGL1/2 differences now with the ongoing WebGL1 branch - is it OK to open PRs that assume WebGL2?


mvaligursky:
Thanks. Yes the main branch is WebGL2 only, no WebGL1 support, so ok to do it here.
Main_v1 branch is WebGL1&2, but that one is for bug fixes only.

mvaligursky (approved):
looks great, thanks, approving

PR #7508, opened 2025-04-01
Refactor the way diffuse, opacity and parallax code is generated for Standard frontend

mvaligursky:
- diffuse, opacity and parallax chunks are now handled using uber shader

**Modified shader chunks:**
```
diffuseDetailMapPS - removed, and it's content added to diffusePS
diffusePS - updated to handle the detail map as well
```


src/scene/materials/standard-material-options-builder.js
```diff
@@ -235,7 +235,7 @@ class StandardMaterialOptionsBuilder {
         options.sheenEncoding = stdMat.sheenEncoding || 'linear';
         options.aoMapUv = stdMat.aoUvSet; // backwards compatibility
         options.aoDetail = !!stdMat.aoMap;
-        options.diffuseDetail = !!stdMat.diffuseMap;
+        options.diffuseDetail = !!stdMat.diffuseDetailMap;
```
willeastcott, line 238: This looks incongruous with the surrounding lines. I'd expect this to be `!!stdMap.diffuseMap` based on the other lines.

slimbuck (approved):

willeastcott (approved):

src/scene/materials/standard-material-options-builder.js
mvaligursky: I will refactor (fix) the other lines as I get to those chunks.

PR #7518, opened 2025-04-03
Ply fetch

slimbuck:
This PR allows the application to provide fetch request when loading PLY asset.

Initiating the fetch request as soon as possible, even before the engine is initialized can save 100's of milliseconds at startup.

This new field is left undocumented, so for now this feature is considered internal.

## Notes
- actually it would be great if the application had access to the fetch request when initiated by the asset system

### DIFF
--- a/src/framework/parsers/ply.js
+++ b/src/framework/parsers/ply.js
@@ -591,6 +591,7 @@ class PlyParser {
      */
     async load(url, callback, asset) {
         try {
+            // either use the fetch request passed in by the application or initiate it ourselves
             const response = await (asset.file?.contents ?? fetch(url.load));
             if (!response || !response.body) {
                 callback('Error loading resource', null);
--- a/src/framework/parsers/ply.js
+++ b/src/framework/parsers/ply.js
@@ -591,7 +591,7 @@ class PlyParser {
      */
     async load(url, callback, asset) {
         try {
-            const response = await fetch(url.load);
+            const response = await (asset.file?.contents ?? fetch(url.load));
             if (!response || !response.body) {
                 callback('Error loading resource', null);
             } else {


src/framework/parsers/ply.js
```diff
@@ -591,7 +591,7 @@ class PlyParser {
      */
     async load(url, callback, asset) {
         try {
-            const response = await fetch(url.load);
+            const response = await (asset.file?.contents ?? fetch(url.load));
```
mvaligursky, line 594: at least add a comment, if not the docs

mvaligursky: I'm not sure it's ideal to add a private way that an external projects depend on though.

marklundin, line 595: Still not sure if this is necessary when you can `<link rel="preload">`  which doesn't need the lib loaded

slimbuck: Ah, two reasons:

According to the gospel of chatgpt:
Note: The value for the as attribute must match the type of resource. While as="fetch" might be used in some experimental cases, it isn’t widely supported or standardized.

But more importantly, the URL is dynamic, not rendered statically into html.

slimbuck: I can document it rn if you prefer?

slimbuck: (comment added)

slimbuck: Some assets already support `file.contents` btw, this change isn't a huge departure.

See https://github.com/playcanvas/engine/blob/main/src/framework/asset/asset.js#L250.

mvaligursky: It'd be great to document if we think this is an API we want to have. We're putting plenty of effort to remove undocumented things that project depend on, so lets not add another one.

marklundin: Definitely should be ok with `as="fetch"`, I've used it before, and [looks well supported](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload#:~:text=11.3-,as%3Dfetch,-50). 

Doing `document.createElement('link')` to inject in the head would be sufficient too.

slimbuck: I have tested this locally and the approach seems to work very well.

But I can't test and profile on DEV and PROD with caching et al and on iOS devices until it's available.

So I would prefer to first make the small internal change and test it broadly, and then if we're sure it's what we want, document and formalize it.

There is the risk that the first part will happen, but not the second...

slimbuck: Lemme try!

slimbuck: Link rel seems to work on chrome, but not safari:

Link:
Screenshot 2025-04-03 at 11 56 30

Fetch:
Screenshot 2025-04-03 at 11 57 03

slimbuck: (This is for scene.compressed.ply)

slimbuck: (And by work I mean do anything. The link doesn't fail, but also doesn't appear to help either).

slimbuck:
Issue created. Feel free to approve https://github.com/playcanvas/engine/issues/7525

mvaligursky (approved):

src/framework/parsers/ply.js
slimbuck: Seems like Safari could be made to work https://stackoverflow.com/a/63814972

willeastcott: Wow, I never knew you could do that - that's amazing! ❤️

PR #7535, opened 2025-04-04
Convert JS based particle shader generator to define based

mvaligursky:
- last particle generator has been moved to use defines based ubershader system
- main vs and ps code was moved to chunks
- allows easy porting to WGSL


willeastcott (approved):
Wondering if we should make a distinction between 'chunks' and 'programs' (top-level with a `main`). That'd make entry points more obvious when browsing our shader code.

mvaligursky:
> Wondering if we should make a distinction between 'chunks' and 'programs' (top-level with a `main`). That'd make entry points more obvious when browsing our shader code.

Yep I though about this few times, an dit would definitely make sense from the user point of view. Refactor is definitely on a list, but perhaps even further than this.

PR #7540, opened 2025-04-07
[Fix] Fix CameraComponent.fog behavior

mvaligursky:
- camera can accept fog parameters during the component construction
- fix so that the rendering actually uses those per camera fog settings
- adjusted the example to use it

### DIFF
--- a/examples/src/examples/test/global-shader-properties.example.mjs
+++ b/examples/src/examples/test/global-shader-properties.example.mjs
@@ -71,12 +71,6 @@ assetListLoader.load(() => {
     app.scene.envAtlas = assets.helipad.resource;
     app.scene.skyboxRotation = new pc.Quat().setFromEulerAngles(0, -70, 0);
 
-    // for params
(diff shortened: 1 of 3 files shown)


slimbuck (approved):

willeastcott:
Does this PR handle fog correctly when cloning the component? See:

https://github.com/playcanvas/engine/blob/efe5ca0de4b6dabd3d048ce5db0a5be393fc3f0c/src/framework/components/camera/system.js#L113-L147

willeastcott:
What does this PR mean for the Editor? We currently have fog on the Scene Render Settings for Engine v2 projects. How do we migrate to this?

willeastcott:
Minor thing while I'm here and looking at this stuff. I'm not keen on the naming for `FogParams` - why not just `Fog` (which is more succinct and aligns nicely with `Sky` for example).

willeastcott:
Is there another example in the engine of a component property being an object rather than a 'simple' type (i.e. number, string, boolean). Since this _feels_ different, are we concerned that this might break anything in our ECS architecture?

PR #7543, opened 2025-04-09
Simplified the ShaderMaterial API to better support GLSL and WGSL shaders

mvaligursky:
This is a non-breaking change. A warning to update to new API is printed, but old format is supported.

Shaders used to be specified this way before, where only a single version (either GLSL or WGSL) could be provided.
```
            vertexCode: wgsl ? vertexShaderWGSL : vertexShaderGLSL,
            fragmentCode: wgsl ? fragmentShaderWGSL : fragmentShaderGLSL,
            shaderLanguage: wgsl ? SHADERLANGUAGE_WGSL : SHADERLANGUAGE_GLSL,
```

Now, both GLSL and WGSL variants can be provided at the same time, and the engine uses the one appropriate for the device.
```
            vertexGLSL: vertexShaderGLSL,
            fragmentGLSL: fragmentShaderGLSL,
            vertexWGSL: vertexShaderWGSL,
            fragmentWGSL: fragmentShaderWGSL,
```


willeastcott:
I'd say drop `Code` from the names.

willeastcott (approved):

PR #7553, opened 2025-04-10
Improve physics API docs

willeastcott:
Improve the API docs covering physics.


src/core/wasm-module.js
```diff
@@ -124,7 +124,26 @@ class Impl {
...
+ *
+ * This class is generally only needed if you are developing against the Engine directly. Editor
+ * projects automatically load WebAssembly modules included in the project's assets.
+ *
+ * Do not use this class to load the Basis WebAssembly module. Instead, please refer to
+ * {@link basisInitialize}.
+ *
+ * @example
+ * // Load the Ammo.js physics engine
+ * pc.WasmModule.setConfig('Ammo', {
```
LeXXik, line 139: If this is an engine-only example, do we still want to add pc proxy?

src/framework/components/rigid-body/system.js
```diff
@@ -69,8 +75,20 @@ class RaycastResult {
...
+ * contact normal, and the collision impulse force.
+ *
+ * Instances of this class are created by the physics engine when collision events occur and are
+ * passed to event handlers only through the global `contact` event on the
+ * {@link RigidBodyComponentSystem}. Individual rigid body components receive instances of
+ * {@link ContactResult} instead.
  *
+ * @example
+ * app.systems.rigidbody.on('contact', (result) => {
+ *     console.log(`Contact between ${result.a.name} and ${result.b.name}`);
```
LeXXik, line 90: There are no a and b properties in a contact result.

src/framework/components/rigid-body/system.js
```diff
@@ -163,7 +181,33 @@ class SingleContactResult {
...
- * Object holding the result of a contact between two Entities.
+ * Represents a single point of contact between two colliding rigid bodies in the physics
+ * simulation. Each contact point stores detailed spatial information about the collision,
+ * including both local and world space coordinates of the exact contact points on both entities,
+ * the contact normal  direction, and the collision impulse force.
+ *
+ * Contact points are generated by the physics engine during collision detection and are typically
+ * accessed through a {@link ContactResult} object, which can contain multiple contact points for a
+ * single collision between two entities. Multiple contact points commonly occur when objects
+ * collide along edges or faces rather than at a single point.
```
LeXXik, line 192: Can't recall Ammo in specific, but generally a contact generated per each vertex, intersecting an object. You can ignore this comment, if Ammo averages a point.

src/framework/components/rigid-body/system.js
```diff
@@ -310,6 +377,9 @@ class RigidBodyComponentSystem extends ComponentSystem {
      * [0, -9.81, 0] which is an approximation of the gravitational force on Earth.
      *
      * @type {Vec3}
+     * @example
+     * // Set the gravity in the physics world to simulate a planet with low gravity
```
LeXXik, line 381: No need //

LeXXik (approved):

src/core/wasm-module.js
willeastcott: All API reference example code prepends pc. before class names. I'm open to changing that, but only if it's globally consistent.

src/framework/components/rigid-body/system.js
willeastcott: It's a code comment, so it's needed (in case someone copies and pastes the code block).

src/framework/components/rigid-body/system.js
willeastcott: I'm not quite sure what you're proposing here. Can you elaborate?

src/framework/components/rigid-body/system.js
willeastcott: This is a SingleContactResult. 🙂

src/framework/components/rigid-body/system.js
LeXXik: Your comment is fine. Thought we could clarify why multiple contacts are generated per single shape. Not necessary.

willeastcott: It already says "Multiple contact points commonly occur when objects collide along edges or faces rather than at a single point." Doesn't that clarify things?

src/framework/components/rigid-body/system.js
```diff
@@ -16,7 +16,13 @@ import { RigidBodyComponentData } from './data.js';
 let ammoRayStart, ammoRayEnd;
 
 /**
- * Object holding the result of a successful raycast hit.
+ * Contains the result of a successful raycast intersection with a rigid body. When a ray
+ * intersects with a rigid body in the physics simulation, this class stores the complete
+ * information about that intersection including the entity, the exact point of impact, the normal
```
mvaligursky, line 21: suggestion: impact -> intersection

mvaligursky (approved):

src/framework/components/rigid-body/system.js
LeXXik: It is also fine to call the distance of the ray where the intersection occurred as time of impact, so both are ok here.

PR #7557, opened 2025-04-11
Remove redundant @type tags from constants

willeastcott:
Remove generic `string` and `number` types from engine constants because `tsc` can discern the constants are literal values and inherently knows the types. This results in 549 lines being removed from the engine source and types becoming stricter.

For example, old example from `playcanvas.d.ts`:

```js
/**
 * Ignores the integer part of texture coordinates, using only the fractional part.
 *
 * @type {number}
 * @category Graphics
 */
export const ADDRESS_REPEAT: number;
```

New example from `playcanvas.d.ts`:

```js
/**
 * Ignores the integer part of texture coordinates, using only the fractional part.
 *
 * @category Graphics
 */
export const ADDRESS_REPEAT: 0;
```

### DIFF
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -1,7 +1,6 @@
 /**
  * Logs a frame number.
  *
- * @type {string}
  * @category Debug
  */
 export const TRACEID_RENDER_FRAME = 'RenderFrame';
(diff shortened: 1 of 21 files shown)


willeastcott:
Merging without approval. Naughty, I know. But this PR is very low risk.

PR #7581, opened 2025-04-17
Envmap and partial lighting works using WGSL

mvaligursky:
more chunks converted to WGSL, clustered lighting partially working (no shadows, cookies ..)

### DIFF
--- a/src/scene/shader-lib/chunks-wgsl/common/frag/envAtlas.js
+++ b/src/scene/shader-lib/chunks-wgsl/common/frag/envAtlas.js
@@ -1,4 +1,8 @@
 export default /* wgsl */`
+
+#ifndef _ENVATLAS_INCLUDED_
+#define _ENVATLAS_INCLUDED_
+
 // the envAtlas is fixed at 512 pixels. every equirect is generated with 1 pixel boundary.
 const atlasSize : f32 = 512.0;
 const seamSize : f32 = 1.0 / atlasSize;
@@ -20,4 +24,6 @@ fn mapShinyUv(uv : vec2f, level : f32) -> vec2f {
     let t : f32 = 1.0 / exp2(level);
     return mapUv(uv, vec4f(1.0 - t, 1.0 - t, t, t * 0.5));
 }
+
+#endif
 `;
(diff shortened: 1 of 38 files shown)


src/scene/shader-lib/chunks-wgsl/lit/frag/lighting/lightFunctionLight.js
```diff
@@ -0,0 +1,228 @@
...
+export default /* glsl */`
+#if defined(LIGHT{i})
+
+void evaluateLight{i}(
+    #if defined(LIT_IRIDESCENCE)
+        vec3 iridescenceFresnel
+    #endif
+) {
+
+    NONE-CLUSTERED LIGHTS ARE NOT SUPPORTED YET
```
willeastcott, line 11:
```suggestion
    NON-CLUSTERED LIGHTS ARE NOT SUPPORTED YET
```

src/scene/shader-lib/chunks-wgsl/common/frag/envAtlas.js
```diff
@@ -1,4 +1,8 @@
 export default /* wgsl */`
+
+#ifndef _ENVATLAS_INCLUDED_
```
slimbuck, line 3: any reason for _'s? don't think other defines around the codebase have these.

mvaligursky: I guess just a habit from c++ days, to separate those double-include guards from normal defines (that people can depend on) to avoid possible collision. We already have _DECODE_INCLUDED_ in code as well.

src/scene/shader-lib/chunks-wgsl/lit/frag/combine.js
```diff
@@ -20,7 +20,7 @@ fn combineColor(albedo: vec3f, sheenSpecularity: vec3f, clearcoatSpecularity: f3
     #endif // LIT_SHEEN
     #ifdef LIT_CLEARCOAT
         let clearCoatScaling: f32 = 1.0 - ccFresnel * clearcoatSpecularity;
-        ret = ret * clearCoatScaling + (ccSpecularLight + ccReflection.rgb) * clearcoatSpecularity;
+        ret = ret * clearCoatScaling + (ccSpecularLight + ccReflection) * clearcoatSpecularity;
```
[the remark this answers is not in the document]
mvaligursky: it's intended, it's vec3 so no need for .rgb

slimbuck (approved):

PR #7594, opened 2025-04-24
WGSL support for picker pass

mvaligursky:
picking uses WGSL shader now

Screenshot 2025-04-24 at 09 34 02


src/scene/shader-lib/chunks-wgsl/common/frag/pick.js
```diff
@@ -0,0 +1,9 @@
+export default /* glsl */`
```
willeastcott, line 1:
```suggestion
export default /* wgsl */`
```

willeastcott (approved):

PR #7599, opened 2025-04-24
WGSL implementation of VSM shadows

mvaligursky:
- both blurring as well as sampling VSM shadows were converted to WGSL

Screenshot 2025-04-24 at 14 39 28

### DIFF
--- a/src/scene/shader-lib/chunks-wgsl/lit/frag/blurVSM.js
+++ b/src/scene/shader-lib/chunks-wgsl/lit/frag/blurVSM.js
@@ -1,0 +1,35 @@
+export default /* glsl */`
+varying vUv0: vec2f;
+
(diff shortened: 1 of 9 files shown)


src/scene/shader-lib/chunks-wgsl/lit/frag/blurVSM.js
```diff
@@ -0,0 +1,35 @@
+export default /* glsl */`
```
willeastcott, line 1:
```suggestion
export default /* wgsl */`
```

willeastcott: Dude, quite making me correct these! 🤣

willeastcott (approved):

PR #7605, opened 2025-04-25
Fix bloom artifacts appearing due to bad specular data

### LINKED ISSUE #7492: Black square artifacts when using CameraFrame Bloom or DoF Post-Processing
## Description

When using either Bloom or Depth of Field (DoF), a black square flashes randomly on the screen. At certain places in my scene I can reliably cause these artifacts to show.

The strange thing is the 3D modeler I work with does not see these artifacts at all.

Another unusual thing is this happens with my much more powerful desktop, and does not happen on my laptop.

Related Forum post from another user: [Artifacts with cameraFrame bloom](https://forum.playcanvas.com/t/artifacts-with-cameraframe-bloom/39530)

## Screenshots

In-scene example

Image

Screenshots from Spector.js. 

| Previous frame | New frame with the artifact
|-|-
| Image | Image

On Discord, I've DM'd Martin the Spector.js save file shown in the screenshots above.

## Reproducibility

Testing with a custom local-compiled engine I'm seeing this on the [latest commit](https://github.com/playcanvas/engine/commit/0d40977498b8278cd915daf6d3731d31c4cb9ade) as of March 28, 2025.

I also went back as far as engine `v2.3.1` (Dec 4, 2024) and was seeing the same artifacts. Testing earlier versions wasn't possible without rewriting the camera-frame script.

### Machine A (Desktop)

Bug reproducible? **`YES` - frequent artifacting, easy to reproduce**

Graphics: **nvidia rtx 3090** (discrete)
Browsers: Firefox 136 (stable, snap), Chrome 134 (stable, deb)
HDR monitor: No
OS: Ubuntu 24.10 (Wayland)

<details><summary>WebGL Report</summary>

Image

</details> 

### Machine B (Laptop)

Can reproduce? **`NO` - no artifacts at all**

Graphics: Intel UHD (integrated) + **nvidia Quadro T1000** (discrete)
Browsers: Firefox 136 (stable, snap), Chrome 134 (stable, deb)
HDR monitor: No
OS: Ubuntu 24.10 (Wayland)

<details><summary>WebGL Report</summary>

Image

</details>

Christopher-Hayes:
## Description

Fixes black square artifacts appearing on screen when the Bloom or Dof effects were used with the CameraFrame script. This bug was only reproduced on Ubuntu with an nvidia GPU.

image

### Solution

This patch adds a clamp to the fresnelSchlick.js functions, `getFresnel()` and `getFresnelCC()`, to avoid bad specular values.

### Explanation

I still don't think I'm 100% sure why it was happening, but from what I can tell:

1. When the angle of light and the camera were exactly parallel it would cause this dot product calculation to be -1.
https://github.com/playcanvas/engine/blob/49ffb62ad79069fa77f987ea257cdcfcc970a585/src/scene/shader-lib/chunks/lit/frag/pass-forward/litForwardBackend.js#L106-L114
2. With the current implementation, a `cosTheta` of `-1` should not break anything. `max()` would set it to `0.0` and the function should work okay.
https://github.com/playcanvas/engine/blob/49ffb62ad79069fa77f987ea257cdcfcc970a585/src/scene/shader-lib/chunks/lit/frag/fresnelSchlick.js#L1-L21
3. But, for whatever reason on certain devices it was causing `getFresnel()` to return a bad specular value.
4. This seems limited to only a pixel or two, but bloom upscaling makes these artifacts a lot more noticeable.

### Before vs After

(Note that the flashing is more visible in-person, the low-FPS recording hides it somewhat)

https://github.com/user-attachments/assets/9b0767c8-105c-41ce-949a-3b7d69a59976

### Testing

Project for testing: https://playcanvas.com/editor/scene/2210459

Fixes #7492

### DIFF
--- a/src/scene/shader-lib/chunks/lit/frag/fresnelSchlick.js
+++ b/src/scene/shader-lib/chunks/lit/frag/fresnelSchlick.js
@@ -9,7 +9,8 @@ vec3 getFresnel(
         float iridescenceIntensity
 #endif
     ) {
-    float fresnel = pow(1.0 - max(cosTheta, 0.0), 5.0);
+    float safeCosTheta = clamp(cosTheta, 0.0, 1.0);
+    float fresnel = pow(1.0 - safeCosTheta, 5.0);
     float glossSq = gloss * gloss;
     vec3 ret = specularity + (max(vec3(glossSq), specularity) - specularity) * fresnel;
 #if defined(LIT_IRIDESCENCE)
@@ -20,7 +21,8 @@ vec3 getFresnel(
 }
 
 float getFresnelCC(float cosTheta) {
-    float fresnel = pow(1.0 - max(cosTheta, 0.0), 5.0);
+    float safeCosTheta = clamp(cosTheta, 0.0, 1.0);
+    float fresnel = pow(1.0 - safeCosTheta, 5.0);
     return 0.04 + (1.0 - 0.04) * fresnel;
 }
 `;


willeastcott (approved):

mvaligursky (approved):
Looks great, thanks!

PR #7630, opened 2025-04-29
Convert top-level constants to enums

### LINKED ISSUE #7589: RFC: Migrate API constants to enum-like API
For example, let's take:

```js
/**
 * Texture data is not stored a specific projection format.
 *
 * @category Graphics
 */
export const TEXTUREPROJECTION_NONE = 'none';

/**
 * Texture data is stored in cubemap projection format.
 *
 * @category Graphics
 */
export const TEXTUREPROJECTION_CUBE = 'cube';

/**
 * Texture data is stored in equirectangular projection format.
 *
 * @category Graphics
 */
export const TEXTUREPROJECTION_EQUIRECT = 'equirect';

/**
 * Texture data is stored in octahedral projection format.
 *
 * @category Graphics
 */
export const TEXTUREPROJECTION_OCTAHEDRAL = 'octahedral';
```

We could rewrite that to:

```js
/**
 * Texture projection modes for storing texture data. These define how the texture's data is
 * interpreted or sampled, such as cubemap or equirectangular projections.
 *
 * @readonly
 * @enum {string}
 * @category Graphics
 */
export const TextureProjection = Object.freeze({
    /** No specific projection format is used. */
    None: 'none',

    /** Cubemap projection format. */
    Cube: 'cube',

    /** Equirectangular projection format. */
    Equirect: 'equirect',

    /** Octahedral projection format. */
    Octahedral: 'octahedral'
});```

So, for properties and functions that took one of those constants and was typed as `@type {string}`, we could update to `@type {TextureProjection}`.

### PR DESCRIPTION
**Pull request by willeastcott**
Fixes #7589 

### CODE REVIEW

File: src/platform/input/mouse-button.js [hunk 1]
```diff
@@ -0,0 +1,20 @@
+/**
+ * Mouse buttons.
+ *
+ * @readonly
+ * @enum {ValueOf<typeof MouseButton>}
```
kungfooman, line 5: Aha, no import for `ValueOf`? We either need to dev-install `type-fest` or just using the inline typedef:

```js
/** @template T @typedef {T[keyof T]} ValueOf */
```

I mean, technically we don't even need it at all, it just looks a bit nicer/shorter than this (which we could also use):

```js
@enum {typeof MouseButton[keyof typeof MouseButton]}
```

Maksims:
Please read: https://github.com/playcanvas/engine/issues/7589#issuecomment-2839650709

willeastcott:
Closing this one. It was a draft for the enum discussion in #7589 and never got finished (the `key.js` module it imports was never committed, and `MouseButton.NONE` should have been `None`).

Since then the codebase has settled on `PascalCase.UPPER_CASE` for object-style enums (see `scripts/esm/camera-frame.mjs` and `src/extras/input/sources/keyboard-mouse-source.js`), and the legacy `Keyboard`/`Mouse` classes this PR targeted have moved to the "Input Devices" category. If we pick the enum idea back up, it should start with the modern input module and keep the existing `UPPER_CASE` constants as the canonical exports. Discussion continues in #7589.

PR #7635, opened 2025-05-01
Update SWC options to target ES5 for UMD builds

marklundin:
The UMD build targeting `fully supports webgl and > 0.1% and not dead` was not wide enough to lower es6 `class X {}` syntax  to es5 `function X(){}`. This meant that super calls such as `PostEffects.call(this)` in user scripts would fail.

This PR ensures UMD build lowers to pseudo classes.

### DIFF
--- a/utils/rollup-swc-options.mjs
+++ b/utils/rollup-swc-options.mjs
@@ -13,6 +13,7 @@ function swcOptions(isDebug, isUMD, minify) {
     return {
         minify,
         jsc: {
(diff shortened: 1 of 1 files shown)


mvaligursky (approved):
Just check there is no significant size increase of the min versions of the build, but I would not expect so.

willeastcott (approved):

willeastcott:
Ideally, we'd cherry pick this and do `2.7.4`.

PR #7638, opened 2025-05-01
Example to demonstrate standard material chunk overrides

mvaligursky:
- initial version, which overrides diffuse color to time-blend colors, and vertex local position function to sway the trees, and also adds a ground fog.
- this will be further updated in a separate PR to add support for wgsl chunk overrides and related testing

Screenshot 2025-05-01 at 15 43 19


LeXXik:
Looks like this creates a new category (Shaders) in the examples list. It seems to clash with the current Graphics category, where some shader examples are housed, like Shader Wobble.

PR #7651, opened 2025-05-07
Shader creation no longer supplies dummy shader in none is specified

mvaligursky:
- internally, this was only used to for transform-feedback. Transform feedback handles it now.
- no need to do WGSL solution here

### DIFF
--- a/src/platform/graphics/shader-utils.js
+++ b/src/platform/graphics/shader-utils.js
@@ -149,7 +149,7 @@ class ShaderUtils {
                 ShaderUtils.precisionCode(device)}
                 ${sharedGLSL}
                 ${ShaderUtils.getShaderNameCode(name)}
-                ${options.fragmentCode || ShaderUtils.dummyFragmentCode()}`;
+                ${options.fragmentCode}`;
         }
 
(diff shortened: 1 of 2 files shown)


willeastcott (approved):

willeastcott:
Wondering if it makes sense to deprecate the TransformFeedback interface.

mvaligursky:
People use it, there's mentioned of that from time to time.

Maksims:
TransformFeedback - is definitely a powerful tool, and it is clearly stated that it is supported only by WebGL2. Since WebGL1 in engine v2 is deprecated, it makes it highly available, while WebGPU is still not commercially viable due to limited support across the board, and if project uses bespoke shaders - also is complicates use of WebGPU.

Due to that reason many developers still are relying exclusively on WebGL2, and have no need/benefit from using WebGPU renderer.

So deprecating it - will definitely be intrusive and unreasonable as the system is widely used, and is a powerful.

If there will be an alternative API - some abstract system for GPU processing of buffers, that is supported by both GAPIs - then deprecating TF is sensible.

PR #7653, opened 2025-05-07
Rename Key and Node classes

willeastcott:
* Rename `Key` to `AnimationKey`
* Rename `Node` to `AnimationNode`

The original names were too generic and prone to collision. For example:

https://developer.mozilla.org/en-US/docs/Web/API/Node

This will group `Node` with other `AnimationXYZ` classes. Note that `Key` was never public API anyway but renaming nevertheless.


mvaligursky (approved):

willeastcott:
> Is the private classes only?

`Node` is pubic, `Key` is private.

PR #7657, opened 2025-05-09
[BREAKING] Deprecated pc.createShader and pc.createShaderFromCode

mvaligursky:
This PR is breaking on in the rare case you were using `pc.createShader`.

Changes:
1. REMOVED `pc.createShader` function, which was very rarely if ever used. The newly provided function can be used instead.
2. DEPRECATED 'pc.createShaderFromCode' function. It can still be used, but logs a deprecated warning in debug build.
3. NEW REPLACEMENT functionality: 'ShaderUtils.createShader'

## NEW API
```
// create a shader, supports both GLSL and WGSL
static ShaderUtils.createShader(device, options);

// parameters
- device – The graphics device.
- options – Object for optional arguments.
- options.uniqueName – Unique shader name; reused if already existing.
- options.attributes – Maps vertex shader attributes to semantics.
- options.useTransformFeedback – Enables transform feedback (WebGL only, default false).
- options.vertexGLSL – Vertex shader code in GLSL.
- options.vertexWGSL – Vertex shader code in WGSL.
- options.vertexIncludes – Includes for resolving #include in vertex shader.
- options.vertexDefines – Defines for resolving #ifdef in vertex shader.
- options.fragmentGLSL – Fragment shader code in GLSL.
- options.fragmentWGSL – Fragment shader code in WGSL.
- options.fragmentIncludes – Includes for resolving #include in fragment shader.
- options.fragmentDefines – Defines for resolving #ifdef in fragment shader.
- options.fragmentOutputTypes – Fragment output types, defaults to vec4.

- returns: Shader
```


src/scene/shader-lib/shader-utils.js
```diff
@@ -0,0 +1,191 @@
...
+
+        return variant;
+    }
+}
+
+function createShader(device, vsName, fsName, useTransformFeedback = false, shaderDefinitionOptions = {}) {
+    Debug.removed('pc.createShader has been removed deprecated. Use ShaderUtils.createShader instead.');
+}
+
+function createShaderFromCode(device, vsCode, fsCode, uniqueName, attributes, useTransformFeedback = false, shaderDefinitionOptions = {}) {
```
slimbuck, line 159: shouldn't this live in deprecated?

slimbuck:
> How can it be rare when so many examples do it?! 

Oh I see they were using `createShaderFromCode`! :D

src/scene/shader-lib/shader-utils.js
mvaligursky: this is the current thinking, to improve three-shaking:
https://github.com/playcanvas/engine/pull/6900

src/scene/shader-lib/shader-utils.js
```diff
@@ -0,0 +1,191 @@
...
+
+        // add pass defines
+        const shaderPassInfo = ShaderPass.get(params.device).getByIndex(params.pass);
+        shaderPassInfo.defines.forEach((value, key) => defines.set(key, value));
+
+        return defines;
+    }
+
+    /**
+     * Process shader using shader processing options, utilizing cache of the ProgramLibrary
```
willeastcott, line 118:
```suggestion
     * Process shader using shader processing options, utilizing the cache of the ProgramLibrary.
```

willeastcott (approved):

PR #7661, opened 2025-05-09
docs: Document i18n's `change` event

Christopher-Hayes:
## Description

Formalizes the `set:locale` event on the `app.i18n` API.

This is helpful for updating your 3D scene in reaction to locale changes.

i18n already has an event that it triggers on locale change, `set:locale`, it is just not public API yet.

## Relevant Discussion

Discord: https://discord.com/channels/408617316415307776/408617316415307778/1370523328074219571

> 
> Chris Hayes
> **There's no way to figure out what events an API can trigger without just looking at the engine code right?**
> 
> **ie—with i18n, I'd like to make a handler for when the locale is changed, but there's no event list documented. Looking at the engine source code, it looks like set:locale is the event I want.**
> [I18n | Engine API Reference - v2.7.4](https://api.playcanvas.com/engine/classes/I18n.html)
>
> yaustar
> **If it's not in the docs, it's not public API and therefore subject to change without warning**
> **When would the locale change without you setting it yourself?**
>
> Chris Hayes
> **I mean I do set it myself, but I figured .i18n would fire an event when locale changed since it has a setter.**
> **I can always just make my own event, but figured I would check if there was a built-in one first.**
>
> yaustar
> **You can also fire the event on i18n object too**
>
> Chris Hayes
> **yeah true**
>
> yaustar
> **That way you know it will work for sure between engine updates**
>
> Chris Hayes
> **It would be a nice event to formalize in the engine though, I can see a lot of situations where you'd want to update your scene if the locale changes, rather than just using the last fetched .locale value.**
>
> yaustar
> **(or you can do a PR to make it public API 😄)**

## Misc

### DIFF
--- a/src/framework/asset/asset-localized.js
+++ b/src/framework/asset/asset-localized.js
@@ -1,4 +1,5 @@
 import { EventHandler } from '../../core/event-handler.js';
+import { I18n } from '../i18n/i18n.js';
 
 import { Asset } from './asset.js';
 
@@ -7,7 +8,7 @@ class LocalizedAsset extends EventHandler {
         super();
 
         this._app = app;
-        app.i18n.on('set:locale', this._onSetLocale, this);
+        app.i18n.on(I18n.EVENT_CHANGE, this._onSetLocale, this);
 
         this._autoLoad = false;
         this._disableLocalization = false;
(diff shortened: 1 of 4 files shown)


Maksims:
Do you think `change` - for the name of this event would be better here? `set:locale` - seems over specific, while being already on a very specific class.

Christopher-Hayes:
It _would_ be similar to `Asset.EVENT_CHANGE = 'change'`

https://api.playcanvas.com/engine/classes/Asset.html#event_change

That would make sense to me. I searched the forum and google search for people using `set:locale` and nothing is coming up, so I expect that name change would break very few projects.

Maksims:
If it wasn't public API, then changing it should not be a problem at all.

PR #7662, opened 2025-05-12
Sogs updates

### LINKED ISSUE #7537: Sogs followup work
We recently added runtime support for loading and rendering SOGS (plas) data in #7510.

The following work is outstanding to match compressed.ply rendering performance:
- reorder data after download (for better rendering performance) into morton order. this requires:
    - generate morton order for gaussian centers
    - reorder the GPU data using this order
- construct chunk data similar to compressed.ply
    - this will allow histogram-based sorting

In addition, the SOGS data could do with more massaging:
- combine labels_l and labels_u
- combine sh0 and opacities
- investigate using avif and possibly squoosh webp -> lossless -> slight loss & max effort where appropriate

slimbuck:
Fixes #7537

This PR updates SOGS runtime as follows:
- combine labels_u and labels_l data into single RG texture
- combine gaussian sh0 (color) and opacity into single texture
- implement more accurate quaternion storage in (2, 8, 8, 8) bits
- reorder data on GPU at load-time to improve runtime memory cache coherency

## SOGS Data Comparison

Examples of SOGS compression. All scenes contain 3 bands of spherical harmonics.

| Scene | Gaussians | PLY | Compressed PLY | SOGS |
|---|---|---|---|---|
| model-viewer | 4.3M | 1GB | 263MB | 56MB |
| model-viewer (1) | 60K | 14.6MB | 3.6MB | 1.2MB | 
|  model-viewer (2) | 720K | 174MB | 43MB | 11MB |

## Notes

- rendering performance is still slower than compressed.ply (on my windows RTX3070 ti machine), but faster than uncompressed ply. haven't investigated why
- load time reordering isn't fast due to the cpu sort. 4M particles takes around 600ms (on laptop). we could perform this sort ahead of time and store the order in another texture, but for now we choose the load-time approach.
- the tooling for generating SOGS scenes is based on a fork of https://github.com/vincentwoo/sogs. we will make this public soon or merge our changes into the mainline, depending on @vincentwoo 's preference.
- creating SOGS scenes requires (currently) a CUDA-enabled python install. we hope to create a browser-based solution.


willeastcott (approved):

mvaligursky (commented):
WGSL chunks need to be updated as well, otherwise this won't work on WebGPU

mvaligursky:
Also, it'd be great to have an engine example using this, so that we can test both platforms easily.

slimbuck:
There is some remaining follwup work:
- implement wgsl chunks
- get device recovery working
- add SOGS loader to engine examples

Maksims:
If the rendering is done by reading from textures, then there might be a significant performance reduction when accessing texture texels at random positions. As it does not allow GPU to cache chunks of it, so the reads from it will require a slow memory trips.
This is especially strong on mobile.

slimbuck:
@Maksims , yep, that's what this PR is addressing :)

Issue #7659, opened 2025-05-09
Shader Chunks public API


mvaligursky:
# Current system

Currently all shader chunks are referenced in a single object:
```
const shaderChunks = {
    alphaTestPS,
    ambientPS,
    aoPS,
    ...
};
```

and Material has similar solution for the overrides:
```
    Material.chunks = { .... }
```

Limitations:
- no support for WGSL, even though temporarily we have `const shaderChunksWGSL = { }`
- no way to tree-shake, even though we manually moved lightmapper list to a separate object, but that allows no overrides.
- no (good) way to detect if chunks were modified, new chunks added - no setter

# Proposal

Implement a pure static class storing chunks:
```
class ShaderChunks {
	static _glslChunks = new Map();  // name -> code
	static _wgslChunks = new Map(); // name -> code

        // assign chunks from an object
	static setGLSL(chunks) {
		for (const [name, code] of Object.entries(chunks)) {
			ShaderChunks._glslChunks.set(name, code);
  		}
	}

	static setWGSL(chunks) {
		for (const [name, code] of Object.entries(chunks)) {
			ShaderChunks._wgslChunks.set(name, code);
  		}
	}

        // remove named chunk
	static removeGLSL(name);
	static removeWGSL(name);

        // access named chunk
	static getGLSL(name) {
		return ShaderChunks._glslChunks.get(name);
	}

	static getWGSL(name) {
		return ShaderChunks._wgslChunks.get(name);
	}

	// @ignore - access all chunks, internal read-only use by the engine
	static glsl() { return ShaderChunks._glslChunks; }
	static wgsl() { return ShaderChunks._wgslChunks; }
}
```

In the engine, individual shader generators will register chunks they use. For example particle-generator could have this:
```
import { ShaderChunks } from ''; // the registry
import { particleStartPS } from 'chunks/particle-start.js'
import { particleEndPS } from 'chunks/particle-end.js'

ShaderChunks.registerGLSL({
	particleStartPS,
	particleEndPS
});
```

Material would have similar API to set up overrides:
```
class Material {
	_glslChunks = new Map();
	_wgslChunks = new Map();

	static setGLSL(chunks) { } // accepts object
	static setWGSL(chunks) { }  // accepts object

	static removeGLSL(name);
	static removeWGSL(name);

	static getGLSL(name) { }
	static getWGSL(name) { }
}
```

Other modules, for example render passes, can add their chunks this way:
```
import { ShaderChunks } from ''; // the registry

[...]

Issue #7588, opened 2025-04-23
Chunk API constants not documented properly


willeastcott:
Currently we have:

https://github.com/playcanvas/engine/blob/2e8c8a03f90249ec3c5576a44443e391ea40b6da/src/platform/graphics/constants.js#L2180-L2198

This results in:

![Image](https://github.com/user-attachments/assets/4e548b6d-a14f-474d-bc88-066100600de6)

I'm not sure what's desired here. Do we want to export and document every single one of these items in this seemingly ever growing list?

PR #7665, opened 2025-05-13
[BREAKING] New public API to customize shader chunks (Material and global)

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/7659
Fixes https://github.com/playcanvas/engine/issues/7588

# Breaking changes

### Material chunks

If you use `Material.chunks` API to replace shader chunks of a Material, you will get a deprecated warning in the debug mode, and you need to adjust your code use the new API. Old behaviour is preserved though in the meantime, so this is not a breaking change in majority of cases.

Error message: 
```
Material.chunks has been removed, please use Material.getShaderChunks instead. For example:
material.getShaderChunks(pc.SHADERLANGUAGE_GLSL).set("chunkName", "chunkCode")
```
### Global chunks

Similarly, if you use `pc.shaderChunks` to replace chunks globally, the API has changed and your use needs to be adjusted. Note that this is again not a breaking change in most cases, as backwards compatibility has been added, but in few cases the behaviour might be different.

# Details of the changes

1. This PR will by **default use WGSL shaders on WebGPU platform**, instead of GLSL transpilation.
2. Chunks are no longer stored on a global object (shared between all Apps using the same context), but per app (based on the GraphicsDevice)
3. Added support for WGSL chunks override.
4. This PR changes the existing public API for setting shader chunk overrides on a Material.
5. This PR changes the existing public API for setting global shader chunk overrides.
6. Updated rollup-shader-chunks plugin to process (remove comments ..) from literal strings tagged with wgsl in addition to existing glsl.
7. internal implementation allows adding chunks to the global maps in the module that depends on them, allowing tree-shaking. This PR moves particle chunks registration to particle component. If ParticleComponent is not used, the chunks can be three-shaken out. Similarly lightmapping and few individual shaders were handled this way. More to follow.

### Global chunks

Before, global GLSL shader chunks were accessible this way
```
	pc.shaderChunks.diffusePS = 'code';
```

now you set GLSL or WGSL shader chunks this way:
```
	pc.ShaderChunks.get(device, pc.SHADERLANGUAGE_GLSL).set('diffusePS', 'code');
	pc.ShaderChunks.get(device, pc.SHADERLANGUAGE_WGSL).set('diffusePS', 'code');
```

### Material chunks

Before, material GLSL shader chunks were accessible this way
```
	material.chunks.diffusePS = 'code';
```

now you set GLSL or WGSL shader chunks this way:
```
        material.getShaderChunks(pc.SHADERLANGUAGE_GLSL).set('diffusePS', 'code');
        material.getShaderChunks(pc.SHADERLANGUAGE_WGSL).set('diffusePS', 'code');
```

## Material class:
Screenshot 2025-05-19 at 12 45 50
Screenshot 2025-05-19 at 12 46 24

## ShaderChunks:
image

## ShaderChunkMap:
image


Maksims:
Please, don't remove `Material.chunks` - as it is a **public API**!
We do need a deprecation path here, not breaking scenario.
People use it.
This should not be breaking change, but provide a fallback to give developers time to migrate.

mvaligursky:
> Please, don't remove [`Material.chunks`](https://api.playcanvas.com/engine/classes/Material.html#chunks) - as it is a **public API**!

Added backwards compatibility now

mvaligursky:
> * Will we ever be in a position to drop the PS? They're fragment shaders.

We could rename them at any time, assuming we consider that change worth the hassle for the users.

mvaligursky:
> * Why have the `wgsl` and `glsl` as properties on the `shaderChunks` instead of a parameter for `set`? Is it conceivable we will deprecate/remove glsl chunks in the future (either because we remove WebGL2 after WebGPU hits 99% adoption or because we have a good way to transpile WGSL to GLSL?

No strong preference. This keeps chunks per language separated on both the API, as well as implementation (two Maps). We expose the maps directly, so that user can work on them, enumerate and similar.

When WebGPU will reach 99% adoption (in 5+ years), we'll refactor the API at that point I suspect.

Maksims:
As this PR implements backwards compatibility, it is no more "BREAKING".

mvaligursky:
> As this PR implements backwards compatibility, it is no more "BREAKING".

I don't think this handles all options the user could do with the previously exposed object (for example enumerate) and so this is breaking in some rare cases.

mvaligursky:
Updated to new API as discussed @willeastcott @slimbuck , please review this when you have some time.

Maksims:
Does new API has to be so verbose?
```javascript
material.getShaderChunks(pc.SHADERLANGUAGE_GLSL).set('diffusePS', 'code');
```
Devs will write this:
```javascript
material.getShaderChunks('glsl').set('diffusePS', 'code');
```

But why do we need extra method to do so, if we just could have an object:
```javascript
material.chunks.glsl.set('diffusePS', 'code');
```

mvaligursky:
This way it aligns better with global overrides
```
pc.ShaderChunks.get(device, pc.SHADERLANGUAGE_GLSL).set('diffusePS', 'code');
```

Maksims:
Should the global overrides be on the device if is requires one?

mvaligursky:
> Should the global overrides be on the device if is requires one?

Could you please clarify the question? I'm not sure what you're asking.

mvaligursky:
Ah I see. Not really, it's only that chunks are associated with the device, but not owned by the device.
Same way many other functions / constructors need device, but are not implemented on the device instead.

Maksims:
In 99%+ developers have a single device, could we simplify life for them, e.g.:
```javascript
pc.shaderChunks.glsl.set(name, code, [device]);
```

mvaligursky:
We do not want to have an API that might suddenly stop working if another instance of the app is created.

It's how the engine works in many cases. You always do
```
new Mesh(device)
```` 

there is no version without the device for some cases.

Majority of uses never overrides chunks, so writing extra few characters when that is needed makes little difference.

mvaligursky:
1. Correct, this will need to be done after the app is initialized, typically done from the script's `initialize`.
2. Chunks are not GPU objects and so the device lost has no impact on them, no user handling needed.
3. Correct. Handling this inside `initialize` should just work.
4. Yes, needs to be assigned per device. Doing so inside `initialize` will do that.

Yes, this is slightly less convenient, but internally we agreed the wins here are worth it, as we get regular issues with multi-app project where the engine uses global objects. We want to remove those as much as possible.

The inconvenience is just a very minor from what I expect.

slimbuck (approved):

PR #7666, opened 2025-05-13
ESM script load URL fix

kpal81xd:
- Fixes issue where query parameters are dropped when loading ESM script in GH-7161 (id and branchId) causing launcher to fail to fetch asset source

### DIFF
--- a/src/framework/handlers/script.js
+++ b/src/framework/handlers/script.js
@@ -82,7 +82,7 @@ class ScriptHandler extends ResourceHandler {
         const isEsmScript = basePath.endsWith('.mjs');
 
         if (isEsmScript) {
-            this._loadModule(basePath, onScriptLoad);
+            this._loadModule(url.load, onScriptLoad);
         } else {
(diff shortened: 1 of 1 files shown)


slimbuck (approved):

marklundin:
Omitting the query string is intentional here. See this comment https://github.com/playcanvas/editor/issues/1236#issue-2711752249.

If the engine loads `utils.mjs?hash=123` and later a user directly imports `import 'utils.mjs'` it will treat these as 2 modules, which is a problem if they have internal state. This issue was reported on Discord originally https://discord.com/channels/408617316415307776/1242436398678671401/1312763719062913065

If scripts are not loading, it's may be because "Bypass for network" is disabled in Dev tools > Application > SW. This will bypass the loading behaviour

image

marklundin:
We should probably handle this better in the editor, potentially adding a check and warning users

marklundin:
Closing this, as it will cause issues if merged

Issue #6135, opened 2024-03-11
Implement KHR_materials_anisotropy glTF extension support


mvaligursky:
https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_anisotropy/README.md

![image](https://github.com/playcanvas/engine/assets/59932779/1758811a-8a27-49b3-b6c2-037313732245)

test models: https://github.com/cx20/gltf-test/tree/master/tutorialModels/AnisotropyBarnLamp

emadurandal:
Hello, I was invited by Mr. Eastcott to contribute PlayCanvas.
https://x.com/willeastcott/status/1920844676648538372

[My library](https://github.com/actnwit/RhodoniteTS) supports KHR_materials_anisotropy, so maybe I can help you.

How can we start collaborating?

mvaligursky:
Hi .. nice to hear you're keen to tackle this one, that'd be fantastic!

The way this probably needs to be done:
- add a new example, perhaps based on [this](https://playcanvas.vercel.app/#/graphics/clear-coat) but loading the anisotropic test model
- figure out what properties we need to add to `StandardMaterial` and add them.
- update GlbParser to parse the extension and set those properties on the material, similarly to other extensions
- update StandardMaterialOptionsBuilder to convert those material settings to options for the shader generator
- update LitShader to convert those options to defines that the shader consumes
- and then update shader chunks / add new chunks as needed, using those defines, to add the shader functionality. In GLSL at least (we can then convert to WGSL), or both.

I skipped lots of details, but this is a rough overview.

emadurandal:
Thank you for the guidance.
First, I will proceed to understand how those PlayCanvas shaders work.

emadurandal:
I'm currently working on understanding the code with reference to Clearcoat's implementation.
However, since the related code has been written for various files (including around lights, etc.), I am wondering where to start.
I need time to explore this a little.

emadurandal:
@mvaligursky 

I learned that StandardMaterial already has a property 'anisotropy'. 
https://github.com/playcanvas/engine/blob/0e1c148decd4a86565c15074690a1d12e80bfbc2/src/scene/materials/standard-material.js#L134-L141

Also, it seems that there is already a shader function in GGX that takes Anisotropy into account.
https://github.com/playcanvas/engine/blob/main/src/scene/shader-lib/chunks-glsl/lit/frag/lightSpecularAnisoGGX.js

[...]

PR #7668, opened 2025-05-14
[BREAKING] Feature: KHR_materials_anisotropy

emadurandal:
This PR is a draft implementation of KHR_materials_anisotropy.
This PR is still in draft form and is intended to serve as a springboard for discussion of implementation at https://github.com/playcanvas/engine/issues/6135 .

Notice
* The visual behavior differs from the existing anisotropy implementation.
* Only WebGL(GLSL) version is implemented. WebGL(WSGL) is not yet.

image
image
image
image

I also noticed during implementation that there seems to be a discrepancy in the results in the implementation variation of the getTBN() shader function. The code for the “object space TBN” case seems to compute a different (possibly incorrect) result than the codes for the other branches. When going through the “object space TBN” branch, the result of the anisotropic direction is the opposite of what is expected.

https://github.com/playcanvas/engine/blob/48a2276427d7860354655fdcc59f10572459e5ff/src/scene/shader-lib/chunks-glsl/lit/frag/TBN.js


emadurandal:
@Maksims I may have used the wrong meaning of “breaking”. There are no API-disruptive changes in this PR. I just wanted to communicate that the appearance of anisotropic reflections could be different than before. If it is inappropriate to add “breaking” in such a case, I will remove it.
(I am new to PlayCanvas development and am still getting used to the conventions.)

mvaligursky:
This looks pretty great! Some feedback:
1. I would only add a single texture, instead of `anisotropyMap` and `anisotropyRotationMap`. Data is always stored in a single texture, so no need for it to consume two samplers.
2. Similarly then for the new chunks - we should just have one, and add defines as needed to control its behavior.

mvaligursky:
I agree this should be marked as BREAKING, as the visuals of this feature enabled will be different then before. We use this flag to bring it up to attention of our users.

emadurandal:
> Some feedback:

Thank you for the feedback.
Please wait a moment while I modify the code.
If we were to use a single Sampler, we would not be able to specify the channel in MapChannel.
It will be fixed in the rg channel for Rotation and in the b channel for Strength.

emadurandal:
I have modified the code to accommodate Feedback.

BTW,
> I confirm I have read the [contributing guidelines](https://github.com/playcanvas/engine/blob/master/.github/CONTRIBUTING.md) and signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

I have included the template text in the Post as it was when I created the PR, do I need to sign the Contributor License Agreement? If necessary, we will sign it.

emadurandal:
> This will affect people's projects indeed, it should be possible to keep the same value range, and do normalization in enginee.

Unfortunately, this time this `anisotropy` has changed not only the value range but also its meaning.
Conventional `anisotropy` affects not only the strength of the effect, but also the direction of the anisotropy.
However, the `anisotropy` in my proposal deals with the direction of anisotropy in the new `anisotropyRotation`, and the `anisotropy` only affects the strength of the effect.

mvaligursky:
I'd suggest to:
- add new properties as `StandardMaterial.anisotropyIntensity` and `StandardMaterial.anisotropyRotation`
- remove `StandardMaterial.anisotropy` from`StandardMaterial`, and add it to deprecated.js - see how `shininess` is handled there which remaps old value to new

emadurandal:
> I'd suggest to:

Indeed, that policy may be a good idea.

One point, I think it would be better to use StandardMaterial.anisotropyStrength instead of StandardMaterial.anisotropyIntensity, in line with the KHR_materials_anisotropy specification.

mvaligursky:
> StandardMaterial.anisotropyIntensity, in line with the KHR_materials_anisotropy specification.

I considered this, but we have multiple properties in StandardMaterial with named xxxIntensity, so it'd be nice to be consistent.

emadurandal:
> I considered this, but we have multiple properties in StandardMaterial with named xxxIntensity, so it'd be nice to be consistent.

I understand. So let's go with StandardMaterial.anisotropyIntensity.

emadurandal:
I have implemented to the various feedbacks.

By the way, here is what it looked like before this PR in the MATERIAL ANISOTROPY example:
スクリーンショット 2025-05-17 18 14 41

On the other hand, here is how the same example looks with anisotropy added to deprecated.js in this PR:
スクリーンショット 2025-05-17 18 17 06

In this pre-PR implementation of anisotropy, materials with high gloss (low roughness) had code that made anisotropy ineffective. On the other hand, KHR_materials_anisotropy now has a specification that allows anisotropy to work even for materials with high gloss (low roughness). Therefore, although there is no complete visual compatibility, you can see that materials with low gloss (high roughness) are compatible in the direction of anisotropy.

willeastcott:
Can we aim to get this included in v.2.8.0, @mvaligursky?

src/scene/materials/standard-material.js
```diff
@@ -130,16 +130,18 @@ const _tempColor = new Color();
...
- *
- * - When anisotropy == 0, specular is isotropic.
- * - When anisotropy < 0, anisotropy direction aligns with the tangent, and specular anisotropy
- * increases as the anisotropy value decreases to minimum of -1.
- * - When anisotropy > 0, anisotropy direction aligns with the bi-normal, and specular anisotropy
- * increases as anisotropy value increases to maximum of 1.
- *
+ * - When anisotropyIntensity == 0, specular is isotropic.
+ * - Specular anisotropy increases as anisotropyIntensity value increases to maximum of 1.
+ * @property {number} anisotropyRotation Defines the rotation of anisotropy. The value is in radian.
```
willeastcott, line 138: Every other angle passed in the PlayCanvas Engine API is in degrees, so this looks strange to me. Any reason not to use degrees?

src/scene/materials/standard-material.js
```diff
@@ -130,16 +130,18 @@ const _tempColor = new Color();
  * @property {string} specularityFactorVertexColorChannel Vertex color channels to use for specularity factor. Can be
  * "r", "g", "b", "a", "rgb" or any swizzled combination.
  * @property {boolean} enableGGXSpecular Enables GGX specular. Also enables
- * {@link StandardMaterial#anisotropy}  parameter to set material anisotropy.
- * @property {number} anisotropy Defines amount of anisotropy. Requires
+ * {@link StandardMaterial#anisotropyIntensity}  parameter to set material anisotropy.
```
willeastcott, line 133:
```suggestion
 * {@link StandardMaterial#anisotropyIntensity} parameter to set material anisotropy.
```

src/scene/materials/standard-material.js
emadurandal: > Every other angle passed in the PlayCanvas Engine API is in degrees,.
I didn't know about that, I'll change it to degrees.

emadurandal: Fixed at https://github.com/playcanvas/engine/pull/7668/commits/052baff38e0e8d473322f391d534203da5ea75a0

emadurandal:
If there is no problem working as is, I will implement the WGSL code as well

willeastcott:
> If there is no problem working as is, I will implement the WGSL code as well

OK, great, thanks!

emadurandal:
I implemented the WGSL code.
And I am using the added examples to verify the operation, but it is a little strange.
It seems to me that the GLSL version of the code is being used (perhaps converted to WGSL), even when in WebGPU mode.
Is there any way to run the WGSL code directly?

emadurandal:
I have confirmed that the WGSL version also works.

src/framework/parsers/glb-parser.js
```diff
@@ -1108,6 +1108,28 @@ const extensionIridescence = (data, material, textures) => {
     }
 };
 
+const extensionAnisotropy = (data, material, textures) => {
+
+    material.enableGGXSpecular = true;
+
+    if (data.hasOwnProperty('anisotropyStrength')) {
+        material.anisotropy = data.anisotropyStrength;
```
mvaligursky, line 1116: Should this set anisotropyIntensity, instead of going via the deprecated API?

src/deprecated/deprecated.js
```diff
@@ -662,6 +663,21 @@ Object.defineProperty(StandardMaterial.prototype, 'useGammaTonemap', {
     }
 });
 
+Object.defineProperty(StandardMaterial.prototype, 'anisotropy', {
```
mvaligursky, line 666: please add Debug.deprecated to both the getter and setter, to warn user there is a newer API to use.

src/scene/shader-lib/chunks-glsl/standard/frag/anisotropy.js
```diff
@@ -0,0 +1,27 @@
...
+    dAnisotropy = 0.0;
+    dAnisotropyRotation = vec2(1.0,0.0);
+
+#ifdef LIT_GGX_SPECULAR
+    dAnisotropy = material_anisotropyIntensity;
+    dAnisotropyRotation = material_anisotropyRotation;
+#endif
+
+    #ifdef STD_ANISOTROPY_TEXTURE
+    dAnisotropy *= texture2DBias({STD_ANISOTROPY_TEXTURE_NAME}, {STD_ANISOTROPY_TEXTURE_UV}, textureBias).b;
```
mvaligursky, line 18: only sample the texture one time to a temporary and get the data out, instead of twice.

src/scene/shader-lib/chunks-glsl/standard/frag/stdFrontEnd.js
```diff
@@ -197,6 +206,14 @@ export default /* glsl */`
                 litArgs_clearcoat_worldNormal = ccNormalW;
             #endif
 
+            #ifdef LIT_SPECULAR
```
mvaligursky, line 209: same here

src/scene/shader-lib/chunks-wgsl/standard/frag/anisotropy.js
```diff
@@ -0,0 +1,26 @@
...
+    dAnisotropy = 0.0;
+    dAnisotropyRotation = vec2f(1.0, 0.0);
+
+#ifdef LIT_GGX_SPECULAR
+    dAnisotropy = uniform.material_anisotropyIntensity;
+    dAnisotropyRotation = uniform.material_anisotropyRotation;
+#endif
+
+#ifdef STD_ANISOTROPY_TEXTURE
+    dAnisotropy *= textureSampleBias({STD_ANISOTROPY_TEXTURE_NAME}, {STD_ANISOTROPY_TEXTURE_NAME}Sampler, {STD_ANISOTROPY_TEXTURE_UV}, uniform.textureBias).b;
```
mvaligursky, line 17: same here, sample the texture once

src/scene/shader-lib/chunks-wgsl/standard/frag/stdFrontEnd.js
```diff
@@ -85,6 +85,15 @@ export default /* wgsl */`
             #include "clearCoatNormalPS"
         #endif
 
+        // anisotropy
+        #ifdef LIT_SPECULAR
```
mvaligursky, line 89: one liner define pls.

src/scene/shader-lib/collections/shader-chunks-glsl.js
```diff
@@ -13,6 +13,7 @@ import blurVSMPS from '../chunks-glsl/lit/frag/blurVSM.js';
 import clearCoatPS from '../chunks-glsl/standard/frag/clearCoat.js';
 import clearCoatGlossPS from '../chunks-glsl/standard/frag/clearCoatGloss.js';
 import clearCoatNormalPS from '../chunks-glsl/standard/frag/clearCoatNormal.js';
+import anisotropyPS from '../chunks-glsl/standard/frag/anisotropy.js';
```
mvaligursky, line 16: these are sorted alphabetically

src/scene/shader-lib/collections/shader-chunks-glsl.js
```diff
@@ -187,6 +188,7 @@ const shaderChunksGLSL = {
     clearCoatPS,
     clearCoatGlossPS,
     clearCoatNormalPS,
+    anisotropyPS,
```
mvaligursky, line 191: these are sorted alphabetically

src/scene/shader-lib/collections/shader-chunks-wgsl.js
```diff
@@ -182,6 +183,7 @@ const shaderChunksWGSL = {
     clearCoatPS,
     clearCoatGlossPS,
     clearCoatNormalPS,
+    anisotropyPS,
```
mvaligursky, line 186: these are sorted alphabetically

src/scene/shader-lib/collections/shader-chunks-wgsl.js
```diff
@@ -13,6 +13,7 @@ import blurVSMPS from '../chunks-wgsl/lit/frag/blurVSM.js';
 import clearCoatPS from '../chunks-wgsl/standard/frag/clearCoat.js';
 import clearCoatGlossPS from '../chunks-wgsl/standard/frag/clearCoatGloss.js';
 import clearCoatNormalPS from '../chunks-wgsl/standard/frag/clearCoatNormal.js';
+import anisotropyPS from '../chunks-wgsl/standard/frag/anisotropy.js';
```
mvaligursky, line 16: these are sorted alphabetically

mvaligursky:
This looks great! I added a bunch or small comments, should be easy to wrap it up now.

src/framework/parsers/glb-parser.js
emadurandal: Fixed at https://github.com/playcanvas/engine/pull/7668/commits/bc2cc20072dd2189de26b7da5ba033f4932714f7

src/deprecated/deprecated.js
emadurandal: Fixed at https://github.com/playcanvas/engine/pull/7668/commits/29f56d6a68c1fe355e96ddb27745d15f08d875c1

src/scene/shader-lib/chunks-glsl/standard/frag/stdFrontEnd.js
```diff
@@ -84,6 +84,15 @@ export default /* glsl */`
             #include "clearCoatNormalPS"
         #endif
 
+        // anisotropy
+        #ifdef LIT_SPECULAR
```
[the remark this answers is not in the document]
emadurandal: Fixed at https://github.com/playcanvas/engine/pull/7668/commits/d9bbc21e4db4db7dc887fa7f82cf9e974af24b26

src/scene/shader-lib/chunks-glsl/standard/frag/stdFrontEnd.js
emadurandal: Fixed at https://github.com/playcanvas/engine/pull/7668/commits/d9bbc21e4db4db7dc887fa7f82cf9e974af24b26

src/scene/shader-lib/chunks-wgsl/standard/frag/stdFrontEnd.js
emadurandal: Fixed at https://github.com/playcanvas/engine/pull/7668/commits/d9bbc21e4db4db7dc887fa7f82cf9e974af24b26

src/scene/shader-lib/collections/shader-chunks-glsl.js
emadurandal: Fixed at https://github.com/playcanvas/engine/pull/7668/commits/6bdd3b2e80f72d748d4aa7afd251b43fdcb8f4f2

src/scene/shader-lib/collections/shader-chunks-wgsl.js
emadurandal: Fixed at https://github.com/playcanvas/engine/pull/7668/commits/6bdd3b2e80f72d748d4aa7afd251b43fdcb8f4f2

src/scene/shader-lib/collections/shader-chunks-wgsl.js
emadurandal: Fixed at https://github.com/playcanvas/engine/pull/7668/commits/6bdd3b2e80f72d748d4aa7afd251b43fdcb8f4f2

emadurandal:
Thank you for the comments. All fixed.

mvaligursky:
by the way, you can run `npm run lint -- --fix` to fix most of the lint issues automatically

mvaligursky (approved):
Great work, thanks!

Issue #1721, opened 2019-10-23
Particle System simulation on 60Hz+ screens.


Maksims:
There is a currently an issue, where simulating particles on screens with higher than 60Hz refresh rate, leads to particles to look too fast. So changing`fixedTimeStep` fixes an issue. But as I assume it has cost associated to it. For example if increasing it to lower number, like `1 / 240` for 240Hz screens, will have performance hit for 60Hz screens and mobile.

Could there be a global flag for it, and developer should manually decide somehow, what it to set to.
There should be a solution to such problem, as monitors and mobile devices with 90, 144, 240, etc. are now becoming a common thing.

https://github.com/playcanvas/engine/blob/f1ae19316054c1bb88a38f9d44177133a14bade0/src/scene/particle-system/particle-emitter.js#L298

Maksims:
I've ended up doing something like this: track average FPS over last 20 frames, if it is showing steady good FPS close to 90, then change fixedTimeStep on particles to `1 / 90`. Then check to 120 then 144 and then 240. It progressively goes up. Averaging - is important, as random spikes in dt are common thing, so single frame time is not reliable.

Something without specific frames (60, 90, 120, 144, 240) steps, but threshold of 20%+ fps than previously steady one - could be good way of switching to better fixedTimeStep for particles.
Can even start low, like from 30fps, this would provide better performance on weak devices.

slimbuck:
Hey @Maksims,

That sounds like a cool solution. In general though changing simulation step time is actually dangerous as you may get subtly different results at different frame rates. Ideally the particle system would always run with a fixed frame step and then interpolate between two steps to match the current frame rate.

I am not familiar with the particle system implementation in playcanvas but I expect making it double buffer wouldn't be difficult (if it doesn't already). Perhaps it's something we can look to do in future.

FeaturedSpace:
I would definitely rather do some sort of hz detection, which is normally possible through other 3d engines (that are not on the web, such as DirectX and OpenGL).

Would it be possible to detect a machine's set HZ level or something like that?

mvaligursky:
is there a strong reason to run it with a fixed time step instead of variable dt? Moar fx system's I worked on / used were using variable dt without a problem.

Maksims:
> In general though changing simulation step time is actually dangerous as you may get subtly different results at different frame rates.

[...]

PR #7675, opened 2025-05-16
[Fix] Fixed particles update speed on high refresh screens

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/1721 (most likely)

- on high screen refresh rates (120Hz+) and multiple particle systems, in some cases a particle simulation with multiple system simulating at the same time would run at higher speed. The issues seems to be that the numSteps is not reset to 0 for each particle, and in case particle does not need update, it would still run number of simulation steps from a previous particle system.

forum thread: https://forum.playcanvas.com/t/is-particles-speed-frame-rate-dependant/39952

### DIFF
--- a/src/framework/components/particle-system/system.js
+++ b/src/framework/components/particle-system/system.js
@@ -200,7 +200,6 @@ class ParticleSystemComponentSystem extends ComponentSystem {
 
     onUpdate(dt) {
         const components = this.store;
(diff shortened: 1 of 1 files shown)


slimbuck (approved):

mvaligursky:
cherry picked to V1

PR #7679, opened 2025-05-19
Improved docs for Material.setDefine

mvaligursky:
handle ‘true’ parameter

### DIFF
--- a/src/scene/materials/material.js
+++ b/src/scene/materials/material.js
@@ -724,7 +724,7 @@ class Material {
      * parts of the shader code.
      *
      * @param {string} name - The name of the define to set.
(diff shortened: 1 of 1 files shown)


src/scene/materials/material.js
```diff
@@ -724,8 +724,8 @@ class Material {
      * parts of the shader code.
      *
      * @param {string} name - The name of the define to set.
-     * @param {string|undefined|false} value - The value of the define. If undefined or false, the
-     * define is removed.
+     * @param {string|undefined|false|true} value - The value of the define. If undefined or false, the
```
willeastcott, line 727: Why `false|true` instead of `boolean`?

mvaligursky: updated

willeastcott (approved):

PR #7690, opened 2025-05-21
Render passes use WGSL chunks

mvaligursky:
- all render passes using shaders have now shaders stored as locally imported chunks
- apart from compose chunk, all those chunks were also converted to WGSL and are used on WebGPU
- compose chunk will be refactored a bit first, before converted to WGSL, all in a separate PR


src/scene/shader-lib/chunks-wgsl/render-pass/frag/depthAwareBlur.js
```diff
@@ -0,0 +1,64 @@
+export default /* glsl */`
```
willeastcott, line 1:
```suggestion
export default /* wgsl */`
```

src/extras/render-passes/render-pass-coc.js
```diff
@@ -19,37 +23,23 @@ class RenderPassCoC extends RenderPassShaderQuad {
...
-                #endif
-
-                gl_FragColor = vec4(cocFar, cocNear, 0.0, 0.0);
-            }`
-        );
+        this.shader = ShaderUtils.createShader(device, {
+            uniqueName: `CocShader-${nearBlur}`,
+            attributes: { aPosition: SEMANTIC_POSITION },
+            vertexChunk: 'RenderPassQuadVS',
+            fragmentChunk: 'CocPS',
```
willeastcott, line 40: Shouldn't this be `cocPS`?

mvaligursky: good point, all chunk names start with a lower case character. Let me fix, there's about 5 of these.

willeastcott (approved):

PR #7702, opened 2025-05-22
Break compose shader to chunks for easier customization + convert to WGSL

mvaligursky:
- reorganise compose shader chunk to handle individual parts on a more unified manner
- break it into multiple chunks per its functionality (bloom, DOF ...)
- convert it all to WGSL as well
- fix to ACES2 WGSL chunk


willeastcott (approved):
Is the `frag` folder in the right place? Looks a bit odd...

Issue #7697, opened 2025-05-21
Proposal for E2E (visual) regression test suite


emadurandal:
## Motivation

It is possible that additional features, specification changes, etc. will cause the rendering results to be different from what they were before. For this reason, we propose the introduction of an E2E (visual) test environment. This enables visual regression testing by comparing a snapshot of the rendered image with the correct image using headless rendering by Chrome.
This mechanism will allow early detection of regression bugs that cannot be covered by unit test.
Since it is headless execution, this mechanism can be incorporated into CI, such as Github Actions.

## Existing tools

The Jest test library has an extension called jest-image-snapshot, which compares the results of the correct image and current drawings and treats any differences as errors. It also outputs the difference images. It is also possible to set a threshold for how much difference is acceptable.

There is a port of jest-image-snapshot for non-Jest test libraries (vitest and mocha).
(I have used both the original jest-image-snapshot and the port for vitest with success on my RhodoniteTS. see the following screen capture).
![Image](https://github.com/user-attachments/assets/76674990-d268-4fff-9c5c-ecb2c826cdbd)

Since PlayCanvas already seems to use mocha for unit tests, you might want to use the porting for mocha.
https://sebastianlandwehr.com/blog/image-snapshot-testing-with-mocha/

## Things to consider

* It would make sense to use existing examples that already exist as visual test items.
  * It may be necessary to add a small code to the existing examples that notifies mocha that the rendering is complete. 
  * However, no other changes will probably be necessary, and the existing examples structure can be appropriated as is.
* Whether to run visual tests with the GPU enabled or in software emulation depends on your objectives.
  * Software emulation can currently test only the WebGL version, but the results will be the same in all environments, so if you want to test locally as well as on CI, software emulation is the way to go.
  * On the other hand, GPU-enabled visual tests can also test WebGPU, but the results will vary slightly depending on the execution environment, so they are basically for CI execution only, and require either a GPU-enabled instance of Github or a dedicated machine and a Self Hosted Github Action.

Thought?

Maksims:
This is definitely great!

Sometimes the change is intentional, so it would be great to ensure that visual tests don't fail, but provide an insight - to what have changed, as some changes are expected.

emadurandal:
I created PR. Please check it out https://github.com/playcanvas/engine/pull/7705

PR #7705, opened 2025-05-23
[Proposal] E2E visual test environment

emadurandal:
This PR proposes an E2E (Visual) Regression testing environment using existing examples. 
(under discussion in https://github.com/playcanvas/engine/issues/7697)

image

## Usage

I have introduced some packages, please npm install them.

Since 'examples/cache/metadata.mjs' is referenced to extract test items, so please npm run build in advance in the examples directory.

Then, run the following command on project root.

```
$ npm run test:e2e 
```

The examples local server will then start automatically and the E2E test will begin.

On the first run, the rendered images are saved as the "correct images".
The second and subsequent runs will compare the images with the correct images created in the first run.

The following command will update the correct images.

```
$ npm run test:e2e:update
```

## Current Issues

* Many existing examples are animated or have a moving camera. They are likely to fail the test due to timing issues with when the screenshot is taken.
  * It would be good to add an option to examples to disable animation for testing purposes.
* I think there are some things that are simply not suitable for testing, so we should be able to specify exclude in some way as well.

This PR is still rough around the edges, so your feedback is welcome.


emadurandal:
Further updates.

* Supports running on Windows
* Switched to headless mode.
* Added correct images

emadurandal:
Thanks for the feedback. I understand the situation. It's not easy to create the best solution.

> One idea I had thought of was instead of storing the screenshots would be a system to compare the examples from two commits through recording them both and comparing the videos.

I tried to find a video version of the jest-imege-snapshot, but unfortunately there does not seem to be such a version. 

I leave it to you all to handle this PR. Hopefully this is the beginning of a discussion on good solutions.

Issue #4315, opened 2022-06-09
Implement color correction


mvaligursky:
Details to be figured out, but based on something equivalent to 3D LUT texture.
This could be done either inlined in the shaders, or as a post-processing.

Maksims:
Inlined in a shader would be faster on mobile, as any post-process on mobile leads to an extra few ms overheads due to mobile GPUs.
Also inline will allow to control it per material and per mesh, just like with tone mapping and gamma correction currently.

mvaligursky:
perhaps we implement support for .CUBE format LUTs, which is the most common color grading format, used by Video Editing Software. Easy to parse. Represents a cube texture, easy to parse. Example attached.

[Cine-4-debarup.zip](https://github.com/playcanvas/engine/files/8926622/Cine-4-debarup.zip)

PR #7720, opened 2025-05-27
Added color correction using LUT texture to CameraFrame postprocessing

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/4315

Added color lookup to CameraFrame postprocessing. Live preview: https://engine-pso3eqdcf-playcanvas.vercel.app/#/graphics/hdr

### New API
```
    cameraFrame.colorLUT.texture = texture;
    cameraFrame.colorLUT.intensity = 1.0;
```

Texture is a common .cube format stored in a horizontal strip
lut-blue

There are multiple free online tools to convert to this format. I used https://www.color.io/free-online-lut-converter - loaded a cube file, and exported to `Cube Strip 2D`.

This is also exposed in the `CameraFrame.mjs` script, but at the moment does not parse correctly. @marklundin is investigating.

https://github.com/user-attachments/assets/304eee07-93c6-4011-b58c-ee39c0f86689


willeastcott:
Isn't this just a new tonemap type? Like, maybe, `TONEMAP_LUT` or `TONEMAP_TEXTURE`?

mvaligursky:
> Isn't this just a new tonemap type? Like, maybe, `TONEMAP_LUT` or `TONEMAP_TEXTURE`?

Tone map converts HDR -> LDR. The look up works on LDR, so it cannot be used instead of tonemap, but along with it.

scripts/esm/camera-frame.mjs
```diff
@@ -217,6 +219,23 @@ class Grading {
...
+class ColorLUT {
+    /**
+     * @attribute
+     * @type {Asset}
+     * @resource texture
+     */
+    texture = null;
+
+    /**
+     * @range [0, 1]
```
marklundin, line 232: You can use `@visibleif {texture}` which uses truthy values. The parser isn't actually working with null at the moment

mvaligursky: I tried that as a first thing, but that didn't work. Does it work on your side?

mvaligursky:
> In the example, should the LUT texture filtering be set to "nearest"?

It needs to be linear, which it is. Otherwise you get banding due to relatively low resolution (32 pixels)

slimbuck:
In the example the controls stop working when HDR is disabled. Is that expected?

mvaligursky:
> https://www.color.io/free-online-lut-converter

I'd say yes as whole postprocessing is disabled. Ideally we hide the slider.

slimbuck (approved):

PR #7725, opened 2025-05-30
Quat dot product and lerp

kpal81xd:
- Adds methods for dot product and linear lerping between two quaternions

### DIFF
--- a/src/core/math/quat.js
+++ b/src/core/math/quat.js
@@ -740,16 +740,59 @@ class Quat {
     }
 
     /**
+     * Set the values of the quaternion from an array.
+     *
+     * @param {number[]|ArrayBufferView} arr - The array to set the quaternion values from.
+     * @param {number} [offset] - The zero-based index at which to start copying elements from the
+     * array. Default is 0.
+     * @returns {Quat} Self for chaining.
+     * @example
+     * const q = new pc.Quat();
+     * q.fromArray([20, 10, 5, 0]);
+     * // q is set to [20, 10, 5, 0]
+     */
+    fromArray(arr, offset = 0) {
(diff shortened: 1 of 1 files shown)


src/core/math/quat.js
```diff
@@ -301,6 +315,35 @@ class Quat {
...
+     * @example
+     * const q1 = new pc.Quat(-0.11, -0.15, -0.46, 0.87);
+     * const q2 = new pc.Quat(-0.21, -0.21, -0.67, 0.68);
+     *
+     * const result = new pc.Quat();
+     * result.lerp(q1, q2, 0);   // Return q1
+     * result.lerp(q1, q2, 0.5); // Return the midpoint interpolant
+     * result.lerp(q1, q2, 1);   // Return q2
+     */
+    lerp(lhs, rhs, alpha) {
```
willeastcott, line 338: Uhm. Why would you ever lerp quaternions instead of slerp them?

willeastcott: OK, fair enough.

src/core/math/quat.js
```diff
@@ -301,6 +315,35 @@ class Quat {
         return this.x * this.x + this.y * this.y + this.z * this.z + this.w * this.w;
     }
 
+
+    /**
+     * Performs a linear interpolation between two quaternions. The result of the interpolation
+     * is written to the quaternion calling the function.
```
willeastcott, line 321: Worth adding a comment here (and in slerp) about when you would use one over the other?

willeastcott (approved):

willeastcott:
@kpal81xd Minor point but adding these kinds of functions should also include unit tests (easily done with ChatGPT).

PR #7729, opened 2025-06-02
Mesh-based Skydome can write depth during prepass, to be comparible with DOF

mvaligursky:
New public API:

```
    // enable depth writing for the sky, for DOF to work on it
    app.scene.sky.depthWrite = true;
```

When `skydome` or `skybox` sky is used (and not `infinite`), it can write depth during the depth-prepass. This allows it to be affected by DOF.
Adjusted example to show how to use this.

Also other small example improvements - darker shadows, automatic shadow distance.

https://github.com/user-attachments/assets/e6bda984-16b4-494d-885a-f1b68c9b1cc5

### DIFF
--- a/src/scene/skybox/sky.js
+++ b/src/scene/skybox/sky.js
@@ -136,6 +136,11 @@ class Sky {
         }
     }
 
+    /**
+     * Returns whether depth writing is enabled for the sky.
+     *
+     * @type {boolean}
+     */
     get depthWrite() {
         return this._depthWrite;
     }
--- a/src/scene/skybox/sky.js
+++ b/src/scene/skybox/sky.js
@@ -38,6 +38,12 @@ class Sky {
     skyMesh = null;
 
     /**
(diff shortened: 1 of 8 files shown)


src/scene/skybox/sky.js
```diff
@@ -109,11 +115,37 @@ class Sky {
...
+    set depthWrite(value) {
+        if (this._depthWrite !== value) {
+            this._depthWrite = value;
+            if (this.skyMesh) {
+                this.skyMesh.depthWrite = value;
+            }
+        }
+    }
+
+    get depthWrite() {
```
willeastcott, line 144: Don't forget to document the getter as well! 😄

willeastcott (approved):

src/scene/skybox/sky.js
mvaligursky: done thanks

PR #7734, opened 2025-06-03
Gsplat reskr

slimbuck:
Fixes #


mvaligursky:
| GSplatResource used to support the function instanciate. This has been removed. We could still support this function if necessary.

If reasonably easy to do, I think we should have it as a deprecated function, to avoid broken projects.

slimbuck:
> If reasonably easy to do, I think we should have it as a deprecated function, to avoid broken projects.

It shouldn't be hard to add, we could probably add the function to `GSplatResourceBase`.

The downside being `GSplatResourceBase` would again need to keep a reference to app. Which I hate.

(In hindsight I wish the app was just passed into `instantiate()` function as a parameter instead of being added as a class member of resource).

src/scene/gsplat/gsplat-sogs-resource.js
```diff
@@ -0,0 +1,39 @@
+import { Vec2 } from '../../core/math/vec2.js';
+import { createGSplatMaterial } from './gsplat-material.js';
+import { GSplatResourceBase } from './gsplat-resource-base.js';
+
+class GSplatSogsResource extends GSplatResourceBase {
+    device;
+
+    destroy() {
+
+    }
```
willeastcott: What's the point of this?

slimbuck:
> So this PR doesn't address custom shader stuff?

Yeah materials will be addressed in a followup PR.

src/scene/gsplat/gsplat-sogs-resource.js
slimbuck: Thanks updated!

Issue #7404, opened 2025-03-05
Experiment with WebXR support using WebGPU


mvaligursky:
see more: https://toji.dev/2025/03/03/experimenting-with-webgpu-in-webxr.html

Maksims:
Here is a relevant explainer: https://github.com/immersive-web/WebXR-WebGPU-Binding/blob/main/explainer.md

mvaligursky:
Update: We now have WebGPU WebXR implementation. I tested all examples on Android and Quest 3, and they work the same way as WebGL version. Differences:
- Quest 3: on WebGL, the plane detection works, but it does not on WebGPU - likely the platform does not implement this yet. The mesh detection works on both.
- Quest 3: Performance when WebGPU is used is much slower than WebGL, especially when MSAA is enabled. Likely related to this being a very new experimental Quest 3 WebGPU implementation. WebGPU examples without XR enabled run fine in browser. I tested Metaballs demo mentioned [here](https://toji.dev/2025/03/03/experimenting-with-webgpu-in-webxr.html) and that suffers the same way.

mvaligursky:
## WebGPU + WebXR — status update (from pre-`XrBridge` baseline)

### Where we started

Before the recent refactor, **there was no `XrBridge`**. WebXR presentation and WebGL-specific details (**`XRWebGLLayer`**, **`XRWebGLBinding`**, default framebuffer handling, camera/depth texture paths, etc.) lived **directly in the general XR / framework layer** (same area as `XrManager` / views), tightly coupled to the WebGL device.

That made **WebGPU immersive XR** awkward: the framework would have had to sprout parallel WebGPU branches or duplicate logic next to every GL call.

### What changed architecturally

We introduced a small **graphics-layer bridge** — [`XrBridge`](https://github.com/playcanvas/engine/blob/main/src/platform/graphics/xr-bridge.js) — constructed with the **`GraphicsDevice`** and delegating to a **per-backend implementation** from `GraphicsDevice#createXrBridgeImpl`:

- **WebGL:** [`WebglXrBridge`](https://github.com/playcanvas/engine/blob/main/src/platform/graphics/webgl/webgl-xr-bridge.js) — owns `XRWebGLLayer` / `XRWebGLBinding`, default framebuffer, GL blits / texture aliasing for camera paths, etc.
[...]

PR #7747, opened 2025-06-09
WebGPU supports internal multi-view rendering used by WebXR

mvaligursky:
WebGPU supports internal multi-view rendering used by WebXR - this is a rendering preparation for WebXR support on WebGPU.

related to https://github.com/playcanvas/engine/issues/7404

Details:
- make `GraphicsDevice.draw` and `.setVerexBuffer` private - there is no point for these to be public and majority of required functionality to be used alongside is private.
- removed `GraphicsDevice.setIndexBuffer`. Index buffer is no longer stored as a state, but supplied to each draw call, to avoid complexities of left-over state.
- modified the way vertex / index buffers are managed for multi-view rendering. This simplifies and unifies the code, and avoids few issues we had to fix over the years related to left-over state with them.
- a hidden example to test multi-view rendering for both WebGL and WebGPU
- if instancing has 0 instances, we skip these draw calls early to avoid expensive set up of render state, and also avoid having to do a custom instancing cleanup.

Other changes:
- glb-instancing example has ground plane receiving shadows now, to test instanced rendering to shadow map

Note: Review the PR without white changes, part of the code is only indented.

Hidden test example:
Screenshot 2025-06-09 at 16 08 02


Maksims:
So this is still issuing multiple draw calls (one per view)?
There is a WebXR Multi-view thing, OVR_multiview2 and OCULUS_multiview, but this is not the one, right?

mvaligursky:
> So this is still issuing multiple draw calls (one per view)? There is a WebXR Multi-view thing, OVR_multiview2 and OCULUS_multiview, but this is not the one, right?

There is no functional modification to WebGL rendering (which those are for I assume), this implements WebGPU baseline equivalent using mutlple draw calls (WebGPU does not yet have custom API for multi-view drawing).

Issue #7733, opened 2025-06-03
Create 3D Gaussian Splatting section in the Examples Browser


willeastcott:
We probably want:

* `loaders/gsplat` > `gaussian-splatting\simple`
* `loaders/gsplat-many` > `gaussian-splatting\multi-splat`

New:

* `gaussian-splatting\sogs` ('lost place' environment with FPS controller)
* `gaussian-splatting\custom-shader` (statues transition effect)

willeastcott:
This now done:

<img width="430" height="360" alt="Image" src="https://github.com/user-attachments/assets/118eae85-134f-4153-809b-a6fb7a0f4589" />

PR #7749, opened 2025-06-10
[BREAKING] Update gsplat component material API

slimbuck:
Further to #7734, this PR updates the `gsplat` component material API.

We drop `materialOptions` property (which indirectly reconfigured the material) in favour of exposing the material instance directly. This matches the `render` and `model` components.

The `gsplat` component will create a default material if none is provided at construction, but this can then be overridden at will.

## Notes
- the internal function `GSplatInstance.configureMaterial` accepts a ShaderMaterial instance and configures it with the gsplat data. This function could be useful as public API, but we must still decide how to expose it.
- cloning a `gsplat` component doesn't clone the material. if it did, then the sort texture would be shared.


mvaligursky:
The PR should probably be maked as [BREAKING]?

mvaligursky:
Also please update the examples to work, I suspect some do not.

src/framework/components/gsplat/component.js
```diff
@@ -190,26 +183,24 @@ class GSplatComponent extends Component {
         return this._instance;
     }
 
-    set materialOptions(value) {
-        this._materialOptions = Object.assign({}, value);
-
-        // apply them on the instance if it exists
+    /**
```
mvaligursky, line 186: This needs descripion, considering it's a public function.

src/framework/components/gsplat/component.js
```diff
@@ -56,22 +55,22 @@ class GSplatComponent extends Component {
     _instance = null;
 
     /**
-     * @type {BoundingBox|null}
+     * @type {ShaderMaterial|null}
      * @private
      */
-    _customAabb = null;
+    _materialStore = null;
```
mvaligursky: why not call it `_material`?

src/scene/gsplat/gsplat-instance.js
```diff
@@ -82,17 +99,15 @@ class GSplatInstance {
...
-            this.sorter.init(this.orderTexture, centers, chunks);
-            this.sorter.on('updated', (count) => {
-                // limit splat render count to exclude those behind the camera
-                this.meshInstance.instancingCount = Math.ceil(count / resource.instanceSize);
-
-                // update splat count on the material
-                this.material.setParameter('numSplats', count);
-            });
-        }
+        this.sorter = new GSplatSorter();
```
mvaligursky, line 102: we should maybe have internal `dither` property set to false, and preserve this conditions, as it might be useful in the future?

src/framework/components/gsplat/component.js
slimbuck: This material is only used temporarily if user sets material before GSplatInstance is constructed.

I thought `_material` naming would be confusing given this temporary purpose. Perhaps `_materialTmp` is a better option?

examples/src/examples/loaders/gsplat-many.example.mjs
```diff
@@ -70,11 +70,12 @@ assetListLoader.load(() => {
     camera.setLocalPosition(-3, 1, 2);
 
     // instantiate guitar with a custom shader
-    const guitar = assets.guitar.resource.instantiate({
-        vertex: files['shader.vert']
+    const guitar = new pc.Entity('guitar');
+    guitar.addComponent('gsplat', {
+        asset: assets.guitar
```
mvaligursky, line 75: we need to preserve the custom vertex shader in this example

src/framework/components/gsplat/component.js
mvaligursky: sure. In that case, maybe when the material is passed to the instance, it's set to null here as well.

examples/src/examples/loaders/gsplat-many.example.mjs
willeastcott: I think we need a dedicated example for splat custom shaders.

willeastcott: See: https://github.com/playcanvas/engine/issues/7733

slimbuck: Yes! Followup PR.

PR #7755, opened 2025-06-11
Add get position and rotation properties for RigidBody

AlexAPPi:
When adding a fixed step, it is very useful to obtain the precise position and rotation of the Rigidbody to ensure accurate and predictable object behavior in the physics simulation.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "playcanvas",
-  "version": "2.8.0-dev.0",
+  "version": "2.9.0-dev.0",
   "lockfileVersion": 3,
   "requires": true,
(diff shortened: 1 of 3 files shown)


Maksims:
Shouldn't the Entity have the right position and rotation if it has a rigid body?

AlexAPPi:
> Shouldn't the Entity have the right position and rotation if it has a rigid body?

For an Entity, the position and rotation are typically represented as interpolated values of the actual positions multiplied by the respective velocities, which is useful for smooth animations and visualization. However, in some cases, it is necessary to obtain precise, non-interpolated values of position and linear velocity to ensure accurate physics calculations.

PR #7757, opened 2025-06-12
[Fix] Re-render all shadows when the device context is restored

mvaligursky:


slimbuck (approved):

src/core/event-handle.js
```diff
@@ -121,6 +121,11 @@ class EventHandle {
     get removed() {
         return this._removed;
     }
+
+    // don't stringify EventHandle to JSON by JSON.stringify
+    toJSON(key) {
+        return undefined;
+    }
```
willeastcott, line 128: Why was this causing a problem? Does this really belong in this PR?

mvaligursky: It does. As part of the shader options file, we serialize light properties .. and light now has the new event handler I added, and this excludes it from being serialized there, otherwise serialize was failing due to some circular reference.

mvaligursky: I added this to GraphicsDevice back then for the same reason - lots of device properties were serialized with the light.

PR #7779, opened 2025-06-18
Inputs overhaul (CameraControls updated)

kpal81xd:
### Input Sources
**SingleGestureSource** - Single touch or virtual joystick input 
**DualGestureSource** - Split screen touch or virtual joystick input
**MultiTouchSource** - Multi touch input (includes pinch handling)
**KeyboardMouseSource** - Keyboard and mouse input
**GamepadSource** - Game pad input

### Input Controllers
**FlyController** - Applies smoothed free-fly transformations to a `Pose`
**OrbitController** - Applies smoothed orbiting transformations to a `Pose`
**FocusController** - Applies smoothed focusing transformations to a `Pose`

### Pose
- `position`, `angles`, `distance` fields
- `lerp` to interpolate poses
- `look` method to construct from  position and focus target using `look`
- `pitchRange` and `yawRange` for angle constraints
- `xRange` `yRange` `zRange` for position constraints


Maksims:
This is not compiled as a part of the engine, and is optional by default, correct?

kpal81xd:
> This is not compiled as a part of the engine, and is optional by default, correct?

It'll be treeshaken out when using the unbundled build of the engine yes

Maksims:
> It'll be treeshaken out when using the unbundled build of the engine yes

What about the Editor with classic scripts?

kpal81xd:
> > It'll be treeshaken out when using the unbundled build of the engine yes
> 
> What about the Editor with classic scripts?

If its a UMD or bundled build all extras are bundled in currently. We are working on solutions to split out extras to reduce the overall engine size.

Maksims:
> If its a UMD or bundled build all extras are bundled in currently. We are working on solutions to split out extras to reduce the overall engine size.

As we rely on classical scripts, it would be great to keep the engine as lean as possible.
If large extras are added, it would be nice to highlight the additional Kbs and % of size difference with/without.

src/extras/input/sources/dual-gesture-source.js
```diff
@@ -0,0 +1,243 @@
...
+const startsWith = (str, prefix) => str.indexOf(prefix) === 0;
+
+/**
+ * @param {string} str - The string to check.
+ * @param {string} suffix - The suffix to check for.
+ * @returns {boolean} - True if the string ends with the suffix, false otherwise.
+ */
+const endsWith = (str, suffix) => str.indexOf(suffix, str.length - suffix.length) !== -1;
+
+/**
```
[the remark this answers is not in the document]
willeastcott: How come this was resolved? Curious as to why you don't want to use the built-ins here.

src/extras/input/sources/dual-gesture-source.js
```diff
@@ -0,0 +1,243 @@
...
+    _leftJoystick;
+
+    /**
+     * @type {VirtualJoystick}
+     * @private
+     */
+    _rightJoystick;
+
+    /**
+     * @param {`${'joystick' | 'touch'}-${'joystick' | 'touch'}`} [layout] - The layout of the dual
```
willeastcott, line 55: Won't this look a bit scary in the generated TypeDocs? Since there's only 4 options, maybe just list the literals?

kpal81xd: It expands them into literals automatically on typedoc:

image

willeastcott: Wow, ok, I didn't expect that!

PR #7782, opened 2025-06-19
Add support for texture async write

slimbuck:
This PR implements an internal function for asynchronous texture writes. It currently only supports WebGL and is flagged as internal.

## Notes:
- The WebGPU implementation will follow.
- We currently use a timer to poll the status of async reads and writes. This is hardly ideal. A better approach would be to test all in-flight transfers for completion at the end of every rendered frame. cc @mvaligursky. This is something we'll address in future.

### DIFF
--- a/src/platform/graphics/texture.js
+++ b/src/platform/graphics/texture.js
@@ -1173,6 +1173,22 @@ class Texture {
     read(x, y, width, height, options = {}) {
         return this.impl.read?.(x, y, width, height, options);
     }
+
+    /**
+     * Upload texture data asynchronously to the GPU.
+     * 
+     * @param {*} x - The left edge of the rectangle.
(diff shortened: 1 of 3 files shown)


src/platform/graphics/texture.js
```diff
@@ -1173,6 +1173,22 @@ class Texture {
     read(x, y, width, height, options = {}) {
         return this.impl.read?.(x, y, width, height, options);
     }
+
+    /**
+     * Upload texture data asynchronously to the GPU.
+     *
+     * @param {*} x - The left edge of the rectangle.
```
mvaligursky: specify types instead of `*`

mvaligursky (approved):

src/platform/graphics/texture.js
slimbuck: done thanks!

PR #7792, opened 2025-06-23
First Person Script & InputSource updates

kpal81xd:
- Updated the script to use the input sources instead of custom handlers
- Updated `KeyboardMouseSource` to use more keys and support pointer lock
- Updated `GamepadSource` to support buttons and triggers
- Updated static code fields in `KeyboardMouseSource` and `GamepadSource` to use camel case

### DIFF
--- a/src/extras/input/sources/gamepad-source.js
+++ b/src/extras/input/sources/gamepad-source.js
@@ -18,7 +18,7 @@ class GamepadSource extends InputSource {
      *
      * @readonly
      */
-    static buttoncode = {
+    static buttonCode = {
         A: 0,
         B: 1,
         X: 2,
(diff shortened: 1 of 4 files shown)


src/extras/input/sources/gamepad-source.js
```diff
@@ -7,45 +7,84 @@ import { InputSource } from '../input.js';
...
  * @property {number[]} rightStick - The right stick deltas, represented as an array of [x, y] coordinates.
  * @augments {InputSource<GamepadSourceDeltas>}
  */
 class GamepadSource extends InputSource {
+    /**
+     * The button codes (based on Xbox controller layout).
+     *
+     * @readonly
+     */
+    static buttoncode = {
```
willeastcott: Should this be `buttonCode` instead?

kpal81xd: Ahh yea ill update the keyboard one too

kpal81xd:
> Just curious...but have you actually tested this with a gamepad?

Yep I tested it on friday at home

PR #7796, opened 2025-06-25
Add resolve step for SOGS sphericla harm

slimbuck:
Fixes #


src/scene/gsplat/gsplat-resolve-sh.js
```diff
@@ -0,0 +1,349 @@
...
+import { QuadRender } from '../graphics/quad-render.js';
+
+const vertexGLSL = /* glsl */`
+    attribute vec2 vertex_position;
+    void main(void) {
+        gl_Position = vec4(vertex_position, 0.0, 1.0);
+    }
+`;
+
+const fragmentGLSL = /* glsl */`
```
mvaligursky, line 24: we should move these to chunks

src/scene/gsplat/gsplat-resolve-sh.js
```diff
@@ -0,0 +1,349 @@
...
+        #define SH_C3_1 2.890611442640554f
+        #define SH_C3_2 -0.4570457994644658f
+        #define SH_C3_3 0.3731763325901154f
+        #define SH_C3_4 -0.4570457994644658f
+        #define SH_C3_5 1.445305721320277f
+        #define SH_C3_6 -0.5900435899266435f
+    #endif
+
+    // see https://github.com/graphdeco-inria/gaussian-splatting/blob/main/utils/sh_utils.py
+    vec3 evalSH(in vec3 sh[SH_COEFFS], in vec3 dir) {
```
mvaligursky: this looks like a copy & paste from another chunk. Should we extract that function to a new chunk and include it in both places.

src/scene/gsplat/gsplat-resource-base.js
```diff
@@ -115,7 +115,13 @@ class GSplatResourceBase {
     configureMaterial(material) {
     }
 
+    /**
+     * @description Evaluates the size of the texture based on the number of splats.
```
willeastcott:
```suggestion
     * Evaluates the size of the texture based on the number of splats.
     *
```

src/scene/gsplat/gsplat-resolve-sh.js
slimbuck: Yes definitely. I decided while the approach is still in flux to keep the feature self contained for now.

src/scene/gsplat/gsplat-resolve-sh.js
slimbuck: yep

willeastcott (approved):

src/scene/shader-lib/glsl/chunks/gsplat/vert/gsplatEvalSH.js
```diff
@@ -0,0 +1,75 @@
...
+        #define SH_COEFFS 8
+    #elif SH_BANDS == 3
+        #define SH_COEFFS 15
+    #else
+        #define SH_COEFFS 0
+    #endif
+
+    #if SH_BANDS > 0
+
+    #define SH_C1 0.4886025119029199f
```
mvaligursky: should we just inline those in the code? No point to get our preprocessor do this for us (at a cost).

src/framework/components/gsplat/component.js
```diff
@@ -469,7 +500,7 @@ class GSplatComponent extends Component {
         // create new instance
         const asset = this._assetReference.asset;
         if (asset) {
-            this.instance = new GSplatInstance(asset.resource, this._materialTmp);
+            this.instance = new GSplatInstance(asset.resource, this._materialTmp, this._fullSH);
```
mvaligursky: I'd stick `this._fullSH` to `options` object, as this is optional and would be easier to add additional properties later

src/framework/components/gsplat/component.js
```diff
@@ -206,6 +212,31 @@ class GSplatComponent extends Component {
         return this._instance?.material ?? this._materialTmp ?? null;
     }
 
+    /**
+     * Sets whether the full spherical-harmonic solver is enabled when rendering SOGS data.
```
mvaligursky: I'd prefer additional info of what this means for the user. Mention quality vs performance, for sogs specifically.

src/scene/gsplat/gsplat-instance.js
```diff
@@ -212,10 +222,30 @@ class GSplatInstance {
...
         }
     }
+
+    setFullSH(value) {
+        const { resource } = this;
+        const { gsplatData } = resource;
+
+        if (gsplatData instanceof GSplatSogsData && gsplatData.shBands > 0) {
+
+            if (this.resolveSH) {
```
mvaligursky, line 242: instead of handling if the value is different in the component, I'd maybe always call this function in the instance, and handle that here. Otherwise it's strange to see that we delete resolveSH regardless of the value, as that assumes the component does not call this function unless the change is needed.

src/framework/components/gsplat/component.js
willeastcott: Yeah, this sounds like it's fully toggling calculation of SH. That's kind of why I suggested `hqsh` for high-quality spherical harmonics. It's more a quality thing right?

src/framework/components/gsplat/component.js
```diff
@@ -61,6 +61,12 @@ class GSplatComponent extends Component {
      */
     _materialTmp = null;
 
+    /**
+     * @type {boolean}
+     * @private
+     */
```
willeastcott:
```suggestion
    /** @private */
```

src/scene/shader-lib/glsl/chunks/gsplat/vert/gsplatEvalSH.js
slimbuck: we could do, but the code would be much more messy. this isn't particularly heavy for preprocessor considering everything else it does?

mvaligursky: maybe just change them to const float to avoid preprocessor completely
`const float SH_C1 = 0.4886025119029199;`

slimbuck: done, thanks. it's also now closer to the wgsl version.

src/framework/components/gsplat/component.js
slimbuck: done

src/framework/components/gsplat/component.js
slimbuck: `fullSH` means more to me than `hqSH` (and fullSH != enableSH), but if you both agree on that naming I'm happy to change it.

willeastcott: @mvaligursky can be tie-breaker.

mvaligursky: I'd probably go with `highQualitySH`. But explain this applies to SOGS only.

slimbuck: how about approxSH defaulting true instead?

mvaligursky: I'll take that.

slimbuck: do we have unanimous agreement @willeastcott ?

src/scene/gsplat/gsplat-instance.js
slimbuck: the component must store the state anyway (for the case where flag is set before instance exists), so I'll keep the logic there, but will also add test in this function so it looks less strange :)

src/framework/components/gsplat/component.js
willeastcott: Fine.

slimbuck: or `fastSH`. sounds positive.

willeastcott: Yeah, I prefer that.

mvaligursky: fastSH does not suggest there is a quality difference. Nobody would turn off faster? Why even have a slower option :)

slimbuck: dunno, we have had 'fastTBN' and 'fastNormalize' etc, which always take short cuts. think it's implied.

mvaligursky: maybe. Cannot say I like those either.
But ok.

willeastcott: Hmm, @mvaligursky makes a good point there...

slimbuck: you could say the same thing about `approxSH`. the name doesn't imply it's fast, so why would anyone use it?

slimbuck: `fastButApproximateSH`

willeastcott: So back to `highQualitySH`? 😏

mvaligursky: yeah
For DOF, I have `highQuality` setting as it seemed the best.

slimbuck: Well except in this case it _isn't_ "high quality" spherical harmonics. If someone asks what's high quality about it, we must say, well it's just the normal quality. The other one is an approximate version that runs fast.

slimbuck: (BTW for me it's an interesting exercise to try tease out what the best name could be in this case, but I'm not actually fussed either way. If you guys agree on a name I will make it so).

willeastcott: It's high quality compared to the low (approximate) quality. So I'm still cool with `highQualitySH`.

slimbuck: done

PR #7798, opened 2025-06-26
Beta builds (manual trigger)

kpal81xd:
### Motivation
The `release-X.XX` branches has support for prereleases however these are solely used for creating preview for bug patches. This PR adds support by creating a manual workflow to create a beta release from current new or breaking features on `main`.

### Overview
- Adds manual workflow to create a prerelease on main branch `X.X.X-beta.X` which **writes** the version bump to `main` (this includes the commit and the tag).
- Adds check for any recent changes before creating release (has force boolean input for override)
- Updates publish workflow to trigger on prereleases with the `beta` id to chain the prerelease workflow to the publish workflow
- Set a new npm publish tag `beta` to track builds
- Updated package json, lockfile and release script to use `beta` instead of `dev` 

### Update
- [x] Cherry pick to branch `main_v1`
- [x] Cherry pick to branch release-2.X
- [x] Cherry pick to branch release-1.7
- [ ] Editor support for `beta` prerelease type

### DIFF
--- a/.github/workflows/beta.yaml
+++ b/.github/workflows/beta.yaml
@@ -21,7 +21,7 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Set up Node.js 18.x
+      - name: Set up Node.js 22.x
         uses: actions/setup-node@v4
         with:
           node-version: 22.x
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v4
 
-    - name: Setup Node.js 18.x
+    - name: Setup Node.js 22.x
       uses: actions/setup-node@v4
       with:
         node-version: 22.x
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -15,7 +15,7 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v4
 
-      - name: Set up Node.js 18.x
+      - name: Set up Node.js 22.x
         uses: actions/setup-node@v4
         with:
           node-version: 22.x


.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,38 @@
+name: Nightly
+
+on:
+  workflow_dispatch:
```
marklundin: consider a daily cron here?

kpal81xd: Yea spoke to Will about it and he suggested we make it manual for better control. Ideally a cron would be the best but ill need to add an extra step for checking if there are any recent changes since this workflow will write a new commit message

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,38 @@
...
+          cache: 'npm'
+          registry-url: 'https://registry.npmjs.org/'
+
+      - name: Configure Git User
+        run: |
+          git config --global user.email "action@github.com"
+          git config --global user.name "GitHub Action"
+        shell: bash
+
+      - name: Bump version
```
marklundin: Could we add additional steps to diff check here? Just wary that multiple dispatches will keep bumping without changes.

eg **!!untested!!**

```
- name: Get last tag
  id: tag
  run: |
    echo "tag=$(git describe --tags --abbrev=0)" >> $GITHUB_OUTPUT

- name: Check for changes since last tag
  id: diffcheck
  run: |
    if git diff --quiet ${{ steps.tag.outputs.tag }}..HEAD; then
      echo "changed=false" >> $GITHUB_OUTPUT
    else
      echo "changed=true" >> $GITHUB_OUTPUT
    fi

- name: Bump version
   if: steps.diffcheck.outputs.changed == 'true'
```

.github/workflows/beta.yaml
```diff
@@ -0,0 +1,56 @@
...
+  beta:
+    runs-on: ubuntu-latest
+    if: github.repository == 'playcanvas/engine' && github.ref_name == 'main'
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Node.js 18.x
```
willeastcott:
```suggestion
      - name: Set up Node.js 22.x
```

PR #7803, opened 2025-06-30
Add asset load event

slimbuck:
This PR adds an event fire on PLY asset load.

The event is called `load:data` and sends the newly created `GSplatData` instance so the user can modify it before constructing the `GSplatInstance`.

This was needed so we can continue to support things like 2dgs in supersplat (https://github.com/playcanvas/supersplat/issues/528).

### DIFF
--- a/src/framework/parsers/ply.js
+++ b/src/framework/parsers/ply.js
@@ -609,6 +609,9 @@ class PlyParser {
                     }
                 );
 
+                // allow application to process the data
+                asset.fire('load:data', data);
+
                 // reorder data
                 if (!data.isCompressed) {
                     if (asset.data.reorder ?? true) {


src/framework/parsers/ply.js
```diff
@@ -609,6 +609,9 @@ class PlyParser {
                     }
                 );
 
+                // allow application to process the data
+                asset.fire('load:data', data);
```
mvaligursky, line 613: So I guess this is just under the hood undocumented call for internal use?

slimbuck: SuperSplat editor will use it immediately so I guess we should be documenting. It's kind of a niche event that I don't expect developers will need...

In general I have left these asset loading details out of documentation because it seems strange to add documentation on `Asset` class that only applies in such specific cases (gsplat asset uncompressed ply load in this case).

PR #7820, opened 2025-07-04
Add Options Object Support to `StandardMaterial#constructor`

kungfooman:
This PR aligns `StandardMaterial#constructor` with other classes like `Texture#constructor` to accept an optional options object, allowing properties (e.g., diffuse, depthTest) to be set during instantiation without being forced to call additional methods like `Material#update` - reducing boilerplate code.

Example:

```js
const entity = new pc.Entity("test");
entity.addComponent("render", {
    type: "box",
    material: new pc.StandardMaterial({diffuse: pc.Color.RED})
});
```

Basically, the aim is to keep the code short and readable and to reduce manual bookkeeping.

@mvaligursky I'm not entirely sure about the requirement for `this.update()` or if these property assignments should happen before e.g. `this.reset()`?

### DIFF
--- a/examples/src/examples/physics/compound-collision.example.mjs
+++ b/examples/src/examples/physics/compound-collision.example.mjs
@@ -62,8 +62,8 @@ app.on('destroy', () => {
 app.scene.ambientLight = new pc.Color(0.2, 0.2, 0.2);
 
 // Create a couple of materials for our objects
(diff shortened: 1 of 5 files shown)


mvaligursky:
Thanks, I'll check into this, not sure about .reset(), not familiar with it.
calling update should not be needed, no problem there, it's only needed after material has been used to render, not when constructing I believe. Again - I'll double check.

PR #7825, opened 2025-07-07
Initial implementation of unified gsplat rendering

mvaligursky:
- initial implementation of GSplatManager, managing all splats. It only implements a temporary API by passing splats to a constructor, that will be replaced.
- gsplat resorces can generate initial LOD at runtime. Only implemented for uncompressed splats, other splats used a single LOD (so no LOD for now). A better solution will be done.
- work buffer stores all active splats in the scene, allowing global sorting. Only splats used by active LOD as used, which adds support for LOD.
- some small refactoring of existing gsplat shader chunks, to allow their more flexible use. New shaders use those to implement copy of splats of various formats to the work buffer and also their rendering from the work buffer.
- simple hidden example to test the functionality and temporary API. Note that the main uncompressed version of the church.ply is excluded due to its size.
- works on WebGL and WebGPU

All this is subject to much improvements.

Screenshot 2025-07-07 at 15 13 20


src/scene/gsplat/gsplat-resource.js
```diff
@@ -286,6 +286,173 @@ class GSplatResource extends GSplatResourceBase {
...
+        // for (let chunkIdx = 0; chunkIdx < numBlocks; chunkIdx++) {
+        //     const base = chunkIdx * 3;
+        //     totalLevel0 += this.lodBlocks.blocksLodInfo[base];
+        //     totalLevel1 += this.lodBlocks.blocksLodInfo[base + 1];
+        //     totalLevel2 += this.lodBlocks.blocksLodInfo[base + 2];
+        // }
+        // const pct0 = ((totalLevel0 / numSplats) * 100).toFixed(1);
+        // const pct1 = ((totalLevel1 / numSplats) * 100).toFixed(1);
+        // const pct2 = ((totalLevel2 / numSplats) * 100).toFixed(1);
+        // console.log(`LOD Distribution: NumBlocks: ${numBlocks}, Level 0 (large): ${pct0}% (${totalLevel0}), Level 1 (medium): ${pct1}% (${totalLevel1}), Level 2 (small): ${pct2}% (${totalLevel2})`);
```
willeastcott, line 421: Do we want to merge this?

mvaligursky: Yes for now, useful logging for me while developing this.
This will be replaced by offline generated LOD, and at that point we'll strip all this out.

src/scene/gsplat/unified/gsplat-unified-sort-worker.js
```diff
@@ -0,0 +1,384 @@
...
+        //     const y = (i & 2 ? boundMin.y : boundMax.y);
+        //     const z = (i & 4 ? boundMin.z : boundMax.z);
+        //     const d = x * dx + y * dy + z * dz;
+        //     if (i === 0) {
+        //         minDist = maxDist = d;
+        //     } else {
+        //         minDist = Math.min(minDist, d);
+        //         maxDist = Math.max(maxDist, d);
+        //     }
+        // }
```
willeastcott, line 97: Again, a heck of a lot of commented out code in this file. Do we want to retain it?

src/scene/gsplat/unified/gsplat-work-buffer.js
```diff
@@ -0,0 +1,217 @@
...
+        this.renderTarget = new RenderTarget({
+            name: 'GsplatWorkBuffer-MRT',
+            colorBuffers: [this.colorTexture, this.centerTexture, this.covATexture, this.covBTexture],
+            depth: false,
+            flipY: true
+        });
+    }
+
+    /**
+     * Estimates the minimal texture size width that can store all splats, using a fixed max texture
```
willeastcott: Did you mean to have `size width` here? Feels like you just need one or the other.

Maksims:
This is a huge step forward for GSplats! 🙏
Are LoDs configurable?

src/scene/shader-lib/glsl/chunks/gsplat/frag/gsplat-copy-to-workbuffer.js
```diff
@@ -0,0 +1,92 @@
+// fragment shader to copy splats in any supported format to MRT work-buffer
```
willeastcott: We have that ugly camel case naming for chunks. Shouldn't we adhere to that?

mvaligursky:
There is a lot of code that is currently commented out, but will be refactored to a useable form in the near future. I'd prefer not to remove it in order to simplify my life.

willeastcott (approved):
Approving with some comments, but I would suggest that @slimbuck's review is required before merging.

mvaligursky:
> This is a huge step forward for GSplats! 🙏 Are LoDs configurable?

There will be public API with a way to configure this in the future. At this stage it's just a tech drop without any public API. Still very much WIP.

src/scene/shader-lib/glsl/chunks/gsplat/frag/gsplat-copy-to-workbuffer.js
mvaligursky: good point, renamed those chunks, thanks

nancywan1004:
Hi @mvaligursky , thanks for introducing the initial LoD implentations! I got to look into your changes and wonder if the chunks(blocks) are divided based on space or memory locations of the splats? Looks like the current implementation is a linear block division in memory (the latter). 

And if that's the case, how would it make sense to calculate the block centers used for the LoD's as the splats in the block might be very spread out in space? Or am I missing out on anything? Would appreciate any clarifications!

Btw, I wonder if an octree-based implementation would make more sense for chunking and LoD's?

mvaligursky:
Hi @nancywan1004 - thanks for taking an interest!
Currently the blocks are directly created from chunks, but chunks use using Morton order, so they are spatially sorted to start with - the LOD works pretty well from my tests. But this is just an initial temporary implementation that allows me to work on the runtime side. We're gonna be adding better offline LOD generation soon, and replacing this generation at load time.

nancywan1004:
> Hi @nancywan1004 - thanks for taking an interest! Currently the blocks are directly created from chunks, but chunks use using Morton order, so they are spatially sorted to start with - the LOD works pretty well from my tests. But this is just an initial temporary implementation that allows me to work on the runtime side. We're gonna be adding better offline LOD generation soon, and replacing this generation at load time.

I see thanks! On that note, is the chunks'(block) generation process also going to change quite a bit? Or this part would stay the same, using the morton sorted chunks for now? The context for me is that I might have to port a [hierarchical gsplat format(octree-based)](https://github.com/graphdeco-inria/hierarchical-3d-gaussians) to the playcanvas viewer. I'm seeking for some good pointers to achieve it, hopefully without too many ground-breaking changes.

mvaligursky:
To be honest, it will most likely change a lot - we're still discussing the best approach here. If you could wait few more weeks, this will get more stable, but no promises till then.

src/scene/gsplat/gsplat-resource.js
```diff
@@ -286,6 +286,195 @@ class GSplatResource extends GSplatResourceBase {
...
+        for (let i = 1; i < numLevels; i++) {
+            const t = i / numLevels; // interpolation factor for all levels
+            const threshold = largeSize - t * (largeSize - smallSize);
+            activeThresholds.push(threshold);
+        }
+
+        const p = new Vec3();
+        const r = new Quat();
+        const s = new Vec3();
+        const iter = gsplatData.createIter(p, r, s);
```
slimbuck, line 318: Specify null for p and r so the iterator doesn't extract the bits you don't use

mvaligursky: good to know, but this code will be removed as soon as we have a new format.

src/scene/gsplat/unified/gsplat-info.js
```diff
@@ -0,0 +1,186 @@
...
+    resource;
+
+    /** @type {number} */
+    numSplats;
+
+    /** @type {GraphNode} */
+    node;
+
+    /**
+     * A state of the splat currently used for rendering. This matches the work buffer.
```
willeastcott, line 36: Should this be "The state"?

src/scene/gsplat/unified/gsplat-unified-sort-worker.js
```diff
@@ -0,0 +1,398 @@
...
+            //             if (z < boundMin.z) boundMin.z = z; else if (z > boundMax.z) boundMax.z = z;
+            //         }
+
+            //         // calculate chunk center and radius from bound min/max
+            //         chunks[c * 4 + 0] = (mx + Mx) * 0.5;
+            //         chunks[c * 4 + 1] = (my + My) * 0.5;
+            //         chunks[c * 4 + 2] = (mz + Mz) * 0.5;
+            //         chunks[c * 4 + 3] = Math.sqrt((Mx - mx) ** 2 + (My - my) ** 2 + (Mz - mz) ** 2) * 0.5;
+            //     }
+            // }
```
willeastcott, line 381: Heck of a lot of commented out code in this file!

src/scene/gsplat/unified/gsplat-manager.js
```diff
@@ -0,0 +1,355 @@
...
+const invModelMat = new Mat4();
+const tempSplats = [];
+
+/**
+ * GSplatManager manages the rendering of splats using a work buffer, where all active splats are
+ * stored and rendered from.
+ *
+ * @ignore
+ */
+class GSplatManager {
```
willeastcott, line 27: Presumably this will be moved to `GSplatComponentSystem` at some point?

src/scene/gsplat/unified/gsplat-info.js
mvaligursky: yep, fixing in follow up PR

src/scene/gsplat/unified/gsplat-unified-sort-worker.js
mvaligursky: It is for now. Not ready to be resurected yet, but will do in the near future.

src/scene/gsplat/unified/gsplat-manager.js
mvaligursky: in the follow up PR I have GSplatDirector which manages GSplatManager instances (per camera / layer) as needed.
GSplatManager stays as is.

PR #7829, opened 2025-07-10
First Person script updates

kpal81xd:
- Added support for jumping on mobile (double jump)
- Updated `DualGestureSource` and `SingleGuestureSource` to have an additional delta `doubleTap`
- Added in UI for joystick 
- Reorganised example script


scripts/esm/first-person-controller.mjs
```diff
@@ -225,6 +227,16 @@ class FirstPersonController extends Script {
      */
     jumpForce = 600;
 
+    /**
+     * The joystick event name for the UI position for the base and stick elements.
+     * The event name is appended with the side: 'left' or 'right'.
```
willeastcott:
```suggestion
     * The event name is appended with the side: ':left' or ':right'.
```

PR #7830, opened 2025-07-10
Engine & Examples build fixes

kpal81xd:
### Engine
- Renamed `buildTarget` to `buildJSOptions`
- Added `buildTypeOption`
- Fixed type imports to use `@import`

### Examples
- Moved all engine rollup deps into one section for examples
- Build types directly from examples instead of via NPM
- Refactored rollup mjs to split up target categories
- Fixed warning for imports from `https://**`


willeastcott:
We haven't updated the NPM dependencies for the examples in a while. Might wanna do that too.

kpal81xd:
> We haven't updated the NPM dependencies for the examples in a while. Might wanna do that too.

Yep ill do it in a separate PR

Issue #7828, opened 2025-07-10
Control mesh shadow in individual cascade.


AlexAPPi:
Currently, the engine provides the ability to control the updating of cascade shadows, but there is no control over the rendering of models for specific cascades. It is proposed to add this feature so that it would be possible to specify whether to render the shadow of a model for each individual cascade.

https://github.com/playcanvas/engine/blob/7b4f9c0c08a3a22e9c9197a300ab75ca613ed24f/src/scene/renderer/shadow-renderer-directional.js#L161

MeshInstance add property:

```
const CASCADE_0 = 1 << 0;
const CASCADE_1 = 1 << 1;
const CASCADE_2 = 1 << 2;
const CASCADE_3 = 1 << 3;
const CASCADE_ALL = 1 << 16;

class MeshInstance  {
...
cascadeShadow = CASCADE_ALL;
...
```

```
class ShadowRendererDirectional {
    ...
    cull(light, comp, camera, casters = null) {
        ...
        this.shadowRenderer.cullShadowCasters(comp, light, lightRenderData.visibleCasters, shadowCam, casters);

        lightRenderData.visibleCasters = lightRenderData.visibleCasters.filter(x => (x.cascadeShadow & (1 << cascade)) !== 0);
```

PR #7837, opened 2025-07-15
Add control mesh shadow in individual cascade.

AlexAPPi:
Solution: https://github.com/playcanvas/engine/issues/7828

Demo: https://launch.playcanvas.com/2281127?debug=true&use_local_engine=http://localhost:51000/playcanvas.js

This pull request introduces the ability to control mesh shadows individually for each cascade, enabling more effective rendering optimization.

**Key Benefits:**
- Allows independent shadow configuration for each cascade, providing greater flexibility in balancing visual quality and performance.
- Shadow settings can be defined separately for each layer of the mesh's level of detail (LOD), allowing precise control over shadow rendering depending on the mesh's importance in the scene.
- This enables developers to fine-tune shadow behavior, reducing unnecessary computations in areas where shadows have minimal visual impact.

**Practical Use:**
- Shadows can be disabled in specific cascades or LOD levels where they are not needed, reducing the system load.
- For key areas in the scene, high-quality shadows can be retained, while less important regions can have shadows minimized or turned off entirely.

In summary, this feature provides flexible and efficient shadow management in graphical scenes, supporting project optimization without sacrificing visual fidelity.

### DIFF
--- a/src/scene/constants.js
+++ b/src/scene/constants.js
@@ -409,6 +409,48 @@ export const shadowTypeInfo = new Map([
 ]);
 
 /**
+ * The flag that controls shadow rendering for the 0 cascade
+ * 
+ * @category Graphics
+ */
+export const SHADOW_CASCADE_0 = 1;
+
+/**
+ * The flag that controls shadow rendering for the 1 cascade
+ * 
+ * @category Graphics
+ */
+export const SHADOW_CASCADE_1 = 2;
+
+/**
+ * The flag that controls shadow rendering for the 2 cascade
+ * 
+ * @category Graphics
+ */
+export const SHADOW_CASCADE_2 = 4;
+
+/**
+ * The flag that controls shadow rendering for the 3 cascade
+ * 
+ * @category Graphics
+ */
+export const SHADOW_CASCADE_3 = 8;
+
+/**
+ * The flag that controls shadow rendering for the 4 cascade
+ * 
+ * @category Graphics
+ */
+export const SHADOW_CASCADE_4 = 16;
+
+/**
+ * The flag that controls shadow rendering for the all cascades
+ * 
+ * @category Graphics
+ */
+export const SHADOW_CASCADE_ALL = 32767;
+
+/**
  * Box filter.
  *
  * @category Graphics
--- a/src/scene/mesh-instance.js
+++ b/src/scene/mesh-instance.js
@@ -9,7 +9,8 @@ import {
     RENDERSTYLE_SOLID,
     SHADERDEF_UV0, SHADERDEF_UV1, SHADERDEF_VCOLOR, SHADERDEF_TANGENTS, SHADERDEF_NOSHADOW, SHADERDEF_SKIN,
     SHADERDEF_SCREENSPACE, SHADERDEF_MORPH_POSITION, SHADERDEF_MORPH_NORMAL, SHADERDEF_BATCH,
-    SHADERDEF_LM, SHADERDEF_DIRLM, SHADERDEF_LMAMBIENT, SHADERDEF_INSTANCING, SHADERDEF_MORPH_TEXTURE_BASED_INT
+    SHADERDEF_LM, SHADERDEF_DIRLM, SHADERDEF_LMAMBIENT, SHADERDEF_INSTANCING, SHADERDEF_MORPH_TEXTURE_BASED_INT,
+    SHADOW_CASCADE_ALL
 } from './constants.js';
 import { GraphNode } from './graph-node.js';
 import { getDefaultMaterial } from './materials/default-material.js';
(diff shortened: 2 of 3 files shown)


src/scene/constants.js
```diff
@@ -408,6 +408,48 @@ export const shadowTypeInfo = new Map([
...
+ * @category Graphics
+ */
+export const SHADOW_CASCADE_3 = 8;
+
+/**
+ * The flag that controls shadow rendering for the 4 cascade
+ *
+ * @category Graphics
+ */
+export const SHADOW_CASCADE_4 = 16;
```
mvaligursky: we only support 4 cascades (0-3), so do not need this one.

src/scene/constants.js
```diff
@@ -408,6 +408,48 @@ export const shadowTypeInfo = new Map([
...
+
+/**
+ * The flag that controls shadow rendering for the all cascades
+ *
+ * @category Graphics
+ */
+export const SHADOW_CASCADE_ALL = 32767;
```
mvaligursky: I'd do 0xff perhaps

src/scene/mesh-instance.js
```diff
@@ -260,6 +261,17 @@ class MeshInstance {
      */
     castShadow = false;
 
+    /**
+     * Mask describing which enable shadow casting in the cascade {@link ShadowRendererDirectional} for this mesh instance.
```
mvaligursky: - I would simplify this a bit, no need for too much detail. Maybe see `MeshInstance.visible` 
- do not link to ShadowRendererDirectional, this is a private implementation user knows nothing about. Mention `LIGHTTYPE_DIRECTIONAL` instead
- mention the default value (Defaults to ..), see other properties in the engine

src/scene/constants.js
AlexAPPi: Removed

src/scene/constants.js
AlexAPPi: Done. I agree, 1 byte is quite enough.

src/scene/renderer/shadow-renderer-directional.js
```diff
@@ -160,17 +161,29 @@ class ShadowRendererDirectional {
...
-                for (let i = 1; i < numVisibleCasters; i++) {
-                    visibleSceneAabb.add(visibleCasters[i].aabb);
+            // exclude all mesh instances that are hidden for this cascade.
+            // find out AABB of visible shadow casters
+
+            for (let i = 0; i < origNumVisibleCasters; i++) {
+                const meshInstance = visibleCasters[i];
+                if (meshInstance.shadowCascadeFlags & cascadeFlag) {
+                    visibleCasters[numVisibleCasters++] = visibleCasters[i];
+                    if (numVisibleCasters === 0) {
```
mvaligursky: will it ever be 0 here? there is ++ on the line before.
I'd move `numVisibleCasters++` later to make it clear.

AlexAPPi: Yes, 0 will never be here.

src/scene/constants.js
```diff
@@ -408,6 +408,41 @@ export const shadowTypeInfo = new Map([
...
+ * @category Graphics
+ */
+export const SHADOW_CASCADE_3 = 8;
+
+/**
+ * The flag that controls shadow rendering for the all cascades
+ *
+ * @category Graphics
+ */
+export const SHADOW_CASCADE_ALL = 255;
```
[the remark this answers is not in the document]
AlexAPPi: It's better to specify the value as 1 byte, as this will allow us to add additional layers in the future if needed, without changing the data structure. This approach provides scalability and leaves room for potential feature expansion.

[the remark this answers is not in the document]
mvaligursky: agreed, ignore the AI here.

src/scene/mesh-instance.js
```diff
@@ -260,6 +261,16 @@ class MeshInstance {
...
 
+    /**
+     * Mask that enables or disables shadow casting for each shadow cascade
+     * used by a mesh instance under a {@link LIGHTTYPE_DIRECTIONAL}.
+     * This property takes effect only when the {@link castShadow} is enabled.
+     * Defaults to {@link SHADOW_CASCADE_ALL}.
+     *
+     * @type {number}
+     */
+    shadowCascadeFlags = SHADOW_CASCADE_ALL;
```
mvaligursky: I'd rename it to `shadowCascadeMask`.
And possible text:
```
Specifies a bitmask that controls which shadow cascades a mesh instance contributes to when rendered with a {@link LIGHTTYPE_DIRECTIONAL} light source. This setting is only effective if the {@link castShadow} property is enabled. Defaults to {@link SHADOW_CASCADE_ALL}, which means the mesh casts shadows into all available cascades.
```

AlexAPPi: Corrected, made changes.

src/scene/renderer/shadow-renderer-directional.js
```diff
@@ -160,14 +162,29 @@ class ShadowRendererDirectional {
...
-            const numVisibleCasters = visibleCasters.length;
+            const origNumVisibleCasters = visibleCasters.length;
+
+            let numVisibleCasters = 0;
+
+            // exclude all mesh instances that are hidden for this cascade.
+            // find out AABB of visible shadow casters
+
+            if (origNumVisibleCasters > 0) {
+                visibleSceneAabb.copy(emptySceneAabb);
```
mvaligursky: this is not a good solution, as that will cause all aabb's to include point 0,0,0.
Lets go back to what it was.

Issue #7824, opened 2025-07-07
Error thrown when using GLTF file with TANGENT and EXT_gpu_mesh_instancing


MAG-AdrianMeredith:
When attempting to load a gltf file with instancing the following error is thrown if it also includes TANGENT accessor.  The attached file will reproduce in the model viewer

WebGL2:
```
task-list.ts:455 Failed to link shader program. Error: Attribute 'vertex_tangent' aliases attribute 'instance_line2' at location 13
```
WebGPU:
```
Error while parsing WGSL: :32:6 error: '@location(13)' appears multiple times
    @location(13) vertex_tangent: vec4f,
     ^^^^^^^^


 - While calling [Device].CreateShaderModule([ShaderModuleDescriptor]).
```

[instancing.glb.zip](https://github.com/user-attachments/files/21103173/instancing.glb.zip)

PR #7849, opened 2025-07-22
[Fix] Change the default semantic to a different slot which does not map to tangents

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/7824

### DIFF
--- a/src/platform/graphics/vertex-format.js
+++ b/src/platform/graphics/vertex-format.js
@@ -3,7 +3,7 @@ import { hashCode } from '../../core/hash.js';
 import { math } from '../../core/math/math.js';
 import { StringIds } from '../../core/string-ids.js';
 import {
-    SEMANTIC_TEXCOORD0, SEMANTIC_TEXCOORD1, SEMANTIC_ATTR12, SEMANTIC_ATTR13, SEMANTIC_ATTR14, SEMANTIC_ATTR15,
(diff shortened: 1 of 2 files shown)


willeastcott (approved):

MAG-AdrianMeredith:
Thanks Martin!  (this came up today again, been playing with rendering the amazon lumberyard bistro)

LeXXik:
If we use some of the slots for instancing, is there a way to tell which are used by the engine and which are free?

mvaligursky:
> If we use some of the slots for instancing, is there a way to tell which are used by the engine and which are free?

These are the default attributes engine uses for standard material (not all of them all the time, depends on material / mesh properties): https://github.com/playcanvas/engine/blob/2cc809eb27796fcc86adffe4ddab003d9283a0c7/src/scene/shader-lib/programs/lit-shader.js#L28-L36

and you can see the fixed mapping of semantic names to slots here:
https://github.com/playcanvas/engine/blob/2cc809eb27796fcc86adffe4ddab003d9283a0c7/src/platform/graphics/constants.js#L2148-L2178

You could enable tracing for shader creation as well, and each created shader attributes it uses here: https://github.com/playcanvas/engine/blob/2cc809eb27796fcc86adffe4ddab003d9283a0c7/src/platform/graphics/shader.js#L51

LeXXik:
Great info, thanks!

Maksims:
This change broke our instancing code in various places, as people likely will be defining custom instancing buffers, and have to update their mapping for semantics.
If such changes can be avoided in the future - will be great.

Issue #7826, opened 2025-07-09
RenderPass gets reset to default when changing camera layers


MAG-AdrianMeredith:
When the layers of a camera are changed, the settings on the cameraFrameRenderPass get reset to default, removing tonemapping etc.

As a result of this PR https://github.com/playcanvas/engine/pull/7719 it calls update on CameraFrameRenderPass, however if you look at CameraFrame.update() it also calls this, but then also reapplies a bunch of settings from the "rendering" prop which is reset to empty when calling the renderpass update function.  Did you mean to call CameraFrame.update()?

https://github.com/user-attachments/assets/8efbebdc-e2a4-4f73-aeb9-c0a1a0728d56

mvaligursky:
I'll investigate, thanks!

mvaligursky:
Done just that, thanks @MAG-AdrianMeredith !

PR #7850, opened 2025-07-23
[Fix] CameraFrame preserves settings when handling layer being added or removed

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/7826

- calls update on the owner class to set up all properties

### DIFF
--- a/src/extras/render-passes/render-pass-camera-frame.js
+++ b/src/extras/render-passes/render-pass-camera-frame.js
@@ -18,6 +18,10 @@ import { RenderPassDownsample } from './render-pass-downsample.js';
 import { Color } from '../../core/math/color.js';
 
 /**
+ * @import { CameraFrame } from './camera-frame.js'
+ */
+
+/**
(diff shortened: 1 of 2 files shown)


src/extras/render-passes/render-pass-camera-frame.js
```diff
@@ -101,17 +105,25 @@ class RenderPassCameraFrame extends RenderPass {
...
+    cameraFrame;
+
     /**
      * @type {RenderTarget|null}
      * @private
      */
     rt = null;
 
-    constructor(app, cameraComponent, options = {}) {
+    constructor(cameraFrame, app, cameraComponent, options = {}) {
```
willeastcott: When a constructor takes the app, don't we normally have that argument first in the args list?

mvaligursky: done, thanks

MAG-AdrianMeredith:
noice

PR #7861, opened 2025-07-25
Unified gsplat architecture is available to GSplatComponent

mvaligursky:
A public way to utilize unified gsplat architecture, using additiona `unified` component property:

```
entity.addComponent('gsplat', {
    asset: assets.logo,
    unified: true
});

```

Note that this property can be changed only when the splat is disabled (it's internal data not created) at the moment, and ignored with a warning otherwise.

Details:
- internal GSplatDirector class, managing instancing of GSplatManager for cameras / layers that have the gsplat component
- Layers stores array of GSplatPlacement (equivanent to MeshInstance) - this is all private, used by GSplatDirector to manage what it needs to render
- added an example to show / test rendering from multiple cameras.

### DIFF
--- a/src/scene/shader-lib/glsl/chunks/gsplat/frag/gsplatCopyToWorkbuffer.js
+++ b/src/scene/shader-lib/glsl/chunks/gsplat/frag/gsplatCopyToWorkbuffer.js
@@ -49,7 +49,7 @@ void main(void) {
         
         // source texture size
         #ifdef GSPLAT_SOGS_DATA
-            uint srcSize = uint(textureSize(scales, 0).x);
+            uint srcSize = uint(textureSize(packedTexture, 0).x);
         #elif GSPLAT_COMPRESSED_DATA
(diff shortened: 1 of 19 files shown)


src/scene/shader-lib/glsl/chunks/gsplat/frag/gsplatCopyToWorkbuffer.js
```diff
@@ -49,7 +49,7 @@ void main(void) {
         
         // source texture size
         #ifdef GSPLAT_SOGS_DATA
-            uint srcSize = uint(textureSize(scales, 0).x);
+            uint srcSize = uint(textureSize(packedTexture, 0).x);
```
slimbuck: remove ifdef?

mvaligursky: added two together using ||

slimbuck (approved):

PR #7869, opened 2025-07-30
Android sogs fix

slimbuck:
Android doesn't load SOGS correctly due to reorder shader not flagging inputs as highp.

### DIFF
--- a/src/scene/shader-lib/glsl/chunks/gsplat/frag/gsplat-sogs-reorder.js
+++ b/src/scene/shader-lib/glsl/chunks/gsplat/frag/gsplat-sogs-reorder.js
@@ -1,9 +1,9 @@
 export default /* glsl */`
-    uniform sampler2D means_l;
-    uniform sampler2D means_u;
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

src/scene/shader-lib/glsl/chunks/gsplat/frag/gsplat-sogs-reorder.js
```diff
@@ -1,9 +1,9 @@
...
-    uniform sampler2D means_l;
-    uniform sampler2D means_u;
-    uniform sampler2D quats;
-    uniform sampler2D scales;
-    uniform sampler2D sh_labels;
+    uniform highp sampler2D means_l;
+    uniform highp sampler2D means_u;
+    uniform highp sampler2D quats;
+    uniform highp sampler2D scales;
+    uniform highp sampler2D sh_labels;
```
willeastcott, line 6: Do all of these need to be highp? Could some be mediump?

PR #7871, opened 2025-07-30
Optimized Mat3 and Mat4 copy and set operations

mvaligursky:
- use TypedArray.set instead of manual copying of elements

### DIFF
--- a/examples/src/examples/graphics/instancing-basic.example.mjs
+++ b/examples/src/examples/graphics/instancing-basic.example.mjs
@@ -106,7 +106,8 @@ assetListLoader.load(() => {
         matrix.setTRS(pos, rot, scl);
 
         // copy matrix elements into array of floats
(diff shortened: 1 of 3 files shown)


slimbuck:
Did you see any difference when profiling?

mvaligursky:
> Did you see any difference when profiling?

Cannot say I profiled it :)

willeastcott:
Be careful - I've tried this in the past and slowed things down according to profiling. Maybe JS engines have moved on....but do careful profiling.

mvaligursky:
> Be careful - I've tried this in the past and slowed things down according to profiling. Maybe JS engines have moved on....but do careful profiling.

ah.

mvaligursky:
Got AI to spin up a test, and it is slower!
Will be reverting this.

Screenshot 2025-07-30 at 16 41 31

mvaligursky:
Reverted this change.

PR #7873, opened 2025-07-31
Export UniformBuffer

slimbuck:
The examples should probably use engine's exported interface to catch these missing exports.


mvaligursky:
Discussed, not needed.

PR #7878, opened 2025-08-01
Expose TextureHandler#parsers as public API with addParser/removeParser methods

### LINKED ISSUE #7564: Should texture parsers be public API?
Currently, we have the following classes exposed in the API reference:

* `TextureParser`
* `BasisParser`
* `DdsParser`
* `HdrParser`
* `ImgParser`
* `KtxParser`
* `Ktx2Parser`

Instances of these classes are created in the `TextureHandler` constructor and saved to `TextureHandler#parsers` but this object is not public API. So the question is:

Should we exposed `TextureHandler#parsers` as public API?
Or should we hide the various parsers from the public API?

killerdevildog:
## Expose TextureHandler#parsers as public API with addParser/removeParser methods

- Add comprehensive JSDoc documentation for parsers property
- Implement addParser() method following ModelHandler pattern for extensibility
- Implement removeParser() method to enable tree-shaking of unused parsers
- Add private _customParsers array to store custom parsers with decider functions
- Update _getParser() method to check custom parsers before default parsers
- Update maxRetries setter to handle both default and custom parsers
- Maintained 100% backward compatibility with existing functionality
- Enable configuration, tree-shaking, and better developer experience

This resolves the architectural inconsistency between TextureHandler and ModelHandler by applying the proven addParser pattern to texture handling. Custom parsers are checked first, allowing developers to override default behavior or add support for new formats like 16-bit PNGs or headerless textures.

Rebased onto latest main (Feb 2026).

Fixes #7564

I confirm I have read the [contributing guidelines](https://github.com/playcanvas/engine/blob/master/.github/CONTRIBUTING.md).

### DIFF
--- a/src/framework/handlers/texture.js
+++ b/src/framework/handlers/texture.js
@@ -20,6 +20,15 @@ import { ResourceHandler } from './handler.js';
  * @import { AppBase } from '../app-base.js'
  */
 
+/**
+ * @callback AddParserCallback
+ * Callback used by {@link TextureHandler#addParser} to decide on which parser to use.
+ * @param {string} url - The resource url.
+ * @param {object} data - The raw texture data.
+ * @returns {boolean} Return true if this parser should be used to parse the data into a
+ * {@link Texture}.
+ */
+
 const JSON_ADDRESS_MODE = {
     'repeat': ADDRESS_REPEAT,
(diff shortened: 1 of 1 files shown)


mvaligursky:
Superseded by #9001, which implemented this as a generic ResourceHandler parser registry and converted TextureHandler to use it.

PR #7881, opened 2025-08-01
Generate intervals texture for unified splat copying on GPU

mvaligursky:
-  gpu can do it much faster

### DIFF
--- a/src/scene/gsplat/unified/gsplat-unified-sort-worker.js
+++ b/src/scene/gsplat/unified/gsplat-unified-sort-worker.js
@@ -121,8 +121,11 @@ function UnifiedSortWorker() {
         //     }
         // }
 
-        const minDist = -1000;
-        const maxDist = 1000;
+        // const minDist = -1000;
+        // const maxDist = 1000;
+        const minDist = -1872;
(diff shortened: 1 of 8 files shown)


src/scene/gsplat/unified/gsplat-unified-sort-worker.js
```diff
@@ -121,8 +121,11 @@ function UnifiedSortWorker() {
         //     }
         // }
 
-        const minDist = -1000;
-        const maxDist = 1000;
+        // const minDist = -1000;
+        // const maxDist = 1000;
+        const minDist = -1872;
+        const maxDist = 1891;
```
willeastcott, line 127: Can numbers get any more magic? 🧙‍♂️

willeastcott (approved):

src/scene/gsplat/unified/gsplat-unified-sort-worker.js
mvaligursky: hah just copied them from non-unified sorter for a scene I was testing. This will go away with octree and stuff.

PR #7886, opened 2025-08-05
[Fix] Reordered bindings when using a StorageBuffer 

samw-improbable:
## Description
I reached a weird bug where the Preprocessor generated code output and generated bind layout were misaligned.

(in `WebgpuShaderProcessorWGSL.processResources`) It seems that the constructor for `BindGroupFormat` assigns 'slots' based on the order of the `formats` arg, and then sorts the bindings into various groups (in my case, textures and buffers were sorted) - and then later `getTextureShaderDeclaration` ([here](https://github.com/playcanvas/engine/blob/371394e19d11dd667e397b217ea60c5639cd89c7/src/platform/graphics/webgpu/webgpu-shader-processor-wgsl.js#L610)) assigns ids to first textures, then storage buffers (which is not the same order as `formats` was).

This ended up with my storage buffer being the first binding in the bind layout (slot 0), but the last one in the actual shader source code (slot 5). I would share a PoC but I haven't got time to write a repro outside of my codebase right now, though if you need one I may be able to write one in the future.

Fixes #

I wrote a naive fix (which makes my code work), but my changes mean that `getTextureShaderDeclaration` no longer does anything with `startBindIndex`, which seems to be used in other places (and seemingly added by this PR https://github.com/playcanvas/engine/pull/7342).

I don't have very much context on Playcanvas or your shader system, so I would appreciate some feedback on the proper way to fix this bug!

## Checklist

### DIFF
--- a/src/platform/graphics/webgpu/webgpu-shader-processor-wgsl.js
+++ b/src/platform/graphics/webgpu/webgpu-shader-processor-wgsl.js
@@ -685,7 +685,7 @@ class WebgpuShaderProcessorWGSL {
             if (format.hasSampler) {
                 // A slot should have been left empty for the sampler at format.slot+1
                 const samplerName = format.sampleType === SAMPLETYPE_DEPTH ? 'sampler_comparison' : 'sampler';
-                code += `@group(${bindGroup}) @binding(${format.slot+1}) var ${format.samplerName}: ${samplerName};\n`;
+                code += `@group(${bindGroup}) @binding(${format.slot + 1}) var ${format.samplerName}: ${samplerName};\n`;
             }
         });
 


mvaligursky:
Hi @samw-improbable - thanks for the report, I'm definitely keen to sort this out, and it'd be amazing to have some repro, at least the shader that you used to create this issue, so that I can see its source and processed output.

willeastcott:
Hey @samw-improbable - just bumping this...do you have a reply for @mvaligursky? Thanks!

src/platform/graphics/webgpu/webgpu-shader-processor-wgsl.js
```diff
@@ -676,26 +676,23 @@ class WebgpuShaderProcessorWGSL {
...
-            code += `@group(${bindGroup}) @binding(${bindIndex}) var ${format.name}: ${textureTypeName};\n`;
-            bindIndex++;
+            code += `@group(${bindGroup}) @binding(${format.slot}) var ${format.name}: ${textureTypeName};\n`;
 
             if (format.hasSampler) {
+                // A slot should have been left empty for the sampler at format.slot+1
                 const samplerName = format.sampleType === SAMPLETYPE_DEPTH ? 'sampler_comparison' : 'sampler';
-                code += `@group(${bindGroup}) @binding(${bindIndex}) var ${format.samplerName}: ${samplerName};\n`;
-                bindIndex++;
+                code += `@group(${bindGroup}) @binding(${format.slot+1}) var ${format.samplerName}: ${samplerName};\n`;
```
mvaligursky:
```suggestion
                code += `@group(${bindGroup}) @binding(${format.slot + 1}) var ${format.samplerName}: ${samplerName};\n`;
```

PR #7893, opened 2025-08-08
Initial implementaion of octree for unified GSplat rendering

mvaligursky:
- initial implementation of an octree for unified splatting, to give us LOD ability
- simple test data hand created using two sogs playcanvas logos
- removed all temporary runtime lod generation for uncompressed splats, octree replaces this
- more work needed till this is made public, so WIP for now


Maksims:
It might be out of the scope, but I wander if there are plans for an efficient Octree and Quadtree implementations in the engine?

willeastcott:
> It might be out of the scope, but I wander if there are plans for an efficient Octree and Quadtree implementations in the engine?

It's a good point - is the octree here developed as a generic thing that can be applied to other use cases?

mvaligursky:
Nope, the octree developed here is not a generic one but very limited and specialised one for the streaming use case, vs visibility culling.

mvaligursky:
There are no current plans for a generic culling one at the moment. I considered it few times, but for the typical small amount of entities most projects have, it never made it to the top of the list.

Maksims:
We have a project where we've implemented Quadtree for LoD's, and batching (batch per node). As without it, we simply had 0.5fps, now we have 60fps+ on the same scenes.

Issue #7840, opened 2025-07-17
Light textures


LeXXik:
This is an engine-only application that is not using lights and doesn't even have a light component system. It looks like the engine still creates several light related textures:

<img width="1371" height="631" alt="Image" src="https://github.com/user-attachments/assets/3e712863-1090-4977-ad2e-843e4d1fa430" />

I've also tried setting clustered lights to false, but it doesn't seem to affect it:

```js
app.scene.clusteredLightingEnabled = false;
```

mvaligursky:
It does, but the textures are pretty tiny. We could avoid it pretty easily though I think.

PR #7894, opened 2025-08-08
Don’t allocate default clustered lighting textures when clustered lighting is disabled

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/7840 in case clustered lighting is disabled.
If it's enabled, those are more complex to avoid the allocation, as so this is ignored.

Has to be merged after https://github.com/playcanvas/engine/pull/7893
(errors should be fixed by merging main here at that point)


src/scene/renderer/renderer.js
```diff
@@ -189,8 +189,10 @@ class Renderer {
         this._shadowRendererDirectional = new ShadowRendererDirectional(this, this.shadowRenderer);
 
         // clustered passes
-        this._renderPassUpdateClustered = new RenderPassUpdateClustered(this.device, this, this.shadowRenderer,
-            this._shadowRendererLocal, this.lightTextureAtlas);
+        if (scene.clusteredLightingEnabled) {
```
slimbuck: this.scene?

willeastcott (approved):

src/scene/renderer/renderer.js
mvaligursky: yep, thanks!

PR #7900, opened 2025-08-11
Examples browser version fixes

kpal81xd:
### Fixes
- Fixes incorrect engine version for examples build

### Additions
- Adds version to Examples browser header

### Preview
image

## Checklist

### DIFF
--- a/utils/rollup-version-revision.mjs
+++ b/utils/rollup-version-revision.mjs
@@ -5,15 +5,14 @@ import { execSync } from 'node:child_process';
  * @returns {string} Version string like `1.58.0-dev`
  */
 function getVersion() {
-    let version;
+    let path;
     try {
-        const root = execSync('git rev-parse --show-toplevel').toString().trim();
-        const pkg = JSON.parse(readFileSync(`${root}/package.json`, 'utf8'));
-        version = pkg.version;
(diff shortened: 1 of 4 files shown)


utils/rollup-version-revision.mjs
```diff
@@ -5,9 +5,15 @@ import { execSync } from 'node:child_process';
...
-    const text = readFileSync('./package.json', 'utf8');
-    const json = JSON.parse(text);
-    return json.version;
+    let version;
+    try {
+        const root = execSync('git rev-parse --show-toplevel').toString().trim();
+        const pkg = JSON.parse(readFileSync(`${root}/package.json`, 'utf8'));
+        version = pkg.version;
+    } catch (e) {
+        version = '0.0.0';
```
willeastcott: Wouldn't it be more accurate to say "Unknown Version" or just don't show it at all?

kpal81xd: Technically since 0.0.0 is still a valid version but would look weird in places where we prepend v

Tbh its probably not even worth adding a catch for the error since the error would be descriptive and the only reason why it would error would be invalid package json or git is not installed. Either way its probably better to let the build fail for user interaction

Issue #7674, opened 2025-05-15
Memory leak inside ParticleEmitter


heretique:
Hey PlayCanvas team, we were doing some memory leaks investigations on our app that uses PlayCanvas engine and we noticed a lot of leaked buffers after the entities containing the particle systems are destroyed. The way we're spawning them is we have some templates with entities with various particle systems and our VFX instantiates them in the scene on demand and then it destroys them when the particle systems have finished playing.
So when doing comparison of heap snapshots before and after spawning those vfx entities a few times we would see something like this:

![Image](https://github.com/user-attachments/assets/81fd8574-fb74-4089-9dd1-8ef3ca085f83)

A lot of `JSArrayBufferData` objects that are being retained by the `meshInstances` property of the `StandardMaterial` for the particular `defaultMaterial`.
Upon investigation we determined that this was due to the fact that the `meshInstance` property of the `ParticleEmitter` is not destroyed when the emitter's `destroy` method is called (and hence `_destroyResources`) and when the code reaches the point of destroying the `material` property it references the `meshInstance` inside the `defaultMaterial` where it remains forever, leaking.

Adding these lines of code to the `_destroyResources` method solves the problem

```
        this.colorParam?.destroy();
        this.colorParam = null;
// newly added
        this.meshInstance?.destroy();
        this.meshInstance = null;

        this.mesh?.destroy();
        this.mesh = null;
///
        this.vertexBuffer?.destroy();
        this.vertexBuffer = undefined; // we are testing if vb is undefined in some code, no idea why

        this.indexBuffer?.destroy();
        this.indexBuffer = undefined;

        this.material?.destroy();
        this.material = null;
```
We didn't create a PR because we were not sure if not destroying the `meshInstance` is intentional or an oversight, but it doesn't seem to have any obvious side effects, except getting rid of the memory leaks

heretique:
To add more info here, we noticed that for some particle systems the mesh was missing when played the second time and we figured that we should not call
```
        this.mesh?.destroy();
        this.mesh = null;
```
only destroy the mesh instance

mvaligursky:
Great investigation, thanks for report. I have a fix PR, and I think it should work correctly with it.

PR #7901, opened 2025-08-11
[Fix] Changes to ParticleEmitter to properly destroy the meshInstance

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/7674

### DIFF
--- a/src/scene/particle-system/particle-emitter.js
+++ b/src/scene/particle-system/particle-emitter.js
@@ -581,7 +581,7 @@ class ParticleEmitter {
         gd.fragmentUniformsCount < 64 || // force CPU if can't use many uniforms; TODO: change to more realistic value (this one is iphone's)
         gd.forceCpuParticles;
 
-        this._destroyResources();
+        const wasVisible = this._destroyResources();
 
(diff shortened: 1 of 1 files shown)


src/scene/particle-system/particle-emitter.js
```diff
@@ -1194,16 +1193,19 @@ class ParticleEmitter {
         this.colorParam?.destroy();
         this.colorParam = null;
 
-        this.vertexBuffer?.destroy();
         this.vertexBuffer = undefined; // we are testing if vb is undefined in some code, no idea why
-
-        this.indexBuffer?.destroy();
         this.indexBuffer = undefined;
 
+        const wasVisible = this.meshInstance?.visible ?? true;
```
willeastcott, line 1199: Wait...isn't `wasVisible` declared at line 584?

mvaligursky: It was. And now it's declared here as well.
Old code was not destroying mesh instance at all .. and so was reading it from it. Now I destroy it, and return it from there.

willeastcott (approved):

PR #7905, opened 2025-08-12
Don’t destroy SOGS packing shader

mvaligursky:
leave it available for additional sogs

### DIFF
--- a/src/scene/gsplat/gsplat-sogs-data.js
+++ b/src/scene/gsplat/gsplat-sogs-data.js
@@ -310,6 +310,7 @@ class GSplatSogsData {
         const { device } = means_l;
         const { scope } = device;
 
+        // Note: do not destroy it, keep it available for the lifetime of the app
         const shader = ShaderUtils.createShader(device, {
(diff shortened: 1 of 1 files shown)


src/scene/gsplat/gsplat-sogs-data.js
```diff
@@ -310,6 +310,7 @@ class GSplatSogsData {
         const { device } = means_l;
         const { scope } = device;
 
+        // Note: do not destroy it, keep it available for the lifetime of the app
         const shader = ShaderUtils.createShader(device, {
```
slimbuck, line 314: the variable should be stored globally then?

mvaligursky: It's in the cache using provided unique name, no need to handle it here too. This matches other shaders used internally in the engine

slimbuck: ok, the name `createShader` is not the best then, because it seems to be equivalent to `new Shader`.

PR #7907, opened 2025-08-13
Gizmo features & fixes

kpal81xd:
Fixes #7892

## Non-breaking changes
- All axes show on translation when dragging
- Add drag modes to set gizmo visibility state when dragging:
   - `show`: shows all shapes on drag
   - `hide`: hides all shapes on drag 
   - `selected`: only shows axes in translation/scale space
- `setTheme` method to update all colors
- Scaling stuttering for negative scaling fixed

## Deprecations
- `xAxisColor`, `yAxisColor`, `zAxisColor`, `colorAlpha` color setter/getters
- `flipShapes` boolean flag

> [!WARNING]
> The deprecation implementations cannot be combined with the new API as new API splits grouped properties into separate values

## Fixes
- Fixes gizmo translation in axis perpendicular to the camera direction to remove warping

## Checklist
- [x] I have read the [contributing guidelines](https://github.com/playcanvas/engine/blob/master/.github/CONTRIBUTING.md)
- [x] My code follows the project's coding standards
- [ ] This PR focuses on a single change

PR #7910, opened 2025-08-18
Gizmo API changes

kpal81xd:
## Breaking changes
- Removed `shading` field

## Non-breaking changes
- Defaults gizmos to use selected mode (only shows active axes)
- Updated hover to highlight arrows when selecting planes or center
- Adds occlusion value between 0 - 1 to change guide line occlusion through selected object (defaults to 0.8)
- Updated theme keys and typings from #7907
- Modified rotate gizmo starting angle guide line from black to the axis color
- Modified `GIZMOAXIS_FACE` constant `face` to `f` (only affects users if using literal value)
- Deprecates `GIZMOSPACE_*` constants in favor of literals
- Deprecates `GIZMOAXIS_*` constants in favor of literals
- Removed Debug assertion for Camera component 

## Fixes
- Fixes rotate gizmo glitching by limiting pitch in example 
- Fixes inverse scaling causing rotation on gizmo

## Refactor
- Added missing typings to shapes

### Preview
image

## Checklist

### DIFF
--- a/src/extras/gizmo/shape/shape.js
+++ b/src/extras/gizmo/shape/shape.js
@@ -105,7 +105,6 @@ tmpG.normals = [];
  * @property {Color} [defaultColor] - The default color of the shape.
  * @property {Color} [hoverColor] - The hover color of the shape.
  * @property {Color} [disabledColor] - The disabled color of the shape.
- * @ignore
  */
 
 /**


mvaligursky:
> Set GIZMOAXIS_FACE="f"
> Set starting guide angle line to match guide axis color

Can you please add details for those breaking changes, allowing users to handle those changes? I'm not sure users would understand what they need to do to upgrade to new version. From reading this it seems the API has not changed, only behavior?

kpal81xd:
> > Set GIZMOAXIS_FACE="f"
> > Set starting guide angle line to match guide axis color
> 
> Can you please add details for those breaking changes, allowing users to handle those changes? I'm not sure users would understand what they need to do to upgrade to new version. From reading this it seems the API has not changed, only behavior?

So in this case its odd because the constant value itself has changed so if you are using the constant with a method such as `enableShape` there will be no change. This only affects users who are using literals directly

mvaligursky:
In that case, can we internally remap old to new, and add Debug.deprecated, to avoid breaking changes.

kpal81xd:
Additionally I might omit the use of constants entirely its unclean and we can filtering options using types instead. Then all constants will be deprecated

mvaligursky:
> Additionally I might omit the use of constants entirely its unclean and we can filtering options using types instead. Then all constants will be deprecated

We discussed this as an option before, but for now it's probably keep to the engine's conventions than going separate way in some cases.

src/extras/gizmo/constants.js
```diff
@@ -1,69 +1,116 @@
...
+ * - 'selected': show only the axis shapes for the affected axes
+ *
+ * @typedef {'show' | 'hide' | 'selected'} GizmoDragMode
+ */
+
 /**
  * Local coordinate space.
  *
  * @category Gizmo
+ * @deprecated Use the literal 'local' instead - {@link GizmoSpace}
```
willeastcott, line 42: If it's deprecated, should we just ignore it from the docs to avoid confusion and bloating the API?

src/extras/gizmo/shape/arc-shape.js
```diff
@@ -39,55 +91,87 @@ class ArcShape extends Shape {
...
     get tolerance() {
         return this._tolerance;
     }
 
-    _updateTransform() {
+    /**
+     * Update the shape's transform.
+     *
+     * @protected
+     * @override
```
willeastcott, line 175: What effect does `@override` have on the generated API ref?

src/extras/gizmo/constants.js
kpal81xd: Yep

src/extras/gizmo/shape/arc-shape.js
kpal81xd: Well shapes should be hidden they are not exposed. Override tag just useful to show there is existing behavior in the base class its overriding so like to put it in

PR #7917, opened 2025-08-22
Gizmo examples camera controls

kpal81xd:
## Description
- Updates individual gizmo examples to use new camera controls

## Checklist

### DIFF
--- a/examples/src/examples/gizmos/transform-rotate.example.mjs
+++ b/examples/src/examples/gizmos/transform-rotate.example.mjs
@@ -79,7 +79,7 @@ camera.addComponent('camera', {
     clearColor: new pc.Color(0.1, 0.1, 0.1),
     farClip: 1000
 });
-const cameraOffset = 4 * camera.camera?.aspectRatio;
+const cameraOffset = 4 * camera.camera.aspectRatio;
 camera.setPosition(cameraOffset, cameraOffset, cameraOffset);
 app.root.addChild(camera);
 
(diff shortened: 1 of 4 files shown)


examples/src/examples/gizmos/transform-rotate.example.mjs
```diff
@@ -66,24 +66,37 @@ box.addComponent('render', {
...
 camera.addComponent('camera', {
     clearColor: new pc.Color(0.1, 0.1, 0.1),
     farClip: 1000
 });
-camera.addComponent('script');
-const orbitCamera = camera.script.create('orbitCamera');
-camera.script.create('orbitCameraInputMouse');
-camera.script.create('orbitCameraInputTouch');
-camera.setPosition(1, 1, 1);
+const cameraOffset = 4 * camera.camera?.aspectRatio;
```
LeXXik: Does it mean the aspectRatio might not be present at the time? If so, camera position would be NaN from that point forward.

LeXXik: Actually, just noticed AI suggested the same in another place.

examples/src/examples/gizmos/transform-scale.example.mjs
```diff
@@ -66,24 +66,37 @@ box.addComponent('render', {
...
 camera.addComponent('camera', {
     clearColor: new pc.Color(0.1, 0.1, 0.1),
     farClip: 1000
 });
-camera.addComponent('script');
-const orbitCamera = camera.script.create('orbitCamera');
-camera.script.create('orbitCameraInputMouse');
-camera.script.create('orbitCameraInputTouch');
-camera.setPosition(1, 1, 1);
+const cameraOffset = 4 * camera.camera?.aspectRatio;
```
LeXXik: Here as well.

examples/src/examples/gizmos/transform-rotate.example.mjs
kpal81xd: It is defined its just typescript cannot assert that `camera.camera` exists since its added with `addComponent` and not assigned directly

PR #7929, opened 2025-08-27
Support SOG v2 format

slimbuck:
## Description
This PR implements SOG v2 bundled and unbundled format (see [here](https://github.com/playcanvas/splat-transform/issues/38) for details):
- the biggest change to runtime is the load-time interleaving of data to support v2 data layout:
    - resolve scale, position and sh0 codebooks into higher bit storage during the load-time interleave step
    - resolve spherical harmonics from rgb888 + codebook to normalized rgb111110
    - support SOG v1 data by interleaving to the same render format at load time
- include small fix to order data correctly when uncompressing compressed.ply files
- update examples to include a '.sog' example file

## Notes
- it's possible that rendering directly from codebook is performant, but needs more testing especially on lower-end devices
- if the codebooks could be used, then we could pack sh0 into the interleaved data (it's currently a separate texture/load)
- the bundled handler currently downloads the entire file before loading the constituent files. in a future PR we can improve this by first downloading the file directory from the end of the file and then downloading constituent files in a range request.

### DIFF
--- a/examples/src/examples/gaussian-splatting/multi-splat.example.mjs
+++ b/examples/src/examples/gaussian-splatting/multi-splat.example.mjs
@@ -99,7 +99,6 @@ assetListLoader.load(() => {
 
     // clone the biker and add the clone to the scene
     const skull = createSplatInstance('skull', assets.skull, 1.5, 0.05, 0, 0.7);
-    skull.setLocalPosition(1.5, 0.05, 0);
     skull.rotate(0, 150, 0);
     app.root.addChild(skull);
 


mvaligursky:
skull.ply was handy to test compressed format rendering with SH. I'd vote for keeping it, and adding sogs version in addition to this, to avoid not having all paths tested.

src/framework/parsers/sogs.js
```diff
@@ -61,6 +97,11 @@ class SogsParser {
     }
 
     async loadTextures(url, callback, asset, meta) {
+        // transform meta to latest shape
+        if (meta.version !== 2) {
+            meta = upgradeMeta(meta);
```
mvaligursky, line 104: do we want to add some Debug.deprecated call to warn people they use deprecated format? 
Or do we officially support both in the future.

src/scene/gsplat/gsplat-sogs-data.js
```diff
@@ -335,9 +377,52 @@ class GSplatSogsData {
...
+        renderTarget.destroy();
+    }
+
+    packShMemory() {
+        const { meta, sh_centroids } = this;
+        const { device } = sh_centroids;
+        const { scope } = device;
+
+        const shader = ShaderUtils.createShader(device, {
+            uniqueName: 'GsplatSogsReorderShShader',
```
mvaligursky: we probably need to make uniqueName more unique - based on defines, meta.version and all those conditions that genereate different shader. Otherwise multiple files loaded that need different shader, will get the same shader.

mvaligursky (approved):

src/framework/parsers/sogs.js
slimbuck: Hmmm yeah maybe, I'll add a warning thanks.

src/scene/gsplat/gsplat-sogs-data.js
slimbuck: I've made it unique, just based on version. Please check that'll work!

mvaligursky: looks good

PR #7935, opened 2025-09-02
Gizmo style updates

kpal81xd:
## Description
- Adds thicker gizmo angle guide lines for better visibility 

## Non breaking changes
- `angleGuideThickness` setter/getter in RotateGizmo

## Preview
image

## Checklist

### DIFF
--- a/src/extras/gizmo/mesh-line.js
+++ b/src/extras/gizmo/mesh-line.js
@@ -118,6 +118,20 @@ class MeshLine {
     }
 
     /**
+     * @type {number}
+     */
+    set thickness(value) {
+        this._thickness = value ?? this._thickness;
+    }
+
+    /**
+     * @type {number}
+     */
+    get thickness() {
+        return this._thickness;
+    }
+
+    /**
      * Draw a line from one point to another with a specific color.
      *
      * @param {Vec3} from - The starting point of the line.
@@ -140,6 +154,9 @@ class MeshLine {
         this._entity.enabled = true;
     }
 
+    /**
+     * Hides the mesh line.
+     */
     hide() {
         this._entity.enabled = false;
     }
--- a/src/extras/gizmo/mesh-line.js
+++ b/src/extras/gizmo/mesh-line.js
@@ -72,6 +72,9 @@ const shaderDesc = {
     `
 };
 
+/**
+ * @ignore
+ */
 class MeshLine {
     /**
      * @type {number}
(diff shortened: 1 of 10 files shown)


willeastcott (approved):
Looks great! 🔥

src/extras/gizmo/mesh-line.js
```diff
@@ -0,0 +1,145 @@
...
+        this._material.update();
+
+        const mesh = Mesh.fromGeometry(app.graphicsDevice, new CylinderGeometry());
+        const meshInstance = new MeshInstance(mesh, this._material);
+
+        this._entity.addComponent('render', {
+            meshInstances: [meshInstance],
+            layers: [layer.id]
+        });
+        app.root.addChild(this._entity);
```
LeXXik: I wonder if the MeshLine class should be responsible only for creating the line entity. Is it going to be a public API? Adding the line to app root would then have to be handled by the script instantiating it.

src/extras/gizmo/mesh-line.js
```diff
@@ -0,0 +1,148 @@
...
+     * @type {number}
+     * @private
+     */
+    _thickness = 0.02;
+
+    /**
+     * @type {ShaderMaterial}
+     * @private
+     */
+    _material = new ShaderMaterial(shaderDesc);
```
mvaligursky: does each line segment have a material? How many draw calls does it cost to render a gizmo, or are those somewhere later batched in some form? If they're batched, we would not need all these materials.

mvaligursky: Or are there only few of these .. and these are not used for curved drawing.

src/extras/gizmo/mesh-line.js
kpal81xd: Ah yes I meant to move that into the rotation gizmo class itself. As for the API for the class itself it will be kept private API until the structure is finalized

src/extras/gizmo/mesh-line.js
kpal81xd: Yea it has a material but looking it can probably reuse the same material from shape since the shader is identical. I will extract that into a separate file for reuse

src/extras/gizmo/shaders.js
```diff
@@ -0,0 +1,57 @@
+import { SEMANTIC_POSITION } from '../../platform/graphics/constants.js';
+
+export const materialDesc = {
```
mvaligursky: I'd vote for less generic name, considering it's being exported.

PR #7951, opened 2025-09-05
Fixes to SOGS data unloading

mvaligursky:
- handle case when SOGS resource is waiting for async GPU processing and gets destroyed - skip early from unfinished processing, don't create resource when device / textures were destroyed
- while still GPU processing, postpone deletion of objects to avoid crashes, and destroy at the end
- when SOGS resource unloads, unload texture resources it created
- similarly, when the device is destroyed, cleanly handle async texture download


saurabh-vh:
Please can anyone explain me what is SOGS. I am newbee

PR #7954, opened 2025-09-08
Gizmo fixes

kpal81xd:
## Fixes
- Adds shader flag to allow for disabling depth testing for mesh line
- Use initial rotation for axis mirroring since inverting scale will flip global rotation
- Fixed editor example deselect when aligning camera

## Checklist

### DIFF
--- a/examples/src/examples/misc/editor.example.mjs
+++ b/examples/src/examples/misc/editor.example.mjs
@@ -195,12 +195,14 @@ data.set('viewCube', {
     lineLength: viewCube.lineLength
 });
 const tmpV1 = new pc.Vec3();
+let aligned = false;
 viewCube.on(pc.ViewCube.EVENT_CAMERAALIGN, (/** @type {pc.Vec3} */ dir) => {
     const cameraPos = camera.getPosition();
     const focusPoint = cc.focusPoint;
     const cameraDist = focusPoint.distance(cameraPos);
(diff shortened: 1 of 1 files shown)


src/extras/gizmo/shaders.js
```diff
@@ -3,62 +3,65 @@ import { SEMANTIC_POSITION } from '../../platform/graphics/constants.js';
...
-        void main(void) {
-            gl_FragColor = vec4(gammaCorrectOutput(decodeGamma(uColor)), uColor.w);
-            gl_FragDepth = 0.0;
-        }
-    `,
-    vertexWGSL: /* wgsl */`
-        attribute vertex_position: vec3f;
+            void main(void) {
+                gl_FragColor = vec4(gammaCorrectOutput(decodeGamma(uColor)), uColor.w);
+                ${depthTest ? '' : 'gl_FragDepth = 0.0;'}
```
mvaligursky: This works, but I'd prefer to keep the string (shader source code) constant, to avoid generating these larger strings each time unlitShader is called.
I should use fragmentDefines, and define 'DEPTH_TEST' or so.
So in the case you already created this shader, there is minimal string operation taking place - and you'd get shader from internal cache.

kpal81xd: Yea i wasn't entirely sure how to do this - the string templating seems messy. Ill try with fragmentDefines

PR #7985, opened 2025-09-18
(fix) Debug.assert raising react error

marklundin:
During @playcanvas/react rendering the camera-frame.js is disable is called `component.destroy()` before the script instance being destroyed itself. This was causing the `Debug.assert()` to raise a console error incorrectly, which looked like a user error. This PR removes the redundant assert.

### DIFF
--- a/src/extras/render-passes/camera-frame.js
+++ b/src/extras/render-passes/camera-frame.js
@@ -355,15 +355,11 @@ class CameraFrame {
     }
 
     enable() {
-        Debug.assert(!this.renderPassCamera);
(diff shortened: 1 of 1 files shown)


lucaheft:
Just stumbled upon this and wanted to add, that this also happens when using camera-frame.mjs and calling `SceneRegistry.changeScene`

mvaligursky:
Hopefully it no longer happens with engine https://github.com/playcanvas/engine/releases/tag/v2.12.0 which has the fix

PR #7992, opened 2025-09-22
Add custom feedback varyings for WebGL Shader with useTransformFeedback

AlexAPPi:
## Add custom feedback varyings
This PR introduces support for passing a list of output varyings for use with transform feedback in WebglShader

Ensured compatibility with the existing shader compilation flow — if varyings are not specified, the behavior remains unchanged.

### DIFF
--- a/src/platform/graphics/shader.js
+++ b/src/platform/graphics/shader.js
@@ -63,6 +63,9 @@ class Shader {
      * vertex shader attribute names to semantics SEMANTIC_*. This enables the engine to match
      * vertex buffer data as inputs to the shader. When not specified, rendering without vertex
      * buffer is assumed.
+     * @param {string[]} [definition.outVaryings] - A list of shader output variable
+     * names that will be captured when using transform feedback. This setting is only effective
+     * if the {@link definition.useTransformFeedback} property is enabled.
      * @param {string} [definition.vshader] - Vertex shader source (GLSL code). Optional when
      * compute shader is specified.
      * @param {string} [definition.fshader] - Fragment shader source (GLSL code). Optional when
--- a/src/platform/graphics/shader.js
+++ b/src/platform/graphics/shader.js
@@ -63,7 +63,7 @@ class Shader {
      * vertex shader attribute names to semantics SEMANTIC_*. This enables the engine to match
      * vertex buffer data as inputs to the shader. When not specified, rendering without vertex
      * buffer is assumed.
-     * @param {string[]} [definition.outVaryings] - A list of shader output variable
+     * @param {string[]} [definition.feedbackVaryings] - A list of shader output variable
      * names that will be captured when using transform feedback. This setting is only effective
      * if the {@link definition.useTransformFeedback} property is enabled.
      * @param {string} [definition.vshader] - Vertex shader source (GLSL code). Optional when
--- a/src/platform/graphics/webgl/webgl-shader.js
+++ b/src/platform/graphics/webgl/webgl-shader.js
@@ -150,13 +150,21 @@ class WebglShader {
         const definition = shader.definition;
         const attrs = definition.attributes;
         if (definition.useTransformFeedback) {
-            // Collect all "out_" attributes and use them for output
-            const outNames = [];
-            for (const attr in attrs) {
-                if (attrs.hasOwnProperty(attr)) {
-                    outNames.push(`out_${attr}`);
+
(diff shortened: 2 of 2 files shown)


src/platform/graphics/shader.js
```diff
@@ -63,6 +63,9 @@ class Shader {
      * vertex shader attribute names to semantics SEMANTIC_*. This enables the engine to match
      * vertex buffer data as inputs to the shader. When not specified, rendering without vertex
      * buffer is assumed.
+     * @param {string[]} [definition.feedbackVaryings] - A list of shader output variable
+     * names that will be captured when using transform feedback. This setting is only effective
+     * if the {@link definition.useTransformFeedback} property is enabled.
```
mvaligursky: not sure this is valid, please check the generated docs:
{@link definition.useTransformFeedback}

mvaligursky:
Lets update the example to use it as well, instead of the current undocumented behavior. I see that the shader is created like this, which is the documented way:

https://github.com/playcanvas/engine/blob/4ed97e4b6da1406b09d5c660e9dea8853be7fc8a/examples/src/examples/graphics/transform-feedback.example.mjs#L145-L149

and so this API needs to be updated to accept the `feedbackVaryings` as well.
The main `Shader` constructor does not need to expose it as a public property there I think.

src/platform/graphics/shader.js
AlexAPPi: +

mvaligursky:
Lets also update the engine example to use it as mentioned above please.

AlexAPPi:
> Lets also update the engine example to use it as mentioned above please.

It doesn’t make sense to change the particle example. Instead, it would be better to add a separate example that demonstrates the use of a custom shader and custom variables. For instance, we could show how to unpack data on the GPU side: take an array of numbers and transform it into a grid or another structure for visualization.

mvaligursky:
We should perhaps just rename this
https://github.com/playcanvas/engine/blob/ae4e537f1f2dd49d55c6c08656c19359c6fbaec1/examples/src/examples/graphics/transform-feedback.shaderFeedback.vert#L6

from `out_vertex_position` to `updated_position`
and use that name when creating the shader .. making this implicit instead of some name matching under the hood.

src/platform/graphics/transform-feedback.js
```diff
@@ -91,21 +93,46 @@ class TransformFeedback {
...
+        }
+
         this.device = inputBuffer.device;
+
         const gl = this.device.gl;
 
-        Debug.assert(inputBuffer.format.interleaved || inputBuffer.format.elements.length <= 1,
-            'Vertex buffer used by TransformFeedback needs to be interleaved.');
+        if (outputBuffer) {
+            Debug.assert(outputBuffer.format.interleaved || outputBuffer.format.elements.length <= 1,
```
mvaligursky: this code is duplicated, just with a different buffer.
I suggest:
```
const outVB = outputBuffer ?? inputBuffer;

// run the code once on outVB
```

src/platform/graphics/transform-feedback.js
```diff
@@ -91,21 +93,46 @@ class TransformFeedback {
...
 
         this._inputBuffer = inputBuffer;
-        if (usage === BUFFER_GPUDYNAMIC && inputBuffer.usage !== usage) {
-            // have to recreate input buffer with other usage
-            gl.bindBuffer(gl.ARRAY_BUFFER, inputBuffer.impl.bufferId);
-            gl.bufferData(gl.ARRAY_BUFFER, inputBuffer.storage, gl.DYNAMIC_COPY);
-        }
 
-        this._outputBuffer = new VertexBuffer(inputBuffer.device, inputBuffer.format, inputBuffer.numVertices, {
+        this._swapEquivalent = inputBuffer.format === outputBuffer?.format;
```
mvaligursky: is this needed?

AlexAPPi: I was planning to throw a warn if swap was passed, but the formats don't match.

src/platform/graphics/transform-feedback.js
AlexAPPi: Fixed/

src/platform/graphics/transform-feedback.js
AlexAPPi: Fixed.

PR #8003, opened 2025-09-26
fix(template-asset): invalidate cached _previewRoot upon receiving new data

### LINKED ISSUE #8002: When updating the raw `data` of a template asset, the associated template does not get updated
### Issue

In the Editor, template assets get updated, when the user applies an override on them. The underlying raw `data` gets updated here:

https://github.com/playcanvas/engine/blob/2d32970cbc83dcc6e0aefcc77956d5fd547bcd26/src/framework/asset/asset.js#L361-L373

But the actual `resource`, which corresponds to a `Template`, does not get touched, so we are working with a stale version of the template within the editor.

Example:

```js
const first = asset.resource.instantiate();
asset.data = data2;
const second = asset.resource.instantiate();
// second has the same structure as first
```

abstrakt8:
## Description

Fixes #8002 

* In the `TemplateResourceHandler`, we had no corresponding `patch` function that would make sure that the assets `Template` gets updated based on the new raw `data`.

## Checklist

### DIFF
--- a/src/framework/handlers/template.js
+++ b/src/framework/handlers/template.js
@@ -52,6 +52,18 @@ class TemplateHandler extends ResourceHandler {
         this.decoder ??= new TextDecoder('utf-8');
         return new Template(this._app, JSON.parse(this.decoder.decode(data)));
     }
+
+    patch(asset, registry) {
+        // only process if this looks like valid template data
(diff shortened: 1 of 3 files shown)


mvaligursky (approved):
Looks great!

PR #8012, opened 2025-10-01
Fix: SOG proper removing of texture assets from asset cache

fimbox:
## Description
When a SOG asset is destroyed, GSplatSogsData.destroy() calls .destroy() directly on the texture resources. This fails to remove the parent Asset from the AssetRegistry cache.

Fixes #
Modify GSplatSogsData.destroy() to call asset.unload() on the texture Asset objects instead of destroying the resources directly.

### DIFF
--- a/src/framework/components/camera/component.js
+++ b/src/framework/components/camera/component.js
@@ -461,6 +461,24 @@ class CameraComponent extends Component {
     }
 
     /**
+     * Sets the depth value to clear the depth buffer to. Defaults to 1.
+     *
+     * @type {number}
+     */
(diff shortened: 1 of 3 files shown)


slimbuck:
I think we only need `orderTexture.destroy()` which was already added in https://github.com/playcanvas/engine/pull/8016?

slimbuck:
Thanks so much for this PR @fimbox!!

I've implemented the same functionality in a (hopefully) slightly cleaner way in #8037.

It wouldn't have happened without your contribution though - thanks again!

PR #8021, opened 2025-10-02
Ring buffer for SkinInstance

AlexAPPi:
## PR Description
This PR introduces a ring buffer for SkinInstance, used when updating the bone matrix texture.
Instead of a single buffer that could cause frequent GPU stalls during data uploads, the system now uses a multi-slot ring buffer across frames.

## Main Changes
* Implemented a ring buffer to store bone matrices, allowing updates to be written into independent slots per frame.
* Bone texture updates no longer wait for the previous frame to complete, reducing the risk of GPU stalls.
* Updated SkinInstance logic to pick the active ring buffer slot based on the current render frame.
* Added configurability for the number of ring buffer slots (default is 1).

## Results
* Reduced GPU blocking during bone data uploads.
* More stable frame times when animating many characters simultaneously.
* Lowered latency between CPU → GPU updates.
* This update is fully backward-compatible and does not affect existing behavior.

I have read the [contributing guidelines](https://github.com/playcanvas/engine/blob/master/.github/CONTRIBUTING.md)

### DIFF
--- a/src/scene/skin-instance.js
+++ b/src/scene/skin-instance.js
@@ -19,6 +19,7 @@ const _invMatrix = new Mat4();
  * @category Graphics
  */
 class SkinInstance {
+
     /**
      * An array of nodes representing each bone in this skin instance.
      *
@@ -33,6 +34,7 @@ class SkinInstance {
      * matrices to generate the final matrix palette.
      */
     constructor(skin) {
+
         this._dirty = true;
 
         // optional root bone - used for cache lookup, not used for skinning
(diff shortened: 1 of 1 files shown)


mvaligursky:
I developed `UploadStream` internal API to help with CPU->GPU uploads here: https://github.com/playcanvas/engine/pull/8053

And so I'm keen to use this for Skin/Batch Instances as well. Which means first switching those to use StorageBuffer instead of texture.

Issue #8015, opened 2025-10-01
WebGPU issue on iOS 26.0.1


LeXXik:
I have iPhone 12 Pro Max with iOS 18.7.1. The WebGPU examples stopped working, showing black screen (both Safari and Chrome on iOS).

After digging a bit, I found that the culprit PR is in 2.8.0 release:

https://github.com/playcanvas/engine/compare/v2.7.7...v2.8.0

<img width="1244" height="368" alt="Image" src="https://github.com/user-attachments/assets/7795ffc4-f6da-4d13-9ff8-ba36be0c7b38" />

If I load the deployment of **Sogs load fix** #7687 (or any other before it), then WebGPU examples work fine.
If I load **[BREAKING] New public API to customize shader chunks (Material and global)** #7665 (or any other after it), then I get black screen.

mvaligursky:
can confirm on iPhone 13 Pro Max

mvaligursky:
I got this from my contact: `It works for me on iOS, it might be a bug we fixed between iOS 26 and iOS 26.1`
But I'll try and see if I can figure out what the problem is and have a workaround, as our examples used to run.

mvaligursky:
Some conclusions at the moment:
- this PR causes it, confirming the initial report: https://github.com/playcanvas/engine/pull/7665
- only StandardMaterials seem affected. ShaderMaterial shaders work fine
- in the mentioned PR we switched to use native WGSL chunks, instead of transpiled GLSL chunks, so the WGSL shader we pass to the device are completely different, making it hard to pin-point specific workaround we need to get it to work. There are no warnings / errors, and those shaders work fine in MacOS / windows / android (tested all just now)
- the issue cannot be reproduced in xcode simulator with 26.0.1 OS

LeXXik:
Was able to get an error by launching a default project from Editor, which adds some debug logs to the html console:

```text
WebGPU BindGroup creation validation error: GPUDevice.createBindGroup: Bind group entry sampleType does not match TextureView sampleType

while rendering

Pass:RenderPassForward RT:WebgpuFramebuffer | Camera: Camera, Layer: World(OPAQUE) | Node: Plane, Material: Default Material

[object Object]

WebGPU BindGroup creation validation error: GPUDevice.createBindGroup: Bind group entry sampleType does not match TextureView sampleType

while rendering

[...]

PR #8038, opened 2025-10-08
[Fix] WebGPU: Use textureLoad for data textures instead of textureSample

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/8015

Converts multiple WGSL shaders to use direct textureLoad instead of sampled access. This allows those shaders to be compatible with the devices that do not support float texture filtering.

This can be forced by commenting out this in the engine code:
```
//this.textureFloatFilterable = requireFeature('float32-filterable');
```

This makes all engine examples compatible with those devices.


willeastcott (approved):
Nice! 👏

PR #8046, opened 2025-10-10
Fix exp perf regression

slimbuck:
## Description
The table introduced in #8030 results in performance regression so use exp instead.

This PR reverts the calc to using build-in exp.


Maksims:
Please add to names of PR's a bit more context. "Perf regression" of what?
There is a lot of GS's PR's, would be good to name them with "GS", so it is not confused with other things.

PR #8053, opened 2025-10-14
Add UploadStream API and optimize GSplat order data uploads with non-blocking transfers

mvaligursky:
### Overview
Introduces a new `UploadStream` API for efficient, non-blocking GPU data uploads and refactors GSplat unified rendering to use storage buffers on WebGPU with optimized upload strategies for both WebGL and WebGPU.

### New API: UploadStream
A platform-agnostic **private** API for uploading data to GPU resources (texture on WebGL / storage buffer on WebGPU) with two modes:
- **Simple mode** (`useSingleBuffer: true`): Direct uploads for low-frequency updates. Blocking if the resource is still in use by the GPU.
- **Optimized mode** (`useSingleBuffer: false`): Non-blocking uploads using PBOs (WebGL) or staging buffers (WebGPU)

**Signature:** `upload(data, target, offset = 0, size = data.length)`

### StorageBuffer API Extensions
- Added an optional `addStorageUsage` parameter to the `StorageBuffer` constructor (defaults to `true`) to support creating staging buffers
- Also added a `copy(srcOffset, dstBuffer, dstOffset, size)` method that wraps `copyBufferToBuffer` for efficient GPU-side buffer copies, enabling the staging buffer pattern used by `UploadStream`.

#### WebGL Optimizations (`WebglUploadStream`)
- PBO-based asynchronous texture uploads with "orphan + bufferSubData" pattern
- Non-blocking fence polling using `clientWaitSync(..., 0, 0)`
- Explicit pixel-store parameter configuration to prevent hidden CPU repacks

#### WebGPU Optimizations (`WebgpuUploadStream`)
- Staging buffer ring for non-blocking uploads via `copyBufferToBuffer`
- Direct `StorageBuffer.write()` for simple mode

#### GSplat Unified Rendering
  - Added `orderBuffer` (StorageBuffer) for WebGPU alongside `orderTexture` for WebGL
  - Integrated `UploadStream` in `setOrderData()` for non-blocking uploads
  
#### Shader Updates
- **`gsplatSource.js` (WGSL)**: Added conditional compilation via `#ifdef STORAGE_ORDER`
  - Storage buffer path for WebGPU: `var<storage, read> splatOrder: array<u32>`
  - Added `splatTextureSize` uniform

#### WebGPU

The impact is huge. The CPU thread is no longer blocked on the upload, and runs 60fps instead of 20 or so. We're still GPU limited, and so the frames are still skipped - but that can be addressed by GPU optimization.

before
webgpu before

now
webgpu now

#### WebGL

A lot lower impact, as there is no completely non-blocking way to upload, and performance depends on the driver implementation. In theory, this should deliver the same performance as WebGPU though.

before
webgl before

now
webgl now

### DIFF
--- a/src/platform/graphics/webgpu/webgpu-upload-stream.js
+++ b/src/platform/graphics/webgpu/webgpu-upload-stream.js
@@ -1,8 +1,10 @@
-import { Debug } from '../../../core/debug.js';
+import { Debug, DebugHelper } from '../../../core/debug.js';
 
 /**
  * @import { UploadStream } from '../upload-stream.js'
  */
+
+let id = 0;
 
 /**
  * WebGPU implementation of UploadStream.
@@ -136,11 +138,13 @@ class WebgpuUploadStream {
         // Get or create a staging buffer (guaranteed to be large enough after recycling)
         const buffer = this.availableStagingBuffers.pop() ?? (() => {
             // @ts-ignore - wgpu is available on WebgpuGraphicsDevice
-            return this.uploadStream.device.wgpu.createBuffer({
+            const newBuffer = this.uploadStream.device.wgpu.createBuffer({
                 size: byteSize,
                 usage: GPUBufferUsage.MAP_WRITE | GPUBufferUsage.COPY_SRC,
                 mappedAtCreation: true
             });
+            DebugHelper.setLabel(newBuffer, `UploadStream-Staging-${id++}`);
+            return newBuffer;
         })();
 
         // Write to mapped range (non-blocking)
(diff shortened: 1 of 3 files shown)


slimbuck:
What's the reason not to use this for the non-managed path as well?

mvaligursky:
No reason, can be done in a separate PR. I tried doing it but it was looking to be a larger change with the way sorter is handling data, so I didn't do it as part of this PR.

slimbuck:
> No reason, can be done in a separate PR. I tried doing it but it was looking to be a larger change with the way sorter is handling data, so I didn't do it as part of this PR.

Ok awesome. This is such a significant improvement that we need it everywhere all at once right now :)

ShenghuGong:
What is the difference between this submission and [#7782](https://github.com/playcanvas/engine/pull/7782) for webgl?  In scenes where textures are frequently updated, the asynchronous solution of [#7782](https://github.com/playcanvas/engine/pull/7782) has limited performance improvement.

mvaligursky:
> What is the difference between this submission and [#7782](https://github.com/playcanvas/engine/pull/7782) for webgl? In scenes where textures are frequently updated, the asynchronous solution of [#7782](https://github.com/playcanvas/engine/pull/7782) has limited performance improvement.

A lot depends on drivers. On WebGPU, the win here is large. On  WebGL, there is no true async upload, and so this all depends on what the driver can do behind the scene. But this PR uses the right API to allow for this.

PR #8065, opened 2025-10-21
Few improvements to unified gsplat rendering

mvaligursky:
- event gets triggered when unified rendering creates a material under the hood
```
GSplatComponentSystem - event called `'material:created'`
```
- public API to access these materials as well
```
GSplatComponentSystem.getGSplatMaterial(camera, layer)
```

- force update LOD when new gsplat is loaded, instead of waiting for other reason to initially update it
- added a helper gsplat shader function to scale it to a round dot with size

### DIFF
--- a/src/framework/components/gsplat/component.js
+++ b/src/framework/components/gsplat/component.js
@@ -242,7 +242,7 @@ class GSplatComponent extends Component {
     /**
      * Gets the material used to render the gsplat.
      *
-     * **Note:** This getter returns `null` when `unified is true`. In unified mode, materials are
+     * **Note:** This getter returns `null` when {@link unified} is `true`. In unified mode, materials are
      * organized per camera/layer combination rather than per component. To access materials in
      * unified mode, use {@link GsplatComponentSystem#getGSplatMaterial}.
      *


src/framework/components/gsplat/component.js
```diff
@@ -221,12 +221,17 @@ class GSplatComponent extends Component {
     /**
      * Sets the material used to render the gsplat.
      *
+     * **Note:** This setter is only supported when `unified is false`. When it's true, multiple
+     * gsplat components share a single material per camera/layer combination.To access materials in
+     * unified mode, use `app.systems.gsplat.getGSplatMaterial(camera, layer)`.
```
willeastcott:
```suggestion
     * **Note:** This setter is only supported when {@link unified} is `false`. When it's true, multiple
     * gsplat components share a single material per camera/layer combination. To access materials in
     * unified mode, use {@link GsplatComponentSystem#getGSplatMaterial}.
```

willeastcott: But test these links by building the docs! 😄

src/framework/components/gsplat/component.js
```diff
@@ -237,9 +242,17 @@ class GSplatComponent extends Component {
     /**
      * Gets the material used to render the gsplat.
      *
+     * **Note:** This getter returns `null` when `unified is true`. In unified mode, materials are
+     * organized per camera/layer combination rather than per component. To access materials in
+     * unified mode, use `app.systems.gsplat.getGSplatMaterial(camera, layer)`.
```
willeastcott: Considering applying my edits from above here too.

src/framework/components/gsplat/system.js
```diff
@@ -105,6 +126,35 @@ class GSplatComponentSystem extends ComponentSystem {
...
+     * Returns null if the material hasn't been created yet. In unified mode, materials are created
+     * during the first frame update when the GSplat is rendered. To be notified immediately when
+     * materials are created, listen to the 'material:created' event on GSplatComponentSystem:
+     *
+     * ```javascript
+     * app.systems.gsplat.on('material:created', (material, camera, layer) => {
+     *     // Material is now available
+     *     material.setParameter('myParam', value);
+     * });
+     * ```
```
willeastcott: Any reason not to put this in an example block?

src/framework/components/gsplat/system.js
```diff
@@ -105,6 +126,34 @@ class GSplatComponentSystem extends ComponentSystem {
...
+     *
+     * Returns null if the material hasn't been created yet. In unified mode, materials are created
+     * during the first frame update when the GSplat is rendered. To be notified immediately when
+     * materials are created, listen to the 'material:created' event on GSplatComponentSystem:
+     *
+     * @example
+     * app.systems.gsplat.on('material:created', (material, camera, layer) => {
+     *     // Material is now available
+     *     material.setParameter('myParam', value);
+     * });
```
willeastcott, line 141: This block needed to go below the param tags.

mvaligursky: I tried and the generated doc is the same.

willeastcott: It is. It's more a consistency thing. Every single JSDoc block we have in the engine is in this order:

* Description (no-tag)
* Params (if any)
* Example (if any)

PR #8088, opened 2025-10-31
[DOCS] add typedef ShaderDefinition and document Shader.definition

aidinabedi:
## Description
This PR takes the type info from the parameter `definition` in `Shader`'s constructor, and makes it into its own typedef called `ShaderDefinition`, so that it can be used to also document the property `Shader.definition`

## Checklist

### DIFF
--- a/src/platform/graphics/shader.js
+++ b/src/platform/graphics/shader.js
@@ -15,6 +15,40 @@ import { ShaderDefinitionUtils } from './shader-definition-utils.js';
 let id = 0;
 
 /**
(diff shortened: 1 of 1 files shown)


mvaligursky (commented):
It'm not sure about exposing `definition` as the public API. Two concerns:
1. It seems the user can modify it's content, but in reality it cannot.
2. The content of it is not the same as what the user passed in. As an example we process shaders and replace the originally supplied shaders. And so we cannot provide guarantees as of what it contains.

What's the reason you need `definition` exposed?

PR #8089, opened 2025-10-31
[DOCS] add documentation for Material.clearVariants

aidinabedi:
## Description
This PR documents the method `clearVariants` in `Material`.

## Checklist

### DIFF
--- a/src/scene/materials/material.js
+++ b/src/scene/materials/material.js
@@ -728,6 +728,9 @@ class Material {
         return this.parameters;
     }
 
+    /**
+     * Clears all cached shader variants from the material and mesh instances that reference it.
+     */
     clearVariants() {
 
         // clear variants on the material


mvaligursky:
What use case do you have that needs this to be public?
Currently it's internal function, and clears variants as needed, without you having to do it manually.

Maksims:
Did not know about this method, and it actually pretty useful in case when updating global shader chunks, and need to force material shader updates.

aidinabedi:
> What use case do you have that needs this to be public? Currently it's internal function, and clears variants as needed, without you having to do it manually.

@mvaligursky We need to call it in cases when we inherit `pc.Material` to create our own "smart" material like `pc.StandardMaterial`, which generate dynamic shader code based on material properties.

mvaligursky:
> Did not know about this method, and it actually pretty useful in case when updating global shader chunks, and need to force material shader updates.

In this case, you just call `material.update();`

src/scene/materials/material.js
```diff
@@ -728,6 +728,9 @@ class Material {
         return this.parameters;
     }
 
+    /**
+     * Clears all cached shader variants from the material and mesh instances that reference it.
```
mvaligursky, line 732: we should add `@protected` tag on it - allowing derived classes to use it, but not making a public API

Issue #7993, opened 2025-09-22
Gaussian Splatting: Significant quality loss when unified is enabled on multiple entities


MotivaCG:
When enabling unified rendering on more than one gaussian entity, image quality degrades noticeably. With a single splat, it looks correct without ordering it.

Sample:

One model without unified https://www.scanmenow.es/samples/dev/tmp/?model=model
Two models with unified https://www.scanmenow.es/samples/dev/tmp/?model=model2

slimbuck:
Such a great scene! ❤️❤️❤️

MotivaCG:
Updated to 2.12 it 'feels' faster and more responsive unfortunately the issue seems to be still there (you can check on the same url)

Btw congrats for the release!

mvaligursky:
Hi .. thanks for the repro, I confirmed this is fixed in my PR where I added needed functionality. Please reopen if you still see some issue. The engine with this should be out in the near future ( a week maybe), but you can test with local engine if needed.

MotivaCG:
Looks great now!

MotivaCG:
@mvaligursky  Unfortunately, although it looks much better now it still has some small glitches (attached video), maybe ordering issues?

[TestMultiGS.zip](https://github.com/user-attachments/files/23483308/TestMultiGS.zip)

MotivaCG:
@slimbuck @mvaligursky  Uppps it was on the desktop browser in the Android one (tested with Chrome and Brave) looks really bad (and it was working with the previous version with some small glitches but working.

<img width="591" height="1280" alt="Image" src="https://github.com/user-attachments/assets/1922aba8-ba51-4cd9-87bc-979998e5b6c4" />

Testing url https://www.scanmenow.es/samples/dev/pc/?model=andreaisasog&controls=1

slimbuck:
My pixel 7 pro seems to work

<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/64377ce3-21a1-4c75-8bea-d2d30e621cd5" />

MotivaCG:
Samsung S24 Ultra here :/

mvaligursky:
do you have the same issue with the engine examples using unified, such as this: https://playcanvas.vercel.app/#/gaussian-splatting/global-sorting ?

mvaligursky:
I can reproduce this on browser stack using your ply
This PR is causing the issue: https://github.com/playcanvas/engine/pull/8055

[...]

PR #8098, opened 2025-11-04
Unified gsplats re-evaluate their Spherical Harmonics when camera moves

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/7993

With the unified gsplats, the spherical harmonics are evaluated to a color, which is stored in the work buffer. Before this PR, this evaluation took place only when LOD changed. This was no often enough, causing visible popping, and also did not work without LOD, so splats without LODs didn't get their colors updated as camera moved.

This PR adds:
- support to render color only part for individual gsplat resources to work buffer
- this is used to regularly update SH to work buffer - but typically not every frame to save on the cost. There are 4 new parameters to control this behaviour, based on camera distance and angle change. Parameters also allow, and by default update SH, with lower frequency for more distant LODs, as the relative change is lower
- SH updates for individual resources is randomized to not take place in a single frame

## New public API on existing GSplatParams class

Screenshot 2025-11-04 at 11 30 19

New example - streaming LOD with SH

Screenshot 2025-11-04 at 11 22 08

### DIFF
--- a/examples/src/examples/gaussian-splatting/lod-streaming-sh.example.mjs
+++ b/examples/src/examples/gaussian-splatting/lod-streaming-sh.example.mjs
@@ -114,22 +114,11 @@ assetListLoader.load(() => {
     app.scene.gsplat.lodUpdateDistance = config.lodUpdateDistance;
     app.scene.gsplat.lodUnderfillLimit = config.lodUnderfillLimit;
 
-
-
-
-
-
-
-
-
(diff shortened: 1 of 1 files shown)


slimbuck:
Are the scale parameters really necessary? It seems a bit complicated API and not sure users would ever actually modify them.

mvaligursky:
I think they could be useful in some cases - for example a bee - you'd probably want to modify the distances to be tiny.

Issue #7933, opened 2025-09-01
Treeshake gsplat chunks


LeXXik:
Currently gsplat shader chunks are included, even if its component system is not used. Would be great to treeshake those off.

mvaligursky:
we need to create a collection for those chunks here (and also in wgsl folder)
https://github.com/playcanvas/engine/tree/main/src/scene/shader-lib/glsl/collections

and then handle imports to not include the collection outside of the gsplat shader generator or similar.

PR #8115, opened 2025-11-10
Refactor gsplat shader chunks for tree-shaking

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/7933

Moves gsplat shader chunk imports into separate collection files to enable tree-shaking when the GSplat component system is not used.

Previously, all gsplat shader chunks were imported directly in shader-chunks-glsl.js and shader-chunks-wgsl.js, preventing them from being tree-shaken during builds. This change follows the same pattern used for particle chunks, where chunks are only imported when their component system is included.

Changes:
- Created gsplat-chunks-glsl.js and gsplat-chunks-wgsl.js collection files
- Removed 23 gsplat chunk imports from main shader chunk files
- Updated GSplatComponentSystem to register chunks on initialization
- Reduces bundle size for projects that don't use Gaussian Splatting

### DIFF
--- a/src/scene/shader-lib/glsl/collections/gsplat-chunks-glsl.js
+++ b/src/scene/shader-lib/glsl/collections/gsplat-chunks-glsl.js
@@ -47,4 +47,3 @@ export const gsplatChunksGLSL = {
     gsplatWorkBufferVS,
     gsplatPackingPS
 };
-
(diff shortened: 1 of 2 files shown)


slimbuck (approved):
<3

LeXXik:
This shaved off 32 Kb from a minified build. Thanks!

PR #8117, opened 2025-11-10
[Fix] Fixed gsplat shader effect to handle limit to a camera

mvaligursky:
Fix to how the `camera` attribute of gsplat shader effect is used.

Typically, only a single camera is used, and so it does not need to be set. But when it's set to limit the effect to a single camera, there was a small bug and the effect would not be applied.

### DIFF
--- a/scripts/esm/gsplat/gsplat-shader-effect.mjs
+++ b/scripts/esm/gsplat/gsplat-shader-effect.mjs
@@ -198,9 +198,10 @@ class GsplatShaderEffect extends Script {
 
         // Determine which cameras to target
         let targetCameras;
-        if (this.camera && this.camera.camera && this.camera.camera.camera) {
+        const cam = this.camera?.camera?.camera;
+        if (cam) {
             // Specific camera specified via attribute
-            targetCameras = [this.camera.camera.camera];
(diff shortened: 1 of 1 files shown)


scripts/esm/gsplat/gsplat-shader-effect.mjs
```diff
@@ -198,9 +198,9 @@ class GsplatShaderEffect extends Script {
 
         // Determine which cameras to target
         let targetCameras;
-        if (this.camera && this.camera.camera) {
+        if (this.camera && this.camera.camera && this.camera.camera.camera) {
```
willeastcott:
```suggestion
        const cam = this.camera?.camera?.camera;
        if (cam) {
```

scripts/esm/gsplat/gsplat-shader-effect.mjs
```diff
@@ -198,9 +198,9 @@ class GsplatShaderEffect extends Script {
...
             // Specific camera specified via attribute
-            targetCameras = [this.camera.camera];
+            targetCameras = [this.camera.camera.camera];
```
willeastcott:
```suggestion
            targetCameras = [cam];
```

PR #8121, opened 2025-11-12
Workaround for packHalf2x16 precision issues on some Android devices

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/7993
It fixes issue introduced here https://github.com/playcanvas/engine/pull/8055 by swapping half formats to uint formats for better compatibility.

### Problem
Gaussian splatting rendering exhibited visible artifacts on Samsung Galaxy S24 (both WebGL and WebGPU), despite working correctly on most other devices (that we tested) The issue was caused by incorrect handling of half-precision subnormal values in the Adreno (TM) 750 GPU's built-in `packHalf2x16` / `pack2x16float` functions.

### Solution
Implemented a hybrid packing strategy that:
1. **Detects subnormal values** before packing (values where `|x| < 2^-14 ≈ 6.1e-5`)
2. **Uses software conversion** for problematic edge cases (correct IEEE 754 ties-to-even rounding)
3. **Falls back to hardware builtin** for normal-range values (99.9%+ of typical data)
4. **Only applies on Android** via platform-specific preprocessor defines

### Performance Impact
- **Negligible overhead**: Subnormals occur in <0.1% of typical normalized scene data
- **Desktop/iOS unaffected**: Fast hardware path used on all non-Android platforms

### Technical Details
- Created reusable half-float packing utilities with Android-specific workarounds:
- Added shader defines for runtime platform detection:
```
- `PLATFORM_DESKTOP`
- `PLATFORM_MOBILE`
- `PLATFORM_ANDROID`
- `PLATFORM_IOS`
```
### Future Considerations
The workaround can be easily removed if/when Android GPU drivers fix their subnormal handling. The `PLATFORM_ANDROID` conditional makes this a simple one-line change.

### DIFF
--- a/src/scene/shader-lib/glsl/chunks/internal/frag/packHalf.js
+++ b/src/scene/shader-lib/glsl/chunks/internal/frag/packHalf.js
@@ -4,97 +4,97 @@ export default /* glsl */`
 
 #if defined(PLATFORM_ANDROID)
 
-// Software pack of one f32 -> f16 (low 16 bits). Ties-to-even, full subnormals.
-uint floatToHalf(float a) {
-    uint u    = floatBitsToUint(a);
-    uint sign = (u >> 16u) & 0x8000u;
-    uint absu = u & 0x7FFFFFFFu;
-    uint man  = u & 0x007FFFFFu;
-    int  e32  = int((u >> 23u) & 0xFFu) - 127;
-    
-    // NaN / Inf
-    if ((absu & 0x7F800000u) == 0x7F800000u) {
-        bool isnan = (man != 0u);
-        return sign | (isnan ? 0x7E00u : 0x7C00u);
-    }
-    
-    // Overflow to Inf
-    if (e32 > 15) return sign | 0x7C00u;
-    
-    // Normal half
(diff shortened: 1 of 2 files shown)


src/platform/graphics/graphics-device.js
```diff
@@ -545,6 +545,12 @@ class GraphicsDevice extends EventHandler {
         if (this.textureFloatFilterable) capsDefines.set('CAPS_TEXTURE_FLOAT_FILTERABLE', '');
         if (this.textureFloatRenderable) capsDefines.set('CAPS_TEXTURE_FLOAT_RENDERABLE', '');
         if (this.supportsMultiDraw) capsDefines.set('CAPS_MULTI_DRAW', '');
+
+        // Platform defines
+        if (platform.desktop) capsDefines.set('PLATFORM_DESKTOP', '');
```
slimbuck, line 550: what's the point of this code?

mvaligursky: This generates defines that are automatically added to each shader - useful to do platform specific shader changes.
Here it's used to limit this precision workaround to android devices only.

PR #8130, opened 2025-11-14
Updated main jsdocs block for GSplatComponent

mvaligursky:
more up to date info

### DIFF
--- a/src/framework/components/gsplat/component.js
+++ b/src/framework/components/gsplat/component.js
@@ -17,7 +17,8 @@ import { GSplatPlacement } from '../../../scene/gsplat-unified/gsplat-placement.
 /**
  * The GSplatComponent enables an {@link Entity} to render 3D Gaussian Splats. Splats are always
  * loaded from {@link Asset}s rather than being created programmatically. The asset type is
- * `gsplat` which are in the `.ply` file format.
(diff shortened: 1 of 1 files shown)


src/framework/components/gsplat/component.js
```diff
@@ -38,11 +39,37 @@ import { GSplatPlacement } from '../../../scene/gsplat-unified/gsplat-placement.
...
+ *   eliminating visibility artifacts and popping effects when splat components overlap.
+ * - **LOD Streaming**: Dynamically loads and renders appropriate levels of detail based on camera
+ *   distance, enabling efficient rendering of massive splat scenes.
+ *
+ * ```javascript
+ * // Enable unified rendering for advanced features
+ * entity.gsplat.unified = true;
+ * ```
+ *
+ * Note: The `unified` property can only be changed when the component is disabled.
```
willeastcott, line 57: I know I mentioned this before, but this still seems super-weird to me. If you change it on an enabled component, we should just do whatever's necessary internally to make it work.

willeastcott (approved):

src/framework/components/gsplat/component.js
mvaligursky: agreed. but if the user already configured non-unified material and then switches to unified, we really don't have a good way to transfer this for example.

Issue #7925, opened 2025-08-25
Memory growth & WebGL context loss when streaming SOGS GSplat frame sequences (double-buffer, serial load)


Heeekdl567:
I’m trying to play a sequence of SOGS GSplat frames in the browser (to approximate “4D” by swapping 3D frames). I have 50 frames and I'm using a double buffer to get a smooth transition between the frames. But after a while I always get the following error in Chrome.

Do you have any idea how to fix that?

<img width="673" height="407" alt="Image" src="https://github.com/user-attachments/assets/8aae84a7-ec81-4c09-a2f9-9ec1a277659d" />

slimbuck:
Hi @Heeekdl567,

Every time a GS scene is loaded, the engine creates a sort web worker as well. The sorter is given the list of gaussian positions to sort and it looks like this allocation is resulting in OOM.

Can you check whether the number of webworkers in chrome devtools are growing as your progress through the animation? If so, then they're likely not being cleaned up correctly.

Thanks!

Heeekdl567:
Hi @slimbuck,

thanks for the pointer. We checked in Chrome DevTools and can confirm that the number of web workers keeps increasing as the sequence plays. They never drop back down across frame swaps, which lines up with the OOM/context-loss we’re seeing.

We tried to workaround this on our side with the attached scripts (double-buffer player; strict serial loading; deferred unloads; device-loss gating; only current + next SOGS in memory). However, we still hit this crash during SOGS loading:

Uncaught (in promise) TypeError: Cannot read properties of null (reading 'readTextureAsync')
    at webgl-texture.js:520
    at texture.js:437
    at gsplat-sogs-data.js:17 ($E / prepareGpuData)
    at gsplat-sogs-data.js:278 (prepareGpuData)
    at sogs.js:86 (loadTextures)


(see screenshot)

At this point we’re not sure how to fix it in our app code. A few questions:

Is there a supported way to explicitly dispose the GSplat sorter worker when a gsplat asset/instance is unloaded or swapped out? (e.g. an API like instance.sorter.terminate() or best practice we should follow)

Is this a known issue with sorter workers not being cleaned up? If so, is there a patch/branch we can test?

Any guidance on the correct unload sequence for gsplat/SOGS (order of detaching assets, waiting for sorter events, etc.) to guarantee worker teardown?

Thanks in advance!

[...]

PR #8141, opened 2025-11-18
Add GSplat flipbook animation script + related functionality + example

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/7925

Adds a reusable script for playing flipbook animations with GSplat assets, supporting multiple playback speeds and efficient asset management.

- Add gsplat-flipbook.mjs: Reusable script with configurable FPS, frame range, and playback modes (loop/bounce)
- Add flipbook.example.mjs: Demo showing multiple animations at different speeds
- Implement reference counting on GSplatResourceBase to prevent premature unloading during async sorting
- Fix AppBase.destroy() order to allow managers to clean up before asset unload
- Assets are shared across instances with automatic cleanup when no longer needed

Example usage: Attach script to entity with unified gsplat component, configure folder and frame range, animation plays automatically.

https://github.com/user-attachments/assets/7deac308-007d-4acc-b87e-16f028e3e0a6

### DIFF
--- a/examples/src/examples/gaussian-splatting/flipbook.example.mjs
+++ b/examples/src/examples/gaussian-splatting/flipbook.example.mjs
@@ -1,4 +1,5 @@
 // @config DESCRIPTION This example demonstrates gsplat flipbook animation using dynamically loaded splat sequences at different playback speeds.
+// @config HIDDEN
 import { deviceType, rootPath, fileImport } from 'examples/utils';
 import * as pc from 'playcanvas';
 


slimbuck (approved):
not sure we want all these assets in the examples repo. please check @willeastcott is happy with it before merging.

willeastcott:
I'm concerned that this example is encouraging the user to use the wrong technique to animate splats. The visuals in this particular dataset can be easily achieved via shaders. If we're going to show off flipbook, it should be for a dataset where there's no other option. And also a dataset that is aesthetically pleasing.

mvaligursky:
yep, communicated the sizes before. I'll wait for @willeastcott 's approval.

mvaligursky:
> I'm concerned that this example is encouraging the user to use the wrong technique to animate splats. The visuals in this particular dataset can be easily achieved via shaders. If we're going to show off flipbook, it should be for a dataset where there's no other option. And also a dataset that is aesthetically pleasing.

I agree. Happy to make the example 'not published' and leave it internal for now.

mvaligursky:
The example is no longer publuished.

mvaligursky:
@willeastcott - are you ok for me to merge this, with the example being private.

PR #8154, opened 2025-11-20
Picker class can (optional) render depth and provide world space pick points

mvaligursky:
## Depth Picking Support for Picker Class

Extends the `Picker` class to optionally capture depth information along with mesh IDs, enabling world position picking.

### New Features
- **Depth picking**: Picker can now render depth to a second color buffer using Multiple Render Targets (MRT)
- **World position picking**: New `getWorldPointAsync()` method converts screen coordinates to 3D world space
- **Per-point depth queries**: New `getPointDepthAsync()` method retrieves depth at a specific screen pixel

### Technical Details
- Depth encoded as `PIXELFORMAT_RGBA8` for cross-platform compatibility (WebGL2 & WebGPU)
- Uses `float2uint` encoding from `float-packing.js` to pack float depth into RGBA8
- New shader pass `SHADER_DEPTH_PICK` generates `DEPTH_PICK_PASS` define for shaders
- Shaders updated: `lit`, `gsplat` (both GLSL and WGSL)
- Fixed WebGL `GL_INVALID_OPERATION` issue with multisampled framebuffer reads by explicitly binding framebuffer in `readTextureAsync`

#### `Picker` class constructor
// BEFORE
new Picker(app, width, height)

// AFTER
new Picker(app, width, height, depth = false)- Added optional `depth` parameter (default: `false`)

#### New public methods
// Get world position at screen coordinates (returns null if no depth)
async getWorldPointAsync(x, y): Promise<Vec3 | null>

// New method
`Picker.destroy()` - Clean up resources

### Examples Updated
- `area-picker.example.mjs` - Demonstrates depth picking with world position visualization
- `picking.example.mjs` - Shows picking on gaussian splats with 3D marker placement

https://github.com/user-attachments/assets/dd8202bc-c776-4902-90ba-00d686c340b9

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -306,6 +306,7 @@ class Resource {
 - **Focus on user-facing changes**: What developers using the engine will see/use
 - **Be concise but complete**: Include all breaking changes and new APIs
 - **Avoid excessive detail**: Group related changes together, don't list every tiny implementation detail or internal refactoring
+- **Only document public APIs**: Do not list functionality tagged with `@ignore`, `@protected`, or `@private` as these are internal implementation details
 
 ## Resources
 


src/framework/graphics/picker.js
```diff
@@ -201,21 +405,45 @@ class Picker {
             magFilter: FILTER_NEAREST,
             addressU: ADDRESS_CLAMP_TO_EDGE,
             addressV: ADDRESS_CLAMP_TO_EDGE,
-            name: 'pick'
+            name: name
         });
+    }
+
+    allocateRenderTarget() {
```
slimbuck, line 412: It would be much more useful if applications were able to provide these buffers. Supersplat editor can't allocate a new set of buffers just for picking.

src/scene/shader-lib/wgsl/chunks/lit/frag/pass-other/litOtherMain.js
```diff
@@ -20,6 +20,9 @@ fn fragmentMain(input: FragmentInput) -> FragmentOutput {
 
     #ifdef PICK_PASS
         output.color = getPickOutput();
+        #ifdef DEPTH_PICK_PASS
```
slimbuck, line 23: is PICK_PASS and DEPTH_PICK_PASS both set?

src/framework/graphics/picker.js
mvaligursky: good possible future extension!

src/scene/shader-lib/wgsl/chunks/lit/frag/pass-other/litOtherMain.js
mvaligursky: yes

src/framework/graphics/picker.js
slimbuck: In general creating apis that allocate and manage large resources without application control or oversight is bad.

It works fine in a demo, but almost nowhere else.

slimbuck (approved):
I'm approving, but consider the picking API still woeful for actual use.

MAG-AdrianMeredith:
Oh I missed this! this will help a lot, we currently still rely on physics raycase for the world position which isn't so great at times

Issue #889, opened 2017-03-27
Rectangle - with position, orientation and size that can be intersected with Ray and other Shapes.


Maksims:
Like with BoundingBox and few other shapes, it would be good to have a rectangle, that is much simpler shape than box, and would be useful in many cases.

It would have position, rotation and size (not halfExtents please).
And would have functionality to check if it intersects with ray or other shapes.

leecommamichael:
Is this as straightforward as adding 'rectangle.js' in 'engine/src/shape' following the practices in the shapes in that directory? For clarification; you're asking for a rectangle, not a rectangular prism, yeah?

guycalledfrank:
I'd also add a circle.

>(not halfExtents please).
That would make it harder to switch between rect/existing aabb. I would keep the halfExtents approach (or change both).

Hitesh-Nagpure:
@Maksims 
I am new to contributing to open source projects. However, this is my approach in solving the issue:


Step-1: Creating a `rectangle.js` in the `src/core/shape` to define the `Rectangle {}` class

<img width="644" height="269" alt="Image" src="https://github.com/user-attachments/assets/e3dfc1ff-b8e1-43d5-8a15-f3753091427d" />

Step-2: Exporting `Rectangle {}` class in `src/index.js` 

<img width="939" height="289" alt="Image" src="https://github.com/user-attachments/assets/ba2a57a3-dddf-40d0-988b-081d5a39f2b5" />

Step-3: Adding `rectangle.test.js` file in `test/core/shape` to add four `Rectangle {}` class-specific test cases

<img width="1055" height="568" alt="Image" src="https://github.com/user-attachments/assets/0db9fdfa-12f8-4e85-aaa2-b25a65772357" />

Step-4: The Result -> As a result, 1580 + 4 (`Rectangle {}` class) test cases are passed.

<img width="794" height="426" alt="Image" src="https://github.com/user-attachments/assets/5fa75c0f-a596-4bcb-ae8b-be7e0dfec8fc" />

Can you assign me this issue?
Also, can I generate the pull request for the issue?

Hitesh-Nagpure:
@Maksims 
The Vercel deployment check fails due to missing authorization, since I’m an external contributor.
Please let me know if I need to do anything else.

Maksims:
Your PR will be picked up and reviewed by one of the PlayCanvas staff when they have time.
I'm an outside contributor just like yourself.

PR #8159, opened 2025-11-21
Adding Rectangle feature (#889)

Hitesh-Nagpure:
## Description
Successfully added the Rectangle feature (#889) by:
1) Creating a new `rectangle.js` file in the `src/core/shapes` folder.
2) Adding `export { Rectangle } from '../src/core/shape/rectangle.js';`  code at line number 104 in `index.js`.
3) Creating a new `rectangle.test.js` file in the `test/core/shapes` folder.

Fixes #
- Added a Rectangle feature in the repository as suggested in (#889)

## Checklist

### DIFF
--- a/src/core/shape/rectangle.js
+++ b/src/core/shape/rectangle.js
@@ -1,0 +1,86 @@
+import { Vec2 } from '../math/vec2.js';
+import { Ray } from '../shape/ray.js';
+
+class Rectangle {
+  constructor(position = new Vec2(), size = new Vec2(1, 1), rotation = 0) {
+    this.position = position;   // Vec2
+    this.size = size;           // Vec2 (width, height)
+    this.rotation = rotation;   // radians
+  }
+
+  // Get rotated corners of rectangle
+  getCorners() {
+    const hw = this.size.x / 2;
+    const hh = this.size.y / 2;
+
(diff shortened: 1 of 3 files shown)


src/core/shape/rectangle.js
```diff
@@ -0,0 +1,86 @@
...
+      }
+    }
+
+    return closestDist !== Infinity;
+  }
+
+  // Axis-Aligned rectangle overlap check (approximate for rotation = 0)
+  intersectsRectangle(other) {
+    if (this.rotation !== 0 || other.rotation !== 0) {
+      // For simplicity, rotation not supported in this example
```
LeXXik, line 67: In this example? Looks like the code is LLM generated.

LeXXik:
This PR does not fix #889. It is hardly useful without a rotation. Even then, most use cases can be solved with pc.Plane, so not sure if this is needed.

willeastcott:
I think most people know my attitude about adding new API. There has to be a very strong case to add new API, particularly classes. We shouldn't add something purely because it's fun or an LLM can solve an issue in 5 minutes. These controls are the reason the Engine API is reasonably compact, even after 15 years.

And ideally, new API should be accompanied by an Engine example that proves out the usefulness of the new API.

PR #8170, opened 2025-11-24
Update the device-cache system so its no longer relies on memory addresses

### LINKED ISSUE #8107: Referencing Device in cache Map can cause errors when the app goes behind a Proxy
I encountered this issue when using a WebComponent that embeds a PlayCanvas application inside a Vue project.  
The component instance is accessed through a Vue `$ref`, which is actually a Proxy object. 
When I call a method that alter the rendering process on the component from outside (for example to add a gizmo), the application fails to retrieve the GraphicsDevice from the internal cache. 
This happens because the Proxy changes the memory reference of the application object and all of its childrens, therefore the device lookup no longer matches the original cached reference. 
https://github.com/playcanvas/engine/blob/4257ba5db9a58b8e57cf1c3c9c3ee3390e6e69c5/src/platform/graphics/device-cache.js#L26-L27
It then try to call the `onCreate` argument, which is often `undefined`  ex : https://github.com/playcanvas/engine/blob/4257ba5db9a58b8e57cf1c3c9c3ee3390e6e69c5/src/scene/shader-lib/get-program-library.js#L20

**Possible fix** : I think changing the DeviceCache map from `Map<GraphicsDevice, any>` to `Map<string, any>` using the `device.canvas.id` as key could solve this issue. Maybe it has side effect on AR/VR devices?

d0rianb:
This PR attempts to fix #8107 when the app object is used  behind a proxy (ex: vue / solidjs).
The cache system use the `canvas.id` as key instead of the memory address of the object.
Maybe it would be useful to fallback on memory address if canvas has undefined id.

Is also add an assertion for to check if onCreate callback  is properly defined when no cached device exists.

### DIFF
--- a/src/platform/graphics/device-cache.js
+++ b/src/platform/graphics/device-cache.js
@@ -26,7 +26,7 @@ class DeviceCache {
     get(device, onCreate) {
         const cacheId = device?.canvas.id
         // Check if the device has a canvas, and if the canvas has a non null id
-        Debug.assert(cacheId, 'Canvas element should have a unique id')
+        Debug.assert(cacheId  || device.isNull, 'Canvas element should have a unique id')
 
         if (!this._cache.has(cacheId)) {
             Debug.assert(onCreate, 'No cached device has been found and create callback is invalid ')


mvaligursky:
The problem is that this is not a correct solution. We can have multiple devices using the same canvas. Also, when some GPU  (WebGL / WebGPU) data are created for a canvas.id, it does not mean they are compatible to be used for a different device using the same canvas.

I think we need to understand the root issue in the original issue. Is the device being silently re-created in your framework or similar? With correct behavior, I'm not sure why your issue is manifested, and we need to understand it before working on a fix.

mvaligursky:
Closing, see the issue for updates.

PR #8171, opened 2025-11-24
GSplat streaming lod - used memory optimization

mvaligursky:
Streaming LOD system uses dynamically loaded SOG format data. Internally, this is converted to a packed format for performance, but original data was left in memory for the case of lost context on the device. In that case, the original data was converted to packed format when the device was restored. This consumes memory which is only used in case of device lost.

This PR implements a solution, where those allocations are removed. When the context is lost, the octree is marked as unloaded, and when the context is restored, the octree reloads resources from url's as originally.

Used VRAM improvements in the streaming-lod example:
Screenshot 2025-11-24 at 09 48 05

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -256,6 +256,14 @@ class Resource {
 }
 ```
 
+### 17. Root Cause Analysis
+
+Always address the root cause of issues rather than implementing workarounds that hide or suppress problems:
+
+- **Identify the root cause**: When you encounter an error or unexpected behavior, investigate why it's happening
+- **Don't mask symptoms**: Avoid solutions that simply hide errors or suppress warnings without fixing the underlying issue
+- **Fix at the source**: When you identify the root cause, fix it where the problem originates, not where it manifests
+
 ## Things to Avoid
 
 ### 18. Anti-Patterns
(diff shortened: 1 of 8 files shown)


src/framework/parsers/sog-bundle.js
```diff
@@ -249,7 +249,11 @@ class SogBundleParser {
             });
 
             // construct the gsplat resource
+            const decompress = asset.data?.decompress;
+            const minimalMemory = asset.options?.minimalMemory ?? false;
```
slimbuck, line 253: Passing `minimalMemory` in `asset.options` doesn't match the existing flags `decompress` and `reorder` which are passed in `asset.data`.

I think the two existing flags should probably have used `asset.options`, not sure why I didn't - my bad.

Anyway it means things are now pretty confusing and I think we should move `decompress` and `reorder` flags to `asset.options` as well.

mvaligursky: yep, sounds like a separate PR kindof refactoring.

AGENTS.md
```diff
@@ -256,6 +256,14 @@ class Resource {
 }
 ```
 
+### 17. Root Cause Analysis
```
slimbuck, line 259: was this meant to be added in this PR?

mvaligursky: sure, it can be added in any PR as it gets adjusted I think. It's not part of the engine release, just some settings for agents.

fimbox:
This makes so much more space for streaming on mobile devices, thanks for this!

Issue #8048, opened 2025-10-10
Refraction attenuation renders wrong


animech-sijo:
PlayCanvas' implementation of refraction doesn't behave correctly. When using attenuation, the entire object gets a bright, milky white look. This doesn't behave as users expect, or how refraction works in the real world, and it doesn't match how glTF renderers renders refraction. Compare these images:

| glTF Sample Viewer      | PlayCanvas     |
| ------------- | ------------- |
| <img width="1456" height="1443" alt="Image" src="https://github.com/user-attachments/assets/c8bac55a-373e-4e04-80b1-2bc785a8993f" /> | <img width="1456" height="1443" alt="Image" src="https://github.com/user-attachments/assets/801a345a-2d93-44cd-bebc-a429ae81d8ef" /> |

Notice especially the 0.25 and 0.5 columns, where the objects are much brighter than expected. It feels like the shader is wrong, and not energy conserving. 

In fact, even just enabling attenuation with a white color (meaning no absorbtion), it immediately becomes milky. It doesn't matter what attenuation distance you set, as long as it's above 0.

| Attenuation distance 0      | Attenuation distance > 0     |
| ------------- | ------------- |
| <img width="1090" height="1182" alt="Image" src="https://github.com/user-attachments/assets/58906db1-00ad-47dd-a5f6-efed80d46e6c" /> | <img width="1090" height="1182" alt="Image" src="https://github.com/user-attachments/assets/cff93b1e-f831-461f-9595-01bdd386e2e7" /> |

We need this fixed to be able to create accurate glass materials that render correctly.

Test project: https://playcanvas.com/project/1405372/overview/attenuation
glTF sample model & viewer: https://github.khronos.org/glTF-Assets/model/AttenuationTest
Renderer comparisons: https://github.khronos.org/glTF-Render-Fidelity/model/AttenuationTest

Issue #6839, opened 2024-07-22
Possible issue with dynamic refraction


mvaligursky:
This can be reproduced in the `dispersion` example on WebGPU. When DISPLAYFORMAT_LDR_SRGB displaFormat is used, the prisms rendered with the dynamic reflection are darker than they are on WebGL. The reason is unclear at this stage, but some observations were made:
- if in evalRefractionColor function of refractionDynamic.js chunk, refraction is converted from linear to gamma space, this works as on WebGL. But as this should all work in linear space, it suggests there is a bug / workaround elsewhere in WebGL code perhaps, or some other issue is causing this.
- the test GLB contains this:
<img width="215" alt="Screenshot 2024-07-22 at 17 26 37" src="https://github.com/user-attachments/assets/4732a936-74ea-485a-9d28-872665ee6602">

and from a quick check, this should affect the refracted colors. Perhaps the issue is related to this somehow, but it's worth noting that all other webgl based glb viewers render it the same way as our WebGL implementation.

WebGL rendering
![Screenshot 2024-07-22 at 16 55 34](https://github.com/user-attachments/assets/d52b131c-aa9d-49d1-9658-b85018e1151a)

WebGPU rendering in DISPLAYFORMAT_LDR_SRGB mode:
![image](https://github.com/user-attachments/assets/c14208c8-7e8d-4623-ac9f-09c3741e7c33)

PR #8173, opened 2025-11-24
[Fix] Correct transmission attenuation rendering

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/8048, https://github.com/playcanvas/engine/issues/6839

Fixes transmission material attenuation rendering issues, making PlayCanvas match the glTF specification and reference implementations.

### Fixed attenuation transmittance calculation
- When attenuation distance is 0 (infinite), transmittance should be 1.0 (full transmission), not the refracted color
- This was causing the refraction color to be squared (multiplied by itself), producing incorrect milky white appearance

### Fixed scene color texture gamma handling
- Added `SCENE_COLORMAP_GAMMA` shader define when camera outputs gamma-corrected values to non-sRGB render targets
- Refraction shader conditionally applies `decodeGamma()` based on define to convert sampled colors to linear space

### Added attenuation test example
- New `examples/src/examples/test/attenuation.example.mjs` using AttenuationTest.glb

Original rendering
original

Fixed rendering
final

Reference implementation
reference

### DIFF
--- a/examples/src/examples/test/attenuation.example.mjs
+++ b/examples/src/examples/test/attenuation.example.mjs
@@ -96,4 +96,3 @@ assetListLoader.load(() => {
 });
 
 export { app };
-


mvaligursky:
> Do we really want an attenuation example? I feel like a gltf models 2.0 browser would probably be more useful.

its an internal (hidden) example, a useful test for now.
But agreed on the model browser, could do one day.

PR #8185, opened 2025-11-26
Fix incorrect time calculation in first two frames

### LINKED ISSUE #8183: Incorrect time calculation in makeTick
In the first 2 frames of the application lifecycle, time is wrongly calculated. This can be verified with the following script:

```ts
import { Script } from 'playcanvas';

export class TimeTest extends Script {

    static scriptName = 'timeTest';

    count = 10;

    /**
     * Called for enabled (running state) scripts on each tick.
     *
     * @param {number} dt - The delta time in seconds since the last frame.
     */
    update(dt) {
        if (this.app.frame > this.count) return;
        // BUG: Frame 2: this.app._time is somehow less than in frame 1??? ¯\_(ツ)_/¯
        console.log(this.app.frame, Number(this.app._time.toFixed(1)), Number(dt.toFixed(4)));
    }
}
```

Sample output (app._time in frame 2 is less than in frame 1):

| Frame | app._time | dt |
|--------|--------|--------|
| 1 | 2299.5 | 0 |
| 2 | 2288.6 | 0 |
| 3 | 2355.3 | 0.0667 |
| 4 | 2381.9 | 0.0266 |
| 5 | 2395.4 | 0.0135 | 
| 6 | 2408.7 | 0.0133 | 
| 7 | 2422 | 0.0133 | 

app._time is private API, but this has also effects on deltaTime, which will be 0 for the first 2 frames. (should only be 0 on first frame)

This happens because of the following calculation:
https://github.com/playcanvas/engine/blob/94166920315e2bbb0f8a57296062a2e550418c3d/src/framework/app-base.js#L2031-L2037

For frame 2 `ms` will be a negative value, because the timestamp passed to the makeTick method is earlier than application._time. This results in clamping the dt to 0.

### Proposed solution

Instead of calling the makeTick function manually in AppBase.start
https://github.com/playcanvas/engine/blob/94166920315e2bbb0f8a57296062a2e550418c3d/src/framework/app-base.js#L980
Queue it with requestFrameAnimation. Similar to how makeTick requeues
https://github.com/playcanvas/engine/blob/94166920315e2bbb0f8a57296062a2e550418c3d/src/framework/app-base.js#L2040-L2045

mvaligursky:
Fixes incorrect delta time calculation in the first two frames of the application lifecycle, where `dt` was 0 for both frames instead of just the first.

**Root Cause:**
- Frame 1: `start()` called `tick()` directly (synchronous), using `performance.now()` as fallback
- Frame 2: RAF provided a timestamp that could be earlier than Frame 1's time, causing negative `ms` value that got clamped to 0

**Solution:**
- Added `_requestAnimationFrame()` helper method to ensure all frames are queued via RAF with consistent timing
- Updated `start()`, `makeTick()`, and `XrManager` to use the helper instead of calling `tick()` directly or duplicating RAF code
- updated test to match other tests in that file that tick the update()

Fixes #8183

### DIFF
--- a/test/framework/components/script/component.test.mjs
+++ b/test/framework/components/script/component.test.mjs
@@ -1668,6 +1668,8 @@ describe('ScriptComponent', function () {
 
             app.start();
 
+            app.update();
+
             expect(window.initializeCalls.length).to.equal(8);
             let idx = 0;
             checkInitCall(e, idx++, 'initialize destroyer');


src/framework/xr/xr-manager.js
```diff
@@ -740,7 +740,7 @@ class XrManager extends EventHandler {
             // old requestAnimationFrame will never be triggered,
             // so queue up new tick
             if (this.app.systems) {
-                this.app.tick();
+                this.app._requestAnimationFrame();
```
willeastcott: Naughty. This is calling a private function.

lucaheft:
I tried fixing it the same way, however this fails 1 unit test. Not sure why , didn't invesitgate further.

Also I think the clamp can be substituted with:
`Math.min(dt, application.maxDeltaTime)`
https://github.com/playcanvas/engine/blob/d60ac3e7ba7431ff128971e95a415319c4d021aa/src/framework/app-base.js#L2048

dt should never be negative anymore.

mvaligursky:
> `Math.min(dt, application.maxDeltaTime)`

It's safer to clamp, browsers occasionally return strange time stamps.

Maksims:
This PR will lead to many apps now that have start-up animation, to be messed up.

We for example have camera on initialize to be calculated, and then animated to focus on the subject. Due to this PR, our animation will skip for likely 400ms, due to first two frame app usually is frozen for shader compilation and batching.

This will lead to start up animation that relies on dt - to be skipped, and it will not be clear to users why.

0 dt for 2 frames - is actually useful and been discussed few times years ago, where it was left at 0 intentionally.

This PR will negatively affect our applications, without clear benefit.

PR #8186, opened 2025-11-26
Add WebGPU support to instancing-custom example - custom WGSL shader

mvaligursky:
Fixes WebGPU rendering in the instancing-custom example by providing WGSL shader chunks alongside GLSL. 

The custom instancing shader code has been extracted to separate files following project conventions:
- instancing-custom.transform-instancing.glsl.vert
- instancing-custom.transform-instancing.wgsl.vert

### DIFF
--- a/src/scene/shader-lib/programs/lit-shader.js
+++ b/src/scene/shader-lib/programs/lit-shader.js
@@ -132,6 +132,7 @@ class LitShader {
             userChunkMap.forEach((chunk, chunkName) => {
 
                 // extract attribute names from the used chunk
+                Debug.assert(chunk);
                 for (const a in builtinAttributes) {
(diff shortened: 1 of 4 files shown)


src/scene/shader-lib/programs/lit-shader.js
```diff
@@ -132,6 +132,7 @@ class LitShader {
             userChunkMap.forEach((chunk, chunkName) => {
 
                 // extract attribute names from the used chunk
+                Debug.assert(chunk);
```
willeastcott, line 135: Delete?

mvaligursky: nope, added on purpose to catch the issue where the user overrides chunk with `undefined`, which otherwise triggers harder to understand crash later on.

PR #8188, opened 2025-11-27
Add init and configure event to AppBase

lucaheft:
## Description
Adds 2 new generic events to AppBase

init -> Fired after AppBase.init
configure -> Fired after successfull AppBase.configure

I've noticed, that there are no jsdocs for events in AppBase, if desired I can add those as well

If this gets merged we also want to reflect these changes on the developer site [Application Lifecycle](https://developer.playcanvas.com/user-manual/scripting/fundamentals/application-lifecycle/)

## Checklist


mvaligursky (commented):
How strange the events are not documented. I'll put on my list to document.

Please document yours though - see how these are done: https://github.com/playcanvas/engine/blob/25effe0e597e9b794a78d00a10a24baf9e60e10f/src/scene/scene.js#L32-L81

Maksims:
`configure` - seems generic in this context. Like application is configured, but with what and by whom?
Really it is something like: `config:loaded` because this is very specific to way Editor outputs the project and provides config.json with data in it. Engine only users usually don't use that approach. A bit more info in description of the event - will be useful.

lucaheft:
I see, not sure how to go forward with this. How about we only add init event for now?

Maksims:
> I see, not sure how to go forward with this. How about we only add init event for now?

I think if you name second event as `config:loaded` - should be fine.

lucaheft:
I just tried this out and it seems like when launching in the editor (use_local_engine), these events are not called. However when downloading they work correctly.

I guess there is some custom editor startup which doesn't call init and configure of the AppBase?
This could be quite confusing for the user.
Can this be changed in the editor, so the behaviour is similiar to the exported editor project?

In my test I subscribed to the events in the loadingscreen

PR #8190, opened 2025-11-28
Test using resolved msaa texture in compute shader

mvaligursky:
- do not merge, test only


mvaligursky:
added similar example to the engine: https://github.com/playcanvas/engine/pull/8192

PR #8203, opened 2025-11-30
Fixed unecessary axes array reallocation in GamePad update - Issue (#…

### LINKED ISSUE #842: game-pads.js allocates a new array every frame
```
        update: function (dt) {
            var pads = this.poll();

            var i, len = pads.length;
            for (i = 0;i < len; i++) {
                this.previous[i] = this.current[i];
                this.current[i] = pads[i];
            }
        },
        poll: function () {
            var pads = [];
            if (this.gamepadsSupported) {
                var padDevices = navigator.getGamepads ? navigator.getGamepads() : navigator.webkitGetGamepads();
                var i, len = padDevices.length;
                for (i = 0; i < len; i++) {
                    if (padDevices[i]) {
                        pads.push({
                            map: this.getMap(padDevices[i]),
                            pad: padDevices[i]
                        });
                    }
                }
            }
            return pads;
        },
```
When poll() gets called, it allocates a new array to store the current gamepad state.

kartik815:
# Optimize Receiving Objects Step to Reduce Initialization Delay (#842)

## Description
This PR addresses issue #842 by improving the performance of the receiving objects step in the engine.  
It reduces the delay when processing incoming objects, ensuring smoother initialization and better overall responsiveness.

Fixes #842

## Checklist


src/platform/input/game-pads.js
```diff
@@ -877,7 +884,7 @@ class GamePads extends EventHandler {
             }
 
             for (let j = 0, m = buttons.length; j < m; j++) {
-                const button = buttons[i];
+                const button = buttons[j];
```
slimbuck, line 899: did you mean to change  i -> j?

kartik815: Yes — this change was intentional.
The inner loop iterates using j, but the original code accessed buttons[i].
That caused incorrect indexing since i belongs to the outer loop.

This was a small existing bug that I fixed while making the main change (in-place axes updates).
If you'd prefer, I can move this fix into a separate PR

mvaligursky:
Thanks for the PR! A few things to address before this can land:

1. **Correctness regression in the axes copy (blocker).** The new code resizes `axes` *before* snapshotting it into `previousAxes`. If `gamepad.axes.length` ever differs from the previous `axes.length`, `previousAxes` ends up with `undefined`s (when growing) or silently drops tail values (when shrinking). Please snapshot `axes` → `previousAxes` using the *old* length first, then resize and copy `gamepad.axes` → `axes`.

2. **Style / lint.** `for(let` → `for (let`, add spaces around `<`, use the cached-length idiom (`for (let i = 0, l = arr.length; i < l; i++)`) like the rest of the file, drop the double blank line, and reword the misleading `// resizing array if size changed` comment (the resize is unconditional). `npm run lint` will flag some of these.

3. **PR title/description** look copy-pasted from a different template and don't describe what this PR actually does. Please update them to cover the gamepad axes change and also mention the unrelated `buttons[i]` → `buttons[j]` fix in `get previous()` so it isn't missed.

willeastcott:
Thanks for taking a look at this. #8203 went quiet after the review above and has since fallen into conflict, so I've opened #9470 which covers the same axes copy (with the previous-axes snapshot taken at the old length), fixes the actual per-frame `[]` allocation in `GamePads#update()` that #842 reported, and adds tests. The `get previous()` typo you caught was already removed from main. Closing this in favour of #9470.

PR #8221, opened 2025-12-04
Fix null values in asset cache for ESM scripts causing error

### LINKED ISSUE #8220: Null value in ScriptHandler cache for ESM scripts
– When using ESM scripts in a project, if you call `app.destroy()`, an error is thrown inside `scriptHandler?.clearCache();` called from `app.destroy()`.
– The error is that `element`, which is the value of `this._cache[key]`, where key is an ESM script id, is null.
– This value is set to null when loading ESM scripts on app start: in the `_loaded` function inside `AssetRegistry.load(asset, options)`, `handler_cache[asset.id]` is set to the value of `extra`, which is null for ESM scripts.

snave333:
## Description
This PR adds a check whether or not a loaded script asset is an ESM script. If so, it does not set the value of the `extra` variable in `AssetRegistry::loaded::_loaded` (which is null for ESM scripts) as the value in `ScriptHandler._cache[asset.id]`. This avoids an uncaught error later when calling `app.destroy()`, which calls `scriptHandler?.clearCache();`.

Fixes #8220

## Checklist

### DIFF
--- a/src/framework/asset/asset-registry.js
+++ b/src/framework/asset/asset-registry.js
@@ -467,8 +467,7 @@ class AssetRegistry extends EventHandler {
                         document.head.removeChild(handler._cache[asset.id]);
                     }
                     // prevents setting a null value in cache for esm scripts
-                    const isEsmScript = asset.file.filename.endsWith('.mjs');
-                    if(!isEsmScript && extra) {
+                    if (extra) {
                         handler._cache[asset.id] = extra;
                     }
                 }


src/framework/asset/asset-registry.js
```diff
@@ -466,7 +466,11 @@ class AssetRegistry extends EventHandler {
                         // remove old element
                         document.head.removeChild(handler._cache[asset.id]);
                     }
-                    handler._cache[asset.id] = extra;
+                    // prevents setting a null value in cache for esm scripts
+                    const isEsmScript = asset.file.filename.endsWith('.mjs');
+                    if(!isEsmScript && extra) {
```
mvaligursky: I think only this is needed:
```suggestion
                    if (extra) {
```

mvaligursky:
Awesome, thanks for the issue and the PR!

mvaligursky (approved):

PR #8232, opened 2025-12-07
Simple gsplat viewer engine example with HDR controls

mvaligursky:
- a simple test viewer for .sog or .ply files
- rendering in linear HDR, exposure and tonemapping at the end - useful for HDR data testing

https://github.com/user-attachments/assets/db0127f6-0c1e-44a5-8013-24ac88cff05c


willeastcott (approved):
Approving, but couldn't we set the thumbnail to something a bit nicer?

PR #8264, opened 2025-12-19
[FIX] Fix ScriptComponent clone not handling awaiting scripts

### LINKED ISSUE #2796: ScriptComponent clone awaiting
https://github.com/playcanvas/engine/blob/bafd41bc227ec45524c3e3085a6de4ace441e967/src/framework/components/script/system.js#L121-L125
`key.awaiting` is always false.

willeastcott:
### Description

When cloning an entity with a ScriptComponent, scripts in an "awaiting" state were not being properly included in the clone's script order.

### Bug

The `cloneComponent` method was iterating over `_scriptsIndex` using `for...in`, but incorrectly checking properties on the string key instead of the script entry object:

```js
for (const key in entity.script._scriptsIndex) {
    if (key.awaiting) {  // ❌ key is a string, key.awaiting is always undefined
        order.splice(key.ind, 0, key);
    }
}
```

Since `key` is a string (the property name), `key.awaiting` was always `undefined` (falsy), making this code block effectively dead code.

### Fix

Access the actual script entry object stored at that key:

```js
for (const key in entity.script._scriptsIndex) {
    const scriptEntry = entity.script._scriptsIndex[key];
    if (scriptEntry.awaiting) {  // ✅ correctly checks the script entry
        order.splice(scriptEntry.ind, 0, key);
    }
}
```

Fixes #2796

## Checklist

### DIFF
--- a/src/framework/components/script/system.js
+++ b/src/framework/components/script/system.js
@@ -121,8 +121,9 @@ class ScriptComponentSystem extends ComponentSystem {
         }
 
         for (const key in entity.script._scriptsIndex) {
-            if (key.awaiting) {
-                order.splice(key.ind, 0, key);
+            const scriptEntry = entity.script._scriptsIndex[key];
+            if (scriptEntry.awaiting) {
+                order.splice(scriptEntry.ind, 0, key);
             }
         }
 


mvaligursky (approved):

mvaligursky:
Closing in favour of #9200.

The diagnosis here is exactly right, but the one-line dereference is not enough on its own — it makes `entity.clone()` throw. Restoring the name into `order` without also adding it to `scripts` means `initializeComponentData` hits:

```js
component.create(data.order[i], {
    enabled: data.scripts[data.order[i]].enabled,   // undefined.enabled
```

```
TypeError: Cannot read properties of undefined (reading 'enabled')
```

which is a worse regression than the silent drop it fixes. #9200 adds the `scripts` entry (from `_scriptsData`, the same place the deferred registry sweep reads the attributes from), plus a running offset on the splice — the stored `ind` is the count of *instantiated* scripts at declaration time, so consecutive awaiting scripts share an index and splicing each at the raw value reverses them. It comes with tests covering all three.

Thanks @willeastcott — the analysis in this PR is what pointed at the rest of it.

PR #8286, opened 2025-12-20
[FIX] Document some missing pixel formats

### LINKED ISSUE #7498: Should PIXELFORMAT_A8, PIXELFORMAT_L8 and PIXELFORMAT_LA8 be public API?
Currently, they are undocumented and so do not make it into the public API reference:

https://github.com/playcanvas/engine/blob/e98943192f03e0bf84f74d5dde8821411084fabe/src/platform/graphics/constants.js#L530-L532

This is despite `PIXELFORMAT_LA8` being referenced here:

https://api.playcanvas.com/engine/classes/XrView.html#texturedepth
https://api.playcanvas.com/engine/classes/XrViews.html#depthpixelformat

willeastcott:
Adds JSDoc documentation to three pixel format constants that were previously undocumented and therefore missing from the public API reference.

### Changes

- Added JSDoc comments with `@category Graphics` to:
  - `PIXELFORMAT_A8` - 8-bit alpha
  - `PIXELFORMAT_L8` - 8-bit luminance (grayscale)
  - `PIXELFORMAT_LA8` - 8-bit luminance with 8-bit alpha

### Why

These constants were already exported and used internally, but `PIXELFORMAT_LA8` was also referenced in the public API documentation for `XrView.texturedepth` and `XrViews.depthPixelFormat` without being documented itself. This caused broken links in the API reference.

Fixes #7498

## Checklist

### DIFF
--- a/src/platform/graphics/constants.js
+++ b/src/platform/graphics/constants.js
@@ -471,8 +471,25 @@ export const INDEXFORMAT_UINT32 = 2;
  */
 export const indexFormatByteSize = [1, 2, 4];
 
+/**
+ * 8-bit alpha.
+ *
+ * @category Graphics
+ */
 export const PIXELFORMAT_A8 = 0;
+
+/**
+ * 8-bit luminance (grayscale).
+ *
+ * @category Graphics
+ */
 export const PIXELFORMAT_L8 = 1;
+
+/**
+ * 8-bit luminance with 8-bit alpha.
+ *
+ * @category Graphics
+ */
 export const PIXELFORMAT_LA8 = 2;
 
 /**


mvaligursky:
I know I removed them for WebGL2 (v2 engine) as there are more modern R channal texture formats that provide the same data, and are compatible with WebGPU as well. These are not, So I'm hoping maybe there is a way to not use these anymore, I think only XR uses those, perhaps the docs could use the replacement formats, but not sure. Maybe we need to check the WebGPU XR spec to see what formats are used there.

willeastcott:
Closing since @mvaligursky is not keen to expose these formats and may take a different approach.

PR #8306, opened 2025-12-22
Make Texture.read() a public API

mvaligursky:
# Make Texture.read() a public API

## Overview

Completes the `Texture.read()` API implementation and makes it public. This method allows downloading texture data from GPU memory to local memory asynchronously.

## Changes

- Add `mipLevel` option support to WebGL backend
- Add `immediate` option support to WebGL backend (calls `gl.flush()` to expedite GPU command submission)
- Remove `@ignore` tag to make the API public

## API

```
texture.read(x, y, width, height, {
    mipLevel: 0,        // which mip level to read
    face: 0,            // which cubemap face to read
    data: null,         // optional pre-allocated buffer
    immediate: false,   // flush commands immediately
    renderTarget: null  // optimization for WebGL (reuse RT)
}).then(data => {
    // data is Uint8Array, Float32Array, etc. based on texture format
});
```

### DIFF
--- a/src/platform/graphics/texture.js
+++ b/src/platform/graphics/texture.js
@@ -1159,8 +1159,6 @@ class Texture {
     /**
      * Download the textures data from the graphics memory to the local memory.
      *
-     * Note a public API yet, as not all options are implemented on all platforms.
-     *
      * @param {number} x - The left edge of the rectangle.
      * @param {number} y - The top edge of the rectangle.
      * @param {number} width - The width of the rectangle.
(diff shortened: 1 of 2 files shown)


Maksims:
If it is immediate, will it call then callback in sync?
````js
// 1
texture.read(x, y, width, height, {
    immediate: true
}).then((data) => {
    // 2
});
// 3
````

**Comment by mvaligursky** [comment]
> If it is immediate, will it call then callback in sync?
> 
> ```js
> // 1
> texture.read(x, y, width, height, {
>     immediate: true
> }).then((data) => {
>     // 2
> });
> // 3
> ```

No. It might just resolve faster, as it will be sent to gpu right away before other commands already prepared to be executed. See this:
```
     * @param {boolean} [options.immediate] - If true, the read operation will be executed as soon as
     * possible. This has a performance impact, so it should be used only when necessary. Defaults
     * to false.
```

Maksims:
So it is not really "immediate" but `"blocking": false` - by default, and blocking path can be used that potentially speeds up the result availability.

mvaligursky:
it's immediate scheduling and execution to run on gpu
not immediate results

none of these options are blocking.

Issue #7872, opened 2025-07-31
Add support for Reading and Writing to texture mip levels in Compute Shaders


MAG-AdrianMeredith:
Hi Team,  As part of my investigation into occlusion culling I have a need to make a HZB (Hierarchical Z Buffer).  I am using a compute shader for this, however in order to do it properly I've need to modify playcanvas in order to create a "TextureView" that can be passed into the compute shader so that I can read from one mip level and write to the other.  

I'm not sure about the best way to "create" the view, as right now you need to do it on the webgpu texture itself which doesn't feel quite right.  Thoughts?

in My Code: 
```
   // create TextureView from texture (WebgpuTexture)
  const mip0View = hzbTextureImpl.createViewForMip(0);
// pass the texture view into setParamer, not the texture
 this.depthCopyCompute.setParameter('destTexture', mip0View);

```
In Playcanvas:
webgpu-texture.js
// A simple class to represent a specific view of a texture
```
class WebgpuTextureView {
    /**
     * @param {WebgpuTexture} gpuTexture - The texture to create a view for.
     * @param {GPUTextureViewDescriptor} viewDescr - The descriptor for the texture view.
     */
    constructor(gpuTexture, viewDescr) {
        this.impl = gpuTexture;
        this.view = gpuTexture.createView(viewDescr);
    }
}

...

    createViewForMip(mipLevel) {
        if (mipLevel < 0 || mipLevel > this.texture.numLevels) {
            Debug.error(`Invalid mip level ${mipLevel} for texture ${this.texture.name} with ${this.desc.mipLevelCount} mip levels`);
            return null;
        }
        const viewDescr = {
            ...this.desc,
            baseMipLevel: mipLevel,
            mipLevelCount: 1
        };

        // create a texture view for the specified mip level
        const view = new WebgpuTextureView(this, viewDescr);
        this.mipViews[mipLevel] = view;
        return view;
    }
}

export { WebgpuTexture, WebgpuTextureView };

```

Then in WengpuBindGroup, check to see if the texture is an instance of TextureView and use that instead of the texture
(webgpu-bind-group.js line 80)
```
        // textures
        const textureFormats = bindGroup.format.textureFormats;
        bindGroup.textures.forEach((tex, textureIndex) => {

            /** @type {WebgpuTexture} */
            const wgpuTexture = tex.impl;
            const textureFormat = format.textureFormats[textureIndex];
            const slot = textureFormats[textureIndex].slot;
            let view;
[...]

PR #8308, opened 2025-12-22
 Add TextureView for mip level and array layer access in compute shaders

mvaligursky:
Fixes https://github.com/playcanvas/engine/issues/7872

Adds support for binding specific mip levels and array layers of textures in compute shaders via a new `TextureView` class.

### Features

- New `TextureView` class representing a view into a subset of a texture (WebGPU only)
- `Texture.getView()` factory method for creating views
- Support for specifying `baseMipLevel`, `mipLevelCount`, `baseArrayLayer`, and `arrayLayerCount`

### Usage

```
// Create views for specific mip levels
const mip0View = texture.getView(0);
const mip1View = texture.getView(1);

// Use with compute shader (e.g., HZB generation)
compute.setParameter('srcTexture', mip0View);
compute.setParameter('destTexture', mip1View);
```

### DIFF
--- a/src/platform/graphics/texture-view.js
+++ b/src/platform/graphics/texture-view.js
@@ -18,34 +18,44 @@ const stringIds = new StringIds();
  */
 class TextureView {
     /**
+     * The texture this view references.
+     *
      * @type {Texture}
-     * @private
+     * @readonly
      */
-    _texture;
+    texture;
 
     /**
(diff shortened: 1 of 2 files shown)


MAG-AdrianMeredith:
I totally missed this (just returning to this work), thank you Martin!

PR #8320, opened 2025-12-25
[FIX] Add mip level for WebGPU device copyRenderTarget

AlexAPPi:
This PR fixes the `RenderTarget` copy process by adding proper `mipLevel` handling in the `copyRenderTarget` method.

Previously, the method always copied only the base mip level (`mipLevel = 0`), which caused incorrect behavior when working with textures containing multiple LOD levels. Now the method correctly uses the `renderTarget.texture.mipLevel` from the source and copies data to the corresponding mip level of the destination render target.

**Changes:**
- Replaced the hardcoded level 0 with the actual mip level from the source.

**Result:**
- Correct data copying between `RenderTarget` objects with multiple mipmap levels.
- Enables using `copyRenderTarget` for mipchain generation and updates.

## Checklist


src/platform/graphics/webgpu/webgpu-graphics-device.js
mvaligursky: please also change it here and bellow

AlexAPPi: Yes, the engine doesn't currently support mipmaps for depth textures, but let's take that into account in our approach here.

PR #8323, opened 2025-12-28
Refactor Annotation System with AnnotationManager

willeastcott:
This PR refactors the annotation system to introduce an `AnnotationManager` that handles all global configuration and rendering resources, while `Annotation` becomes a lightweight data-only script. The two classes are completely decoupled and communicate via app-level events.

https://github.com/user-attachments/assets/9fd7b09e-5808-46e7-9259-b4abbe6ffa12

### AnnotationManager

A new manager script that handles:
- **Global configuration** with Editor-configurable attributes:
  - `hotspotSize` - Size of hotspots in screen pixels
  - `hotspotColor` - Default color of hotspots
  - `hoverColor` - Color when hovered
  - `opacity` - Opacity when visible
  - `behindOpacity` - Opacity when behind geometry
- **Shared resources**: CSS stylesheet, rendering layers, shared mesh, tooltip DOM
- **Per-annotation resources**: Entities, materials, textures, DOM elements
- **Interactions**: Hover states, click handling, tooltip display

The manager listens for app-level events:
- `annotation:add` - Automatically registers annotations and creates their rendering resources
- `annotation:remove` - Cleans up resources when annotations are destroyed

### Annotation

Simplified to a pure data class with three attributes:
- `label` - Short text on the hotspot (e.g., "1", "A")
- `title` - Tooltip title
- `text` - Tooltip description

Fires app-level events in `postInitialize()` to ensure the manager is ready:
- `annotation:add` - When created
- `annotation:remove` - When destroyed

Also fires script-level events for the manager to handle attribute changes:
- `label:set`, `title:set`, `text:set`

### Decoupled Architecture
- `Annotation` has **zero knowledge** of `AnnotationManager`
- Communication happens entirely through app-level events
- Manager subscribes to individual annotation events after registration

### Event Flow
```
AnnotationManager.initialize()
    → Listens for 'annotation:add' on app

Annotation.postInitialize()
    → Fires 'annotation:add' on app

AnnotationManager._registerAnnotation()
    → Creates texture, materials, entities, DOM
    → Subscribes to annotation's label:set, title:set, text:set events
```

### Resource Management
- All rendering resources are created and owned by the manager
- Clean separation: Annotation is just data, Manager handles visuals
- Proper cleanup on destroy for both manager and individual annotations

## Updated Example

The annotations example has been updated to demonstrate the new architecture:
- Creates `AnnotationManager` on the bicycle entity
- Annotations are children of the bicycle, so they move with it
- Includes control panel for adjusting all manager properties:
  - Hotspot size slider
  - Hotspot color picker
  - Hover color picker
  - Opacity sliders

### AnnotationManager

```javascript
// Create manager
const manager = entity.script.create(AnnotationManager, {
    properties: {
        hotspotSize: 30,
        hotspotColor: new Color(1, 1, 1),
        hoverColor: new Color(1, 0.5, 0),
        opacity: 1,
        behindOpacity: 0.25
    }
});

// Update at runtime
manager.hotspotSize = 40;
manager.hotspotColor = new Color(0.9, 0.9, 0.9);
```

### Annotation

```javascript
// Create annotation - automatically registered with manager via events
const annotation = entity.script.create(Annotation, {
    properties: {
        label: '1',
        title: 'Point of Interest',
        text: 'Description text here'
    }
});

// Update at runtime - manager handles visual updates
annotation.label = 'A';
annotation.title = 'New Title';
```

## Checklist

### DIFF
--- a/scripts/esm/annotation.mjs
+++ b/scripts/esm/annotation.mjs
@@ -1046,7 +1046,6 @@ export class Annotation extends Script {
         // Clean up on destroy
         this.once('destroy', () => {
             this.app.fire('annotation:remove', this);
-            this.off();
         });
     }
 }


scripts/esm/annotation.mjs
```diff
@@ -381,239 +455,598 @@ export class Annotation extends Script {
...
+        hotspotDom.addEventListener('wheel', onWheel);
 
-        Annotation.parentDom.appendChild(this.hotspotDom);
+        // Document click to dismiss tooltip
+        const onDocumentPointerDown = () => {
+            if (this._activeAnnotation === annotation) {
+                this._hideTooltip(annotation);
+            }
+        };
+        document.addEventListener('pointerdown', onDocumentPointerDown);
```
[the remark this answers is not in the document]
willeastcott: Fixed.

scripts/esm/annotation.mjs
```diff
@@ -381,239 +455,598 @@ export class Annotation extends Script {
...
+     * @param {Annotation} annotation - The annotation that was active
+     * @private
+     */
+    _hideTooltip(annotation) {
+        this._activeAnnotation = null;
+        this._tooltipDom.style.opacity = '0';
+
+        // Wait for fade out before hiding
+        setTimeout(() => {
+            if (this._tooltipDom && this._tooltipDom.style.opacity === '0') {
```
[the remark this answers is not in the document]
willeastcott: Fixed.

scripts/esm/annotation.mjs
```diff
@@ -381,239 +455,597 @@ export class Annotation extends Script {
...
+     * Called after all scripts have initialized, ensuring the AnnotationManager
+     * is ready to receive the annotation:add event.
      */
-    _calculateScreenSpaceScale() {
-        const cameraPos = Annotation.camera.getPosition();
-        const toAnnotation = this.entity.getPosition().clone().sub(cameraPos);
-        const distance = toAnnotation.length();
+    postInitialize() {
+        // Notify any listeners that this annotation has been created
+        this.app.fire('annotation:add', this);
```
slimbuck, line 1044: Might be more flexible if annotations searched up the tree for the first AnnotationManager and used that instead?

willeastcott: This is a philosophical preference, I guess, but generally I like scripts to be fully decoupled, not make assumptions and communicate using application events where possible. Forcing the manager to be an ancestor in the hierarchy is an unnecessary restriction to impose IMHO.

slimbuck: Maybe, but it means you can have more than one manager. Not strongly advocating though.

willeastcott: That's fair. I can address in a follow up PR.

slimbuck: I take it back. Using the scene hierarchy probably isn't ideal for assigning managers, because then a single object can't two different looking annotations. Perhaps just adding a manager name or similar to the existing add event could work better?

willeastcott: That's an option. Needs brainstorming really. This is really just a baby step in making the annotation system better - let's more we can do to evolve it further.

PR #8326, opened 2025-12-30
Examples mobile orientation canvas resize

kpal81xd:
## Description
- Resizes canvas when orientation changes (mobile)

Fixes #

## Checklist

### DIFF
--- a/examples/src/examples/animation/blend-trees-1d.example.mjs
+++ b/examples/src/examples/animation/blend-trees-1d.example.mjs
@@ -52,8 +52,10 @@ app.setCanvasResolution(pc.RESOLUTION_AUTO);
 // Ensure canvas is resized when window changes size
 const resize = () => app.resizeCanvas();
 window.addEventListener('resize', resize);
+window.addEventListener('orientationchange', resize);
 app.on('destroy', () => {
     window.removeEventListener('resize', resize);
+    window.removeEventListener('orientationchange', resize);
 });
 
 const assetListLoader = new pc.AssetListLoader(Object.values(assets), app.assets);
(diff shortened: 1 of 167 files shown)


willeastcott:
Am I right in saying the PlayCanvas Editor app boilerplate code doesn't listen for orientationchange? Was wondering why it's needed here. (AI generated PR descriptions will normally cover this type of thing for you).

kpal81xd:
> Am I right in saying the PlayCanvas Editor app boilerplate code doesn't listen for orientationchange? Was wondering why it's needed here. (AI generated PR descriptions will normally cover this type of thing for you).

Found that sometimes the canvas does not resize for me on iPhone Chrome so have to manually check for orietnation. Will investigate first.

Maksims:
The reason why orientation change is not reliable, is that different OS's and browsers can report pre-event size of the window withing orientation change event. For that reason you cannot rely on it. Also, after orientation change, some OS's browsers, have extra animations of panels sliding off the screen, which lead to window size changes, but of course will not be reported by orientation change event.

The most reliable, is an additional timed (setTimeout/rAF) check if window.innerWidth/innerHeight has changed vs previous sizes.

kpal81xd:
> The reason why orientation change is not reliable, is that different OS's and browsers can report pre-event size of the window withing orientation change event. For that reason you cannot rely on it. Also, after orientation change, some OS's browsers, have extra animations of panels sliding off the screen, which lead to window size changes, but of course will not be reported by orientation change event.
> 
> The most reliable, is an additional timed (setTimeout/rAF) check if window.innerWidth/innerHeight has changed vs previous sizes.

That is a good point i have noticed that too acc while doing some testing. Ill close this PR then

PR #8329, opened 2026-01-03
[FIX] Fix glTF node matrix decomposition with negative scale

### LINKED ISSUE #8328: Node with transform doesn't scale properly
When I open a model ([link](https://playcanvas.com/viewer/?load=https://jochem.cc/scale.glb)) in the PlayCanvas Viewer that scales the root node along the X-axis by using the `scale` property it's displayed correctly:
image

When I modify the model ([link](https://playcanvas.com/viewer/?load=https://jochem.cc/matrix.glb)) to use the `matrix` property instead of `scale`, the entire model is flipped along the vertical axis:
image

Both models validate without errors and are displayed correctly in almost every other viewer I tested (only Cesium had trouble with the normals, but this is a known issue).

willeastcott:
Fixes #8328

image

### Description

When a glTF node uses a `matrix` property (instead of separate `translation`, `rotation`, `scale` properties) that contains a negative scale component (creating a reflection/flip), the model was displayed incorrectly - appearing flipped or upside down.

### Changes

- Use `Quat.setFromMat4()` for rotation extraction instead of `Mat4.getEulerAngles()` when decomposing node matrices
- Apply `scaleSign` to the X scale component to preserve negative determinant (mirrored) transforms

### Technical Details

The root cause was that `Mat4.getScale()` returns only positive values (it computes vector lengths), and `Mat4.getEulerAngles()` doesn't coordinate with negative determinant handling. This could cause the flip to be absorbed into the rotation as an extra 180° rotation, or be lost entirely.

The fix uses `Quat.setFromMat4()` which already properly handles negative determinant matrices by normalizing the X axis before rotation extraction (see lines 589-597 in `quat.js`). The negative scale is then preserved by multiplying the X scale by `Mat4.scaleSign`, which returns -1 for mirrored matrices.

This approach is consistent with how the engine handles mirrored transforms elsewhere (e.g., batch manager, `GraphNode._worldScaleSign`).

## Checklist

### DIFF
--- a/src/framework/parsers/glb-parser.js
+++ b/src/framework/parsers/glb-parser.js
@@ -1517,6 +1517,7 @@ const createAnimation = (gltfAnimation, animationIndex, gltfAccessors, bufferVie
 
 const tempMat = new Mat4();
 const tempVec = new Vec3();
+const tempQuat = new Quat();
 
 const createNode = (gltfNode, nodeIndex, nodeInstancingMap) => {
     const entity = new GraphNode();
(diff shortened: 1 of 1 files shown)


willeastcott:
@mvaligursky I know this already has 1 approval but I think it's best if I get your approval too since you commented on the original issue. Let me know if you think this fix is correct.

PR #8334, opened 2026-01-05
Allow changing unified property without disabling component

mvaligursky:
Allow the `GSplatComponent.unified` property to be changed at runtime without requiring manual disable/enable of the component.

### Changes

- Modified `unified` setter to automatically destroy and recreate internal data (instance/placement) when the value changes
- Simplified global-sorting example to demonstrate direct property assignment

### Before
```
comp.enabled = false;
comp.entity.enabled = false;
comp.unified = true;
comp.enabled = true;
comp.entity.enabled = true;
```

### After
```
comp.unified = true;
```

### DIFF
--- a/src/framework/components/gsplat/component.js
+++ b/src/framework/components/gsplat/component.js
@@ -54,8 +54,6 @@ import { GSplatPlacement } from '../../../scene/gsplat-unified/gsplat-placement.
  * entity.gsplat.unified = true;
  * ```
  *
- * Note: The `unified` property can only be changed when the component is disabled.
- *
  * Relevant Engine API examples:
  *
  * - [Simple Splat Loading](https://playcanvas.github.io/#/gaussian-splatting/simple)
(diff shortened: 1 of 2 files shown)


src/framework/components/gsplat/component.js
```diff
@@ -439,21 +437,19 @@ class GSplatComponent extends Component {
     }
 
     /**
-     * Sets whether to use the unified gsplat rendering. Can be changed only when the component is
-     * not enabled. Default is false.
+     * Sets whether to use the unified gsplat rendering. Default is false.
+     *
+     * Note: Material handling differs between modes. When unified is false, use
+     * {@link GSplatComponent#material}. When unified is true, materials are shared per
+     * camera/layer - use {@link GSplatComponentSystem#getGSplatMaterial} instead.
```
willeastcott, line 444: I'm not a fan of this naming. I prefer something like `GSplatComponentSystem#getMaterial`.

willeastcott (approved):

src/framework/components/gsplat/component.js
mvaligursky: yep good point, will address separately.

PR #8338, opened 2026-01-06
WebGPU lightmapper support for shadow casting local lights and omni shadow shader cleanup

mvaligursky:
## Summary

Enables lightmapper to work correctly on WebGPU with shadow casting omni/spot lights and cleans up shadow shader handling for omni lights by properly separating shadow coordinate calculation based on light type.

### Lightmapper WebGPU Support

- Unified WebGL and WebGPU rendering paths to both use render passes, and this adds support for local lights on WebGPU
- Non-clustered shadow rendering now uses `RenderPassShadowLocalNonClustered` instead of deprecated `shadowRenderer.render()`

### Dead Code Removal

**`renderer.js`**
- Removed `setCamera()` and `clearView()` methods (only used by old lightmapper path)

**`shadow-renderer.js`**
- Removed `render()` method (replaced by render passes)
- Removed `insideRenderPass` parameter from `renderFace()` - now always assumes inside render pass

### Shader Cleanup - Omni Shadow Matrix

Omni lights use cubemap shadows that sample by direction, not matrix transformation. The `shadowMatrix` uniform was incorrectly handled.

**`lightFunctionShadow.js` (GLSL & WGSL)**
- Added `getShadowSampleCoordOmni{i}()` - dedicated function for omni lights without `shadowMatrix` parameter
- `getShadowSampleCoord{i}()` now only handles directional/spot lights
- `getShadow{i}()` dispatches to appropriate function based on light type

**`lightDeclaration.js` (GLSL & WGSL)**
- Shadow section: Skip declaring `shadowMatrix` for omni lights (not used)
- Cookie section (GLSL only): Always declare `shadowMatrix` for omni cookies (needed for cube orientation)

### Example Updates

**`lights-baked.example.mjs`**
- Update to be easily tell if something does not work, which was not possible with the original random positioning
- Added orbit camera with controls
- Added visual representations for lights (emissive sphere/cone/cylinder)
- Added UI controls for toggling lights and soft shadows
- Objects arranged in deterministic 8x8 grid

Screenshot 2026-01-06 at 11 39 40


examples/src/examples/graphics/lights-baked.controls.mjs
```diff
@@ -0,0 +1,53 @@
...
+                jsx(BooleanInput, {
+                    type: 'toggle',
+                    binding: new BindingTwoWay(),
+                    link: { observer, path: 'data.lights.directional' },
+                    value: observer.get('data.lights.directional')
+                })
+            ),
+            jsx(
+                LabelGroup,
+                { text: '    🟡 Soft' },
```
willeastcott, line 43: What're the additional spaces at the start of the label for?

examples/src/examples/graphics/lights-baked.example.mjs
```diff
@@ -1,9 +1,14 @@
-import { deviceType } from 'examples/utils';
+import { data } from 'examples/observer';
+import { deviceType, rootPath } from 'examples/utils';
 import * as pc from 'playcanvas';
 
 const canvas = /** @type {HTMLCanvasElement} */ (document.getElementById('application-canvas'));
 window.focus();
 
+const assets = {
+    script: new pc.Asset('script', 'script', { url: `${rootPath}/static/scripts/camera/orbit-camera.js` })
```
willeastcott, line 9: Can you use camera-controls.mjs (as I do in the new annotations example)?

examples/src/examples/graphics/lights-baked.controls.mjs
mvaligursky: it was supposed to be indented .. as it applies to directional light only, but it didn't work :)

PR #8343, opened 2026-01-07
Lazy mesh creation for GSplat resources

mvaligursky:
Defers mesh and instanceIndices allocation in `GSplatResourceBase` until actually needed by `GSplatInstance`, eliminating unnecessary GPU resource allocation when unified rendering mode is used.

### Changes

- `GSplatResourceBase` no longer creates mesh and instanceIndices in constructor
- Added `ensureMesh()` for lazy creation when `GSplatInstance` needs them
- Added `releaseMesh()` with ref counting to clean up instanceIndices when all instances are destroyed
- Mesh destruction is handled automatically by `MeshInstance` via its internal refCount

### Benefits

- Zero mesh/instanceIndices allocation when only unified mode is used
- Proper cleanup when switching from non-unified to unified mode
- Shared mesh across multiple instances of the same resource (existing behavior preserved)

### DIFF
--- a/src/scene/gsplat/gsplat-resource-base.js
+++ b/src/scene/gsplat/gsplat-resource-base.js
@@ -37,11 +37,11 @@ class GSplatResourceBase {
     /** @type {BoundingBox} */
     aabb;
 
-    /** @type {Mesh} */
-    mesh;
+    /** @type {Mesh|null} */
+    mesh = null;
 
-    /** @type {VertexBuffer} */
-    instanceIndices;
+    /** @type {VertexBuffer|null} */
+    instanceIndices = null;
 
     /** @type {number} */
     id = id++;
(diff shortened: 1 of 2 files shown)


src/scene/gsplat/gsplat-resource-base.js
```diff
@@ -114,6 +111,36 @@ class GSplatResourceBase {
         return this._refCount;
     }
 
+    /**
+     * Ensures mesh and instanceIndices exist. Creates them lazily on first call. Must be paired
+     * with a call to releaseMesh() when done.
+     *
+     * @ignore
+     */
+    ensureMesh() {
```
LeXXik, line 120: Perhaps .mesh could be a getter.

mvaligursky: it would make it less clear that under the hood it increments refCount, that needs to be decremented.

PR #8353, opened 2026-01-09
Small updates to the BlurredPlanarReflection script

mvaligursky:
- SceneColorMap instead of _sceneColorMap is exposed attribute
- heightRange is a value field and not a slider

### DIFF
--- a/scripts/esm/blurred-planar-reflection.mjs
+++ b/scripts/esm/blurred-planar-reflection.mjs
@@ -409,6 +409,7 @@ class BlurredPlanarReflection extends Script {
      * Height range for distance-based blur calculation.
      *
      * @attribute
+     * @range [0]
      */
     heightRange = 1;
 


scripts/esm/blurred-planar-reflection.mjs
```diff
@@ -409,9 +409,6 @@ class BlurredPlanarReflection extends Script {
      * Height range for distance-based blur calculation.
      *
      * @attribute
```
willeastcott, line 411: What about this to disallow negative numbers:
```js
    /**
     * Height range for distance-based blur calculation.
     *
     * @attribute
     * @range [0]
     */
    heightRange = 10;
```

willeastcott (approved):
Approving with a suggestion.

scripts/esm/blurred-planar-reflection.mjs
mvaligursky: great!

PR #8366, opened 2026-01-11
Add DrawCommands resize interface

### LINKED ISSUE #8345: Fix a memory leak when calling setMultiDraw
I noticed a memory leak when calling setMultiDraw several times in a row.
Allocate is always called, which results in the creation of a StorageBuffer while the previous buffer is not deleted.

https://github.com/playcanvas/engine/blob/d09fd9f600720164a880628b0c0fcd38ca4099c5/src/scene/mesh-instance.js#L1224
https://github.com/playcanvas/engine/blob/d09fd9f600720164a880628b0c0fcd38ca4099c5/src/platform/graphics/webgpu/webgpu-draw-commands.js#L42

**It would also be good if old data was saved and not reset then allocate.**

```
    /**
     * Allocate AoS buffer and backing storage buffer.
     * @param {number} maxCount - Number of sub-draws.
     */
    allocate(maxCount) {
        this.storage?.destroy(); // <--- DISPOSE EXISTS
        this.gpuIndirect = new Uint32Array(5 * maxCount);
        this.gpuIndirectSigned = new Int32Array(this.gpuIndirect.buffer);
        this.storage = new StorageBuffer(this.device, this.gpuIndirect.byteLength, BUFFERUSAGE_INDIRECT | BUFFERUSAGE_COPY_DST);
    }
```

AlexAPPi:
**Add `preserve` parameter to `setMultiDraw` for safe resizing**

Added a new optional `preserve` parameter to the `setMultiDraw` method, which preserves previous draw commands when changing the `maxCount` size.

### Changes
- **New parameter**: `preserve: boolean` (defaults to `false`)
  - `true` - copies existing draw commands when increasing/decreasing `maxCount`
  - `false` - clears commands on each call (previous behavior)
- Updated JSDoc documentation with parameter description

### Usage
```javascript
// Previous behavior (clears commands)
meshInstance.setMultiDraw(camera, 16);

// New behavior with command preservation
meshInstance.setMultiDraw(camera, 32, true); // Increases to 32, preserves old commands
meshInstance.setMultiDraw(camera, 8, true);  // Decreases to 8, preserves commands
```

### Benefits
- **Safe resizing** without data loss
- **Backward compatibility** - existing code works unchanged
- Useful for **dynamic object count** changes in the scene

***

Fixes #8345

## Checklist

### DIFF
--- a/src/platform/graphics/draw-commands.js
+++ b/src/platform/graphics/draw-commands.js
@@ -114,6 +114,18 @@ class DrawCommands {
     }
 
     /**
+     * Resize persistent storage for the draw commands.
+     *
+     * @param {number} maxCount - Maximum number of draw calls to allocate storage for.
+     * @param {boolean} preserve - Whether to copy previous draw commands.
+     * @ignore
+     */
+    resize(maxCount, preserve) {
+        this._maxCount = maxCount;
+        this.impl.resize?.(maxCount, preserve);
+    }
+
+    /**
      * Writes one draw command into the allocated storage.
      *
      * @param {number} i - Draw index to update.
--- a/src/platform/graphics/webgl/webgl-draw-commands.js
+++ b/src/platform/graphics/webgl/webgl-draw-commands.js
@@ -28,9 +28,35 @@ class WebglDrawCommands {
      * @param {number} maxCount - Number of sub-draws.
      */
     allocate(maxCount) {
-        this.glCounts = new Int32Array(maxCount);
-        this.glOffsetsBytes = new Int32Array(maxCount);
-        this.glInstanceCounts = new Int32Array(maxCount);
+        this.resize(maxCount, false);
+    }
+
+    /**
+     * Resize SoA arrays for multi-draw.
+     * @param {number} maxCount - Number of sub-draws.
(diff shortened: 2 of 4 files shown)


mvaligursky:
Hi @AlexAPPi - I'm not super keen on this, it creates potential complications I'd prefer to not have in a simple API:

There's a potential GPU timeline hazard with copying buffer content during resize.

`StorageBuffer.copy()` uses `commandEncoder.copyBufferToBuffer()` which only executes when the command buffer is submitted (typically at frame end). However, subsequent `write()` calls use `queue.writeBuffer()` which executes immediately when queued.

If a user resizes and then writes new data in the same frame:
1. `copyBufferToBuffer` is recorded (not executed)
2. `queue.writeBuffer` executes immediately
3. Command buffer submits at frame end → copy overwrites the new data

This could cause the preserved old data to overwrite newly written entries.

AlexAPPi:
@mvaligursky Yes, there are certain problems here, I would correct the description in your PR, since data is reset if a resize occurs.

mvaligursky:
That's a fast I do not want to be guaranteed by the API, as the implementation might change. As far as the user is concerned, when this function is called, the data is invalidated.

PR #8389, opened 2026-01-20
[FIX] Fix XrInputSource crash when grip unavailable

willeastcott:
## Summary

- Fixed `TypeError: Cannot read properties of null (reading 'setTRS')` crash in VR mode
- `getPosition()` and `getRotation()` now correctly return `null` when grip data is not yet available, as documented
- Added defensive null checks in `XrControllers` script

## Technical Details

The `getPosition()` and `getRotation()` methods had incorrect null guards:

```javascript
// Before (broken)
getPosition() {
    if (!this._position) return null;  // _position is always initialized, never null
    this._updateTransforms();          // crashes when _localTransform is null
}

// After (fixed)
getPosition() {
    if (!this._grip) return null;      // correctly checks if grip data exists
    this._updateTransforms();
}
```

The crash occurred when `getPosition()` was called before the first valid `gripPose` was received from WebXR. The `_localTransform` matrix is only created when `_grip` becomes true, but the old guard checked `_position` which is always a valid `Vec3`.

This aligns the implementation with the existing JSDoc contract: *"Get the world space position of input source if it is handheld (grip is true). Otherwise it will return null."*

## Checklist

### DIFF
--- a/scripts/esm/xr-controllers.mjs
+++ b/scripts/esm/xr-controllers.mjs
@@ -98,8 +98,10 @@ class XrControllers extends Script {
                         jointEntity.setRotation(joint.getRotation());
                     }
                 } else {
-                    entity.setPosition(inputSource.getPosition());
-                    entity.setRotation(inputSource.getRotation());
+                    const position = inputSource.getPosition();
+                    const rotation = inputSource.getRotation();
+                    if (position) entity.setPosition(position);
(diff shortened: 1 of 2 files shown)


Maksims:
Good catch! 🙏

slimbuck (approved):

PR #8414, opened 2026-01-28
Exposed the color band width parameter for radial reveal effect

mvaligursky:
- replaced hardcoded number by attribute / uniform

### DIFF
--- a/scripts/esm/gsplat/reveal-radial.mjs
+++ b/scripts/esm/gsplat/reveal-radial.mjs
@@ -11,6 +11,7 @@ uniform vec3 uDotTint;
 uniform vec3 uWaveTint;
 uniform float uOscillationIntensity;
 uniform float uEndRadius;
+uniform float uBandWidth;
 
 // Shared globals (initialized once per vertex)
 float g_dist;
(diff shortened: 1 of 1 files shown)


scripts/esm/gsplat/reveal-radial.mjs
```diff
@@ -37,7 +38,7 @@ void modifySplatCenter(inout vec3 center) {
     if (g_dist > uEndRadius) return;
     
     // Only apply oscillation if lift wave hasn't fully passed
-    bool wavesActive = g_liftTime <= 0.0 || g_dist > g_liftWavePos - 1.5;
+    bool wavesActive = g_liftTime <= 0.0 || g_dist > g_liftWavePos - 1.5 * uBandWidth;
```
LeXXik, line 41: That's just a `scalar * scalar `and can be probably shortened by simply `scalar` value that the uniform holds. Save a bit of ALU.

mvaligursky: sure, but there are two multiplies. 
tiny difference.

willeastcott (approved):

PR #8421, opened 2026-01-30
Unified GSplat Picking Support

mvaligursky:
## Summary

- Adds picking support for Gaussian Splats in unified rendering mode
- Introduces centralized `PickerId` class for consistent ID allocation across pickable objects
- Extends the `Picker` class to return `GSplatComponent` for unified gsplats alongside `MeshInstance` for other objects

### Centralized ID Generation

- New `PickerId` class (`src/scene/picker-id.js`) with static `get()` method for allocating unique IDs
- `MeshInstance` now uses `PickerId.get()` instead of internal counter
- `GSplatComponent` allocates its own ID at construction, passed to placement

### GSplat Unified Picking

- `GSplatParams.id` property enables per-component ID storage in the work buffer
- When enabled, a `pcId` stream (R32U format) is added to the work buffer format

### Picker Updates

- `Picker.getSelection()` and `getSelectionAsync()` now return `(MeshInstance | GSplatComponent)[]`

### Examples

- Updated `gaussian-splatting/picking.example.mjs` to demonstrate unified picking
- Created `gaussian-splatting-legacy/picking.example.mjs` for non-unified picking reference

### New: `GSplatParams.id`

```javascript
// Enable gsplat ID for unified picking
app.scene.gsplat.id = true;
```

Once enabled, cannot be disabled. The ID stream persists for the lifetime of the app.

### Updated: `Picker.getSelectionAsync()` / `Picker.getSelection()`

Return type changed from `MeshInstance[]` to `(MeshInstance | GSplatComponent)[]`:

```javascript
picker.getSelectionAsync(x, y, 1, 1).then((results) => {
    const picked = results[0];
    if (picked instanceof pc.GSplatComponent) {
        // Unified gsplat component
        console.log('Picked gsplat:', picked.entity.name);
    } else {
        // MeshInstance
        console.log('Picked mesh:', picked.node.name);
    }
});
```


examples/src/examples/gaussian-splatting/picking.example.mjs
```diff
@@ -81,6 +82,9 @@ assetListLoader.load(() => {
         });
     }
 
+    // Enable gsplat ID for unified picking
+    app.scene.gsplat.id = true;
```
willeastcott, line 86: Oh, I hadn't noticed this property before. Couple of things:

1. I still feel like configuration props for rendering gsplats should live on the `GSplatComponentSystem`.
2. `id` suggests to me it's an actual ID (e.g. a GUID or number). So naming feels weird. Something like `enableIds` would be better IMHO.

mvaligursky: 1. perhaps.
2. will do, great point.

PR #8441, opened 2026-02-04
Gaussian Splatting: Use half-precision for spherical harmonics evaluation on WebGPU

mvaligursky:
## Summary
- Converts spherical harmonics (SH) evaluation to use half-precision types for improved performance on GPUs with f16 support
- Changes SH coefficient array type from `array<vec3f, SH_COEFFS>` to `array<half3, SH_COEFFS>`
- Updates all SH format readers (uncompressed, compressed, SOG) to output half3 coefficients

## Performance Impact
- Halves memory usage for SH coefficient arrays when f16 is supported (90 bytes vs 180 bytes for 15 coefficients)
- Keeps all SH math in half precision end-to-end, reducing register pressure
- Falls back to f32 automatically on devices without f16 support via the `half` type aliases


willeastcott:
Does this mean the work buffer can use float16 to store coefficients as well?
Are we compromising visual in any way?

mvaligursky:
Work buffer stores only color .. we resolve SH every few degrees and store resolved values to keep rendering fast. But those are float16.

PR #8442, opened 2026-02-04
[Fix] GsplatResourceBase.getTexture needs to be public

mvaligursky:
- part of public API


src/scene/gsplat/gsplat-resource-base.js
```diff
@@ -358,7 +358,6 @@ class GSplatResourceBase {
      *
      * @param {string} name - The name of the texture.
      * @returns {Texture|undefined} The texture, or undefined if not found.
```
willeastcott: Should this return `null` if not found? What's the typical return value in this sort of case?

mvaligursky: yep null is more common, thanks, done

PR #8450, opened 2026-02-09
fix: Inline WGSL resolveCodebook to fix Firefox naga compatibility

mvaligursky:
## Summary

- we were pass uniform arrays by reference, which WebGPU spec does not allow (Chrome is ok with it, Firefox correctly is rejecting this)
- Inlines the WGSL `resolveCodebook` function at all 3 call sites to fix Firefox WebGPU shader validation errors
- Firefox's naga WGSL compiler rejects `ptr<uniform, ...>` as a function parameter per the WGSL spec (only `function`, `private`, and `workgroup` address spaces are allowed for pointer parameters)
- Codebook uniform arrays are now accessed directly at each call site instead of being passed through a function

## Files changed

- `src/scene/shader-lib/wgsl/chunks/gsplat/frag/gsplatPacking.js` — Removed `resolveCodebook` function
- `src/scene/shader-lib/wgsl/chunks/gsplat/frag/gsplatSogReorder.js` — Inlined codebook resolution for `scales_codebook` and `sh0_codebook`
- `src/scene/shader-lib/wgsl/chunks/gsplat/frag/gsplatSogReorderSh.js` — Inlined codebook resolution for `shN_codebook`

## Test plan

- [ ] Verify SOG gaussian splat rendering works correctly on Chrome (WebGPU)
- [ ] Verify SOG gaussian splat rendering works correctly on Firefox (WebGPU)
- [ ] Verify no shader validation errors in Firefox console

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/scene/shader-lib/wgsl/chunks/gsplat/frag/gsplatPacking.js
+++ b/src/scene/shader-lib/wgsl/chunks/gsplat/frag/gsplatPacking.js
@@ -24,16 +24,4 @@ fn unpack101010(v: u32) -> vec3f {
 
 fn unpack111110(v: u32) -> vec3f {
     return vec3f((vec3u(v) >> vec3u(21u, 10u, 0u)) & vec3u(0x7ffu, 0x7ffu, 0x3ffu)) / vec3f(2047.0, 2047.0, 1023.0);
-}
-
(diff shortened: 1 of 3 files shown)


slimbuck (approved):
any performance regression?

PR #8457, opened 2026-02-13
Refactor data texture creation using internal Texture helper

mvaligursky:
## Summary
- add internal `Texture.createDataTexture2D(graphicsDevice, name, width, height, format, levels)` helper for nearest-filtered, clamp-to-edge, no-mipmap 2D data textures
- replace repeated inline `new Texture(...)` option blocks with the helper across picker, post-effect queue, gsplat-unified paths, light atlas, depth grab, prepass, and SSAO
- inline `GSplatWorkBuffer._createTexture` since it became a direct pass-through to the new helper

### DIFF
--- a/src/extras/render-passes/render-pass-prepass.js
+++ b/src/extras/render-passes/render-pass-prepass.js
@@ -1,6 +1,4 @@
 import {
-    FILTER_NEAREST,
-    ADDRESS_CLAMP_TO_EDGE,
     PIXELFORMAT_R32F,
     PIXELFORMAT_RGBA8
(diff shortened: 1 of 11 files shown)


AlexAPPi:
I would create a separate DataTexture class; I can provide an example. This class would have an API for updating cells/rows and other functionality.

mvaligursky:
> I would create a separate DataTexture class; I can provide an example. This class would have an API for updating cells/rows and other functionality.

Hmm, not sure this is something particularly needed for the internal use in the engine?
create an issue with some details for this though.

PR #8466, opened 2026-02-20
Use half precision (f16) in gsplat WGSL shaders

slimbuck:
## Summary

Use half precision types throughout the gsplat WGSL shader pipeline to reduce ALU cost and varying bandwidth on devices that support the `shader-f16` WebGPU feature. On devices without f16 support, all `half` types automatically fall back to f32 via the engine's alias system, so there is no change in behavior.

### Changes

**Fragment shader** (`gsplat.js` frag)
- `gaussianUV` and `gaussianColor` varyings changed from `vec2f`/`vec4f` to `half2`/`half4`, halving interpolation bandwidth
- `normExp()` and alpha computation operate entirely in half precision
- Final output color promoted to f32 only at the output assignment

**Vertex shader - core** (`gsplat.js` vert, `gsplatStructs.js`, `gsplatSource.js`, `gsplatCommon.js`)
- `SplatSource.cornerUV` changed from `vec2f` to `half2`
- `SplatCorner.uv` changed from `vec2f` to `half2`, `aaFactor` from `f32` to `half`
- Color pipeline uses `half4` throughout; promoted to f32 only for the `modifySplatColor` user hook and `clipCorner` math
- `clipCorner` computes clip factor in f32 (for log/sqrt range), stores result as `half`

**Quaternion & covariance** (`gsplatQuatToMat3.js`, `gsplatCorner.js`)
- `quatToMat3` now returns `half3x3` and computes entirely in half precision
- `quatMul` computes in half, returns f32
- `computeCovariance` operates in `half3x3`; result promoted to f32 at the `initCornerCov` boundary for projection math

**Spherical harmonics** (`gsplatEvalSH.js`)
- `evalSH` return type changed from `vec3f` to `half3`, avoiding an unnecessary promotion since all SH constants and intermediates were already half

**SOG format** (`sog.js`, `sogSH.js`)
- `getColor()` dequantization (`mix`) and SH_C0 multiply now operate in half3
- `readSHTexel()` dequantization (`mix`) now operates in half3 instead of f32, benefiting the 15 calls per vertex (SH band 3)
- SH data read loop unrolled to avoid Dawn's forward progress volatile wrapper on loop-based texture reads

### Technical details

All half-precision types use the engine's alias system (`half`, `half2`, `half3`, `half4`, `half3x3`) defined in `half-types.js`. When `CAPS_SHADER_F16` is not defined, these resolve to their f32 equivalents (`f32`, `vec2f`, `vec3f`, `vec4f`, `mat3x3f`), ensuring the shaders compile and run correctly on all devices. No native WGSL f16 types (`f16`, `vec2h`, `mat3x3h`, etc.) are used directly.

f32 is intentionally retained for:
- Position dequantization and world-space calculations
- View/projection matrix math
- Eigenvalue decomposition in `initCornerCov` (precision-sensitive subtraction)
- `clipCorner` log/sqrt (dynamic range)

### DIFF
--- a/src/scene/shader-lib/wgsl/chunks/gsplat/vert/gsplatCommon.js
+++ b/src/scene/shader-lib/wgsl/chunks/gsplat/vert/gsplatCommon.js
@@ -14,7 +14,7 @@ export default /* wgsl */`
 
 // modify the gaussian corner so it excludes gaussian regions with alpha less than 1/255
 fn clipCorner(corner: ptr<function, SplatCorner>, alpha: half) {
-    let clip = min(half(1.0), sqrt(-log(half(1.0) / (half(255.0) * alpha))) / half(2.0));
+    let clip = min(half(1.0), sqrt(log(half(255.0) * alpha)) * half(0.5));
     corner.offset = corner.offset * f32(clip);
     corner.uv = corner.uv * clip;
 }
--- a/src/scene/shader-lib/glsl/chunks/gsplat/vert/gsplat.js
+++ b/src/scene/shader-lib/glsl/chunks/gsplat/vert/gsplat.js
@@ -75,6 +75,12 @@ void main(void) {
 
     modifySplatColor(modelCenter, clr);
 
+    // discard splats with alpha too low to contribute any visible pixel
+    if (255.0 * clr.w <= 1.0) {
+        gl_Position = discardVec;
+        return;
+    }
+
     clipCorner(corner, clr.w);
 
     // write output
--- a/src/scene/shader-lib/glsl/chunks/gsplat/vert/gsplatCommon.js
+++ b/src/scene/shader-lib/glsl/chunks/gsplat/vert/gsplatCommon.js
@@ -14,7 +14,7 @@ export default /* glsl */`
 
 // modify the gaussian corner so it excludes gaussian regions with alpha less than 1/255
 void clipCorner(inout SplatCorner corner, float alpha) {
-    float clip = min(1.0, sqrt(-log(1.0 / (255.0 * alpha))) / 2.0);
+    float clip = min(1.0, sqrt(log(255.0 * alpha)) * 0.5);
     corner.offset *= clip;
     corner.uv *= clip;
 }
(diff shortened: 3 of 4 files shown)


src/scene/shader-lib/wgsl/chunks/gsplat/vert/gsplatCommon.js
```diff
@@ -14,8 +14,8 @@ export default /* wgsl */`
 
 // modify the gaussian corner so it excludes gaussian regions with alpha less than 1/255
 fn clipCorner(corner: ptr<function, SplatCorner>, alpha: f32) {
-    let clip: f32 = min(1.0, sqrt(-log(1.0 / (255.0 * alpha))) / 2.0);
-    corner.offset = corner.offset * clip;
+    let clip: half = half(min(1.0, sqrt(-log(1.0 / (255.0 * alpha))) / 2.0));
```
mvaligursky: should this whole line use half precision, instead of just the result. Currently we just cast result to half and cast it back to float on the next line

slimbuck: Yes good catch, thanks!

src/scene/shader-lib/wgsl/chunks/gsplat/vert/gsplatCommon.js
```diff
@@ -13,9 +13,9 @@ export default /* wgsl */`
 #include "gsplatOutputVS"
 
 // modify the gaussian corner so it excludes gaussian regions with alpha less than 1/255
-fn clipCorner(corner: ptr<function, SplatCorner>, alpha: f32) {
-    let clip: f32 = min(1.0, sqrt(-log(1.0 / (255.0 * alpha))) / 2.0);
-    corner.offset = corner.offset * clip;
+fn clipCorner(corner: ptr<function, SplatCorner>, alpha: half) {
+    let clip = min(half(1.0), sqrt(-log(half(1.0) / (half(255.0) * alpha))) / half(2.0));
```
mvaligursky: I wonder ...

Image

slimbuck: Nice, I've included this too.

PR #8588, opened 2026-04-10
Move FisheyeProjection to scene/graphics for shared use

mvaligursky:
Move FisheyeProjection out of gsplat-unified into scene/graphics, since it is also used by skybox rendering.

**Changes:**
- Move `fisheye-projection.js` from `src/scene/gsplat-unified/` to `src/scene/graphics/`
- Update imports in gsplat renderer and skybox


willeastcott:
Doesn't this feel more like a Camera property?

mvaligursky:
In a way, but only two systems honor it (unified gsplat & infinite skydome), nothing else. Geometry, particles, UI, debug lines ... , none of that.

LeXXik:
Feels like a camera thingy, though

mvaligursky:
> Feels like a camera thingy, though

Might feel like it, but it's not :)
Even camera matrices are not updated to handle this in any way,
This is a custom implementation in two shaders.

PR #8601, opened 2026-04-15
findScript|findScripts returns Script instead of the deprecated

saidmoya12:
### Description
findScript and findScripts returns Script instead of the deprecated ScriptType.

### Changes
Now we can cast in typescript using

```const myScript = entity.findScript("scriptName") as MyExtendedScript```

### Checklist

### DIFF
--- a/src/framework/entity.js
+++ b/src/framework/entity.js
@@ -23,7 +23,7 @@ import { getApplication } from './globals.js';
  * @import { RigidBodyComponent } from './components/rigid-body/component.js'
  * @import { ScreenComponent } from './components/screen/component.js'
  * @import { ScriptComponent } from './components/script/component.js'
- * @import { ScriptType } from './script/script-type.js'
+ * @import { Script } from './script/script.js'
  * @import { ScrollViewComponent } from './components/scroll-view/component.js'
  * @import { ScrollbarComponent } from './components/scrollbar/component.js'
  * @import { SoundComponent } from './components/sound/component.js'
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Maksims** [comment]
This PR is potentially will break some advanced code.
First of all properties of a class should start from a lowercase letter, not uppercase.
Additionally it introduces a variable, that is not in private notation, that likely will collide with people's code who uses that property. Polluting script space - should be last resort.

Also, ScriptType extends Script, so there should not be a problem with existing code, as user can easily check for both.
If it is related to TypeScript, then TypeScript related issue should be solved without changing JS side.

**Comment by willeastcott** [comment]
Thanks for raising this, @saidmoya12. The underlying problem is real: because `ScriptType` declares private fields, TypeScript treats it nominally, so `entity.findScript('x') as MyScript` fails, and so does `entity.findScript(MyScript)` and `entity.script.get(...)`, which this PR did not cover.

We've taken it further in #9472, which types `findScript`, `findScripts`, `script.get`, `script.create` and the `ScriptRegistry` against `Script`, and adds a generic overload so passing the class returns that class directly with no cast:

```ts
const controller = entity.findScript(PlayerController); // PlayerController | undefined
```

Two parts of this PR we could not take: renaming `ScriptAttributes#scriptType` is a public API break, and renaming the reserved attribute name `__scriptType` to `__Script` would have stopped the reserved-name guard from protecting the real `__scriptType` field on `Script`.

Closing in favor of #9472.

PR #8606, opened 2026-04-16
Add shader defines for prepass to support alpha test and dithered opacity

slimbuck:
## Summary

The prepass shader pass was sharing the same empty code path as the pick pass, which meant it had no shader defines set. This caused materials with alpha testing or opacity dithering to render incorrectly during the prepass because the relevant shader code was never compiled in.

This PR separates the prepass from the pick pass and adds a dedicated `preparePrepassPass()` method to `LitShader` that sets:

- `LIT_ALPHA_TEST` — enables alpha-test discard in the prepass
- `LIT_BLEND_TYPE` — sets the blend type so the shader can handle transparency correctly
- `STD_OPACITY_DITHER` — enables opacity dithering (matching the shadow pass behavior via `opacityShadowDither`)

## Technical Details

- Adds the `ditherNames` import from `constants.js`
- Splits the `SHADER_PICK` / `SHADER_PREPASS` combined branch in `generateFragmentShader()` so each pass gets its own preparation
- The new `preparePrepassPass()` follows the same pattern as `prepareShadowPass()`

### DIFF
--- a/src/scene/shader-lib/programs/lit-shader.js
+++ b/src/scene/shader-lib/programs/lit-shader.js
@@ -13,6 +13,7 @@ import {
     SPRITE_RENDERMODE_SLICED, SPRITE_RENDERMODE_TILED, shadowTypeInfo, SHADER_PREPASS,
     lightTypeNames, lightShapeNames, spriteRenderModeNames, fresnelNames, blendNames, lightFalloffNames,
     cubemaProjectionNames, specularOcclusionNames, reflectionSrcNames, ambientSrcNames,
+    ditherNames,
     REFLECTIONSRC_NONE
 } from '../../constants.js';
 import { ChunkUtils } from '../chunk-utils.js';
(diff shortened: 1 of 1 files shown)


src/scene/shader-lib/programs/lit-shader.js
```diff
@@ -513,6 +514,13 @@ class LitShader {
         this._setupLightingDefines(hasAreaLights, options.clusteredLightingEnabled);
     }
 
+    preparePrepassPass() {
+        const { options } = this;
+        this.fDefineSet(options.alphaTest, 'LIT_ALPHA_TEST');
+        this.fDefineSet(true, 'LIT_BLEND_TYPE', blendNames[options.blendType]);
```
mvaligursky: The other two lines are great, but this BLEND one should be removed, as the pre-pass does not blend at all. Only alpha testing functionality can be used.

slimbuck: Done thank you!

mvaligursky (approved):

PR #8618, opened 2026-04-19
docs: fix JSDoc link warnings for ShaderUtils and GSplatComponent

willeastcott:
## Summary

Fixes three TypeDoc `{@link}` resolution warnings surfaced by `npm run docs`:

- `src/platform/graphics/shader.js` - change `{@link ShaderUtils#createShader}` to `{@link ShaderUtils.createShader}` (static members use `.`).
- `src/scene/shader-lib/shader-utils.js` - add a class-level JSDoc block to `ShaderUtils`. Without it, `excludeNotDocumented: true` strips the class from the generated docs, which is why the link above failed to resolve even after the syntax fix.
- `src/framework/components/gsplat/component.js` - change `{@link workBufferModifier}` (no such symbol) to `{@link setWorkBufferModifier}`.

## Test plan

- [x] `npm run docs` - the three targeted warnings are gone.
- [x] Spot-check the generated `ShaderUtils` page and `Shader` constructor docs to confirm the link now resolves.

### DIFF
--- a/src/framework/components/gsplat/component.js
+++ b/src/framework/components/gsplat/component.js
@@ -562,8 +562,8 @@ class GSplatComponent extends Component {
      * - {@link WORKBUFFER_UPDATE_ONCE}: Force update this frame, then switch to AUTO.
      * - {@link WORKBUFFER_UPDATE_ALWAYS}: Update every frame.
      *
-     * This is typically useful when using custom shader code via {@link workBufferModifier} that
-     * depends on external factors like time or animated uniforms.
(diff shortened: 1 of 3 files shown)


willeastcott:
@mvaligursky This PR requires we make `ShaderUtils` class public API. Is that what you want? The alternative is to remove the comment recommending the use of `ShaderUtils.createShader`.

LeXXik:
Its useful for post effects, and some of the engine-only examples use it, so I'd say we make it public.

mvaligursky:
Yep, that function should be public.

PR #8632, opened 2026-04-22
Calculate camera aspect ration in CameraComponent onEnable

### LINKED ISSUE #5825: camera.screenToWorld calculates incorrectly for orthographic camera until first update is done
### Description

Example: https://playcanvas.com/project/1163442/overview/incorrect-screentoworld
(projection and orthographic cameras. Both calculate the left center screen border screenToWorld position 4 times. During initialize, postInitialize, first and second update)

1. launch scene
2. open dev tools
3. reload page
4. check logs

camera with perspective projection calculates world position correctly all 4 times

camera with orthographic projection calculates incorrectly during initialize and postInitialize
on the first update only **y** is calculated correctly
on the second and all subsequent updates **x** and **y** are calculated correctly

Platform: Windows
Browser: Chrome

lucaheft:
## Description
Fixes an issue where a `CameraComponent` with `aspectRatioMode` set to `ASPECT_AUTO` would report a stale/incorrect `aspectRatio` until the first frame was rendered. The aspect ratio is now additionally calculated in `onEnable`, so it is correct from the moment the component is enabled.

### Problem
When a camera is enabled with `aspectRatioMode === ASPECT_AUTO`, `aspectRatio` was only refreshed by the render loop. Any code that read `camera.aspectRatio` before the first frame — e.g. during scene setup, picking, frustum tests, or screen-space math — would see the default value of `16/9`, not the actual current value for the target render surface.

Fixes #5825 

## Checklist

### DIFF
--- a/src/framework/components/camera/component.js
+++ b/src/framework/components/camera/component.js
@@ -1207,6 +1207,10 @@ class CameraComponent extends Component {
             this.addCameraToLayers();
         }
 
+        if (this.aspectRatioMode === ASPECT_AUTO) {
+            this.aspectRatio = this.calculateAspectRatio();
+        }
+
         this.postEffects.enable();
     }
 


mvaligursky (approved):
This works, thanks. We'd ideally handle additional cases, for example renderTarget setter as well, but that can be done separatetely.

Thanks!

PR #8656, opened 2026-04-27
Use texImage2D for gsplat splat-order uploads

slimbuck:
## Summary

On WebGL, the gsplat CPU-sort path uploads a per-frame `Uint32Array` of sorted
splat indices into an `R32UI` texture via `UploadStream.uploadPBO` →
`gl.bufferSubData` (PBO orphan) → `gl.texSubImage2D`. For multi-MB integer
textures, this stalls the main thread on Chrome's renderer→GPU IPC.

A `?cpu-sort` (WebGPU CPU sort, same JS sorter) vs `?webgl` (WebGL2 CPU sort)
comparison on the same splat scene:

|                        | WebGL2  | WebGPU CPU-sort |
|------------------------|---------|-----------------|
| FireAnimationFrame p90 | 74 ms   | 2.2 ms          |
| FireAnimationFrame max | 93 ms   | 2.4 ms          |
| Top hotspot            | `bufferSubData` (~25% of CPU) | `uploadStaging` (negligible) |

Switching `WebglUploadStream.uploadDirect` to call `gl.texImage2D` directly
(full-buffer, fresh allocation each frame) takes a different — much cheaper —
IPC path on Chrome and eliminates the stall. WebGPU is unchanged.

## Changes

- `WebglUploadStream.uploadDirect` rewritten to call `gl.texImage2D` directly,
  bypassing `Texture.upload()` (which uses `texSubImage2D` after the first
  frame). Wraps the source `TypedArray` as `Uint8Array` when the texture's
  pixel type is `UNSIGNED_BYTE`.
- `GSplatWorkBuffer` (unified) and `GSplatSorter` (legacy) construct their
  `UploadStream` with `useSingleBuffer = !device.isWebGPU`, so WebGL routes
  through the new `uploadDirect`. WebGPU continues using the staging path.

## Test plan

- [ ] WebGL2 gsplat scene (unified + legacy) renders correctly
- [ ] WebGPU gsplat scene unchanged (no visual or perf regression)
- [ ] Profile confirms main-thread stall is eliminated on WebGL
- [ ] Smoke-test on Windows / Linux WebGL2 to confirm no platform regresses

### DIFF
--- a/src/platform/graphics/webgl/webgl-upload-stream.js
+++ b/src/platform/graphics/webgl/webgl-upload-stream.js
@@ -133,9 +133,6 @@ class WebglUploadStream {
         device.setUnpackFlipY(false);
         device.setUnpackPremultiplyAlpha(false);
         device.setUnpackAlignment(data.BYTES_PER_ELEMENT);
-        gl.pixelStorei(gl.UNPACK_ROW_LENGTH, 0);
-        gl.pixelStorei(gl.UNPACK_SKIP_ROWS, 0);
-        gl.pixelStorei(gl.UNPACK_SKIP_PIXELS, 0);
 
         // Wrap the source TypedArray to match the texture's pixel type when needed
         // (e.g. RGBA8UI expects UNSIGNED_BYTE → Uint8Array, even if the caller passes Uint32Array).
--- a/src/platform/graphics/webgl/webgl-upload-stream.js
+++ b/src/platform/graphics/webgl/webgl-upload-stream.js
@@ -104,7 +104,11 @@ class WebglUploadStream {
     }
 
     /**
-     * Direct texture upload (simple, blocking).
+     * Direct texture upload via gl.texImage2D — uploads the full buffer with a
+     * fresh storage allocation each call. Bypasses the engine's Texture.upload
+     * path (which uses texSubImage2D after the first frame). This avoids the
+     * texSubImage2D path's stall on multi-MB integer-format uploads through
+     * Chrome's renderer→GPU IPC on some drivers.
      *
      * @param {Uint8Array|Uint32Array|Float32Array} data - The data to upload.
      * @param {Texture} target - The target texture.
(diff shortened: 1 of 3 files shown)


src/platform/graphics/webgl/webgl-upload-stream.js
```diff
@@ -114,10 +118,42 @@ class WebglUploadStream {
...
+        // Ensure the GL texture object exists and is bound.
+        // @ts-ignore - setTexture is available on WebglGraphicsDevice
+        device.setTexture(target, 0);
+        device.activeTexture(0);
+        device.bindTexture(target);
+
+        device.setUnpackFlipY(false);
+        device.setUnpackPremultiplyAlpha(false);
+        device.setUnpackAlignment(data.BYTES_PER_ELEMENT);
+        gl.pixelStorei(gl.UNPACK_ROW_LENGTH, 0);
```
mvaligursky: it should use functions similar to device.setUnpackAlignment instead of pixelStorei directly
but we need to be sure these calls will have no side-effects for normal functionality, as we're changing some settings. Do we need to reset those back?

slimbuck: Good point.

I'm already using the cached wrappers (`setUnpackFlipY` / `setUnpackPremultiplyAlpha` / `setUnpackAlignment`) for the three unpack values the engine caches. The other three (`UNPACK_ROW_LENGTH` / `UNPACK_SKIP_ROWS` / `UNPACK_SKIP_PIXELS`) aren't cached on the device, but grepping `src/` confirms `uploadPBO` and `uploadDirect` are the only writers anywhere in the engine. Their WebGL2 default is `0` and nothing else touches them, so the three `pixelStorei(..., 0)` calls were no-ops. Dropped them — the side-effect concern goes with them.

The three values set via cached wrappers (FlipY=false, PremultiplyAlpha=false, Alignment=BYTES_PER_ELEMENT) do persist in the cache, but every other upload site in the engine sets these per-call too, so the pattern is unchanged.

(`uploadPBO` has the same redundant `pixelStorei` lines — pre-existing and left alone for this PR; happy to clean up in a follow-up if useful.)

mvaligursky (approved):

PR #8675, opened 2026-05-03
refactor: Rename legacy _addedModel/_showModel/_hideModel on SpriteComponent

willeastcott:
## Summary

Renames private legacy-named state and methods in `SpriteComponent` to reflect what they actually do (managing mesh instance membership in scene layers):

- `_addedModel` -> `_inLayers`
- `_showModel()` -> `_addToLayers()`
- `_hideModel()` -> `_removeFromLayers()`

## Details

These names were holdovers from when `SpriteComponent` owned a `Model` container (removed in #8674). The flag and methods never managed a `Model` directly — they manage the single `_meshInstance` to/from scene layers.

All renamed symbols are private. Updates two call sites in `SpriteAnimationClip` (`_component._hideModel()` / `_component._showModel()`).

Left alone: the public `removeModelFromLayers()` method, since it matches the cross-component convention (`ModelComponent.removeModelFromLayers`, `ElementComponent.removeModelFromLayers`) used by `BatchManager`.

## Test plan

- `npm run lint` — clean.
- `npm test` — 1672 passing, 0 failing.

### DIFF
--- a/src/framework/components/sprite/component.js
+++ b/src/framework/components/sprite/component.js
@@ -219,7 +219,7 @@ class SpriteComponent extends Component {
     _updateAabbFunc = null;
 
     /** @private */
-    _addedModel = false;
+    _inLayers = false;
 
     // animated sprites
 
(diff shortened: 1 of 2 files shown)


willeastcott:
Folding into #8674 per request.

PR #8730, opened 2026-05-15
fix(xr): visionOS MSAA resolve into XR framebuffer

### LINKED ISSUE #7410: WebXR: Apple Vision Pro rendering issues
It is still a problem that WebXR VR does not work well with Apple Vision Pro. It is likely due to a specific implementation differences of AVP's WebXR, but I believe we can do a workaround on our end to ensure it works well.

The only way to run VR session on AVP is to disable anti-aliasing.

mvaligursky:
Fixed #7410

Working WebXR with MSAA on Apple Vision Pro. Previously the only way to get a usable image on AVP was to disable anti-aliasing, because the engine's MSAA resolve uses `gl.blitFramebuffer` directly into the `XRWebGLLayer`'s opaque framebuffer — a path that visionOS WebKit handles incorrectly, producing a small zoomed-in sub-region instead of full stereo content.

**Fix**: on visionOS only, MSAA color is resolved into an internal scratch texture via `blitFramebuffer`, then copied 1:1 into the XR framebuffer with a single fullscreen textured quad. The rasterization path is what AVP renders correctly.

**Changes:**
- Added `pc.platform.visionos` boolean — detects Apple Vision Pro. visionOS Safari spoofs a macOS UA (no `"visionOS"` string), so detection combines `Macintosh` UA + `maxTouchPoints > 0` + no iPhone/iPad/iPod token.
- New internal `WebglXrMsaaCopy` helper — owns a scratch `Texture`/`RenderTarget` and a fullscreen-quad copy shader, built via `Shader`, `ShaderDefinitionUtils`, and `device.scope`. Lazily created on `WebglGraphicsDevice`, only on visionOS.
- `WebglRenderTarget.resolve` routes XR color resolves through the helper when `platform.visionos` is true; all other XR platforms (Quest, Chrome, etc.) keep the existing direct-blit path with zero overhead.
- Depth resolve is unchanged on all platforms.

**API Changes:**
- Added `pc.platform.visionos` (`boolean`).

**Performance:**
- visionOS only: one extra `blitFramebuffer` (MSAA → scratch texture) plus one fullscreen quad draw per frame. Required because direct blits into the AVP XR FBO are broken; `WEBGL_multisampled_render_to_texture` is not available on AVP.
- Other platforms: zero impact, gated on `platform.visionos`.

### DIFF
--- a/src/core/platform.js
+++ b/src/core/platform.js
@@ -38,6 +38,12 @@ const browserName =
                     'other')));
 
 const xbox = /xbox/i.test(ua);
+// visionOS Safari reports a macOS-style UA (no "visionOS" string) but uniquely has
+// maxTouchPoints > 0 on a "Macintosh" UA without any iPhone/iPad/iPod token.
+const visionos = /Macintosh/i.test(ua) &&
+    typeof navigator !== 'undefined' &&
+    navigator.maxTouchPoints > 0 &&
+    !/iPhone|iPad|iPod/i.test(ua);
 const touch = (environment === 'browser') && ('ontouchstart' in window || ('maxTouchPoints' in navigator && navigator.maxTouchPoints > 0));
 const gamepads = (environment === 'browser') && (!!navigator.getGamepads || !!navigator.webkitGetGamepads);
 const workers = (typeof Worker !== 'undefined');
@@ -122,6 +128,13 @@ const platform = {
      * @type {boolean}
      */
     android: platformName === 'android',
+
+    /**
+     * True if running on Apple Vision Pro (visionOS).
+     *
+     * @type {boolean}
+     */
+    visionos: visionos,
 
     /**
      * True if running on an Xbox device.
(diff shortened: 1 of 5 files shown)


src/core/platform.js
```diff
@@ -123,6 +129,13 @@ const platform = {
      */
     android: platformName === 'android',
 
+    /**
+     * True if running on Apple Vision Pro (visionOS).
+     *
+     * @type {boolean}
+     */
+    visionos: visionos,
```
willeastcott, line 137: Gotta say - I'm not a fan of explanding the `platform` object further. This user agent sniffing technique is really scary (should the browser change it someday) - people's published app behavior can suddenly change/break.

mvaligursky: yeah those are dirty.
but I need it internally .. I'll make it @ignore.

mvaligursky: making it non-public here https://github.com/playcanvas/engine/pull/8765

PR #8731, opened 2026-05-15
fix: Setting renderSceneColorMap or renderSceneDepthMap to true when already set, sets it to false

MAG-AdrianMeredith:
Setting renderSceneColorMap or renderSceneDepthMap to true when already set, sets it to false

## Description
I don't have a proveable issue here but I saw this function and it sounds dodgy af.  Passing true can cause it to set it to false, it seems unlikely to be intentional

Fixes #
n/a
## Checklist


mvaligursky (approved):
Great find, thanks!

PR #8795, opened 2026-05-27
docs(entity): make getGuid() part of the public API

### LINKED ISSUE #7394: Public version of entity._guid
Currently, I use the private API `entity._guid` a lot, as it is the only entity-specific option available.

Personally, I think it's strange that this is a private API, especially since we already have a public option like [`findByGuid()`](https://api.playcanvas.com/classes/Engine.Entity.html#findbyguid).

It would be nice to have something public, such as `entity.getGuid()` or `entity.guid`, as well.

mvaligursky:
## Summary

- Removes `@ignore` from `Entity.getGuid()` so it shows up in the generated API docs.
- Adds a note that the GUID is generated lazily on first access.

## Why

Users have been relying on the private `entity._guid` because there was no public accessor — even though `findByGuid()` is fully public. `getGuid()` is already widely used internally (scene parser, element-input, component systems, scroll-view, button, render, anim, script, rigid-body, screen, collision), so the contract is stable; it just wasn't documented for external use.

`setGuid()` is intentionally left `@ignore` — its existing JSDoc warns that misuse "will corrupt the graph this Entity is in", so it should stay an internal API.

Resolves #7394.

## Test plan

- [ ] Confirm `getGuid()` appears in the generated API reference.
- [ ] No behavior change — existing internal callers are unaffected.

### DIFF
--- a/src/framework/entity.js
+++ b/src/framework/entity.js
@@ -473,10 +473,10 @@ class Entity extends GraphNode {
     }
 
     /**
-     * Get the GUID value for this Entity.
+     * Get the GUID value for this Entity. The GUID is generated lazily on first access if one has
+     * not been assigned.
      *
      * @returns {string} The GUID of the Entity.
-     * @ignore
      */
     getGuid() {
         // if the guid hasn't been set yet then set it now before returning it


willeastcott:
If we're going to make this public, this would be better done as `Entity#guid` (getter).

willeastcott:
And there could be an argument to make the setter public too. I'm guessing our own Editor has to circument the API when loading scenes.

mvaligursky:
yep, I'll create a getter. Setter is risky due to this comment:

     * Set the GUID value for this Entity. Note that it is unlikely that you should need to change
     * the GUID value of an Entity at run-time. Doing so will corrupt the graph this Entity is in.

PR #8797, opened 2026-05-27
docs(entity): expose guid as a public getter

mvaligursky:
## Summary

- Promotes `Entity#guid` to a public read/write property — the lazy GUID generation and `_entityIndex` bookkeeping now live in the getter/setter directly.
- `Entity#getGuid()` and `Entity#setGuid()` are retained as `@ignore` + `@deprecated` wrappers that emit `Debug.deprecated(...)` and delegate to the property, so existing external callers (notably the editor) keep working.
- Migrates every in-engine caller (14 src files, 12 test/asset files) off the deprecated methods so the engine doesn't trigger its own deprecation warnings.

## Why

Follow-up to #8795 and #7394. Per review on the previous follow-up:
- @willeastcott suggested making `getGuid/setGuid` deprecated wrappers that call the new property accessors.
- Exposing the setter publicly is debatable, but keeping `setGuid` working (via `@ignore` + deprecation) doesn't add public API surface and avoids breaking the editor, which calls it in many places.

## Test plan

- [x] Full test suite passes (`npm test` — 1720 passing).
- [x] Lint clean across all modified files.
- [ ] Confirm `Entity#guid` appears in the generated API reference and `getGuid`/`setGuid` do not.
- [ ] Confirm calling `getGuid()`/`setGuid()` from user code logs a single deprecation warning and still works.

### DIFF
--- a/src/framework/entity.js
+++ b/src/framework/entity.js
@@ -473,10 +473,20 @@ class Entity extends GraphNode {
     }
 
     /**
+     * Gets the GUID for this Entity.
+     *
+     * @type {string}
+     */
+    get guid() {
+        return this.getGuid();
+    }
+
+    /**
      * Get the GUID value for this Entity. The GUID is generated lazily on first access if one has
      * not been assigned.
      *
      * @returns {string} The GUID of the Entity.
+     * @ignore
      */
     getGuid() {
         // if the guid hasn't been set yet then set it now before returning it
--- a/src/framework/entity.js
+++ b/src/framework/entity.js
@@ -473,28 +473,48 @@ class Entity extends GraphNode {
     }
 
     /**
+     * Sets the GUID for this Entity. Note that it is unlikely that you should need to change the
+     * GUID value of an Entity at run-time. Doing so will corrupt the graph this Entity is in.
+     *
+     * @type {string}
+     * @ignore
+     */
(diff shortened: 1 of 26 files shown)


src/framework/entity.js
```diff
@@ -472,11 +472,21 @@ class Entity extends GraphNode {
         return entities.map(entity => entity.c.script.get(nameOrType));
     }
 
+    /**
+     * Gets the GUID for this Entity.
+     *
+     * @type {string}
+     */
+    get guid() {
+        return this.getGuid();
```
willeastcott: You could deprecate `getGuid/setGuid` and have them call the getter/setter.

mvaligursky: Consider that majority of uses should not call the setter, I'm not sure I'd want to expose it at all.
We can deprecate getGuid, but Editor calls it in many places as well. I do not think it's worth breaking Editor's compatibility with old versions of the engine for this.

willeastcott: It's fine - it won't break the editor. It'll just be calling depreacted functions and I'll migrate it to the getter sometime soon.

mvaligursky: yeah I guess so, the Editor never runs with older version of engine now, even if projects use older version.

PR #8818, opened 2026-06-01
[BREAKING] World-space PCSS penumbra, plus gsplat & cascade fixes

mvaligursky:
**Breaking change — `light.penumbraSize` is now interpreted in world-space units.**

Directional PCSS soft shadows previously interpreted `penumbraSize` in shadow-map texel/UV space, so the resulting softness depended on the shadow resolution and on how the shadow camera happened to be fitted to the scene. It is now interpreted in **world units**, matching the documented meaning ("the area size of the light"): the penumbra width scales with the real blocker→receiver distance in the scene and is independent of shadow resolution and light direction.

As a result the useful numeric range has changed dramatically — values that previously read as roughly 1–100 now correspond to roughly 0.01–0.2. **Applications that set `penumbraSize` (and, to a lesser degree, `penumbraFalloff`) will need to re-tune them**; old values produce extremely over-soft, washed-out shadows. As a guideline, ~0.05 is a sensible starting point for a typical scene, ~0.01–0.02 approximates a sharp sun, and ~0.5+ a broad area light. `penumbraFalloff` keeps its meaning (higher = softens faster with distance, 1 = linear) but its interaction with the new world-space curve may warrant minor adjustment.

**Changes:**
- **World-space directional PCSS** — penumbra is computed from world-space blocker→receiver distance instead of normalized shadow-map depth, making softness consistent regardless of light direction, shadow resolution, or shadow-camera fit. (GLSL + WGSL)
- **Gaussian splat shadows** — splats now write correct depth in the shadow pass (previously a constant 0), so they cast proper PCSS contact-hardening shadows. The unified gsplat manager now reports a real world-space AABB for its shared mesh instance, so directional shadow fitting/culling uses the actual splat bounds.
- **Cascaded PCSS stability** — the directional shadow camera depth range is derived from a stable cross-cascade union of caster bounds, eliminating sudden softness jumps as casters cross cascade boundaries or the camera moves. Per-cascade ortho radii are passed to the shader so each cascade softens correctly instead of inheriting cascade 0's radius.
- PCSS-specific shader uniforms are gated behind the PCSS shadow type, so the common PCF/VSM paths carry no extra cost.

**API Changes:**
- `light.penumbraSize` — same signature, **behavioral/units change** (now world-space). Compiles unchanged; values need re-tuning per the note above.

**Examples:**
- New `gaussian-splatting/shadow-soft` example — gsplat bikes casting soft shadows onto a ground disc.
- `graphics/shadow-cascades` — added PCSS Penumbra/Falloff HUD sliders; tripled camera zoom-out range.
- Re-tuned `penumbraSize`/`penumbraFalloff` defaults in `shadow-soft`, `shadow-cascades`, `shadow-catcher`, `gaussian-splatting/simple`, and `gaussian-splatting/spherical-harmonics` to the new world-space scale.

Screenshot 2026-06-01 at 14 40 10


d0rianb:
Hi, in the [shadow example](https://engine-7xh3hefv1-playcanvas.vercel.app/#/gaussian-splatting/shadow-soft), at a some camera distance, the shadows clip. 
Is there a specific parameter controlling this behavior ?
screenshot

mvaligursky:
shadowDistance on the light controls this distance.

d0rianb:
Hi, on mobile (tested on iphone, Safari, latest iOS), the shadows are not soft, with the default example parameters : 
F5230C78-7AB9-4B83-A661-622D7E0B7B21_1_201_a

mvaligursky:
It's only soft on devices that support float-32 filterable textures, and unfortunately many iPhones do not have support for this, and so it falls back to hard shadows.

Issue #8824, opened 2026-06-02
Adding ESM script to ScriptRegistry - not added by it's name (bug).


Maksims:
Using engine-only.
Creating ESM scripts through extending Script class.
[User Manual](https://developer.playcanvas.com/user-manual/scripting/) explains how to define script, but no mention on how to add it to registry.

```js
import { Script } from 'playcanvas';

export class Test extends Script {
    static scriptName = 'test';

    update(dt) {
        this.entity.rotate(10 * dt, 10 * dt, 10 * dt);
    }
}
```

API docs for [ScriptRegistry](https://api.playcanvas.com/engine/classes/ScriptRegistry.html) have `add` method, which works perfect on ScriptType (non ESM scripts). But when adding ESM script using such method, then adding script to entity using script name does not work.
`static scriptName` seems to be ignored when adding to registry.

So this code will not result in script added:
```js
entity.addComponent('script');
entity.script.create('test');
```

As well as instantiating entities from templates or any other procedural load.
It is only possible to add script to entity by class reference, which is highly limiting and prevents async loading of scripts in any form.

This is a regression in async capabilities of engine for scripting.

# Solution:
When adding ESM script to registry, ensure it is registered by its static name.

LeXXik:
Curious what do you use scripts for in an engine-only project? I mean there is probably a good reason, I just didn't face one. In my case it seemed enough to simply sub to update events the methods that needed those. I always felt Script and ScriptType scripts were for the Editor.

Maksims:
Many benefits:
1. Organizational and splitting the structure between isolated scripts - good for long term.
2. Subscribing to app.update for every Entity - would be an overkill.
3. ScriptComponent have 2 layers of execution order: hierarchical and within component order of them affects order of execution. So you have control over it.
4. Async loading of scripts on larger project, works great also.

mvaligursky:
Hi @Maksims - please check the fix. I'd appreciate if you could also test it, if possible, with your use case.

mvaligursky:
Closing this as its done. Please comment if that's not correct

Issue #2881, opened 2021-02-17
Subclass of a script replaces the superclass script


detzt:
Consider this scenario where `Derived` is a subclass of `Base` (in TypeScript):

```TypeScript
class Base extends pc.ScriptType { ... }
pc.registerScript(Base);

class Derived extends Base { ... }
pc.registerScript(Derived);
```

When now this line is executed the wrong name is taken:

https://github.com/playcanvas/engine/blob/38649760a3afd22161d5bdae59f1e25abce4cbeb/src/script/script.js#L138

Since `Derived` is a `Base` and `Base.__name` was set to "Base" before, `Derived.__name` returns "Base" as well.
When a few lines later the `Derived` script is registered, it replaces the already registered `Base` script and calls the `swap` method.

mvaligursky:
This has been fixed by #8831 (released in 2.19.3), with a follow-up in #8844.

The root cause was that script name resolution read `__name`/`scriptName` through the prototype chain, so a derived class inherited its base class's already-assigned name and overwrote the base in the registry on registration.

Name resolution now considers **only own properties** of the class (centralized in `getScriptRegistryName`: own `__name` → own `scriptName` → lowerCamelCase class name), and the same logic is used by `registerScript`, `ScriptRegistry.add`, `ScriptComponent.create`, and the ESM script handler. A subclass therefore registers under its own name and no longer collides with (or hot-swaps) its base. The explicit-name workaround (`pc.registerScript(Derived, 'Derived')`) is no longer required.

Covered by dedicated subclassing tests in `test/framework/script/script-registry.test.mjs`.

Closing as resolved.

PR #8831, opened 2026-06-03
Fix ESM script registration by name and subclass name collision

mvaligursky:
## Summary

Fixes two long-standing script name-resolution bugs that share one root cause: names were read through the prototype chain instead of from the class itself.

### #8824 — ESM script added to the registry not keyed by its name
ESM scripts declare their name with a static `scriptName` field. Because it's a class field, it shadows the inherited `scriptName` accessor (`[[Define]]` semantics), so the setter that assigns `__name` never runs. `ScriptRegistry.add()` read the unset `__name` and registered the script under a `null` key — so it couldn't be created by name:

```js
class Test extends Script { static scriptName = 'test'; }
app.scripts.add(Test);
entity.script.create('test'); // silently failed
```

This broke template instantiation, procedural creation, and async/dynamic script loading for engine-only projects.

### #2881 — subclass of a script replaces the superclass
A subclass inherited its base class's already-assigned `__name`/`scriptName`, so registering the subclass overwrote the base in the registry (and could trigger an unintended hot-swap):

```js
registerScript(Base);     // registered as 'Base'
registerScript(Derived);  // also resolved to 'Base' -> overwrote Base
```

## Fix

Name resolution now considers **only own properties** of the class. The logic is centralized in a new internal `getScriptRegistryName()` helper — own `__name` → own `scriptName` → lowerCamelCase class name — and used consistently by `registerScript`, `ScriptRegistry.add`, `ScriptComponent.create`, and the ESM script handler.

## Compatibility

- No public API signatures change. `registerScript`, `createScript`, `ScriptRegistry`/`.add()`, `Script` are unchanged; the new helpers are internal (not exported from `index.js`).
- Conforming scripts (with a declared `scriptName`, or a name passed to `createScript`/`registerScript`) are unaffected.
- Only observable change in the undocumented nameless path: `registerScript`/`add` now resolve to the lowerCamelCase class name (matching the asset loader and `create`, which already did) instead of PascalCase.

## Tests

Adds `test/framework/script/script-registry.test.mjs` covering both fixes (ESM `add` by name, subclass non-collision, explicit-name precedence, nameless camelCase unification). Full suite passes.

### DIFF
--- a/src/framework/components/script/component.js
+++ b/src/framework/components/script/component.js
@@ -725,6 +725,13 @@ class ScriptComponent extends Component {
                 scriptType.__name = getScriptRegistryName(scriptType);
             }
             scriptName = scriptType.__name;
+
+            // bail out rather than index the script under an `undefined`/empty name (which would
+            // attach it as `this.undefined` and could mask other scripts)
+            if (!scriptName) {
+                Debug.error(`The script class could not be added to entity '${this.entity.name}' because its name could not be resolved. Add a static "scriptName" property to the class.`);
+                return null;
+            }
         }
 
         if (scriptType) {
--- a/src/framework/components/script/component.js
+++ b/src/framework/components/script/component.js
@@ -8,14 +8,12 @@ import {
     SCRIPT_POST_UPDATE, SCRIPT_SWAP
 } from '../../script/constants.js';
 import { ScriptType } from '../../script/script-type.js';
-import { getScriptName } from '../../script/script.js';
+import { getScriptName, getScriptRegistryName, toLowerCamelCase } from '../../script/script.js';
 
 /**
  * @import { ScriptComponentSystem } from './system.js'
  * @import { Script } from '../../script/script.js'
  */
-
-const toLowerCamelCase = str => str[0].toLowerCase() + str.substring(1);
 
 /**
  * The ScriptComponent enables an {@link Entity} to have custom behavior by attaching scripts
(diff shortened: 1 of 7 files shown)


mvaligursky:
Thanks for the review @copilot. Addressed the name-resolution edge cases (anonymous class with an empty `.name` and no static `scriptName`) in `05318dc`:

**Addressed (threads resolved):**
- **`toLowerCamelCase`** now returns the input unchanged for empty/falsy values instead of throwing on `str[0]`. This also defuses the two warning-string crashes (`component.js` and `handlers/script.js`).
- **`registerScript`** logs an error and returns early when no name resolves, so it no longer assigns `__name = undefined`, calls `add()`, or pushes a nameless entry to `ScriptTypes`.
- **`ScriptComponent.create`** returns `null` (and logs) instead of indexing the instance under an `undefined`/empty key (`this.undefined`).
- **ESM script handler** skips exports whose name cannot be resolved rather than registering under an invalid key.

Added tests for the `registerScript` and `ScriptComponent.create` fail-fast paths.

**Not addressed (left open):**
- The `this._scripts.hasOwnProperty(...)` suggestion in `script-registry.js`. It's a valid hardening point, but that line is pre-existing (untouched by this PR) and unrelated to the name-resolution fix, so it's out of scope here. Happy to do it as a separate small PR.

Maksims:
`app.scripts.add(Script)` - now works as intended.
Also tested on large project if any regressions, no regressions found.
Thank you!

Maksims (approved):

PR #8834, opened 2026-06-03
#8833 fix: ensure dof effect is resolution independent

### LINKED ISSUE #8833: Depth of Field effect not resolution independent
When using Depth of Field.  The affect is dependent on the resolution, as such the effect varies greatly depending on the resolution
pixel ratio 1
Image
pixel ratio 3
Image

MAG-AdrianMeredith:
## Description

Fixes #8833

- Change the blur radius to be resolution independent against a reference resolution of 540.

High resolution
Screenshot 2026-06-03 at 15 16 12

Low resolution
Screenshot 2026-06-03 at 15 16 04

## Checklist


MAG-AdrianMeredith:
@mvaligursky What do you suggest we do with "high quality" switch.  From what I can tell it was a workaround for this issue and as such no longer makes sense.  Remove it?  Deprecate it?

src/extras/render-passes/frame-pass-dof.js
```diff
@@ -183,9 +183,10 @@ class FramePassDof extends FramePass {
         this.cocPass.focusDistance = this.focusDistance;
         this.cocPass.focusRange = this.focusRange;
 
-        // adjust blur sizes to give us the same results regardless of the quality (resolution)
+        // blur radius is expressed as a fraction of frame height in the blur shader, so the effect
+        // is now resolution- and quality-independent and needs no per-quality compensation
```
mvaligursky: The normalization actually happens in `RenderPassDofBlur` (against `referenceHeight`), not in the blur shader — this comment reads as if the shader does it. Minor reword:

```suggestion
        // blur radius is normalized against the blur pass reference height, so the effect is now
        // resolution- and quality-independent and needs no per-quality compensation
```

src/extras/render-passes/render-pass-dof-blur.js
```diff
@@ -107,8 +118,12 @@ class RenderPassDofBlur extends RenderPassShaderQuad {
 
         this.kernelId.setValue(this.kernel);
         this.kernelCountId.setValue(this.kernel.length >> 1);
-        this.blurRadiusNearId.setValue(this.blurRadiusNear);
-        this.blurRadiusFarId.setValue(this.blurRadiusFar);
+
+        // express the blur radius as a fraction of the reference height, so the effect is
+        // resolution-independent (the shader then aspect-corrects it to stay circular in pixels)
```
mvaligursky: If `referenceHeight` is ever set to 0 or a negative value, this uploads `Infinity`/`NaN` blur radii and breaks rendering. Guard it with a fallback to the default:

```suggestion
        // guard against an invalid reference height to avoid Infinity/NaN blur radii
        const referenceHeight = this.referenceHeight > 0 ? this.referenceHeight : 540;
        const invReferenceHeight = 1 / referenceHeight;
```

mvaligursky (commented):
Looks good — the approach is correct. Expressing the blur as a fraction of a reference height (rather than in source-texture texels) is the right way to make this resolution- and pixel-ratio-independent, and the `height/width` factor on the X step correctly keeps the kernel circular in screen pixels. The math checks out: in pixels both axes resolve to `kernel * blurRadius/540 * height`. Picking `referenceHeight = 540` to reproduce the old high-quality blur at 1080p is a sensible calibration that preserves previously authored `blurRadius` values. Perf impact is effectively nil (one scalar divide swapped for a vec2 divide; no change to tap count or texture sizes).

On the **high quality flag** question: I'd keep it. It's not the workaround for this issue — it controls the *internal resolution* of the DoF textures (half-of-full-res far texture + full-res composite vs. quarter-res), which is a genuine quality/perf trade-off independent of the resolution-dependence bug. The actual workaround was the `* (highQuality ? 1 : 0.5)` factor on `blurRadiusFar`, and you've already correctly removed that — now that blur is reference-height-relative, the far-texture resolution no longer changes blur strength, so the compensation is obsolete. So nothing further to deprecate.

Two small suggestions inline (both optional, ready to accept).

PR #8722, opened 2026-05-12
build: migrate js builds to esbuild

kpal81xd:
## Description
Migrate the engine JavaScript build targets from Rollup to esbuild for faster clean and watch builds.

- Adds esbuild target helpers and transform plugins for JSCC preprocessing, debug stripping, shader chunk
  processing, dynamic import handling, and import validation.
- Routes JS build and watch targets through esbuild, requires explicit JS type/format for direct `build.mjs`
  leaf commands, and uses Turbo for aggregate build/watch orchestration.
- Speeds type declaration generation by invoking TypeScript and Rollup APIs directly.
- Preserves Rollup-style build logging, including watch rebuild start and completion logs.
- Moves `build:treemap`, `build:treenet`, and `build:treesun` to esbuild metafiles via
  `esbuild-visualizer`; drops `build:treeflame` because there is no esbuild-compatible flamegraph
  template.

## Build comparison
Deltas compare current PR head `2f80a12a2` against `origin/main` `6df566a37` in isolated worktrees. Build speed is
a clean one-shot target build; watch build speed is the watcher rebuild after a content touch to `src/core/guid.js`.
JS artifact sizes were measured from target outputs before type generation is mixed into ESM tree totals.

This table uses `~same` for negligible changes: timing deltas under 10%, and size deltas that are byte-identical
or round below 0.1 KiB.

| Artifact / Target | Raw size Δ | Gzip size Δ | Build speed | Watch rebuild speed |
|---|---:|---:|---:|---:|
| rel UMD | -21.5 KiB (-0.6%) | +27.1 KiB (+3.7%) | 5.9× faster | 4.7× faster |
| rel ESM bundle+tree | -339.7 KiB (-4.5%) | +18.3 KiB (+1.0%) | 2.5× faster | 5.1× faster |
| dbg UMD | -3424.5 KiB (-15.2%) | -1054.5 KiB (-20.3%) | 11.1× faster | 8.1× faster |
| dbg ESM bundle+tree | -4698.2 KiB (-16.1%) | -1296.4 KiB (-18.6%) | 3.4× faster | 14.3× faster |
| prf UMD | -21.7 KiB (-0.6%) | +27.1 KiB (+3.7%) | 6.4× faster | 5.1× faster |
| prf ESM bundle+tree | -341.2 KiB (-4.5%) | +18.0 KiB (+1.0%) | 2.5× faster | 6.3× faster |
[...]

PR #8837, opened 2026-06-04
fix(build): preserve live bindings of mutable UMD exports (pc.app)

### LINKED ISSUE #8836: pc.app - is not available. Breaking projects.
Since 2.19, using `pc.app` is null, which breaks many of our projects, as we have use it a lot.

mvaligursky:
## Summary

Since 2.19.0, `pc.app` is permanently `null` in the **UMD / global build** (`playcanvas.js`), breaking any project or script that reads the global `pc.app` (e.g. the classic orbit-camera `mouse-input` script used by the WebXR AR Starter Kit). The ESM build is unaffected.

Fixes #8836. Also explains [forum thread #42337](https://forum.playcanvas.com/t/pc-app-null/42337).

## Root cause

2.19.0 migrated the JS builds to esbuild (#8722). The UMD output wraps the engine as an esbuild IIFE whose exports are live getters (`app: () => app`), then bridges that namespace to the outer UMD `exports` (= `window.pc`) via a hand-written footer:

```js
var pc = (() => { … __export(index_exports, { app: () => app, … }); return __toCommonJS(index_exports); })();
Object.assign(exports, pc);   // ← snapshots every getter to its load-time value
```

`Object.assign` invokes each getter once and copies the **value**. At module-load time no application exists, so `app` is `null`. `window.pc.app` becomes a frozen `null` data property; when an app is later constructed and the inner module's `app = this` runs, `window.pc.app` never updates.

This only affects **mutable** exports. `app` is the one that bites in practice. Scripts using `this.app` are unaffected (which is why the regression looked project-specific), but scripts reading the global `pc.app` break.

## Fix

Copy the property **descriptors** rather than the values, preserving the getter (and thus the live binding) on the UMD namespace:

```js
Object.defineProperties(exports, Object.getOwnPropertyDescriptors(pc));
```

## Verification

Built both UMD bundles and loaded them as a browser global (`window.pc`) in a `vm` sandbox, then constructed an `AppBase`:

| build | `pc.app` descriptor | `pc.app === new AppBase(canvas)` |
|-------|--------------------|----------------------------------|
| before (`Object.assign`) | `value = null` (frozen) | `false` — stays `null` |
| after (`Object.defineProperties`) | getter (live) | `true` — updates correctly |

All 1272 exports remain present; only the binding semantics for mutable exports change.

A regression test would ideally assert `pc.app` updates after construction in the UMD build, but the current mocha suite runs against `src/` (ESM, where live bindings already work), so there's no build-output test harness to hook into — noting it here as a follow-up.

### DIFF
--- a/utils/esbuild-build-target.mjs
+++ b/utils/esbuild-build-target.mjs
@@ -133,7 +133,11 @@ const getUmdBanner = (banner) => {
 };
 
 const getUmdFooter = () => {
-    return /* js */ `Object.assign(exports, pc);
+    // Copy the property descriptors rather than the values, so that live bindings of mutable
+    // exports (e.g. `app`, which is null until an application is created) are preserved on the
+    // UMD namespace. `Object.assign` would snapshot each getter to its load-time value, leaving
+    // `pc.app` permanently null. See https://github.com/playcanvas/engine/issues/8836
+    return /* js */ `Object.defineProperties(exports, Object.getOwnPropertyDescriptors(pc));
 }));`;
 };
 


kpal81xd (approved):
Looks good and makes sense since values were copied and not the references (which the descriptors include via getters/setters). Just testing in the Editor to confirm

kpal81xd:
Verified all good in the Editor - Maybe worth adding a test

PR #8842, opened 2026-06-04
Fix AssetRegistry URL lookups by type

### LINKED ISSUE #8489: AssetRegistry.getByUrl() returns wrong asset type for same URL (texture vs cubemap)
Hi team,
I encountered an issue with AssetRegistry.getByUrl() when working with cubemap assets:
A PNG file URL is used for both a texture asset and a cubemap asset in the registry.
getByUrl(url) only returns the last added asset (e.g., texture instead of cubemap), even though the cubemap asset with the same URL exists.
Since _urlToAsset is a Map that overwrites entries with the same URL, getByUrl() can’t distinguish assets by type, which makes it impossible to reliably retrieve cubemap assets via URL (a common use case for skyboxes).
Is there a recommended way to get an asset by URL + type (instead of just URL) to avoid this conflict? Or could getByUrl() be enhanced to support type filtering?
Thanks!

puneetdixit200:
## Description
Store all assets registered for the same file URL so `AssetRegistry#getByUrl(url, type)` can return the asset matching a requested type. Untyped `getByUrl(url)` keeps the existing last-added behavior for compatibility.

Fixes #8489

## Testing
- `npm test -- test/framework/asset/asset-registry.test.mjs --grep "#getByUrl"`
- `npm test -- test/framework/asset/asset-registry.test.mjs`
- `npm run lint`
- `npm run build:types`
- `npm run test:types`
- `git diff --check`

## Checklist

### DIFF
--- a/src/framework/asset/asset-registry.js
+++ b/src/framework/asset/asset-registry.js
@@ -179,7 +179,7 @@ class AssetRegistry extends EventHandler {
     _idToAsset = new Map();
 
     /**
-     * @type {Map<string, Asset>}
+     * @type {Map<string, Set<Asset>>}
      * @private
      */
     _urlToAsset = new Map();
@@ -264,8 +264,12 @@ class AssetRegistry extends EventHandler {
 
         this._idToAsset.set(asset.id, asset);
 
-        if (asset.file?.url) {
-            this._urlToAsset.set(asset.file.url, asset);
+        const url = asset.file?.url;
+        if (url) {
+            if (!this._urlToAsset.has(url)) {
+                this._urlToAsset.set(url, new Set());
+            }
+            this._urlToAsset.get(url).add(asset);
         }
 
         if (!this._nameToAsset.has(asset.name)) {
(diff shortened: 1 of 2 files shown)


src/framework/asset/asset-registry.js
```diff
@@ -356,12 +365,24 @@ class AssetRegistry extends EventHandler {
...
-    getByUrl(url) {
-        return this._urlToAsset.get(url);
+    getByUrl(url, type) {
+        const assets = this._urlToAsset.get(url);
+        if (!assets) {
+            return undefined;
+        }
+
+        let result;
+        assets.forEach((asset) => {
```
Maksims, line 380: Please replace with a simple for..of loop, with early return inside of it.

mvaligursky (commented):

---

Looks correct and well-tested. Untyped `getByUrl(url)` preserves the last-added semantics (Set iteration is insertion order, `forEach` keeps overwriting `result`), all external callers go through the public method, and the `Set`-per-url removal is actually more correct than the old map — removing one of two assets sharing a url no longer wipes the other. Nothing blocking. Two minor notes:

1. **`getByUrl` iterates the whole set with no early exit** and returns the *last* match. That's intentional to keep untyped compat, but for the typed branch a `for...of` with an early `break` would be cheaper and clearer about intent (and would return the first match of that type). Sets are tiny so the perf difference is negligible — purely a readability nit.

2. **Pre-existing (not introduced here), but now more relevant:** the registry indexes by `asset.file.url` at `add` time and never re-indexes if `file.url` changes later (`remove` then uses the *current* url, so a changed-url asset leaves a stale set entry). Harmless single-asset before; with multiple assets per url now supported, a stale entry could shadow a live one. Out of scope for this PR — just flagging in case it's worth a follow-up.

PR #8865, opened 2026-06-10
fix(mini-stats): render correctly with multiple cameras

kpal81xd:
## Problem

MiniStats drew itself into the UI layer via the immediate-draw system. That injects the overlay into the per-camera culled list of only the first camera rendering the UI layer, and its clip-space quads map to that camera's viewport/scissor rather than the canvas. With two or more cameras the overlay gets squeezed into the first camera's sub-viewport and, depending on camera rects/clears/render targets, can be hidden entirely.

## Fix

The overlay shader uses no camera uniforms, so it is now decoupled from cameras and layers entirely:

- `AppBase.renderComposition` fires a new `framegraphbuild` event between building and rendering the frame graph, so passes can be appended for the frame.
- `Render2d` draws its quad buffer directly to the backbuffer in a dedicated `RenderPassMiniStats` (load, no clear, full-canvas viewport/scissor since the frame-graph compiler may merge it into the previous camera's pass), with bind-group support for WebGPU. Adding it via the frame graph lets the compiler handle the MSAA store/resolve chain.
- `MiniStats` appends the pass on `framegraphbuild`, and `destroy()` now releases the `Render2d` GPU resources (previously leaked).

The overlay no longer requires any camera to render the UI layer, and always draws on top, after all cameras.

## Verification

Tested a split-screen two-camera repro on WebGL2 and WebGPU: previously squished/garbled into the left viewport, now full-size at the bottom-left of the canvas. Single-camera placement is unchanged, expanded graph mode renders correctly, and unit tests/lint pass.

### DIFF
--- a/src/extras/mini-stats/mini-stats.js
+++ b/src/extras/mini-stats/mini-stats.js
@@ -1,7 +1,6 @@
 import { math } from '../../core/math/math.js';
 import { Texture } from '../../platform/graphics/texture.js';
 import { ADDRESS_REPEAT, FILTER_NEAREST } from '../../platform/graphics/constants.js';
-import { LAYERID_UI } from '../../scene/constants.js';
 import { CpuTimer } from './cpu-timer.js';
 import { GpuTimer } from './gpu-timer.js';
 import { StatsTimer } from './stats-timer.js';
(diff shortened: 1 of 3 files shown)


mvaligursky:
Lets just close this and don't mention a full screen pass suggestion again :)

kpal81xd:
Nobody saw this

PR #8876, opened 2026-06-11
refactor(components): eliminate ComponentData classes and engine schema usage

willeastcott:
﻿## Description

Completes the multi-PR effort to eliminate the legacy `ComponentData` classes (latest prior step: #8873). After that PR, all 23 `*ComponentData` classes defined only `enabled = true`, so the schema/buildAccessors machinery was generating exactly one accessor per component — `enabled`, backed by the store data object.

### What changed

- **Base `Component`** now has a `_enabled = true` field and a real `enabled` getter/setter that fires the same `'set'` event the generated accessor did (the unconditional fire is load-bearing for script and zone, which listen for `set_enabled` even when the value is unchanged).
- **Base `ComponentSystem`**: `DataType` is now optional in `addComponent` (an empty object is stored when absent), and `initializeComponentData`'s `properties` parameter is now optional — the descriptor loop is the legacy path for external schema-based components, and when it is omitted the base initializer applies the `enabled` state from the data block directly. Engine systems just call `super.initializeComponentData(component, data)`; element, sprite and the layout systems, which deliberately initialize `enabled` before their asset-loading setters, opt out by passing an explicit empty list.
- **All 23 systems**: removed `_schema`, `this.schema`, `this.DataType`, the `data.js` imports and the module-level `Component._buildAccessors(...)` calls; **deleted all 23 `data.js` files**. They were never exported from `index.js`, so there is no public API break.
- **Targeted fixes** for paths the migration would otherwise have silently broken:
  - The animation and sprite system update loops read `enabled` off the (now-empty) store record — rewritten to read the component.
  - `JointComponent.initFromData` raw-wrote `this._enabled`, which would clobber the new backing field and suppress the `onDisable` transition (leaking the Ammo constraint when a joint is added disabled) — `enabled` is removed from its property list.
  - audio-listener and joint were the only systems relying on the default `cloneComponent` (which clones from store data); both get explicit overrides.
  - Removed the now-redundant custom `enabled` accessors from `ElementComponent` and `ScriptComponent` (the base accessor is identical to ScriptComponent's old one).

### playcanvas-spine compatibility

The legacy machinery (`buildAccessors`/`_buildAccessors`, the `data` getter, the constructor schema fallback, store data records, schema-driven init, default `cloneComponent`) is **kept** in the base classes and annotated as a legacy path. playcanvas-spine defines a 9-property schema + `DataType` and relies on all of it via the base constructor fallback. Since no engine component exercises that path anymore, a new test (`test/framework/components/legacy-component.test.mjs`) locks it down: per-instance accessor building, data-backed `enabled`, `set_*` events, and default cloning from live store data. Once spine is modernized, a follow-up PR can delete the whole mechanism.

### Behavior notes

- `_enabled` is now an own enumerable property on every component (previously the value lived in the store and the accessors were non-enumerable).
- `component.data` returns `{}` for engine components; external code that raw-wrote `component.data.enabled` (already documented as not firing events) now has no effect.
- Reading `component.enabled` inside a subclass constructor previously threw (no store record yet); it now returns `true`.

### Verification

- `npm test` — 1,814 passing, including the 5 new legacy-path tests and the "clones the enabled state" tests for all 23 component types
- `npm run lint`, `npm run build:types`, `npm run test:types` — clean
- Spine smoke test in the examples browser (spineboy example with the bundled `playcanvas-spine.3.8.js`): accessors built via the constructor fallback, portal animation renders and advances, `entity.spine.enabled` toggling round-trips through the data store, and `entity.clone()` produces a working second spine instance

## Checklist


LeXXik:
Just a note from my experience creating custom components for the engine. I found component schemas useful, as it was able to catch malformed options early. For example:

```js
entity.addComponent('component', {
    optionA: true,
    opionB: true
});
```

Internally the engine would initialize component data using the provided options object. I would then verify each property of the object against the schema and issue a warning, if the property is not found in the schema (e.g. `opionB` typo). Perhaps, we could have something similar in debug-only builds.

willeastcott:
Thanks for your thoughts on this, @LeXXik. We can probably address this in a follow up PR. 🤔

PR #8886, opened 2026-06-12
refactor(collision): deduplicate rigid body / trigger rebuild

willeastcott:
Continues the component-consistency series (#8871, #8876, #8878-#8882). Sibling of #8883 and #8884.

## Changes

The block that rebuilds the rigid body or re-creates the trigger after a collision shape change appeared nearly verbatim in two places:

- `CollisionSystemImpl.recreatePhysicalShapes`
- `CollisionMeshSystemImpl.doRecreatePhysicalShape`

This extracts the two bodies into `_recreateBody(entity)` and `_recreateTrigger(entity, component)` on the base `CollisionSystemImpl`, while deliberately keeping the branch conditions at each call site.

## Why the gating stays at the call sites

The two call sites are not identical: the base implementation skips trigger creation for compound children (`else if (!component._compoundParent)`), while the mesh implementation creates the trigger unconditionally. That difference is observable behavior - a mesh collider *can* be a compound child, since `_addEachDescendant` assigns `_compoundParent` to any descendant collision without a rigidbody regardless of type. Extracting only the bodies preserves behavior by construction; the asymmetry is now called out with a comment on the mesh side.

Verified with `npm test` / `npm run lint` / `npm run build` / `npm run build:types` + `npm run test:types`. The base-implementation block sits behind the Ammo gate, so unit tests cannot reach it; the mesh trigger branch is exercised Ammo-less by the existing model tests.


LeXXik:
I am not sure I see the value in this PR. Is it to reduce the code base? +25 lines vs -22.

willeastcott:
> I am not sure I see the value in this PR. Is it to reduce the code base? +25 lines vs -22.

Fair question. The goal isn't line count. The two copies of this block had actually drifted: the base path skips trigger creation for compound children (`!component._compoundParent`), the mesh path doesn't. Before this change you'd only spot that by diffing two blocks ~400 lines apart, with no way to tell if it was intentional. Now the shared logic lives in one place and the only thing left at each call site is the difference, with a comment marking it as deliberate.

It also gives the disableSimulation → createBody → enableSimulation sequence a single home, so any future change to the rebuild ordering (there's a known latent bug in this area) gets made once instead of twice.

Happy to close it if you don't think that's worth the indirection, though

mvaligursky:
I usually de-duplicate to keep all paths consistent even if its just few lines. I'd merge.

PR #8891, opened 2026-06-12
feat(joint): promote JointComponent to public API with typed joints

willeastcott:
## Description

Rewrites the private (`@ignore`) JointComponent into a public, type-based joint API, ready for editor exposure. The whole surface is tagged `@alpha` to signal it may still change before it stabilizes.

### API

- **`joint.type`** ∈ `fixed | ball | hinge | slider | 6dof` (`JOINTTYPE_*` constants), each backed by the most appropriate Bullet constraint (`btFixedConstraint`, `btConeTwistConstraint`, `btHingeConstraint`, `btSliderConstraint`, `btGeneric6DofSpringConstraint`). Per-type backends are plain module-scope dispatch tables next to the component, not system-hosted impl classes.
- **The joint entity's own transform defines the joint frame** — its local X axis is the primary axis (hinge rotation, slider travel, ball twist). `entityA`/`entityB` are the constrained bodies (accepting `Entity` or GUID); `entityB = null` pins to the world.
- **Per-type properties**: `enableLimits`/`limits` + `motorSpeed`/`maxMotorForce` (hinge/slider — the motor engages whenever `maxMotorForce > 0`), `swingLimitY`/`swingLimitZ`/`twistLimit` (ball), and per-axis `linearMotionX/Y/Z`-style enums (`MOTION_LOCKED/LIMITED/FREE`), `linearLimitsX/Y/Z` Vec2 pairs and Vec3 `linearStiffness`/`linearDamping`/`linearEquilibrium` spring families (6dof; a spring acts on an axis when its stiffness component is > 0). Angular families mirror linear, in degrees.
- **`breakImpulse`** with a `break` event and read-only `isBroken`; **`refreshFrames()`** re-captures frames from current transforms and re-arms broken joints. The raw Ammo constraint is intentionally **not** exposed (kept `@ignore`, like `RigidBodyComponent.body`) so the public surface stays backend-agnostic for a future Jolt/PhysX backend.
- No redundant `enable*` booleans: a spring acts when its stiffness is > 0 and a motor when its `maxMotorForce` is > 0, so on/off is derived from the magnitude rather than a separate flag.

### Implementation notes for reviewers

- **World pinning routes through a shared static fixed body** (`JointComponentSystem.getFixedBody()`, Bullet's own pattern) because the single-body `btConeTwistConstraint` constructor does not world-convert its frame origin — verified empirically: world-pinned ball joints anchored at the world origin and flew toward it.
- **Break detection is three-tier**: `isEnabled()` → `getAppliedImpulse()` → anchor-separation measurement. The bundled ammo build binds *neither* of the first two on constraints, so the stock build relies on anchor separation (works for fixed/ball/hinge/slider; 6dof warns once and still breaks physically, just without the event).
- **Self-healing lifecycle**: joints wait for rigid bodies to enter the simulation (pending set drained pre-step) and tear down synchronously before any `btRigidBody` is destroyed, via two new internal `RigidBodyComponent` events (`simulationenabled`/`simulationdisabled`) fired from `enableSimulation`/`disableSimulation` — every body-destruction path passes through the latter.
- **Frames are derived from unscaled world position + rotation**, matching how rigid bodies treat entity scale.
- Live property changes apply to existing constraints where Bullet allows and **activate the constrained bodies** (sleeping islands otherwise ignore motor/limit changes).
- Entity references track destruction, **clone with full fidelity and remap on entity duplication** (button-style `resolveDuplicatedEntityReferenceProperties`).
- Fixes several latent bugs in the old component, including `!enableCollision` being passed as Bullet's `useLinearReferenceFrameA` constructor argument, angular equilibrium applied in radians while limits were converted from degrees, setter crashes before constraint creation, and scene-data initialization never having worked (GUIDs were written to backing fields unresolved).

### Breaking changes

The component was `@ignore`'d, so no public API breaks. For anyone using it via forum recipes: `type` now exists and defaults to `fixed` (old behavior = `type: '6dof'`), `enableCollision` default flipped `true` → `false` (Unity/Godot convention), `breakForce` renamed `breakImpulse` (it always was an impulse threshold), and the 6dof spring booleans are gone (stiffness > 0 enables a spring).

### Verification

- 19 new ammo-less unit tests (`test/framework/components/joint/`); full suite, lint, `build:types`/`test:types` all pass.
- Two new examples, used as live verification rigs: **physics/joints** (door hinge with limits, motorized windmill, ball chain with asymmetric swing limits, motorized slider patrol, 6dof spring, breakable wall firing `break` events) and **physics/ragdoll** (three ragdolls cloned from one template — exercising duplication remap — with mouse/touch grab-and-drag via a ball joint to a kinematic anchor, and a 1/120 fixed timestep to prevent limb tunnelling).
- Behaviors verified by stepping the simulation programmatically: hinge motor runs at exactly the requested speed, door slams clamp at the configured limit, ball swing anisotropy matches the documented Y/Z mapping, slider respects travel limits, the motor free/drive/brake states all behave (force 0 = free, force > 0 + speed 0 = brake), welds break with events on impact, rigidbody disable/enable destroys and resurrects constraints, and `refreshFrames()` re-attaches broken joints.

### Follow-ups (separate repos)

- **editor**: joint inspector, schema entry, component menu, reference docs.
- **playcanvas/ammo.js**: bind `btTypedConstraint::isEnabled()` (one-line idl change) for exact break detection — the engine feature-detects and upgrades automatically; optionally bind `btGeneric6DofSpring2Constraint` for physical spring damping.

## Checklist


src/framework/components/joint/component.js
```diff
@@ -3,512 +3,1633 @@ import { math } from '../../../core/math/math.js';
...
+        return this._enableLimits;
+    }
+
+    /**
+     * Sets the lower and upper limit of the joint's primary degree of freedom. For hinge joints,
+     * these are angles of rotation about the joint's X axis, in degrees. For slider joints, these
+     * are distances along the joint's X axis, in meters. Only used when {@link enableLimits} is
+     * true. Defaults to `[-45, 45]`.
+     *
+     * @type {Vec2}
```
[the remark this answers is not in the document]
willeastcott: Intentional - the documented type is deliberately `Vec2`. The array form is supported only for scene/JSON deserialization (where vectors arrive as `[x, y]`), matching CollisionComponent, whose `halfExtents`/`linearOffset` are typed `Vec3` while also accepting arrays. Keeping the public surface prescriptive rather than widening it to `Vec2 | number[]`.

src/framework/components/joint/system.js
```diff
@@ -5,35 +5,156 @@ import { JointComponent } from './component.js';
...
+            const inertia = new Ammo.btVector3(0, 0, 0);
+            const info = new Ammo.btRigidBodyConstructionInfo(0, motionState, shape, inertia);
+            this._fixedBody = new Ammo.btRigidBody(info);
+            Ammo.destroy(info);
+            Ammo.destroy(inertia);
+            Ammo.destroy(transform);
+        }
+
+        return this._fixedBody;
     }
```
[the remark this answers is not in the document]
willeastcott: Done in 945dec8 - renamed to `_getFixedBody` and tagged `@ignore`, so the shared Ammo body is no longer part of the public surface. Used `@ignore` rather than `@private` because the component calls it cross-class (same as the other internal component/system seam methods).

src/framework/components/rigid-body/component.js
```diff
@@ -741,6 +745,10 @@ class RigidBodyComponent extends Component {
             body.forceActivationState(BODYSTATE_DISABLE_SIMULATION);
 
             this._simulationEnabled = false;
+
+            // internal event consumed by the joint system to destroy constraints before
+            // the body they reference is removed from the dynamics world or destroyed
```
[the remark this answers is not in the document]
willeastcott: Done in 945dec8 - corrected the comment. You are right: the body has already been removed from the dynamics world by the time the event fires. The teardown's purpose is to drop the constraint while the body is still a valid object, before it is later destroyed or rebuilt - the comment now says that.

src/framework/components/joint/system.js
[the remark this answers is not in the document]
willeastcott: Follow-up: on maintainer feedback, kept the plain name `getFixedBody` (no underscore) with `@ignore` instead - the underscore would imply class-private, but it is part of the component/system contract. `@ignore` keeps it off the public docs/API surface.

src/framework/components/joint/component.js
```diff
@@ -3,512 +3,1612 @@ import { math } from '../../../core/math/math.js';
...
+const _vec3b = new Vec3();
+const _quat = new Quat();
+
+// anchor separation, in meters, beyond which a constraint that bullet has internally broken is
+// considered detectably broken - far above the violation a live constraint exhibits under load
+const _breakSeparation = 0.2;
+
+// assigns an incoming Vec2 or [x, y] array to target, returning true if it changed
+function setVec2(target, value) {
+    const x = value instanceof Vec2 ? value.x : value[0];
```
LeXXik, line 32: Do we need those checks? What is the case when value is an Array?

willeastcott: It's when loading from scene data...you get an Array then, not a Vec2.

LeXXik: I suppose Editor needs this.

test/framework/components/joint/component.test.mjs
```diff
@@ -0,0 +1,417 @@
...
+            expect(cloneJoint.entityB).to.equal(cloneRoot.findByName('bodyB'));
+            expect(cloneJoint.entityA).to.not.equal(bodyA);
+            expect(cloneJoint.entityB).to.not.equal(bodyB);
+        });
+
+    });
+
+    describe('lifecycle', function () {
+
+        it('survives a disable and enable round trip', function () {
```
LeXXik: This test is actually misleading, as it doesn't go through the critical code path (when a constraint exists).

willeastcott: Updated - thanks.

src/framework/components/joint/component.js
willeastcott: Yes.

willeastcott:
@LeXXik I think I'm happy with this PR now. Any further concerns from you about it?

LeXXik:
Looks good! I like your approach with multi-joints. Looking forward to see Editor's support for it.

willeastcott:
> Looks good! I like your approach with multi-joints. Looking forward to see Editor's support for it.

Yeah, I tried to shape the API to map well onto an Editor component UI. It should work pretty well! 😄

PR #8895, opened 2026-06-13
fix(text): render Devanagari and other complex scripts correctly

### LINKED ISSUE #4102: Hindi text does not render correctly
Forum: https://forum.playcanvas.com/t/hindi-language-in-game-text-issues/24712
Project: https://playcanvas.com/editor/scene/1358357

It should look like this: 
image

But looks like this
image

I've check the font on Google and it looks correct there: https://fonts.google.com/noto/specimen/Noto+Sans+Devanagari

image

willeastcott:
Fixes #4102.

`string.getSymbols` split Devanagari into one symbol per code point, so vowel signs (matras) detached from their consonant and conjuncts never formed — text rendered garbled and out of order as reported.

`getSymbols` now forms proper grapheme clusters: a base character plus its trailing combining marks (`\p{M}`), `ZWJ`/`ZWNJ` joiners and virama-joined consonants. Because `CanvasFont` rasterizes each symbol with the browser's text shaper, whole syllables now shape into a single correctly-composed glyph (Devanagari, plus Arabic, Tamil, etc.).

Bitmap/MSDF fonts have per-code-point atlases that don't contain a multi-code-point cluster, so `TextElement` expands any font-missing cluster back into its code points — they render exactly as before.

Tested: added `getSymbols` unit tests for Devanagari vowel signs and conjuncts; existing emoji/flag/ZWJ cases are unchanged; verified end-to-end that a real `CanvasFont` atlas now contains correctly-shaped Hindi clusters.

### DIFF
--- a/src/core/string.js
+++ b/src/core/string.js
@@ -6,6 +6,7 @@ const HIGH_SURROGATE_BEGIN = 0xD800;
 const HIGH_SURROGATE_END = 0xDBFF;
 const LOW_SURROGATE_BEGIN = 0xDC00;
 const LOW_SURROGATE_END = 0xDFFF;
+const ZERO_WIDTH_NON_JOINER = 0x200C;
 const ZERO_WIDTH_JOINER = 0x200D;
 
 // Flag emoji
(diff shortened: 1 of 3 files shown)


mvaligursky (commented):

---

Looks correct and the design (cluster in `getSymbols`, expand-on-missing in `TextElement`, self-consistent with `CanvasFont._renderAtlas` which keys its atlas via the same `getSymbols`) is sound. A few things to consider, none blocking:

1. **Behavior change for bitmap/MSDF fonts that the description says is unchanged.** A multi-code-point ZWJ emoji cluster (e.g. `👨‍👩‍👧`) was previously a single symbol → one missing-glyph fallback in a bitmap atlas. After this PR `TextElement` expands it into its individual code points, so each is now looked up separately (components may render, the ZWJ falls back). That's arguably an improvement, but it's *not* "exactly as before" for that case — worth confirming it's intended.

2. **`COMBINING_MARK.test(ch)` only sees one UTF-16 unit.** Combining marks outside the BMP (e.g. variation-selector supplement U+E0100–U+E01EF) are passed a lone surrogate and won't match `\p{M}`, so they won't cluster. Pre-existing limitation, not a regression — just noting the boundary.

3. **Optional / future:** `Intl.Segmenter` (`granularity: 'grapheme'`) now has broad support and would replace the hand-maintained `VIRAMAS` set with full UAX-29 segmentation. Understandable to avoid it for browser/test-env support reasons, but it'd be the lower-maintenance path long-term.

willeastcott:
I'm not happy to merge this. Closing.

PR #8898, opened 2026-06-14
fix(physics): clear cached collisions when a collision component is removed

### LINKED ISSUE #5797: The triggerleave event doesn't work as expected after reloading scene
The `triggerleave` event doesn't work as expected after reloading the scene, when the trigger entity is inside a rigidbody entity.

A workaround is to enable the trigger entity (or collision component) a frame later, but I don't think this is possible in all situations.

I think this bug can easily break projects that use a lot of trigger entities.

Here is a project with the issue:
https://playcanvas.com/editor/scene/1894025

Here is a project with a workaround:
https://playcanvas.com/editor/scene/1894025

willeastcott:
## Description

The rigid-body `collisions` map is keyed by entity GUID and was never cleared when an entity is destroyed. Reloading the *same* scene recreates entities with identical GUIDs (from `resource_id`) within a single task — no physics step runs in between — so the stale entry, still referencing the destroyed entity, was reused by the new same-GUID entity.

From then on `triggerleave` / `collisionend` never fired, because `_cleanOldCollisions` decides which leave event to fire by inspecting the cached (now destroyed) entity, whose `trigger` / `body` had already been deleted on destroy.

This clears the entity's cached collision on collision `beforeremove`. Every physics body entity has a collision component, so this single hook covers both triggers and rigid bodies. Added a regression test.

Fixes #5797

## Checklist

### DIFF
--- a/src/framework/components/collision/system.js
+++ b/src/framework/components/collision/system.js
@@ -732,6 +732,12 @@ class CollisionComponentSystem extends ComponentSystem {
     onBeforeRemove(entity, component) {
         this.implementations[component.type].beforeRemove(entity, component);
         component.onBeforeRemove();
+
+        // discard any stored collisions keyed to this entity so a later entity that reuses the
+        // same GUID (e.g. after reloading the same scene) does not inherit stale tracking
+        if (this.app.systems.rigidbody) {
+            this.app.systems.rigidbody.clearEntityCollisions(entity);
+        }
     }
 
     onRemove(entity) {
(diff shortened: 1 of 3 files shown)


willeastcott:
Credit to @defektu, who diagnosed the root cause in the issue thread — that reloading the *same* scene recreates entities with the same GUIDs and leaves stale entries in the rigid-body collision tracking ("a flaw with collision objects' others list… not propagating").

This PR fixes that at the source (clearing the cached collision on `beforeremove`) rather than via the `changeScene` timing change proposed there. That keeps it robust for `loadSceneHierarchy` and programmatic destroy/recreate too, and avoids regressing `changeScene`'s behavior of applying scene settings and keeping the old scene visible until the new data is ready.

PR #8907, opened 2026-06-15
Move XR per-view data from the XR manager to the Camera (RenderView)

### LINKED ISSUE #5011: Move XR Manager views to the Camera refactor
From https://github.com/playcanvas/engine/pull/4837#pullrequestreview-1272339732

> Currently, when XR session is active, the XR manager modifies the camera to point to itself. The reason is to allow forward-renderer to access an array of views that the manager stores. I'd want to move the views to camera, and the XR manager would just update them. But that's a separate PR.

mvaligursky:
Closes #5011.

### Background

When an XR session was active, the XR manager pointed the scene `Camera` at itself (`camera.camera.xr = manager`) purely so the renderer could reach the manager's per-view array. That made the **scene** layer (`renderer`, `forward-renderer`, unified gsplat) reach up into the **framework** XR module. This moves the per-view data onto the `Camera`, so the XR manager just updates it.

### Changes

- **New `RenderView` class** (`src/scene/render-view.js`) — a single view's projection / view matrices, viewport, and the derived "off" matrices, plus `updateTransforms()`. `XrView` now extends it and populates it from the WebXR frame (`setView` / `setViewport`). `RenderView` stays public because `XrView.viewport` was already public.
- **`Camera` owns the views** — `camera.xrViews` is the per-view list (`null` when not in XR), and `xrActive` is derived from it, replacing the manager back-pointer. `Camera.updateViewTransforms()` refreshes the per-view derived matrices (called by the renderer and the gsplat passes).
- **Consumers updated** — `renderer.js`, `forward-renderer.js` (`_isMultiview`), and the unified gsplat code read `camera.xrViews` / `xrActive` instead of `camera.xr.*`. `CameraComponent.endXr` now goes through `app.xr`.
- **Examples** — the two `xr-views` examples construct `pc.RenderView` directly instead of mocking the manager's `xr` object.

### Notes

- Relies on the dynamic-uniform-buffer bind-group fix from #8906 (already merged) for the WebGPU multiview `xr-views` example to render all views correctly.
- Net `-94` lines (the `RenderView` extraction removes duplication).

### DIFF
--- a/src/scene/camera.js
+++ b/src/scene/camera.js
@@ -522,7 +522,6 @@ class Camera {
      * null when not rendering an XR session. Set by the XR manager.
      *
      * @param {RenderView[]|null} value - The per-view list, or null when not in XR.
-     * @ignore
      */
     set xrViews(value) {
         // the projection source switches between XR and non-XR when XR activeness changes, so the
@@ -535,7 +534,6 @@ class Camera {
 
     /**
      * @type {RenderView[]|null}
-     * @ignore
      */
     get xrViews() {
         return this._xrViews;
@@ -545,7 +543,6 @@ class Camera {
      * True while an XR session owns this camera (equivalent to {@link Camera#xrViews} being set).
      *
      * @type {boolean}
-     * @ignore
      */
     get xrActive() {
         return this._xrViews !== null;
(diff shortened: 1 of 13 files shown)


src/scene/camera.js
```diff
@@ -512,15 +517,38 @@ class Camera {
...
 
-    set xr(newValue) {
-        if (this._xr !== newValue) {
-            this._xr = newValue;
+    /**
+     * Sets the list of {@link RenderView}s this camera renders with (one per XR eye/screen), or
+     * null when not rendering an XR session. Set by the XR manager.
+     *
+     * @param {RenderView[]|null} value - The per-view list, or null when not in XR.
+     * @ignore
```
willeastcott: You don't need to mark any of the interface as ignore because the class has it.

PR #2797, opened 2021-01-29
Remove masking examples

willeastcott:
The UI masking examples located in `examples/masking` have never been exposed on playcanvas.github.io. They are completely non-obvious in what they are showing. This PR deletes the examples. Instead, we should submit a subsequent PR that introduces clear, good-looking examples that illustrate the UI masking feature.

I confirm I have signed the [Contributor License Agreement](https://docs.google.com/a/playcanvas.com/forms/d/1Ih69zQfJG-QDLIEpHr6CsaAs6fPORNOVnMv5nuo0cjk/viewform).

PR #8923, opened 2026-06-18
Add UI element masking example

### LINKED ISSUE #2799: Add an example showing UI masking
We don't have any engine-only examples showing how to achieve masking with the element system.

mvaligursky:
Adds an engine-only example demonstrating UI masking with the element system, addressing the gap noted in #2799 (the original masking examples were removed in #2797).

**Examples:**
- New `user-interface/masking` example with two demos on a screen-space UI:
  - Rectangular mask — a row of square tiles scrolls horizontally behind a rectangular mask element and is clipped to its bounds.
  - Alpha-shaped mask — content is clipped to the opaque pixels of a heart texture (the mask material's alpha test drives the shape), with the heart "beating" via animated size.
- `Mask` toggle (default on) to compare clipped vs unclipped content, plus an `Animate` toggle to pause motion.
- Reuses existing assets (`colors.webp`, `heart.png`) and includes generated thumbnails.

Closes #2799.


willeastcott:
This looks like a test rather than an example. Can we try to make new examples look a bit more relatable to engine users that are trying to achieve specific features/effects in their apps?

PR #8929, opened 2026-06-19
feat(assets): throttle concurrent asset requests

### LINKED ISSUE #5258: Throttle large number of assets requests?
Some browsers like chrome, will get network resource errors if there are too many network requests at once (usually around 1000 on my Macbook Pro). This will need to network errors and some assets not being loaded.

Projects that are commonly affected by this are those that preload many assets at once (eg https://www.warbands.io/)

Perhaps we should throttle the number (as a project option?) that are currently in flight to reduce possible errors on devices?

mvaligursky:
Adds a limit on the number of concurrent asset requests, so large preloads no longer fail with `net::ERR_INSUFFICIENT_RESOURCES` — browsers reject requests once too many are in flight at once (empirically reproducible well below the asset counts some projects preload). Fixes #5258.

**Changes:**
- Throttle requests at the HTTP layer — the single choke point all resource handlers go through — so it applies transparently to every asset type with no changes to the registry, handlers, or preload. Requests beyond the limit are queued and dispatched as earlier ones complete.
- The `XMLHttpRequest` is still created and returned synchronously; only the `.send()` call is deferred, so progress/abort behaviour is unchanged.
- Default limit is **128**; `0` disables throttling (previous behaviour). It is a process-global limit, matching the browser's per-process resource limit, and covers all XHR-based loads (the large majority of assets; the ``-element texture fallback, `<audio>`, `<script>` and bundle `fetch` paths are browser-managed and not affected).
- Editor projects can set it via a new `maxConcurrentRequests` application property, parsed alongside the existing `maxAssetRetries`.

**API Changes:**
- Add `app.loader.maxConcurrentRequests` (getter/setter) to read or override the limit at runtime:
  ```javascript
  app.loader.maxConcurrentRequests = 50; // at most 50 requests in flight at once
  app.loader.maxConcurrentRequests = 0;  // disable throttling
  ```

**Related:**
- Editor follow-up to surface this in the Network settings panel: playcanvas/editor#2119
- User-manual docs update tracked separately on the developer-site.

### DIFF
--- a/src/framework/app-base.js
+++ b/src/framework/app-base.js
@@ -805,6 +805,11 @@ class AppBase extends EventHandler {
             this.loader.enableRetry(props.maxAssetRetries);
         }
 
+        // configure the concurrent request limit - overrides the default (0 is valid, meaning unlimited)
+        if (typeof props.maxConcurrentRequests === 'number' && props.maxConcurrentRequests >= 0) {
+            this.loader.maxConcurrentRequests = props.maxConcurrentRequests;
+        }
+
         // TODO: remove this temporary block after migrating properties
         if (!props.useDevicePixelRatio) {
             props.useDevicePixelRatio = props.use_device_pixel_ratio;
--- a/src/framework/handlers/loader.js
+++ b/src/framework/handlers/loader.js
@@ -1,4 +1,5 @@
 import { Debug } from '../../core/debug.js';
+import { http } from '../../platform/net/http.js';
 
 /**
  * @import { AppBase } from '../app-base.js'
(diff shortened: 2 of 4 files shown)


mvaligursky:
Thanks for the review — addressed all five points in `50f2efc2e`:

1. **Per-request slot state stored on `options`** → moved to the per-request `xhr` (`xhr._slotHeld`). Reusing an `options` object across requests (the `retry: false` path, which isn't cloned) can no longer corrupt accounting. Added a regression test that fires 4 concurrent requests sharing a single `options` object and asserts `_activeRequests` returns cleanly to 0.
2. **Unconditional `options.error(...)` in the send-catch** → guarded with `typeof options.error === 'function'`.
3. **`_activeRequests` doc** → clarified it only counts slot-accounted (throttled async) requests and stays 0 when throttling is disabled.
4. **`_pump()` O(n²) `Array#shift`** → now dequeues via a `_sendQueueHead` index with periodic compaction, so draining a large queue is O(n).
5. **Setter validation** → `ResourceLoader#maxConcurrentRequests` clamps to a non-negative integer (`Math.max(0, Math.floor(value)) || 0`), with `Infinity` still supported, matching `enableRetry()`.

Full test suite passes (1850) including the new per-request slot-tracking test.

@copilot review

Issue #830, opened 2017-01-09
SDF fonts: possible solution for downscaling


yard:
Hi guys,

I've been recently playing around with rasterizing the fonts with msdf textures in playcanvas. SDF is well-known for its ability to handle upscaling well, but performs poorly when it comes to downscaled shapes. There are solutions using supersampling of various sorts, but they don't seem to be quite universal and require a lot of adjustments.

An idea: why don't we use alpha channel to store original glyph's pixel opacity and fall back to it when font size is lower than the char size in msdf map? Bitmap fonts look pretty good when chars are rendered at the scale of 1 and below (and thanks to bilinear filtering we even get a poor man's AA). 

Looks like alpha channel is not used in msdf calculations in any way, so we can use it to render glyphs as if we had a bitmap font. The threshold can be controlled by material's parameter "enabling" the alpha channel to provide pixel opacity for low font sizes.

P.S. Could you please hint me which tool is supposed to be used to create JSON and PNG files for Playcanvas fonts?

willeastcott:
@vkalpias or @daredevildave know which tool is used to generate the font files. Guys?

daredevildave:
Interesting idea about storing the bitmap version in the alpha channel. My original thought was that we'd offer two types of font asset, MSDF and Bitmap. If you needed a small font size you could generate a bitmap font of the size you want. I haven't got around to trying it yet. For MSDF fonts we generate them at 32x32 px. If you think anything smaller than that would work OK scaled down then it's worth trying out.

The tool we use for generating is a modified version of [msdfgen](https://github.com/Chlumsky/msdfgen). [Our branch](https://github.com/daredevildave/msdfgen) has fixes for generating glyphs of consistent scale and then we do atlas and json generation in the backend pipeline.

yard:
Thank you for the hints. Let me try the fallback approach - I will post the results shortly and we can decide if it's worth mainlining. 

Btw I've recently encountered this article - http://wdobbie.com/post/gpu-text-rendering-with-vector-textures/. The demo looks particulary impressive; I was very surprised such a sophisticated shader actually does perform quite well in terms of performance. Would be curious to hear your thoughts on this one :)

yard:
Hi guys,

I've been playing with various tweaks to fonts to make it look better in smaller sizes.

[...]

PR #8935, opened 2026-06-19
fix(element): render MSDF text at the true glyph edge

### LINKED ISSUE #2948: Can we increase quality of font assets?
We have had a number of issues with some fonts where there are artifacts in the rendering of glyphs.

This thread for example: https://forum.playcanvas.com/t/how-to-display-cyrillic-and-glyphs-characters/18631/14

image

Is there any way we can increase the quality or have the user have some control about the resolution per glyph to allow for more complex fonts?

Some of the issues looks similar to what have been mentioned here: https://lambdacube3d.wordpress.com/2014/11/12/playing-around-with-font-rendering/

willeastcott:
## Problem

Magnified MSDF text rendered via an `ElementComponent` (text) shows a **hole in the `f` crossbar** and **notches in the `x` crossing** — MSDF median sign-errors at thin junctions, not aliasing.

Roboto Light Italic, "fox" at fontSize 256–480 (≈ 4× the 64px atlas glyphs):

| Before | After |
|:---:|:---:|
|  |  |

## Cause

`applyMsdf` remapped the field with `map(0.05, …)`, placing the 50%-coverage threshold at `sigDist ≈ 0.525` instead of `0.5` — a fixed inward **erosion of every glyph**. Where the MSDF median plateaus just above 0.5 (thin crossbars / crossings) the erosion cuts through, producing the hole and notches. This is the `0.05` threshold @mvaligursky flagged in #2948.

Separately, the AA width came from the **horizontal uv derivative only** (`fwidth(vUv0).x`), so diagonal/italic edges were under-anti-aliased.

## Fix

- Threshold at the true glyph edge: `edge = 0.5 - 0.5 * font_sdfIntensity` — intensity `0` renders at the real edge; `font_sdfIntensity` still fattens with the same slope; this also removes a latent divide-by-zero `map()` had at high intensity.
- Anti-alias from the **gradient of the distance field** (`fwidth(sigDist)` / `abs(dpdx)+abs(dpdy)`), correct for edges of any orientation.
- Remove the now-unused `map()`, `font_pxrange` / `font_textureWidth` uniforms and the `_getPxRange()` helper.

GLSL and WGSL chunks kept in parity.

## Compatibility

Removing the erosion means text now renders at its true (designer-intended) weight, which is marginally **heavier** than before. The shift is uniform and equals the erosion that was applied (≈ 0.2 atlas texels per edge). It is most visible on very thin fonts stored at low atlas resolution — the Roboto Light Italic above is near a worst case (≈ +25% stroke width, because its strokes are only ≈ 2 texels wide); most fonts will change far less. Deliberate weight adjustments should use the per-font `font_sdfIntensity` (or `outlineThickness`), not a global threshold bias.

Because the fix removes the `font_pxrange` and `font_textureWidth` uniforms from the `msdfPS` chunk, `msdfPS` is registered in the chunk-version table (API `2.20`) so a custom override of that chunk is flagged by the chunk validator.

## Testing

- **WebGL2** (repro project) and **WebGPU** (examples → User Interface ▸ Text): clean rendering; outline, drop shadow, markup colours and small text all verified.
- `npm test` passes (incl. the `text-element` suite); lint clean.

Fixes #2948

Related to #830 (small-size MSDF readability): this removes the erosion and improves AA, but that issue's atlas-side asks — uniform glyph scale, sub-pixel quad placement, and a bitmap/MTSDF fallback — are out of scope here.


willeastcott:
Just to prove the point, here is the Roboto Light Italic font rendered by Windows on the left and the same font rendered in current PlayCanvas Engine on the right:

image

Again, Windows on left and PlayCanvas Engine with fixes applied on the right:

image

So yes, fonts will appear heavier than before but they will be correct and match the source font asset.

My expectation is that we might have some developers noticing but they can probably just set their font intensity back to 0 and their text should match the source font _and_ no longer exhibit noticable artifacts. I propose we merge and see how developers respond. This fix is simply too good to pass up!

mvaligursky (approved):
all good, the versioning is required as well, as old chunk will no longer work (missing uniforms).

mvaligursky:
Docs follow-up: added an **Engine v2.20** entry to the shader chunk migrations page (covering the `msdfPS` rework and the removed `font_pxrange` / `font_textureWidth` uniforms + `map()` helper) in both the English and Japanese versions — playcanvas/developer-site#1067.

PR #8945, opened 2026-06-22
fix(particles): upload CPU vertex buffer on rebuild/reset so paused emitters don't render garbage

### LINKED ISSUE #5993: Particle System Issue - Texture stretched to fullscreen
Hey hey

We have encountered some weird behaviour in the particle systems we usually set up. When playing a particle system, which uses a texture for its particles, in the first frame of playing it, the texture is scaled up to span across the whole screen. It looks like a short flickering, after which the particle system behaves normal. It also only happens the first time the particle system is played. If we stop and reset it, the next time we play it it’s fine.

To reproduce this issue we had to:
- Set up a particle system component
- Set the autoplay flag to true
- Set the sort order to anything other than NONE
- Add a texture asset as the color map
- Set the preload flag of the texture asset to false
- Set a timeout to 0 seconds, when the particle system is paused

You will see that the particles are huge. I set up a reproduction [project](https://playcanvas.com/editor/scene/1940319) and posted about it in the [forum](https://forum.playcanvas.com/t/particle-system-issue-first-frame-fullscreen/34649).

willeastcott:
## Description

CPU particle emitters (`sort != PARTICLESORT_NONE`) fill their vertex buffer in `addTime()` but defer the GPU upload to `ParticleEmitter.finishFrame()`, which `ParticleSystemComponentSystem.onUpdate` only calls while `!component._paused`.

When a colorMap texture with `preload: false` finishes loading **after** the system has been paused, the resulting `colorMap` setter → `_setComplexProperty` → `rebuild()` + `reset()` leaves the freshly-allocated vertex buffer unflushed. The GPU keeps rendering the stride-4 quad template uploaded by `_allocate()`, reinterpreted as the 15-float CPU vertex format → garbage scale/positions → **billboards stretched fullscreen** on the first frame.

Fix: flush the buffer with `finishFrame()` after the initial `addTime(0)` fill in `rebuild()` and `reset()`. It's a no-op for GPU particles and safe to call without a matching `lock()` (the non-paused update loop already does so every frame).

Verified in-browser against the issue's repro (autoPlay + `sort:2` + non-preloaded colorMap + `pause()` on `setTimeout(0)`): the fullscreen stretch is gone and the playing path is unaffected. Existing particle-system tests pass. (The headless test harness uses `NullGraphicsDevice`, which disables particle systems, so the emitter vertex-buffer path can't be unit-tested there.)

Fixes #5993

## Checklist


mvaligursky (approved):

mvaligursky (commented):

---

**Verdict: correct and minimal. No blocking issues.**

I traced the lock/upload lifecycle to confirm the fix is safe:
- In CPU mode, `addTime()` (`particle-emitter.js:1082`) calls `this.vertexBuffer.lock()` and deliberately leaves it unlocked (line 1084 is commented out) — the GPU upload is deferred to `finishFrame()`.
- `finishFrame()` (line 1013) only unlocks when `useCpu`, so it's a genuine no-op for GPU particles. ✓
- `VertexBuffer.unlock()` (`vertex-buffer.js:147`) is stateless — `lock()` just returns `storage`, there's no lock flag — so calling it after the `addTime(0)` fill simply uploads the current contents. Safe with no matching explicit `lock()`, exactly as the PR notes. ✓

So adding `finishFrame()` after the initial fill in both `rebuild()` and `reset()` correctly flushes the freshly-allocated CPU buffer instead of leaving the stale stride-4 quad template on the GPU for a paused emitter.

Other angles:
- **Breaking changes:** none.
- **Performance:** in the non-paused path this causes one redundant upload per `rebuild()`/`reset()` (the frame's regular `finishFrame()` uploads again). These are infrequent (asset load / property change), so it's negligible — worth being aware of but not worth guarding.
- **Tests:** understandable that the `NullGraphicsDevice` headless harness can't exercise this; in-browser verification against the #5993 repro is the right call. A line in CHANGELOG/PR noting the manual-verification-only status is already present.

PR #8946, opened 2026-06-22
fix(particles): support non-interleaved meshes as particle meshes

### LINKED ISSUE #5637: Cannot use mesh from procedural in particle system
The particle system cannot use the mesh generated by the procedural methods. For example:

```js
this.entity.particlesystem.mesh = pc.createSphere(this.app.graphicsDevice);
```

Repro:
https://playcanvas.com/project/1139607/overview/particle-system-issue

image

willeastcott:
## Description
Particle systems can now use non-interleaved meshes (e.g. `pc.createSphere(device)`) as the particle mesh.

`ParticleEmitter._allocate` copied the source mesh assuming an interleaved vertex layout (position in the first 3 floats of each vertex, one global `totalFloats / numVertices` stride). Meshes from the procedural generators are non-interleaved (`Mesh.fromGeometry` builds the format with a `vertexCount`, giving a planar `PPP…NNN…UUU…` layout), so the copy read scrambled and out-of-bounds data — only vertex 0 landed correctly, and UVs ran off the end into NaN.

It now reads positions and UVs using each element's own `offset` and `stride`, which is correct for both interleaved and non-interleaved meshes. UVs are zero-filled when the source mesh has none (previously NaN).

Fixes #5637

## Checklist


mvaligursky (approved):
Nice!

mvaligursky (commented):

---

**Verdict: correct fix, addresses the root cause cleanly. No blocking issues.**

I verified the indexing math against `VertexFormat` (`src/platform/graphics/vertex-format.js:181-182`):
- **interleaved** → `offset` = element offset within a vertex, `stride` = full vertex size;
- **non-interleaved** (`vertexCount` specified, e.g. `Mesh.fromGeometry` / `createSphere`) → `offset` = start of that element's planar block, `stride` = the element's own size.

So `meshData[vert * stride + offset]` (in floats) is correct for **both** layouts, which is exactly the bug — the old single global `totalFloats/numVertices` stride only worked for interleaved meshes. Reading position and UV via their own element offset/stride is the right generalization, and zero-filling missing UVs (vs. the previous NaN run-off) is a strict improvement.

Other angles:
- **Breaking changes:** none — interleaved meshes resolve to identical indices.
- **Performance:** `.find()` runs twice per `_allocate`, not per vertex. Negligible.
- **Cleanliness:** clearer than the prior loop; good explanatory comment.

One minor latent assumption noted inline (non-blocking).

src/scene/particle-system/particle-emitter.js
```diff
@@ -904,16 +904,20 @@ class ParticleEmitter {
...
-                    }
-                }
+
+                // Copy positions and UVs using each element's own offset and stride (in floats),
+                // which handles both interleaved and non-interleaved source meshes. Procedurally
+                // generated meshes (e.g. createSphere) are non-interleaved.
+                const elements = this.mesh.vertexBuffer.format.elements;
+                const posElement = elements.find(el => el.name === SEMANTIC_POSITION);
+                const uvElement = elements.find(el => el.name === SEMANTIC_TEXCOORD0);
+                posOffset = posElement.offset / 4;
```
mvaligursky, line 917: Minor / non-blocking: `posElement.offset / 4` and `/ 4` throughout assume (a) `SEMANTIC_POSITION` is always present and (b) both position and UV are 32-bit floats.

(a) — if a mesh has no position element, `posElement` is `undefined` and this throws. A particle mesh without positions is degenerate, and the old code also assumed positions existed (at offset 0), so this isn't a regression — but unlike the UV path it's unguarded. A one-line guard or assert would make the failure mode explicit.

(b) — the `/ 4` (bytes→floats) is correct for the procedural generators this PR targets, but would mis-index a mesh with non-float positions/UVs (e.g. normalized int16 from a compressed/quantized source). Also a pre-existing limitation, not introduced here — just worth a comment so it isn't silently assumed universal.

PR #8947, opened 2026-06-23
fix(anim): support cloning components with dynamically-added layers

### LINKED ISSUE #6395: Undefined access when anim component with multiple layers is cloned
Application crashes, when trying to clone an entity with an Anim Component that has more than 1 animation layer.

Repro:
Open engine example: https://playcanvas.github.io/#/animation/layer-masks
And add to the end:
```
modelEntity.clone();
```

image

image

willeastcott:
### Fixes #6395

Cloning an entity whose `anim` component had a layer added at runtime via `addLayer()` (rather than via a loaded state graph) threw `TypeError: Cannot read properties of undefined (reading 'assignAnimation')`. This reproduces by opening the [`animation/layer-masks`](https://playcanvas.github.io/#/animation/layer-masks) example and appending `modelEntity.clone()`.

**Cause:** layers (and states) added dynamically via `addLayer()` / `assignAnimation()` are not recorded in the component's `_stateGraph`. On clone, `initializeComponentData` rebuilds the clone's layers from that stale state graph, then iterates the source's *live* layers doing `component.layers[i].assignAnimation(...)` — for any layer missing from the state graph, `component.layers[i]` is `undefined`.

**Fix:** resolve the clone's layer by name (creating it via `addLayer` if the state graph didn't recreate it) and assign onto that layer instead of the index. No behavior change for the state-graph/Editor-driven path, where all layers already exist.

Verified with a new regression test in `test/framework/components/anim/component.test.mjs` that mirrors the issue's repro (base layer + a dynamically-added additive second layer). It throws the documented error before the fix and passes after.


mvaligursky (commented):

---

**Verdict: looks correct and well-scoped. No blocking issues.**

I verified the core assumption: `AnimComponent.addLayer()` (`src/framework/components/anim/component.js:383`) is idempotent — it calls `findAnimationLayer(name)` and returns the existing layer if present, only creating one otherwise. So resolving the clone's layer by name handles both paths cleanly:
- **state-graph / Editor-driven layers** — already recreated by `loadStateGraph()`, so `addLayer()` returns the existing instance (the `weight/mask/blendType` args are ignored, which is fine since the existing layer already carries the graph's values).
- **dynamically-added layers** — absent from `_stateGraph`, so `addLayer()` creates them with the source layer's `weight/mask/blendType`, which is exactly what the old index-based code couldn't do.

Other angles checked:
- **Breaking changes:** none. Behavior is identical for the existing all-layers-present path.
- **Layer ordering / index integrity:** preserved — iteration is in `data.layers` order and `_layerIndices` stays consistent.
- **Performance:** negligible (one extra name lookup per layer at clone time).
- **Test:** the regression test mirrors the issue repro and also adds a single-layer guard. Good coverage of the additive blend-type round-trip and independent-instance check.

Minor (optional, non-blocking): the `weight`/`mask`/`blendType` passed to `addLayer()` are silently dropped for layers that already exist. That's correct today, but a short comment noting "args only apply when the layer is newly created here" could save a future reader a double-take.

PR #8957, opened 2026-06-24
Remove unused Camera.beforePasses

mvaligursky:
Removes the `Camera.beforePasses` extension point, which is no longer used by anything in the engine.

It was added in #8531 for the compute-based tiled GSplat renderer (`GSplatComputeGlobalRenderer`) to schedule compute work before a camera's main render pass. That renderer was removed in #8558, which deleted the only writer but left the field and its reader behind. This finishes that cleanup.

**Changes:**
- Remove the `beforePasses` field from `Camera`.
- Remove `RenderPassForward.updateCameraBeforePasses()` and its call in `frameUpdate()`.

**API Changes:**
- Removed `Camera.beforePasses` (`FramePass[]`). It was always empty in normal usage — the engine never populated it after #8558.

### DIFF
--- a/src/scene/camera.js
+++ b/src/scene/camera.js
@@ -117,14 +117,6 @@ class Camera {
      */
     framePasses = [];
 
-    /**
-     * Frame passes that execute before this camera's main scene rendering. Entries are picked up
-     * by the RenderPassForward that renders this camera's layers.
-     *
-     * @type {FramePass[]}
-     */
-    beforePasses = [];
-
     /** @type {number} */
     jitter = 0;
 
--- a/src/scene/renderer/render-pass-forward.js
+++ b/src/scene/renderer/render-pass-forward.js
@@ -168,26 +168,6 @@ class RenderPassForward extends RenderPass {
         return index;
     }
 
-    // Collect before-passes from cameras whose first render action lives in this
-    // RenderPassForward. Uses the existing firstCameraUse flag (set by LayerComposition)
-    // to guarantee each camera's before-passes are scheduled exactly once, even when
-    // multiple RenderPassForward instances reference the same camera (e.g. CameraFrame's
-    // scenePass vs afterPass).
-    updateCameraBeforePasses() {
-        for (let i = 0; i < this.renderActions.length; i++) {
-            const ra = this.renderActions[i];
-            if (ra.firstCameraUse) {
-                const camera = ra.camera?.camera;
-                if (camera) {
-                    const { beforePasses } = camera;
-                    for (let j = 0; j < beforePasses.length; j++) {
-                        this.beforePasses.push(beforePasses[j]);
-                    }
-                }
-            }
-        }
-    }
-
     updateDirectionalShadows() {
         // add directional shadow passes if needed for the cameras used in this render pass
         const { renderer, renderActions } = this;
@@ -237,7 +217,6 @@ class RenderPassForward extends RenderPass {
 
     frameUpdate() {
         super.frameUpdate();
-        this.updateCameraBeforePasses();
         this.updateDirectionalShadows();
         this.updateClears();
     }


MAG-AdrianMeredith:
Woh there!  Is there a good reason to remove this? I'm using it.  Whats the alternative? @mvaligursky

mvaligursky:
No alternative. I'm looking at larger under the hood refactoring, and so cleaned it as it was at the moment blocking the refactoring. Not sure about alternative, I guess it depends on what it is used. It was never public, just an internal thing I added. I liked it, but it's in a way.

Share some details of what you use it for please.

MAG-AdrianMeredith:
Put simply its the only way to answer the question of "this render pass must happen exactly before this one". This is used extensively for my gpu driven rendering rnd.

I'll do a quick ai summary if that helps

MAG-AdrianMeredith:
Why we need camera.beforePasses
beforePasses is the only injection point that lets us interleave custom render passes at exact positions inside the engine's per-camera frame graph — between scenePass, composePass, TAA, and the directional-shadow pass. The GPU-driven meshlet path (?cleanCull, default ON) depends on three orderings that nothing else can express:

1. Meshlet raster must land in the opaque slot (right after scenePass)

Composites with non-meshlet opaque geometry, shares depth.
Visible to colorGrabPass (refraction) + the transparent pass.
Critically runs before taaPass/composePass → temporal accumulation + tonemap include the meshlets.
afterPasses is wrong: it runs after compose = after TAA/tonemap, so compose would source a meshlet-less scene.
2. HZB occlusion chain must immediately follow the meshlet raster

cleanDepthGrab → cleanHzbFeed → cleanCompleteHZB grab post-meshlet scene-RT depth and build this frame's complete HZB = next-frame pass-1 occlusion reference. Needs the slot directly after the raster, before colorGrab/compose.
3. Meshlet directional-shadow pass must sit AFTER engine dir-shadow, BEFORE scenePass

Engine dir-shadow pass clears the shadow map + draws standard casters; scenePass samples it.
We draw meshlet casters into the freshly-cleared map between those two. The frame graph flattens a pass's beforePasses just before that pass — so ordering within beforePasses is the mechanism that places our caster draw correctly (we even hook scenePass.frameUpdate to append after the engine's own dir-shadow push).
Core point
beforePasses gives us deterministic positional insertion relative to engine built-in passes. The two alternatives both fail:

afterPasses → everything lands after compose (post-TAA/tonemap) → wrong for color compositing, refraction, and shadow-map ordering.
engine main-forward path → engine owns the draw; we can't do GPU-driven indirect raster + two-pass HZB occlusion + meshlet shadow injection.
Remove beforePasses and there is no API surface to put a custom pass between two specific engine passes. The meshlet raster, the occlusion HZB chain, and the meshlet shadow caster all break.

mvaligursky:
Thanks for the detail! I'll definitely add this back or equivalent when I'm done with current refactors, before the next engine version ships.

MAG-AdrianMeredith:
Thanks appreciate it, worst case I can do something nasty in our fork but I'm trying to reduce custom changes and move closer back to upstream

mvaligursky:
@MAG-AdrianMeredith `beforePasses` is back in #9073, same shape as before (`camera.beforePasses` as a plain `FramePass[]`), ported to the refactored internals.

One deliberate change worth checking against your fork: the before-passes are now collected **after** the engine pushes its directional shadow passes, so the execution order is `engine dir-shadow → camera.beforePasses → scene`. That means your meshlet shadow-caster pass can now just be pushed into `camera.beforePasses` in the order you need — no more hooking `scenePass.frameUpdate` to append after the engine's dir-shadow push. If you keep that hook alongside the new collection, watch out for the pass getting scheduled twice.

This PR intentionally covers only the "before scene" injection points (your shadow-caster and pre-scene cases). The mid-frame positions — after the opaque scene pass, before color-grab / TAA / compose (your meshlet raster + HZB chain) — aren't addressed here yet; I'm considering follow-up options for those, likely along the lines of making `FramePassCameraFrame` officially subclassable (`CameraFrame.createRenderPass()` is already an override hook). Would be useful to hear if that shape would cover the rest of your setup.

Issue #8958, opened 2026-06-24
Improve engine API ergonomics for generated PlayCanvas apps


kpal81xd:
## Summary

Generated PlayCanvas apps repeatedly trip over public APIs that look like plain mutable data, but actually have engine ownership, lifecycle, readiness, hierarchy, or setter/helper side effects.

This issue tracks engine/API-layout fixes only. Do not solve this in the app template.

## Current Status

| Area | PR | Status | Notes |
| --- | --- | --- | --- |
| Readonly typings/JSDoc for live engine-owned getters | #8959 | Merged | First pass for arrays, maps, transform values, physics/UI values, material state, and related setters. |
| Entity render traversal and bounds helpers | #8960 | Draft | Adds `Entity#forEachRenderMeshInstance(cb)` and `Entity#getAabb(result?)`. |

## Workstreams

| Priority | Theme | Scope | Status |
| --- | --- | --- | --- |
| P0 | Render/model mesh ownership | `meshInstances` readonly getter types, engine-owned docs, explicit replacement path, debug errors for current/stale/destroyed instances. | Partly covered by #8959; replacement/debug work remains. |
| P0 | Entity render helpers | Subtree mesh traversal and subtree render bounds for common GLB edits. | Draft in #8960. |
| P0 | Setter-only object values | Transform getters, script order, and other values where component-wise mutation bypasses engine updates. | First typings/JSDoc pass merged in #8959; more docs may be needed. |
| P0 | Asset/container readiness | Failed-load semantics, `loadFromUrlAndFilename` retry errors, and container render-resource readiness. | Todo. |
| P0 | Mesh lifecycle clarity | `Mesh#clear` dynamic flags and mesh format changes after `MeshInstance` construction. | Todo. |
| P1 | Collections and hierarchy | `children`, component `layers`, layer mesh lists, current-entity component lookup, tag search behavior. | Partly covered by #8959; helper/docs work remains. |
| P1 | Live component values | Physics vectors, UI/render vectors/colors/rects, material/blend state, particle graphs, sound slots, anim layers/masks. | Partly covered by #8959; helper/docs work remains. |
| P1 | Loading and cloning semantics | Container/GLB APIs, generated sub-resource lifetime, clone/shared-resource rules. | Todo. |
| P1 | Input ergonomics | Nullable input devices and common keyboard code strings. | Todo. |
[...]

PR #8959, opened 2026-06-24
fix: mark live API getters readonly

kpal81xd:
Refs #8958

## Description
- Marks live engine-owned getter surfaces as readonly in generated typings, including mesh instances, layer arrays, graph children, transform values, physics values, UI vectors, sound slots, animation layers, material blend state, and skybox rotation.
- Updates JSDoc examples around transform getters to clone before mutating.

## Manual smoke test
1. In a TypeScript PlayCanvas app using this branch's generated declarations, type `entity.getPosition().x = 1` or `entity.render.meshInstances.push(meshInstance)`.
   Expected: the editor reports a readonly/type error for the direct mutation.
2. Replace the transform edit with `const position = entity.getPosition().clone(); position.x = 1; entity.setPosition(position);`.
   Expected: the editor accepts the setter-based update.

## Checklist


mvaligursky (commented):
🤖 **Automated PR review** — posted on my behalf by Claude Code (Opus 4.8). Not a human review. The points below are suggestions to weigh as possible improvements, not changes that necessarily need to be addressed. (Reviewed retrospectively — this PR is already merged, so these are follow-up considerations.)

---

Doc/typings-only, no runtime change, and a genuine improvement. I confirmed the `anchor`/`pivot` getters narrowing to `Readonly<Vec4>`/`Readonly<Vec2>` is safe — their setters carry their own `@type {... | number[]}` JSDoc, so passing `number[]` to the setter still type-checks. Two things worth being aware of:

1. **`Readonly<T>` only blocks property assignment, not mutating method calls.** It stops `entity.getPosition().x = 1` (the documented footgun), but `entity.getPosition().add(v)` / `.set(...)` / `.copy(...)` / `.mulScalar(...)` still type-check and will mutate the internal vector with no error. Same for `Readonly<Quat>`/`Readonly<Mat4>`/`Readonly<Color>`/`Readonly<BlendState>`. So the protection is partial — valuable for the most common mistake, but not a full guard. Worth a note in the issue if method-based mutation is a concern (a branded/opaque type would be needed to fully prevent it).

2. **This is a breaking change for TypeScript consumers.** `Readonly<T>` / `ReadonlyArray<T>` are not assignable back to `T` / `T[]`, so existing downstream code that stores or forwards these return values into mutable-typed variables or parameters (e.g. `const p: Vec3 = entity.getPosition()` or `fn(entity.render.meshInstances)` where `fn(x: MeshInstance[])`) will newly fail to compile — even when it only reads. That's the intended behavior for catching mutation, but it's worth calling out explicitly in the release notes so users aren't surprised by new type errors on upgrade.

src/framework/components/camera/component.js
```diff
@@ -740,7 +740,7 @@ class CameraComponent extends Component {
     /**
      * Gets the array of layer IDs ({@link Layer#id}) to which this camera belongs.
      *
-     * @type {number[]}
+     * @type {ReadonlyArray<number>}
```
willeastcott, line 743: What's the difference between using that and `@readonly`? We already use `@readonly` across the codebase.

mvaligursky: Different levels — one is about the reference, the other about the value.

@readonly marks the property/binding as non-reassignable. It's the JSDoc equivalent of TS's readonly modifier on a member:

/** @readonly */
this.position = new Vec3();
// entity.position = otherVec;  ❌ can't rebind the slot
// entity.position.x = 1;       ✅ still allowed — the Vec3's contents are mutable
It says nothing about whether the object it points at can be mutated.

Readonly<T> (in @type) changes the value's type so all of T's properties become readonly:

/** @type {Readonly<Vec3>} */
get position() { ... }
// pos.x = 1;  ❌ field is readonly
// (but pos.add(v), pos.set(...) ✅ still allowed — see below)
It says nothing about whether you can reassign the slot.

For this PR the goal was to stop entity.getPosition().x = 1 — i.e. mutating the contents of the returned object — so Readonly<T> is the correct tool. @readonly wouldn't help here: these are getters with no setter, so the slot is already non-reassignable, and @readonly wouldn't touch .x.

Two extra nuances worth keeping in mind:

They're orthogonal, not alternatives — you could have both (a readonly field whose value is also Readonly<T>). They guard different things.
ReadonlyArray<T> is stronger than Readonly<Vec3>. TS defines ReadonlyArray specially, so it actually removes the mutating methods (push/pop/splice) from the type. But Readonly<Vec3> is just the generic mapped type — it makes .x/.y/.z readonly but leaves Vec3's mutating methods (.set, .add, .copy) fully callable. That's why the array surfaces in this PR are well-protected, but the Vec/Quat/Mat/Color ones are only partially protected (the point I raised in the review). Fully locking those down would need a branded/opaque type, not Readonly<> or @readonly.

PR #8960, opened 2026-06-24
feat: add entity render mesh helpers

kpal81xd:
Refs #8958

## Description
- Adds `Entity#forEachRenderMeshInstance(cb)` to traverse render/model mesh instances across an entity subtree.
- Adds `Entity#getAabb(result?)` to compute subtree render bounds without reading live `meshInstances` arrays.
- Covers runtime behavior with entity tests.

## Manual smoke test
1. Create an entity hierarchy with render or model components on parent and child entities.
2. Call `entity.forEachRenderMeshInstance(cb)` and confirm `cb` receives each mesh instance with its owning component and entity.
3. Call `entity.getAabb()` and confirm it returns a bounding box covering the rendered subtree, or `null` for an empty hierarchy.

## Checklist
- [x] My code follows the project coding standards

### DIFF
--- a/src/framework/entity.js
+++ b/src/framework/entity.js
@@ -1,5 +1,6 @@
 import { Debug } from '../core/debug.js';
 import { guid } from '../core/guid.js';
+import { BoundingBox } from '../core/shape/bounding-box.js';
 import { GraphNode } from '../scene/graph-node.js';
 import { getApplication } from './globals.js';
 
@@ -28,6 +29,14 @@ import { getApplication } from './globals.js';
  * @import { ScrollbarComponent } from './components/scrollbar/component.js'
  * @import { SoundComponent } from './components/sound/component.js'
  * @import { SpriteComponent } from './components/sprite/component.js'
+ * @import { MeshInstance } from '../scene/mesh-instance.js'
+ */
+
+/**
+ * @callback RenderMeshInstanceCallback
+ * @param {MeshInstance} meshInstance - The render mesh instance.
+ * @param {RenderComponent|ModelComponent} component - The component that owns the mesh instance.
+ * @param {Entity} entity - The entity that owns the component.
  */
 
 /**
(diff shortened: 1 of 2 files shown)


src/framework/entity.js
```diff
@@ -466,6 +475,63 @@ class Entity extends GraphNode {
...
+     *
+     * @param {RenderMeshInstanceCallback} callback - The function to execute on each mesh
+     * instance.
+     * @param {object} [thisArg] - Optional value to use as this when executing callback function.
+     * @example
+     * entity.forEachRenderMeshInstance((meshInstance) => {
+     *     meshInstance.visible = false;
+     * });
+     */
+    forEachRenderMeshInstance(callback, thisArg) {
```
mvaligursky, line 490: why is it called `forEachRenderMeshInstance` when it works on model components.
it should be `forEachMeshInstance`

src/framework/entity.js
```diff
@@ -466,6 +475,63 @@ class Entity extends GraphNode {
...
+     *
+     * @param {BoundingBox} [result] - The bounding box to receive the result.
+     * @returns {BoundingBox|null} The bounding box, or null if no render mesh instances were found.
+     * @example
+     * const aabb = entity.getAabb();
+     * if (aabb) {
+     *     // use aabb
+     * }
+     */
+    getAabb(result) {
```
mvaligursky, line 518: not sure this function is good as is, I think meshInstance.aabb is only updated during rendering, so on the first frame it's not valid. we should not expose this for now till that is sorted. separate PR.

src/framework/entity.js
kpal81xd: This is just a draft for now - the exact names we can modify. I am just collating findings from my experiments with agents

LeXXik:
There is also related: https://github.com/playcanvas/engine/pull/7844

src/framework/entity.js
kpal81xd: Still working on this - this is not merge ready at all

PR #8961, opened 2026-06-24
Refactor light & directional-shadow culling into separate phases

mvaligursky:
Internal renderer refactor that splits directional-shadow and light-visibility culling into smaller, single-responsibility phases. This is preparation for a later change that builds the frame graph before mesh culling; there is no change to rendering behavior.

**Changes:**
- `ShadowRendererDirectional`: split shadow-map allocation out of `cull()` into a new `prepareShadowMap()` — the minimal, caster- and camera-independent step needed before a directional shadow pass can be created.
- `Renderer`: extract `collectDirectionalShadowLights()` (builds the per-camera directional shadow-light set and allocates shadow maps) from `cullShadowmaps()`, separate from the mesh-dependent caster cull.
- `Renderer`: split the cull phase into `updateLightVisibility()` (per-frame light reset, per-camera frustum update + light-visibility culling, clustered light-atlas allocation, and directional shadow-light collection — all independent of mesh culling) and a slimmed `cullComposition()` (mesh-instance + shadow-caster culling only). The per-frame light-visibility reset moves from `Renderer.beginFrame()` into `updateLightVisibility()`.

All affected methods are internal (`@ignore`). Behavior-neutral: every producer still precedes its consumer; the only observable change is that `EVENT_PRECULL` now fires after light culling (the frustum is still updated beforehand).

### DIFF
--- a/src/framework/lightmapper/lightmapper.js
+++ b/src/framework/lightmapper/lightmapper.js
@@ -849,6 +849,7 @@ class Lightmapper {
             }
 
             if (light.type === LIGHTTYPE_DIRECTIONAL) {
+                this.renderer._shadowRendererDirectional.prepareShadowMap(light);
                 this.renderer._shadowRendererDirectional.cull(light, comp, this.camera, casters);
 
                 const shadowPass = this.renderer._shadowRendererDirectional.getLightRenderPass(light, this.camera);
--- a/src/scene/renderer/forward-renderer.js
+++ b/src/scene/renderer/forward-renderer.js
@@ -1087,7 +1087,11 @@ class ForwardRenderer extends Renderer {
         // update gsplat director
         this.gsplatDirector?.update(comp);
 
-        // visibility culling of lights, meshInstances, shadows casters
+        // light visibility culling, light atlas allocation and directional shadow light collection
+        // (mesh-independent, so it can run before the frame graph is built in a later refactor)
+        this.updateLightVisibility(comp);
+
+        // visibility culling of meshInstances and shadow casters
         // after this the scene culling is done and script callbacks can be called to report which objects are visible
         this.cullComposition(comp);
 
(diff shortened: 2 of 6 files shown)


mvaligursky (commented):

---

Refactor looks correct and behavior-neutral as claimed. Verified the key invariants hold: every `_shadowRendererDirectional.cull()` caller (lightmapper + the new `cameraDirShadowLights` loop) is now preceded by `prepareShadowMap()`; `updateLightTextureAtlas()` still runs after all `cullLights` and before any shadow culling; local-light shadow culling in `cullShadowmaps` is untouched; and per-camera light culling is OR-accumulated, so hoisting it into its own loop is equivalent. Two observations, neither blocking:

1. **`EVENT_PRECULL` change is slightly broader than "fires after light culling."** Because all cameras' frustum updates + light culls now complete in `updateLightVisibility` before *any* `PRECULL` fires, a listener also sees every *other* camera already frustum-updated and culled when `PRECULL` fires for the first camera (previously cameras were processed one-at-a-time, interleaved). The practical impact is the same class of listener — one that toggles light enabled/visibility in `PRECULL` expecting to influence that frame's light culling no longer can. Worth a line in the migration/release notes if any public-facing code relies on `PRECULL` ordering.

2. **New ordering invariant for the light reset.** Moving the per-frame `light.beginFrame()` reset out of `Renderer.beginFrame()` and into `updateLightVisibility()` means it now runs *after* `setSceneConstants()` and `gsplatDirector.update()` rather than before. I confirmed neither currently reads the per-frame light state that's reset (`visibleThisFrame`, `maxScreenSize`, `atlasViewportAllocated`, `atlasSlotUpdated`), so it's benign today — but anything inserted between `beginFrame()` and `updateLightVisibility()` in future that touches that state would silently see stale (prev-frame) values. A one-line comment at the reset site noting "must run before any consumer of per-frame light state" could guard against that regressing later.

PR #8962, opened 2026-06-24
Rework shadow update-mode lifecycle; keep THISFRAME pending until rendered

mvaligursky:
Reworks how a light's `shadowUpdateMode` is handled during the frame, fixing a behavior bug and preparing for an upcoming change that builds the frame graph before mesh culling.

**Behavior fix:**
- A one-shot `SHADOWUPDATE_THISFRAME` request is now consumed only when the shadow actually renders this frame. Previously it was consumed unconditionally, so a light flagged while not rendered (for example an off-screen local light) could have its request dropped without ever updating the shadow. It now stays pending and the shadow updates the next frame the light is rendered.

**Refactor (no behavior change; prep for the frame-graph reorder):**
- `ShadowRenderer.needsShadowRendering` is now a pure predicate (no `shadowUpdateMode` flip, no stat side-effect).
- The `THISFRAME → NONE` consume and the shadow-map-update stat moved into `Renderer.consumeOneShotShadows`, applied once after the frame graph is built and shadow casters are culled.
- The `NONE → THISFRAME` re-flag (atlas slot reassigned / shadow map not yet allocated) moved into `updateLightVisibility`, ahead of the frame graph build and shadow culling.

**Behavior note:**
- A directional shadow-casting light shared by multiple cameras with `SHADOWUPDATE_THISFRAME` now renders its shadow per camera (each fit to that camera's cascade frustum) instead of only for the first camera. This is the correct result, since the directional shadow frustum is camera-dependent.

All affected methods are internal (`@ignore`).

### DIFF
--- a/src/scene/renderer/forward-renderer.js
+++ b/src/scene/renderer/forward-renderer.js
@@ -1030,6 +1030,11 @@ class ForwardRenderer extends Renderer {
                 }
             }
         }
+
+        // consume one-shot (THISFRAME) shadow updates - the frame graph is now built and shadow
+        // casters were culled earlier this frame, so both have read the shadow update mode before
+        // it is changed here
+        this.consumeOneShotShadows();
     }
 
     /**
--- a/src/scene/renderer/renderer.js
+++ b/src/scene/renderer/renderer.js
@@ -969,28 +969,12 @@ class Renderer {
      */
     cullShadowmaps(comp) {
 
-        const isClustered = this.scene.clusteredLightingEnabled;
-
-        // shadow casters culling for local (point and spot) lights
+        // shadow casters culling for local (point and spot) lights. The shadow-update-mode forcing
+        // (atlas slot reassigned / shadow map not yet allocated) was applied earlier, in
+        // updateLightVisibility, so the update mode is already final here.
         for (let i = 0; i < this.localLights.length; i++) {
             const light = this.localLights[i];
             if (light._type !== LIGHTTYPE_DIRECTIONAL) {
-
-                if (isClustered) {
-                    // if atlas slot is reassigned, make sure to update the shadow map, including the culling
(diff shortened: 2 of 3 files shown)


mvaligursky (commented):

---

Solid rework. I verified the core safety property: all three `needsShadowRendering` callers (`prepareLights`, `buildNonClusteredRenderPasses`, `getLightRenderPass`) run at frame-graph *build* time and capture the render decision into the created passes; no `render-pass-shadow-*` pass re-checks `shadowUpdateMode`/`needsShadowRendering` at execute time. So consuming the flag at the end of `buildFrameGraph` (after the passes are built, before the graph executes) is safe. The consume predicates also match the build gating exactly (clustered local `needs && atlasViewportAllocated`, non-clustered `needs`, directional `needs`), and the behavior fix (keep THISFRAME pending when not rendered) is correct. One minor issue:

**`_shadowMapUpdates` stat undercounts for a `SHADOWUPDATE_THISFRAME` directional light shared by multiple cameras.** In `consumeOneShotShadows`, the directional loop both counts the stat and consumes `THISFRAME -> NONE` in the same pass. A directional light shared by N cameras appears in `cameraDirShadowLights` N times and (per the behavior note) now builds/renders a shadow pass for each. But on the first camera's iteration the consume flips the mode to `NONE`, so `needsShadowRendering` returns `false` for the remaining N-1 cameras and their (already-built, already-rendered) passes are never added to the stat — it reports 1 face-set where N rendered. Functionally harmless (rendering is unaffected, since passes were built before the consume), but the profiler stat is now inconsistent with the new per-camera behavior, and inconsistent with the `REALTIME` directional case which correctly counts all N. Consider accumulating the stat independently of the consume (count per built pass / per camera occurrence, then consume the one-shot flag separately).

PR #8967, opened 2026-06-25
Route view uniforms through dynamic buffers; remove view bind groups

mvaligursky:
Replaces the per-render-action view bind group storage and the dedicated view bind group format with the engine's existing dynamic uniform buffer mechanism, and lets the standard forward layer path own all view-uniform setup. This is a step toward making `RenderPassForward` a generic, layer/camera-agnostic pass.

**View uniforms:**
- Remove `viewBindGroups` storage from `RenderAction`, `LightRenderData`, and the picker / prepass / lightmapper passes.
- The view uniform buffer is now a shared non-persistent buffer (per uniform format, cached on the renderer) bound via the dynamic buffer system — no per-render-action bind groups and no manual lifetime / pooling management (the dynamic buffer system owns recycling).
- `setupViewUniformBuffers(viewUniformFormat, viewList)` is the single entry point; multiview captures the per-view bind group + dynamic offset.

**View bind group format:**
- Drop `viewBindGroupFormat` entirely. The view bind group has no textures, so its single-UB layout comes from the actually-bound dynamic buffer; shaders only need the view *uniform* format. `ShaderProcessorOptions` no longer takes a bind group format, and it is removed from the renderers, `MeshInstance.getShaderInstance`, and the material `getShaderVariant` path. `initViewBindGroupFormat` is renamed `initViewUniformFormat`.

**Forward layer path:**
- `renderForwardLayer` ensures the view uniform format lazily (so the picker and lightmapper no longer pre-seed it) and accepts optional `drawCallback` and `viewUniformFormat` arguments. A custom view format is threaded into both the view buffer setup and shader generation so they always match.
- The picker now renders through `renderForwardLayer` (instead of the low-level `renderForward`) with its own minimal view format, dropping the `dispatchGlobalLights` / `shadowAtlasParams` workaround.

**Cleanup:**
- Remove `LayerComposition#destroyRenderActions` / `destroy` (render actions no longer hold GPU resources) and inline the reset into `_update`.

**API:**
- Internal renderer refactor; no public API change. `RenderPassForward` is unchanged; `renderForwardLayer` only gains optional arguments.

Verified across ~30 graphics examples plus picker (color + depth pick) on WebGL2 and WebGPU.

### DIFF
--- a/src/extras/render-passes/render-pass-prepass.js
+++ b/src/extras/render-passes/render-pass-prepass.js
@@ -31,9 +31,6 @@ const DEPTH_UNIFORM_NAME = 'uSceneDepthMap';
  * @ignore
  */
 class RenderPassPrepass extends RenderPass {
-    /** @type {BindGroup[]} */
-    viewBindGroups = [];
-
     /** @type {Texture} */
     linearDepthTexture;
 
@@ -56,12 +53,6 @@ class RenderPassPrepass extends RenderPass {
         this.renderTarget = null;
         this.linearDepthTexture?.destroy();
         this.linearDepthTexture = null;
-
-        this.viewBindGroups.forEach((bg) => {
-            bg.defaultUniformBuffer.destroy();
-            bg.destroy();
-        });
-        this.viewBindGroups.length = 0;
     }
 
     setupRenderTarget(options) {
@@ -126,7 +117,7 @@ class RenderPassPrepass extends RenderPass {
                         }
                     }
 
-                    renderer.renderForwardLayer(camera, renderTarget, null, undefined, SHADER_PREPASS, this.viewBindGroups, {
+                    renderer.renderForwardLayer(camera, renderTarget, null, undefined, SHADER_PREPASS, {
                         meshInstances: tempMeshInstances
                     });
 
--- a/src/framework/app-base.js
+++ b/src/framework/app-base.js
@@ -1892,11 +1892,6 @@ class AppBase extends EventHandler {
 
         this.systems.destroy();
 
-        // layer composition
-        if (this.scene.layers) {
-            this.scene.layers.destroy();
-        }
-
         // destroy bundle registry
         this.bundles.destroy();
         this.bundles = null;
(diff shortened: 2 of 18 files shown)


mvaligursky (commented):

---

Reviewed the whole diff; the refactor is sound and the central claim holds — the WGSL/GLSL processors build the view UB binding from `uniformFormats[BINDGROUP_VIEW]` alone (the view bind group never had textures), so dropping `viewBindGroupFormat` is safe, and the rename/removal is clean (no dangling `viewBindGroupFormat`/`initViewBindGroupFormat`/`.viewBindGroups`/`destroyRenderActions` references, no orphaned `destroy()` callers). Single- vs multi-view binding is consistent (`_viewBindGroups[v]` is only read under `if (viewList)`; single-view binds directly in `setupViewUniformBuffers`). A couple of things worth a look:

1. **The multiview (XR) path is the highest-risk area and likely not covered by the example suite.** It's the part the refactor changed most: per-view bind-group + dynamic-offset are now captured into shared renderer-instance arrays (`_viewBindGroups` / `_viewBindGroupOffsets`) in `setupViewUniformBuffers` and consumed later in `renderForwardInternal`. I traced it and believe it's correct (populate→consume happens within one `renderForwardLayer` with no intervening `setupViewUniformBuffers` to clobber the arrays, and offsets stay valid since the dynamic buffer isn't recycled mid-frame) — but "~30 graphics examples + picker" almost certainly doesn't exercise WebXR multiview. Recommend an explicit XR multiview pass on both backends before merge, since a regression there wouldn't show up in the verified set.

2. **Picker minimal view format + custom-material pick shaders.** The picker's format only carries `matrix_viewProjection` / `matrix_view`; the comment notes any other view uniform "falls back to the per-mesh uniform buffer." That's correct for the built-in pick shaders, but a *custom* material whose pick/depth-pick pass references another view-scope uniform (e.g. a camera world-position uniform) would now route it through the mesh UB. It should still resolve via scope (set by `setCameraUniforms`), so functionally fine — just worth a sanity check if any custom pick materials exist in the wild, since the old full-forward format would have placed those in the view UB.

Minor: the shared per-format view UBs in `_viewUniformBuffers` are never explicitly destroyed, but since they're non-persistent (CPU-side staging only, GPU storage owned by the dynamic pool) there's no real leak — noting only for completeness.

PR #8969, opened 2026-06-26
Free font textures when a font asset is unloaded

### LINKED ISSUE #7033: When a font asset is unloaded the font’s textures are not destroyed and still consume VRAM
When a font asset is unloaded the font’s textures are not destroyed and still consume VRAM.
Is there a reason for that or is it a bug?

The font class does not have a destroy method so asset.unload cannot call it.

I attempted a work around  in the test project https://playcanvas.com/project/1264902/overview/unloading-a-font-asset.  
In this project the user can press the space bar to toggle between a large VRAM font and a small VRAM font.
To show the bug, comment out the workaround in this function.

```
Test.prototype.unloadFontAsset = function(fontAsset)  {
    // When unloading a font asset the font's texture(s) are not destroyed.
    // which is a waste of VRAM.
    let textures = fontAsset.resource.textures;
    for (let i = 0; i < textures.length; i++) {
        let texture = textures[i];

        // Destroy to free resources associated with this texture.
        // see https://api.playcanvas.com/classes/Engine.Texture.html#destroy
        texture.destroy();

        // Clear the font texture from the resource loader cache.
        // see https://api.playcanvas.com/classes/Engine.ResourceLoader.html#clearCache
        this.app.loader.clearCache(texture.name,'texture');
    }
    fontAsset.unload();
};
```

mvaligursky:
Unloading a font asset (`asset.unload()`) did not release the GPU memory used by its textures. The `Font` resource had no `destroy()` method, so the asset system's generic `resource.destroy?.()` cleanup was a no-op and the textures stayed resident in VRAM. The textures were also left in the resource loader cache, so a subsequent load could return a stale entry.

Fixes #7033.

**Changes:**
- Add `Font.destroy()`, which destroys the font's textures and removes them from the resource loader cache.
- `FontHandler` now gives the created `Font` a reference to the resource loader so it can clear the cached texture entries on unload.

**API Changes:**
- Added `Font#destroy()`. It is called automatically when the owning font asset is unloaded; existing user workarounds that manually destroy the font's textures remain safe (double-destroy is a no-op).

### DIFF
--- a/src/framework/font/font.js
+++ b/src/framework/font/font.js
@@ -1,6 +1,7 @@
 import { FONT_MSDF } from './constants.js';
 
 /**
+ * @import { ResourceLoader } from '../handlers/loader.js'
  * @import { Texture } from '../../platform/graphics/texture.js'
  */
 
@@ -33,9 +34,42 @@ class Font {
          */
         this.intensity = 0.0;
 
+        /**
+         * The resource loader used to load the font textures. Set by the {@link FontHandler} so
+         * that {@link Font#destroy} can release the textures from the loader cache. Null for fonts
+         * created without going through the resource loader.
+         *
+         * @type {ResourceLoader|null}
+         * @ignore
+         */
+        this._loader = null;
+
         // json data
         this._data = null;
         this.data = data;
+    }
+
+    /**
+     * Frees the GPU textures owned by the font and removes them from the resource loader cache.
+     * Called automatically when the owning font asset is unloaded (see {@link Asset#unload}).
+     */
+    destroy() {
+        const textures = this.textures;
+        if (textures) {
+            for (let i = 0; i < textures.length; i++) {
+                const texture = textures[i];
+
+                // font textures are loaded directly through the resource loader (keyed by their
+                // url, which is stored as the texture name), so remove the cache entry to avoid
+                // handing back a destroyed texture on a subsequent load
+                this._loader?.clearCache(texture.name, 'texture');
+                texture.destroy();
+            }
+            this.textures = null;
+        }
+
+        this._loader = null;
+        this._data = null;
     }
 
     set data(value) {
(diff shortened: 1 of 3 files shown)


mvaligursky (commented):

---

Fix is correct for the standard load → unload → reload case (I confirmed `texture.name` reliably equals the load URL = cache key, since every texture parser sets `name: url`, and `clearCache`/`getFromCache` key on the same `makeKey(url, type)`). Three things to consider, in rough order of likelihood:

1. **Active-use behavior change worth confirming.** Previously the texture leak masked it; now `asset.unload()` actually destroys the GPU textures. If a `TextElement` (or `CanvasFont` consumer) still references the font when its asset is unloaded, it will hold destroyed textures and render with dead resources. This is the intended semantics of unload, but worth confirming the text element drops/handles its font reference gracefully on asset unload rather than erroring — the PR doesn't mention that path.

2. **Shared texture, no reference counting.** `Font.destroy()` unconditionally calls `texture.destroy()` and `clearCache(...)`, and the loader cache (`_cache[key]`) isn't ref-counted. If the same texture URL is ever resolved for more than one live `Font` (e.g. two font assets whose derived `.png` URL collides, or any future path that shares a cached font texture), unloading one font destroys the texture out from under the others. Low probability since each font's atlas PNG is normally unique — but previously the leak hid this, so it's newly reachable. A guard ("only destroy if no other resource holds it") or accepting the limitation with a note would cover it.

3. **Coupling to the parser naming convention.** Keying cache removal on `texture.name` works only because all texture parsers happen to set `name: url`. That's true today across all six parsers, but it couples `Font.destroy` to that convention; if a parser ever set a different `name`, the cache entry would silently leak (texture still freed, but a reload returns a dead entry). Storing the actual loaded URLs on the `Font` at load time would decouple it. Minor / robustness only.

PR #8987, opened 2026-06-30
Use uniform buffers for view-level uniforms on WebGL2

mvaligursky:
View-level uniforms (camera/view matrices, viewport, exposure, clustered-lighting params, etc.) now flow through a uniform buffer on WebGL2, matching the WebGPU path, instead of being committed as individual GL uniforms.

**Changes:**
- Add `WebglShaderProcessorGLSL`, a WebGL2 GLSL processor that injects a std140 `ub_view` block and removes the standalone view-uniform declarations. Attributes, varyings, textures and mesh-level uniforms are left to the existing `gles3` macro path. It runs from `WebglShader` when shader processing options are present.
- Add the WebGL2 uniform-buffer building blocks: `WebglUniformBuffer` (persistent), `WebglBindGroup` / `WebglBindGroupFormat`, a per-size whole-buffer dynamic-buffer pool (`bufferData` orphaning, bound with `bindBufferBase`), and `device.setBindGroup`. Uniform blocks are linked to their binding points in `WebglShader`.
- Make the renderer's view uniform path unconditional on all backends (previously gated on `supportsUniformBuffers`, i.e. WebGPU-only). The mesh-UB and quad-render paths remain gated. The null device gains no-op uniform-buffer support so it runs harmlessly.
- Handle WebGL/WebGPU context loss for both persistent and dynamic uniform buffers.
- XR multiview renders correctly: each eye binds its own view buffer.

GLSL shader authoring is unaffected — view uniforms are still declared as individual `uniform`s (e.g. `uniform mat4 matrix_view;`) and are grouped into the buffer transparently. WebGL2 only (WebGL1 is no longer supported).

**Performance:**
- On WebGL2, view uniforms are uploaded once per view and bound with a single `bindBufferBase`, rather than committed individually per shader. Mesh-level uniforms are unchanged.

### DIFF
--- a/src/framework/graphics/render-pass-picker.js
+++ b/src/framework/graphics/render-pass-picker.js
@@ -194,7 +194,7 @@ class RenderPassPicker extends RenderPass {
                     meshInstances: tempMeshInstances,
                     splitLights: lights,
                     lightClusters: this.emptyWorldClusters,
-                    viewUniformFormat: device.supportsUniformBuffers ? this.getViewUniformFormat() : undefined,
+                    viewUniformFormat: this.getViewUniformFormat(),
                     drawCallback: () => device.setBlendState(this.blendState)
                 });
 
--- a/src/platform/graphics/dynamic-buffer.js
+++ b/src/platform/graphics/dynamic-buffer.js
@@ -33,6 +33,14 @@ class DynamicBuffer {
         ]);
     }
 
+    /**
+     * Upload the buffer's data to the GPU. A no-op on backends (such as WebGPU) that copy the data
+     * to the GPU separately; WebGL overrides this to eagerly upload, as it has no buffer mapping and
+     * executes draws immediately.
+     */
+    upload() {
+    }
+
     getBindGroup(ub) {
         const ubSize = ub.format.byteSize;
         let bindGroup = this.bindGroupCache.get(ubSize);
--- a/src/platform/graphics/null/null-bind-group-format.js
+++ b/src/platform/graphics/null/null-bind-group-format.js
@@ -1,0 +1,11 @@
+/**
+ * A Null implementation of the BindGroupFormat.
+ *
+ * @ignore
+ */
+class NullBindGroupFormat {
+    destroy() {
+    }
+}
+
+export { NullBindGroupFormat };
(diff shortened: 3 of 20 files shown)


mvaligursky (commented):

---

Read the whole PR; the design is sound and consistent with the existing WebGPU UB path. I verified the multiview correctness (each eye's `alloc()` hands out a distinct whole `WebglDynamicBuffer`, and its per-buffer bind-group cache captures that specific buffer, so each eye binds its own — the lazy `bufferId` read at bind time is correctly ordered after `endUpdate`→`upload`), context-loss handling (`device.buffers` is a `Set`, `restoreContext` re-uploads persistent UBs, non-persistent re-allocate next update), and that the std140 block is generated from the same `UniformBufferFormat` the renderer uploads. Two things to consider:

1. **`setBindGroup` binds every buffer in the bind group to the *same* binding point.** The loop iterates `bindGroup.impl.buffers` but always passes `index` to `bindBufferBase`, so for a bind group with more than one uniform buffer, each bind overwrites the previous and only the last survives. This is fine today (the comment notes the single-UB-per-group assumption, and on WebGL2 only the view UB uses a binding point), but the loop *looks* like it supports N buffers while silently only honoring one. A `Debug.assert(buffers.length <= 1)` — or binding to `index + i` — would make the assumption explicit and fail loudly if a future multi-UB bind group reaches this path.

2. **Per-view-setup allocation cost scales with the number of view setups, which is dominated by shadow passes.** Each `setupViewUniformBuffers` (every camera layer + every directional/local shadow render) now allocates a whole pooled GL buffer and does a `bufferData` respecify, versus the old per-shader uniform commits. Pooling avoids cross-frame churn, but a scene with many dynamic shadow-casting lights does N `bufferData` uploads/frame (N = view setups). The view UB for shadows is tiny (single mat4), so this is likely still a net win, but the verified example set may not include heavy many-dynamic-shadow-light scenes — worth a quick confirm there before relying on the perf claim universally.

Minor: the unsupported-block fallback in `WebglShader` (`bindingPoint = bindGroupIndex >= 0 ? bindGroupIndex : i`) links an unrecognized block to its iteration index `i`, which could alias binding point 0 (view) if such a block ever coexisted with `ub_view`. Only reachable via the asserted "user uniform buffers not supported" path, so not a real concern today — noting for completeness.

PR #8995, opened 2026-07-01
Warn about unknown options passed to addComponent

### LINKED ISSUE #8994: Add property checks against the component schema
We can add a property check to early catch typos and mistyped options that we pass when adding a new component to the entity. At the moment those are silently ignored.

For example:

https://github.com/playcanvas/engine/blob/258682678415ac4bdc0234b223af4a532899a11d/src/framework/components/render/system.js#L65-L69

could become:

```js
_properties = new Set([ 'material', 'meshInstances', 'asset', ... ]);

// iterate over smaller _data, instead of larger _properties (we rarely/never pass all props)
for (const [key, value] of Object.entries(_data)) {
    const exists = _properties.has(key);

    Debug.assert(exists, 'Invalid option passed when creating Render Component: ${key}');

    if (exists) {
        component[key] = value;
    }
}
```

mvaligursky:
Options passed to `addComponent` (or `entity.addComponent(type, data)`) with a mistyped name — e.g. `castShadow` instead of `castShadows` — are currently silently ignored. This adds a debug-build check that warns when an option name doesn't correspond to a settable property of the component, so typos are easy to catch. Fixes #8994.

**Changes:**
- `ComponentSystem.addComponent` now validates the supplied option keys and emits a `Debug.warnOnce` for any key that is neither a settable property of the component nor an explicitly allowed, system-consumed key.
- The check is reflective — it looks for a real setter or writable field on the component — so it needs no hand-maintained per-property lists and stays correct as components gain or lose properties. It intentionally errs toward false negatives (never false positives): a getter-only/read-only property is flagged, but a typo that happens to collide with a real member is not.
- The entire check is wrapped in `Debug.call`, so it is stripped from release builds (verified: zero occurrences of the validation code in a release bundle).
- A small `extraDataProperties` allow-list is declared on the few systems that accept options which are not settable component properties:
  - render / model / gsplat: `aabbCenter`, `aabbHalfExtents` (used to build `customAabb`)
  - light: `enable` (deprecated alias for `enabled`)
  - anim: `layers`, `masks` (consumed directly during init)
  - script: `order` (consumed alongside `scripts`)

**Notes:**
- Debug-only and non-fatal: in release builds there is no behavior or size change. In debug builds, previously-ignored bad options now produce a one-time console warning per option name + component type.

### DIFF
--- a/src/framework/components/anim/system.js
+++ b/src/framework/components/anim/system.js
@@ -25,6 +25,10 @@ class AnimComponentSystem extends ComponentSystem {
 
         this.ComponentType = AnimComponent;
 
+        // 'layers' is read-only (managed via addLayer) and 'masks' is not a component property; both
+        // are consumed directly in initializeComponentData
+        this.extraDataProperties = ['layers', 'masks'];
+
         this.on('beforeremove', this.onBeforeRemove, this);
         this.app.systems.on('animationUpdate', this.onAnimationUpdate, this);
     }
--- a/src/framework/components/gsplat/system.js
+++ b/src/framework/components/gsplat/system.js
@@ -132,6 +132,9 @@ class GSplatComponentSystem extends ComponentSystem {
 
         this.ComponentType = GSplatComponent;
 
+        // consumed to build customAabb, not settable component properties
+        this.extraDataProperties = ['aabbCenter', 'aabbHalfExtents'];
+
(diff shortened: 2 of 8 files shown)


LeXXik:
`extraDataProperties` look like a hack.

mvaligursky:
> `extraDataProperties` look like a hack.

A workaround. There are some properties that the scene stores as vec3 for example (due to not having a Bounds inspector), so those are converted on load to it. 

Not ideal, but that's what we have currently.

PR #9001, opened 2026-07-02
Add pluggable parser registry to ResourceHandler and convert handlers

mvaligursky:
Adds a public, uniform way to register custom file-format parsers on any resource handler — previously only `ModelHandler` had an (undocumented) `addParser`, and formats like SOG, compressed PLY or custom texture codecs could not be plugged in without forking handlers.

**Changes:**
- `ResourceHandler` gains a parser registry: `addParser(parser)`, `removeParser(parser)` and a read-only `parsers` getter. Parsers are selected newest-first, so a later registration overrides a built-in parser for the same format
- A parser implements `canParse(context)` (context provides `url`, `ext`, `basename`, `asset`, `app`) plus `load`/`open`; on registration the handler assigns itself to `parser.handler`, and parsers fetch via the new `ResourceHandler#fetch` helper
- Converted handlers: model, json, gsplat, container, css/html/text/shader/binary, animation, texture. Retry configuration (`maxRetries`) now lives on the handler as the single source of truth
- `Texture` constructor gains an `srgb` option — the texture is created directly in the sRGB variant of the requested format, replacing per-parser format translation
- Texture parser fixes surfaced by the new characterization example: dds created linear then format-switched at patch time (expensive switch + warning); single-level ktx/ktx2/basis files rendered black (treated as an incomplete mip chain)
- The hdr parser now enforces RGBE requirements (linear format, no mipmaps) over asset-provided options

**API Changes:**
- New: `ResourceHandler#addParser`, `#removeParser`, `#parsers`, `#fetch`; `ResourceParser` / `ParserContext` typedefs; `Texture` constructor `srgb` option

```js
// register a custom texture format
class TgaParser {
    canParse(context) { return context.ext === 'tga'; }
    load(url, callback, asset) { this.handler.fetch(url, pc.Http.ResponseType.ARRAY_BUFFER, callback, asset); }
    open(url, data, device, textureOptions) { /* decode → new pc.Texture(...) */ }
}
app.loader.getHandler('texture').addParser(new TgaParser());
```

- Breaking: the `decider` argument of `addParser` is removed — implement `canParse(context)` on the parser instead (the argument is still accepted but ignored, with a debug warning); the `AddParserCallback` type is removed
- Breaking (undocumented internals): the texture/gsplat/container handlers' internal `parsers` maps are replaced by the registry — code such as `handler.parsers.tga = ...` must use `addParser`
- Behavior: loading a gsplat with an unrecognized extension now fails with a clear "no parser found" error instead of silently attempting the PLY parser

**Examples:**
- New hidden `test/texture-formats` example loading png/dds/ktx2/basis/hdr in a grid, exercising every texture parser
- `loaders/obj` updated to the new `addParser` signature (the OBJ sample parser implements `canParse`)


src/framework/handlers/handler.js
```diff
@@ -69,7 +114,122 @@ class ResourceHandler {
...
+    /**
+     * Selects a parser for the given context, consulting registered parsers newest-first.
+     *
+     * @param {ParserContext} context - The context built by {@link ResourceHandler#_makeContext}.
+     * @returns {ResourceParser|null} The first parser whose `canParse` returns true, or null.
+     * @ignore
+     */
+    _selectParser(context) {
+        for (let i = this._parsers.length - 1; i >= 0; i--) {
+            if (this._parsers[i].canParse(context)) {
```
mvaligursky, line 238: The announced breaking change (`addParser(parser, decider)` removed) has a harsher failure mode in release builds than the PR intends. `Debug.assert`/`Debug.removed` are stripped in release, so a legacy parser that lacks `canParse` still gets pushed into `_parsers` — and since selection is newest-first, this line then calls `undefined(context)` and throws a TypeError for **every** subsequent load of that handler type (all models, not just the custom format), with no hint of the cause.

Suggestion: make selection defensive so a non-conforming parser is inert rather than fatal:
```js
const p = this._parsers[i];
if (typeof p.canParse === 'function' && p.canParse(context)) {
    return p;
}
```
(or refuse the registration outside the Debug-only path). That preserves the intended "debug warning + custom format stops working" outcome instead of breaking the whole handler in production.

mvaligursky (commented):

---

Reviewed the registry core, all handler conversions, the texture parsers and the srgb path; ran the PR's test files (65 passing) and lint (clean) locally. The design is consistent — every converted parser implements `canParse`/`load`, no leftover references to the old parser-map internals, registration order gives correct newest-first precedence (incl. the documented `lod-meta.json` octree-vs-SOG case), and `TextureHandler.patch` no longer triggers the expensive sRGB format switch since parsers now create textures in the right variant. One substantive point, two minors:

1. **The removed `decider` back-compat fails much harder in release builds than intended** — see inline comment on `_selectParser`. A legacy parser without `canParse` breaks *every* load for that handler type with an opaque TypeError, instead of the intended debug-only warning.

2. **Parser selection runs twice per load (in `load` and again in `open`) and assumes the registry is unchanged in between.** If a user calls `addParser`/`removeParser` while loads are in flight, `open` can select a different parser than the one whose `load` fetched the data (e.g. a newly registered catch-all claiming data fetched by a built-in). Extremely unlikely in practice, but a one-line note on `addParser`'s docs ("register parsers before starting loads") — or caching the selected parser per in-flight load — would close it.

3. **`TextureHandler.open` calls `parser.open` without a null guard.** Safe today because `ImgParser.canParse()` is an unconditional catch-all, but `removeParser` is now public API — a user removing the img parser (e.g. to force-fail unknown formats) plus a mid-flight registry change could make `_selectParser` return null here and throw a raw TypeError. A `if (!parser) return undefined;` (or Debug.assert) would keep the failure explicit. Minor.

PR #9013, opened 2026-07-02
Support pen input in touch sources (fix #9012)

### LINKED ISSUE #9012: Touch input : Apple Pencil input is ignored in PlayCanvas touch controls
Apple Pencil events come through as pointerType: `pen`, but MultiTouchSource only handles pointerType: `touch`. As a result, stylus drag/orbit gestures are ignored even though the same interaction works with fingers.

Could MultiTouchSource treat "pen" like touch input, or expose a way to opt into pen support ?

d0rianb:
## Support pen input in touch gesture sources
  - Treat `pen` pointer events like `touch` in `MultiTouchSource` and `SingleGestureSource`.
  - Preserve existing mouse and touch behavior.

Tested with the "Orbit" example on an iPad Pro (ios 26.5, with an apple pencil)

### DIFF
--- a/src/extras/input/sources/multi-touch-source.js
+++ b/src/extras/input/sources/multi-touch-source.js
@@ -1,6 +1,6 @@
 import { Vec2 } from '../../../core/math/vec2.js';
 import { InputSource } from '../input.js';
-import { movementState } from '../utils.js';
+import { isTouchPointer, movementState } from '../utils.js';
 
 const tmpVa = new Vec2();
 
@@ -56,7 +56,7 @@ class MultiTouchSource extends InputSource {
         const { pointerId, pointerType } = event;
         this._movementState.down(event);
 
-        if (pointerType !== 'touch') {
+        if (!isTouchPointer(pointerType)) {
             return;
         }
         this._element?.setPointerCapture(pointerId);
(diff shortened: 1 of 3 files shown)


mvaligursky (commented):

---

Fix looks correct. I checked for double-handling: `KeyboardMouseSource` filters on `pointerType === 'mouse'`, so pen events were previously ignored by *every* source and are now claimed only by the touch sources — no conflict. One gap and one nit:

1. **`DualGestureSource` was left out.** It has the same three `pointerType !== 'touch'` early-outs (`_onPointerDown` / `_onPointerMove` / `_onPointerUp`, around lines 137 / 174 / 211 of `dual-gesture-source.js`) as the two sources you converted. As it stands, a pen works with orbit/fly (single/multi-touch sources) but is still dead in the dual-joystick gesture source. Converting it to `isTouchPointer` too would make pen support consistent across all touch-style sources — or if the exclusion is deliberate, a brief note in the PR would help.

2. Naming nit: `isTouchPointer` returning `true` for `'pen'` reads slightly off at call sites. Something like `isTouchLikePointer` (or a short comment noting pen is intentionally treated as touch) would make the intent self-evident. Entirely optional.

mvaligursky:
Looks great, please fix the lint.

mvaligursky (approved):

d0rianb:
Not related but have you considered oxlint in favor of eslint ? 
Perf improvement is 145ms vs 21s on my m5

PR #9014, opened 2026-07-02
style: normalize examples source structure

kpal81xd:
## Description
- Adds tracked Prettier config and format scripts for examples source.
- Stacks `eslint-config-prettier` for examples source, then re-enables JS equivalents of the beta TypeScript style rules where applicable.
- Extends example linting to keep asset-loading setup flat with top-level await.
- Keeps explicit arrow parameter parentheses and converts safe example function expressions to arrows.
- Rewrites examples to import PlayCanvas symbols directly and enforces no `pc.*` namespace usage in examples source.
- Reformats examples source to use the same clean structure consistently.

## Checklist

## Manual smoke test
1. Open the examples browser and load `misc/hello-world`.
   Expected: the cube renders and rotates normally.
2. Open `loaders/glb`.
   Expected: the GLB scene loads and camera/light switching still works.
3. Open `test/texture-copy`.
   Expected: the overlay reports all texture copy checks passed.


mvaligursky:
Do we have different eslint rules than in the engine code here?

examples/src/examples/misc/html-texture.example.mjs
```diff
@@ -6,12 +6,12 @@
...
 
 //
-// This example demonstrates the HTML-in-Canvas API: a styled HTML element with
-// CSS animations is appended to a canvas marked with the "layoutsubtree"
-// attribute, then captured into a GPU texture (texElementImage2D on WebGL,
-// copyElementImageToTexture on WebGPU).
+// this example demonstrates the html-in-canvas api: a styled html element with
+// css animations is appended to a canvas marked with the "layoutsubtree"
+// attribute, then captured into a gpu texture (texelementimage2d on webgl,
+// copyelementimagetotexture on webgpu).
```
willeastcott: What's with the aggressive lower-casing?

kpal81xd:
> Do we have different eslint rules than in the engine code here?

Yes these only apply to the examples themselves so their formatting is consistent

examples/src/examples/misc/html-texture.example.mjs
kpal81xd: Dropped

PR #9027, opened 2026-07-05
Add CircleGeometry procedural geometry class

mvaligursky:
Adds a procedural circle (flat disc) geometry to the existing geometry family (Plane, Box, Sphere, Cylinder, ...), useful for water surfaces, ground discs and similar radial meshes.

**Changes:**
- New `CircleGeometry` class (`src/scene/geometry/circle-geometry.js`) - a disc in the XZ plane with a +Y normal, matching the conventions of `PlaneGeometry` (CCW winding, planar 0-1 UVs across the bounding square, `uvs1 = uvs`, optional tangent generation)
- Options: `radius`, `sectors` (divisions around the circumference), `rings` (concentric vertex rings) and `ringExponent` - values above 1 concentrate the rings towards the center, so a large disc can carry dense tessellation near the viewer and cheap large triangles towards the rim (the main use case: water / terrain discs reaching the horizon)
- Exported from `src/index.js`

**API Changes:**
- New public class `pc.CircleGeometry`

**Examples:**
- Used by an upcoming water example (separate PR)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -196,6 +196,7 @@ export { StandardMaterialOptions } from './scene/materials/standard-material-opt
 // SCENE / PROCEDURAL
 export { calculateNormals, calculateTangents } from './scene/geometry/geometry-utils.js';
 export { CapsuleGeometry } from './scene/geometry/capsule-geometry.js';
+export { CircleGeometry } from './scene/geometry/circle-geometry.js';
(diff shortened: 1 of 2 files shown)


LeXXik:
It could have an inner circle radius, allowing for hollow shapes, like a nimbus.

PR #9035, opened 2026-07-06
Fix crash when unloading sog assets still referenced by the unified world

slimbuck:
Unloading a sog-based gsplat asset destroyed its source webp textures immediately, even while the unified gsplat world was still rendering from them. The sog parsers register an asset.once('unload') handler that unloads the child texture assets, which bypasses resource ref-counting entirely: GSplatResourceBase.destroy() correctly defers GPU destruction while refCount > 0 (via GSplatResourceCleanup), but the parser-level handler killed the textures out-of-band.

This crashes when an octree (lod-meta) asset is unloaded to switch views: GSplatOctreeResource.destroy() → octree.destroy() → assetLoader.destroy() force-unloads every node-file asset while world states still hold refs to their resources, and the work-buffer bake re-reads the sog textures (camera movement alone queues SH color re-renders). On WebGPU the bake binds the destroyed textures: NULL texture view asserts, an uncaught exception inside GSplatWorkBufferRenderPass, and unrecoverable pass-state corruption ("submit while inside a pass" cascades). The streaming per-file unload on LOD transitions has the same hazard. It only reproduces when other unified splats keep the world alive across the unload (e.g. permanent standalone splats in the scene); with no remaining placements the manager is torn down and nothing bakes the stale state.

The fix moves texture ownership to the resource: once GSplatSogResource is constructed, the parser's unload handler only removes the child texture assets from the registry, and the textures are destroyed solely by the resource's ref-count deferred _actualDestroy() (via GSplatSogData.destroy(), which already covers all source textures plus the codebook). While the load is still in flight the handler behaves as before, unloading the textures to cancel the load. The decompress path is unchanged, as ownership never transfers to the resource there.

Applies to both the sog.js and sog-bundle.js parsers.

### DIFF
--- a/src/framework/parsers/sog-bundle.js
+++ b/src/framework/parsers/sog-bundle.js
@@ -261,8 +261,9 @@ class SogBundleParser {
             let ownedByResource = false;
 
             // When the parent gsplat asset unloads, remove child texture assets from the
-            // registry. Destroy their texture resources only while the load is still in flight
-            // (cancellation); once owned by the resource, destruction is left to it.
+            // registry. Destroy their texture resources only until GSplatSogResource ownership
+            // has been transferred (cancellation before ownership transfer); once owned by the
+            // resource, destruction is left to it.
             asset.once('unload', () => {
                 Object.values(textures).forEach((t) => {
                     // remove from registry
--- a/src/framework/parsers/sog-bundle.js
+++ b/src/framework/parsers/sog-bundle.js
@@ -255,13 +255,23 @@ class SogBundleParser {
 
             const { assets } = this.app;
 
+            // Set once the GSplatSogResource takes ownership of the textures: from then on they
+            // may only be destroyed through the resource's (ref-count deferred) destroy, as the
+            // unified world may still be rendering from them when the asset unloads.
+            let ownedByResource = false;
+
+            // When the parent gsplat asset unloads, remove child texture assets from the
+            // registry. Destroy their texture resources only while the load is still in flight
+            // (cancellation); once owned by the resource, destruction is left to it.
             asset.once('unload', () => {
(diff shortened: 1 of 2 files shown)


mvaligursky (commented):

---

Fix looks correct. I verified the ownership transfer is complete and safe: all seven parser-loaded textures (`means_l/u`, `quats`, `scales`, `sh0`, `sh_centroids`, `sh_labels`) are exactly the set destroyed by `GSplatSogData._destroyGpuResources()`; `Texture.destroy()` is `if (device)`-guarded, so the cancellation path's double-destroy (unload handler `t.unload()` followed later by `resource.destroy()` → `data.destroy()`) is a no-op; and there is no await between the `GSplatSogResource` constructor and the `ownedByResource = true` assignment, so the unload handler can't observe a half-transferred state. One adjacent gap worth considering:

**`sog-bundle.js` has no mid-flight abort handling (`unloaded` flag / `_shouldAbort`), unlike `sog.js`.** If the parent asset unloads while a bundle load is in flight, the unload handler correctly cancels the textures (`ownedByResource` is still false), but the load then continues: it constructs a `GSplatSogResource` over the already-destroyed textures, sets `ownedByResource = true`, and delivers it via `callback(null, resource)` to an asset that's already unloaded — so the resource (and its codebook texture) is never destroyed and leaks, and any consumer that picks it up binds dead textures. This is pre-existing rather than introduced here, but the PR touches exactly this code in both parsers and the fix mirrors sog.js in every other respect — porting the `unloaded`/`_shouldAbort` pattern over (destroy the resource and call back null, as sog.js does) would close the same class of crash for the bundle path.

mvaligursky (approved):

PR #9037, opened 2026-07-06
Fix sog bundle parser continuing after asset unloads mid-load

mvaligursky:
Follow-up to #9035. The `.sog` bundle parser (`sog-bundle.js`) had no mid-flight abort handling, unlike `sog.js`. If the parent asset unloaded while a bundle load was in flight, the unload handler correctly destroyed the texture resources, but the load then continued: it constructed a `GSplatSogResource` over the already-destroyed textures, set `ownedByResource = true`, and delivered the resource to the unloaded asset — leaking the resource (and its codebook texture) and letting any consumer bind dead textures.

**Changes:**
- Ported the `unloaded` flag / `_shouldAbort` pattern from `sog.js` to `SogBundleParser`, with abort checkpoints after the archive download, after texture loading, around GPU data preparation, and after resource construction — each cleaning up what exists at that point (texture assets, `GSplatSogData`, or the resource) and cancelling via `callback(null, null)`, same as `sog.js`.
- Moved the `asset.once('unload')` handler registration to before the first `await`. Previously it was registered only after all textures finished loading, so an unload during the zip download or texture loading left the child texture assets in the registry and leaked their resources.

Verified in the browser: normal `.sog` bundle loading unaffected (gaussian-splatting/editor example), and an unload triggered mid-download now aborts cleanly with no leaked texture assets and no errors.

### DIFF
--- a/src/framework/parsers/sog-bundle.js
+++ b/src/framework/parsers/sog-bundle.js
@@ -157,6 +157,20 @@ class SogBundleParser {
     }
 
     /**
+     * Checks if loading should be aborted due to asset unload or invalid device.
+     *
+     * @param {Asset} asset - The asset being loaded.
+     * @param {boolean} unloaded - Whether the asset was unloaded during async loading.
+     * @returns {boolean} True if loading should be aborted.
+     * @private
+     */
+    _shouldAbort(asset, unloaded) {
+        if (unloaded || !this.app.assets.get(asset.id)) return true;
+        if (!this.app?.graphicsDevice || this.app.graphicsDevice._destroyed) return true;
+        return false;
+    }
+
+    /**
      * @param {object} url - The URL of the resource to load.
      * @param {string} url.load - The URL to use for loading the resource.
      * @param {string} url.original - The original URL useful for identifying the resource type.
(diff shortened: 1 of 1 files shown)


mvaligursky:
Re the Copilot comments: all three describe the same edge case — an abort triggered *without* the parent's `unload` event firing (parent removed via `assets.remove()` without `unload()`, or graphics device destroyed) at one of the later checkpoints, which destroy the `GSplatSogData`/resource but leave the child texture assets in the registry. Leaving this as-is intentionally, for these reasons:

- **The main cancellation path is already covered.** Mid-flight the parent asset is never `loaded`, so `Asset.unload()` early-returns and the `unload` event cannot fire during a load anyway — cancellation is detected via the `!assets.get(asset.id)` registry check in `_shouldAbort`. The checkpoint right after `Promise.allSettled` handles that case with full cleanup (removes and unloads all child texture assets). The remaining exposure is only the short `prepareGpuData()` / `decompress()` awaits after that point.

- **The device-teardown variant is negligible.** The graphics device is only destroyed when the application itself is being destroyed — `AppBase.destroy()` unloads assets and tears down the asset registry along with the device. A texture asset left registered for the final moments of app teardown has no observable effect; there is no ongoing rendering or asset lookup that could bind it, and everything is released together.

- **This mirrors `sog.js` by design.** The equivalent checkpoints in `sog.js` behave identically today (`data.destroy()` / `resource.destroy()` only, no texture-asset registry cleanup). The goal of this PR is to bring the bundle parser in line with the existing, shipped pattern.

If we decide the later checkpoints should also remove/unload the child texture assets, that's best done as a small follow-up covering both parsers at once rather than diverging them here.

PR #9043, opened 2026-07-06
Decouple physics into a backend abstraction with Ammo and Null implementations

### LINKED ISSUE #2801: Decouple physics engine

willeastcott:
## Description

Decouples the engine from ammo.js by introducing an internal physics backend contract, as proposed in #2801: `PhysicsWorld`, `PhysicsBody` and `PhysicsJoint` base classes under `src/framework/physics/`, with all Ammo code ported behind them (`physics/ammo/`) and a no-op `NullPhysicsWorld` that lets the physics component lifecycle run in tests without a physics engine. Adding a new backend (Jolt, Rapier, a custom engine) is now a matter of implementing the three classes.

Closes #2801

**Zero public API change.** Component APIs, events and constants are untouched. Ammo is auto-selected exactly as before (`typeof Ammo` after WasmModule load) - now the only reference to the Ammo global outside `physics/ammo/`. Backends install through a single `@ignore`d seam, `RigidBodyComponentSystem#setPhysicsWorld()`; a public selection API (e.g. via `AppOptions`) is deferred until a second real backend lands. The unsupported native escape hatches keep returning raw Ammo objects (`rigidbody.body`, `dynamicsWorld`, `joint.constraint`, `collision.shape`), so `scripts/physics/vehicle.js` runs untouched.

Notes for reviewers:

- Shapes are opaque handles (the raw `btCollisionShape` for Ammo) created from typed descriptors; per-type creation inside the backend uses module-scope dispatch tables. Mesh sources expose vertex/index data through lazy accessors, so geometry cached per mesh id is never re-extracted.
- Contact reporting crosses the boundary via a listener (`onContactsBegin`/`onContactPair`/`onContactsEnd`) with one reused pair object; all pooling, collision maps and event firing stay in the system, and per-substep semantics (and the no-tick-callback fallback ordering) are preserved.
- Joint frames cross the boundary as scale-free `Mat4`s with X as the primary axis; the backend applies its native axis correction (bullet's hinge Z remap) in matrix space, keeping results identical to the previous single-sided computation.
- Behavior verified bit-exact against master with a real-Ammo A/B harness (~20 scenarios: falling/resting bodies, raycasts, contact/trigger event frame ordering and impulses, all 5 joint types, teleport interpolation from #8915, compound child moves, mesh/hull colliders, raw-vehicle escape hatches), plus all 7 physics examples in the browser and 26 new null-backend unit tests.
- Deliberate deviations, all internal: bt math temps are now per-world instead of cross-app statics (fixes a latent multi-app teardown bug); mesh scale-change rebuilds compare against the recorded build scale (removes redundant render-path rebuilds for non-unit scales); scene-settings gravity is applied without the Ammo-global gate; hinge motors use the last-stepped `fixedTimeStep`; the undocumented `system.createBody/destroyBody`, `collision.getCompoundChildShapeIndex` and `joint system getFixedBody` were removed (the world-anchor fixed body lives on the backend now).

Follow-ups: public backend-selection API, and flattening the `CollisionSystemImpl` class hierarchy (its `createPhysicalShape` methods are now one-liners).

## Checklist

### DIFF
--- a/src/framework/components/collision/component.js
+++ b/src/framework/components/collision/component.js
@@ -182,6 +182,14 @@ class CollisionComponent extends Component {
     _hasOffset = false;
 
     /**
+     * The entity world scale a mesh shape was last built with - a change triggers a rebuild.
+     *
+     * @type {Vec3|null}
+     * @private
+     */
+    _builtWorldScale = null;
+
+    /**
      * Create a new CollisionComponent.
      *
      * @param {CollisionComponentSystem} system - The ComponentSystem that created this Component.
@@ -607,35 +615,12 @@ class CollisionComponent extends Component {
     }
 
     /**
-     * @param {object} shape - The Ammo collision shape to find.
-     * @returns {number|null} The shape's index in the child array of the compound shape.
-     * @ignore
-     */
-    getCompoundChildShapeIndex(shape) {
-        const compound = this._shape;
-        const shapes = compound.getNumChildShapes();
-
-        for (let i = 0; i < shapes; i++) {
-            const childShape = compound.getChildShape(i);
-            if (Ammo.getPointer(childShape) === Ammo.getPointer(shape)) {
-                return i;
-            }
-        }
-
-        return null;
-    }
-
-    /**
      * @param {GraphNode} parent - The parent node.
      * @private
      */
     _onInsert(parent) {
-        // TODO
-        // if is child of compound shape
-        // and there is no change of compoundParent, then update child transform
-        // once updateChildTransform is exposed in ammo.js
-
-        if (typeof Ammo === 'undefined') {
+        const world = this.system.physicsWorld;
+        if (!world) {
             return;
         }
 
@@ -645,7 +630,7 @@ class CollisionComponent extends Component {
             let ancestor = this.entity.parent;
             while (ancestor) {
                 if (ancestor.collision && ancestor.collision.type === 'compound') {
-                    if (ancestor.collision.shape.getNumChildShapes() === 0) {
+                    if (world.getCompoundChildCount(ancestor.collision.shape) === 0) {
                         this.system.recreatePhysicalShapes(ancestor.collision);
                     } else {
                         this.system.recreatePhysicalShapes(this);
@@ -736,12 +721,12 @@ class CollisionComponent extends Component {
                 this.entity.rigidbody.enableSimulation();
             }
         } else if (this._compoundParent && this !== this._compoundParent) {
-            if (this._compoundParent.shape.getNumChildShapes() === 0) {
+            if (this.system.physicsWorld.getCompoundChildCount(this._compoundParent.shape) === 0) {
                 this.system.recreatePhysicalShapes(this._compoundParent);
             } else {
-                const transform = this.system._getNodeTransform(this.entity, this._compoundParent.entity);
-                this._compoundParent.shape.addChildShape(transform, this._shape);
-                Ammo.destroy(transform);
+                // the shape is known to be absent from the compound (it was removed on
+                // disable), so a forced child update adds it back at the current pose
+                this.system.updateCompoundChildTransform(this.entity, true);
 
                 if (this._compoundParent.entity.rigidbody) {
                     this._compoundParent.entity.rigidbody.activate();
--- a/src/framework/components/collision/system.js
+++ b/src/framework/components/collision/system.js
@@ -628,7 +628,9 @@ class CollisionComponentSystem extends ComponentSystem {
     }
 
     onRemove(entity) {
-        if (entity.rigidbody && entity.rigidbody.body) {
+        // gate on the backend body, not the public getter - the getter surfaces the NATIVE
+        // body, which backends without native handles keep null
+        if (entity.rigidbody && entity.rigidbody._body) {
             entity.rigidbody.disableSimulation();
         }
 
--- a/src/framework/components/collision/system.js
+++ b/src/framework/components/collision/system.js
@@ -16,8 +16,9 @@ import { Trigger } from './trigger.js';
 const mat4 = new Mat4();
 const p1 = new Vec3();
 const p2 = new Vec3();
+const p3 = new Vec3();
 const quat = new Quat();
-const tempGraphNode = new GraphNode();
+const quat2 = new Quat();
 
 // Note that `shape` is deliberately absent from this list - it is runtime
 // state created and owned by the type implementation, not component data
@@ -66,8 +67,9 @@ class CollisionSystemImpl {
     // Re-creates rigid bodies / triggers
     recreatePhysicalShapes(component) {
         const entity = component.entity;
+        const world = this.system.physicsWorld;
 
-        if (typeof Ammo !== 'undefined') {
+        if (world) {
             if (entity.trigger) {
                 entity.trigger.destroy();
                 delete entity.trigger;
@@ -111,7 +113,7 @@ class CollisionSystemImpl {
 
             if (component._compoundParent) {
                 if (component !== component._compoundParent) {
-                    if (firstCompoundChild && component._compoundParent.shape.getNumChildShapes() === 0) {
+                    if (firstCompoundChild && world.getCompoundChildCount(component._compoundParent.shape) === 0) {
                         this.system.recreatePhysicalShapes(component._compoundParent);
                     } else {
                         this.system.updateCompoundChildTransform(entity, true);
@@ -165,7 +167,7 @@ class CollisionSystemImpl {
 
     destroyShape(component) {
         if (component._shape) {
-            Ammo.destroy(component._shape);
+            this.system.physicsWorld.destroyShape(component._shape);
             component._shape = null;
         }
     }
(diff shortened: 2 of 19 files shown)


src/framework/components/collision/system.js
```diff
@@ -756,30 +643,13 @@ class CollisionComponentSystem extends ComponentSystem {
...
-            const idx = parentComponent.getCompoundChildShapeIndex(entity.collision.shape);
-            if (idx === null) {
-                parentComponent.shape.addChildShape(transform, entity.collision.shape);
-            } else {
-                parentComponent.shape.updateChildTransform(idx, transform, true);
-            }
-            Ammo.destroy(transform);
+            this._getNodeTransform(entity, parentComponent.entity, p3, quat2);
+            this.physicsWorld.updateCompoundChild(parentComponent.shape, entity.collision.shape, p3, quat2);
         }
```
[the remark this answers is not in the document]
willeastcott: Not reachable: `_compoundParent` is only ever assigned inside world-gated paths (`recreatePhysicalShapes`/`_onInsert`), so a registered compound child implies an installed backend. This matches the previous code's implicit invariant, where compound shapes could only exist once Ammo had loaded.

src/framework/components/collision/system.js
```diff
@@ -756,30 +643,13 @@ class CollisionComponentSystem extends ComponentSystem {
...
-        if (collision.shape.removeChildShape) {
-            collision.shape.removeChildShape(shape);
-        } else {
-            const ind = collision.getCompoundChildShapeIndex(shape);
-            if (ind !== null) {
-                collision.shape.removeChildShapeByIndex(ind);
-            }
-        }
+        this.physicsWorld.removeCompoundChild(collision.shape, shape);
     }
```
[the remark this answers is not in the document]
willeastcott: Same invariant as above: `component._shape` is non-null only when the world that created it is installed, and entities are torn down before systems on app destroy.

src/framework/components/rigid-body/system.js
```diff
@@ -269,41 +276,7 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
-                    new Vec3(normal.x(), normal.y(), normal.z()),
-                    rayCallback.get_m_closestHitFraction()
-                );
-            }
-        }
-
-        Ammo.destroy(rayCallback);
-
-        return result;
+        return this._world.raycastFirst(start, end, options);
```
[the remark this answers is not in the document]
willeastcott: Behavior-identical by design: the previous code also threw here without Ammo (`ammoRayStart` was never created). The no-backend contract remains unchanged - raycasts require a loaded physics library.

src/framework/components/rigid-body/system.js
```diff
@@ -353,58 +326,12 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
-            }
-
-            if (options.sort) {
-                results.sort((a, b) => a.hitFraction - b.hitFraction);
-            }
-        }
-
-        Ammo.destroy(rayCallback);
-
         return results;
```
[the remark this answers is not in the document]
willeastcott: Same as `raycastFirst` - the previous code threw on `Debug.assert(Ammo.AllHitsRayResultCallback)` / `new Ammo...` without Ammo, so returning `[]` here would be a behavior change rather than a preservation.

src/framework/physics/physics-world.js
```diff
@@ -0,0 +1,362 @@
...
+ * @property {number[]|Uint16Array|Uint32Array} indices - Triangle indices. Ignored when
+ * convexHull is true.
+ * @property {number} base - The first index to read.
+ * @property {number} count - The number of indices to read (a multiple of 3). Ignored when
+ * convexHull is true - hulls consume every position.
+ * @property {boolean} convexHull - Build a convex hull instead of a triangle mesh.
+ * @property {boolean} checkDuplicates - Weld duplicate vertices while building triangle data.
+ * @property {Vec3|null} shapeScale - Scaling to apply to the built sub-shape, or null.
+ * @property {Vec3} position - The sub-shape position within the mesh shape. Collision component
+ * offsets are already applied by the caller.
```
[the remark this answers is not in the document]
willeastcott: Good catch - fixed in db487bfbe.

src/framework/components/rigid-body/component.js
```diff
@@ -287,7 +261,7 @@ class RigidBodyComponent extends Component {
     }
 
     get body() {
-        return this._body;
+        return this._body ? this._body.nativeBody : null;
     }
```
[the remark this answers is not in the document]
willeastcott: Good catch. A sweep found exactly one remaining internal call site gating on the public getter (`CollisionComponentSystem#onRemove`) - fixed in db487bfbe to gate on `_body`, with a regression test. Everything else already checks `_body`/`_simulationEnabled`.

src/framework/components/collision/component.js
```diff
@@ -736,12 +721,12 @@ class CollisionComponent extends Component {
                 this.entity.rigidbody.enableSimulation();
             }
         } else if (this._compoundParent && this !== this._compoundParent) {
-            if (this._compoundParent.shape.getNumChildShapes() === 0) {
+            if (this.system.physicsWorld.getCompoundChildCount(this._compoundParent.shape) === 0) {
                 this.system.recreatePhysicalShapes(this._compoundParent);
             } else {
```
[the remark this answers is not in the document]
willeastcott: Not reachable: this branch requires `_compoundParent`, which is only assigned under an installed backend (see the `updateCompoundChildTransform` thread).

src/framework/physics/ammo/ammo-physics-shape.js
```diff
@@ -0,0 +1,348 @@
...
+    compound.addChildShape(getTransform(world, source.position, source.rotation), triMeshShape);
+}
+
+/**
+ * Per-type native shape creation. Sizes use engine conventions (full heights) and are
+ * converted to Bullet conventions here.
+ *
+ * @type {Object<string, (world: AmmoPhysicsWorld, desc: PhysicsShapeDesc) => object>}
+ */
+const shapeFactories = {
```
AlexAPPi, line 169: Need Make method for add other factories for examples btHeightfieldTerrainShape

src/framework/physics/physics-body.js
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * The base class for a rigid body owned by a {@link PhysicsWorld}. Backends subclass it and
+ * override every method. The base implementation is inert: setters do nothing and getters
+ * leave their out-parameters untouched, so callers fall back to their own cached values -
+ * matching the engine's behavior when no physics library is loaded.
+ *
+ * @ignore
+ */
+class PhysicsBody {
```
AlexAPPi, line 15: It would be great to add a custom interpolation type.

src/framework/components/collision/trigger.js
```diff
@@ -70,26 +70,18 @@ class Trigger {
...
 
-    _getEntityTransform(transform) {
-        const bodyPos = this.component.getShapePosition();
-        const bodyRot = this.component.getShapeRotation();
-        _ammoVec1.setValue(bodyPos.x, bodyPos.y, bodyPos.z);
-        _ammoQuat.setValue(bodyRot.x, bodyRot.y, bodyRot.z, bodyRot.w);
-
-        transform.setOrigin(_ammoVec1);
-        transform.setRotation(_ammoQuat);
+    _getEntityTransform(position, rotation) {
```
AlexAPPi, line 77: Consider a faster way to read the transformation.
For example, when you can fill in both the position and the rotation at once.

src/framework/components/rigid-body/system.js
```diff
@@ -138,37 +137,73 @@ class RigidBodyComponentSystem extends ComponentSystem {
...
-            ammoRayEnd = new Ammo.btVector3();
-            RigidBodyComponent.onLibraryLoaded();
-
-            this.contactPointPool = new ObjectPool(ContactPoint, 1);
-            this.contactResultPool = new ObjectPool(ContactResult, 1);
-            this.singleContactResultPool = new ObjectPool(SingleContactResult, 1);
-
-            this.app.systems.on('update', this.onUpdate, this);
-        } else {
+        if (!this._world && typeof Ammo !== 'undefined') {
```
AlexAPPi, line 140: What if I have a different handler? For example, I only use heap pointers? And I want to replace the default handler with my own?

src/framework/components/rigid-body/component.js
```diff
@@ -1096,54 +1033,43 @@ class RigidBodyComponent extends Component {
...
+                const ao = component.angularOffset;
+
+                // Un-rotate the angular offset and then use the new rotation to
+                // un-translate the linear offset in local space
+                // Order of operations matter here
+                const invertedAo = _quat2.copy(ao).invert();
+                const entityRot = _quat1.copy(_rotation).mul(invertedAo);
+
+                entityRot.transformVector(lo, _vec3);
+                entity.setPosition(_position.x - _vec3.x, _position.y - _vec3.y, _position.z - _vec3.z);
```
AlexAPPi: Replace to entity.setPositionAndRotation(....)

src/framework/components/rigid-body/component.js
```diff
@@ -1096,54 +1033,43 @@ class RigidBodyComponent extends Component {
...
+                entity.setRotation(entityRot);
+
+            } else {
+                entity.setPosition(_position);
```
AlexAPPi: entity.setPositionAndRotation(...)

AlexAPPi (commented):
Quick notes

Maksims:
As I understand, currently the public API will stay the same. Since we've did implement Jolt as alternative physics implementation and re-used Collision and Rigidbody components for it, we've realized that there are some fundamental and conceptual differences between physics engines, that might prove trickly to implement some of the APIs.

So it definitely worth investigating API layouts and features of various physics engines: Ammo.js, Jolt, Box3D, Cannon.js and others. And figuring out if current API layout is good, or there might be some improvements to be done, as now is the best opportunity for it.

AlexAPPi:
> As I understand, currently the public API will stay the same. Since we've did implement Jolt as alternative physics implementation and re-used Collision and Rigidbody components for it, we've realized that there are some fundamental and conceptual differences between physics engines, that might prove trickly to implement some of the APIs.
> 
> So it definitely worth investigating API layouts and features of various physics engines: Ammo.js, Jolt, Box3D, Cannon.js and others. And figuring out if current API layout is good, or there might be some improvements to be done, as now is the best opportunity for it.

Practice shows that in production today only a few physics engines are still used — over recent years sustained demand has essentially remained for just Ammo and Jolt; the others have fallen out of use or are rarely encountered. Given that, and considering the complexity and amount of work required to create your own engine or to study a niche one in depth, I have serious doubts about whether it’s worth spending time and resources on this now.

Example: https://github.com/Gamebop/physics

src/framework/physics/ammo/ammo-physics-shape.js
willeastcott: The shape dispatch table is deliberately internal for this first step - the goal of this PR is a behavior-identical port, so it only covers the types the engine supported before. Custom native shapes keep working exactly as they did (build the raw shape and assign it via `collision.shape`). A registration hook for custom shape types like heightfields is a good candidate for when the backend contract goes public - noted as a follow-up.

src/framework/physics/physics-body.js
willeastcott: Agreed that would be useful, but it is out of scope for this PR, which deliberately changes no behavior. The seam for it is now in place though: `getTransform` is defined as "the simulated pose to present to the scene", so a backend can implement whatever interpolation policy it wants once the contract is public and configurable.

src/framework/components/collision/trigger.js
willeastcott: This matches the pre-PR code (separate `getShapePosition`/`getShapeRotation` reads) and runs once per trigger per frame, so I kept it verbatim to stay behavior-identical. A combined offset-aware read is a reasonable micro-optimization for a follow-up.

src/framework/components/rigid-body/system.js
willeastcott: That is exactly what this seam enables: call `app.systems.rigidbody.setPhysicsWorld(new MyPhysicsWorld())` before `app.start()` and the Ammo auto-detect is skipped - it only runs when no world is installed. The `PhysicsWorld` base class defines the full contract (world lifecycle, bodies, shapes, joints, raycasts, contact reporting), so a pointer-based implementation is free to represent natives however it wants. The classes are internal for now while the contract bakes; exporting them so third-party backends can be built against the npm package is the explicit next step (see the PR description).

src/framework/components/rigid-body/component.js
willeastcott: Done in 0a876591b.

src/framework/components/rigid-body/component.js
willeastcott: Done in 0a876591b (applied to both branches).

willeastcott:
@AlexAPPi I've followed up on your feedback.
@Maksims Thanks for chiming in with your thoughts.
Let me know if there's anything still outstanding, guys.

Maksims:
> Practice shows that in production today only a few physics engines are still used — over recent years sustained demand has essentially remained for just Ammo and Jolt; the others have fallen out of use or are rarely encountered. Given that, and considering the complexity and amount of work required to create your own engine or to study a niche one in depth, I have serious doubts about whether it’s worth spending time and resources on this now.

Box3D is a serious contender to become a mainstream physics engine also. It has very small build size, it performs really well (better than Ammo), supports multi-threading, determinism and much more modern.

AlexAPPi:
> > Practice shows that in production today only a few physics engines are still used — over recent years sustained demand has essentially remained for just Ammo and Jolt; the others have fallen out of use or are rarely encountered. Given that, and considering the complexity and amount of work required to create your own engine or to study a niche one in depth, I have serious doubts about whether it’s worth spending time and resources on this now.
> 
> Box3D is a serious contender to become a mainstream physics engine also. It has very small build size, it performs really well (better than Ammo), supports multi-threading, determinism and much more modern.

Thanks for the info, definitely check out this engine! I agree, it looks quite promising.

PR #9049, opened 2026-07-07
Rebuild mesh collision shapes when the entity world scale changes

### LINKED ISSUE #3062: Multiple Mesh Colliders of Different Scales don't behave properly
When using differently scaled mesh colliders of the same mesh, only 1 instance of the collider will be used on all entities.

As LeXXik put it (forum username):
"Placing a mesh into the physics world is a computationally expensive operation - the engine generates an Ammo compound shape from a number of bvh triangle meshes, one for each mesh instance. Once generated the trimeshes then get cached for future re-use. As a result, when you use the same mesh for another collider, the trimeshes are taken from the cache, instead of generating new ones."

Provide as much information as possible. Include (where applicable):
[The forum discussion](https://forum.playcanvas.com/t/mesh-collider-broken/19187)

Example:

### Steps to Reproduce

1. Add a mesh and ammo to your project.
2. Create 2 or more entities of different scales.
3. Give them both the same mesh collider.

Result: all colliders have the same scale.
Desired: all colliders have their own scale.

### LINKED ISSUE #6212: Mesh collider no longer scales in real time
### Description

I have been unable to update the mesh collider scale since around March of this year.

I have prepared a minimal project on this issue, please see
[https://playcanvas.com/project/1204845/overview/meshcolliderscaletest](https://playcanvas.com/project/1204845/overview/meshcolliderscaletest)

This project has 4 spheres in a box.
A tween changes the size of the box.
I want to smoothly scale the mesh collider to match the scale of that box.
If I can correctly scale the mesh collider, the movement of the spheres should change.

I suspect the reason is that the following code in realtimeScaling.js is not working for some reason.
`col.system.implementations.mesh.doRecreatePhysicalShape(col);`

Previously, in order to resize the mesh collider in real time, the above code could be called every frame to do the update.
However, one day it stopped working.
I understand that this is a very costly process, but I really need this.

I have been developing a project that uses this program for a long time and it did not work on previous projects either.

Any help would be much appreciated.
Thank you very much.

willeastcott:
Scaling an entity with a mesh collider at runtime never rebuilt the shape: the `updateMeshTransform` scale check was only reachable through `onTransformChanged`, whose last caller (the Designer livelink) was removed in 2015. This PR wires that check up for real: the collision system keeps a list of mesh components with a built shape, and the rigid body update loop rebuilds any whose entity world scale has drifted from the scale the shape was built with (compared with a relative tolerance, so rotation-induced float noise in the extracted scale doesn't churn shapes).

Two supporting fixes in the Ammo backend:

- Triangle data built with a bake scale no longer enters the per-mesh-id cache. The cache key ignores scale, so the first builder's baked scale leaked into every other user of the same mesh (#3062, render path) and made rebuilds return stale geometry — which is why the manual `doRecreatePhysicalShape` workaround stopped working after #6087 (#6212). Unit-scale sources still share cached data as before.
- Scaled triangle data is owned by the shape and destroyed with it, so continuously tweening a scale doesn't grow the cache.

Compound children are skipped by the watch (their rebuild path doesn't detach the old shape from the parent compound), and the legacy model-asset path is untouched.

Verified headlessly against real ammo.js with raycast probes: rescale 1→2 moves the hit surface accordingly (and only after the physics tick), same-mesh entities at scales 2 and 3 coexist, a 200-step scale tween leaves the cache at one entry, convex hulls rebuild too, and rotation alone doesn't recreate the shape. All checks fail on main as described in the issues.

Fixes #6212
Fixes #3062

### DIFF
--- a/src/framework/components/collision/system.js
+++ b/src/framework/components/collision/system.js
@@ -20,6 +20,9 @@ const p2 = new Vec3();
 const p3 = new Vec3();
 const quat = new Quat();
 const quat2 = new Quat();
+const worldScale = new Vec3();
+
+const SCALE_CHANGE_TOLERANCE = 1e-5;
 
 // Note that `shape` is deliberately absent from this list - it is runtime
 // state created and owned by the type implementation, not component data
@@ -87,8 +90,7 @@ const collisionImpls = {
         // the model comes from assets - no placeholder is created
         beforeInitialize() {},
         createPhysicalShape: createMeshShape,
-        recreatePhysicalShapes: recreateMeshShapes,
-        updateTransform: updateMeshTransform
+        recreatePhysicalShapes: recreateMeshShapes
     },
 
     compound: {
@@ -150,6 +152,11 @@ function destroyShape(system, component) {
 }
 
 function beforeRemove(system, entity, component) {
+    const watchIndex = system._meshComponents.indexOf(component);
+    if (watchIndex !== -1) {
+        system._meshComponents.splice(watchIndex, 1);
+    }
+
     if (component._shape) {
         if (component._compoundParent && !component._compoundParent.entity._destroying) {
             system._removeCompoundChild(component._compoundParent, component._shape);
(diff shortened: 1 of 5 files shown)


willeastcott:
Superseded by https://github.com/playcanvas/engine/pull/9311, which takes a different approach: instead of keeping scaled triangle data out of the cache, it uses Bullet's `btScaledBvhTriangleMeshShape` (newly bound in kripken/ammo.js#448) so every collider shares one BVH per mesh and carries its own scale, and it also fixes the model-asset path and the compound-child rebuild. The scale watch and its tests from this PR live on there.

PR #9065, opened 2026-07-16
Support for KTX2 Cubemaps

MotivaCG:
Fixes KTX2 cubemap loading by detecting six-face textures and transcoding every face at every mip level.

This creates a native cubemap texture instead of loading only the first face.

Validated with a 2048×2048 KTX2 cubemap in a WebGL2 application in PC/Android/iOS

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,7 +34,7 @@
         "rollup-plugin-dts": "6.4.1",
         "serve": "14.2.6",
         "sinon": "22.0.0",
-        "turbo": "^2.10.4",
+        "turbo": "2.10.4",
         "typedoc": "0.28.20",
         "typedoc-plugin-mdn-links": "5.1.1",
         "typedoc-plugin-merge-modules": "7.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,7 +34,7 @@
         "rollup-plugin-dts": "6.4.1",
         "serve": "14.2.6",
         "sinon": "22.0.0",
-        "turbo": "2.10.4",
+        "turbo": "^2.10.4",
         "typedoc": "0.28.20",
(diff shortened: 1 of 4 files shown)


mvaligursky (commented):

---

Looks correct and well-contained. I traced the full path: the cubemap changes live only in `transcodeKTX2` (the `.basis` path is untouched and still reports `cubemap: false`), the flattened transfer list correctly handles the nested face arrays for `postMessage`, `BasisQueue.handleResponse` is the only response path (no inline transcode to miss), the KTX2 face order matches the engine's +X/−X/+Y/−Y/+Z/−Z convention, and the ktx2 parser's `open()` already passes `data.cubemap` through to the `Texture` (with clamp addressing and the level-count-derived `mipmaps` flag), so no changes were needed there. Disabling GGGR unswizzle for cubemaps is also right — the unswizzle loop would otherwise crash on the per-mip face arrays. Two minor notes:

1. **Cubemap arrays are silently truncated to layer 0.** A KTX2 file with `faceCount === 6` and `layerCount > 1` transcodes only the first cubemap (every call passes `layer = 0`). That mirrors the existing 2D-array behavior, so it's not a regression — but since this PR introduces face iteration, a debug-build warning when `layerCount > 1` would make the limitation visible instead of silent.

2. **Validation was WebGL2-only** (per the description). The engine's compressed-cubemap upload path is exercised by the legacy DDS/KTX1 parsers so WebGPU is likely fine, but a quick WebGPU smoke test of a transcoded cubemap would close the gap — transcoded target formats (e.g. ETC/ASTC vs BC) differ per platform and this is the first compressed-cubemap producer on the basis path.

mvaligursky (approved):
Great fix, thanks!

Issue #5955, opened 2024-01-19
Incoherency among lerp functions


NewboO:
I noticed `math.lerp` is clamped but `VecX.lerp` and `Quat.slerp` aren't. I don't know if there's a way to fix this without breaking lots of projects but if you use easing functions like back or elastic, it gives unexpected results when tweening a single float compared to tweening a whole vector.

PR #9087, opened 2026-07-20
Add unclamped scalar interpolation

mvaligursky:
## Overview

Addresses #5955 by adding an unclamped scalar interpolation helper consistent with the extrapolating vector, color, and quaternion interpolation APIs.

- Adds `math.lerpUnclamped` with JSDoc examples.
- Covers interpolation and extrapolation behavior with unit tests.
- Clarifies that `math.lerp` clamps its interpolation factor.
- Documents the existing unclamped extrapolation behavior of `Quat.lerp` and `Quat.slerp`.

## Public API changes

Before, scalar extrapolation required an inline expression:

```javascript
const result = a + (b - a) * alpha;
```

It can now use the new helper:

```javascript
const result = math.lerpUnclamped(a, b, alpha);
```

The existing `math.lerp` behavior is unchanged and remains clamped.

## Performance

The helper is allocation-free and evaluates the interpolation expression directly.

## Testing

- Focused core math unit tests
- ESLint
- Type declaration generation and validation

### DIFF
--- a/src/core/math/math.js
+++ b/src/core/math/math.js
@@ -130,9 +130,7 @@ const math = {
      *
      * @param {number} a - Number to linearly interpolate from.
      * @param {number} b - Number to linearly interpolate to.
-     * @param {number} alpha - The value controlling the result of interpolation. When alpha is 0,
-     * a is returned. When alpha is 1, b is returned. Between 0 and 1, a linear interpolation
-     * between a and b is returned. alpha is clamped between 0 and 1.
+     * @param {number} alpha - The interpolation factor, clamped to the range 0 to 1.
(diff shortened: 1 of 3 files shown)


willeastcott:
Did you consider adding a final param to control clamping to avoid adding a new function to the API?
Won't this now mismatch against other APIs have have lerp but no lerpUnclamped?

mvaligursky:
> Did you consider adding a final param to control clamping to avoid adding a new function to the API? Won't this now mismatch against other APIs have have lerp but no lerpUnclamped?

I have, but preferred to avoid condition in such a basic operation.
also, some other engines also have two variants

PR #9090, opened 2026-07-21
feat: add instanced wide line renderer

mvaligursky:
## Summary

- Add extras WideLine and WideLineRenderer APIs with per-point positions, RGB colors, widths, caps, joins, closure, and dash styling.
- Batch all owned segments in one instanced draw call per camera/layer pass, with automatic capacity growth.
- Add screen-space and world-space width shader variants, depth/layer controls, and opaque rendering.
- Add wide-line, wide-lines-styles, and wide-lines-dynamic examples with controls and thumbnails.

Screenshot 2026-07-21 at 10 38 01


Maksims:
Awesome.
One of significant use cases we have with one project, where we had to implement wide lines, is rendering dashed line in form of arrowed rather than 90 degrees cut.
And moving them along the path, with shader.
That way it looks like flowing arrows.

MAG-AdrianMeredith:
Nice, my first ever rendering engine was a webgl version of canvas 2d.  Spent so long getting lines right!  Look forward to trying this.  (dont currently have much use case but I'm sure I can find one)

mvaligursky:
> Nice, my first ever rendering engine was a webgl version of canvas 2d. Spent so long getting lines right! Look forward to trying this. (dont currently have much use case but I'm sure I can find one)

I look forward to some testing. I had 3 examples, but I'm sure it's not perfect yet.

AlexAPPi:
> > Nice, my first ever rendering engine was a webgl version of canvas 2d. Spent so long getting lines right! Look forward to trying this. (dont currently have much use case but I'm sure I can find one)
> 
> I look forward to some testing. I had 3 examples, but I'm sure it's not perfect yet.

I just started working on a spline system, and here is such an implementation, thanks!

PR #9105, opened 2026-07-23
Migrate internal render targets to origin option, complete flipY deprecation

mvaligursky:
Follow-up to #9101. Migrates all engine-internal render targets off the deprecated `flipY` flag to the `origin` option, completes the deprecation, and reworks CameraFrame and the gsplat projector so no engine code depends on `flipY` any more.

**Changes:**
- Migrated internal render targets that replicate the WebGL storage layout on all APIs to `origin: RENDERTARGET_ORIGIN_BOTTOM`: shadow maps (removing the long-standing temporary `flipY` workaround and its TODO), the clustered cookie atlas, texture reprojection (`reprojectTexture`), the outline renderer, and the gsplat processor and work buffers.
- **CameraFrame** no longer inherits the output target's `flipY` into its intermediate scene render target. The scene chain always renders in the API-native orientation, and the compose pass flips its sampling into the requested orientation via a new `composeTargetFlipY` uniform. As a result the scene, TAA, and grab passes always take the common code path regardless of the final target's orientation.
- **Unified gsplat renderer**: the projector now builds a canonical (unflipped) projection and caches canonical clip positions; the hybrid raster vertex shader applies the per-pass target orientation to the final vertex position using the existing `projectionFlipY` uniform. This keeps the projection cache valid for any raster target (forward, prepass, pick) without the projector needing to know the target, and fixes picking into a flipped render target. `projectionFlipY` is now refreshed on all render passes (including XR).
- Completed the deprecation: the constructor `flipY` option now emits a runtime deprecation warning (previously only the `flipY` property setter did). No engine code passes `flipY` any more, so debug builds are free of internal deprecation warnings.

**API Changes:**
- Added: public `RenderTarget#origin` getter, resolved at construction from the `origin` option and kept consistent when the deprecated `flipY` option or property is used.
- Deprecated (runtime warning): the constructor `flipY` option, matching the already-deprecated `RenderTarget#flipY` property.

**Examples:**
- Updated the hidden `test/render-target-orientation` example to construct its bottom-origin render target via `origin` instead of `flipY`.

### DIFF
--- a/src/platform/graphics/render-target.js
+++ b/src/platform/graphics/render-target.js
@@ -121,6 +121,12 @@ class RenderTarget {
     _flipY;
 
     /**
+     * @type {string}
+     * @private
+     */
+    _origin;
+
+    /**
      * Creates a new RenderTarget instance. A color buffer or a depth buffer must be set.
      *
      * @param {object} [options] - Object for passing optional arguments.
@@ -315,13 +321,21 @@ class RenderTarget {
         });
         if (options.origin === RENDERTARGET_ORIGIN_TOP) {
             this._flipY = !device.isWebGPU;
+            this._origin = RENDERTARGET_ORIGIN_TOP;
         } else if (options.origin === RENDERTARGET_ORIGIN_BOTTOM) {
             this._flipY = device.isWebGPU;
+            this._origin = RENDERTARGET_ORIGIN_BOTTOM;
         } else if (options.origin === RENDERTARGET_ORIGIN_NATIVE) {
             this._flipY = false;
+            this._origin = RENDERTARGET_ORIGIN_NATIVE;
         } else {
-            // origin not specified - honor the deprecated flipY option
+            // origin not specified - honor the deprecated flipY option, and derive the
+            // equivalent origin from it
+            if (options.flipY !== undefined) {
+                Debug.deprecated('RenderTarget "flipY" option is deprecated, use the "origin" option instead. Typical migration: flipY: !device.isWebGPU -> origin: RENDERTARGET_ORIGIN_TOP, flipY: device.isWebGPU -> origin: RENDERTARGET_ORIGIN_BOTTOM.');
+            }
             this._flipY = options.flipY ?? false;
+            this._origin = this._flipY ? (device.isWebGPU ? RENDERTARGET_ORIGIN_BOTTOM : RENDERTARGET_ORIGIN_TOP) : RENDERTARGET_ORIGIN_NATIVE;
         }
 
         this._mipLevel = options.mipLevel ?? 0;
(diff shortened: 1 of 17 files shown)


src/platform/graphics/render-target.js
```diff
@@ -315,13 +321,21 @@ class RenderTarget {
...
             this._flipY = false;
+            this._origin = RENDERTARGET_ORIGIN_NATIVE;
         } else {
-            // origin not specified - honor the deprecated flipY option
+            // origin not specified - honor the deprecated flipY option, and derive the
+            // equivalent origin from it
+            if (options.flipY !== undefined) {
+                Debug.deprecated('RenderTarget "flipY" option is deprecated, use the "origin" option instead. Typical migration: flipY: !device.isWebGPU -> origin: RENDERTARGET_ORIGIN_TOP, flipY: device.isWebGPU -> origin: RENDERTARGET_ORIGIN_BOTTOM.');
+            }
             this._flipY = options.flipY ?? false;
```
mvaligursky, line 337: Two silent fall-throughs in the `origin` resolution worth hardening while this is being finalized:

1. **An unrecognized `origin` string ends up here silently.** A typo like `origin: 'bottom '` or `'Bottom'` doesn't match any branch, falls into the deprecated-flipY path, and resolves to NATIVE with no warning — an orientation bug that's hard to trace back to the typo. A debug assert would catch it:
```js
Debug.assert(options.origin === undefined,
    `RenderTarget: unknown origin '${options.origin}'`);
```
2. **`flipY` is silently ignored when `origin` is also specified** — someone migrating incrementally who passes both gets no deprecation warning (the warning only fires in this else branch) and no hint that `flipY` lost. Moving the `options.flipY !== undefined` deprecation check above the branch chain would cover both cases.

Both are debug-build-only additions; behavior for valid inputs is unchanged.

mvaligursky (commented):

---

Reviewed the full diff plus the surrounding code at the PR head. The core mechanisms check out: every `flipY: true` → `ORIGIN_BOTTOM` conversion is behavior-preserving per API (I enumerated all `rt.flipY` consumers — the WebGL-side value change on the quad-driven targets like the cookie atlas and gsplat work buffers has no WebGL consumer, and the camera-driven consumers — projection flip, `flipFaces`, the shadow-renderer `flipFactor` — see identical values on both APIs); the compose flip is applied consistently to every scene-texture-sampling effect (DOF/SSAO/bloom/fringing all use the flipped uv, and vignette is flip-symmetric so its uv choice is inert); and the canonical projector cache pairs consistently with the canonical `_computeClipToViewZ` inverse, with the flip applied to the output position only. One small API-hardening suggestion inline.

test/platform/graphics/render-target.test.mjs
```diff
@@ -91,4 +91,35 @@ describe('RenderTarget', function () {
...
+
+        it('is derived from the deprecated flipY setter', function () {
+            const rt = createRenderTarget({ origin: RENDERTARGET_ORIGIN_BOTTOM });
+            rt.flipY = true;
+            expect(rt.origin).to.equal(RENDERTARGET_ORIGIN_TOP);
+            rt.flipY = false;
+            expect(rt.origin).to.equal(RENDERTARGET_ORIGIN_NATIVE);
+            destroyRenderTarget(rt);
+        });
+    });
```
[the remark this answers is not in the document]
mvaligursky: Addressed in 54f583732 — added a WebGPU-path describe block that stubs `isWebGPU` on the null device and asserts the mirrored resolution: origin top does not flip, origin bottom flips, and the deprecated flipY option/property derives origin bottom (not top).

src/platform/graphics/render-target.js
mvaligursky: Both are already covered by the `Debug.call` validation block just above the resolution chain: an unrecognized origin trips the assert at line 316 (listing the valid values), and passing both origin and flipY emits the warnOnce at line 318 ("origin takes precedence"), which is the hint that flipY was dropped. Leaving the guards as-is rather than adding redundant checks.

PR #9111, opened 2026-07-24
Remove specularTint and avoid unnecessary color shader invalidation

mvaligursky:
The StandardMaterial color properties now update uniforms without invalidating shader variants just by being read. Specular shader structure changes remain tracked explicitly when the specular color crosses the black/non-black boundary.

**Changes:**
- Remove the obsolete StandardMaterial#specularTint property and its parser/type-fixup references.
- Preserve the removed-property compatibility accessor for legacy callers.
- Always apply the specular color for active specular-color shading while preserving reflection-only behavior.
- Add focused tests for color mutation, gamma conversion, shader invalidation, and specular shader defines.

**API Changes:**
- Remove StandardMaterial#specularTint. Legacy access reports the property as removed and behaves as if it were always enabled.

### DIFF
--- a/src/scene/materials/standard-material.js
+++ b/src/scene/materials/standard-material.js
@@ -658,6 +658,9 @@ class StandardMaterial extends Material {
         const specularIsBlack = isBlack(this._specular);
         if (this._specularIsBlack !== specularIsBlack) {
             this._specularIsBlack = specularIsBlack;
+            // This is intentionally conservative: another material property might already force
+            // specular shading, in which case this transition only changes a uniform. We can avoid
+            // that redundant variant clear later by tracking the derived useSpecular state here.
             this._dirtyShader = true;
         }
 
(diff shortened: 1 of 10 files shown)


src/scene/materials/standard-material.js
```diff
@@ -1083,19 +1091,12 @@ function _defineColor(name, defaultValue) {
...
         defaultValue: defaultValue,
-        getterFunc: function () {
-            // HACK: since we can't detect whether a user is going to set a color property
-            // after calling this getter (i.e doing material.ambient.r = 0.5) we must assume
-            // the worst and flag the shader as dirty.
-            // This means currently animating a material color is horribly slow.
-            this._dirtyShader = true;
-            return this[`_${name}`];
-        }
+        dirtyShaderFunc: () => false
```
mvaligursky, line 1097: This is the big win here. Changing a color property no longed forces shader options evaluation.

Issue #7451, opened 2025-03-19
Engine v2 fog and clear color differences.


Maksims:
With engine v2, colors are converted to linear workflow. So the fog color is not used as provided, and set value as fog color will look different to lets camera clear color, which is not converted to linear.

Should the fog color appear as it is set?

Here the camera clear and fog colors are set to the same value, but they look vastly different.

![Image](https://github.com/user-attachments/assets/3578885c-6134-43ef-85c3-36dafaf6d64d)

mvaligursky:
this is related:
https://github.com/playcanvas/engine/issues/7181

mvaligursky:
So the solution here would be to convert the clear color to linear space in the simplest case.

Maksims:
> So the solution here would be to convert the clear color to linear space in the simplest case.

That would probably work. Would be great to have Editor pickers that would render colors in both non and linear spaces, so we can visually know what color will look like. As currently we choose one color, but everything end up looking differently.

Also, it is important to have 0,0,0,x and 1,1,1,x colors not to be tonemapped, as such clear colors are often optimized by GPUs as a fast-clear.

mvaligursky:
> Also, it is important to have 0,0,0,x and 1,1,1,x colors not to be tonemapped, as such clear colors are often optimized by GPUs as a fast-clear.

we do not tonemap clear colors, so no problem here.

agreed on picker, in this case it'd be handy.

Maksims:
Also noticed, when using CameraFrame script for post-effects, the clear color is then tonemapped, so when CameraFrame is enabled, the sky looks vastly different.

mvaligursky:
> Also noticed, when using CameraFrame script for post-effects, the clear color is then tonemapped, so when CameraFrame is enabled, the sky looks vastly different.

Yep, this is unavoidable, as tonemapping is postprocessing there and the whole screen is processed.

mvaligursky:
Closing as a duplicate of #7181 — same root cause (the clear color is a raw framebuffer value while scene colors like fog are linearized, exposed and tone mapped). Full analysis and the V3 plan, including the CameraFrame difference reported here and the editor color-picker suggestion, is written up in https://github.com/playcanvas/engine/issues/7181#issuecomment-5069650213 with a reference implementation in #9113.

Issue #7181, opened 2024-12-07
Fog + Tonemapping


willeastcott:
Often, you want to fade your scene via a fog color so that it matches the clear color of the camera.

This works fine if you have tonemapping disabled. But let's say you enable linear tonemapping and have `0.1, 0.1, 0.1` clear color set, you need to set the fog color to roughly `0.21, 0.21, 0.21` for the scene to match up with the camera's clear color.

Is this because we need to be applying fog after the tonemap? Or should we be tonemapping the fog color?

mvaligursky:
I think the problem is that the clear color is not tonemapped. A workaround here would be to render an inverted sphere with a color as this color would be tonemapped. Not surer convenient though.
This should not be an issue if post-effects / CameraFrame rendering is used, as the clear color is used to clear the color and the tone mapping is applied over it as well.

mvaligursky:
Investigated this in depth, together with #7451 (closed as duplicate — same root cause; its CameraFrame observation is covered below).

**Root cause** — `camera.clearColor` is a raw framebuffer value, while every rendered color (including `fog.color`) is a scene color. Fog is linearized on the CPU, blended in linear HDR space, then exposure + tone mapping + gamma encoding are applied in the shader. The clear color skips all of that. A fully fogged pixel shows `gamma(tonemap(exposure · linear(c)))` while the background shows `c` — they match only with `TONEMAP_NONE`, which is exactly what this issue reports. With `CameraFrame` there is an additional facet (reported in #7451): the raw clear color is written **un-linearized** into the linear HDR scene buffer and then tone mapped by the compose pass, so the background matches neither the authored color nor fog, and the same scene renders a different background with vs without post-processing.

**Proposed solution (prototyped and verified in #9113)** — a camera flag `clearColorTonemapped` that treats the clear color as a scene color: convert to linear, apply exposure + tone mapping (a small CPU mirror of the tonemap shader chunks), and encode to the render pass's output gamma. The transform respects the pass's gamma/tonemap overrides, so under `CameraFrame` (scene pass = `GAMMA_NONE`/`TONEMAP_NONE`) it reduces to linearizing the color into the HDR buffer, where the compose pass tone maps it like everything else. Verified: fogged geometry vs background matches with **delta 0 for all 7 tonemappers × WebGL2/WebGPU × with and without CameraFrame**, and CameraFrame vs no-CameraFrame backgrounds agree within 1/255.

[...]

PR #9113, opened 2026-07-24
Add camera.clearColorTonemapped to match clear color with scene colors

mvaligursky:
> [!NOTE]
> **Reference PR — closed intentionally, deferred to engine V3.** See the investigation in #7181. Shipping this as an opt-in flag on v2 was rejected because the off-state is inherently inconsistent under `CameraFrame` (see below); for V3 this behavior should become the default (`true`), which is a breaking visual change.

Camera clear color is a raw framebuffer value while every rendered color (including `fog.color`) is a scene color — linearized, exposed, tone mapped and gamma encoded. As a result the background matches fog only when tone mapping is disabled (#7181), and with `CameraFrame` the raw clear color lands un-linearized in the HDR scene buffer and gets tone mapped by the compose pass, so it matches neither the authored value nor fog (#7451).

This PR adds an opt-in `clearColorTonemapped` flag (default `false`, non-breaking) that treats the clear color as a scene color: converted to linear space, processed by the camera's exposure and tone mapping, and encoded to the render pass's output gamma — the same pipeline shaded pixels go through.

**Changes:**
- `CameraComponent#clearColorTonemapped` / `Camera#clearColorTonemapped` — opt-in flag, default `false`
- `src/scene/tonemapping.js` — `toneMapColor()`, a CPU mirror of all 7 tone mapping shader chunks (`TONEMAP_NONE` applies neither tone mapping nor exposure, matching the shaders)
- `Camera#getRenderPassClearColor()` — computes the clear value for a pass; respects the pass's gamma / tone mapping overrides, so under `CameraFrame` (scene pass renders with `GAMMA_NONE`/`TONEMAP_NONE`) it reduces to linearizing the color into the HDR buffer, and the compose pass tone maps it like everything else
- Applied in `RenderPassForward.updateClears` and in `Renderer#clear` (partial-viewport clears)
- `test-clear-color/index.html` — verification harness (fully-fogged geometry vs cleared background pixel comparison)

**Verification:**
- Flag on: fogged geometry and background match with delta 0 for all 7 tonemappers × WebGL2/WebGPU × with and without `CameraFrame` (exposure 1.5); `CameraFrame` vs no-`CameraFrame` backgrounds agree within 1/255
- Flag off: reproduces the reported mismatches (delta up to 53/255 forward, 119/255 with `CameraFrame`)
- `TONEMAP_NONE` + flag on yields the exact authored color (no behavior change without tone mapping)

**Known trade-offs:**
- The `false` state cannot be made consistent under `CameraFrame`: showing the authored color unprocessed through an HDR compose pass would require inverting the tonemap curves, which are non-invertible. Hence the V3 plan of defaulting to `true` instead of shipping the flag on v2.
- Tone mapped white is not exactly 1.0 (e.g. ACES), so pure-white clears lose the driver fast-clear optimization when the flag is enabled.

Relates to #7181, #7451.

### DIFF
--- a/src/framework/components/camera/component.js
+++ b/src/framework/components/camera/component.js
@@ -502,6 +502,32 @@ class CameraComponent extends Component {
     }
 
     /**
+     * Sets whether the clear color is tone mapped. When false, the clear color is written to the
+     * render target exactly as specified, so it is not affected by the camera's exposure, tone
+     * mapping or gamma correction. When true, the clear color is treated as a scene color: it is
+     * converted to linear space and processed by the camera's exposure, tone mapping and gamma
+     * correction, the same way shaded pixels are. Enable this to make the cleared background
+     * match other scene colors - most notably {@link FogParams#color}, which otherwise only
+     * matches the clear color when tone mapping is disabled. This also makes the background
+     * appear identical whether or not HDR post-processing (e.g. `CameraFrame`) is used. Defaults
+     * to false.
+     *
+     * @type {boolean}
+     */
+    set clearColorTonemapped(value) {
+        this._camera.clearColorTonemapped = value;
+    }
+
+    /**
+     * Gets whether the clear color is tone mapped.
+     *
+     * @type {boolean}
+     */
+    get clearColorTonemapped() {
+        return this._camera.clearColorTonemapped;
+    }
+
+    /**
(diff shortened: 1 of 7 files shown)


mvaligursky:
Closing intentionally — deferred to engine V3 where this behavior should become the default. Full investigation and plan: #7181.

PR #9114, opened 2026-07-24
Move StandardMaterial transform grouping to update

mvaligursky:
Move StandardMaterial texture-transform grouping out of shader option generation so uniform-only transform changes do not repeatedly evaluate or invalidate shader variants.

**Changes:**
- Cache texture transform group IDs when StandardMaterial is updated, preserving exact transform deduplication and mutable Vec2 behavior.
- Make shader option generation consume the cached transform IDs.
- Track transform property changes so unchanged materials skip grouping work.
- Add focused coverage for shared, distinct, animated, identity, and implicitly prepared transforms.

**Examples:**
- Add a hidden Material Texture Transforms example for visually checking individual texture channels, combined output, grouping changes, animation, and shader allocations.

**Performance:**
- Avoid transform generation and comparison during shader option evaluation.
- Avoid rebuilding shader variants when transform values animate without changing their grouping.
- Skip transform scans when relevant material properties have not been accessed or modified.


Maksims:
In Editor, project got updated to 2.22, and now it does not run and we have exceptions in this code:

image

image

After inspecting data, _matTex2D is assumed to be an array or object? As it has keys like an object, but it is instantiated as an array. This breaks the loop.

{B36A0332-58AE-4EBA-BD3F-AF019B674F77}

mvaligursky:
@Maksims thanks for the report. I can reproduce it, and the array itself isn't the problem: `for...in` over `_matTex2D` has been used in the engine for years. What's new in 2.22 is that `StandardMaterialMapTransforms.update()` reads `material._${p}MapTiling` for every key the loop yields, and `for...in` also yields **inherited** enumerable keys. So if anything in the project adds an enumerable property to `Array.prototype` (e.g. `Array.prototype.remove = function () {...}`), 2.21 skipped that key silently while 2.22 throws exactly this error on that line. `Object.keys` in your screenshot only lists own keys (25), which is why it looked fine there.

Could you run this in the launch page console and paste the output?

```js
Object.keys(Array.prototype)
```

It should be `[]`. Anything listed is the polluter; my guess from your Sources tab is `graphene.min.js`. Regardless, that loop should iterate own keys only so the engine stops depending on a clean prototype.

Maksims:
We do extend Array.prototype with additional methods for convenience.
And I believe engine should not rely that nobody would not do that.

mvaligursky:
Nobody is suggesting you cannot extend it .. this is purely a question to confirm the cause.
As mentioned above, this is years preexisting issue in the engine, suddenly exposed, and I have a fix already.

Issue #7480, opened 2025-03-25
Shadow catcher exhibits circular artifacts


willeastcott:
Check out this Web Components example:

https://playcanvas.github.io/web-components/examples/#animation.html

You should see this:

https://github.com/user-attachments/assets/692b6fa5-c0b0-4db8-8bfe-8abe76475659

The light is configured thusly:

```html
<pc-light cast-shadows shadow-distance="30" shadow-intensity="0.6" shadow-resolution="1024" shadow-type="vsm-16f"></pc-light>
```

willeastcott:
I believe this could be related to this code:

https://github.com/playcanvas/engine/blob/main/src/scene/renderer/shadow-renderer-directional.js#L153-L158

Needs further investigation.

mvaligursky:
Investigated this. There are actually **two separate artifacts** with different causes, and only one of them is fixed so far.

### 1. Flickering square artifacts - fixed by #9124

Animated meshes (skinning / morphing) occasionally rasterize a degenerate triangle in light space, producing a fragment with depth outside of the `[0, 1]` range (measured `z` of 2 to 3.4) or NaN. The VSM exponential warp turns those into enormous values - up to 1e31 - and the separable VSM blur then spreads a single such texel over a `vsmBlurSize`² area, which the receiver reads as an occluder behind it, so the whole block turns lit. Hence hard-edged squares that flicker with the animation.

This is strictly animation driven: 34 of 1248 frames contained such texels while the animation played, and 0 of 1227 with it paused. #9124 discards those fragments in the shadow caster, which makes the invalid values impossible by construction.

### 2. Rounded / circular light leak - NOT fixed, to be improved later

The circular artifacts in the title are a different, inherent property of the current EVSM implementation, and #9124 does not change them. Because `exp()` grows so steeply, a linear blur of the warped moments behaves as a soft *maximum*, so the deepest texel within the filter kernel dominates the filtered occluder depth. Where that ends up behind the receiver, the one-tailed Chebyshev early-out returns fully lit. It is therefore worst where caster and receiver depths are close, such as around a character's feet, and never higher up.

Evidence that this is the mechanism, all measured at a fixed animation pose:
- `vsmBias` makes no difference at all (identical artifact area for every value from 0.0005 to 0.05), confirming the hard `mean <= moments.x` early-out is deciding, not the variance / Chebyshev path.
[...]

PR #9125, opened 2026-07-27
Reference (do not merge): EVSM4 for VSM_32F

mvaligursky:
> [!NOTE]
> **Reference only - closed immediately, not intended to be merged.** Opened to preserve an EVSM4 implementation and its measurements for a future improvement of #7480. The narrow, mergeable part of this exploration was split out into #9124.

`SHADOW_VSM_32F` currently uses a single (positive) exponential warp - EVSM2. Because `exp()` grows so steeply, a linear blur of those moments behaves as a soft maximum, so the deepest texel within the filter kernel dominates the filtered result. When that filtered occluder depth ends up behind the receiver, the one-tailed Chebyshev early-out returns fully lit, which shows up as a rounded light leak inside the shadow - most visible where caster and receiver depths are close, such as around the feet of a character.

**What this branch contains:**
- `SHADOW_VSM_32F` reimplemented as EVSM4 - the map stores a positive and a negative exponential warp of the depth, each with its second moment, using `kPos = 15` (unchanged) and `kNeg = 5`. A Chebyshev bound is evaluated per warp and the tighter of the two is used. The negative warp is a soft minimum, so it does not suffer from the domination above and prevents the false "lit" result. The public `SHADOW_VSM_32F` constant is unchanged, so user code is unaffected. `SHADOW_VSM_16F` stays on EVSM2.
- Because EVSM4 uses all four channels, the coverage channel is gone, so the VSM_32F map is cleared to the warped depth of the far plane instead - an occluder behind any receiver, and so unshadowed.
- The VSM blur now filters all four channels.
- The caster depth guard (as in #9124, but here applied to all shadow types rather than only VSM), and temporary animation scrubbing / logging in the example, used to find reproduction poses.

**Measured** on `shaders/shader-material-shadows` at a fixed pose, comparing the positive-only bound against the minimum of both, against a PCF5 reference: light leak energy 4.5x lower, compact leak blobs 24 -> 6, blob area 12362 -> 1380 px, largest blob 143x210 -> 58x97 px, and slightly less over-darkening than VSM_16F.

**Why it is not proposed for merging as is:** VSM_32F still leaks more than VSM_16F, the clear-to-far change is more leak prone than the coverage approach when receivers are far from the shadow camera far plane, WebGPU was not re-verified after the last change, and the example still contains the debug scaffolding.

### DIFF
--- a/examples/src/examples/shaders/shader-material-shadows.controls.jsx
+++ b/examples/src/examples/shaders/shader-material-shadows.controls.jsx
@@ -1,4 +1,11 @@
-import { BindingTwoWay, LabelGroup, Panel, SelectInput } from '@playcanvas/pcui/react';
+import {
+    BindingTwoWay,
+    BooleanInput,
+    LabelGroup,
+    Panel,
+    SelectInput,
+    SliderInput
+} from '@playcanvas/pcui/react';
 
 import {
     SHADOW_PCF1_16F,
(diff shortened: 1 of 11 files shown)


mvaligursky:
Closing as reference only - this is preserved for a future EVSM4 based improvement of #7480. The mergeable part is #9124.

PR #9131, opened 2026-07-28
Support multiple transform feedback buffers and add a visual example

mvaligursky:
## Description

Follow-up to #9109, which added separate-buffer transform feedback so each captured varying can be written into its own buffer. This makes the feature usable from `TransformFeedback` and adds the visual example.

### Buffer roles

`TransformFeedback` now accepts an array of descriptors, each giving a buffer an explicit role:

```js
const tf = new TransformFeedback([
    { input: positions, output: positionsOut },  // read and written, swapped each step
    { input: constants },                        // read only, never modified
    { output: instances }                        // written only, for the render pass
]);
```

- **Read-only** suits per-item data which never changes. Without it, such data has to be copied through transform feedback every step and double-buffered for no reason.
- **Write-only** suits data only a later pass consumes — the instancing stream in the example is a good fit, since the simulation never reads it back.

Descriptors with an `output` are assigned transform feedback indices in the order they appear, skipping those without one, so the order matches the captured varyings. The existing `(inputBuffer)` and `(inputBuffer, outputBuffer)` forms are unchanged, as is the deprecated `usage` shim.

### Vertex array object cache

Swapping a pair exchanges the GPU buffers behind the `VertexBuffer` objects while leaving the objects themselves in place. A single input buffer caches its vertex array object on the buffer, so the swap carries it along — but a multi-buffer draw takes its vertex array object from `_vaoMap`, which is keyed on the buffer objects and their formats. Neither changes when the GPU buffers move underneath, so the cached vertex array object would keep reading the pre-swap buffers.

`GraphicsDevice#removeVertexArrayFromCache` drops the entry after each swap so the next step rebuilds it. It is only called from `TransformFeedback#process`, and only when more than one input buffer is in play — nothing on the normal draw path changes.

### validateAttributes

`validateAttributes` took exactly two vertex formats:

```js
validateAttributes(shader, vb0Format, vb1Format)
```

With three or more vertex streams the rest were invisible to it, so it reported a false `Vertex attribute [...] at location N required by the shader is not present in the currently assigned vertex buffers` on every frame. Previously unreachable, since nothing in the engine bound that many. Now takes the buffer list; both the WebGL and WebGPU call sites updated.

Also adds the missing `@category Graphics` to `TRANSFORM_FEEDBACK_INTERLEAVED` and `TRANSFORM_FEEDBACK_SEPARATE` — they were the only annotated constants in `graphics/constants.js` without it.

### Example

`graphics/transform-feedback-separate` simulates a flock entirely on the GPU in one pass, exercising all three roles:

| Buffer | Role | Contents |
|---|---|---|
| positions | read + write | `vec4(pos.xyz, 0)` |
| velocities | read + write | `vec4(vel.xyz, 0)` |
| constants | **read only** | seed, max speed, wander strength |
| instances | **write only** | `vec4(pos.xyz, yaw)` → instancing stream |

The render pass binds only the instancing buffer — the simulation buffers never enter it. Cone colour is derived from the heading, so the value stored in the instancing buffer is directly visible. Sits alongside the existing `graphics/transform-feedback`, which shows the single interleaved buffer case, making the contrast easy to see.

## Testing

Verified on WebGL2:

- Buffer counts confirm the roles at runtime — two buffers each for positions and velocities, **one** each for constants and instances. The read-only buffer is byte-identical across frames.
- The instancing buffer carries the current simulation position plus the heading.
- No console errors, `gl.getError() === 0` across rendered frames.
- Single-buffer regression check: `graphics/transform-feedback` still animates correctly, with particle lifetime decrementing exactly `0.083` over 5 frames at 1/60.
- Full unit suite passes (2119), engine and examples lint clean, examples prettier clean.

Transform feedback is WebGL2 only, so the example is flagged `WEBGPU_DISABLED`.

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/playcanvas/engine/blob/main/.github/CONTRIBUTING.md)


mvaligursky:
Hi @AlexAPPi - please check this does what you need, and report any issues if not, thanks!

PR #9145, opened 2026-07-31
Fix component event leak when removing an enabled component

### LINKED ISSUE #9144: Memory leak in Render Component
The rendering component will subscribe to layer changes in onEnable and unsubscribe in onDisable. If you remove an active render component from an entity, it will not unsubscribe causing a memory leak.

mvaligursky:
Fixes #9144.

`ComponentSystem#removeComponent` only fires `beforeremove` - it never disables the component. So anything a component subscribes to in `onEnable()` has to be torn down in `onBeforeRemove()` as well. Five components did not do this, leaking three subscriptions each:

| leaked subscription | owner | lifetime |
|---|---|---|
| `scene.on('set:layers')` | `Scene` | app |
| `layers.on('add')` | `LayerComposition` | app |
| `layers.on('remove')` | `LayerComposition` | app |

Each holds `scope = component`, and `component.entity` then keeps the entity and its subtree alive for the life of the app. The leaked `set:layers` handler also kept re-subscribing the dead component to every `LayerComposition` assigned from then on.

`entity.destroy()` was never affected, since `Entity#destroy` sets `enabled = false` on every component before removing them. Only the direct `entity.removeComponent(...)` / `system.removeComponent(entity)` path leaked.

**Changes:**
- `render`, `model`, `gsplat`, `sprite` and `element` components now undo their `onEnable` state in `onBeforeRemove`, following the existing `camera` / `light` precedent:
  ```js
  if (this.enabled && this.entity.enabled) {
      this.onDisable();
  }
  ```
  The guard mirrors the condition under which `onEnable` ran, so an already-disabled component is not torn down twice, and no `disable` / `state` events are fired. It runs first in the method, as sprite's and element's `onDisable` use state that the rest of `onBeforeRemove` destroys.
- `render`, `model` and `gsplat` now also unsubscribe from `removehierarchy` and `inserthierarchy`. Their constructors subscribe to four entity hierarchy events but only two were being unsubscribed.
- As a side effect of routing through `onDisable`, an entity with a `batchGroupId` is now removed from its batch group when its component is removed - previously the batch group kept the entity node after the component was gone.
- New `test/framework/components/component-teardown.test.mjs` covering all seven components with layer subscriptions across removal while enabled, removal while disabled, entity destruction, and layer composition reassignment after removal. `camera` and `light` act as the controls.
- New `test/framework/components/render/component.test.mjs` covering the hierarchy events, the batch group and layer removal.

`particlesystem` follows the same pattern and is already correct (it sets `this.enabled = false`), but cannot be covered by the tests because its `onEnable` early-outs on the null graphics device.

### DIFF
--- a/src/framework/components/element/component.js
+++ b/src/framework/components/element/component.js
@@ -2613,6 +2613,12 @@ class ElementComponent extends Component {
     }
 
     onBeforeRemove() {
+        // removing a component does not disable it first, so undo what onEnable set up. This runs
+        // before the image / text elements are destroyed below, as onDisable uses them.
+        if (this.enabled && this.entity.enabled) {
+            this.onDisable();
+        }
+
         this.entity.off('insert', this._onInsert, this);
         this._unpatch();
         if (this._image) {
--- a/src/framework/components/gsplat/component.js
+++ b/src/framework/components/gsplat/component.js
@@ -819,13 +819,20 @@ class GSplatComponent extends Component {
     }
 
     onBeforeRemove() {
+        // removing a component does not disable it first, so undo what onEnable set up
+        if (this.enabled && this.entity.enabled) {
+            this.onDisable();
+        }
+
         this.destroyInstance();
 
         this.asset = null;
         this._assetReference.id = null;
 
         this.entity.off('remove', this.onRemoveChild, this);
+        this.entity.off('removehierarchy', this.onRemoveChild, this);
         this.entity.off('insert', this.onInsertChild, this);
+        this.entity.off('inserthierarchy', this.onInsertChild, this);
     }
 
     onLayersChanged(oldComp, newComp) {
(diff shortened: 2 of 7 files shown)


mvaligursky (commented):

No blocking findings.

I traced the removal path through ComponentSystem, Entity hierarchy state changes, and each affected component lifecycle. The enabled and in-hierarchy guard matches the condition under which onEnable runs, and the teardown is ordered before sprite clips and element renderables are destroyed. The render, model, and gsplat changes also now balance all four hierarchy-event subscriptions. The cleanup remains allocation-free outside the removal path and introduces no public API change.

Validation: the 31 focused teardown and render-component tests pass locally, and all reported CI checks are green.

PR #9146 fixes the separate event-handle issue after LayerComposition replacement; that does not block this focused removal-path fix.

PR #9146, opened 2026-07-31
Fix layer event leak when the scene layer composition is replaced

mvaligursky:
Found while investigating #9144, but independent of it.

Every component with layer subscriptions re-subscribes to the new `LayerComposition` in its `onLayersChanged` handler, but did so with a bare `newComp.on(...)` and discarded the returned event handles. `onDisable` was therefore left holding `_evtLayerAdded` / `_evtLayerRemoved` handles for the *previous* composition and could no longer unsubscribe from the current one.

Measured on the `render` component, though all eight behaved the same way:

```
newComp[add] after scene.layers = newComp : 2
newComp[add] after component disabled     : 2   <-- onDisable cannot remove untracked subs
```

The subscriptions survived both disabling the component and destroying the entity, so the composition kept the component - and through it the entity - alive. `scene.layers` is reassigned by the engine itself in `AppBase#_parseApplicationProperties` when applying scene settings, as well as by anyone using the public `Scene#layers` setter.

**Changes:**
- `render`, `model`, `gsplat`, `sprite`, `element`, `camera`, `light` and `particlesystem` now track the re-subscription:
  ```js
  this._evtLayerAdded?.off();
  this._evtLayerAdded = newComp.on('add', this.onLayerAdded, this);
  ```
  Releasing the stored handle replaces the previous `oldComp.off(...)` calls, which did the same thing for the old composition only.
- `ElementComponent#onLayersChanged` no longer throws for group elements. It did `this._image ? this._image._renderable.model : this._text._model`, and a group element has neither, so any live group element threw whenever `scene.layers` was reassigned. It is now guarded the same way the `batchGroupId` setter is.
- New `test/framework/components/component-layers-changed.test.mjs` covering unsubscription after composition reassignment on disable, on entity destruction, and after repeated reassignment.

`particlesystem` is fixed alongside the others but cannot be covered by the tests, because its `onEnable` early-outs on the null graphics device.

### DIFF
--- a/src/framework/components/camera/component.js
+++ b/src/framework/components/camera/component.js
@@ -1198,10 +1198,12 @@ class CameraComponent extends Component {
      */
     onLayersChanged(oldComp, newComp) {
         this.addCameraToLayers();
-        oldComp.off('add', this.onLayerAdded, this);
-        oldComp.off('remove', this.onLayerRemoved, this);
-        newComp.on('add', this.onLayerAdded, this);
-        newComp.on('remove', this.onLayerRemoved, this);
+
+        // store the new handles, so that onDisable can unsubscribe from the current composition
+        this._evtLayerAdded?.off();
+        this._evtLayerAdded = newComp.on('add', this.onLayerAdded, this);
+        this._evtLayerRemoved?.off();
+        this._evtLayerRemoved = newComp.on('remove', this.onLayerRemoved, this);
     }
 
     /**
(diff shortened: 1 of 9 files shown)


mvaligursky (commented):

No blocking findings.

I checked the event-handle lifecycle across all eight affected component types, including repeated LayerComposition replacement, disable, entity destruction, and the particle-system early-return path. Releasing the retained handle before storing the new subscription correctly follows the active composition and uses the constant-time handle removal path. The ElementComponent guard also handles group elements without changing image or text behavior. There is no public API or backend-specific impact.

Validation: all 21 focused layer-composition tests pass locally, the two PRs merge cleanly at the source level, and all reported CI checks are green.

Issue #7514, opened 2025-04-02
Invalid shader is generated when refraction is enabled without metalness


mvaligursky:
Refraction (both cubemap and dynamic) can only be used with metalness. Otherwise we get shader compilation errors. Ideally this should work as well, or at least force useMetalness perhaps.

**Note that this issues exist and can be reproduce both in engine V2 and V1.**

Repros:
1. engine example Material-refraction, just comment out 'material.useMetalness = true;'
2. Editor. Create a new material, open Refraction section, and set Refraction value to non-zero.

<img width="1133" alt="Image" src="https://github.com/user-attachments/assets/ffc2ed97-3082-4e7b-8f90-e83d626b8c31" />

related to https://github.com/playcanvas/engine/pull/7513

PR #9147, opened 2026-07-31
Support refraction without the metalness workflow

mvaligursky:
Refraction (both cubemap and dynamic) could only be used together with `useMetalness`. Without it, the generated shader either failed to compile or read uninitialized shader globals, as reported in #7514. This makes refraction work in both workflows.

This is also the direction glTF specifies: `KHR_materials_transmission` defines transmission on the **dielectric** part of the material and notes that for `metallicFactor = 1.0` the transmission factor no longer matters, so refraction was never conceptually tied to the metalness toggle - it needs the IOR-driven dielectric fresnel, which the engine only happened to compute in the metalness path.

**Changes:**
- include and evaluate `ior` in the refraction front end when the metalness path does not, and upload `material_refractionIndex` to match
- gate specularity and gloss on refraction as well, since `addRefraction` consumes both
- include `refractionCubePS` based on the reflection source instead of the specular reflection path, which additionally requires `useSpecular`. This is what produced the `'addRefraction' : no matching overloaded function found` error in the issue
- declare `STD_SPECULAR_CONSTANT` for refractive materials, so the fresnel uses the material specular color rather than defaulting to white, which cancelled the refraction out entirely
- evaluate refraction outside of the lighting / reflections block, so it is no longer silently skipped in scenes with neither
- `needsNormal` now accounts for refraction

All of the above are applied to both the GLSL and WGSL chunks.

**Examples:**
- Material Refraction: added a Metalness toggle (on by default) to exercise both workflows

**Notes:**
Every shader variant touched here previously either failed to compile or read uninitialized values, so no working variant changes. Verified by hashing the generated fragment shaders against a build of the base commit: variants without refraction are byte identical, and the metalness refraction variants are identical modulo the whitespace of the moved `addRefraction` call.

A follow up worth doing separately is giving the `litArgs_*` globals default initializers, to rule out this class of uninitialized reads. It is kept out of this PR because it changes the generated source of every lit shader.

### DIFF
--- a/examples/src/examples/materials/material-refraction.controls.jsx
+++ b/examples/src/examples/materials/material-refraction.controls.jsx
@@ -20,6 +20,13 @@ export function Controls({ observer }) {
                         link={{ observer, path: 'data.dynamic' }}
                     />
                 </LabelGroup>
+                <LabelGroup text='Metalness'>
+                    <BooleanInput
+                        type='toggle'
+                        binding={new BindingTwoWay()}
+                        link={{ observer, path: 'data.metalness' }}
+                    />
+                </LabelGroup>
             </Panel>
         </>
     );
--- a/examples/src/examples/materials/material-refraction.example.mjs
+++ b/examples/src/examples/materials/material-refraction.example.mjs
@@ -133,7 +133,7 @@ material.metalness = 0.0; // low metalness, otherwise it's reflective
 material.gloss = 1.0;
 material.glossMap = assets.other.resource;
 material.glossMapChannel = 'g';
-material.useMetalness = true; // refractive materials are currently supported only with metalness
+material.useMetalness = true; // metalness workflow by default, can be toggled off in the UI
 material.refraction = 0.8;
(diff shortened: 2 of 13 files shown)


mvaligursky (commented):

No blocking findings.

I traced the full shader path from StandardMaterial option generation and uniform upload through fragment declarations, front-end argument initialization, shared lit defines, reflection-source chunk selection, and backend evaluation. The refraction-only path now initializes normal, IOR, gloss, and specularity consistently; cube refraction no longer depends on LIT_REFLECTIONS, and moving addRefraction outside the lighting/reflections block preserves its ordering relative to the existing area-light reset and AO processing. The GLSL and WGSL changes remain structurally matched, with no effect on non-refraction variants or public APIs.

Runtime validation on the deployed preview covered cube and dynamic refraction, metalness enabled and disabled, on both WebGL 2 and WebGPU. Every variant rendered correctly and produced no preview-origin warnings or shader compilation errors. All reported CI checks are also green.

PR #9149, opened 2026-07-31
Rebind anim component when render mesh instances change

### LINKED ISSUE #5225: Morph Animations do not load properly when Container is loaded at runtime
### Description

When loading assets for a 3D Model at runtime with an animation track that includes Morph Animation, the Morph animations seem to be lost. This is an issue, since you may not want to load any assets of a 3D model until needed.

But when using Preload on the container and template of the 3D Model, the morph animations do load. When then all asstes are unloaded and then loaded again an error is thrown when launching with debug mode with versions prior to 1.62 (the error vanishes in 1.62 but the problem doesn't): 
```

morph-instance.js:197 Uncaught TypeError: Cannot read properties of null (reading '_targets')
    at MorphInstance.setWeight (morph-instance.js:197:55)
    at Array.func (default-anim-binder.js:94:47)
    at AnimTarget.callSetters [as _set] (default-anim-binder.js:104:39)
    at AnimEvaluator.update (anim-evaluator.js:264:35)
    at AnimController.update (anim-controller.js:582:29)
    at AnimComponentLayer.update (component-layer.js:257:26)
    at AnimComponent.update (component.js:732:28)
    at AnimComponentSystem.onAnimationUpdate (system.js:84:31)
    at ComponentSystemRegistry.fire (event-handler.js:201:26)
    at Application.update (app-base.js:1171:22)
```

### Steps to Reproduce

Morph anim Lost after asset unload & load with Container and Template preloaded
1. Execute https://playcanv.as/b/ca15bdcc 
2. Press "Load" Button to load other assets of the 3D Avatar and wait for avatar to show
3. Press "Unload" Button
4. Press Load Button again and wait for avatar to show again.
5. Morph Animations on face (i.e. blinking and correct lips movements) are lost.

Morph anim never loaded when Container and Template are not preloaded
1. Execute https://playcanv.as/b/bcd88dc3
2. Press "Load" Button to load all assets of the 3D Avatar and wait for avatar to show
3. Notice how no moprh animations (no blinking or intricate lips movements are present)
4. Press "Unload" Button
5. Press "Load" Button again and wait for avatar to show again.
6. Morph Animations are still not present. 

Project for reference:
https://playcanvas.com/project/1055497/overview/morphtester

mvaligursky:
Fixes #5225 — morph animations are lost when a container asset is loaded at runtime rather than preloaded.

`DefaultAnimBinder`'s `weight` handler captures the morph instances **once**, when the anim controller resolves its curve paths, and returns `null` when there are none. Resolution is never retried, and nothing told the anim component that a render component's mesh instances had changed. That produces both symptoms in the issue:

- **Morph animation never plays.** The render asset is still downloading, so `render.meshInstances` is empty and the morph weight curves resolve to `null` and are dropped. When the asset arrives and mesh instances are created, nothing rebinds. Bone curves are unaffected because they resolve against graph nodes, which exist from the start — hence the reported "body animates, face frozen".
- **Crash after unload and reload.** `MorphInstance.destroy()` nulls its `morph`, but the binder still holds the destroyed instances, so the next update throws `Cannot read properties of null (reading '_targets')` in `MorphInstance.setWeight`. In release builds the `Debug.assert` is stripped, so weights are silently written into the dead instance instead — which is why the reporter saw the error disappear in 1.62 while the bug remained.

The binder's `materialTexture` handler captures a mesh instance the same way, so animated material textures were affected too.

**Why not the `model` component's approach.** `ModelComponent` calls `this.entity.anim.rebind()` directly when its model is assigned, and that is complete for a model: the Model owns every mesh instance on a private graph node hierarchy parented under its own entity, so an anim component on that entity is the only one that can bind them. A render hierarchy is a public entity hierarchy and any descendant can carry its own render component. The layout `instantiateRenderEntity` produces for an imported character puts the anim component on the root and render components on mesh descendants, so a direct co-located lookup never reaches the owning component.

**Changes:**
- `RenderComponent` announces mesh instance changes as a `meshInstancesChange` event on the component system registry, from `destroyMeshInstances()` and from the `meshInstances` setter.
- `AnimComponentSystem` subscribes once and rebinds every component whose animated hierarchy contains the changed entity. The subscription lives on the system, not on each component, so there is a single listener removed by the existing system `destroy()`, and disabled components are still rebound — they keep their bindings and are not rebound when re-enabled.
- `AnimComponent#animatesEntity()` performs the containment test against the root bone when one is assigned and the component's entity otherwise, matching what the binder actually walks.
- The destroy is split into a private `_destroyMeshInstances()` so replacing the mesh instances notifies once, rather than also notifying against the intermediate empty list.
- The notification is skipped while the entity is being destroyed, and tolerates the systems registry already being torn down — `AppBase#destroy` nulls it before unloading assets, which still drives render component asset callbacks.

**Tests:** five regression cases — mesh instances created after binding, mesh instances replaced, both again with the anim component on a parent and the render component on a child, a root bone hierarchy, plus an assertion that an unrelated hierarchy is not rebound.

Note that a render asset load performs two rebinds, because `_onRenderAssetLoad()` destroys and then `_cloneMeshes()` assigns. That is deliberate: the destroy must notify, otherwise a component whose `render.meshes` is not yet populated keeps stale bindings to destroyed morph instances between the destroy and the later `set:meshes` event. The extra rebind is a bounded one-off next to parsing the asset.

**API Changes:**
- `AnimComponent#animatesEntity(entity)` — new, `@ignore`. `_destroyMeshInstances` and `_onMeshInstancesChanged` on `RenderComponent` are private.
- `meshInstancesChange` is a new event on the component system registry, alongside the existing `update` / `animationUpdate` / `postUpdate`.

`ModelComponent` keeps its direct co-located rebind and is not moved onto the new event; worth a follow-up if the same non-co-located layout matters for the legacy component.

### DIFF
--- a/src/framework/components/render/component.js
+++ b/src/framework/components/render/component.js
@@ -330,7 +330,7 @@ class RenderComponent extends Component {
     set meshInstances(value) {
         Debug.assert(Array.isArray(value), 'MeshInstances set to a Render component must be an array.');
 
-        // don't notify the anim component yet, it's notified once the new mesh instances are in place
+        // don't notify yet, the change is announced once the new mesh instances are in place
         this._destroyMeshInstances();
 
         this._meshInstances = value;
@@ -357,7 +357,7 @@ class RenderComponent extends Component {
             }
         }
 
-        this._rebindAnim();
+        this._onMeshInstancesChanged();
     }
 
     /**
@@ -767,11 +767,11 @@ class RenderComponent extends Component {
     /** @private */
     destroyMeshInstances() {
         this._destroyMeshInstances();
-        this._rebindAnim();
+        this._onMeshInstancesChanged();
     }
 
     /**
-     * Destroys the mesh instances without notifying an anim component. Used by the meshInstances
+     * Destroys the mesh instances without firing a change notification. Used by the meshInstances
      * setter, which notifies once the replacement mesh instances are in place.
      *
      * @private
@@ -792,17 +792,21 @@ class RenderComponent extends Component {
     }
 
     /**
-     * Notifies an anim component on the same entity that the set of mesh instances has changed, so
-     * that it re-resolves the animation targets referencing them - morph target weights and
-     * animated material textures. Without this, morph animations are silently dropped when the
-     * render asset loads after the animation was bound, and keep driving destroyed morph instances
-     * after the asset is unloaded. See #5225.
+     * Fires a notification that the set of mesh instances has changed, so that systems which
+     * reference them can re-resolve. An anim component binds morph target weights and animated
+     * material textures to specific mesh instances, and unlike a model component - which owns its
+     * mesh instances on a private graph node hierarchy under its own entity - a render hierarchy is
+     * a public entity hierarchy, so the anim component driving these mesh instances can live on any
+     * ancestor entity. The notification is therefore broadcast rather than delivered directly. See
+     * #5225.
      *
      * @private
      */
-    _rebindAnim() {
+    _onMeshInstancesChanged() {
         if (!this.entity._destroying) {
-            this.entity.anim?.rebind();
+            // the systems registry is torn down before assets are unloaded when the app is
+            // destroyed, and by then there is nothing left to notify
+            this.system.app.systems?.fire('meshInstancesChange', this);
         }
     }
 
--- a/src/framework/components/render/component.js
+++ b/src/framework/components/render/component.js
@@ -329,7 +329,9 @@ class RenderComponent extends Component {
      */
     set meshInstances(value) {
         Debug.assert(Array.isArray(value), 'MeshInstances set to a Render component must be an array.');
-        this.destroyMeshInstances();
+
+        // don't notify the anim component yet, it's notified once the new mesh instances are in place
+        this._destroyMeshInstances();
 
         this._meshInstances = value;
 
@@ -354,6 +356,8 @@ class RenderComponent extends Component {
                 this.addToLayers();
             }
         }
+
+        this._rebindAnim();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/framework/components/render/component.js
```diff
@@ -776,6 +791,21 @@ class RenderComponent extends Component {
...
+     * that it re-resolves the animation targets referencing them - morph target weights and
+     * animated material textures. Without this, morph animations are silently dropped when the
+     * render asset loads after the animation was bound, and keep driving destroyed morph instances
+     * after the asset is unloaded. See #5225.
+     *
+     * @private
+     */
+    _rebindAnim() {
+        if (!this.entity._destroying) {
+            this.entity.anim?.rebind();
```
mvaligursky: [P1] Rebind the anim component that owns this hierarchy

AnimComponent binders resolve graph targets throughout their root hierarchy, but this lookup only finds an anim component co-located with the render component. In a typical imported character, the anim component is on the root while render components live on mesh descendants, so assigning or replacing a child render mesh never reaches the owning anim component and the original missing/stale morph binding remains. I reproduced this on the PR commit with a parent anim entity and a child render entity: after binding before the child mesh existed, assigning the mesh and advancing the animation left the morph weight at 0 instead of 0.5. The same applies to material-texture targets. Please invalidate anim binders whose graph contains this entity (including rootBone configurations), and add a non-co-located regression case.

mvaligursky (commented):

I found one blocking correctness issue. The new tests and the existing render teardown suite pass, but both regression tests place the anim and render components on the same entity. AnimComponent bindings can target descendants, which is the usual layout for imported character hierarchies, and that path is not notified by this implementation.

Validation: 35 existing focused tests passed. An additional parent-anim / child-render regression test against this commit failed with the morph weight remaining at 0 instead of reaching 0.5.

mvaligursky (commented):
Automated follow-up PR review — Codex (GPT-5)

No blocking findings on the updated commit. The previous P1 hierarchy-ownership issue is addressed.

The registry notification now reaches every AnimComponent whose actual binder graph contains the changed render entity, including the common parent-anim / child-render layout and explicit rootBone hierarchies, while avoiding unrelated components. The system-level subscription avoids per-component listeners, disabled anim components remain correctly rebound, entity destruction suppresses unnecessary work, and the optional systems lookup covers late asset callbacks during application teardown. The replacement setter still emits only once after the new mesh instances are installed.

Validation: all 39 focused anim, render, and component-teardown tests pass locally, including the parent/child case that failed against the previous commit. All reported CI checks are green.

PR #9151, opened 2026-07-31
docs: harden the JSDoc and published types that LLMs read

kpal81xd:
An agent writing PlayCanvas code reads two surfaces: the JSDoc blocks and `build/playcanvas.d.ts`. Today neither says that a `StandardMaterial` change needs `update()` to reach the GPU, that `Vec3#scale` has been dead since 2.0, or even that `scale` takes a number. This makes both surfaces say what the source actually does.

### Behaviour that fails silently

Twelve members whose failure is currently only discoverable by debugging. glTF import enables `glossInvert`, so `gloss` holds roughness and `0` is shiny, not rough. `removeChild` detaches without disabling, so the subtree keeps rendering and running scripts. `Asset#ready` fires on success only, and fires immediately with a null resource if registered after a failure. `CameraFrame` groups cannot be passed through `ScriptComponent#create`, whose `properties` merge is shallow and drops each group's `enabled` flag.

Four entry-point classes get the same treatment, since an agent forms its plan there before reading any member: `StandardMaterial` (assignments need `update()`), `Scene` (tone mapping is per camera; `fog` is a read-only `FogParams`), `AppBase` (`init()` is required), `Application` (input devices stay `null` unless passed to the constructor).

### Capability that already ships, from where it is needed

~34 scripts under `scripts/esm/**`, the input-source layer, and the `playcanvas/debug` and `playcanvas/profiler` subpaths. None of these appeared anywhere in `src/`, so nothing reading the package could discover them. Each pointer sits in the block a reader is already in when the need arises — `CameraComponent` points at `camera-controls.mjs`, `RigidBodyComponent` at the character controllers, `ResourceLoader` at the `.obj` and `.spz` parsers.

### 62 legacy members marked deprecated

They warn at runtime through `Debug.deprecated` but carried no doc block, so they reached the declarations as bare signatures with nothing to indicate they are dead — no strikethrough in an editor, no signal to a model trained on the pre-2.x API. Each now uses the pattern already in `GSplatComponent`:

```js
/**
 * @deprecated Use Vec3#mulScalar instead.
 * @param {number} scalar - The number to multiply by.
 * @returns {Vec3} Self for chaining.
 * @ignore
 */
scale(scalar) {
```

`@ignore` keeps them out of the API reference exactly as an absent block did, so the reference is unchanged. The runtime message stays intact, since a console line has to name its own member; the tag carries only the migration.

### 17 signatures hardened

Those blocks initially omitted `@param`/`@returns`, which preserved `any` on every legacy member taking an argument. `any` is the absence of a type, so:

| Change | Members |
| --- | --- |
| `any` → `number` | `scale` (Vec2/3/4), `getTarget`, `setBlendFunction`/`Separate`, `setBlendEquation`/`Separate`, `setDepthFunc` |
| `any` → `boolean` | `setColorWrite`, `setBlending`, `setDepthWrite`, `setDepthTest` |
| `any` → `string` | `setName` |
| `any` → `null` | `VertexFormat.defaultInstancingFormat` |
| `{}` → `{ [x: string]: string }` | `Material#chunks` |
| `GraphNode` → `GraphNode \| null` | `getParent` |

Every row but the last only rejects calls that were already wrong at runtime. **`getParent()` is breaking for TypeScript callers that chain off it**, and is a fix: it returns `this.parent`, which the field and the property both declare `GraphNode|null`, and it is null for the root of any hierarchy, for a node never added to one, and after `removeChild` or a parent's `destroy()`. The deprecated alias had been claiming to be safer than the property that replaces it.

The other 45 members are untouched — tsc already infers them correctly from bodies like `return this.fog.color`, so tags there would be inert.

### Verification

Rebased onto and measured against `main` at d88289fb0:

| Gate | Result |
| --- | --- |
| API reference surface (`utils/api-surface.mjs`) | byte-identical, 5984 lines |
| Published declarations changed | the 17 members above, and nothing else |
| `@deprecated` in `.d.ts` | 53 → 115 |
| TypeDoc | 0 errors; warnings 35 → 34, none introduced |
| `npm run lint` / `npm test` / `npm run test:types` | clean / 2216 passing / passes |

No control flow, default value or export change. A plain `.d.ts` line diff is not a usable gate here: two builds from identical source differ by ~20 lines, because tsc emits unions and class declarations in an unstable order. The declaration row above compares the multiset of tokens with comment lines removed.

### Deliberately not fixed

Two bugs are documented rather than repaired, since this PR touches no logic. `createAttributesDefinition` leaves `attributes` as `undefined`, so a `ShaderMaterial` on a skinned or morphed mesh throws (`shader-generator-shader.js:37`, `: {}` fixes it). And `MiniStats`' render timing subtracts a start timestamp recorded only under `_PROFILER`, so a release build reports time since page load rather than a frame time.

Also excluded: the four behaviour bugs from the original review, which need code and tests, and declarations for `scripts/**`, which would change published artifacts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,12 +80,6 @@ app.on('update', dt => box.rotate(10 * dt, 20 * dt, 30 * dt));
 app.start();
 ```
 
-Elsewhere in your code, or from the browser console, retrieve the running application with
-`AppBase.getApplication()`. You do not need to assign it to `window` yourself.
-
-Ready-made `Script` classes — camera and character controllers, post-processing, water, sky, XR and
-more — ship in the same package under the `playcanvas/scripts/esm/` subpath.
-
 Want to play with the code yourself? Edit it on [CodePen](https://codepen.io/playcanvas/pen/NPbxMj).
 
 A full guide to setting up a local development environment based on the PlayCanvas Engine can be found [here](https://developer.playcanvas.com/user-manual/engine/standalone/).
@@ -145,8 +139,3 @@ Now you can run various build options:
 | ------- | ----------- | ---------- |
 | `npm run build` | Build all engine flavors and type declarations | `build` |
 | `npm run docs` | Build engine [API reference docs](https://api.playcanvas.com/engine/) | `docs` |
-
-The ESM builds are emitted as per-module trees — `build/playcanvas/src/**`,
-`build/playcanvas.dbg/src/**` and `build/playcanvas.prf/src/**` — mirroring `src/` one file per
-module, each with its own `.d.ts`. These ship in the npm package and are the easiest way to read
-engine source or types for a single class without loading the whole `playcanvas.d.ts` bundle.
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ app.start();
 ```
 
 Elsewhere in your code, or from the browser console, retrieve the running application with
-`Application.getApplication()`. You do not need to assign it to `window` yourself.
+`AppBase.getApplication()`. You do not need to assign it to `window` yourself.
 
 Ready-made `Script` classes — camera and character controllers, post-processing, water, sky, XR and
 more — ship in the same package under the `playcanvas/scripts/esm/` subpath.
--- a/README.md
+++ b/README.md
@@ -80,6 +80,12 @@ app.on('update', dt => box.rotate(10 * dt, 20 * dt, 30 * dt));
 app.start();
 ```
 
+Elsewhere in your code, or from the browser console, retrieve the running application with
+`Application.getApplication()`. You do not need to assign it to `window` yourself.
+
+Ready-made `Script` classes — camera and character controllers, post-processing, water, sky, XR and
+more — ship in the same package under the `playcanvas/scripts/esm/` subpath.
+
 Want to play with the code yourself? Edit it on [CodePen](https://codepen.io/playcanvas/pen/NPbxMj).
 
 A full guide to setting up a local development environment based on the PlayCanvas Engine can be found [here](https://developer.playcanvas.com/user-manual/engine/standalone/).
@@ -139,3 +145,8 @@ Now you can run various build options:
 | ------- | ----------- | ---------- |
 | `npm run build` | Build all engine flavors and type declarations | `build` |
 | `npm run docs` | Build engine [API reference docs](https://api.playcanvas.com/engine/) | `docs` |
+
+The ESM builds are emitted as per-module trees — `build/playcanvas/src/**`,
+`build/playcanvas.dbg/src/**` and `build/playcanvas.prf/src/**` — mirroring `src/` one file per
+module, each with its own `.d.ts`. These ship in the npm package and are the easiest way to read
+engine source or types for a single class without loading the whole `playcanvas.d.ts` bundle.
--- a/src/framework/app-base.js
+++ b/src/framework/app-base.js
@@ -117,6 +117,11 @@ let app = null;
  *
  * It is the preferred entry point for new code - {@link Application} is a convenience subclass
  * that registers everything for you, and is expected to be deprecated in a future release.
+ *
+ * `new AppBase(canvas)` only constructs the instance and its root entity. You must then call
+ * {@link AppBase#init} with an {@link AppOptions} supplying at minimum `graphicsDevice`,
+ * `componentSystems` and `resourceHandlers` before adding components or calling
+ * {@link AppBase#start}. {@link Application} assembles those options for you.
  */
 class AppBase extends EventHandler {
     /**
--- a/src/framework/app-base.js
+++ b/src/framework/app-base.js
@@ -121,7 +121,7 @@ let app = null;
  * `new AppBase(canvas)` only constructs the instance and its root entity. You must then call
  * {@link AppBase#init} with an {@link AppOptions} supplying at minimum `graphicsDevice`,
  * `componentSystems` and `resourceHandlers` before adding components or calling
- * {@link AppBase#start}. {@link Application} assembles those options for you.
+ * {@link AppBase#start}. Create the `graphicsDevice` with {@link createGraphicsDevice}.
  */
 class AppBase extends EventHandler {
     /**
--- a/src/framework/app-base.js
+++ b/src/framework/app-base.js
@@ -1185,7 +1185,10 @@ class AppBase extends EventHandler {
     }
 
     /**
-     * Controls how the canvas fills the window and resizes when the window changes.
+     * Controls how the canvas fills the window. The canvas is sized when this is called and on
+     * every {@link AppBase#resizeCanvas}; the engine installs no window `resize` listener of its
+     * own, so call `resizeCanvas` from your own handler to keep the window-relative modes tracking
+     * the window.
      *
      * @param {string} mode - The mode to use when setting the size of the canvas. Can be:
      *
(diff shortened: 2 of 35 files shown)


src/platform/graphics/graphics-device.js
```diff
@@ -951,71 +951,129 @@ class GraphicsDevice extends EventHandler {
...
+    /**
+     * @deprecated Always returns true.
+     * @ignore
+     */
     get supportsVolumeTextures() {
         Debug.deprecated('GraphicsDevice#supportsVolumeTextures is deprecated as it is always true.');
         return true;
     }
 
+    /**
```
mvaligursky, line 1036: why do we need to add jsdocs for what is basically no longer an API?

kpal81xd: Because agents will just start using old and outdated API based on what its been trained on. Also its common for it to start grepping the types file to find out what the types signature is so this directs it towards newer API

src/framework/components/camera/component.js
```diff
@@ -1149,6 +1153,15 @@ class CameraComponent extends Component {
...
      * Convert a point from 3D world space to 2D screen space.
      *
+     * The returned `z` is the unnormalized clip space depth, not a behind-the-camera flag: it also
+     * goes negative for points in front of a perspective camera that are nearer than twice the
+     * near clip, and for an orthographic camera it is negative across the whole near half of the
+     * depth range. To reject points behind the camera, test the view space depth instead - pass
+     * the world position through {@link CameraComponent#viewMatrix} and discard it when the
+     * resulting `z` is zero or greater. For DOM labels anchored to world positions,
+     * `playcanvas/scripts/esm/annotations.mjs` provides `Annotation` and `AnnotationManager`,
+     * which handle behind-camera rejection and occlusion fading.
```
mvaligursky: I don't think reference to annotations belongs on the camera component

src/framework/app-base.js
```diff
@@ -114,6 +114,11 @@ let app = null;
  * {@link ResourceHandler}s yourself. This facilitates
  * [tree-shaking](https://developer.mozilla.org/en-US/docs/Glossary/Tree_shaking) when bundling
  * your application.
+ *
+ * `new AppBase(canvas)` only constructs the instance and its root entity. You must then call
+ * {@link AppBase#init} with an {@link AppOptions} supplying at minimum `graphicsDevice`,
+ * `componentSystems` and `resourceHandlers` before adding components or calling
+ * {@link AppBase#start}. {@link Application} assembles those options for you.
```
mvaligursky: I would not recommend Application here, as this is getting deprecated, and does not support WebGPU / tree-shaking

src/scene/materials/shader-material.js
```diff
@@ -21,7 +21,9 @@ import { Material } from './material.js';
  * @property {string} [fragmentWGSL] - The fragment shader code in WGSL.
  * @property {Object<string, string>} [attributes] - Object detailing the mapping of vertex shader
  * attribute names to semantics SEMANTIC_*. This enables the engine to match vertex buffer data as
- * inputs to the shader. Defaults to undefined, which generates the default attributes.
+ * inputs to the shader. Defaults to undefined, which generates the default attributes. Must be
+ * supplied when the material is applied to a skinned or morphed mesh, as the skinning and morph
+ * attributes the engine adds automatically are merged into this object.
```
mvaligursky: I don't think this is correct regarding to skinned or morphed mesh.

src/scene/materials/standard-material.js
```diff
@@ -61,6 +61,12 @@ const isBlack = (color) => {
  * Most maps can use 3 types of input values in any combination: constant ({@link Color} or number),
  * mesh vertex colors and a {@link Texture}. All enabled inputs are multiplied together.
  *
+ * A property assignment only reaches the GPU once {@link Material#update} is called. The uniform
+ * values below are recomputed solely inside `updateUniforms`, which the renderer skips unless
```
mvaligursky: updateUniforms - we should not expose private function name / functionality in a public API docs.

src/scene/scene.js
```diff
@@ -28,6 +28,13 @@ import { getDefaultMaterial } from './materials/default-material.js';
  * A scene is a graphical representation of an environment. It manages the scene hierarchy, all
  * graphical objects, lights, and scene-wide properties.
  *
+ * Tone mapping and gamma correction are configured per camera, not on the scene: use
+ * {@link CameraComponent#toneMapping} and {@link CameraComponent#gammaCorrection}. `Scene` has no
+ * such properties and no deprecation shim, so assigning to them silently does nothing. Fog remains
+ * scene-wide, but {@link Scene#fog} is a read-only {@link FogParams} object rather than a mode
+ * constant — set its `type`, `color`, `start` and `end` — and {@link CameraComponent#fog} can
+ * override it for a single camera.
+ *
```
mvaligursky: we should not document "property does not exist here", only document what properties exist. Otherwise we could add lots of these statements all over the place.

src/scene/scene.js
```diff
@@ -451,7 +458,10 @@ class Scene extends EventHandler {
     }
 
     /**
-     * Sets the environment lighting atlas.
+     * Sets the environment lighting atlas, an octahedral atlas of prefiltered mip levels. To build
```
mvaligursky: I don't think we support 'an octahedral atlas '

src/scene/scene.js
```diff
@@ -628,6 +638,11 @@ class Scene extends EventHandler {
     /**
      * Sets the base cubemap texture used as the scene's skybox when skyboxMip is 0. Defaults to null.
      *
+     * For a sky that needs no cubemap asset, `playcanvas/scripts/esm/sky/procedural-sky.mjs`
+     * renders an analytic daylight sky and keeps a directional light aligned with the sun so
+     * direct lighting and shadows match. Related scene-dressing scripts ship alongside it:
+     * `water.mjs`, `grid.mjs`, `shadow-catcher.mjs` and `blurred-planar-reflection.mjs`.
```
mvaligursky: `blurred-planar-reflection.mjs` is not related, remove that one.

README.md
```diff
@@ -80,6 +80,12 @@ app.on('update', dt => box.rotate(10 * dt, 20 * dt, 30 * dt));
 app.start();
 ```
 
+Elsewhere in your code, or from the browser console, retrieve the running application with
+`Application.getApplication()`. You do not need to assign it to `window` yourself.
```
mvaligursky: not Application.getApplication() but AppBase.getApplication()

src/scene/materials/shader-material.js
kpal81xd: The sentence describes what the code does today rather than what it should do. With `skin` set and no `attributes`, `createAttributesDefinition` dereferences `undefined`, on this branch and on current `main` (`shader-generator-shader.js:38`):

```js
shaderGeneratorShader.createAttributesDefinition({}, { skin: true, shaderDesc: {} });
// TypeError: Cannot set properties of undefined (setting 'vertex_boneWeights')
```

That is a bug, not a contract, so documenting it as a requirement was the wrong call. Dropping the sentence; the fix belongs in its own PR. Note that defaulting to `{}` is not that fix: for GLSL, `shader.js:193` only collects attributes from the vertex source when the field is `undefined`, so an empty object would skip that path and leave the shader with only the skin and morph semantics. It needs to collect first, then add them.

mvaligursky (approved):

Issue #9051, opened 2026-07-08
Diagonal rendering artifacts on sprites and textures in Exynos 2500 device


andvaribekho:
I’m experiencing some weird rendering artifacts. In a Exynos based device. 

I got this report from a user with a Galaxy Z Flip smf766u1

The game is built using webgl1  and engine ver v1.77

Some UI sprites have a strange diagonal artifact, where the upper-left part of the image appears corrupted. I’m also seeing the same issue on some textures applied to 3D objects. For example, half of the table texture appears corrupted.

In both cases, with sprites and 3D textures, the artifact is split along a diagonal line.

We tested the game in PC, Apple Silicon, iphone, ipad and snapdragon based android phones and it looks fine in all of them.  The problem only occurs so far in Exynos 2500 based devices. 

Any ideas what could be causing this?

Thanks

<img width="512" height="1035" alt="Image" src="https://github.com/user-attachments/assets/5f8b892b-b138-4036-b835-beec073edd85" />

mvaligursky:
We've had multiple issues with Exynos based devices due to their drivers having bugs, and so I suspect this is one of those issues.
Are you able to create a simple repro project for this? Just a single quad / element that reproduces this issue?
Does the same project when using WebGL2 (engine 2) still reproduces the issue?

andvaribekho:
I  prepared a test project  where I'm experiencing the  graphic glitch over diagonals I shared some weeks ago on Exynos based devices.

https://playcanvas.com/editor/scene/2565759


The problem happens only in Google Chrome,  it doesn't happen when using Firefox or other chromium based browsers like Edge or Brave.,

The problem happens both in UI elements using sprites and in 3D geometry using textures too.,

I tested different file formats ( jpg, png, webp) different resolutions ( 1024, 512, 2048 ) and different compression algorithms ( basis and legacy ),  I also tested using the textures as "repeat, clamp or mirror repeat"  and nothing seems to solve the issue.

I also tried disabling Vulkan and Skia Graphite but the problem persists

The weirdest thing is that the glitch happens randomly.  Some sprites or textures are displayed correctly  ( like the cup pong logo or the "play online" and "play with friends" buttons )  but others are glitched.    And there is no difference between the working assets and the glitched assets.  It happens randomly.  In sprites from texture atlases  or textures.

The problem is affecting not only our games, but games from other developers too ( attached a screenshot from Om Nom Run)

[...]

PR #9156, opened 2026-08-07
Draw UI image element quads as indexed triangles, not a triangle strip

willeastcott:
## Description

UI image elements built their quad as a non-indexed 4 vertex `GL_TRIANGLE_STRIP`. On Exynos 2500 / Xclipse (RDNA) devices in Chrome for Android, the interpolated uvs come out corrupted across **exactly the strip's second triangle**, so half of every UI image is smeared along the quad diagonal.

The vertex order is BR, TR, BL, TL, so the strip emits:

- triangle 0 = (BR, TR, BL) — the bottom-right half
- triangle 1 = (TR, BL, TL) — the top-left half, and the *odd* triangle, whose winding the driver has to flip to keep facing consistent

In the reporter's screenshots the bottom-right half of every affected sprite is correct and the top-left half is smeared along the BL→TR diagonal. That is triangle 1, every time.

The control case is in the same screenshots: **all the text renders correctly.** Text elements use the same shaders, the same atlas textures and the same materials — they differ only in that `TextElement` already emits indexed triangles. The strip's implicit winding flip for odd triangles is the one thing the two paths do not share.

Rather than detect the device, this drops the strip. It buys nothing here (both are a single draw of two triangles), it is a legacy topology that little else on the web still exercises, and the batcher already had to convert these meshes to indexed triangles anyway. The index order matches the batcher's, so batched and non-batched images rasterize identically.

Changes:

- `image-element.js` — 4 vertices plus a 6 index buffer, `PRIMITIVE_TRIANGLES`. The index buffer is shared by every image element via a `DeviceCache`, and `ImageElement#destroy` detaches it before the renderable tears the mesh down, since `Mesh#destroy` would otherwise destroy a buffer the remaining elements still use.
- `immediate.js` — same for the `app.drawTexture` quad, the only other content-path strip in the engine
- `batch-manager.js` — comment no longer claims the non-indexed fan / strip case is the UI path
- `component.test.mjs` — regression tests for the topology and for the shared buffer surviving a single element's destruction

## Correctness

Built the engine before and after and diffed real GPU output on WebGL 2. For plain, stencil-masked and batched UI images, and for `app.drawTexture`, the rendered pixels are **identical**, while the draw call changes:

```
before: drawArrays(TRIANGLE_STRIP, 4)
after:  drawElements(TRIANGLES, 6)
```

Batching still merges 3 quads into a single `drawElements(TRIANGLES, 18)`, and a stencil-masked child still clips to exactly 100x100 / 10000 px at the same pixel coordinates as before.

Test suite: 2044 passing, with the same 11 pre-existing failures as `main` (asset fetches that need network access).

## Performance

Benchmarked because an indexed draw is not obviously free. 2000 UI image elements at 8x8px (small and non-overlapping, so this is draw bound rather than fill bound), 90 frames x 8 rounds with the variant order alternating each round, `gl.finish()` per frame. Intel Arc Pro 140T, ANGLE/D3D11:

| variant | draw call | median ms/frame | vs strip |
|---|---|---|---|
| strip (before) | `drawArrays(TRIANGLE_STRIP, 4)` | 3.40 | +0.0% |
| indexed, index buffer per element | `drawElements(TRIANGLES, 6)` | 4.20 | +23.5% |
| **indexed, one shared index buffer (this PR)** | `drawElements(TRIANGLES, 6)` | **3.40** | **+0.0%** |
| 6 non-indexed vertices | `drawArrays(TRIANGLES, 6)` | 3.40 | +0.0% |

The cost is the per-element buffer, not the indexed draw — holding everything else constant and swapping 2000 distinct index buffers for one shared one recovers all of it. That matches the code: `WebglGraphicsDevice#setBuffers` rebinds `ELEMENT_ARRAY_BUFFER` on every draw regardless, since it cannot cache the state (the VAO captures the last bind). A distinct buffer per element turns each of those into a real rebind, on top of a driver side buffer object per element. With one shared buffer it is the same handle every time.

Worth noting the corollary: going indexed adds no extra WebGL call, because non-indexed draws already pay that `bindBuffer(…, null)`.

6 non-indexed vertices measured at parity too, so it was not worth the larger diff — it would need `_updateMesh` rewritten to write duplicated corners, and `batch-manager.js` taught about non-indexed `PRIMITIVE_TRIANGLES` (today the non-indexed branch only handles TRIFAN/TRISTRIP, and anything else falls through leaving `numIndices` / `indexData` from the previous loop iteration — a latent bug that change would start hitting).

This is desktop ANGLE/D3D11, not a mobile tiler, so the ranking there may differ — though "one buffer object instead of N" should hold at least as well.

## Note on the 3D case

I do not have an Exynos device, so this is not confirmed on hardware — but the artifact lines up with the strip's odd triangle too precisely for that to be coincidence.

The 3D table in the first screenshot of #9051 is a separate question: indexed triangle meshes do not go through the strip path, so this change will not affect them. The one route by which a 3D mesh would hit the same driver bug is a glTF with `mode: 5` primitives, which `gltf-accessor.js` maps straight to `PRIMITIVE_TRISTRIP`. Worth asking the reporter to check `meshInstance.mesh.primitive[0].type` on the affected mesh — if it is `5`, the same reasoning extends there; if it is `4`, that artifact has a different cause.

Refs #9051

## Checklist

### DIFF
--- a/src/framework/components/element/image-element.js
+++ b/src/framework/components/element/image-element.js
@@ -6,12 +6,15 @@ import { Vec2 } from '../../../core/math/vec2.js';
 import { Vec3 } from '../../../core/math/vec3.js';
 import { Vec4 } from '../../../core/math/vec4.js';
 import {
+    BUFFER_STATIC,
     FUNC_EQUAL,
-    PRIMITIVE_TRISTRIP,
+    INDEXFORMAT_UINT16,
+    PRIMITIVE_TRIANGLES,
     SEMANTIC_POSITION, SEMANTIC_NORMAL, SEMANTIC_TEXCOORD0,
     STENCILOP_DECREMENT,
     TYPE_FLOAT32
 } from '../../../platform/graphics/constants.js';
+import { IndexBuffer } from '../../../platform/graphics/index-buffer.js';
 import { VertexBuffer } from '../../../platform/graphics/vertex-buffer.js';
 import { VertexFormat } from '../../../platform/graphics/vertex-format.js';
 import { DeviceCache } from '../../../platform/graphics/device-cache.js';
@@ -37,6 +40,10 @@ import { Asset } from '../../asset/asset.js';
 
 const _tempColor = new Color();
 const _vertexFormatDeviceCache = new DeviceCache();
+
+// Triangulation of the 4 vertex quad the image element is built from. This matches the triangulation
+// the batcher applies to these meshes, so batched and non-batched images rasterize identically.
+const _quadIndices = [0, 1, 3, 0, 3, 2];
 
 class ImageRenderable {
     constructor(entity, mesh, material) {
@@ -433,7 +440,7 @@ class ImageElement {
         const r = this._rect;
         const device = this._system.app.graphicsDevice;
 
-        // content of the vertex buffer for 4 vertices, rendered as a tristrip
+        // content of the vertex buffer for the 4 corners of the quad
         const vertexData = new Float32Array([
             w, 0, 0,                        // position
             0, 0, 1,                        // normal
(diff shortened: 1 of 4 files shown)


mvaligursky (approved):

Reviewed topology and winding correctness, WebGL/WebGPU behavior, batching consistency, shared index-buffer ownership across element destruction and device loss/restoration, immediate-mode rendering, API compatibility, and performance impact. No blocking issues found.

The indexed triangle order is consistent with the batcher, the shared DeviceCache allocation avoids per-element index buffers, and the explicit detach prevents Mesh destruction from invalidating other image elements. The focused ElementComponent suite passes (8/8), and all repository CI checks are green.

willeastcott:
Noooooo! I meant to close this.

mvaligursky:
> Noooooo! I meant to close this.

Hah! Need a revert PR now :)

PR #9163, opened 2026-08-12
Fix StandardMaterial property types in the generated .d.ts

### LINKED ISSUE #9152: StandardMaterial: occludeDirect is typed as number in the generated .d.ts, and opacityShadowDither's JSDoc says boolean
Two `StandardMaterial` properties disagree between their class-level `@property` JSDoc and the emitted `.d.ts`, in opposite directions. In each case the runtime value tells you which side is wrong.

Found in **playcanvas 2.21.3** while adding typed attribute bindings over `StandardMaterial` in [`@playcanvas/web-components`](https://github.com/playcanvas/web-components).

## The disagreement

| Property | `@property` JSDoc | Emitted `.d.ts` | Runtime default | Wrong |
|---|---|---|---|---|
| `occludeDirect` | `{boolean}` | `number` | `false` (boolean) | **the declaration** |
| `opacityShadowDither` | `{boolean}` | `string` | `'none'` (string) | **the JSDoc** |

## Reproduction

```js
import { StandardMaterial } from 'playcanvas';

const m = new StandardMaterial();
console.log(typeof m.occludeDirect);       // 'boolean'
console.log(typeof m.opacityShadowDither); // 'string'
```

Emitted declarations (`build/playcanvas.d.ts`):

```ts
set occludeDirect(arg: number);          // 17642
get occludeDirect(): number;

set opacityShadowDither(arg: string);    // 17662
get opacityShadowDither(): string;
```

Class-level JSDoc:

```
@property {boolean} occludeDirect Tells if AO should darken directional lighting. Defaults to false.
@property {boolean} opacityShadowDither Used to specify whether shadow opacity is dithered, which ...
```

## Impact

`occludeDirect` is the one that bites consumers: TypeScript rejects the correct assignment, so setting it from TS needs a `@ts-ignore`.

```ts
material.occludeDirect = true;
// error TS2322: Type 'boolean' is not assignable to type 'number'.
```

`opacityShadowDither` is documentation-only — the declaration is already right — but the JSDoc says `boolean` for what is actually a dither mode (`DITHER_NONE`/`DITHER_BAYER8`/`DITHER_BLUENOISE`/`DITHER_IGNNOISE`), so the API reference is misleading. Its sibling `opacityDither` is documented correctly.

## Possibly related: a missing doc comment

`occludeDirect` is one of only three `StandardMaterial` accessors emitted with no doc comment at all — a blank line where every neighbour has one:

```ts
	/** Main (primary) normal map UV channel. */
	set normalMapUv(arg: number);
	get normalMapUv(): number;

	
	set occludeDirect(arg: number);
	get occludeDirect(): number;

	/** Uses ambient occlusion to darken specular/reflection. ... */
	set occludeSpecular(arg: number);
	get occludeSpecular(): number;
```

The other two are `opacityShadowDither` and `sheenVertexColorChannel` (the latter has no `@property` entry at all, so its missing doc is expected). Losing the doc comment and falling back to the neighbouring `number` look like they may share a cause in whatever associates the class-level `@property` block with the generated accessors — but that's a guess, and the two type fixes stand on their own regardless.

Every other boolean flag on the class (`useMetalness`, `useSheen`, `enableGGXSpecular`, `opacityFadesSpecular`, `specularTint`, `twoSidedLighting`) emits `boolean` correctly and keeps its doc comment, so this is isolated rather than systemic.

willeastcott:
## Description

`StandardMaterial` defines its accessors dynamically, so tsc emits none of them — they are declared by hand from `STANDARD_MAT_PROPS` in `utils/plugins/rollup-types-fixup.mjs`. Each doc comment was looked up with ``contents.match(`@property {${type}} ${name}`)``, **passing the type through as a regex**. That one line caused three distinct defects.

**1. Type disagreements silently dropped the doc comment and emitted the wrong type.** When the type in the list disagreed with the class JSDoc, the lookup found nothing, so the accessor was emitted with an empty doc comment and the (wrong) type from the list:

| Property | Emitted `.d.ts` | Runtime default | Correct type |
|---|---|---|---|
| `occludeDirect` | `number` | `false` | `boolean` |
| `alphaFade` | `boolean` | `1` (`_defineFloat`) | `number` |

Both broke the correct assignment from TypeScript:

```ts
material.occludeDirect = true;  // TS2322: Type 'boolean' is not assignable to type 'number'.
material.alphaFade = 0.5;       // TS2322: Type 'number' is not assignable to type 'boolean'.
```

`occludeDirect` is the one reported in #9152; `alphaFade` is the same defect found by generalizing it.

**2. `Texture|null` was treated as a regex alternation.** It matched `@property {Texture` at its *first* occurrence in the file and then sliced from the wrong offset, so every texture property inherited a mangled fragment of `diffuseMap`'s description:

```ts
/** e main (primary) diffuse map of the material (default is null). */
set anisotropyMap(arg: Texture|null);

/** seMap The main (primary) diffuse map of the material (default is null). */
set aoMap(arg: Texture|null);
```

**3. `@property {number} anisotropy` prefix-matched `anisotropyIntensity`**, so the deprecated alias inherited its neighbour's documentation.

### Changes

- `occludeDirect` → `boolean`, `alphaFade` → `number`, matching their runtime defaults.
- `opacityShadowDither`'s JSDoc said `boolean` for what is a dither mode string (`DITHER_NONE`/`DITHER_BAYER8`/…). The emitted declaration was already correct, so the JSDoc is corrected to `string` — its sibling `opacityDither` was already documented correctly.
- The doc lookup now parses the `@property` block into a map keyed by property name instead of building a regex out of the type. This restores the correct description on ~30 texture properties and on `anisotropy`.
- **The build now throws** on a type disagreement or a missing `@property` tag, so this cannot silently regress into a dropped doc comment again:
  ```
  StandardMaterial types disagree with its JSDoc - fix the @property tag in
  src/scene/materials/standard-material.js or the type in STANDARD_MAT_PROPS:
    occludeDirect: declared as 'number' but documented as 'boolean'
  ```
- Added the missing `@property` tag for `sheenVertexColorChannel` (verified `'rgb'` at runtime) and an explicit `@deprecated` doc for the `anisotropy` alias.

Every `StandardMaterial` accessor now emits with its own correct doc comment — previously `occludeDirect`, `opacityShadowDither`, `alphaFade` and `sheenVertexColorChannel` had none at all.

Fixes #9152

### Notes for reviewers

- The `guard` no longer encodes a type (`set alphaFade(arg: boolean);` → `set alphaFade(arg:`), so it won't need updating the next time a type changes. Re-running `build:types` is idempotent — verified.
- The new check firing was verified by re-introducing both failure modes against a temp copy of the plugin.
- **Out of scope, worth a follow-up:** checking the reverse direction turned up 49 *documented* properties with no declaration in the `.d.ts` at all (`thickness`, `attenuation`, `sheenGloss`, `alphaDither`, and all the iridescence/refraction ones) — the hand-maintained list holds 184 entries against 233 `@property` tags. That's missing declarations rather than wrong ones, and adding 49 entries would swamp this change, so it is deliberately left out.

### Testing

- `npm run build:types` → `npm run test:types` clean.
- The issue's repro type-checks under `--strict`, with `occludeDirect = 1` and `alphaFade = true` still correctly rejected (asserted via `@ts-expect-error`).
- `npm run lint` clean.
- `npm test`: 2044 passing / 11 failing — **identical with the changes stashed**. The 11 are pre-existing asset-loading timeouts in the font/sprite handler tests, unrelated to this change.

## Checklist

### DIFF
--- a/src/scene/materials/standard-material.js
+++ b/src/scene/materials/standard-material.js
@@ -338,6 +338,8 @@ const isBlack = (color) => {
  * "g", "b", "a", "rgb" or any swizzled combination.
  * @property {boolean} sheenVertexColor Use mesh vertex colors for sheen. If sheen map or
  * sheen tint are set, they'll be multiplied by vertex colors.
+ * @property {string} sheenVertexColorChannel Vertex color channels to use for sheen. Can be "r",
+ * "g", "b", "a", "rgb" or any swizzled combination.
  * @property {number} sheenGloss The glossiness of the sheen (fabric) microfiber structure.
  * This color value is a single value between 0 and 1.
  * @property {boolean} sheenGlossInvert Invert the sheen gloss component (default is false).
@@ -389,7 +391,7 @@ const isBlack = (color) => {
  * - {@link DITHER_IGNNOISE}: Opacity is dithered using an interleaved gradient noise.
  *
  * Defaults to {@link DITHER_NONE}.
- * @property {boolean} opacityShadowDither Used to specify whether shadow opacity is dithered, which
+ * @property {string} opacityShadowDither Used to specify whether shadow opacity is dithered, which
  * allows shadow transparency without alpha blending. Can be:
  *
  * - {@link DITHER_NONE}: Opacity dithering is disabled.
--- a/utils/plugins/rollup-types-fixup.mjs
+++ b/utils/plugins/rollup-types-fixup.mjs
@@ -7,10 +7,15 @@ const REGULAR_OUT = '\x1b[22m';
 
 const TYPES_PATH = './build/playcanvas/src';
 
+// StandardMaterial defines its accessors dynamically, so TypeScript emits none of them. They are
+// declared by hand from this list, and their doc comments are taken from the matching
+// `@property {type} name` tag in the class JSDoc - so the type here must match the type documented
+// there, or the build fails. An optional third element supplies the doc comment for properties that
+// have no `@property` tag (deprecated aliases, which are defined in src/deprecated/deprecated.js).
 const STANDARD_MAT_PROPS = [
-    ['alphaFade', 'boolean'],
+    ['alphaFade', 'number'],
     ['ambient', 'Color'],
-    ['anisotropy', 'number'],
+    ['anisotropy', 'number', 'Defines amount of anisotropy. @deprecated Use {@link StandardMaterial#anisotropyIntensity} and {@link StandardMaterial#anisotropyRotation} instead.'],
     ['anisotropyIntensity', 'number'],
     ['anisotropyRotation', 'number'],
     ['anisotropyMap', 'Texture|null'],
@@ -137,7 +142,7 @@ const STANDARD_MAT_PROPS = [
     ['normalMapRotation', 'number'],
     ['normalMapTiling', 'Vec2'],
     ['normalMapUv', 'number'],
-    ['occludeDirect', 'number'],
+    ['occludeDirect', 'boolean'],
     ['occludeSpecular', 'number'],
     ['occludeSpecularIntensity', 'number'],
     ['opacity', 'number'],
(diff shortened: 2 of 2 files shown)


willeastcott:
Superseded by #9295, which fixes the same two type errors (alphaFade is now number, occludeDirect is now boolean) and replaces the regex-based description lookup with a plain string search, so the Texture|null alternation no longer mangles the texture-map doc comments. The JSDoc side of this branch (opacityShadowDither typed as string, sheenVertexColorChannel documented) landed in #9294. The one piece not carried over is the build-time check that the hand-declared list agrees with the @property tags; that can be revisited separately if wanted.

PR #9189, opened 2026-08-18
Destroy the asset registry with the application

### LINKED ISSUE #3886: Asset memory is not free correctly because registry attribute is still binded
https://github.com/playcanvas/engine/blob/771201ad5d421eefacce0e5bd83e6a7585122f3d/src/asset/asset.js#L490-L515

to

```
 unload() {
        if (!this.loaded && this._resources.length === 0)
            return;

        this.fire('unload', this);
        this.registry.fire('unload:' + this.id, this);

        const old = this._resources;

        // remove resource from loader cache
        if (this.file) {
            this.registry._loader.clearCache(this.getFileUrl(), this.type);
        }

        // clear resources on the asset
        this.resources = null;
        this.registry = null;
        this.loaded = false;

        // destroy resources
        for (let i = 0; i < old.length; ++i) {
            const resource = old[i];
            if (resource && resource.destroy) {
                resource.destroy();
            }
        }
    }
```

mvaligursky:
Fixes #3886.

Every `Asset` holds a reference to its `AssetRegistry` - `AssetRegistry#add` assigns it and nothing ever cleared it - and the registry itself was never cleaned up, as it had no `destroy` method and `AppBase#destroy` only called `off()` on it. An asset which outlived the application therefore kept the registry alive, and with it every other asset in the registry, including their `data` and `file.contents` payloads.

Verified with `WeakRef` probes on the null device. Holding an asset after `app.destroy()`:

| | before | after |
| --- | --- | --- |
| asset was in the registry | registry, loader and app all retained | all collected |
| asset was removed earlier | registry and all other assets retained | other assets collected, registry left as an empty shell |

The application itself is no longer reachable in either case since #9190.

### Changes

- New `AssetRegistry#destroy`, called from `AppBase#destroy`: unsubscribes from the `name` and `tags` events of all assets, clears their `registry` back-reference, empties the four lookup maps and releases the tags cache and bundle registry. `AppBase#assets` is cleared as well, which is what the commented out assertion in `application.test.mjs` was waiting for.
- `Asset#unload` no longer assumes a registry, since assets which outlive the application no longer have one.

### Compatibility

No behaviour change for a running application - only teardown is affected:

- `AssetRegistry#remove` is untouched, so `remove()` followed by `unload()` still clears the loader cache entry as before, and `asset.registry` stays valid for as long as the application lives.
- `app.assets` is now `null` after `app.destroy()`, consistent with `app.root`, `app.scene` and `app.loader`, which are already cleared there.

An earlier revision of this PR also cleared `Asset#registry` in `AssetRegistry#remove`. That turned out to be a breaking change for little benefit - it makes `asset.data`, `asset.loadFaces` and `asset.reload()` throw on a removed asset, and silently stops `unload()` from clearing the loader cache entry - while freeing nothing extra, because the registry it points at is emptied by `destroy` anyway.

### DIFF
--- a/src/framework/asset/asset-registry.js
+++ b/src/framework/asset/asset-registry.js
@@ -353,8 +353,12 @@ class AssetRegistry extends EventHandler {
             asset.tags.off('remove', this._onTagRemove, this);
 
             // clear the back-reference, so that an asset which outlives the application does not
-            // keep the registry - and through it all other assets - alive
-            asset.registry = null;
+            // keep the registry - and through it all other assets - alive. An asset added to
+            // another registry since points at that one instead, which cannot retain this
+            // registry, so it is left alone.
+            if (asset.registry === this) {
+                asset.registry = null;
+            }
         }
 
         this._assets.clear();
--- a/src/framework/asset/asset-registry.js
+++ b/src/framework/asset/asset-registry.js
@@ -340,6 +340,35 @@ class AssetRegistry extends EventHandler {
     }
 
     /**
+     * Destroys the registry, releasing all assets held by it. Called by {@link AppBase#destroy}.
+     * Note that this does not destroy the resources of the assets - {@link Asset#unload} needs to
+     * be called for each asset before the registry is destroyed.
+     *
+     * @ignore
+     */
+    destroy() {
+        for (const asset of this._assets) {
+            asset.off('name', this._onNameChange, this);
+            asset.tags.off('add', this._onTagAdd, this);
+            asset.tags.off('remove', this._onTagRemove, this);
+
+            // clear the back-reference, so that an asset which outlives the application does not
+            // keep the registry - and through it all other assets - alive
+            asset.registry = null;
+        }
+
+        this._assets.clear();
+        this._idToAsset.clear();
+        this._urlToAsset.clear();
+        this._nameToAsset.clear();
+
+        this._tags = null;
+        this.bundles = null;
+
+        this.off();
+    }
+
+    /**
      * Retrieve an asset from the registry by its id field.
      *
      * @param {number} id - The id of the asset to get.
--- a/src/framework/app-base.js
+++ b/src/framework/app-base.js
@@ -1953,7 +1953,8 @@ class AppBase extends EventHandler {
             assets[i].unload();
             assets[i].off();
         }
-        this.assets.off();
+        this.assets.destroy();
+        this.assets = null;
 
         // destroy scene after assets are unloaded (components need scene.layers during asset cleanup)
         this.scene.destroy();
--- a/src/framework/asset/asset.js
+++ b/src/framework/asset/asset.js
@@ -598,7 +598,7 @@ class Asset extends EventHandler {
         }
 
         this.fire('unload', this);
-        this.registry.fire(`unload:${this.id}`, this);
+        this.registry?.fire(`unload:${this.id}`, this);
 
         const old = this._resources;
 
@@ -613,7 +613,7 @@ class Asset extends EventHandler {
 
         // remove resource from loader cache
         if (this.file) {
-            this.registry._loader.clearCache(this.getFileUrl(), this.type);
+            this.registry?._loader.clearCache(this.getFileUrl(), this.type);
         }
 
         // destroy resources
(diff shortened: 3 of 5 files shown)


src/framework/asset/asset-registry.js
```diff
@@ -295,7 +295,8 @@ class AssetRegistry extends EventHandler {
     }
 
     /**
-     * Remove an asset from the registry.
+     * Remove an asset from the registry. This clears {@link Asset#registry}, so call
+     * {@link Asset#unload} before removing an asset if its resource needs to be destroyed.
```
mvaligursky: [P1] Update the remaining remove-before-unload callers. This new contract is not yet followed by several shipped repository paths: scripts/esm/xr/xr-controllers.mjs (two sites), scripts/esm/gsplat/gsplat-flipbook.mjs, and the gaussian-splatting lod-streaming and relighting examples still call remove() and then unload(). With this change, unload() destroys the resource but cannot clear ResourceLoader._cache; a later load of the same URL therefore returns the cached, already-destroyed resource. I reproduced that exact state on this head. Please update every remaining call site (or make remove() evict the loader cache without destroying the resource) before changing this public lifecycle behavior.

src/framework/asset/asset-registry.js
```diff
@@ -336,9 +337,43 @@ class AssetRegistry extends EventHandler {
             this.fire(`remove:url:${asset.file.url}`, asset);
         }
 
+        // clear the back-reference to this registry, so that code holding on to the asset does not
+        // keep the registry - and through it the whole application - alive. Skipped if a listener
+        // of the events fired above has added the asset back to the registry.
+        if (!this._assets.has(asset)) {
```
mvaligursky: [P2] Preserve a registry assigned by a remove listener. The guard only detects re-addition to this same registry. If an asset remove listener migrates the asset with otherRegistry.add(asset), otherRegistry now contains the asset and sets asset.registry correctly, but this block subsequently overwrites that back-reference with null. I reproduced the inconsistent state (otherRegistry.get(id) returns the asset while asset.registry is null). Only clear the reference when asset.registry still equals this and the asset was not re-added here.

mvaligursky (commented):
Automated PR review by Codex (GPT-5). Two compatibility issues need addressing. Focused affected-area tests pass (41/41), but targeted lifecycle probes reproduce both findings below.

src/framework/asset/asset-registry.js
```diff
@@ -339,6 +339,35 @@ class AssetRegistry extends EventHandler {
...
+     */
+    destroy() {
+        for (const asset of this._assets) {
+            asset.off('name', this._onNameChange, this);
+            asset.tags.off('add', this._onTagAdd, this);
+            asset.tags.off('remove', this._onTagRemove, this);
+
+            // clear the back-reference, so that an asset which outlives the application does not
+            // keep the registry - and through it all other assets - alive
+            asset.registry = null;
```
mvaligursky: [P2] Only clear this registry own back-reference. AssetRegistry.add currently permits the same Asset object to be added to a second registry and updates asset.registry to that newer registry. The first registry still contains the asset, so destroying it reaches this unconditional assignment and resets the back-reference to null even though the asset remains in the live second registry. I reproduced the resulting inconsistent state: secondRegistry.get(asset.id) returns the asset while asset.registry is null. Guard this with asset.registry === this before clearing; when the reference belongs elsewhere it cannot retain this registry, so the leak fix is preserved.

mvaligursky (commented):
Fresh full automated PR review by Codex (GPT-5), performed from the current base and head rather than as a delta from the earlier review. Reviewed all changed files, application teardown ordering, listener ownership, loader/cache behavior, removed and retained assets, re-entrancy, multi-application behavior, memory collection, compatibility, and performance. The focused application/asset/i18n suite passes (123 passing, 2 pending), the targeted remove→unload compatibility probe passes, the WeakRef/GC probe collects the application, registry, and loader, and all repository CI checks are green. One remaining correctness issue is noted inline.

src/framework/asset/asset-registry.js
mvaligursky: Good catch - fixed in 6f24009b1: the back-reference is now only cleared when `asset.registry === this`, with a test covering the two-registry case. As you say, a reference which belongs elsewhere cannot retain this registry, so the leak fix is unaffected.

Worth noting the mirror image is still ill-defined: destroy the *newer* registry and the check passes, so the reference is cleared while the asset remains in the older registry's maps. Resolving that properly needs a surviving-owner concept which the class does not have - `add` already clobbers `asset.registry` silently - so I have left it there.

Issue #239, opened 2015-04-19
Flat shading


willeastcott:
It would be good if there was an easy way to enable flat shading on a mesh.

detzt:
Thanks, that would be really helpful!

Maksims:
In WebGL 2.0 it is possible to define shading `smooth` or `flat` for `in`/`out`. It means it will be possible by altering shader to enable flat shading for just normals. Although this will not work for WebGL 1.0.

willeastcott:
Closing mainly due to age. Since I originally submitted this and I no longer think this is much of an issue, I'm going to close.

kungfooman:
> Due to inability of making this shader shading approach backwards compatible without affecting a vertex buffer, the only way would be to preceduraly edit vertex and index buffers splitting vertices into separate triangles.

Not really, you can use `GL_OES_standard_derivatives`. Based on [this](https://registry.khronos.org/OpenGL/extensions/OES/OES_standard_derivatives.txt) it is OpenGL ES 2.0 (aka WebGL 1.0).

Normal, "non flat":

![Image](https://github.com/user-attachments/assets/0b13f453-ab5b-48e1-85f7-043a60e792dc)

Standard derivative flat shading:

![Image](https://github.com/user-attachments/assets/bf5e7cfc-c04e-48b4-84f2-05d7624f6e87)

PlayCanvas version 2 code:

```js
  const material = new pc.ShaderMaterial({
    attributes: {
      aPosition: pc.SEMANTIC_POSITION
    },
    vertexGLSL: `
      attribute vec3 aPosition;
      uniform mat4 matrix_model;
      uniform mat4 matrix_viewProjection;
      varying vec3 vWorldPos;
      void main(void) {
        vec4 worldPos = matrix_model * vec4(aPosition, 1.0);
        vWorldPos = worldPos.xyz;
        gl_Position = matrix_viewProjection * worldPos;
      }
    `,
    fragmentGLSL: `
      varying vec3 vWorldPos;
      void main(void) {
        vec3 dx = dFdx(vWorldPos);
        vec3 dy = dFdy(vWorldPos);
        vec3 normal = normalize(cross(dx, dy));
        gl_FragColor = vec4(normal * 0.5 + 0.5, 1.0);
      }
    `
  });
```


I know this issue is old, but... there is still no builtin solution? Or at least I can't find anything. Maybe we can reopen @willeastcott ?

[...]

PR #9191, opened 2026-08-18
Add flat shading support

mvaligursky:
Adds flat shading, resolving #239. When enabled, each fragment is shaded using the geometric normal of its triangle instead of the normal interpolated from the vertex normals, giving the surface a faceted look.

The normal is taken from the screen space derivatives of the world position. This is exact rather than an approximation - both derivatives lie in the plane of the triangle, so their cross product is the normal of that plane - and it works on skinned and morphed geometry at no additional cost, as it operates on the final world space positions. It also does not need the mesh to supply vertex normals at all.

The returned normal is oriented to match the triangle winding, so it agrees with correctly authored vertex normals on both front and back faces. Flat shading therefore only changes the faceting: `cull`, `frontFace` and `twoSidedLighting` all behave as they do for smooth shading.

**Changes:**
- New `flatNormalPS` chunk (GLSL and WGSL) providing `getFlatNormal(worldPos)`
- `Material#flatShading`, backed by a `FLAT_SHADING` define. As all material types resolve their defines through `ShaderUtils.getCoreDefines`, this needs no lit option plumbing, and it reaches `StandardMaterial`, `LitMaterial` and `ShaderMaterial` alike
- The `tbnBasis` uniform declaration in `TBN.js` is now behind an `#ifndef` guard so it can be shared with the new chunk
- Two independent sign corrections that do not double count: `tbnBasis` compensates for the flipY projection flip and WebGPU's Y-down framebuffer space, then `gl_FrontFacing` / `pcFrontFacing` restores the winding orientation

**API Changes:**
- Added `Material#flatShading` (boolean, defaults to `false`). As with other material properties, call `Material#update()` after changing it
- Added `flatShading` to the standard material asset parameters
- Added the `flatNormalPS` shader chunk, usable from `ShaderMaterial` shaders via `#include "flatNormalPS"` together with `#ifdef FLAT_SHADING`

**Examples:**
- `graphics/asset-viewer` - flat shading toggle, exercising it across transmission, volume, IOR, iridescence, sheen and clearcoat materials
- `graphics/lights` - flat shading toggle on the statue and ground, showing that shadows and light cookies remain correct
- `shaders/shader-hatch` - flat shading toggle demonstrating the `ShaderMaterial` route. Its lighting moved from the vertex to the fragment shader, which derivatives require, so it is now evaluated per pixel

### DIFF
--- a/src/scene/shader-lib/wgsl/chunks/lit/frag/pass-forward/litForwardMain.js
+++ b/src/scene/shader-lib/wgsl/chunks/lit/frag/pass-forward/litForwardMain.js
@@ -22,7 +22,11 @@ fn fragmentMain(input: FragmentInput) -> FragmentOutput {
     #endif
 
     #ifdef LIT_NEEDS_NORMAL
-        dVertexNormalW = normalize(vNormalW);
+        #ifdef FLAT_SHADING
+            dVertexNormalW = getFlatNormal(vPositionW);
+        #else
+            dVertexNormalW = normalize(vNormalW);
+        #endif
 
         #ifdef LIT_TANGENTS
             #if defined(LIT_HEIGHTS) || defined(LIT_USE_NORMALS) || defined(LIT_USE_CLEARCOAT_NORMALS) || defined(LIT_GGX_SPECULAR)
--- a/examples/assets/scripts/misc/hatch-material.mjs
+++ b/examples/assets/scripts/misc/hatch-material.mjs
@@ -54,15 +54,9 @@ const createHatchMaterial = (device, textures) => {
                 // Functions added: getNormalMatrix, getLocalNormal
                 #include "normalCoreVS"
 
-                // engine supplied uniforms
-                uniform vec3 view_position;
-
-                // out custom uniforms
-                uniform vec3 uLightDir;
-                uniform float uMetalness;
-
-                // variables we pass to the fragment shader
-                varying float brightness;
+                // variables we pass to the fragment shader, where the lighting is evaluated
+                varying vec3 worldPos;
+                varying vec3 worldNormal;
 
             #endif
 
@@ -71,25 +65,18 @@ const createHatchMaterial = (device, textures) => {
                 // use functionality from transformCore to get a world position, which includes skinning and morphing as needed
                 mat4 modelMatrix = getModelMatrix();
                 vec3 localPos = getLocalPosition(vertex_position.xyz);
-                vec4 worldPos = modelMatrix * vec4(localPos, 1.0);
+                vec4 worldPosition = modelMatrix * vec4(localPos, 1.0);
 
                 #ifndef SHADOW_PASS
 
                     // use functionality from normalCore to get the world normal, which includes skinning and morphing as needed
                     mat3 normalMatrix = getNormalMatrix(modelMatrix);
                     vec3 localNormal = getLocalNormal(vertex_normal);
-                    vec3 worldNormal = normalize(normalMatrix * localNormal);
 
-                    // simple wrap-around diffuse lighting using normal and light direction
-                    float diffuse = dot(worldNormal, uLightDir) * 0.5 + 0.5;
-
-                    // a simple specular lighting
-                    vec3 viewDir = normalize(view_position - worldPos.xyz);
-                    vec3 reflectDir = reflect(-uLightDir, worldNormal);
-                    float specular = pow(max(dot(viewDir, reflectDir), 0.0), 9.0);
-
-                    // combine the lighting
-                    brightness = diffuse * (1.0 - uMetalness) + specular * uMetalness;
+                    // the lighting is evaluated per-pixel, so simply pass the world space position
+                    // and normal along to the fragment shader
+                    worldPos = worldPosition.xyz;
+                    worldNormal = normalMatrix * localNormal;
 
                 #endif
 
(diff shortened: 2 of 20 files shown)


src/scene/shader-lib/wgsl/chunks/lit/frag/pass-forward/litForwardMain.js
```diff
@@ -22,7 +22,11 @@ fn fragmentMain(input: FragmentInput) -> FragmentOutput {
     #endif
 
     #ifdef LIT_NEEDS_NORMAL
-        dVertexNormalW = normalize(vNormalW);
+        #ifdef FLAT_SHADING
+            dVertexNormalW = getFlatNormal(vPositionW);
```
mvaligursky, line 26: [P1] Stop requiring vertex normals for flat-shaded lit variants

This replaces the fragment normal, but `LitShader.needsNormal` still makes `generateVertexShader()` declare `vertex_normal` and `vNormalW`. On WebGPU, a flat-shaded `StandardMaterial` applied to a mesh containing only positions and indices therefore creates a shader input that is absent from the vertex layout: at this head, the same normal-free mesh renders on WebGL2 but produces no frame on WebGPU, while restoring `mesh.setNormals(...)` makes WebGPU render immediately. This contradicts the new public API guarantee that flat shading does not require vertex normals. Please separate the need for a fragment-stage lighting normal from the need for the vertex-normal attribute (so the `NORMALS` input/varying is omitted for flat-shaded variants), and add a WebGPU regression covering a normal-free mesh.

mvaligursky (commented):

I found one blocking WebGPU compatibility issue in the advertised normal-free mesh path. Beyond that finding, I reviewed the public API and serialization plumbing, material define propagation/copying, GLSL/WGSL chunk integration, front-face and render-target sign handling, ShaderMaterial usage, pass behavior, and hot-path/variant impact.

Validation performed: all current CI checks are green; the focused StandardMaterial suite passes (22 tests); the StandardMaterial and ShaderMaterial examples compile and render on WebGL2 and WebGPU; and I exercised a positions/indices-only mesh on both backends. The latter exposes the inline issue below.

src/scene/shader-lib/wgsl/chunks/lit/frag/pass-forward/litForwardMain.js
mvaligursky: Good catch on the mismatch - the docs were wrong and I have fixed that in dfcb1234d. But I do not think the diagnosis of "blocking WebGPU compatibility issue" holds, so I have taken the narrower fix for now.

**Not a regression.** `attributes.vertex_normal` is gated only on `LitShader.needsNormal`, which is true for *any* lit material - lighting, reflections, specular, ambient SH. `flatShading` does not touch it. So a positions-and-indices-only mesh with an ordinary, non-flat `StandardMaterial` already declares `vertex_normal` and already produces no frame on WebGPU on `main` today. This PR does not change that path. What it did do was *advertise* it: my JSDoc claimed flat shading "does not require the mesh to supply vertex normals", which invited people into a pre-existing engine limitation. That claim is now removed from the property docs and both chunk comments, and replaced with an explicit note of the limitation.

**Why the attribute skip is not landing here.** It is not localised to the fragment normal:

- `dNormalMatrix` is declared and assigned inside `getNormal()` in `lit/vert/normal.js`, which is only included under `#ifdef NORMALS` - but `litMain.js` uses it for `vTangentW` and for `vObjectSpaceUpW` under `GGX_SPECULAR`.
- `vBinormalW` is `cross(vNormalW, vTangentW)`, so the tangent path needs the varying too.
- `ltc.js` reads `vNormalW` directly for area light billboarding.

So a correct version hoists the normal matrix out of the `NORMALS` include, switches `ltc.js` to `dVertexNormalW` (which incidentally makes area lights honour flat shading), and gates the attribute on something like `!flatShading || hasTangents || enableGGXSpecular || hasAreaLights`. That touches vertex-normal plumbing shared by every lit variant, and deserves its own PR and its own cross-backend verification rather than being appended to this one. It is queued alongside the glTF compliance work, where it is directly needed - glTF requires flat normals when `NORMAL` is absent, and the parser currently generates smooth ones instead.

**On the requested test.** A WebGPU regression is not achievable in the current suite: `test/app.mjs` builds every app on `NullGraphicsDevice`, so no backend executes in unit tests. The achievable equivalent is asserting that `vertex_normal` is absent from the generated attribute map for a flat shaded variant, in the same spirit as the existing shader-generation assertions in the ShaderMaterial suite. That belongs with the follow-up, where the behaviour actually changes.

mvaligursky: Automated re-review by Codex (GPT-5): agreed with this clarification. The WebGPU behavior is pre-existing rather than introduced by this PR, and commit dfcb1234d removes the unsupported normal-free guarantee from the public JSDoc while documenting the current lit-shader limitation in both chunks. Keeping the vertex-normal plumbing refactor as a separately tested change is reasonable, so the original P1 is addressed. I am noting the still-stale PR description separately.

mvaligursky (commented):
Automated re-review by Codex (GPT-5) of changes through `dfcb1234d1ea8f05bc5cd30fc40411daaf81ab1a`.

I reviewed the response and the three-file follow-up diff. The clarification is correct: a lit shader requiring `vertex_normal` for a normal-free mesh is pre-existing behavior on `main`, not a regression introduced by `flatShading`. The narrower fix is clean: the public JSDoc no longer promises support for meshes without normals, and both shader chunks explicitly document the current limitation. The broader vertex-normal dependency refactor is reasonably left to a dedicated, cross-backend change. The previous P1 thread is therefore resolved.

One non-code item remains: please update the PR description before merging. It still says that flat shading “does not need the mesh to supply vertex normals at all,” which now contradicts both the implementation and the updated chunk comments.

No additional code findings. The follow-up diff passes `git diff --check`, and all current CI checks are green.

PR #9192, opened 2026-08-18
Recalculate particle local bounds when the spawn volume changes

### LINKED ISSUE #811: particlesystem changing emitterExtents does not recalculates bounds
If pack8 is used, then changing `emitterExtents` is not triggering recalculation of normalization box. Calling `emitter.rebuild` or switching `emitterShape` to Sphere and then back to Box triggers recalculation.

mvaligursky:
Fixes #811.

Changing `emitterExtents` (or `emitterRadius`) at runtime did not update the particle emitter's bounds, so the mesh instance kept a stale culling AABB: the emitter can pop out of view while its particles are plainly on screen, and the editor's selection box for a particle system shows the wrong size.

The original 2016 issue is about pack8, whose lossy RGBA8 state textures normalized particle positions into that box - a stale box then visibly corrupted the positions. #8926 removed pack8 (v2.20.0) and nothing in the particle shaders depends on emitter bounds any more, so all that is left of the issue is the culling AABB, and it is still broken in three ways:

- `prevEmitterExtents` was assigned `this.emitterExtents`, a reference to the very `Vec3` the component handed the emitter. Mutating the extents in place - `ps.emitterExtents.set(...)`, `.x = n`, `.copy(v)` - compared the object with itself, so the change was never noticed.
- The snapshot was never refreshed after a recalculation, so assigning a new `Vec3` (what the component setter and the editor do) worked once and then re-ran `calculateLocalBounds` on every frame, forever.
- The check was wrapped in `if (!this.useCpu)`, added in #2059 with no stated reason, so CPU emitters never tracked the spawn volume at all.

**Changes:**
- `prevEmitterExtents` is now the emitter's own `Vec3`, and the snapshot of both it and `prevEmitterRadius` moved to the top of `calculateLocalBounds`, so every recalculation records the inputs it was calculated for. That keeps in place mutation working - which a setter driven fix cannot - and `rebuild()` gets the bookkeeping for free.
- Dropped the `useCpu` gate, so CPU emitters recalculate too.
- `_setGraphProperty` now calls `calculateLocalBounds()` after `rebuildGraphs()`. The velocity and scale graphs are its main inputs, and unlike the spawn volume they cannot be cheaply compared each frame, so assigning a curve at runtime used to leave the bounds stale with no check to catch it.

Note the bounds stay deliberately conservative: `calculateLocalBounds` pads every axis by the integrated world velocity magnitude, which is isotropic because that velocity is in world space while the bounds are local, and `worldBounds` is a union over a trail that swaps every `lifetime`, so growth applies within a frame while shrinking lags. Both predate this PR - they were simply invisible while the bounds never moved. Tightening the padding would need the node rotation folded in and is left out of here.

**API Changes:**
- Removed `ParticleEmitter#prevWorldBoundsSize`, `ParticleEmitter#prevWorldBoundsCenter` and `ParticleEmitter#worldBoundsSize`. All three were pack8 machinery, added in 2016 to feed the `inBoundsSize`/`outBoundsSize` uniforms that mapped positions into the bounds box, and have been written but never read since #8926. Undocumented, and unused by the editor.

**Performance:**
- `calculateLocalBounds` no longer runs on every frame for the lifetime of an emitter whose extents were assigned once.

### DIFF
--- a/src/framework/components/particle-system/component.js
+++ b/src/framework/components/particle-system/component.js
@@ -1836,6 +1836,10 @@ class ParticleSystemComponent extends Component {
         if (this.emitter) {
             this.emitter[name] = arg;
             this.emitter.rebuildGraphs();
+
+            // velocity and scale graphs are inputs to the local bounds, and unlike the spawn
+            // volume they cannot be cheaply compared each frame, so refresh the bounds here
+            this.emitter.calculateLocalBounds();
             this.emitter.resetMaterial();
         }
     }
--- a/src/scene/particle-system/gpu-updater.js
+++ b/src/scene/particle-system/gpu-updater.js
@@ -145,9 +145,6 @@ class ParticleGPUUpdater {
 
         emitter.swapTex = !emitter.swapTex;
 
-        emitter.prevWorldBoundsSize.copy(emitter.worldBoundsSize);
-        emitter.prevWorldBoundsCenter.copy(emitter.worldBounds.center);
-
         DebugGraphics.popGpuMarker(device);
     }
 }
--- a/src/scene/particle-system/particle-emitter.js
+++ b/src/scene/particle-system/particle-emitter.js
@@ -519,7 +519,9 @@ class ParticleEmitter {
             accumR[1] += this.qRadialSpeed2[index] * stepWeight;
             maxR = Math.max(maxR, Math.max(Math.abs(accumR[0]), Math.abs(accumR[1])));
 
-            maxScale = Math.max(maxScale, this.qScale[index]);
+            // both scale curves, as the rendered size is interpolated between them, and by
+            // magnitude, as a negative scale mirrors the particle without shrinking it
+            maxScale = Math.max(maxScale, Math.abs(this.qScale[index]), Math.abs(this.qScale2[index]));
         }
 
         if (this.emitterShape === EMITTERSHAPE_BOX) {
--- a/src/scene/particle-system/particle-emitter.js
+++ b/src/scene/particle-system/particle-emitter.js
@@ -354,12 +354,10 @@ class ParticleEmitter {
         this.worldBoundsTrail = [new BoundingBox(), new BoundingBox()];
         this.worldBounds = new BoundingBox();
 
-        this.worldBoundsSize = new Vec3();
-
-        this.prevWorldBoundsSize = new Vec3();
-        this.prevWorldBoundsCenter = new Vec3();
-        this.prevEmitterExtents = this.emitterExtents;
-        this.prevEmitterRadius = this.emitterRadius;
+        // spawn volume the local bounds were last calculated for, used to detect changes to it.
+        // Note this must be a copy, as emitterExtents can be modified in place by the user.
+        this.prevEmitterExtents = new Vec3();
+        this.prevEmitterRadius = 0;
         this.timeToSwitchBounds = 0;
 
         // simulation shaders - do not destroy those, as they're cached and shared between emitters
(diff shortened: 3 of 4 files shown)


src/framework/components/particle-system/component.js
```diff
@@ -1836,6 +1836,10 @@ class ParticleSystemComponent extends Component {
         if (this.emitter) {
             this.emitter[name] = arg;
             this.emitter.rebuildGraphs();
+
+            // velocity and scale graphs are inputs to the local bounds, and unlike the spawn
+            // volume they cannot be cheaply compared each frame, so refresh the bounds here
+            this.emitter.calculateLocalBounds();
```
mvaligursky, line 1842: [P2] Include the secondary scale curve in the refreshed bounds

This recalculation does not make `scaleGraph2` changes safe: `calculateLocalBounds()` still updates `maxScale` from only `this.qScale[index]`, while both the CPU and GPU particle paths interpolate the rendered scale through `qScale2`. On this head, creating a default emitter and assigning `component.scaleGraph2 = new Curve([0, 100, 1, 100])` leaves `localBounds.halfExtents.x` at `1`; a focused assertion that it grows past `41` fails, although all five tests added by this PR pass. Those larger particles can therefore still be culled with the stale/small AABB. Please calculate the maximum from both quantized scale curves (and use the magnitude if negative scales are supported), with a regression for the secondary curve.

mvaligursky (commented):

I found one correctness gap in the graph-driven bounds refresh. Apart from that finding, I reviewed the spawn-volume snapshot lifecycle, replacement and in-place mutation paths, CPU/GPU parity, trail/world-AABB propagation, graph setter behavior, obsolete pack8 state removal, allocation/per-frame cost, and test coverage. The snapshot copy and refresh placement are sound, the CPU gate removal is appropriate, and the removed world-bound fields have no remaining consumers.

Validation performed: all current CI checks are green, `git diff --check` passes, and the five added particle-emitter tests pass locally. I also added a focused `scaleGraph2` regression case; it fails on the current head because the recalculation still derives the particle size from only the first scale curve, as detailed inline.

src/framework/components/particle-system/component.js
mvaligursky: Good catch, fixed in 59b840bda.

Confirmed the mechanism: the rendered size is interpolated between both curves - `scale = lerp(scale, scale2, rnd) * uniformScale` in `cpu-updater.js`, and `scale += (scaleDiv * 2.0 - 1.0) * scaleDivMult * fract(rndFactor * 10000.0)` in the `particle` vertex chunk - while `maxScale` came from `qScale` alone. The padding is tight per face rather than slack (`quadXY` spans [-1, 1], so a particle's half size is `scale`), so this really was under-coverage: with `scaleGraph2` at 100 the box padded 1.

Worth noting the negative case is worse than under-sized. `maxScale` starts at `0`, so a wholly negative curve never beats the seed and the padding collapses to zero:

| | `localBounds.half.x` before | after |
| --- | --- | --- |
| `scaleGraph2 = 100` | 1.00 | 101.00 |
| `scaleGraph = -100` | 0.00 | 100.00 |

So the maximum is now taken over both curves by magnitude:

```js
maxScale = Math.max(maxScale, Math.abs(this.qScale[index]), Math.abs(this.qScale2[index]));
```

This also lines the function up with the rest of its inputs, which already pair both curves - `wVels = [qVelocity, qVelocity2]`, `lVels = [qLocalVelocity, qLocalVelocity2]`, and `accumR[0]`/`accumR[1]` from `qRadialSpeed`/`qRadialSpeed2`. Scale was the only exception.

Added two regression tests, for the secondary curve and for the negative magnitude, both of which fail on the previous head. Note the `Math.abs` on the primary curve is a behaviour change beyond the secondary curve alone: emitters using negative scale curves now get bounds where they previously got none. Happy to split that out if you would rather keep this PR to the secondary curve.

mvaligursky: Automated re-review by Codex (GPT-5): verified on `59b840bda`. The original `scaleGraph2 = 100` reproduction now grows the local bounds, and the full focused suite passes all 8 cases. Taking the maximum absolute value across both curves is the correct conservative bound for their interpolation; the negative-primary correction is the same root-cause fix and is small enough to keep in this PR. No split is needed. This finding is resolved.

mvaligursky (commented):
Automated re-review by Codex (GPT-5) of changes through `59b840bdaa203e192ec03c87e675be3546800ad9`.

The response and update fully address the previous finding. `calculateLocalBounds()` now takes the maximum magnitude across both quantized scale curves, which safely covers every value produced by interpolation and correctly handles negative scales as mirrored geometry. The new tests cover primary-curve growth, the exact secondary-curve regression, and negative magnitude. The negative-scale correction belongs in this PR because it is the same `maxScale` under-coverage defect and does not add meaningful scope or risk.

Validation: the focused particle-emitter suite passes 8/8, ESLint passes for the changed implementation and test, `git diff --check` passes, and all current CI checks are green. The original review thread has been resolved. No additional findings.

PR #9195, opened 2026-08-19
Add shadowCascadeMask to the render component

### LINKED ISSUE #9126: Improvement shadowCascadeMask
After adding support for shadowCascadeMask, both usability and performance have improved significantly. However, controlling this parameter currently requires handling each MeshInstance via script. I suggest adding this parameter to the RenderComponent, similar to castShadow, and exposing it in the editor so it can also be controlled visually.

mvaligursky:
Exposes the per-mesh directional shadow cascade mask on the render component, so selecting which cascades an entity casts into no longer requires walking its mesh instances from a script. Fixes #9126.

**Changes:**
- `RenderComponent#shadowCascadeMask` is assigned to every mesh instance the component manages, including ones created later by an asset load or a `type` change.
- The property is part of the component's property list, so it is initialized from `addComponent` data and preserved when an entity is cloned.
- Added render component tests covering the default value, initialization from component data, propagation to existing and to newly created mesh instances, and cloning.

**API Changes:**
- Added `RenderComponent#shadowCascadeMask`, defaulting to `SHADOW_CASCADE_ALL`. Combine `SHADOW_CASCADE_0` .. `SHADOW_CASCADE_3` to select individual cascades, and note that this only filters the meshes per cascade - use `castShadows` to disable shadow casting completely.

```javascript
// only cast shadows into the two cascades closest to the camera
entity.render.shadowCascadeMask = pc.SHADOW_CASCADE_0 | pc.SHADOW_CASCADE_1;
```

`ModelComponent` is deliberately left alone as it is legacy, and the gsplat component cannot support this in unified mode, where shadow casters are registered per placement and rendered through a single shared mesh instance.

### DIFF
--- a/src/framework/components/render/component.js
+++ b/src/framework/components/render/component.js
@@ -1,5 +1,5 @@
 import { Debug } from '../../../core/debug.js';
-import { LAYERID_WORLD, RENDERSTYLE_SOLID } from '../../../scene/constants.js';
+import { LAYERID_WORLD, RENDERSTYLE_SOLID, SHADOW_CASCADE_ALL } from '../../../scene/constants.js';
 import { BatchGroup } from '../../../scene/batching/batch-group.js';
 import { MeshInstance } from '../../../scene/mesh-instance.js';
 import { MorphInstance } from '../../../scene/morph-instance.js';
@@ -67,6 +67,9 @@ class RenderComponent extends Component {
 
     /** @private */
     _castShadows = true;
+
+    /** @private */
+    _shadowCascadeMask = SHADOW_CASCADE_ALL;
 
     /** @private */
     _receiveShadows = true;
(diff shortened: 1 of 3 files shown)


mvaligursky (commented):

No actionable issues found.

Reviewed the change from the merge base for correctness, public API compatibility, component initialization and clone ordering, primitive and asset-driven mesh replacement, directional-shadow semantics, batching interactions, performance, documentation, and test coverage. The component-level value is retained independently of the current mesh instances and is consistently propagated through the shared `meshInstances` setter, so later primitive recreation and render-asset mesh updates inherit it. Adding the property to the ordered system property list also covers initialization and cloning without changing existing defaults.

Local verification on the exact head commit: the focused RenderComponent test suite passed (8 tests), ESLint passed for all changed files, generated TypeScript declarations include the new getter/setter, `test:types` passed, and `git diff --check` passed. All current GitHub checks are green.

PR #9196, opened 2026-08-20
[BREAKING] Store frustum planes in a typed array

mvaligursky:
`Frustum` kept its six planes as `Plane` objects, so reading a plane walked from the frustum to a `Plane` to a `Vec3` to its components. The planes now live in a single `Float32Array`, packed as four floats per plane - the normal's x, y and z followed by the distance - in the existing right, left, bottom, top, far, near order.

**API Changes:**

The public `planes` array is gone, replaced by an accessor pair:

```javascript
// before
const normal = entity.camera.frustum.planes[0].normal;

// after
const plane = new pc.Plane();
entity.camera.frustum.getPlane(0, plane);
const normal = plane.normal;
```

- `getPlane(index, result)` writes the plane at `index` into the supplied `Plane` and returns it.
- `setPlane(index, plane)` stores a plane, normalizing it on the way in - `containsSphere` compares a plane distance against a sphere radius, so it requires unit length normals. `setFromMat4` already normalized, so the two entry points now agree.
- The packed array is available as `planeData` for internal use, tagged `@ignore`.

**Changes:**
- `containsPoint`, `containsSphere`, `copy` and `setFromMat4` read and write the packed array directly.
- `add` is unchanged in behaviour: it reads the other frustum's planes into scratch `Plane` objects and reuses the existing three-plane corner intersection. It only runs for stereo XR, a couple of times per frame, so the unpacking costs nothing there.
- The gsplat frustum culler and gsplat shadow renderer both already kept their own `Float32Array(24)` in exactly this packed layout for their cull compute shaders, so both now copy the array in one go instead of unpacking six plane objects. They keep their own buffer rather than sharing the frustum's, since each may be handed a shared scratch frustum that a later call overwrites.
- Added `test/core/shape/frustum.test.mjs`, covering the `getPlane` / `setPlane` round trip and normalization, `copy` and `clone` independence, `containsPoint`, `containsSphere`, and `add` containment over sampled points.

**Performance:**

Measured in a browser on a scene of 5072 mesh instances lit by 10 shadow casting omni lights, so 61 frustums are built per frame - one for the camera and one per shadow map face. Timings are of the culling work itself rather than whole frames, which in this scene it is far too small a share of to resolve. Each figure is the best of nine rounds, and the runs were stable to within a couple of percent.

| | before | after | |
| --- | --- | --- | --- |
| frustum cull all 5072 mesh instances | 0.245 ms | **0.199 ms** | 1.23x |
| build the frame's 61 frustums | 0.0071 ms | **0.0028 ms** | 2.5x |

The frustum build is the larger relative win, mostly from dropping the `Plane.set` plus `Vec3.normalize` call chain rather than from the storage itself, and it scales with the number of shadow casting lights.

Float32 rather than Float64 storage costs no real precision here, since `Mat4` is itself backed by a `Float32Array` and so the planes were already derived from float32 inputs. The extra rounding works out at roughly 5e-6 of the camera's distance from the origin - tens of microns at this scene's scale. It is measurable at the frustum boundary: this scene reports 3550 mesh instances visible after the change and 3548 before, a difference of two objects sitting within that band of a plane. A smaller scene of 1542 mesh instances reported an identical count either way.

### DIFF
--- a/src/core/shape/frustum.js
+++ b/src/core/shape/frustum.js
@@ -47,8 +47,6 @@ function intersectPlanes(p1, p2, p3, out) {
  * visibility of points and bounding spheres. Typically, you would not create a Frustum shape
  * directly, but instead query {@link CameraComponent#frustum}.
  *
- * @example
- * const frustum = new Frustum();
  * @category Math
  */
 class Frustum {
@@ -65,6 +63,15 @@ class Frustum {
      * @ignore
      */
     planeData = new Float32Array(24);
+
+    /**
+     * Create a new Frustum instance.
+     *
+     * @example
+     * const frustum = new Frustum();
+     */
+    // eslint-disable-next-line no-useless-constructor
+    constructor() { }
 
     /**
      * @type {Plane[]}
--- a/src/core/shape/frustum.js
+++ b/src/core/shape/frustum.js
@@ -12,6 +12,9 @@ const _c31 = new Vec3();
 const _c12 = new Vec3();
 const _corner = new Vec3();
 
+// scratch planes, used to read the planes of another frustum in add()
+const _scratchPlanes = [new Plane(), new Plane(), new Plane(), new Plane(), new Plane(), new Plane()];
+
 /**
  * Intersects three planes and writes the intersection point to out.
  *
@@ -43,27 +46,24 @@ function intersectPlanes(p1, p2, p3, out) {
  * visibility of points and bounding spheres. Typically, you would not create a Frustum shape
  * directly, but instead query {@link CameraComponent#frustum}.
  *
+ * @example
+ * const frustum = new Frustum();
  * @category Math
  */
 class Frustum {
     /**
-     * The six planes that make up the frustum.
+     * The six planes of the frustum, packed as four floats each - the normal's x, y and z followed
+     * by the plane's distance from the origin - in the order right, left, bottom, top, far, near.
+     * The normals point inwards, so a point is outside a plane when
+     * `normal.dot(point) + distance` is negative.
      *
-     * @type {Plane[]}
-     */
-    planes = [];
-
-    /**
-     * Create a new Frustum instance.
+     * This is the frustum's storage, exposed for internal use where the packed form avoids
+     * per-plane object access. Use {@link Frustum#getPlane} and {@link Frustum#setPlane} instead.
      *
-     * @example
-     * const frustum = new Frustum();
+     * @type {Float32Array}
+     * @ignore
      */
-    constructor() {
-        for (let i = 0; i < 6; i++) {
-            this.planes[i] = new Plane();
-        }
-    }
+    planeData = new Float32Array(24);
 
     /**
      * Returns a clone of the specified frustum.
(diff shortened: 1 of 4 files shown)


src/core/shape/frustum.js
```diff
@@ -43,27 +46,24 @@ function intersectPlanes(p1, p2, p3, out) {
...
-     * const frustum = new Frustum();
+     * @type {Float32Array}
+     * @ignore
      */
-    constructor() {
-        for (let i = 0; i < 6; i++) {
-            this.planes[i] = new Plane();
-        }
-    }
+    planeData = new Float32Array(24);
```
mvaligursky, line 65: [P1] Preserve `Frustum#planes` through the deprecation cycle. This has been a documented public mutable field for years, and existing applications can read or modify `camera.frustum.planes[i]`; after this minor release those accesses become `undefined` and fail at runtime. Marking the PR as breaking and documenting a migration does not provide backward compatibility. Please either retain a deprecated compatible view for the required cycle or reserve this storage change for the next major release.

mvaligursky (commented):

I found one blocking compatibility issue and one smaller public-API reporting issue. The API report also removes `Frustum.constructor()` even though runtime construction remains supported and the examples still show `new Frustum()`; please retain an explicit empty constructor so this unrelated documented API entry does not disappear.

The packed implementation itself looks correct: extraction order and normalization match the previous code, point/sphere tests use the same boundary conditions, `copy` remains independent, XR `add` retains its algorithm without per-call allocation, and both gsplat consumers make the required defensive copy.

Local verification on the exact head commit: 44 focused Frustum, Camera, and local-shadow-renderer tests passed; ESLint passed for all changed files; generated TypeScript declarations and `test:types` passed; and `git diff --check` passed. All current GitHub checks are green.

AlexAPPi:
After this update, using custom frustum culling will become quite inconvenient. I suggest adding support for testing whether a Bounding Box intersects the frustum.

AlexAPPi:
```
    public check(frustumPlanes: pc.Plane[], patchIndex: number, mask: number): number {

        // [minX, maxX, minY, maxY, minZ, maxZ]
        const box = this._boxes[patchIndex];

        for (let i = 0; i < 6; i++) {

            const bit = 0b100000 >> i;

            if ((mask & bit) === 0) continue;

            const pp = frustumPlanes[i];
            const pn = pp.normal;

            const px = pn.x;
            const py = pn.y;
            const pz = pn.z;
            const planeConstant = pp.distance;

            const ix = px > 0 ? 1 : 0;
            const iy = py > 0 ? 3 : 2;
            const iz = pz > 0 ? 5 : 4;

            const xMin = box[ix];
            const yMin = box[iy];
            const zMin = box[iz];

            const minDot = (px * xMin) + (py * yMin) + (pz * zMin);
            if (minDot < -planeConstant) {
                return -1;
            }

            const xMax = box[ix ^ 1];
            const yMax = box[iy ^ 1];
            const zMax = box[iz ^ 1];

            const maxDot = (px * xMax) + (py * yMax) + (pz * zMax);
            if (maxDot > -planeConstant) {
                mask ^= bit;
            }
        }

        return mask;
    }
```

AlexAPPi:
Updates of this scale can break a large number of components. Unfortunately, backward compatibility was not preserved at all.

mvaligursky:
it's pretty much impossible to make original direct access to internal array backwards compatible.

And yep, in a follow up PR I'm switching culling from spheres to aabbs, so there will be a test function.

PR #9197, opened 2026-08-20
Cull mesh instances by their bounding box instead of a bounding sphere

mvaligursky:
Frustum culling built a bounding sphere from a mesh instance's world AABB using `halfExtents.length()` — the circumscribing sphere, which is the loosest bound a box has — and tested that against the frustum planes. For a cube that sphere is 2.7x the box's volume; for anything elongated it is far worse, a beam of half extents 60, 1, 1 getting a sphere of radius 60. This tests the box itself.

**Changes:**
- `Frustum#containsAabb(aabb)` tests a bounding box against the frustum, using the box's extent along each plane normal: `r = |n.x|ex + |n.y|ey + |n.z|ez`, with the box outside a plane when `n·c + d <= -r`. Public, because reading the planes to hand-roll this is the common case for application side culling — it also covers the request in #9196.
- `MeshInstance#_isVisible` uses it, so camera, spot light and directional cascade culling all pick it up. The temporary `BoundingSphere` and the `halfExtents.length()` square root it needed are gone.
- The omni shadow caster classification switches from the caster's bounding sphere radius to its per-axis extents. The `1 / sqrt(1 + slope²)` that normalizes the side plane normals cancels on both sides of a box test, so it drops out and that path gets cheaper as well as tighter.
- Tests: six `containsAabb` cases, including one that checks it never reports a box visible that the sphere test rejects across 20k random boxes, and one that demonstrates it is tighter; plus the omni suite's reference moved to box semantics and a new elongated-caster case.

**API Changes:**

```javascript
frustum.containsAabb(meshInstance.aabb);  // true if it intersects or is inside
```

Returns a boolean rather than the 0/1/2 of `containsSphere`. Distinguishing "fully inside" would be cheap — a box's extent is symmetric about its centre, so the same extent serves both tests for one extra comparison per plane — but no caller needs it, and it measured 11% slower where all six planes are evaluated. Adding it later would be non-breaking.

**Performance:**

Measured on a clustered omni shadow scene of 1542 casters and 32 shadow casting omni lights, at 1600x900, medians of 40 deterministically driven frames.

| | before | after | |
| --- | --- | --- | --- |
| frustum cull the whole scene, isolated | 0.067 ms | **0.035 ms** | 1.9x |
| render time | 44.7 ms | **41.9 ms** | −6% |
| forward draw calls | 1041 | **1029** | −1.2% |
| shadow draw calls | 37554 | **35207** | −6.3% |

Culling got faster rather than slower, which is worth explaining: on top of the packed planes from #9196 the box test drops the square root and the temporary sphere write that `_isVisible` needed, and does one comparison per plane where `containsSphere` does two. So it is both tighter and cheaper.

A hash of the framebuffer after a fixed frame sequence is identical before and after — every object no longer submitted was entirely outside the frustum it was being drawn into, contributing no pixels.

How much the draw call count improves is scene dependent, and scales with how large the casters are relative to the frustum. A synthetic sweep gives 1–12% for a camera frustum and 20–55% for the small face frusta of a short range omni light, so scenes with large floors and walls lit by short range lights gain considerably more than this one, and scenes of small compact props gain less.

**Notes:**
- Light culling still uses `Light#getBoundingSphere`, which is a genuine sphere and already tight for spot cones.
- The gsplat cull compute shaders still do a sphere test — the same opportunity, in a separate area.
- Against exact SAT, the omni classification now has zero false positives at any caster size, because its slab rejection supplies exactly the three box-axis separating axes a plane test lacks. `containsAabb` on a general frustum still leaves the usual plane-test false positives near frustum corners, from 0% for small boxes up to a few percent for boxes large relative to the frustum. Adding the frustum's own AABB as a fourth rejection would remove most of those and is actually faster for large boxes, but it needs new per-frustum state, so it is left for a follow-up.

### DIFF
--- a/src/core/shape/frustum.js
+++ b/src/core/shape/frustum.js
@@ -3,6 +3,7 @@ import { Debug } from '../debug.js';
 import { Vec3 } from '../math/vec3.js';
 
 /**
+ * @import { BoundingBox } from './bounding-box.js'
  * @import { BoundingSphere } from './bounding-sphere.js'
  * @import { Mat4 } from '../math/mat4.js'
  */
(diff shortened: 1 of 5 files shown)


mvaligursky (commented):

No actionable issues found.

Reviewed the change from the merge base for geometric correctness, boundary semantics, public API design, world-AABB validity, camera/spot/directional culling behavior, the specialized six-face omni path, custom visibility and culling overrides, performance, and test coverage. `containsAabb` uses the standard conservative plane projection test with the same outside/touching convention as `containsSphere`; `MeshInstance#_isVisible` supplies its evaluated world-space AABB; and the expanded omni equations use the correct axial/lateral extent for every face and sign.

Local verification on the exact head source: 51 focused Frustum, Camera, and local-shadow tests passed, including an additional randomized anisotropic-box stress run across all six omni faces. ESLint passed for all changed files, generated TypeScript declarations and `test:types` passed, and `git diff --check` passed. All current GitHub checks are green.

Issue #643, opened 2016-06-17
Cull 6 point light projections separately


guycalledfrank:
Currently visible (at least partially) shadowed point lights always require to render the scene 6 times for each cubemap face, but we could cull each projection separately and skip some faces.

willeastcott:
It looks like a point light does now cull the scene against each face of the shadow cubemap:

https://github.com/playcanvas/engine/blob/master/src/scene/forward-renderer.js#L1749

So I'm closing this. @guycalledfrank, if I've misunderstood something, please reopen. 😄

guycalledfrank:
I think what I meant was to cull 6 point light frustums by the camera frustum to skip some faces completely.

willeastcott:
Reassigning this to you @mvaligursky since you've working in this area at the moment. Any thoughts?

mvaligursky:
I implemented this and measured it, then decided to drop it for now. Recording the findings here so the next attempt does not start from scratch. The work is on the `mv-omni-face-cull` branch, and #9199 has the full diff and numbers.

**It does help, but only in a specific situation.** All six face frusta share the light's position as their apex, so a light whose position is inside the camera frustum can never have a face rejected. The saving is confined to lights sitting outside the frustum whose range still reaches into it. Measured on the clustered omni shadows example, shadow draw calls:

- camera far away with the scene centred, 14 forward draws against 433 shadow draws: **433 → 343 (−20.8%)**
- scene pushed to the edge of the view: **628 → 544 (−13.4%)**
- from inside the scene: **628 → 620 (−1.3%)**
- pointed away from the scene: already 0, since the light itself gets culled

So it is worth the most exactly where the forward pass sees almost nothing but the lights are still visible, and very little when you are looking straight at everything. No pixel changed at any of those poses.

[...]

PR #9199, opened 2026-08-20
Cull omni shadow cube faces the camera cannot sample

mvaligursky:
Addresses #643.

An omni light renders all six faces of its shadow cube map whenever the light is visible, and the light's bounding sphere passing the camera frustum test is the only gate. A light whose range reaches into the view from off screen therefore pays for six faces even when most of them cover space no camera can see.

**Changes:**
- Each face frustum is tested against the frustum of every camera that may sample the shadow map, and faces none of them reach are skipped. All six face frusta share the light's position as their apex, so a light *inside* a camera frustum still needs every face — the saving is for lights outside a frustum whose range reaches in, which is the case that costs the most.
- The mask is only applied to lights whose shadow re-renders every frame. Face culling can only ever save per-frame work, and per-frame shadow work only comes from those lights, since a one shot update renders once ever. Restricting it that way gives up nothing and avoids retiring a one shot update after a *partial* render, which would leave faces that could never be filled in when the camera later moves to sample them. It also means the mask needs no persistent state — it is recomputed each frame, so a face that becomes visible is simply rendered on that frame.
- Skipped faces are still cleared by the render pass and just receive no casters, so a face wrongly skipped costs a missing shadow rather than leaving a previous atlas tenant's depth to be sampled.
- Culling without a camera list keeps all six faces, which is what the lightmapper gets — it bakes with its own cameras.
- `Frustum#getCorners` is extracted for the test, and `Frustum#add` now uses it rather than repeating the plane triplet intersection with its own scratch planes. `intersectPlanes` works on the packed plane data directly, which drops the scratch `Plane` and `Vec3` objects from `frustum.js`.

**Performance:**

Measured on the clustered omni shadows example by toggling the mask within a single page load, so the scene, camera and frame state are identical on both sides:

| camera | shadow draw calls | |
| --- | --- | --- |
| far away, scene centred (14 forward draws) | 433 → 343 | −20.8% |
| scene at the edge of the view | 628 → 544 | −13.4% |
| from inside the scene | 628 → 620 | −1.3% |
| pointed away from the scene | 0 → 0 | light culling already handles this |

No pixel differs at any of those poses, with an unchanged-versus-unchanged control to confirm the comparison is stable. The saving is largest exactly where the forward pass sees almost nothing but the lights are still visible, and it can never reach zero, since a light whose position is inside the frustum cannot have any face rejected.

Six tests cover the apex case, the union across several cameras, disabled cameras, the update mode gating, and the no-camera-list default.

### DIFF
--- a/src/core/shape/frustum.js
+++ b/src/core/shape/frustum.js
@@ -1,46 +1,60 @@
-import { Plane } from './plane.js';
 import { Debug } from '../debug.js';
-import { Vec3 } from '../math/vec3.js';
 
 /**
  * @import { BoundingBox } from './bounding-box.js'
  * @import { BoundingSphere } from './bounding-sphere.js'
  * @import { Mat4 } from '../math/mat4.js'
+ * @import { Plane } from './plane.js'
+ * @import { Vec3 } from '../math/vec3.js'
  */
 
-// temporaries for three-plane intersection
-const _c23 = new Vec3();
-const _c31 = new Vec3();
-const _c12 = new Vec3();
-const _corner = new Vec3();
-
-// scratch planes, used to read the planes of another frustum in add()
-const _scratchPlanes = [new Plane(), new Plane(), new Plane(), new Plane(), new Plane(), new Plane()];
+// corners of the frustum being added in add()
+const _corners = new Float64Array(24);
 
 /**
- * Intersects three planes and writes the intersection point to out.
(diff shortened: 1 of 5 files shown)


mvaligursky:
Closing for now, not because it does not work but as a priority call - the correctness surface around cached shadows and multi-camera setups is more than this is worth at the moment. The branch stays for whenever it is picked up again, and I have left the reasoning on #643.

PR #9200, opened 2026-08-20
Fix ScriptComponent clone dropping scripts awaiting their script type

### LINKED ISSUE #2796: ScriptComponent clone awaiting

mvaligursky:
## Description

`ScriptComponentSystem#cloneComponent` iterated `_scriptsIndex` with `for...in` and checked `key.awaiting`, where `key` is the script name string — so the check was always `undefined` and the whole block was dead code:

```js
for (const key in entity.script._scriptsIndex) {
    if (key.awaiting) {              // always falsy
        order.splice(key.ind, 0, key);
    }
}
```

Cloning an entity therefore silently dropped any script whose script type was not in the registry yet. When the type was later registered, the deferred sweep in `ScriptRegistry#add` created the script on the original entity, but the clone had no awaiting entry so it never got one:

```
source: _scriptsIndex = [scriptA, loadedLater(awaiting,ind=1), scriptB]
clone : _scriptsIndex = [scriptA, scriptB]
--> after loadedLater is registered
source.script.loadedLater: EXISTS
clone.script.loadedLater : MISSING
```

## Fix

Dereferencing the index entry is not enough on its own:

- **The awaiting script also needs an entry in `scripts`.** `initializeComponentData` looks every name from `order` up in `scripts` (`data.scripts[data.order[i]].enabled`), so a name in `order` but missing from `scripts` throws `TypeError: Cannot read properties of undefined (reading 'enabled')` — a hard throw out of `entity.clone()` in place of the silent drop. The data comes from `_scriptsData`, which is also where the registry sweep reads the attributes from when it eventually creates the script.

- **The awaiting script needs a well-defined slot.** It is placed directly after the script it was declared after, which is where the deferred creation puts it on the source entity (see #9202), so source and clone converge on the same order.

That needs the declaration order of *all* scripts, created or not. Neither existing record is that: `_scripts` only holds created scripts and is reordered by `move()`, and the key order of `_scriptsIndex` is not declaration order for every accepted script name (integer-like keys are enumerated first, so `['scriptA', '10', '2', 'scriptB']` enumerates as `2, 10, scriptA, scriptB`). So it is recorded explicitly as `_declarationOrder`, maintained in `create()` / `destroy()`.

Supersedes #8264, which is the one-line dereference and hits the `TypeError` above.

Fixes #2796

## Testing

Five tests in `test/framework/components/script/component.test.mjs`, all five fail on `main`:

- the clone keeps the awaiting entry, its `ind`, and its declared `enabled` / attribute data
- consecutive awaiting scripts keep their relative order
- ...including with integer-like script names
- an awaiting script is placed after the script it was declared after, when that has been moved
- the clone gets the instance, in the right slot, once the script type is registered

## Checklist

### DIFF
--- a/src/framework/components/script/component.js
+++ b/src/framework/components/script/component.js
@@ -198,6 +198,10 @@ class ScriptComponent extends Component {
         this._postUpdateList = new SortedLoopArray({ sortBy: '__executionOrder' });
 
         this._scriptsIndex = {};
+        // the names of all scripts declared on this component, created or still awaiting their
+        // script type, in the order they were declared. Unlike the key order of `_scriptsIndex`
+        // this is reliable for any script name, and unlike `_scripts` it is not affected by move()
+        this._declarationOrder = [];
         this._destroyedScripts = [];
         this._destroyed = false;
         this._scriptsData = null;
@@ -751,6 +755,10 @@ class ScriptComponent extends Component {
 
                 this._insertScriptInstance(scriptInstance, ind, len);
 
+                if (!this._scriptsIndex[scriptName]) {
+                    this._declarationOrder.push(scriptName);
+                }
+
                 this._scriptsIndex[scriptName] = {
                     instance: scriptInstance,
                     onSwap: function () {
@@ -793,6 +801,10 @@ class ScriptComponent extends Component {
 
             Debug.warn(`script '${scriptName}' is already added to entity '${this.entity.name}'`);
         } else {
+            if (!this._scriptsIndex[scriptName]) {
+                this._declarationOrder.push(scriptName);
+            }
+
             this._scriptsIndex[scriptName] = {
                 awaiting: true,
                 ind: this._scripts.length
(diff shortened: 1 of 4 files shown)


mvaligursky (commented):

No actionable issues found in this PR in isolation.

The fix correctly dereferences each `_scriptsIndex` entry, reconstructs the missing `scripts` data required by component initialization, copies awaiting attribute data, and offsets equal captured indices so consecutive awaiting declarations are not reversed during cloning. The clone then retains an awaiting entry and participates in the later registry sweep.

Local verification on the exact head: the full focused ScriptComponent suite passed (91 tests), ESLint passed for both changed files, and `git diff --check` passed. All current GitHub checks are green.

mvaligursky (commented):

Re-reviewed the changes since commit cc49f49d. No actionable issues found.

The explicit `_declarationOrder` record removes the integer-like-key enumeration problem and gives clone reconstruction the same predecessor rule used by the stacked ordering fix. It is maintained across create/destroy, protected as a reserved script name, and reconstructs moved-plus-awaiting clones consistently. The expanded focused ScriptComponent suite passed (93 tests); changed-file ESLint, diff validation, and all current CI checks are green.

PR #9203, opened 2026-08-20
Copy the ESM script attribute data when cloning a script component

mvaligursky:
## Description

Hardening, no user-visible bug today.

`cloneComponent` passed the source component's own `_attributeDataMap` entry straight into the clone's component data:

```js
const attributes = entity.script._attributeDataMap?.get(scriptName) || { };
for (const key in scriptInstance.__attributes) {
    attributes[key] = scriptInstance.__attributes[key];   // writes into the source's object
}
```

`initializeComponentData` then does `component._scriptsData = data.scripts`, so the clone's `_scriptsData` ends up holding the very object the source reads its attribute values back from — anything writing through one writes through the other.

Nothing mutates it as things stand: `__attributes` only exists on `ScriptType` instances, and those never get an `_attributeDataMap` entry (`create` only populates the map for non-`ScriptType` scripts), so the copy loop is a no-op for exactly the scripts the map holds data for. The two happen to be disjoint, which is what keeps the loop above from writing into the source component. That is a fragile thing to rely on, so copy.

## Testing

One test in `test/framework/components/script/component.test.mjs`, fails on `main`: the clone's attribute data is a distinct object, and writing to it does not change the source's.

## Checklist

### DIFF
--- a/src/framework/components/script/system.js
+++ b/src/framework/components/script/system.js
@@ -110,7 +110,8 @@ class ScriptComponentSystem extends ComponentSystem {
             const scriptName = scriptInstance.__scriptType.__name;
             order.push(scriptName);
 
-            const attributes = entity.script._attributeDataMap?.get(scriptName) || { };
+            // copy, the map holds the source component's own attribute data
+            const attributes = { ...entity.script._attributeDataMap?.get(scriptName) };
             for (const key in scriptInstance.__attributes) {
                 attributes[key] = scriptInstance.__attributes[key];
             }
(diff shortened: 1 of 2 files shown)


mvaligursky (commented):

No actionable issues found.

The shallow object copy is at the correct ownership boundary: it prevents the clone's component data from aliasing the source `_attributeDataMap` entry while preserving the existing value semantics for attribute payloads, which are subsequently interpreted or cloned according to their schemas. The regression test verifies both identity separation and mutation isolation.

Local verification on the exact head: the full focused ScriptComponent suite passed (89 tests), ESLint passed for both changed files, and `git diff --check` passed. All current GitHub checks are green.

PR #9204, opened 2026-08-20
Do not throw when an ESM script has attribute data but no schema

mvaligursky:
## Description

`ScriptComponent#initializeAttributes` warns that a schema is required for the data to be instantiated, then falls straight through and dereferences it anyway:

```js
const schema = this.system.app.scripts?.getSchema(name);
if (!schema) {
    Debug.warnOnce(`No schema exists for the script '${name}'. ...`);
}

assignAttributesToScript(this.system.app, schema.attributes, data, script);
```

So creating a `Script` class with attribute data but no registered schema throws `TypeError: Cannot read properties of undefined (reading 'attributes')` out of `ScriptComponent#create`, rather than just warning:

```js
class MyScript extends Script {
    static scriptName = 'myScript';
}
entity.script.create(MyScript, { attributes: { speed: 42 } });   // throws
```

Return after the warning, which is what the message already says happens. Note the `Debug` call is stripped from release builds while the dereference is not, so the throw is there in both.

Found while writing tests for #9203, unrelated to the clone path.

## Testing

One test in `test/framework/components/script/component.test.mjs`, fails on `main`: creating a `Script` with attribute data and no schema warns, does not throw, and leaves the script instance usable with no attributes assigned.

## Checklist

### DIFF
--- a/src/framework/components/script/component.js
+++ b/src/framework/components/script/component.js
@@ -431,10 +431,12 @@ class ScriptComponent extends Component {
                 return;
             }
 
-            // Fetch schema and warn if it doesn't exist
+            // Fetch schema and warn if it doesn't exist. Without one there is nothing to assign
+            // the data against, so there is no point in carrying on
             const schema = this.system.app.scripts?.getSchema(name);
             if (!schema) {
                 Debug.warnOnce(`No schema exists for the script '${name}'. A schema must exist for data to be instantiated on the script.`);
+                return;
             }
 
             // Assign the attributes to the script instance based on the attribute schema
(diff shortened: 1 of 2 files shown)


mvaligursky (commented):

No actionable issues found.

The early return matches the existing warning contract and is required outside debug builds as well, where the `Debug` call is stripped but the schema dereference would otherwise remain. It leaves the script instance usable, avoids fabricating an attribute layout without a schema, and preserves the stored raw data should initialization be retried after a schema is supplied.

Local verification on the exact head: the full focused ScriptComponent suite passed (89 tests), ESLint passed for both changed files, and `git diff --check` passed. All current GitHub checks are green.

Issue #5264, opened 2023-04-24
1.62 is unable to load JSON files on FB Instant


yaustar:
Discord link: https://discord.com/channels/408617316415307776/408617316415307778/1099987986470420481

On FB, JSON requests respond with `application/octet-stream` despite being a text string:

![image](https://user-images.githubusercontent.com/16639049/233970094-154512c6-7eb5-4d13-818c-457c67a0b554.png)

PR that caused this is: https://github.com/playcanvas/engine/pull/5165/

yaustar:
While this fix is being tested for 1.63, here is a project with a patch that reverts the PR: https://playcanvas.com/project/1067556/overview/json-162-patch

Copy the script json-load-patch.js and make sure Loading Type is set to be 'After Engine'
<img width="452" alt="image" src="https://user-images.githubusercontent.com/16639049/234874927-bd420990-c1c4-45c6-b98d-be43cc7a868d.png">

PR #9205, opened 2026-08-21
[BREAKING] Report malformed JSON responses as load errors

### LINKED ISSUE #1175: Invalid json in case of failed load
https://github.com/playcanvas/engine/blob/18bdc36d108031beca3fbfab948feb911eea2159/src/net/http.js#L372

If I try to load a JSON file and I have a loading error (for example "No Access-Control-Allow-Origin header is present") the xhr status is 0 and is considered as a success. I don't know if this is correct, but supposing it is, the result of the get request is an empty string that cannot be parsed:

> Uncaught SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at Object._onSuccess (playcanvas-stable.dbg.js:20204)
    at Object._onReadyStateChange (playcanvas-stable.dbg.js:20174)
    at Object.<anonymous> (playcanvas-stable.dbg.js:20138)

So a simple patch would be to check for a not empty string before parsing, but I'm not sure if it's not better to understand why the xhr status zero is considered a successful request.
I had also an issue when the network was off. The get method doesn't return an error, but a successful empty result.

mvaligursky:
Fixes #1175.

A JSON asset whose body is malformed or empty loaded "successfully" with `asset.resource` set to `null` and no error fired. Asking XHR to parse the body itself makes a parse failure unreportable: it represents a failure and the valid document `null` identically — as a null response — and the raw text cannot be read back once the response type is JSON.

**Not merged.** See *Risks* below: the fix is sound and tested, but it changes the behaviour of the most heavily used loading path in the engine, and the risk was judged not worth taking for an edge case that has been benign since 2018.

**Changes:**
- Request the raw text for JSON responses and parse it in the engine, so a malformed or empty body fails the load with a `SyntaxError` (carrying the character position) while the valid document `null` stays a successful load. Through the asset pipeline a broken body now fires `error` with `Error loading JSON resource: <url> [SyntaxError: ...]` instead of loading with a null resource. The JSON intent is checked before the content type, so a `.json` served with a binary content type (#5264) still parses. `overrideMimeType('text/plain; charset=utf-8')` forces UTF-8 decoding, since reading text would otherwise honour a declared charset where the JSON response type always decodes as UTF-8.
- Deliver one callback per network failure. A failed request reaches the engine twice, because the XHR spec's request error steps fire `readystatechange` before firing `error`, and the existing guard only covered the retry path — so with retries off (the `http.get` default) the callback ran twice with `'Network error'`.
- Report, rather than silently replace, an exception thrown while handling a successful response. It was being caught and re-delivered as a load error with the original stack discarded; it is now logged, and the request still fails so that a consumer which threw part way through its own chain is not left with a load that never completes.

**Breaking changes:**
- The XHR returned by `Http#get` / `Http#request` for a JSON request reports `responseType === 'text'`, and its `response` / `responseText` hold the unparsed JSON text rather than the parsed value. The callback — the documented way to receive the data — is unaffected. This is the observable cost of parsing in the engine: reading the raw text is the only way to tell a parse failure apart from the valid document `null`. Documented on the `options.responseType` parameter. Nothing in `src/`, `scripts/` or `utils/` captures the return value of `http.get`, so this affects external callers only.
- A malformed, truncated or empty JSON body at HTTP 200 now fails the load. Previously it resolved to `null`. Every valid JSON document is unchanged, including `null`, `false` and `0`.

**Risks:**
- *Blast radius.* Every JSON asset load in the engine goes through this path — scenes, hierarchies, templates, materials, animations, sprite/font/texture atlases, sog and gsplat-octree metadata, and `config.json`. A regression here breaks loading rather than degrading it.
- *Peak memory on large JSON, unmeasured.* Reading the text materialises the whole body as a JS string alongside the parsed object, where the JSON response type kept only bytes plus the object. For a multi-megabyte scene JSON that raises peak memory by roughly the size of the body (up to 2 bytes per character). This could not be quantified under jsdom, whose XHR is pure JS, and was not measured in a real browser.
- *Charset behaviour is untested.* The `overrideMimeType` call is verified only by hand against a live server — nise's fake XHR does not implement charset-dependent decoding, so no unit test covers it. A future refactor of the response-type handling would not notice it regressing.
- *Content previously tolerated now fails.* Anything served at a `.json` url that is not valid JSON used to resolve to `null` and may have been silently ignored by an app; it now fails the asset load.
- *`test/assets/test-malformed.json`* is deliberately invalid JSON committed to the repo. CI is green today (no JSON lint or formatter gate), but adding one later would fail on it, and the `.json` extension is load-bearing for the test.
- Residual cleanups not applied, since this is not being merged: `options.retrying` is now redundant with the per-request completion flag, and the JSON branch of the test helper's fake XHR is unreachable.


src/platform/net/http.js
```diff
@@ -610,17 +633,31 @@ class Http {
                 response = xhr.responseText;
             }
 
-            options.callback(null, response);
+            // A JSON response type is parsed by the browser, which reports a parse failure as a
+            // null response rather than as an error - so without this an unparseable body would be
+            // delivered as a successful load with a null resource. The one false positive is a body
+            // of literal `null`, which is valid JSON but indistinguishable here: the raw text is
+            // not readable once the response type is JSON.
+            if (xhr.responseType === Http.ResponseType.JSON && response === null) {
```
mvaligursky: [P1] Preserve valid JSON `null`

This condition makes every successful response containing the valid JSON document `null` fail through the public `Http` API. Existing callers can legitimately request nullable JSON, and the accompanying test explicitly locks in the breaking behavior. The ambiguity comes from asking XHR to parse the response: please retain/request the raw text and run `JSON.parse` in the engine (including the `.json`/octet-stream case), or otherwise detect malformed bodies without redefining valid JSON as an error.

mvaligursky (commented):

I traced JSON decoding, duplicate XHR terminal events, retries, concurrency-slot release, and callback exception propagation. The focused Http and AssetRegistry suites pass; changed-file ESLint, diff validation, and all CI checks are green.

One backward-compatibility issue remains: the public Http API now rejects a valid JSON document. Because the new test explicitly codifies this behavior, this would be a permanent API regression rather than an uncovered edge case.

src/platform/net/http.js
```diff
@@ -469,10 +469,19 @@ class Http {
...
+        // text cannot be read back once the response type is JSON. Request the text instead and
+        // parse it in _onSuccess, so a malformed body is reported as a SyntaxError while `null`
+        // stays a successful load. Tracked on the xhr rather than on options, which callers may
+        // reuse across requests.
         const xhr = new XMLHttpRequest();
+        xhr._jsonResponse = responseType === Http.ResponseType.JSON;
         xhr.open(method, url, options.async);
         xhr.withCredentials = options.withCredentials !== undefined ? options.withCredentials : this.withCredentials;
-        xhr.responseType = options.responseType || this._guessResponseType(url);
+        xhr.responseType = xhr._jsonResponse ? Http.ResponseType.TEXT : responseType;
```
mvaligursky, line 499: [P1] Preserve the returned XHR's JSON contract

`Http#get` and `Http#request` publicly return this XHR, so substituting `text` is observable beyond the callback: a request made with `responseType: Http.ResponseType.JSON` now immediately reports `xhr.responseType === 'text'`, and after completion `xhr.response` is the raw JSON string instead of the parsed value. Existing callers that attach their own load handler or inspect the returned request therefore break even though the engine callback still receives parsed data. A targeted probe expecting the previously requested type failed with `expected 'text' to equal 'json'`. Please keep raw-text parsing behind a loader-specific/internal path or otherwise preserve the documented returned-XHR behavior; if that cannot be preserved, this needs to be treated as an explicit breaking public API change rather than shipped as transparent error handling.

mvaligursky (commented):

Re-reviewed the changes since commit 8c4682cc. The previous finding is resolved: valid JSON `null` now succeeds while malformed and empty bodies produce parse errors, including the octet-stream case. The focused Http suite (25 tests) and AssetRegistry suite (24 tests) pass; changed-file ESLint, diff validation, and all CI checks are green.

A targeted compatibility probe found one new public-API regression in the raw-text implementation, noted inline.

mvaligursky:
Closing unmerged. The fix works and is covered by tests, but it changes the behaviour of every JSON load in the engine, and the risks listed in the description — chiefly the unmeasured peak-memory cost on large scene JSON and the untested charset handling — outweigh fixing an edge case that has been benign since 2018. Branch `mv-json-load-errors` is kept for reference if this is revisited.

Issue #155, opened 2015-01-30
Parallax Occlusion Mapping


Maksims:
We have parallax mapping, but it is the cheapest one, and does not produces great results.  
Consider implementing option for material to enable `parallaxOcclusion` with ability to specify number of `parallaxOcclusionLayers`.

PR #9207, opened 2026-08-21
Fix parallax mapping correctness and promote its example

### LINKED ISSUE #5209: Confirm our simple parallax mapping functions as expected
There were reports it does not work, one example is here:
https://forum.playcanvas.com/t/bug-in-parallax-mapping-and-possible-fix/30338

Possible implementation we have / should have for this simple solution:
https://github.com/marcusstenbeck/tncg14-parallax-mapping/blob/master/documents/Parallax%20Mapping%20with%20Offset%20Limiting%20-%20A%20Per-Pixel%20Approximation%20of%20Uneven%20Surfaces.pdf

attached as well
[Parallax Mapping with Offset Limiting - A Per-Pixel Approximation of Uneven Surfaces.pdf](https://github.com/playcanvas/engine/files/11102832/Parallax.Mapping.with.Offset.Limiting.-.A.Per-Pixel.Approximation.of.Uneven.Surfaces.pdf)

possible other options
https://twitter.com/SebAaltonen/status/1641051599878475776

Additionally, we could add an example (engine or editor) demonstrating this.

### LINKED ISSUE #7821: Applying Height Map Makes Mesh Transparent
When applying a height map to a model in the PlayCanvas Editor, the object becomes transparent and disappears from the viewport. This appears to be an engine bug, possibly the underlying shader is failing to compile.

- Open a project in the PlayCanvas Editor
- Apply a height map to a model
- Observe the model becoming transparent or vanishing

Expected behavior
The model should display correctly with the height map applied.

Image
Image

### LINKED ISSUE #8689: StandardMaterial Parallax does not work without normal map.
If material does not have a normal map, then parallax map is ignored.
Expected: still affect the albedo map regardless of normal map slot.

mvaligursky:
`StandardMaterial#heightMap` was a single-tap offset mapping implementation with a `viewDirT.z += 0.42` fudge added in 2016 and never revisited. Several parts of it were broken, and the parts that worked were about 4x too weak to see.

**Changes:**

- The parallax offset now reaches the opacity map. `getParallax()` ran *after* `getOpacity()`, so the opacity map sampled un-offset uvs while every other map of the material was offset (GLSL and WGSL).
- A height map with no normal map now works. `hasTBN` excluded `useHeights`, so no tangent frame was generated and the offset was silently zero. `LIT_HEIGHTS` now also selects the derivative-based frame, and the frame's uv falls back to the height map's uv channel.
- The uv used to build the tangent frame no longer includes the parallax offset, which was a circular dependency - that frame is what the offset is computed from. `_getUvSourceExpression` takes a new `allowParallaxOffset` argument for this.
- `dUvOffset` is no longer referenced by passes which do not compute it. It is declared under `FORWARD_PASS` only, so any other pass with a height map enabled would emit a shader that fails to compile. The renderer does not reach that combination today, so this is an invariant rather than a live bug.
- The offset is now real offset limiting (Welsh 2004), `height * normalize(viewDirT).xy`, replacing the `viewDirT.z += 0.42` fudge which both understated the effect and skewed it as the view direction changed.
- New unit tests in `test/scene/shader-lib/standard-parallax.test.mjs` covering the four fixes above, each checked to fail against the unfixed source.

**API Changes:**

- `heightMapFactor` scales the offset by `0.1` rather than `0.025`, so the default factor of 1 displaces the texture by up to 5% of a uv tile instead of ~1.25%. **Existing content using a height map will show a noticeably stronger effect** - divide the previous value by 4 to keep the old look. Useful values are now roughly 0 to 2.
- `StandardMaterial#heightMap` and `#heightMapFactor` docs now state that mid-grey is the level of the original geometry, that the offset applies to every other map of the material (so the height map needs the same tiling and offset as those maps), and what the factor means in uv units.

**Examples:**

- `materials/parallax-mapping` - promoted out of `test/` and un-hidden. Four shapes in a 2x2 grid compare diffuse only, normal map, height map alone, and normal map + height map, with a shape selector (box or sphere) so the offset is exercised on differently-oriented tangent frames, and a slider driving `heightMapFactor`. Thumbnails regenerated.

#7821 could not be reproduced - the shader compile failure theorised in that report is not reachable through the renderer, as noted above - but the parallax path it exercises is fixed here, so it is closed by this change too.

Parallax occlusion mapping (#155) is not part of this change.

Fixes #8689
Fixes #7821
Closes #5209


willeastcott:
If parallax mapping was broken, why not just upgrade it to something more modern like displacement mapping?

mvaligursky:
> If parallax mapping was broken, why not just upgrade it to something more modern like displacement mapping?

Working on that in a follow up PR, as an optional (and more expensive) option. But this is a valid option for lower end devices.

PR #9213, opened 2026-08-24
Add parallax occlusion mapping as an opt-in parallax mode

### LINKED ISSUE #155: Parallax Occlusion Mapping

mvaligursky:
`StandardMaterial#heightMap` has only ever had one implementation: a single tap of the map which shifts the uv of the other maps. That is cheap and stable, but it cannot represent deep relief - the texture slides across the surface rather than reading as depth. This adds marching the view ray through the height field as an opt-in second mode.

**Changes:**

- `StandardMaterial#parallaxMode` selects between `PARALLAX_OFFSET` (the existing single tap, still the default) and `PARALLAX_OCCLUSION`. It follows the `opacityDither` pattern - string constants plus a `STD_PARALLAX` value define - so the mode only chooses the body of `getParallax()`. Everything downstream consumes the resulting offset exactly as before.
- `StandardMaterial#parallaxSamples` (default 16) caps the height map taps taken along the ray. It is a uniform rather than a define, so changing it generates no extra shader variants. The number actually taken follows the uv distance being marched, so it accounts for the depth scale rather than the view angle alone.
- The march is a linear search for the first step below the height field, followed by a five iteration binary search which resamples inside the bracketing step. That refinement is what makes the mode usable at shallow angles: interpolating between the ends of the step instead assumes the field runs straight across it, and at a shallow angle one step can span several features of the map, which snaps the hit to the step grid and terraces. Measured against a 256 sample reference on a floor seen from low down, 16 samples goes from 11.36 error to 1.81 with the refinement - also beating 64 samples without it (5.07) at a quarter of the linear taps.
- The march distance is limited to ten times the depth. Past that the relief stops deepening rather than stretching further across the texture than any sample budget can resolve, in the same spirit as the offset limiting of the single tap mode.
- Both GLSL and WGSL. The height map is sampled at an explicit mip level inside the march, as that is non-uniform control flow, where the implicit derivatives an automatic level needs are undefined and WGSL raises a `derivative_uniformity` diagnostic.
- Five more tests in `test/scene/shader-lib/standard-parallax.test.mjs` (11 total): the march only exists in occlusion mode, the refinement resamples rather than interpolates, the explicit mip level is used, each mode generates a distinct shader (`parallaxMode` is assigned onto the shared options object rather than declared on it, so this guards the cache key), the sample uniform is only declared when marching, and the mode is ignored when no height map is assigned.

**API Changes:**

- New: `PARALLAX_OFFSET`, `PARALLAX_OCCLUSION`, `StandardMaterial#parallaxMode`, `StandardMaterial#parallaxSamples`. Defaults preserve current behaviour exactly - existing materials are untouched.
- Occlusion mode reads the height map as **depth below** the original geometry, where white sits at the level of the geometry and black is `heightMapFactor` deep. Offset mode instead pivots the surface around the mid-grey level. This is documented on `parallaxMode`, as it is a genuine difference in how the same map is interpreted, and switching modes shifts the apparent zero level.

**Examples:**

- `materials/parallax-mapping` replaces the four shape comparison example, which this covers: a brick room with a sphere resting on the floor, a spot and an omni light both casting shadows, and an environment atlas. Mode selects None, Offset or Occlusion, with Samples and Height, plus per light toggles and an environment intensity slider. Thumbnails regenerated.
- New `examples/assets/textures/bricks076a/` - the ambientCG Bricks076A set at 1K, CC0, converted to webp at 1.4MB against 5.1MB for the equivalent JPG set, with a text file recording the source in case more of the set is wanted later.

Screenshot 2026-08-24 at 11 00 27

The room is built from six inward facing planes rather than a box seen from the inside, because a tangent frame derived from screen space derivatives is mirrored on a back face while `handleTwoSidedLighting` only flips the normal of that frame, which inverts the marched relief. Worth fixing separately for the `TBN_DERIVATIVES` path, since it also silently mirrors normal mapping on two sided surfaces.

Silhouette clipping and self shadowing from the height field are not part of this. The depth buffer still sees the flat surface, so shadows and the pre-pass are unaffected by the relief.

Fixes #155


src/scene/shader-lib/glsl/chunks/standard/frag/parallax.js
```diff
@@ -1,25 +1,122 @@
...
+        float maxMarchLength = parallaxScale * parallaxMaxSlope;
+        vec2 uvSpan = marchLength > maxMarchLength ? march * (maxMarchLength / marchLength) : march;
+
+        // The step count follows the distance actually being marched rather than the view angle, so
+        // that it accounts for the depth scale too - a shallow height map needs fewer taps than a
+        // deep one at the same angle. material_parallaxSamples is the budget it is capped to. The
+        // count is deliberately not rounded up: quantizing it makes the step size jump from one
+        // whole number of steps to the next, which shows up on a curved surface as bands of constant
+        // sample count with a visible seam between them.
+        float steps = clamp(length(uvSpan) / parallaxUvPerStep, 1.0, material_parallaxSamples);
```
mvaligursky, line 53: [P2] Keep `parallaxSamples` aligned with its public tap-budget contract. The new property is documented as the *maximum number of height-map taps*, but this value only caps the linear-search iterations: the shader also takes the entry sample on line 61 and up to five refinement samples. Consequently the documented default of 16 can issue 22 taps per pixel (and a value of 1 can issue 7), which is material for a performance-tuning API in a fragment hot path. Please either include those fixed taps in the budget, or redefine/document the property and example control as linear march steps plus the fixed overhead; the WGSL copy needs the same adjustment.

mvaligursky (commented):

I found one actionable performance/API-contract issue. The overall integration is clean: the existing offset mode remains the default, the mode participates in shader identity only when relevant, the sample count remains a uniform, both GLSL and WGSL implementations match, serialization/copy metadata and public API documentation are present, and the example and tests cover the important shader-variant paths.

Validation performed: the focused parallax suite passes (11 tests), ESLint passes for all changed engine/test source files, `git diff --check` passes, and the current GitHub build, unit, lint, types, docs, API, examples, and deployment checks are green.

PR #9216, opened 2026-08-24
Alias buffer storage correctly when it is a typed array

### LINKED ISSUE #9212: Vertex Buffer constructor, jsdoc issue
Looks like a jsdoc issue, but needs a check.

The instancing example is passing a Float32Array to vertex buffer constructor, while it expects an ArrayBuffer:

https://github.com/playcanvas/engine/blob/0aaef5e432c4797ed23c1f966e2cc43232ec0613/examples/src/examples/graphics/instancing-basic.example.mjs#L140-L142

https://github.com/playcanvas/engine/blob/0aaef5e432c4797ed23c1f966e2cc43232ec0613/src/platform/graphics/vertex-buffer.js#L29-L43

mvaligursky:
Follow-up to #9215 (which closed #9212), which documented that vertex and index buffer storage can be either an `ArrayBuffer` or a typed array. Constructing a typed array from `lock()` directly only handles the former: for a typed array the copy constructor is selected instead, which ignores the byte offset and the length and copies the elements rather than viewing them, silently detaching the result from the buffer.

`IndexBuffer#writeData` was affected in released code. The GLB parser supplies typed arrays as index data, so `Mesh#setIndices` followed by `Mesh#update` on any GLB-loaded mesh wrote into a detached copy — `unlock()` then uploaded the untouched original, with no error. `Mesh#getIndices` happened to read correct values, as the copy holds the same elements. The mesh collision shape and mesh particle paths have the same problem for vertex buffers built from typed array data.

**Changes:**
- Add `BufferUtils.createStorageView(buffer, arrayType, byteOffset, length)`, which resolves the underlying `ArrayBuffer` and base offset from either storage form, confines the default view length to the buffer's own range rather than running to the end of the underlying buffer, and asserts on the one case that cannot be satisfied at all - a misaligned sub-view being asked for a wider element type.
- Use it in `IndexBuffer#writeData` and `#readData`, replacing the `_lockTypedArray` helper and its per-format ternary.
- Use it in the mesh collision shape path (`CollisionComponentSystem`) and the mesh particle path (`ParticleEmitter`).
- `VertexIteratorAccessor` has the same problem but is left as is, as it takes the storage rather than the buffer, and every buffer the engine iterates today is `ArrayBuffer` backed. It gains a `TODO` describing the failure and an assert that the view it built actually aliases the requested memory at the requested offset, so the failure is loud rather than silent. Testing the result rather than the input catches the non-zero offset case too, which the copy constructor collapses to zero.

**API Changes:**
None - `BufferUtils` is `@ignore`.

**Tests:**
New suites for `BufferUtils` (plain `ArrayBuffer`, whole typed array, sub-view into a larger buffer, byte offset on top of the storage offset, explicit length, element type distinct from the storage type), for `IndexBuffer#writeData`/`#readData` (two of which fail without this change), and for the `VertexIterator` assert.

### DIFF
--- a/src/platform/graphics/vertex-iterator.js
+++ b/src/platform/graphics/vertex-iterator.js
@@ -84,6 +84,8 @@ class VertexIteratorAccessor {
      * Create a new VertexIteratorAccessor instance.
      *
      * @param {ArrayBuffer} buffer - The vertex buffer containing the attribute to be accessed.
+     * Note that this must be an {@link ArrayBuffer} - see the limitation described in the
+     * constructor body.
      * @param {object} vertexElement - The vertex attribute to be accessed.
      * @param {string} vertexElement.name - The meaning of the vertex element. This is used to link
      * the vertex data to a shader input. Can be:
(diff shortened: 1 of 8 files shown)


src/platform/graphics/vertex-iterator.js
```diff
@@ -134,12 +136,27 @@ class VertexIteratorAccessor {
         this.numComponents = vertexElement.numComponents;
 
         // create the typed array based on the element data type
+        // TODO: this only handles an ArrayBuffer. VertexBuffer storage can also be a typed array
```
mvaligursky, line 139: [P1] Fix typed-array-backed VertexIterator access instead of only asserting. `VertexBuffer` explicitly accepts and retains typed-array storage, and `VertexIterator` is a public API; with that valid combination this constructor selects the typed-array copy overload, so iterator writes mutate a detached copy and `end()` uploads the untouched original. `Debug.assert` is stripped from production builds, leaving the same silent data loss in releases (and reads can also be wrong when the storage and attribute types differ). Please pass the `VertexBuffer` into the accessor and use the new `BufferUtils.createStorageView` for both layouts, then make the regression test verify reads/writes through typed-array storage and sub-views rather than codifying the assertion.

mvaligursky (commented):

I found one blocking correctness gap in the same storage-aliasing path. The new utility and the IndexBuffer, collision, and particle integrations are otherwise clean: they preserve sub-view offsets, confine default view lengths to the logical buffer, cover differing storage/element types, and avoid allocations beyond the required view objects.

Validation performed: focused affected tests **14 passing**, changed-file ESLint, and `git diff --check` all pass. All GitHub build, unit, lint, types, docs, API, examples, and deployment checks are green; however, the new VertexIterator test currently verifies the diagnostic for the known broken behavior rather than a working typed-storage path.

Issue #9059, opened 2026-07-10
WebGPU: allow sampling multisampled render targets (custom MSAA resolves for HDR color + depth)


SashaRX:
**Area:** graphics · WebGPU

#### Problem

On WebGPU there is currently no supported way to bind a multisampled render
target as a texture in a user shader. Two things block a custom MSAA resolve:

1. The multisampled COLOR buffer is created with `RENDER_ATTACHMENT` usage only —
   there is no option to add `TEXTURE_BINDING`, so its samples can never be read
   in a shader. See `WebgpuRenderTarget.initColor`
   (`src/platform/graphics/webgpu/webgpu-render-target.js`):

   ```js
   const multisampledTextureDesc = {
     size: [width, height, 1], dimension: '2d', sampleCount: samples, format,
     usage: transientColor
       ? GPUTextureUsage.RENDER_ATTACHMENT | GPUTextureUsage.TRANSIENT_ATTACHMENT
       : GPUTextureUsage.RENDER_ATTACHMENT   // <- no TEXTURE_BINDING option
   };
   ```

2. `BindGroupFormat`/`BindTextureFormat` cannot express a multisampled texture
   binding. `BindTextureFormat`'s constructor has no `multisampled` parameter, and
   `WebgpuBindGroupFormat` hardcodes it off
   (`src/platform/graphics/webgpu/webgpu-bind-group-format.js`):

   ```js
   const multisampled = false;   // <- never true
   ...
   texture: { sampleType: gpuSampleType, viewDimension, multisampled }
   ```

   So even though the WGSL reflector already recognises `texture_multisampled_2d`
   and `texture_depth_multisampled_2d`
   (`webgpu-shader-processor-wgsl.js`, `textureBaseInfo`), a shader that binds one
   gets a bind-group layout that says `multisampled: false` → validation error.

The engine already does exactly this resolve internally — `WebgpuResolver`
(`src/platform/graphics/webgpu/webgpu-resolver.js`) binds
`texture_depth_multisampled_2d`, does `textureLoad(img, coord, 0u)`, and runs a raw
`layout: "auto"` pipeline that bypasses `BindGroupFormat`. The capability exists;
it just isn't reachable from user code. (Internal MS *depth* buffers even opt into
`TEXTURE_BINDING` already — `else if (samples > 1)` in the depth init path — but MS
*color* never does, and neither is bindable through a `BindGroupFormat`.)

This is the failure class the WebGPU spec deliberately pushed to userland: there is
[...]

PR #9217, opened 2026-08-24
Add samples option to Texture for multisampled textures on WebGPU

mvaligursky:
Adds first-class multisampled textures: `new Texture(device, { samples: 4 })` creates a texture that can be used as a render target attachment and whose individual samples can be read in a shader using `textureLoad` (WebGPU only). This is the first part of the render-target surface designed in #9059 (see https://github.com/playcanvas/engine/issues/9059#issuecomment-5368086964); a follow-up adds `RenderTarget` support for multisampled attachments with per-attachment resolve control. Refs #9059.

**Changes:**
- `Texture` accepts a `samples` option (WebGPU only, ignored with a warning on other devices, normalized to the device's supported sample count). Multisampled textures are validated to be 2D, non-array, non-storage, with no mipmaps or initial data, and to use a format that supports multisampling.
- `pixelFormatInfo` gains multisample-capability flags per the WebGPU format capability table (8/16-bit unorm / float / integer formats, `r32float`, `rgb10a2` variants, depth formats), with an internal `isMultisampleCapablePixelFormat` helper. Multisample and resolve capability are independent - integer formats and `r32float` can be multisampled but not resolved, which the follow-up RenderTarget PR supports via optional per-attachment resolve.
- `Texture#copy` supports copies between multisampled textures with matching sample counts (full-texture copies only, per WebGPU requirements), and rejects copies between differing sample counts.
- WebGPU texture creation passes the sample count; usage stays `RENDER_ATTACHMENT | TEXTURE_BINDING | COPY_SRC | COPY_DST`, the legal set for multisampled textures.

**API Changes:**
- `Texture` constructor option `samples` (number, defaults to 1).
- `Texture#samples` getter.

### DIFF
--- a/src/platform/graphics/constants.js
+++ b/src/platform/graphics/constants.js
@@ -1095,7 +1095,7 @@ export const pixelFormatInfo = new Map([
     [PIXELFORMAT_DEPTH,         { name: 'DEPTH', size: 4, msaa: true }],
     [PIXELFORMAT_DEPTH16,       { name: 'DEPTH16', size: 2, msaa: true }],
     [PIXELFORMAT_DEPTHSTENCIL,  { name: 'DEPTHSTENCIL', size: 4, msaa: true }],
-    [PIXELFORMAT_111110F,       { name: '111110F', size: 4 }],
+    [PIXELFORMAT_111110F,       { name: '111110F', size: 4, msaa: true }],
     [PIXELFORMAT_SRGB8,         { name: 'SRGB8', size: 4, ldr: true, srgb: true }],
     [PIXELFORMAT_SRGBA8,        { name: 'SRGBA8', size: 4, ldr: true, srgb: true, msaa: true }],
     [PIXELFORMAT_BGRA8,         { name: 'BGRA8', size: 4, ldr: true, msaa: true }],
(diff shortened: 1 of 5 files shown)


src/platform/graphics/texture.js
```diff
@@ -307,7 +323,29 @@ class Texture {
...
         this._premultiplyAlpha = options.premultiplyAlpha ?? false;
 
-        this._mipmaps = options.mipmaps ?? true;
+        // multisampled (MSAA) texture, WebGPU only - can only be rendered into, and read in
+        // shaders using textureLoad
+        const requestedSamples = options.samples ?? 1;
+        if (requestedSamples > 1) {
+            if (graphicsDevice.isWebGPU) {
+                // WebGPU only supports sample counts of 1 or 4
+                this._samples = graphicsDevice.maxSamples;
```
mvaligursky, line 332: [P2] Account for the extra samples in texture VRAM size. `Texture#gpuSize` still returns only `calcGpuSize(...)`, and `WebgpuTexture.uploadData()` copies that value into `_gpuSize` / `device._vram`. Consequently every new 4x texture is reported as consuming only one sample's storage in `AppStats.vram`, texture tracing, and the per-texture memory listing. Please multiply the base level size by `_samples` (with a regression assertion for a 4x texture), or otherwise include the sample count in the accounting path.

src/platform/graphics/constants.js
```diff
@@ -1165,6 +1165,21 @@ export const isIntegerPixelFormat = (format) => {
...
+ * WebGPU (per the WebGPU texture format capabilities table, core features). Note that support for
+ * hardware resolve is a separate, narrower capability: integer formats and
+ * {@link PIXELFORMAT_R32F} are multisample-capable but cannot be resolved, and depth formats have
+ * no hardware resolve at all.
+ *
+ * @param {number} format - The pixel format.
+ * @returns {boolean} True if the format supports multisampling.
+ * @ignore
+ */
+export const isMultisampleCapablePixelFormat = (format) => {
```
mvaligursky: [P3] Include feature-gated formats in this capability check. This static flag covers only the core table, but this backend already requests and exposes `rg11b10ufloat-renderable` and `texture-formats-tier1`. On devices where those are enabled, `PIXELFORMAT_111110F`, `PIXELFORMAT_RG8S`, and `PIXELFORMAT_RGBA8S` are valid multisampled render-attachment formats, yet this helper returns false and the constructor emits an incorrect debug assertion. Please make validation device-aware (or explicitly add the enabled optional cases) and test at least the existing `textureRG11B10Renderable` path.

mvaligursky (commented):

The WebGPU descriptor and multisampled copy constraints match the current specification, and the public API is documented clearly. I found two gaps in the surrounding accounting/capability handling. The deferred `RenderTarget` attachment integration is understood to be follow-up work and is not treated as a finding here.

Validation performed: focused affected tests **15 passing**, changed-file ESLint, and `git diff --check` all pass. All GitHub build, unit, lint, types, docs, API, examples, size, and deployment checks are green.

PR #9219, opened 2026-08-24
Sample the parallax march at an explicit mip level

mvaligursky:
The occlusion mode added in #9213 sampled the height map at mip 0 for every tap of the march. That is the sharpest data available, but it is the wrong level once the surface is minified: it aliases, it defeats the texture cache, and it makes the shader resolve detail that cannot be seen. This derives the level the march should read, uses it for every tap, and lets the same number drive the step count and a fade.

**Changes:**

- The mip level is computed once before the loop from the derivatives of the height map's uv in texels. Every tap - the entry sample, the linear search and the five refinement halvings - reads that level rather than 0. Taking it before the loop keeps it in uniform control flow, and the height map's own uv expression excludes the parallax offset, so the derivatives are well defined.
- The step count now counts detail rather than distance in uv: `marchTexels = length(uvSpan * heightMapSize) / exp2(lod)`, and one step per couple of texels of the mip being read. A coarser mip holds proportionally less detail, so the march gets cheaper as the surface recedes. `parallaxSamples` is still the cap.
- The offset fades out once the march covers barely a texel of that mip, and the march is skipped entirely when the fade reaches zero. There is nothing left to displace at that point, so this removes both the shimmer and the cost.
- `parallaxUvPerStep` is replaced by `parallaxTexelsPerStep`, which is only a meaningful unit now that the texture size is known.

WGSL uses `textureDimensions`; GLSL uses `textureSize`, which is core in GLSL ES 3.00 and already used by several chunks.

**Examples:**

- `materials/parallax-mapping` builds its sphere from `SphereGeometry` at 128 bands instead of using the primitive, whose sixteen bands leave the silhouette and shading faceted - the facet edges shimmer as the offset steps across them. Thumbnails regenerated.

**Measured**, painting each pixel by the number of height map taps it took, over a sphere at increasing distance:

| distance | sphere radius | mean taps | peak taps |
| --- | --- | --- | --- |
| 4 | 335 px | 10.75 | 20.6 |
| 16 | 84 px | 6.96 | 9.8 |
| 64 | 21 px | 1.87 | 6.9 |
| 128 | 10 px | 0.04 | 3.7 |
| 256 | 5 px | 0 | 0 |

The falloff is monotonic and reaches zero, so distant surfaces stop marching altogether rather than settling at a fixed cost. Note that most of that comes from the fade rather than the step count: the step count alone shifts the total by about one tap over the three doublings a room-sized scene spans.

One thing tried and rejected: skipping the refinement halvings once a step covers only a couple of texels, on the grounds that the bracket midpoint is then already sub-texel. Measured against a 256-sample reference it cost 4.6 against a 0.2 baseline, on an effect whose whole magnitude is around 20, and there was no threshold that both fired and stayed cheap. The refinement is what carries the accuracy, not the linear step density, so it always runs - there is a comment at the loop saying so.


src/scene/shader-lib/glsl/chunks/standard/frag/parallax.js
```diff
@@ -44,64 +49,89 @@ void getParallax() {
...
-                break;
+        // The mip level the march reads. Every tap uses an explicit level because the loop below is
+        // non-uniform control flow, where the implicit derivatives an automatic level needs are
+        // undefined, and because it says how much detail the map can resolve here - which sets both
+        // the step count and the fade.
+        vec2 heightMapSize = vec2(textureSize({STD_HEIGHT_TEXTURE_NAME}, 0));
+        vec2 uvTexels = {STD_HEIGHT_TEXTURE_UV} * heightMapSize;
+        vec2 texelsDx = dFdx(uvTexels);
+        vec2 texelsDy = dFdy(uvTexels);
+        float lod = max(0.0, 0.5 * log2(max(dot(texelsDx, texelsDx), dot(texelsDy, texelsDy))));
```
mvaligursky, line 60: [P3] Add a regression that distinguishes the derived LOD from the old fixed level. The existing `samples the height map with an explicit mip level` test only checks that `texture2DLod` exists and `texture2DBias` does not, so it already passed when every tap hard-coded `0.0`; it would also keep passing if this derivative calculation, the fade/skip, or one of the search/refinement `lod` arguments regressed. Please assert that generated GLSL and WGSL contain the footprint calculation and that the entry, march, and refinement taps all consume the derived level (plus the fade/zero-tap path).

mvaligursky (commented):

I found no rendering-correctness or backend-parity blocker. The footprint-to-LOD calculation is standard, remains outside divergent control flow, is mirrored correctly in GLSL/WGSL, and the same explicit level is used by the entry, linear-search, and refinement taps. The continuous fade and early skip are coherent with the measured performance goal, and the regenerated thumbnail looks clean. I left one non-blocking test-quality finding because the current regression test does not distinguish this implementation from the old fixed-LOD shader.

Validation performed: focused parallax tests **11 passing**, changed shader-file ESLint, and `git diff --check` pass. All GitHub build, unit, lint, types, docs, API, examples, size, and deployment checks are green.

PR #9221, opened 2026-08-25
Support multisampled color buffers and custom resolves on RenderTarget

mvaligursky:
`RenderTarget` accepts multisampled textures (created with `samples` > 1, added in #9217) as color buffers, rendering directly into their samples (WebGPU only). An optional per-attachment resolve buffer receives the hardware resolve; without one, the samples are stored and can be read in a shader using `textureLoad` - enabling custom resolves the hardware cannot express, and multisampling of formats that cannot be hardware-resolved at all (integer formats, `r32float`). This is part 2 of the design in #9059 (https://github.com/playcanvas/engine/issues/9059#issuecomment-5368086964); a follow-up adds multisampled depth buffers. Refs #9059.

```javascript
const msColor = new Texture(device, { width, height, format: PIXELFORMAT_RGBA16F, samples: 4 });
const renderTarget = new RenderTarget({ colorBuffer: msColor, depth: true });
// samples are stored - a later pass reads them using textureLoad on a texture_multisampled_2d
```

**Changes:**
- `RenderTarget` explicit multisampled mode: when the color buffers are multisampled textures, the render target renders directly into them and the sample count is inferred from the textures. New `resolveBuffer` / `resolveBuffers` options provide optional per-attachment hardware resolve targets, validated for matching format and dimensions, single-sampled, resolve-capable format, and consistent sample counts across MRT attachments. `resize` and `destroyTextureBuffers` cover the resolve buffers.
- `RenderPass` per-attachment default color ops: attachments with a resolve buffer resolve and discard the samples (matching the implicit multisampled path); attachments without one store them. The defaults remain overridable per pass, and a multisampled attachment that is neither stored nor resolved warns (its output would be discarded).
- WebGPU implementation: the user texture is the render pass view (no internal multisampled buffer is allocated, and the user texture is not owned or destroyed by the render target), the resolve buffer is wired as the resolve target, and the per-pass resolve flag attaches / detaches it.
- `copyRenderTarget` validates equal sample counts for color copies - copies between multisampled targets work, copies between multisampled and single-sampled targets error with a pointer to use a resolve instead.
- `pixelFormatInfo` gains hardware-resolve capability flags (`isMultisampleResolveCapablePixelFormat`) - a narrower capability than multisampling: integer formats and `r32float` can be multisampled but not resolved.
- Expanded the `RenderTarget` class documentation: usage, MSAA, and a section on explicit multisampled buffers with custom resolves.

**API Changes:**
- `RenderTarget` constructor options `resolveBuffer` and `resolveBuffers`.
- `RenderTarget#resolveBuffer` getter and `RenderTarget#getResolveBuffer(index)`.
- `RenderTarget` `colorBuffer` / `colorBuffers` options accept multisampled textures (WebGPU).

**Examples:**
- New `graphics-advanced` category (further examples to migrate to it separately), with a new `custom-msaa-resolve` example: a split-screen comparison of the hardware box resolve against a custom per-sample tonemapped resolve on HDR content, with a pixel-scale control to magnify the per-pixel differences. Thumbnails included.

Screenshot 2026-08-25 at 11 11 24


src/platform/graphics/render-target.js
```diff
@@ -227,12 +323,25 @@ class RenderTarget {
...
+        // samples. When the provided color buffers are multisampled textures (explicit
+        // multisampled attachments, WebGPU only), the render target renders directly into them
+        // and the sample count is inferred from the textures.
         const { maxSamples } = this._device;
-        this._samples = Math.min(options.samples ?? 1, maxSamples);
-        if (device.isWebGPU) {
-            // WebGPU only supports values of 1 or 4 for samples
-            this._samples = this._samples > 1 ? maxSamples : 1;
+        const firstColorBuffer = options.colorBuffer ?? options.colorBuffers?.[0];
+        const explicitMsaa = (firstColorBuffer?.samples ?? 1) > 1;
```
mvaligursky: [P2] Validate mixed sample counts regardless of attachment order

`explicitMsaa` is derived only from attachment 0, and the same-sample-count validation below runs only when that value is true. As a result, `colorBuffers: [singleSampled, multisampled]` takes the implicit path, keeps `RenderTarget.samples` at the option/default value, emits no same-count assertion, ignores any `resolveBuffers`, and later gives WebGPU attachment views with different sample counts, which makes `beginRenderPass` fail validation. The added test covers only `[multisampled, singleSampled]`. Please validate sample-count consistency independently of `explicitMsaa` (or validate all attachments before selecting the mode) so the reversed ordering is rejected with the intended diagnostic as well.

mvaligursky (commented):

Reviewed the exact head `4324f3a0abcae99fa9c14e57f149cea5bc8c4496` from scratch, including public API and generated declarations, RenderTarget ownership/lifecycle and resize behavior, per-pass store/resolve semantics, WebGPU attachment setup and copy constraints, MRT behavior, format capability checks, example code/assets, tests, discussion, and CI.

Verification: focused unit tests pass (43/43), changed source/test files pass ESLint, `npm run build:types` and `npm run test:types` pass, `git diff --check` passes, and all current GitHub checks are green. I also ran the deployed custom-MSAA-resolve example on WebGPU; both hardware and custom resolve paths rendered successfully with no engine/WebGPU console errors.

I found one actionable MRT validation gap inline. The core valid configurations and the new example otherwise look clean and well documented.

mvaligursky (commented):
Automated follow-up PR review by Codex (GPT-5).

Re-reviewed commit `86e72f99d642ba57b31e56a99e00f6d6b1b84f79` against the previously reviewed head. The prior P2 concern is resolved: explicit-MSAA detection now considers every supplied color attachment, so `[single-sampled, multisampled]` selects the validation path and reports the mixed sample counts instead of reaching an invalid WebGPU render pass. The added regression test exercises that exact ordering and also verifies the inferred render-target sample count.

I checked the constructor ordering, single-buffer and MRT paths, resolve-buffer gating, transient handling, and allocation impact of the change and found no new actionable issues. Verification: focused graphics tests pass (44/44), the changed source and test files pass ESLint, `git diff --check` passes, and all current GitHub CI/deployment checks are green.

PR #9223, opened 2026-08-25
Add a parallax terrain example, and tighten the occlusion march limit

mvaligursky:
### New example: `materials/parallax-terrain`

Screenshot 2026-08-25 at 14 08 37

A cpu generated displaced grid with a single ground material over all of it, so the marched parallax mode is seen across a wide range of distances and viewing angles at once - the case it earns its cost in over the single tap one. A directional light with cascaded shadows and 40 trees give the relief something to be shadowed by, and the camera is the hybrid `CameraControls` (orbit + fly) so the whole terrain can be visited.

Controls: parallax **Mode / Samples / Height**, **Shadow Cascades** (filtering, count) and **Soft Shadows** (falloff, samples, blocker samples).

Textures are [ambientCG Ground092C](https://ambientcg.com/view?id=Ground092C), CC0, 1K converted to webp - 0.7 MB against 3.3 MB for the equivalent jpg maps. The tree model is the existing `low-poly-tree.glb`, credited in the example header.

### Behaviour change: `parallaxMaxSlope` 10 → 5

⚠️ **This is a visible change for any material using `PARALLAX_OCCLUSION`.**

The constant caps the march at that multiple of the relief depth, which bounds how far the lookup slides along the surface at a grazing angle. That slide is what makes the relief appear to swim as the camera moves: parallax fakes depth by displacing the texture lookup by `depth × tan(θ)` while the pixel itself never moves, and `d(tan θ)/dθ = sec²θ` is ~13 at 74° off the normal, so a small camera move slides the texture a long way.

Measured on this terrain, on a slope at ~74°, with the relief only 0.25 world units deep:

| `parallaxMaxSlope` | lateral slide (world) | variation over a 1.8 unit camera move |
|---|---|---|
| 10 (before) | 0.81 → 0.55 | **0.26** |
| 3 | 0.65 → 0.52 | 0.17 |
| 1.5 | 0.44 → 0.43 | 0.03 |

5 was picked by eye as the point where grazing relief still reads without the visible swim. The trade is less apparent depth at extreme grazing angles in exchange for a stable image. Ruled out as causes by measurement, for the record: the distance fade (forcing it to 1 changed the result by 0.0001), the explicit mip level (lod 0 vs +2 moves the depth ~7%), and mesh tessellation (nothing in the march depends on triangle density).

Verified on WebGL2 and WebGPU, all five shadow filtering modes and cascade counts 1-4.


Maksims:
Feels like there is a height fade out based on fragment distance is required, to fade into no-parallax based on proximity.

mvaligursky:
> Feels like there is a height fade out based on fragment distance is required, to fade into no-parallax based on proximity.

Both number of taps for parallax and number of taps for shadows fade out based on pixel size .. so in the distance the parallax is cheaper all the way to disabled.

See this https://x.com/ValigurskyM/status/2091916601293496351

PR #9224, opened 2026-08-25
fix: disable blending when generating SOG centers into an integer render target

slimbuck:
## Problem

On WebGPU, picking a SOG scene (e.g. double-clicking in SuperSplat Viewer) produces Dawn validation errors and a broken pick:

```
Blending is enabled but color format (TextureFormat::RGBA32Uint) is not blendable.
 - While validating targets[0] framebuffer output.
 - While encoding [RenderPassEncoder "SogGenerateCenters-PassEncoder RT:sogCentersTexture"].SetPipeline(...)
```

`GSplatSogData.generateCenters()` lazily renders splat centers into an RGBA32U texture via `RenderPassQuad`. Since #8535, `RenderPassQuad` deliberately preserves the device blend state (callers own it), so the pass inherits the alpha blending left over from splat rendering. WebGPU rejects blending on integer formats, the pipeline is invalid, the pass silently no-ops, and the subsequent centers readback returns garbage — so picking on SOG scenes under WebGPU fails entirely. WebGL ignores blend state on integer targets, which is why only WebGPU surfaces this.

## Fix

Set `BlendState.NOBLEND` on the device before rendering the quad, matching the caller contract established in #8535 and the pattern used by other `drawQuadWithShader`/`RenderPassQuad` callers.

### DIFF
--- a/src/scene/gsplat/gsplat-sog-data.js
+++ b/src/scene/gsplat/gsplat-sog-data.js
@@ -3,6 +3,7 @@ import { Quat } from '../../core/math/quat.js';
 import { Vec3 } from '../../core/math/vec3.js';
 import { Vec4 } from '../../core/math/vec4.js';
 import { GSplatData } from './gsplat-data.js';
+import { BlendState } from '../../platform/graphics/blend-state.js';
 import { RenderTarget } from '../../platform/graphics/render-target.js';
 import { Texture } from '../../platform/graphics/texture.js';
 import {
@@ -403,6 +404,9 @@ class GSplatSogData {
         renderPass.init(renderTarget);
         renderPass.colorOps.clear = false;
         renderPass.depthStencilOps.clearDepth = false;
+        // RenderPassQuad preserves the device blend state; blending must be off
+        // for the integer render target
+        device.setBlendState(BlendState.NOBLEND);
         renderPass.render();
         quad.destroy();
 
--- a/src/scene/gsplat/gsplat-sog-data.js
+++ b/src/scene/gsplat/gsplat-sog-data.js
@@ -3,7 +3,6 @@ import { Quat } from '../../core/math/quat.js';
 import { Vec3 } from '../../core/math/vec3.js';
 import { Vec4 } from '../../core/math/vec4.js';
 import { GSplatData } from './gsplat-data.js';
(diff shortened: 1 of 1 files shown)


src/scene/gsplat/gsplat-sog-data.js
```diff
@@ -403,6 +404,9 @@ class GSplatSogData {
         renderPass.init(renderTarget);
         renderPass.colorOps.clear = false;
         renderPass.depthStencilOps.clearDepth = false;
+        // RenderPassQuad preserves the device blend state; blending must be off
+        // for the integer render target
+        device.setBlendState(BlendState.NOBLEND);
```
mvaligursky: Suggestion: use `RenderPassShaderQuad` instead, and then no explicit blend call is needed at all.

It handles this declaratively, and not just for blending: it exposes `blendState` (defaulting to `BlendState.NOBLEND`), `depthState`, `cullMode`, `frontFace` and stencil as pass properties, and feeds all of them into `setDrawStates()` itself. So the pass guarantees every render state — with `RenderPassQuad` we'd have to keep setting each one on the device by hand to be sure. It accepts a pre-built `quadRender` (as `GSplatProcessor` does), so it's a drop-in here:

```js
        const quad = new QuadRender(shader);
        const renderPass = new RenderPassShaderQuad(device);
        renderPass.quadRender = quad;
        renderPass.name = 'SogGenerateCenters';
        renderPass.init(renderTarget);
        renderPass.colorOps.clear = false;
        renderPass.depthStencilOps.clearDepth = false;
        renderPass.render();
        quad.destroy();
```

with the imports swapped accordingly - drop `BlendState`, and replace the `RenderPassQuad` import with:

```js
import { RenderPassShaderQuad } from '../../scene/graphics/render-pass-shader-quad.js';
```

(couldn't attach this as a one-click suggestion, as the `const renderPass` line is just outside the diff hunk)

slimbuck: Done in 9d258a7 — switched to `RenderPassShaderQuad` with the pre-built `quadRender` (same pattern as `GSplatProcessor`) and dropped the explicit blend call and `BlendState` import. Agreed it's the better shape: every draw state is now guaranteed by the pass itself.

mvaligursky (approved):

PR #9225, opened 2026-08-25
Add soft self shadowing of the height field to parallax occlusion mapping

mvaligursky:
Relief lit by a directional light now shadows itself. `StandardMaterial#parallaxShadowSamples` marches the height map towards the light from the point the view ray hit, and **zero (the default) takes the march out of the generated shader entirely** - so the count both budgets the effect and switches it, in the same shape as `alphaTest`.

### Soft, not a first-hit test

The march accumulates how far the field stands above the light ray, weighted by distance from the hit, rather than stopping at the first blocker. More taps buy a smoother penumbra which hardens towards contact, at a fixed cost for a given budget.

`parallaxShadowHardness` is 16 (a chunk constant, not API). Measured mean darkening on rippled ground is 4.2 / 7.4 / 10.1 / 12.1 at 4 / 8 / 16 / 32, and **nothing clips to black at any of them** because the ambient fills the shadow - it is an authority knob rather than a clipping risk, and returns flatten after 16.

### Directional lights only

Not just a cost decision. Directional lights are never clustered - `world-clusters.js` excludes `LIGHTTYPE_DIRECTIONAL` and the renderer keeps them out of `localLights` - so they always run the non-clustered `evaluateLight{i}` path. That means one file touched, no changes to `clusteredLight.js`, and identical behaviour whether clustered lighting is on or off. Extending it to clustered omni/spot would make per-pixel cost scale with light count, which deserves its own opt-in.

The hook sits **outside** `#ifdef LIGHT{i}CASTSHADOW`: relief shadowing needs no shadow map, so a light with `castShadows: false` still gets it.

### What the view march hands over

Two scalars in globals, since the light march runs per light after the front end is done: the hit depth (already scaled by the distance fade, so the shadow fades out with the relief for free) and the mip level, so no tap in the light march needs a derivative. That last point is load-bearing - WGSL rejects a derivative in non-uniform control flow outright.

### Notes

- `parallaxShadowSamples` is declared with `definePropInternal` and a hand-written setter rather than `_defineFloat`, because the generic number rule invalidates the shader when a value crosses 0 **or 1**, and only zero matters here - otherwise dragging the count from 1 to 2 rebuilt for nothing. A test covers that, and another covers clone, which is what keeps a hand-declared property honest.
- Both factors in the accumulation are clamped non-negative. That is not defensive: the step count is never a whole number, so the last iteration overshoots the surface where the penetration is negative - and a negative penetration times a negative weight is a positive, which showed up as rings of false shadow on the crests (21.8% of pixels at a distant oblique view).
- The terrain example gains a **Self Shadow** control, and a **Light** panel with a rotation slider and slow auto rotate. Which side the light comes from matters as much as how low it is: measured self-shadowed coverage is 8.8% at 25 deg azimuth against 1.2% at 160 deg. Lowering the sun actually *reduces* coverage, because more surface falls to `N·L <= 0` where the direct light is already zero.

6 new tests. Verified on WebGL2 and WebGPU.


src/scene/materials/standard-material.js
```diff
@@ -470,6 +470,11 @@ const isBlack = (color) => {
  * when {@link parallaxMode} is {@link PARALLAX_OCCLUSION} (default is 16). Fewer taps are taken as
  * the view direction approaches the surface normal, where the ray barely moves. Has no effect in
  * {@link PARALLAX_OFFSET} mode.
+ * @property {number} parallaxShadowSamples The maximum number of height map taps taken towards each
```
mvaligursky, line 473: [P2] Register the property with standard-material parsing

`parallaxShadowSamples` is public but is missing from `standardMaterialParameterTypes` alongside `parallaxSamples` in `standard-material-parameters.js`. Consequently, every JSON/material asset containing it logs `Ignoring unsupported input property to standard material: parallaxShadowSamples`, is marked invalid by the validator, and bypasses number validation (the parser currently happens to assign it through its unknown-type fallback). Please add `parallaxShadowSamples: 'number'` to that registry and ideally cover JSON material parsing with a regression test.

mvaligursky (commented):
Automated PR review by Codex (GPT-5) at exact head `42b67a33e082cc0188cd7174653ed5ac2220213f`.

Reviewed the public material API and dirtying behavior, shader-option generation, GLSL/WGSL height-field marching and light-direction conventions, per-light performance/lifecycle implications, examples, tests, CI, and current discussion. I also ran the focused shader/material tests (39 passing), linted the changed engine/test files, and exercised the deployed example on WebGL2 and WebGPU with self-shadowing disabled and with higher sample counts; both backends rendered without relevant console errors. All required CI checks are green.

The core implementation looks coherent and remains opt-in, but I found one actionable integration gap in the standard-material JSON/asset path, noted inline.

mvaligursky:
Good catch on the parameter registry — fixed in 154c9f4.

`parallaxShadowSamples: 'number'` is now in `standardMaterialParameterTypes` next to `parallaxSamples`, and there is a regression test covering both halves of what was broken: the validator staying valid (rather than warning and marking the material invalid) and the json parser assigning the value through the typed path. I verified the test fails with the registry entry removed.

PR #9226, opened 2026-08-25
Add RenderTarget depthResolveMode controlling multisampled depth resolve

mvaligursky:
Adds control over how the samples of a multisampled depth buffer are resolved into a single depth value by the WebGPU shader-based depth resolve - used by the depth grab pass (`sceneDepthMap`), a depth `RenderTarget#copy`, and the automatic resolve into a provided `depthBuffer`.

```javascript
const rt = new RenderTarget({ colorBuffer, depth: true, samples: 4, depthResolveMode: DEPTHRESOLVE_MAX });
rt.depthResolveMode = DEPTHRESOLVE_SAMPLE0;  // mutable at any time, used at resolve time
```

**Behavior change:** the default resolve changes from sample 0 to **`DEPTHRESOLVE_MIN`**. Sample 0 is an arbitrary sample position - at geometry silhouettes it lands on the foreground or background essentially at random per pixel, causing edge noise in depth-consuming effects such as soft particles. MIN (the nearest surface, with a standard depth buffer) is deterministic and conservative, and matches what depth consumers such as Hi-Z occlusion expect. `DEPTHRESOLVE_SAMPLE0` remains available to restore the previous behavior.

**Changes:**
- New constants `DEPTHRESOLVE_MIN` (default), `DEPTHRESOLVE_MAX` and `DEPTHRESOLVE_SAMPLE0`, and a mutable `RenderTarget#depthResolveMode` property with a matching constructor option. WebGPU only - on WebGL2 the depth resolve sample selection is defined by the implementation and the property is ignored.
- `WebgpuResolver` supports all three modes, with shaders and pipelines lazily cached per mode and format. Its WGSL source is extracted into a shader chunk (`shader-chunks/frag/webgpu-depth-resolve.js`), with the mode selected by a `DEPTH_RESOLVE_***` define handled by the shader preprocessor - the first of the internal device-level renderers to move its shader to the platform chunks.
- Fixed a crash in `copyRenderTarget` when copying depth from a render target with a user-provided `depthBuffer` and `samples` > 1 - the multisampled depth of such a target is stored separately from the internally-allocated case, and the copy read a null texture.

Verified on WebGPU (Metal): a known-depth scene rendered into an MSAA target and resolved with each mode - at all 100 mixed-sample edge pixels, MIN returned the near surface depth, MAX the far clear value, and SAMPLE0 a value within that range; mode switching at runtime works. The `ground-fog` and `particles-snow` examples exercise the new default through the MSAA depth grab.

## Checklist
[...]

PR #9230, opened 2026-08-26
Support multisampled depth buffers and depth resolve targets on RenderTarget

### LINKED ISSUE #9059: WebGPU: allow sampling multisampled render targets (custom MSAA resolves for HDR color + depth)

mvaligursky:
Completes the multisampled render target surface designed in #9059 (https://github.com/playcanvas/engine/issues/9059#issuecomment-5368086964): `RenderTarget` accepts a multisampled depth-format texture as its `depthBuffer`, rendering directly into its depth samples (WebGPU only). The samples can be read in a shader using `textureLoad` on a `texture_depth_multisampled_2d` - the custom depth resolves that motivated the issue - or resolved into a new optional `depthResolveBuffer` at the end of a render pass, using the shader-based resolve controlled by `depthResolveMode` (#9226), since WebGPU has no hardware depth resolve. Together with the color support in #9221, this fixes #9059.

```javascript
const msDepth = new Texture(device, { format: PIXELFORMAT_DEPTH, width, height, samples: 4 });
const resolvedDepth = new Texture(device, { format: PIXELFORMAT_R32F, width, height, mipmaps: false });
const rt = new RenderTarget({ colorBuffer: msColor, depthBuffer: msDepth, depthResolveBuffer: resolvedDepth });
```

**Changes:**
- `RenderTarget` accepts a depth-format texture with `samples` > 1 as `depthBuffer` - it becomes the depth-stencil attachment directly, user-owned and using its actual format (a true `depth32float` MSAA attachment is now possible). The sample count is inferred from any multisampled attachment, so depth-only multisampled render targets work, and mixed sample counts across color and depth attachments are validated. The single-sampled `R32F` depthBuffer path (the legacy shader-resolve destination) is unchanged - the two are disambiguated purely by format class and sample count.
- New `depthResolveBuffer` option and getter - a single-sampled `R32F` destination for the end-of-pass shader resolve, validated for format and dimensions, covered by `resize` and `destroyTextureBuffers`. `RenderPass` resolves depth by default when it is present (the multisampled depth remains stored either way, as the resolve reads it after the pass).
- The end-of-pass depth resolve routes its source and destination by mode: legacy (internally allocated multisampled depth into the user `R32F` buffer) or explicit (the user multisampled depth into `depthResolveBuffer`), both honoring `depthResolveMode`.
- `copyRenderTarget` supports depth copies between multisampled render targets with equal sample counts (a depth snapshot via `copyTextureToTexture`), with guards for mismatched counts; the resolver path for multisampled-to-single-sampled copies is unchanged.

**API Changes:**
- `RenderTarget` constructor option `depthResolveBuffer` and matching getter.
- `RenderTarget` `depthBuffer` option accepts multisampled depth-format textures (WebGPU).

**Examples:**
- New `graphics-advanced/msaa-depth-fog` example: a split-screen comparison of depth fog computed from resolved depth (with a live `depthResolveMode` dropdown - each mode fails differently at silhouettes) against fog evaluated per-sample from the multisampled depth, which matches the antialiased color exactly. Includes an in-depth explanation of why MSAA and depth-based effects conflict, a pixel-scale magnifier, and thumbnails.

## Checklist


mvaligursky (commented):
Automated PR review by Codex (GPT-5) at exact head `33b6bb1552f40c5b9a2a540800a8016c127a232d`.

Reviewed sample-count inference and validation, explicit depth attachment formats/ownership, render-pass store and resolve policy, legacy versus explicit resolver routing, shader binding, resize/destruction, multisampled copy constraints, public API/types, example behavior, performance, CI, and current discussion.

Verification performed:
- Full unit suite: 2,446 passing, 2 pending
- Focused graphics suite: 75 passing
- Engine and example lint: clean; new example formatting: clean
- Type generation and type tests: passing
- Deployed `msaa-depth-fog` example: rendered correctly on WebGPU; MIN, MAX, and SAMPLE0 switched at runtime without relevant validation errors; resize path remained clean
- Direct WebGPU validation: equal-format 4x multisampled `depth32float` textures copied successfully with `copyTextureToTexture`
- All required CI and deployment checks: green

The core attachment, shader resolve, and copy implementation works, but I found one behavioral API gap and one public documentation mismatch, noted inline.

src/platform/graphics/render-pass.js
```diff
@@ -258,6 +258,13 @@ class RenderPass extends FramePass {
         // assume the user wants to use its content, and so store it by default
         if (rt?.depthBuffer) {
             this.depthStencilOps.storeDepth = true;
+
+            // a depth resolve buffer is the output of the shader-based depth resolve, which runs
+            // after the pass - resolve by default (the multisampled depth is stored either way, as
+            // the resolve reads it after the pass has finished)
+            if (rt.depthResolveBuffer) {
+                this.depthStencilOps.resolveDepth = true;
```
mvaligursky, line 266: [P2] Keep explicit depth resolves usable with `autoResolve: false`

This per-pass flag is not sufficient to trigger the new explicit resolve: `WebgpuGraphicsDevice.endRenderPass` also requires `target.autoResolve`. Therefore a valid target constructed with both `depthResolveBuffer` and `autoResolve: false` never writes its resolve buffer even when a caller explicitly sets `depthStencilOps.resolveDepth = true`. There is no documented manual fallback—`RenderTarget.resolve(false, true)` delegates to the empty `WebgpuRenderTarget.resolve` implementation—so the buffer can only remain stale. This also differs from explicit color resolve buffers, whose per-pass `resolve` flag is honored independently. Please either let the per-pass flag drive explicit depth resolves (retaining `autoResolve` only for the legacy path), implement the WebGPU manual resolve, or reject/document this constructor combination.

src/platform/graphics/webgpu/webgpu-graphics-device.js
```diff
@@ -1637,10 +1643,30 @@ class WebgpuGraphicsDevice extends GraphicsDevice {
 
             if (sourceRT.samples > 1) {
 
-                // resolve the depth to a color buffer of destination render target, using the
-                // resolve mode of the source render target
-                const destTexture = dest.colorBuffer.impl.gpuTexture;
-                this.resolver.resolveDepth(commandEncoder, sourceTexture, destTexture, sourceRT.depthResolveMode);
+                // multisampled destination depth buffer - a plain copy between the multisampled
```
mvaligursky, line 1648: [P3] Update the public copy documentation for this WebGPU path

`RenderTarget#copy` still says that depth buffers can only be copied on WebGL2. This branch makes that public documentation directly incorrect by supporting WebGPU multisampled depth snapshots here (and MSAA-to-single-sample shader resolves in the sibling branch). Please update the method's JSDoc to describe the supported WebGPU depth-copy cases and their sample/format requirements so users can discover this functionality.

mvaligursky (commented):
Automated follow-up PR review by Codex (GPT-5) at exact head `a5556c1db5bd3e7b7d22920440741e10cfb7eec2`, covering the commit added after previously reviewed head `33b6bb1552f40c5b9a2a540800a8016c127a232d`.

No new actionable issues found. Both findings from the previous review are resolved:

- Explicit multisampled depth attachments now follow the per-pass `resolveDepth` flag even when `RenderTarget.autoResolve` is false, while the legacy implicit multisampled-depth path remains gated by `autoResolve`.
- The public `RenderTarget#copy` documentation now describes the supported WebGL2 and WebGPU depth-copy/resolve cases.

Verification at this head:
- Focused render target, render pass, WebGPU render target, and WebGPU resolver tests: 60 passing.
- Full test suite: 2446 passing, 2 pending.
- ESLint passed for both changed files.
- Type declaration generation and type tests passed.
- All current CI and deployment checks are green.

There is no dedicated regression test for the exact explicit-depth plus `autoResolve: false` branch, so that remains a non-blocking coverage opportunity; the updated condition and surrounding resolve paths were inspected directly.

PR #9231, opened 2026-08-26
Fix looping sounds that specify a start time and duration

### LINKED ISSUE #4712: Audio doesn't loop on Firefox when Start Time and Duration are used
Forum: https://forum.playcanvas.com/t/looped-audio-stops-in-mozilla/27825

Project: https://playcanvas.com/project/993494/overview/sound-loop

image

willeastcott:
## Description

A `SoundInstance` (or sound slot) that has both a **start time** and a **duration** set does not loop — it plays a single iteration, fires `end` and stops.

### Root cause

`SoundInstance` passed `this._duration` as the third argument to `AudioBufferSourceNode.start()` even when the instance was looping. Per the Web Audio spec that argument is a hard limit on *total* playback — the source stops after that many seconds of buffer content, "including any whole or partial loop iterations" — so the source stopped at the end of the first iteration, `onended` fired, and `_onEnded()` turned that into an `end` event plus `stop()`.

This only ever worked because the engine assigns `source.loop` *after* `start()` (the "moved to be after start() because of Chrome bug" comment), and Blink decided whether to honour the grain duration at `start()` time:

```cpp
// third_party/blink/renderer/modules/webaudio/audio_buffer_source_handler.cc
// AudioBufferSourceHandler::ClampGrainParameters(), called from StartSource()
if (is_duration_given_ && Loop()) {
  end_time_ = start_time_ + grain_duration_;   // spec behaviour
} else {
  grain_duration_ = ClampTo(grain_duration_, 0.0, buffer_duration - grain_offset_);
}
```

With `loop` still false at `start()`, no end time was scheduled and the duration was effectively discarded. That accident is now gone everywhere:

| Engine | Behaviour |
|---|---|
| Firefox ≥ 84 (Nov 2020) | honours the duration regardless of ordering ([bug 1618225](https://bugzilla.mozilla.org/show_bug.cgi?id=1618225)) — the original report |
| Chromium ≤ 149 | legacy `is_duration_given_ && Loop()` gate — engine works by accident |
| Chromium ≥ **150** (branch 7871, stable 2026-06-30) | handler rewrite applies the limit unconditionally in the render loop — engine breaks |

which matches the recent report on the issue that Chrome and Brave now fail too. I bisected the milestone branches to confirm 149 → 150 is the flip.

### Fix

1. Only pass the duration to the source for non-looping instances (`_startSource()`). For a looping instance, `loopStart`/`loopEnd` already define the region to repeat, which every browser honours identically.
2. Since a looping source is now started without a duration, disabling `loop` mid-playback would let it run to the end of the buffer. `_stopAtEndOfLoopRegion()` schedules a stop at the end of the current iteration so the duration is still respected.
3. Clamp `loopEnd` instead of wrapping it with a modulo — see below.

### Second, independent bug fixed here

```js
this.source.loopEnd = Math.max(this.source.loopStart, capTime(this._startTime + this._duration, buffer.duration));
```

`capTime()` is a modulo, so whenever `startTime + duration` landed on or past a multiple of the buffer duration the wrapped value collapsed onto `loopStart`. Both the spec and Blink's `ResolveLoopPoints()` treat `loopEnd <= loopStart` as "loop the whole buffer". Measured on a 4 s clip, `startTime 1 / duration 3` ("loop the last 3 s"), `startTime 2 / duration 2` and `startTime 3 / duration 2` all looped the entire clip. This one is browser-independent and was simply masked by looping being broken anyway.

## Verification

Verified against a real Web Audio implementation using `OfflineAudioContext` (4 s buffer with a distinct DC level per second, so the rendered samples identify which part of the buffer is playing):

- Reproduced the ordering dependence directly — `loop = true` before `start(0, 1, 2)` stops after 2 s and fires `onended`; set after `start()` it loops forever.
- With the fix, the loop region renders correctly for `startTime`/`duration` of 1/1, 1/2, 1/3, 2/2, 3/2, 0/2 and 0/4, including the three cases that previously fell back to looping the whole buffer.
- Validated `_stopAtEndOfLoopRegion()` by suspending the render mid-playback: for `startTime 1, duration 2`, disabling loop at t=3.5 s schedules the stop at exactly 4.0 s and every sampled output value matches the prediction.

ESLint is clean on the changed file.

## Notes for reviewers

- The `source.loop` assignment is still left after `start()` — unchanged, since that ordering also relates to the volume/gain setup next to it.
- The inverse transition (enabling `loop` mid-playback on an instance that was *started* non-looping with a duration) still stops at the duration, because the duration was baked into `start()`. Fixing that would need the source to be recreated; it is the same root cause but a separate, much rarer case, so it is left alone here.
- Two adjacent pre-existing bugs found while investigating are deliberately **not** touched: `resume()` passes the full `_duration` rather than the remaining duration, so pause/resume extends a non-looping grain; and `_updateCurrentTime()` derives `_currentTime` from the absolute buffer offset while the `currentTime` setter treats the value as relative to `_startTime`, so `currentTime` is wrong unless `startTime` is an integer multiple of `duration`. Happy to file these separately.
- No unit test added — there is no existing `SoundInstance` test and covering this needs a real (or offline) `AudioContext` rather than a mock. Say the word if you'd like a test harness for it.

Fixes #4712

## Checklist


src/platform/sound/instance.js
```diff
@@ -706,6 +712,50 @@ class SoundInstance extends EventHandler {
...
+        }
+
+        // work out how far into the loop region playback currently is
+        const context = this._manager.context;
+        let position = this._currentOffset + (context.currentTime - this._startedAt) * this._pitch;
+        if (position > loopEnd) {
+            position = loopStart + ((position - loopStart) % region);
+        }
+
+        source.stop(context.currentTime + Math.max(0, loopEnd - position) / this._pitch);
```
mvaligursky: [P2] Reschedule or cancel this stop when playback state changes

`AudioBufferSourceNode.stop(when)` installs an absolute context-time deadline that remains active after this method returns. If an app disables looping and then re-enables it before the current region ends, the `loop` setter changes `source.loop` back to `true` but this old stop still fires; a Chromium `OfflineAudioContext` reproduction with start/duration 1/2, `loop = false` at 0.5 s, and `loop = true` at 1 s became silent and stopped at 2 s even though `instance.loop` was true. Changing `pitch` after this call similarly leaves a deadline calculated with the old rate, cutting the region early or late. Please track the pending region stop and replace/cancel/reschedule it when loop or pitch changes (or recreate the source with equivalent state), and cover the off/on transition with a real/offline audio regression test.

mvaligursky (commented):
Automated PR review by Codex (GPT-5) at exact head `eb28263204109ccb92350c76e4d760c7c7eed836`.

The primary fix is well targeted: looping sources no longer receive a finite `start()` duration, and clamping `loopEnd` avoids the whole-buffer fallback when the previous modulo collapsed the loop points. I found one actionable issue in the new mid-playback loop-disable path.

Verification:
- Reviewed the complete PR diff and surrounding `SoundInstance` play, resume, pause, pitch, loop, and `onended` state transitions.
- Reproduced the changed behavior in a real Chromium `OfflineAudioContext`: the basic start-time/duration loop remained active, but an off/on loop toggle still stopped at the previously scheduled deadline, and a post-schedule pitch change also stopped at the stale deadline.
- `git diff --check` passed.
- All GitHub build, lint, type, unit-test, docs, example, API, size, and deployment checks are green.

The absence of a real-Web-Audio regression test currently leaves this state-transition case uncovered.

src/platform/sound/instance.js
willeastcott: Good catch - fixed in 1fa201f76. The pending region stop is now tracked and revisited on both transitions: re-enabling `loop` pushes the deadline out of reach (the Web Audio API can only replace a scheduled stop, not cancel it) and a `pitch` change recomputes it against the new rate, using a buffer position tracked separately from `_currentTime` (which is capped to the duration of the instance, not the buffer) and re-baselined on every rate change. Verified in a Chromium `OfflineAudioContext` that your exact repro (start/duration 1/2, `loop = false` at 0.5s, `loop = true` at 1.0s) now loops indefinitely instead of going silent at 2s, that `loop = false` at 1.0s followed by `pitch = 0.5` at 1.5s slides the stop from 2.0s to 2.5s, that a pitch change while still looping is accounted for when looping is later disabled, and that the plain loop-off case still stops exactly at `loopEnd`. No offline-audio regression test, though: the mocha suite runs in Node, which has no Web Audio implementation, so this would need a new browser-based harness - happy to add one if you want it.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #9233, opened 2026-08-27
Error-driven GSplat LOD selection with a derived fallback

mvaligursky:
LOD selection for streamed SOG has been purely geometric: each node picks a level from camera-distance bands, and the budget balancer redistributes by distance bucket. Distance says nothing about how much a node *actually* loses at a given level — a node of near-duplicate splats and one carrying fine detail at the same distance are treated alike.

This spends the budget by measured quality loss instead.

**How it works.** Every node starts at the cheapest level it can render, which is the coarsest the scene can be and therefore always within budget. Each single-level upgrade available anywhere in the scene is then ranked by `coverage * error removed / splats added` and bought best-first until one does not fit. Stopping at the first upgrade that does not fit — rather than skipping it and continuing — keeps a node's outcome from depending on whether some unrelated cheaper upgrade happened to be considered first, which is what made small camera movements flip levels on and off.

**Errors always exist.** They are read from the manifest when splat-transform wrote them (`lodErrors`), and derived from splat counts otherwise, so there is one code path rather than a fallback to distance selection. The derived measure is `ln(finestCount / count)`: the allocator only consumes differences between adjacent levels and decimation is geometric, so a log gives equal error steps for equal count ratios.

Measured against authored errors as ground truth on three captures, over multiple cameras, budgets and weightings: the derived proxy lands within 2–17%, where the distance-based system it replaces is 20–790% worse. Accuracy tracks how finely an asset is partitioned rather than its level count.

**Precomputation.** Per node, the levels are reduced once to the Pareto frontier over (splat count, error) and stored as a chain of upgrades with their cost and value. Because coverage is a single non-negative per-node scalar, the running minimum that keeps a node's returns non-increasing can be taken at build time too, leaving one multiply per upgrade at selection time. Only a node's next unbought upgrade is ever queued, so at most one entry per node is live and the buckets are intrusive lists over preallocated typed arrays with no per-update allocation.

**Changes:**
- New `GSplatLodTable`, built per octree and LOD range, holding the per-node upgrade chains
- `GSplatBudgetBalancer` rewritten: value-ranked greedy over upgrades, 512 fixed-scale buckets, lazy insertion, early exit
- `GSplatOctree` reads or derives per-level errors; `lodErrorSource` reports which
- Underfill and prefetch step along a node's LOD chain rather than raw level indices, so a node's splat count cannot exceed what was budgeted for it — real captures do contain levels where a coarser one holds *more* splats
- Distance-bucket balancing, `NodeInfo.budgetBucket`, `computeGlobalMaxDistance` and the `_budgetScale` feedback loop removed

**API Changes:**
- `GSplatComponent#lodBaseDistance` and `GSplatComponent#lodMultiplier` are removed (`Debug.removed`). LOD levels are now chosen to fit `app.scene.gsplat.splatBudget`, so use that to control quality. The equivalents on the internal `GSplatPlacement` and the component schema entries are gone too.
- `GSplatParams#splatBudget` defaults to `1000000` rather than `0`, and budgeted selection can no longer be disabled. A non-positive value would pin every node to its coarsest level, so it warns and the default is used instead. Any content relying on `splatBudget = 0` to mean "no cap" needs a real budget.

**Examples:**
- All streamed-LOD examples moved onto assets exported with error metadata
- `billions` sets an explicit budget, having previously relied on `splatBudget = 0` plus the per-instance distance ramp

### DIFF
--- a/src/framework/components/gsplat/component.js
+++ b/src/framework/components/gsplat/component.js
@@ -98,20 +98,6 @@ class GSplatComponent extends Component {
     _materialTmp = null;
 
     /**
-     * Base distance for the first LOD transition (LOD 0 to LOD 1).
-     *
-     * @private
-     */
-    _lodBaseDistance = 5;
-
-    /**
-     * Geometric multiplier between successive LOD distance thresholds.
-     *
-     * @private
-     */
-    _lodMultiplier = 3;
-
-    /**
      * Minimum allowed LOD index (inclusive).
      *
      * @private
@@ -385,54 +371,41 @@ class GSplatComponent extends Component {
     }
 
     /**
-     * Sets the base distance for the first LOD transition (LOD 0 to LOD 1). Objects closer
-     * than this distance use the highest quality LOD. Each subsequent LOD level transitions
-     * at a progressively larger distance, controlled by {@link lodMultiplier}. Clamped to a
-     * minimum of 0.1. Defaults to 5.
-     *
      * @type {number}
+     * @deprecated LOD level selection is driven by `app.scene.gsplat.splatBudget`.
+     * @ignore
      */
     set lodBaseDistance(value) {
-        this._lodBaseDistance = Math.max(0.1, value);
-        if (this._placement) {
-            this._placement.lodBaseDistance = this._lodBaseDistance;
-        }
+        Debug.removed('GSplatComponent#lodBaseDistance is removed. LOD levels are chosen to fit app.scene.gsplat.splatBudget; use that to control quality.');
     }
 
     /**
-     * Gets the base distance for the first LOD transition.
-     *
      * @type {number}
+     * @deprecated LOD level selection is driven by `app.scene.gsplat.splatBudget`.
+     * @ignore
      */
     get lodBaseDistance() {
-        return this._lodBaseDistance;
+        Debug.removed('GSplatComponent#lodBaseDistance is removed. LOD levels are chosen to fit app.scene.gsplat.splatBudget; use that to control quality.');
+        return 0;
     }
 
     /**
-     * Sets the multiplier between successive LOD distance thresholds. Each LOD level
-     * transitions at this factor times the previous level's distance, creating a geometric
-     * progression. Lower values keep higher quality at distance; higher values switch to
-     * coarser LODs sooner. Clamped to a minimum of 1.2 to avoid degenerate logarithmic LOD
-     * computation. LOD distances are automatically compensated for the camera's field of
-     * view — a wider FOV makes objects appear smaller on screen, so LOD switches to coarser
-     * levels sooner to match the reduced screen-space detail. Defaults to 3.
-     *
      * @type {number}
+     * @deprecated LOD level selection is driven by `app.scene.gsplat.splatBudget`.
+     * @ignore
      */
     set lodMultiplier(value) {
-        this._lodMultiplier = Math.max(1.2, value);
-        if (this._placement) {
-            this._placement.lodMultiplier = this._lodMultiplier;
-        }
+        Debug.removed('GSplatComponent#lodMultiplier is removed. LOD levels are chosen to fit app.scene.gsplat.splatBudget; use that to control quality.');
     }
 
     /**
-     * Gets the geometric multiplier between successive LOD distance thresholds.
-     *
      * @type {number}
+     * @deprecated LOD level selection is driven by `app.scene.gsplat.splatBudget`.
+     * @ignore
      */
     get lodMultiplier() {
-        return this._lodMultiplier;
+        Debug.removed('GSplatComponent#lodMultiplier is removed. LOD levels are chosen to fit app.scene.gsplat.splatBudget; use that to control quality.');
+        return 0;
     }
 
     /**
@@ -485,24 +458,20 @@ class GSplatComponent extends Component {
 
     /**
      * @type {number[]|null}
-     * @deprecated Use {@link lodBaseDistance} and {@link lodMultiplier} instead.
+     * @deprecated LOD level selection is driven by `app.scene.gsplat.splatBudget`.
      * @ignore
      */
     set lodDistances(value) {
-        Debug.removed('GSplatComponent#lodDistances is removed. Use lodBaseDistance and lodMultiplier instead.');
-        if (Array.isArray(value) && value.length > 0) {
-            this.lodBaseDistance = value[0];
-            this.lodMultiplier = 3;
-        }
+        Debug.removed('GSplatComponent#lodDistances is removed. LOD levels are chosen to fit app.scene.gsplat.splatBudget; use that to control quality.');
     }
 
     /**
      * @type {number[]}
-     * @deprecated Use {@link lodBaseDistance} and {@link lodMultiplier} instead.
+     * @deprecated LOD level selection is driven by `app.scene.gsplat.splatBudget`.
      * @ignore
      */
     get lodDistances() {
-        Debug.removed('GSplatComponent#lodDistances is removed. Use lodBaseDistance and lodMultiplier instead.');
+        Debug.removed('GSplatComponent#lodDistances is removed. LOD levels are chosen to fit app.scene.gsplat.splatBudget; use that to control quality.');
         return [];
     }
 
@@ -1008,8 +977,6 @@ class GSplatComponent extends Component {
             this._placement = null;
 
             this._placement = new GSplatPlacement(resource, this.entity, 0, this._parameters, null, this._id);
-            this._placement.lodBaseDistance = this._lodBaseDistance;
-            this._placement.lodMultiplier = this._lodMultiplier;
             this._placement.lodRangeMin = this._lodRangeMin;
             this._placement.lodRangeMax = this._lodRangeMax;
             this._placement.workBufferUpdate = this._workBufferUpdate;
--- a/src/scene/gsplat-unified/gsplat-lod-table.js
+++ b/src/scene/gsplat-unified/gsplat-lod-table.js
@@ -1,5 +1,3 @@
-import { Debug } from '../../core/debug.js';
-
 /**
  * @import { GSplatOctree } from './gsplat-octree.js'
  */
@@ -11,20 +9,21 @@ import { Debug } from '../../core/debug.js';
  * Each node is reduced to a chain of single-level *upgrades*, ordered cheapest level first. The
  * chain is the Pareto frontier over (splat count, error): a level is kept only when it strictly
  * improves on the cheapest error seen so far, which discards levels that cost more and look worse
- * than something else the node already offers. Requiring a *strict* improvement also collapses
- * levels identical in both, so consecutive entries always differ in both and every upgrade's cost
- * stays above zero.
+ * than something else the node already offers - a small fraction in practice, and redundant by
+ * definition. Requiring a *strict* improvement also collapses levels identical in both, so
+ * consecutive entries always differ in both and every upgrade's cost stays above zero.
+ *
+ * Every level that survives is a level the node can render, and the allocator moves one level at a
+ * time. Nothing else is dropped: levels that are poor value for their splats are still real
+ * improvements, and the chain doubles as the set of states streaming and underfill may pass
+ * through, so removing them would deny a loaded level to underfill and turn prefetch's one-level
+ * climb into a multi-level jump.
  *
  * Per update the allocator needs one number per node - its projected screen coverage - and the
- * value of an upgrade is `coverage * error removed / splats added`. The second factor is fixed, and
- * coverage is a single non-negative scalar multiplying every upgrade of that node equally, so the
- * running minimum that keeps a node's returns non-increasing can be taken here rather than each
- * update:
- *
- *   min over j<=k of (coverage * r_j)  ===  coverage * min over j<=k of r_j
- *
- * That is what {@link GSplatLodTable#upgradeRatio} stores, leaving one multiply per upgrade at
- * selection time.
+ * value of an upgrade is `coverage * error removed / splats added`. The second factor is fixed and
+ * coverage is a single non-negative scalar multiplying every upgrade of that node equally, so
+ * {@link GSplatLodTable#upgradeRatio} holds the fixed part and selection costs one multiply per
+ * upgrade.
  *
  * @ignore
  */
@@ -92,9 +91,13 @@ class GSplatLodTable {
     upgradeCost;
 
     /**
-     * Per upgrade, error removed per additional splat, clamped to the running minimum along the
-     * node's chain so returns never increase as a node gets finer. Multiplying by the node's
-     * coverage yields the upgrade's value.
+     * Per upgrade, error removed per additional splat over the best run this upgrade opens up -
+     * `max` over the levels reachable from where it starts, rather than its own slope. Multiplying
+     * by the node's coverage yields the upgrade's value.
+     *
+     * Note this is not monotone along a chain: once a poorly-valued step has been taken, what
+     * remains can be worth more than what was just bought. The allocator tolerates that, see
+     * GSplatBudgetBalancer.
      *
      * @type {Float32Array}
      */
@@ -187,35 +190,6 @@ class GSplatLodTable {
                 continue;
             }
 
-            // Reduce the frontier to its upper concave hull over (count, -error). The frontier only
-            // guarantees that error falls as cost rises - the slopes between consecutive levels can
-            // still *rise* towards the finer end, and a level sitting below the chord between its
-            // neighbours is never worth stopping at. Dropping it pools the levels either side into
-            // one compound upgrade priced at the chord, which is the value actually on offer.
-            //
-            // Without this the allocator would have to clamp each marginal ratio to the running
-            // minimum to keep its greedy order valid, and that understates the compound step: for
-            // levels (20,10) -> (90,8) -> (100,0) both ratios clamp to 2/70, hiding the 10-error
-            // reduction available for 80 splats at 0.125. Real authored error curves are not
-            // concave - this binds on 9-27% of upgrades across measured captures - so the clamp
-            // could leave a node stranded at its coarsest level while cheaper, less valuable
-            // upgrades elsewhere won the budget. Compacts in place, as above.
-            let hullCount = 0;
-            for (let i = 0; i < frontierCount; i++) {
-                const lod = scratch[i];
-                while (hullCount >= 2) {
-                    const a = scratch[hullCount - 2];
-                    const b = scratch[hullCount - 1];
-                    // sign of (b - a) x (lod - a) with y = -error; >= 0 means b is not above the
-                    // chord a -> lod, so it cannot be an optimal stopping point
-                    const cross = (lods[b].count - lods[a].count) * (lods[a].error - lods[lod].error) -
-                                  (lods[a].error - lods[b].error) * (lods[lod].count - lods[a].count);
-                    if (cross >= 0) hullCount--; else break;
-                }
-                scratch[hullCount++] = lod;
-            }
-            frontierCount = hullCount;
-
             const startLod = scratch[0];
             this.startLod[n] = startLod;
             this.startCount[n] = lods[startLod].count;
@@ -225,20 +199,28 @@ class GSplatLodTable {
             for (let i = 1; i < frontierCount; i++) {
                 const coarseLod = scratch[i - 1];
                 const fineLod = scratch[i];
-                const cost = lods[fineLod].count - lods[coarseLod].count;
-                const benefit = lods[coarseLod].error - lods[fineLod].error;
+
+                // Value this step by the best deal reachable by carrying on from where it starts,
+                // not by its own slope. A step can be poor on its own while the run it opens is
+                // excellent - for levels (20,10) -> (90,8) -> (100,0) the first step removes 2
+                // error for 70 splats, but reaching 100 removes 10 for 80. Priced locally the node
+                // looks worthless and loses the budget to genuinely inferior upgrades elsewhere;
+                // priced by its reach it competes on what it is actually worth, then climbs one
+                // level at a time. O(levels) per step over a handful of levels.
+                let ratio = 0;
+                for (let j = i; j < frontierCount; j++) {
+                    const reach = scratch[j];
+                    const r = (lods[coarseLod].error - lods[reach].error) /
+                              (lods[reach].count - lods[coarseLod].count);
+                    if (r > ratio) ratio = r;
+                }
+
                 upgradeToLod[upgradeCount] = fineLod;
-                upgradeCost[upgradeCount] = cost;
-                upgradeRatio[upgradeCount] = benefit / cost;
+                upgradeCost[upgradeCount] = lods[fineLod].count - lods[coarseLod].count;
+                upgradeRatio[upgradeCount] = ratio;
                 upgradeCount++;
             }
 
-            // The hull makes this hold by construction, and both the allocator's single sweep and
-            // its lazy re-insertion depend on it.
-            const emitted = upgradeCount;
-            Debug.call(() => {
-                this._assertNonIncreasing(n, upgradeRatio, emitted);
-            });
         }
 
         this.firstUpgrade[nodeCount] = upgradeCount;
@@ -250,21 +232,6 @@ class GSplatLodTable {
         this.upgradeToLod = upgradeToLod.subarray(0, upgradeCount).slice();
         this.upgradeCost = upgradeCost.subarray(0, upgradeCount).slice();
         this.upgradeRatio = upgradeRatio.subarray(0, upgradeCount).slice();
-    }
-
-    /**
-     * Asserts that a node's emitted marginal ratios never rise towards the finer end.
-     *
-     * @param {number} nodeIndex - The node whose chain was just emitted.
-     * @param {Float32Array} ratios - The ratio array being filled.
-     * @param {number} end - One past the node's last emitted upgrade.
-     * @private
-     */
-    _assertNonIncreasing(nodeIndex, ratios, end) {
-        for (let k = this.firstUpgrade[nodeIndex] + 1; k < end; k++) {
-            Debug.assert(ratios[k] <= ratios[k - 1],
-                `GSplatLodTable: node ${nodeIndex} has rising marginal returns (${ratios[k - 1]} -> ${ratios[k]}); the concave hull should have pooled these levels.`);
-        }
     }
 
     /**
--- a/src/scene/gsplat-unified/gsplat-lod-table.js
+++ b/src/scene/gsplat-unified/gsplat-lod-table.js
@@ -1,3 +1,5 @@
+import { Debug } from '../../core/debug.js';
+
 /**
  * @import { GSplatOctree } from './gsplat-octree.js'
  */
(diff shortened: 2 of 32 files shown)


mvaligursky (commented):
Automated PR review by Codex (GPT-5) at exact head `51149c88040e4acdb0e9f0db38d30867e9178233`.

This is a thoughtful redesign: the count/error frontier, no-allocation runtime queue, count-monotone underfill path, malformed-metadata fallback, and focused tests are all strong. I found three actionable implementation/compatibility issues and one public documentation mismatch. The change to `splatBudget`'s default and zero semantics is also a breaking behavioral change and should be handled alongside the public accessor removals in the release/versioning decision.

Additional P3 documentation note: the public `lodRangeMin` and `lodRangeMax` descriptions still say the optimal LOD is selected by distance. Both should describe clamping the new global error/budget allocator's selected chain instead.

Verification:
- Reviewed the full 31-file diff and the surrounding world update, streaming, placement, range, and allocator lifecycle paths.
- Focused LOD table/balancer tests: 33 passing.
- Full suite: 2481 passing, 2 pending.
- ESLint passed for the changed core implementation and new tests.
- Added a minimal allocation reproduction for a non-concave authored error curve; the current allocator spent 80/120 splats with residual error 10 while a feasible 120-splat choice had residual error 4.
- Ran the updated authored-error `lod-streaming` example in the browser on WebGPU: it loaded and rendered cleanly at 3.953M splats under its 4M budget, with no warning/error logs.
- `git diff --check` passed, and all GitHub build, lint, type, unit-test, docs, examples, API, size, and deployment checks are green.

src/scene/gsplat-unified/gsplat-lod-table.js
```diff
@@ -0,0 +1,284 @@
...
+            totalStartCount += lods[startLod].count;
+            totalFinestCount += lods[scratch[frontierCount - 1]].count;
+
+            let previousRatio = Infinity;
+            for (let i = 1; i < frontierCount; i++) {
+                const coarseLod = scratch[i - 1];
+                const fineLod = scratch[i];
+                const cost = lods[fineLod].count - lods[coarseLod].count;
+                const benefit = lods[coarseLod].error - lods[fineLod].error;
+                previousRatio = Math.min(previousRatio, benefit / cost);
```
mvaligursky: [P2] Build the concave frontier instead of clamping marginal returns

A Pareto frontier does not guarantee non-increasing marginal benefit, and replacing an increasing successor ratio with the running minimum loses the value of buying the required upgrades as a compound step. For example, one node with `(count,error)` levels `(20,10) -> (90,8) -> (100,0)` gets both ratios clamped to `2/70`, although the direct coarse-to-fine upgrade removes 10 error for 80 splats (`0.125`). Against another `20 -> 60` node removing 4 error (`0.1`) with a total budget of 120, the current balancer buys the second node, stops at 80 splats, and leaves residual error 10; choosing the first node's full 80-splat upgrade fits exactly and leaves residual 4. Authored error curves are not guaranteed concave, so this can defeat the stated quality objective substantially. Please reduce each node to the appropriate upper concave hull (or pool adjacent increasing-slope upgrades into compound steps) before emitting the chain, and add this shape to the allocator tests.

src/scene/gsplat-unified/gsplat-octree.js
```diff
@@ -226,6 +286,94 @@ class GSplatOctree {
...
+     * the sub-range's cheapest entry.
+     *
+     * @param {number} rangeMin - Finest allowed LOD index.
+     * @param {number} rangeMax - Coarsest allowed LOD index.
+     * @returns {GSplatLodTable} The selection table.
+     */
+    getLodTable(rangeMin, rangeMax) {
+        let table = this._lodTable;
+        if (!table || table.rangeMin !== rangeMin || table.rangeMax !== rangeMax) {
+            table = new GSplatLodTable(this, rangeMin, rangeMax);
```
mvaligursky, line 359: [P2] Avoid rebuilding tables for supported per-instance ranges

`lodRangeMin`/`lodRangeMax` are per placement, so two instances of the same octree can legitimately use different ranges. With this single cache slot, every LOD cycle alternates ranges in `resolveLodRange()`, then the balancer calls `getLodTable()` for both again: two instances produce four full table builds per update, and multiple duplicate typed-array tables remain referenced by the instances and balancer. Each build is proportional to nodes × levels and allocates both maximum-size and trimmed arrays, so large streamed captures will incur repeated allocation/GC hitches during camera movement. The debug warning acknowledges the case but does not make a supported component configuration safe. Cache active tables by range, or let each instance retain its table and have the balancer consume that resolved table, rebuilding only when that instance's configured range actually changes.

src/framework/components/gsplat/component.js
```diff
@@ -385,54 +371,41 @@ class GSplatComponent extends Component {
...
      * @type {number}
+     * @deprecated LOD level selection is driven by `app.scene.gsplat.splatBudget`.
+     * @ignore
      */
     set lodBaseDistance(value) {
-        this._lodBaseDistance = Math.max(0.1, value);
-        if (this._placement) {
-            this._placement.lodBaseDistance = this._lodBaseDistance;
-        }
+        Debug.removed('GSplatComponent#lodBaseDistance is removed. LOD levels are chosen to fit app.scene.gsplat.splatBudget; use that to control quality.');
```
mvaligursky, line 379: [P1] Keep these public APIs through a deprecation cycle

`lodBaseDistance` and `lodMultiplier` are currently documented public accessors, but this change marks them `@ignore`, removes them from component serialization, makes setters no-ops/getters return 0, and calls `Debug.removed` (which is stripped in production). The API report confirms four public members disappear, so existing projects silently lose configured behavior after a normal 2.x update. Adding `@deprecated` at the same time as removal is not a deprecation period. Per the engine's API stability policy, retain typed/serialized compatibility with a development deprecation warning for at least a major-version cycle, or land this together with an explicitly versioned breaking release and migration plan. The new default and `splatBudget = 0` behavior change needs the same treatment.

src/scene/gsplat-unified/gsplat-lod-table.js
```diff
@@ -0,0 +1,335 @@
...
+                    const b = scratch[hullCount - 1];
+                    // sign of (b - a) x (lod - a) with y = -error; >= 0 means b is not above the
+                    // chord a -> lod, so it cannot be an optimal stopping point
+                    const cross = (lods[b].count - lods[a].count) * (lods[a].error - lods[lod].error) -
+                                  (lods[a].error - lods[b].error) * (lods[lod].count - lods[a].count);
+                    if (cross >= 0) hullCount--; else break;
+                }
+                scratch[hullCount++] = lod;
+            }
+            frontierCount = hullCount;
```
mvaligursky: [P1] Preserve feasible intermediate LODs in the discrete allocator and streaming chain

A point below the chord is redundant only for a fractional allocation; these LODs are indivisible. With this file's own `(count,error)` example `(20,10) -> (90,8) -> (100,0)`, a budget of 90 can render the middle level exactly, but this compaction emits `[20,100]`, so the balancer remains at 20 splats/error 10 and leaves 70 splats unused instead of choosing error 8. Because `findCoarserAccepted` and `finerOnChain` use this same compacted chain, an already-loaded 90-splat level is also unavailable for underfill and prefetch jumps directly from 20 to 100. Please keep a count-monotone Pareto chain for feasible/streaming states and represent pooled allocator offers separately (or otherwise handle affordable prefixes), with tests for a budget smaller than the compound endpoint and for a loaded pooled-out intermediate.

src/scene/gsplat-unified/gsplat-octree.js
```diff
@@ -226,6 +294,98 @@ class GSplatOctree {
...
+     * @returns {GSplatLodTable} The selection table.
+     */
+    getLodTable(rangeMin, rangeMax) {
+        const key = rangeMin * 256 + rangeMax;
+        let table = this._lodTables.get(key);
+        if (!table) {
+            // Evict the oldest rather than growing without bound - Map iterates in insertion order.
+            // The cap only needs to cover the ranges live at one time, which is a handful of quality
+            // presets in practice.
+            if (this._lodTables.size >= MAX_LOD_TABLES) {
```
mvaligursky: [P2] Do not evict tables that are still used by live instances

The fixed cap reintroduces the rebuild loop as soon as five legitimate per-placement ranges are live. After warming ranges A-E, the map contains only B-E; resolving A evicts B, resolving B evicts C, and so on, causing all five node×level tables and their typed arrays to be rebuilt every LOD update. A five-range reproduction produces five cache misses on every subsequent pass. The debug warning only diagnoses the resulting allocation/GC hitch. Please retain all currently referenced ranges (for example with per-range instance refcounts and eviction of inactive entries), rather than using a FIFO cap that can evict live tables.

mvaligursky (commented):
Automated re-review by Codex (GPT-5) at `8690339030d2edbe99763f7cea6f94918b07528a`.

The follow-up correctly removes the running-minimum distortion in the previously reported 120-splat case and makes the balancer consume each instance's already-resolved table. However, I found two remaining issues in those fixes:

- Removing below-chord Pareto levels assumes splat upgrades are divisible. They are discrete, so an intermediate can be the best—and only—upgrade that fits a smaller budget. Removing it also makes the shared streaming navigation skip useful loaded/prefetch levels.
- The four-entry FIFO table cache still rebuilds every live range on every update as soon as five distinct per-instance ranges coexist.

Non-blocking documentation cleanup: `GSplatLodTable`'s class/property comments still describe single-level upgrades and running-minimum clamping, while the implementation now emits compound hull steps. The public `lodRangeMin`/`lodRangeMax` descriptions also still say selection is distance-driven. Per instruction, I did not reassess or repeat the separate deprecation-cycle concern.

Validation:
- Focused LOD table/balancer suite: 36 passing.
- Full suite: 2484 passing, 2 pending.
- Full ESLint and `git diff --check`: pass.
- Targeted discrete-budget reproduction: chain becomes `[2, 0]`; at budget 90 the allocator spends 20 with residual error 10 despite LOD 1 fitting exactly at 90 with residual error 8. The same chain cannot select loaded LOD 1 for underfill and prefetches directly from LOD 2 to LOD 0.
- Targeted cache reproduction: five live ranges with capacity four produced five identity misses/rebuilds per subsequent pass.
- All GitHub CI, API, build-size, docs, type, examples, test, and deployment checks are green.

src/scene/gsplat-unified/gsplat-lod-table.js
```diff
@@ -0,0 +1,312 @@
...
+                // not by its own slope. A step can be poor on its own while the run it opens is
+                // excellent - for levels (20,10) -> (90,8) -> (100,0) the first step removes 2
+                // error for 70 splats, but reaching 100 removes 10 for 80. Priced locally the node
+                // looks worthless and loses the budget to genuinely inferior upgrades elsewhere;
+                // priced by its reach it competes on what it is actually worth, then climbs one
+                // level at a time. O(levels) per step over a handful of levels.
+                let ratio = 0;
+                for (let j = i; j < frontierCount; j++) {
+                    const reach = scratch[j];
+                    const r = (lods[coarseLod].error - lods[reach].error) /
```
mvaligursky, line 213: [P1] Do not price a prefix using a suffix that may not fit

This ratio ranks only the next adjacent upgrade, but it can include the error reduction of a farther endpoint without checking that the cumulative run is affordable. For the documented chain `(20,10) -> (90,8) -> (100,0)` plus a rival `(20,4) -> (60,0)`, equal coverage and budget 110, the floor costs 40. This code prices the first 70-splat step at `10/80 = 0.125`, ahead of the rival's `4/40 = 0.1`, so the balancer buys it and reaches `(90,8)`; the 10-splat suffix then does not fit and the early exit leaves residual error `8 + 4 = 12`. Buying the rival instead is feasible at total cost 80 and leaves residual error `10 + 0 = 10`. The existing test only uses budget 120, where the whole run fits, and `exactGreedy` repeats this same pricing assumption. Please make the run whose value is used an atomic/affordability-aware purchase, or otherwise rank a prefix only by quality that is actually committed under the remaining budget, and add the 110-boundary regression case.

src/scene/gsplat-unified/gsplat-octree-instance.js
```diff
@@ -609,84 +575,26 @@ class GSplatOctreeInstance {
...
-                const lod = nodes[nodeIndex].lods[optimalLodIndex];
-                if (lod && lod.count) {
-                    totalSplats += lod.count;
-                }
-            }
+            // Squared projected radius. Floored just above zero so a degenerate node still has a
+            // well-defined, lowest-possible priority rather than a value the allocator has to
+            // special-case.
+            const radius = nodes[nodeIndex].boundingSphere.w;
+            const projectedRadius = radius / Math.max(radius + fovAdjustedDistance, 1e-12);
```
mvaligursky: [P2] Handle orthographic coverage without distance attenuation

This is the perspective projected-radius formula, but it is used for every camera projection. Under an orthographic camera, two equal-radius nodes occupy the same screen area regardless of depth (`Camera#getScreenSize` uses `radius / orthoHeight`), while this expression gives the farther node lower coverage and therefore less LOD budget; moving the camera along its view axis also reshuffles quality even though neither node's rendered footprint changes. Orthographic cameras are supported by the engine and GSplat examples, so branch on `camera.projection` and derive the orthographic coverage without `fovAdjustedDistance` (with a focused test for equal-size nodes at different depths).

src/scene/gsplat-unified/gsplat-octree.js
```diff
@@ -226,6 +282,99 @@ class GSplatOctree {
...
+     * sub-range's Pareto frontier is not the full frontier filtered down to it - when `rangeMax`
+     * lands inside a run of levels with equal error, a level that the full range discards becomes
+     * the sub-range's cheapest entry.
+     *
+     * @param {number} rangeMin - Finest allowed LOD index.
+     * @param {number} rangeMax - Coarsest allowed LOD index.
+     * @returns {GSplatLodTable} The selection table, with its reference count incremented.
+     */
+    acquireLodTable(rangeMin, rangeMax) {
+        const key = rangeMin * 256 + rangeMax;
```
mvaligursky: [P3] Use a collision-free key for the LOD range

`rangeMin * 256 + rangeMax` is only unique while `rangeMax < 256`, but neither the manifest parser nor `resolveLodRange()` imposes that bound. For example, `[0, 300]` and `[1, 44]` both produce key 300; on the current code, acquiring them in that order returns the `[0, 300]` table for the `[1, 44]` request. That also defeats `resolveLodRange()` because acquiring the aliased table and releasing the previous reference leaves the instance on the wrong range. Please use a collision-free pair representation (for example, a string key or nested maps), and use the same representation in `releaseLodTable`.

mvaligursky (commented):
Automated in-depth re-review by Codex (GPT-5) at exact head `da3f2a0ad1ff6820a0663f6e415bd050ab9c948b`.

The latest commits correctly fix the two findings from the previous review: the full Pareto chain is retained for discrete allocation/streaming, and LOD tables are now reference-counted by live instances instead of using an evicting cap. I re-audited those implementations from scratch, including range changes, instance/world destruction, device loss, underfill, and prefetch; their normal ownership and navigation invariants look sound.

I found three remaining issues: the new best-reachable-run price can purchase a poor prefix even when the suffix that justifies its price cannot fit; orthographic cameras are still evaluated with a perspective distance formula; and the numeric table-cache key aliases ranges above 255. The first two affect selection quality in supported runtime scenarios; the cache issue is a lower-frequency correctness edge.

Non-blocking documentation cleanup: the PR description still explains the superseded running-minimum precomputation, and the public `lodRangeMin` / `lodRangeMax` descriptions still say selection is distance-driven. Per instruction, I did not reassess or repeat the separate deprecation-cycle concern.

Verification:
- Reviewed the complete 31-file base-to-head diff and all six commits, not only the follow-ups.
- Focused LOD table/balancer suite: 38 passing.
- Full suite: 2486 passing, 2 pending.
- Full ESLint and `git diff --check`: pass.
- Reproduced the unaffordable-prefix allocation on the current classes and exhaustively enumerated its feasible assignments: current result is 110 splats / residual error 12; a feasible 80-splat assignment has residual error 10.
- Reproduced the cache alias on the current `GSplatOctree`: requesting `[0, 300]` then `[1, 44]` returns the same `[0, 300]` table.
- Checked the coverage derivation against the engine camera's projection-specific screen-size behavior; orthographic screen size is explicitly distance-independent.
- All current GitHub build, lint, type, unit-test, docs, examples, API, size, and deployment checks are green.

src/scene/gsplat-unified/gsplat-lod-table.js
mvaligursky: Addressed in 5bb9fd54f, though not via the hull in the end: pooling levels out of the chain removed feasible render/stream states (20-25% of Pareto levels on measured captures, every one a strict error improvement), so instead each step is priced by the best run reachable from its start — max over reachable levels of Δerror/Δcost — while the node still climbs one level at a time. Your (20,10)→(90,8)→(100,0) example prices at 0.125 and outbids the 0.1 rival at budget 120; measured 1-17% lower residual than the running-min clamp across four captures. The shape is covered in the allocator tests ('buys a compound upgrade that beats a cheaper rival outright').

src/scene/gsplat-unified/gsplat-lod-table.js
mvaligursky: Addressed in 5bb9fd54f: the hull no longer prunes the chain. The full count-monotone Pareto chain is what the allocator, streaming, underfill and prefetch all walk (only dominated levels drop, ~2-4% on measured captures), and the compound value lives purely in the per-step pricing.

src/scene/gsplat-unified/gsplat-octree.js
mvaligursky: Addressed in 6e5470979 (with 603598deb fixing the key): tables are per range, reference counted by the instances holding them, and the balancer consumes the table resolveLodRange() already resolved (inst.lodTable) instead of resolving again — so two instances on different ranges build each table once and share thereafter.

src/scene/gsplat-unified/gsplat-octree.js
mvaligursky: Addressed in 6e5470979: the FIFO cap is gone in favour of reference counting — acquireLodTable/releaseLodTable, with an instance releasing its previous table on range change and on destroy. A table lives exactly while some instance is on its range; tests cover more live ranges than the old cap held.

src/scene/gsplat-unified/gsplat-lod-table.js
mvaligursky: Verified — at budget 110 this leaves residual 12 vs the optimal 10, and at 100 the hard stop leaves 14 vs 10. We built the affordability-aware drain (atomic affordable runs, demotion to the best fitting run, pass-over when nothing fits) and measured it across four captures: ~0.2% aggregate residual improvement, +13-25% drain time from the per-pop chain walk, and the pass-over semantics measured equal-or-worse temporal churn on 3 of 4 assets — it relaxes the hard early exit we keep deliberately for camera-motion stability. Decision: keep the simple drain and accept the boundary; both cases are pinned in tests with this rationale (603598deb). Happy to revisit if a real capture surfaces the boundary visibly.

src/scene/gsplat-unified/gsplat-octree-instance.js
mvaligursky: Fixed in 603598deb: coverage branches on camera.projection — orthographic uses min(radius / orthoHeight, 1)², depth-independent, mirroring Camera#getScreenSize, with the behind-camera penalty still applied so invisible content cannot win budget; FOV compensation is now perspective-only. Tests cover equal-size nodes at different depths, dolly invariance, window sizing and the behind penalty.

src/scene/gsplat-unified/gsplat-octree.js
mvaligursky: Fixed in 603598deb: string key `${rangeMin},${rangeMax}` in both acquireLodTable and releaseLodTable, with a regression test for the [0,300] / [1,44] alias.

src/scene/gsplat-unified/gsplat-octree-instance.js
```diff
@@ -594,99 +569,51 @@ class GSplatOctreeInstance {
...
+            // well-defined, lowest-possible priority rather than a value the allocator has to
+            // special-case.
+            const radius = nodes[nodeIndex].boundingSphere.w;
+            let coverage;
+            if (ortho) {
+                // No distance attenuation: the footprint is the radius against the ortho window,
+                // clamped to a full-window 1 as the perspective ratio is bounded by 1. The behind
+                // penalty divides squared, matching how a penalized distance scales the far-field
+                // perspective coverage.
+                const projectedRadius = Math.min(radius * invOrthoHeight, 1);
```
mvaligursky, line 602: [P2] Convert the local radius to world units for orthographic coverage

`radius` comes from the octree-local bounding sphere, while `camera.orthoHeight` is a world-space window size. The perspective path is scale-correct because both radius and the inverse-transformed camera distance are local, but this orthographic ratio mixes spaces and never uses the `uniformScale` already computed above. In a direct current-code reproduction, a unit leaf at scale 1 produces coverage `0.12`; changing its placement to uniform scale 2 still produces `0.12`, although the projected radius doubles and coverage should become `0.48` before clamping. This also misranks otherwise identical octrees placed at different scales under the shared budget. Multiply the radius by the placement's world scale for this branch (equivalent to `radius * uniformScale / orthoHeight`) and add a non-identity-scale regression test.

mvaligursky (commented):
Automated follow-up review by Codex (GPT-5) at exact head `603598debc811498e4a21336f8b45638974d899f`, incorporating the replies on all prior reports.

The replies materially change the assessment of the allocator boundary. The 110/100-splat outcomes are now explicitly accepted heuristic behavior, the tests pin both cases, and the author reports an implemented affordability-aware comparison showing only ~0.2% aggregate residual improvement at +13–25% drain time with equal-or-worse temporal churn on most measured assets. Given that evidence and the allocator's deliberate hard-stop stability objective, I consider that report resolved by an informed design decision rather than an outstanding defect.

The collision-free string cache key correctly fixes the reported range alias. The orthographic branch fixes depth attenuation, dolly invariance, ortho-window sizing, and the behind-camera weighting at identity scale, but it introduces one remaining unit-space error: local node radius is compared directly with world-space `orthoHeight`, so scaled placements receive the wrong coverage. I left one inline P2 for that case.

Non-blocking documentation remains stale: the PR description still explains the superseded running-minimum precomputation, and public `lodRangeMin` / `lodRangeMax` text still says selection is distance-driven. Per instruction, I again excluded the separate deprecation-cycle concern.

Verification:
- Read and evaluated every new inline reply, then reviewed the complete five-file follow-up and its surrounding projection/transform paths.
- Focused allocator/table/coverage suites: 46 passing.
- Full suite: 2494 passing, 2 pending.
- Full ESLint and `git diff --check`: pass.
- Cache-alias regression is covered and passes.
- Targeted transform reproduction: changing an orthographic placement from scale 1 to scale 2 leaves coverage at `0.12` instead of increasing it to `0.48` (4× screen area).
- All current GitHub build, lint, type, unit-test, docs, examples, API, size, and deployment checks are green.

src/scene/gsplat-unified/gsplat-octree-instance.js
mvaligursky: Fixed in 7ab4243d4: the placement's uniform scale is folded into the precomputed inverse window height, so the ortho ratio is world-space (radius * uniformScale / orthoHeight). Regression test pins the scale-2 case — coverage quadruples, and it reproduces your 0.12-stuck value on the old code.

mvaligursky (commented):
Automated follow-up review by Codex (GPT-5) at exact head `7ab4243d4c3754b55b118948bb402b61e72a2450`.

No findings in the final P2 fix. The commit correctly converts the octree-local radius into the orthographic camera's world-space window by folding the placement's accumulated uniform world scale into `invOrthoHeight`. This is done once outside the per-node loop, preserves the existing no-allocation hot path, includes parent/inherited scale through `getWorldTransform()`, and leaves the perspective calculation unchanged because its radius and camera distance remain in the same local space.

The new regression test avoids the coverage clamp and verifies the important invariant directly: doubling placement scale quadruples projected-area coverage. It reproduces the former `0.12 -> 0.12` failure and now produces the expected `0.12 -> 0.48` behavior.

Verification:
- Reviewed the complete two-file commit and the surrounding perspective/orthographic transform paths.
- Focused allocator/table/coverage suites: 47 passing.
- Full suite: 2495 passing, 2 pending.
- Full ESLint and `git diff --check`: pass.
- Confirmed accumulated parent scale is returned by the world transform used by the fix.
- All GitHub build, lint, type, unit-test, docs, examples, API, size, and deployment checks are green.

The previously reported P2 is resolved. Per instruction, the separate deprecation-cycle concern remains outside this review.

Ben-Mack:
Hi @mvaligursky , Is it possible to keep using the old distance-based LOD behavior?
I've tested 2.22.0-beta.25 which included this new error-based LOD, and in my scene, compared same scene/budget, it look more blurry than old distance-based version. 
I think because the old distance based can be more guaranteed that if i move close to any chunk, that nearest chunk will have higher detail LOD loaded. But with this new error-based behavior, that is not guaranteed, so for some chunk that has fine details, but not high contrast, even though it is nearest to camera, it's still more blurry than before.
My naive guess is error-driven prefer to load higher detail LOD for chunk with high contrast details even if farther away, which because i guess high contrast details usually have higher error score so it is usually preferred to select finer LOD.
Unfortunately i can't share the scene, but i think keeping both behavior for a while should be safe. 
Do you think it is possible?

mvaligursky:
Thanks for testing @Ben-Mack . What you say makes sense, and I think the best solution would a single slider (0-1 range, default to perhaps 0.5 for balanced), to give more priority between foreground and background. Thoughts?

Ben-Mack:
Would set distance priority to the max would give similar result to current distance-based?  I want the behavior to be mainly based on distance, with error weighting as optional/bonus.

In my scene, error-driven is clear dowgrade in visual, because the old distance-based has a very useful behavior that neighbor chunks that at the same distance to camera tends to use the same LOD level, which minimize artifacts, cleaner visual because neighbors display at the same details, avoid artifacts/difference at chunk seams when mixing different levels. 
Current error driven doesn't have this behavior, neighbor chunks even though same distance to camera, LOD selection for each chunk still mixed from various different LOD level, which look visibly worse, visual look not consistent, can see a high detail chunk beside low detail chunk.

mvaligursky:
Have you had a chance to try it with the latest splat-transform that generates the error metrics? That'd be a super useful feedback if you had, to see if you find this better in that respect.

mvaligursky:
I exposed new settings to allow distance based mode. See here @Ben-Mack 
https://github.com/playcanvas/engine/pull/9267

That is pretty close to the old behavior, and should suite your needs.

Ben-Mack:
@mvaligursky Thanks a lot for such quick update! 
Visually new distance mode works quite well, but i noticed a blocking problem of new v2.22 engine:
On ios chrome/safari (ip 14 pro), the new v2.22 is much more prone to crash on ios (auto reload page) compared to v2.21, are there some config I can tune?  (tested the release, non-beta version)
I guess maybe LOD update may pressure mem too much, or maybe bring fps too low, so safari decided to refresh the page. 

iOS especially prone to crash when camera moving quickly between multi position across the scene, which bring fps down heavily because many LOD need to be updated, then it crash. 
v2.21 also prone to crash in this case, but v2.22 is much worse, even crash right at first loading even though splatBudget is 500k/1M, not even moving the camera.

I think slower lod streaming/update is totally acceptable cost to avoid crashing on ios, ideally making it even more reliable than 2.21, do you have any idea?

mvaligursky:
Hi @Ben-Mack . I made a small change https://github.com/playcanvas/engine/pull/9326 which updates the distance memory use to be in line with 2.21, in my tests it is equivalent now, while before it has perhaps up to 10% higher memory use. I'll be releasing 2.22.1 with this most likely tomorrow.

In order to further lower the memory, you could lower this value https://api.playcanvas.com/engine/classes/GSplatParams.html#cooldownticks in order to unload unreferenced assets sooner. 

You can also set this to a higher number than default 0 on those devices, in order to not load the highest quality splats at all: https://api.playcanvas.com/engine/classes/GSplatComponent.html#lodrangemin

and perhaps even adjust this: https://api.playcanvas.com/engine/classes/GSplatComponent.html#lodfalloff

As always, your feedback is welcome, please do report back your findings / issues.

PR #9235, opened 2026-08-27
DrawCommands optimization

AlexAPPi:
## Skip empty multi-draw / keep GPU-driven indirect draws

Forward and shadow rendering no longer submit a mesh instance when its bound `DrawCommands` have no useful sub-draws. GPU-driven indirect draws (`MeshInstance#setIndirect`) are still issued when the slot count is > 0, because the CPU cannot inspect GPU-written instance counts.

- `DrawCommands#update` now computes `hasDraws`: a sub-draw counts if both index/vertex count and instance count are > 0
- Forward and shadow renderers skip the mesh when `hasDraws` is false (avoids material/shader setup and the draw)
- Indirect vs multi-draw is distinguished by `slotIndex`: `-1` is multi-draw (`0` is a valid indirect slot). `allocate()` resets `slotIndex` to `-1`; `setIndirect` assigns a real slot (`>= 0`)
- WebGPU `draw` uses `slotIndex` only for the indirect path; multi-draw still starts at base slot 0

## Public API

`DrawCommands#hasDraws` — `true` if any CPU-side sub-draw has work. For indirect draws (`slotIndex >= 0`), `true` when `count > 0`.

```javascript
const cmd = meshInstance.setMultiDraw(null, n);
cmd.add(0, indexCount, instanceCount, firstIndex);
cmd.update(write); // hasDraws is false when write === 0 or all instance/index counts are 0
```

Indirect usage is unchanged:

```javascript
meshInstance.setIndirect(null, slot, count); // hasDraws === count > 0
```

## Performance

Empty multi-draw lists (`update(0)` or all-zero instance/index counts) skip the whole draw path: no shader bind, no uniform setup, no multi-draw / indirect API call.

## Checklist

### DIFF
--- a/src/platform/graphics/draw-commands.js
+++ b/src/platform/graphics/draw-commands.js
@@ -148,7 +148,7 @@ class DrawCommands {
      */
     update(count) {
         this._count = count;
-        this.primitiveCount = this.impl.update?.(count) ?? 0;
+        this.primitiveCount = this.impl.update(count);
 
         // For indirect draws, the CPU cannot inspect the instance counts,
         // so we treat the draw count as the hasDraws flag.
@@ -157,7 +157,7 @@ class DrawCommands {
             return;
         }
 
-        this.hasDraws = this.impl.hasDraws ?? (count > 0);
+        this.hasDraws = this.impl.hasDraws;
     }
 }
 
--- a/src/platform/graphics/draw-commands.js
+++ b/src/platform/graphics/draw-commands.js
@@ -66,11 +66,21 @@ class DrawCommands {
     }
 
     /**
-     * Slot index of the first indirect draw call. Ignored for multi-draw commands.
+     * First indirect slot, or `-1` for multi-draw.
      *
      * @ignore
      */
-    slotIndex = 0;
+    slotIndex = -1;
+
+    /**
+     * True if this container is bound to GPU-driven indirect slots.
+     *
+     * @type {boolean}
+     * @ignore
+     */
+    get isIndirect() {
+        return this.slotIndex >= 0;
+    }
 
     /**
      * Total number of primitives across all sub-draws (pre-calculated).
@@ -78,6 +88,14 @@ class DrawCommands {
      * @ignore
      */
     primitiveCount = 0;
+
+    /**
+     * True if any sub-draw has index/vertex count > 0 and instance count > 0.
+     * For indirect draws (`isIndirect`), true when {@link count} > 0.
+     *
+     * @type {boolean}
+     */
+    hasDraws = false;
 
     /**
      * @param {import('./graphics-device.js').GraphicsDevice} device - The graphics device.
@@ -104,6 +122,7 @@ class DrawCommands {
      */
     allocate(maxCount) {
         this._maxCount = maxCount;
+        this.slotIndex = -1;
         this.impl.allocate?.(maxCount);
     }
 
@@ -123,13 +142,22 @@ class DrawCommands {
     }
 
     /**
-     * Finalize and set draw count after all commands have been added.
+     * Sets the draw count and {@link hasDraws}.
      *
      * @param {number} count - Number of draws to execute.
      */
     update(count) {
         this._count = count;
         this.primitiveCount = this.impl.update?.(count) ?? 0;
+
+        // For indirect draws, the CPU cannot inspect the instance counts,
+        // so we treat the draw count as the hasDraws flag.
+        if (this.isIndirect) {
+            this.hasDraws = count > 0;
+            return;
+        }
+
+        this.hasDraws = this.impl.hasDraws ?? (count > 0);
     }
 }
 
--- a/src/platform/graphics/null/null-draw-commands.js
+++ b/src/platform/graphics/null/null-draw-commands.js
@@ -7,9 +7,21 @@ class NullDrawCommands {
     /** @type {boolean} */
     hasDraws = false;
 
+    /**
+     * Write a single draw entry.
+     * @param {number} i - Draw index.
+     * @param {number} indexOrVertexCount - Count of indices/vertices.
+     * @param {number} instanceCount - Instance count.
+     * @param {number} firstIndexOrVertex - First index/vertex.
+     */
     add(i, indexOrVertexCount, instanceCount, firstIndexOrVertex) {
     }
 
+    /**
+     * Sets {@link hasDraws} and calculate total primitives for stats (profiler builds only).
+     * @param {number} count - Number of active draws.
+     * @returns {number} Total primitive count.
+     */
     update(count) {
         this.hasDraws = count > 0;
         return 0;
--- a/src/platform/graphics/null/null-draw-commands.js
+++ b/src/platform/graphics/null/null-draw-commands.js
@@ -23,7 +23,7 @@ class NullDrawCommands {
      * @returns {number} Total primitive count.
      */
     update(count) {
-        this.hasDraws = count > 0;
+        this.hasDraws = false;
         return 0;
     }
 }
--- a/src/platform/graphics/null/null-draw-commands.js
+++ b/src/platform/graphics/null/null-draw-commands.js
@@ -4,7 +4,15 @@
  * @ignore
  */
 class NullDrawCommands {
+    /** @type {boolean} */
+    hasDraws = false;
+
     add(i, indexOrVertexCount, instanceCount, firstIndexOrVertex) {
+    }
+
+    update(count) {
+        this.hasDraws = count > 0;
+        return 0;
     }
 }
 
--- a/src/platform/graphics/webgpu/webgpu-graphics-device.js
+++ b/src/platform/graphics/webgpu/webgpu-graphics-device.js
@@ -956,15 +956,16 @@ class WebgpuGraphicsDevice extends GraphicsDevice {
             }
 
             // draw
-            if (drawCommands) { // indirect draw path
+            if (drawCommands) { // indirect / multi-draw path
 
+                const baseSlot = drawCommands.isIndirect ? drawCommands.slotIndex : 0;
                 const storage = drawCommands.impl?.storage ?? this.indirectDrawBuffer;
                 const indirectBuffer = storage.impl.buffer;
                 const drawsCount = drawCommands.count;
 
                 // TODO: when multiDrawIndirect is supported, we can use it here instead of a loop
                 for (let d = 0; d < drawsCount; d++) {
-                    const indirectOffset = (drawCommands.slotIndex + d) * _indirectEntryByteSize;
+                    const indirectOffset = (baseSlot + d) * _indirectEntryByteSize;
                     if (indexBuffer) {
                         passEncoder.drawIndexedIndirect(indirectBuffer, indirectOffset);
                     } else {
--- a/src/scene/renderer/forward-renderer.js
+++ b/src/scene/renderer/forward-renderer.js
@@ -572,8 +572,12 @@ class ForwardRenderer extends Renderer {
             // Skip hardware-instanced rendering with 0 instances. When draw commands (indirect /
             // multi-draw) are bound, they are the source of truth for the number of draws and
             // per-draw instance counts, so instancingData.count must not gate the draw.
+            const drawCommands = drawCall.getDrawCommands(camera);
+            if (drawCommands && !drawCommands.hasDraws) {
+                continue;
+            }
             const instancingData = drawCall.instancingData;
-            if (instancingData && instancingData.count <= 0 && !drawCall.getDrawCommands(camera)) {
+            if (instancingData && instancingData.count <= 0) {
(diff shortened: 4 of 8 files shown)


src/scene/renderer/forward-renderer.js
```diff
@@ -572,8 +572,12 @@ class ForwardRenderer extends Renderer {
             // Skip hardware-instanced rendering with 0 instances. When draw commands (indirect /
             // multi-draw) are bound, they are the source of truth for the number of draws and
             // per-draw instance counts, so instancingData.count must not gate the draw.
+            const drawCommands = drawCall.getDrawCommands(camera);
+            if (drawCommands && !drawCommands.hasDraws) {
+                continue;
+            }
             const instancingData = drawCall.instancingData;
-            if (instancingData && instancingData.count <= 0 && !drawCall.getDrawCommands(camera)) {
+            if (instancingData && instancingData.count <= 0) {
```
mvaligursky: [P1] Preserve command-driven draws when instancing count is zero

This gate now runs even when `drawCommands` exists. `setInstancing(true)` initializes `instancingData.count` to 0, while `setIndirect(...)` does not change it, so a valid indirect command with `hasDraws === true` is removed from the prepared list. The same regression is present in `ShadowRenderer` at its corresponding gate. This also contradicts the `MeshInstance` contract that command counts take precedence over `instancingCount`. Please apply the instancing check only when no draw commands are bound (for example, `instancingData && instancingData.count <= 0 && !drawCommands`).

src/platform/graphics/webgpu/webgpu-graphics-device.js
```diff
@@ -956,15 +956,16 @@ class WebgpuGraphicsDevice extends GraphicsDevice {
             }
 
             // draw
-            if (drawCommands) { // indirect draw path
+            if (drawCommands) { // indirect / multi-draw path
 
+                const baseSlot = drawCommands.isIndirect ? drawCommands.slotIndex : 0;
                 const storage = drawCommands.impl?.storage ?? this.indirectDrawBuffer;
```
mvaligursky, line 962: [P1] Select the command buffer using the new mode discriminator

The buffer choice still ignores `isIndirect`. If a mesh/camera calls `setMultiDraw(...)` and then `setIndirect(...)` in the same frame, `MeshInstance` reuses the same `DrawCommands`; `slotIndex` becomes indirect, but `impl.storage` remains the CPU multi-draw buffer. This line therefore reads the requested indirect slot from stale CPU storage (and can address beyond that buffer) instead of `this.indirectDrawBuffer`. Choose the storage by mode—device indirect buffer for `isIndirect`, implementation storage for multi-draw—and consider avoiding `impl.update()` or recreating the implementation when transitioning to indirect.

src/platform/graphics/draw-commands.js
```diff
@@ -123,13 +142,22 @@ class DrawCommands {
...
         this.primitiveCount = this.impl.update?.(count) ?? 0;
+
+        // For indirect draws, the CPU cannot inspect the instance counts,
+        // so we treat the draw count as the hasDraws flag.
+        if (this.isIndirect) {
+            this.hasDraws = count > 0;
+            return;
+        }
+
+        this.hasDraws = this.impl.hasDraws ?? (count > 0);
```
mvaligursky: [P2] Keep hasDraws semantics consistent on NullGraphicsDevice

`NullDrawCommands` has neither `update` nor `hasDraws`, so this fallback makes every `update(count > 0)` report true even when all submitted vertex/index or instance counts are zero. Because `hasDraws` is a new documented public property, its value should not depend on the selected backend. Please make the null implementation track enough command data to implement the same predicate, or keep that state in the backend-independent wrapper.

mvaligursky (commented):
Automated PR review by Codex (GPT-5) at `f6931ec1024679225b239fc064027c28143ce5ef`.

I found two rendering correctness issues and one public-API consistency issue:

- The new renderer gate drops valid command-driven rendering whenever legacy `instancingCount` is zero. This affects both forward and shadow submission and directly contradicts the documented precedence rules.
- WebGPU still selects the command buffer independently of the new `isIndirect` discriminator, so switching the same mesh/camera from multi-draw to indirect in one frame reads the old CPU buffer instead of the device indirect buffer.
- `DrawCommands#hasDraws` does not have its documented meaning on `NullGraphicsDevice`.

Validation: `npm test` passes (2450 passing, 2 pending), `npm run lint` passes, and `git diff --check` passes. The PR adds no targeted coverage for the new skip behavior or multi-draw/indirect transitions. Type generation could not be run in the isolated worktree because the build script requires a worktree-local `node_modules/.bin/tsc`; this is an environment limitation, not a reported PR failure. The two visible Vercel checks fail at Git authorization rather than in the code.

Please add regression tests covering: command draws with `instancingCount === 0` in forward and shadow paths, all-empty versus partially non-empty command lists, indirect slot 0, and reusing a command key across multi-draw/indirect modes.

src/scene/renderer/forward-renderer.js
AlexAPPi: Fixed

mvaligursky (commented):
Automated re-review by Codex (GPT-5) at `08a8bbfa1bc4fe05a331397c33932d021b105394`.

The original regression where valid command-driven draws were rejected solely because `instancingCount === 0` is fixed, and I treated the outdated/dismissed thread as closed. However, the replacement condition now has the opposite truth-table error: empty command lists are not skipped unless legacy instancing also exists and has a non-positive count. This means the advertised optimization does not run for the normal non-instanced `setMultiDraw` path, and the PR pays the new O(command-count) CPU scan without receiving its intended submission/setup saving in that primary case. The same predicate is used in forward and shadow rendering.

The earlier WebGPU buffer-selection finding is unchanged at this head: a same-key multi-draw to indirect transition still prefers the retained CPU `impl.storage` over `indirectDrawBuffer`. I have not duplicated that existing inline thread.

Validation: the targeted renderer reproduction returns one prepared draw for both `hasDraws: false` cases (no `instancingData`, and positive legacy instancing), while correctly retaining a valid command with zero legacy instancing. `npm test` passes (2450 passing, 2 pending), `npm run lint` passes, and `git diff --check` passes. The Vercel failures remain Git-authorization failures rather than code failures.

src/scene/renderer/forward-renderer.js
```diff
@@ -573,7 +573,10 @@ class ForwardRenderer extends Renderer {
             // multi-draw) are bound, they are the source of truth for the number of draws and
             // per-draw instance counts, so instancingData.count must not gate the draw.
             const instancingData = drawCall.instancingData;
-            if (instancingData && instancingData.count <= 0 && !drawCall.getDrawCommands(camera)) {
+            const drawCommands = drawCall.getDrawCommands(camera);
+
+            if (instancingData && instancingData.count <= 0 &&
+                (!drawCommands || !drawCommands.hasDraws)) {
```
mvaligursky: [P1] Branch on draw-command state before legacy instancing

This conjunction consults `hasDraws` only when `instancingData` exists and its count is non-positive. Consequently, `setMultiDraw(...); update(0)` on a normal non-instanced mesh (`instancingData === null`) is still prepared and submitted, as is an all-empty command list accompanied by a positive legacy instancing count. That defeats the PR's core optimization and leaves its unconditional O(command-count) scan as pure overhead in the primary example. The intended precedence is `if (drawCommands ? !drawCommands.hasDraws : instancingData?.count <= 0) continue;`. Please apply the same truth table in `ShadowRenderer` and cover the command-present/absent × has-draws × instancing-count cases with tests.

src/platform/graphics/draw-commands.js
AlexAPPi: Fixed.

src/platform/graphics/null/null-draw-commands.js
```diff
@@ -4,8 +4,16 @@
...
  */
 class NullDrawCommands {
+    /** @type {boolean} */
+    hasDraws = false;
+
     add(i, indexOrVertexCount, instanceCount, firstIndexOrVertex) {
     }
+
+    update(count) {
+        this.hasDraws = count > 0;
```
mvaligursky: [P2] Derive hasDraws from active command contents

This is the same `count > 0` fallback that the commit is intended to replace. Because `add()` still discards `indexOrVertexCount` and `instanceCount`, `add(0, 0, 0, 0); update(1)` reports `hasDraws === true`, contrary to the public contract; it also cannot distinguish valid entries inside versus outside the first `count` slots. Please retain a per-slot work flag (or the two counts) during `add()`, reset/resize it in `allocate()`, and have `update(count)` scan only the active prefix. A null-backend unit test for empty, mixed, truncated, and zero-count lists would catch this directly.

mvaligursky (commented):
Automated re-review by Codex (GPT-5) at `1399f7cfd0d917979b8eeadd9e1cbc9e9fe411be`.

The latest `Fix P2` commit does not yet implement the documented null-backend semantics. `NullDrawCommands` still ignores the submitted command data and equates a positive active-command count with useful work, so the original empty-command case remains observable. I added one focused inline comment on the new implementation.

The other two active findings are also unchanged at this head, and I have not duplicated their existing inline threads:

- The forward/shadow predicate still skips empty commands only when legacy instancing also has a non-positive count, so normal non-instanced empty multi-draw lists are submitted.
- WebGPU still selects retained CPU `impl.storage` when a command key transitions from multi-draw to indirect, instead of selecting `indirectDrawBuffer` by mode.

The previously dismissed/outdated concern remains treated as closed.

Validation: targeted null-backend checks produce `hasDraws === true` both for an explicitly empty active command and when the only valid command lies outside `update(count)`'s active range. `npm test` passes (2450 passing, 2 pending), `npm run lint` passes, and `git diff --check` passes. No targeted regression tests were added. The visible Vercel failures remain Git-authorization failures, not code failures.

src/platform/graphics/null/null-draw-commands.js
AlexAPPi: Fixed.

mvaligursky (commented):
A few things to address:

**1. Release-build cost of the `hasDraws` scan** — in both `WebglDrawCommands.update` and `WebgpuDrawCommands.update`, the per-sub-draw loop used to be inside `// #if _PROFILER` and was compiled out of release builds; it now always runs. That's a new O(count) cost per update paid by everyone, including users with very large batches and no empty draws. The cost should be paid by whoever submits empty draws, not by every user — in non-profiler builds, exit the scan as soon as `hasDraws` becomes true (keep the full loop for the profiler stats), so the typical case exits on the first sub-draw.

**2. `webgpu-graphics-device.js` storage selection** — now that `isIndirect` exists, the storage pick should use it: `drawCommands.isIndirect ? this.indirectDrawBuffer : (drawCommands.impl?.storage ?? this.indirectDrawBuffer)`. Without it, a command that was ever multi-draw and later switched to indirect on the same camera key reads from the stale impl storage while the offsets index the global indirect buffer. One-line fix, worth doing in this PR.

**3. Forward-renderer prepare loop** — `drawCall.getDrawCommands(camera)` is now called unconditionally for every draw call every frame, where previously it only ran in the rare zero-instance case. This is one of the hottest loops in the engine and the goal is to make draw call submission faster, not slower — gate it on a cheap `drawCall.drawCommands` null check so the common path pays a single property read.

**4. Docs / tests** — `hasDraws` JSDoc links `isIndirect`, which is `@ignore`d, so the generated API docs get a dangling reference. The `update()` doc also lost the "call after all commands have been added" guidance, which is exactly the contract `hasDraws` depends on. Please also add unit tests covering `update(0)`, all-zero sub-draws, and the indirect vs multi-draw paths.

src/platform/graphics/webgpu/webgpu-graphics-device.js
AlexAPPi: For indirect storage not initialized. But here we will deliberately leave a loophole so that we can specify another buffer of our own.

AlexAPPi:
@mvaligursky  Thanks for the review.

1 / 3. Agreed on the cost concern. For shadows this should be effectively free, because we already did similar work there. For the main render path we still need to think it through — I don’t want the common case to pay extra for something it never uses. I’ll look at gating the scan so typical non-empty batches (and the forward-renderer prepare loop) stay cheap, and the cost sits with whoever actually submits empty draws.

On isIndirect / storage selection: I’d rather keep a way to use a custom buffer. Forcing isIndirect ? this.indirectDrawBuffer : … closes that off, and custom storage is occasionally useful. I’m open to a safer pick that prefers the global indirect buffer when appropriate, as long as custom buffers remain possible.

Docs/tests — will fix the dangling isIndirect link, restore the “call after all commands have been added” note on update(), and add tests for update(0), all-zero sub-draws, and the indirect vs multi-draw paths.

AlexAPPi:
After some discussion, we decided to leave visible as the source of truth for whether to display the mesh or not.

PR #9243, opened 2026-08-28
Register module scripts with the owning application

mvaligursky:
Ensure ESM scripts loaded by ScriptHandler register with the handler’s application instead of whichever application is globally current.

**Changes:**
- Pass the handler application to the existing registerScript call
- Add multi-application coverage for ESM script registration

### DIFF
--- a/src/framework/handlers/script.js
+++ b/src/framework/handlers/script.js
@@ -153,7 +153,7 @@ class ScriptHandler extends ResourceHandler {
                     }
 
                     // Register the script name
-                    registerScript(scriptClass, scriptName);
+                    registerScript(scriptClass, scriptName, this._app);
 
                     // Store any schema associated with the script
                     if (scriptSchema) this._app.scripts.addSchema(scriptName, scriptSchema[scriptName]);
--- a/test/framework/handlers/script-handler.test.mjs
+++ b/test/framework/handlers/script-handler.test.mjs
@@ -1,0 +1,49 @@
+import { expect } from 'chai';
+
+import { AppBase } from '../../../src/framework/app-base.js';
+import { createApp } from '../../app.mjs';
+import { jsdomSetup, jsdomTeardown } from '../../jsdom.mjs';
+
+describe('ScriptHandler', function () {
+
+    let app;
+
+    beforeEach(function () {
+        jsdomSetup();
+        app = createApp();
+    });
+
+    afterEach(function () {
+        app?.destroy();
+        app = null;
+        jsdomTeardown();
+    });
+
+    it('registers module scripts with the handler application', async function () {
+        const otherApp = createApp();
+        const scriptUrl = new URL('../../../src/framework/script/script.js', import.meta.url);
+        const source = `
+            import { Script } from '${scriptUrl}';
+            export class ModuleScript extends Script {
+                static scriptName = 'moduleScript';
+            }
+            // module.mjs`;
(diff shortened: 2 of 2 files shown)


src/framework/handlers/script.js
```diff
@@ -153,7 +153,7 @@ class ScriptHandler extends ResourceHandler {
                     }
 
                     // Register the script name
-                    registerScript(scriptClass, scriptName);
+                    registerScript(scriptClass, scriptName, this._app);
```
mvaligursky, line 156: This fixes the ESM path, but the classic-script path in the same `load()` method has the identical bug and is untouched. A classic script calls `createScript(name)` from inside its own file with no `app` argument, so `createScript` → `registerScript(type, name, undefined)` → `AppBase.getApplication().scripts` — exactly the global-current-app resolution this PR is removing here. Since classic scripts are the Editor default, that's arguably the more commonly hit half.

What makes this look like an unfinished thread rather than a deliberate scope choice: `load()` sets `script.app = this._app` a few lines up (line 56), and the field's own declaration comments it as *"set during script load to be used for initializing script"* — but I grepped the whole of `src/`, and `script.app` is only ever **written** (here, and nulled in `app-base.js:1967`). Nothing in the engine ever reads it. So the intended plumbing exists but is dead.

A natural completion, with no API change and no new plumbing: have `registerScript` prefer `script.app` over the global when no explicit `app` is passed —
```js
const registry = (app ?? script.app ?? AppBase.getApplication()).scripts;
```
That would make the classic path resolve to the owning application for the duration of its load, and give `script.app` an actual consumer. (Worth checking the import direction for a cycle between `script-create.js` and `script.js` first.) Equally fine to leave for a follow-up — but the PR title claims module scripts specifically, so it'd be worth saying the classic path is knowingly out of scope.

mvaligursky (commented):

---

Fix is correct (`registerScript`'s third arg selects `app.scripts` over `AppBase.getApplication().scripts`) and the test is a genuine regression test — the `expect(AppBase.getApplication()).to.equal(otherApp)` guard makes it fail against the old code rather than passing vacuously. One gap and one test-robustness note, both inline.

The gap in short: the ESM half of this bug is now fixed, but the **classic-script half in the same `load()` method is not** — and the `script.app = this._app` line right above it looks like the vestige of the mechanism that was meant to fix it.

test/framework/handlers/script-handler.test.mjs
```diff
@@ -0,0 +1,49 @@
...
+
+    it('registers module scripts with the handler application', async function () {
+        const otherApp = createApp();
+        const scriptUrl = new URL('../../../src/framework/script/script.js', import.meta.url);
+        const source = `
+            import { Script } from '${scriptUrl}';
+            export class ModuleScript extends Script {
+                static scriptName = 'moduleScript';
+            }
+            // module.mjs`;
```
mvaligursky, line 31: Worth a one-line comment here: this trailing `// module.mjs` token is load-bearing, not a label. The handler chooses the ESM branch with `basePath.endsWith('.mjs')`, and the only reason this data URL satisfies that is that `encodeURIComponent` leaves `.` and alphanumerics unescaped, so the encoded source ends in `.mjs`.

If someone reformats the template literal or drops the comment, the test silently routes to `_loadScript` instead — appending a real `<script>` element to the jsdom document — and fails for a reason with nothing to do with app registration. A short note (`// the trailing token makes the data URL end in '.mjs' so the handler takes the ESM branch`) would save that debugging session.

PR #9245, opened 2026-08-28
Add heightMapBase to unify the parallax zero level across modes

mvaligursky:
Adds `StandardMaterial#heightMapBase` - the height map value that sits at the level of the original geometry - and unifies the zero-level convention of the two parallax modes before 2.22 ships.

Previously the two modes read the same map differently: `PARALLAX_OFFSET` pivoted the relief around mid-grey (the behavior shipped for years), while the new `PARALLAX_OCCLUSION` treated white as the geometry level and could only carve downwards. Toggling between modes shifted the whole surface by half the relief depth, and content authored against the POM reading would have baked that inconsistency in.

**Changes:**
- New `StandardMaterial#heightMapBase` (0-1, default 0.5): relief above the base stands out of the surface, relief below it sinks in. The default matches the long-shipped OFFSET pivot, so existing height maps look unchanged; `1` reproduces the depth-only POM reading this replaces (for crevice-style maps authored white = flat), `0` treats the map as pure elevation.
- Both parallax modes now honor the same base: OFFSET remaps the height relative to it, and the POM march enters the height volume at the ray's top-of-volume crossing, extrapolated back along the view ray, so relief above the base moves the lookup towards the camera. `heightMapBase: 1` produces the exact march this replaces.
- The distance fade now pulls the hit onto the geometry plane, so relief flattens in place rather than sliding to the top of the volume; the self-shadow climb keeps fading to zero length so shadows leave together with the relief.
- The property only feeds a uniform, so it never invalidates the shader (the generic float rule would recompile whenever a slider lands on exactly 0 or 1).

**Examples:**
- `materials/parallax-mapping` and `materials/parallax-terrain` both gain a Base slider.

Verified on WebGL2 and WebGPU in both examples: base 0 / 0.5 / 1 in both modes, with self shadowing active on the terrain.

### DIFF
--- a/examples/src/examples/materials/parallax-mapping.controls.jsx
+++ b/examples/src/examples/materials/parallax-mapping.controls.jsx
@@ -53,6 +53,15 @@ export function Controls({ observer }) {
                         precision={2}
                     />
                 </LabelGroup>
+                <LabelGroup text='Base'>
+                    <SliderInput
+                        binding={new BindingTwoWay()}
+                        link={{ observer, path: 'data.base' }}
+                        min={0.0}
+                        max={1}
+                        precision={2}
+                    />
+                </LabelGroup>
             </Panel>
             <Panel headerText='Lighting'>
                 <LabelGroup text='Spot'>
--- a/examples/src/examples/materials/parallax-mapping.example.mjs
+++ b/examples/src/examples/materials/parallax-mapping.example.mjs
@@ -264,22 +264,28 @@ data.set('data', {
     // reads as spikes when the floor is viewed from low down. This is the depth the material suits.
     height: 0.4,
 
+    // the height map value that sits at the level of the geometry - the engine default pivots the
+    // relief around mid-grey, and 1 treats the map as pure depth carved below the surface
+    base: 0.5,
+
     spot: true,
     omni: true,
     env: envIntensity
 });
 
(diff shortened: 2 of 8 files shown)


mvaligursky (commented):

No blocking findings.

I reviewed the new `StandardMaterial.heightMapBase` API, material reset/copy/asset parsing and uniform-only update behavior, the OFFSET and OCCLUSION equations, distance fade and self-shadow coordinates, GLSL/WGSL parity, examples, compatibility, and performance impact. The default `0.5` preserves OFFSET behavior, base `1` preserves the prior POM march, and a constant map at the selected base produces zero displacement as intended. The implementation adds no per-fragment samples or new shader variants.

Validation performed:
- `npm run lint`
- focused StandardMaterial/parallax tests: 40 passing
- full unit suite: 2,496 passing, 2 pending
- debug ESM build
- live parallax-mapping and parallax-terrain examples on WebGL 2 and WebGPU, including bases 0, 0.5, and 1 and the self-shadow path; no new shader or console errors
- all current PR checks are green

Non-blocking test suggestion: the existing parallax tests do not directly pin the new property default/copy/asset parsing/uniform value or the base-dependent shader equations. Adding focused regression coverage would make this mathematically sensitive behavior safer to evolve.

PR #9249, opened 2026-08-28
Correct sound duration and playback timing

mvaligursky:
Keep sound duration reporting and playback position consistent with configured start times and the underlying audio resource.

**Changes:**
- Clamp `SoundInstance.duration` and `SoundSlot.duration` to the playable time after the normalized start time.
- Track `SoundInstance.currentTime` relative to `startTime` and wrap assigned values immediately.
- Resume paused or seeked sounds at the correct buffer offset and limit non-looping playback to the remaining configured duration.
- Add regression coverage for shorter, equal, longer, non-zero-start, wrapped-start, seek, and pause/resume cases.

**API Changes:**
- `SoundInstance#duration` and `SoundSlot#duration` now return the playable duration after `startTime` when a duration is configured.
- `SoundInstance#currentTime` now consistently reports a position relative to `startTime`, including immediately after assignment and across pause/resume.

```javascript
slot.startTime = 2;
slot.duration = 6; // Audio asset duration is 4 seconds

slot.duration; // Before: 4
slot.duration; // After: 2

instance.currentTime = 5; // Effective duration is 3 seconds
instance.currentTime; // Before playback: 5
instance.currentTime; // After: 2
```

### DIFF
--- a/src/framework/components/sound/slot.js
+++ b/src/framework/components/sound/slot.js
@@ -588,8 +588,8 @@ class SoundSlot extends EventHandler {
     }
 
     /**
-     * Gets the duration of the sound that the slot will play starting from startTime. The returned
-     * value is clamped to the duration of the audio asset.
+     * Gets the duration of the sound that the slot will play starting from {@link startTime}. The
+     * returned value is clamped to the time available after the normalized start time.
      *
      * @type {number}
      */
@@ -602,7 +602,8 @@ class SoundSlot extends EventHandler {
 
         // != intentional
         if (this._duration != null) {
-            return Math.min(this._duration, assetDuration);
+            const startTime = (this._startTime % assetDuration) || 0;
+            return Math.min(this._duration, assetDuration - startTime);
         }
         return assetDuration;
     }
--- a/src/framework/components/sound/slot.js
+++ b/src/framework/components/sound/slot.js
@@ -588,7 +588,8 @@ class SoundSlot extends EventHandler {
     }
 
     /**
-     * Gets the duration of the sound that the slot will play starting from startTime.
+     * Gets the duration of the sound that the slot will play starting from startTime. The returned
+     * value is clamped to the duration of the audio asset.
      *
      * @type {number}
      */
@@ -601,7 +602,7 @@ class SoundSlot extends EventHandler {
 
         // != intentional
         if (this._duration != null) {
-            return this._duration % (assetDuration || 1);
+            return Math.min(this._duration, assetDuration);
         }
         return assetDuration;
     }
--- a/src/platform/sound/instance.js
+++ b/src/platform/sound/instance.js
@@ -216,7 +216,11 @@ class SoundInstance extends EventHandler {
          */
         this._currentTime = 0;
 
-        /** @private */
+        /**
+         * The playback position relative to startTime when _startedAt was recorded.
+         *
+         * @private
+         */
         this._currentOffset = 0;
 
         /**
@@ -302,13 +306,17 @@ class SoundInstance extends EventHandler {
     }
 
     /**
-     * Sets the current time of the sound that is playing. If the value provided is bigger than the
-     * duration of the instance it will wrap from the beginning.
+     * Sets the current time of the sound that is playing, relative to {@link startTime}. If the
+     * value provided is bigger than the duration of the instance it will wrap from the beginning.
      *
      * @type {number}
      */
     set currentTime(value) {
+        value = Number(value) || 0;
         if (value < 0) return;
+
+        const duration = this.duration;
+        const currentTime = duration ? capTime(value, duration) : value;
 
         if (this._state === STATE_PLAYING) {
             const suspend = this._suspendInstanceEvents;
@@ -318,19 +326,19 @@ class SoundInstance extends EventHandler {
             this.stop();
 
             // set _startOffset and play
-            this._startOffset = value;
+            this._startOffset = currentTime;
             this.play();
             this._suspendInstanceEvents = suspend;
         } else {
             // set _startOffset which will be used when the instance will start playing
-            this._startOffset = value;
+            this._startOffset = currentTime;
             // set _currentTime
-            this._currentTime = value;
+            this._currentTime = currentTime;
         }
     }
 
     /**
-     * Gets the current time of the sound that is playing.
+     * Gets the current time of the sound that is playing, relative to {@link startTime}.
      *
      * @type {number}
      */
@@ -375,8 +383,8 @@ class SoundInstance extends EventHandler {
     }
 
     /**
-     * Gets the duration of the sound that the instance will play starting from startTime. The
-     * returned value is clamped to the duration of the sound resource.
+     * Gets the duration of the sound that the instance will play starting from {@link startTime}.
+     * The returned value is clamped to the time available after the normalized start time.
      *
      * @type {number}
      */
@@ -385,7 +393,9 @@ class SoundInstance extends EventHandler {
             return 0;
         }
         if (this._duration) {
-            return Math.min(this._duration, this._sound.duration);
+            const soundDuration = this._sound.duration;
+            const startTime = capTime(this._startTime, soundDuration);
+            return Math.min(this._duration, soundDuration - startTime);
         }
         return this._sound.duration;
     }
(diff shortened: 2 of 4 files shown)


src/platform/sound/instance.js
```diff
@@ -384,7 +385,7 @@ class SoundInstance extends EventHandler {
             return 0;
         }
         if (this._duration) {
-            return capTime(this._duration, this._sound.duration);
+            return Math.min(this._duration, this._sound.duration);
```
mvaligursky: The clamp bounds against the whole resource, but the value being described is the duration *from `startTime`* — as this getter's own doc says ("the duration of the sound that the instance will play starting from startTime"). Once `startTime > 0` the longest playable span is `sound.duration - startTime`, so the getter still over-reports.

I confirmed this against the real class (4s sound, mock `AudioContext` capturing `source.start` args):

```
_duration=6 startTime=0 | getter=4 | start(offset=0, dur=6) -> audible=4 | AGREE
_duration=6 startTime=2 | getter=4 | start(offset=2, dur=6) -> audible=2 | DISAGREE
_duration=9 startTime=1 | getter=4 | start(offset=1, dur=9) -> audible=3 | DISAGREE
_duration=3 startTime=2 | getter=3 | start(offset=2, dur=3) -> audible=2 | DISAGREE
```

(`offset` is the `capTime(this._startTime + offset, this._sound.duration)` result from `_playAudioImmediate`; the audible span is bounded by the buffer end regardless of the length handed to `source.start`.)

Note the last row: `_duration=3 < sound.duration=4`, so `Math.min` is a no-op there — meaning this is pre-existing and not introduced here. But it's the same invariant the PR is fixing, and the fix is the same shape:

```js
return Math.min(this._duration, this._sound.duration - this._startTime);
```

Worth confirming against `startTime`'s clamping at line 528 (`Math.max(0, ...)`) that the subtraction can't go negative for a loaded sound — if `startTime` can exceed `sound.duration`, wrap it in `Math.max(0, ...)`.

src/framework/components/sound/slot.js
```diff
@@ -601,7 +602,7 @@ class SoundSlot extends EventHandler {
 
         // != intentional
         if (this._duration != null) {
-            return this._duration % (assetDuration || 1);
+            return Math.min(this._duration, assetDuration);
```
mvaligursky: Same `startTime` gap as the instance getter — `SoundSlot` has its own `_startTime` (line 161, settable at line 753) and this getter carries the identical "starting from startTime" wording, so whatever bound is chosen for `SoundInstance#duration` should be mirrored here to keep the two in step.

mvaligursky (commented):

---

Ran the new tests and lint locally — both clean, and the change is correctly scoped to the getters (`_createInstance` passes the raw `_duration`, so no playback config shifts).

One gap, detailed inline: **the clamp ignores `startTime`**, so the "getters agree with playback" invariant still doesn't hold whenever `startTime > 0`. Not a regression (it predates the modulo→clamp change), but it's the same invariant this PR is establishing and it's a one-token extension of the same expression.

Secondary: **neither new test suite exercises `startTime`.** The shorter/equal/longer matrix is otherwise thorough, but with `startTime` left at its default of 0 the interaction is entirely uncovered — which is why the above slips through green. A case like `{ startTime: 2, duration: 6 }` on the 4s fixture would pin it (expected 2, currently 4).

src/framework/components/sound/slot.js
mvaligursky: Addressed in b36f6a7ee. SoundSlot now mirrors SoundInstance by clamping against the time remaining after normalized startTime, with regression cases for non-zero and wrapped start times.

src/platform/sound/instance.js
mvaligursky: Addressed in b36f6a7ee. The duration now subtracts the normalized start time, including start times at or beyond the resource length. The same update also makes currentTime relative to startTime and corrects seek and pause/resume offsets.

src/platform/sound/instance.js
```diff
@@ -302,14 +306,18 @@ class SoundInstance extends EventHandler {
...
+     * value provided is bigger than the duration of the instance it will wrap from the beginning.
      *
      * @type {number}
      */
     set currentTime(value) {
+        value = Number(value) || 0;
         if (value < 0) return;
 
+        const duration = this.duration;
+        const currentTime = duration ? capTime(value, duration) : value;
```
mvaligursky, line 319: This is the eager normalization worth calling out in the PR's API Changes section. Before this commit the setter stored the raw value and `get currentTime` returned it verbatim via `_startOffset`, so `currentTime = 5` on a 3s duration read back as `5` and only wrapped at `play()`. It now reads back `2`.

Matching the documented "it will wrap from the beginning" immediately is the better behavior, and the new test pins it — this is purely about the description not mentioning a changed property round-trip.

One ordering consequence, if you think it's worth caring about: because the wrap is applied at assignment time using the duration *as of that moment*, the caller's original intent is discarded. Setting `currentTime = 5` and then widening `duration` afterwards leaves the offset at `2`, whereas the old lazy wrap would have re-derived `5 % newDuration`. Both setters restart playback so there's no torn state — it's just that assignment order now matters where it didn't. Probably fine to accept; a `@remarks` note would cover it.

src/platform/sound/instance.js
```diff
@@ -384,7 +393,9 @@ class SoundInstance extends EventHandler {
             return 0;
         }
         if (this._duration) {
-            return capTime(this._duration, this._sound.duration);
+            const soundDuration = this._sound.duration;
+            const startTime = capTime(this._startTime, soundDuration);
```
mvaligursky, line 397: "the normalized start time" in the new doc wording is carrying more meaning than a reader will unpack. `capTime` is a modulo wrap, so two non-obvious rules follow:

- `startTime = 6` on a 4s sound normalizes to `2` — playback silently begins 2s in rather than erroring or clamping to the end.
- `startTime === sound.duration` wraps to `0`, i.e. asking to start exactly at the end restarts from the beginning (your `{ startTime: 4, duration: 6 }` test encodes this).

Both match what `_playAudioImmediate` already did with `capTime(this._startTime + currentOffset, this._sound.duration)`, so aligning the getter to it is right. Just worth stating the wrap outright in the JSDoc ("start times beyond the resource length wrap modulo its duration") so the behavior is discoverable from the API docs rather than only from this expression.

src/framework/components/sound/slot.js
```diff
@@ -601,7 +602,8 @@ class SoundSlot extends EventHandler {
 
         // != intentional
         if (this._duration != null) {
-            return this._duration % (assetDuration || 1);
+            const startTime = (this._startTime % assetDuration) || 0;
```
mvaligursky, line 605: This inlines the same rule `SoundInstance#duration` expresses as `capTime(this._startTime, soundDuration)` — presumably because `capTime` is module-private to `instance.js`. Two encodings of one rule, and the `|| 0` is load-bearing in both: it converts the `NaN` from `x % 0` when the asset isn't loaded, and collapses the `startTime === duration` case to `0`.

The two getters are now required to agree (that's the invariant this PR establishes), so the duplication is the kind that drifts quietly — a future tweak to one wouldn't fail any test that compares them, since none does. Exporting `capTime` from `instance.js` (or lifting it somewhere shared) and calling it from both would make them structurally impossible to diverge.

mvaligursky (commented):
🤖 **Automated re-review** (commit `b36f6a7`) — posted on my behalf by Claude Code (Opus 4.8). Not a human review. The points below are suggestions to weigh as possible improvements, not changes that necessarily need to be addressed.

---

The `startTime` gap is resolved. I re-ran the getter-vs-playback comparison at this head across 14 configurations and it now agrees in every one — including the loop path, which the commit doesn't touch directly (`loopEnd = min(loopStart + _duration, buffer.duration)` turns out to be algebraically identical to `loopStart + this.duration`, so the loop window matches the getter for free). 18 new tests pass, 116 across the adjacent sound suites, lint clean.

The `_currentOffset` re-basing is a nice catch beyond what I raised: `set pitch` does `this._currentOffset = this.currentTime`, and before this commit those two were in different coordinate spaces (buffer-absolute vs `startTime`-relative), so that assignment was silently wrong whenever `startTime > 0`. It's now consistent.

Three points, all minor and mostly about surfacing the new semantics rather than the code.

**The PR description is now stale.** The **API Changes** section still lists only the original resource-length clamp, but this commit adds two further observable changes:

- The `duration` getters now subtract the normalized `startTime`, so the headline example is incomplete — the same `slot.duration = 45` on that 38.22s asset returns something different again once `startTime > 0`.
- `set currentTime` now normalizes eagerly (see inline). Previously `instance.currentTime = 5` with `duration = 3` read back as `5` and only wrapped when playback started; it now reads back `2` immediately. The new test asserts this deliberately, so it's clearly intended — it just deserves a line in the API notes, since it changes a property round-trip.

PR #9267, opened 2026-08-31
Add a LOD mode and a per-splat detail falloff for streamed GSplats

mvaligursky:
Follow-up to #9233, adding user control over how the splat budget is spent.

## `scene.gsplat.lodMode` — global selection mode

- `GSPLAT_LODMODE_ERROR` (`'error'`, default): unchanged behaviour — the budget goes where it removes the most approximation error per splat, using the manifest's error tables when present and derived errors otherwise.
- `GSPLAT_LODMODE_DISTANCE` (`'distance'`): ignores error metadata and guarantees a coarser-with-distance progression — detail steps down in clean concentric bands around the camera, with band edges adapting to the budget. Useful when a capture's quality makes its error tables unreliable (e.g. sky-heavy nodes that pin themselves to LOD 0). Internally this synthesizes per-level errors whose per-step ratio is a per-level constant, so node content cancels out of the ranking and the same single allocator path serves both modes.
- In distance mode, coverage is pure inverse-square world distance under both projections: node size is factored out, so equal-distance nodes always pick equal levels regardless of the leaf sizes the octree cut produced, and orthographic cameras gain a distance ordering (their footprint carries no depth term to rank by).
- The mode is part of the cached LOD table identity, and is also settable via scene settings (`render.gsplatLodMode`).

## `GSplatComponent#lodFalloff` — per-splat near/far tilt

- Range [0, 8], default 1. Higher values concentrate detail near the camera at the cost of the far field; 0 spreads the budget evenly with no view preference. In distance mode this controls the band spacing.
- The tilt pivots around a mid-field reference, so it redistributes the splat's detail rather than changing its overall share of the budget — useful for balancing foreground splats against large environments.
- Implemented as a multiply in the balancer's key space: the default path is bit-identical to before, and a non-default falloff costs ~0.03 ms at 39k nodes (vs ~0.9 ms for the naive `Math.pow` per node), so it is safe on mobile.

## Example

The `lod-streaming` example gains a LOD Mode dropdown and a LOD Falloff slider.

Tests: distance-mode table pricing and cache identity, falloff behaviour against an exact greedy oracle (including bit-identical default), size-independence and ortho/scale behaviour of distance-mode coverage, and component default/init/clamp/clone.

### DIFF
--- a/src/framework/components/gsplat/component.js
+++ b/src/framework/components/gsplat/component.js
@@ -381,9 +381,9 @@ class GSplatComponent extends Component {
      * Sets how quickly this splat's level of detail drops with distance from the camera. The
      * default of 1 gives a balanced falloff. Higher values concentrate detail near the camera at
      * the cost of the far field, while values towards 0 spread it evenly across the scene
-     * regardless of the view. This redistributes the detail this splat receives from the global
-     * {@link GSplatParams#splatBudget} without changing its overall share of it. Clamped to
-     * [0, 8].
+     * regardless of the view. This primarily redistributes the detail this splat receives from
+     * the global {@link GSplatParams#splatBudget} between its near and far field, though it can
+     * also shift how the budget divides between splats. Clamped to [0, 8].
      *
      * @type {number}
      */
--- a/src/framework/components/gsplat/component.js
+++ b/src/framework/components/gsplat/component.js
@@ -98,6 +98,13 @@ class GSplatComponent extends Component {
     _materialTmp = null;
 
     /**
+     * How fast quality falls off away from the camera, exponent on projected coverage.
+     *
+     * @private
+     */
+    _lodFalloff = 1;
+
+    /**
      * Minimum allowed LOD index (inclusive).
      *
      * @private
@@ -368,6 +375,32 @@ class GSplatComponent extends Component {
      */
     get castShadows() {
         return this._castShadows;
+    }
+
+    /**
+     * Sets how quickly this splat's level of detail drops with distance from the camera. The
+     * default of 1 gives a balanced falloff. Higher values concentrate detail near the camera at
+     * the cost of the far field, while values towards 0 spread it evenly across the scene
+     * regardless of the view. This redistributes the detail this splat receives from the global
+     * {@link GSplatParams#splatBudget} without changing its overall share of it. Clamped to
+     * [0, 8].
+     *
+     * @type {number}
+     */
+    set lodFalloff(value) {
+        this._lodFalloff = Math.min(Math.max(value, 0), 8);
+        if (this._placement) {
+            this._placement.lodFalloff = this._lodFalloff;
+        }
+    }
+
+    /**
+     * Gets how quickly this splat's level of detail drops with distance from the camera.
+     *
+     * @type {number}
+     */
+    get lodFalloff() {
+        return this._lodFalloff;
     }
 
     /**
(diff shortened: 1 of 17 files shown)


mvaligursky (commented):

I found three actionable issues, reported inline: one public TypeScript API regression and two mismatches between the documented LOD guarantees and the allocator behavior.

Validation performed:
- `npm run lint`
- focused GSplat component/allocator/table/coverage tests: 63 passing
- full unit suite: 2,528 passing, 2 pending
- direct allocator reproductions for equal-distance ties and cross-placement falloff allocation
- all current PR checks are green

src/scene/constants.js
```diff
@@ -1287,6 +1287,25 @@ export const GSPLAT_RENDERER_COMPUTE = 3;
  * @type {number}
  * @category Graphics
  */
+/**
```
mvaligursky, line 1284: [P1] Preserve `GSPLAT_DEBUG_NONE` in the public typings. The new constant docs were inserted between that existing symbol and its JSDoc, leaving `GSPLAT_DEBUG_NONE` undocumented. The API report already confirms the resulting breaking change as `-const GSPLAT_DEBUG_NONE: number`, so TypeScript consumers importing this existing public constant will fail after the release. Move these declarations before the `GSPLAT_DEBUG_NONE` comment, or move its JSDoc back immediately above the export.

src/scene/gsplat-unified/gsplat-lod-table.js
```diff
@@ -25,6 +33,12 @@
  * {@link GSplatLodTable#upgradeRatio} holds the fixed part and selection costs one multiply per
  * upgrade.
  *
+ * In {@link GSPLAT_LODMODE_DISTANCE} the node's error metadata is deliberately ignored and a
+ * synthetic table is used instead, shaped so that error-per-splat is a per-level constant across
+ * all nodes. Content then cancels out of the ranking entirely and selection degenerates to
+ * distance bands: every node steps coarser at fixed distance ratios, which is the guarantee that
```
mvaligursky, line 39: [P2] The allocator does not actually guarantee complete distance bands at the budget boundary. Equal priorities are queued node-by-node, and `GSplatBudgetBalancer` returns as soon as the next upgrade does not fit. With two identical nodes at equal coverage, each going from 5 to 10 splats, distance mode and a budget of 15 produce `[LOD 0, LOD 1]`: one node is upgraded and the other is not. With only 512 value buckets, nearby unequal distances can tie this way too, making the boundary depend on node order instead of forming the promised concentric band. Either batch/apply tied level upgrades consistently (or select a band threshold first), or weaken the public guarantee and add a boundary test.

src/framework/components/gsplat/component.js
```diff
@@ -370,6 +377,32 @@ class GSplatComponent extends Component {
         return this._castShadows;
     }
 
+    /**
+     * Sets how quickly this splat's level of detail drops with distance from the camera. The
+     * default of 1 gives a balanced falloff. Higher values concentrate detail near the camera at
+     * the cost of the far field, while values towards 0 spread it evenly across the scene
+     * regardless of the view. This redistributes the detail this splat receives from the global
+     * {@link GSplatParams#splatBudget} without changing its overall share of it. Clamped to
```
mvaligursky: [P2] `lodFalloff` does not preserve a placement's overall budget share. The global `1e-4` pivot only leaves nodes at that one coverage unchanged; it does not normalize the transformed priorities per placement. In a direct reproduction with one node in each of two placements at coverage `0.01`, placement A ratio 1 and B ratio 2, neutral falloff gives the only affordable upgrade to B, while setting only A to falloff 2 gives it to A. Since A has no near/far nodes to redistribute between, its share has changed outright. Please either normalize per placement if this invariant is required, or document the control as a pivoted tilt that can also change allocation between splats.

